PR #2947, opened 2014-12-10
bug 1082798

m8ttyB:


groovecoder:
Ping @darkwing - what do you think about this? Merge?

darkwing:
If it blocks merging of PRs, I'd say no, since we noticed some inconsistencies in tests passing locally vs. on Sauce Labs.

.travis.yml
```diff
@@ -32,12 +39,18 @@ install:
...
+    - sudo apt-get install -y nodejs
 
 before_script:
     - mysql -e 'create database kuma;'
 
 script:
     - coverage run --include=`pwd`/kuma/*,`pwd`/apps/* manage.py test --noinput -v2 access actioncounters authkeys contentflagging devmo kpi kuma landing sumo
+    - cd tests/ui
+    - npm install intern
+    - node_modules/.bin/intern-runner config=intern-sauce-labs d=developer.allizom.org
```
stof: ``if [ "$TRAVIS_SECURE_VARS" = 1 ]; then node_modules/.bin/intern-runner config=intern-sauce-labs d=developer.allizom.org; fi;`` will allow running these tests only when secure vars are available on Travis

.travis.yml
```diff
@@ -32,12 +39,18 @@ install:
...
+    - curl -sL https://deb.nodesource.com/setup | sudo bash -
+    - sudo apt-get install -y nodejs
 
 before_script:
     - mysql -e 'create database kuma;'
 
 script:
     - coverage run --include=`pwd`/kuma/*,`pwd`/apps/* manage.py test --noinput -v2 access actioncounters authkeys contentflagging devmo kpi kuma landing sumo
+    - cd tests/ui
+    - npm install intern
```
stof: the npm install should not be part of the testing scripts but of the setup (probably ``install`` in that case). A failure during one of the scripts does not stop Travis from running other ones (this is a main difference with other steps of the build)

.travis.yml
```diff
@@ -32,12 +39,18 @@ install:
     - pip install --no-index --find-links=travis_wheels -r requirements/compiled.txt
     - pip install --find-links=base_wheels -r requirements/prod.txt -r requirements/dev.txt
     - pip install coveralls
+    # JS test requirements
+    - curl -sL https://deb.nodesource.com/setup | sudo bash -
+    - sudo apt-get install -y nodejs
```
stof: nodejs is already available on Travis (for the python language, you don't have the choice of the nodejs version, but you still have it available)

.travis.yml
m8ttyB: thank you .. done

.travis.yml
m8ttyB: Good point -- done

.travis.yml
m8ttyB: That's good to know -- I'll remove the installation of nodejs from the `.travis.yaml` config and see if presents a problem

darkwing:
Continued pause on this until Q2; going to close for now but we'll come back to it @m8ttyB !

PR #2970, opened 2014-12-23
Bug 1104260: Add switch for researching on prod

openjck:


openjck:
https://github.com/mozilla/kuma/pull/2983

PR #2978, opened 2014-12-30
fix bug 1115738 - move IP to author column in dashboard

groovecoder:


groovecoder:
Fixed tests.

openjck:
Looks awesome! One last thing. We could reduce the amount of markup by giving the table an ID and matching the ID in the CSS. Something like this...

```html
<table id="revisions" ... >
   ...
</table>
```

```css
table#revisions td {
   ...
}
```

groovecoder:
When I made that change, the `class="title"` column [became uppercase from the base mixin](https://github.com/groovecoder/kuma/blob/move-ip-to-author-column-1115738/media/redesign/stylus/includes/mixins.styl#L392). Not sure the best way to exclude this particular `class="title"` element without having to re-scan a ton of content. :/

groovecoder:
Here's what I tried: https://github.com/groovecoder/kuma/commit/acfdef2bf6ce8be8ee1b80873d8196710e290ee2

openjck:
Ah, okay. In that case we can make it even simpler, and save even more bytes.

We can start with where the pull request is now (866927e53437fe388c9a916920727e29a492c445), remove `dashboard-col` from the markup, and then change line 53 of dashboards.styl from `.dashboard-col` to `td`. That should be all we need to reduce the size of the markup/CSS and maintain the look that we have on master.

PR #2982, opened 2015-01-02
Bug 1116419: Highlight both 'bug' and 'Bug' in Wiki comments and history

adusca:


groovecoder:
This will be a nice bug to fix. But we might try it like this, so it will be completely case-insensitive. (e.g., "BUG 1116419" or even "bUg 1116419"

https://gist.github.com/02bfab4e1eb146ee1cd2

adusca:
Done!

groovecoder:
Since we don't have one, would you mind adding a `test_bugize_text` to [the `HelpTests`](https://github.com/adusca/kuma/blob/master/kuma/wiki/tests/test_helpers.py#L12-L16) for this?

groovecoder:
Works and has tests!! :+1:

PR #2984, opened 2015-01-06
fix bug 1115767 - SMTP headers for url & username

groovecoder:
https://bugzilla.mozilla.org/show_bug.cgi?id=1115767

Spot-check instructions are the same as https://github.com/mozilla/kuma/pull/2966, but check for the custom SMTP headers too.

I'm not super-happy with this code so I could use a critical review on it.

### DIFF
--- a/kuma/wiki/views.py
+++ b/kuma/wiki/views.py
@@ -2190,8 +2190,8 @@ def _make_first_edit_email(new_rev, request):
     article_url = "%s%s" % (Site.objects.get_current().domain,
                             doc.get_absolute_url())
     email = EmailMessage(subject, message, settings.DEFAULT_FROM_EMAIL,
-                         headers = {'X-Article-Url': article_url,
-                                    'X-Editor-Username': user.username})
+                         headers={'X-Kuma-Document-Url': article_url,
+                                  'X-Kuma-Editor-Username': user.username})
     return email
 
 
--- a/kuma/wiki/views.py
+++ b/kuma/wiki/views.py
@@ -17,6 +17,7 @@ from tower import ugettext_lazy as _lazy, ugettext as _
 
 from django.conf import settings
 from django.contrib import messages
+from django.contrib.sites.models import Site
 from django.core.exceptions import PermissionDenied
 from django.core.cache import cache
 from django.core.mail import EmailMessage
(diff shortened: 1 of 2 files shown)


groovecoder:
Either @ubernostrum or @jezdez should tell me a better way to do this ...

kuma/wiki/views.py
```diff
@@ -2186,7 +2187,11 @@ def _make_first_edit_email(new_rev, request):
     template = 'wiki/email/edited.ltxt'
     context = context_dict(new_rev)
     message = render_to_string(request, template, context)
-    email = EmailMessage(subject, message, settings.DEFAULT_FROM_EMAIL)
+    article_url = "%s%s" % (Site.objects.get_current().domain,
+                            doc.get_absolute_url())
+    email = EmailMessage(subject, message, settings.DEFAULT_FROM_EMAIL,
+                         headers = {'X-Article-Url': article_url,
+                                    'X-Editor-Username': user.username})
```
jezdez: While RFC 822 is not mentioning it, it's good practice to let the extension headers contain the product they are from, so let's use `X-Kuma-Document-URL` and `X-Kuma-Document-Editor`.

kuma/wiki/views.py
```diff
@@ -2186,7 +2187,11 @@ def _make_first_edit_email(new_rev, request):
     template = 'wiki/email/edited.ltxt'
     context = context_dict(new_rev)
     message = render_to_string(request, template, context)
-    email = EmailMessage(subject, message, settings.DEFAULT_FROM_EMAIL)
+    article_url = "%s%s" % (Site.objects.get_current().domain,
+                            doc.get_absolute_url())
+    email = EmailMessage(subject, message, settings.DEFAULT_FROM_EMAIL,
+                         headers = {'X-Article-Url': article_url,
```
jezdez: No spaces around `=`

jezdez:
Other than the comments above, r+.

groovecoder:
Updated.

jezdez:
@groovecoder My point was to use Kuma since that's the software, and MDN the platform implemented with it. The head should have Kuma in there as that's how headers behave usually.

Also please mind my other proposal to use "document" instead of "article" in the header name since that's more in line what the wiki app in Kuma contains.

r-

groovecoder:
Updated again.

PR #2988, opened 2015-01-07
fix bug 1115738 - (again) with dashboard-col class

groovecoder:
[The blanket `td` style caused issues with the diff table inside the dashboard](https://bugzilla.mozilla.org/show_bug.cgi?id=1115738#c8). :(

So, I'm back to `dashboard-col` classes, and also `dashboard-col-title` and `dashboard-col-author` for identifying the special columns. And now I understand how Twitter Bootstrap class naming conventions evolved. :frowning:


openjck:
Nice catch. I forgot to consider the inner table. Thanks for fixing this.

I'm going to supersede with a slightly different approach. Child selectors can also be used to match only the outer table. They solve the same problem, but in a more concise way and without needing more classes.

PR #2989, opened 2015-01-07
Fix bug 1115738 again: Select only outer cells

openjck:
Supersedes #2988.


groovecoder:
Looks good; better code. Merging as soon as Travis passes. (Had to kick it)

PR #2991, opened 2015-01-08
[fix bug 1118656] Flag spammers in UI

hoosteeno:


hoosteeno:
Needs rebase, not ready.

hoosteeno:
Rebased.

kuma/users/models.py
```diff
@@ -166,6 +166,13 @@ def beta_tester(self):
                 self.user.groups.values_list('name', flat=True))
 
     @property
+    def is_banned(self):
+        return self.user.bans.filter(is_active=True).exists()
+
+    def active_ban(self):
+        if self.is_banned:
+            return self.user.bans.filter(is_active=True)[:1][0]
+
```
groovecoder: Nice refactor. We should add a couple corresponding tests to `kuma/users/tests/test_models.py` ... I can help do that if needed.

apps/devmo/helpers.py
```diff
@@ -206,8 +206,12 @@ def datetimeformat(context, value, format='shortdatetime', output='html'):
     Returns date/time formatted using babel's locale settings. Uses the
     timezone from settings.py
     """
+    if isinstance(value, datetime.date):
+        # Turn a date into a datetime
+        value = datetime.datetime.combine(value, datetime.datetime.min.time())
+
     if not isinstance(value, datetime.datetime):
-        # Expecting date value
+        # Expecting datetime value
```
groovecoder: Need to update [a failing test](https://travis-ci.org/mozilla/kuma/builds/46384489#L2367) related to this change.

kuma/dashboards/templates/dashboards/includes/revision_dashboard_body.html
```diff
@@ -45,6 +45,12 @@
                     {% if show_ips %}
                     <br/><a class="dashboard-ip-link" href="{{ url('admin:wiki_revisionip_changelist') }}?revision={{revision.id}}" target="_blank">{{ _('View IP') }}</a>
                     {% endif %}
+                    {% if revision.creator.get_profile().is_banned and request.user.is_superuser %}
```
groovecoder: To reduce code duplication, add:
```
{% set active_ban = revision.creator.get_profile().active_ban() %}
```
here, and then use it below:
```
banned {{ datetimeformat(active_ban.date, format='date') }}
by {{ active_ban.by }}
```

kuma/users/templates/users/profile.html
```diff
@@ -94,6 +96,13 @@ <h1 class="page-title">
           {% endif %}
       </div>
 
+      {% if profile.is_banned and request.user.is_superuser %}
+      <div class="banned">
+          User banned {{ datetimeformat(profile.active_ban().date, format='date') }} 
+          by {{ profile.active_ban().by }}.
+      </div> 
+      {% endif %}
```
groovecoder: Instead of just hiding the ban button and showing the ban info somewhere else, maybe we should just replace the "Ban" button with this "Already banned" info?

groovecoder:
Looking good. Couple issues to fix. And @openjck or @stephaniehobson should check out the stylus changes here. Probably will want to move the banned style up so we don't duplicate it in both `dashboard.styl` and `profile.styl`

stephaniehobson:
I think the thing to do with .banned is to create a file in components with the class in it and to @require it into both dashboard.style and profile.styl. This would be in keeping with the re-org changes I've been making :)

stephaniehobson:
Also, since we do not have any other instances of the class .banned, we do not need to nest the .banned declaration inside #profile-head and .dashboard-table (unless you are running into specificity conflicts).

kuma/users/templates/users/profile.html
hoosteeno: I'm not convinced. I don't think they belong in the same place just because they both relate to bans. I think textual information about the profile is lined up on the left under the username, and actions to take on the profile are in buttons on the right. 

This approach has the added benefit of not requiring UI experimentation to try and cram textual data in a space normally occupied by a button. 

I nominate @stephaniehobson to break the tie.

apps/devmo/helpers.py
hoosteeno: The test is right, the code is wrong.

kuma/users/models.py
hoosteeno: Added some tests, let me know if they're OK!

hoosteeno:
I think I addressed all earlier comments in the latest commit.

groovecoder:
@hoosteeno - can you re-word the 2nd commit message to include the bug number? Or squash the 2 commits together?

kuma/users/tests/test_models.py
```diff
@@ -110,6 +110,8 @@ def test_ban_user(self):
         ban.save()
         testuser_banned = User.objects.get(username='testuser')
         ok_(not testuser_banned.is_active)
+        ok_(testuser_banned.get_profile().is_banned == True)
+        ok_(testuser_banned.get_profile().active_ban().by == admin)
```
groovecoder: cheers and :beers:

groovecoder:
@stephaniehobson - what's your call on the placement of the "is banned" info? Should we put it on the left side with other textual information, or put it where the "Ban" button is otherwise displayed?

hoosteeno:
Squashed!

stephaniehobson:
Leaving my comment on the banned info here instead of on the outdated diff:

I think that the line of text is best with the textual information. If we want a sign that the user has been banned in place of the banned button we should be working within the button framework. We could either disable the "Ban User" button and change the text to "User Banned" or we could make it into a "Unban User" button. To make the change easier to spot at a glance we could make the ban button red and make it grey in either of my post-banning suggestions.

There's no reason we can't do both a button and the line of text :)

groovecoder:
+1 to a disabled "Ban" button ... maybe with hover info about the active ban?

hoosteeno:
I moved the ban status on the profile into an inactive button, which required refactoring the ban_link helper. This was tricky; please review closely.

media/redesign/stylus/components/userflags.styl
```diff
@@ -0,0 +1,8 @@
+/*
+    Some flags that adorn users, generally for display to site managers
+*/
+
+.banned {
+    color: #f00;
+    set-smaller-font-size();
+}
```
stephaniehobson: I'm trying to move us towards having our stylus files based on individual components, could we name this file banned.styl after the component? Other user flags will get their own files. (Yes it feels weird to have a stylus file that has only one class declaration in it).

hoosteeno: Done.

stephaniehobson:
Superseded by #3004.

PR #2992, opened 2015-01-09
Fix bug 747805: Added the scrollback widget to MDN.

amalantony:


amalantony:
Going through: https://bugzilla.mozilla.org/show_bug.cgi?id=747805, there seems to be a few more changes to be made before the pull request can be merged:

1. User's name on MDN should reflect as the user's name on Scrollback.
2. There should be some kind of a notice to the user that the widget is not a substitute for a tech support tool and should be just a platform for community conversations. 
3. Some sort of way to track users' engagement with the widget.
4. Perhaps, a CTA on the widget which prompts the user to log into MDN.

Perhaps, this pull request should be labelled incomplete, by someone with appropriate permissions.

kuma/wiki/templates/wiki/document.html
```diff
@@ -73,6 +73,25 @@
...
+        el.async = 1;
+        el.src = (location.protocol === "https:" ? "https:" : "http:") + "//scrollback.io/client.min.js";
+        document.getElementsByTagName("script")[0].parentNode.appendChild(el);
+        scrollbackAdded = true;
+      } else {
+        var iframe = document.getElementsByClassName("scrollback-stream")[0];
+        iframe.src = iframe.src.replace(/\/[^/]*\?/, '/' + encodeURIComponent(roomName) + '?');
+      }
+  	}
+  </script>
```
groovecoder: Can this move down to the `{% block js %}` block? That's where we try to do any custom JS on pages.

groovecoder:
Thanks @amalantony! First code scan looks pretty good - just that single comment. I'll dig into this more and try to spot-check it ASAP, but it may be next week.

kuma/wiki/templates/wiki/document.html
```diff
@@ -399,6 +416,15 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
...
   </menu>
+
+  <!-- embed scrollback widget based on the tags on the page -->
+  {% set tags = document.tags.all() %}
+  {% for tag in tags %}
+    {% if settings.SCROLLBACK_ROOM_MAP[tag.name] and waffle.flag('scrollback_widget') %}
+      <button class = "scrollback-button" onclick='addScrollback("{{settings.SCROLLBACK_ROOM_MAP[tag.name]}}")'> Chat about {{tag.name}} </button>
+    {% endif %}
+  {% endfor %}
+
```
jezdez: I don't understand this block, why do we have to go through the list of document tags if the setting for this only contains one item?

The call to `settings.SCROLLBACK_ROOM_MAP[tag.name]` will also raise a KeyError as soon as one of the tags don't match.

kuma/wiki/templates/wiki/document.html
```diff
@@ -399,6 +416,15 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
     <menuitem data-action="$edit" label="{{_('Edit page')}}"></menuitem>
     <menuitem data-action="$history" label="{{_('View page history')}}"></menuitem>
   </menu>
+
+  <!-- embed scrollback widget based on the tags on the page -->
+  {% set tags = document.tags.all() %}
+  {% for tag in tags %}
+    {% if settings.SCROLLBACK_ROOM_MAP[tag.name] and waffle.flag('scrollback_widget') %}
+      <button class = "scrollback-button" onclick='addScrollback("{{settings.SCROLLBACK_ROOM_MAP[tag.name]}}")'> Chat about {{tag.name}} </button>
```
darkwing: Text needs to be localized

kuma/wiki/templates/wiki/document.html
darkwing: The indentation badly need to be fixed.

darkwing: This should also be moved to the bottom of the template, waffled, and put into an external JS file with `async` set.

kuma/wiki/templates/wiki/document.html
amalantony: @jezdez 
> I don't understand this block, why do we have to go through the list of document tags if the setting for this only contains one item?

The settings can have more tag to room mappings in the future. 

>  The call to settings.SCROLLBACK_ROOM_MAP[tag.name] will also raise a KeyError as soon as one of the tags don't match

If I'm not mistaken, this approach works fine with Jinja2 templates. Is there a better way to check if a key exists in a dict in Jinja2?

kuma/wiki/templates/wiki/document.html
amalantony: @darkwing Where do you usually place external JS files in the project?

groovecoder: @amalantony we put JS files in [`media/redesign/js`](https://github.com/mozilla/kuma/tree/master/media/redesign/js)

Then, we add a reference to the file under the appropriate section of [the `MINIFY_BUNDLES['js']` element in `settings.py`](https://github.com/mozilla/kuma/blob/master/settings.py#L710-L784).

This template extends the `wiki/base.html` template, [which includes the '`wiki`' script bundle](https://github.com/mozilla/kuma/blob/master/kuma/wiki/templates/wiki/base.html#L13-L15), so ...

* [ ] Add a `scrollback.js` file to `media/redesign/js`
* [ ] Add `redesign/js/scrollback.js` to the `wiki` section of `MINIFY_BUNDLES['js']`

Does that make sense?

amalantony: @groovecoder Yes, that makes sense. Also, how do I get the username of the user who is logged into Kuma within the template? I need this to implement single sign on, where the user gets the same Scrollback username as his username in Kuma.

groovecoder:
To get the username for the currently logged in user:
```
{{ if request.user.username }}
```
Where are you trying to include it?

amalantony:
@groovecoder The username will be added to the window.scrollback object as a new property (`nick`): https://github.com/amalantony/kuma/blob/scrollback-integration/kuma/wiki/templates/wiki/document.html#L83

jezdez:
Hi there, I've added a few questions in the Bugzilla ticket and would like to ask  @amalantony to pause work on this for now.

kuma/wiki/templates/wiki/document.html
```diff
@@ -399,10 +397,24 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
...
+
 {% endblock %}
 
 {% block js %}
-  <script type="text/javascript">
-      if(window.jQuery) jQuery('.from-search-toc').mozSearchResults('{{ search_url|safe }}');
-  </script>
-{% endblock %}
+	{% if waffle.flag('scrollback_widget') %}
+		<script type="text/javascript" src="../../../../media/redesign/js/scrollback.js" async="true"></script>
```
darkwing: Don't think this path will work...

groovecoder:
Assigning back to @jezdez - he's going thru the legal & privacy details that @hoosteeno dug up to make sure this works. But we still may not get to this for a while yet ...

groovecoder:
We have [our final legal guidance](https://bugzilla.mozilla.org/show_bug.cgi?id=1120531#c8) to add a notice to the widget:
"Welcome! This public chat is provided by Scrollback and is subject to it's Privacy Policy (http://web.scrollback.io/privacy)."

Then we'll need to rebase this before it can merge.

groovecoder:
@amalantony - any time to work on this since we got our final legal guidance cleared?

amalantony:
@groovecoder Since I do not work at Scrollback anymore, I'm not sure if I'll be the right person to look into this. I'll notify someone at Scrollback regarding the status of this issue and they could take it up from here.

aravindet:
@groovecoder we’ll update the PR soon, thanks!

aravindet:
This is done, can we remove the “not ready” label @amalantony ?

groovecoder:
Thanks @aravindet. This PR includes an update to the vendor/src/django submodule for some reason? Can you remove that so the only changes here are the small, waffled front-end changes?

@darkwing - how do those changes look to you?

harishkumarv:
@groovecoder, I have removed the django submodule update.

openjck:
Looks like this is ready. Removing the *not ready* tag.

darkwing:
Going to make a few small updates to this PR for standards reasons

darkwing:
Superceding this with another...

PR #2994, opened 2015-01-12
Fix bug 1119348 - Moved rendering of first edit email to Celery task

jezdez:
This fixes a bad isolation of the send_first_edit_email task that was introduced in bug 1115767.

This prevents passing in the pre-made email object which may contain stateful information.

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -1,20 +1,21 @@
 import logging
 
-from celery.task import task, group
-
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.core.mail import send_mail
-from django.db import transaction
-from django.dispatch import receiver
-from django.db import connection
+from django.contrib.sites.models import Site
 from django.core.cache import get_cache
+from django.core.mail import EmailMessage, send_mail
+from django.db import connection, transaction
+from django.dispatch import receiver
+from django.template.loader import render_to_string
 
+from celery.task import task
 from constance import config
 
 from devmo.utils import MemcacheLock
+from .events import context_dict
 from .exceptions import StaleDocumentsRenderingInProgress, PageMoveError
-from .models import Document, RevisionIP
+from .models import Document, RevisionIP, Revision
 from .signals import render_done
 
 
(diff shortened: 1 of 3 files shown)


kuma/wiki/tasks.py
```diff
@@ -202,6 +203,18 @@ def delete_old_revision_ips(immediate=False, days=30):
...
 
 @task
-def send_first_edit_email(email):
-    email.to = [config.EMAIL_LIST_FOR_FIRST_EDITS,]
+def send_first_edit_email(revision_pk):
+    """ Make an 'edited' notification email for first-time editors """
+    revision = Revision.objects.get(pk=revision_pk)
+    user, doc = revision.creator, revision.document
+    subject = ("[MDN] %(user)s made their first edit, to: %(doc)s" %
+               {'user': user.username, 'doc': doc.title})
```
robhudson: Do you need a `u""` string if username or title have unicode characters?

kuma/wiki/tasks.py
```diff
@@ -202,6 +203,18 @@ def delete_old_revision_ips(immediate=False, days=30):
...
+    message = render_to_string('wiki/email/edited.ltxt',
+                               context_dict(revision))
+    current_site = Site.objects.get_current()
+    doc_url = "https://%s%s" % (current_site.domain, doc.get_absolute_url())
```
robhudson: Surprised there's no helper for this. It seems like you'd do this a lot? In zamboni we have `absolutify` which is used in both Python and template code.

robhudson:
r+wc

kuma/wiki/tasks.py
groovecoder: Surprisingly, these emails are the first place where we've had to generate absolute url's for docs.

jezdez: Yeah, probably a good idea. Filed ticket https://bugzilla.mozilla.org/show_bug.cgi?id=1121051

kuma/wiki/tasks.py
jezdez: Yeah, good catch

PR #2996, opened 2015-01-13
fix bug 1118318 - return 410 for deleted docs

groovecoder:
Spot-check: Delete a document, then go back to its url. Make sure net console shows 410.

### DIFF
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -34,7 +34,7 @@ from kuma.wiki.constants import DOCUMENT_LAST_MODIFIED_CACHE_KEY_TMPL
 from kuma.wiki.content import get_seo_description
 from kuma.wiki.events import EditDocumentEvent
 from kuma.wiki.models import (Document, Revision, RevisionIP, DocumentZone,
-                              DocumentTag)
+                              DocumentTag, DocumentDeletionLog)
 from kuma.wiki.tests import (doc_rev, document, new_document_data, revision,
                              normalize_html, create_template_test_users,
                              make_translation)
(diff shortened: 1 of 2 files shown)


kuma/wiki/tests/test_views.py
```diff
@@ -462,6 +462,18 @@ def test_deletion_clears_last_modified(self):
         doc.delete()
         ok_(not cache.get(cache_key))
 
+    def test_deleted_doc_returns_410(self):
+        """Requesting a deleted doc returns 410"""
+        doc, rev = doc_rev()
+        doc.delete()
+        DocumentDeletionLog.objects.create(locale=doc.locale, slug=doc.slug,
+                                           user=rev.creator, reason="test")
+        self.url = reverse('wiki.document',
```
jezdez: Why `self` here? Couldn't this also use `doc.get_absolute_url()`?

jezdez:
I'm not sure if using a 410 is the best idea to be honest: http://searchenginewatch.com/sew/how-to/2340728/matt-cutts-on-how-google-handles-404-410-status-codes

"If a page is gone and you think it's temporary, go ahead and use a 404. If the page is gone and you know no other page that should substitute for it, you don't have anywhere else that you should point to, and you know that that page is going to be gone never come back, then go ahead and serve a 410."

groovecoder:
I'll have to bring in @teoli2003 for the `410` vs. `404` question.

@teoli2003 - this will serve `410` for all deleted docs, which might be too broad? I.e., should we only serve `410` for *spam* urls? Do we delete docs for other reasons?

groovecoder:
Also, updated to use `doc.get_absolute_url()`.

jezdez:
To be clear, I'm r- on using 410 for this.

teoli2003:
Most of the time we don't delete page, we make them redirects to more relevant pages (30x).
We delete mostly "sandbox" and spam.
Sandbox may be pages we recreated later when they have a correct name (that is when the wannabe editor follow a red link but put garbage in it). I don't know if this case is problematique: 200 -> 410 and long later again a 200 (with a completely new content).

groovecoder:
We're going to 4xx to get spam documents out of search indices. 404 will leave deleted documents in Google's index for 24 hours, but since we're currently serving 200 that's still a big improvement.

I'll change to 404. We can watch it to see if we still get too many visits from search engines to deleted spam docs. If so, let's re-consider using 410 aggressively when we know the urls are spam.

groovecoder:
Updated to 404.

PR #2997, opened 2015-01-13
Bug 962617 - Remove cronjobs

jezdez:


robhudson:
How does a migration removal work to remove the unused kpi app tables?

jezdez:
@robhudson Deleting the kpi app won't work with a migration since there is obviously not an app to store the migrations. I'll just delete the tables manually :)

robhudson:
> Deleting the kpi app won't work with a migration since there is obviously not an app to store the migrations. I'll just delete the tables manually :)

Ok. I wasn't sure if there was an app that was sort of a global app that you put these into. We're going to be moving towards Django's native migrations in 1.7 in the near distant future and I'm not very familiar with it as we've always used "schematic".

robhudson:
This looks good to me but where are the crontab changes? I.e. where is the removal of the `./manage.py cron` calls and where are the additions of the celery crons defined? I imagined something saying "Run this task at 11am daily" or something similar.

jezdez:
The crontab is maintained outside of the repo and :cyliang is going to help me with this. The ticket has the cronjobs we need to remove.

jezdez:
@robhudson Is this a r+?

PR #2999, opened 2015-01-14
Fix bug 1119314: Upgrade Django to 1.4.18.

ubernostrum:


jezdez:
doesn't that require an update of the requirements files as well?

jezdez:
Django doesn't seem to be in there, so let's punt on this and fix it when we are moving to peep

r+

PR #3001, opened 2015-01-14
Bug 1121440 - Link to article from revision view

SebastianZ:


kuma/wiki/templates/wiki/revision.html
```diff
@@ -8,7 +8,7 @@
 {% block content %}
 
   <article>
-    <h1>{{ document.title }}</h1>
+    <h1><a href="{{ url('wiki.document', document.full_path, locale=document.locale) }}">{{ _('{title}')|fe(title=document.title) }}</a></h1>
```
groovecoder: Spot-check looks good but I'd like @jezdez, @robhudson, or @ubernostrum to double-check me that this safe use of [`fe`](https://jingo.readthedocs.org/en/latest/#jingo.helpers.fe).

robhudson: Is `document.title` trusted input? Or does it require `|fe` because it may contain HTML that shouldn't be escaped? Or both?

Otherwise, you should probably use `|f` instead...
```
>>> t = env.from_string('<a href="">{{ title|fe(title=title) }}</a>')
>>> t.render({'title': '<script>alert("xss")</script>'})
u'<a href=""><script>alert("xss")</script></a>'
```

groovecoder: I always get this stuff confused. Bottom line is that `document.title` comes from users, so we don't trust it. We store any markup un-escaped, so it needs to be escaped in the template.

I.e., @SebastianZ - please revert the element content back to `{{ document.title }}`

SebastianZ: Sorry for the late reply.

> Bottom line is that document.title comes from users, so we don't trust it. We store any markup un-escaped, so it needs to be escaped in the template.
> 
> I.e., @SebastianZ - please revert the element content back to {{ document.title }}

Isn't the conclusion from your sentences above to rather keep the escaping? HTML needs to be escaped like for [](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img$revision/684477) for example. Note that https://github.com/mozilla/kuma/blob/dcb90e5263d432ea4a654910faf6d071f7f7dd82/kuma/wiki/templates/wiki/document_revisions.html#L14 is using the same code.

Sebastian

robhudson: TBH, I'm not even sure how `{{ _('{title}')|fe(title=document.title) }}` is supposed to work? `document.title` is pulled from the database and already localized right? Why does it need to be passed through `_` (aka `gettext`)?

I would think simply doing the follow is all you need:
`<a href="{{ url('wiki.document', document.full_path, locale=document.locale) }}">{{ document.title }}</a>`
and that should be escaped by default.

groovecoder: Okay, I told you I always get confused with this. :smile:

Both ways are safe on a local spot-check: `{{ document.title }}` and `{{ _('{title}')|fe(title=document.title) }}` with a title of `<script>alert("ouch")</script>` because [we've enabled Jinja autoescaping](https://github.com/mozilla/kuma/blob/master/settings.py#L559). A search in the code shows we use both methods thru-out, so there's no tie-breaker there. :frowning:

With that said, I prefer this `|fe` usage because it's more explicit.

@robhudson - I think yours is un-safe because you did `title|fe(title=title)` and not `'{title}'|fe(title=title)` ? I.e., you didn't name the var for string interpolation?

groovecoder: Duh, @robhudson is right - these titles are locale-specific already. So it doesn't need `_()`

SebastianZ: Ok, changed that now. Should I change the [other occurrances](https://github.com/mozilla/kuma/search?utf8=%E2%9C%93&q=%22{{+_%28%27{title}%27%29|fe%28title%3Ddocument.title%29+}}%22&type=Code), too?

Sebastian

kuma/wiki/templates/wiki/revision.html
```diff
@@ -8,7 +8,7 @@
 {% block content %}
 
   <article>
-    <h1>{{ document.title }}</h1>
+    <h1><a href="{{ url('wiki.document', document.full_path, locale=document.locale) }}">{{ document.title }}</a></h1>
```
jezdez: Please use `{{ document.get_absolute_url() }}` for the href here

SebastianZ: Fixed in https://github.com/SebastianZ/kuma/commit/937a37551e8b959f7fa0f6c4464ed61a353553c5.

Sebastian

jezdez:
Thanks @SebastianZ!

SebastianZ:
You're welcome!

Asking for the [other occurrances of such links](https://github.com/mozilla/kuma/search?utf8=%E2%9C%93&q=%22{{+_%28%27{title}%27%29|fe%28title%3Ddocument.title%29+}}%22&type=Code) again. Should I also change them to just `<a href="{{ document.get_absolute_url() }}">{{ document.title }}</a>`?

Sebastian

jezdez:
@SebastianZ Good idea, yeah!

SebastianZ:
Here's the pull request for it: https://github.com/mozilla/kuma/pull/3008

Sebastian

PR #3002, opened 2015-01-14
Bug 1119397: Install jQuery with Bower

openjck:


openjck:
Especially interested in what @jezdez and @groovecoder think. We can talk more about deployment strategy in the AWS meeting.

darkwing:
Nice work @openjck !

gulpfile.js
```diff
@@ -0,0 +1,16 @@
+var gulp = require('gulp');
+
+gulp.task('default', ['move-assets']);
+
+// Move assets to where other front-end build tasks expect them to be. This is a
+// temporary stopgap toward using Gulp for all front-end build tasks.
+gulp.task('move-assets', function() {
+
+    // Just a helper function to make the code below more clear
```
groovecoder: nice.

.gitignore
```diff
@@ -40,3 +40,7 @@ wheelhouse
 james.ini
 celeryev.pid
 media/js/libs/ckeditor/source/ckbuilder
+lib/front-end
+
+# Only needed until Gulp manages all front-end build tasks
+media/js/libs/jquery.js
```
groovecoder: I assume this will eventually be just `media/js/libs` ?

groovecoder:
Looks good to me. Do we need to work out how to use (or not) `gulp` and/or `bower` during deploy before we add more dependencies to bower?

Is it possible to use `gulp` to first try `bower install --offline` and only try to download dependencies if it sees `bower ENOTFOUND`? That would work kinda like `pip install --download` and `pip install --find-links` ...

.gitignore
openjck: Exactly. Depending on how our Gulp infrastructure evolves, it might even turn into something like `media/js/mdn.js.min` or similar.

jezdez:
I strongly suggest to *not* add bower/gulp to the deploy scheme, run those in development when working on a branch and checking in the results into git. That not only allows improved dev/prod parity (no surprises) with regard to the compiled result but also prevents out-of-sync-versions between the various servers.

jezdez:
@groovecoder asked on IRC the other day whether this isn't the same with Python packages, and the answer is yes and no. We use a few Python packages that require a C compiler and OS dependent libraries to compile, so we have to install those on demand during deployment. For the pure Python dependencies we'll ship a folder of release files to be used during deploy to reduce the number of vendored packages.

I don't see much overlap of this conceptually with the frontend library dependency tracking in this PR since the frontend files are assets that need to be processed before they are delivered to the user.

openjck:
Glad we're having this discussion. I'm not sure I have a strong opinion one way or the other. @jezdez would that approach extend to versioning compiled CSS, concatenated JavaScript, etc.?

jezdez:
@openjck Yeah, it would, there is no need to repeat the various steps to create the shippable product over and over on each deploy we create a build only if we change things. Here's what a good summary of that approach looks like: https://lincolnloop.com/blog/integrating-front-end-tools-your-django-project/ (gulp file here: https://gist.github.com/mlouro/8886076)

openjck:
Updated! Whew.

groovecoder:
Note: See also [the related dev-webdev thread](https://groups.google.com/d/msg/mozilla.dev.webdev/h--HrwGaJUU/CarZy_nwCP8J).

jezdez:
Hm, I think I miscommunicated my point of view, apologies everyone. I think committing we have two issues at hand:

1. dependencies and libraries to compile CSS and JS (etc) files
2. the compiled CSS and JS (etc) files

In my opinion, we should use the provisioning system to install (1) in the development environment, so that we can create (2) to commit them to the repo. We should use the most specific set of version identifiers the install tools allow so that we can easily recreate the state of (2) if needed.

groovecoder:
@jezdez :+1: and also seems to be the consensus in the dev-webdev thread.

I.e., commit the dependencies & libraries, don't commit compiled/generated files - re-compile and re-generate them during deployment.

jezdez:
Argh, that's *not* what I mean!

1. *don't* commit the dependencies but only package.json of 'kuma' to be used with a installer when provisioning the dev environment
2. *do* commit the compiled files created with the tools from 1.

I don't want us to rely on a super complex deploy server setups since we don't have control over it and compiling files slows deploys down. Instead compile only when needed -- when files change, locally.

openjck:
I'll switch back to downloading the dependencies during provision. I'll also tackle JavaScript minification/concatenation on the same branch, since it's going to chance some of this code anyway.

Closing this pull request for now. I'll open a replacement later this week.

PR #3003, opened 2015-01-14
Fix bug 1121051: Add absolutify helper

robhudson:


apps/contentflagging/models.py
```diff
@@ -71,9 +67,8 @@ def flag(self, request, object, flag_type, explanation, recipients=None):
             subject = subject.format(object=object)
             t = loader.get_template('contentflagging/email/flagged.ltxt')
             url = '/admin/contentflagging/contentflag/' + str(object.pk)
-            host = Site.objects.get_current().domain
-            content = t.render(Context({'url':url,'host':host,
-                                        'object':object,'flag_type':flag_type}))
+            content = t.render(Context(
+                {'url': url, 'object': object, 'flag_type': flag_type}))
```
jezdez: Please don't indent just one tab but visually to the next item, so this is correct:

    content = t.render(Context({'url': url,
                                'object': object,
                                'flag_type': flag_type}))

kuma/wiki/events.py
```diff
@@ -60,24 +55,6 @@ def context_dict(revision):
     return context
 
 
-def notification_mails(revision, subject, template, url, users_and_watches):
```
jezdez: Because this is unused?

kuma/wiki/helpers.py
```diff
@@ -112,12 +113,13 @@ def format_comment(rev):
...
     comment = bugize_text(rev.comment if rev.comment else "")
 
-    #  If a page move, say so
+    # If a page move, say so
     if prev_rev and prev_rev.slug != rev.slug:
-        comment += (jinja2.Markup('<span class="slug-change">'
-                                  '<span>%s</span>'
-                                  ' <i class="icon-long-arrow-right" aria-hidden="true"></i> '
-                                  '<span>%s</span></span>') % (prev_rev.slug, rev.slug))
+        comment += (jinja2.Markup(
```
jezdez: The indentation was correct

kuma/wiki/helpers.py
```diff
@@ -277,3 +279,15 @@ def document_zone_management_links(user, document):
                                   args=(zone.id,))
 
     return links
+
+
+@register.filter
+def absolutify(url, site=None):
```
jezdez: Could you check if `url` is just a slash and set it to an empty string here? So we don't have to do `absolutify('')` but `absolutify('/')`?

kuma/wiki/management/commands/render_document.py
```diff
@@ -39,12 +40,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         self.options = options
-
-        self.base_url = options['baseurl']
-        if not self.base_url:
-            from django.contrib.sites.models import Site
-            site = Site.objects.get_current()
-            self.base_url = 'http://%s' % site.domain
+        self.base_url = options['baseurl'] or absolutify('')
```
jezdez: I've left a comment for the absolutify helper, I think this shouldn't be `absolutify('')` but `absolutify('/')` to signal that it's the root URL path.

kuma/wiki/tests/test_helpers.py
```diff
@@ -15,6 +17,15 @@ def test_tojson(self):
...
             '{"title": "&lt;script&gt;alert(&quot;Hi!&quot;)&lt;/script&gt;"}')
 
+    @mock.patch_object(Site.objects, 'get_current')
+    def test_absolutify(self, get_current):
+        get_current.return_value.domain = 'testserver'
+
+        eq_(absolutify('/woo'), 'https://testserver/woo')
+        eq_(absolutify('http://domain.com'), 'http://domain.com')
+        site = Site(domain='otherserver')
+        eq_(absolutify('/woo', site), 'https://otherserver/woo')
```
jezdez: Needs a test for `absolutify('')` and `absolutify('/')`, maybe even `absolutify('//')`

kuma/wiki/helpers.py
```diff
@@ -277,3 +279,15 @@ def document_zone_management_links(user, document):
...
+    """Takes a URL and prepends the `site.domain` to it."""
+    if not site:
+        site = Site.objects.get_current()
+
+    if url.startswith('http'):
+        return url
+    else:
+        return 'https://{domain}{url}'.format(domain=site.domain, url=url)
```
jezdez: This should use urljoin.urlunparse to prevent any nasty issues with URL joining

kuma/wiki/views.py
```diff
@@ -471,9 +471,8 @@ def document(request, document_slug, document_locale):
             raise Http404
 
         # Check if we should fall back to default locale.
-        fallback_doc, fallback_reason, redirect_url = _default_locale_fallback(request,
-                                                                               document_slug,
-                                                                               document_locale)
+        fallback_doc, fallback_reason, redirect_url = _default_locale_fallback(
+            request, document_slug, document_locale)
```
jezdez: The indentation was correct

kuma/wiki/views.py
```diff
@@ -525,16 +524,14 @@ def document(request, document_slug, document_locale):
...
+        rendering_params[param] = request.GET.get(param, False) is not False
     rendering_params['section'] = request.GET.get('section', None)
     rendering_params['render_raw_fallback'] = False
     rendering_params['use_rendered'] = kumascript.should_use_rendered(doc, request.GET)
 
     # Step 4: Get us some HTML to play with.
-    doc_html, ks_errors, render_raw_fallback = _get_html_and_errors(request,
-                                                                    doc,
-                                                                    rendering_params)
+    doc_html, ks_errors, render_raw_fallback = _get_html_and_errors(
```
jezdez: The indentation was correct.

kuma/wiki/views.py
```diff
@@ -929,17 +932,18 @@ def new_document(request):
...
-                             'parent_slug': parent_slug,
-                             'parent_id': initial_parent_id,
-                             'document_form': doc_form,
-                             'revision_form': rev_form,
-                             'WIKI_DOCUMENT_TAG_SUGGESTIONS': constance.config.WIKI_DOCUMENT_TAG_SUGGESTIONS,
-                             'initial_tags': initial_tags,
-                             'allow_add_attachment': allow_add_attachment,
-                             'attachment_form': AttachmentRevisionForm(),
-                             'parent_path': parent_path})
+        return render(
```
jezdez: Could you move the definition of the context to a variable? that would make this indentation a lot more readable, I've cleaned this up in some places here and there

kuma/wiki/events.py
robhudson: Yes. I found no use of it in the code. The last usage was removed a couple years ago by Ricky from what I could find.

kuma/wiki/helpers.py
robhudson: Good idea.

kuma/wiki/helpers.py
robhudson: I like that idea.

kuma/wiki/views.py
robhudson: In this case it was more than 80 characters

robhudson:
@jezdez Rebased and addressed comments.

kuma/wiki/views.py
jezdez: Ah, ok, then leave it

kuma/wiki/views.py
robhudson: Wrapping for 80 characters.

jezdez:
Looks like a legitimate test failure

robhudson:
Tests fixed. When I switched to `urlparse` it lost the query parameters. I added those back and am using `urlunparse`.

jezdez:
:heart_eyes_cat:

PR #3006, opened 2015-01-15
bug 1113260 - add django-banish to ban by IP

groovecoder:
:disappointed: to add another dependency via submodule, but ...

This will help us fight spammers more efficiently. We can ban IP addresses ourselves with the django admin site interface rather than filing a NetOps bug every time.

To check:

1. `git submodule update --init` to get `django-banish` into `vendor/src`
2. `vagrant provision` to run `syncdb` to create the `banish` models
3. Activate the `store_revision_ips` [waffle switch](https://developer-local.allizom.org/admin/waffle/switch/)
4. Make a revision
5. Go to [the revision dashboard](https://developer-local.allizom.org/en-US/dashboards/revisions)
6. [ ] Click the "Toggle IPs" button
7. [ ] Click "Ban IP"
8. [ ] Enter a reason for the ban
9. [ ] Click Save
10. You should get "Permission Denied" errors for every request

Now, to un-banish yourself:

1. Set `BANISH_ENABLED = False` in `settings.py`
2. Go back to [the banishments admin interface](https://developer-local.allizom.org/admin/banish/banishment/) and delete the banishment record

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1269,6 +1269,6 @@ SOCIALACCOUNT_PROVIDERS = {
 # django-banish defaults; listing here to be explicit
 BANISH_ENABLED = True
 BANISH_EMPTY_UA = True
-BANISH_ABUSE_THRESHOLD = 75
+BANISH_ABUSE_THRESHOLD = 9999999 # TODO: https://bugzil.la/1122658
 BANISH_USE_HTTP_X_FORWARDED_FOR = True
 BANISH_MESSAGE = _("This connection has been banned for suspicious activity.")
--- a/settings.py
+++ b/settings.py
@@ -1270,4 +1270,5 @@ SOCIALACCOUNT_PROVIDERS = {
 BANISH_ENABLED = True
 BANISH_EMPTY_UA = True
 BANISH_ABUSE_THRESHOLD = 75
+BANISH_USE_HTTP_X_FORWARDED_FOR = True
 BANISH_MESSAGE = _("This connection has been banned for suspicious activity.")
--- a/settings.py
+++ b/settings.py
@@ -1,9 +1,10 @@
 # Django settings for kuma project.
+from collections import namedtuple
+import json
 import logging
 import os
 import platform
-import json
-from collections import namedtuple
+import sys
 
 from django.utils.functional import lazy
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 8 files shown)


jezdez:
I haven't spotchecked this yet, but I think we should first find out how to get the real IP of a user in production. If it's done via the `HTTP_X_FORWARDED_FOR` header then we should set the `BANISH_USE_HTTP_X_FORWARDED_FOR` setting to `True`

jezdez:
If there is another way then we may have to subclass the middleware and reimplment _get_ip

groovecoder:
Good catch. We've [started by using `X-Forwarded-For`](https://bugzilla.mozilla.org/show_bug.cgi?id=1110904#c4). It's been working so far in [the `get_ip` function](https://github.com/groovecoder/kuma/blob/f56eae800a5e84c422c2561eb58412b22e9926e1/apps/actioncounters/utils.py#L11-L40) we use [while saving revisions](https://github.com/mozilla/kuma/pull/2960/files#diff-08089c7053931e5d5c6afbb6fb3e2533R2164) to store the revision IPs in the production environment. So, I've updated to set `BANISH_USE_HTTP_X_FORWARDED_FOR` to `True`. (Also disabled banish during tests; it was automatically banishing the tests for sending requests too fast)

jezdez:
Hm, speaking of which, given the fact that kumascript GETs the raw document data (and others?) from kuma, I wonder if that will mean we'll have to unban it all the time?

groovecoder:
@jezdez - good point. I'll dig into that.

groovecoder:
Since we are first using django-banish to implement IP-based banning, I'm re-opening this with a commit that effectively unlimits the request threshold so that KumaScript can still function. We could still merge and push this for the IP banning and then follow it up with another change to implement the automatic throttling.

jezdez:
r+

groovecoder:
Updated with link to show all IPs in the revision dashboard, and then to Ban the IP with a new button in each row.

groovecoder:
spam

settings.py
```diff
@@ -1263,3 +1265,10 @@ def get_user_url(user):
         }
     }
 }
+
+# django-banish defaults; listing here to be explicit
+BANISH_ENABLED = True
+BANISH_EMPTY_UA = True
+BANISH_ABUSE_THRESHOLD = 9999999 # TODO: https://bugzil.la/1122658
```
ubernostrum: Maybe use `sys.maxint` instead? :)

groovecoder:
updated threshold to `sys.maxint`

ubernostrum:
Looks good to me. r+

PR #3010, opened 2015-01-19
Fix Bug 1122997 - Ban user button display

stephaniehobson:
Changed if statement that displays ban button to use same permission check that generates the ban button.


stephaniehobson:
No sure why we need to check permissions here as well as in the helper that generates the ban link?

PR #3011, opened 2015-01-20
Bug 756533 - For a handful of fixes

jezdez:


jezdez:
I've renamed a few loggers from mdn.* to kuma.* and that needs to happen unless we want to see the logging module to raise warning all the time.

robhudson:
r+

robhudson:
The one note I have, and maybe it's specific to zamboni, is we have a test similar to this that Django 1.7 complained about b/c the `INSTALLED_APPS` python path differed from the import path during test run. We had something like `'zamboni.mkt.translations.tests.models.TestModel' != 'mkt.translations.tests.models.TestModel'`. What I had to do was to get the model from the app config instead of importing it.

jezdez:
@robhudson Ah, good to know, I think we may be able to fix this by not using nose anymore ;)

PR #3012, opened 2015-01-20
Fix various scripts after apps/ removal

robhudson:


groovecoder:
post-merge nit: all commits are supposed to reference a bug #. We can reference bugs that are already closed are bugs that are simply tracking bugs; just need some reference to what motivated the change.

PR #3013, opened 2015-01-21
bug 957802 - import actioncounters.utils from kuma

groovecoder:
Spot-check == `vagrant provision` should work without errors.


robhudson:
I grepped for other occurrences and this was the only one I could find. r+.

KE95:
Fyck

PR #3015, opened 2015-01-22
A few more fixes for bug 756533.

jezdez:


jezdez:
I'm a bit unsure if that will mean that prod and stage will use the debug logger. The whole settings setup is so messed up I don't know what to do

settings.py
```diff
@@ -208,8 +210,22 @@
     if delivered_lang in PROD_LANGUAGES:
         LANGUAGE_URL_MAP[requested_lang.lower()] = delivered_lang
 
-# Override Django's built-in with our native names
+
+def get_locales():
+    locales = {}
+    file = os.path.join(ROOT, 'kuma', 'languages.json')
+    json_locales = json.loads(open(file, 'r').read())
```
robhudson: Simpler might be to just use `json.load`?

jezdez:
The circular logging filter issue was fixed in Django 1.5 in 2012 (https://github.com/django/django/commit/fc69fff9ab5bba8a6cff3eaf51f02a3204b1c015)

settings.py
jezdez: Done.

settings.py
```diff
@@ -208,8 +210,22 @@
     if delivered_lang in PROD_LANGUAGES:
         LANGUAGE_URL_MAP[requested_lang.lower()] = delivered_lang
 
-# Override Django's built-in with our native names
+
+def get_locales():
+    locales = {}
+    file = os.path.join(ROOT, 'kuma', 'languages.json')
```
robhudson: On my filesystem this is at `kuma/core/languages.json`. So I would think you need to add the extra `'core'` in there.

settings.py
```diff
@@ -208,8 +210,22 @@
...
+    json_locales = json.load(open(file, 'r'))
+    for locale, meta in json_locales.items():
+        locales[locale] = _Language(meta['english'],
+                                    meta['native'],
+                                    meta['iso639_1'])
```
robhudson: Generally curious why this needs to be a namedtuple and not just the dict imported from the json file?

settings.py
jezdez: The commit moves the file as well

settings.py
jezdez: I have no idea

settings.py
robhudson: Ah, I see that now. :+1:

robhudson:
r+

PR #3019, opened 2015-01-23
Fix bug 1125295: Remove sticky A/B test activation

openjck:
This reverts commit aeda5373.

### DIFF
--- a/media/redesign/js/wiki.js
+++ b/media/redesign/js/wiki.js
@@ -734,15 +734,4 @@
             });
         };
     })();
-
-    /*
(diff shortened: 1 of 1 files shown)


stephaniehobson:
I seem to lost my dev environment somehow. It is still my intention to review this PR but I'm not going to get to it before at least tomorrow afternoon :/

openjck:
No problem. This is a super simple change, so even if it bit-rots it should be easy to restore. Take as much time as you need.

stephaniehobson:
Seems legit.

PR #3020, opened 2015-01-23
bug 1121918 - put authkeys behind permission

groovecoder:
This puts the Add & Delete auth keys views behind the respective permissions. It also includes a data migration to delete all existing authkeys.
api-key-permission

### DIFF
--- a/kuma/authkeys/migrations/0003_remove_authkeys.py
+++ b/kuma/authkeys/migrations/0003_remove_authkeys.py
@@ -8,7 +8,9 @@ class Migration(DataMigration):
 
     def forwards(self, orm):
         "Write your forwards methods here."
-        orm.Key.objects.all().delete()
+        (orm.Key.objects.all().exclude(
+            id__in=orm.KeyAction.objects.values_list('key_id', flat=True))
+         .delete())
 
     def backwards(self, orm):
         "Write your backwards methods here."
--- a/kuma/authkeys/migrations/0003_remove_authkeys.py
+++ b/kuma/authkeys/migrations/0003_remove_authkeys.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        "Write your forwards methods here."
+        orm.Key.objects.all().delete()
+
+    def backwards(self, orm):
+        "Write your backwards methods here."
+        # No way to go back
+
+    models = {
+        'auth.group': {
+            'Meta': {'object_name': 'Group'},
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'name': ('django.db.models.fields.CharField', [], {'unique': 'True', 'max_length': '80'}),
+            'permissions': ('django.db.models.fields.related.ManyToManyField', [], {'to': "orm['auth.Permission']", 'symmetrical': 'False', 'blank': 'True'})
+        },
+        'auth.permission': {
+            'Meta': {'ordering': "('content_type__app_label', 'content_type__model', 'codename')", 'unique_together': "(('content_type', 'codename'),)", 'object_name': 'Permission'},
+            'codename': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            'content_type': ('django.db.models.fields.related.ForeignKey', [], {'to': "orm['contenttypes.ContentType']"}),
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'name': ('django.db.models.fields.CharField', [], {'max_length': '50'})
+        },
+        'auth.user': {
+            'Meta': {'object_name': 'User'},
+            'date_joined': ('django.db.models.fields.DateTimeField', [], {'default': 'datetime.datetime.now'}),
+            'email': ('django.db.models.fields.EmailField', [], {'max_length': '75', 'blank': 'True'}),
+            'first_name': ('django.db.models.fields.CharField', [], {'max_length': '30', 'blank': 'True'}),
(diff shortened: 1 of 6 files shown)


kuma/authkeys/views.py
```diff
@@ -12,7 +11,7 @@
 ITEMS_PER_PAGE = 15
 
 
-@login_required
+@permission_required('authkeys.add_key', raise_exception=True)
```
groovecoder: Should we keep both decorators?

kuma/users/helpers.py
```diff
@@ -42,10 +42,20 @@ def ban_link(context, ban_user, banner_user):
...
+    return Markup(link)
+
+
+@register.function
+@contextfunction
+def admin_link(context, user):
+    """Returns a link to admin a user"""
+    link = ''
+    url = reverse('admin:auth_user_change', args=(user.id,))
+    link = '<a href="%s" class="button neutral">%s<i aria-hidden="true" class="icon-lock"></i></a>' % (url, _('Admin'))
```
groovecoder: Can @openjck or @stephaniehobson verify this markup is good?

groovecoder:
Ping @jezdez, @ubernostrum, or @robhudson for review of the back-end bits.

kuma/users/helpers.py
stephaniehobson: Markup looks good. Not sure lock is the right icon here, but I don't have any better suggestions.  Gavel maybe? I'd suggest the gear icon but we're already using it for manage when a user is on their own profile page - maybe we should be showing the manage menu when viewing other  people's profiles?

kuma/authkeys/templates/authkeys/list.html
```diff
@@ -22,14 +24,21 @@
...
 
         </li>
         {% endfor %}
     </table>
 {% endif %}
+
+{% else %}
+<p>{% trans %}
+  You need a special permission to use API keys.
+  You may request permission by <a href="https://bugzilla.mozilla.org/enter_bug.cgi?format=__standard__&product=Mozilla Developer Network&component=User management">filing a bug</a>.
```
robhudson: Would it be good to urlencode that string?

robhudson: I'd also move the URL outside the `trans`, like:
```
{% trans bugzilla_url='https://bugzilla.mozilla.org/enter_bug.cgi?format=__standard__&product=Mozilla Developer Network&component=User management' %}
...
You may request permission by <a href="{{ bugzilla_url }}">filing a bug</a>.
{% endtrans %}
```

kuma/authkeys/views.py
robhudson: The `permission_required` decorator doesn't check it, but also an anonymous user would not have those permissions. But I think I would add `login_required` b/c a login redirect is nicer than a 403.

kuma/authkeys/templates/authkeys/list.html
groovecoder: When I moved the url up into the `trans` declaration it was automatically url-encoded by jinja. :gift:

kuma/users/helpers.py
groovecoder: Yeah, first I reached for the gear. But you got me thinking in those memes and I found `icon-wrench`!

We would have to re-do lots of url mapping to make the manage menu viewable on other users' profiles. All those urls are coded to work on the current `request.user` object.

kuma/authkeys/views.py
groovecoder: Restored; it *is* nicer.

groovecoder:
Updated with first review fixes.

groovecoder:
Also updated the description with a spiffy GIF of how the new perm and API page works.

kuma/users/helpers.py
stephaniehobson: +1 for icon wrench :)

kuma/authkeys/migrations/0003_remove_authkeys.py
```diff
@@ -0,0 +1,76 @@
...
+import datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        "Write your forwards methods here."
+        orm.Key.objects.all().delete()
```
jezdez: Shouldn't that also delete the keyactions?

kuma/authkeys/tests/test_views.py
```diff
@@ -135,3 +146,52 @@ def test_delete_key(self):
...
+        _cache_bust_user_perms(self)
+
+        resp = self.client.get(url)
+        eq_(200, resp.status_code)
+
+
+def _cache_bust_user_perms(self):
+    # Need to cache-bust the user perms by re-fetching from DB
+    # https://docs.djangoproject.com/en/1.7/topics/auth/default/#permissions-and-authorization
+    self.user = User.objects.get(username=self.user.username)
```
jezdez: No need to put this in a pseudo private function, just put it in a base test class and inherit from it in both tests.

jezdez:
LGTM other than the two comments above.

kuma/authkeys/migrations/0003_remove_authkeys.py
groovecoder: Good point. When I ran it locally it deleted them automatically via the default cascade behavior - is that good enough?

But this reminds me that there is a `Key` with existing `KeyAction` records. Should I change this migration to only delete Keys that have not performed any actions? That will preserve the existing API action log.

jezdez: Ah, I thought for some reason that South's fake ORM wouldn't trigger the default cascade delete, sorry.

Yeah, I think changing it to only delete the really unused keys makes sense +1

groovecoder:
Updated again. Should be ready. Unless we want to clean up the commit history so the migration file is only touched in a single commit?

jezdez:
@groovecoder Nah, works for me :) r+

PR #3026, opened 2015-01-27
Fix bug 1126461: Fix overlapping sticky buttons

openjck:
https://bugzilla.mozilla.org/show_bug.cgi?id=1126461#c0

### DIFF
--- a/media/redesign/stylus/components/wiki/page-buttons.styl
+++ b/media/redesign/stylus/components/wiki/page-buttons.styl
@@ -5,6 +5,8 @@
     width: auto;
 
     &.fixed {
+        background-color: $default-background-color;
+        border-radius: 0 0 0 10px;
         overflow-y: inherit;
         z-index: 50;
     }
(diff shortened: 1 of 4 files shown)


openjck:
@shobson: Especially interested in your thoughts on the design when the table of contents is collapsed (the "small desktop" media query size).

A screenshot of page buttons appearing in the top-right corner of the page, blocking other content that would otherwise be seen there

This works well in some situations (like the one pictured) but can be annoying when it covers text. I also considered extending the width to `100%`, so that the buttons appear to be inside a white horizontal bar that covered the whole top of the page, but that was annoying in a different way.

openjck:
There was also a layout edge-case in the original feature. See [this bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1126461#c2) for more information.

@darkwing: If you have the time, I'd appreciate your review of the layout fix (the changes to wiki.js).

openjck:
I think it would be good to get something in sooner than later. The bugs on production are pretty annoying. Maybe we could follow this PR up with one that improves the design/UX further?

media/redesign/stylus/components/wiki/page-buttons.styl
```diff
@@ -5,6 +5,8 @@
     width: auto;
 
     &.fixed {
+        background-color: $default-background-color;
+        border-radius: 0 0 0 10px;
```
stephaniehobson: Needs to be bidi friendly.

PR #3028, opened 2015-01-28
Bug 1124199: Revert default sticky buttons

openjck:
This reverts commit d9369db700f2f1e1e226b5fb57936582fe5c418c.

After turning sticky page buttons on by default, a number of bugs were
discovered. This commit turns the sticky page buttons back off by
default so that the bugs can be fixed.


stephaniehobson:
Buttons are unsticky, TOC is still sticky, sticky buttons can be triggered by changing data attribute. Looks good.

PR #3032, opened 2015-01-29
Fix bug 1120526 - Move to using elasticsearch-dsl

robhudson:
This also requires the backend Elasticsearch server be upgraded to 1.3.x and refactors search code in general.


robhudson:
ES Query Comparison: https://gist.github.com/robhudson/2b49597977f0f9576b77 (The top JSON is master, the lower JSON is this branch)

There are some things to fix there, like the function score query inside the bool query and the missing facets.

robhudson:
I looked at the coveralls data for these and things seem pretty well tested. There are a few areas where facet tests could be improved but I may save those for when we switch to using aggregations.

As such, I'm removing the 'not-ready' label. Let the criticism flow forth in great abundance.

kuma/search/commands.py
```diff
@@ -3,47 +3,58 @@
...
+    index_name = index.prefixed_name
 
-    if index is None:
-        index = get_index()
+    es = cls.get_connection('indexing')
 
-    log.info('Wiping and recreating %s....', index)
-    recreate_index(es=es, index=index)
+    log.info('Wiping and recreating %s....', index_name)
+    recreate_index(es=es, index=index_name)
```
jezdez: Let's move recreate_index into the Index model

kuma/search/index.py
```diff
@@ -1,69 +1,13 @@
...
-                       if k.startswith(settings.ES_INDEX_PREFIX))
+from .models import Index
 
-    indexes = [(k, v['docs']['num_docs']) for k, v in indexes.items()]
 
-    return indexes
+log = logging.getLogger('kuma.search.index')
 
 
 def delete_index_if_exists(index):
```
jezdez: Let's move this into the Index model manager

kuma/search/index.py
```diff
@@ -72,138 +16,36 @@ def delete_index_if_exists(index):
...
-
-    """
-    stats = {}
-    for name, cls in get_mapping_types().items():
-        stats[name] = S(cls).count()
-
-    return stats
-
-
 def recreate_index(es=None, index=None):
```
jezdez: Let's move this into the Index model class

kuma/search/paginator.py
```diff
@@ -0,0 +1,58 @@
...
+
+        """
+        number = self.validate_number(number)
+        bottom = (number - 1) * self.per_page
+        top = bottom + self.per_page
+
+        # Force the search to evaluate and then attach the count. We want to
+        # avoid an extra useless query even if there are no results, so we
+        # directly fetch the count from hits.
+        resp = self.object_list[bottom:top].execute()
```
jezdez: can we call that "result" instead of "resp"?

kuma/search/commands.py
```diff
@@ -3,47 +3,58 @@
...
 from .utils import chunked, format_time
 
+
 log = logging.getLogger('kuma.search.commands')
 
 
 @requires_good_connection
-def es_reindex_cmd(percent=100, mapping_types=None,
-                   chunk_size=1000, index=None):
+def es_reindex_cmd(chunk_size=1000, index=None, percent=100):
```
jezdez: Let's rename that to `reindex_all` and move into the `WikiDocumentType` class (or a parent class if you want to keep it agnostic)

kuma/search/tests/test_types.py
```diff
@@ -1,61 +1,49 @@
...
-
     def test_current_locale_results(self):
         self.refresh()
-        results = (DocumentType.search().query(or_={'title': 'article',
-                                                    'content': 'article'})
-                                        .filter(locale='en-US')
-                                        .highlight('content'))
-        for doc in results:
+        results = (WikiDocumentType.search()
+                   .query(query.Match(title='article') |
```
jezdez: Indent to matching dot

kuma/search/views.py
```diff
@@ -48,15 +49,20 @@ def initial(self, request, *args, **kwargs):
...
     def get_queryset(self):
-        return DocumentS(DocumentType,
-                         url=self.request.get_full_path(),
-                         current_page=self.current_page,
-                         serialized_filters=self.serialized_filters,
-                         selected_filters=self.selected_filters)
+        return WikiDocumentType.search()
+
+    def list(self, request, *args, **kwargs):
+        """
```
jezdez: Add comment to method

kuma/search/views.py
```diff
@@ -48,15 +49,20 @@ def initial(self, request, *args, **kwargs):
...
 
     def get_queryset(self):
-        return DocumentS(DocumentType,
-                         url=self.request.get_full_path(),
-                         current_page=self.current_page,
-                         serialized_filters=self.serialized_filters,
-                         selected_filters=self.selected_filters)
+        return WikiDocumentType.search()
+
+    def list(self, request, *args, **kwargs):
```
jezdez: I wonder if list is the correct method to override here, not dispatch?

kuma/wiki/models.py
```diff
@@ -1471,8 +1472,7 @@ def is_watched_by(self, user):
         return EditDocumentEvent.is_notifying(user, self)
 
     def get_mapping_type(self):
```
jezdez: Rename that to `get_document_type`?

kuma/wiki/search.py
```diff
@@ -0,0 +1,255 @@
...
+from django.utils.html import strip_tags
+
+from elasticsearch.helpers import bulk
+from elasticsearch_dsl import document, field
+from elasticsearch_dsl.connections import connections
+from elasticsearch_dsl.mapping import Mapping
+from elasticsearch_dsl.search import Search
+
+
+# Configure Elasticsearch connections for connection pooling.
```
jezdez: I think we should prevent accidentally calling this multiple times, how about putting that into a function and call it in the models.py? Then once we're on Django 1.7 we can move this to a AppConfig class's setup method

kuma/wiki/search.py
```diff
@@ -0,0 +1,255 @@
...
+    def get_settings(cls):
+        settings = {
+            'mappings': cls.get_mapping(),
+            'settings': {
+                'index': {
+                    'analysis': cls.get_analysis()
+                }
+            }
+        }
+
```
jezdez: Remove empty line

kuma/wiki/search.py
```diff
@@ -0,0 +1,255 @@
...
+        for exclude in cls.exclude_slugs:
+            excludes.append(Q(slug__icontains=exclude))
+
+        qs = (model.objects
+                   .filter(is_template=False,
+                           is_redirect=False,
+                           deleted=False)
+                   .exclude(reduce(operator.or_, excludes)))
+
+        percent = float(percent) / 100
```
jezdez: We should use the new division behavior here to prepare ourselves for future Python version. So add `from __future__ import division` and use `//`

kuma/wiki/tasks.py
```diff
@@ -266,3 +268,57 @@ def build_sitemaps():
...
+    :arg ids: Iterable of `Document` pks to remove.
+    :arg index_pk: The `Index` pk of the index to remove items from.
+
+    """
+    cls = WikiDocumentType
+    es = cls.get_connection('indexing')
+    index = Index.objects.get(pk=index_pk)
+
+    for pk in ids:
+        es.delete(index=index.prefixed_name, doc_type=cls.get_doc_type(),
```
jezdez: I think there is a way to bulk delete documents, so maybe add a `WikiDocumentType.bulk_delete` method and call that?

puppet/manifests/dev-vagrant.pp
```diff
@@ -34,10 +34,14 @@
...
         'index' => {
           'number_of_replicas' => '0',
-          'number_of_shards'   => '1'
+          'number_of_shards'   => '3'
         },
         'network' => {
           'host' => '0.0.0.0'
+        },
+        'discovery.zen.ping' => {
+            'multicast.enabled' => false,
```
jezdez: Nice!

scripts/travis-install
```diff
@@ -13,7 +13,7 @@ then
     aria2c -j 10 https://s3-us-west-2.amazonaws.com/pkgs.mozilla.net/locale/mdn/locale.tar.gz
     tar -C .. -zxf product_details_json.tar.gz
     tar xvfz locale.tar.gz
-    wget https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-0.90.13.deb
-    sudo dpkg --force-confdef -i elasticsearch-0.90.13.deb
+    wget https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-1.3.7.deb
```
jezdez: The Puppet manifest above uses 1.2.4, which one are we moving to?

settings.py
```diff
@@ -1156,8 +1146,17 @@ def JINJA_CONFIG():
 
 KUMASCRIPT_URL_TEMPLATE = 'http://developer.mozilla.org:9080/docs/{path}'
 
+# Elasticsearch related settings.
+ES_DEFAULT_NUM_REPLICAS = 1
+ES_DEFAULT_NUM_SHARDS = 5
```
jezdez: This seems to not being used, right?

jezdez:
Amazing work @robhudson. Things are not only much better in the library usage but also generally the code quality improved dramatically. Thank you!

I've left a few comments, most of which are minor issues.

kuma/search/views.py
jezdez: ok

kuma/search/commands.py
robhudson: Ok. I filed a bug for the `@requires_good_connection` change.

scripts/travis-install
robhudson: Updated to 1.3.7

robhudson:
Updated all the things. Thanks for the review @jezdez! Ready for another pass or merging when backend is ready.

jezdez:
Okay, almost there, this only needs fixes for the two issues above and then we can squash and merge this. The commit message needs a "Fixes bug 1120526" prefix.

PR #3035, opened 2015-01-30
Bug 756533 - Slightly revert logging changes from 96d0d221bb3d297b4af1bcb9e66ebe614f351533.

jezdez:
This increases the logging threadshold to just errors and reintroduces
the RequireDebugFalse filter for the admin email log handler.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1165,6 +1165,11 @@ SYSLOG_TAG = 'http_app_kuma'
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
+    'filters': {
+        'require_debug_false': {
+            '()': 'django.utils.log.RequireDebugFalse',
+        },
+    },
     'formatters': {
         'default': {
             'format': '{0}: %(asctime)s %(name)s:%(levelname)s %(message)s: '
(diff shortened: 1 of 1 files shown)


robhudson:
While we're at it could we change `LOG_LEVEL = logging.WARN` to `logging.DEBUG`? This prevents the console logger from showing any logs that are at `DEBUG` level. Either that or change the console logger to not use this variable.

settings.py
```diff
@@ -1186,18 +1192,16 @@ def JINJA_CONFIG():
...
             'handlers': ['console'],
             'propagate': True,
-            # Use the most permissive setting. It is filtered in the handlers.
-            'level': logging.DEBUG,
+            'level': logging.ERROR,
         },
         'elasticsearch': {
-            'level': logging.ERROR,
             'handlers': ['console'],
+            'level': logging.ERROR,
```
robhudson: I would love it if MDN had a logging facility set up so I can view the logs from the staging server. Then in my ES commit I would change this to DEBUG to check on this for a week or so.

jezdez: We've talked to webops but it always didn't work out. With cloudops that was promised to be different.

PR #3038, opened 2015-01-30
Fix bug 1035569 - Clean out query parameters that are empty.

jezdez:
This prevents a hard exception on the search page when given a query parameter that doesn't have a value.

### DIFF
--- a/kuma/search/tests/test_utils.py
+++ b/kuma/search/tests/test_utils.py
@@ -43,6 +43,12 @@ class URLTests(test_utils.TestCase):
         eq_(url.merge_query_param('foo', [None]),
             'http://example.com/?foo=&foo=bar&spam=eggs')
 
+    def test_clean_params(self):
+        for url in ['http://example.com/?spam=',
+                    'http://example.com/?spam']:
+            url_object = QueryURLObject(url)
+            eq_(url_object.clean_params(url_object.query_dict), {})
+
     def test_referer_bad_encoding(self):
         class _TestRequest(object):
             # In order to test this we just need an object that has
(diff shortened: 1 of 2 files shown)


kuma/search/tests/test_utils.py
```diff
@@ -43,6 +43,19 @@ def test_merge_query_param(self):
...
+        url = QueryURLObject(original)
+
+        eq_(url.clean_params(url.query_dict),
+            {})
+
+        original = 'http://example.com/?spam'
+        url = QueryURLObject(original)
+
+        eq_(url.clean_params(url.query_dict),
+            {})
```
robhudson: This could be in a loop... `for original_url in ('http://example.com/?spam=', 'http://example.com/?spam'): ...`

kuma/search/utils.py
```diff
@@ -30,10 +30,20 @@ def merge_query_param(self, name, value):
...
+        return self.without_query().set_query_params(self.clean_params(params))
+
+    def clean_params(self, params):
+        """
+        Cleans query parameters that don't have a value to not freak out
+        urllib's quoting and Django's form system.
+        """
+        clean_params = {}
+        for param, defaults in params.items():
+            if defaults not in ('', None, [None], ['']):
```
robhudson: for string types would we want to `.strip()`, for cases like `' '`?

kuma/search/utils.py
```diff
@@ -30,10 +30,24 @@ def merge_query_param(self, name, value):
...
+        """
+        clean_params = {}
+        for param, default in params.items():
+            if isinstance(default, list) and len(default) == 1:
+                default = default[0]
+            if isinstance(default, basestring):
+                default = default.strip()
+            if default not in ('', None):
+                clean_params[param] = default
+        return clean_params
```
robhudson: I think that's nice.

PR #3040, opened 2015-01-30
Fix bug 1104127 - Use Memcache for devmo_url and slight simplification.

jezdez:


kuma/wiki/helpers.py
```diff
@@ -318,22 +320,20 @@ def devmo_url(context, path):
...
     """
+    fallback_locale = settings.WIKI_DEFAULT_LANGUAGE
+
     if hasattr(context['request'], 'locale'):
         locale = context['request'].locale
     else:
-        locale = settings.WIKI_DEFAULT_LANGUAGE
-    try:
-        url = cache.get('devmo_url:%s_%s' % (locale, path))
-    except:
```
jezdez: I removed the blank except here since that's something we need to know when it happens so we can find the problem behind it.

kuma/wiki/helpers.py
```diff
@@ -318,22 +320,20 @@ def devmo_url(context, path):
     Create a URL pointing to Kuma.
     Look for a wiki page in the current locale, or default to given path
     """
+    fallback_locale = settings.WIKI_DEFAULT_LANGUAGE
+
     if hasattr(context['request'], 'locale'):
         locale = context['request'].locale
     else:
-        locale = settings.WIKI_DEFAULT_LANGUAGE
```
robhudson: Could this be rewritten as `locale = getattr(context['request'], 'locale', fallback_locale)`?

kuma/wiki/helpers.py
```diff
@@ -343,9 +343,8 @@ def devmo_url(context, path):
                 """
                 child = Document.objects.get(locale=locale,
                                              parent=parent)
```
robhudson: This looks like it could fit on one line?

kuma/wiki/helpers.py
```diff
@@ -318,22 +320,20 @@ def devmo_url(context, path):
...
+
+    url = memcache.get('devmo_url:%s_%s' % (locale, path))
+
     if not url:
-        url = reverse('wiki.document',
-                      locale=settings.WIKI_DEFAULT_LANGUAGE,
-                      args=[path])
-        if locale != settings.WIKI_DEFAULT_LANGUAGE:
+        url = reverse('wiki.document', locale=fallback_locale, args=[path])
+        if locale != fallback_locale:
```
robhudson: I have to admit I don't understand this `if` section at all.

kuma/wiki/helpers.py
jezdez: Yes

kuma/wiki/helpers.py
jezdez: If the currently selected language isn't the default language (English) then we try to get the wiki document in English because apparently the language specific page didn't exist.

robhudson: It could have not existed b/c of a cache miss tho, right?

jezdez: not sure

robhudson: ISTM that this is conflating a cache miss with a document/locale combination that doesn't exist. I won't merge as long as that seems the case. The rest of the changes look good tho.

jezdez: I don't understand what you're trying to say, maybe @groovecoder can help as he wrote that thing IIRC

kuma/wiki/models.py
```diff
@@ -1478,6 +1481,11 @@ def get_contributors(self):
         return User.objects.filter(pk__in=list(top_creator_ids))
 
 
+@receiver(render_done)
+def purge_wiki_url_cache(sender, instance, **kwargs):
+    memcache.delete(u'wiki_url:%s:%s' % (instance.locale, instance.slug))
```
robhudson: Would it make sense to set the cache here instead of just deleting it? This works but I'm just curious.

kuma/wiki/helpers.py
```diff
@@ -22,7 +21,7 @@
 
 from kuma.core.urlresolvers import reverse
 from .constants import DIFF_WRAP_COLUMN
-from .models import Document
+from .models import Document, memcache
```
robhudson: I'm curious why you do this instead of calling `get_cache`?

jezdez: get_cache in old Django's had the problem of creating cache clients all the time, so the fewer we do that the less connections are open against the memcache cluster

kuma/wiki/models.py
jezdez: I'm reluctant to cache warm this since the render system is known to be having race conditions. The less state I add the better. Then there is also the question what "render_done" is compared to just save and delete for example. E.g. imagine a document translation that would be deleted, the URL should point at the original document from that point on, but we'd have both the parent document rendered as well as the old document being deleted. I sure this would create some kind of conflict eventually.

jezdez: OTOH we could drastically reduce the number of SQL gets :-/

PR #3041, opened 2015-01-30
Build JavaScript using Gulp and install some JavaScript libraries with Bower

openjck:
The more testing we can do, the better. This changes a lot of important resources.

Some things to think about:

* This ends the use of jingo-minify for building JavaScript assets. Has it been replaced completely? Have all of the `{{ js() }}` calls been replaced? Aside from the `{{ js() }}` calls, is there anything else that needs to be updated?
* Are JavaScript-powered features still working well?
* Are we on board with this strategy of committing generated files (like minified/concatenated JavaScript) but not the tools that generated them (like Gulp plugins and Bower packages)? Do we like how the approach was put into practice (using `gulp watch` in foreman)?

Other notes:

* You'll need to `vagrant provision` before reviewing this to get the latest copy of ~/.bash_profile in the VM


package.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "MDN",
```
jezdez: Could you name that kuma as that's the name of the software while MDN is the product?

openjck: Sure

openjck:
I was adding commits to this PR yesterday and today, but I'll stop now so that we can review something more stable. So this is ready to be reviewed whenever.

jezdez:
@openjck I'll look tomorrow

darkwing:
On "/en-US/docs/new?slug=Tools", I get this error:

```
File "/home/vagrant/src/templates/base.html", line 206, in block "site_js"
19:37:25 web.1        |     {{ js(script) }}
19:37:25 web.1        |   File "/home/vagrant/src/vendor/src/jingo-minify/jingo_minify/helpers.py", line 134, in js
19:37:25 web.1        |     urls = get_js_urls(bundle, debug)
19:37:25 web.1        |   File "/home/vagrant/src/vendor/src/jingo-minify/jingo_minify/helpers.py", line 67, in get_js_urls
19:37:25 web.1        |     item in settings.MINIFY_BUNDLES['js'][bundle]]
19:37:25 web.1        | KeyError: 'js'
```

openjck:
> On "/en-US/docs/new?slug=Tools", I get this error:

Nice catch. Fixed.

openjck:
Just stumbled across some instances of `BUILD_ID_JS` that need to be cleaned up. That'll take some thought.

openjck:
This introduces a new approach to cache-busting our assets.

On master, we append `BUILD_ID_JS` to asset filenames to cache-bust them. The value of `BUILD_ID_JS` changes every time we do a push, whether or not any files changed.

On this branch, we cache-bust each asset with a hash of the file contents. That way, assets are only cache-bust when they really change.

Sounds like a small change, but it could be a big help considering how often we push and how infrequently some assets change.

kuma/core/helpers.py
```diff
@@ -349,3 +351,17 @@ def number(context, n):
...
     return format_decimal(n, locale=_babel_locale(_contextual_locale(context)))
+
+
+@register.function
+@jinja2.contextfunction
+def cache_bust(context, path):
+    """Return the versioned filename of an asset in media/, given its
+    non-versioned filename."""
+    with open('media/rev-manifest.json') as manifest_data:
+        manifest = json.loads(manifest_data.read())
```
jezdez: use `json.load(manifest_data)` here

kuma/core/helpers.py
```diff
@@ -349,3 +351,17 @@ def number(context, n):
...
+
+        try:
+            return static(settings.MEDIA_URL + manifest[path])
```
jezdez: uh, that doesn't make sense, the `MEDIA_URL` settig shouldn't be used with the static template tag as that just does a URL join with `STATIC_URL`. `MEDIA_URL` is `/media/` by default so this will be effectively doing:

urljoin('/media/' + path_to_file)

Also, static files should *not* be built into the MEDIA_ROOT (and accessed via the MEDIA_URL setting). Use STATIC_ROOT and STATIC_URL instead.

jezdez:
This makes me realize how fucked up the static files setup is for kuma. It completely ignore the fact that Django has introduces `STATIC_ROOT` as the location for static assets like js libraries contrary to `MEDIA_ROOT` only being there for user uploaded files. See https://docs.djangoproject.com/en/1.4/ref/templates/builtins/#static

jezdez:
In other words, the gulp build files (including the manifest file) should be written to and served from the `STATIC_ROOT` folder.

kuma/core/helpers.py
```diff
@@ -8,6 +9,7 @@
 from django.utils.encoding import smart_str
 from django.utils.http import urlencode as urlencode_util
 from django.utils.tzinfo import LocalTimezone
+from django.templatetags.static import static
```
jezdez: Sorry I missed that earlier, that must be `from django.contrib.staticfiles.templatetags.staticfiles import static`

openjck:
Updated

openjck:
> Nice, but we should do another move, move the static files to a "static" folder to separate it from uploads.

Which static files?

jezdez:
@openjck The ones in the "media" folder. Just move that folder to "static" and update the gulp file. To summarize Django's best practices: MEDIA_ROOT/MEDIA_URL are only used for uploads, STATIC_ROOT/STATIC_URL for static assets such as js/img/css. The idea is to separate the user uploaded content from the infrastructure files.

jezdez:
@openjck Ah, there are of course a few non-static files in "media": "media/uploads", "media/attachments", etc. So renaming may not do the trick, but actual forceful separation :)

openjck:
Definitely an improvement worth making to the codebase. It would be a lot easier to do once all JavaScript and CSS has been migrated to Gulp. How do we feel about postponing until then?

openjck:
> Seems like there is a release on http://cdnjs.cloudflare.com/ajax/libs/jquery-scrollTo/1.4.14/jquery.scrollTo.js, any reason we're using not that?

We could use cloudfare, but it only hosts versions newer than the version we're using. We should upgrade the plugin eventually, but the scope of this pull request is just to move requirements to Bower.

openjck:
> Let's use https://raw.githubusercontent.com/briancherne/jquery-hoverIntent/master/jquery.hoverIntent.js here to not depend on a private hoster

Same with this plugin. GitHub only hosts newer versions of the plugin, and the scope at the moment is to migrate requirements to Bower. Because the requirements are pulled down locally, if the host goes down we will have other options, including restoring the copied source code from our Git history.

groovecoder:
This is looking good to me. I especially like the `cache-bust` method to only change the asset names when the underlying assets change.

I was able to rebase pretty easily too.

How can I make my local dev vm serve up the **un-bundled, non-minified** assets?

jingo-minify does this with the `js` and `css` helpers. I.e., `{{ js('bundle-name') }}` and `{{ css('bundle-name') }}` automatically check when `settings.DEBUG == True` and render separate `<script>` and `<link>` elements for each asset listed in the bundle. When `settings.DEBUG == False` they render the single `<script>` and `<link>` elements for the bundled, minifed asset.

Maybe some similar helpers on top of the new `cache-bust` helper here?

gulpfile.js
```diff
@@ -0,0 +1,203 @@
...
+    gulp.watch(npmDependencies, ['install-and-shrinkwrap-npm-dependencies']);
+
+    // Bower
+    gulp.watch(jsDependencies, ['install-javascript-dependencies']);
+
+    // JavaScript
+    for(var bundleName in jsBundles) {
+        if(jsBundles.hasOwnProperty(bundleName)) {
+            var bundle = jsBundles[bundleName];
+            watchJSBundle(bundleName, bundle);
```
groovecoder: This didn't seem to work for me as I expected. Steps to reproduce:

1. `gulp watch`
2. Change `analytics.js` (I just added an `alert("gulp changed me!");` line to it)
3. Refresh home page

Expected results:
Should see the "gulp changed me!" alert box.

Actual results:
Nothing. :frowning: 

And only this showing on the console:
```
(env)vagrant@developer-local:~/src$ gulp watch                                 [55/1609]
[21:16:11] Using gulpfile ~/src/gulpfile.js
[21:16:11] Starting 'watch'...
[21:16:11] Finished 'watch' after 83 ms
```

It wasn't until I did `gulp` (to run `build-javascript`) that the changes took effect.

openjck: Thanks for explaining that. I can't seem to reproduce locally. How long was the wait between steps 2 and 3? This sounds like [bug 1054257](https://bugzilla.mozilla.org/show_bug.cgi?id=1054257).

There are a few ways around that: we could fix [bug 1054257](https://bugzilla.mozilla.org/show_bug.cgi?id=1054257), tell developers to run Gulp on the host machine, or tell developers to wait for a few moments after making file changes. The first option could become a rabbit hole, but one of the others might work in the short term. On the plus side, the delay is no worse than the delay in compiling stylesheets, which we've become used to.

openjck:
> How can I make my local dev vm serve up the un-bundled, non-minified assets?

Good question. I never found unbundled assets to be especially helpful, but I can see a case for them if someone really needs to use the debugger with the original symbols. This will take some thought. It will need to be something more expansive than changing `cache_bust`, since `cache_bust` takes the filename of a minified resource as input.

stephaniehobson:
I find unbundled CSS to be quite helpful and would be sad it it went away.

gulpfile.js
groovecoder: I ran `gulp watch` and then went and did some emails for about 20 minutes. :wink: Still no action. :frowning:

openjck: Weird, I can't reproduce. Are you sure `media/js/analytics.js` was edited, and not one of the files also named `analytics.js` from the Bower package? Was `gulp watch` run locally or in the VM?

groovecoder: Yup, `media/js/analytics. Ran `gulp watch` in the VM.

openjck: Do you have NFS disabled by any chance? I hear that under some conditions (I think disabled NFS is one of them) Vagrant doesn't pick up on file changes.

What happens if you run `gulp watch` on the host?

groovecoder:
My McDonald's idea:

1. Create new `js(bundle_name)` and `css(bundle_name)` helpers that replace the jingo-minify helpers
2. Each helper:
  1. `if settings.DEBUG == True:` render a `<script>` or `<link>` element for each file in the bundle
  2. `if settings.DEBUG == False:` render a single `<script>` or `<link>` element for the combined, minified file using `cache_bust` to create the filename.

Seem logical?

jezdez:
@groovecoder -1 on writing our own Python stuff to handle bundled, we should follow the best practices of the community and use django-pipeline. It's what bedrock has been moved to, too.

groovecoder:
:+1: using best practices python.

That would mean moving the bundle declarations out of `gulpfile.js` and back into `settings.py`.

openjck:
django-pipeline also does concatenation and minification. Is it worth the effort to get django-pipeline and Gulp to talk to each other if django-pipeline does the same job anyway?

groovecoder:
That's a big question. Let's start a thread with dev-mdn and dev-webdev about it?

openjck:
It would be easiest to do this either entirely with a Django plugin or entirely with Gulp. If we do this entirely with Gulp, and we need uncompressed assets locally, we'll need to run Gulp at deploy time to treat different environments differently.

jezdez:
@openjck Agreed, I don't think we should mix gulp and another tool that does asset compilation, concatenation, minification, etc – that was my point earlier really. I'm happy to use gulp, but we should use it completely and not compromise. Looking back at https://lincolnloop.com/blog/integrating-front-end-tools-your-django-project/ this is what they went with, staticfiles only for the management of Python side static files and gulp for any processing.

openjck:
https://groups.google.com/d/msg/mozilla.dev.mdn/xtZ6d1pjKQ0/3nvwoXjMka4J

PR #3044, opened 2015-02-03
fix bug 1124355 - use gravatar_url helper in profiles

groovecoder:


kuma/users/templates/users/profile.html
```diff
@@ -121,7 +121,7 @@ <h1 class="profile-title">
       </div>
       <div class="column-2 ">
         <figure>
-          <img src="{{ profile.gravatar }}" alt="{{ profile.user.username }}" width="200" height="200" class="profile-photo avatar" />
+          <img src="{{ gravatar_url(profile.user) }}" alt="{{ profile.user.username }}" width="200" height="200" class="profile-photo avatar" />
```
jezdez: `{{ profile.gravatar() }}` is enough as that wraps the helper, https://github.com/mozilla/kuma/blob/785870517ddaaa6bb8b148be88c462c62ceddb09/kuma/users/models.py#L176-L177

groovecoder: Yeah, that was the other way to fix. I.e., change `profile.gravatar` (bound method) to `profile.gravatar()` (method call). I went with this because "gravatar_url" was more explicit about what it's returning - i.e., a url. But I'm good with either way.

groovecoder:
Updated to use `profile.gravatar()`

PR #3046, opened 2015-02-03
Bug 1125850 - Use a Celery chord for rendering stale documents instead of process inline.

jezdez:


kuma/wiki/tasks.py
```diff
@@ -37,6 +37,12 @@ def render_document(pk, cache_control, base_url):
     return document.rendered_errors
 
 
+@task()
+def clear_render_stale_document_lock():
+    lock = MemcacheLock('render-stale-documents-lock', expires=60 * 60)
```
robhudson: This is set multiple times. Perhaps use a constant? Also, it doesn't appear that you need it here since you just release it.

kuma/wiki/tasks.py
```diff
@@ -37,6 +37,12 @@ def render_document(pk, cache_control, base_url):
...
+    lock.release()
+
+
 @task(throws=(StaleDocumentsRenderingInProgress,))
 def render_stale_documents(log=None):
     """Simple task wrapper for rendering stale documents"""
```
robhudson: I can't comment on the line b/c it's not close enough to a change, but is `if lock.acquired` the right conditional below? `lock.acquired` sounds like the successful path? Looking at the code it says there's already a lock so I guess it is just the wording that is awkward.

kuma/wiki/tasks.py
```diff
@@ -56,15 +62,17 @@ def render_stale_documents(log=None):
...
     if lock.acquire():
-        try:
-            for doc in stale_docs:
-                doc.render('no-cache', settings.SITE_URL)
-                log.info("Rendered stale %s" % doc)
-        finally:
-            lock.release()
+        stale_pks = stale_docs.values_list('pk', flat=True)
+        log.info('Scheduling rendering of documents: %s' %
+                 stale_pks)
```
robhudson: Looks like that can fit on a single line.

kuma/wiki/tasks.py
```diff
@@ -56,15 +62,17 @@ def render_stale_documents(log=None):
...
+    log.info('Found %s stale documents' % stale_docs_count)
     response = None
     if lock.acquire():
-        try:
-            for doc in stale_docs:
-                doc.render('no-cache', settings.SITE_URL)
-                log.info("Rendered stale %s" % doc)
-        finally:
-            lock.release()
+        stale_pks = stale_docs.values_list('pk', flat=True)
```
robhudson: How big can this get? Would you ever need to chunk it? Can it have multiple pks that might need a `.distinct()`?

robhudson:
Celery changes look good. This will parallelize the tasks so they happen faster, hopefully. :+1:

kuma/wiki/tasks.py
jezdez: Just trying to keep the implementation details centralized in the lock class

kuma/wiki/tasks.py
jezdez: could be pretty big, depends on how much work is being done

kuma/wiki/tasks.py
jezdez: Gonna clean it up a bit.

kuma/wiki/tasks.py
jezdez: That's alright, `lock.acquired` says that a different process has currently acquired a lock, so it'll raise an exception noting that state instead of continuing.

jezdez:
@robhudson I've now used a chord of a chunked task, can you double check if that makes sense?

kuma/wiki/tasks.py
jezdez: I've renamed it to `locked` and made it a method. So `lock.locked` returns either `True` or `False`

jezdez:
Gnaaaah now the tests fail again :(

jezdez:
@robhudson I wonder if we can't just not use chunks, I just want to distribute the load of the render_stale_documents task to all workers

kuma/wiki/tasks.py
```diff
@@ -56,15 +81,14 @@ def render_stale_documents(log=None):
...
-    if lock.acquire():
-        try:
-            for doc in stale_docs:
-                doc.render('no-cache', settings.SITE_URL)
-                log.info("Rendered stale %s" % doc)
-        finally:
-            lock.release()
+    if render_lock.acquire():
+        stale_pks = stale_docs.values_list('pk', flat=True)
+        render_tasks = render_document_chunk.chunks(stale_pks, 10)
```
robhudson: A task has a `chunks` method?

jezdez: Yeah, http://celery.readthedocs.org/en/latest/userguide/canvas.html#chunks

PR #3047, opened 2015-02-03
fix bug 1125648

EmiXLabs:
Azerbaijani Language for MDN
Bug 1125648
https://bugzilla.mozilla.org/show_bug.cgi?id=1125648
Have no idea about orders in languages.json
+ Documentation for kuma (http://kuma.readthedocs.org/en/latest/localization.html#adding-a-new-locale) should be changed, adding new locale to settings.py is not enough and will give error. It should be added to kuma/languages.json too

### DIFF
--- a/kuma/languages.json
+++ b/kuma/languages.json
@@ -1,4 +1,9 @@
 {
+    "az": {
+        "iso639_1": "az",
+        "english": "Azerbaijani",
+        "native": "Az\u0259rbaycanca"
+    },
     "pa-IN": {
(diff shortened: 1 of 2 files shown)


groovecoder:
Thanks @EmiXLabs. I'm currently working thru the way we update `languages.json` in [bug 1124423](https://bugzilla.mozilla.org/show_bug.cgi?id=1124423). I'll update the docs when I figure out the best way to update `languages.json`

PR #3049, opened 2015-02-04
Fixes bug 931377: Adding button to toggle <pre> wrapped code samples

jamonation:
I chatted a bit with @jezdez about this - JavaScript seems to be the lowest barrier to implementing this show/hide pre code sample functionality. Here's a basic patch to do that.

It only works in English, and the button appears above each and every pre element, so it may not be quite ready to merge.

My Dev Kuma doesn't have a large enough sample of pages in it to be sure this behaves exactly as expected so some testing/review would be welcome.

### DIFF
--- a/media/redesign/js/main.js
+++ b/media/redesign/js/main.js
@@ -212,6 +212,19 @@
         mdn.Notifier.discover();
     })();
 
+
+    /*
+        Inject Toggle Code button before <pre> elements
+        TODO: Make it translatable into any detected browser language
+    */
+    (function() {
+        $.each($('pre'), function() {
+            var $pre = $(this);
+            $pre.before('<button class="button toggle-pre">Toggle code <i class="icon-caret-down" aria-hidden="true"></i></button>');
+        });
+    })();
+
+
     /*
         Tabzilla
     */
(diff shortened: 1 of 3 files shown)


stephaniehobson:
Hi @jamonation I talked to the writing team to get a bit more context for this bug and added the extra info to https://bugzilla.mozilla.org/show_bug.cgi?id=931377#c3

stephaniehobson:
Basically, the behaviour needs to be triggered on very specific code samples and only at the behest of the writer. In light of the additional context, I think you need to re-think your solution.

Let me know if you have any questions or if I can help provide any extra context, it's a big complicated system sometimes :)

media/redesign/js/main.js
```diff
@@ -212,6 +212,19 @@
...
+        Inject Toggle Code button before <pre> elements
+        TODO: Make it translatable into any detected browser language
+    */
+    (function() {
+        $.each($('pre'), function() {
+            var $pre = $(this);
+            $pre.before('<button class="button toggle-pre">Toggle code <i class="icon-caret-down" aria-hidden="true"></i></button>');
+        });
+    })();
+
```
darkwing: Can this be reduced to...
```
$('pre').before('...')
```

jamonation:
Going to close this and come up with an alternate implementation that takes the extra feedback from @stephaniehobson into account.

PR #3050, opened 2015-02-04
Fixes bug 1061626: Adding explicit link to tagging guidelines to new_document template

jamonation:
I don't want to step on any toes, but it appears that the contributor to https://bugzilla.mozilla.org/show_bug.cgi?id=1061626 has disappeared. Here's a patch to close the bug per @stephaniehobson's suggestion.

### DIFF
--- a/kuma/wiki/templates/wiki/new_document.html
+++ b/kuma/wiki/templates/wiki/new_document.html
@@ -65,17 +65,19 @@
         </div>
       {% endif %}
 
+      <section id="page-tags" class="page-tags wiki-block">
+        <h3><i aria-hidden="true" class="icon-tags"></i>{{ _('Tags') }} <a href="{{ wiki_url('Project:MDN/Contributing/Editor_guide/Editing#The_tags_box') }}"><i aria-hidden="true" title="{{ _('Learn how to tag an article') }}" class="icon-question-sign editor-help-icon"></i></a></h3>
+        <p>
+        {% trans url=wiki_url('Project:MDN/Contributing/Tagging_standards') %}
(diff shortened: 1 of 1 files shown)


kuma/wiki/templates/wiki/new_document.html
```diff
@@ -65,6 +65,16 @@
...
+        {% trans url=wiki_url('Project:MDN/Contributing/Tagging_standards') %}
+        Categorize the article. It will make the article findable under the right filters in the search engine. <a target="_blank" href="{{ url }}">Read the Tagging Guide</a>.
+        {% endtrans %}
+        </p>
+        <input id="tagit_tags" type="text" name="tags" value="{% for tag in tags %}{{ tag.name }},{% endfor %}" maxlength="255" class="tags" />
+      </section>
+
       {% include 'wiki/includes/revision_comment.html' %}
 
       {% include 'wiki/includes/review_tags.html' %}
```
stephaniehobson: We already have the tags form on this page, the code you added should replace the review_tags include otherwise we prompt for tags twice on the page ;)

jamonation:
I think I've found all the duplicate code. I put the tags section in the correct location in the template to match the order/layout of the edit view.

stephaniehobson:
Changes look great :) Thanks for the speedy changes.

One last thing, now that it's read to go could you please "squash" your 3 commits into one? This make it easier for us to manage the changes if we have to cherry pick or revert something, we can be sure we're interacting with a commit that doesn't rely on any others. Thanks!

jamonation:
Makes sense, I'll make a point of doing that next time I open a PR with a whack of small incremental changes.

stephaniehobson:
Looking good. Thanks for the contribution!

PR #3051, opened 2015-02-04
Fixes bug 1108474: Make compare revisions button on history page sticky

jamonation:
This is to address the feature requested in https://bugzilla.mozilla.org/show_bug.cgi?id=1108474 - make the compare revisions button sticky when scrolling the page.

### DIFF
--- a/media/redesign/js/wiki.js
+++ b/media/redesign/js/wiki.js
@@ -578,6 +578,18 @@
         }
     });
 
+    /*
+      Make Compare selected revisions button sticky when scrolling history page
+    */
+    (function() {
+        var button = $('.revision-list-controls .link-btn');
+        var revisionButtonOffset = $(button).offset().top;
+        $(window).on('scroll', function() {
+            var $compareButton = $(button);
+            var scroll = $(this).scrollTop();
+            $compareButton.toggleClass('fixed', scroll >= revisionButtonOffset);
+        });
+    })();
 
     function debounce(func, wait, immediate) {
         var timeout;
--- a/media/redesign/stylus/components/wiki/revision-list.styl
+++ b/media/redesign/stylus/components/wiki/revision-list.styl
@@ -97,3 +97,7 @@ html[dir=rtl] .revision-list-controls:before {
     width: 460px;
 }
 
+/* make the Compare selected revisions button align to layout */
+.revision-list-controls .fixed {
+    bidi-style(left, 24px, right, auto);
+}
--- a/kuma/wiki/templates/wiki/document_revisions.html
+++ b/kuma/wiki/templates/wiki/document_revisions.html
@@ -73,11 +73,6 @@
             </li>
             {% if loop.last %}</ul>{% endif %}
           {% endfor %}
-          {% if revision_count > 1 %}
-            <div class="revision-list-controls revision-list-controls-bottom">
-              <input type="submit" class="link-btn" value="{% if document.parent %}{{ _('Compare Selected Translations') }}{% else %}{{ _('Compare Selected Revisions') }}{% endif %}" />
-            </div>
-          {% endif %}
         </form>
       {% else %}
         {{ _('This document has no revisions.') }}


media/redesign/stylus/components/wiki/revision-list.styl
```diff
@@ -97,3 +97,7 @@ html[dir=rtl] .revision-list-controls:before {
     width: 460px;
 }
 
+/* make the Compare selected revisions button align to layout */
+.revision-list-controls .fixed {
+    left: 24px;
+}
```
stephaniehobson: We need a new line at the end of files or GIT complains :/

media/redesign/stylus/components/wiki/revision-list.styl
```diff
@@ -97,3 +97,7 @@ html[dir=rtl] .revision-list-controls:before {
     width: 460px;
 }
 
+/* make the Compare selected revisions button align to layout */
+.revision-list-controls .fixed {
+    left: 24px;
```
stephaniehobson: Needs to be RTL text friendly (example:https://developer.mozilla.org/he/docs/%D7%9E%D7%95%D7%A9%D7%92%D7%99%D7%9D$history)

We have a stylus mixin to help accomplish this:
https://github.com/mozilla/kuma/blob/d6692dff1d51c4529c4754874608041213250eca/media/redesign/stylus/includes/mixins.styl#L311-L327

stephaniehobson:
@darkwing Can you review the javascript please?

media/redesign/stylus/components/wiki/revision-list.styl
jamonation: Added :)

media/redesign/stylus/components/wiki/revision-list.styl
jamonation: I think I've added this correctly in my recent commit. It's a bit hard to test though. Mind taking another look now?

media/redesign/stylus/components/wiki/revision-list.styl
```diff
@@ -97,3 +97,7 @@ html[dir=rtl] .revision-list-controls:before {
     width: 460px;
 }
 
+/* make the Compare selected revisions button align to layout */
+.revision-list-controls .fixed {
+    bidi-style(left, 24px, right, 24px);
```
stephaniehobson: The 4th parameter needs to be the inverse value for left, in this case "auto".

The way the bidi mixin works is that the first two values are compiled into the regular style sheet:
```
.revision-list-controls .fixed {
   left: 24px;
}
```
And then an over-ride is made for the RTL text, over-riding the value of the first property with the value provided in the 4th parameter.
```
html[dir='rtl'] .revision-list-controls .fixed {
   left: auto;
   right: 24px;
}
```

Thanks for the quick turn around on my comments!

media/redesign/stylus/components/wiki/revision-list.styl
stephaniehobson: I had to jump through so many hoops to get the internationalization stuff set up on my testing machine >_<  I'm happy to test the RTL :)

media/redesign/stylus/components/wiki/revision-list.styl
jamonation: That makes sense now that you've explained it, hopefully I can remember for future commits :)

media/redesign/stylus/components/wiki/revision-list.styl
jamonation: I realized that I can just edit the DOM in Inspector and set dir ="rtl" - but don't let that take away from you enjoying testing on your local system heh.

media/redesign/js/wiki.js
```diff
@@ -578,6 +578,21 @@
...
+    (function() {
+        var $revisionButtonOffset = $('.revision-list-controls .link-btn').offset().top;
+        $(window).scroll( function() {
+            var $compareButton = $('.revision-list-controls .link-btn');
+            var $scroll = $(window).scrollTop();
+            if ($scroll >= $revisionButtonOffset) {
+                $compareButton.addClass('fixed');
+            } else {
+                $compareButton.removeClass('fixed');
+            }
```
robhudson: Alternatively: `$compareButton.toggleClass('fixed', $scroll >= $reviesionButtonOffset);`

jamonation: That makes sense @robhudson, I'll try to opt for the more compact toggle/toggleClass methods where it makes sense in the future.

media/redesign/js/wiki.js
```diff
@@ -578,6 +578,17 @@
         }
     });
 
+    /*
+      Make Compare selected revisions button sticky when scrolling history page
+    */
+    (function() {
+        var $revisionButtonOffset = $('.revision-list-controls .link-btn').offset().top;
```
darkwing: You use `$('.revision-list-controls .link-btn')` twice;  I would create a reference to that button so you aren't constantly retrieving it:

```
var button = $('.revision-list-controls .link-btn');
```

media/redesign/js/wiki.js
```diff
@@ -578,6 +578,17 @@
...
+        $(window).scroll( function() {
```
darkwing:
```
$(window).on('scroll', function() {
```

media/redesign/js/wiki.js
```diff
@@ -578,6 +578,17 @@
...
+            var $compareButton = $('.revision-list-controls .link-btn');
+            var $scroll = $(window).scrollTop();
```
darkwing:
```
var scroll = $(this).scrollTop();
```

Unless the var returns a jQuery object, don't use $ in the var name

jamonation:
@darkwing @stephaniehobson I think that I've covered all the feedback from both of you. Mind one last review before calling this one fully baked?

stephaniehobson:
Squashed and everything. Thanks @jamonation :) R!

PR #3054, opened 2015-02-04
Pin ES puppet module

robhudson:


openjck:
Looks good. Just waiting for Travis.

groovecoder:
Little late, but this commit message should have started with `bug 957802` - the bug we use as the catch-all for maintaining the dev environment.

openjck:
Ah, I missed that. I added a comment to [bug 957802](https://bugzilla.mozilla.org/show_bug.cgi?id=957802) so we can at least have a paper trail in one direction.

PR #3056, opened 2015-02-05
Fix bug 1084503: extend UserAdmin to allow sorting in admin/auth/user view

jamonation:
I *think* that this does what is expected in the main bug description. It adds a field Date Joined to the admin display of users, along with a Filter to narrow the result down by date. The details are in https://bugzilla.mozilla.org/show_bug.cgi?id=1084503

### DIFF
--- a/kuma/users/admin.py
+++ b/kuma/users/admin.py
@@ -6,6 +6,20 @@ from taggit.forms import TagWidget
 from kuma.core.managers import NamespacedTaggableManager
 from .models import UserBan, UserProfile
 
+from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.models import User
+
+
+class ExtendedUserAdmin(UserAdmin):
+    # extend the admin view of users to show date_joined field; add a filter on the field too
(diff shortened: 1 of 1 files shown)


kuma/users/admin.py
```diff
@@ -6,6 +6,19 @@
...
 
+from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.models import User
+
+
+class ExtendedUserAdmin(UserAdmin):
+    # extend the admin view of users to show date_joined field; add a filter on the field too
+    list_display = ('username', 'email', 'first_name', 'last_name', 'date_joined', 'is_staff', 'is_active')
+    list_filter = ('is_staff', 'is_superuser', 'is_active', 'date_joined',)
+
```
jezdez: You may want to add a `ordering` attribute as well to define to auto sort it by `date_joined` instead of `username` (which is the default)

jamonation:
Thanks for the feedback @jezdez, I've added the -date_joined ordering so that newest users show up first.

PR #3060, opened 2015-02-05
Bug 1120526 - Clean up facets a little and use post_filter

robhudson:


kuma/search/filters.py
```diff
@@ -162,50 +161,36 @@ def filter_queryset(self, request, queryset, view):
...
+                    active_filters.append(F(filter_operator, tag_filters))
                 else:
                     active_filters.append(F('term', tags=filter_tags[0]))
 
             if len(filter_tags) > 1:
-                facet_params = {
-                    'or': {
-                        'filters': [{'term': {'tags': tag}}
-                                    for tag in filter_tags],
-                        '_cache': True,
```
jezdez: IIRC that was important to keep speed high

kuma/search/views.py
```diff
@@ -43,8 +42,6 @@ class SearchView(ListAPIView):
     def initial(self, request, *args, **kwargs):
         super(SearchView, self).initial(request, *args, **kwargs)
         self.current_page = self.request.QUERY_PARAMS.get(self.page_kwarg, 1)
-        self.drilldown_faceting = flag_is_active(request,
-                                                 'search_drilldown_faceting')
```
jezdez: :wink:

kuma/search/filters.py
robhudson: When I read the docs it was the default. They offer `_cache` if you want to disable it.

robhudson: For reference: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-term-filter.html#_caching_18

jezdez: ok, that was different I'm sure

PR #3061, opened 2015-02-06
Fixes bug 1130378 - Faster reindexing

robhudson:
This updates reindexing to parallelize the indexing tasks. It also updates the tests a little since a few failed when I did this.

I'm not sure if there's a pre-existing bug this should go to?

### DIFF
--- a/kuma/search/management/commands/reindex.py
+++ b/kuma/search/management/commands/reindex.py
@@ -23,4 +23,6 @@ class Command(BaseCommand):
         if not 1 <= percent <= 100:
             raise CommandError('percent should be between 1 and 100')
 
-        WikiDocumentType.reindex_all(options['chunk_size'], percent=percent)
+        message = WikiDocumentType.reindex_all(options['chunk_size'],
+                                               percent=percent)
+        self.stdout.write(unicode(message) + '\n')
--- a/kuma/search/models.py
+++ b/kuma/search/models.py
@@ -107,8 +107,7 @@ class Index(models.Model):
         return '%s-%s' % (settings.ES_INDEX_PREFIX, self.name)
 
     def populate(self):
-        from .tasks import populate_index
-        populate_index.delay(self.pk)
+        return WikiDocumentType.reindex_all(index=self, chunk_size=500)
 
     def record_outdated(self, instance):
         if self.successor:
@@ -130,6 +129,8 @@ class Index(models.Model):
         self.outdated_objects.all().delete()
         self.promoted = True
         self.save()
+        # Allow only a single index to be promoted.
+        Index.objects.exclude(pk=self.pk).update(promoted=False)
 
     def demote(self):
         self.promoted = False
--- a/kuma/wiki/search.py
+++ b/kuma/wiki/search.py
@@ -3,22 +3,21 @@ from __future__ import division
 
 import logging
 import operator
-import time
+from math import ceil
 
 from django.conf import settings
 from django.db.models import Q
 from django.utils.html import strip_tags
-from django.db import reset_queries
+from django.utils.translation import ugettext_lazy as _
 
+from celery import chain
 from elasticsearch.helpers import bulk
-from elasticsearch.exceptions import NotFoundError
 from elasticsearch_dsl import document, field
 from elasticsearch_dsl.connections import connections
 from elasticsearch_dsl.mapping import Mapping
 from elasticsearch_dsl.search import Search
 
-from kuma.core.utils import chunked
-from kuma.search.utils import format_time
+from kuma.core.utils import chord_flow, chunked
 
 
 log = logging.getLogger('kuma.wiki.search')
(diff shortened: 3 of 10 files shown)


kuma/wiki/search.py
```diff
@@ -288,86 +287,39 @@ def reindex_all(cls, chunk_size=1000, index=None, percent=100):
...
+        pre_task = prepare_index.si(index.pk)
+        post_task = finalize_index.si(index.pk)
+
+        if not total:
+            # If there's no data we still create the index and finalize it.
+            chain(pre_task, post_task).apply_async()
+        else:
+            index_tasks = [index_documents.si(chunk, index.pk)
+                           for chunk in chunked(indexable, chunk_size)]
+            if settings.CELERY_ALWAYS_EAGER:
```
jezdez: I think it'd be neat to refactor that to a general helper function, e.g.:

```python
def flow(pre_task, main_tasks, post_task):
    if settings.CELERY_ALWAYS_EAGER:
        # Eager mode and chords don't get along. So we serialize
        # the tasks as a workaround.
        main_tasks.insert(0, pre_task)
        main_tasks.append(post_task)
        task = chain(*main_tasks)
    else:
        task = chain(pre_task, chord(header=main_tasks, body=post_task))
    return task
```

The use it like this:
```python
index_tasks = [index_documents.si(chunk, index.pk)
               for chunk in chunked(indexable, chunk_size)]

flow(pre_task, index_tasks, post_task).apply_async()
```

jezdez:
https://bugzilla.mozilla.org/show_bug.cgi?id=1130378

kuma/wiki/search.py
robhudson: Sure, I can try that. We do have 2 places in the code that could use it.

kuma/search/models.py
```diff
@@ -128,6 +127,8 @@ def promote(self):
             index_objects.delay(mapping_type, [instance.id])
             rescheduled.append(label)
         self.outdated_objects.all().delete()
+        # Allow only a single index to be promoted.
+        Index.objects.all().update(promoted=False)
```
robhudson: @jezdez Any problem here? I think it helps lead us to using index aliases but realize it may go against your plan. There's more simplifying we can do if this is ok, like `get_current()` doesn't need to sort by `-created` any longer, just get the promoted index.

kuma/wiki/search.py
robhudson: I added the helper.

kuma/search/models.py
jezdez: Ah, good point, this is a change for a race condition since `get_current()` is used often

kuma/search/management/commands/reindex.py
```diff
@@ -23,4 +23,6 @@ def handle(self, *args, **options):
         if not 1 <= percent <= 100:
             raise CommandError('percent should be between 1 and 100')
 
-        WikiDocumentType.reindex_all(options['chunk_size'], percent=percent)
+        message = WikiDocumentType.reindex_all(options['chunk_size'],
+                                               percent=percent)
+        print unicode(message)
```
jezdez: Can you use something else than print here?

kuma/wiki/search.py
```diff
@@ -288,86 +287,31 @@ def reindex_all(cls, chunk_size=1000, index=None, percent=100):
...
-
-        # Disable automatic refreshing.
-        temporary_settings = {
-            'index': {
-                'refresh_interval': '-1',
-                'number_of_replicas': '0',
-            }
-        }
+        total = len(indexable)
+        total_chunks = int(ceil(total / float(chunk_size)))
```
jezdez: again with the future division

kuma/wiki/search.py
```diff
@@ -288,86 +287,31 @@ def reindex_all(cls, chunk_size=1000, index=None, percent=100):
...
+        pre_task = prepare_index.si(index.pk)
+        post_task = finalize_index.si(index.pk)
+
+        if not total:
+            # If there's no data we still create the index and finalize it.
+            chain(pre_task, post_task).apply_async()
+        else:
+            index_tasks = [index_documents.si(chunk, index.pk)
+                           for chunk in chunked(indexable, chunk_size)]
+            chord_flow(pre_task, index_tasks, post_task).apply_async()
```
jezdez: yayayaya!

kuma/search/management/commands/reindex.py
robhudson: `sys.stdout.write` perhaps? I think I've seen that used in django management commands. Any reason why to not use print?

jezdez: I think the management commands have a `self.stdout` IIRC

kuma/wiki/search.py
robhudson: This file already has future division in it. So I'll just remove the `float` wrapper there.

kuma/search/models.py
robhudson: Currently it still does the filter and orders by `-created`. I could swap the order so it updates the `self` promoted value, then removes promoted from all excluding `self`?

kuma/wiki/search.py
jezdez: cool

kuma/search/models.py
jezdez: wfm

PR #3062, opened 2015-02-06
Refactor settings.

jezdez:
This requires some extended handholding when deploying and help from webops to put the prod and stage specific settings overrides into the appropriate places (probably just ``.env`` files)

The bugs for this are:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1119433
- https://bugzilla.mozilla.org/show_bug.cgi?id=1147427
- https://bugzilla.mozilla.org/show_bug.cgi?id=1146915
- https://bugzilla.mozilla.org/show_bug.cgi?id=1146890


kuma/settings/common.py
```diff
@@ -530,14 +519,12 @@ def lazy_language_deki_map():
     'badger',
 )
 
-TEST_RUNNER = 'test_utils.runner.RadicalTestSuiteRunner'
+TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
```
robhudson: Saving pytest for later? :grin:

jezdez: yeah, till after the Django upgrade

kuma/settings/local.py
```diff
@@ -0,0 +1,113 @@
...
+DOMAIN = 'developer-local.allizom.org'
+
+KUMASCRIPT_URL_TEMPLATE = 'http://localhost:9080/docs/{path}'
+
+ATTACHMENT_HOST = 'mdn-local.mozillademos.org'
+
+# Elasticsearch related settings.
+ES_URLS = ['127.0.0.1:9200']
+ES_DEFAULT_NUM_REPLICAS = 0
+ES_DEFAULT_NUM_SHARDS = 5
```
robhudson: I'd make shards=1 for local development.

groovecoder:
Is this PR worth keeping, or should we start a new one just for the settings refactor?

jezdez:
@groovecoder Yeah, I'll rebase and work on a slightly different implementation.

docs/installation.rst
```diff
@@ -95,37 +95,22 @@ projects simultaneously.
...
-Note the two settings ``TEST_CHARSET`` and ``TEST_COLLATION``. Without these,
-the test suite will use MySQL's (moronic) defaults when creating the test
-database (see below) and lots of tests will fail. Hundreds.
+At a minimum, you will need to define a database connection. The default
+database configuration is::
+
+    DATABASE_URL = 'mysql://kuma:kuma@localhost:3306/kuma'
+
+In other words, it uses MySQL default, the username and password of 'kuma'
+when trying to access the database 'kuma'. We automatically MySQL's InnoDB
```
groovecoder: grammar nit: "We automatically MySQL's ..." ?

docs/installation.rst
```diff
@@ -198,10 +174,10 @@ To start the dev server, run ``./manage.py runserver``, then open up
 ``http://localhost:8000``. If everything's working, you should see
 the MDN home page!
 
-You might need to first set ``LC_CTYPE`` if you're on Mac OS X until
-`bug 754728 <https://bugzilla.mozilla.org/show_bug.cgi?id=754728>`_ is fixed::
+You might need to first set ``LC_CTYPE`` in your ``.env`` file if you're on
+Mac OS X until `bug 754728 <https://bugzil.la/754728>`_ is fixed::
```
groovecoder: https://bugzil.la/754728 is marked as fixed. Can we drop this section?

jezdez: I don't know since the ticket doesn't really explain why this is needed in the first place. The commit referenced in the ticket (5f745996c009f658fda9ab359d6e0d7c4e56fd2e) doesn't really explain it either. I don't see a reason to not keep that info.

groovecoder: Makes sense. Looks like the bug is "fixed" by including the docs. :neutral_face:

kuma/settings/local.py
```diff
@@ -0,0 +1,73 @@
...
+    'devserver.modules.profile.ProfileSummaryModule',
+    'devserver.modules.profile.MemoryUseModule',
+    'devserver.modules.cache.CacheSummaryModule',
+
+    # Modules not enabled by default
+    # 'devserver.modules.ajax.AjaxDumpModule',
+    # 'devserver.modules.profile.LineProfilerModule',
+)
+
+LOG_LEVEL = logging.ERROR
```
robhudson: :heart:  Thanks for changing this for local configs.

kuma/settings/prod.py
```diff
@@ -0,0 +1,25 @@
+from .common import *  # noqa
+
+ATTACHMENT_HOST = 'mdn.mozillademos.org'
+
+DATABASES['default']['OPTIONS'].update({
+    'charset': 'utf8',
+    'use_unicode': True,
+})
```
robhudson: Why isn't this included in the default?

kuma/users/tests/__init__.py
```diff
@@ -27,8 +27,7 @@ def user(save=False, **kwargs):
     password = kwargs.pop('password', 'password')
     user = User(**kwargs)
     user.set_password(password)
-    if save:
-        user.save()
+    user.save()
```
robhudson: Why this change? Not saving seems like a good way to avoid a db write if you don't need it. But if you want to always save we should remove the unused `safe=False` variable as well?

kuma/wiki/tests/test_models.py
```diff
@@ -97,13 +98,13 @@ def test_json_data(self):
         # Check fields of translated doc
         ok_('translations' in data)
         eq_(de_doc.locale, data['translations'][0]['locale'])
-        result_l10n_tags = sorted([str(x) for x
-                           in data['translations'][0]['localization_tags']])
+        result_l10n_tags = sorted([str(x)
+                                  for x in data['translations'][0]['localization_tags']])
```
robhudson: 1 more space for the indent

puppet/files/etc/apache2/conf.d/mozilla-kuma-apache.conf
```diff
@@ -14,11 +14,7 @@ Listen 8080
     SetEnv HTTPS on
     RequestHeader set X-Forwarded-Proto "https"
 
-    # Covering all pre-bug-702988 past-known server aliases
     ServerName developer-local.allizom.org
-    ServerAlias developer-kumadev.mozilla.org
-    ServerAlias developer-mdndev.mozilla.org
-    ServerAlias developer-dev.mozilla.org
     ServerAlias mdn-local.mozillademos.org
```
robhudson: Any reason we still need this one? I've always wondered why there are two in the setup docs.

robhudson:
I added a few comments but in general this is going to be a great change. r+wc.

docs/installation.rst
jezdez: :+1:

kuma/settings/local.py
jezdez: :heart_eyes_cat:

kuma/settings/prod.py
jezdez: That's some oddity between the various settings, the prod settings have it, but the local ones don't. I'm afraid of upgrading the charset to utf8 and unicode locally to not break the database somehow.

kuma/users/tests/__init__.py
jezdez: Not sure, I think some tests didn't pass without it? I'll try again without that change.

kuma/wiki/tests/test_models.py
jezdez: :eyeglasses:

puppet/files/etc/apache2/conf.d/mozilla-kuma-apache.conf
jezdez: `mdn-local.mozillademos.org` is used for serving demos locally under a different domain than the main one `developer-local.allizom.org`, similar to `mdn.mozillademos.org` in production.

docs/installation.rst
jezdez: Ok, I looked into this a bit more, this originates here: http://babel.edgewall.org/ticket/200 and was fixed in Babel trunk here: http://babel.edgewall.org/changeset/586

And we're indeed still vulnerable to that problem: https://github.com/mozilla/kuma/blob/827fbdfc77773021b8f2fdcd452d73d91f913e6e/vendor/packages/Babel/babel/core.py#L642

So as long as we use a stone age old version of Babel we're stuck on this section.

jezdez:
@robhudson Any chance for another pass?

celery.py
```diff
@@ -1,46 +0,0 @@
...
-
-from django.conf import settings
-from celery import Celery
-
-app = Celery('kuma')
-
-# Using a string here means the worker will not have to
-# pickle the object when using Windows.
-app.config_from_object('django.conf:settings')
-app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
```
robhudson: You don't need the celery app any longer? I was setting up something similar on zamboni recently b/c celery 3.1.x had some changes.

jezdez: Nah, we're using celery via the management commands instead of as a standalone Python script. That means we don't need that whole celery.py at all. The celery config is all applied via django-celery's Celery loader.

robhudson: I thought django-celery was no longer needed b/c celery now has the django bits in it?

robhudson:
I'm not sure what changed but I did a quick skim and it looked good to me.

celery.py
jezdez: Yes and no, it's not a requirement since the ability to use it with Django is now built-in. But if you want to use the admin integration and easier setup (via the management command "celery") you still need it. In other words, when they "added" Django support in 3.x again, it was only part of the full feature set of django-celery that got ported over.

groovecoder:
So, how's this looking for a potential merge & push tomorrow during "WebOps Sprint" time? (Along with https://bugzilla.mozilla.org/show_bug.cgi?id=1147427)

jezdez:
@groovecoder I think it's looking good, I'm currently discussing the details with Celia.

groovecoder:
I'm going to close this. We'll keep the branch up for any code we can potentially re-use in a post-update settings refactor.

PR #3064, opened 2015-02-06
Fix Bug 1130026: Code styling in .warning messages

stephaniehobson:
- Beefed up comments in syntax/base.styl
- Added an explicit text-color declaration to the code-block mixin (it's good practice to declare text and bgcolor together)
- Switched to variables for a few things
- Moved margin-bottom declaration off code and moved it to just pre (as code sometimes appears as an inline element)

A page to help with testing: https://developer.mozilla.org/en-US/docs/User:stephaniehobson:code


openjck:
Thanks for always writing such detailed instructions! This should be a breeze to test.

openjck:
Looks great! :fireworks:

PR #3065, opened 2015-02-06
fix bug 1119532 - add honeypot field to account signup form

groovecoder:


kuma/users/views.py
```diff
@@ -418,5 +419,11 @@ def get_context_data(self, **kwargs):
         })
         return context
 
+    def dispatch(self, request, *args, **kwargs):
+        ret = verify_honeypot_value(request, None)
+        if type(ret) is HttpResponseBadRequest:
```
jezdez: `isinstance` may be better

groovecoder:
Updated to use `isinstance` instead of `type`

kuma/users/views.py
```diff
@@ -418,5 +419,11 @@ def get_context_data(self, **kwargs):
         })
         return context
 
+    def dispatch(self, request, *args, **kwargs):
+        ret = verify_honeypot_value(request, None)
```
jezdez: `s/ret/response/g`

jezdez:
LGTM otherwise

kuma/users/views.py
groovecoder: I had `response` in there at first, but then I saw that `verify_honeypot_value` can also return `None`, so I changed it. I.e., the returned value may not be a response, so I just named it `ret`. I could go either way on it though. I'll leave it up to you.

jezdez: a `response` can be None, that's alright

groovecoder:
Updated to `response` instead of `ret`. Just need Travis to pass ...

jezdez:
Yay!

PR #3067, opened 2015-02-10
Fix bug 1104119 - Cache refill instead of invalidation for document zone remaps.

jezdez:
This resets the cached document zone remaps when saving documents instead just deleting them when saving the zone.

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -24,7 +24,7 @@ from kuma.search.models import Index
 from .events import context_dict
 from .exceptions import PageMoveError, StaleDocumentsRenderingInProgress
 from .helpers import absolutify
-from .models import Document, Revision, RevisionIP
+from .models import Document, Revision, RevisionIP, DocumentZone
 from .search import WikiDocumentType
 from .signals import render_done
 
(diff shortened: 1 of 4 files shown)


kuma/wiki/tasks.py
```diff
@@ -112,11 +112,7 @@ def build_json_data_for_document_task(pk, stale):
 
 @receiver(render_done)
 def build_json_data_handler(sender, instance, **kwargs):
-    try:
-        build_json_data_for_document_task.delay(instance.pk, stale=False)
-    except:
-        logging.error('JSON metadata build task failed',
-                      exc_info=True)
+    build_json_data_for_document_task.delay(instance.pk, stale=False)
```
groovecoder: Is this safe? Or was it a stray commit from other logging work?

jezdez: well, this was never safe but actually hid a potential scheduling error behind a blank except. if there is a reason why scheduling the task doesn't work we want to know via an exception handling, not a log entry

groovecoder: Makes sense.

groovecoder:
Still spot-checking locally ...

The bug mentions increasing the cache timeout from 60s to 1 day? As I read our current `CACHES` `memcache` setting, it will cache for 1 hour? Is that enough or should we bump it all the way up to 1 day while we're in here?

jezdez:
Hmm, good point, since we're actively resetting the cache now this should be safe to be increased indeed. Fixing..

jezdez:
For the record, I only increased the cache timeout for this specific cache setting, not the general memcache backend.

jezdez:
argh argh argh, heisenci

groovecoder:
Spot-check worked. I edited the "Add-ons" zone page and saw this in the celery log:
```
[2015-02-10 15:03:28,534: INFO/MainProcess] Task kuma.wiki.tasks.invalidate_zone_cache[60a0ee14-7e6a-43a6-8c7a-ca98e52
ac11d] succeeded in 0.154088480998s: None
```

PR #3069, opened 2015-02-10
Bug 1125850 - Use lxml to render sitemap XML.

jezdez:
That also reduces the risk of memory leakage when building all sitemaps.

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -3,7 +3,6 @@ from __future__ import with_statement
 import logging
 import os
 from datetime import datetime
-from xml.dom.minidom import parseString
 
 from django.conf import settings
 from django.contrib.auth.models import User
@@ -16,6 +15,7 @@ from django.utils.encoding import smart_str
 
 from celery import task
 from constance import config
+from lxml import etree
 
 from kuma.core.cache import memcache
 from kuma.core.utils import chord_flow, chunked, MemcacheLock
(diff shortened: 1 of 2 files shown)


kuma/wiki/tasks.py
```diff
@@ -265,13 +265,17 @@ class WikiSitemap(GenericSitemap):
...
+SITEMAP_ELEMENT = u'<sitemap><loc>%s</loc><lastmod>%s</lastmod></sitemap>'
+
+
 @task
 def build_sitemaps():
-    sitemap_element = '<sitemap><loc>%s</loc><lastmod>%s</lastmod></sitemap>'
-    sitemap_index = ('<sitemapindex xmlns="http://www.sitemaps.org/'
-                     'schemas/sitemap/0.9">')
+    sitemap_parts = [(u'<sitemapindex xmlns="http://www.sitemaps.org/'
+                      u'schemas/sitemap/0.9">')]
```
robhudson: looks like you don't need the parents now that you have the list brackets

kuma/wiki/tasks.py
```diff
@@ -281,25 +285,29 @@ def build_sitemaps():
...
             if not os.path.exists(directory):
                 os.makedirs(directory)
+
             with open(os.path.join(directory, 'sitemap.xml'), 'w') as f:
-                f.write(xml)
+                f.write(smart_str(render_to_string('wiki/sitemap.xml',
+                                  {'urls': WikiSitemap(info).get_urls(page=1)})))
+
+            # Force the gc to cleanup
+            del queryset
```
robhudson: This only happens if count > 0. Should it be outside that loop?

kuma/wiki/tasks.py
```diff
@@ -281,25 +285,29 @@ def build_sitemaps():
...
+            del info
 
             sitemap_url = absolutify('/sitemaps/%s/sitemap.xml' % locale)
-            sitemap_index = (sitemap_index + sitemap_element %
-                             (sitemap_url, timestamp))
+            sitemap_parts.append(SITEMAP_ELEMENT % (sitemap_url, timestamp))
 
-    sitemap_index = sitemap_index + "</sitemapindex>"
+    sitemap_parts.append(u"</sitemapindex>")
```
robhudson: everywhere else you've used single quotes

robhudson:
r+wc.

kuma/wiki/tasks.py
jezdez: god damnit

kuma/wiki/tasks.py
jezdez: ah, yeah

kuma/wiki/tasks.py
jezdez: good point

PR #3071, opened 2015-02-10
Fix Bug - 1125803 - Good & bad code examples.

stephaniehobson:
Add styling for good and bad code examples.

Colour blind friendly but not screen reader friendly. Needs to be used in combination with headings.

- added .example-good and .example-bad to be used on code examples
- made some code-block stuff into variables for reuse in new classes
- moved colours from components.styl into vars.styl
- polished comments

Usage: add .example-good or .example-bad to a `<pre>` or `<code>` tag that also has `class="brush:html"` declared on it (needs both the example class AND the prism styling to work).

### DIFF
--- a/media/redesign/stylus/components/components.styl
+++ b/media/redesign/stylus/components/components.styl
@@ -99,24 +99,13 @@
     reverse-link-decoration();
 }
 
-/* hoping to one day use these colours to generate most colours we use on MDN but just here for now */
+/*
+Notifications
+====================================================================== */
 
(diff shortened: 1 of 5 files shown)


stephaniehobson:
Filling in the comments for the commit I decided to re-examine my decision to only make this work on prism-processed code blocks. Making this as "not ready" until I decide whether or not to re-work it.

stephaniehobson:
Urgh. Manged to commit a vendor file >:(

PR #3073, opened 2015-02-11
Django upgrade

ubernostrum:
This is the tracking PR for upgrading kuma to Django 1.7. This PR contains our project meeting notes & tasks.

The process:
1. `git fetch mozilla`
2. `get checkout mozilla/django-upgrade-1116335`
3. `git submodule sync && git submodule update --init`
4. `find . -name "*.pyc" | xargs rm -f`
5. (vagrant-only) `vagrant provision`
6. Either:
   * [django-compat-lint](https://github.com/ubernostrum/django-compat-lint) ([bash script](https://gist.github.com/ubernostrum/0fb56cafab16bb8af546)) kuma with the 1.7 ruleset
   * `./manage.py test 2> test-stderr.txt`

7. Analyze common errors
8. Pick an error to fix
9. Add yourself to a "Tasks for code" item below

Tasks for code:
- [x] @groovecoder verify updated `django-statici18n` usage
- [x] @groovecoder - throw new enhanced intern tests at the branch
- [x] @jezdez - convert puppet provisioning to ansible
  - [x] debug rabbitmq install
- [x] @jezdez - Update kuma setup docs
  - [x] Remove host machine install docs
  - [x] Convert vagrant/vm docs from puppet to ansible (e.g., install ansible on host machine)
  - [x] mark submodules as deprecated
  - [x] pip install `compiled.txt`
- [ ] @openjck and/or @darkwing and/or others - spot-check ansible vm build

Tasks for deploy:
Anytime:
- [x] with Sheeri - Add utf8_distinct_ci collation to MySQL ([bug 1161107](https://bugzilla.mozilla.org/show_bug.cgi?id=1161107))

May 26
- [x] With Jake and cyliang: Create new python 2.7 `virtualenv` ([bug 1148435](https://bugzilla.mozilla.org/show_bug.cgi?id=1148435))
- [x] With Jake and cyliang: Install dependencies from this branch's `compiled.txt` requirements file into the new `virtualenv`

June 2
- [x] Ping VM team re: snapshot & potential restore
- [x] Ping MDN teams re: stage server freeze
- [x] Deploy branch to stage ([logs](http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/?C=M;O=D))
  - http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/3c95d5eabe720b5ca9e25ececfba7da9786325e3.1433259513
    - error: `err: fatal: destination path 'vendor/src/elasticsearch-dsl-py' already exists and is not an empty directory.` ([cause](https://github.com/mozilla/kuma/commit/061a1e602d85cb8225ba8efc9378e6ea4ed1c4f4))
    - fix: `rm -rf vendor/src/elasticsearch-dsl-py`
  - http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/3c95d5eabe720b5ca9e25ececfba7da9786325e3.1433259741
    - error: `ImportError: cannot import name widgets`
    - fix: `find . -name "*.pyc" | xargs rm -f`
  - http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/3c95d5eabe720b5ca9e25ececfba7da9786325e3.1433262124
    - error: `ImportError: No module named Image`
    - fix: add VENV_DIR to the commander_settings.py and https://github.com/mozilla/kuma/commit/2d8b7d7fac66d413e26397960a43793ce8a1a94f
  - http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/2d8b7d7fac66d413e26397960a43793ce8a1a94f.1433271998
    - error: `ImportError: No module named git`
    - fix: https://github.com/mozilla/kuma/commit/0836005e29b798bea544be625c88c903a0b40919
  - http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/0836005e29b798bea544be625c88c903a0b40919.1433276127
    - error: `django.db.utils.OperationalError: (1060, "Duplicate column name 'user_id'")`
    - fix: Continue with fake migrations below ...
- [x] With Jake and cyliang: Convert web heads Apache to use python 2.7 path
  - error: https://cliang.pastebin.mozilla.org/8835477
    - fix: <cyliang> Hrm.  Didn't make it from the admin host to the webhead.  Will try to force the files out ... W00t.  that did it.
  - error: all-servers.conf file moved.  (Was in kuma/etc/apache/all-servers.conf; now in kuma/provisioning/roles/kuma/files/apache2/all-servers.conf)
    - fix: Altering include line
  - error: https://cliang.pastebin.mozilla.org/8835480
    - fix: Remove https://github.com/mozilla/kuma/blame/master/wsgi/kuma.wsgi#L36-L41 (https://github.com/mozilla/kuma/commit/0276e9ff5b338e845943d8d7cc952676e7afb822)
- [x] With Jake and cyliang: Manually do [fake migrations](https://github.com/mozilla/kuma/pull/3073#issuecomment-99107431) during the first deployment
- [x] Deploy branch to stage again
  - http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/0276e9ff5b338e845943d8d7cc952676e7afb822.1433282257
    - error: `Unknown command: 'update_badges'`
    - fix: https://github.com/mozilla/kuma/commit/8068ff61f1698383dd8adf36b1d99cf03918dbbd
- [x] Update cron jobs to use new python2.7 venv

June 2-4
- [x] Stage Test, test, test with writers, WebQA, devs, et al.
  -  [Bug 1171977](https://bugzilla.mozilla.org/show_bug.cgi?id=1171977) - ISE when submitting a demo on staging
- [x] Fix Travis builds on branch

June 9
- [x] Merge branch into master

June 9
- [ ] Repeat deployment steps on Prod

Tasks for post-deploy:
- [ ] Clean out db tables for removed apps and/or libraries, including stale content type records
- [ ] @groovecoder - notify dev-events team that MDN events page is gone
- [ ] @groovecoder - notify mdn lists that features are gone: demo comments
- [ ] @jezdez - send email to dev-mdn re: big dev environment changes (e.g., new `vendor` scheme and `pip` requirements usage)
- [ ] @groovecoder - schedule mdn-dev (+ Cloud Ops?) post-mortem on the update project, including: 
  - Policy for on-going updates to MDN python dependencies to prevent another huge gap
    - requires.io
  - Discuss pip-ifying dependencies, peep-like checks, installing from tarballs, adding pip install compiled to chief script
  - Resuming django-pipeline (and removing GitPython in the process)
  - Discuss a new tagging system
  - Discuss html5lib usage
  - Discuss PostgreSQL
- [ ] Update `django-rest-framework`
- [ ] @jezdez - re-base, merge, & push https://github.com/mozilla/kuma/pull/3062
  - [ ] @groovecoder - schedule mdn-dev meeting re: new config design


groovecoder:
@ubernostrum - how's this branch going? Are you updating more libraries from `vendor/src` to 1.6-compatible versions?

groovecoder:
Ping @ubernostrum - 8 days without an update here. :frowning:

jezdez:
Ok, I've found a series of issues that need to be addressed before moving along with this. That said, I think this looks great in general, lots of changes, but all worth it. We may end up in the present eventually!!

ubernostrum:
I've cleaned up most of the submodule version issues (and even removed one!), but there are a few things that need further poking or decisions:

- [ ] Get `requirements.txt` up to date with the versions of things we're actually using (for everything that can be installed that way), and put hashes in it for peep.
- [ ] Decide on a safe version of django-rest-framework.
- [ ] Try out mainline bleach instead of Les' fork.
- [ ]  django-statici18n upgrade cleanup.
- [ ] django-taggit cleanup.
- [ ] Decide if we really do need httplib2.
- [ ] Decide if we're going to get rid of grappelli.
- [ ] Get me to do a proper PyPI release of soapbox.

groovecoder:
I investigated DRF and [emailed the summary to dev-mdn](https://groups.google.com/d/msg/mozilla.dev.mdn/sWge_fYafo0/V6eshe8PP7wJ)

I suggest we update to 2.3.14 immediately, and probably 2.4.4 soon after.

groovecoder:
Also, mdn-dev and devdoc-writers confirm we can remove grappelli.

groovecoder:
https://github.com/mozilla/kuma/pull/3109 is now merged, so this could rebase on master and then it's ready to run tests?

groovecoder:
I switched to the latest to try to run tests and I got these:
```
    import constance.config
ImportError: No module named config
```
Looks like all those calls need to be:
```
from constance import config
```
?

Should that already be done? Or am I just catching up?

ubernostrum:
@groovecoder I'm working on clearing out a bunch of import-related errors. Some of them come from circular import problems.

groovecoder:
I added https://github.com/mozilla/kuma/commit/7d23be4f67f43f7fb5e30bd0f7474aa9b39a4148 which updates the constance import statements. It was enough to get the test command to actually run ...
```
Ran 610 tests in 50.568s

FAILED (SKIP=6, errors=436, failures=5)
```

groovecoder:
Summary of the most common errors I see in the test output:
Using `django.utils.html`:
```
...
  File "/home/vagrant/src/kuma/wiki/search.py", line 82, in from_django
    'content': strip_tags(obj.rendered_html),
  File "/home/vagrant/src/vendor/src/django/django/utils/functional.py", line 203, in wrapper
    return func(*args, **kwargs)
  File "/home/vagrant/src/vendor/src/django/django/utils/html.py", line 156, in strip_tags
    if not ('<' in value or '>' in value):
TypeError: argument of type 'NoneType' is not iterable
```
Using django-threadedcomments ([0.5.2](https://github.com/HonzaKral/django-threadedcomments/tree/0.5.2)):
```
...
  File "/home/vagrant/src/vendor/src/django/django/conf/urls/__init__.py", line 26, in include
    urlconf_module = import_module(urlconf_module)
  File "/home/vagrant/src/vendor/src/django/django/utils/importlib.py", line 40, in import_module
    __import__(name)
  File "/home/vagrant/src/vendor/packages/django-threadedcomments/threadedcomments/urls.py", line 1, in <module>
    from django.conf.urls.defaults import patterns, url
ImportError: No module named defaults
```
`django.utils.text` error:
```
  File "/home/vagrant/src/vendor/src/django/django/utils/functional.py", line 203, in wrapper
    return func(*args, **kwargs)
  File "/home/vagrant/src/vendor/src/django/django/utils/text.py", line 409, in slugify
    value = unicodedata.normalize('NFKD', value).encode('ascii', 'ignore').decode('ascii')
TypeError: normalize() argument 2 must be unicode, not str
```
These 3 errors seems to account for ~80% of all the test errors.

ubernostrum:
Started pushing the Django 1.7 bits; upgraded the Django submodule and removed South and migrations. Next step will be another quick dependency scan.

groovecoder:
Filed https://github.com/jbalogh/jingo/issues/66 for a Django 1.7 issue with jingo.

groovecoder:
The latest jingo code includes a fix for https://github.com/jbalogh/jingo/issues/66 already. So I updated the jingo submodule to it.

groovecoder:
@jezdez - re: django-threadedcomments, how involved do you think that could be? I have more than half a mind to just disable the feature on the demo studio and get rid of the library completely.

jezdez:
@groovecoder Great question, I was going to raise this actually as well. The issue is not only django-threadedcomments but actually the underlying framework django.contrib.comments that was removed from Django and now is only unusually badly maintained at https://github.com/django/django-contrib-comments. So if there is no other use of the comments system I'd suggest to remove it (but keep the data) and migrate to something like disqus at some point.

jezdez:
@groovecoder Oh, to answer your question: it's non-trivial.

groovecoder:
Today's sanity-check report. `update_product_details` task fails during vagrant provision because the `django-debug-toolbar` version is incompatible. Should we remove `django-debug-toolbar` or update it?

jezdez:
@groovecoder Definitely update it, not get rid of the toolbar.

jezdez:
@groovecoder Seems like we actually don't know which version we updated to :( https://github.com/mozilla/kuma-lib/commit/15b93cc1bd6a970f6031c8267044f7479d21500d

jezdez:
1.3.0 was just released though with 1.8 compatibility http://django-debug-toolbar.readthedocs.org/en/1.3/changes.html#id1

groovecoder:
Updated debug_toolbar, our smuggler usage, and some other various bits which let me run `update_product_details` successfully:
```
vagrant@developer-local:~/src$ /home/vagrant/env/bin/python manage.py update_product_details
/home/vagrant/src/kuma/core/managers.py:23: RemovedInDjango18Warning: `_PrefetchTaggableManager.get_prefetch_query_set` method should be renamed `get_prefetch_queryset`.
  class _PrefetchTaggableManager(_TaggableManager):

/home/vagrant/src/kuma/search/admin.py:46: RemovedInDjango18Warning: Creating a ModelForm without either the 'fields' attribute or the 'exclude' attribute is deprecated - form IndexModelForm needs updating
  class IndexModelForm(forms.ModelForm):

/home/vagrant/src/vendor/src/django/django/forms/widgets.py:143: RemovedInDjango18Warning: `DocumentAdmin.queryset` method should be renamed `get_queryset`.
  .__new__(mcs, name, bases, attrs))

/home/vagrant/src/vendor/src/django/django/forms/widgets.py:143: RemovedInDjango18Warning: `SubmissionAdmin.queryset` method should be renamed `get_queryset`.
  .__new__(mcs, name, bases, attrs))

Grabbing list of JSON files from the server from http://svn.mozilla.org/libs/product-details/json/
No last update timestamp found.
Updating firefox_primary_builds.json from server
Writing new copy of firefox_primary_builds.json to /home/vagrant/product_details_json.
Updating firefox_history_major_releases.json from server
```
...
```
Writing last-updated timestamp (Mon, 16 Mar 2015 13:50:57 GMT).
Product Details update run complete.
```

groovecoder:
And combined with using [@jezdez branch of `django-tidings`](https://github.com/erikrose/django-tidings/pull/25), I was able to run a full `vagrant provision`!

groovecoder:
Newest errors trying to run tests. Tons of these:
```
AttributeError: Problem installing fixture '/home/vagrant/src/kuma/users/fixtures/test_users.json': 'NoneType' object has no attribute 'using'
```
Bunch of these too:
```
UndefinedError: 'kuma.users.models.User object' has no attribute 'get_absolute_url'
```
I think they need more digging than I can give them this evening. :)

groovecoder:
Found a scary [Bunny Cave](http://dont-starve-game.wikia.com/wiki/Bunnyman) ...
underground
The @lmorchard branch of bleach we were using was never merged (and probably shouldn't have been.) [We use it to make bleach skip its full CSS gauntlet on wiki content](https://github.com/mozilla/kuma/blob/b8b55316cc885916e6555088f1409054dcb35b3c/kuma/wiki/managers.py#L41). So, we'll need to investigate if we still need to do that, or if we can change to the more strict [`BLEACH_ALLOWED_STYLES`](https://github.com/mozilla/kuma/blob/b8b55316cc885916e6555088f1409054dcb35b3c/settings.py#L1077-L1082) method.

lmorchard:
If I recall, the reason for that branch is that there's some strange inline CSS hanging out somewhere on MDN that threw Bleach into an infinite loop. Like, I let a single doc test run overnight and it never finished. 

What I don't recall is what doc that was or what the CSS looked like :/

groovecoder:
@lmorchard was that overnight fail on a page view, a page save, or a migration run?

lmorchard:
Memory's fuzzy, but to isolate it from Kuma, I think I'd taken the page source out of the database and wrote a quick Bleach unit test that got hung up on it. But, that was a few laptops ago and I don't think it ever got committed anywhere

lmorchard:
Actually, [these PRs I submitted to Bleach might offer some context](https://github.com/jsocol/bleach/pulls?q=is%3Apr+is%3Aclosed+author%3Almorchard). What I think happened is I *thought* I fixed the infinite loop, then ran into it again, and then just gave up and decided to throw in a flag to skip the CSS regex gauntlets "for now" (2 years ago)

groovecoder:
Today's sanity-check contribution ...

I'm still able to vagrant build (on @jezdez fork of `django-tidings`) and run tests. 

I also made a commit to get past 1 big group of test errors [I mentioned last week](https://github.com/mozilla/kuma/pull/3073#issuecomment-91373155) by updating our `ABSOLUTE_URL_OVERRIDES` setting to match the `users.user` model. (Django 1.7 dropped `get_absolute_url` from `AbstractUser`).

groovecoder:
Another commit to get past the `AttributeError: Problem installing fixture` errors. Those 2 fixes have fixed 150+ tests. Down from 476 test errors to 329 ...
```
Ran 612 tests in 64.720s

FAILED (SKIP=6, errors=329, failures=5)
```

groovecoder:
Nice updates. Fixed another 34 tests.
```
Ran 627 tests in 78.210s

FAILED (SKIP=6, errors=295, failures=4)
```

groovecoder:
Another update; another 14 tests fixed.
```
Ran 627 tests in 74.324s

FAILED (SKIP=6, errors=281, failures=6)
```

groovecoder:
4 more tests fixed. Starting to feel like ...
treadwall

jezdez:
Yay :-/

```
FAILED (SKIP=6, errors=213, failures=8)
```

jezdez:
@ubernostrum whoops, accidentally removed the `skip_gauntlet` parameter already since I assumed it was something in their API. It's a one line change though and shouldn't conflict with your work, feel free to reverse it.

jezdez:
@ubernostrum @groovecoder Just merged master into the branch

jezdez:
```
FAILED (SKIP=10, errors=73, failures=34)
```

jezdez:
```
FAILED (SKIP=10, errors=40, failures=44)
```

ubernostrum:
Following up on bleach: I grabbed the source of [this revision from MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Build_Documentation$revision/39242), which was current at the time @lmorchard filed his bleach bugs, and I tried manually doing bleach.clean() on it with kuma's defined set of allowed tags/attributes/styles. It returned cleaned HTML immediately, so I'm going to go with "whatever it was, it's no longer a problem".

jezdez:
@ubernostrum Yay!!

jezdez:
```
FAILED (SKIP=10, errors=21, failures=35)
```

groovecoder:
FWIW, today's sanity check:
```
Ran 636 tests in 173.701s

FAILED (SKIP=10, errors=21, failures=34)
```
* I still had to change django-tidings to @jezdez branch. (Though I could have done master now)
* I still saw the django-soapbox warnings

I poked into the `events.tests.test_models:TestCalendar` errors. Looks like we're using an old `toxml()` method of html5lib that has been removed in version 0.999. I couldn't quite see how to fix it before I stopped for the day.

ubernostrum:
After a bit of fighting with packaging stuff, [soapbox is on PyPI](https://pypi.python.org/pypi/django-soapbox). Will update our copy accordingly.

jezdez:
I'm almost done with tidings, will push to PyPI today.

groovecoder:
I took a shot at merging `master` into this branch in my repo:

https://github.com/groovecoder/kuma/commit/2c6ded4e9eddda713a21acd7f356acdcb9517c6a

There was a conflict in `kuma/wiki/views.py` where I introduced a new `_check_doc_revision_for_alteration` where we had also removed a `@transaction.autocommit` decorator. But, [I cleaned it up](https://github.com/groovecoder/kuma/commit/2c6ded4e9eddda713a21acd7f356acdcb9517c6a#diff-08089c7053931e5d5c6afbb6fb3e2533). Okay to add that merge commit to this branch?

jezdez:
@groovecoder Why wouldn't it, I merged master last week and had way more conflicts

groovecoder:
Merged master. I'm digging into the events code use of html5lib 0.999 now.

groovecoder:
After digging a little bit, I checked to see if/how much the events app is even used ... We have no links to it, and there's been < 300 page-views on it this year.

So, now I'm removing the `kuma.events` app completely to see how that affects/improves tests.

jezdez:
@groovecoder Isn't that used on the homepage?

groovecoder:
Nope. You might be thinking of the Hacks blog feed on the home page? Or it might have been on the home page before the redesign.

In any case, I have [a branch with `kuma.events`](https://github.com/groovecoder/kuma/tree/remove-events-1116335) removed that fixes (or removes) another 7 errors and 1 failure:
```
Ran 629 tests in 168.090s

FAILED (SKIP=10, errors=14, failures=34)
```

Waiting to hear back from dev-events@mozilla.com if we need to keep the page around before I remove it. Or we might have to leave behind a redirect at least.

groovecoder:
I spot-checked a bunch of inline style docs on the new version of bleach and [emailed dev-mdn list](https://groups.google.com/d/msg/mozilla.dev.mdn/WQWE3aEkIM8/P069C4Tau8wJ). So far it looks fine. I might get some more responses with specific pages to check, but I feel good keeping the updated mainline bleach library.

jezdez:
```
FAILED (SKIP=10, errors=20, failures=35)
```

jezdez:
I'm seeing an interesting bug, caused by https://github.com/alex/django-taggit/commit/b32c6cbeb37d40d00d687948ccc4854205b59b7c.

In short, the way MySQL compares strings means it'll think that "CSS Reference" and "CSS Référence" are the same and taggit will raise an `IntegrityError` to prevent a infinite loop and let the app deal with that case. What this boils down is that we're stuck with the question whether we should switch the taggit tables to using a different collation that respect the fact that the tag name contains non-ASCII characters. I think we should be switching to `utf8_unicode_ci` with a migration and see if that helps.

See https://docs.djangoproject.com/en/dev/ref/databases/#collation-settings for more info and https://github.com/alex/django-taggit/pull/220/files#r11341702 for some prior discussion in a proposal to fix that in taggit. @ubernostrum @groovecoder I need your help to sanity check this plan, this is a serious regression.

jezdez:
Oh, btw, the error traceback is https://gist.github.com/jezdez/e6a508a943f6caeb935d

jezdez:
```
FAILED (SKIP=10, errors=10, failures=35)
```

groovecoder:
Removed the events app.
```
FAILED (SKIP=10, errors=3, failures=34)
```

groovecoder:
@jezdez is it a regression or is it still [bug 776048](https://bugzilla.mozilla.org/show_bug.cgi?id=776048)?

jezdez:
@groovecoder I think that bug is probably another symptom, yeah. Either way, this shows up hard now in the tests so we need to fix it.

jezdez:
```
FAILED (SKIP=10, errors=3, failures=30)
```

groovecoder:
Yeah, strange. I'm getting:
```
FAILED (SKIP=10, errors=2, failures=29)
```

jezdez:
FTR, I just checked in stage and prod we use `utf8_general_ci` as the collation for the taggit name and slug columns.

jezdez:
So again a day wasted to those tests and brittle software design. We had a talk with Sheeri and she recommended using the utf8_bin collation for the column of the tag table.

Sadly that won't do since that collation doesn't only take Unicode characters into account when sorting (and by that when adding tags to the table) but also is case sensitive.

But what we want are case insensitive tags that take unicode characters into account, so that "rest" and "rést" would not be the same tag, but "rest" and "REST" would.

The rest of my day shall be explained by a IRC transcript:

```
— @jezdez arrived at http://dev.mysql.com/doc/refman/5.5/en/adding-collation.html
23:41:14 <@jezdez> sheeri: what would it take to implement a utf8_distinct_ci collation?
23:41:36 <@jezdez> as we want case insensitive tags that take unicode characters into account when comparing them
23:41:47 <@jezdez> similar to https://mariadb.com/blog/adding-case-insensitive-distinct-unicode-collation
```

jezdez:
So I think we need to add a new collation to MySQL.

lmorchard:
FWIW, "CSS Reference" and "CSS Référence" (ie. the tag and DB collation issue) sounds like good ole [bug 776048](https://bugzilla.mozilla.org/show_bug.cgi?id=776048#c13). 

Tried solving that back in summer 2013, but it also required an upgrade to the MySQLdb module. Turned out the DB changes triggered a unicode bug in the driver. But, maybe what you're proposing won't cause the same issue

jezdez:
@lmorchard Yeah, that's a symptom, we're running with utf8_generical_ci right now, which evaluates "é" as "e" during string comparison, hence the `IntegrityError`. That this shows up in a the tests is simply because django-taggit has stopped trying to automagically work around that exception in a newer version and just passes it back to the app – which I think is the correct way of handling issues like this.

That said, upgrading the MySQL driver also will be required since if we ever want to use utf8_bin we'll need that newer version to not accidently have to handly `array.array` instances as the return data type. 

That said, according to Django's docs (and the trustworthy ORM people there) that's not what's recommended nowadays anyway, but a fork of the original db library. https://docs.djangoproject.com/en/1.7/ref/databases/#mysql-db-api-drivers So https://pypi.python.org/pypi/mysqlclient is what I'll be upgrading us to.

jezdez:
Alternatively we should just move to Postgres and stop having to worry about all this shit.

groovecoder:
Down to 20 errors/failures.

```
Ran 630 tests in 187.461s

FAILED (SKIP=10, errors=2, failures=18)
```

groovecoder:
@ubernostrum - got an update on the ContentSectionTool tests? Think we're down to those and/or other wiki view tests after the unicode tag issue.

ubernostrum:
Good news: wiki/test_content is now at 0 errors, 0 failures for me. I also think I understand how SectionIDFilter works.

Bad news: my brain poured out through my ears about halfway through that.

groovecoder:
Awesome work @ubernostrum. We're down to 6 errors/failures:
```
Ran 629 tests in 183.592s

FAILED (SKIP=10, errors=1, failures=5)
```

groovecoder:
I dug into the `NewRevisionTests.test_new_revision_GET_based_on` failure. The feature [was actually disabled](https://github.com/mozilla/kuma/commit/05e4e20c), so we might be able to clean it out completely now.

But ...

I narrowed the failure down to - I think - updated [Django 1.7.2 now using `__html__` methods](https://code.djangoproject.com/ticket/23831). The `RevisionForm`'s `content` `BoundField` has an `__html__` method and that's what adds the extra `\r\n` characters that makes the test fail.

I spot-checked some regular edits and nothing else seems to cause any errors. We should discuss at our next meeting.

ubernostrum:
@groovecoder Is there a reason to not just add the `\r\n` to the expected output?

ubernostrum:
Or I guess more correctly, to not just compare output after doing `.strip()`, since that gets rid of spurious whitespace and newline characters? Adding that gets the test passing for me.

ubernostrum:
Actually, never mind. I keep forgetting `assertHTMLEqual` exists in Django now, and that's the method we should be using any time we're comparing HTML. Switched the test to use `assertHTMLEqual` and now it passes.

ubernostrum:
@groovecoder and @jezdez FYI: as of that last spree of commits, I see only the tag-case-sensitivity issue in the wiki tests. Some skips that I can't be 100% sure of, but every other test in wiki is passing for me now.

groovecoder:
w00t! :clap: For anyone still keeping score ... we're down to 2 errors/failures.
```
Ran 629 tests in 190.551s

FAILED (SKIP=10, errors=1, failures=1)
```

groovecoder:
FWIW, the test error & failure I'm seeing:
```
======================================================================
ERROR: Verify the tagged documents list view, even for duplicate tags
----------------------------------------------------------------------
  File "/home/vagrant/src/kuma/wiki/tests/test_templates.py", line 710, in test_tag_list_duplicates
    fr_tag.save()
  File "/home/vagrant/src/vendor/src/django-taggit/taggit/models.py", line 76, in save
    return super(TagBase, self).save(*args, **kwargs)
  File "/home/vagrant/src/vendor/src/django/django/db/models/base.py", line 589, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/vagrant/src/vendor/src/django/django/db/models/base.py", line 617, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/home/vagrant/src/vendor/src/django/django/db/models/base.py", line 698, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/home/vagrant/src/vendor/src/django/django/db/models/base.py", line 731, in _do_insert
    using=using, raw=raw)
  File "/home/vagrant/src/vendor/src/django/django/db/models/manager.py", line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/vagrant/src/vendor/src/django/django/db/models/query.py", line 921, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/home/vagrant/src/vendor/src/django/django/db/models/sql/compiler.py", line 920, in execute_sql
    cursor.execute(sql, params)
  File "/home/vagrant/src/vendor/src/django/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/vagrant/src/vendor/src/django/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/home/vagrant/src/vendor/src/django/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/vagrant/src/vendor/src/django/django/db/backends/mysql/base.py", line 129, in execute
    return self.cursor.execute(query, args)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 173, in execute
    self.errorhandler(self, exc, value)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
IntegrityError: (1062, "Duplicate entry 'CSS R\xc3\xa9f\xc3\xa9rence' for key 'name'")

======================================================================
FAIL: test_toc_view (kuma.wiki.tests.test_views.ViewTests)
----------------------------------------------------------------------
  File "/home/vagrant/src/kuma/wiki/tests/test_views.py", line 219, in test_toc_view
    eq_(resp.content, '<ol><li><a href="#Head_2" rel="internal">Head 2</a>'
AssertionError: '<ol><li><a rel="internal" href="#Head_2">Head 2</a><ol><li><a rel="internal" href="#Head_3">Head 3</a></ol></li></ol>' != '<ol><li><a href="#Head_2" rel="internal">Head 2</a><ol><li><a href="#Head_3" rel="internal">Head  3</a></ol></li></ol>'

----------------------------------------------------------------------
```

jezdez:
@groovecoder @ubernostrum I just added Django migrations, so you need to run `./manage.py migrate --fake``. BUT:

One thing to consider is that the 0002 migration in the wiki app does the conversion of the collation from `utf8_general_ci` to `utf8_distinct_ci` which assumes that you have that collation installed. That's *not* done yet as we're working through the issue I've seen locally (https://bugzilla.mozilla.org/show_bug.cgi?id=1161107), but in case you'd like to run that migration make sure to *not* run the general fake migration above but this first:

```
./manage.py migrate --fake wiki 0001
./manage.py migrate wiki 0002
./manage.py migrate --fake
```
That should first recreate the migration story correctly then apply the change in collation, then fake migrate the rest.

groovecoder:
For posterity, here's how I took a crack at adding this the MySQL collation to my vm ...

1. Check no collation exists
```
mysql> SELECT * FROM INFORMATION_SCHEMA.COLLATIONS WHERE id=252 OR collation_name='utf8_distinct_ci';
Empty set (0.01 sec)
```
2. Added this to my `/user/share/mysql/charsets/Index.xml`:
```
<collation name="utf8_distinct_ci"    id="252">
</collation>
```
3. Added the [`utf8_distinct_ci.xml`](ftp://ftp.skysql.com/downloads/hartmut/utf8_distinct_ci.xml) file to my `/usr/share/mysql/charsets/` directory.
4. I restarted MySQL via `sudo service mysql restart`
5. Checked for the collation:
```
mysql> SELECT * FROM INFORMATION_SCHEMA.COLLATIONS WHERE id=252 OR collation_name='utf8_distinct_ci';
+------------------+--------------------+-----+------------+-------------+---------+
| COLLATION_NAME   | CHARACTER_SET_NAME | ID  | IS_DEFAULT | IS_COMPILED | SORTLEN |
+------------------+--------------------+-----+------------+-------------+---------+
| utf8_distinct_ci | utf8               | 252 |            |             |       8 |
+------------------+--------------------+-----+------------+-------------+---------+
1 row in set (0.00 sec)
```
6. Ran the first fake migration: `./manage.py migrate --fake wiki 0001`
7. Ran the migration to update the columns to the new collation: `./manage.py migrate wiki 0002`
8. Got error:
```
django.db.utils.OperationalError: (1273, "Unknown collation: 'utf8_distinct_ci'")
```

groovecoder:
I got the collation installed correctly. When I ran the test, I saw the migrations applied, but got the same error. :cry: 
https://pastebin.mozilla.org/8832638
So, I'm not sure where else to go in this rabbit hole? Can we defer this fix to replacing the taggit-tag system completely after we update?

jezdez:
The steps that @groovecoder posted weren't enough for me to get this running locally, as the `utf8_distinct_ci.xml` *content* needed to be added to the `Index.xml` file. So here are the steps that worked for me:

1. Check that no collation exists with the new name

	```
	mysql> SELECT * FROM INFORMATION_SCHEMA.COLLATIONS WHERE collation_name='utf8_distinct_ci';
	Empty set (0.01 sec)
	```

2. Find out where the character files are stored:

	```
	mysql> SHOW VARIABLES LIKE 'character_sets_dir';
	+--------------------+------------------------------------------------------+
	| Variable_name      | Value                                                |
	+--------------------+------------------------------------------------------+
	| character_sets_dir | /usr/local/Cellar/mysql/5.6.24/share/mysql/charsets/ |
	+--------------------+------------------------------------------------------+
	1 row in set (0.01 sec)
	```

3. Download the [`utf8_distinct_ci.xml`](ftp://ftp.skysql.com/downloads/hartmut/utf8_distinct_ci.xml) file.

4. Copy the content of the `utf8_distinct_ci.xml` file to the `<charset name="utf8">` element of the `Index.xml` file inside the `character_sets_dir` so that it's a different sub element.

5. Figure out the max ID given to the installed collations:

	```
	mysql> SELECT MAX(ID) FROM INFORMATION_SCHEMA.COLLATIONS;
	+---------+
	| MAX(ID) |
	+---------+
	|     247 |
	+---------+
	1 row in set (0.02 sec)
	```

6. Modify the id of the added collation in the `Index.xml` with the max ID + 1:

	```
	<collation name="utf8_distinct_ci" id="248">
	...
	</collation>
	```

7. Modify the ``<charsets max-id="xx">`` root element on top of the file with the ID given in step 6:

	```
	<charsets max-id="248">
	...
	</charsets>
	```

8. Restart MySQL via `sudo service mysql restart` or whatever other means

9. Checked for the collation:

	```
	mysql> SELECT * FROM INFORMATION_SCHEMA.COLLATIONS WHERE collation_name='utf8_distinct_ci';
	+------------------+--------------------+-----+------------+-------------+---------+
	| COLLATION_NAME   | CHARACTER_SET_NAME | ID  | IS_DEFAULT | IS_COMPILED | SORTLEN |
	+------------------+--------------------+-----+------------+-------------+---------+
	| utf8_distinct_ci | utf8               | 248 |            |             |       8 |
	+------------------+--------------------+-----+------------+-------------+---------+
	1 row in set (0.01 sec)
	```

10. Ran the first fake migration: `./manage.py migrate --fake wiki 0001`

11. Ran the migration to update the columns to the new collation: `./manage.py migrate wiki 0002`

12. Run the tests that errors out before:

	```
	./manage.py test kuma.wiki.tests.test_templates:DocumentListTests
	```

That should not raise the same exception as shown here: https://github.com/mozilla/kuma/pull/3073#issuecomment-98838102

jezdez:
FTR, this is my ``Index.xml``: https://gist.github.com/jezdez/44ec01d92d17918c259e

ubernostrum:
@jezdez set things up per your explanation, and now I don't get the tag uniqueness errors but I do get one test failure. Full output of wiki tests is in [this gist](https://gist.github.com/ubernostrum/069a2981eb858deb146d).

jezdez:
Yay @ubernostrum! I'll take a look at the other test failure later today.

groovecoder:
:exclamation:

jezdez:
```
Ran 629 tests in 162.905s

OK (SKIP=10)
```

groovecoder:
dance

jezdez:
I'll need more time weeding out vendor/lib and vendor/packages as there are overlaps. After talking with Jake and C they'll want to have a requirements file that they can install for at least the compiled dependencies.

I want us to have three things:

1) compiled.txt with complied dependencies for dev/travis/futureprod
2) source.txt with analog versions for the depenencies in vendor/src
3) packages.txt with the same for vendor/packages

vendor/lib/python should be merged with vendor/packages
E.g. did you know we have pytz both in vendor/lib/python and vendor/packages?

jezdez:
those requirements files can easily also contain the peep hashes btw

ubernostrum:
Oops, I flake8'd.

jezdez:
@ubernostrum There are still a few things unfixed: https://travis-ci.org/mozilla/kuma/jobs/61698438
The files with the missing `gettext` function, that's django-dbgettext's fault as it just generates those dummy files so the translation message extraction system picks those things up. What I would recommend is importing django's gettext there.

ubernostrum:
While I'm thinking of it, sometime in the next couple days (i.e., once I'm home), I'm going to do a run through and 1.7-ify all our urls.py files. Right now we're still using `patterns` + prefixes in a bunch of places, and fixing that is a quick and easy future-proofing thing.

jezdez:
```
Ran 629 tests in 187.877s

OK (SKIP=10)
```

groovecoder:
bender-disco-dance

ubernostrum:
@jezdez:
The new `vendor/` and `requirements/` are working fine for me locally, and although I can't say I've reviewed *everything* you did, it looks clean and correct to me from a dig through the commits. Which is not something I thought I'd ever say about our `vendor/` or `requirements/`.

groovecoder:
This needs another merge from master now. :/

jezdez:
I'll handle it on Monday.

groovecoder:
First try of `vagrant up` with ansible:
```
TASK: [kuma | Ensure local settings files] ************************************
fatal: [developer-local] => input file not found at /Users/lcrouch/code/kuma/provisioning/roles/kuma/files/vagrant/settings_local.py or /Users/lcrouch/code/kuma/provisioning/vagrant/settings_local.py

FATAL: all hosts have already failed -- aborting
```
I don't have a `vagrant/settings_local.py` file. Just the `settings_local.py` file at the root of the repo?

groovecoder:
Ansible/provisioning is coming along. After [I restored `kumascript_settings_local`](https://github.com/mozilla/kuma/pull/3073#commitcomment-11386471), I hit this during a provision run: https://gist.github.com/groovecoder/710ad4207af5379736a3

So I commented `debug_toolbar` out of the `INSTALLED_APPS` in `settings_local.py` and the provisioning finished!

When I ran the server and hit the home page, it gave me another `AttributeError: 'tuple' object has no attribute 'regex'` from `urlresolvers`. :cry:

Then when I try to `vagrant halt`, `vagrant up`, and `vagrant provision` again I'm getting `socket.error: [Errno 111] Connection refused` from py-amqp: https://gist.github.com/groovecoder/e52a6584217646232acb

I'll keep working with it though.

groovecoder:
https://github.com/mozilla/kuma/commit/9049bdeb1027fa1f7e37378d661837bc1f26fad5 caused the `AttributeError` from `urlresolvers` so I'm working without it now.

groovecoder:
FWIW, I get a bunch of test errors in the vm like so:
```
OSError: [Errno 40] Too many levels of symbolic links: '/home/vagrant/src/media/uploads/demos'
```

groovecoder:
I got rid of the `Too many levels of symolic links` error by removing my existing `media/uploads` file. It's also in `.gitignore`.

groovecoder:
Down to 2 other `OSError`s related to demo uploads:
```
======================================================================
ERROR: Demo files should be deleted when the demo is censored.
----------------------------------------------------------------------
  File "/home/vagrant/src/kuma/demos/tests/test_models.py", line 484, in test_censored_demo_files_are_deleted
    s.censor(url="http://example.com/censored-explanation")
  File "/home/vagrant/src/kuma/demos/models.py", line 494, in censor
    rmtree(root)
  File "/usr/lib/python2.7/shutil.py", line 256, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 254, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/home/vagrant/src/media/uploads/demos/t/e/tester/2095312189753de6ad47dfe20cbe97ec'

======================================================================
ERROR: Ensure that demo files are deleted along with submission record
----------------------------------------------------------------------
  File "/home/vagrant/src/kuma/demos/tests/test_models.py", line 336, in test_demo_deletion
    s.delete()
  File "/home/vagrant/src/kuma/demos/models.py", line 539, in delete
    rmtree(root)
  File "/usr/lib/python2.7/shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 250, in rmtree
    os.remove(fullname)
OSError: [Errno 16] Device or resource busy: '/home/vagrant/src/media/uploads/demos/t/e/tester/2095312189753de6ad47dfe20cbe97ec/.nfs0000000002e2e4820000001e'

----------------------------------------------------------------------
Ran 624 tests in 156.908s

FAILED (SKIP=10, errors=2)
```

groovecoder:
I tested merging master into this branch, and I started to get this error during `vagrant provision`: https://gist.github.com/groovecoder/1a62f34523b1e97f63bc

ubernostrum:
@groovecoder I have what looks like a clean merge (only a couple of conflicts, mostly from languages added since we last did this). Do you want me to push it up here so we can all debug from the same version?

ubernostrum:
Also, the `AttributeError: 'tuple' object has no attribute 'regex'` message means there's a URL pattern somewhere that's still a raw tuple instead of a call to `url()`. I thought I caught all of those, but will do another run looking for them.

groovecoder:
I'd rather revert the url commit and then do a merge so we don't have another fix to make. The current fixes are already:

* [ ] Restore `kumascript_settings_local.json` for provisioning
* [ ] Fix amqp `socket.error` during provision
* [ ] Fix `OSError` during tests
* [ ] Fix Travis runs

I've been working on a merged branch here: https://github.com/groovecoder/kuma/tree/merged-django-upgrade-1116335 went back to https://github.com/mozilla/kuma/commit/e4aa3a20e2451dfd7bd28cc4d67b37884e7d3b0d to avoid the `urlsresolvers` error and then cherry-picked the ansible commits to test it. (I scanned our `urls.py` files first to see if I could spot the missing `url()` call.)

I'm going to work in a branch where I revert the url commit, merge, and then try to fix the above errors.

ubernostrum:
I'd really rather not revert the URL work especially since [I found and fixed the problem](https://github.com/ubernostrum/kuma/commit/29414a91c983c9c74bbdc9a9d0a67eff2474aee1).

groovecoder:
Added merge commit, fix for urlconfs, and restoring of kumascript_settings_local for ansible.

groovecoder:
Current fix-list:

* [x] Restore `kumascript_settings_local.json` for provisioning (https://github.com/mozilla/kuma/commit/5251456a848f6291fe525e36c966908ddadfc92a)
* [ ] vm test fixes:
  * [ ] `OSError: [Errno 39] Directory not empty` in `test_censored_demo_files_are_deleted`
  * [ ] `OSError: [Errno 16] Device or resource busy` in `test_demo_deletion`
  * [ ] 5 x `TypeError: 'str' object is not callable` during search view tests
* [ ] Fix Travis runs

groovecoder:
@jezdez can you think of a way to re-use the `mysql.yml` ansible task file on Travis?

jezdez:
@groovecoder Yeah, lemme work on it on Monday please. In short, either we'll do this with [tags](http://docs.ansible.com/playbooks_tags.html) or with a separate playbook. Since we already have some script to set up Travis I'd say we port that to a Ansible role and use it in a separate playbook.

groovecoder:
Can we put a time limit on porting the travis scripts to ansible role? It seems like there are potential yak-holes under that, and we really need to merge this.

groovecoder:
I haven't been able to fix it yet, but on Friday I found that the ansible playbook is creating `rabbitmq.conf` with `tcp_listeners, []`. I think we need to add a `rabbitmq_conf_tcp_listeners_address` to our `vars/rabbitmq.yml` because [the `rabbitmq.conf.j2` template](https://github.com/mozilla/kuma/commit/889431918343714090575e1a2ff121ef2a079482#diff-bb5a1b199e62f6f1a4bc034e600b7896) is looking for it? (When I tried that on Friday it didn't seem to work, but maybe I need to re-build the box and not just re-provision?)

groovecoder:
Note: https://github.com/groovecoder/kuma/commit/0334f1e7da9427998d1701744759a434817f7e2d fixes the amqp connection issue in the vm, but I'm waiting to see what we do with https://github.com/mozilla/kuma/commit/b8b8f837149789afdd2a00309b4524f347e61acd before I push it to this branch.

groovecoder:
I incorporated [the fix for amqp connection in the vm](https://github.com/mozilla/kuma/commit/880cf5ff6d028a9bb972c3f08052e5da7068229a). I ran the new intern client-side test suite against this branch and [fixed a bug I found in the newsfeed template](https://github.com/mozilla/kuma/commit/3c95d5eabe720b5ca9e25ececfba7da9786325e3). (I also sent [a PR to improve the intern test suite messages.](https://github.com/mozilla/kuma/pull/3229))

Only thing I know we have to do now is fix Travis CI. I'll work on that tomorrow.

groovecoder:
I'm using https://github.com/mozilla/kuma/pull/3232 against this branch to let me test Travis builds without committing here (yet).

I couldn't find a path to follow with running ansible connections to Travis CI localhost, so I started a python script that re-uses those same Jinja2 templates as ansible to install the `Index.xml` file into MySQL.

ubernostrum:
When trying to follow vagrant install instructions, getting this at the submodule init step:

    $ git submodule update --init --recursive
    Submodule 'testdata' (https://github.com/html5lib/html5lib-tests.git) registered for path 'html5lib/tests/testdata'
    Cloning into 'html5lib/tests/testdata'...
    remote: Counting objects: 3953, done.
    remote: Total 3953 (delta 0), reused 0 (delta 0), pack-reused 3953
    Receiving objects: 100% (3953/3953), 4.91 MiB | 2.74 MiB/s, done.
    Resolving deltas: 100% (2028/2028), done.
    Checking connectivity... done.
    Submodule path 'vendor/src/html5lib/html5lib/tests/testdata': checked out 'f6a1b202de14fc057b196044c5ebef4672be3dd0'
    fatal: Needed a single revision
    Unable to find current revision in submodule path 'vendor/src/urlobject'

groovecoder:
I destroyed and re-built my vm last night and had no issue. Probably wifi connection.

jezdez:
292432986db702a42ffa72274547f9388976e58b is wrong. `settings_test.py` is committed to the repo (until we merge the settings-refactor), so overwriting it with a different file from Ansible will result in an unclean git index (aka you have to commit it). Since we're deprecating non-Vagrant installations (at least from a supported maintenance perspective).

jezdez:
woops, part of the sentence missing.

..I'm going to hardcode that path in the settings_test.py.

jezdez:
So apparently there are problems with MySQL on Travis, https://github.com/travis-ci/travis-ci/issues/2353 (reported by the author of the Ansible roles we're using for MySQL). So I'm going to uninstall the Travis provided MySQL install the community server from MySQL itself (https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/index.html). **Because I haven't explored a rabbit cave in recent weeks**

jezdez:
```
Ran 625 tests in 214.740s

OK (SKIP=10)
```

groovecoder:
Another merge from master.
```
Ran 625 tests in 162.653s

OK (SKIP=10)
```

jezdez:
giphy

PR #3081, opened 2015-02-13
Fix bug 1132962: Use <mark> for matched terms

openjck:
Quoting MDN...

    The HTML <mark> Element represents highlighted text, i.e., a run of
    text marked for reference purpose, due to its relevance in a
    particular context. For example it can be used in a page showing
    search results to highlight every instance of the searched for word.

-- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark

### DIFF
--- a/kuma/search/fields.py
+++ b/kuma/search/fields.py
@@ -34,7 +34,7 @@ class DocumentExcerptField(serializers.Field):
     """
     A serializer field that given a wiki DocumentType object returns
     a cleaned version of the excerpt fields with the highlighting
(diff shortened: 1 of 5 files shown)


robhudson:
Nice. r+ on the backend changes.

stephaniehobson:
I like the proposed changes but I don't have time this week to rebuild my machine to get search working to test them. Going to unassign myself.

robhudson:
I checked it out locally and it looked good.

PR #3083, opened 2015-02-18
fix bug 1124423 - add African languages

groovecoder:
To spot-check:

1. `cd locale && svn update` to get [the new languages' message files from SVN](http://viewvc.svn.mozilla.org/vc?view=revision&revision=137752)
2. Check that [/ee/](https://developer-local.allizom.org/ee/), [/ha/](https://developer-local.allizom.org/ha/), [/ig/](https://developer-local.allizom.org/ig/), [/ln/](https://developer-local.allizom.org/ln/), [/sw/](https://developer-local.allizom.org/sw/), [/wo/](https://developer-local.allizom.org/wo/), and [/yo/](https://developer-local.allizom.org/yo/) all work.


groovecoder:
Note to self: Travis is failing because [its install script needs an updated `product_details_json.tar.gz` from S3](https://github.com/mozilla/kuma/blob/master/scripts/travis-install#L12-L13).

groovecoder:
@jezdez - now that our wheels docs are gone, do we document our amazon s3 storage stuff anywhere else? I need to update those `.tar.gz` files on S3, unless you think there's a better way to get those onto Travis? (e.g., just commit them to the repo?)

jezdez:
@groovecoder Yeah, I think we should try to find out if we can put this tarball somewhere else. How often do we really update the data behind the products? Same for the locale?

jezdez:
@groovecoder To be clear, I think the S3 is not a good idea as it complicates maintenance and is a inconsistency between dev and prod.

groovecoder:
I say we just check it into the kuma repo. (http://12factor.net/codebase ?)

jezdez:
@groovecoder go for it

groovecoder:
@jezdez r? I just put the files into `scripts/` along with `travis-install`

jezdez:
@groovecoder @ubernostrum wrote in June last year about how we should structure the repo, including a new directory called `/etc`:

	* ``etc/`` would be the location for non-Django-related configuration
	 files and other miscellaneous scripts, similar to ``/etc`` on a Unix
	 system. This would include any necessary web-server configuration
	 files, ``pip`` requirements files, Vagrant configuration (if
	 possible), maintenance scripts and so on.

I think those files should land there as well, maybe under `/etc/data` or something. @ubernostrum any opinion on this?

groovecoder:
Since this has minimal routine changes outside of the Travis `/etc/data` change, I'm opening this up to other assignees.

PR #3085, opened 2015-02-19
Fixes bug 1132159 - Allow highlight to be disabled

robhudson:


kuma/search/filters.py
```diff
@@ -192,3 +194,21 @@ def filter_queryset(self, request, queryset, view):
...
+class HighlightFilterBackend(BaseFilterBackend):
+    """
+    A django-rest-framework filter backend that adds search term highlighting.
+
+    If the query string `highlight=0` is passed, we bypass highlighting.
+
+    """
+    def filter_queryset(self, request, queryset, view):
+        highlight = request.GET.get('highlight', True)
+        if highlight != '0':
```
jezdez: I dislike GET parameters that assume that 0 is false-y, let's make that `?highlight=false`

kuma/search/serializers.py
```diff
@@ -118,12 +117,15 @@ def field_to_native(self, obj, field_name):
...
 
 class DocumentSerializer(BaseDocumentSerializer):
-    excerpt = DocumentExcerptField(source='*')
+    excerpt = serializers.SerializerMethodField('get_excerpt')
     tags = serializers.ChoiceField(read_only=True, source='tags')
     score = serializers.FloatField(read_only=True, source='_meta.score')
     explanation = serializers.SerializerMethodField('get_explanation')
     parent = BaseDocumentSerializer(read_only=True, source='parent')
 
+    def get_excerpt(self, obj):
```
jezdez: Hm, why not keep using the field subclass?

kuma/search/filters.py
```diff
@@ -192,3 +194,21 @@ def filter_queryset(self, request, queryset, view):
...
 
         return queryset
+
+
+class HighlightFilterBackend(BaseFilterBackend):
+    """
+    A django-rest-framework filter backend that adds search term highlighting.
+
+    If the query string `highlight=0` is passed, we bypass highlighting.
+
```
jezdez: Remove empty line in comment

kuma/search/filters.py
```diff
@@ -192,3 +194,21 @@ def filter_queryset(self, request, queryset, view):
...
+    """
+    def filter_queryset(self, request, queryset, view):
+        highlight = request.GET.get('highlight', True)
```
jezdez: This should be `request.QUERY_PARAMS.get('highlight') or True`
- request.GET is not the DRF way of things
- the 'highlight' query parameter could be an empty string, e.g. `?highlight=&q=yeah+yeah`

kuma/search/filters.py
```diff
@@ -192,3 +194,21 @@ def filter_queryset(self, request, queryset, view):
         queryset = queryset.extra(facets=facets)
 
         return queryset
+
+
+class HighlightFilterBackend(BaseFilterBackend):
+    """
+    A django-rest-framework filter backend that adds search term highlighting.
+
+    If the query string `highlight=0` is passed, we bypass highlighting.
```
jezdez: I was wondering if we should "hide" that parameter in a more subtle name, e.g. "h"? So that `?q=yeah+yeah&h=false` works?

jezdez:
rwc+

kuma/search/serializers.py
robhudson: Because the code was duplicated it two places, once in the serializer field class and once in `WikiDocumentType.get_excerpt`, which will also return the summary if no highlight.

kuma/search/filters.py
robhudson: Hmm, I always thought that was a pep8 thing but now that I look I can't find it.

kuma/search/serializers.py
jezdez: ah, okidoki :+1:

robhudson:
Updated with refactored DRF serializers for query string validation.

kuma/search/filters.py
```diff
@@ -16,6 +20,17 @@ def get_filters(getter_func):
     return filters.keys()
 
 
+class QueryStringValidationMixin(object):
+    """
+    A mixin to validate query string parameters.
+    """
+    def validate_query_params(self, request):
```
jezdez: This seems be called three times for each request, which seems wasteful. Couldn't we run this once per request in the view and set a view parameter? The filter methods do get the view instance passed along after all.

kuma/search/filters.py
```diff
@@ -84,13 +98,14 @@ class SearchQueryBackend(BaseFilterBackend):
     ]
 
     def filter_queryset(self, request, queryset, view):
-        search_param = request.QUERY_PARAMS.get(self.search_param, None)
+        qs = self.validate_query_params(request)
```
jezdez: technically that's not a `qs` (~query string) but query params, any chance to call that variable `validated_query_params` or `validated_params`?

kuma/search/filters.py
robhudson: Ooh yeah, good call. That will also remove the need for this mixin. I thought that was a bad code smell.

kuma/search/filters.py
robhudson: sure

robhudson:
Updated. Back to you @jezdez.

jezdez:
LGTM r+

jezdez:
tumblr_inline_njhdq1yxtk1raprkq

PR #3089, opened 2015-02-24
Fixes bug 1129119 - Convert to using elasticsearch aggregations

robhudson:


jezdez:
Wait, that's it?

robhudson:
Yep, that's it. The aggregations are similar to the facets and we were already using `post_filter` so there ended up not being much to it.

jezdez:
err, that was a r+ btw

PR #3090, opened 2015-02-24
fix bug 1136293 - move persona host to settings.py

groovecoder:
To spot-check using the Persona stage server, add this to your `settings_local.py` file:
```
PERSONA_VERIFIER_URL = 'https://verifier.login.anosrep.org/verify'
PERSONA_INCLUDE_URL = 'https://login.anosrep.org/include.js'
```
Then, when you sign in with Persona, make sure the `include.js` asset is coming from `login.anosrep.org`. When you sign in with Persona, you'll probably need to make a new account on the stage server. And if you use a gmail account, you'll be using the new OAuth bridge.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1261,6 +1261,8 @@ SOCIALACCOUNT_PROVIDERS = {
         }
     }
 }
+PERSONA_VERIFIER_URL = 'https://verifier.login.persona.org/verify'
+PERSONA_INCLUDE_URL = 'https://login.persona.org/include.js'
 
 # django-banish defaults; listing here to be explicit
 BANISH_ENABLED = False # TODO: https://bugzil.la/1126412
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -1261,6 +1261,9 @@ def get_user_url(user):
         }
     }
 }
+PERSONA_HOST = 'login.persona.org'
+PERSONA_VERIFIER_URL_PATH_FORMAT = 'https://verifier.%s/verify'
+PERSONA_INCLUDE_URL_PATH_FORMAT = 'https://%s/include.js'
```
jezdez: Let's just have two settings, PERSONA_VERIFIER_URL and PERSONA_INCLUDE_URL. That's easier to debug than using string substitution.

groovecoder:
Updated; and updated spot-check instructions too.

jezdez:
LGTM r+

openjck:
Also looking...

openjck:
Looks good to me, and the tests are now passing. Merging...

openjck:
Gotta say, the bridge is a nice improvement. The old way of authenticating with Google never made much sense.

PR #3094, opened 2015-02-26
Bug 1119397: Migrate straightforward dependencies to Bower

openjck:
Installed Bower packages are version-controlled as part of this change. Once we begin watching and automatically re-building compressed bundles, we can version-control the bundles instead and ignore the Bower packages.


groovecoder:
jQuery and html5shiv seem to be working fine.
On the home page I get:
```
TypeError: $list.owlCarousel is not a function home.js:7:0
```
bower-owlcarousel-typeerror
On a wiki page with prism, I get:
```
GET https://developer-local.allizom.org/media/js/syntax-prism-min.js?build=dev [HTTP/1.1 404 Not Found 1ms]
```
bower-404-syntax-prism-min

openjck:
Like the other work on this project, the code here can still be used, but we can avoid adding the libraries to the repository wholesale by switching to django-pipeline *first*. We'll come back to this later.

PR #3095, opened 2015-03-03
Bug 1137020: Track homepage search with Optimizely

openjck:
The easiest way to test this is to insert an `alert()` between lines 16 and 17. I can never seem to figure out how to test `trackEvent` locally, but I'm also not worried about it since it's just like other `trackEvent` calls.

### DIFF
--- a/media/redesign/js/home.js
+++ b/media/redesign/js/home.js
@@ -10,4 +10,12 @@
         $list.css('height', 'auto');
     }
 
+    // Track search submissions with Optimizely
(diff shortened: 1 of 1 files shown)


openjck:
Stephanie, you mentioned that you saw an Owl Carousel error on this branch. I can't seem to reproduce. Is it still happening for you?

openjck:
I opened #3098 to fix the Owl Carousel bug. The bug only manifests locally, so we don't necessarily need for #3098 to be merged before merging this.

PR #3097, opened 2015-03-04
Fix bug 1139670: Remove the maintenance page

openjck:
This is nice to have, and I honestly wish we used it more than we do.
It's much nicer than the standard black and white error message we
usually serve up when the site is down. It worked well, too, even though
IT had an unrelated hosting issue the last time they used it.

That said, we haven't been maintaining it (no pun intended) and we don't
have any maintenance planned. We can always use this code as inspiration
for future maintenance pages, but there's no good reason to let it
bitrot in the repository in the meantime.


groovecoder:
Can I call **DNM** on this until we figure out if there will be site downtime during the RabbitMQ physical-to-virtual process?

openjck:
Works for me. Happy to have a use for it. :smile:

groovecoder:
It looks like we **will** use this sometime soon. Should we close this PR? Or maybe change it to update the maintenance page?

openjck:
I'll make this a big rock for the week. When do we need the page to be ready?

groovecoder:
May have some down-time on Saturday. :frowning:

openjck:
Sounds good to me. That's plenty of time.

PR #3098, opened 2015-03-05
Fix bug 1139947: Fix homepage Owl Carousel error

openjck:
This bug only manifests locally. The "home" bundle is attached to the
homepage asynchronously. The scripts in the bundle still need to be
loaded in order, however.

On production, the scripts *are* loaded in order because the are
concatenated before the whole thing is attached asynchronously.

Locally, each script within the bundle is individually attached
asynchronously, so they're sometimes loaded out of order, causing the
error.

### DIFF
--- a/kuma/landing/templates/landing/homepage.html
+++ b/kuma/landing/templates/landing/homepage.html
@@ -181,7 +181,11 @@
 
 {% block js %}
   {{ super() }}
-  {{ js('home', async=True) }}
+  {% if settings.TEMPLATE_DEBUG %}
+      {{ js('home') }}
+  {% else %}
+      {{ js('home', async=True) }}
+  {% endif %}
   {% if waffle.flag('search_suggestions') %}
     {{ js('search-suggestions') }}
     <script>


openjck:
It might help to have @jezdez or @groovecoder review this, too, especially with regard to Django conventions.

stephaniehobson:
Error is gone. R+ from me. Tagging @jezdez to review the Django.

kuma/landing/views.py
```diff
@@ -33,6 +33,7 @@ def home(request):
         'updates': updates,
         'stats': community_stats,
         'current_challenge_tag_name': devderby_tag,
+        'template_debug': settings.TEMPLATE_DEBUG,
```
jezdez: No need to pass that into the template, the `settings` object is passed to all template contexts automatically via the `global_settings` template context processor (https://github.com/mozilla/kuma/blob/ae6c3d233c3503994f0d30671108b6d0d756effd/kuma/core/context_processors.py#L5-L7). Just refer to the setting in question as ``settings.TEMPLATE_DEBUG``.

jezdez:
@openjck No no, using the `TEMPLATE_DEBUG` setting is fine.

jezdez:
rwc+

openjck:
Thanks for review, @stephaniehobson. Great catch, @jezdez.

Updated! :smile:

PR #3100, opened 2015-03-05
Bug 1094452 - Convert render_document command to use chain

robhudson:


kuma/wiki/management/commands/render_document.py
```diff
@@ -2,100 +2,115 @@
...
-        make_option('--all', dest="all", default=False,
-                    action="store_true",
-                    help="Render ALL documents"),
-        make_option('--min-age', dest="min_age", default=600,
-                    help="Documents rendered less than this many seconds ago "
-                         "will be skipped"),
-        make_option('--baseurl', dest="baseurl",
+        make_option('--all', dest='all', default=False, action='store_true',
+                    help='Render ALL documents'),
+        make_option('--min-age', dest='min_age', default=600,
```
jezdez: When passing in an override value, will that be converted into int or does datetime.timedelta accept strings as well?

kuma/wiki/management/commands/render_document.py
```diff
@@ -2,100 +2,115 @@
...
-                         Document.objects.count())
-            docs = Document.objects.order_by('-modified').iterator()
-            for doc in docs:
-                self.do_render(doc)
+            # Query all documents, excluding those whose `last_rendered_at` is
+            # within `min_render_age` or NULL.
+            min_render_age = (
+                datetime.datetime.now() -
+                datetime.timedelta(seconds=self.options['min_age']))
+            docs = Document.objects.filter(
```
jezdez: Is there a model manager method or model method that returns this type of query already?

kuma/wiki/management/commands/render_document.py
```diff
@@ -2,100 +2,115 @@
...
-            render_age = now - doc.last_rendered_at
-            min_age = datetime.timedelta(seconds=self.options['min_age'])
-            if (render_age < min_age):
-                logging.debug(u"Skipping %s (%s) - rendered %s sec ago" %
-                              (doc, doc.get_absolute_url(), render_age))
-                return
-
-        if self.options['force']:
-            doc.render_started_at = None
+                doc = Document.objects.get(locale=locale, slug=slug)
```
jezdez: Could we factor out that query into a command method so that we don't have to do a query per passed arg? Something like `get_documents_for_args` or something?

kuma/wiki/management/commands/render_document.py
```diff
@@ -2,100 +2,115 @@
...
+                percent_complete = int(ceil((count / total) * 100))
+                tasks.append(
+                    email_render_document_progress.si(percent_complete,
+                                                      total))
+
+        # Add a 100% complete email task.
+        tasks.append(email_render_document_progress.si(100, total))
+
+        # Make it so.
+        chain(*tasks).apply_async()
```
jezdez: I think it'd be worthwhile to have a `link_error` handler passed to `apply_async` here to catch any errors while running through that chain, like mentioned here http://celery.readthedocs.org/en/latest/userguide/canvas.html#chains. We could send a mail to mdndev about it for each error, to be able to track the documents that don't render properly as part of the full render. Or is there some related feature already on the document level? @groovecoder

kuma/wiki/tasks.py
```diff
@@ -42,6 +42,19 @@ def render_document(pk, cache_control, base_url):
 
 
 @task
+def email_render_document_progress(percent_complete, total):
+    """
+    Task to send email for render_document progress notification.
+    """
+    subject = ('The command `render_document` is %s%% complete'
+               % percent_complete)
```
jezdez: Please put the `%` on the previous line

jezdez:
Looks great already, just a few minor comments

kuma/wiki/management/commands/render_document.py
robhudson: `TypeError: unsupported type for timedelta seconds component: str`

robhudson: I'll fix. Thanks.

kuma/wiki/management/commands/render_document.py
robhudson: I looked but didn't find anything.

kuma/wiki/management/commands/render_document.py
robhudson: That will be tricky b/c of the manipulation happening on the args to get them into the locale and slug, but I could try. If there are more than 1 would you want these chained as well (if `--defer` is provided)?

kuma/wiki/management/commands/render_document.py
robhudson: Inside the `document_render` task I can wrap the actual render call in a `try/except` block. But, I'd need to know what exceptions we want to retry the task and what exceptions we want to log or email?

kuma/wiki/management/commands/render_document.py
jezdez: Ok.

kuma/wiki/management/commands/render_document.py
jezdez: Not sure if it's worth the hassle tbh

kuma/wiki/management/commands/render_document.py
jezdez: Hmm, any kind of "didn't render" would match, so anything except that?

robhudson: I'm having trouble understanding your comment, sorry. Any kind of "didn't render" should send an email? The rest should retry the task?

jezdez: Sorry, I mean there are cases in which we want to retry the task like if kumascript is down, I think there is an exception raised for that. We should catch it and do the retry. For other errors we should just do a blank exceptin handling and send the email, right?

robhudson:
Hi @jezdez. I updated this to use chunking to avoid a recursion error with pickle/celery and added exception catching.

jezdez:
LGTM

groovecoder:
There's another commit here without a reference to the bug #. Should we add something to our Travis script that fails on commits without bug #'s?

robhudson:
> There's another commit here without a reference to the bug #.

I separated it b/c it wasn't related to the code change for the bug. Unless there's a general pep8 cleanup bug already then I should have used that.

> Should we add something to our Travis script that fails on commits without bug #'s?

If that's your policy but it seems overkill to me. I think it's ok if commits have no bug number for things like pep8 or typos or things like that.

kuma/wiki/management/commands/render_document.py
```diff
@@ -2,100 +2,112 @@
...
-                return
-
-        if self.options['force']:
-            doc.render_started_at = None
+                doc = Document.objects.get(locale=locale, slug=slug)
+                if self.options['force']:
+                    doc.render_started_at = None
+                log.info(u'Rendering %s (%s)' % (doc, doc.get_absolute_url()))
+                try:
+                    render_document(doc, self.cache_control, self.base_url)
```
groovecoder: This gave me:
```
...
  File "/data/www/developer.mozilla.org/kuma/kuma/wiki/management/commands/render_document.py", line 90, in handle
    render_document(doc, self.cache_control, self.base_url)
...
  File "/data/www/developer.mozilla.org/kuma/kuma/wiki/tasks.py", line 39, in render_document
    document = Document.objects.get(pk=pk)
...
TypeError: int() argument must be a string or a number, not 'Document'
```
When I ran:
```
./manage.py render_document en-US/docs/User:groovecoder
```
Looks like we need to call it with `doc.pk` like the old `do_render` method did for us?

robhudson: Woops, good catch.

groovecoder:
I generally say commit changes against the bug # you're working on when you
make them. I.e., not every change needs its own bug.

In this case, "bug 1094452 - PEP8 in wiki/management/"

-L

PR #3101, opened 2015-03-06
Fix Bug 1139883: New styles for .spec-RC

stephaniehobson:
Matching .spec-CR except for letters.


groovecoder:
Assigning @openjck

PR #3103, opened 2015-03-06
bug 1113260 - use django-ratelimit for ip bans

groovecoder:
**DNM** Sanity-check PR. Here's the approach so far:

* Adding a new `IPBan` model to `kuma.core` (like `UserBan` in `kuma.users`)
* Decorate views with `ratelimit` of a new `limit_banned_ip_to_0` util function

Pro: Simpler code that uses `django-ratelimit` (a well-supported popular library) as its designed

Con: Have to decorate each view (or add mixin to CBV's)

Overall I'm liking this approach more than `django-banish` or any others we've tried so far.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -148,6 +148,3 @@
 [submodule "vendor/src/django-ratelimit"]
 	path = vendor/src/django-ratelimit
 	url = git://github.com/jsocol/django-ratelimit.git
-[submodule "vendor/src/django-cacheback"]
-    path = vendor/src/django-cachebackk
-    url = https://github.com/codeinthehole/django-cacheback.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -148,3 +148,6 @@
 [submodule "vendor/src/django-ratelimit"]
 	path = vendor/src/django-ratelimit
 	url = git://github.com/jsocol/django-ratelimit.git
+[submodule "vendor/src/django-cacheback"]
+    path = vendor/src/django-cachebackk
+    url = https://github.com/codeinthehole/django-cacheback.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -136,9 +136,6 @@
 [submodule "vendor/src/django-celery"]
 	path = vendor/src/django-celery
 	url = git://github.com/celery/django-celery.git
-[submodule "vendor/src/django-banish"]
-	path = vendor/src/django-banish
-	url = git://github.com/yourabi/django-banish.git
 [submodule "vendor/src/django-honeypot"]
 	path = vendor/src/django-honeypot
 	url = git://github.com/sunlightlabs/django-honeypot.git
@@ -148,3 +145,6 @@
 [submodule "vendor/src/django-cacheback"]
 	path = vendor/src/django-cacheback
 	url = https://github.com/codeinthehole/django-cacheback.git
+[submodule "vendor/src/django-ratelimit"]
+	path = vendor/src/django-ratelimit
+	url = git://github.com/jsocol/django-ratelimit.git
--- a/kuma/contentflagging/utils.py
+++ b/kuma/contentflagging/utils.py
@@ -1,40 +1,6 @@
-import re
 import hashlib
 
-# this is not intended to be an all-knowing IP address regex
-IP_RE = re.compile('\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}')
-
-
-def get_ip(request):
-    """
-    Retrieves the remote IP address from the request data.  If the user is
-    behind a proxy, they may have a comma-separated list of IP addresses, so
-    we need to account for that.  In such a case, only the first IP in the
-    list will be retrieved.  Also, some hosts that use a proxy will put the
-    REMOTE_ADDR into HTTP_X_FORWARDED_FOR.  This will handle pulling back the
-    IP from the proper place.
-
-    **NOTE** This function was taken from django-tracking (MIT LICENSE)
-             http://code.google.com/p/django-tracking/
-    """
-
-    # if neither header contain a value, just use local loopback
-    ip_address = request.META.get('HTTP_X_FORWARDED_FOR',
-                                  request.META.get('REMOTE_ADDR', '127.0.0.1'))
-    if ip_address:
-        # make sure we have one and only one IP
-        try:
-            ip_address = IP_RE.match(ip_address)
-            if ip_address:
-                ip_address = ip_address.group(0)
-            else:
-                # no IP, probably from some dirty proxy or other device
-                # throw in some bogus IP
-                ip_address = '10.0.0.1'
-        except IndexError:
-            pass
-
-    return ip_address
+from kuma.core.utils import get_ip
 
 
 def get_unique(content_type, object_pk, request=None, ip=None, user_agent=None, user=None):
--- a/kuma/core/jobs.py
+++ b/kuma/core/jobs.py
@@ -31,4 +31,4 @@ class IPBanJob(KumaJob):
         return "60/m"
 
     def empty(self):
-        return []
+        return "60/m"
--- a/kuma/core/jobs.py
+++ b/kuma/core/jobs.py
@@ -20,7 +20,7 @@ class KumaJob(Job):
         return hashlib.md5(six.b(':').join(value)).hexdigest()
 
 
-class IPBanJob(Job):
+class IPBanJob(KumaJob):
     lifetime = 60 * 60 * 3
     refresh_timeout = 60
 
@@ -29,11 +29,6 @@ class IPBanJob(Job):
         if IPBan.objects.active(ip=ip).exists():
             return "0/s"
         return "60/m"
-
-    def key(self, ip):
-        # override the default way to make sure we handle unicode,
-        # bytestring and integer versions of the pk the same
-        return 'kuma:core:ipban:%s' % ip
 
     def empty(self):
         return []
--- a/kuma/core/jobs.py
+++ b/kuma/core/jobs.py
@@ -18,3 +18,22 @@ class KumaJob(Job):
         if isinstance(value, tuple):
             value = tuple(to_bytestring(v) for v in value)
         return hashlib.md5(six.b(':').join(value)).hexdigest()
+
+
+class IPBanJob(Job):
+    lifetime = 60 * 60 * 3
+    refresh_timeout = 60
+
+    def fetch(self, ip):
+        from .models import IPBan
+        if IPBan.objects.active(ip=ip).exists():
+            return "0/s"
+        return "60/m"
+
+    def key(self, ip):
+        # override the default way to make sure we handle unicode,
+        # bytestring and integer versions of the pk the same
+        return 'kuma:core:ipban:%s' % ip
+
+    def empty(self):
+        return []
--- a/kuma/core/managers.py
+++ b/kuma/core/managers.py
@@ -7,9 +7,11 @@ TODO:
 - Permissions for tag namespaces (eg. system:* is superuser-only)
 - Machine tag assists
 """
+from datetime import date, timedelta
 import operator
-from django.db import router
 
+from django.db import router
+from django.db import models
 from django.db.models.fields import BLANK_CHOICE_DASH
 from django.contrib.auth.models import AnonymousUser
 
@@ -239,6 +241,16 @@ def resolve_allowed_tags(model_obj, tags_curr, tags_new,
             tags_out.extend(ns_tags_curr[ns])
 
     return tags_out
+
+
+class IPBanManager(models.Manager):
+    def active(self, ip):
+        return super(IPBanManager, self).filter(ip=ip, deleted__isnull=True)
+
+    def delete_old(self, days=30):
+        cutoff_date = date.today() - timedelta(days=days)
+        old_ip_bans = self.get_query_set().filter(created__lte=cutoff_date)
+        old_ip_bans.delete()
 
 
 # Tell South to ignore our fields, if present.
--- a/kuma/core/managers.py
+++ b/kuma/core/managers.py
@@ -245,11 +245,11 @@ def resolve_allowed_tags(model_obj, tags_curr, tags_new,
 
 class IPBanManager(models.Manager):
     def active(self, ip):
-        return super(IPBanManager, self).filter(ip=ip, deleted__isnull=True)
+        return self.filter(ip=ip, deleted__isnull=True)
 
     def delete_old(self, days=30):
         cutoff_date = date.today() - timedelta(days=days)
-        old_ip_bans = self.get_query_set().filter(created__lte=cutoff_date)
+        old_ip_bans = self.filter(created__lte=cutoff_date)
         old_ip_bans.delete()
 
 
--- a/kuma/core/models.py
+++ b/kuma/core/models.py
@@ -48,7 +48,7 @@ class IPBan(models.Model):
     def delete(self, *args, **kwargs):
         self.deleted = timezone.now()
         self.save()
-        IPBanJob.delete(self.ip)
+        IPBanJob().invalidate(self.ip)
 
     def __unicode__(self):
         return u'%s banned on %s' % (self.ip, self.created)
--- a/kuma/core/models.py
+++ b/kuma/core/models.py
@@ -1,11 +1,10 @@
-from datetime import datetime
-
 from django.db import models
 from django.utils import timezone
 
 from south.modelsinspector import add_introspection_rules
 
 from .managers import IPBanManager
+from .jobs import IPBanJob
 
 
 class ModelBase(models.Model):
@@ -49,6 +48,7 @@ class IPBan(models.Model):
     def delete(self, *args, **kwargs):
         self.deleted = timezone.now()
         self.save()
+        IPBanJob.delete(self.ip)
 
     def __unicode__(self):
         return u'%s banned on %s' % (self.ip, self.created)
--- a/kuma/core/models.py
+++ b/kuma/core/models.py
@@ -1,6 +1,9 @@
+from datetime import datetime
 from django.db import models
 
 from south.modelsinspector import add_introspection_rules
+
+from .managers import IPBanManager
 
 
 class ModelBase(models.Model):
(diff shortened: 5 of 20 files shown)


groovecoder:
Can I get a sanity check from @jezdez, @ubernostrum, @robhudson ... or even @jsocol? :wink:

kuma/core/utils.py
```diff
@@ -284,3 +286,11 @@ def chord_flow(pre_task, tasks, post_task):
         return chain(*tasks)
     else:
         return chain(pre_task, chord(header=tasks, body=post_task))
+
+
+def limit_banned_ip_to_0(group, request):
+    try:
+        IPBan.objects.get(ip=get_ip(request))
```
jsocol: If this ever gets applied to more-frequent endpoints, I'd put some layer of caching here.

jsocol:
This is exactly the kind of thing I hoped people would build when I made the decorator accept callables!

Middleware might be more effective if you are willing to be coarser (i.e. block _all_ requests, or all unsafe HTTP methods, from a given IP) in exchange for writing way less code. That said, putting things that can fail (like DB/cache lookups) in middleware is discouraged because the error handling is rough, so it's important to think about what those failure cases look like.

The nice thing about doing this at the application layer is that the application can control things relatively easily. The downside is that the application still consumes (some) resources serving banned users, so if you are willing to be coarser, something at a higher level, like fail2ban, may also be an option.

Finally, I made some specific comments, because I miss you folks.

kuma/core/managers.py
```diff
@@ -241,6 +243,13 @@ def resolve_allowed_tags(model_obj, tags_curr, tags_new,
     return tags_out
 
 
+class IPBanManager(models.Manager):
+    def delete_old(self, days=30):
```
jsocol: IP addresses aren't _that_ stable, 30 days seems high as a default.

kuma/core/managers.py
```diff
@@ -241,6 +243,13 @@ def resolve_allowed_tags(model_obj, tags_curr, tags_new,
...
+        cutoff_date = date.today() - timedelta(days=days)
+        old_ip_bans = self.get_query_set().filter(created__lte=cutoff_date)
+        old_ip_bans.delete()
```
jsocol: I've taken to adding `deleted = models.DateTimeField(null=True,blank=True)` instead of actually deleting things, so there's some history. Then instead of actually deleting, you can make the model's/manager's `delete()` method just set the `deleted` attribute, and since you have a manager here anyway, maybe make an `active()` that does `super().filter(deleted__isnull=True)`.

kuma/core/models.py
```diff
@@ -34,6 +37,14 @@ def update(self, **kw):
                                           created=False)
 
 
+class IPBan(models.Model):
+    ip = models.CharField(max_length=40, editable=False, db_index=True,
+                          blank=True, null=True, unique=True)
```
robhudson: Would using https://docs.djangoproject.com/en/1.7/ref/models/fields/#genericipaddressfield have any benefit?

kuma/core/models.py
```diff
@@ -34,6 +37,14 @@ def update(self, **kw):
...
+    created = models.DateTimeField(default=datetime.now, db_index=True)
```
jsocol: Re: the caching comment, I've been doing something like:

```python
@classmethod
def get(cls, some_natural_key):
    obj = cache.get(cls._cache_key(some_natural_key))
    if obj == EMPTY_SENTINEL:  # e.g. '-'
        raise cls.DoesNotExist  # or return None
    # ... etc
```

IP address, and `deleted__isnull` make a pretty natural key here.

groovecoder:
Updated to fully functional code. I'm sticking with the single decorator because this is first and foremost meant to implement https://bugzilla.mozilla.org/show_bug.cgi?id=1113260. We may add middleware when we go on to implement automatic banning & limiting in https://bugzilla.mozilla.org/show_bug.cgi?id=1122658.

Still needs tests ... that's next.

groovecoder:
Okay, this is ready with tests and everything. An outstanding question for @jezdez @ubernostrum - how should we clean out the db tables from `django-banish`? Data migration in `kuma.core`? Or manually?

kuma/core/managers.py
```diff
@@ -241,6 +243,16 @@ def resolve_allowed_tags(model_obj, tags_curr, tags_new,
     return tags_out
 
 
+class IPBanManager(models.Manager):
+    def active(self, ip):
+        return super(IPBanManager, self).filter(ip=ip, deleted__isnull=True)
+
+    def delete_old(self, days=30):
+        cutoff_date = date.today() - timedelta(days=days)
+        old_ip_bans = self.get_query_set().filter(created__lte=cutoff_date)
```
jezdez: `self.filter(..)` is enough, the call to `get_query_set` is implied when using `self`

kuma/core/managers.py
```diff
@@ -241,6 +243,16 @@ def resolve_allowed_tags(model_obj, tags_curr, tags_new,
     return tags_out
 
 
+class IPBanManager(models.Manager):
+    def active(self, ip):
+        return super(IPBanManager, self).filter(ip=ip, deleted__isnull=True)
```
jezdez: Just use `self.filter` here, no need to use `super`

kuma/core/models.py
```diff
@@ -34,6 +37,21 @@ def update(self, **kw):
                                           created=False)
 
 
+class IPBan(models.Model):
+    ip = models.GenericIPAddressField()
+    created = models.DateTimeField(default=datetime.now, db_index=True)
```
jezdez: Let's use `django.utils.timezone.now` for the default in preparation to make kuma timezone aware (a feature in current Django)

kuma/core/models.py
```diff
@@ -34,6 +37,21 @@ def update(self, **kw):
...
+    deleted = models.DateTimeField(null=True, blank=True)
+
+    objects = IPBanManager()
+
+    def delete(self, *args, **kwargs):
+        self.deleted = datetime.now()
```
jezdez: Use `timezone.now` as well

kuma/core/models.py
```diff
@@ -34,6 +37,21 @@ def update(self, **kw):
...
+        self.save()
```
jezdez: This mean's we can never delete that db entry, which is not a good pattern. There are apps out there implementing that logical delete pattern much better, I'd strongly recommend using that instead.

kuma/core/tasks.py
```diff
@@ -51,3 +52,8 @@ def clean_sessions():
     else:
         logger.error('The clean_sessions task is already running since %s' %
                      memcache.get(LOCK_ID))
+
+
+@task
+def delete_old_ip_bans(immediate=False, days=30):
```
jezdez: Unused `immediate` param

kuma/core/tasks.py
```diff
@@ -7,6 +7,7 @@
 import constance.config
 
 from kuma.core.cache import memcache
+from .models import IPBan
```
jezdez: Please add empty line before local import.

kuma/core/tests/test_models.py
```diff
@@ -0,0 +1,36 @@
+from datetime import date, timedelta
+
+from nose.tools import eq_
+
+from django.test import TestCase
+
+from kuma.core.models import IPBan
```
jezdez: Please use local import, `from ..models import IPBan`

kuma/core/tests/test_models.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class RevisionIPTests(TestCase):
+    def test_delete_older_than_default_30_days(self):
+        old_date = date.today() - timedelta(days=31)
+        IPBan.objects.create(ip='127.0.0.1', created=old_date).save()
```
jezdez: `create` calls `save` automatically, no need for duplication

kuma/core/tests/test_models.py
```diff
@@ -0,0 +1,36 @@
...
+        eq_(1, IPBan.objects.all().count())
+        IPBan.objects.delete_old()
+        eq_(0, IPBan.objects.all().count())
+
+    def test_delete_older_than_days_argument(self):
+        ban_date = date.today() - timedelta(days=5)
+        IPBan(ip='127.0.0.1', created=ban_date).save()
```
jezdez: either use `objects.create` or `save`

kuma/core/tests/test_models.py
```diff
@@ -0,0 +1,36 @@
...
+from django.test import TestCase
+
+from kuma.core.models import IPBan
+
+
+class RevisionIPTests(TestCase):
+    def test_delete_older_than_default_30_days(self):
+        old_date = date.today() - timedelta(days=31)
+        IPBan.objects.create(ip='127.0.0.1', created=old_date).save()
+        eq_(1, IPBan.objects.all().count())
```
jezdez: `all()` is not needed here, it's implied when running `count()`, also happens in lines below

kuma/core/utils.py
```diff
@@ -284,3 +286,9 @@ def chord_flow(pre_task, tasks, post_task):
         return chain(*tasks)
     else:
         return chain(pre_task, chord(header=tasks, body=post_task))
+
+
+def limit_banned_ip_to_0(group, request):
+    if IPBan.objects.active(ip=get_ip(request)).count() > 0:
```
jezdez: Please use `exists` instead of `count() > 0` since `exists` has a slightly improved internal cache

kuma/wiki/tests/test_views.py
```diff
@@ -547,6 +548,34 @@ def test_banned_users(self):
...
+class BannedIPTests(UserTestCase, WikiTestCase):
+    """Tests readonly scenarios"""
+    fixtures = UserTestCase.fixtures + ['wiki/documents.json']
+    localizing_client = True
+
+    def setUp(self):
+        super(BannedIPTests, self).setUp()
+        self.ip = '127.0.0.1'
+        self.ip_ban = IPBan.objects.create(ip=self.ip)
+        self.d, r = doc_rev()
```
jezdez: please don't use single character variable names

kuma/wiki/tests/test_views.py
```diff
@@ -547,6 +548,34 @@ def test_banned_users(self):
...
+        self.edit_url = reverse('wiki.edit_document', args=[self.d.full_path])
+
+    def test_banned_ip_cant_get_edit(self):
+        self.client.login(username='testuser', password='testpass')
+        resp = self.client.get(self.edit_url, REMOTE_ADDR=self.ip)
```
jezdez: s/resp/response/g

jezdez: as well as below

kuma/wiki/tests/test_views.py
```diff
@@ -547,6 +548,34 @@ def test_banned_users(self):
...
+        eq_(403, resp.status_code)
+
+    def test_banned_ip_cant_post_edit(self):
+        self.client.login(username='testuser', password='testpass')
+        resp = self.client.get(self.edit_url, REMOTE_ADDR=self.ip)
+        eq_(403, resp.status_code)
+
+    def test_banned_ip_can_still_get_articles(self):
+        resp = self.client.get(reverse('wiki.document',
```
jezdez: Please use `self.document.get_absolute_url`

kuma/wiki/tests/test_views.py
```diff
@@ -2354,6 +2383,9 @@ def test_store_revision_ip(self):
         rev_ip = RevisionIP.objects.get(revision=rev)
         eq_('127.0.0.1', rev_ip.ip)
 
+    def test_banned_ip(self):
```
jezdez: Something you want to fill?

kuma/core/utils.py
```diff
@@ -284,3 +286,9 @@ def chord_flow(pre_task, tasks, post_task):
...
+        return "0/s"
+    return "60/m"
```
jezdez: This *definitely* needs caching, no way around it. This is going to be called *a lot* and no matter how optimized the MySQL cluster is, counts and checks are slow. Merge #3115 and add a django-cacheback job to handle the cache invalidation nicely.

jezdez:
Ok, I think we're on the right track, but this needs some updates, mostly code smell but one biggy: caching. See #3115 for a new way to handle cache refills via Celery nicely that should be used here as well.

kuma/core/models.py
groovecoder: The 2 simple pluggable apps I've found so far are:

https://github.com/paltman/django-logicaldelete
https://github.com/hearsaycorp/django-livefield

[Others](https://www.djangopackages.com/grids/g/deletion/) seem like overkill. I'll check them out next week.

groovecoder:
Updated with all the smaller updates. Going to check out https://github.com/mozilla/kuma/pull/3115 caching strategy before adding the caching here.

groovecoder:
Added https://github.com/groovecoder/kuma/commit/53b749611daef03f68ca2e9c730e57aec55a11ac with a `django-cacheback` `IPBanJob`. It also includes some of the same changes in #3115, so we should finish and merge that first. Then I can rebase this PR.

kuma/core/jobs.py
```diff
@@ -18,3 +18,17 @@ def hash(self, value):
...
+    refresh_timeout = 60
+
+    def fetch(self, ip):
+        from .models import IPBan
+        if IPBan.objects.active(ip=ip).exists():
+            return "0/s"
+        return "60/m"
+
+    def empty(self):
+        return []
```
jezdez: I don't think it should return a list here since you're expecting a string in fetch. the docstring says:

        Return the appropriate value for a cache MISS (and when we defer the
        repopulation of the cache)

So probably something like `"60/m"` as well?

kuma/core/models.py
```diff
@@ -34,6 +38,22 @@ def update(self, **kw):
...
+    ip = models.GenericIPAddressField()
+    created = models.DateTimeField(default=timezone.now, db_index=True)
+    deleted = models.DateTimeField(null=True, blank=True)
+
+    objects = IPBanManager()
+
+    def delete(self, *args, **kwargs):
+        self.deleted = timezone.now()
+        self.save()
+        IPBanJob.delete(self.ip)
```
jezdez: This needs to be using an instance of a job, so `IPBanJob().delete(self.ip)`. That said, I think we should use the `refresh` method since that not only deletes the cache entry but fills the cache synchronously again right away and doesn't wait till the next request to fill it.

kuma/core/utils.py
```diff
@@ -19,7 +19,9 @@
 
 from taggit.utils import split_strip
 
+from kuma.actioncounters.utils import get_ip
```
jezdez: Let's move that function into core.

kuma/wiki/tests/test_views.py
```diff
@@ -547,6 +548,39 @@ def test_banned_users(self):
...
+        self.doc, rev = doc_rev()
+        self.edit_url = reverse('wiki.edit_document',
+                                args=[self.doc.full_path]
+                               )
+
+    def test_banned_ip_cant_get_edit(self):
+        self.client.login(username='testuser', password='testpass')
+        response = self.client.get(self.edit_url, REMOTE_ADDR=self.ip)
+        eq_(403, response.status_code)
+        cache.clear()
```
jezdez: Move the call to `cache.clear` into the `setUp` method instead.

groovecoder:
Updated.

I moved `cache.clear` to `tearDown` because it's cleaning up the cache for tests **following** the `BannedIPTests` tests - including tests in other classes.

.gitmodules
```diff
@@ -148,3 +145,9 @@
 [submodule "vendor/src/django-cacheback"]
 	path = vendor/src/django-cacheback
 	url = https://github.com/codeinthehole/django-cacheback.git
+[submodule "vendor/src/django-ratelimit"]
+	path = vendor/src/django-ratelimit
+	url = git://github.com/jsocol/django-ratelimit.git
+[submodule "vendor/src/django-cacheback"]
```
jezdez: Hm, why is that adding that twice?

kuma/core/models.py
```diff
@@ -34,6 +38,22 @@ def update(self, **kw):
...
 
+class IPBan(models.Model):
+    ip = models.GenericIPAddressField()
+    created = models.DateTimeField(default=timezone.now, db_index=True)
+    deleted = models.DateTimeField(null=True, blank=True)
+
+    objects = IPBanManager()
+
+    def delete(self, *args, **kwargs):
+        self.deleted = timezone.now()
```
jezdez: What's the verdict about logicial deletes btw?

jezdez: I mean apps.

kuma/wiki/tests/test_views.py
```diff
@@ -547,6 +548,40 @@ def test_banned_users(self):
...
+    localizing_client = True
+
+    def setUp(self):
+        super(BannedIPTests, self).setUp()
+        self.ip = '127.0.0.1'
+        self.ip_ban = IPBan.objects.create(ip=self.ip)
+        self.doc, rev = doc_rev()
+        self.edit_url = reverse('wiki.edit_document',
+                                args=[self.doc.full_path]
+                               )
```
jezdez: Move the closing bracket up one line

kuma/wiki/tests/test_views.py
```diff
@@ -547,6 +548,40 @@ def test_banned_users(self):
...
+
+    def test_banned_ip_cant_post_edit(self):
+        self.client.login(username='testuser', password='testpass')
+        response = self.client.get(self.edit_url, REMOTE_ADDR=self.ip)
+        eq_(403, response.status_code)
+
+    def test_banned_ip_can_still_get_articles(self):
+        response = self.client.get(self.doc.get_absolute_url(),
+                                   REMOTE_ADDR=self.ip
+                                  )
```
jezdez: Bracket up one line

jezdez:
LGTM other those last minor issues r+

kuma/core/models.py
groovecoder: All the ones I saw in https://www.djangopackages.com/grids/g/deletion/ looked either too heavyweight or unstable/un-maintained. :frowning: Unless one of them looks right to you?

groovecoder:
Updated. Waiting on Travis ...

kuma/core/models.py
jezdez: Hm, I haven't looked for real. If that's not the way you wanna go now (we can do that later (soon)), can you add the deleted field to the admin list display?

groovecoder:
Added the custom `list_display` for the admin site.

groovecoder:
Read for final review & merge ...

kuma/contentflagging/utils.py
```diff
@@ -1,40 +1,6 @@
...
-                ip_address = ip_address.group(0)
-            else:
-                # no IP, probably from some dirty proxy or other device
-                # throw in some bogus IP
-                ip_address = '10.0.0.1'
-        except IndexError:
-            pass
-
-    return ip_address
+from kuma.core.utils import get_ip
```
jezdez: Yay for code deduping!

jezdez:
r+

PR #3104, opened 2015-03-06
fix bug 1140436: add period url param & select box to revisions dashboard

jamonation:
I think that this needs some i18n polish (because I don't know how that system works in Kuma), but it filters by hour/day/week/month per @a2sheppy's feature request.

### DIFF
--- a/kuma/dashboards/forms.py
+++ b/kuma/dashboards/forms.py
@@ -6,7 +6,12 @@ from kuma.core.form_fields import StrippedCharField
 
 
 LANG_CHOICES = [('', _('All Locales'))] + settings.LANGUAGES
-
+PERIOD_CHOICES = [('', _('None')),
+                  ('hour', _('Hour')),
+                  ('day', _('Day')),
+                  ('week',_('Week')),
+                  ('month',_('30 days')
+               )]
 
 class RevisionDashboardForm(forms.Form):
 
@@ -31,3 +36,8 @@ class RevisionDashboardForm(forms.Form):
     end_date = forms.DateField(required=False, label=_(u'End Date:'),
                     input_formats=['%m/%d/%Y'],
                     widget=forms.TextInput(attrs={'pattern': '\d{1,2}/\d{1,2}/\d{4}'}))
+
+    preceding_period = forms.ChoiceField(choices=PERIOD_CHOICES,
+                    required=False,
+                    label=_(u'Preceding Period:'))
+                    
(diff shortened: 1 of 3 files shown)


kuma/dashboards/forms.py
```diff
@@ -6,7 +6,7 @@
 
 
 LANG_CHOICES = [('', _('All Locales'))] + settings.LANGUAGES
-
+PERIOD_CHOICES = (('', 'None'), ('hour','Hour'), ('day','Day'), ('week','Week'), ('month','30 days'))
```
groovecoder: The human-readable option labels need the tower gettext_lazy `_` wrapper. Like the `LANG_CHOICES` tuple above.

jamonation:
Ah that's what I was missing, thanks @groovecoder - I've added the _ wrapper, and switched from a tuple to a list of tuples to be consistent with the LANG_CHOICES above.

kuma/dashboards/forms.py
```diff
@@ -6,7 +6,7 @@
 
 
 LANG_CHOICES = [('', _('All Locales'))] + settings.LANGUAGES
-
+PERIOD_CHOICES = [('', _('None')), ('hour', _('Hour')), ('day', _('Day')), ('week',_('Week')), ('month',_('30 days'))]
```
jezdez: Mind putting new lines after each element?

kuma/dashboards/views.py
```diff
@@ -48,6 +48,24 @@ def revisions(request):
...
+            if preceding_period == 'week':
+                seconds = 7 * 24 * 60 * 60
+            if preceding_period == 'day':
+                seconds = 24 * 60 * 60
+            if preceding_period == 'hour':
+                seconds = 60 * 60
+            # use the form date if present, otherwise, offset from now
+            end_date = (filter_form.cleaned_data['end_date'] or
+                        datetime.datetime.now())
+            start_date = end_date - datetime.timedelta(seconds=seconds)
```
jezdez: Please use `django.utils.timezone.now()` here instead in preparation for enabling Django's timezone support.

jamonation:
Thanks for the review @jezdez - I've incorporated your suggestions.

A note: this is the wrong branch name - I can open another pull request if that would make things easier than editing merge commit messages.

Last thing - would you like me to make a separate bug & branch for doing timezone migrations from datetime to django.utils version, or is there one already that I could take on as a task?

groovecoder:
Spot-check looks good. Up to you if you want to open another PR or amend the commit message on this one.

A new bug in the "Code Cleanup" component would be great.

jamonation:
Sure, I've amended the message. Just waiting on Travis and then we should be all set?

PR #3105, opened 2015-03-06
Fix Bug 931377: Add a class to hide when viewing article but show in editor

stephaniehobson:
Altered current .hidden class to display in editor and have a dotted black border around it and floating label.

This page already has some instances of .hidden on it and it helpful for testing: https://developer-local.allizom.org/en-US/docs/Web/API/Canvas_API/Tutorial/Basic_animations

I had trouble getting the editor to pull the newly generated stylesheets. Sometimes running `./manage.py compress_assets` helped.


openjck:
Awesome! Looks great. :tada:

PR #3109, opened 2015-03-10
Fix bug 1130383 - Get rid of test-utils.

jezdez:


robhudson:
Don't forget settings.py... settings.INSTALLED apps and settings.TEST_RUNNER.

robhudson:
r+ when tests pass.

ubernostrum:
Should have sent... a poet...

jodileansm

manage.py
```diff
@@ -24,15 +24,24 @@ def path(*parts):
...
+        try:
+            import settings_local  # noqa
+        except ImportError:
+            # or use the plain settings.py module
+            settings_mod = 'settings'
+        else:
+            settings_mod = 'settings_local'
+
+    # override the env var with what we want
+    os.environ['DJANGO_SETTINGS_MODULE'] = settings_mod
```
robhudson: Are you happy with this?

settings_travis.py
```diff
@@ -1,10 +1,4 @@
-from settings import *
+from settings_test import *
 
-CELERY_ALWAYS_EAGER = True
-ASYNC_SIGNALS = False
-ES_URLS = ['localhost:9200']
-ES_INDEX_PREFIX = 'mdn'
-ES_INDEXES = {'default': 'main_index'}
-ES_INDEXING_TIMEOUT = 30
```
robhudson: The `ES_LIVE_INDEX` is now different. That ok?

tox.ini
```diff
@@ -7,7 +7,7 @@ deps=
     -rrequirements/compiled.txt
     -rrequirements/dev.txt
 commands =
-    coverage run manage.py test --noinput -v2 kuma
+    coverage run manage.py test --noinput -v2 {posargs:kuma}
     coverage report
 setenv =
     FORCE_DB = 1
```
robhudson: I think this `FORCE_DB` can go away now. That's a test-utils only thing iirc. The nose test runner has something equivalent but it's different... you have to tell it to reuse the db explicitly.

robhudson:
Back to you with questions and one minor fix.

manage.py
jezdez: Hahaha, "happy". No

settings_travis.py
jezdez: We want to test against ES on Travis, yeah

tox.ini
jezdez: Not the nose test runner from 2010 or 2011 we're using :'(

jezdez: And yeah I'll remove FORCE_DB

jezdez: Or should I say FARCE_DB

jezdez: Sigh

groovecoder:
To run the tests in my vm, I had to copy the `DATABASES` setting from `settings_local.py` into `settings_test.py` now that we only load one or the other in `manage.py`? Is that the right way to run locally, or do we need to update test instructions somehow?

groovecoder:
ping @jezdez - what do you think? Should we copy the `DATABASES` setting into `settings_test.py` so dev environments can run the tests out-of-the-box?

jezdez:
@groovecoder No, we should update the `settings.py` to have the database setting that we expect for local development including testing, I don't understand why we even have an "empty" `DATABASES` setting in there that won't work anyway. That means that the Travis settings need an overridden `DATABASES` setting using the root user following http://docs.travis-ci.com/user/database-setup/#MySQL

groovecoder:
Makes sense. So you'll fill in the `DATABASES` setting in `settings.py` as part of this PR? Or is that a step for another PR?

jezdez:
@groovecoder Part of this PR

groovecoder:
I checked this out and it works locally! Just needs a rebase and then it's good to merge?

jezdez:
@groovecoder Yeah, IMO

PR #3110, opened 2015-03-10
Fix bug 1141314: Update the maintenance page

openjck:
To test this locally:

1. Run `python -m SimpleHTTPServer` in the host's `maintenance` directory
2. Visit <a href="http://localhost:8000">http://localhost:8000</a>


openjck:
Ah, I almost forgot. One nice thing about this page is that it doesn't need Vagrant to be tested. Running the `python` command in the VM or even locally should work.

groovecoder:
We need to merge & push this today so that WebOps can put it up for tomorrow's down-time. @stephaniehobson - okay if I review & merge this myself?

PR #3115, opened 2015-03-13
Cache document zone stack and simplify caching strategy in conjunction with Celery.

jezdez:


kuma/wiki/jobs.py
```diff
@@ -0,0 +1,56 @@
...
+    def fetch(self, locale):
+        from .models import DocumentZone
+        zones = (DocumentZone.objects.filter(document__locale=locale,
+                                             url_root__isnull=False)
+                                     .exclude(url_root=''))
+        remaps = [('/docs/%s' % zone.document.slug, '/%s' % zone.url_root)
+                  for zone in zones]
+        return remaps
+
+    def key(self, locale):
```
groovecoder: Should we make a `KumaJob(Job)` super-class to hold some of the common `key` and `empty` logic? Is there enough identical behavior to make us want to re-use them?

kuma/wiki/models.py
```diff
@@ -333,31 +340,40 @@ def regenerate_cache_with_fields(self):
...
-                .extractSection(section_id, ignore_heading=ignore_heading)
-                .serialize())
+        """
+        Convenience method to extract the rendered content for a single section
+        """
+        if self.rendered_html:
+            content = self.rendered_html
+        else:
+            content = self.html
+        return self.extract_section(content, section_id, ignore_heading)
```
groovecoder: :+1: for more readable code

groovecoder: oh, **and** for re-using code!

kuma/wiki/jobs.py
jezdez: Sure why not. FTR, the only reason why I even override the `key` method is because the automatic approach implemented in cacheback doesn't correctly take unicode vs. byte string differences into account. Which this simple key method does.

kuma/wiki/tests/test_models.py
```diff
@@ -19,18 +19,19 @@
...
+from . import (document, revision, doc_rev, normalize_html,
+               create_template_test_users, create_topical_parents_docs)
+from .. import tasks
+from ..constants import REDIRECT_CONTENT
+from ..exceptions import (PageMoveError,
+                          DocumentRenderedContentNotAvailable,
+                          DocumentRenderingInProgress)
+from ..jobs import DocumentZoneStackJob
+from ..models import (Document, Revision, RevisionIP,
+                      DocumentZone, TaggedDocument)
```
groovecoder: nit: could put `DocumentZone` up on the preceding line now?
```
from ..models import (Document, Revision, RevisionIP, DocumentZone,
                      TaggedDocument)
```

tox.ini
```diff
@@ -7,7 +7,7 @@ deps=
     -rrequirements/compiled.txt
     -rrequirements/dev.txt
 commands =
-    coverage run manage.py test --noinput -v2 kuma
+    coverage run manage.py test --noinput -v2 {posargs:kuma}
```
groovecoder: This change is also in https://github.com/mozilla/kuma/pull/3109/files#diff-b91f3d5bd63fcd17221b267e851608e8 so it may potentially conflict?

kuma/wiki/jobs.py
groovecoder: Maybe send an upstream pull request :question:

groovecoder:
Code review done ... moving on to spot check.

groovecoder:
Found a bug where a new zone isn't displayed as such ...

Steps to reproduce:
1. Make a new document
2. Convert it into a zone
3. Refresh it

Expected result:
The doc should show up as a content zone. (i.e., including the `zone-landing` classes, and the "Manage content zone" link in the gear menu)

Actual result:
The doc displays as a regular doc.

jezdez:
Aha, thanks @groovecoder!

kuma/wiki/tests/test_models.py
jezdez: :+1:

tox.ini
jezdez: Nah, git will auto merge that since there is nothing else in the other branch changing on that line.

kuma/wiki/jobs.py
jezdez: Did that, I think that solved the problem, https://github.com/codeinthehole/django-cacheback/pull/39

jezdez: In other words, I'm not going to add a `KumaJob` for now since we can rely on the automatic key generation.

jezdez:
@groovecoder I've fixed the things you saw, I think -- using the `post_save` signal for documents instead of the `render_done` one. Let me know if it works for you. I did find an issue with cacheback actually, as you suggested and fixed it locally. PR: https://github.com/codeinthehole/django-cacheback/pull/40

groovecoder:
Looks good. I was able to make a new doc, convert it to zone, and see it rendered as a zone. I also made a child page of the zone landing page and it immediately rendered as the right zone too.

PR #3116, opened 2015-03-13
bug 1142639 - move maintenance/ to media/ for static serving

groovecoder:
Moving the maintenance page into our own `media/` directory means we can serve it at `/media/maintenance` via Apache before any python is executed.

E.g. locally, go to https://developer-local.allizom.org/media/maintenance/ without `runserver` running.

This makes it easier to update in the future, and we can simply redirect all MDN requests to /media/maintenance.


openjck:
LGTM :shipit:

PR #3117, opened 2015-03-16
DO NOT MERGE:  bug 875062 - Add initial framework for social sharing on MDN

darkwing:
Stephanie:  This is the initial code for adding social sharing to MDN.  We still need some styling but the links are working great!


darkwing:
Actually, @stephaniehobson, we may want to make these buttons instead of links..

PR #3118, opened 2015-03-16
Fix Bug 1137656: Overlapping icons in quick links

stephaniehobson:
- Consolidated quick links code into one file.
- Moved p and div declaration outside of the icon selector.
- Consolidated link style declarations
- Moved .see-also declaration inside of .title declaration so relationship between the two is clearer.
- Restricted toggle icon styles to toggle icon so they don't have to be over-written by sidebar-icon styles anymore
- Altered sidebar-icon styles to be more RTL friendly
- Moved some things into variables or made use of existing ones
- Instead of absolute positioning, icons are placed in the gutter by having a negative margin on the first icon. Subsequent icons appear inline so they don't overlap anymore.

Testing pages:
- https://developer.mozilla.org/en-US/docs/Web/API/AnimationEvent/initAnimationEvent
- https://developer-local.allizom.org/en-US/docs/Web/HTML/Element

### DIFF
--- a/media/stylus/components/wiki/customcss.styl
+++ b/media/stylus/components/wiki/customcss.styl
@@ -1649,14 +1649,6 @@ dd.landingPageList {
     background: rgba-fallback(rgba(230,96,0,0.45));
     border-color: rgba-fallback(rgba(0,0,0,0.05));
 }
-/* sidebar quick links stuff */
-.quick-links .sidebar-icon {$selector-icon} {
-    top: 10px;
-    position: absolute;
-    left: 12px;
-    display: block;
-    opacity: 0.3;
-}
 /* new community box */
 .communitybox {
     background-color: #00539f;
--- a/media/stylus/components/wiki/quick-links.styl
+++ b/media/stylus/components/wiki/quick-links.styl
@@ -1,11 +1,17 @@
-/* quick links */
+$see-also-outdent = 6px;
+
 .quick-links {
     set-smaller-font-size();
-    position: relative;
     margin-bottom: $grid-spacing;
+    position: relative;
 
     a {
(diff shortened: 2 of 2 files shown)


darkwing:
Kicked this, failure looked bogus

groovecoder:
Failure is on Travis/wget. There's fix in https://github.com/mozilla/kuma/pull/3120 but we can merge thru this failure.

groovecoder:
Needs rebase after the media/redesign merge.

groovecoder:
Does @openjck or @darkwing want to give the stylus a review before I spot-check to merge?

darkwing:
Spotted right away that the arrow position is moving between "Hide Sidebar" and "Show Sidebar"

stephaniehobson:
Toggle link fixed.

darkwing:
Seeing this on single-icon'd view:

http://screencast.com/t/WQRnkPu1sIp4

The icons are "below" the text a tiny bit.  

I added `vertical-align: super` to `.quick-links .sidebar-icon i[class^="icon-"]`, and that looks great for single icons, but double don't look great.  So then I added `vertical-align: super` to the `s.obsoleteElement` element which I took from your sample (https://gist.github.com/stephaniehobson/908ff59df89128fa24d6) and they all lined up.   Really odd.

stephaniehobson:
Tried to fix it with vertical-align: top and a bit of absolute positioning. Not sure why the spacing goes wacky when there's a second icon but I was seeing it too.

darkwing:
It was the `s.obsoleteElement` next to the two icons causing that issues, I think.

darkwing:
Looks good! R+

PR #3121, opened 2015-03-17
New issue 1060633

abhidoeslinux:
Added a link to getting started page, appears the first time user signs up.

### DIFF
--- a/kuma/landing/templates/landing/homepage.html
+++ b/kuma/landing/templates/landing/homepage.html
@@ -181,7 +181,11 @@
 
 {% block js %}
   {{ super() }}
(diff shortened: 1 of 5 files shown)


openjck:
Thanks so much for submitting this! This looks good.

The message will appear even without the lines in `profile.html` and `base.html`. We can safely remove those.

Do you have any experience squashing Git commits? It would be great to squash these so that we only have one commit related to this feature in our version control history.

abhidoeslinux:
Hi openjck. I messed up while squashing so I deleted that branch and created a new one. I have opened a PR with the new branch

PR #3122, opened 2015-03-18
fix bug 1060633 using django messaging framework and mdn notification system

abhidoeslinux:
New pull request with a single commit message

### DIFF
--- a/kuma/users/models.py
+++ b/kuma/users/models.py
@@ -1,5 +1,6 @@
 import datetime
 
+from django.contrib import messages
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
 from django.dispatch import receiver
@@ -17,6 +18,7 @@ from kuma.core.fields import LocaleField, JSONField
 from kuma.core.managers import NamespacedTaggableManager
 from kuma.core.models import ModelBase
 from kuma.wiki.models import Revision
+from kuma.wiki.helpers import wiki_url
 
 from .helpers import gravatar_url
 from .tasks import send_welcome_email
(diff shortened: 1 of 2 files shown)


openjck:
Awesome work. This looks great.

I mentored @abhidoeslinux on this contribution. Would someone else, maybe @jezdez and/or @darkwing be able to review this?

kuma/users/models.py
```diff
@@ -195,6 +197,9 @@ def create_user_profile(sender, instance, created, **kwargs):
 
 @receiver(user_signed_up)
 def on_user_signed_up(sender, request, user, **kwargs):
+    context={'request':request}
```
jezdez: Please add a whitespace after the colon.

kuma/users/models.py
```diff
@@ -195,6 +197,9 @@ def create_user_profile(sender, instance, created, **kwargs):
...
+    msg='You have completed the first step of <a href="%s">getting started with MDN</a>' % wiki_url(context, 'MDN/Getting_started')
```
jezdez: The message needs to be marked for translation so that we can have our translators turn it into translations.

kuma/users/models.py
abhidoeslinux: @jezdez Added.

kuma/users/models.py
abhidoeslinux: @jezdez @openjck Not sure how do I go ahead with this? help?

openjck: To make this change, you could edit the file and add another commit. Once all of the changes have been made, we could squash all of the commits into one.

abhidoeslinux: Thanks! I have added a new commit. Looking for any options that allow squashing.

groovecoder:
Uplifting from IRC:
15:07 jezdez: groovecoder: Abhishek: we have a work around for this in place
15:07 jezdez: lemme find that code
15:09 jezdez: Abhishek: see https://github.com/mozilla/kuma/blob/master/kuma/core/tests/__init__.py#L180-L186
15:10 jezdez: Abhishek: then use it like this: https://github.com/mozilla/kuma/blob/7de757d1710c770c831ba9be47705b28e3db3458/kuma/users/tests/test_adapters.py#L89-L97

abhidoeslinux:
Working on it..

groovecoder:
Superseded by https://github.com/mozilla/kuma/pull/3130

PR #3127, opened 2015-03-19
Bug 1139508: Redirect media/redesign with htaccess

openjck:
We need to redirect from /media/redesign to /media. This htaccess
redirect won't have any affect locally (because we alias /media in
mozilla-kuma-apache.conf), but it might work on production where the
alias isn't active.

Unfortunately, the only way to know for sure is to try it out. If this
doesn't do the trick, we can revert this change and work with WebOps to
set up the redirects in whatever form their configuration requires.


groovecoder:
Since this is going to production and we don't know for sure that it's right, can we start this as a 302 instead of a 301?

openjck:
Good question. We should be sure we're not permanently redirecting users to the wrong destination. Thankfully, the destination isn't uncertain or dependent on our configuration, just whether or not the redirect will be picked up at all. In other words, the redirect will either work as expected or have no effect.

PR #3129, opened 2015-03-19
bug 1113260 - manual IP ban & un-ban

groovecoder:
Spot check:
0. Make sure you're running the `python2.6 manage.py celery worker --events --beat --autoreload --concurrency=4` process
1. Activate [waffle switch](https://developer-local.allizom.org/admin/waffle/switch/) `store_revision_ips`
2. Make an edit
3. Go to [the revisions dashboard](https://developer-local.allizom.org/en-US/dashboards/revisions), click "Toggle IPs" and then click "Ban IP from Editing"
4. [ ] Go back to the edit page, verify you get permission denied error page
5. Go to [the IP Bans admin page](https://developer-local.allizom.org/admin/core/ipban/), click the IP Ban, click "Delete" and then "Yes, I'm sure"
6. [ ] Go back to the edit page, verify you get the edit page

### DIFF
--- a/kuma/core/models.py
+++ b/kuma/core/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.dispatch import receiver
 from django.utils import timezone
 
 from south.modelsinspector import add_introspection_rules
@@ -48,10 +49,15 @@ class IPBan(models.Model):
     def delete(self, *args, **kwargs):
         self.deleted = timezone.now()
         self.save()
-        IPBanJob().invalidate(self.ip)
 
     def __unicode__(self):
         return u'%s banned on %s' % (self.ip, self.created)
+
+
+@receiver(models.signals.post_save, sender=IPBan)
+@receiver(models.signals.pre_delete, sender=IPBan)
+def invalidate_ipban_caches(sender, instance, **kwargs):
+    IPBanJob().invalidate(instance.ip)
 
 
 add_introspection_rules([], [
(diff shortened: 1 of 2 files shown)


openjck:
Looks good. Just waiting for Travis.

kuma/core/models.py
```diff
@@ -48,12 +49,19 @@ class IPBan(models.Model):
...
     def __unicode__(self):
         return u'%s banned on %s' % (self.ip, self.created)
 
 
+def invalidate_ipban_caches(sender, instance, **kwargs):
+    IPBanJob().invalidate(instance.ip)
+
+
+models.signals.post_save.connect(invalidate_ipban_caches, sender=IPBan)
+models.signals.post_delete.connect(invalidate_ipban_caches, sender=IPBan)
```
jezdez: I'm not sure if `pre_delete` wouldn't be better here since the `post_delete` works under the assumption that the table row is already gone and only the python ORM object has the info needed. To prevent future mistakes that extend the `invalidate_ipban_caches` signal handler somehow I'd rather go with that.

Also you should use the `receiver` decorator if you're importing it, like so:

```
@receiver(models.signals.post_save, sender=IPBan)
@receiver(models.signals.pre_delete, sender=IPBan)
def invalidate_ipban_caches(sender, instance, **kwargs):
    IPBanJob().invalidate(instance.ip)
```

groovecoder:
Good calls @jezdez. Updated.

PR #3132, opened 2015-03-20
Bug 1137020: Improve callout box design

openjck:


media/stylus/home.styl
```diff
@@ -21,7 +21,11 @@ $home-involved-background-y = 50px;
     minimize-header-search();
 
     .home-callouts {
-        background: #4d4e53;
+        background: #fff;
+
+        .column-container {
+            padding: $grid-spacing;
```
stephaniehobson: Clashes with code on line 466 to give twice as much padding as we need on tablet and mobile view. 465 to 467 can probably just get deleted (but I can't leave a comment on those lines because they weren't edited).

openjck:
Great catch! Updated.

stephaniehobson:
:+1:

PR #3134, opened 2015-03-23
bug 1144213 - pin tox version 1.9.0 for travis

groovecoder:
1.9.1 bug: https://bitbucket.org/hpk42/tox/issue/230/tox-191-ignores-package-versions-specified


jezdez:
This was now fixed in tox 1.9.2

groovecoder:
Updated this to pin tox to 1.9.2. We should keep it pinned to avoid this in the future, right?

jezdez:
Yeah, makes sense pinning.

PR #3135, opened 2015-03-24
fix bug 1115071 - Ensure correct addresses for context menu items

darkwing:
Should've just used the full URL in the `data-` attribute in the first place.

### DIFF
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -424,8 +424,8 @@
     </div> <!-- ends "main-content" -->
 
   <menu type="context" id="edit-history-menu">
-    <menuitem data-action="$edit" label="{{_('Edit page')}}"></menuitem>
-    <menuitem data-action="$history" label="{{_('View page history')}}"></menuitem>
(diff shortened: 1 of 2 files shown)


kuma/wiki/templates/wiki/document.html
```diff
@@ -424,8 +424,8 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
     </div> <!-- ends "main-content" -->
 
   <menu type="context" id="edit-history-menu">
-    <menuitem data-action="$edit" label="{{_('Edit page')}}"></menuitem>
-    <menuitem data-action="$history" label="{{_('View page history')}}"></menuitem>
+    <menuitem data-action="{{ url('wiki.edit_document', document.full_path, locale=document.locale) }}" label="{{_('Edit page')}}"></menuitem>
+    <menuitem data-action="{{ url('wiki.document_revisions', document.full_path) }}" label="{{_('View page history')}}"></menuitem>
```
openjck: Out of curiosity, why no locale argument for the revision page?

darkwing: I'm using the same URL for the `page-buttons` revision page, so I cannot provide a given reason, but since we've not had bugs about the other URL, this shouldn't be an issue either.  Ask @jezdez maybe?

PR #3137, opened 2015-03-25
fix bug 1126461 - Don't sticky buttons when on smaller screens

darkwing:
I'm setting a different `position` on the `small-desktop`, like we do for the TOC, to signal the buttons should not follow if on a smaller screen.

I've also eased up on the debounce.  All the scroll functionality is taxing right now.

### DIFF
--- a/media/js/wiki.js
+++ b/media/js/wiki.js
@@ -302,6 +302,8 @@
         var $pageButtons = $('.page-buttons');
         var pageButtonsOffset = $pageButtons.offset();
 
+        var stickyFeatureEnabled = $pageButtons.attr('data-sticky') === 'true';
+
         // Get button alignment according to text direction
         var buttonDirection = ($('html').attr('dir') === 'rtl') ? 'left' : 'right';
 
@@ -310,6 +312,8 @@
             var pageButtonsHeight = 0;
             var $mainContent = $('.wiki-main-content');
 
+            var pointerEvents = $toggler.css('pointer-events');
+
             if(!e || e.type === 'resize') {
                 // Calculate right and offset for page buttons on resize and page load
                 if(buttonDirection === 'right'){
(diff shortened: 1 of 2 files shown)


media/js/wiki.js
```diff
@@ -302,6 +302,8 @@
         var $pageButtons = $('.page-buttons');
         var pageButtonsOffset = $pageButtons.offset();
 
+        var stickyFeatureEnabled = $pageButtons.attr('data-sticky') === 'true';
```
openjck: We could think about removing the `data-sticky` attribute altogether, unless we plan to disable the sticky buttons dynamically in the future.

openjck: Also, yay for strict equals! :smile:

media/stylus/components/wiki/page-buttons.styl
```diff
@@ -34,6 +34,15 @@
...
 
+@media $media-query-small-desktop {
+
+    .page-buttons, #page-buttons {
+        &.fixed {
+            position: relative;
+        }
+    }
+}
+
```
openjck: Another option to consider is nesting this media query under the existing occurrence of the selector:

https://github.com/mozilla/kuma/pull/3137/files#diff-e2b6c4642cf93c8eab58fc0ecb41f3adR7

darkwing: I'll defer to @stephaniehobson on this.

media/js/wiki.js
darkwing: Do we want to make this waffled?

media/stylus/components/wiki/page-buttons.styl
stephaniehobson: Since we already have a couple other media queries separated out I think where it is is the best place for it.

media/js/wiki.js
stephaniehobson: I don't think we need to waffle it. We test it, we like the results, we're rolling it into production.

openjck: Yep, agreed.

stephaniehobson: Not ready for public consumption yet because of RTL problems so it's not blocking this PR :)

PR #3138, opened 2015-03-26

openjck:
The more testing we can do, the better. This touches so many areas of the codebase.

**To do**
* [ ] Fix the Travis errors, apparently related to Travis not running `collectstatic`
* [ ] Fix local 404s ([more info](https://github.com/mozilla/kuma/pull/3138#issuecomment-88998467))
* [ ] Move `media/styles/` to `styles/` and `media/js/` to `js/` and so on ([more info](https://github.com/mozilla/kuma/pull/3138#issuecomment-89006752))
* [ ] Look into and consider refactoring `for style in styles` ([more info](https://github.com/mozilla/kuma/pull/3138#issuecomment-91285938))
* [ ] Fix issue with display of homepage demos
* [x] Move the Pipeline template to `templates/` and remove `media/templates` from `TEMPLATE_DIRS`


openjck:
The error appears to be related to Travis not running `collectstatic`. I'll try to find a way to make it do that.

groovecoder:
This looks awesome. The changes to docs suggest I can just run it? Or are there spot-check instructions?

Travis shows a lot of these:
```
The file 'js/wiki.js' could not be found with <pipeline.storage.PipelineCachedStorage object at 0x5b67910>
```
?

Also, since this removes 17.8k lines of code, we'll need to include 17.8 animated GIFs in the pull request. I'll start us off ...
pipe

groovecoder:
Hmm ... well, the spot-check is working so well so far that I almost don't trust it ...

groovecoder:
Bunch of these on my local box:
```
[02/Apr/2015 11:17:20] "GET /en-US/fonts/OpenSans-Italic-webfont.ttf HTTP/1.1" 404 9239 (time: 3.25s; sql: 11ms (6q))
```

groovecoder:
When I ran `manage.py collectstatic` it seemed to copy over `media/attachments` files? Are we supposed to skip that directory?

openjck:
> This looks awesome. The changes to docs suggest I can just run it? Or are there spot-check instructions?

No special steps are needed to run Pipeline locally. To test the changes as if they were being run on production, change `DEBUG` and `DEV` to `False`. I'm actually having trouble getting Django to serve up compressed assets in that latter case. It can find them, but not serve them. Though that may be due to me "emulating" production incorrectly.

> Travis shows a lot of these:

I think that's due to Travis [not running connectstatic](https://github.com/mozilla/kuma/pull/3138#issuecomment-88275994). I have been looking into how to address that, but haven't found a solution yet.

> When I ran manage.py collectstatic it seemed to copy over media/attachments files? Are we supposed to skip that directory?

Great catch. I can update `STATICFILES_DIRS` to call out `media/styles` individually.

groovecoder:
Not sure if this is a separate bug ... but in `base.html` we have:
```
  {% block site_css %}
    {% compressed_css 'mdn' %}

    {% for style in styles %}
      {% compressed_css style %}
    {% endfor %}
  {% endblock %}
```
And then in `homepage.html` we have:
```
{% block site_css %}
    {{ super() }}
    {% compressed_css 'home' %}
{% endblock %}
```
To include an extra bundle, it seems like we should use **either** the `styles` list, **or** the `super()` method?

openjck:
> Not sure if this is a separate bug ... but in base.html we have:

Good question. We were doing that before django-pipeline, but I'll take a look anyway. I recall refactoring this away back when we were using Gulp.

openjck:
When I change `STATICFILES_DIRS` to name `media/styles` and `media/js` individually, I see this error about `blueprint.png` not being found.

https://gist.github.com/openjck/993a99d43db7977561b5

Usually the error means a stylesheet is referring to the file by absolute address, but that's not the case with `blueprint.png`.

groovecoder:
Are those the only directories you added to it? You may need to add `media/img` for it to find the files in that directory.

openjck:
> Are those the only directories you added to it? You may need to add media/img for it to find the files in that directory.

Same error unfortunately. I think collectstatic finds and processes in-source references automatically, anyway.

openjck:
@jezdez: Any ideas on what could be causing [this error](https://github.com/mozilla/kuma/pull/3138#issuecomment-90108056)?

jezdez:
@openjck I've said it before, please separate uploads from static files. Attachments are **not** support to be stored along side with static files.

In other words, `MEDIA_ROOT` should *not* be listed in `STATICFILES_DIRS` as it would defeat the purpose of deployment time file copying. You don't want to copy around *all* attachments to the `STATIC_ROOT` right?

The error you were point at is quite simple: the staticfiles finder system can't find the file in the location it is provided, in our case all ``static`` subdirectories in apps and the folders listed in ``STATICFILES_DIRS``. You **don't** need the folder `media/img` to the setting as that's not how the file is referred to. It looks for a file that starts with`img/` not without. I see the file listed in `customcss.styl` in the `.communitybox` class.

I saw that you added `path('media/templates')` to `TEMPLATES_DIRS`, why?

openjck:
> In other words, MEDIA_ROOT should not be listed in STATICFILES_DIRS as it would defeat the purpose of deployment time file copying. You don't want to copy around all attachments to the STATIC_ROOT right?

Definitely. Luke [made the same point](https://github.com/mozilla/kuma/pull/3138#issuecomment-89006752), and I agree. In my working directory, `media/styles` and `media/js` are individually added to `STATICFILES_DIRS`. I haven't pushed that change up yet because it caused (revealed?) [the other problem](https://github.com/mozilla/kuma/pull/3138#issuecomment-90108056).

> I saw that you added path('media/templates') to TEMPLATES_DIRS, why?

On master, we [hard-code the filenames of CKEditor dependencies](https://github.com/openjck/kuma/blob/master/kuma/wiki/templates/wiki/ckeditor_config.js#L89-93) in a configuration file. On this branch, we [use compressed_css](https://github.com/openjck/kuma/blob/0291008212523d02cf47a56b640ee41adb6a0ccc/kuma/wiki/templates/wiki/ckeditor_config.js#L89) along with a [custom template](https://github.com/openjck/kuma/blob/0291008212523d02cf47a56b640ee41adb6a0ccc/media/templates/pipeline/javascript-array.jinja) and a [special configuration in settings.py](https://github.com/openjck/kuma/blob/0291008212523d02cf47a56b640ee41adb6a0ccc/settings.py#L789) that tells Pipeline to use the custom template when the CKEditor dependencies are included. If we would prefer to do this differently, let me know.

jezdez:
But even a subdirectory of `media` shouldn't be in STATICFILES_DIRS since collectstatic will recursively copy things around. Just separate them completely please.

jezdez:
About the template, please move the file outside of the `media` folder as well, it has nothing to do with static files after all. And remove `media/templates` from the `TEMPLATES_DIRS` setting please.

jezdez:
The best place is in the templates folder.

openjck:
> But even a subdirectory of media shouldn't be in STATICFILES_DIRS since collectstatic will recursively copy things around. Just separate them completely please.

Sure. Where would you recommend putting `media/styles` and `media/js` instead? Is there an example in another open-source project we could follow?

jezdez:
Sure thing, check out https://github.com/django/djangoproject.com/blob/master/djangoproject/settings/common.py

It uses a few helpers for the path construction, but the gist is this, translated to our setup:

```python
MEDIA_ROOT = path(uploads')
MEDIA_URL = '/uploads/'

DEMO_UPLOADS_ROOT = os.path.join(MEDIA_ROOT, 'demos')
DEMO_UPLOADS_URL = '/uploads/demos/'

STATICFILES_DIRS = [
    path('static'),  # for files not requiring compilation
    path('stylus'),  # for sourcing the ``PIPELINE_CSS`` and ``PIPELINE_JS`` packs
]

STATIC_ROOT = path('assets')
STATIC_URL = '/assets/'

STATICFILES_FINDERS = (
    'django.contrib.staticfiles.finders.FileSystemFinder',
    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
)
HUMANSTXT_ROOT = STATIC_ROOT
```

The problem with all of this is that "media" is an incredibly ambiguous term and that Django for a long term left it to the users to handle asset management and file upload handling. In the future kuma may be required to separate those two, e.g. have uploads on a different S3 bucket than the site assets. The only way to isolate them is to have different locations. The names above reflect that, the `MEDIA_ROOT` setting should really be called `UPLOADS_ROOT` but that ship has sailed.

As you can see I've added a new folder called "stylus" which lives in the root directory next to the "static" folder. That still means the folder will be picked up by pipeline when running collectstatic, but newer version of pipeline have an own finder implementation that allows to ignore those files matching a certain file extensions, like .stylus. Once we're on a newer version of Django we can use that. (https://github.com/cyberdelia/django-pipeline/blob/master/pipeline/finders.py#L69-L117)

One thing I've been wanting to do as part of the settings refactor in #3062 is move the templates and static assets of the kuma project into the new kuma namespace package. So that we'd have all file the logically belong to kuma-the-software-package are in one place and not litered in the root directory. That would give you then a directory structure of `kuma/static`, `kuma/stylus`, `kuma/templates`, but `uploads` and `assets` stay in the root directory.

groovecoder:
So, will this need some web-head work to merge & push? We have an opportunity to do a sprint with WebOps on Thursday, so if there's something to do for this, we can plan it by then.

openjck:
I don't think we'll need to coordinate with WebOps, but I'll know more soon once I dig back into this.

groovecoder:
@openjck - what's the status on this? Is it waiting for a specific PR to merge first?

openjck:
> @openjck - what's the status on this? Is it waiting for a specific PR to merge first?

No dependencies, just a lot happening in parallel. I've been working pull request reviews so far. There's also the [search filter rework](https://github.com/mozilla/kuma/pull/3172), the social sharing button A/B tests, and the [static code example redirects](https://bugzilla.mozilla.org/show_bug.cgi?id=1078186). There's also a Mozilla-sponsored event this weekend that needs me to present.

I definitely over-committed myself this week. :smile:

I'll finish off the pull request reviews today after rebuilding my Vagrant box, and then move back onto django-pipeline from there.

openjck:
Starting a new branch.

PR #3139, opened 2015-03-26
fix bug 886016 - Add pagination to the top and bottom of list pages

darkwing:
Easy enough.  The double-bottom-spacing I created is eye-pleasing but I didn't use any science.

### DIFF
--- a/kuma/wiki/templates/wiki/list_documents.html
+++ b/kuma/wiki/templates/wiki/list_documents.html
@@ -43,6 +43,7 @@
     <div id="document-list" class="boxed">
       <h1>{{ title }}</h1>
       {% if documents.object_list %}
+        {{ documents|paginator }}
         <p>{{ counter }}</p>
         <ul class="document-list">
           {% for doc in documents.object_list %}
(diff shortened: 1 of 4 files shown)


stephaniehobson:
I have some science to confirm your double-bottom-spacing :)

Before David:
screen shot 2015-03-27 at 10 20 50

After David:
screen shot 2015-03-27 at 10 20 50

It IS more pleasing! And I'd like to suggest we drop it to 2em or 30px.

Because the internal elements are floated the changes to the padding is actually helping the element take up the appropriate amount of vertical space. Yay! I think we should change it to 2em to make the padding reflect the actual vertical space without adding any extra to the bottom. 

I suggest ems because the vertical space will change if the text size does. We can go with 30px if you'd rather stick to px though, it gets a little out of vertical alignment with bigger text but nothing breaks.

stephaniehobson:
Silly Github, that's the same image twice.

stephaniehobson:
After David:

screen shot 2015-03-27 at 10 20 13

darkwing:
Updated with `1.5`; sticking with px because it's more consistent with all other spacing!

stephaniehobson:
:+1:

PR #3140, opened 2015-03-27
Bug 1148435 - Added Python 2.7 to Travis config.

jezdez:
It currently is allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ services:
 env:
     matrix:
         - TOXENV=py26
+        - TOXENV=py27
         - TOXENV=flake8
         - TOXENV=docs
     global:
@@ -34,3 +35,4 @@ notifications:
 matrix:
     allow_failures:
         - env: TOXENV=flake8
+        - env: TOXENV=py27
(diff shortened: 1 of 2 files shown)


jezdez:
I don't quite understand how this works on 2.6 and not on 2.7. Is the product json app only Python 2.6 compatible?

jezdez:
@groovecoder Any idea?

groovecoder:
Strange. It should work in both. I've seen that error before in python 2.6. The `django-mozilla-product-details` app is looking for the `product_details_json` directory. That should be fixed with the `travis-install` script extracting the new [`etc/data/product_details_json.tar.gz`](https://github.com/mozilla/kuma/tree/6bf3e40ea79517a1683020dfdfbde80867623cd3/etc/data) into the travis root directory.

The stranger(?) thing is that the py27 job is still using 2.6.9? (https://travis-ci.org/mozilla/kuma/jobs/56108876)

jezdez:
That's only because the main python we install tox with is 2.6.9, but the python *inside* tox will be python 2.7, you can see it in the output.

jezdez:
I'm inclined to let this slide in favor of #3073, what do you think @groovecoder?

groovecoder:
Yup. I almost asked you the same thing yesterday.

PR #3142, opened 2015-03-27
Bug 1145630: Social Sharing A/B tests

stephaniehobson:
Styling for 3 different types of buttons going in 3 different places.

Classes:
.share
.share.share-group
.share.share-thin

All hidden by default at the moment, show by adding class `show` to the element with class `share`

Here's a screen shot of the `share.share-thin` positioned under the title, `share.share-group` positioned in the TOC, and the default `.share` under the article.
screen shot 2015-03-30 at 15 47 49

### DIFF
--- a/media/stylus/components/share-link.styl
+++ b/media/stylus/components/share-link.styl
@@ -1,0 +1,124 @@
+/*
+share-link
+====================================================================== */
+.share-link {
+    border: 1px solid rgba-fallback(rgba(255,255,255,0.2));
+    border-radius: 2px;
+    box-shadow: none;
+    color: #fff;
+    bidi-value(margin, 0 6px 6px 0, 0 0 6px 6px);
+    padding: 3px;
+
+    &.facebook {
+        background-color: rgb(59, 89, 152);
+    }
+    &.twitter {
+        background-color: rgb(0, 172, 237);
+    }
+    &.gplus {
+        background-color: rgb(221, 75, 57);
+    }
+
+    {$selector-icon} {
+        bidi-style(margin-left, 0, margin-right, 4px);
+        bidi-style(margin-right, 4px, margin-left, 0);
+    }
+
+
+    /*
+    share-thin (can be removed if we decid not to use the share-thin style)
+    ====================================================================== */
+
+    .share-thin & {
+        display: inline-block;
+        width: 0;
+        height: 20px;
+        bidi-value(padding, 0 0 0 25px, 0 25px 0 0);
+        overflow: hidden;
+        vertical-align: middle;
+        position: relative;
+        border: 2px solid transparent;
+
+        &:focus,
+        &:hover {
+            border: 2px solid rgba(255,255,255, 0.5);
+        }
+
+        {$selector-icon} {
+            position: absolute;
+            top: 4px;
+            left: 50%;
+            bidi-value(margin-left, -0.4em, -0.4em); /* using ems because relative to text size */
+        }
+
+        &.facebook {$selector-icon} {
+            bidi-value(margin-left, -0.25em, -0.25em); /* using ems because relative to text size */
+        }
+
+        &.gplus {$selector-icon} {
+            bidi-value(margin-left, -0.35em, -0.35em); /* using ems because relative to text size */
+        }
+
+
+        @media $media-query-small-desktop {
+            height: 40px;
+            bidi-style(padding-left, 40px, padding-right, 0);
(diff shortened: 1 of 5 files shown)


stephaniehobson:
This is not ready yet. It lacks one significant thing... a way to make the share links visible. This is a tini bit complicated for the one in the .toc.

Putting it here because I know @openjck was talking about working on optimizely tests at the hack event, thought he might want to steal it.

I intend to come back to this after my doctor's appointment today and fix the one missing thing but, it's a Friday and I might get captured by a wild beer.

stephaniehobson:
Cleaned up and ready for review.

openjck:
This. Looks. Awesome. :smiley:

Will we plan to keep all three styles around after the experiment ends? If not, maybe we could add some comments to code that might later be eligible to be removed.

media/stylus/components/share-link.styl
```diff
@@ -0,0 +1,122 @@
...
+        &.facebook {$selector-icon} {
+            bidi-value(margin-left, -0.25em, -0.25em); /* using ems because relative to text size */
+        }
+
+        &.gplus {$selector-icon} {
+            bidi-value(margin-left, -0.35em, -0.35em); /* using ems because relative to text size */
+        }
+
+
+        @media all and (max-width: 1200px) {
```
openjck: We could alternatively use `$media-query-small-desktop` here.

media/stylus/components/share-link.styl
```diff
@@ -0,0 +1,122 @@
+/*
+share-link
+====================================================================== */
+.share-link {
+    border: 1px solid rgba-fallback(rgba(255,255,255,0.2));
+    border-radius: 2px;
+    box-shadow: none;
+    color: #fff;
+    bidi-value(margin, 0px 6px 6px 0px, 0px 0px 6px 6px);
```
openjck: I'm pretty sure our compressor does this, but just to be safe, maybe we could replace `0px` with `0` in this file.

media/stylus/components/share.styl
```diff
@@ -12,32 +13,78 @@
...
+
+        @media $media-query-small-desktop {
+            background-color: transparent;
+            top: 0;
+        }
 
-    &.facebook {
-        background-color: rgb(59, 89, 152);
+        /* hide title if there isn't enough space for it in TOC */
+        @media all and (min-width:1025px) and (max-width: 1400px),
```
openjck: It's one pixel off, but we could try the similar `$media-query-tablet` for a `min-width` of `1024px`.

media/stylus/components/share.styl
```diff
@@ -12,32 +13,78 @@
...
+        all and (max-width: 479px) {
```
openjck: `$media-query-small-mobile` could potentially be used for a `max-width` of `480px`. Unless I'm mistaken, and we want to intentionally react just before everything else does.

media/stylus/components/share.styl
```diff
@@ -12,32 +13,78 @@
...
+            }
+        }
     }
-    &.twitter {
-        background-color: rgb(0, 172, 237);
+
+    /* TODO: this will over ride hide by default above, optimizely needs to add a class to swap */
+    /* TODO: display by default at full screen width when toggle not possible */
+    .toggle-container + &.show {
+      display: block;
```
openjck: Looks like EditorConfig might have missed this and a couple of rules below.

openjck:
Added a few nitpick line comments. All very minor. Otherwise this looks fantastic. I'm excited to run this experiment.

stephaniehobson:
Updated and ready for re-review.

media/stylus/components/share.styl
stephaniehobson: Couldn't use the variable because I needed min-width and it's max-width.

media/stylus/components/share.styl
stephaniehobson: Updated part of this to use a variable but Stylus didn't evaluate line 40 if I switched to use our variable there :(

openjck:
Awesome! I'll take a look.

Will we plan to keep all three styles around after the experiment ends? If not, maybe we could add some comments to code that might later be eligible to be removed.

openjck:
When we run the experiment, would we want to show all buttons at once or experiment with different placements in addition to different styles?

stephaniehobson:
The buttons should only show up in one place on the page at a time.

PR #3143, opened 2015-03-28
fix bug 1148738 - Formal syntax box may be longer than the screen

tregagnon:
- Switching `.syntaxbox`'s `white-space` property from `pre-line` to `pre-wrap`. A quick solution until we find a more advanced solution. (an expandable box triggered by the user?)
- Adding ellipsis at the start and the end of line when text is cut off.

I checked with @teoli2003. This is one of the expected changes for the new CSS Reference article template . It also improves JavaScript articles when viewed through a small viewport.

### DIFF
--- a/media/stylus/components/syntax/syntaxbox.styl
+++ b/media/stylus/components/syntax/syntaxbox.styl
@@ -18,7 +18,8 @@ pre.twopartsyntaxbox {
 
 pre.syntaxbox code[class*="language-"],
 pre.twopartsyntaxbox code[class*="language-"] {
-    white-space: pre-line;
+    white-space: pre-wrap;
+    text-overflow: ellipsis ellipsis;
 }
 
 .syntaxbox-help {


groovecoder:
@stephaniehobson and/or @openjck take a look at this?

groovecoder:
Giving to @openjck - he has fewer PR's already assigned. :wink:

openjck:
Awesome! Thanks for the contribution, @tregagnon.

What would be a good page to test this against? I notice that the rule modified here doesn't select the [display formal syntax box](https://developer.mozilla.org/en-US/docs/Web/CSS/display#Formal_syntax) Jean-Yves referred to in the bug.

stephaniehobson:
Here's a good testing page for syntax related stuff. Decrease the size of your browser to invoke the wrapping stuff on most samples. https://developer.mozilla.org/en-US/docs/User:stephaniehobson:code

I think @tregagnon has found a bug in how this was set up initially because his change isn't being picked up and neither is the original white-space declaration.

Instead of editing the declaration on line 21 let's move the original white-space declaration up to line 11 and delete the declaration at lines 19-23.

stephaniehobson:
Thanks for helping us solve this one @tregagnon. I thought it was an important fix to get into production so I integrated your branch into #3197.

PR #3144, opened 2015-03-28
Fix Bug 1148752 - Main heading of wiki articles creates horizontal scroll with small viewports

tregagnon:
- Breaks all words in wiki articles main heading. The long API method names create ugly horizontal scrollbars when pages are viewed from tablets or mobiles. Broken words are ugly but horizontal scrolls are too inconvenient.
- Reduce `line-height` as so much spaces between each lines of the same word was confusing.
- Keep whitespace in design by increasing margin's sizes
- Switch margin's length to em so they will stay coherent with the font size even if `font-size` changes

I look around in the Style code and `em` doesn't seem to be used. I think margins which contribute to vertical rhythm should be in `em` to stay coherent with the text size.

Before | After
------------ | -------------
main-heading-horizontal-scroll | fixed-main-heading-horizontal-scroll

### DIFF
--- a/media/stylus/components/wiki/document-head.styl
+++ b/media/stylus/components/wiki/document-head.styl
@@ -3,6 +3,9 @@
     clear: both;
 
     h1 {
-        margin-bottom: 20px;
+        word-break: break-all;
+        line-height: 1.2;
+        margin-top: 7px;
+        margin-bottom: 27px;
     }
 }


media/stylus/components/wiki/document-head.styl
```diff
@@ -3,6 +3,9 @@
     clear: both;
 
     h1 {
-        margin-bottom: 20px;
+        word-break: break-all;
+        line-height: 1.2;
+        margin-top: .1458em;
```
stephaniehobson: Sorry, the vertical rhythm has been specified by the designer in pixels, we can't deviate from the design in one place without reconsidering all margins in the entire wiki and consulting with the design team. Please leave the original margins.

stephaniehobson:
Please go back to the margins in the original design otherwise this is a great simple fix :)

tregagnon:
The margins respect visually those defined originally in the design (values slightly increased because of the reduced line-height).

The relatively sized margins insure there will always be a coherent white space for the current font-size.
For example, when dealing with Arabic or logograms (Chinese, Japanese) that might need an increased font size for the same readability.

But I understand you want to keep the ongoing convention. I only regret that the design didn't take in account the adaptability expected from the web. After all, MDN is supposed to be a reference. :wink: 

I will change it.

darkwing:
I did not implement this because of the following:

Broken words

That, IMO, looks much worse for *every* heading than the rare times that an API name is super long and pushes the screen beyond device.  This problem isn't as easy as the PR suggests and I don't recommend it be merged.

stephaniehobson:
Well, darn, I agree that's a show stopper :(

We could look at `hyphens` instead. Combined with `hyphenate-limit-zone` and `hyphenate-limit-chars` to try to avoid problems like the one @darkwing found. 

Is that something you would be interested in trying @tregagnon?

darkwing:
I know it wont fix the issue but we can probably make H1's smaller.

openjck:
Looks like this also needs a rebase. :frowning:

darkwing:
Closing this as I believe @stephaniehobson created a PR I merged yesterday to address the issue.  Please reopen if I'm wrong, @stephaniehobson !

PR #3145, opened 2015-03-29
Add cacheback job for search filters.

jezdez:


robhudson:
django-cacheback looks awesome and this was perfect.

PR #3146, opened 2015-03-29
Fix bug 1148894: Show featured demos on homepage

openjck:


kuma/landing/views.py
```diff
@@ -12,9 +12,9 @@
 
 def home(request):
     """Home page."""
-    demos = (Submission.objects
-                       .all_sorted('upandcoming')
-                       .exclude(hidden=True))[:12]
+    demos = (Submission.objects.filter(featured=True)
+                               .exclude(hidden=True)
+                               .order_by('-modified').all()[:12])
```
jezdez: The `all()` at the end is not needed here.

As there is some infrastructure for sorting is already in place, I wonder if you want to add that type of filtered queryset to the `all_sorted` method as another option? Something like `Submission.objects.all_sorted('freshfeatured')` that then filters on featured and orders by modified as you have it here?

openjck:
I think that's a great idea. Will do.

openjck:
Updated! :smile:

kuma/demos/models.py
```diff
@@ -361,6 +361,10 @@ def all_sorted(self, sort=None):
             return queryset.order_by('-likes_total')
         elif sort == 'upandcoming':
             return queryset.order_by('-likes_recent', '-launches_recent')
+        elif sort == 'recentfeatured':
+            return queryset.filter(featured=True) \
```
jezdez: Please don't use a backslash but parentheses.

openjck:
Updated! :fireworks:

jezdez:
Nicely done! LGTM

jezdez:
The test breaking is our heisenbug with regard to uploaded files.

PR #3147, opened 2015-03-29
Bug 1148743 - Add button to open samples in JSFiddle

iakshay:
Code to insert buttons using javascript.

```js    
    $('.sample-code-frame, .sample-code-table').parent().prepend(function() {
        return '<div><a class="button open-in-host" data-host="jsfiddle" data-section="' + 
        $(this).children('iframe').attr('id').substring("frame_".length) + '">Open in JSFiddle</a>'
        + '<a style="margin-left:10px" class="button open-in-host" data-host="codepen" data-section="' + 
        $(this).children('iframe').attr('id').substring("frame_".length) + '">Open in Codepen</a></div>';
    });

### DIFF
--- a/media/js/wiki.js
+++ b/media/js/wiki.js
@@ -848,7 +848,7 @@
         initDetailsTags();
     }
     
-    function create_jsfiddle(title, htmlCode, cssCode, jsCode) {
+    function openJSFiddle(title, htmlCode, cssCode, jsCode) {
        var $form = $('<form method="post" target="_blank" action="https://jsfiddle.net/api/mdn/">'
             + '<input type="hidden" name="html" />'
             + '<input type="hidden" name="css" />'
@@ -864,7 +864,7 @@
        $form.submit();
     }
     
-    function create_codepen(title, htmlCode, cssCode, jsCode) {
+    function openCodepen(title, htmlCode, cssCode, jsCode) {
        var $form = $('<form method="post" target="_blank" action="http://codepen.io/pen/define">'
             + '<input type="hidden" name="data">'
             + '<input type="submit" />'
@@ -874,11 +874,11 @@
        $form.submit();
     }
            
-    function create_sample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
+    function openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
        if(sampleCodeHost == 'jsfiddle') {
-            create_jsfiddle(title, htmlCode, cssCode, jsCode);
+            openJSFiddle(title, htmlCode, cssCode, jsCode);
         } else if(sampleCodeHost == 'codepen') {
-            create_codepen(title, htmlCode, cssCode, jsCode);
+            openCodepen(title, htmlCode, cssCode, jsCode);
         } 
     }
 
@@ -904,10 +904,33 @@
             var jsCode = $sample.find('.brush\\:.js, .brush\\:.js\\;').text();
             var title = $sample.find('h2[name=' + section + ']').text();
             
-            create_sample(sampleCodeHost, title, htmlCode, cssCode, jsCode);
+            openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode);
             
             $button.removeAttr('disabled');
         });
     });
 
+    function createSampleButtons(section, sites) {
+        var buttons = '<div style="margin-bottom:10px;">';
+        sites.forEach(function(site){
+            // convert sitename to lowercase for icon name and host identifier
+            buttons += '<button class="open-in-host" data-host="'+ site.toLowerCase() +'" data-section="' + 
+                section + '"><i aria-hidden="true" class="icon-'+ site.toLowerCase() +'"></i> Open in ' + 
+                site +'</button>';
+        });
+        buttons += '</div>';
+        return buttons;
+    }
+
+    mdn.insertSampleButtons = function(sites) {
+        // using id to get sample code section since macros discard other attributes
+        $('.sample-code-frame').before(function() {
+            var section = $(this).attr('id').substring("frame_".length);
+            return createSampleButtons(section, sites);
+        });
+        $('.sample-code-table').before(function(){
+            var section = $(this).find('iframe').attr('id').substring("frame_".length);
+            return createSampleButtons(section, sites);
+        }); 
+    }
 })(window, document, jQuery);
--- a/media/js/wiki.js
+++ b/media/js/wiki.js
@@ -847,5 +847,67 @@
     if($('details').length){
         initDetailsTags();
     }
+    
+    function create_jsfiddle(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="https://jsfiddle.net/api/mdn/">'
+            + '<input type="hidden" name="html" />'
+            + '<input type="hidden" name="css" />'
+            + '<input type="hidden" name="js" />'
+            + '<input type="hidden" name="title" />'
+            + '<input type="hidden" name="wrap" value="b" />'
+            + '<input type="submit" />'
+        + '</form>');
+       $form.find('input[name=html]').val(htmlCode);
+       $form.find('input[name=css]').val(cssCode);
+       $form.find('input[name=js]').val(jsCode);
+       $form.find('input[name=title]').val(title);
+       $form.submit();
+    }
+    
+    function create_codepen(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="http://codepen.io/pen/define">'
+            + '<input type="hidden" name="data">'
+            + '<input type="submit" />'
+        + '</form>');
+       var data = {'title': title, 'html': htmlCode, 'css': cssCode, 'js': jsCode};
+       $form.find('input[name=data]').val(JSON.stringify(data));
+       $form.submit();
+    }
+           
+    function create_sample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
+       if(sampleCodeHost == 'jsfiddle') {
+            create_jsfiddle(title, htmlCode, cssCode, jsCode);
+        } else if(sampleCodeHost == 'codepen') {
+            create_codepen(title, htmlCode, cssCode, jsCode);
+        } 
+    }
+
+    $('#wikiArticle').on('click', 'button.open-in-host', function(){
+        var $button = $(this);
+        var section = $button.attr('data-section');
+        var sampleCodeHost = $button.attr('data-host');
+        
+        // track the click and sample code host
+        mdn.analytics.trackEvent({
+            category: 'Samples',
+            action: 'open-' + sampleCodeHost,
+            label: section
+        });
+        
+        // disable the button, till we open the fiddle
+        $button.attr('disabled', 'disabled');
+        var sampleUrl = window.location.href.split('#')[0] + '?section=' + section + '&raw=1';
+        $.get(sampleUrl).then(function(sample) {
+            var $sample = $('<div />').append(sample);
+            var htmlCode = $sample.find('.brush\\:.html, .brush\\:.html\\;').text();
+            var cssCode = $sample.find('.brush\\:.css, .brush\\:.css\\;').text();
+            var jsCode = $sample.find('.brush\\:.js, .brush\\:.js\\;').text();
+            var title = $sample.find('h2[name=' + section + ']').text();
+            
+            create_sample(sampleCodeHost, title, htmlCode, cssCode, jsCode);
+            
+            $button.removeAttr('disabled');
+        });
+    });
 
 })(window, document, jQuery);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by stephaniehobson** [comment]
We can add a line to forms.styl which adds the margin to any button after the first button and then we can localize it for right to left display as well.

.open-in-host + .open-in-host {
    bidi-style(margin-left, ($grid-spacing / 2), margin-right, 0);
}

File: media/js/wiki.js [hunk 1]
```diff
@@ -847,5 +847,73 @@
     if($('details').length){
         initDetailsTags();
     }
+    
+    function create_jsfiddle(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="https://jsfiddle.net/api/post/library/pure/">'
+            + '<input type="hidden" id="html" name="html" />'
```
darkwing: Do we need any of these `id` attributes?  These seem unnecessary.  Instead of `$form.find('#html')`, can you do `$form.find('input[name=html')`?

media/js/wiki.js
```diff
@@ -847,5 +847,73 @@
...
+            + '<input type="hidden" id="css" name="css" />'
+            + '<input type="hidden" id="js" name="js" />'
+            + '<input type="hidden" id="panel_html" name="panel_html" value="0" />'
+            + '<input type="hidden" id="panel_css" name="panel_css" value="0" />'
+            + '<input type="hidden" id="panel_js  name="panel_js value="0" />'
+            + '<input type="hidden" id="title" name="title" value="' + title +'" />'
```
darkwing: This seems unsafe because if the title includes `"`, that could become an injection problem.  Should use `val()` like you do below.

media/js/wiki.js
```diff
@@ -847,5 +847,73 @@
...
+        
+        // track the click and sample code host
+        mdn.analytics.trackEvent({
+            category: 'Samples',
+            action: 'open-' + sampleCodeHost,
+            label: section
+        });
+        
+        // disable the button, till we open the fiddle
+        if($button.hasClass('disabled')) return;
```
darkwing: Should this go above the event tracking code?

media/js/wiki.js
```diff
@@ -847,5 +847,73 @@
...
+           
+    function create_sample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
+       if(sampleCodeHost == 'jsfiddle') {
+            create_jsfiddle(title, htmlCode, cssCode, jsCode);
+        } else if(sampleCodeHost == 'codepen') {
+            create_codepen(title, htmlCode, cssCode, jsCode);
+        } 
+    }
+
+    $('#wikiArticle').on('click', 'a.open-in-host', function(){
```
stephaniehobson: Since these links will not work without javascript enabled we should use the HTML button element instead of links. (links without the href element cannot receive keyboard focus).

media/js/wiki.js
```diff
@@ -847,5 +847,73 @@
...
+        var $button = $(this);
+        var isDisabled = $button.attr('data-disabled');
```
stephaniehobson: When we switch to using a `<button>` instead of a `<a>` we can also use the HTML attribute disabled instead of creating our own. The browser then disables the button for us and we don't need to worry about writing code to manage the disabled button, so you could remove line 902 as well :)

As a note, the disabled attribute is boolean, which means have to add and remove the attribute not switch between true and false.

stephaniehobson:
Wanted to add: super excited to see a PR for this. It's a feature everyone has been talking about for a while!

media/js/wiki.js
iakshay: I thought `id` would be better performance-wise, although in this case since we have very few elements - it shouldn't make much difference.

media/js/wiki.js
iakshay: Fixed

media/js/wiki.js
iakshay: You are right :)

media/js/wiki.js
iakshay: `<button>` isn't allowed in macros. So if keep it as a link, we can add it using macros in future. 

I'm switching to `<button>` for now.

iakshay:
Good for another review :)

I also made minor change in `EmbedLiveSample` macro which I have posted [here][0]

[0]: https://gist.github.com/iakshay/18dba5dac3d6ce99967d

iakshay:
Do we want to insert the buttons using optimizely?

groovecoder:
IIRC, @openjck wanted to use optimizely to change the order of the buttons. So we may not need to insert them via optimizely, but we may run an Optimizely experiment that re-orders them for 50% of viewers.

openjck:
Having had some time to think about it, I would recommend just measuring clicks with Google Analytics. Switching the order of the buttons with Optimizely (or switching between showing one and showing another) would be [bad for habituation](http://blog.openjck.com/context-menus-and-the-design-of-design/).

If one button is clicked way more than the other, we could decide to support only that service. If the buttons receive similar numbers of clicks, we could decide to support both services. I don't expect button placement will have a very significant effect on number of clicks.

groovecoder:
If you'd like to try it out, go for it. For now, @iakshay - could we put this behind a waffle flag so we can merge and push it while @openjck decides (or not) to try out Optimizely?

media/stylus/base/elements/forms.styl
```diff
@@ -114,3 +114,7 @@ label {
         text-decoration: none;
     }
 }
+
+.open-in-host + .open-in-host {
+    bidi-style(margin-left, ($grid-spacing / 2), margin-right, 0);
+}
```
stephaniehobson: Needs a line break at end of file (picky GitHub is picky).

stephaniehobson:
All my comments have been addressed, thank you :) It would be nice if the sample code in the commit message was also updated.

openjck:
I'm definitely up for doing an A/B test. I'll coordinate with @iakshay offline to get it going.

iakshay:
Update code to insert buttons

```js
$('.sample-code-frame, .sample-code-table').parent().prepend(function() {
        return '<div><button class="open-in-host" data-host="jsfiddle" data-section="' + 
        $(this).children('iframe').attr('id').substring("frame_".length) + '"><i aria-hidden="true" class="icon-jsfiddle"></i> Open in JSFiddle</i></button>'
        + '<button class="open-in-host" data-host="codepen" data-section="' + 
        $(this).children('iframe').attr('id').substring("frame_".length) + '"><i aria-hidden="true" class="icon-codepen"></i> Open in Codepen</button></div>';
    });

File: media/js/wiki.js [hunk 7]
```diff
@@ -847,5 +847,67 @@
...
+        });
+        
+        // disable the button, till we open the fiddle
+        $button.attr('disabled', 'disabled');
+        var sampleUrl = window.location.href.split('#')[0] + '?section=' + section + '&raw=1';
+        $.get(sampleUrl).then(function(sample) {
+            var $sample = $('<div />').append(sample);
+            var htmlCode = $sample.find('.brush\\:.html, .brush\\:.html\\;').text();
+            var cssCode = $sample.find('.brush\\:.css, .brush\\:.css\\;').text();
+            var jsCode = $sample.find('.brush\\:.js, .brush\\:.js\\;').text();
```
iakshay: @darkwing Will these selectors match all our code sample snippets?

I have written selectors for the following 2 variations
```html
// Case 1
<pre class="brush: js"></pre>
// Case 2
<pre class="brush: js; highlight:[7]"></pre>

Comment (iakshay) [inline media/js/wiki.js]: Also, I'm not sure if converting code to base64 is necessary.

Comment (openjck) [inline media/js/wiki.js]: I was wondering the same thing. What do you think, @darkwing?

Comment (zalun) [inline media/js/wiki.js]: It is not - see example in https://jsfiddle.net/zalun/mk29ttc2/

Comment (openjck) [inline media/js/wiki.js]: > Will these selectors match all our code sample snippets?

12:09 <davidwalsh> I believe so
12:09 <davidwalsh> I think taht's right

**Comment by openjck** [comment]
How would you feel about writing the [insertion code](https://github.com/mozilla/kuma/pull/3147#issue-65059302) into the repository, so that we would run something like `mdn.injectFiddleButton()` and `mdn.injectCodepenButton` from Optimizely?

The code is non-trivial, so it makes sense for us to revision it. We also want to be sure that it stays synchronized with other changes happening in the codebase.

**Comment by openjck** [comment]
Maybe we could even add the buttons, but use `display: none` on them by default. That would be easier to maintain. We used the same approach in the A/B testing groundwork in #3142.

**Comment by iakshay** [comment]
I have created  `mdn.insertSampleButtons(sites)`. To insert the buttons you refer use this snippet

````js
// insert JSFiddle button
mdn.insertSampleButtons(['jsfiddle']);
// insert JSFiddle and Codepen buttons
mdn.insertSampleButtons(['jsfiddle', 'codepen']);
// other order
mdn.insertSampleButtons(['codepen', 'jsfiddle']);

**Comment by openjck** [comment]
What would be a good page to spot-check this on? I'm having trouble finding pages that use `.sample-code-frame` or `.sample-code-table`.

Since this involves the live samples system, which @darkwing has done the most work with, I think it would be great to get his opinion, too.

File: media/js/wiki.js [hunk 8]
```diff
@@ -847,5 +847,90 @@
...
+            var title = $sample.find('h2[name=' + section + ']').text();
+            
+            openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode);
+            
+            $button.removeAttr('disabled');
+        });
+    });
+
+    function createSampleButtons(section, sites) {
+        var buttons = '<div style="margin-bottom:10px;">';
```
darkwing: We shouldn't inline styles -- a CSS class should be created.

media/js/wiki.js
```diff
@@ -847,5 +847,90 @@
     if($('details').length){
         initDetailsTags();
     }
+    
+    function openJSFiddle(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="https://jsfiddle.net/api/mdn/">'
```
darkwing: Instead of a large string, let's create the elements with JS.  We retrieve most of them to set their values anyway.

media/js/wiki.js
```diff
@@ -847,5 +847,90 @@
...
+        + '</form>');
+       $form.find('input[name=html]').val(htmlCode);
+       $form.find('input[name=css]').val(cssCode);
+       $form.find('input[name=js]').val(jsCode);
+       $form.find('input[name=title]').val(title);
+       $form.submit();
+    }
+    
+    function openCodepen(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="http://codepen.io/pen/define">'
```
darkwing: Same here -- let's create these elements programatically

iakshay:
@openjck - I modified the `EmbedLiveSample` macro to add those classes. Update macro is posted [here] [0]

[0]:https://gist.github.com/iakshay/18dba5dac3d6ce99967d

media/js/wiki.js
iakshay: I think creating elements programatically will make the code too long and unreadable. But I'm happy to do it, if this breaks our style guide.

Also, should I move my code into separate file. IIRC, we want to waffle flag this feature as well as A/B test.  Is this part as simple as creating separate js file, or do I need to do waffle flag checking in js also?

Let me know if there is anything else that needs to be fixed. I want this feature in production asap :)

iakshay: @darkwing ^^

groovecoder:
ping @darkwing - are you able to review this again?

media/js/wiki.js
darkwing: Let's waffle this, please

openjck: Just to be sure, did you see that the buttons need to be manually inserted with JS?

iakshay: Would it be suffice to move the code to a new file for waffle flag or is there more to it? Can you link me to a commit that does something similar?

groovecoder:
@iakshay - are you able to waffle this?

iakshay:
@groovecoder I'm not sure, how we waffle features client side. Do I just move the code to new file?

groovecoder:
We use a `django-waffle` helper to create a `window.waffle` object. We use(d) it in [`demos/home.html`](https://github.com/mozilla/kuma/blob/d9f09acc3badba3c05cbba9e05448b20e83e1505/kuma/demos/templates/demos/home.html#L317-L325) and in [`ckeditor_config.js`](https://github.com/mozilla/kuma/blob/d9f09acc3badba3c05cbba9e05448b20e83e1505/kuma/wiki/templates/wiki/ckeditor_config.js#L105-L107)

openjck:
Do we need to Waffle this? Even on the branch, the buttons need to be manually inserted with JS.

darkwing:
This does need to be waffled, yes.

zalun:
any update?

darkwing:
Superceded by: https://github.com/mozilla/kuma/pull/3240

PR #3149, opened 2015-03-30
bug 1145793 - Remove references to old redesign class

darkwing:
We don't need this class anmore \o/

### DIFF
--- a/media/stylus/components/wiki/customcss.styl
+++ b/media/stylus/components/wiki/customcss.styl
@@ -909,22 +909,11 @@ dl {
 .indicatorInHeadline,
 .blockIndicator,
 .overheadIndicator {
-/*box-shadow: 1px 1px 1px #CCC*/
     margin: 0px 0px 1.5em;
     border: 5px solid;
-    padding: 10px 10px 0px 10px;
+    padding: 10px;
     background: none repeat scroll 0 0;
     font-size: 12px;
-}
-.redesign .inlineIndicator,
-.redesign .indicatorInHeadline,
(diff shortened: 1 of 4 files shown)


stephaniehobson:
Lines 1609-1629 of the customcss.style file need to be updated as well.

stephaniehobson:
Glad to see this going! We need to update the sphinx templates to remove the class as well but we can create a separate bug for that, it's not merge blocking.

darkwing:
Good call Stephanie!  Updated!

media/stylus/components/wiki/customcss.styl
```diff
@@ -909,23 +909,12 @@ dl {
 .indicatorInHeadline,
 .blockIndicator,
 .overheadIndicator {
-/*box-shadow: 1px 1px 1px #CCC*/
     margin: 0px 0px 1.5em;
     border: 5px solid;
-    padding: 10px 10px 0px 10px;
+    padding: 10px;
```
stephaniehobson: These styles are being over-ridden by a declaration on line 948 and the inlineIndicators don't look right anymore. Before the added specificity of the redesign class was preventing this conflict.

I suspect but am not certain that we can just remove inlineIndicator from the other declaration.

Good testing page for this: https://developer.mozilla.org/en-US/docs/User:teoli/In_content

stephaniehobson: Looks like it's also a problem with `.overheadIndicator` when used in combination with `.standardNoteBlock` from line 979.

PR #3153, opened 2015-04-01
Bug 1145793: Remove .htaccess redesign redirect

openjck:
We added this redirect to .htaccess experimentally. We knew it had no
effect locally and were wondering if it would have any effect on our
stage and production servers.

This ultimately had no effect on our servers, so we asked WebOps [1] to
define this redirect in our main Apache config instead.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1145793


darkwing:
R+, works for me.

PR #3156, opened 2015-04-03
Fix bug 1117059 $revision doesn't list the tags of the given revision

nhenezi:


groovecoder:
Looks good. Since this is mostly copy/paste from `document.html`, can you move it into a template include to eliminate code duplication? So create a new `kuma/wiki/templates/wiki/includes/document_tags.html` and include it in both `document.html` and `revision.html`?

nhenezi:
So the problem is that `tags` in `document.html` are objects while `tags` in `revision.html` are strings.

Since expression statements aren't enabled in Jinja2 I cannot do this in `document.html`:
```python
    {% set tags = [] %}
    {% for tag in document.tags.all() %}
      {% do tags.append(tag.name %}
    {% endfor %}
    {% include "wiki/includes/document_tag.html" with context %}
```

And since Jinja2 `v2.5.2` doesn't have filter `map` (it's added in `v2.6`) I cannot do something like
```python
    {% set tags = document.tags.all() | map(attribute='name') %}
    {% include "wiki/includes/document_tag.html" with context %}
```

Another approach would be to define a macro with `tag_name` parameter which then would be evaluated (e.g. in `document.html` you would pass `'tag.name`, while in `revision.html` you would pass `tag`), but that seems like overkill.

How do you handle when you have to convert a list of objects to a list of object attributes in templates? E.g in python:

```python
   tags = [t.name for t in document.tags.all()]
```


Thanks!

groovecoder:
Hmm ... I mean the code in each file starting with:
```
{% set tags = document.tags.all() %}
```
in both [revision.html](https://github.com/nhenezi/kuma/blob/e2df63ef1275cd666a3fc5c64999686914832c77/kuma/wiki/templates/wiki/revision.html#L56), and [document.html](https://github.com/nhenezi/kuma/blob/e2df63ef1275cd666a3fc5c64999686914832c77/kuma/wiki/templates/wiki/document.html#L370). The only part of the code using the tag is the `url()` helper, and it seems to work with either? So could we just change both `document.html` and `revision.html` to pass the tag object?

nhenezi:
Ah, didn't know that url helper worked with both. I've pushed a fixup - let me know if it's okay and if you want it squashed.

kuma/wiki/templates/wiki/revision.html
```diff
@@ -52,6 +52,11 @@
       </div>
     </details>
 
+    <div class="wiki-block contributors">
+      {% set tags = document.tags.all() %}
+      {% include "wiki/includes/document_tag.html" with context %}
```
groovecoder: `with context` is the default behavior for `include`s, so we typically leave it off our `include` tags.

kuma/wiki/templates/wiki/document.html
```diff
@@ -368,13 +368,7 @@
...
                   <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
                   {% set tags = document.tags.all() %}
-                  {% if tags.exists() %}
-                  <!-- tags if present -->
-                  <div class="tag-attach-list contributors-sub">
-                    <i aria-hidden="true" class="icon-tags"></i><strong>{{ _('Tags:') }}</strong>&nbsp;
-                    <ul class="tags tags-small">{% for tag in tags %}<li><a href="{{ url('wiki.tag', tag.name) }}">{{ tag.name }}</a></li>{% endfor %}</ul>
-                  </div>
-                  {% endif %}
+                  {% include "wiki/includes/document_tag.html" with context %}
```
groovecoder: `with context` is the default behavior for `include`s, so we typically leave it off our `include` tags.

groovecoder:
Looks good, with comments. Go ahead and squash please. Thanks!

PR #3131, opened 2015-03-20
Fixes bug 632204 - Remove reCAPTCHA

utkbansal:
* removes the captcha form field from forms and template
* removes django-recaptcha submodule
* remove the recaptcha settings from settings.py

PR #3158, opened 2015-04-05
fix bug 632204 - Removed captcha field from demo submission form

kvsingh:
Removed captcha field from demo submission form and from the corresponding template. Also removed the assertion for captcha in test_views.py


groovecoder:
@utkbansai already has https://github.com/mozilla/kuma/pull/3131 in for this bug.

PR #3161, opened 2015-04-06
Experiment with lazyloading Tabzilla

darkwing:
@stephaniehobson:  This is the best course for lazy loading Tabzilla from a logical perspective.  A few things:

1.  Appending the stylesheet to the `head` tag would give Tabzilla priority over our styles so it seems a bad idea

2.  I've used `prependTo` to put the stylesheet at the top of the head, but I think that's causing another repaint.

Let me know what you think of this.  Obviously not ready to merge, but just wanted to put this out there.

### DIFF
--- a/media/js/main.js
+++ b/media/js/main.js
@@ -221,11 +221,17 @@
         Tabzilla
     */
     if($('#tabzilla').length) {
-        $.ajax({
-            url: '//mozorg.cdn.mozilla.net/en-US/tabzilla/tabzilla.js',
-            dataType: 'script',
-            cache: true
-        });
+        $('<link />').attr({
+            href: '//mozorg.cdn.mozilla.net/media/css/tabzilla-min.css',
+            type: 'text/css',
+            rel: 'stylesheet'
+        }).on('load', function() {
+            $.ajax({
+                url: '//mozorg.cdn.mozilla.net/en-US/tabzilla/tabzilla.js',
+                dataType: 'script',
+                cache: true
+            });
+        }).prependTo(doc.head);
     }
 
 })(window, document, jQuery);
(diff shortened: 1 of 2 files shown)


darkwing:
I could probably make tabzilla `opacity: 0` until loaded but it appears there's a "jump" too, we may have to modify height.

stephaniehobson:
I suggest we add this CSS to our own files. If tabzilla changes it'll be big news and we'll have time to make changes :)

```
#tabzilla {
    bidi-value(float, right, left); /* already in our CSS */
    display: block;
    height: 44px;
    width: 150px;
    overflow: hidden;
}

#tabzilla-panel {
    height: 0px; /* in case the javascript loads before the CSS */
}
```

The lazy loading is working for me without any repaints so I'm not sure you need to make changes. But you sounded unsure if you'd found the best place to attach it so I thought I'd mention another possibility is appending it to the `<body>`.

groovecoder:
Is there a mechanism for monitoring tabzilla changes, other than "whoa the prod site looks funky"? :wink:

darkwing:
@groovecoder I have that same concern, which is why I mentioned we may just want to leave things be.

stephaniehobson:
They've been announcing changes in webdev before making them (though, I don't think they've actually made any since I was hired).

To be honest, not copying the CSS doesn't protect us against "whoa the prod site looks funky".

stephaniehobson:
Thinking about this I'm going to go a step further and say adding these lines protects us against the production site looking funky.  Right now we have no protection if the tabzilla JS loads and the CSS does not.

I also think that lazy loading the fonts is more important than loading tabzilla so if this is a blocker we should remove tabzilla from MDN completely.

groovecoder:
That makes sense to me.

darkwing:
@stephaniehobson Wanna throw your CSS toward this per your ideas so we can push this forward?

stephaniehobson:
Superseded by #3181

PR #3174, opened 2015-04-11
fix bug 1104289 - added footer to display document id in document.html template 

kvsingh:


stephaniehobson:
Thanks for re-submitting @kvsingh :)

Looks good to me from a front-end, standards perspective. @groovecoder Can you look this over from a python perspective?

kuma/wiki/templates/wiki/document.html
```diff
@@ -367,6 +368,12 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
                       {{ datetimeformat(current_revision.created, format='datetime') }}
                     </div>
                   {% endif %}
+		 {% if is_admin %}
+		    <div class="contributors-sub">
+		      Document number: {{ document.id }}
```
groovecoder: This needs l10n marking to enable translation ...
```
{% trans id=document.id %}
Document ID: {{ id }}
{% endtrans %}
```

stephaniehobson: There's a how to guide for this here: http://kuma.readthedocs.org/en/latest/localization.html

PR #3177, opened 2015-04-13
Bug 1151523 - intrinsic ratio youtube iframes

stephaniehobson:
Using the switch to the macro for youtube imbeds to take this opportunity to improve how the videos display at different screen sizes.

Works in concert with https://developer.mozilla.org/en-US/docs/Template:EmbedYouTube
Here's some content with videos you can alter to test: https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Work_with_animations

Testing:
- create the macro on your local test environment
- fire up kumascript
- embed some videos like so `{{ EmbedYouTube('mDHtLK88ZW4') }}`
- resize your browser window and watch the ratio be maintained

### DIFF
--- a/media/stylus/components/wiki/intrinsic.styl
+++ b/media/stylus/components/wiki/intrinsic.styl
@@ -1,0 +1,31 @@
+/*
+Creates container that preserves aspect ratio, and fills container with iframe
+http://alistapart.com/article/creating-intrinsic-ratios-for-video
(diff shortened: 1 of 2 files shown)


stephaniehobson:
The macro changes are blocking the writers from converting youtube iframes to the macro. This PR should be considered a high priority for reviewing.

Tagging @darkwing because the iframe conversions are his idea but anyone can jump in :)

darkwing:
Works for me!  R+

PR #3178, opened 2015-04-13
Fix Bug 1149205 - Add <wbr> to article headings

stephaniehobson:
In which I write python code which is probably not suitable for publication but does at least come with a test.

We need to add intelligent line breaks to really long article headings, especially on smaller screens. Inserting <wbr> tags before periods seems like a good option.

Two things that likely have to change:
1) I used safe on the document titles assuming that what we do to them on the way into the database makes them safe. I could be wrong. Maybe there is a better way to do this?
2) I assumed there will never be any "." in code in the document titles. The find replace is at least smart enough not to break up decimal numbers but not smart enough to leave something like a url in a src attribute alone.

### DIFF
--- a/kuma/wiki/helpers.py
+++ b/kuma/wiki/helpers.py
@@ -24,6 +24,8 @@ from .constants import DIFF_WRAP_COLUMN
 from .jobs import DocumentZoneStackJob
 from .models import Document, memcache
 
+BREAKS_RE = re.compile(r'(?<=[a-z,A-Z])\.(?=[a-z,A-Z])')
+
 register.function(build_policy_admin_links)
 
 
@@ -355,3 +357,12 @@ def wiki_url(context, path):
     # finally cache the reversed document URL for a bit
     memcache.set(u'wiki_url:%s:%s' % (locale, path), url, 60 * 5)
     return url
+
+
+@register.filter
+def addbreaks(title):
+    """
+    adds <wbr> tags before . that appear between letters
+    no numbers should be split
+    """
+    return BREAKS_RE.sub('<wbr>.', title)
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -172,7 +172,8 @@
             <a class="button from-search-next only-icon disabled" title="{{ _('Next Search Result') }}" data-empty-title="{{ _('No Previous Search Result') }}"><i aria-hidden="true" class="icon-chevron-right"></i></a>
           </span>
 
-        <h1>{{ document.title }}</h1>
+        {% set safe_title = document.title|forceescape %}
+        <h1>{{ safe_title|addbreaks|safe }}</h1>
 
         {% if waffle.flag('share_links') %}
           {{ share_links(document) }}
--- a/kuma/wiki/tests/test_helpers.py
+++ b/kuma/wiki/tests/test_helpers.py
@@ -7,7 +7,7 @@ from django.contrib.sites.models import Site
 from kuma.core.cache import memcache
 from kuma.users.tests import UserTestCase
 from kuma.wiki.helpers import (absolutify, document_zone_management_links,
-                               revisions_unified_diff, tojson)
+                               revisions_unified_diff, tojson, addbreaks)
 from kuma.wiki.models import DocumentZone
 from kuma.wiki.tests import revision, WikiTestCase
 
@@ -17,6 +17,9 @@ class HelpTests(WikiTestCase):
     def test_tojson(self):
         eq_(tojson({'title': '<script>alert("Hi!")</script>'}),
             '{"title": "&lt;script&gt;alert(&quot;Hi!&quot;)&lt;/script&gt;"}')
+
+    def test_addbreaks(self):
+        eq_(addbreaks('this.is.a.test.1.1'), 'this<wbr>.is<wbr>.a<wbr>.test.1.1')
 
     @mock.patch.object(Site.objects, 'get_current')
     def test_absolutify(self, get_current):


kuma/wiki/helpers.py
```diff
@@ -355,3 +355,12 @@ def wiki_url(context, path):
     # finally cache the reversed document URL for a bit
     memcache.set(u'wiki_url:%s:%s' % (locale, path), url, 60 * 5)
     return url
+
+
+@register.function
```
jezdez: I'd rather use a `@register.filter` here since that makes this function be available as something akin to shell pipes, something that is more in line with the template language and the actual thing you're doing in the function. The call would then be: `{{ document.title|addbreaks|safe }}`

kuma/wiki/helpers.py
```diff
@@ -355,3 +355,12 @@ def wiki_url(context, path):
...
+def addbreaks(title):
+    """
+    adds <wbr> tags before . that appear between letters
+    no numbers should be split
+    """
+    return re.sub(r"(?<=[a-z,A-Z])\.(?=[a-z,A-Z])", "<wbr>.", title)
```
jezdez: That looks sensible, except that we should separate the compilation of the regex pattern and the call to sub into two different pieces. What happens when you call `re.sub` is that it takes the raw string you pass and then compiles it into a pattern object, which can be time costly. If you instead compile the pattern manually on the module level (outside the function), it'll only do that compilation once per server process instead on every call. So this should be:
```python
BREAKS_RE = re.compile(r'(?<=[a-z,A-Z])\.(?=[a-z,A-Z])')

def addbreaks(title):
    """
    adds <wbr> tags before . that appear between letters
    no numbers should be split
    """
    return BREAKS_RE.sub('<wbr>.', title)
```
Please put `BREAKS_RE` to the top of the module.

jezdez:
@stephaniehobson! I don't know what you're talking about, that looks like totally legit Python code. Good work!

About your question regarding the safety of the title, I'm not sure about that. Maybe @groovecoder can shed some light on that? In general I lean towards to always strip tags on the way in, but I'm not sure about old legacy content.
I also don't know about your second question, sadly. Paging @groovecoder as the resident person-who-has-been-around-longest.

groovecoder:
1. The content has contained un-restricted markup since MindTouch, so we haven't sanitized markup on the way in to preserve integrity of what editors enter. It's an on-going challenge. :neutral_face: 

2. Not sure what you mean by ' "." in code in the document titles' ? Can you give some examples?

stephaniehobson:
@groovecoder Long titles [like on this page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/propertyIsEnumerable) are being treated like a single word and breaking the page at the moment. The "." is a logical break point if we can manage it.

kuma/wiki/helpers.py
stephaniehobson: Wasn't sure what the top of the module is, I put it at the top of helpers.py after the import calls, is that okay?

kuma/wiki/helpers.py
stephaniehobson: Done! The test still runs fine with it called as a function, is it okay to leave the test?

stephaniehobson:
I'd be nice if there was a way to treat the title as hostile but still add some HTML to be rendered.

jezdez:
@stephaniehobson Could we use something like `document.title|striptags|addbreaks|safe`?

stephaniehobson:
@jezdez I don't think we can strip them because we have titles that *are* tags: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

stephaniehobson:
Is there an HTML escape filter?

jezdez:
@stephaniehobson That's the default, that "|safe" turns the escaping off.

jezdez:
All this reminds me that our data sanitation strategy is bad and we need to change it.

kuma/wiki/templates/wiki/document.html
```diff
@@ -172,7 +172,8 @@
             <a class="button from-search-next only-icon disabled" title="{{ _('Next Search Result') }}" data-empty-title="{{ _('No Previous Search Result') }}"><i aria-hidden="true" class="icon-chevron-right"></i></a>
           </span>
 
-        <h1>{{ document.title }}</h1>
+        {% set safe_title = document.title|forceescape %}
+        <h1>{{ safe_title|addbreaks|safe }}</h1>
```
stephaniehobson: @groovecoder Can you explain to me why this doesn't work?

The `<wbr>` tags are added and not escaped in this code: 
```<h1>{{ "this.is.a.test"|addbreaks|safe }}</h1>```

But in this code the `<wbr>` tags are escaped!?!?
```<h1>{{ safe_title|addbreaks|safe }}</h1>```

groovecoder: Above this line there's a `set safe_title = document.title|forceescape` line?

stephaniehobson: Yeah but why is the line before it effecting the output of that line? Why is the filter effecting the variable here instead of when I apply it?

stephaniehobson: This would totally work in JavaScript, is all all I'm saying.

groovecoder: Okay, I think I found out why ...

When `jingo` [registers a filter](https://github.com/jbalogh/jingo/blob/ad909e4830bff6acb1c3f5ca983b6e9c7bf45454/jingo/__init__.py#L119-L137), it [initializes its own jinja2 execution environment](https://github.com/jbalogh/jingo/blob/ad909e4830bff6acb1c3f5ca983b6e9c7bf45454/jingo/__init__.py#L46-L74), with [`autoescape` set to `True`](https://github.com/jbalogh/jingo/blob/ad909e4830bff6acb1c3f5ca983b6e9c7bf45454/jingo/__init__.py#L56). The only way to override the `autoescape` setting of filters would be a global `settings.py` value, which we don't want to do. :cry:

I'll assign this to me and try to think of something we can do here. Or maybe I'll ask dev-webdev.

jezdez: @groovecoder Please don't just assign it to you or ask dev-webdev. Auto escaping is a core ability of kuma and should be handled by its developers. In other words, this is too serious to do in between.

stephaniehobson: @groovecoder Thanks for the explanation :)

We should time box this. There's a couple of other potential solution to this problem (though none as elegant) and I don't want to be introducing security problems.

PR #3179, opened 2015-04-14
Bug 1154283 - Update htmlelt style to fit well with real HTML table

JeremiePat:


darkwing:
Best way to view/test @JeremiePat ?

JeremiePat:
@darkwing : In production I have this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr (with a real table) vs. any other HTML reference page (let's say https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b). 

I updated https://developer.allizom.org/en-US/docs/Web/HTML/Element/abbr to have the same as in production but I don't know why style on allizom doesn't seams to match those on production.

JeremiePat:
Ok, my bad, thanks allizom. I had a bad css cascading rule.

media/stylus/components/wiki/customcss.styl
```diff
@@ -373,6 +377,13 @@ div.index ol {
     clear: right;
     border-right: 6px solid rgba-fallback(rgba(255,149,0,0.5));
 }
+.htmlelt th, .htmlelt td {
```
stephaniehobson: Each selector should be on it's own line, please.

media/stylus/components/wiki/customcss.styl
```diff
@@ -365,6 +365,10 @@ div.index ol {
     padding: 11px 22px;
     background-color: rgba-fallback(rgba(255,149,0,0.1));
 }
+// Necessary to overload .text-content table but not [dir] .htmlelt
+table.htmlelt {
+    border:none;
```
stephaniehobson: Space between the property  and value please.

media/stylus/components/wiki/customcss.styl
```diff
@@ -365,6 +365,10 @@ div.index ol {
     padding: 11px 22px;
     background-color: rgba-fallback(rgba(255,149,0,0.1));
 }
+// Necessary to overload .text-content table but not [dir] .htmlelt
```
stephaniehobson: Please use CSS comments.

stephaniehobson:
This is a great improvement! Can't wait to see more of the tables in production. Much more mobile friendly :) (and I was worried they wouldn't be - but it is much better).

stephaniehobson:
@JeremiePat I took this opportunity to clean up the HTML a little and made the table structure available in all colours, not just htmlel. I hope that's okay!

Superseded by #3196.

JeremiePat:
Thanks @stephaniehobson , I was about to get back to this. Your improvement looks amazing :D

PR #3182, opened 2015-04-15
fix bug 1104289 - added footer to display document id in document.html 

kvsingh:


darkwing:
Thank you for the contribution but the original bug was invalid :  https://bugzilla.mozilla.org/show_bug.cgi?id=1104289

:(

Great work though!

PR #3124, opened 2015-03-18
bug 875062 - Add initial doc sharing

groovecoder:
Supersedes https://github.com/mozilla/kuma/pull/3119 with fixed tests, gettext-wrapped text, and new `utm_medium` per @darkwing's [rider](https://en.wikipedia.org/wiki/Rider_%28theater%29).

PR #3184, opened 2015-04-16
"Google +" to "Google+"

yfdyh000:
I think this "Google +" is strange, usually "Google+" or "Google Plus".

Related pull requests: #3124, #3142.


groovecoder:
@yfdyh000 - thanks for the contribution. Next time, please remember to include a bug # in the commit message, per our [CONTRIBUTING guidlines](https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md).

yfdyh000:
@groovecoder Yes, I know. It is a minimal modification, so I want to skip this step.
> Discuss large changes on the dev-mdn mailing list or on a bugzilla bug before coding.

I think this should apply to this situation.

But anyway, thank you for reminding.

groovecoder:
@yfdyh000 for sure, this didn't need an email to dev-mdn or its own bug. But, we could have re-used bug 875602 in the commit message.

I.e., `bug 875062 - "Google +" to "Google+"`

yfdyh000:
@groovecoder I feel this would be strange if I did not mention this modification in the bug. I mentioned the relevant pull in the pull, and I think this is enough.

groovecoder:
@yfdyh000 - we're certainly a little strange in the way we use both bugzilla and GitHub, so I totally understand. :smile: 

The nice thing about mentioning the bug # in the commit message, is that we have a GitHub hook that will automatically make a comment on the bug when the commit lands in master. So, the bug will automatically get a mention of the modification.

yfdyh000:
@groovecoder Ok, thank you for telling me about this feature.

PR #3185, opened 2015-04-16
bug 1152386 - Adding tests for new, edit, and translate within the wiki

darkwing:
Adds testing for edit and translate screens, ensuring CKEditor loads

### DIFF
--- a/tests/ui/_cli.js
+++ b/tests/ui/_cli.js
@@ -30,15 +30,21 @@ define({
         // If we weren't provided username and password, let's set a grep to avoid login tests
         if(args.u == undefined && args.p == undefined) {
             greps.push('requires-login');
+            console.log('No username (-u) and password (-p) provided.  Tests requiring login will be skipped.');
         }
 
         // Set a document for wiki testing
         if(args.wd == undefined) {
             greps.push('requires-doc');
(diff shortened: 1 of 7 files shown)


darkwing:
Merged the other PRs into this.

tests/ui/_cli.js
```diff
@@ -30,16 +30,22 @@ define({
...
         if(args.wd == undefined) {
             greps.push('requires-doc');
+            console.log('No wiki document (-wd) provided.  Most wiki tests will be skipped.');
         }
 
         // Set the final GREP value
         args.grep = greps.length ? ('^(?!.*?\\[' + greps.join('|') + '\\])') : '';
 
+        if(args.grep) {
+            console.log('Command line arguments have forced a grep to skip tests:  ' + args.grep);
```
openjck: Hm, I'm not sure I understand. Is there another way of phrasing this? What will be different from the point of view of the person running the test?

darkwing: It's basically summarizing the two comments above.

openjck: Ah, so it will appear only when another error message also appears. Okay, that makes sense then.

PR #3189, opened 2015-04-17
fix bug 1155300 - Open revision content by default

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1155300

### DIFF
--- a/kuma/wiki/templates/wiki/revision.html
+++ b/kuma/wiki/templates/wiki/revision.html
@@ -57,19 +57,19 @@
       {% include "wiki/includes/document_tag.html" %}
     </div>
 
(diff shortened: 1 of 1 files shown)


groovecoder:
This technically fixes the bug as written, but the bigger issue is still there (as described in [comment 1 on the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1155300#c1)) - i.e., when the section is closed, the toggle to open it is gone.
revision-content

darkwing:
Updated!  Headers where in the wrong spot!

stephaniehobson:
:+1:

PR #3191, opened 2015-04-20
fix bug 1146711 - add Fulah to MDN

groovecoder:
Spot-check on this branch:
1. [Get the localizations](http://kuma.readthedocs.org/en/latest/localization.html#get-localizations) if you don't already have them.
2.
```
cd locale
svn up
./compile-mo.sh .
```
3. Go to https://developer-local.allizom.org/ff/ and make sure it pulls up.
4. Translate a test document into Pulaar-Fulfulde

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -98,6 +98,7 @@ MDN_LANGUAGES = (
                  'el',
                  'es',
                  'fa',
+                 'ff',
                  'fi',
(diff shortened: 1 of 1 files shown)


darkwing:
Checking

darkwing:
R+  Works as expected.

PR #3192, opened 2015-04-21
Bug 1150118 - lazyload fonts

stephaniehobson:
First pass at lazy loading fonts.  Only lazy loading `Open Sans` and `Open Sans Light` at the moment.

How it works:
1. User visits for first time
2. User does not have right data attributes on `html` element so webfonts do not display
3. Javascript runs loading as many webfonts as possible before either a) downloading all and displaying them, or  b) timing out at 3 seconds and displaying what it has loaded so far. When all fonts are loaded a cookie is set.
4. User goes to new page, if the cookie is set the server includes the data attributes on the `html` tag and webfonts display immediately (if cookie is not set javascript runs again, attempting to load all fonts again)

To test:
- clear cache
- load page, watch for: fallback displays immediately, site and heading fonts swap in simultaneously
- reload page without clearing cache, watch for: site and heading fonts should display immediately
- get a really slow connection
- clear cache
- load page, watch for: on a slow enough connection the fonts should time out, the page should never swap in the site and heading fonts
- reload page without clearing cache, watch for: site and heading fonts may display immediately or may time out again (watch the network panel, if they loaded they should display)

I'm really happy with this. I think it is an improvement on the current first load user experience :D 

Current (prod) in Firefox:
prod

Lazyload in Firefox:
ffo

I know everyone is really skeptical that we should be messing with the experience so I wanted to get some feedback before solving the edge cases.

To do:
- [x] waffle
- [x] check license on FontFaceObserver is compatible with MDN
- [x] find solution to revert to default behavior if JavaScript disabled
- [x] see if anything can be done about hard refresh load when cookie set
- [x] check all uses of `Open Sans` and `Open Sans Light` have fallbacks declared.

To do that might have to wait for a waffled launch:
- [ ] get feedback on user experience to decide on go-ahead
- [ ] test in IE


media/js/libs/fontfaceobserver/fontfaceobserver-standalone.js
```diff
@@ -0,0 +1,41 @@
+/*
+
+https://github.com/bramstein/fontfaceobserver
+v1.4.5
+
+Copyright (c) 2014 - Bram Stein
```
stephaniehobson: @gerv I could use a 3rd opinion on the inclusion of this library if you have a moment. The original repo is here: https://github.com/bramstein/fontfaceobserver I copied only fontfaceobserver-standalone.js from the repo but the file did not include the license so I added the license into the file. I need to double check that 1) the license is compatible with MDN and 2) I am adhering to it.

gerv: The license is compatible. If you add a copy of the full license text to the top of the file, that will be fine.

stephaniehobson:
Woot! Ready for review!

stephaniehobson:
I'm not sold on the stylus technique I'm using for the fallback, they are essentially each being hand-coded, but I'm not sure we can automate this either. What gets tweaked for display is different in each place. This allows the maximum amount of control while also requiring slightly more work.

One problem with the current technique is that if a fallback is not specified the webfonts will be loaded in the regular order. I could fix that by removing the font declarations from the main style sheet and lazy loading the font declarations as well.

I'd like to see this get into production so I can get feedback from a wider audience on it. Several people were quite concerned that we'd be fussing with the default browser behaviour.

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
+(function(win, doc, $) {
+    /*
+        Font Loading
+    */
+
+    if(window.waffle && window.waffle.flag_is_active('web_font_loader')) {
```
darkwing: On line 1 we have `win` which is an alias for window, so we can use that instead of `window` throughout this file.

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+
+        // to collect the results of the observers as they return
+        var fontsLoaded = {};
+
+        // timout for all observers
+        var ffoTimeout = 3000;
+
+        // kick off font loading
+        function ffoStart() {
+            ffoLoad();
```
darkwing: I don't think we need `ffoStart` at all, just call `ffoLoad` within `done`

darkwing: Actually, we can keep it, if it's for mental/maintenance reasons.  That might be good.

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+                "class" : "ffo-opensansextrabold",
+                "varient" : [
+                    {"weight" : "normal"}
+                ]
+            }*/
+        };
+
+        // load the observer plug in
+        $.getScript('/media/js/fontfaceobserver-min.js')
+        .done(function() { ffoStart(); })
```
darkwing: No need for the `function() { ffoStart(); }`, can just do `ffoStart`

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+        .fail(function() { ffoFinished(); });
```
darkwing: Can just be `ffoFinished`

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+
+        // to collect the results of the observers as they return
+        var fontsLoaded = {};
```
darkwing: You'll want to move these vars above the AJAX call (hoisting:  http://www.adequatelygood.com/JavaScript-Scoping-and-Hoisting.html)

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
+(function(win, doc, $) {
+    /*
+        Font Loading
+    */
+
+    if(window.waffle && window.waffle.flag_is_active('web_font_loader')) {
+
+        // what fonts are we loading?
+        var fonts = {
```
darkwing: I think you want this to be an array and not an object.  So....

```
var fonts = [{  }, {  }]
```

Objects are good for key/value

darkwing:
Functionality-wise, this is working out pretty well so I'm fine with piloting it.  :)

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+
+        // kick off font loading
+        function ffoStart() {
+            ffoLoad();
+        }
+
+        // starts observers for all fonts in fonts array
+        function ffoLoad() {
+            // for each font
+            $.each(fonts, function(index) {
```
darkwing: You aren't using this `index` arg at this point, so I'd remove it.

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+
+                });
+            });
+        }
+
+        function ffoCheck(){
+            var allLoaded = true;
+
+            // loop through fontsLoaded looking to see if they've loaded
+            $.each(fontsLoaded, function(index){
```
darkwing: You aren't using this `index` arg at this point, so I'd remove it.

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+                        isLoaded = false;
+                    }
+                })
+                if(isLoaded == false) {
+                    allLoaded = false;
+                }
+            });
+
+            // if all fonts are loaded swap font in
+            if(allLoaded == true) {
```
darkwing: You can just do `if(allLoaded)`

media/js/main.js
```diff
@@ -217,27 +217,25 @@
...
-    */
     (function($tabzilla) {
-        if(!$tabzilla.length) return;
-
-        $('<link />').attr({
-            href: '//mozorg.cdn.mozilla.net/media/css/tabzilla-min.css',
-            type: 'text/css',
-            rel: 'stylesheet'
-        }).on('load', function() {
+        if(!(window.waffle) || window.waffle && !(window.waffle.flag_is_active('web_font_loader'))) {
```
darkwing: These can all be `win`

media/js/fonts.js
```diff
@@ -0,0 +1,158 @@
...
+            },
+            "2" : {
+                "name" : "Open Sans",
+                "class" : "ffo-opensans",
+                "varient" : [
+                    {"weight" : "normal"},
+                    {"weight" : "bold"},
+                    {"style" : "italic"}
+                ]
+            } /*,
```
darkwing: If we aren't using this, we should remove it

media/js/fonts.js
stephaniehobson: #faceplam I had a good reason for doing that at some point. No need to be nice. I think it's silly to leave it.

stephaniehobson:
Updated :)

media/js/fonts.js
stephaniehobson: JS is a real mind something or other sometimes.

darkwing:
R+

darkwing:
@stephaniehobson If I can merge this, can you push to staging and do IE testing?  Just to make sure we aren't bricking something?

PR #3194, opened 2015-04-21
Make edit button dark blue

openjck:
A recent A/B test showed that around 31% more people click the edit
button when it's a darker shade of blue.

Full results: http://optimize.ly/~KSOEmr?token=63f0bef5c746e00caa31#view=2
Analysis: https://groups.google.com/d/msg/mozilla.dev.mdn/98GP_Vh30Cc/GSL_L-f-pJYJ

### DIFF
--- a/media/stylus/components/wiki/page-buttons.styl
+++ b/media/stylus/components/wiki/page-buttons.styl
@@ -34,6 +34,15 @@
     }
 }
 
+/* Use a dark blue background color for all document edit buttons except those
+   that appear on zone landing pages. Superusers can customize zone landing page
+   style and we can't be sure that dark blue will look good on the background
+   colors they choose. */
+body:not(.zone-landing) #edit-button {
+    background-color: $blue;
+    color: $blue-background-text-color;
+}
+
 @media $media-query-small-desktop {
 
     .page-buttons, #page-buttons {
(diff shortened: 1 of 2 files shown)


media/stylus/components/wiki/page-buttons.styl
```diff
@@ -34,6 +34,13 @@
     }
 }
 
+/* Users can style zone landing pages. We can't be sure that dark blue will look
+   good on background colors they choose. */
+body:not(.zone-landing) #edit-button {
```
darkwing: `:not` sometimes worries me.  Thoughts @stephaniehobson ?

openjck: https://twitter.com/chriscoyier/status/414149535800127488 :smile:

If measurement shows that this noticeably affects performance, though, we should address it.

darkwing: Not from a perf standpoint, from a maintenance/cleanliness standpoint.

openjck: Ah. Gotcha.

I spent some time thinking about that. We could alternatively give non-landing pages a new class. Something like `.standard-article`. But then if we create a third category of page later on, selecting `.standard-article` won't be enough.

We ultimately want to select everything that *isn't* a zone landing page, so I think the `:not` selector is a natural fit.

stephaniehobson: I think it makes the code more readable. I'm okay with it.

stephaniehobson:
:+1:

PR #3196, opened 2015-04-21
Fix Bug Bug 1154283: Use summary box on tables

stephaniehobson:
- made styles applicable to both ul and table for all colours of summary box
- standardized styles for both ul and table to match
- moved summary box classes into separate .styl file
- converted declarations to pre-processor syntax (variables! nesting! FTW!)
- RTL bugs

Supersedes #3179

Testing: 
Old style summary box made on a list: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b
New style summary box made on a table: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr
Testing page with both side by side: https://developer.allizom.org/en-US/docs/User:stephaniehobson:summary-box

### DIFF
--- a/media/stylus/components/wiki/customcss.styl
+++ b/media/stylus/components/wiki/customcss.styl
@@ -321,122 +321,8 @@ div.index ol {
     padding-left: 0;
     list-style-type: none;
 }
-/* Stuff used for the syntax information about CSS, HTML, and so forth */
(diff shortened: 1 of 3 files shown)


darkwing:
Miles, miles better.  Yesssss. R+

PR #3197, opened 2015-04-22
Fix Bug 1148738 - Syntax boxes should wrap

stephaniehobson:
White-space declaration wasn't being applied because the selector was too specific. This moves it up to a more general selector.

Supersedes #3143.


darkwing:
So I'm seeing the wrapping but never seeing any `ellipsis` -- what should I be seeing with that?

stephaniehobson:
I would be triggered by a line that can't wrap, you could test it by copying a URL into a code block.

I'm not 100% sold on needing it though, we can remove it if you think it's extraneous.

darkwing:
Ahhh yes, that looks great!  Approved!  R+

PR #3198, opened 2015-04-22
Bug 1143084 remove unused fonts from code base

davidyezsetz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1143084


stephaniehobson:
Yay! Super excited to see this! This is the lion's share of the work and it looks solid.

There's a little fine tuning to do. @davidyezsetz are you interested in doing that work too? 

Stuff like:
- [ ] setting letter spacing on .demo-title to 0
- [ ] including moztt in one of the gaia style sheets
- [ ] changing buttons to match buttons elsewhere on MDN
- [ ] vertically centering and uppercasing .gallery-head .count

If not I can take over the branch and make them. It's probably just as fast to fix them as it is to report them.

Again, really excited to see this PR. Thanks!

davidyezsetz:
Sure thing. 1,2,4 are clear, but what do you mean with changing buttons?

stephaniehobson:
For example the edit / remove / hide buttons on a [demo detail page](https://developer.mozilla.org/en-US/demos/detail/paperfold-css) should match the edit button on [wiki pages](https://developer.mozilla.org/en-US/docs/Web/HTML). Other examples of the button style can be found in places like the save button when editing a profile.

It looks like the buttons are already inheriting and over-riding the default site button styles so hopefully this is just a matter of deleting more lines of code :)

davidyezsetz:
So we're talking about all instances of class="button" ? As long as I can grep something it shouldn't be a problem.

stephaniehobson:
It's more complicated than that. Maybe we should leave that for a future improvement.

davidyezsetz:
I was afraid that was case.
I think there needs to be a bit of groundwork i.e. digging through the pages to find all implementation of buttons that should look alike and listing them  [like this](http://bradfrost.com/wp-content/uploads/2013/07/pnc-buttons-2.jpg).  Then a designer can pick one and it can be implemented.
I will finish your other 3 points in the meantime, so this can be merged in.

davidyezsetz:
About moztt inclusion for Gaia demos: I don't see a single css file that is included in every demo. Should I create a moztt.css file and include it in all the demos?

stephaniehobson:
Editing the demos is complicated because they sit on the demos server. But that reminds me... I'm not sure they were picking up the font from the global style sheets (iframes don't inherit styles from their parents) if the font isn't being included directly in any of those files right now it's probably never been displaying. I'll talk to the content team and see if they have any thoughts (or might be willing to do the work of editing the demos).

stephaniehobson:
I'm going to file a separate bug for the moztt stuff too. We won't be breaking anything with this PR that isn't already broken.

davidyezsetz:
Great. Baring mistakes in my last two commits this should be complete then.

stephaniehobson:
:+1: This is ready to go.

openjck:
Superseded by #3202 which uses the same changes but prepends bug numbers to the commit messages.

stephaniehobson:
Thanks for doing that @openjck I really wanted to merge this one today :)

PR #3201, opened 2015-04-23
Improve Intern Tests

darkwing:
This will be a running PR, with several commits, which can be periodically merged.


openjck:
Awesome! Could this be merged now, if we wanted to? What does the *needs tests* label indicate?

darkwing:
Sure, if all tests were passing for you.  The label is there just so you don't feel the need to merge / evaluate right away.

PR #3203, opened 2015-04-27
Bug 1157937 - Track TOC clicks

stephaniehobson:


darkwing:
How about this instead?

```
/*
    Track Clicks on TOC links
*/
$('#toc').on('click', 'a', function() {
    var $thisLink = $(this);

    mdn.analytics.trackEvent( {
        category: 'TOC Links',
        action: $thisLink.text(),
        label: $thisLink.attr('href')
    });
});
```

darkwing:
Kicked the Travis job, looks like a transient error.

groovecoder:
:+1: for the `$('#toc').on('click', 'a')` event binding.

stephaniehobson:
I'm not arguing, I just wanting to understand but I don't know how to ask this question without sounding petulant so, sorry: Could someone please explain to my why this is better?

groovecoder:
I like it because it's less nested code and easier for me to read.

darkwing:
I prefer my method because:

-  I'm using event delegation on the TOC, assigning one event handler on the parent instead of assigning an event to every link within it.

-  You don't need to check for `.length` for methods directly on the jQuery return object -- nothing will happen if the selector matches nothing.

-  You don't need an `each` for jQuery object methods -- it's an implied loop for all matches.

groovecoder:
This is assigned back to me ... want me to pick it up?

stephaniehobson:
Updated, leaving it to @groovecoder to review.

Sorry for the slow update, I've had other things on my mind. And thinks for the explanations :)

PR #3212, opened 2015-05-11
Intern stability 1152386

darkwing:
Simply PR'ing this so WebQA can look.

### DIFF
--- a/docs/tests-ui.rst
+++ b/docs/tests-ui.rst
@@ -30,7 +30,7 @@ Running Tests
 
     node_modules/.bin/intern-runner config=intern-local
 
-The above runs the entire suite of tests. Custom functionality has been added to allow for command line arguments to be passed to modify configuration, namely `b` to set which browsers to run in, `t` for which test suites to run, `u` to provide a username for Persona, `p` to provide a password for Persona, `d` for which domain to run on, and `wd` which represents the slug of an existing article to test::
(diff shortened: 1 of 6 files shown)


darkwing:
This is ready for a look.  This makes login more reliable and also adds destructive tests to the wiki.

Run as follows to test:

./node_modules/intern/bin/intern-runner.js config=intern-local.js b=firefox u=yourwikiuser@domain.com p=YourPass d=developer-local.allizom.org destructive=true t=wiki

groovecoder:
Kicked Travis job.

openjck:
Travis looks good now. Testing...

openjck:
Looks good aside from one thing. The destructive tests assume the existence of `docs/Web`, but we've also talked about not running these tests on production. Can we instead create a page so that the tests don't have a working database import as a dependency?

darkwing:
>> I was unable to run the destructive tests because they assume the existence of docs/Web

It doesn't at all.  The destructive tests create a page at `test_2938743924` or whatever the timestamp is.  `docs/Web` doesn't come into play at all.

openjck:
Something else must have went wrong, then. Could have sworn I saw `docs/Web` show up in the log output. I'll look again.

groovecoder:
Been using these to test the django update on stage. I think http://kuma.readthedocs.org/en/latest/tests-ui.html needs an update with better display of all the test options.

PR #3214, opened 2015-05-13
bug 1156930 - add Tagalog, Burmese, Songhay, Bambara languages

groovecoder:
Spot-check on this branch:
1. [Get the localizations](https://kuma.readthedocs.org/en/latest/localization.html#get-localizations) if you don't already have them.
2. `./manage.py update_product_details`
3. Update your `locale` directory and compile new messages files
```
cd locale
svn up
./compile-mo.sh .
```

1. Go to [/tl/](https://developer-local.allizom.org/tl/), [/my/](https://developer-local.allizom.org/my/), [/son/](https://developer-local.allizom.org/son/), [/bm/](https://developer-local.allizom.org/bm/).
  * [ ] Make sure the home page works.
2. Translate a test document into Tagalog, Burmese, Songhay, and Bambara languages.


stephaniehobson:
Passing this back. Not able to determine if the problem is with my local environment or the PR. Languages appeared as expected in locale directory, languages.json, and settings.py.

Homepage: renders in all but tl (404)
Translate into: lists all new locales.
Actually translating a document: in any locale generates the error:
```
  File "/home/vagrant/src/settings.py", line 230, in lazy_langs
    for lang in langs])
KeyError: 'bm'
```

openjck:
The homepage loads for me in all new locales. When I visit an existing page like [HTML](https://developer-local.allizom.org/en-US/docs/Web/HTML), I get [this traceback](https://pastebin.mozilla.org/8833460).

I can't create new pages in Vagrant (even on master) due to an unexplained `IOError: Socket closed` error.

groovecoder:
Can either of you try to do a `./manage.py update_product_details` in the vm and see if that fixes it? That's the only step I can think of that I did that isn't in these spot-check instructions.

openjck:
Yep, that fixed the `KeyError`.

groovecoder:
Ah, so after that does it work to translate docs in the languages?

openjck:
> Ah, so after that does it work to translate docs in the languages?

Trying to test that now, but I'm getting the same `IOError`. I don't see it mentioned in the (troubleshooting document)[https://kuma.readthedocs.org/en/latest/troubleshooting.html]. I'll ask in IRC and do some digging.

stephaniehobson:
Fixed the key error!

So the only problem I'm experiencing now is a 404 for the homepage in tl.

openjck:
Looks good `son`.

groovecoder:
:cry: /tl/ works on my local, but doesn't work [on stage](https://developer.allizom.org/tl/) either. Will dig into this ...

groovecoder:
The good news is that translating docs to /tl/ works at least:

https://developer.allizom.org/tl/docs/User:groovecoder

groovecoder:
Wait, nvm. My homepage redirect was cached in Firefox. Other browsers show it fine.

PR #3215, opened 2015-05-13
bug 1164032 - Move to using trackEvent for JS, Ajax, and other errors.

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1164032

### DIFF
--- a/media/js/analytics.js
+++ b/media/js/analytics.js
@@ -91,12 +91,23 @@
             Track specific clientside errors create by our code
         */
         trackClientErrors: function() {
+
             $(win).on('error', function(e) {
                 var originalEvent = e.originalEvent;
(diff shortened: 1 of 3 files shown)


groovecoder:
Would it be simpler to change the `trackError` method to:
```
        trackError: function(description, label) {
            if(win.ga && ga.create) {
                var data = {
                    category: 'Error',
                    action: description,
                    label: label || null
                }
                analytics.trackEvent(data);
            }
        }
```
That way we can keep using `trackError` and change the calls to:
```
mdn.analytics.trackError('404');
mdn.analytics.trackError(originalEvent.message, originalEvent.filename + ':' + originalEvent.lineno);
mdn.analytics.trackError('AJAX', settings.url + ' : ' + e.result);
```
And we'll get an "Error" event category report to give us a comprehensive report of all errors in a single view.

darkwing:
Closing out, will submit a new one.

PR #3218, opened 2015-05-15
Bug 775569 and other small fixes

a2sheppy:
Implements Ctrl+0 (Cmd-0) as a key shortcut for the "Remove formatting" option in
CKEditor (bug 775569).

There are also some fixes which are parts of other general editing clean-up
bugs:

* Removed the "Right sidebar" and "Callout box" styles from
the Style drop-down menu.

* Added the key shortcut Ctrl+Shift+K (Cmd-Shift-K) for "Remove link".


darkwing:
R+

PR #3219, opened 2015-05-15

openjck:
Current issues:

* [ ] Django is not serving anything in `/static/assets/compressed`
    * Steps to reproduce
        * `vagrant up`
        * `vagrant ssh`
        * `foreman start`
        * `./manage.py collectstatic --noinput`
        * `cat static/assets/compressed/css/gaia.css`. The file contents will be printed.
        * Visit https://developer-local.allizom.org/static/assets/compressed/css/gaia.css. The file will not be found.
* [ ] Travis tests are failing. This might be due to `collectstatic` not being run, or it might be due to the issue above.
* [ ] [Some code](https://pastebin.mozilla.org/8833664) is still using jingo-minify's `{{ BUILD_ID_JS }}`. We need to find a different way of efficiently cache-busting assets that use this template tag.


openjck:
@jezdez: Any thoughts on the current issues listed above? I can't figure them out.

groovecoder:
I'm going to close this. We'll keep the branch up to preserve the code for re-use in a potential new PR for django-pipeline after we merge the python & django updates.

openjck:
Photograph of a broken physical oil pipeline

PR #3224, opened 2015-05-26
Bug 1150118: Dropping time out to 2 seconds

stephaniehobson:
Let's experiment with this for a week. It's not quite the "Can we just not swap out the fonts on first load" some people are asking for but it decreases how far someone can get before the swap happens, if it happens at all.


darkwing:
R+, works for me!

PR #3226, opened 2015-05-27
Fix bug 1078186: Redirect static canvas examples

openjck:


openjck:
A code review would be good, but we don't need to worry about spot-checking this. The code isn't going to have any effect locally, but Cylia [already confirmed](https://bugzilla.mozilla.org/show_bug.cgi?id=1078186#c34) that the rules work on our servers.

PR #3227, opened 2015-05-31
Fix Bug 1159634 - Set h5 font-size

davidyezsetz:


groovecoder:
@stephaniehobson or @openjck ?

media/stylus/base/elements/typography.styl
```diff
@@ -48,6 +48,11 @@ h4 {
     letter-spacing: -0.014em;
 }
 
+h5 {
+    set-font-size(16px);
+    letter-spacing: -0.014em;
```
stephaniehobson: 16px looks good :)

We're going to start lazy loading our web fonts. As part of this, there's now a way to hand-craft the display of the fallback fonts. You can see how it was declared for the h4 tag here: https://github.com/mozilla/kuma/blob/master/media/stylus/base/elements/typography.styl#L69-L75 Could you add fallbacks for the h5? 

I usually test this with dev tools by creating a block of multi-line text in each font and then absolutely positioning them over each other and fine tuning.

Alternatively, copying the h4 declaration would still get an R+ from me.

davidyezsetz: Sure thing.

davidyezsetz:
@stephaniehobson this is what you had in mind right? (Green is a fallback font)
fallback_font

stephaniehobson:
It looks good from here but review is going to be delayed until I have the new VM running, sorry!

PR #3228, opened 2015-06-01
bug 1170235 - Add initial assets for Compat Tables

darkwing:
Initial work for bringing compat tables into MDN.  My test template is as follows:

https://gist.github.com/darkwing/6fccdfd9852f338799eb

Seeing lots of display issues (i.e. `abbr` not working as intended, sprite not working as intended) so we'll need to figure out what's wrong.


stephaniehobson:
Superseded by #3235

PR #3230, opened 2015-06-01
Fix Bug 1148752 - CSS break long article titles.

stephaniehobson:
Added word-wrap: break-word to auto-break when necessary. Added hyphens: auto to add hyphen at break for supporting browser/locale combos.

Moved line height 1 declaration to apply to all headings by default and over ride for h1 (instead of applying to all and being over ridden by all but h1).

Decreased line height on document head. Increased margins on document head to account for decreased line height.

Before:
screen shot 2015-06-01 at 15 21 22

After:
screen shot 2015-06-01 at 15 26 10


darkwing:
Works beautifully and looks awesome!  R+!

PR #3232, opened 2015-06-03
bug 1116335 - utf8_distinct_ci collation for Travis CI MySQL

groovecoder:
Just sending a PR to trigger a Travis build ...

### DIFF
--- a/scripts/travis-before-script.py
+++ b/scripts/travis-before-script.py
@@ -1,0 +1,17 @@
+import codecs
+import os
+
+from jinja2 import Environment, FileSystemLoader
+
+
+env = Environment(
+    loader=FileSystemLoader(
+        os.path.join(
+            os.path.dirname(__file__), "../provisioning/roles/kuma/templates"
+        )
+    )
+)
+template = env.get_template('Index.xml.j2')
+
+with codecs.open('/usr/share/mysql/charsets/Index.xml', 'w', encoding='utf-8') as f:
+    f.write(template.render(collation_id=999))
--- a/settings.py
+++ b/settings.py
@@ -52,7 +52,7 @@ DATABASES = {
         'ATOMIC_REQUESTS': True,
         'TEST': {
             'CHARSET': 'utf8',
-            'COLLATION': 'utf8_general_ci',
+            'COLLATION': 'utf8_distinct_ci',
         },
     },
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,12 @@ env:
 before_install: scripts/travis-install
 install:
     - pip wheel -r requirements/compiled.txt coveralls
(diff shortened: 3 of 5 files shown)


settings.py
```diff
@@ -52,7 +52,7 @@
         'ATOMIC_REQUESTS': True,
         'TEST': {
             'CHARSET': 'utf8',
-            'COLLATION': 'utf8_general_ci',
+            'COLLATION': 'utf8_distinct_ci',
```
jezdez: Why is this needed? The test runner will run the migrations and therefore apply the collation to the tables needing it.

scripts/travis-before-script.py
```diff
@@ -0,0 +1,17 @@
+import codecs
```
jezdez: Ough man, that's janky. I'm r- on this.

tox.ini
```diff
@@ -19,7 +19,7 @@ setenv =
 deps=
     -rrequirements/compiled.txt
 commands =
-    coverage run manage.py test --noinput -v2 {posargs:kuma}
+    sudo coverage run manage.py test --noinput -v2 {posargs:kuma}
```
jezdez: Why sudo?

groovecoder:
For all of these comments: I'm just trying things out to see what the best way to make Travis work is.

settings.py
groovecoder: The test runner fails to apply the collation migration with a MySQL `OperationalError` so I'm trying some other approaches.

scripts/travis-before-script.py
groovecoder: `codecs` is janky?

settings.py
jezdez: It didn't for me, is this in your VM? What's the traceback?

scripts/travis-before-script.py
jezdez: The whole file is :(

settings.py
groovecoder: In the Travis CI environment. I'm still poking at it trying things today

scripts/travis-before-script.py
groovecoder: It's not my favorite. On the plus side, it makes use of the Jinja templates we have for Ansible. But, it adds another discrepancy between Travis CI and dev environments. I was searching for a way to running ansible on Travis CI, but all I could find were guides to test ansible **roles** with Travis.

Though I recall we might be able to make an inventory file just for Travis with a single "localhost" entry? I'll try that too.

groovecoder:
Closing. https://github.com/mozilla/kuma/commit/ed67af1a55b5a4ce937d1688ff875899277edc84 and https://github.com/mozilla/kuma/commit/5c89cc2b00dc13220c909a013682f9ebb32de712 are the better code to do this.

PR #3234, opened 2015-06-03
bug 1133709 - use '!' to blank password field

groovecoder:
This makes it easier for DBA to run the anonymize script.

### DIFF
--- a/scripts/anonymize.sql
+++ b/scripts/anonymize.sql
@@ -23,11 +23,12 @@
 -- (PII) data must be removed such that no data would be specifically tied to
 -- an individual.
 
-SET @common_hash_secret=rand();
+SET @common_hash_secret := rand();
 
 SET FOREIGN_KEY_CHECKS=0;
 
 TRUNCATE actioncounters_actioncounterunique;
+TRUNCATE account_emailconfirmation;
 TRUNCATE auth_message;
 TRUNCATE authkeys_key;
 TRUNCATE authkeys_keyaction;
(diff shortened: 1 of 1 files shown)


openjck:
+1 from Sheeri:

https://bugzilla.mozilla.org/show_bug.cgi?id=1133709#c28

groovecoder:
Assigning @jezdez ...

scripts/anonymize.sql
```diff
@@ -47,11 +48,14 @@ TRUNCATE tidings_watch;
...
 UPDATE auth_user SET
     -- username left alone, because it's public info
-    password = NULL,
+    password = '!',
     email = CONCAT('user-', id, '@example.com'),
     first_name = ROUND(RAND()*1000000),
     last_name = ROUND(RAND()*1000000);
 
+UPDATE account_emailaddress SET
+    email = CONCAT('user-', id, '@example.com');
```
jezdez: I think this shouldn't use the `id` of the account email address row but the user id the email address refers to, to continue have some relation to the data in the `auth_user` table. So this should work:

`UPDATE account_emailaddress SET email = CONCAT('user-', user_id, '@example.com');`

jezdez:
rwc+

scripts/anonymize.sql
groovecoder: I started with that, but then the records violate a unique constraint, since the same user_id can have multiple account_emailaddress records. E.g., for user ID 8, `user@gmail.com` and `user@yahoo.com` both become `user-8@example.com` which violates a unique key constraint.

So I wasn't sure how else to anonymize the email address. :neutral_face:

jezdez: The problem is that I'm not sure how allauth will react to the data integrity broken between the account_emailaddress and auth_user tables. If we'd use a simple hashing function on the full email address plus a random identifier that is the same for both update statements that should prevent rainbow table attacks since that secret is unknown. Something like:

```
UPDATE auth_user SET
    -- username left alone, because it's public info
    password = '!',
    email = CONCAT(MD5(CONCAT(email, @common_hash_secret)), '@example.com'),
    first_name = ROUND(RAND()*1000000),
    last_name = ROUND(RAND()*1000000);
UPDATE account_emailaddress SET
	email = CONCAT(MD5(CONCAT(email, @common_hash_secret)), '@example.com');
```
untested..

groovecoder: I'll test it ...

groovecoder:
Updated and spot-checked locally. r? again.

jezdez:
LGTM

PR #3238, opened 2015-06-05
Fix Bug 1126320 - Increase Inline Code Tag Visibility

stephaniehobson:
Changed `<code>` and `<kbd>` appearing outside of `<pre>` to use Courier New instead of Consolas. I tried to standardize on Consolas a while back but it's too hard to differentiate it from open-sans.

Also increased the weight of Courier New.

Test page: https://developer.mozilla.org/en-US/docs/User:stephaniehobson:code Only the inline tags in the first paragraph and heading should change (and they should get easier to spot).

### DIFF
--- a/media/stylus/base/elements/typography.styl
+++ b/media/stylus/base/elements/typography.styl
@@ -148,9 +148,9 @@ blockquote {
 pre/code/kbd
 ====================================================================== */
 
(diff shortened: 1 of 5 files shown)


darkwing:
R+, looks good.

PR #3239, opened 2015-06-05
Bug 1070001: Reorg stylus - .spec

stephaniehobson:
- separated `.spec` into its own file.
- `color` and `border-color` on `:before` are inherited from `color` declaration on parent, so removed.
- changed `background-color` to background for fewer characters
- combined declarations where possible
- checked all classes are in use
- RTL improvements
- removed `min-width` since not applied to inline elements (so it was never applied)
- moved border width and style declaration to short hand on group declaration
- snugged bottom padding up a bit to try to make top and bottom spacing above letters equal
- converted rgba values to hex

Testing page: https://developer.mozilla.org/en-US/docs/User:teoli/In_content#Specification_boxes 
Only difference between prod and local should be slightly shorter badges.


stephaniehobson:
443 characters in savings, if we're counting (I might be counting. It's the little victories, you know?).

media/stylus/components/wiki/spec.styl
```diff
@@ -0,0 +1,108 @@
...
+    font-weight: 700;
+    border-radius: 2px;
+    border-left: 4px solid;
+    letter-spacing: 1px;
+    padding: 0.45em 0.35em 0.30em;
+    /* bidi for this is complicated, the tag itself is LTR but the
+       name of the spec could be either LTR or RTL (depending on
+       if it's been translated) so we add margins to both sides.
+       Not perfect but good compromise.  */
+    bidi-value(margin, 0.2em 0.6em 0 0, 0.2em 0.6em 0);
```
darkwing: Why use `bidi-value` if you set `direction: ltr;` below?

darkwing:
Outside of my comment, this looks awesome!

groovecoder:
Selfish ask: Could we wait until we add YouShouldUse to kuma repo to merge this? That will test that this commit sends real Stylus payload to YouShouldUse and we can check for any errors.

media/stylus/components/wiki/spec.styl
stephaniehobson: The content inside the tag is reliably LTR. The content outside the tag is not reliably RTL or LTR and so the tag is unpredictably placed on either the left or the right of the content it sits before. On RTL pages it needs margins on both side but on LTR pages it is predictably placed on the left and does not need the left margin.

stephaniehobson:
Yes, we can wait for YouShouldUse.

openjck:
I'll be adding YSY today. :smile:

openjck:
> I'll be adding YSY today. :smile:

Added!

PR #3240, opened 2015-06-09
Bug 1148743 - Open jsfiddle and CodePen for Samples

darkwing:
Supercedes https://github.com/mozilla/kuma/pull/3147

@iakshay I don't see where we use `.sample-code-frame` or `.sample-code-table` -- is that something that needs to be added to the Embed template or...?


iakshay:
Yep, you can see the updated template in the gist -https://gist.github.com/iakshay/18dba5dac3d6ce99967d

darkwing:
OK, I've updated this so it works.   \o/

Something to note:  we're going to fight popup blockers, and of course, we can't prevent that.  @hoosteeno 

@stephaniehobson Wanna do a style or usability check here?

darkwing:
@hoosteeno :  Also of note re: popup blockers, if the user chooses "allow popups", the window opens *but* the post data (the CSS, HTML, and JS) don't get delivered, so our functionality will appear broken, at least in Firefox.

stephaniehobson:
Review is going to be delayed until I have the new VM running, sorry!

media/js/wiki-samples.js
```diff
@@ -0,0 +1,88 @@
...
+            var title = $sample.find('h2[name=' + section + ']').text();
+            openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode);
+
+            $button.removeAttr('disabled');
+        });
+    });
+
+    function openJSFiddle(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="https://jsfiddle.net/api/mdn/" class="hidden">'
+            + '<input type="hidden" name="html" />'
```
stephaniehobson: JSHint suggests the "+" for each of these multi-line declarations should go on the preceding line.

media/js/wiki-samples.js
```diff
@@ -0,0 +1,88 @@
...
+            // convert sitename to lowercase for icon name and host identifier
+            buttons += '<button class="open-in-host" data-host="'+ site.toLowerCase() +'" data-section="' +
+                section + '"><i aria-hidden="true" class="icon-'+ site.toLowerCase() +'"></i> Open in ' +
+                site +'</button>';
+        });
+        buttons += '</div>';
+        return buttons;
+    }
+
+})(window, document, jQuery)
```
stephaniehobson: Missing semi-colon.

stephaniehobson:
I'm not able to get it to run locally. What am I missing:
- downloaded PR and complied style sheets
- added and enabled the wiki_samples waffle flag
- updated the EmbedLiveSample macro and refreshed the page

The class `live-sample-frame` is appearing on the `iframe` and `wiki-samples.js` is being included on the page.

PR #3242, opened 2015-06-10
Scrollback integration

darkwing:
Supercedes https://github.com/mozilla/kuma/pull/2992

I completely rewrote the system

@stephaniehobson Will probably have an opinion about implementation, however.

PR #3243, opened 2015-06-10
Scrollback integration fixed

darkwing:
https://github.com/mozilla/kuma/pull/3242

I completely rewrote the system

@stephaniehobson Will probably have an opinion about implementation, however.


groovecoder:
While re-writing, we should re-word all commit messages to include the bug #.

darkwing:
Donezo!

stephaniehobson:
Review is going to be delayed until I have the new VM running, sorry!

aravindet:
@darkwing Thanks for refactoring this.

I see you've removed the overlay notice. This was added based on a couple of discussions on bugzilla:

1. A link to the community guidelines (IRC etiquette) page was discussed here: https://bugzilla.mozilla.org/show_bug.cgi?id=747805#c61 and it is in response to Justin's concern about people treating it as a helpdesk;

2. A link to Scrollback’s privacy policy was a requirement from the legal team: https://bugzilla.mozilla.org/show_bug.cgi?id=1120531#c8

I don't mind these being removed but you might want to confirm with the folks involved, especially about the privacy policy thing.

stephaniehobson:
**Some bugs:**
- [ ] I was not prompted to okay the guidelines or review the privacy policy as per https://bugzilla.mozilla.org/show_bug.cgi?id=1120531
- [ ] Clicking the tag a second time to close the overlay leaves the outline of it still in view
- [ ] Clicking the tag multiple times (to open and close the overlay) results in errors.
- [ ] The background of the top bar matches the homepage but it should match the community boxes instead (example: CSS landing page)
- [ ] The MDN room (reached from clicking the arrow next to "all messages") should have an icon and description added to it.
- [ ] Can we change the full screen view to have a top bar background to match the overlay?

**Some UI/UX concerns:**
- [Comment 54](https://bugzilla.mozilla.org/show_bug.cgi?id=747805#c54) and [Comment 65](https://bugzilla.mozilla.org/show_bug.cgi?id=747805#c65) still need to be addressed, there was lots of discussion but no final decisions.
-  This implementation welcomes "Help, my code is broken" questions. That seems like inviting a lot more work in for our staff and while the bugzilla discussion doesn't seem to be firmly against it I think this is a terrible idea. I like the widget but I think it should only be promoted to users who are actively editing MDN docs.
- We need to find a way to display the widget separate from the tags so we can explain what the widget is and so user knows what they're clicking into. We need to explain: 
 - who are these people? 
 - what is this place?
 - why would I want to be here?
 - this is a 3rd party tool (Scrollback looks like it was built on the Google visual standards guide, probably useful to identify it is not a Google thing).
- The all messages pages is confusing. Most of the "discussions" are one line. Can we disable this page?
- I expected it to stay open as I browsed from page to page (I've been spoiled by Facebook apparently). We should warn people before they leave the page and suggest they pop it open full screen if they want to keep it.
- It's not localized, we should probably warn our users that the IRC channel is mostly English as well.

I'm going to take these UX/UI questions back to the bug to see if we can sort them out.

stephaniehobson:
Actually [Comment 71](https://bugzilla.mozilla.org/show_bug.cgi?id=747805#c71) covers some of that. 

- [ ] Should only be available in the "Contribute to MDN" pages.

stephaniehobson:
Added a comp to the bug to solve some of the UI/UX concerns.

groovecoder:
Closing this stale PR. We should avoid this feature again in the future since this is the 2nd or 3rd time it's been volunteered and abandoned.

PR #3245, opened 2015-06-10
Bug 1169673 - Lower complexity of db queries in RevisionFeed.

jezdez:


robhudson:
Code looks good but there's a couple test failures. r+ with fixed tests.

PR #3231, opened 2015-06-01
Fix Bug 1156233 - zebra stripe tables

stephaniehobson:
Slightly darker bg in tables with 5 or more rows.

Examples:
screen shot 2015-06-01 at 15 56 36
screen shot 2015-06-01 at 15 57 45

PR #3247, opened 2015-06-10
Fix Bug 1156233 - zebra stripe tables

openjck:
Supersedes #3231 with a rebase so that we can avoid switching between Puppet and Ansible when testing this change.

This is ready for testing but needs approval from the writing team before being merged.

### DIFF
--- a/media/stylus/components/content.styl
+++ b/media/stylus/components/content.styl
@@ -24,7 +24,14 @@ $table-blue = #d4dde4;
                 box-shadow: inset 0 -1px 0 0 rgba-fallback(rgba($table-blue, 0.5));
             }
 
-            td.header, th {
+            /* zebra strip if 5 or more rows */
+            tr:first-child:nth-last-child(n+5) ~ tr:nth-child(2n-2) td {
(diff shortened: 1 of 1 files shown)


darkwing:
Why does this need writing team approval?  They aren't creative directors.

openjck:
From #3231:

> Needs an okay from the content team since it is indiscriminately applied to all tables.

stephaniehobson:
Because I'm not a creative directory either?

darkwing:
I think of you as one, but meh, OK

openjck:
Looks great to me.

I agree that there isn't a clear decision-maker here. I would feel comfortable merging it without consulting the writers. I think we have as much of a stake as they do.

stephaniehobson:
They are the ones who asked for it, I don't want to cause them problems or fail to fix the problem. I've already had feedback that it shouldn't be applied to singled column tables (that complicated selector is about to get more complicated!)

openjck:
Works for me! :smile:

Everything looks good to me, so we can merge if/when we get approval from the writers.

stephaniehobson:
Gonna adjust this and open a new one.

PR #3249, opened 2015-06-10
Bug 1173514 - Move gravatar URL function into cacheback job to reduce hashing time impact.

jezdez:


kuma/users/jobs.py
```diff
@@ -0,0 +1,29 @@
+import hashlib
+import urllib
+
+from django.conf import settings
+
+from kuma.core.jobs import KumaJob
+
+
+DEFAULT_AVATAR = getattr(settings, 'DEFAULT_AVATAR',
+                         settings.MEDIA_URL + 'img/avatar-default.png')
```
robhudson: Why does the `settings.DEFAULT_AVATAR` have a default here? lol

jezdez: I have no idea, it's.. legacy?

kuma/users/models.py
```diff
@@ -215,6 +216,12 @@ def create_user_profile(sender, instance, created, **kwargs):
         p, created = UserProfile.objects.get_or_create(user=instance)
 
 
+@receiver(models.signals.post_save, sender=settings.AUTH_USER_MODEL)
+def invalidate_gravatar_url(sender, instance, created, **kwargs):
+    if instance.email:
+        UserGravatarURLJob().invalidate(instance.email)
```
robhudson: Is it possible for the email to go from having a real value to `None` and we should invalidate that also?

kuma/users/jobs.py
robhudson: Do both paths exist? I would try to just have one default. That's too confusing.

robhudson:
r+wc (and passing tests)

kuma/users/models.py
jezdez: Ah, hm, good question, I think that should delete the entry in that case.

jezdez:
@darkwing provided a new default avatar image

robhudson:
:heart_eyes:

PR #3251, opened 2015-06-10
Fix bug 1173506 - Get rid of Document.full_path in favor of the real value Document.slug.

jezdez:


robhudson:
r+ but looks like legit test failures.

PR #3252, opened 2015-06-11
Bug 1116335 - Use Ubuntu's keyserver to get the MySQL PGP key as it's more reliable.

jezdez:


jezdez:
The MIT servers simply seems to be instable.

jezdez:
Closing in favor of #3253

PR #3259, opened 2015-06-12
Fix Bug 1126320 - Inline code tag visibility

stephaniehobson:
Dialing back the visibility a bit.

Changed font to Courier since Courier New is too fine at the regular weight. 
Removed default bold.

### DIFF
--- a/media/stylus/includes/vars.styl
+++ b/media/stylus/includes/vars.styl
@@ -54,9 +54,9 @@ $site-font-family = 'Open Sans', Arial, sans-serif;
 $site-font-family-fallback = Arial, sans-serif;
 $heading-font-family = 'Open Sans Light', 'Helvetica', Arial, sans-serif;
 $heading-font-family-fallback = 'Helvetica', Arial, sans-serif;
-$code-inline-font-family = 'Courier New', 'Andale Mono', monospace; /* inline */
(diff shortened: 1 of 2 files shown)


media/stylus/includes/vars.styl
```diff
@@ -54,9 +54,9 @@ $site-font-family = 'Open Sans', Arial, sans-serif;
 $site-font-family-fallback = Arial, sans-serif;
 $heading-font-family = 'Open Sans Light', 'Helvetica', Arial, sans-serif;
 $heading-font-family-fallback = 'Helvetica', Arial, sans-serif;
-$code-inline-font-family = 'Courier New', 'Andale Mono', monospace; /* inline */
+$code-inline-font-family = 'Courier', 'Andale Mono', monospace; /* inline */
 $code-block-font-family = Consolas, Monaco, 'Andale Mono', monospace; /* prism */
-$diff-font-family = 'Courier New', 'Andale Mono', monospace;
+$diff-font-family = 'Courier', 'Andale Mono', monospace;
```
darkwing: Can we do:
```
$code-inline-font-family = $diff-font-family = 'Courier', 'Andale Mono', monospace;
```
?

If not, can we put those two vars next to each other so we remember that they're essentially the same?

stephaniehobson:
Ready for re-review.

darkwing:
R+!

PR #3262, opened 2015-06-15
Fix bug 1119433 - Refactor settings.

jezdez:
Replaces #3062.


docs/email.rst
```diff
@@ -17,25 +17,29 @@ So now you know you aren't emailing real users, but you'd still like to email
...
+By default kuma is configured to print the email it sends in your console
+when you run it as plain text. That means the full raw source of every email
+will be visible where you run ``foreman start`` or ``./manage.py runserver``
+manually.
 
-::
+In case you want to override that default (not recommended) to have the emails
+really being sent out, you need to override a setting by modifying your
+``.env`` file in the root directory of the kuma source code. Set the
+``EMAIL_URL`` variable in ``.env`` like so::
```
groovecoder: I'm a little nervous to even document this ability. Should we leave it un-documented so that only `EMAIL_URL`-savvy (and therefore email-conscious) contributors know how to send real emails?

kuma/demos/tests/test_helpers.py
```diff
@@ -14,9 +12,9 @@ def test_tag_description_no_description(self):
             "title": "Testing tag without description",
         }
         description = tag_description("test_tag")
-        eq_("test_tag", description)
+        self.assertEqual("test_tag", description)
```
groovecoder: I like this here, but are we setting a precedent to remove nose `eq_` everywhere?

kuma/settings/common.py
```diff
@@ -867,14 +815,19 @@ def JINJA_CONFIG():
 FLAG_REASONS = DEMO_FLAG_REASONS + WIKI_FLAG_REASONS
 
 # bit.ly
-BITLY_API_KEY = "SET ME IN SETTINGS_LOCAL"
-BITLY_USERNAME = "SET ME IN SETTINGS_LOCAL"
+BITLY_USERNAME = config('BITLY_USERNAME', default='lmorchard')
+BITLY_API_KEY = config('BITLY_API_KEY',
+                       default='R_2653e6351e31d02988b3da31dac6e2c0')
```
groovecoder: Now that we have our own, should we update this too? (@hoosteeno - is there a way to get a dev/sandbox API key for our bitly account?) Or maybe we register for bitly from our mdn-dev@mozilla.com address and use that one?

kuma/settings/common.py
```diff
@@ -867,14 +815,19 @@ def JINJA_CONFIG():
...
 
-GOOGLE_MAPS_API_KEY = "ABQIAAAAijZqBZcz-rowoXZC1tt9iRT5rHVQFKUGOHoyfP_4KyrflbHKcRTt9kQJVST5oKMRj8vKTQS2b7oNjQ"
+GOOGLE_MAPS_API_KEY = config('GOOGLE_MAPS_API_KEY',
```
groovecoder: IIRC, this was only ever used on the events page, which is now removed? We could probably drop this.

kuma/settings/common.py
```diff
@@ -1092,19 +1073,19 @@ def JINJA_CONFIG():
 BASKET_URL = 'https://basket.mozilla.com'
 BASKET_APPS_NEWSLETTER = 'app-dev'
 
-KUMASCRIPT_URL_TEMPLATE = 'http://developer.mozilla.org:9080/docs/{path}'
+KUMASCRIPT_URL_TEMPLATE = 'http://localhost:9080/docs/{path}'
```
groovecoder: Should we put this in `config()` too? Since it changes between environments?

kuma/urls.py
```diff
@@ -75,9 +74,6 @@
         name='users.honeypot'),
 ]
 
-if settings.DEBUG:
-    urlpatterns += staticfiles_urlpatterns()
```
groovecoder: We can remove the corresponding `import` too, right?

kuma/demos/tests/test_helpers.py
jezdez: Yeah, nose's `eq_` is silly compared

jezdez: woops.

compared to `assertEqual`. My secret long term plan is to move us to pytest as it's becoming the defacto standard for testing while nose's big time is in the past.

kuma/settings/common.py
jezdez: Sure, this seems orthogonal to this ticket, but sure.

kuma/demos/tests/test_helpers.py
groovecoder: :+1:

groovecoder:
Code is looking really good. Still want to spot-check a vm build with all this.

Also, the wsgi file-name change will need WebOps to help us push, yeah?

kuma/settings/common.py
jezdez: It's the same in all environments, we can wrap it in a `config` call when needed (e.g. when we separate the web heads from the kumascript servers). This is me being careful about adding configuration variables.

docs/email.rst
jezdez: The ability exists so it should be documented. It's a fairly edge case settings so I don't expect this to be an issue.

kuma/settings/common.py
groovecoder: :+1:

kuma/settings/common.py
jezdez: Ah, good catch!

docs/email.rst
groovecoder: :+1:

kuma/urls.py
jezdez: :+1:

jezdez:
@groovecoder Thanks, the changes are far smaller after the upgrade branch.

I think we could in theory also create a symlink from the old to the new (Django compat) path, but tbh since webops needs to change the env var/config loading anyways it seems trivial to just change the path to the wsgi file.

kuma/demos/tests/test_helpers.py
robhudson: If pytest is the eventual goal, why not just `assert 'test_tag' == description` since pytest monkey patches assert to make it nice?

groovecoder: Wait, pytest monkey-patches? :disappointed:

jezdez: This isn't the place to have this discussion.

kuma/search/tasks.py
```diff
@@ -77,7 +77,6 @@ def finalize_index(index_pk):
     index.populated = True
     index.save()
 
-    subject = ('[%s] Index %s completely populated' %
-               (settings.PLATFORM_NAME, index.prefixed_name))
+    subject = ('Index %s completely populated' % index.prefixed_name)
```
robhudson: No longer needs parens.

robhudson:
Can we schedule some time with ops to roll this out? Or is there a bigger ops dependency we're waiting on? I :heart: these changes and would :heart: to see them merged.

groovecoder:
:+1: though this needs a rebase first too.

Our other pending WebOps tasks are:

* [Bug 1184670](https://bugzilla.mozilla.org/show_bug.cgi?id=1184670) - Automatically run Intern test suite against stage & prod servers after deployments
* [Bug 1005013](https://bugzilla.mozilla.org/show_bug.cgi?id=1005013) - Add New Relic to Kumascript

We should be able to finish the Intern test suite on Monday or Tuesday of next week. And this could probably go out in front of New Relic + KS since we're waiting on a New Relic support ticket.

jezdez:
@robhudson r?

Basically that will allow us to use the new settings structure on the old system by adding a sub `settings.py` that imports the prod settings, since the `settings.py` is imported by the `settings_local.py` on stage and prod, depending on the hostname. Since the migrations are run on the admin node, it'll load the prod settings even if we deploy stage, so I'm not sure if that's ideal. Sadly Chief doesn't provide an environment variable or something of which environment it runs at the moment. Any tips?

robhudson:
I like the looks of this. Since kuma/settings/local.py is checked into source, if you wanted to override something would you create a settings.py file and `from kuma.settings.local import *` at the top, followed by overrides? Is that how you envision the servers doing it? Or something else?

jezdez:
@robhudson No, there is no Python level settings overrides anymore, everything is done through env vars, in the `.env` file. If there is a setting that can't be overriden via an env var we need to customize the `common.py` once to be able to override it using python-decouple's casters.

jezdez:
I removed some code cleanup changes and then did the following:

- add a legacy `settings.py` that will be imported by the `settings_local.py` files that exists in prod and stage right now
- in the legacy `settings.py` file it will import either the `kuma.settings.stage` or `kuma.settings.prod` settings depending on one of the two environment markers:
  - an environment variable called `CHIEF_REMOTE_HOST` that is set in `chief_deploy.py` for the case in which `manage.py` is called from there on the admin node
  - inspect the result of the python code `platform.node()` and see if it contains `'stage'`
- once we're away from the current infrastructure we can safely set `DJANGOS_SETTINGS_MODULE` to either `kuma.settings.stage` or `kuma.settings.prod` and remove the legacy `settings.py` file again. the default is `kuma.settings.local`
- the legacy WSGI script `wsgi/kuma.wsgi` is set to look for `settings_local.py` is existing on the stage/prod environments; it can be removed when we're migrated away from the current infrastructure

@willkg @robhudson r? sorry for another round, I've tested these changes on stage and they work (\o/)

Makefile
```diff
@@ -7,7 +7,7 @@ endif
 
 # Note: these targets should be run from the kuma vm
 django-tests: in_vagrant
-	python manage.py test
+	DJANGO_SETTINGS_MODULE=kuma.settings.testing python manage.py test
```
willkg: Does anyone use this make rule? I ask because it's not mentioned in the docs anywhere and it's kind of awful--no one should ever run the tests this way without all the doohickey flags.

docs/celery.rst
```diff
@@ -66,13 +64,12 @@ Configuring and Running
...
 ``manage.py celeryd``.
 
-In ``settings_local.py`` you should set at least this, if you want to use
-Celery::
+In your ``.env`` file you should set at least this, if you want to use Celery::
 
-    CELERY_ALWAYS_EAGER = False
+    CELERY_ALWAYS_EAGER = false
 
 This defaults to ``True``, which causes all task processing to be done online.
```
willkg: Assuming the casing of the above is correct, then this should be ``true``.

docs/email.rst
```diff
@@ -17,25 +17,29 @@ So now you know you aren't emailing real users, but you'd still like to email
...
-    EMAIL_HOST_PASSWORD = '<your gmail password>'
+    EMAIL_URL = 'smtps://<your gmail address>:<your gmail password>@smtp.gmail.com:587/'
 
 Yeah, you need to put your Gmail password in a plain text file on your
 computer. It's not for everyone. Be **very** careful copying and pasting
-settings from ``settings_local.py`` if you go this route.
+settings from ``.env`` if you go this route.
+
+The ``.env`` file should **never** checked into Git and is therefore listed
+the ``.gitignore`` file.
```
willkg: How often does anyone set the email backend up like this chapter talks about? If it's like never, then we should nix this chapter or reduce it to the bare minimum of "see the django docs".

docs/email.rst
```diff
@@ -17,25 +17,29 @@ So now you know you aren't emailing real users, but you'd still like to email
...
-    EMAIL_HOST_USER = '<your gmail address>@gmail.com'
-    EMAIL_HOST_PASSWORD = '<your gmail password>'
+    EMAIL_URL = 'smtps://<your gmail address>:<your gmail password>@smtp.gmail.com:587/'
 
 Yeah, you need to put your Gmail password in a plain text file on your
 computer. It's not for everyone. Be **very** careful copying and pasting
-settings from ``settings_local.py`` if you go this route.
+settings from ``.env`` if you go this route.
+
+The ``.env`` file should **never** checked into Git and is therefore listed
```
robhudson: should never *be* checked into Git and is therefore listed *in* the ...

manage.py
```diff
@@ -2,27 +2,9 @@
...
-            import settings_local  # noqa
-        except ImportError:
-            # or use the plain settings.py module
-            settings_mod = 'settings'
-        else:
-            settings_mod = 'settings_local'
-
-    # override the env var with what we want
-    os.environ['DJANGO_SETTINGS_MODULE'] = settings_mod
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'kuma.settings.local')
```
willkg: This requires changes to the testing chapter in the docs. It also probably requires some thinking. Doing:

```
make django-tests
```

Doesn't allow you to pass along args (and doesn't have the flags) so it's much harder to use for development. Maybe we should instead create a `test.sh` bash script something like this?:

```
#!/bin/bash

export DJANGO_SETTINGS_MODULE=settings_test
./manage.py test -s --noinput --logging-clear-handlers $@
```

scripts/chief_deploy.py
```diff
@@ -16,6 +16,14 @@
 
 VENV_BIN = os.path.join(settings.VENV_DIR, 'bin')
 
+
+# Set local settings to legacy name until we can safely use environment
+# specifc one.
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings_local'
+
+# Use commander's remote hostname to differ between stage and prd in setting.py
```
willkg: Nit: "prd" is spelled wrong here. Not a big deal or anything.

willkg:
My biggest concern is that the current changes don't update the testing documentation or change how it works. Seems like adding a test.sh shell script would go a long way towards reducing clutter in the docs and making it easy to run tests. I'm sure there are other possibilities.

Other than that, this seems ok to me. I suspect we'll land it and then have to fiddle with the dev environment stuff a bit to streamline things.

Makefile
jezdez: I don't know, I personally don't, don't recall why it was added.

docs/celery.rst
jezdez: :+1:

docs/email.rst
jezdez: Yeah, I think for a while we didn't have a proper console email backend, so this seems like historical info. Gonna remove it and point to the Django docs.

docs/email.rst
jezdez: Well spotted!

jezdez: :eyes:

manage.py
jezdez: I'm very against adding random shell scripts, since we've in the past had trouble maintaining them (or writing good ones in the first place). We used fabric for a while, but that had similar problems. I'm a fan of well written Makfiles since they hide complexity well enough. Lemme update the django-tests Make task then to accept parameters.

scripts/chief_deploy.py
jezdez: Good catch.

jezdez:
@willkg Yeah, lemme update the docs a bit about how to run tests. I considered using tox for a while, but then we don't need its matrix feature or anything else it provides. Then we used to use Fabric and Invoke for a while, but it didn't work nicely either. We have a `scripts/` folder full of weirdly written shell scripts, which I'm not a fan of. I'd rather spend an hour today to fix up the Makefile instead.

docs/celery.rst
jezdez: Wait, that's correct, CELERY_ALWAYS_EAGER is enabled in local development. The text is a bit odd, I'll fix.

willkg:
+1 for Makefile becoming *the* way.

kuma/search/tasks.py
jezdez: :+1:

manage.py
jezdez: We've moved on to py.test, so this is moot.

jezdez:
Ok, rebased over the py.test fixes in master. I need another pass at this. @willkg r?

The general gist is that we'll default to using `make test` (and other make tasks) and then document how to use special options for py.test. When we see a recurring test command we can add a Make task later. I've done that already for the case of getting test coverage.

manage.py
willkg: Hooray!

willkg:
I eyeballed everything and it looks fine to me. I don't know what the rollout plan is, but I think it should include another round of testing this on -stage.

I'm going to run through building a new dev environment and double-checking things, but I don't think that should block merging this. If we hit issues, we should fix them as we hit them.

So, r+ from me.

willkg:
I'm hitting one issue with the dev environment and I don't understand what's going on--it's really puzzling.

Namely, I do:

```
./manage.py runserver
```

in the vagrant vm, then I go to https://developer-local.allizom.org/en-US/ in my browser and there's no CSS served. Same thing happens when I do:

```
gunicorn kuma.wsgi -b 0.0.0.0:8000
```

which launches the server via the WSGI path.

The weird part is that when I look at the source for the page, there's no CSS links in the head. There are JS links in the foot, so I'm not sure what's going on.

Does this work for you?

jezdez:
@willkg Summarizing what I said on IRC, I don't have the problem and it seems as if you were able to fix the missing styles by running the `compilecss` and `collectstatic` step. But I wasn't able to reproduce your slowness of page rendering which was dramatic. Until we have a fix for this we can't risk deploying this.

willkg:
I'm looking into my weird slowness problems with a fresh vagrant vm.

1. I've got a macbook pro, so I'm setting it all up there to see what it does on osx. Maybe it's a linux vs. osx platform issue?
2. Depending on how that goes, I'll check to see if it's a timeout trying to connect to ES or Redis or something.
3. Or maybe an empty db (this is a hand-wavey theory).
4. Then I'll look at what the django process is actually doing.

I'll keep working on this tonight and tomorrow.

willkg:
I get exactly the same slowness issue with a fresh Vagrant vm based on the master branch (with some tweaks due to pip releasing 8 last night) on my Linux box.

However, on the Macbook I've got with a fresh Vagrant vm, it takes 5 seconds to render both with the master branch and this settings refactor branch.

Given that, I claim it's a platform issue. Maybe due to how I'm mounting src/ or something like that. Maybe something finicky with django-pipeline or something else that causes problems with my Linux box but not my Macbook.

We should move forward with this.

docs/celery.rst
```diff
@@ -66,21 +64,20 @@ Configuring and Running
...
-overriden either in ``settings_local.py`` or via the command line when running
+overriden either in your ``.env`` file or via the command line when running
 ``manage.py celeryd``.
 
-In ``settings_local.py`` you should set at least this, if you want to use
-Celery::
+In your ``.env`` file you should set at least this, if you want to run Celery
+tasks asynchronously::
 
-    CELERY_ALWAYS_EAGER = False
```
butang88: On

docs/email.rst
```diff
@@ -1,41 +0,0 @@
...
-If you have ``sendmail`` installed and working, that should do it. However, you
-might get caught in spam filters. An easy workaround for spam filters or not
-having sendmail working is to send email via a Gmail account.
-
-::
-
-    EMAIL_USE_TLS = True
-    EMAIL_PORT = 587
-    EMAIL_HOST = 'smtp.gmail.com'
-    EMAIL_HOST_USER = '<your gmail address>@gmail.com'
```
butang88: alipboulala@gmail.com

docs/celery.rst
```diff
@@ -66,21 +64,20 @@ Configuring and Running
...
+    CELERY_ALWAYS_EAGER = false
```
butang88: True

robhudson:
I tried rebasing this but got lost when the changing requirements/* files and what is supposed to go where. Things changed recently on master and conflicting changes happened in this branch. I can continue tomorrow when people are around to ask questions.

robhudson:
Ok, rebase is complete and tests pass.
party!

willkg:
I did the following:

1. skimmed the changes again--they look fine as far as I can tell
2. created a new vagrant vm--works fine
2. ran the tests in the vm--passed

I think this works fine for dev environments. Landing it on -stage and then -prod will be interesting. Then we're in the two-settings-systems situation.

What's the plan for that? Is it stay in two-settings-systems situation until AWS? Are we planning to migrate with webops' help?

jezdez:
@willkg The only thing that is done in stage/prod differently to the new settings structure is the use of a `settings_local.py`. So unless we get some way to easily override the stage/prod specific settings via env vars by porting the `settings_local.py` files that exist there I'd suggest to keep the separation until AWS.

robhudson:
I tested this on stage and was met with success. This is ready to be merged. @jezdez would you like to do the honors?

jezdez:
@robhudson ZOMG yes!!

jezdez:
wqgd7ykebnswy

PR #3263, opened 2015-06-15
Fix bug 1174820 - Revert the get_previous changes made in bug #1169673.

jezdez:


jezdez:
Seems like the query didn't 100% work the way the rest of the code is expecting it is. I want to roll this part back to prevent the dashboard being half broken right now.

groovecoder:
:+1:

PR #3265, opened 2015-06-15
Fix bug 1173395 - Don't pass in non-ASCII pseudo filenames to difflib.

jezdez:
This basically works around the upstream issue in difflib, which is documented in the many 3rd party bug reports:
- https://bugzilla.redhat.com/show_bug.cgi?id=1221169
- https://github.com/wbsoft/frescobaldi/issues/674
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=784940
- https://bugzilla.mozilla.org/show_bug.cgi?id=1149667

### DIFF
--- a/kuma/wiki/helpers.py
+++ b/kuma/wiki/helpers.py
@@ -90,8 +90,18 @@ seqm is a difflib.SequenceMatcher instance whose a & b are strings"""
 
 
 def _massage_diff_content(content):
-    tidy_options = {'output-xhtml': 0, 'force-output': 1}
-    content = tidy_document(content, options=tidy_options)
+    tidy_options = {
+        'output-xhtml': 0,
+        'force-output': 1,
+    }
+    try:
+        content = tidy_document(content, options=tidy_options)
+    except UnicodeDecodeError:
+        # In case something happens in pytidylib we'll try again with
+        # a proper encoding
+        content = tidy_document(content.encode('utf-8'), options=tidy_options)
+        tidied, errors = content
+        content = tidied.decode('utf-8'), errors
     return content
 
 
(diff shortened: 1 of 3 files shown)


groovecoder:
A spot-check edit on https://developer-local.allizom.org/fr/Apps/Design/Contenu$edit still hit a `UnicodeDecodeError`, though I can't tell from the traceback where it originated. :neutral_face: https://pastebin.mozilla.org/8836961

kuma/wiki/helpers.py
```diff
@@ -131,21 +131,16 @@ def format_comment(rev):
...
     if from_revision is None or to_revision is None:
         return "Diff is unavailable."
-    fromfile = u'[%s] %s #%s' % (from_revision.document.locale,
-                                 from_revision.document.title,
-                                 from_revision.id)
-    tofile = u'[%s] %s #%s' % (to_revision.document.locale,
-                               to_revision.document.title,
-                               to_revision.id)
+    fromfile = '[%s] #%s' % (from_revision.document.locale, from_revision.id)
+    tofile = '[%s] #%s' % (to_revision.document.locale, to_revision.id)
```
groovecoder: So this effectively removes the title from the unified diff display?

groovecoder:
Ah, https://github.com/mozilla/kuma/blob/e7c06dcb376a3dac640b9b9eac73859ecea35b01/kuma/wiki/helpers.py#L92-L95 is probably where the traceback originates, since the traceback includes `tidylib` and that's the only place we use `tidylib`

kuma/wiki/helpers.py
jezdez: Yep

jezdez:
@groovecoder Ok, taking a look at tidy

groovecoder:
FWIW, fuller console output from the celery worker process: https://pastebin.mozilla.org/8837181

jezdez:
@groovecoder r?

groovecoder:
I'm afraid I got the error again :crying_cat_face: 

I set `CELERY_ALWAYS_EAGER = True` and debugged thru the code. From what I can tell, the `UnicodeDecodeError` is thrown in `pytidylib` in a way that the calling code in `helpers.py` can't catch it? There's a clue that during the exception, the debugger starts at `_ctypes/callbacks.c` and thru `sink.pyc` and `utf_8.pyc`? Does that suggest the error is thrown at a compiled/C level that we can't catch in calling code? Maybe we need to update `tidylib` c library in the new python 2.7 environment?

We could merge this as-is, and we'll have fixed the `UnicodeEncodeError` with `difflib`, but we'll lose visibility of these `UnicodeDecodeError`s thrown in `pytidylib`.

jezdez:
I don't know which version of tidylib we're using tbh, why are we using that in the first place? Doesn't lxml have something like that?

jezdez:
I prefer to merge this and open a new ticket for investigating tidy.

groovecoder:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1175937 for the tidy follow-up.

PR #3271, opened 2015-06-17
fix bug 1175681 - use unicode format string

groovecoder:
To fix errors: https://rpm.newrelic.com/accounts/263620/applications/3172075/traced_errors/3066ef-1d471519-153e-11e5-9387-f8bc124250a8


jezdez:
LGTM

PR #3272, opened 2015-06-18
Fix bug 930300 - Don't stumble over querystrings with empty parameter names.

jezdez:


groovecoder:
Looks good to me. For what it's worth, I couldn't reproduce the original traceback on master with the original url in the bug. I did reproduce it on master with the url included in the test case here.

PR #3276, opened 2015-06-23
Bug 1148743 - Open jsfiddle and CodePen for Samples

darkwing:
Supercedes https://github.com/mozilla/kuma/pull/3240

I needed to redo this due to the VM change -- things go weird.

How to test:

1.  Create a macro called `EmbedLiveSample`:  https://gist.github.com/darkwing/cbd69a8756081d482474  (this was changed)
2.  Create a macro called `LiveSampleURL`:  https://gist.github.com/darkwing/4b09b0b1d1e6e41a66d2  (this is unchanged)
3.  Create a document with the following content:  https://gist.github.com/darkwing/0dd8cdaba22497efb76c

You should see the "Open" buttons above the rendered sample

### DIFF
--- a/media/js/wiki-samples.js
+++ b/media/js/wiki-samples.js
@@ -1,0 +1,98 @@
+(function(win, doc, $) {
+    if(!win.waffle || !win.waffle.flag_is_active('wiki_samples')) return;
+
+    var sites = ['codepen'];
+    var frameLength = 'frame_'.length;
+
+    var sourceURL = $('link[rel=canonical]').attr('href') || win.location;
+    var plug = '<!-- Learn about this code on MDN: ' + sourceURL + ' -->\n\n';
+
+    // using id to get sample code section since macros discard other attributes
+    $('.sample-code-frame').after(function() {
+        var section = $(this).attr('id').substring(frameLength);
+        return createSampleButtons(section, sites);
+    });
+    $('.sample-code-table').after(function(){
+        var section = $(this).find('iframe').attr('id').substring(frameLength);
+        return createSampleButtons(section, sites);
+    });
+
+    $('#wikiArticle').on('click', 'button.open-in-host', function(){
+        var $button = $(this);
+        var section = $button.attr('data-section');
+        var sampleCodeHost = $button.attr('data-host');
+        var sampleUrl = win.location.href.split('#')[0] + '?section=' + section + '&raw=1';
+
+        // track the click and sample code host
+        mdn.analytics.trackEvent({
+            category: 'Samples',
+            action: 'open-' + sampleCodeHost,
+            label: section
+        });
+
+        // disable the button, till we open the fiddle
+        $button.attr('disabled', 'disabled');
+        $.get(sampleUrl).then(function(sample) {
+            var $sample = $('<div />').append(sample);
+            var htmlCode = $sample.find('.brush\\:.html, .brush\\:.html\\;').text();
+            var cssCode = $sample.find('.brush\\:.css, .brush\\:.css\\;').text();
+            var jsCode = $sample.find('.brush\\:.js, .brush\\:.js\\;').text();
+            var title = $sample.find('h2[name=' + section + ']').text();
+            openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode);
+
+            $button.removeAttr('disabled');
+        });
+    });
+
+    function openJSFiddle(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="https://jsfiddle.net/api/mdn/" class="hidden">' +
+            '<input type="hidden" name="html" />' +
+            '<input type="hidden" name="css" />' +
+            '<input type="hidden" name="js" />' +
+            '<input type="hidden" name="title" />' +
+            '<input type="hidden" name="wrap" value="b" />' +
+            '<input type="hidden" name="source" value="mdn" />' +
+            '<input type="hidden" name="medium" value="code-sample" />' +
+            '<input type="submit" />' +
+        '</form>').appendTo(doc.body);
+
+       $form.find('input[name=html]').val(htmlCode);
+       $form.find('input[name=css]').val(cssCode);
+       $form.find('input[name=js]').val(jsCode);
+       $form.find('input[name=title]').val(title);
+       $form.get(0).submit();
+    }
+
+    function openCodepen(title, htmlCode, cssCode, jsCode) {
+       var $form = $('<form method="post" target="_blank" action="https://codepen.io/pen/define" class="hidden">' +
+            '<input type="hidden" name="data">' +
+            '<input type="submit" />' +
+        '</form>').appendTo(doc.body);
+
+       var data = {'title': title, 'html': plug + htmlCode, 'css': cssCode, 'js': jsCode};
+       $form.find('input[name=data]').val(JSON.stringify(data));
+       $form.get(0).submit();
+    }
+
+    function openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
+       if(sampleCodeHost === 'jsfiddle') {
+            openJSFiddle(title, htmlCode, cssCode, jsCode);
+        } else if(sampleCodeHost === 'codepen') {
+            openCodepen(title, htmlCode, cssCode, jsCode);
+        }
+    }
+
+    function createSampleButtons(section, sites) {
+        var buttons = '<div>';
+        $.each(sites, function(){
+            // convert sitename to lowercase for icon name and host identifier
+            var host = this.toLowerCase();
+            buttons += '<button class="open-in-host" data-host="'+ host +'" data-section="' +
+                section + '"><i aria-hidden="true" class="icon-'+ host +'"></i> Open in ' +
+                this +'</button>';
+        });
+        buttons += '</div>';
+        return buttons;
+    }
+
+})(window, document, jQuery);
--- a/media/stylus/base/elements/forms.styl
+++ b/media/stylus/base/elements/forms.styl
@@ -114,3 +114,7 @@ label {
         text-decoration: none;
     }
 }
+
+.open-in-host + .open-in-host {
+    bidi-style(margin-left, ($grid-spacing / 2), margin-right, 0);
+}
--- a/settings.py
+++ b/settings.py
@@ -719,6 +719,7 @@ MINIFY_BUNDLES = {
         'wiki': (
             'js/search-navigator.js',
             'js/wiki.js',
+            'js/wiki-samples.js',
         ),
         'wiki-edit': (
             'js/wiki-edit.js',


stephaniehobson:
It works!

stephaniehobson:
It works with the demo content provided in step 3.

I could not get it working with this template: https://gist.github.com/stephaniehobson/345c04668a1e2107427c but it was also not firing any errors.

darkwing:
Looking now!

darkwing:
Fighting with KumaScript at the moment...

darkwing:
Going to try to fight with this again :/

darkwing:
OK, so what I did to get this working on other docs is:

1.  Re-saved each of the templates.
2.  Hard-refreshed my sample pages
3.  Basic refresh to see the buttons.

:+1:

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
+(function(win, doc, $) {
+    if(!win.waffle || !win.waffle.flag_is_active('wiki_samples')) return;
+
+    var sites = ['jsfiddle', 'codepen'];
+    var frameLength = 'frame_'.length;
+
+    // using id to get sample code section since macros discard other attributes
+    $('.sample-code-frame').before(function() {
```
stephaniehobson: Buttons should be inserted after the live sample instead of before.

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+        var section = $(this).attr('id').substring(frameLength);
+        return createSampleButtons(section, sites);
+    });
+    $('.sample-code-table').before(function(){
```
stephaniehobson: Buttons should be inserted after the live sample instead of before.

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+    function openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
+       if(sampleCodeHost === 'jsfiddle') {
+            openJSFiddle(title, htmlCode, cssCode, jsCode);
+        } else if(sampleCodeHost === 'codepen') {
+            openCodepen(title, htmlCode, cssCode, jsCode);
+        }
+    }
+
+    function createSampleButtons(section, sites) {
+        var buttons = '<div style="margin-bottom:10px;">';
```
stephaniehobson: No inline styles please. I would suggest moving this margin-bottom declaration on to the buttons themselves instead of on the wrapper div and then you can add the declaration to forms.styl with the other margin declarations for these buttons.

darkwing: Grrr, that was debug, sorry.  Missed it.

stephaniehobson:
It's working for me now and it looks great! This is a feature I'd really like to see go live :) A few nits for UX and code maintainability. 

And we're waiting on a decision about whether or not to open in a popup or just take over the window, right?

groovecoder:
Decision ... from @hoosteeno? Or should we split-test it? :wink:

darkwing:
Updated!

kleberveridiano:
Yes it is.

2015-06-30 14:56 GMT-03:00 David Walsh <notifications@github.com>:

openjck:
It looks like the *not ready* label no longer applies, so I'm going to remove it. Let me know if I'm mistaken. :-)

openjck:
We talked about this in the meeting this morning. Did we decide on some concrete next steps for this pull request in particular? Should we begin reviewing now or hold off?

hoosteeno:
We agreed that the code needs review, and I need to answer a question about whether to open in a popup or in the same tab. I think they can happen in either order.

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+
+    // using id to get sample code section since macros discard other attributes
+    $('.sample-code-frame').after(function() {
+        var section = $(this).attr('id').substring(frameLength);
+        return createSampleButtons(section, sites);
+    });
+    $('.sample-code-table').after(function(){
+        var section = $(this).find('iframe').attr('id').substring(frameLength);
+        return createSampleButtons(section, sites);
+    });
```
openjck: I may be missing something, but can this be condensed to the following? It uses lines 8-11 with a different selector.

```javascript
$('.sample-code-frame, .sample-code-table frame').after(function() {
    var section = $(this).attr('id').substring(frameLength);
    return createSampleButtons(section, sites);
});
```

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+        var section = $button.attr('data-section');
+        var sampleCodeHost = $button.attr('data-host');
+        var sampleUrl = win.location.href.split('#')[0] + '?section=' + section + '&raw=1';
+
+        // track the click and sample code host
+        mdn.analytics.trackEvent({
+            category: 'Samples',
+            action: 'open-' + sampleCodeHost,
+            label: section
+        });
```
openjck: I definitely don't have any disagreements with these choices, but also want to invite @groovecoder to share any thoughts. Custom event naming is an art... a confusing art. Like performance art or something.

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+
+        // disable the button, till we open the fiddle
+        $button.attr('disabled', 'disabled');
+        $.get(sampleUrl).then(function(sample) {
+            var $sample = $('<div />').append(sample);
+            var htmlCode = $sample.find('.brush\\:.html, .brush\\:.html\\;').text();
+            var cssCode = $sample.find('.brush\\:.css, .brush\\:.css\\;').text();
+            var jsCode = $sample.find('.brush\\:.js, .brush\\:.js\\;').text();
```
openjck: We could also consider the `^=` attribute selector if we haven't already. Both have tradeoffs&mdash;up to you really.

```javascript
$sample.find('[class^=brush\\:.js]').text();
```

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+            '<input type="hidden" name="data">' +
+            '<input type="submit" />' +
+        '</form>').appendTo(doc.body);
+       var data = {'title': title, 'html': htmlCode, 'css': cssCode, 'js': jsCode};
+       $form.find('input[name=data]').val(JSON.stringify(data));
+       $form.get(0).submit();
+    }
+
+    function openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
+       if(sampleCodeHost === 'jsfiddle') {
```
openjck: Super nitpick. Needs one extra space. :smile:

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+            openJSFiddle(title, htmlCode, cssCode, jsCode);
+        } else if(sampleCodeHost === 'codepen') {
+            openCodepen(title, htmlCode, cssCode, jsCode);
+        }
+    }
```
openjck: You had a neat suggestion on Discord that could also apply here. Since the functions have the same signature, we could do something like this:

```javascript
function openSample(sampleCodeHost, title, htmlCode, cssCode, jsCode) {
    var openFn;

    if(sampleCodeHost === 'jsfiddle') {
        openFn = openJSFiddle;
    } else if(sampleCodeHost === 'codepen') {
        openFn = openCodepen;
    } else {
        return;
    }

    openFn(title, htmlCode, cssCode, jsCode);
}
```

Just a stylistic thing. It really depends on how many providers we expect to have in the future. I'll leave the decision up to you.

hoosteeno:
I have tested this with a _blank form target and without. Comments:
* The _blank target experience with a popup blocker enabled is bad. We should not use a _blank target.
* Replacing the MDN page with the destination is OK for a waffled release; I'd like to establish some way to assess whether that experience works for users who see it. I think we need insight into analytics from the destination sites -- I think we can infer success of this feature from the answer to the question, "do MDN users on code sample sites bounce immediately, or do they tinker"? I will reach out to Chris Coyier and Piotr.
* JSFiddle does an additional POST after MDN's initial POST, which means MDN is actually 2 back-button clicks away when JSFiddle loads. I think this is also a bad experience, particularly as we are citing "back button" as a reason it is OK to overwrite the tab. I suggest we should only use CodePen until we learn whether JSFiddle's behavior can be changed. I will reach out to Piotr.

zalun:
The second POST request is necessary to display the result in an iframe. If there is a way to work around it, I can fix it straight on. There is a need to display the result from a different domain.

groovecoder:
Here we're proposing to show the sample in a pop-up or in the MDN window itself. Not sure if an iframe is possible here without a major re-write.

zalun:
showing the result iframe within MDN should be done in-house (we've discussed it on the MDN meetup)
pop-up for editing is better, but it requires user interaction...
as for the moment we allow to place the JSFiddle edit page within iframe. If I will make an exception if I decide to disallow it.

media/js/wiki-samples.js
```diff
@@ -0,0 +1,90 @@
...
+            '<input type="submit" />' +
+        '</form>').appendTo(doc.body);
+       $form.find('input[name=html]').val(htmlCode);
+       $form.find('input[name=css]').val(cssCode);
+       $form.find('input[name=js]').val(jsCode);
+       $form.find('input[name=title]').val(title);
+       $form.get(0).submit();
+    }
+
+    function openCodepen(title, htmlCode, cssCode, jsCode) {
```
hoosteeno: For codepen, can we...
* send utm parameters with the POST? We'd like to send the equivalent of `?source=mdn&medium=code-sample`
* append a comment to the HTML of any code sample like `<!-- Learn about this code on MDN: https://article.url.goes/here -->`

darkwing: 1.  Yes

2.  Yes

darkwing:
Update:  Implement POST params, remove JSFiddle for now

darkwing:
Updates made!

stephaniehobson:
R!

zalun:
JSFiddle is now live with one request MDN API

hoosteeno:
@darkwing, Piotr has made changes to JSFiddle that will enable a workflow identical to CodePen. Since this isn't merged yet, can we re-implement JSFiddle with the same utm parameters as CodePen? Then I'll do one more quick functional review, someone can r? the code, and we can release.

darkwing:
@hoosteeno So you want `source=mdn` and `medium=code-sample`, yes?  Anything else?

hoosteeno:
@darkwing before answering I quickly checked the GA docs. Turns out "campaign name" is required.

For both JSFiddle and CodePen, can we send...
?utm_source=mdn&utm_medium=code-sample&utm_campaign=external-samples

hoosteeno:
@darkwing also the HTML comment we append to samples should be on both sites (https://github.com/mozilla/kuma/pull/3276/files#r34311242)

hoosteeno:
Our experiment plan[0] suggets we need to add a custom dimension for this feature so we can know who saw it. 

@groovecoder, you have the privileges to do that -- can you add one just like "helpfulness"?
@darkwing, once he does, can you add a custom event like the one in helpfulness[1] that will trigger if the person is seeing codesamples?

[0] https://docs.google.com/document/d/1woTfL6RMo4iCN-5EZQzwHOk_GV8Nx7Zq6H0Lb6cO4Mw/edit#
[1] https://github.com/mozilla/kuma/pull/3282/files#diff-25a3d06887fce3344ebdad7742003900R16

groovecoder:
I created the new dimension in GA.
```
var dimensionValue = 'SOME_DIMENSION_VALUE';
ga('set', 'dimension8', dimensionValue);
```

PR #3280, opened 2015-06-24
bug 1177264 - use doc.get_contributors in document view

groovecoder:
This change seems to remove a number of duplicate queries to the `wiki_revision` table. According to New Relic, those queries add an average of 80-90ms to (20% of) our backend response time on document views.

### DIFF
--- a/kuma/wiki/views.py
+++ b/kuma/wiki/views.py
@@ -529,10 +529,7 @@ def document(request, document_slug, document_locale):
     # the contributors right into the Document's html field.)
     # NOTE: .only() avoids a memcache object-too-large error for large wiki
     # pages when an attempt is made to cache all revisions
-    contributors = set([r.creator for r in doc.revisions
-                                              .filter(is_approved=True)
-                                              .only('creator')
-                                              .select_related('creator')])
+    contributors = doc.get_contributors()
 
     # TODO: Port this kitsune feature over, eventually:
     #     https://github.com/jsocol/kitsune/commit/


groovecoder:
32% better performance on the doc views now ...
get_contributors_screenshot

openjck:
Wow! What is `get_contributors()` doing that's so different?

kuma/wiki/views.py
```diff
@@ -529,10 +529,7 @@ def document(request, document_slug, document_locale):
     # the contributors right into the Document's html field.)
     # NOTE: .only() avoids a memcache object-too-large error for large wiki
     # pages when an attempt is made to cache all revisions
```
jezdez: the code comment would need to be removed as well, gents

jezdez:
also what's up with merging this with Travis failing? :(

kuma/wiki/views.py
groovecoder: Doh, good call.

groovecoder:
The [build passed on my fork](https://travis-ci.org/groovecoder/kuma/builds/68242809), and I didn't want to get back into the Travis queue. :cry: But that's a good point - if a fork build passes, should we wait to merge until the PR build passes too? In theory they are the exact same commit and build, right?

jezdez:
> On 25 Jun 2015, at 17:41, luke crouch <notifications@github.com> wrote:
> 
> The build passed on my fork, and I didn't want to get back into the Travis queue.  But that's a good point - if a fork build passes, should we wait to merge until the PR build passes too? In theory they are the exact same commit and build, right?
> 
We should have the PR specific build pass to have a consistent merge history and no chance of retroactive criticism by people looking at our contributing strategy. Also I simply had no way of knowing that your fork build passed.

groovecoder:
Sounds good. Should we add it to CONTRIBUTING.md ?

openjck:
My bad on Travis. I must not have noticed the warning message.

jezdez:
@groovecoder Yeah, if that's really something that needs to be said? If the PR doesn't look like it's passing, don't merge it.

PR #3281, opened 2015-06-24
fix bug 1156233 - zebra stripe tables

stephaniehobson:
Final word on striped tables. Stripes all tables with class .standard-table. Stripes can be removed by adding 'nostripe' class to table.


darkwing:
R+.  Nice!

PR #3282, opened 2015-06-25
[bug 1032455] Add helpfulness interface using GA

hoosteeno:
This is an experiment to determine whether people use the feature and whether we use the data from it.

### DIFF
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -416,4 +416,5 @@
       {# This can be removed once the social network button multivariate test is completed. #}
       mdn.optimizely.push(['activate', 2796520257]);
   </script>
+  {{ js('document') }}
 {% endblock %}
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -416,5 +416,7 @@
       {# This can be removed once the social network button multivariate test is completed. #}
       mdn.optimizely.push(['activate', 2796520257]);
   </script>
-  {{ js('document') }}
+  {% if waffle.flag('share_links') or waffle.flag('helpfulness') %}
+    {{ js('document') }}
+  {% endif %}
 {% endblock %}
--- a/media/js/helpfulness.js
+++ b/media/js/helpfulness.js
@@ -1,0 +1,32 @@
+(function(win, doc, $) {
+    if (document.cookie.replace(/(?:(?:^|.*;\s*)helpful-stfu\s*\=\s*([^;]*).*$)|^.*$/, "$1") !== 'true') {
+        if (window.waffle && window.waffle.flag_is_active('helpfulness')) {
+            // if they stay a minute, ask if it was helpful
+            setTimeout(function() {
+                var question = mdn.Notifier.growl('Did this page help you?<br> <a href="#" class="notification-button" id="helpful-yes">Yes</a><a href="#" class="notification-button" id="helpful-no">No</a><a id="helpful-stfu" href="#">Never ask me again</a>', { closable: true, duration: 0}).question();
+
+                $('#helpful-yes').click(function() {
+                   confirm("Thanks for your feedback!", 'success', 'Yes');
+                });
+                $('#helpful-no').click(function() {
+                   confirm("Thanks! We\'ll improve it.", 'error', 'No');
+                });
+                $('#helpful-stfu').click(function() {
+                   confirm("We won't bug you again.", 'info', 'STFU');
+                   document.cookie = "helpful-stfu=true; expires=Fri, 31 Dec 9999 23:59:59 GMT;";
+                });
+
+                function confirm(msg, type, label) {
+                    question.close(question.item);
+                    mdn.Notifier.growl(msg, { closable: true, duration: 2500 })[type]();
+                    mdn.analytics.trackEvent({
+                        category: 'Helpful',
+                        action: 'Clicked',
+                        label: label
+                    });
+                }
+
+            }, 60000);
+        }
+    }
+})(window, document, jQuery);
--- a/media/js/helpfulness.js
+++ b/media/js/helpfulness.js
@@ -1,30 +1,59 @@
 (function(win, doc, $) {
     if (document.cookie.replace(/(?:(?:^|.*;\s*)helpful-stfu\s*\=\s*([^;]*).*$)|^.*$/, "$1") !== 'true') {
         if (window.waffle && window.waffle.flag_is_active('helpfulness')) {
-            // if they stay a minute, ask if it was helpful
+             
             setTimeout(function() {
-                var question = mdn.Notifier.growl('Did this page help you?<br> <a href="#" class="notification-button" id="helpful-yes">Yes</a><a href="#" class="notification-button" id="helpful-no">No</a><a id="helpful-stfu" href="#">Never ask me again</a>', { closable: true, duration: 0}).question();
+                var notification = mdn.Notifier.growl(gettext('Did this page help you?<br> <a href="#" class="notification-button" id="helpful-yes">Yes</a><a href="#" class="notification-button" id="helpful-no">No</a><a id="helpful-stfu" href="#">Never ask me again</a>'), { closable: true, duration: 0}).question();
 
                 $('#helpful-yes').click(function() {
                    confirm("Thanks for your feedback!", 'success', 'Yes');
                 });
                 $('#helpful-no').click(function() {
-                   confirm("Thanks! We\'ll improve it.", 'error', 'No');
+                   confirm("", 'error', 'No');
                 });
                 $('#helpful-stfu').click(function() {
                    confirm("We won't bug you again.", 'info', 'STFU');
                    document.cookie = "helpful-stfu=true; expires=Fri, 31 Dec 9999 23:59:59 GMT;";
                 });
 
+                // create a dropdown in case the page is unhelpful 
+                var unhelpful_options = [
+                    {val: 'Translate', text: gettext('Translate it into my language')},
+                    {val: 'Make-Simpler', text: gettext('Make it less confusing')},
+                    {val: 'Needs-More-Info', text: gettext('It needs more information')},
+                    {val: 'Incorrect', text: gettext('It contains incorrect information')},
+                    {val: 'Other', text: gettext('Something else')}
+                ]
+                var $select = $('<select />').attr({
+                    id: 'helpful-detail',
+                    value: 'Choose one...'
+                }).append($('<option>').attr('selected', 'selected').text('Choose one...'));
+                $(unhelpful_options).each(function() {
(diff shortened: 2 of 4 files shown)


hoosteeno:
I think this is not ready -- needs l10n at least.

media/js/helpfulness.js
```diff
@@ -0,0 +1,62 @@
+(function(win, doc, $) {
+    if (document.cookie.replace(/(?:(?:^|.*;\s*)helpful-stfu\s*\=\s*([^;]*).*$)|^.*$/, "$1") !== 'true') {
```
groovecoder: tumblr_ne2s3xznzu1rlkbimo1_400

settings.py
```diff
@@ -715,6 +714,10 @@ def JINJA_CONFIG():
             'js/profile.js',
             'js/moz-jquery-plugins.js',
         ),
+        'document': (
```
darkwing: Can we change this to `helpfulness` instead?  `document` seems too general.

kuma/wiki/templates/wiki/document.html
```diff
@@ -416,4 +416,5 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
       {# This can be removed once the social network button multivariate test is completed. #}
       mdn.optimizely.push(['activate', 2796520257]);
   </script>
+  {{ js('document') }}
```
darkwing: This should have a waffle wrapper so we aren't including needless JS files.

media/js/helpfulness.js
darkwing: Maybe we use `localStorage` instead?

media/js/helpfulness.js
```diff
@@ -0,0 +1,71 @@
...
+                    else {
+                        notification[type](gettext(msg), 2000);
+                        mdn.analytics.trackEvent({
+                            category: 'Helpful',
+                            action: 'Clicked',
+                            label: label
+                        });
+                    }
+                }
+            }, 60000);
```
darkwing: Why are we settting this to a timeout?

hoosteeno: Luke and I agreed that we shouldn't ask people if a page is helpful until they've had a chance to look at it -- I chose a minute because it's a long time to read, but still inside our avg. bounce rate.

settings.py
hoosteeno: Stephanie and I noticed we'd both written JS features (her: social, me: helpfulness) that should (eventually) appear on all document pages, but not on editing pages. We agreed this might be a general case for scripts in the future.

kuma/wiki/templates/wiki/document.html
groovecoder: i.e.,
```
{% if waffle.flag('share_links') or waffle.flag('helpfulness') %}
{{ js('document') }}
{% endif %}
```
:question:

groovecoder:
Looks good to me, with 1 nit-pick: an article always asks me if it was helpful even if I've already voted 'Yes' or 'No' on it? Since users often navigate back and forth on pages on MDN, should we suspend the prompt if the user has already answered for the article in the last NN days? @stephaniehobson ?

If we wrap the `js('document')` call in the waffle flag I'll feel good merging and pushing this. It will start collecting some great data for us to decide if it's worth investing a full engineering project into this idea.

hoosteeno:
That's a good point -- an oversight, really. I'd rather fix it (making our interaction with localstorage a bit more complex) than bug people about the same article. Maybe a flag that expires after 6(?) months.

Will also fix the waffle setup.

kleberveridiano:
ok!

2015-06-30 12:22 GMT-03:00 Justin Crawford <notifications@github.com>:

hoosteeno:
O dang. Merge commit in there. Will extricate.

groovecoder:
Also need to re-word the commit message to be prefixed with "bug NNNNNNN"

media/js/helpfulness.js
```diff
@@ -0,0 +1,80 @@
...
+                    gettext('Yes') +
+                    '</a><a href="#" class="notification-button" id="helpful-no">' +
+                    gettext('No') +
+                    '</a><a id="helpful-stfu" href="#">' +
+                    gettext('Never ask me again') +
+                    '</a>';
+                var notification = mdn.Notifier.growl(ask, {closable: true, duration: 0}).question();
+
+                // answers to the simple question include...
+                $('#helpful-yes').click(function() {
```
darkwing: The new syntax is `.on('click', function() {'

media/js/helpfulness.js
```diff
@@ -0,0 +1,80 @@
...
+                    confirm("Thanks for your feedback!", 'success', 'Yes');
+                });
+                $('#helpful-no').click(function() {
+                    confirm("", 'error', 'No');
+                });
+                $('#helpful-stfu').click(function() {
+                    localStorage.setItem('helpful-stfu', 'true');
+                    confirm("We won't bug you again.", 'info', 'STFU');
```
darkwing: Single quotes, por favor

media/js/helpfulness.js
```diff
@@ -0,0 +1,80 @@
...
+                            action: 'Clicked',
+                            label: label
+                        });
+                        notification.error(gettext('Uh oh. What would make it better?') + '<br>' + $select[0].outerHTML, 0);
+                        $('#helpful-detail').change(function() {
+                            confirm(gettext("Thanks! We'll fix it."), 'info', $(this).val());
+                        });
+                    }
+                    else {
+                        ask_again_later();
```
darkwing: Use camel-case with JavaScript :)

media/js/helpfulness.js
```diff
@@ -0,0 +1,80 @@
...
+                function confirm(msg, type, label) {
+                    // if "No" ask for more info
+                    if (type === 'error') {
+                        mdn.analytics.trackEvent({
+                            category: 'Helpful',
+                            action: 'Clicked',
+                            label: label
+                        });
+                        notification.error(gettext('Uh oh. What would make it better?') + '<br>' + $select[0].outerHTML, 0);
+                        $('#helpful-detail').change(function() {
```
darkwing: `on('change', function() {`

media/js/helpfulness.js
```diff
@@ -0,0 +1,80 @@
+(function(win, doc, $) {
+    // no localStorage, no helpfulness rating
+    if (('localStorage' in win)) {
+        var stfu = localStorage.getItem('helpful-stfu') === 'true'; // true if ever clicked stfu
+        var asked_recently = parseInt(localStorage.getItem(document.location + '#answered-helpful')) > Date.now();
```
darkwing: You can use `doc` instead of document and for JS validation purposes, you should add `10` as the second argument to `parseInt`

groovecoder:
@hoosteeno - want to assign someone specific to review?

hoosteeno:
@darkwing has been giving great feedback so far.

groovecoder:
@darkwing - got a chance to re-review this soon? I'd love to launch this tomorrow if possible so we can start collecting the data.

media/js/helpfulness.js
```diff
@@ -0,0 +1,84 @@
+(function(win, doc, $) {
+    // no localStorage, no helpfulness rating
+    if (('localStorage' in win)) {
+        var stfu = localStorage.getItem('helpful-stfu') === 'true'; // true if ever clicked stfu
```
darkwing: For the sake of positivity, we should probably change the `stfu` references to `ignore` or something like that.  `stfu` may reflect bad on us.

media/js/helpfulness.js
```diff
@@ -0,0 +1,84 @@
...
+        var asked_recently = parseInt(localStorage.getItem(doc.location + '#answered-helpful')) > Date.now();
+        if (!stfu && !asked_recently) {
+
+            setTimeout(function() {
```
darkwing: Big anonymous functions make the code hard to follow -- I recommend making this a named function and using that as the first argument to `setTimeout`.

```
setTimeout(myFn, 60000);

function myFn() {
//... body
}
```

darkwing:
Outside of these two nits, I really like what you've done here!

media/js/helpfulness.js
```diff
@@ -0,0 +1,91 @@
...
+        if (!ignore && !asked_recently) {
+            // ask about helpfulness after 1 min
+            setTimeout(inquire, 60000);
+        }
+    }
+
+    // create a notification
+    function inquire() {
+        // dimension7 is "helpfulness"
+        ga('set', 'dimension7', 'Yes');
```
darkwing: My last feedback is that this line should be moved to:

`templates/includes/google_analytics.html`

hoosteeno: I think I understand why this is a good idea in principle. But we're setting this dimension flag for users flagged by waffle, who have not said "Never ask me again" and who have not already responded to the inquiry on the page they're looking at. Replicating all that non-GA business logic in the GA file seems like a violation of DRY. Do you see another way to do it, or do you think maybe this should be an exception to the rule?

groovecoder: My :dollar: 0.02:

The point of the custom dimension is to analyze between 2 segments/cohorts in GA: 1. Those who saw the helpfulness feature and 2. Those who did not.

Given that, only setting the dimension when we create the (initial) notification makes sense to me. Otherwise, we'll be putting visitors into the "Saw helpfulness dialog" cohort even if they never saw the notification.

darkwing: OK.  In that case, I suggest:

```
if(ga) ga('set', 'dimension7', 'Yes');
```

..since we don't use GA locally.  Threw an error for me.

hoosteeno:
I added code to resolve @darkwing's most recent comment (https://github.com/hoosteeno/kuma/commit/964782802104c3beff9fcc33dbac82485e5d6183#diff-25a3d06887fce3344ebdad7742003900R16). Unless there are any further comments, this code is ready to merge. We can switch the flag after further coordination.

groovecoder:
:shipit:

groovecoder:
Any chance we can merge this in the next 30-60m? I'd like to push it (along with https://github.com/mozilla/kuma/pull/3346) today if possible ...

darkwing:
Looks good, let's give it a try

darkwing:
R+

groovecoder:
Just a post-deploy note ... there was confusing behavior in prod where the `share_links` flag is enabled for super-users: I saw the helpfulness feature immediately without having to enable its waffle flag. Do we need to clean that up with a follow-up PR? I.e., to de-couple the logic in `helpfulness.js` from the `or` logic around `js('document')`.

stephaniehobson:
Yeah, something needs to be done about that. I think a check for the waffle needs to be added to helpfulness.js

hoosteeno:
@stephaniehobson @groovecoder r? https://github.com/mozilla/kuma/pull/3349

PR #3283, opened 2015-06-25
bug 1150118 - A bit of JS cleanup

darkwing:


media/js/fonts.js
```diff
@@ -111,12 +111,12 @@
             var fontAttribute = index;
             var fontLoaded = true;
             $.each(this, function() {
-                if(this == false) {
+                if(!this) {
```
darkwing: I wonder if we just do `fontLoaded = this;`

stephaniehobson: Right now it loops through each font and checks if it is loaded, changing `fontLoaded` to false if it hasn't. If we just do `fontLoaded = this` then `fontLoaded` will have the value of the last font in the loop, won't it? Or am I misunderstanding where we'd insert it?

media/js/fonts.js
```diff
@@ -111,12 +111,12 @@
...
                     fontLoaded = false;
                 }
             })
-            if(fontLoaded == true) {
-                $('html').attr("data-"+fontAttribute, true);
+            if(fontLoaded) {
+                $('html').attr('data-' + fontAttribute, true);
```
stephaniehobson: At least I'm internally consistent with my incorrect use of quotes. >_<

media/js/fonts.js
stephaniehobson: Can we make the same replacement on lines 83 and 87?

stephaniehobson:
Tested, and it works, so we could merge and have better code :) Or we could make additional changes, mostly those suggested by jshint.

darkwing:
Updated!

stephaniehobson:
R+ once Travis gives the okay.

darkwing:
Kicked!

media/js/fonts.js
kleberveridiano: to me it´s fine!

kleberveridiano:
Great!

PR #3284, opened 2015-06-25
bug 1177264 - Optimize (some more) of document.html rendering

groovecoder:
Some tweaks to document.html logic based on [New Relic profiling data](https://rpm.newrelic.com/accounts/263620/applications/3172075/profiles/1706748).

### DIFF
--- a/kuma/wiki/helpers.py
+++ b/kuma/wiki/helpers.py
@@ -18,6 +18,7 @@ from django.utils.html import conditional_escape
 from constance import config
 from jingo import register
 from teamwork.shortcuts import build_policy_admin_links
+import waffle
 
 from kuma.core.urlresolvers import reverse
 from .constants import DIFF_WRAP_COLUMN
@@ -324,8 +325,12 @@ def wiki_url(context, path):
     Create a URL pointing to Kuma.
     Look for a wiki page in the current locale, or default to given path
     """
+    request = context['request']
+    if waffle.flag_is_active(request, 'dumb_wiki_urls'):
+        return reverse('wiki.document', args=[path])
+
     default_locale = settings.WIKI_DEFAULT_LANGUAGE
-    locale = getattr(context['request'], 'locale', default_locale)
+    locale = getattr(request, 'locale', default_locale)
 
     # let's first check if the cache is already filled
     url = memcache.get(u'wiki_url:%s:%s' % (locale, path))
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -26,7 +26,7 @@
 {% set zone_stack = document.zone_stack %}
 {% set is_zone = zone_stack|length %}
 {% set is_zone_root = is_zone and zone_stack[0].document == document %}
-{% set is_translated = document.other_translations.exists() %}
+{% set other_translations = document.other_translations %}
 
 {% set show_left = zone_subnav_html or quick_links_html or approvals_html %}
 {% set show_right = (not is_zone_root and toc_html) %}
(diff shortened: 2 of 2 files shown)


groovecoder:
https://gist.github.com/stephaniehobson/9966e7f051c10e784bb9

stephaniehobson:
In ja the current behaviour on Prod is:
- dropdown menu link is to /en-US/Firefox

In ja on this branch the behaviour is:
- dropdown menu link is ja/docs/Firefox
- page doesn't exist so user is forwarded to en-US version

Server behaviour is different, use experience is the same :(

darkwing:
This is really, really great Luke!  Well done!

kuma/wiki/templates/wiki/document.html
```diff
@@ -58,8 +59,8 @@
   <link rel="alternate" type="application/json" href="{{ url('wiki.json_slug', document.slug) }}">
   <link rel="canonical" href="{{ canonical }}" >
 
-  {% if is_translated %}
-    {% for translation in document.other_translations %}
+  {% if other_translations %}
```
jezdez: This `if` will not work correctly since `document.other_translations ` is the related manager and not some truthy object. You must use `exists()`

kuma/wiki/helpers.py
```diff
@@ -324,8 +325,12 @@ def wiki_url(context, path):
     Create a URL pointing to Kuma.
     Look for a wiki page in the current locale, or default to given path
     """
+    request = context['request']
+    if waffle.flag_is_active(request, 'dumb_wiki_urls'):
```
jezdez: Hm, how is this dumb? More "simplistic"?

groovecoder: Right - it just skips all the "smart" logic bellow here. Per New Relic, the following calls to `memcache.get`, `reverse`, and `Document.objects`, and/or `redirect_url` make up as much as 5-10% of the time spent in `document.html`

jezdez: Yeah, I think we should make the "dumb" the default as soon as possible and live with the redirecting nature of the URLs

kleberveridiano: o yeah!

2015-06-29 13:52 GMT-03:00 Jannis Leidel <notifications@github.com>:

kuma/wiki/templates/wiki/document.html
groovecoder: It seems return some kind of truthy object, because the spot check works fine here and for [the other boolean check for the lang switcher down on line 81](https://github.com/groovecoder/kuma/commit/8795d34d3aff5eb2a51c5f442c1bf1b764b88f06#diff-09328eed22f2f3060a415388b98060ccR81)

I removed the call to `exists` because [the docs](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#django.db.models.query.QuerySet.exists) say...
> if some_queryset has not yet been evaluated, but you know that it will be at some point, then using some_queryset.exists() will do more overall work (one query for the existence check plus an extra one to later retrieve the results) than simply using bool(some_queryset), which retrieves the results and then checks if any were returned.
We can't call `bool` directly in jinja, but if we want to be explicit, I could add a `has_translations = bool(doc.other_translations)` type casting call to the view and pass it along to the context?

jezdez: Wait, you're right, `other_translations` is a cached property of the document object and *not* a related manager as I thought. So forget what I said.

jezdez:
r+

groovecoder:
Dang - NR shows 20 waffle_flag SELECT queries now. :frowning: Need to move the check up higher if I can.

groovecoder:
Note: I enabled `dumb_wiki_urls` for everyone on production. I asked #mdn to re-ping bugs in the [zones tracking bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1158239) if they notice any extra issues pop up. If nothing extra comes up this week, I propose we remove this flag and make these dumb urls permanent to avoid their logic **and** the memcache:get to the waffle flag table.

jezdez:
@groovecoder Works for me

PR #3285, opened 2015-06-26
Fix Bug 1177571: Standardize and variablize item spacing 

stephaniehobson:
Also fix bug 1177223: change list item spacing to margin instead of padding.

Created variables $content-horizontal-spacing and $content-vertical-spacing and attempted to replace references to 6px and 8px with them. Please note: I have attempted to replicate the most common current usage of these values, discussing if there is value in these design decisions is out of the scope of this PR.

Minor display changes in some places, no more than 2px difference, all of which I think are acceptable.

Test pages:
https://developer-local.mozilla.org/en-US/docs/User:teoli/In_content
- quick links
- specification tables (small change)
- summary boxes (small change, partially due to li change)
- there's also a few lists where you can see the effect of the switch from padding to margin
Your profile page
- form items should not change appearance
https://developer-local.allizom.org/en-US/docs/User%3Astephaniehobson%3Atest
- has a tripple nested list so you can observe switch from padding to margin on lis
https://developer-local.allizom.org/en-US/docs/MDN/Contribute/Localize/RSS_feeds
- scroll down to the bottom to look at the "fullwidth-table"s
https://developer-local.allizom.org/en-US/docs/Web/CSS/display
- left sidebar (slightly less space)
- small tags in the footer (should not be any change)
https://developer-local.allizom.org/en-US/docs/Web/API/AnimationEvent/initAnimationEvent
- left sidebar (links should be closer, but icons should not be mis-aligned)
A page that needs an editorial review
- box in left hand column should have a bit more spacing between the form and button 
https://developer-local.allizom.org/en-US/dashboards/revisions
- shouldn't look substantially different
A zone with a left nav
- shouldn't look any different.

### DIFF
--- a/media/stylus/base/elements/forms.styl
+++ b/media/stylus/base/elements/forms.styl
@@ -15,7 +15,7 @@ input, button, textarea, select, optgroup, option {
 input[type=text], input[type=password], input[type=search], input[type=email], input[type=url], textarea {
     background: #fff;
     border: 1px solid #cbc8b9;
-    padding: 6px 8px;
+    padding: $content-vertical-spacing $content-horizontal-spacing;
 }
 
 input[type='search'] {
--- a/media/stylus/components/components.styl
+++ b/media/stylus/components/components.styl
@@ -55,7 +55,7 @@
         a {
             set-smaller-font-size();
             display: block;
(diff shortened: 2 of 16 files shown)


darkwing:
R+, not seeing any issues.  \o/

PR #3287, opened 2015-06-29
Only show active users as contributors

jezdez:


jezdez:
This also removes the code comment as mentioned in #3280.

PR #3288, opened 2015-06-29
Fix bug 1148727 - Reenable user after removing a ban.

jezdez:


kuma/users/admin.py
```diff
@@ -21,7 +21,8 @@ class UserAdmin(BaseUserAdmin):
 
 class UserBanAdmin(admin.ModelAdmin):
     fields = ('user', 'by', 'reason', 'is_active')
-    list_display = ('user', 'by', 'reason')
+    list_display = ('user', 'by', 'reason', 'is_active')
+    list_editable = ('is_active',)
```
jezdez: This also adds the ability to toggle a ban easily from the admin list.

robhudson: Having both `UserBan.is_active` and `User.is_active` that are essentially opposites of each other is very confusing here.

kuma/users/tests/test_models.py
```diff
@@ -108,10 +108,19 @@ def test_ban_user(self):
...
         ok_(testuser_unbanned.is_active)
+
+        ban.is_active = True
+        ban.save()
+        testuser_unbanned = self.user_model.objects.get(username='testuser')
+        ok_(not testuser_unbanned.is_active)
+
+        ban.delete()
+        testuser_unbanned = self.user_model.objects.get(username='testuser')
+        ok_(testuser_unbanned.is_active)
```
robhudson: Seems like a good idea to also assert the `is_banned` and `active_ban` methods here as well.

robhudson:
r+wc

kuma/users/admin.py
kleberveridiano: ok!

kuma/users/tests/test_models.py
kleberveridiano: ok!

jezdez: Good idea!

kuma/users/admin.py
jezdez: Yeah, I don't like it either, should we rename it?

robhudson: I don't think it is necessary for this PR but it was a cause of major confusion for me at first glance.

jezdez: Understood, I'll file a ticket

jezdez: https://bugzilla.mozilla.org/show_bug.cgi?id=1178399

PR #3289, opened 2015-06-29
Fix bug 1178252 - Stop updated the json blob when a wiki document is logically deleted like when it's marked as spam.

jezdez:
This also fixes and reenables some deferred tests that were skipped.

### DIFF
--- a/kuma/wiki/tests/test_models.py
+++ b/kuma/wiki/tests/test_models.py
@@ -7,10 +7,10 @@ from xml.sax.saxutils import escape
 import mock
 from nose.tools import eq_, ok_
 from nose.plugins.attrib import attr
-from nose import SkipTest
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.test.utils import override_settings
 
 from constance import config
 from waffle.models import Switch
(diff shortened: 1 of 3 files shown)


kuma/wiki/tests/test_models.py
```diff
@@ -891,43 +891,61 @@ def test_get_rendered(self, mock_kumascript_get):
...
         # Initially empty json field should be filled in after render()
-        eq_(None, self.d1.json)
+        eq_(self.d1.json, None)
         self.d1.render()
+        # reloading from db to get the updates done in the celery task
+        self.d1 = Document.objects.get(pk=self.d1.pk)
         ok_(self.d1.json is not None)
 
-        time.sleep(0.1)  # Small clock-tick to age the results.
+        time.sleep(1.0)  # Small clock-tick to age the results.
```
jezdez: This was needed for the test to pass :(

kuma/wiki/tests/test_models.py
```diff
@@ -891,43 +891,61 @@ def test_get_rendered(self, mock_kumascript_get):
...
+        """
         config.KUMASCRIPT_TIMEOUT = 1.0
         mock_kumascript_get.return_value = ('<p>summary!</p>', None)
-
         ok_(not self.d1.rendered_html)
         result_summary = self.d1.get_summary()
+        ok_(not mock_kumascript_get.called)
+        ok_(not self.d1.rendered_html)
+
+        self.d1.render()
```
jezdez: This is needed now that get_summary doesn't call get_rendered anymore.

kuma/wiki/tests/test_models.py
kleberveridiano: ok!

kuma/wiki/tests/test_models.py
kleberveridiano: ok!

kleberveridiano:
I´m very happy now!

kuma/wiki/tests/test_models.py
groovecoder: :cry:

kuma/wiki/tests/test_models.py
groovecoder: :+1:

PR #3290, opened 2015-06-29
Bug 1125850 - Optimize sitemap celery task

jezdez:
This does a bunch of things:

- limit the documents we want to write sitemap files for (by using an already existing model manager)
- stop fetching the html blob from the database
- split the sitemap building into tasks per locale to distribute load to all celery nodes
- use a chord to write the sitemap index after building all individual sitemap files
- write sitemap files that are above the 50k items per sitemap limit instead of just the first one

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -14,7 +14,7 @@ from django.dispatch import receiver
 from django.template.loader import render_to_string
 from django.utils.encoding import smart_str
 
-from celery import task
+from celery import task, chord
 from constance import config
 from lxml import etree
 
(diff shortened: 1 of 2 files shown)


kuma/wiki/tasks.py
```diff
@@ -296,44 +296,69 @@ class WikiSitemap(GenericSitemap):
...
+def build_index_sitemap(results):
+    """
+    A chord callback task that writes a sitemap index file for the
+    given results of :func:`~kuma.wiki.tasks.build_locale_sitemap` task.
+    """
+    sitemap_parts = [SITEMAP_START]
 
-        del queryset  # Force the gc to cleanup
+    for result in results:
+        # result can be empty of no documents were found
```
robhudson: s/of/if/

robhudson:
Looks great. r+wc

PR #3300, opened 2015-07-02
Bug 1177798 - Order contributors by recency of last edit.

jezdez:
In other words, the active user with the most recent edit shows up first.

This also moves the fetching of the contributors into a cacheback job to be able to invalidate the cache asynchronously.

### DIFF
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -1,3 +1,6 @@
+import collections
+
+from django.contrib.auth import get_user_model
 from kuma.core.jobs import KumaJob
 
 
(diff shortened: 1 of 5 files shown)


kuma/wiki/models.py
```diff
@@ -1806,7 +1808,7 @@ def make_current(self):
     def __unicode__(self):
         return u'[%s] %s #%s: %s' % (self.document.locale,
                                      self.document.title,
-                                     self.id, self.content[:50])
```
jezdez: I find the fact we show revision content in the string version of the revision silly.

groovecoder: :+1:

kuma/wiki/jobs.py
```diff
@@ -44,3 +47,45 @@ def fetch(self, locale):
...
+                                                .values_list('creator_id',
+                                                             flat=True))
+
+        if not recent_creator_ids:
+            return self.empty()
+
+        # then return the ordered results given the ID list, MySQL only syntax
+        select = collections.OrderedDict([
+            ('ordered_ids',
+             'FIELD(id,%s)' % ','.join(map(str, recent_creator_ids))),
```
robhudson: So this will end up looking like `FIELD(id,5,3,2,6,7,1)`? That looks strange to me.

jezdez: Strange indeed, MySQL.. http://dev.mysql.com/doc/refman/5.5/en/string-functions.html#function_field

jezdez: Blast from the past, http://davedash.com/2010/02/11/retrieving-elements-in-a-specific-order-in-django-and-mysql/

robhudson: Why an ordered dict here if it's the only thing in it?

robhudson:
r+

kuma/wiki/jobs.py
jezdez: It's what's recommended in the Django docs since order matters here and we may want to extend it in the future.

PR #3305, opened 2015-07-03
Bug 1177264 - Remove un-used teamwork app.

jezdez:
Signed-off-by: Jannis Leidel <jannis@leidel.info>


kuma/wiki/migrations/0001_initial.py
```diff
@@ -285,7 +284,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='document',
             name='team',
-            field=models.ForeignKey(blank=True, to='teamwork.Team', null=True),
+            field=models.ForeignKey(blank=True, to='wiki.DocumentTag', null=True),
```
jezdez: This is needed to be able to build the fake models during migration up and not fail with an import error.

kuma/wiki/migrations/0003_auto_20150703_0419.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
```
jezdez: This migration file is basically changes that we've done in the past weeks and gone unnoticed, e.g. extension of the locale setting, change of the related names of the tag fields.

kuma/wiki/migrations/0001_initial.py
groovecoder: Seems like a hack? Do we plan to get rid of this in a later squashing of the migrations back into a single `0001` migration without referencing teamwork at all?

groovecoder:
r+ after we make a decision (any decision) on https://github.com/mozilla/kuma/pull/3305#discussion_r33966144

groovecoder:
r+ after we make a call on https://github.com/mozilla/kuma/pull/3305#discussion_r33966144

kuma/wiki/migrations/0001_initial.py
jezdez: Yeah, that's the plan, squashing migrations is still an important part of the migration system, even after it landed in Django.

groovecoder: Okay, this is good-to-go as far as I can tell ... @openjck - why the "not ready" label?

openjck: Looked like [this comment](https://github.com/mozilla/kuma/pull/3305#issuecomment-118953720) still needs to be addressed?

I like using the *not ready* label to help me ignore certain pull requests when doing a PR run, but if it's not helpful to the back-end team I can stop adding it when I triage our pull request backlog.

groovecoder: Ah, I've been treating "not ready" as "not ready *to be reviewed*" ...

openjck: Ah, yeah. The label is definitely ambiguous. I chose the name *not ready to be merged* in Discord for that reason.

I'll stay out of the way now. :smile:

PR #3306, opened 2015-07-03
Fix bug 1179897 - Reduce number of db queries on profile page.

jezdez:
This does a bunch of things:

- Move revision helper to custom user model (ref bug 1180208).
- Use else clause for for loop for getting rid of slow count query.
- Prefetch revision documents to reduce related queries.
- Use a cached property for the previous revision for in-page load caching.
- Refactor the gravatar method with better caching.

### DIFF
--- a/kuma/attachments/feeds.py
+++ b/kuma/attachments/feeds.py
@@ -16,8 +16,7 @@ class AttachmentsFeed(DocumentsFeed):
         return item.title
 
     def item_description(self, item):
-        previous = item.get_previous()
-        if previous is None:
+        if item.previous is None:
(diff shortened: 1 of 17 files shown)


groovecoder:
Needs rebase. Are we scared of the gravatar cacheback job?

jezdez:
@groovecoder Not particularly since we do md5 hashing in the gravatar function I think the computational overhead of the function is higher than the impact the traffic from Memcached.

groovecoder:
I'll merge & push this. Then I'll try re-enabling the contributor bar for beta users to see if it causes a spike in memcache traffic from the cacheback job generating the gravatars in the bar.

PR #3310, opened 2015-07-03
Bug 815693: Rename .user-state

stephaniehobson:
Renamed user-state to oauth-logged-in to mirror oauth-login class as the two components swap places depending on user state. (I know this seems picky but it's to support some other more complex changes).

Testing:
- sign in, make sure signout widget displays correctly


openjck:
Nice! This looks like all instances of `user-state` alright. Spot-checking now...

PR #3313, opened 2015-07-03
Fix Bug 1180280: Mobile search field

stephaniehobson:
When viewing the mobile site with the header site search expanded it should appear centered.

Test in RTL as well.


openjck:
Nice!

PR #3329, opened 2015-07-08
Bug 1125850 - Reduce the chunk size for rendering stale documents by 50%.

jezdez:
This is a test to see if that reduces the memory/swap usage of the celery workers.


robhudson:
r+

PR #3330, opened 2015-07-08
bug 957802 - Fix issues running tests in new vagrant install

jwhitlock:
Updates for issues found when running tests (`./manage.py test`) in a new vagrant install, with just a `git submodule update` and a superuser:

* Some content tests failed due to attribute order.  Adjusted the content tool to alpha-sort attributes in tests, but leave them alone in regular content flow.
* Some submodules don't ignore *.pyc files, so I set them to ignore non-tracked files, so that `git status` is clean.
* A test required setting up translated strings.  Fixed the mock so it works w/o translations installed.

### DIFF
--- a/kuma/users/tests/test_tasks.py
+++ b/kuma/users/tests/test_tasks.py
@@ -35,15 +35,16 @@ class TestWelcomeEmails(UserTestCase):
         eq_(expected_to, welcome_email.to)
         ok_(u'utm_campaign=welcome' in welcome_email.body)
 
-    @mock.patch('kuma.core.utils.strings_are_translated')
+    @mock.patch('kuma.users.tasks.strings_are_translated')
     def test_dont_send_untranslated_language_email(self,
                                                    strings_are_translated):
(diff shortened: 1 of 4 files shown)


kuma/users/tests/test_tasks.py
```diff
@@ -35,15 +35,16 @@ def test_default_language_email(self):
...
         ok_(u'utm_campaign=welcome' in welcome_email.body)
 
-    @mock.patch('kuma.core.utils.strings_are_translated')
+    @mock.patch('kuma.users.tasks.strings_are_translated')
     def test_dont_send_untranslated_language_email(self,
                                                    strings_are_translated):
         strings_are_translated.return_value = False
         testuser = self.user_model.objects.get(username='testuser')
-        send_welcome_email(testuser.pk, 'tlh')  # Qapla'
+        send_welcome_email(testuser.pk, 'tlh')  # mev!
```
jezdez: Nice ;)

jezdez:
r+

PR #3332, opened 2015-07-09
Fix bug 1174804 - Fix rendering of choice fields with Django >=1.7.

jezdez:
This is a case of a monkeypatch in jingo being out of date and a regression from us updating to a current Django release while jingo hasn't been released yet.

This makes use of https://github.com/jbalogh/jingo/commit/0d293ae8aefa0419961ab6480f38a3d2dfbadc44 and its explicit monkey patching of the `django.forms.widgets.RadioChoiceInput` class that replaces the old `django.forms.widgets.RadioInput`.

Note: There is https://github.com/jbalogh/jingo/issues/67 to ask the maintainers to do a proper release with little results yet.

### DIFF
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -462,6 +462,17 @@ class ProfileViewsTest(UserTestCase):
                 response.context['profile_form'].fields[field].label, basestring),
                 'Field %s is a string!' % field)
 
+    def test_bug_1174804(self):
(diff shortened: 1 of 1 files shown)


jezdez:
@stephaniehobson Let me know if you need help with the review of this.

stephaniehobson:
Works! User experience for that signup could use a little work... but that's out of scope.

PR #3334, opened 2015-07-09
Fixes bug 1142203: refactor datetime.now into django.utils timezone.now

jezdez:


jamonation:
Hey @jezdez not to worry - the rewrite is kind of a big deal. Happy that this will get done and that I'm not the one to refactor it again :D

jezdez:
@jamonation Thanks for your understanding, much appreciated :heart:

robhudson:
Picking this up again as part of the move to Postgresql. I rebased as best I could and will await a test run to see what might be left to do.

jezdez:
@robhudson Awesome!

robhudson:
Tests pass. I'm going to look for any new code setting datetimes that might have been added in the past 6 mos when this was last worked on. Any preliminary reviews would be nice.

I added the "webops" label b/c we can't merge this until we move to AWS.

kuma/core/tests/test_helpers.py
```diff
@@ -96,6 +97,9 @@ def test_message_with_url_is_link(self):
 class TestDateTimeFormat(UserTestCase):
     def setUp(self):
         super(TestDateTimeFormat, self).setUp()
+        self.default_timezone = timezone.get_default_timezone()
+        self.old_times = (datetime.fromordinal(733900)
```
jezdez: I don't mean to be obtuse, but I'm not a fan of using the `datetime.fromordinal` function here, even if it was used before. To quote the docs: `Return the datetime corresponding to the proleptic Gregorian ordinal, where January 1 of year 1 has ordinal 1.` Can't we just use a specific time and be done with it?

kuma/wiki/tests/test_templates.py
```diff
@@ -245,7 +246,7 @@ def test_revision_view(self):
         """Load the revision view page and verify the title and content."""
         d = _create_document()
         r = d.current_revision
-        r.created = datetime(2011, 1, 1)
+        r.created = datetime(2011, 1, 1, tzinfo=timezone.utc)
```
jezdez: Why utc here?

robhudson: You should as @jezdez it looks like he made the change in https://github.com/mozilla/kuma/commit/c276efff168af976de499c3b1a4b46717417bff4 ;)

jezdez: Who the hell is this guy and why did he use utc for god's sake? For real, this should use the default timezone I think.

robhudson: Hehe, ok I can take a closer look.

jezdez:
FTR, this depends on moving to Postgres ([bug 1159930](https://bugzilla.mozilla.org/show_bug.cgi?id=1159930))

PR #3335, opened 2015-07-09
bug 997788 - fix ga event label to track position clicked

groovecoder:
As we roll out the contributor bar, we want to record which contribution **position** is clicked so we know if users tend to click the contributors listed first. Not sure why this was changed in https://github.com/mozilla/kuma/commit/42a74bed25f578a05da32a12b32cb92eadf6c430

### DIFF
--- a/media/js/wiki.js
+++ b/media/js/wiki.js
@@ -518,10 +518,11 @@
         $contributors.on('click', 'a', function(e) {
             var newTab = (e.metaKey || e.ctrlKey);
             var href = this.href;
+            var index = $(this).parent().index() + 1;
             var data = {
                 category: 'Top Contributors',
                 action: 'Click position',
-                label: href
+                label: index
             };
 
             if (newTab) {


media/js/wiki.js
```diff
@@ -518,10 +518,11 @@
         $contributors.on('click', 'a', function(e) {
             var newTab = (e.metaKey || e.ctrlKey);
             var href = this.href;
+            var $li = $(this).parent();
```
darkwing: I think we should either not use a var, and just make it `label: $(this).parent().index()` or make the var `index` and set it to that value, using that var in the data object.  In any case, `$li` isn't really used anywhere and it seems "half-way" to what we actually need, which is weird.

darkwing:
I guess one thing I'd mention is that we may want to increment the index by 1 so we mentally think "first " position.  Just a thought.  We aren't computers! :D

groovecoder:
Updated.

groovecoder:
:up: dated.

darkwing:
Nice! R+

darkwing:
...waiting on tests...

PR #3336, opened 2015-07-09
fix bug 1173122 - fix footer license link

groovecoder:


templates/base.html
```diff
@@ -155,7 +155,7 @@
       <div class="column-main">
         {% block footer_copyright_redesign %}
         <p>
-          {% trans copyright_url=wiki_url('Project:MDN/About#Copyrights_and_licenses'), about_url=wiki_url('Project:MDN/About'), thisyear=thisyear() %}
+          {% trans copyright_url='/docs/MDN/About#Copyrights_and_licenses', about_url='/docs/MDN/About', thisyear=thisyear() %}
```
jezdez: When opening the URL `/docs/MDN/About#Copyrights_and_licenses` I get redirected to the `en-US` version but the hash is lost in the redirect :(

jezdez:
rwc+

groovecoder:
Updated to use `wiki_url` which uses the locale prefixer to avoid the redirect that loses the url hash.

jezdez:
r+

PR #3339, opened 2015-07-12
bug 957802 - Improve mysql init script

jwhitlock:
/etc/init.d/mysql greps for the mysql process to determine if mysql is running.  When running, the process count includes the mysqld process and the grep for that process (count=2).  However, 5% - 10% of the time (and maybe more common under vagrant / ansible), the grep is not in the process list, causing false negatives when detecting the mysqld process is running.  When provisioning, this results in the utf8_distinct_ci coalition not being applied, migrations failing, and provision failing until mysql is restarted (maybe by rebooting the vagrant instance).

This new init script ignores the grep process, making the run detection more reliable.  Tested by cleanly running `vagrant up` from scratch.

### DIFF
--- a/provisioning/roles/geerlingguy.mysql/files/etc/init.d/mysql
+++ b/provisioning/roles/geerlingguy.mysql/files/etc/init.d/mysql
@@ -1,0 +1,182 @@
+#!/bin/bash
+#
+### BEGIN INIT INFO
+# Provides:          mysql
+# Required-Start:    $remote_fs $syslog
+# Required-Stop:     $remote_fs $syslog
+# Should-Start:      $network $time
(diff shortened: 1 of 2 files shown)


jwhitlock:
Submitted upstream as http://bugs.mysql.com/bug.php?id=77696

jezdez:
Woah! That is awesome!

jezdez:
r+

jezdez:
Oh crap, I only now noticed that this changes a vendored Ansible role. D'oh!

jezdez:
Opened https://github.com/mozilla/kuma/pull/3340 to fix this. D'oh

PR #3340, opened 2015-07-13
Bug 957802 - Reverted modification of vendored MySQL role and moved it into kuma role.

jezdez:


jezdez:
I would recommend filing an issue upstream with the role author to ship the init file there instead.

jwhitlock:
+r, checked with ``vagrant destroy && vagrant up``.  Submitted upstream as https://github.com/geerlingguy/ansible-role-mysql/pull/44 .

PR #3341, opened 2015-07-13
Bug 1183082 - Fix attachment feeds.

jezdez:


darkwing:
@jezdez "attachment"  !  :)

kuma/attachments/feeds.py
```diff
@@ -10,15 +10,19 @@ class AttachmentsFeed(DocumentsFeed):
     subtitle = _("Recent revisions to MDN file attachments")
 
     def items(self):
-        return AttachmentRevision.objects.order_by('-created')[:50]
+        return (AttachmentRevision.objects.prefetch_related('creator',
+                                                            'attachment')
+                                          .order_by('-created')[:50])
```
groovecoder: Would `select_related` be more efficient here, since `creator` and `attachment` are both `ForeignKey` relations? `select_related` would get them in the same query with a SQL join, rather than `prefetch_related` doing separate lookups for each and joining in Python?

groovecoder:
r+ with comments

kuma/attachments/feeds.py
jezdez: Yeah, `select_related` would do this in one query, but since we're running on a MySQL cluster with query cache on it would essentially invalidate the query cache if any of the inner join tables is changed, e.g. if a new user is added or even just updated.

So assuming a case in which a new user signs up, using `prefetch_related` here would only trigger the invalidation of the query cache for the `creator` prefetch query. While the other two for the attachment revision and the attachment would come straight from memory. All in all this is a micro optimization that given the number of queries we do against MySQL may not matter, but it's nicer to use the query cache as much as possible and live with the slight overhead on the Python side.

groovecoder: Makes sense. r+

PR #3342, opened 2015-07-13
bug 1038036

ronakkhunt:
Allowed space in interest tag and removed repeated insertion of JavaScript.
So the JS bundle "jquery-ui", "demostudio" and "profile" were repeatedly adding following JS file.
1) "js/libs/jquery-ui-1.10.3.custom/js/jquery-ui-1.10.3.custom.min.js"  AND
2)"js/moz-jquery-plugins.js"

That was creating the problem.
Also space needed to allowed in "tag-it" Library.

### DIFF
--- a/kuma/users/templates/users/profile_edit.html
+++ b/kuma/users/templates/users/profile_edit.html
@@ -20,7 +20,6 @@
 
 {% block site_js %}
     {{ super() }}
-    {{ js('jquery-ui') }}
     {{ js('demostudio') }}
     {{ js('libs/tagit') }}
     {{ js('profile') }}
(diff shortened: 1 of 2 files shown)


groovecoder:
@darkwing @stephaniehobson @openjck - who wants this one?

openjck:
I'll take it!

openjck:
Looks awesome. Thanks so much!

ronakkhunt:
@openjck Do I need to mark this as resolved in Bugzilla

openjck:
Good catch. Yes, you're welcome to mark it as `RESOLVED` and `FIXED`. GitHub automatically closes the bug when the commit message starts with the word *Fix*.

ronakkhunt:
@openjck So, I tried to mark it as resolved, But there was no option "Fixed" in resolution dropdown. May be i do not have enough privileges.

openjck:
Ah yes, that's probably the reason. I went ahead and marked it as `RESOLVED` and `FIXED`. Thanks for your contribution!

PR #3345, opened 2015-07-14
Bug 957802: Ignore sanitized database dump

openjck:
I always worry when copying the sanitized database dump to the project
root before importing it. Let's be sure it can never be committed.


groovecoder:
Needs a bug # in commit message. Can use bug 957802

openjck:
Updated!

darkwing:
R+, added "prod_devmo.sql" to my repo and `git status` did not list it.

PR #3346, opened 2015-07-15
bug 1182627 - add nofollow to profile compare link

groovecoder:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1182627#c0, the Yahoo crawler is getting into `compare_revisions` and triggering slow transactions that are affecting python response times.

This adds `rel="nofollow"` to the last `compare_revisions` link on the site that was missing it. It should discourage more slow transactions from crawlers like Yahoo.


jezdez:
r+ assuming tests pass

PR #3347, opened 2015-07-16
Fix bug 1173178 - Catch Unicode decode errors in the search API.

jezdez:
This shows a 404 for the search if it's not working because of a decoding error of search query parameters.

### DIFF
--- a/kuma/search/utils.py
+++ b/kuma/search/utils.py
@@ -14,6 +14,9 @@ log = logging.getLogger('kuma.search.utils')
 SEARCH_DOWN_DETAIL = _('Search is temporarily unavailable. '
                        'Please try again in a few minutes.')
 
+SEARCH_ERROR_DETAIL = _('Something went wrong with the search query. '
+                        'Please try again in a few minutes.')
+
 
 class QueryURLObject(URLObject):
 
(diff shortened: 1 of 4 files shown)


kuma/search/utils.py
```diff
@@ -72,13 +75,17 @@ def search_exception_handler(exc):
...
+            # FIXME: This really should return a 503 error instead but Zeus
+            # doesn't let that through and displays a generic error page in that
+            # case which we don't want here
+            log.error('Elasticsearch exception: %s' % exc)
+            return Response({'detail': SEARCH_DOWN_DETAIL},
+                            status=status.HTTP_200_OK)
+        elif isinstance(exc, UnicodeDecodeError):
+            log.error('UnicodeDecodeError exception: %s' % exc)
+            return Response({'detail': SEARCH_ERROR_DETAIL},
+                            status=status.HTTP_404_NOT_FOUND)
```
groovecoder: A plain 400 is probably more appropriate for these kinds of errors.

Also, this path isn't reached via [the url originally posted in the bug](https://developer-local.allizom.org/en-US/search?q=%BF%27%22%28&topic=api&skill=intermediate&type=howto%2F), so I assume this is defensive coding in case something else raises a `UnicodeDecodeError`?

jezdez: I've done some research and 404 are more appropriate for this type of issue, wrong parameters.

The code in question *is* handled by this DRF exception handler, there were just two spots where this would show tracebacks.

groovecoder: :+1:

PR #3349, opened 2015-07-16
[bug 1032455] Fix post-merge issues

hoosteeno:


kuma/wiki/templates/wiki/document.html
```diff
@@ -416,7 +416,10 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
       {# This can be removed once the social network button multivariate test is completed. #}
       mdn.optimizely.push(['activate', 2796520257]);
   </script>
-  {% if waffle.flag('share_links') or waffle.flag('helpfulness') %}
-    {{ js('document') }}
+  {% if waffle.flag('share_links') %}
+    {{ js('social') }}
+  {% endif %}
```
jezdez: In case those flags are exclusive there is also `{% elif %}`

hoosteeno: Both/either/neither are all valid. That's why the original code didn't work -- it coupled them.

jezdez: Understood :+1:

stephaniehobson:
R+

PR #3350, opened 2015-07-16
Bug 1177264 - Cache the available search filters much longer.

jezdez:
That should reduce the db queries a notch.


groovecoder:
Spot-check looks good; DjDT reports this eliminates 2 extra cache calls during typical document view.

PR #3353, opened 2015-07-21
Fix Bug 1184218: Tabzilla won't stay open

stephaniehobson:
Tabzilla is loading multiple times, changing the true/false evaluations back to this format stops it. Javascript FTW!


darkwing:
R+

PR #3354, opened 2015-07-21
Bug 1185627: MDN 10th anniversary zone styles

stephaniehobson:
Initial assets for 10th anniversary zone. Covers quotes, learn more buttons, inverted colour promos, hero image and headline.

Designs are here: https://bugzilla.mozilla.org/show_bug.cgi?id=1140601 (we're not going for 1:1 match).

Testing: 
1) Recreate [this zone page](https://developer.allizom.org/en-US/docs/MDN_at_ten) locally. 
2) Include the generated css file in the"styles" section of the zone configuration `@import url(/media/css/mdn10.css);`


darkwing:
I've completed this locally and it looks col!  Outside of ensuring the design looks good and the audio plays, what is the scope of this review?

One issue I have noticed is that the history page IFrame doesn't look good at mobile width.  Can we simply inject that content into the page instead of using an IFrame?

stephaniehobson:
@darkwing The history page is not connected to this PR.

darkwing:
R+!  Looks awesome, great work @stephaniehobson !

PR #3355, opened 2015-07-22
A couple of documentation tweaks

davehunt:


docs/tests-ui.rst
```diff
@@ -12,6 +12,7 @@ Installing Dependencies
 
 3. From the `tests/ui/` directory, use NPM or another package manager to install Intern::
 
+    mkdir node_modules
```
groovecoder: Weird. I didn't have to do this. :confused:

groovecoder:
Thanks @davehunt. Per [our contributing guidelines](https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md), can you prefix your commit messages with a bug number? For these kinds of docs/setup issues, I always (re-)use [Bug 957802 - Streamline new contributor setup](https://bugzilla.mozilla.org/show_bug.cgi?id=957802). E.g.:
```
bug 957802 - Fix Intern link
bug 957802 - Create node_modules directory to force local package install
```

docs/tests-ui.rst
darkwing: I'd prefer we left that out; `npm` is meant to create the directory if it's not there so you may need to configure your specific machine differently.

docs/tests-ui.rst
```diff
@@ -1,7 +1,7 @@
 Client-side Testing with Intern
 ===============================
 
-The client-side testing tool for the MDN front-end is `Intern <http://intern.io>`_, a Selenium WebDriver API which allows developers to write automated testing via JavaScript. Intern is an open source project created and maintained by `SitePen <http://sitepen.com>`_.
+The client-side testing tool for the MDN front-end is `Intern https://theintern.github.io/>`_, a Selenium WebDriver API which allows developers to write automated testing via JavaScript. Intern is an open source project created and maintained by `SitePen <http://sitepen.com>`_.
```
darkwing: Good catch, they moved the site since I started tests! :)

docs/tests-ui.rst
davehunt: Okay, I wonder why it wasn't created on my machine then.

davehunt:
Updated.

docs/tests-ui.rst
darkwing: Ooops, I think you're missing a `<` at the beginning of the link?

davehunt: Gah! Well spotted. Fixed now.

darkwing:
R+, thank you!

PR #3357, opened 2015-07-22
Fix Bug 1186655: 10th anniversary logo

stephaniehobson:
White and grey versions included, waffled as mdn10_logo.

Testing: 
- [x] white logo appears on home page and zones
- [x] grey logo appears on regular pages
- [x] logo collapses appropriately on tablet and mobile screens


stephaniehobson:
Now with crisper logo! Maybe. The "d" is still fuzzy on my laptop but the "10" is much crisper. Let me know how it looks to you.

darkwing:
R+; I think we may be able to do better with logo but let's move forth for now.

PR #3359, opened 2015-07-22
Fix Bug 1186660: MDN 10 homepage tile

stephaniehobson:
Assets and homepage title waffled as `mdn10_tile`.

The page that this tile points to has not yet been created on production, it can be launched but not waffled on yet.

@groovecoder the strings in this PR need to be extracted and sent to verbatim.


darkwing:
R+, looks awesome!

PR #3363, opened 2015-07-23
Bug 1185627: MDN 10th - chapter audio and quote tweaks

stephaniehobson:
Hide "Open in Codepen" button and prevent quotation mark from messing with text line height.


darkwing:
Ballin!  This looks awesome, @stephaniehobson !  R+

PR #3364, opened 2015-07-23
Various Vagrant fixes.

jezdez:
The call to `config.env.enable` was needed to be done before reading the env vars.


jezdez:
The call to `config.env.enable` was needed to be done before reading the env vars.

jwhitlock:
r+ 

Fresh install with VirtualBox 5.0 and Vagrant 1.7.4 with overrides in .env:
```
export VAGRANT_MEMORY_SIZE=8192
export VAGRANT_CPU_CORES=4
export VAGRANT_ANSIBLE_VERBOSE=true
```
Got the requested 8 GB memory, 4 processors, verbose ansible, and a 64-bit Ubuntu.

PR #3366, opened 2015-07-24
Bug 1187363 - "Open in CodePen" option issues

darkwing:
This:

-  More reliably gets sample contents 
-  Prevents duplicated "Open in..." buttons for tables.

Can copy this content to test:  https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Drawing_DOM_objects_into_a_canvas


stephaniehobson:
R!

PR #3367, opened 2015-07-24
fix bug 1185753 - Fix text color in languages dropdown

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1185753

### DIFF
--- a/media/stylus/components/wiki/page-buttons.styl
+++ b/media/stylus/components/wiki/page-buttons.styl
@@ -14,6 +14,7 @@
         display: inline-block;
         position: relative;
         bidi-value(text-align, left, right);
+        color: $text-color;
 
         a, button, input[type='submit'], input[type='button'] {
             set-smaller-font-size();


media/stylus/components/wiki/page-buttons.styl
```diff
@@ -14,6 +14,7 @@
         display: inline-block;
         position: relative;
         bidi-value(text-align, left, right);
+        color: $text-color;
```
stephaniehobson: The problems with the styles originate with the .submenu class, let's associate this declaration with that class so it fixes the problems in other places it's used as well.

In mixin_submenu.styl could go: 
a) with the background-colour declaration in submenu-set
b) with the link color declaration in $submenu

a is better for maintainability
b is better for performance

If you don't have a preference I guess b is the right place for now.

stephaniehobson:
Superseded by #3379.

PR #3369, opened 2015-07-27
Bug 1177798 - Order contributors by recency of last edit.

jezdez:
This also moves the contributor bar fetching into a cacheback job.

### DIFF
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -1,7 +1,7 @@
 {% extends "wiki/base.html" %}
 {% block title %}{{ page_title(document.title + seo_parent_title) }}{% endblock %}
 
-{% from "wiki/includes/document_macros.html" import build_document_crumbs, get_document_quick_links, get_document_buttons, get_document_subnav, get_approvals_html, document_watch, share_links with context %}
+{% from "wiki/includes/document_macros.html" import build_document_crumbs, get_document_quick_links, get_document_buttons, get_document_subnav, get_approvals_html, document_watch, share_links, contributor_links with context %}
 
 {% if fallback_reason == 'no_translation' %}
   {% set help_link = url('wiki.translate', document_path=document.slug, locale=document.locale)|urlparams(tolocale=request.locale) %}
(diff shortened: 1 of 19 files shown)


jezdez:
To be clear, there is another contributor bar in the footer of document pages, which will use the same cacheback job to populate a list of links. That list is **not** waffled and will use the cacheback job and therefor Memcache and Celery by default. If you think we should waffle the contributor footer bar as well to prevent potential issues with using Memcache/Celery let me know.

kuma/wiki/templates/wiki/document.html
```diff
@@ -179,11 +184,9 @@
...
-        {% if waffle.flag('top_contributors') %}
-            {% set contrib_limit = 13 %}
-            {% set contrib_count = contributors.count() %}
+        {% if has_contributors and waffle.flag('top_contributors') %}
 
-        <div class="contributor-avatars" data-all-text="{{ _('Show all') }}&hellip;<span class='hidden'>{{ _('contributors') }}</span>" {% if contrib_count > contrib_limit %}data-has-hidden="1"{% endif %}>
+        <div class="contributor-avatars" data-all-text="{{ _('Show all') }}&hellip;<span class='hidden'>{{ _('contributors') }}</span>" {% if contributors_count > contributor_show_limit %}data-has-hidden="1"{% endif %}>
             <span class="quickstat">
             {% trans count=contrib_count %}
             by {{ count }} contributor:
```
groovecoder: Needs to be `{{ contributors_count }}` now?

groovecoder:
Spot-check looks good. I also put a debug in the middle of the `DocumentContributorsJob.fetch` to make sure it was only called once and that subsequent pageviews didn't trigger it.

I am still concerned about the risk that the initial spike of cacheback jobs for every document could trigger memcache overload issues again. It looks like this `document.html` call is the only place that calls `get_contributors`, and 0.0002% of visitors to JavaScript clicked a contributor link in the footer.

So, if we put the entire logic of `get_contributors`, the header, and the footer displays behind the waffle flag, we should be able to throttle this up slowly and watch memcache activity via New Relic, right? (On both production & stage now that stage has memcache)

jezdez:
@groovecoder So I just made both the header and footer bar depending on the "top_contributors" waffle flag. I moved the check for it to the view since it's easier for the else branch. Can you take a look again?

jezdez:
@groovecoder BTW we could rate limit the cacheback task to something like 120/m or something

groovecoder:
This looks good to me. I toggled the flag and saw only the first request make the SQL call. If we rate-limit the cacheback job to 120/m, what happens when it hits 121/m? Will it make the next request wait, or just skip refresh from cache?

jezdez:
The rate limitation happens on Celery's backend level, so what will happen is that Celery will actively prevent from sending the task info to its workers. In other words, the task messages will pile up in RabbitMQ until Celery workers are able to work through them. Since we track the number of RabbitMQ messages with Graphite we'll see if something is wrong.

Cacheback on the other side is smart enough that if a job hasn't produced results it'll set a cache value to an empty value (in our case an empty list) with a TTL of 60 seconds by default. That's the assumed period of time we expect cacheback tasks to be resulting in properly filling the cache with an actual result (even though it may be an empty list just as well). In other words, in those 60 seconds cacheback won't try scheduling another Celery task.

groovecoder:
Cool. We also have NR RabbitMQ monitoring: https://rpm.newrelic.com/accounts/263620/dashboard/4681572/page/3

PR #3374, opened 2015-07-29
bug 1118735 - only build master branch on push

groovecoder:
I guess we'll see if Travis still builds PRs even though they're not the master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+    only:
+        - master
 cache:
     apt: true
     directories:


groovecoder:
This PR still built, and it did *not* trigger a push build, even though push builds are enabled now. (See https://github.com/mozilla/kuma/pull/3376 and https://github.com/mozilla/kuma/pull/3377 for PR that were double-built because they did not include this updated `.travis.yml`)

So, I think this is good to merge.

jezdez:
r+

PR #3376, opened 2015-07-30
bug 1177264 - remove search_filters context processor (12.8% of app server time)

groovecoder:
Per [the latest New Relic profile](https://rpm.newrelic.com/accounts/263620/applications/3172075/profiles/1716288), at least 12.8% of app server time was spent in `kuma/search/context_processors.py.search_filters`.

On top of the fact that [the search filter feature is currently waffled off](https://developer.mozilla.org/admin/waffle/flag/65/), the `searchFilters` json is only needed on the home page, where search filter suggestions are displayed (when the feature is waffled on.)

So, I moved its logic into the `homepage` view to cut out the time spent processing this context during other views.

Spot-check:
* [x] Enable the `search_suggestions` waffle flag
* Go to home page
  * [x] Verify search filter suggestions still work
* Go to a document page
  * [x] Verify document page still works
* [x] Disable the `search_suggestions` waffle flag
  * [x] Verify search filter suggestions disappear from home page
  * [x] Verify document page still works

### DIFF
--- a/kuma/landing/templates/landing/homepage.html
+++ b/kuma/landing/templates/landing/homepage.html
@@ -209,6 +209,9 @@
       {{ js('home', async=True) }}
   {% endif %}
   {% if waffle.flag('search_suggestions') %}
+    <script>
+        window.mdn.searchFilters = {{ command_search_filters|jsonencode }};
+    </script>
     {{ js('search-suggestions') }}
     <script>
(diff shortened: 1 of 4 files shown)


groovecoder:
@stephaniehobson, @openjck, or @darkwing want to take a look at this? I think you've done the most with this feature?

stephaniehobson:
R :+1:

PR #3377, opened 2015-07-30
bug 1177264 - remove extra tags.exists() check (1.1% of app server time)

groovecoder:
Per [the latest New Relic profile](https://rpm.newrelic.com/accounts/263620/applications/3172075/profiles/1716288), 1.1% of app server time is spent in `django/db/models/query.py.exists` [called by `kuma/wiki/templates/wiki/includes/document_tag.html`](https://github.com/mozilla/kuma/blob/374c12e203f80ade3d77aa401e3b6a99b1fa1a36/kuma/wiki/templates/wiki/includes/document_tag.html#L1).

This check is superfluous, because:
1. Both calling templates: [`document.html`](https://github.com/mozilla/kuma/blob/374c12e203f80ade3d77aa401e3b6a99b1fa1a36/kuma/wiki/templates/wiki/document.html#L356-L357) and [`revision.html`](https://github.com/mozilla/kuma/blob/374c12e203f80ade3d77aa401e3b6a99b1fa1a36/kuma/wiki/templates/wiki/revision.html#L56-L57) call `tags.all()` anyway
2. Most requests are to documents with tags

This PR removes the `exists()` call which removes a superfluous database query from almost every document view.

### DIFF
--- a/kuma/wiki/templates/wiki/includes/document_tag.html
+++ b/kuma/wiki/templates/wiki/includes/document_tag.html
@@ -1,4 +1,4 @@
-{% if tags.exists() %}
+{% if tags %}
   <div class="tag-attach-list contributors-sub">
     <i aria-hidden="true" class="icon-tags"></i>
     <strong>{{_('Tags:')}}</strong>&nbsp;


jezdez:
For the record, we must be vigilante about assuming that removing `exists()` is a good idea, the query is a lot faster than what `bool(queryset)` does, which is to fetch all results from the database and put them into a Python list - as that can be both heavy on the database *and* memory intensive on the Python process.

But if like in this case there is no other condition in the application code (in this case the `document_tag.html` template) as to whether to show the fetched data we can rely on the queryset cache filled by the evaluation of the if condition. As soon as there is a deviation from that rule though like not only iterating over the results we need to reintroduce the `exists()` call.

jezdez:
r+

groovecoder:
Yeah, I noticed the same thing as I was digging thru some of the rest of the thread profile. There are other areas where `exists()` shows up, but after I looked at a few them and re-read the docs on exists, I saw that we can only remove it under those conditions.

jezdez:
BTW, I want to mention that MySQL is known to be slow with counts (which exists() basically does). So with PG we wouldn't have to discuss this.

jwhitlock:
I thought it was the other way around - MySQL is fast at counts and PG is slow:

https://wiki.postgresql.org/wiki/Why_PostgreSQL_Instead_of_MySQL:_Comparing_Reliability_and_Speed_in_2007#Counting_rows_in_a_table

Not sure if an alternate query is used for exists() in PG.

jezdez:
@jwhitlock Ah right, Postgres is also slow for counts without WHERE clause, but not so much for those that have one. In case of this particular queryset we would indeed also be slow on Postgres. I think that's another sign where it's dangerous to just assume a particular pattern to be true with regard to the speed of ORM querysets on the various backends.

I opened a ticket a while ago to make us use approximated counts where we just need a number, along the lines of http://adamj.eu/tech/2014/07/16/extending-djangos-queryset-to-return-approximate-counts/ But I haven't gotten to it yet. For other cases it's best to use `exists()` as it only fetches one result item and skips select related and other potentially applied queryset modifications.

jezdez:
That article explains in a later sentence that a blank count on all items of the table is slow for InnoDB as it doesn't cache the value. MySQL will have to do a slow table scan on even with a WHERE clause: https://www.percona.com/blog/2006/12/01/count-for-innodb-tables/

I opened a ticket a while ago to make us use approximated counts where possible, along the lines of http://adamj.eu/tech/2014/07/16/extending-djangos-queryset-to-return-approximate-counts/ But I haven't gotten to it yet.

PR #3378, opened 2015-07-30
fix bug 1189222 - $subscribe: skip zone middleware

groovecoder:
Redirecting POST $subscribe requests to zone urls causes a GET to $subscribe,
which has @require_POST decorator.

I'm not sold on this approach, though. :confused: 

Checks:
* [x] Edit a document still works
* [x] Subscribe to a zone document now works


jwhitlock:
+r on local host, was able to subscribe to [Firefox_OS/Developing_Gaia](https://developer.mozilla.org/en-US/Firefox_OS/Developing_Gaia) and edit it.

PR #3384, opened 2015-08-03
Bug 1184662 - Update Intern docs to include intern version

darkwing:
Since intern just upgraded, we should tell people to use the version we've been developing on

### DIFF
--- a/docs/tests-ui.rst
+++ b/docs/tests-ui.rst
@@ -12,7 +12,7 @@ Installing Dependencies
 
 3. From the `tests/ui/` directory, use NPM or another package manager to install Intern::
 
-    npm install intern
+    npm install intern@^2.0.0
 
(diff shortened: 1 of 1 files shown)


openjck:
Mind if I grab this, @groovecoder?

groovecoder:
:+1:

docs/tests-ui.rst
```diff
@@ -12,7 +12,7 @@ Installing Dependencies
 
 3. From the `tests/ui/` directory, use NPM or another package manager to install Intern::
 
-    npm install intern
+    npm install intern@2.2.2
```
openjck: We could alternatively recommend that people run `npm install intern@^2.0.0`. That will install the latest version in the 2.0.0 branch, which should be API compatible. That way we won't need to update the file for every version.

darkwing:
John wins.  Updated.

PR #3386, opened 2015-08-04
Bug 1190171 - Revision content denormalization and other cleanups of the compare revision view.

jezdez:
- Adds ratelimiting to the compare revision view.
- Refactors the signal handling to use new patterns and less confusing code.
- Prefetch revision tags and document for fewer queries.
- Denormalize the revision content in a separate field that is filled on save and uses tidy to do the CPU-intense work once instead of on every request.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -47,8 +47,7 @@ from .content import (extract_code_sample, extract_css_classnames,
                       extract_html_attributes, extract_kumascript_macro_names,
                       get_content_sections, get_seo_description, H2TOCFilter,
                       H3TOCFilter, SectionTOCFilter)
-from .jobs import (DocumentZoneStackJob, DocumentZoneURLRemapsJob,
-                   DocumentContributorsJob)
+from .jobs import DocumentZoneStackJob, DocumentContributorsJob
 from .exceptions import (DocumentRenderedContentNotAvailable,
                          DocumentRenderingInProgress, PageMoveError,
                          SlugCollision, UniqueCollision)
@@ -57,6 +56,7 @@ from .managers import (DeletedDocumentManager, DocumentAdminManager,
                        TaggedDocumentManager, TransformManager)
 from .search import WikiDocumentType
 from .signals import render_done
+from .utils import tidy_content
 
 
 def cache_with_field(field_name):
@@ -1613,6 +1613,7 @@ class Revision(models.Model):
 
     summary = models.TextField()  # wiki markup
     content = models.TextField()  # wiki markup
+    tidied_content = models.TextField(blank=True)  # wiki markup tidied up
 
     # Keywords are used mostly to affect search rankings. Moderators may not
     # have the language expertise to translate keywords, so we put them in the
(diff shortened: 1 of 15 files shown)


jezdez:
So this does a bunch of things, which we don't have to do all at once, but I wanted to offer it anyway since I think we shouldn't run tidy on every request of that view but keep the tidied revision content on file instead.

The other changes are probably less troubling.

kuma/wiki/models.py
```diff
@@ -1663,6 +1613,7 @@ class Revision(models.Model):
 
     summary = models.TextField()  # wiki markup
     content = models.TextField()  # wiki markup
+    tidied_content = models.TextField(blank=True)  # wiki markup tidied up
```
groovecoder: Ah, this is a good way to preserve the submitted data (in the `content` field) **and** store the tidied content. But should we populate the `tidied_content` field when it's accessed via `get_tidied_content()` or should we just populate it when the `Revision` is saved?

jezdez: It's populated both, see https://github.com/mozilla/kuma/pull/3386/files#diff-a103da980103930e43dab15a81a2a4f5R128

jezdez: The idea is to not have to go through all the hundred of thousands revisions at once but spread it a bit. Updating the tidied content on access is only useful for old revisions, the post_save hook is for all new revisions.

groovecoder: Doh, sorry I over-looked that; I jumped straight to this `models.py` diff to see what you did here.

kuma/wiki/models.py
```diff
@@ -1805,6 +1756,14 @@ def get_section_content(self, section_id):
         """Convenience method to extract the content for a single section"""
         return self.document.extract_section(self.content, section_id)
 
+    def get_tidied_content(self):
+        if self.tidied_content:
+            return self.tidied_content
+        else:
+            from .tasks import tidy_revision_content
+            tidy_revision_content.delay(self.pk)
+            return tidy_content(self.content)
```
groovecoder: Could a crawler still request a `compare_revisions` view for some old revisions and trigger this CPU-intensive tidying of content with its first request?

kuma/wiki/views.py
```diff
@@ -1512,11 +1512,12 @@ def get_previous(revisions):
 @xframe_options_sameorigin
 @process_document_path
 @prevent_indexing
+@ratelimit(key='user_or_ip', rate='60/m', block=True)
```
groovecoder: :+1: Do we want an even slower limit? `30/m` or even `15/m` seems like the limit that a human could do.

kuma/wiki/views.py
```diff
@@ -1533,8 +1534,9 @@ def compare_revisions(request, document_slug, document_locale):
         # Punt any errors in parameter handling to a 404
         raise Http404
 
-    revision_from = get_object_or_404(Revision, id=from_id, document=doc)
-    revision_to = get_object_or_404(Revision, id=to_id, document=doc)
+    revisions = Revision.objects.prefetch_related('document', 'tags')
```
groovecoder: Nice optimization.

groovecoder:
:heart: using Django 1.7 app system.

kuma/wiki/models.py
jezdez: Yeah, absolutely, this is a compromise to be able to continue to show the diff for revisions without denormalized content now as it's the main feature of the view. It's basically only there for the time in which not all of the revisions have been tidied. Alternatively we could run a full tidying manually via Celery.

kuma/wiki/views.py
jezdez: Works for me, let's go with the most conservative of `15/m`

groovecoder:
Needs a rebase now.

groovecoder:
Intern tests pass. Spot-check on `get_tidied_content()` from a django shell works as expected - empty at first, triggers a celery task, then it's populated next time. Saving a new revision populates `tidied_content` before `get_tidied_content` is called.

kuma/wiki/tasks.py
```diff
@@ -423,10 +416,25 @@ def unindex_documents(ids, index_pk):
...
-
     """
     cls = WikiDocumentType
     es = cls.get_connection('indexing')
     index = Index.objects.get(pk=index_pk)
 
     cls.bulk_delete(ids, es=es, index=index.prefixed_name)
+
+
+@task(rate_limit='60/m')
```
groovecoder: This is still `60/m`?

jezdez:
I actually decided to increase the rate limit of the tidy_revision_content Celery task to `120/s`. The reasoning is simple: we have 550k revisions right now, with a rate limit of `15/m` we'd wait 12 days for even running the tasks. On the other hand, MySQL is currently able to handle `UPDATE` queries very fast and hasn't hit performance ceilings (not even close) with that when I looked at the last week's stats. We also are able to have the Celery nodes run through the CPU intense tidying easier given we can spread it to their collective CPUs, in other words, they aren't fully utilized right now as well.

That said I agree that there is a big risk with scheduling Celery tasks on every call to `get_tidied_content` since if assuming we create a big stack of tasks by normal site browsing it's guarenteed that we'll create duplicates of the tasks. That's why I've added a simple cache based check to the `get_tidied_content` function that prevents duplicate scheduling. It's set to timeout after 3 days, after which a call to `get_tidied_content` would again schedule a task. The calculation for that timeout is: `(546235 / 120) / 60 / 24 = 3.1 days` -- the number of days it takes in theory to run through all revisions.

groovecoder:
Note: It's set to `120/m`, which I assume is correct. (Not `120/s` written in the comment. :wink:)

jezdez:
@groovecoder woops, good catch!

PR #3388, opened 2015-08-05
bug 1170235 - Increase animation speed of compat tables

darkwing:
Increasing animation speed so tables look better.


media/js/wiki-compat-tables.js
```diff
@@ -145,10 +152,9 @@
                 $openCell.addClass('active').attr('aria-expanded', true);
 
                 // add measured height to history and to the cell/row it is being displayed beneath (CSS handles transition)
+                displayDetect = $table.find('thead td:first').first().css('z-index');
 
-                var displayDetect = $('.bc-table thead td').css('z-index');
```
darkwing: I think this was the line that slowed stuff down; this grabs every `table` and every `thead td` cell;  I've changed it to get the first `thead td` in this current table.

stephaniehobson:
The tray is still disappearing instead of sliding up but that's not what you were trying to fix so R+

PR #3389, opened 2015-08-05
Bug 1139490: Improve path-related code in Stylus

openjck:
See the individual commit messages for more information.

### DIFF
--- a/media/stylus/base/fonts/compat-icons.styl
+++ b/media/stylus/base/fonts/compat-icons.styl
@@ -1,10 +1,10 @@
 @font-face {
     font-family: 'compat-icons';
-    src:url('/media/fonts/compat-icons.eot?-ehmwve');
-    src:url('/media/fonts/compat-icons.eot?#iefix-ehmwve') format('embedded-opentype'),
-        url('/media/fonts/compat-icons.woff?-ehmwve') format('woff'),
-        url('/media/fonts/compat-icons.ttf?-ehmwve') format('truetype'),
-        url('/media/fonts/compat-icons.svg?-ehmwve#compat-icons') format('svg');
+    src:url($path-to-assets + 'fonts/compat-icons.eot?-ehmwve');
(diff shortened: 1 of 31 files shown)


darkwing:
So here's what I did to evaluate this:

1.  On master I ran:  ./scripts/compile-stylesheets && ./manage.py compress_assets
2.  On your branch I ran:  ./scripts/compile-stylesheets && ./manage.py compress_assets
3.  I see the following:

```
Minifying /home/vagrant/src/media/css/mdn-all.css (using clean-css)
Minifying /home/vagrant/src/media/css/demostudio-all.css (using clean-css)
Minifying /home/vagrant/src/media/css/gaia-all.css (using clean-css)
Unchanged files skipped for minification: 45
```

In theory there should be no changes, right?

openjck:
The `compile-stylesheets` script assumes that there was only one commit. It won't work here because there are four commits. As we move to django-pipeline, I'm going to need to either update the script or remove it.

openjck:
Nice catch. Some of the variables were leaking into the compiled CSS.

Updated! There are still some differences in `main.css` and `demos.css` but those are only rules that Stylus decided to move around, probably because of the new `@requires`.

darkwing:
Still seeing:

```
src:url("//mozorg.cdn.mozilla.net$path-to-assetsfonts/OpenSans
```

openjck:
Hm, I'm not seeing that on my end. Be sure to compile the Stylus in the VM. Try these steps:

1. On the host machine, run `rm -rf css`
2. On the host machine, run `git co -- css`
3. In the virtual machine, run `compile-stylesheets`
4. Search for the term *path-to-assets* in the **css** directory

openjck:
Turns out I hadn't pushed the latest code up. Fixed now.

darkwing:
Any reason not to squash all of this?

openjck:
We could. I tend to like keeping separate changes in separate commits so that we can individually reverse/bisect/etc. them later.

darkwing:
Everything looking good to me!  R+

PR #3390, opened 2015-08-06
Bug 1180208 - Post user profile cleanup.

jezdez:


kuma/users/apps.py
```diff
@@ -0,0 +1,110 @@
...
+                                  dispatch_uid='users.user.post_save')
+        user_signed_up.connect(self.on_user_signed_up)
+        email_confirmed.connect(self.on_email_confirmed)
+        social_account_removed.connect(self.on_social_account_removed)
+
+        # the user ban signal handlers to connect to
+        UserBan = self.get_model('UserBan')
+        signals.post_delete.connect(self.on_ban_delete,
+                                    sender=UserBan,
+                                    dispatch_uid='users.user_ban.delete')
```
groovecoder: Hmm ... I can't remember why we re-activate the user with a `UserBan.post_delete` signal handler, but we de-activate user a user in a custom `UserBan.save` method?

groovecoder:
Branch migrations worked on my vm that had just been migrated to move profile data over to user models. So this looks good except the 1 (small) question, and needs a rebase.

Also, we should probably push https://github.com/mozilla/kuma/pull/3386 separately from this merge & push since both of them include migrations that could be tricky in stage and/or prod.

kuma/users/apps.py
jezdez: I don't quite remember but you're right, I just moved this into the apps module as a signal callback, too.

groovecoder:
Needs updating now that another PR has already used up the wiki migration 0006.

jezdez:
Updated and rebased.

groovecoder:
I ran all migrations on master, then switched to this branch to run migrations and I got:
```
Your models have changes that are not yet reflected in a migration, and so won't be applied.
Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

jezdez:
huh, hmm, can you run ./manage.py makemigrations wiki to see what changed?
could be updated locales since the list of locales is used in the locale storing fields and Django’s migration system freezes all parameters in migrations

groovecoder:
It was this: https://gist.github.com/groovecoder/e590fa8152dd1f0bd6a9

Not sure how that change only registers now?

jwhitlock:
It looks like a quirk of ``django.core.files.storage.FileSystemStorage``, that the local base path shows up in the field definition: ``location=b'/home/vagrant/uploads/demos'``.  We could create the production version of this path in the vagrant box, and set the uploads to that folder, so that ``makemigrations`` will be happier, but feels like a [bug 957802](https://bugzilla.mozilla.org/show_bug.cgi?id=957802) task, out of scope for this.

groovecoder:
Alright the good news is that the `users` migrations work fine, and all of the selenium tests pass - existing user is able to sign in, edit, etc.

So this is good to merge & push. :clap:

groovecoder:
AIUI, we're still waiting on staging celery & MySQL fixes to push this?

jezdez:
@groovecoder Yeah, waiting on stage celery/db issues to clear first.

PR #3393, opened 2015-08-10
Bug 957802 - Make sure the language mapping isn't exectured until Django is ready.

jezdez:


darkwing:
I can confirm that I was able to get my server running and avoid provisioning errors with this PR

PR #3396, opened 2015-08-10
Bug 1190284 - Whitespace at the bottom of languages drop down menu

stephaniehobson:
Removed min-height from submenus.


darkwing:
R+!  Nice!

PR #3398, opened 2015-08-10
fix bug 1190582 - add Malagasy locale

groovecoder:
Spot-check on this branch:
1. [Get the localizations](http://kuma.readthedocs.org/en/latest/localization.html#get-localizations) if you don't already have them.
2.
```
cd locale
svn up
./compile-mo.sh .
```
3. Go to https://developer-local.allizom.org/mg/ and make sure it pulls up.
4. Translate a test document into Malagasy.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -125,6 +125,7 @@ MDN_LANGUAGES = (
     'ka',
     'ko',
     'ln',
+    'mg',
     'ml',
     'ms',
     'my',


openjck:
Can you say more about the changes to the tar-gzip files?

openjck:
Test failures look legitimate.

groovecoder:
[Travis uses them to create the product details languages.json and the `locale/` directory needed for languages.](https://github.com/mozilla/kuma/blob/master/scripts/travis-install#L13-L15) I updated them [per our docs](http://kuma.readthedocs.org/en/latest/localization.html#adding-a-new-locale) to make sure that Travis has all the latest languages. They aren't technically needed, since Malagasy has been in `languages.json` for a while.

Test failure is legit; need to dig into it. :cry:

jezdez:
@groovecoder It's too bad we have to maintain an own `kuma/languages.json` instead of being able to use the one offered by the product-details app. It seems the old reason we need that at all is a circular import issue when needing to use the `languages.json` to populate a setting. It provides a FAQ for that even, https://github.com/mozilla/django-product-details#caveats--known-issues (https://github.com/mozilla/django-product-details/commit/534845270f73a69ec480e6b6664e7ee97ec49662) Should we try that again? Since we use the `LOCALES` setting to build the `LANGUAGES` setting, we would need to make the creation of `LANGUAGES` also lazy.

groovecoder:
ZOMG Travis is finally green! Quick merge merge merge!
:shipit:

jezdez:
GO GO GO!!

PR #3400, opened 2015-08-10
bug 1186091 - performance smoke test with locust

groovecoder:


docs/tests-load.rst
```diff
@@ -0,0 +1,38 @@
...
+against other environments.
+
+.. note:: **DO NOT RUN LOCUST TESTS AGAINST PRODUCTION**
+
+Installing
+----------
+
+See `Installation Docs for Locust
+<http://docs.locust.io/en/latest/installation.html>`_. (Notice there
+are some platform-specific instructions.)
```
jezdez: We should install Locust in the VM to have a common base to run it from

docs/tests-load.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Running Tests
+-------------
+
+.. note:: **DO NOT RUN LOCUST TESTS AGAINST PRODUCTION**
+
+1. Start locust with one of our test files and our target host::
+
+    locust -f tests/load/smoke.py --host=https://developer.allizom.org
```
jezdez: I think I'd like to wrap that in a Makefile and hardcode the host so that we don't accidentally use prod.

tests/load/smoke.py
```diff
@@ -0,0 +1,54 @@
...
+
+class SmokeBehavior(TaskSet):
+    """
+    Request behavior that represents some standard production behavior.
+    Tasks uris are hard-coded for now, but could be stored or even dynamically
+    generated later.
+    See https://bugzil.la/1186085
+    """
+
+    @task(265)
```
jezdez: what are those numbers in the @task decorators?

docs/tests-load.rst
groovecoder: Good point. :+1:

docs/tests-load.rst
groovecoder: :+1:

tests/load/smoke.py
groovecoder: They are weights for task distribution during the test. These values are based on sampled traffic so this test sends requests proportionate to a snapshot of traffic levels on production. (See https://bugzilla.mozilla.org/show_bug.cgi?id=1186085#c1)

jezdez: Ah, cool, thanks! Could you make the parameter passed as `@task(weight=265)` so it's easier to understand what the number means?

groovecoder:
Updated.

requirements/test.txt
```diff
@@ -0,0 +1 @@
+locustio==0.7.3
```
jezdez: Let's call this file `performance.txt` or something so we have a distinction between the "roles" of requirements files, not use ambiguous terms (e.g. vs. python tests, intern tests etc).

jezdez:
Cool! wc+

Makefile
```diff
@@ -0,0 +1,5 @@
+load-smoke-test:
+	locust -f tests/load/smoke.py --host=https://developer.allizom.org
+
+browser-tests:
+	cd tests/ui ; ./node_modules/.bin/intern-runner config=intern-local d=developer.allizom.org b=firefox
```
jezdez: nice!

Makefile
```diff
@@ -0,0 +1,5 @@
+load-smoke-test:
```
jezdez: I think having consistent make task names would be cool here, e.g. `make smoke-tests` and `make browser-tests`. Maybe even `make django-tests`?

Makefile
jezdez: Ah, one thing, could you cd back to the main directory? Or use [pushd/popd](https://en.wikipedia.org/wiki/Pushd_and_popd)?

groovecoder:
Updated again.

PR #3401, opened 2015-08-11
Fix bug 1193315 - Prevent 404 on grabbing image

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1193315


stephaniehobson:
R+

PR #3405, opened 2015-08-11
fix bug 1193455 - Include font observer in main package

darkwing:
Since this is no longer waffled, we don't need to load the item via AJAX.  Save a request.


stephaniehobson:
Needs a rebase :(

stephaniehobson:
Works as expected. R+ after rebase.

darkwing:
Updated!

PR #3407, opened 2015-08-12
Bug 1187645 - Refactored check for invalid slug characters.

jezdez:
That includes adding the percentage sign to it.

### DIFF
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -1303,23 +1303,20 @@ class DocumentEditingTests(UserTestCase, WikiTestCase):
                              reverse('wiki.document', args=[data['slug']],
                                      locale=settings.WIKI_DEFAULT_LANGUAGE))
 
-        # Slashes should not be acceptable via form input
-        data['title'] = 'valid with slash'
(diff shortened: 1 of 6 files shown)


groovecoder:
Code looks good. The bug seems to really want a regression test to make sure we don't re-introduce this later?

jezdez:
Agreed! Test forthcoming

jezdez:
@groovecoder Done

kuma/wiki/tests/test_views.py
```diff
@@ -1361,9 +1358,15 @@ def _createAndRunTests(slug):
             locale = settings.WIKI_DEFAULT_LANGUAGE
             foreign_locale = 'es'
             new_doc_url = reverse('wiki.new_document')
-            invalid_slug = invalid_slug1 = "some/thing"
-            invalid_slug2 = "some?thing"
-            invalid_slug3 = "some thing"
+            invalid_slug = "some/thing"
```
groovecoder: Do we still need this `invalid_slug` varaible? Or can we use `invalid_slugs[0]` below?

jezdez: Ah, hm, not sure

PR #3408, opened 2015-08-12
Fix bug 1193815 - Fixed rendering of JavaScript translation catalogs.

jezdez:
This makes use of the "javascript" domain name used by the non-standard i18n setup Mozilla uses.
Also move to using a cacheback job to profit from thundering herd protection etc.

### DIFF
--- a/docs/localization.rst
+++ b/docs/localization.rst
@@ -263,6 +263,10 @@ you need them for something, they're pretty easy to get::
 
 if you're a git fan.)
 
+Then run the Django management command to update the static JavaScript
+translation catalogs::
+
+    python manage.py compilejsi18n
 
 Updating the Localizations
 ==========================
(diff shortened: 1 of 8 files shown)


darkwing:
` you need to run the compilejsi18n management command to build the js files statically`

darkwing:
R+ from me in that it's working properly

@groovecoder :  Want to do python code review?

darkwing:
Great work @jezdez !

darkwing:
Test failures appear legit.

jezdez:
Seems like some view didn't have access to the request's locale, just added a condition to use the default langauge (en-US) as a fallback.

groovecoder:
Looks good. And `compilejsi18n` is already in our deployment script, so this should work out of the box! Very nice!

(Note: For some reason, many lanugages' translations are commented out in their `javascript.po` file. [e.g., German](http://viewvc.svn.mozilla.org/vc/projects/mdn/trunk/locale/de/LC_MESSAGES/javascript.po?view=markup) So while this is a good fix for removing the extra request for the js catalog, the catalog is still empty for many languages. :cry: )

PR #3411, opened 2015-08-13
Make newsletter and users resources async

darkwing:
A few more resources which are standalone and can be async.

### DIFF
--- a/kuma/search/templates/search/base.html
+++ b/kuma/search/templates/search/base.html
@@ -4,9 +4,6 @@
 {% if not styles %}
   {% set styles = ('search',) %}
 {% endif %}
-{% if not scripts %}
-  {% set scripts = ('search',) %}
-{% endif %}
 {% set hide_header_search = True %}
 
 {% block site_css %}
@@ -15,4 +12,5 @@
 
 {% block site_js %}
     {{ super() }}
+    {{ js('search', async=True) }}
 {% endblock %}
--- a/kuma/wiki/templates/wiki/base.html
+++ b/kuma/wiki/templates/wiki/base.html
@@ -10,6 +10,7 @@
     {% endif %}
 {% endblock %}
 
-{% if not scripts %}
-  {% set scripts = ('wiki',) %}
-{% endif %}
+{% block site_js %}
+    {{ super() }}
+    {{ js('wiki', async=True) }}
+{% endblock %}
--- a/kuma/wiki/templates/wiki/document.html
+++ b/kuma/wiki/templates/wiki/document.html
@@ -42,7 +42,7 @@
 {% endif %}
 
 {% block bodyclass %}document {% if is_zone %}zone{% endif %} {% if is_zone_root %}zone-landing{% endif %}{% endblock %}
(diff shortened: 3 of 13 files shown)


kuma/wiki/templates/wiki/document.html
```diff
@@ -42,7 +42,7 @@
 {% endif %}
 
 {% block bodyclass %}document {% if is_zone %}zone{% endif %} {% if is_zone_root %}zone-landing{% endif %}{% endblock %}
-{% block body_attributes %}data-slug="{{ document.slug }}" contextmenu="edit-history-menu"{% endblock body_attributes %}
+{% block body_attributes %}data-slug="{{ document.slug }}" contextmenu="edit-history-menu" data-search-url="{{ search_url|safe }}"{% endblock body_attributes %}
```
darkwing: This should be safe, right @jezdez ?

jezdez: I think so, yeah

kuma/search/templates/search/base.html
```diff
@@ -4,9 +4,6 @@
 {% if not styles %}
   {% set styles = ('search',) %}
 {% endif %}
-{% if not scripts %}
-  {% set scripts = ('search',) %}
-{% endif %}
```
openjck: Interesting. Could you say more about this? I'm wondering why we checked for the non-existence of `scripts` in the first place.

kuma/wiki/templates/wiki/base.html
```diff
@@ -10,6 +10,7 @@
     {% endif %}
 {% endblock %}
 
-{% if not scripts %}
-  {% set scripts = ('wiki',) %}
-{% endif %}
```
openjck: Wondering about the same thing here. Was there a reason we were checking for the non-existence of `scripts`?

openjck:
Otherwise this looks great! Nice work.

kuma/search/templates/search/base.html
groovecoder: Technically, this would allow each child of `search/base.html` to declare a custom set of `scripts` for itself. But if a child template did not declare its own `scripts`, this `base` template would include the `search` script.

kuma/wiki/templates/wiki/base.html
groovecoder: Same [note as above re: search](https://github.com/mozilla/kuma/pull/3411/files#r37025763).

darkwing:
@groovecoder Do your comments endorse this or no?

groovecoder:
Since none of the templates that extend from these base templates override the `scripts`, it should be fine to merge this. Make sure to run selenium tests and check front-end functionality of wiki and search. :smile:

darkwing:
Tests worked for me in Chrome and Firefox.  Did you see different?

openjck:
Passed for me in Firefox. I think @stephaniehobson said she wanted a chance to review, too.

stephaniehobson:
A already said it was good to go on IRC :)

openjck:
> I already said it was good to go on IRC :)

Ah, awesome. I missed that. R+ from me then.

PR #3397, opened 2015-08-10
bug 1145225 - Optimize images with TinyPng

darkwing:
Used TinyPNG to crunch the hell out of our core images.  Saved MBs.  Excellent.

PR #3414, opened 2015-08-13
bug 1194461 - Change tabzilla images to data URIs

darkwing:
Hardcoding an image used on every page eliminates a request from every page!


groovecoder:
If we do this and it's the only place we reference those images, we should remove them from the code-base too.

darkwing:
Don't agree with that, tbh, but I'm just getting started; I may make stylus do something fancy with this.

groovecoder:
Sorry, yeah, that should have been a question/suggestion. :smile:

darkwing:
We should wait on this, actually:  https://github.com/mozilla/kuma/pull/3397

darkwing:
Closing, more coming after the PR 3397

PR #3415, opened 2015-08-13
fix bug 1191348 - Anonymize harder

jwhitlock:
scripts/anonymize.py updated with current table list, minus stale tables slated to be dropped from production.  It is now noisier on errors.

Many changes to scripts/anonymize.sql:
- Drops stale tables
- Truncates more tables, include bans, social accounts, constance
  config, currently empty tables.
- Legacy profile fields (first name, last name, homepage) are blanked
- Anonymizes user profile fields if populated, otherwise leaves them
  blank.
- Drops hidden and censored demos
- Converts revision IPs to 192.168.x.x private IP space

My test, in a vagrant box with a production database dump:

```
(env)vagrant@developer-local:~/src$ time scripts/anonymize.py kuma -u kuma -p kuma
Dumping input DB to kuma-20150813.sql.gz
Warning: Using a password on the command line interface can be insecure.
Creating temporary DB kuma_anontmp_20150813
Warning: Using a password on the command line interface can be insecure.
Warning: Using a password on the command line interface can be insecure.
Importing the input dump into the temporary DB
Warning: Using a password on the command line interface can be insecure.
Applying scripts/anonymize.sql to the temporary DB
Warning: Using a password on the command line interface can be insecure.
Dumping temporary DB to kuma-anon-20150813.sql.gz
Warning: Using a password on the command line interface can be insecure.
Dropping temporary db kuma_anontmp_20150813
Warning: Using a password on the command line interface can be insecure.
Database "kuma_anontmp_20150813" dropped
Deleting input DB dump kuma-20150813.sql.gz

real	12m19.172s
user	5m11.216s
sys	2m37.164s
(env)vagrant@developer-local:~/src$ gunzip kuma-anon-20150813.sql.gz 
(env)vagrant@developer-local:~/src$ mysql -u kuma -p kuma
...
mysql> source kuma-anon-20150813.sql
...
mysql> ^DBye
(env)vagrant@developer-local:~/src$ ./manage.py collectstatic --noinput
(env)vagrant@developer-local:~/src$ ./manage.py compilejsi18n
(env)vagrant@developer-local:~/src$ foreman start
```

Here's a anonymized profile:
anon sheppy

Because I wiped out the social accounts, login will be difficult.  However, there's too much PII in there, so I don't think we can keep it.


groovecoder:
I'm not sure why we even have a `skip_anonymize` option for `anonymize.py`? We should remove that so nobody can run `anonymize.py` and get anything except an anonymized DB.

scripts/anonymize.sql
```diff
@@ -41,27 +90,102 @@ TRUNCATE djcelery_periodictask;
...
+    WHERE mozillians_url != "";
+UPDATE auth_user SET
+    stackoverflow_url = CONCAT("https://stackoverflow.com/users/1/", MD5(CONCAT(stackoverflow_url, @common_hash_secret)))
+    WHERE stackoverflow_url != "";
+UPDATE auth_user SET
+    twitter_url = CONCAT("https://twitter.com/", MD5(CONCAT(twitter_url, @common_hash_secret)))
+    WHERE twitter_url != "";
+UPDATE auth_user SET
+    website_url = CONCAT("https://example.com/", MD5(CONCAT(website_url, @common_hash_secret)))
+    WHERE website_url != "";
```
groovecoder: What do we gain by doing all these `UPDATE auth_user` statements separately instead of one big one?

scripts/anonymize.sql
```diff
@@ -41,27 +90,102 @@ TRUNCATE djcelery_periodictask;
...
 
-UPDATE account_emailaddress SET
-    email = CONCAT(MD5(CONCAT(email, @common_hash_secret)), '@example.com');
+DELETE FROM demos_submission WHERE hidden=1;
+DELETE FROM demos_submission WHERE censored=1;
```
groovecoder: :+1:

groovecoder:
IIRC, we only use `anonymize.sql` for anonymization? Did we decide not to remove the `anonymize.py` script altogether? If we keep it, we should add canary records before the anonymization, and then test for the canary records at the end, and make sure DBA always use the script and not just the `.sql`.

Doing spot-check ...

scripts/anonymize.sql
jwhitlock: If the user did not set a URL, then the anonymized user also doesn't have an URL.  I'm not sure how to do that in 1 line of SQL.

jwhitlock:
``anonymize.py`` could be renamed ``clone_db.py``, and more of the options would make sense.

scripts/anonymize.py
```diff
@@ -56,34 +63,40 @@
...
     tagging_tag
     tagging_taggeditem
     taggit_tag
     taggit_taggeditem
-    threadedcomments_freethreadedcomment
-    threadedcomments_testmodel
-    threadedcomments_threadedcomment
     tidings_watch
+    tidings_watchfilter
     user_profiles
```
groovecoder: On latest db dump, migrated up thru https://github.com/mozilla/kuma/pull/3390 ...
```
Dumping input DB to kuma-20150818.sql.gz
Warning: Using a password on the command line interface can be insecure.
mysqldump: Couldn't find table: "user_profiles"
```

scripts/anonymize.sql
```diff
@@ -27,13 +27,62 @@ SET @common_hash_secret := rand();
...
+-- wiki_editortoolbar
+-- wiki_helpfulvote
+-- wiki_localizationtag
+-- wiki_localizationtaggedrevision
+-- wiki_reviewtag
+-- wiki_reviewtaggedrevision
+-- wiki_revision
+-- wiki_taggeddocument
+
 TRUNCATE account_emailconfirmation;
```
groovecoder: On latest DB dump migrated up thru #3390 ...
```
Applying scripts/anonymize.sql to the temporary DB
Warning: Using a password on the command line interface can be insecure.
ERROR 1146 (42S02) at line 73: Table 'kuma_anontmp_20150818.account_emailconfirmation' doesn't exist
```

scripts/anonymize.sql
groovecoder: Ah, I see. This preserves blank fields as blank - i.e., it doesn't put an md5 hash into the field if it was blank.

groovecoder:
:+1: change `anonymize.py` to `clone_db.py`, but do we even need to keep `anonymize.py`? If we want to, let's make sure we communicate with DBA that they should use this script. If they're not going to use the `.py` script, let's not try to maintain it or create the situation where we think something is covered in this `.py` that isn't actually being used.

scripts/anonymize.py
jwhitlock: Removing from anonymize.py

scripts/anonymize.sql
jwhitlock: This is interesting - EmailConfirmation is part of django-allauth, and is present in my production DB dump.  Maybe the user_profiles step caused it to fail.

jwhitlock:
I've fixed the ``user_profiles`` issue, renamed to ``clone_db.py``, and made it exit early and loudly on errors.

It doesn't have much more functionality than a bash script, but it's way easier than running the commands manually.  If DBA uses a different process, then we can drop it to remove confusion, but I'd rather sync on the same process.

groovecoder:
Updates look good. Should we merge now, or wait for staging issues to clear up?

jwhitlock:
Let's wait - I think we need to concentrate on fixing staging.

groovecoder:
Staging fixed. Prod pushed.

PR #3418, opened 2015-08-17
Fix Bug 1185769 - Feedback menu item names

stephaniehobson:


darkwing:
Shouldn't we be localizing these strings?

stephaniehobson:
Yup. Don't know how I missed that. *face, palm*

stephaniehobson:
Updated.

openjck:
Nice!

PR #3419, opened 2015-08-17
Bug 1195335 - Split wiki views into separate modules.

jezdez:
This includes some cleanups and code deletions.


jwhitlock:
Done with the code review, still need to do live testing. LOTS of views, so it will probably be another day.

This would have been easier to review as small, checkable commits.  Something like:

* Rename url wiki.get_children to wiki.children
* Rename url wiki.subscribe_document to wiki.subscribe
* Drop unused ui_locale param from Document.get_absolute_url
* Move kuma/wiki/views.py to kuma/wiki/views/__init__.py
* Move views to views/delete...
* Remove unused view delete_revision
* etc. etc.

Or, if that is too detailed, at least
* Move views
* Move templates
* Cleanup code

With the move + code cleanup squashed into a single commit, I had to check out
views.py and visually confirm that each view got moved and that it
roughly does the same thing.  Too late for this one, but maybe next time.

As for the actual code changes,
* :+1: better organization of views
* :+1: more consistent doc strings
* :+1: modernize redirects and other Django updates
* :+1: dropping some dead code and 1-liner util functions

There's additional cleanup I'd want to do, but this commit is already ambitious enough.

jwhitlock:
I believe the decision on IRC was to split this into multiple commits in a new branch and pull request.

jezdez:
Superceded by #3431

PR #3421, opened 2015-08-18
Bug 957802 - Fix name of make target for load tests.

jezdez:
Also made make tasks phony and added a clean task.


jezdez:
@groovecoder Just a few minor updates for the new load tester.

Procfile
```diff
@@ -1,4 +1,4 @@
-web: find kuma -name '*.pyc' -exec rm {} \; && python2.7 manage.py runserver 0.0.0.0:8000
+web: make clean && python2.7 manage.py runserver 0.0.0.0:8000
```
groovecoder: Good call. Can we change `docs/troublshooting.rst` to match this too?

Makefile
```diff
@@ -2,9 +2,15 @@
 django-tests:
 	python manage.py test
 
-performance-tests:
+load-tests:
```
groovecoder: nit-picky, but I've heard "load tests" imply tests on a system to see how much load it can take before it breaks. While this test is more like a performance regression test?

Procfile
jezdez: Good idea!

Makefile
jezdez: Ah, fair point, let's update the docs then, too.

Makefile
```diff
@@ -2,9 +2,15 @@
...
 	locust -f tests/load/smoke.py --host=https://developer.allizom.org
 
 # Note: this target should be run from the host machine with selenium running
 browser-tests:
 	pushd tests/ui ; ./node_modules/.bin/intern-runner config=intern-local d=developer.allizom.org b=firefox; popd
+
+clean:
+	find kuma -name '*.pyc' -exec rm {} \;
```
jwhitlock: :+1: for a ``make clean``.  I'll add the "really clean *.pyc except for that one python-magic file" later.

jezdez: :sob:

Procfile
jezdez: Updated it with "performance"

groovecoder:
Looks good. I kicked Travis but at this point I'm ready to merge on the failures because they're all Travis connectivity issues.

jezdez:
@groovecoder Yeah, I think that's fine, it doesn't touch files that the tests cover

PR #3422, opened 2015-08-18
bug 1194482 - Live sample don't send code to service if code comes from another page

JeremiePat:
The `EmbedLiveSample` allows to embed a sample that comes from a remote page. This patch make sure that in such case the proper code is retrieve from the right page before sending it.

### DIFF
--- a/media/js/wiki-samples.js
+++ b/media/js/wiki-samples.js
@@ -1,4 +1,6 @@
 (function(win, doc, $) {
+    "use strict";
+
     if(!win.waffle || !win.waffle.flag_is_active('wiki_samples')) return;
 
     var sites = ['codepen', 'jsfiddle'];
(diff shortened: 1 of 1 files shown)


darkwing:
Can you put together a page on staging that has a variety of code sample formats so we can test this out?

JeremiePat:
Sir, yes sir.
Here it is:
* https://developer.allizom.org/en-US/docs/User:Jeremie/Live_Sample_Tests (same page call)
* https://developer.allizom.org/en-US/docs/User:Jeremie/Live_Sample_Remote_Call (remote call)

JeremiePat:
Arg! I screw up my branch with unwanted merge. I'll fix that tomorrow.

darkwing:
Checking this out

darkwing:
@JeremiePat The first test looks great but it looks the the second one got nuked from staging :(  Can you provide me the content or re-add the doc?  Sorry :(

JeremiePat:
Done :)

darkwing:
Works for me!  Thank you @JeremiePat !

PR #3423, opened 2015-08-18
Bug 1187967 - Extract code samples via cacheback.

jezdez:


jwhitlock:
I got a failure when testing my copy of this document:

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/How_to_build_custom_form_widgets?raw

It has this macro:
 ``{{EmbedLiveSample("Basic_state",120,130, "", "HTML/Forms/How_to_build_custom_form_widgets/Example_1")}}``

When I run kuma.wiki.tasks.refresh_code_samples against the document ID, I get this traceback:

```
>>> from kuma.wiki.tasks import refresh_code_samples
>>> refresh_code_samples(58349)
  File "<console>", line 1, in <module>
  File "/home/vagrant/src/vendor/src/celery/celery/app/task.py", line 420, in __call__
    return self.run(*args, **kwargs)
  File "/home/vagrant/src/kuma/wiki/tasks.py", line 58, in refresh_code_samples
    macro_parameters = document.extract.macro_parameters(macro)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/newrelic-2.52.0.40/newrelic/api/function_trace.py", line 104, in literal_wrapper
  File "/home/vagrant/src/kuma/wiki/content.py", line 93, in macro_parameters
    for part in match.split(',')])
  File "/home/vagrant/src/vendor/src/django/django/utils/functional.py", line 214, in wrapper
    return func(*args, **kwargs)
  File "/home/vagrant/src/vendor/src/django/django/utils/text.py", line 430, in unescape_string_literal
    raise ValueError("Not a string literal: %r" % s)
ValueError: Not a string literal: u'120'
```

``unescape_string_literal`` is expecting something like ``'"abc"'``, and can't handle an 'integer' string like ``120``.

jwhitlock:
Also, ``{{EmbedLiveSample}}` can take an external URL now:

https://github.com/mozilla/kuma/pull/3422

jezdez:
@jwhitlock Ough, yeah, same as `EmbedDistLiveSample`, guess we have to keep extracting the code samples for them on every request if that parameter is given.

jezdez:
I scratched using `unescape_string_literal` as it wasn't doing what I thought it would.

jwhitlock:
I think the TravisCI failed tests are real errors (not out-of-memory errors).  Also, there might be a redesign as noted in a [previous commit](https://github.com/mozilla/kuma/commit/3ae196ee94ff54e35dfbda9b69d31cf49aac8088#commitcomment-12832562):

> Hm, that makes me wonder if the mode of operation makes sense to be honest.

> Right now we extract all code samples when a document is saved, while instead we could treat the whole thing as lazy (since we already accept the fact that we won't cover all code samples given the remote source ability of code sample macros). Instead we could just have the view trigger a Celery task with the specific code sample name and have the table be populate over lazily. That would ease the roll out of this feature as well lower the impact of saving a document since it wouldn't have to extract the code sample on save time but only asynchronously on request time. We would have to delete all code samples of a document when it's saved though. Hm, not sure.

jezdez:
@jwhitlock I've reconsidered and reverted the database denorm part in favor of a cacheback based solution. It uses a generation based caching algorithm (LOL) to invalidate all code samples of a document on save.

groovecoder:
Rebased this on master to get Travis to go green ... or at least not fail on ES memory/connection errors.

kuma/wiki/jobs.py
```diff
@@ -100,3 +100,17 @@ def fetch(self, pk):
     def empty(self):
         # the empty result needs to be an empty list instead of None
         return []
+
+
+class DocumentCodeSampleJob(GenerationJob):
+    lifetime = 60 * 60 * 12
+    refresh_timeout = 60
+    generation_based = True
```
jwhitlock: What does ``generation_based`` do?

kuma/wiki/content.py
```diff
@@ -50,6 +49,117 @@
...
+        highlighting already does that with <pre>'s
+
+        Given the name of a code sample, attempt to extract it from rendered
+        HTML with a fallback to non-rendered in case of errors.
+        """
+        parts = ('html', 'css', 'js')
+        data = dict((x, None) for x in parts)
+
+        try:
+            src, errors = self.document.get_rendered()
```
jwhitlock: {{EmbedLiveSample}}, as used on [MDN/Contribute/Howto/Create_an_interactive_exercise_to_help_learning_the_web
](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Howto/Create_an_interactive_exercise_to_help_learning_the_web), uses a live sample on the same page.   So, to render that page, we need to load the code samples from that page, which requires rendering that page, which requires loading the code sample from that page...

It obviously works (I can view the page on the site), I'm just not sure how.  I think I need to trace through the code and see what's happening.

Also, I'm curious if anything is lost by using the raw version of a page rather than the rendered version.

kuma/wiki/apps.py
```diff
@@ -94,11 +94,17 @@ def on_document_save(self, sender, instance, **kwargs):
...
+        # cleanup document zones
         async = kwargs.get('async', True)
         invalidate_zone_urls_cache(instance, async=async)
         invalidate_zone_stack_cache(instance, async=async)
+
+        # refresh the document's contributor bar data
         DocumentContributorsJob().invalidate(instance.pk)
 
+        # new generation of document's code samples
+        DocumentCodeSampleJob(generation_args=[instance.pk]).renew()
```
jwhitlock: This should be added to the method docstring, or the docstring should be less wordy. Also, are you worried about creating a long-lived generation key for every page, even ones without code samples?

jwhitlock:
I was hoping to have the code review finished by your morning, but there is more to go for tomorrow.

[Template:LiveSampleURL](https://developer.mozilla.org/en-US/docs/Template:LiveSampleURL) includes a revision ID, which appears unused in ``code_sample``.  Should a particular revision be looked up?  Or is this wishful thinking in ``LiveSampleURL`` that should be removed?

kuma/core/jobs.py
```diff
@@ -37,6 +39,46 @@ def hash(self, value):
...
+    @cached_property
+    def generation_key(self):
+        generation_args = ':'.join([str(key) for key in self.generation_args])
+        return '%s:%s:generation' % (self.class_path, generation_args)
+
+    def key(self, *args, **kwargs):
+        key = super(GenerationJob, self).key(*args, **kwargs)
+        return '%s@%s' % (key, self.generation())
+
+    def generation(self):
```
jwhitlock: I suspect you are implementing a subset of cacheback inside of a cacheback job.  I don't know whether to be impressed or frightened.

kuma/wiki/jobs.py
jezdez: Just a leftover from when this was part of the `KumaJob` class.

kuma/wiki/content.py
jezdez: I'd like to fix the code sample system at some point, but this is just a copy of the current code sample extraction. Let's investigate this at a later state.

kuma/wiki/apps.py
jezdez: ok

kuma/core/jobs.py
jezdez: Not really a subset, but an addition, which is why I wanted to put this into a subclass. This still uses the main features of cacheback but simply adds another invalidation scheme via a cache key generation additionally to the default one-cache-key-per-job-parameters technique.

jezdez:
@jwhitlock Fixed nitbits

kuma/core/jobs.py
jwhitlock: What I mean is that you could use a cacheback job to create and maintain the random generation key, instead of direct cache manipulation.  Something like:

```python
class GenerationJob(KumaJob):
    generation_age = 60 * 60 * 24 * 365

        generation_args = kwargs.pop('generation_args', [])
        super(KumaJob, self).__init__(*args, **kwargs)
        self.generation_key = GenerationKeyJob(
            age=self.generation_age, for_class=self.class_path,
            generation_args=generation_args)

    def key(self, *args, **kwargs):
        key = super(GenerationJob, self).key(*args, **kwargs)
        return '%s@%s' % (key, self.generation_key.get())

    def invalidate_generation(self):
        self.generation_key.delete()


class GenerationKeyJob(Job):
    def __init__(self, age, for_class, generation_args, *args, **kwargs):
        self.age = age
        self.for_class = for_class
        self.generation_args = generation_args

    def key(self, *args, **kwargs):
        generation_args = ':'.join([str(key) for key in self.generation_args])
        return '%s:%s:generation' % (self.for_class, generation_args)

    def fetch(self, *args, **kwargs):
        return crypto.get_random_string(length=12)
```

This isn't cut and paste code, but it only requires understanding cacheback jobs, not Django cache interfaces and other bits.

kuma/wiki/content.py
jwhitlock: I'm wrong about ``{{EmbedLiveSample}}``.  It adds an iframe whose source is ``$samples/(sample_name)``, which renders the HTML document with the CSS, JS, and body HTML.  I still think there's latent bugs in here (need to test what happens when rendered_html is an empty string), but it should match current behavior.

kuma/wiki/apps.py
```diff
@@ -93,12 +93,16 @@ def on_document_save(self, sender, instance, **kwargs):
...
           document
+        - trigger the renewal of the code sample job generation
         """
         async = kwargs.get('async', True)
         invalidate_zone_urls_cache(instance, async=async)
         invalidate_zone_stack_cache(instance, async=async)
+
         DocumentContributorsJob().invalidate(instance.pk)
 
+        DocumentCodeSampleJob(generation_args=[instance.pk]).renew()
```
jwhitlock: *(moving from commit-level comment)*

This will cause a new cryptographically strong random number to be generated, and stored in memcached for a very long time, on **every** page save, not just those with code sample source.  Instead, it should be something like:

```
cache.delete(DocumentCodeSampleJob(generation_args=[instance.pk]).generation_key)
```
which should probably be a helper method on ``GenerationJob``.  This will delete any existing generation key, causing invalidation and ``.renew()`` on use.

jwhitlock:
A couple things left:
* The optional change of ``GenerationJob`` to two cacheback job classes
* Delete the generation key on document save, not ``.renew()``

robhudson:
@jezdez @jwhitlock Is there anything I can help with here?

jwhitlock:
@robhudson I think @jezdez has the next step, but I'm not sure what happens when a code reviewer suggests big changes.

jezdez:
@jwhitlock I'm working on spam stuff right now and will pick this up again when I'm done

jwhitlock:
Rebased in new PR #3810.

PR #3427, opened 2015-08-19
bug 1055078 - Fellowship and auth cleanup

darkwing:
Just some cleanup I thought I could do.

### DIFF
--- a/media/js/fellowship.js
+++ b/media/js/fellowship.js
@@ -1,30 +1,20 @@
-(function (win, $) {
+(function ($) {
     'use strict';
 
-        var $cta = $('.cta:not(a)');
-        var $cta_links = $cta.find('a');
-        // allow focus
-        $cta.attr('tabindex', 0).addClass('js');
-        // when activated add the class to show and remove listeners which preventDefault
-        function activate(caller) {
-            var $caller = $(caller);
-            var $calling_cta;
-            if($caller.hasClass('cta')){
-                $calling_cta = $caller;
-            } else {
-                $calling_cta = $caller.parents('.cta');
-            }
-            $calling_cta.addClass('active').off();
-            $calling_cta.find('a').off();
-        }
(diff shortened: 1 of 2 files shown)


media/js/fellowship.js
```diff
@@ -1,30 +1,20 @@
...
+        .on('focus touchstart', function(ev) {
+            ev.preventDefault();
             activate(this);
         });
 
-})(window, jQuery);
+    // when activated add the class to show and remove listeners which preventDefault
+    function activate(caller) {
+        var $caller = $(caller);
+        var $callingCta = $caller.hasClass('cta') ? $caller : $caller.parents('.cta');
```
darkwing: Can you detail why it's the clicked item or caller?  I think we may be able to do this better.

stephaniehobson: I didn't write all of this code so I'm not 100% sure about this but...

The widget works one of three ways:
1) User hovers over it, exposing the links and clicks on a .cta > a. The javascript doesn't actually have to do anything in this case.
2) User attempts to tap on the "share this" text but because the links are positioned over top of it but invisible, they tap on a link without seeing what it is. JS interfears on the first tap, prevents the link click, and exposes the links so they can see what they're doing.
3) User tabs to the .cta parent and, when it receives focus, it triggers the js to make the 3 links visible.

groovecoder:
Travis failures here are the connection issues with Travis - not the code. Okay to merge thru it.

stephaniehobson:
@darkwing Given https://github.com/mozilla/kuma/pull/3427#discussion_r37562276 are you okay moving forward with this code or did you want to try to refine it?

darkwing:
Yeah, I'm fine with it.

darkwing:
Rebased so all tests pass.

stephaniehobson:
R :+1:

PR #3429, opened 2015-08-20
bug 1194461 - Use data URIs for the tabzilla image

darkwing:
Not removing images from the repo in the case that we prefer to use stylus' JS utility (eventually) to implement data URIs.  Also dependent upon what pipeline does.


groovecoder:
Image looks good in many window sizes. Waiting for Travis ...

PR #3432, opened 2015-08-21
Bug 1144134 - editorial reviews should be yellow

stephaniehobson:
Upping the specificity on this to override the cascade.

I want to fix the mess that is the 40 classes we're using quite badly.


groovecoder:
Travis errors are connection issues not related to this code.

PR #3433, opened 2015-08-21
bug 1191348 - updated to allow use of a socket instead of a host

Sheeri:
change is so that if socket isn't specified, host is used. If socket is specified, host is *not* used.

### DIFF
--- a/scripts/clone_db.py
+++ b/scripts/clone_db.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 """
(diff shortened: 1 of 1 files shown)


jwhitlock:
Thanks for the fix!  I tested this in my developer VM, and it works with the socket.

Can you make two changes?
* Change the commit message to start with "bug 1191348", so it will get added to the bugzilla history on merge
* Change the top line from:

```shell
#!/usr/bin/env python
```

to 
```shell
#!/usr/bin/env python2.7
```

to reflect that it used Python 2.7's ``subprocess.check_output``.

PR #3434, opened 2015-08-21
bug 1184662 - Allow time for scrolling CSS to apply

darkwing:
Since we debounce scroll events, we should provide an allowance time before testing a given CSS property.


darkwing:
I believe this is the last transient issue we'll have.

groovecoder:
Failures are old ES memory/connection issues. Looks good; works well; less filling.

PR #3438, opened 2015-08-24
fix bug 1189596 - only remap urls under zones

groovecoder:


kuma/wiki/middleware.py
```diff
@@ -31,7 +31,8 @@ def process_request(self, request):
         remaps = DocumentZoneURLRemapsJob().get(request.locale)
         for original_path, new_path in remaps:
 
-            if request.path_info.startswith(original_path):
+            if (request.path_info == original_path or
+                request.path_info.startswith(''.join([original_path, '/']))):
```
jezdez: the original path could be a Unicode string, so this should be `u''.join([original_path, '/'])` or just `u'%s/' % original_path`

groovecoder:
Good catch. Updated code & test.

groovecoder:
Rebased on master to see if latest Travis fix works well. :smile:

jwhitlock:
+r

Tried:
```shell
curl -kI https://developer.mozilla.org/en-US/docs/Mozilla/Firefox_for_IOS # 302 Found
curl -kI https://developer-local.allizom.org/en-US/docs/Mozilla/Firefox_for_iOS # 200 OK
curl -kI https://developer-local.allizom.org/en-US/docs/Mozilla/Firefox # 302 Found
curl -kI https://developer-local.allizom.org/en-US/docs/Mozilla/Firefox/ # 302 Found
curl -kI https://developer-local.allizom.org/en-US/docs/Mozilla/Firefox_foo # 404 Not Found
```

PR #3426, opened 2015-08-19
Bug 1196342 - Install Elasticsearch and Memcache via Ansible.

jezdez:

PR #3439, opened 2015-08-24
bug 1196342 - use Travis elasticsearch service

groovecoder:
Alternative to https://github.com/mozilla/kuma/pull/3426: try using Travis CI's built-in elastic search service. Doesn't version-match to prod, but it's better than the all-out failures we have now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
     - "2.7"
 services:
     - memcached
+    - elasticsearch
 env:
     matrix:
         - TOXENV=py27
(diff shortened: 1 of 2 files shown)


groovecoder:
So, this is green. It's using a different version of ElasticSearch from what we have in prod, but like I said - I think this is better than constant test failure? @jezdez, @jwhitlock ?

jezdez:
Yeah, I think so, too. I wish we could just use the same Ansible provisioning as in #3426 to have parity, but until then I guess we have to bite the bullet. r+

PR #3440, opened 2015-08-24
1197858 travis intern

darkwing:
Testing automated Travis + intern tests.  Not ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,20 +13,32 @@ services:
     - elasticsearch
 env:
     matrix:
-        - TOXENV=py27
-        - TOXENV=flake8
-        - TOXENV=docs
+        #- TOXENV=py27
+        #- TOXENV=flake8
+        #- TOXENV=docs
+        - TOXENV=intern
     global:
         - DJANGO_SETTINGS_MODULE=settings_travis
         - DEBIAN_FRONTEND=noninteractive
         - CFLAGS=-O0
+        - secure: "SnWizKd9zRohcjRO7Dpslz2VHLo3pYruz5p1KWcHDa3ttVxMXGhNz1jBUD/Lazn1uifqSoI8o8nMC503VXc+hu9Gq33wDjjoQRAU4srz13Acr6uRe7mNnGt3P9dKqODucMMqn0FvcZv+7z9vD73KNWLYk8n70Bd3+Wsmcre4kyE="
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -19,13 +19,18 @@ env:
...
 before_install:
     - scripts/travis-install
     - pip install -U pip
+    - cd tests/ui && npm install intern@2.2.2 && cd ../../
 install:
     - pip install -r requirements/travis.txt
     - if [[ $TOXENV == 'py27' ]]; then ansible-playbook -vvv --tags=mysql --connection=local --inventory-file=provisioning/inventory provisioning/travis.yml; fi
-script: tox -v
+script:
+    - cd tests/ui && node_modules/.bin/intern-runner config=intern-sauce-labs destructive=true wd=Web
```
groovecoder: When we get it working, we should see if we can move the command into the `commands` section of `tox.ini`: that should let all the commands run in parallel so the python tests and the intern tests can run simultaneously and hopefully not take so long to finish them all.

groovecoder: Hmm ... actually it might help to move this into tox.ini to get it working in the first place? Looks like Travis is running the intern tests for each of our build jobs configured in tox, rather than just running them once.

.travis.yml
```diff
@@ -13,20 +13,32 @@ services:
...
     - pip install -U pip
+    - cd tests/ui && npm install intern@2.2.2 && cd ../../
 install:
     - pip install -r requirements/travis.txt
+    - pip install -r requirements/compiled.txt # Make me conditional for intern
     - if [[ $TOXENV == 'py27' ]]; then ansible-playbook -vvv --tags=mysql --connection=local --inventory-file=provisioning/inventory provisioning/travis.yml; fi
-script: tox -v
+    - if [[ $TOXENV == 'intern' ]]; then ansible-playbook -vvv --tags=mysql --connection=local --inventory-file=provisioning/inventory provisioning/travis.yml; fi
+before_script:
+    - python manage.py runserver 0.0.0.0:80 # Make me conditional for intern
```
jezdez: That command will block and needs a sudo (since it uses a privileged port). I would recommend using `sudo python manage.py runserver 0.0.0.0:80 &` That should do the trick

jezdez: Now that I realize what you try to do, this command needs to run inside the tox config as that's where the rest of the dependencies are installed.

darkwing:
@jezdez Any insight as to why this keeps hanging on the same MySQL piece?

groovecoder:
This was a good go at it, but we're going to change to running the Intern test suite against the stage & prod servers during our deployment process.

PR #3441, opened 2015-08-24
bug 1189451 - add 'Host' header

groovecoder:
To by-pass the internal load balancer from the locust cluster to the stage web-heads, we run locust with: `--host=<IP of stage web server>`.

So, we need to add an explicit 'Host' HTTP header on the requests so Apache will respond to them on the correct vhost.

### DIFF
--- a/tests/performance/smoke.py
+++ b/tests/performance/smoke.py
@@ -1,3 +1,5 @@
+import os
+
 from locust import HttpLocust, TaskSet, task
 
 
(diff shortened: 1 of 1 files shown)


tests/performance/smoke.py
```diff
@@ -9,6 +11,12 @@ class SmokeBehavior(TaskSet):
     See https://bugzil.la/1186085
     """
 
+    def on_start(self):
+        locust_host = 'developer.allizom.org'
+        if 'LOCUST_HOST' in os.environ:
+            locust_host = os.environ['LOCUST_HOST']
```
jezdez: Let's do this in one line: `locust_host = os.environ.get('LOCUST_HOST', 'developer.allizom.org')`

groovecoder: D'oh, yeah good idea.

groovecoder:
Updated.

jwhitlock:
+r

PR #3442, opened 2015-08-25
bug 1197971 - dont render previews of deferred docs

groovecoder:
To help mitigate risk of [disruptions of service caused by previewing documents with a large number of macro calls](https://bugzilla.mozilla.org/show_bug.cgi?id=1197971#c1), skip KumaScript rendering when previewing a document with deferred rendering. ([We automatically set documents to defer rendering when they take too long to render.](https://github.com/mozilla/kuma/blob/1de4dd081432647c8eb3f8c519a6d141cbf24be5/kuma/wiki/models.py#L547-L553)


jwhitlock:
+r

PR #3443, opened 2015-08-25
bug 957802 - set db logging to warn during tests

groovecoder:
This drives me crazy when I'm trying to debug tests and have to scroll all the way up thru a bunch of `DEBUG` level output from `django.db.backends.schema`. I even [asked on Stack Overflow](https://stackoverflow.com/questions/32158773/how-do-i-disable-django-migration-debug-logging) and I wonder if this is an upstream bug in/between django or django-nose? (@jwhitlock - you do django-nose now right?)

### DIFF
--- a/settings_test.py
+++ b/settings_test.py
@@ -12,3 +12,9 @@ INSTALLED_APPS += (
 BANISH_ENABLED = False
 
 DEMO_UPLOADS_ROOT = '/home/vagrant/uploads/demos'
+
+LOGGING['loggers']['django.db.backends.schema'] = {
+    'handlers': ['console'],
+    'propagate': True,
+    'level': 'WARNING',
+}


jwhitlock:
Yeah, but I've been too busy on https://github.com/mdn/browsercompat/pull/38 to actually fix django-nose.  I've added an [answer](http://stackoverflow.com/a/32210265/10612), but the commit may still be an improvement - I've rarely if ever needed the database traceback.

jwhitlock:
+r - the code works.  Tested by adding an always-fail test:

```python
    def test_fail(self):
        self.fail("Testing logging.")
```
and getting it to run first:
```
./manage.py test --failed -sx  # Run up to the failing test
./manage.py test --failed -sx  # Run just the failing test
```

I'll leave it to you if you still want to merge, or if ``--nologcapture`` is a better solution.

groovecoder:
Going to merge. That works quickly, but I **want** to capture the other logging output. I just don't want the `DEBUG` level of output.

PR #3444, opened 2015-08-25
Bug 1144134 - indicator refinements

stephaniehobson:
Adding language icon to in translation warning.

Removing first-line styling (beginning with a strong tag is a convention I would like the writers to follow but not one I should be enforcing with CSS)


darkwing:
R+

PR #3446, opened 2015-08-26
bug 1198931 - Add Archive subtree on all locales to robots.txt

a2sheppy:
This patch adds the Archive subtree in all docs locales to the
robots.txt file so they shouldn’t be indexed by the major search
engines.


darkwing:
This site says it will work:  http://tools.seobook.com/robots-txt/analyzer/

groovecoder:
Closing until the decision is made about the change in https://bugzilla.mozilla.org/show_bug.cgi?id=1198931

PR #3447, opened 2015-08-27
bug 1199166 - disallow robots from comparing

groovecoder:
[Today's down-time incident](https://rpm.newrelic.com/accounts/263620/applications/3172075/downtime/16882252) was [caused by Yahoo Slurp! crawling onto a bad and expensive compare transaction](https://bugzilla.mozilla.org/show_bug.cgi?id=1199166#c1).

To help mitigate the risk of search crawlers triggering overload transactions, we want to add a number of new `Disallow` directives to our `robots.txt`, so we can expect well-behaved robots to stay away from them.

I used Google Webmaster Tools to verify that this `robots.txt` file will not disallow regular docs urls. I couldn't find equivalent webmaster tools for Yahoo, so I emailed Kev Needham (our Yahoo liaison) and set up a Google Analytics alert to email me if our traffic from Yahoo drops below our minimum observed threshold, just in case.


media/robots.txt
```diff
@@ -1,4 +1,4 @@
-User-Agent: *
+User-agent: *
```
jwhitlock: was this a suggested change?

media/robots.txt
```diff
@@ -6,5 +6,30 @@ Request-rate: 1/5
 Disallow: /*feed=rss
 Disallow: /*type=feed
 Disallow: /skins
-Disallow: /template:
+Disallow: /*docs/feeds
+Disallow: /*docs/templates
+Disallow: /*docs*Template
```
jwhitlock: will this block https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Tutorial/Templates?

media/robots.txt
```diff
@@ -6,5 +6,30 @@ Request-rate: 1/5
...
+Disallow: /*docs*$move
+Disallow: /*docs*$quick-review
+Disallow: /*docs*$revert
+Disallow: /*docs*$repair_breadcrumbs
+Disallow: /*docs*$delete
+Disallow: /*docs*$restore
+Disallow: /*docs*$purge
+Disallow: /*docs*$subscribe
+Disallow: /*docs*$vote
+Disallow: /*docs*.json
```
jwhitlock: will this block https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json ?

media/robots.txt
groovecoder: Yes, suggested by Google webmaster tools

media/robots.txt
groovecoder: Good catch. Updated.

media/robots.txt
groovecoder: Nope. robots.txt directives aren't regex :neutral_face: 
mdn-robots-json-okay

jwhitlock:
+r would crawl again

groovecoder:
giphy

PR #3449, opened 2015-08-27
bug 1199166 - select_related for revision.document

groovecoder:
This saves a couple of extra queries on the compare view. Would like @jezdez to make sure I'm not crazy.


jezdez:
We've discussed this before, `prefetch_related` is preferred here since it will automatically span multiple tables when a related queryset result will ask for other related data (such as tags). It also works better since it doesn't fetch everything with a very specific query that is hard to cache with MySQL's query cache as done by `select_related`. We rely on that query cache heavily and should not abandon it in favor of a reduced number of queries. Closing as wontfix.

PR #3450, opened 2015-08-27
bug 1180208 - Manually load JSON in migration 0004

jwhitlock:
Commit 874db113 changed profile.misc to a nullable TextField, so JSONField could be removed. This changes the follow-on commit to manually load the JSON string.

### DIFF
--- a/kuma/users/migrations/0004_auto_20150722_1243.py
+++ b/kuma/users/migrations/0004_auto_20150722_1243.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+from json import loads
 
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models, migrations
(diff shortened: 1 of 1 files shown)


groovecoder:
Test failure may be legit?
```
A document's json field is refreshed on render(), but not on save()
```

jwhitlock:
The ``document.json`` field was always a text field, so not the same issue.  I'm not sure why the test failed, but it passed on rebuild.

groovecoder:
`provision` works well on existing vm. I'm going to make sure we're clear of deployments and then destroy and rebuild my vm to make sure this works on a fresh/blank vm too.

jezdez:
Aha, seems like someone didn't do the migration in time eh? Well, this seems sensible r+

jwhitlock:
Or you could say "never change a migration".  In general, I'd prefer to live with past mistakes until we do a periodic migration reset (maybe once a year). In this case, I think we've found all the repercussions.

jezdez:
That sounds like a nice notion to have one day, a *sane* migration story. I need to tell you about the time we used multiple migration systems in kuma at once, probably when we have alcoholic beverages close by.

In the meantime I'm going to be not sad to break everyone's VM to get to a state that we don't have to think about custom written stuff from the past like that JSONField.

FWIW I can live with past mistakes as well, heck that's software development, that's a given. There is a breaking point though where dealing with cruft becomes more costly than just burning some bridges every once in a while. At MDN I learned to tolerate a lot more cruft than usual but also how to build bigger fires when the time comes to burn them down.

jwhitlock:
Agreed, with real code it is compromises all the way down, and kuma is realer than most. I'm chastising myself too - I saw PR #3390, I saw the changed migration and heard that nagging voice in the back of my head, and didn't track it down any farther. It's sad that fixing code is easier than getting a new production database.  It's sad that you need production databases to do effective work with kuma.  These are all fronts where we are making progress though. This feels like patching a leak rather than buying a pump, which is progress.

PR #3452, opened 2015-08-27
bug 1199166 - block_user_agents decorator for views

groovecoder:
Not sure this `settings.BLOCKABLE_USER_AGENTS` approach is the best one. Need some r? on this overall approach ...

### DIFF
--- a/kuma/core/decorators.py
+++ b/kuma/core/decorators.py
@@ -1,6 +1,8 @@
 from functools import wraps
 import inspect
+import re
 
+from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.contrib.auth.decorators import user_passes_test
 from django.core.exceptions import PermissionDenied
@@ -136,3 +138,21 @@ def is_superuser(u):
 
 superuser_required = user_passes_test(is_superuser)
 #: A decorator to use for requiring a superuser
+
+
+def block_user_agents(view_func):
+    blockable_user_agents = getattr(settings, 'BLOCKABLE_USER_AGENTS', [])
+    blockable_ua_patterns = []
+    for agent in blockable_user_agents:
+        blockable_ua_patterns.append(re.compile(agent))
+
+    def agent_blocked_view(request, *args, **kwargs):
+        http_user_agent = request.META.get('HTTP_USER_AGENT', None)
+        if http_user_agent is not None:
+            for pattern in blockable_ua_patterns:
+                if pattern.search(request.META['HTTP_USER_AGENT']):
+                    return HttpResponseForbidden()
+        return view_func(request, *args, **kwargs)
+
+    return wraps(view_func,
+                 assigned=available_attrs(view_func))(agent_blocked_view)
--- a/kuma/wiki/views/list.py
+++ b/kuma/wiki/views/list.py
@@ -3,6 +3,7 @@ from django.http import Http404
 from django.shortcuts import get_object_or_404, get_list_or_404, render
 from django.views.decorators.http import require_GET
 
+from kuma.core.decorators import block_user_agents
 from kuma.core.utils import paginate
 
 from ..constants import DOCUMENTS_PER_PAGE
@@ -12,6 +13,7 @@ from ..models import (Document, DocumentTag, Revision, ReviewTag,
 from ..queries import MultiQuerySet
 
 
+@block_user_agents
 @require_GET
 def documents(request, category=None, tag=None):
     """
@@ -47,6 +49,7 @@ def documents(request, category=None, tag=None):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
 @require_GET
 def templates(request):
     """
@@ -62,6 +65,7 @@ def templates(request):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
 @require_GET
 def tags(request):
     """
@@ -72,6 +76,7 @@ def tags(request):
     return render(request, 'wiki/list/tags.html', {'tags': tags})
 
 
+@block_user_agents
 @require_GET
 def needs_review(request, tag=None):
     """
@@ -90,6 +95,7 @@ def needs_review(request, tag=None):
     return render(request, 'wiki/list/needs_review.html', context)
 
 
+@block_user_agents
 @require_GET
 def with_localization_tag(request, tag=None):
     """
@@ -108,6 +114,7 @@ def with_localization_tag(request, tag=None):
     return render(request, 'wiki/list/with_localization_tags.html', context)
 
 
+@block_user_agents
 @require_GET
 def with_errors(request):
     """
@@ -123,6 +130,7 @@ def with_errors(request):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
 @require_GET
 def without_parent(request):
     """Lists wiki documents without parent (no English source document)"""
@@ -137,6 +145,7 @@ def without_parent(request):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
 @require_GET
 def top_level(request):
     """Lists documents directly under /docs/"""
@@ -151,6 +160,7 @@ def top_level(request):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
 @require_GET
 @process_document_path
 @prevent_indexing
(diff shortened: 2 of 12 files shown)


kuma/core/decorators.py
```diff
@@ -136,3 +137,17 @@ def is_superuser(u):
 
 superuser_required = user_passes_test(is_superuser)
 #: A decorator to use for requiring a superuser
+
+
+def block_user_agents(view_func):
+    blockable_user_agents = []
+    if hasattr(settings, 'BLOCKABLE_USER_AGENTS'):
+        blockable_user_agents = settings.BLOCKABLE_USER_AGENTS
```
jezdez: Those three lines can be written as `blockable_user_agents = getattr(settings, 'BLOCKABLE_USER_AGENTS', [])`

kuma/core/decorators.py
```diff
@@ -136,3 +137,17 @@ def is_superuser(u):
...
+
+    def agent_blocked_view(request, *args, **kwargs):
+        if 'HTTP_USER_AGENT' in request.META:
```
jezdez: Don't do a containment check and then get the data again two lines later in every loop over the patterns, this is a high traffic decorator, stuff like that will safe cycles. Use instead:

```
HTTP_USER_AGENT = request.META.get('HTTP_USER_AGENT')
for user_agent_pattern in blockable_user_agents:
    if user_agent_regex.match(HTTP_USER_AGENT):
        raise PermissionDenied
# ...
```

The use of match instead of search should make this faster as well.

settings.py
```diff
@@ -1188,6 +1189,15 @@ def get_user_url(user):
 
 HONEYPOT_FIELD_NAME = 'website'
 
+BLOCKABLE_USER_AGENTS = (
```
jezdez: Please make that setting a list of names and then turn them into a list of regex patterns in the module where the decorator is defined.

kuma/core/decorators.py
```diff
@@ -136,3 +137,17 @@ def is_superuser(u):
...
 superuser_required = user_passes_test(is_superuser)
 #: A decorator to use for requiring a superuser
+
+
+def block_user_agents(view_func):
+    blockable_user_agents = []
+    if hasattr(settings, 'BLOCKABLE_USER_AGENTS'):
+        blockable_user_agents = settings.BLOCKABLE_USER_AGENTS
+
+    def agent_blocked_view(request, *args, **kwargs):
```
jezdez: Apple `functools.wraps` here like so: `@wraps(view_func, assigned=available_attrs(view_func))` where `available_attrs` is `django.utils.decorators.available_attrs`.

kuma/core/decorators.py
jezdez: Wait, that wasn't quite right, this should be:

```
HTTP_USER_AGENT = request.META.get('HTTP_USER_AGENT', None)
if HTTP_USER_AGENT is not None: 
    for user_agent_pattern in blockable_user_agents:
        if user_agent_regex.match(HTTP_USER_AGENT):
            raise PermissionDenied
# ...
```

groovecoder:
Updated.

groovecoder:
Oh, still needs tests ...

groovecoder:
Adding tests with real user agent values showed me that we need to use `pattern.search` to catch the user agent strings. This is ready for final review & merge.

groovecoder:
@jwhitlock - can you take this one now?

jwhitlock:
Is kumascript blocked by this?  I'm pretty sure ``document.get_children`` and ``document.as_json`` are used by KS macros.

groovecoder:
Not likely but it could potentially block KumaScript if it does something crazy like send 'curl' or 'get' as the user agent string. I'll verify before merging & deploying. (I'll also add `YandexBot` to the blockable agents)

kuma/wiki/views/list.py
```diff
@@ -12,6 +13,7 @@
 from ..queries import MultiQuerySet
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/all and https://developer.mozilla.org/en-US/docs/tag/* to robots.txt as well?

kuma/wiki/views/list.py
```diff
@@ -62,6 +65,7 @@ def templates(request):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/tags to robots.txt?

kuma/wiki/views/list.py
```diff
@@ -72,6 +76,7 @@ def tags(request):
     return render(request, 'wiki/list/tags.html', {'tags': tags})
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/needs-review and https://developer.mozilla.org/en-US/docs/needs-review/* to robots.txt?

kuma/wiki/views/list.py
```diff
@@ -90,6 +95,7 @@ def needs_review(request, tag=None):
     return render(request, 'wiki/list/needs_review.html', context)
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/localization-tag and https://developer.mozilla.org/en-US/docs/localization-tag/* to robots.txt?

kuma/wiki/views/list.py
```diff
@@ -108,6 +114,7 @@ def with_localization_tag(request, tag=None):
     return render(request, 'wiki/list/with_localization_tags.html', context)
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/with-errors to robots.txt?

kuma/wiki/views/list.py
```diff
@@ -123,6 +130,7 @@ def with_errors(request):
     return render(request, 'wiki/list/documents.html', context)
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/without-parent to robots.txt?

kuma/wiki/views/list.py
```diff
@@ -137,6 +145,7 @@ def without_parent(request):
...
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/top-level to robots.txt?

kuma/wiki/views/list.py
```diff
@@ -47,6 +49,7 @@ def documents(request, category=None, tag=None):
...
```
jwhitlock: *(I didn't miss this one - it's in robots.txt already)*

kuma/wiki/views/misc.py
```diff
@@ -42,6 +42,7 @@ def ckeditor_config(request):
                   content_type='application/x-javascript')
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/get-documents to robots.txt?

kuma/wiki/views/misc.py
```diff
@@ -157,6 +160,7 @@ def helpful_vote(request, document_path):
     return redirect(document)
 
 
+@block_user_agents
```
jwhitlock: Should we add https://developer.mozilla.org/en-US/docs/load/ to robots.txt? (superuser only)

jwhitlock:
+r on code, with possible additions to robots.txt

kuma/wiki/views/list.py
groovecoder: :stuck_out_tongue:

jwhitlock:
+r on robots.txt additions. Remember to confirm kumascript before it goes to production.

groovecoder:
* Added the `Disallow` entries to `robots.txt`.
* Added `YandexBot` to `BLOCKABLE_USER_AGENTS` since it triggered a db-heavy transaction yesterday.

Yup; still need to verify the kumascript user agent sent to Apache/django.

groovecoder:
After looking thru access logs for other crawling user agents, I also added:
```
"Baiduspider",
"CCBot",
"ScoutJet",
```

groovecoder:
I verified locally that KumaScript requests are sent to Apache without a blocked user agent string:
```
developer-local.allizom.org:443 127.0.0.1 - - [03/Sep/2015:16:57:06 +0000] "GET /en-US/docs/en-US/Template:dekiscript:page?raw=1 HTTP/1.1" 404 3947 "-" "-"
```
And cyliang sees the internal requests to production Apache are sent similarly - i.e., with `"-"` user agent:
```
 - - [03/Sep/2015:09:30:04 -0700] "GET /en-US/docs/MDN/Kuma/API?raw=1 HTTP/1.1" 200 6722 "-" "-"
```
So, this is good for merge + deploy.

PR #3453, opened 2015-08-27
fix bug 1194308 - Prevent Wrapping of code buttons

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1194308


stephaniehobson:
How does this prevent them from wrapping?

darkwing:
It doesn't -- I took the gist of sheppy's complaint as the odd left spacing for the second line

PR #3455, opened 2015-08-31
fix bug 1200259 - Implement static homepage

darkwing:
This is (1) controversial and (2) incomplete, but I wanted to throw this out there.  Potential problems:

*  Waffle flags on the homepage
*  User logs in, homepage refreshes, user still appears logged out.

I don't want to load the login block with AJAX because it's another request every time on the homepage.  Let's all think on this.  My solution to the second bullet point is redirecting users to their profile page if they've logged in from the homepage

### DIFF
--- a/kuma/landing/views.py
+++ b/kuma/landing/views.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.shortcuts import render
 from django.views import static
+from django.views.decorators.cache import cache_page
 
 from kuma.core.sections import SECTION_USAGE
 from kuma.core.cache import memcache
@@ -10,6 +11,7 @@ from kuma.search.models import FilterGroup
 from kuma.search.serializers import GroupWithFiltersSerializer
 
 
+@cache_page(HOMEPAGE_CACHE_TIMEOUT, key_prefix=HOMEPAGE_CACHE_PREFIX)
 def home(request):
     """Home page."""
     demos = Submission.objects.all_sorted(sort='recentfeatured', max=12)
@@ -30,7 +32,8 @@ def home(request):
         'demos': demos,
         'updates': updates,
         'stats': community_stats,
-        'command_search_filters': serializer.data
+        'command_search_filters': serializer.data,
+        'hide_header_login': True,
     }
     return render(request, 'landing/homepage.html', context)
 
--- a/scripts/chief_deploy.py
+++ b/scripts/chief_deploy.py
@@ -97,6 +97,7 @@ def update_info(ctx):
             ctx.local("svn status")
 
         ctx.local("git rev-parse HEAD > media/revision.txt")
+        ctx.local("git show --format='%ai' > media/deploytime.txt")
 
 
 @task
--- a/templates/includes/login.html
+++ b/templates/includes/login.html
@@ -1,4 +1,4 @@
-{% if user.is_authenticated() %}
+{% if user.is_authenticated() and not hide_header_login %}
     <div class="oauth-logged-in">
         {% include 'socialaccount/snippets/login_service_icon.html' %}
         <ul>
@@ -7,7 +7,6 @@
         </ul>
     </div>
 {% else %}
-
     {% set github_url = provider_login_url('github', next=next_url) %}
 
     <form>
--- a/settings.py
+++ b/settings.py
@@ -74,6 +74,10 @@ CACHES = {
 }
 
 CACHEBACK_CACHE_ALIAS = 'memcache'
+CACHE_MIDDLEWARE_ALIAS = 'memcache'
+
+HOMEPAGE_CACHE_TIMEOUT = 60 * 15
+HOMEPAGE_CACHE_PREFIX = 'v1'
 
 # Addresses email comes from
 DEFAULT_FROM_EMAIL = 'notifications@developer.mozilla.org'


darkwing:
Assigning to jezdez for first round of thoughts/problems.

kuma/landing/views.py
```diff
@@ -30,7 +32,8 @@ def home(request):
         'demos': demos,
         'updates': updates,
         'stats': community_stats,
-        'command_search_filters': serializer.data
+        'command_search_filters': serializer.data,
+        'hide_header_login': True
```
jezdez: Please add a trailing comma

templates/includes/login.html
```diff
@@ -1,4 +1,4 @@
-{% if user.is_authenticated() %}
+{% if (not hide_header_login) and user.is_authenticated() %}
```
jezdez: Let's turn this around and make it `{% if user.is_authenticated() and not hide_header_login %}`, much easier to read

jezdez:
The `@cache_page` decorator uses the cache middleware behind the scenes which makes it use the "default" cache backend. For historical reasons the "default" cache backend is still a server process local memory backend -- one that stores cache values in a localized process cache instead of a global memcache that is shared by all server processes across all servers.

We need to tell the decorator to use the proper cache backend, please define `CACHE_MIDDLEWARE_ALIAS = 'memcache'` next to the `CACHES` setting to do that.

kuma/landing/views.py
```diff
@@ -10,6 +11,7 @@
 from kuma.search.serializers import GroupWithFiltersSerializer
 
 
+@cache_page(60 * 15)
```
jezdez: Setting a short timeout like this should work fine with the caching that happns on the load balancer every 60 minutes. So we can say that changes that impact things on the frontpage will take a minimum of 15 minutes and a maximum of 60 minutes to reach it.

What I want to add though is an optional parameter to the decorator called `key_prefix` that would allow us to invalidate the frontpage cache on deploy if we ever need an urgent change to apply for certain. That's really needed since by default Django's cache middleware (the backend of the decorator) doesn't take a build ID or something into account. So let's add this:

    @cache_page(60 * 15, key_prefix='v1')
    def home(request):
        # ...

Or even better move both values into settings so we have all cache related values in one spot:

    @cache_page(HOMEPAGE_CACHE_TIMEOUT, key_prefix=HOMEPAGE_CACHE_PREFIX)
    def home(request):
        # ...

Then in `settings.py` next to the `CACHES` setting:

    HOMEPAGE_CACHE_TIMEOUT = 60 * 15
    HOMEPAGE_CACHE_PREFIX = 'v1'

jezdez:
Two things we need to figure out:

1. Are there any other things happening during template rendering in the `homepage.html` template (and its parent template the `base.html`) that we would basically let only happen every 15 minutes when the cache is timing out? E.g. look at the waffle flag conditions please to see if this can have a negative impact on operations.

2. Would we profit from making the homepage view also return 304 responses instead of 200s all the time? There are deorators for conditional view processing that set last modified and etag response headers according to a value we provide: https://docs.djangoproject.com/en/1.8/topics/conditional-view-processing/

We use the last_modified decorator in the wiki document view using the time the document was last rendered. Maybe we could use something similar for the frontpage as well? That would assume that we somehow can provide a date to the decorator, so something that we need to set during the deploy, similar to a build ID we set. On the Python side we could then use that deploy datetime as the basis for the last-modified and etag headers, effectively turning the homepage to only change on deploys. That would be a drastic step I know, but something maybe that would be worthwhile compared to the speed gains.

darkwing:
1.  We do currently have a host of homepage waffles but none are time-specific.
2.  I'll give the 304 thing a go.
3.  We'll need to modify the login backend to redirect to the user to the user profile page if they login from the homepage

darkwing:
Updated this PR so that the deploy script creates a file that stores the last commit; we may want to rethink this but I thought it would be a start.

kuma/landing/views.py
```diff
@@ -10,6 +11,7 @@
 from kuma.search.serializers import GroupWithFiltersSerializer
 
 
+@cache_page(HOMEPAGE_CACHE_TIMEOUT, key_prefix=HOMEPAGE_CACHE_PREFIX)
```
jezdez: That'll need to be:

    @cache_page(settings.HOMEPAGE_CACHE_TIMEOUT,
                key_prefix=settings.HOMEPAGE_CACHE_PREFIX)
    def home(request):
    # ...

scripts/chief_deploy.py
```diff
@@ -97,6 +97,7 @@ def update_info(ctx):
             ctx.local("svn status")
 
         ctx.local("git rev-parse HEAD > media/revision.txt")
+        ctx.local("git show --format='%ai' > media/deploytime.txt")
```
jezdez: Let's not use git here actually, since that can be misleading. E.g. check out the difference between:

`git show -s --format=%ai 356032ab82bea4248977d255820a9a5f8652dc4f`

and

`git show -s --format=%ci 356032ab82bea4248977d255820a9a5f8652dc4f`

Both of which refer to 356032ab82bea4248977d255820a9a5f8652dc4f which because it was reviewed by me changes the time of the commit versus the time of the authoring. The times differ though and therefore we should prevent our deploy system to be dependent on this type of thing.

Instead use this to generate a UTC datetime that we can easily parse later in Python

    ctx.local("date --utc --rfc-822 > media/deploytime.txt")

Here's the code for parsing it with Python:

    from datetime import datetime
    import email.utils

    date_from_file = 'Mon, 07 Sep 2015 13:37:00 +0000'
    date_in_python = datetime.utcfromtimestamp(email.utils.mktime_tz(email.utils.parsedate_tz(date_from_file)))

This isn't beautiful but we can do this with built-in features from Python.

jezdez:
Okay the next step is to use the date from the datetime.txt in a function for the [condition decorator](https://docs.djangoproject.com/en/1.8/topics/conditional-view-processing/) or the only [last_modified](https://docs.djangoproject.com/en/1.8/topics/conditional-view-processing/#shortcuts-for-only-computing-one-value) decorator if we want to only set the `Last-Modified` headers and not the `ETag`.

groovecoder:
Do we need to keep this open? Or has this work fallen off our priority list?

darkwing:
I'll reluctantly close this.

PR #3456, opened 2015-08-31
Bug 1144134: Summary box border colour.

stephaniehobson:
Summary box border colour was being over-ridden by .text-content table declarations so I gave it more specificity.


darkwing:
R+, works for me

PR #3457, opened 2015-09-01
bug 1177264 - optimize zone root summary display

groovecoder:
Assigning to @darkwing in case he can remember a reason to keep the `selector_content_find` helper?


darkwing:
R+, works for me

PR #3461, opened 2015-09-01
Bug 1197444: Move images to kuma/static/img

openjck:


openjck:
Also cc @stephaniehobson @darkwing @groovecoder

darkwing:
Kicked tests

darkwing:
@openjck I think you have a python code style issue here.

openjck:
> @openjck I think you have a python code style issue here.

Yep, you're right. Tests should pass now.

openjck:
Tests passing! :smile:

darkwing:
It appears to me that the zone background gradient is missing when I create a zone page.

kuma/landing/templates/landing/promote_buttons.html
```diff
@@ -274,7 +274,7 @@
         // Update the button preview image.
         $('#button-image')
             .attr('alt', sel.alt_text)
-            .attr('src', '{{ MEDIA_URL }}img/promote/promobutton_'+sel.id+'.png');
+            .attr('src', '{{ STATIC_URL }}img/promote/promobutton_'+sel.id+'.png');
```
jezdez: You should be able to do `{{ static('img/promote/promobutton_' + sel.id + '.png'') }}` here since we're using Jinja2 and the limitations of Django's template language doesn't apply here

kuma/landing/templates/landing/promote_buttons.html
```diff
@@ -284,7 +284,7 @@
         // Update the button code.
         $('#button-code').val([
             '<a href="', sel.tracked_url, '" title="', sel.alt_text, '">',
-            '<img src="{{ request.build_absolute_uri(MEDIA_URL) }}img/promote/promobutton_', sel.id, '.png" ', 
+            '<img src="{{ request.build_absolute_uri(STATIC_URL) }}img/promote/promobutton_', sel.id, '.png" ',
```
jezdez: The use of `request.build_absolute_uri` isn't needed here since the staticfiles storage should handle such decisions. Also same case of using the `static` tag here is needed

kuma/search/templates/search/plugin.html
```diff
@@ -4,7 +4,7 @@
 {# L10n: Search Mozilla Developer Network documentation #}
 <Description>{{ _('Search Mozilla Developer Network documentation') }}</Description>
 <InputEncoding>UTF-8</InputEncoding>
-<Image width="16" height="16" type="image/png">{{ MEDIA_URL }}img/favicon32.png</Image>
+<Image width="16" height="16" type="image/png">{{ STATIC_URL }}img/favicon32.png</Image>
```
jezdez: Let's use the static tag here.

settings.py
```diff
@@ -420,6 +421,10 @@ def get_locales():
     'django.contrib.staticfiles.finders.FileSystemFinder',
 )
 
+STATICFILES_DIRS = (
+    'kuma/static',
```
jezdez: Please use `path('kuma', 'static')` here to have cross platform compatible paths

jezdez: (path being a helper function define at the top of the file)

media/stylus/includes/vars.styl
```diff
@@ -173,7 +173,7 @@ $selector-heading-font-fallback = 'html[data-ffo-opensanslight="false"]:not(.no-
 paths
 ====================================================================== */
 $path-to-assets = '/media/';
-$path-to-images = $path-to-assets + 'img/';
+$path-to-images = '/static/img/';
```
jezdez: This should be a relative path from the compiled CSS files to the image file so we can use of staticfiles' included replacement of full file names (`url('../img/header.png')` to `url('../img/header.ddaac5ce6246.png')`. The relativeness is needed to accurately calculate a map of files in all CSS files something if we'd hard code the root path would not work depending on the storage backend we use.

settings.py
```diff
@@ -1179,7 +1184,7 @@ def get_user_url(user):
         'AUDIENCE': 'https://developer.mozilla.org',
         'REQUEST_PARAMETERS': {
             'siteName': 'Mozilla Developer Network',
-            'siteLogo': MEDIA_URL + 'img/opengraph-logo.png',
+            'siteLogo': STATIC_URL + 'img/opengraph-logo.png',
```
jezdez: Sadly that's not something we can easily make it use the static tag as there is a circular import issue preventing this (we can't import static to use it here because that loads the settings, which imports static, which imports the settings etc). There is a place in allauth where those values are actually passed along to the Persona login template, so I think we can inject it there properly with the static tag, but I'm afraid we have to live with this one direct `STATIC_URL` use. I think filing a ticket would make sense for that case.

.gitignore
```diff
@@ -18,7 +18,7 @@ build.py
 media/uploads
 media/attachments
 media/sitemap*
-static/
+/static/
```
jezdez: Nice! :+1:

kuma/attachments/templates/attachments/includes/attachment_list.html
```diff
@@ -57,7 +57,7 @@
           <td colspan="4">
             <input type="submit" name="submit" value="Upload" />
 
-            {% set spinner_path = MEDIA_URL + 'img/upload-spinner.gif' %}
+            {% set spinner_path = static('img/upload-spinner.gif') %}
             <img src="{{ spinner_path }}" id="page-attachments-spinner" alt="{{ _('Uploading') }}" />
```
jezdez: You can probably save this one `set` call

jezdez:
So far this is looking good @openjck! I need to dive into the code later again to look for other leftovers, this is exciting!

kuma/landing/templates/landing/promote_buttons.html
openjck: I'm not sure that would be possible. `sel.id` is defined on the client-side.

settings.py
openjck: Great catch! Will do.

kuma/attachments/templates/attachments/includes/attachment_list.html
openjck: Yep. Nice catch!

kuma/landing/templates/landing/promote_buttons.html
openjck: Same as above. Unfortunately, it looks like we'll need to use `STATIC_URL` directly in some form unless we refactor this.

media/stylus/includes/vars.styl
openjck: That's great! I was wondering about how to get staticfiles to replace paths. Thanks for the info.

openjck:
Updated!

> It appears to me that the zone background gradient is missing when I create a zone page.

Interesting. The gradient appears for me. Is the homepage missing anything? Are any errors reported on the `foreman` output?

darkwing:
@openjck I guess I lied, it's there now.  Sorry! :(

openjck:
> @openjck I guess I lied, it's there now. Sorry! :(

No problem at all. Happy it's working now. Thanks for checking it out.

stephaniehobson:
- [ ] missing icons on [demo pages](https://developer-local.allizom.org/en-US/demos/detail/paperfold-css/launch)
screen shot 2015-09-04 at 09 58 08
- [ ] remove icon missing from tag-it instances (example from user profile) 
screen shot 2015-09-04 at 10 09 47
- [ ] Firefox logo missing from [FF zone landing](https://developer-local.allizom.org/en-US/Firefox)
- [ ]  Images are not showing up on the MDN 10 history page... not sure this is caused by your PR or some weirdness in my local set up but don't have time to dig into it, can you look?

openjck:
Awesome, awesome catch. The missing demo and tag icons have been fixed.

The in-content problems have been fixed with a temporary workaround. Long story short: those pages have hardcoded references to the old addresses. So in the long term, we need to redirect from `/media/img` to `/static/img`. The only problems is that we can't redirect away from `/media` on production without assistance from WebOps (see [bug 1145793](https://bugzilla.mozilla.org/show_bug.cgi?id=1145793)).

For the time being, I added a redirect that will work in Vagrant. I also added a symlink from the old path to the new path. Shortly after this code is pushed to production, I'll coordinate with WebOps to redirect from the old path to the new path. Once the redirect is working on production, I'll submit a pull request to remove the symlink. It's a wacky workaround, but it's the only way to keep the images available without downtime.

kuma/landing/templates/landing/promote_buttons.html
jezdez: You're totally right, my fault

kuma/landing/templates/landing/promote_buttons.html
jezdez: I think we're fine with this now, but we should open a follow-up ticket to remove all uses of `STATIC_URL` to be able to switch out the staticfiles storage backend when we move to Cloudfront or some other form of backend that generates URLs to files programmatically via an API.

media/stylus/includes/vars.styl
```diff
@@ -173,7 +173,7 @@ $selector-heading-font-fallback = 'html[data-ffo-opensanslight="false"]:not(.no-
 paths
 ====================================================================== */
 $path-to-assets = '/media/';
-$path-to-images = $path-to-assets + 'img/';
+$path-to-images = '../../static/img/';
 $logo-sprite-url = $path-to-images + 'logo_sprite.svg?2014-01';
```
jezdez: I think this should be:

```
$path-to-assets = '../';
$path-to-images = $path-to-assets + 'img/';
```

I hope I'm not missing something, let's look at an example. There is this in compat-icons.styl:

```
url($path-to-assets + 'fonts/compat-icons.eot?-ehmwve')
```

The Stylus file will be compiled into `<STATIC_ROOT>/css/wiki-compat-tables.css` so eventually be accessible via `/static/css/wiki-compat-tables.css`. With the current code it'll have this hardcoded:

```
url("/media/fonts/compat-icons.eot?-ehmwve")
```

But we want to make the Stylus files completely unaware how the compiled files are hosted (as that's an implementation detail and outside the scope of what Stylus should know) and so we need to make all paths in the compiled files relative.

So with `$path-to-assets = '../';` we'll get this:

```
url("../fonts/compat-icons.eot?-ehmwve")
```

Which will work since the browser will just traverse the URL path up from the location of the compiled file at `/static/css/wiki-compat-tables.css`

openjck: Definitely. That's the plan down the line. At the moment there are some uses ([example](https://github.com/mozilla/kuma/blob/master/media/stylus/base/fonts/opensans.styl#L5-L6)) where `$path-to-assets` needs to be more explicit.

A great point though. I'll be sure to remember this once everything is moved to static.

jezdez: Ah, sorry, that was me not having the whole plan in my head and erring on the side of repeating, apologies.

openjck: No problem at all. It's a great point. I made a note to address that as part of the Stylus move.

openjck:
Anything else needed on this before we can merge?

jezdez:
Assigning back to @openjck after we talked about it on IRC with webops

openjck:
Blocked on WebOps [bug 1203631](https://bugzilla.mozilla.org/show_bug.cgi?id=1203631).

openjck:
[Bug 1203631](https://bugzilla.mozilla.org/show_bug.cgi?id=1203631) has been resolved, so this is all set.

openjck:
Jannis is out for the rest of the week. @groovecoder @robhudson @jwhitlock any takers? :smile:

openjck:
Gonna make one quick update to *vhosts.conf* before we merge.

openjck:
Updated. I updated *vhosts.conf* to redirect *directly* from `media/redesign/img` to `static/img`. I'll have the WebOps use the same redirect on production.

PR #3462, opened 2015-09-02
fix bug 1201251 - Ensure nav stays on screen on small devices

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201251


stephaniehobson:
R +1

PR #3463, opened 2015-09-03
fix bug 1201584 - Make footer links larger tap targets

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201584


stephaniehobson:
Better R+

PR #3464, opened 2015-09-03
fix bug 1201594 - Hide contributor bar on mobile

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201594


jwhitlock:
+r I'd rather not send the bytes at all, but this is a step forward for usefulness on mobile.

PR #3465, opened 2015-09-03
fix bug 1201604 - Hide CodePen and Stylus buttons upon print

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201604

### DIFF
--- a/media/stylus/wiki.styl
+++ b/media/stylus/wiki.styl
@@ -64,6 +64,7 @@ Styling for article content
 @require 'components/wiki/content/moreinfo';
 @require 'components/wiki/content/summary';
 @require 'components/wiki/customcss';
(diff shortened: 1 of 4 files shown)


media/stylus/wiki.styl
```diff
@@ -64,6 +64,7 @@ Styling for article content
 @require 'components/wiki/content/moreinfo';
 @require 'components/wiki/content/summary';
 @require 'components/wiki/customcss';
+@require 'components/wiki/samples';
```
stephaniehobson: Yes, separate file++++ 

Can we name the file open-in-host instead of samples, please.

stephaniehobson:
Works as expected R after nit.

darkwing:
Updated!

PR #3466, opened 2015-09-03
fix bug 1201612 - hide notification tray upon print

darkwing:


stephaniehobson:
OMG our print view of the home page could use some TLC.

This PR. This was a good call :)

R+!

PR #3468, opened 2015-09-03
bug 1177264 - Revision dashboard optimizations

groovecoder:
Superseding https://github.com/mozilla/kuma/pull/3467 with a better version, including:

aad6fce: Don't query for userbans unless it's a super-user looking at the dashboard (saves ~50 queries for non-super-users)

9fc6309: eliminate an expensive duplicate count query

### DIFF
--- a/kuma/dashboards/views.py
+++ b/kuma/dashboards/views.py
@@ -70,18 +70,10 @@ def revisions(request):
 
     if query_kwargs:
         revisions = revisions.filter(**query_kwargs)
-        total = revisions.count()
-    else:
-        # If no filters, just do a straight count(). It's the same
-        # result, but much faster to compute.
-        total = Revision.objects.count()
 
-    # Only bother with this if we're actually going to get
-    # some revisions from it. Otherwise it's a pointless but
-    # potentially complex query.
     revisions = paginate(request, revisions, per_page=PAGE_SIZE)
 
-    context = {'revisions': revisions, 'page': page, 'total': total}
+    context = {'revisions': revisions, 'page': page}
 
     # Serve the response HTML conditionally upon reques type
     if request.is_ajax():
--- a/kuma/dashboards/views.py
+++ b/kuma/dashboards/views.py
@@ -21,7 +21,8 @@ def revisions(request):
     filter_form = RevisionDashboardForm(request.GET)
     page = request.GET.get('page', 1)
 
-    revisions = (Revision.objects.select_related('creator')
+    revisions = (Revision.objects.prefetch_related('creator',
+                                                   'document')
                                  .order_by('-created')
                                  .defer('content'))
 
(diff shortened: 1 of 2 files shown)


groovecoder:
@jwhitlock - check this out?

kuma/dashboards/views.py
```diff
@@ -69,7 +70,6 @@ def revisions(request):
 
     if query_kwargs:
         revisions = revisions.filter(**query_kwargs)
-        total = revisions.count()
     else:
```
jwhitlock: Should the else: path go away now?

jwhitlock:
I'm usually against squashing commits, but at least the first three commits, if not all four, are yelling "squash me please"

When I look at DJDR, the majority of the query time is spent counting revisions twice.  I think the total calculation shouldn't be duplicated.

groovecoder:
Squashed.

Did you mean DjDT? d753e8b should remove the duplicate revision counting.

groovecoder:
Oh, I see what you mean - it was still double-counting for dashboard requests without any url param filters. Yeah, that's also fixed in d753e8b

jwhitlock:
Yes, I was looking at DJDT.

This code is strange.  They are doing a ``COUNT()``, which gets cached on the queryset so that the pagination doesn't need to reuse it.  However, the only use of ``total`` is to check if there are any records.  My guess is they are trying to work around some performance issue in Django's Paginator, but the performance issue is ``COUNT()`` on huge tables like this, and that's solved by using an estimate, or a different pagination method, or applying an indexed filter to the default page query.

Check out https://github.com/mozilla/kuma/commit/167fe92b6b9c - can you think of a reason not to eliminate the total entirely?

jwhitlock:
And, if you disagree with my modifications, :+1: on the PR as is.  This view needs a lot more help than a few modifications, but this is a step forward.

kuma/dashboards/views.py
```diff
@@ -69,16 +70,12 @@ def revisions(request):
 
     if query_kwargs:
         revisions = revisions.filter(**query_kwargs)
-        total = revisions.count()
-    else:
-        # If no filters, just do a straight count(). It's the same
-        # result, but much faster to compute.
-        total = Revision.objects.count()
 
     # Only bother with this if we're actually going to get
```
jezdez: Is this comment still accurate?

jwhitlock: Here's where the comment was added: https://github.com/mozilla/kuma/commit/39dd99f487e6c503e5dcddeb331f4121348c44b2

ubernostrum was attempting to short-circuit pagination if the requested page was past the total number of pages.  This was back when the revisions dashboard was implemented with JSON, rather than Django pagination.  The code in question is no longer executed, so there is no reason to efficiently count the number of revisions, filtered or not.

Django pagination calls ``COUNT(*)`` to get the total, and does its own work to validate page numbers.  The total item count is available from the page object.  I think adding total to the context is definitely a artifact from the old way of paginating this page, and should be removed in favor of the changes in https://github.com/mozilla/kuma/commit/167fe92b6b9c

groovecoder:
Updated to remove the stale comment and the superfluous template context.

jwhitlock:
Looks good to me. Bouncing to jezdez to see if he agrees the ``total`` stuff is now unneeded.

jezdez:
Yep, works for me, thanks <3

PR #3469, opened 2015-09-04
bug 1174209 - catch, log, and retry DNE exceptions

groovecoder:
Not my favorite way to brush this under the rug, but we're getting too many of these errors and emails.


groovecoder:
Like that?

jwhitlock:
Yep, perfect.  I haven't used retry before, so I want to confirm it does what I expect locally.

jwhitlock:
+r It retries every 120 seconds, and sends an email on 5 failures.  Also, I was able to replicate the missing revision on page save, and the task worked on first retry.

PR #3470, opened 2015-09-04
bug-1199840-style-move-error-message

stephaniehobson:
Fix Bug 1199840: white text on red background in move page error message.


darkwing:
R+, looks great!

PR #3472, opened 2015-09-05
Fix Bug 1201584 - footer links

stephaniehobson:
Line height for vertical spacing. Grammatical corrections to footer text.
Border to separate footer from content starting at tablet size when width no longer does it.

### DIFF
--- a/media/stylus/base/elements/sectioning.styl
+++ b/media/stylus/base/elements/sectioning.styl
@@ -29,40 +29,14 @@ main {
 }
 
 footer {
-
     background: #fff;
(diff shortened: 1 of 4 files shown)


darkwing:
So what I'm seeing at desktop width is "(bullet) About MDN" on the next line.  We should axe the bullet or bring the links back up...I think

stephaniehobson:
Re-wrote the PR to give us more fine-grain control. Marked it up as two paragraphs and a list. I think the list looks good on its own line with a leading bullet but I can remove it if it's still bugging you.

darkwing:
R+!   Wonderful!

PR #3473, opened 2015-09-07
Refactor wiki forms and other things I found while travelling

jezdez:


jezdez:
Just wanted to say that this makes the test output look like this:

```
(env)vagrant@developer-local:~/src$ ./manage.py test kuma
nosetests kuma --traverse-namespace --verbosity=1
Creating test database for alias 'default'...
...................................................................................................S...............................................................................................................................................................................................................................................................................................................................................S..........................S............................S..S............................................................................................................................................
----------------------------------------------------------------------
Ran 634 tests in 168.150s

OK (SKIP=5)
```

No weird log output anymore \o/

kuma/wiki/forms.py
```diff
@@ -327,63 +356,57 @@ def clean_current_rev(self):
...
-        return new_rev
+        Persists the revision and returns it.
+        Takes the view request and document of the revision.
+        Does some specific things when the revision is fully saved.
+        """
+        # have to check for first edit before we save
+        is_first_edit = request.user.wiki_revisions().count() == 0
 
+        # Popping the commit parameter here to make sure there are no surprises
+        kwargs.pop('commit', False)
```
jwhitlock: This feels wrong to me. Should we be finding the caller that is passing commit=True?  Is there some ```catch TypeError``` at some other level of code?  I'm willing to ignore for another day.

kuma/core/tests/test_views.py
```diff
@@ -1,40 +1,47 @@
...
         try:
-            response = self.client.get('/en-US/test_exception/')
-            eq_(500, response.status_code)
+            self.logger.handlers = [logging.NullHandler()]
+            response = self.client.get(self.suspicous_path)
+            eq_(response.status_code, 400)
             eq_(0, len(mail.outbox))
-        except:
-            pass
+        finally:
```
jwhitlock: Are you against ``setup()`` / ``teardown()``, or is there something that requires ``try``/``finally`` instead?

kuma/core/tests/test_views.py
```diff
@@ -1,40 +1,47 @@
...
 from django.core import mail
+from django.test import override_settings
+from django.utils.log import AdminEmailHandler
 
 from kuma.core.tests import KumaTestCase
 
 from ..urlresolvers import reverse
 
 
+@override_settings(
```
jwhitlock: Nice use of ``@override_settings``, and happy to see logging garbage out of test output

kuma/actioncounters/tests/tests.py
```diff
@@ -79,7 +80,7 @@ def test_bad_multiple_counters(self):
                                      user_agent=user_agent, user=user)
             u2.save()
             ok_(False, "This should have triggered an IntegrityError")
-        except:
+        except IntegrityError:
```
jwhitlock: +:100: for narrowing exception handling here and everywhere

kuma/wiki/views/revision.py
```diff
@@ -84,19 +84,15 @@ def compare(request, document_slug, document_locale):
     The ids are passed as query string parameters (to and from).
     """
     locale = request.GET.get('locale', document_locale)
+    if 'from' not in request.GET or 'to' not in request.GET:
+        raise Http404
```
jwhitlock: I'd use ``400 Bad Request``, but others say [422 Unprocessable Entity](http://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required) and others say other values and any change would change current behavior. Maybe ``404 Not Found`` makes sense because it's a query string parameter.  Probably shouldn't change unless Roy Fielding stops by and tells us we're wrong.

kuma/core/utils.py
```diff
@@ -57,7 +57,7 @@ def smart_int(string, fallback=0):
     """Convert a string to int, with fallback for invalid strings or types."""
     try:
         return int(float(string))
-    except (ValueError, TypeError):
+    except (ValueError, TypeError, OverflowError):
```
jwhitlock: Finally safe against ``smart_int("inf")`` (which is now zero)

kuma/authkeys/models.py
```diff
@@ -47,7 +46,9 @@ def generate_secret(self):
         return secret
 
     def check_secret(self, secret):
-        return hash_secret(secret) == self.hashed_secret
+        if not self.hashed_secret:
+            return False
+        return constant_time_compare(hash_secret(secret), self.hashed_secret)
```
jwhitlock: Finally safe against timing attacks

kuma/attachments/templates/attachments/includes/attachment_list.html
```diff
@@ -1,27 +1,34 @@
-{% set show_attach_button = allow_add_attachment %}
+{% set show_attach_button = allow_add_attachment_by(request.user) %}
```
jwhitlock: :+1: cleaner and centralized

kuma/authkeys/decorators.py
```diff
@@ -9,11 +9,12 @@
 
 
 def accepts_auth_key(func):
-    """Enable a view to accept an auth key via HTTP Basic Auth.
+    """
+    Enable a view to accept an auth key via HTTP Basic Auth.
```
jwhitlock: I think we'll need a style guide, because I'm not sure of the rules.  I've been following PEP 257, which allows one-line docstrings on one line, but you seem to correct them to three lines.  And here, I'm used to PEP 257's example of an empty line between the one-line docstring and the elaborate description.  However, I'm not sure what standard you're aiming for, so I feel I can't judge if you are moving toward the standard or not.

kuma/wiki/templates/wiki/create.html
```diff
@@ -53,6 +53,7 @@
           {% endif %}
         </ul>
         <input type="hidden" name="parent_topic" value="{{ parent_id | safe }}" />
+        {# FIXME check what's parentSlugPrefix doing here #}
```
jwhitlock: ``git log -SparentSlugPrefix`` finds: 80302fec85a9ae60963d50a246733f2d3869e66d (fix bug 753584).  My guess is some client-slide form validation was not implemented.

kuma/wiki/models.py
```diff
@@ -1295,52 +1295,16 @@ def show_toc(self):
...
-    @staticmethod
-    def from_url(url, required_locale=None, id_only=False):
-        """
-        Return the approved Document the URL represents, None if there isn't
-        one.
+        return reverse(endpoint, locale=self.locale, args=[self.slug])
 
-        Return None if the URL is a 404, the URL doesn't point to the right
-        view, or the indicated document doesn't exist.
+    def get_edit_url(self):
```
jwhitlock: :+1: makes a big difference in code constructing edit URLs

kuma/wiki/views/edit.py
```diff
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import json
```
jwhitlock: Nice catch, and we probably need to add pyflakes to our toolkit

kuma/wiki/templates/wiki/select_locale.html
```diff
@@ -8,9 +8,9 @@
 
     <h1>{{ _('Select a language to translate to:') }}</h1>
     <ul class="locales">
-      {% for locale in settings.LANGUAGES %}
-        {% if locale[0] != settings.WIKI_DEFAULT_LANGUAGE %}
-          <li><a href="{{ url('wiki.translate', document_path=document.slug)|urlparams(tolocale=locale[0]) }}"><bdi>{{ locale[1] }} ({{ locale[0] }})</bdi></a></li>
+      {% for code, name in settings.LANGUAGES %}
```
jwhitlock: :+1: easier to understand loop

kuma/wiki/forms.py
```diff
@@ -233,47 +250,55 @@ def __init__(self, *args, **kwargs):
...
+        # happens, changes to the slug are ignored for an iframe submissions
         if self.is_iframe_target:
-            # Since these collidables can change the URL of the page, changes
-            # to them are ignored for an iframe submission
-            return getattr(self.instance.document, name)
+            return self.instance.document.slug
 
-        error_message = {'slug': SLUG_COLLIDES}.get(name, OTHER_COLLIDES)
+        # first check if the given slug doesn't contain slashes and other
+        # charaters not allowed in a revision slug component (without parent)
```
jwhitlock: s/charaters/characters/

kuma/wiki/forms.py
```diff
@@ -233,47 +250,55 @@ def __init__(self, *args, **kwargs):
...
+
+        # edits can come in without a slug, so default to the current doc slug
         try:
-            existing_doc = Document.objects.get(
-                locale=self.instance.document.locale,
-                **{name: value})
+            if not self.cleaned_data['slug']:
+                document_slug = self.instance.document.slug
+                self.cleaned_data['slug'] = self.instance.slug = document_slug
+        except ObjectDoesNotExist:
```
jwhitlock: Which line(s) could raise ``ObjectDoesNotExist``?

kuma/wiki/forms.py
```diff
@@ -233,47 +250,55 @@ def __init__(self, *args, **kwargs):
...
+            pass
+
+        # then if there is a parent document we prefix the slug with its slug
+        if self.parent_slug:
+            self.cleaned_data['slug'] = u'/'.join([self.parent_slug,
```
jwhitlock: The [standard way](https://docs.djangoproject.com/en/1.7/ref/forms/validation/#cleaning-a-specific-field-attribute) to do this is to return the changed slug, rather than manipulate ``cleaned_data['slug']`` directly, and let Django take care of storing the cleaned value.

kuma/wiki/views/utils.py
```diff
@@ -36,22 +33,17 @@ def split_slug(slug):
...
 
-    return {
-        'specific': specific,
-        'parent': parent,
-        'full': slug,
-        'parent_split': slug_split,
-        'length': length,
-        'root': root,
-        'seo_root': seo_root,
+    return {                         # with this given: "some/kind/of/Path"
```
jwhitlock: :+1: hard-won comments

kuma/wiki/views/utils.py
```diff
@@ -36,22 +33,17 @@ def split_slug(slug):
...
+        'seo_root': seo_root,        # 'some'
     }
 
 
-def join_slug(parent_split, slug):
-    parent_split.append(slug)
-    return '/'.join(parent_split)
-
-
 @newrelic.agent.function_trace()
```
jwhitlock: :see_no_evil: gross code glad it's gone

darkwing: I WAS NEW TO PYTHON LEAVE ME ALONE

kuma/wiki/views/edit.py
```diff
@@ -167,11 +162,11 @@ def edit(request, document_slug, document_locale, revision_id=None):
...
                     # if must be here for section edits
                     if 'slug' in post_data:
-                        post_data['slug'] = join_slug(
-                            slug_dict['parent_split'], post_data['slug'])
+                        post_data['slug'] = u'/'.join([slug_dict['parent'],
+                                                       post_data['slug']])
 
                     # Get the possibly new slug for the imminent redirection:
-                    doc = doc_form.save(None)
+                    doc = doc_form.save(parent=None)
```
jwhitlock: :smile: This is why Python has named parameters - make this sort of thing easier to understand.

kuma/wiki/views/translate.py
```diff
@@ -193,36 +190,28 @@ def translate(request, document_slug, document_locale, revision_id=None):
...
-
-                    save_revision_and_notify(rev_form, request, doc)
-                    return redirect(doc)
+                parent_id = request.POST.get('parent_id', '')
+
+                # Attempt to set a parent
+                if parent_id:
+                    try:
+                        parent_doc = get_object_or_404(Document,
+                                                       id=parent_id)
```
jwhitlock: :nit: After dedent, ``get_object_or_404`` should fit on one line

kuma/wiki/views/utils.py
jwhitlock: :sunglasses: dude code not the coder

jwhitlock:
+r the code looks a lot better, and breaking this into small commits made it MUCH easier to review.  There are a few nitpicks and style questions, which may or may not result in changes.

kuma/wiki/forms.py
jezdez: Yeah, it's wrong, lemme find the culprit.

kuma/core/tests/test_views.py
jezdez: Ah, right, I can do this with those, lemme update..

kuma/wiki/views/revision.py
jezdez: FWIW the 404 was there already, but I get what you mean. 422 seems like an odd choice though as it has little to no precedence in Django land and comes from the sub protocol WebDAV. Also, I don't freaking know how the load balancer treats 422s, so 404 will have to do for now.

kuma/core/utils.py
jezdez: ;)

kuma/authkeys/models.py
jezdez: Yeah, that was a surprise when I saw that, tbh. Since it's the first thing you'd look for when building a token based system.

kuma/authkeys/decorators.py
jezdez: Ough, yeah. In general I like having a visual block on top of functions that moves the payload of the function away from the name. Also, one line docstrings are in my experience almost always a sign that docs are missing so I err on the side of promoting writing prose in there that just write short half sentences.

Since PEP 257 is ambiguous about this ("The summary line may be on the same line as the opening quotes or on the next line." https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) I'm following Django's example of putting the summary line one the next line as it was my school of Python style. I'm eager to follow that but I can understand if you're reluctant.

kuma/wiki/templates/wiki/create.html
jezdez: Aha, safe to remove apparently. Unless @darkwing objects?

kuma/wiki/models.py
jezdez: One day I want to centralize the URL construction for the domain objects so we can do `document.urls.edit()` or something, since we do *a lot* of URL reversal during runtime.

kuma/wiki/views/edit.py
jezdez: This was just catching up with a previous commit it seems. We do have flake8 in place that should show unneeded imports as errors in Travis.

kuma/wiki/forms.py
jezdez: `self.instance.document` may raise a `RelatedObjectDoesNotExist` which is constructed at runtime and I'd rather catch the base class: https://github.com/django/django/blob/1bbca7961cee20c4ddd453a7d74d316e84f4bbb5/django/db/models/fields/related.py#L433-L441

kuma/wiki/forms.py
jezdez: Fair enough, I can assign it to a separate variable, I agree with the standard way, wrote part of those docs after all. I think I got a bit engrossed in the way these forms were validated before :(

kuma/wiki/views/utils.py
jezdez: @darkwing It was amazing work at the time given constraints and time given to you. I'm impressed and will call you JavaScript *and* Python developer from now on.

kuma/wiki/views/translate.py
jezdez: Awww man, thanks for finding this, this makes me happy

jezdez:
So many thanks @jwhitlock for reviewing this. Can't stress this enough

groovecoder:
Die weird log output! :clap:

jwhitlock:
Thanks, @jezdez! And thanks to my review partner, http://www.emoji-cheat-sheet.com !

kuma/wiki/forms.py
```diff
@@ -327,63 +356,57 @@ def clean_current_rev(self):
             # If there's no document yet, just bail.
             return current_rev
 
-    def save_section(self, creator, document, **kwargs):
```
groovecoder: Looks like we're removing the section-editing functionality, but we've still left in some comments, logical paths (`if (self.section_id ...`) and tests for it?

groovecoder: Bah, now I see where it moved. Sorry.

kuma/wiki/models.py
groovecoder: :+1:

kuma/wiki/tasks.py
```diff
@@ -149,52 +150,60 @@ def move_page(locale, slug, new_slug, email):
...
-    Move was requested for document with slug %(slug)s in locale
-    %(locale)s, but no such document exists.
-    """ % {'slug': slug, 'locale': locale}
+            Move was requested for document with slug %(slug)s in locale
+            %(locale)s, but no such document exists.
+        """ % {'slug': slug, 'locale': locale}
         logging.error(message)
-        send_mail('Page move failed', message, settings.DEFAULT_FROM_EMAIL,
+        send_mail('Page move failed',
+                  textwrap.dedent(message),
```
groovecoder: :+1: `textwrap.dedent` to make the code **and** the email text more read-able!

kuma/wiki/models.py
```diff
@@ -1358,22 +1322,10 @@ def redirect_url(self):
...
                     if url.startswith(settings.SITE_URL):
                         return url
-                    elif (url[0] == '/' and url[1] != '/'):
+                    elif url[0] == '/' and url[1] != '/':
                         return url
-                elif (len(url) == 1 and url[0] == '/'):
+                elif len(url) == 1 and url[0] == '/':
                     return url
-                else:
-                    return None
```
groovecoder: We lost this `else` block that explicitly returned `None`? So the tests pass here because of an implicit return of `None` which feels kinda shaky for code that is meant to protect users against malicious redirects.

jezdez: I don't believe it's a problem since returning None is implied in Python functions and it's documented in the docstring.

kuma/wiki/tasks.py
jezdez: \o/

kuma/wiki/forms.py
jezdez: Yeah, I think we should remove it since it makes the editing flow still quite cumbersome. Have we agreed that saving sections only doesn't make sense? I've never seen the feature in the frontend, does it even exist there?

jezdez:
eee

jezdez:
Sorry, kid on the keyboard

settings_test.py
```diff
@@ -13,8 +13,15 @@
 
 DEMO_UPLOADS_ROOT = '/home/vagrant/uploads/demos'
 
-LOGGING['loggers']['django.db.backends.schema'] = {
-    'handlers': ['console'],
-    'propagate': True,
-    'level': 'WARNING',
-}
+LOGGING['loggers'].update({
+    'django.db.backends.schema': {
```
groovecoder: nit: I recently updated my `settings_test.py` to `django.db.backends` and not just `schema` because some other DB logs were cluttering my test output.

groovecoder:
r+ from me too! Looking great.

kuma/wiki/models.py
groovecoder: Okay so long as it's a language construct and not part of a library or something.

/me hates magic

kuma/wiki/forms.py
groovecoder: It did at a time but it was causing bugs. I'm fine to remove it all now. I have an idea for a front-end experiment that can validate demand for the feature in the future. If we get that validation, we can look at re-implementing it on this new and better code.

settings_test.py
jezdez: ok, I update it

jezdez:
@jwhitlock @grovecoder I've made the changes you proposed, do you want to take a look at it again?

kuma/wiki/forms.py
```diff
@@ -327,63 +356,58 @@ def clean_current_rev(self):
...
+        Persists the revision and returns it.
+        Takes the view request and document of the revision.
+        Does some specific things when the revision is fully saved.
+        """
+        # have to check for first edit before we save
+        is_first_edit = request.user.wiki_revisions().count() == 0
 
+        # Making sure we don't commit the saving right away since we
+        # want to do other things here.
+        kwargs['commit'] = False
```
jwhitlock: Works for me

kuma/core/tests/test_views.py
```diff
@@ -1,40 +1,47 @@
...
 
     def setUp(self):
-        self.old_logging = settings.LOGGING
+        super(LoggingTests, self).setUp()
+        self.old_handlers = self.logger.handlers[:]
 
     def tearDown(self):
-        settings.LOGGING = self.old_logging
+        super(LoggingTests, self).tearDown()
+        self.logger.handlers = self.old_handlers
```
jwhitlock: Thanks for moving to ``setUp``/``tearDown``

kuma/wiki/templates/wiki/create.html
```diff
@@ -52,9 +52,7 @@
             {% include 'wiki/includes/document_toc_field.html' %}
           {% endif %}
         </ul>
-        <input type="hidden" name="parent_topic" value="{{ parent_id | safe }}" />
-        <div class="hidden" id="parentSlugPrefix">{{ parent_slug | safe }}</div>
-
+        <input type="hidden" name="parent_topic" value="{{ parent_id|safe }}" />
```
jwhitlock: :+1: for removing cruft

kuma/wiki/forms.py
```diff
@@ -233,47 +250,55 @@ def __init__(self, *args, **kwargs):
...
+        # then if there is a parent document we prefix the slug with its slug
+        if self.parent_slug:
+            slug = u'/'.join([self.parent_slug, slug])
 
-        error_message = {'slug': SLUG_COLLIDES}.get(name, OTHER_COLLIDES)
         try:
-            existing_doc = Document.objects.get(
-                locale=self.instance.document.locale,
-                **{name: value})
+            doc = Document.objects.get(locale=self.instance.document.locale,
```
jwhitlock: I'm trying to wrap my brain around this try/except block, so bear with me.

- No existing ``Document`` in locale with the slug - ``Document.DoesNotExist`` is raised, slug is OK to use
- There is an existing ``Document`` with slug...
  - and ``Revison`` has a ``Document``...
    - and the existing ``Document`` is a redirect: slug was used in the past, but is OK to reuse
    - and the existing ``Document`` is not a redirect...
       - and does not match ``Revision``'s ``Document``: ``ValidationError(SLUG_COLLIDES)`` is raised
       - and matches ``Revision``'s ``Document``: slug is OK, colliding document is same document
  - and ``Revision`` doesn't exist yet: ``ValidationError(SLUG_COLLIDES)`` is raised
  - and ``Revision`` doesn't have a document yet: ``ValidationError(SLUG_COLLIDES)`` is raised
  - and accessing ``Revision``'s document raises a ``Document.DoesNotExist`` exception: slug is OK :interrobang:

It feels like the last case is a (pre-existing) bug, if ``self.instance.document`` can raise a ``Document.DoesNotExist`` error.  Which maybe it doesn't.  If it is a potential issue, then we'd need something like:

```python
        # Is the slug in use?
        try:
            doc = Document.objects.get(locale=self.instance.document.locale,
                                       slug=slug)
        except Document.DoesNotExist:
            # No existing document for this value, so we're good here.
            pass
        else:
            if not doc.get_redirect_url():
                rev_doc = None
                try:
                    rev_doc = self.instance and self.instance.document
                except Document.DoesNotExist:
                    pass

                if rev_doc is None:
                    # This document-and-revision doesn't exist yet, so there
                    # shouldn't be any collisions at all.
                    raise forms.ValidationError(SLUG_COLLIDES)
                elif doc.pk != rev_doc.pk:
                    # There's another document with this value,
                    # and we're not a revision of it.
                    raise forms.ValidationError(SLUG_COLLIDES)
```

Even if ``self.instance.document`` can't raise an exception, I like my except blocks as tight as possible.

jwhitlock:
+r on updated PR, with possible exception of try/except block looking for slug collisions

kuma/wiki/forms.py
jezdez: `self.instance.document` can raise a `RelatedObjectDoesNotExist` which is a subclass of `AttributeError` and `ObjectDoesNotExist`. I think that should only ever happen if we have faulty data in the database, when a revision didn't get a document assigned for some reason

jwhitlock:
Talked to @jezdez about the try/except block last Friday.  If there is a bug, then it's an existing bug, and would appear as two ``Document`` objects with the same ``locale`` and ``slug``.  I wasn't able to find any in my local database.  @jezdez, assign back to me when you're happy with the code, or go ahead an merge.

jezdez:
Ok, thanks @jwhitlock!

PR #3474, opened 2015-09-08
fix bug 1197779 - restore username to UserEditForm

groovecoder:
This fixes the bug in the site UI, but it seems to cause errors for the tests that test the edit view and I can't figure out why. :cry: Need @jwhitlock, @jezdez, or maybe even @robhudson to see if/why I'm crazy here?


jwhitlock:
Much more work needed.  See #3476

PR #3475, opened 2015-09-08
bug 1194461 - Make home promo a data URI

darkwing:
Image is only used once and since the homepage has its own stylesheet it's worth saving the request for a little bit more of CSS load.


stephaniehobson:
Sure.

PR #3478, opened 2015-09-09
fix bug 1203230 - Don't re-prism the prisms

darkwing:
Sample doc html;  https://gist.github.com/darkwing/c395a73fbfcb1cc650c8


openjck:
Nice!

PR #3480, opened 2015-09-10
Fix bug 1203625 - Prevent JS error in edit interface

darkwing:
Currently the most frequent JS error on the site, which doesn't happen very often.  \o/  But hey, let's fix it.

https://bugzilla.mozilla.org/show_bug.cgi?id=1203625


openjck:
I got the error to appear once, but now I'm having trouble reproducing it. Any tips?

PR #3482, opened 2015-09-15
Bug 1188029 - Spam prevention via Akismet

jezdez:


jezdez:
This will break horribly right now and should only considered to be a state in between..

kuma/spam/forms.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, request, *args, **kwargs):
+        self.request = request
+        self.akismet = akismet.Akismet()
+        super(AkismetFormMixin, self).__init__(*args, **kwargs)
+
+    def akismet_parameters(self):
+        """
+        When using this mixin make sure you implement this method,
+        get the parent class' value and and return a dictionary of
```
groovecoder: "value and and return"

groovecoder:
Only 1 tiny nit-pick so far. Looking really good - great work @jezdez!

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+        return ('%s response, debug help: %s, full response: %s' %
+                (self.status_code, self.debug_help, self.response.text))
+
+
+class Akismet(object):
+    """
+    The Akismet client implementation wrapping a request session and
+    having API specific methods for easy spam and ham handling.
+    """
+    submission_success = 'Thanks for making the web a better place.'
```
robhudson: l10n?

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Required:
+
+        :param str user_ip: The IP address of the user submitting the content
+        :param str user_agent: The user agent of the user submitting the data
+
+        Optional:
+
+        :param str referrer: The content of the HTTP_REFERER header should
+                             be sent here.
```
robhudson: You spelled `HTTP_REFERER` wrong. :see_no_evil: :smiling_imp: :trollface:

kuma/spam/forms.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        When using this mixin make sure you implement this method,
+        get the parent class' value and return a dictionary of
+        parameters matching the ones of the Akismet.check_comment method.
+
+        Use the self.instance or self.request variables to build it.
+        """
+        if not hasattr(self, 'cleaned_data'):
+            raise forms.ValidationError(
+                'The form data has not yet been validated. Please try again'
```
robhudson: Let's end this sentence with a period.

robhudson: Also, l10n.

kuma/spam/models.py
```diff
@@ -0,0 +1,18 @@
+from django.conf import settings
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
```
robhudson: I prefer the habit of importing this as `_lazy`. Otherwise you always have to look at how `_` is aliased to know whether it is the lazy one or not.

kuma/wiki/forms.py
```diff
@@ -356,14 +392,77 @@ def clean_current_rev(self):
...
+        # wrapping this in a try/finally to make sure that even if
+        # creating a spam attempt object fails we call the parent
+        # method that raises a ValidationError
+        try:
+            DocumentSpamAttempt(
+                title=self.cleaned_data['title'],
+                slug=self.cleaned_data['slug'],
+                user=self.request.user,
+                document=document,
+            ).save()
```
robhudson: Why not `DocumentSpamAttempt.objects.create(...)`? I'm asking honestly, not suggesting one over the other.

jezdez: I had that before and it hid an `IntegrityError`, I can turn this again now.

kuma/spam/akismet.py
jezdez: :laughing:

kuma/spam/akismet.py
jezdez: Nah, that's the message coming *from* Akismet that we use to check for success.

kuma/spam/forms.py
jezdez: Yeah, that makes sense.

kuma/spam/models.py
jezdez: I'm fairly strongly opinionated about `_` being an alias for the lazy version of the gettext function since it works for clases – both import time (e.g. class attributes like model fields or in views classes) and process time (e.g. inside of view functions). It's a best practice that I've adopted from Django and other projects I've worked on. If – *and only if* – the non-lazy version of gettext is needed it's good to import it explicitly. I consider the current way how Mozilla project use "_lazy" to be an antipattern.

robhudson: I like that too. As long as it is standard across the project so we don't have to double check the alias everywhere.

jezdez:
@groovecoder @robhudson @jwhitlock This would be ready to be looked at if you have time.

jwhitlock:
I would support a quick merge of the first 4 commit (the cleanup commits), if you wanted to open a new PR.

kuma/spam/akismet.py
jwhitlock: Could you add a comment that this is the [documented response](http://akismet.com/development/api/#submit-spam), so that an enthusiastic future dev doesn't localize it?

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @property
+    def url(self):
+        return 'https://%s.rest.akismet.com/1.1/' % self.key
+
+    def send(self, method, user_ip='', user_agent='', **payload):
+        payload.update({
+            'blog': self.domain,
+            'user_ip': user_ip,
+            'user_agent': user_ip,
```
jwhitlock: This should be ``'user_agent': user_agent``.

jezdez: Oh man, good catch!

kuma/spam/akismet.py
jezdez: I shall!

jezdez:
@jwhitlock Sure, coming right up..

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+        except AkismetError:
+            # just take a shortcut when there was an
+            # error accessing Akismet
+            self.ready = False
+
+    @property
+    def url(self):
+        return 'https://%s.rest.akismet.com/1.1/' % self.key
+
+    def send(self, method, user_ip='', user_agent='', **payload):
```
jwhitlock: It looks like ``user_ip`` and ``user_agent`` are optional so that ``verify_key`` will work.  Should they be added to the payload in this case?  Maybe they should be removed from ``send`` and be explicitly added as kwargs in the other methods, or only added when non-empty strings.

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+        An API call method to test the provided API key.
+
+        This is automatically called when the Akismet client is
+        instantiated so that the app code can decide if it's usable right
+        away.
+        """
+        # quick bailout when the key isn't given
+        if not self.key:
+            return False
+        response = self.send('verify-key', **{'key': self.key})
```
jwhitlock: How about ``self.send('verify-key', key=self.key)`` instead?

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+        try:
+            return {'valid': True, 'invalid': False}[response.text]
+        except KeyError:
+            self.error(response)
+
+    def check_comment(self, user_ip, user_agent, **optional):
+        """
+        Submits content to Akismet to check if it's spam or not.
```
jwhitlock: :star: gold star for complete doc string

kuma/spam/forms.py
```diff
@@ -0,0 +1,63 @@
+from django import forms
+from tower import ugettext_lazy as _
+
+from . import akismet
+
+
+class AkismetFormMixin(object):
```
jwhitlock: :star: gold star for good doc strings, nice override and integration points

kuma/spam/tests/test_akismet.py
```diff
@@ -0,0 +1,151 @@
+from django.test import SimpleTestCase
+from constance.test import override_config
+
+import requests
+import responses
```
jwhitlock: :open_mouth: I did not know about ``responses``.  Super useful

kuma/spam/akismet.py
```diff
@@ -0,0 +1,174 @@
...
+            'user_agent': user_ip,
+        })
+        url = urlparse.urljoin(self.url, method)
+        try:
+            return self.session.post(url, data=payload)
+        except exceptions.RequestException:
+            # let this exception bubble up for now since the mail_admins
+            # log handler and Sentry will let us know about this
+            raise
+
```
jwhitlock: Should we remove the ``try``/``except`` block, since it does nothing?  Or add a test that raises an exception?

kuma/spam/models.py
```diff
@@ -0,0 +1,18 @@
...
+    user = models.ForeignKey(settings.AUTH_USER_MODEL)
+    created = CreationDateTimeField(_('created'), db_index=True)
+
+    class Meta:
+        get_latest_by = 'created'
+        ordering = ('-created',)
+        abstract = True
+
+    def __unicode__(self):
+        return self.created
```
jwhitlock: This is both 1) not used, and 2) doesn't return a unicode string

kuma/spam/akismet.py
jezdez: Yeah, let's add an exception raising test and a TODO for later if/when we move away from Sentry

kuma/spam/models.py
jezdez: :facepalm:

kuma/spam/tests/test_akismet.py
jezdez: It is! Basically just a convenience wrapper around mock, but useful nevertheless

.gitmodules
```diff
@@ -164,3 +164,9 @@
 [submodule "vendor/src/django-sundial"]
 	path = vendor/src/django-sundial
 	url = https://github.com/charettes/django-sundial
+[submodule "vendor/src/responses"]
+	path = vendor/src/responses
+	url = https://github.com/getsentry/responses.git
+[submodule "vendor/src/cookies"]
+	path = vendor/src/cookies
+	url = https://github.com/sashahart/cookies.git
```
jwhitlock: https://github.com/sashahart/cookies has a note that the project moved to https://gitlab.com/sashahart/cookies.  Should we use the gitlab URL?

.gitmodules
```diff
@@ -164,3 +164,9 @@
 [submodule "vendor/src/django-sundial"]
 	path = vendor/src/django-sundial
 	url = https://github.com/charettes/django-sundial
+[submodule "vendor/src/responses"]
+	path = vendor/src/responses
+	url = https://github.com/getsentry/responses.git
+[submodule "vendor/src/cookies"]
```
jwhitlock: This project lacks a ``.gitignore`` line for ``*.pyc``.  Can you add ``ignore = untracked`` so that ``cookies.pyc`` won't show up in ``git status``?

.gitmodules
jezdez: Ah, huh, that must have been recent. Updating..

.gitmodules
jezdez: Sure

kuma/spam/akismet.py
jezdez: Yep!

kuma/spam/akismet.py
jezdez: Ah, ok, I'll take another look and will get back to you

jwhitlock:
+rwc This is really high-quality code for a new feature.  My impression is that kuma/spam is being written for spin-off into a separate library, and it would be a pretty good one.

kuma/spam/akismet.py
jezdez: I've thought about this a bit more and think it actually needs to pass without raising the exception since the premise is that content submission won't be hindered if there is a connectivity problem. Instead I'll record the exception with New Relic's agent and move on.

jezdez:
@jwhitlock So I've applied all the changes you asked for, but had to make some changes in the form API to better handle cases in which an Akismet client instance become unavailable. The readiness is now calculated until a successful verification of the key happened. Also, I now record requests exceptions via New Relic instead of bubbling it up since there was a requirement in the spec to never fail due to Akismet API issues.

kuma/spam/akismet.py
```diff
@@ -0,0 +1,215 @@
...
+    def ready(self):
+        """"
+        Returns whether this client is usable by verifying the API key.
+        """
+        if self._verified is None:
+            self._verified = self.verify_key()
+            if self._verified is None:
+                return False
+        return self._verified
+
```
jwhitlock: How about ``@cached_property``?

jwhitlock:
+rwc.  Signed up for free Kismet key, added to constance.  Tested ``Akismet.ready``, ``.check_comment`` (returns ``True`` if spam), and ``.submit_spam``.  Editing a document still works, but not 100% sure that Akismet was in the loop.

kuma/spam/akismet.py
jezdez: The property needs to be able to evaluated multiple time, e.g. when the key verification doesn't work the first time for some reasons (e.g. connectivity issue) the result will be `None` and by that tricker a retry of the verification on the next request (form instance).

jezdez:
I've added a new waffle switch called `spam_checks_enabled` that is *off* by default. It's the kill switch for all (future) spam checks, including the Akismet one we're adding here.

jezdez:
@jwhitlock Yet another round of checking :-/

kuma/spam/forms.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def akismet_enabled(self):
+        """
+        Decides whether to even check for spam during the form validation.
+
+        Checks the API client if it's ready by default.
+        """
+        return (self.akismet_client.ready and
+                waffle.switch_is_active(SPAM_CHECKS_SWITCH))
+
```
jwhitlock: The order should be reversed - check if the switch is active, then make a network request to the Akismet service

jwhitlock:
+rwc

Practical testing - Akismet is not checked when ``spam_checks_enabled`` is not preset or not active (although it does check that the akismet key is valid).  When both ``spam_checks_enabled`` is active and the flag ``wiki_spam_exempted`` is inactive, the content is sent to Akismet.

It appears that adding a spammy link to a good page will not trigger an Akismet rejection, though.  This may require training, or other facts (posting from localhost, an @mozilla.com account) override the spammy text. I for one welcome our new arbitrary robot overlords.

kuma/spam/forms.py
jezdez: Good catch!

PR #3483, opened 2015-09-15
fix bug 1174498 - error message for revert dupes

groovecoder:
When a user reverts a document to an old revision, we want to catch the
IntegrityError if it collides with the document's redirect, so the user knows
they can fix it by deleting the redirect.

### DIFF
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -2469,6 +2469,21 @@ class DocumentEditingTests(UserTestCase, WikiTestCase):
         ok_(mock_kumascript_get.called,
             "kumascript should have been used")
 
+    def test_revert_moved(self):
+        doc = document(slug='move-me', save=True)
+        rev = revision(document=doc, save=True)
+        prev_rev_id = rev.id
+        doc._move_tree('moved-doc')
+        self.client.login(username='admin', password='testpass')
+
+        resp = self.client.post(reverse('wiki.revert_document',
+                                        args=[doc.slug, prev_rev_id],
+                                        locale=doc.locale),
+                                follow=True)
+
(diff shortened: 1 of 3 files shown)


groovecoder:
@robhudson - can you check me on this?

kuma/wiki/views/delete.py
```diff
@@ -34,7 +35,21 @@ def revert_document(request, document_path, revision_id):
...
+        except IntegrityError:
+            return render(
+                request,
+                'wiki/confirm_revision_revert.html',
+                {
+                    'revision': revision,
+                    'document': revision.document,
+                    'error': "Document already exists. Note: You cannot "
+                             "revert a document that has been moved until you "
+                             "delete its redirect."
```
robhudson: l10n on this string? Also don't need double quotes.

robhudson:
Could/Should we unit test this change?

groovecoder:
Updated with `_()` and a test.

kuma/wiki/tests/test_views.py
```diff
@@ -2469,6 +2469,21 @@ def test_revert(self, mock_kumascript_get):
...
+        doc = document(slug='move-me', save=True)
+        rev = revision(document=doc, save=True)
+        prev_rev_id = rev.id
+        doc._move_tree('moved-doc')
+        self.client.login(username='admin', password='testpass')
+
+        resp = self.client.post(reverse('wiki.revert_document',
+                                        args=[doc.slug, prev_rev_id],
+                                        locale=doc.locale),
+                               follow=True)
```
robhudson: shift over 1 more space

groovecoder: :rage: Fixed.

PR #3487, opened 2015-09-16
bug 1196365 - Add OpenSans woff2 files

darkwing:
To test you may need to do a search/replace of `'//mozorg.cdn.mozilla.net' + `


darkwing:
Once https://github.com/mozilla/kuma/pull/3501/files#diff-5b5d9b0aeaf130e8bfe7810e6831cfcaR19 is merged, we should add woff2 to that font list extension

stephaniehobson:
This branch name made my day :dog:

stephaniehobson:
Looking good in FF, Chrome, and Safari iOS. Can't test in IE any more for some reason. We'll have to test when it gets pushed to staging.

Do we need to do anything special to make sure the files get to the CDN?

R+

stephaniehobson:
Oh yeah, there's a thought...

Do we need to add woff2 to this too? https://github.com/mozilla/kuma/blob/8f3241ddf5563d235f319fda51701825ab977f1e/media/fonts/.htaccess

darkwing:
Updated!

stephaniehobson:
Working when serving locally. R+

Falls back nicely when looking to the CDN for the files that aren't there yet. I think we're good to go. Do we need to do anything special to get the fonts to the CDN?

openjck:
I'd love to help get this in before moving the fonts to kuma/static. I don't think anything special needs to be done to get the fonts on the CDN.

PR #3488, opened 2015-09-16
fix bug 1205381 - Remove the demos block from the homepage

darkwing:
PERF PERF PERF


openjck:
Awesome! We can also get rid of [line 7](https://github.com/mozilla/kuma/blob/42ab101ca1a5cbb16eeb6f7e3bddb90ecae61e4e/kuma/landing/views.py#L7), [line 15](https://github.com/mozilla/kuma/blob/42ab101ca1a5cbb16eeb6f7e3bddb90ecae61e4e/kuma/landing/views.py#L15), and [line 30](https://github.com/mozilla/kuma/blob/42ab101ca1a5cbb16eeb6f7e3bddb90ecae61e4e/kuma/landing/views.py#L30) of *kuma/landing/views.py*.

darkwing:
Good call, forgot the view file -- will do!

openjck:
Can we also safely remove this?

https://github.com/mozilla/kuma/blob/42ab101ca1a5cbb16eeb6f7e3bddb90ecae61e4e/kuma/demos/templates/demos/elements/submission_thumb.html#L1-L4

PR #3490, opened 2015-09-16
Bug 1197444: Use same expires headers for static/

openjck:


openjck:
Weird, not seeing an expires header locally. Any idea @groovecoder?

PR #3494, opened 2015-09-16
Bug 957802: Allow all hosts in settings_local.py

openjck:
The ALLOWED_HOSTS setting needs to be set to disable debug mode and
emulate production locally.

### DIFF
--- a/provisioning/roles/kuma/files/vagrant/settings_local.py
+++ b/provisioning/roles/kuma/files/vagrant/settings_local.py
@@ -3,6 +3,9 @@ import logging
 
 INTERNAL_IPS = ('127.0.0.1', '192.168.10.1',)
 
+# ALLOWED_HOSTS must be set whenever DEBUG = False
+ALLOWED_HOSTS = '*'
+
 DEBUG = True
 DEV = True
 TEMPLATE_DEBUG = DEBUG


provisioning/roles/kuma/files/vagrant/settings_local.py
```diff
@@ -3,6 +3,8 @@
 
 INTERNAL_IPS = ('127.0.0.1', '192.168.10.1',)
 
+ALLOWED_HOSTS = '*'
+
```
robhudson: Django docs state, "When DEBUG is True or when running tests, host validation is disabled; any host will be accepted. Thus it’s usually only necessary to set it in production."

Is this causing a problem somewhere?

groovecoder: Yesterday I was trying to run with `DEBUG = False` to re-create production behavior and I hit this.

openjck: That's interesting. Behavior doesn't match the documentation for me either. I get this error from runserver when `DEBUG = False` and there are no allowed hosts defined.

```
14:30:53 web.1        | CommandError: You must set settings.ALLOWED_HOSTS if DEBUG is False.
14:30:53 web.1        | exited with code 1
14:30:53 system       | sending SIGTERM to all processes
```

robhudson:
r+. But perhaps with a code comment to explain this is needed if `DEBUG = False`?

openjck:
Updated!

PR #3497, opened 2015-09-17
bug 1205667 - get_tidied_content can return blank

groovecoder:
When a $compare request is made for a large revision,
we want to skip tidy_content and return a warning to the user,
so we don't block requests on the expensive tidy operation.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -1714,11 +1714,15 @@ class Revision(models.Model):
         """Convenience method to extract the content for a single section"""
         return self.document.extract_section(self.content, section_id)
 
-    def get_tidied_content(self):
+    def get_tidied_content(self, allow_none=False):
         """
         Return the revision content parsed and cleaned by tidy.
 
-        This will first check for a denormalized field,
+        First, check in denormalized db field. If it's not available, schedule
+        an asynchronous task to store it.
+
+        allow_none -- To prevent CPU-hogging calls, return None instead of
+                      calling tidy_content in-process.
         """
         # we may be lucky and have the tidied content already denormalized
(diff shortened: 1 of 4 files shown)


kuma/wiki/models.py
```diff
@@ -1735,7 +1735,10 @@ def get_tidied_content(self):
                 # already and don't need to schedule it the next time
                 # we use 3 days as a limit to try it again
                 memcache.set(tidying_scheduled_cache_key, 1, 60 * 60 * 24 * 3)
-            tidied_content, errors = tidy_content(self.content)
+            if allow_blank:
+                tidied_content = ''
```
jwhitlock: I'm worried that an empty string is a valid tidied content.  I'd rather return something non-string, like ``None`` (changing param to ``allow_null`` or ``async``), to specify that no content was returned.

kuma/wiki/models.py
```diff
@@ -1714,7 +1714,7 @@ def get_section_content(self, section_id):
         """Convenience method to extract the content for a single section"""
         return self.document.extract_section(self.content, section_id)
 
-    def get_tidied_content(self):
+    def get_tidied_content(self, allow_blank=False):
```
jwhitlock: docstring could use update for new functionality (and to remove that trailing comma after ``field,`` :grimacing:

jwhitlock:
Oh, and test coverage :smile:

groovecoder:
Updated with changes and tests.

jwhitlock:
+r was able to get the "Server is rendering the content comparison. Please refresh this page in a few minutes." message, refreshed and got it again, and on third refresh got the content comparison.

PR #3499, opened 2015-09-17
fix bug 1205750 - Limit precision of REM units

darkwing:
No more .32894732943rem


stephaniehobson:
Does what it says on the tin :+1:

PR #3500, opened 2015-09-17
Add "make coverage" target

jwhitlock:
Add a 'make coverage' target.  If run from the host, it enters VM, runs tests with coverage, builds HTML coverage docs, and opens them in the browser.  If run from the VM, it runs coverage and prints coverage to console.

Opening browser works on OS X, needs confirmation on desktop Linux.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,4 +3,4 @@ source = kuma
 branch = 1
 
 [report]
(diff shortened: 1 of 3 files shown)


robhudson:
Looks and works well for me.

PR #3501, opened 2015-09-17
bug 957802 - serve fonts to local.allizom for devs

groovecoder:
spot-check: `vagrant provision` and then check a page that tries to load `compat-icons.woff`: the request should succeed instead of a 403.

PR #3504, opened 2015-09-19
bug 1206309 - waffled api driven compat tables

stephaniehobson:
Display api-driven compat tables to users with a waffle enabled on pages that have been identified as ready.

Testing:
1) download PR
2) `./manage.py compress_assets`
3) create waffle flag named `compat-api` and enable it for yourself
4) make/update a local copy of: https://developer.allizom.org/en-US/docs/Template%3AEmbedCompatTable
5) embed the macro and call it `{{EmbedCompatTable('web-css-background')}}`
6) verify that new table appears
7) disable waffle flag
8) verify that new table does not appear

This PR does not hide the old table yet. For debugging/testing purposes it is useful to see both side by side. 

Skateboard for @lonnen:

### DIFF
--- a/media/js/wiki-compat-tables.js
+++ b/media/js/wiki-compat-tables.js
@@ -1,5 +1,7 @@
 (function(win, doc, $) {
 
+    $('#compat-api').removeClass('hidden');
+
     // Private var to assign IDs to history for accessibility purposes
     var historyCount = 0;
 
--- a/media/js/wiki.js
+++ b/media/js/wiki.js
@@ -304,8 +304,12 @@
         Load and build compat tables if present in page
     */
     (function() {
-        var $compatTables = $('.bc-table');
+        // don't run if no compat table on page
+        var $compatTables = $('#compat-api table');
         if(!$compatTables.length) return;
+
+        // don't run if waffle not active
+        if(!win.waffle || !win.waffle.flag_is_active('compat_api')) return;
 
         $('<link />').attr({
                 href: mdn.mediaPath + 'css/wiki-compat-tables-min.css',


stephaniehobson:
Github skateboard gif fail. Trying again:

stephaniehobson:
I think this is github's gif fail, not mine. Perhaps it will appear later.

renoirb:
The gif apears :)

I'll try to test it myself
Bouncing ping-pong balls

darkwing:
Sample doc code?  I think it would be helpful if we had the same base

stephaniehobson:
https://gist.github.com/stephaniehobson/8b6039f8f167c782d7a3

darkwing:
This worked for me outside of:

1.  Our convention for waffle flags is `_`, not `-`
2.  Images and fonts aren't showing up for me

stephaniehobson:
Updated to `_`

#3501 should take care of the fonts and I think a collect static should fix the missing images.

stephaniehobson:
`-` in CSS/JS and `_` in the waffle flag.

stephaniehobson:
I'm 1/2 through a companion PR, I'm just going to combine the two.

renoirb:
I could not do the full run with an empty kuma database yet. There's many KumaScript templates to bring up manually. <aside>It makes me wonder how you systematically deploy changes between production and other environments. So many manual steps, so many opportunities for manipulation mistake.</aside>

stephaniehobson:
We very rarely tie kumascript to the files stored in kuma. This one is unusual in a lot of ways. The writers usually test their kumascript changes on staging (which has the database updated every month or so) and then copy to production when they're ready but there are no other dependencies on the kumascripts they write.

PR #3505, opened 2015-09-21
fix bug 1182210 - Store bitly URLs on model

robhudson:


kuma/core/tests/test_utils.py
```diff
@@ -21,3 +25,26 @@ def test_empty_string(self):
     def test_wrong_type(self):
         eq_(0, smart_int(None))
         eq_(10, smart_int([], 10))
+
+
+@override_settings(USE_SHORT_URLS=True)
+class BitlyTestCase(KumaTestCase):
```
jezdez: The `KumaTestCase` isn't needed here and we can just use Django's standard test class

kuma/wiki/migrations/0008_auto_20150921_1154.py
```diff
@@ -0,0 +1,32 @@
...
...), (b'ha', 'Hausa'), (b'he', '\u05e2\u05d1\u05e8\u05d9\u05ea'), (b'hi-IN', '\u0939\u093f\u0928\u094d\u0926\u0940 (\u092d\u093e\u0930\u0924)'), (b'hr', 'Hrvatski'), (b'hu', 'magyar'), (b'id', 'Bahasa Indonesia'), (b'ig', 'Igbo'), (b'it', 'Italiano'), (b'ja', '\u65e5\u672c\u8a9e'), (b'ka', '\u10e5\u10d0\u10e0\u10d7\u10e3\u10da\u10d8'), (b'ko', '\ud55c\uad6d\uc5b4'), (b'ln', 'Ling\xe1la'), (b'mg', 'Malagasy'), (b'ml', '\u0d2e\u0d32\u0d2f\u0d3e\u0d33\u0d02'), (b'ms', 'Melayu'), (b'my', '\u1019\u103c\u1014\u103a\u1019\u102c\u1018\u102c\u101e\u102c'), (b'nl', 'Nederlands'), (b'pl', 'Polski'), (b'pt-BR', 'Portugu\xeas (do\xa0Brasil)'), (b'pt-PT', 'Portugu\xeas (Europeu)'), (b'ro', 'rom\xe2n\u0103'), (b'ru', '\u0420\u0443\u0441\u0441\u043a\u0438\u0439'), (b'son', 'So\u014bay'), (b'sq', 'Shqip'), (b'sv-SE', 'Svenska'), (b'sw', 'Kiswahili'), (b'ta', '\u0ba4\u0bae\u0bbf\u0bb4\u0bcd'), (b'th', '\u0e44\u0e17\u0e22'), (b'tl', 'Tagalog'), (b'tn', 'Setswana'), (b'tr', 'T\xfcrk\xe7e'), (b'vi', 'Ti\u1ebfng Vi\u1ec7t'), (b'wo', 'Wolof'), (b'xh', 'isiXhosa'), (b'yo', 'Yor\xf9b\xe1'), (b'zh-CN', '\u4e2d\u6587 (\u7b80\u4f53)'), (b'zh-TW', '\u6b63\u9ad4\u4e2d\u6587 (\u7e41\u9ad4)'), (b'zu', 'isiZulu')]),
```
jezdez: I think it would be nicer to have thise locale related change in a separate migration

kuma/wiki/models.py
```diff
@@ -1467,6 +1473,32 @@ def contributors(self):
     def zone_stack(self):
         return DocumentZoneStackJob().get(self.pk)
 
+    def set_share_url(self, save=True):
+        share_url = bitly_shorten('/'.join([settings.SITE_URL,
```
jezdez: Please use `urlparse.urljoin` here

kuma/wiki/models.py
```diff
@@ -1467,6 +1473,32 @@ def contributors(self):
...
+        if save and share_url:
+            self.share_url = share_url
+            if self.pk:
+                self.save(update_fields=['share_url'])
+            else:
+                self.save()
+
+    def get_share_url(self):
+        """
+        Returns the `share_url` or generates it if None.
```
jezdez: Inline code is two backticks in rST:

    ``share_url``

kuma/wiki/models.py
```diff
@@ -1467,6 +1473,32 @@ def contributors(self):
...
+
+        If not `settings.USE_SHORT_URLS`, we return the full URL.
+
+        """
+        if settings.USE_SHORT_URLS:
+            if self.share_url:
+                return self.share_url
+
+            self.set_share_url()
```
jezdez: let's add the else clause here for clarity

kuma/wiki/models.py
```diff
@@ -1467,6 +1473,32 @@ def contributors(self):
...
+
+        else:
+            return '/'.join([settings.SITE_URL, self.get_absolute_url()])
```
jezdez: urlparse.urljoin again please

kuma/wiki/templates/wiki/includes/document_macros.html
```diff
@@ -85,26 +85,22 @@
...
 
-    {% set short_share_url_twitter = share_url_twitter | bitly_shorten %}
-    {% set short_share_url_facebook = share_url_facebook | bitly_shorten %}
-    {% set short_share_url_gplus = share_url_gplus | bitly_shorten %}
-
-    {% set twitter_url = 'http://twitter.com/share'|urlparams(url=short_share_url_twitter, text=share_text) %}
-    {% set facebook_url = 'http://facebook.com/sharer.php?s=100&p[url]=' + short_share_url_facebook %}
-    {% set gplus_url = 'https://plus.google.com/share'|urlparams(url=short_share_url_gplus) %}
+    {% set twitter_url = 'http://twitter.com/share'|urlparams(url=share_url_twitter, text=share_text) %}
+    {% set facebook_url = 'http://facebook.com/sharer.php?s=100&p[url]=' + share_url_facebook %}
```
jezdez: I was wondering if we should replace `&` with `&amp;`?

kuma/core/utils.py
```diff
@@ -321,3 +326,15 @@ def get_unique(content_type, object_pk, name=None, request=None,
...
+
+def bitly_shorten(url):
+    if not settings.USE_SHORT_URLS:
+        return None
+
+    try:
+        return bitly.shorten(url)['url']
+    except (bitly_api.BitlyError, KeyError) as e:
+        # If bit.ly service fails or the API key isn't configured.
+        log.error('Bitly URL shortening failed: %s' % e)
```
jezdez: I think we should use `log.exception` here (this was a bug in the previous code as well). Without it we don't get the traceback object passed to the logger.

kuma/demos/helpers.py
```diff
@@ -345,3 +347,24 @@ def _contextual_locale(context):
     if not localedata.exists(locale):
         locale = settings.LANGUAGE_CODE
     return locale
+
+
+bitly = bitly_api.Connection(login=getattr(settings, 'BITLY_USERNAME', ''),
```
jezdez: Couldn't we reuse the core bitly client connection here?

kuma/wiki/models.py
```diff
@@ -920,6 +923,9 @@ def save(self, *args, **kwargs):
             # If this is a translation without a topic parent, try to get one.
             self.acquire_translated_topic_parent()
 
+        # Set the shortened URL (e.g. bit.ly).
+        self.set_share_url(save=False)
```
jezdez: I thought we'd run this as a Celery task to be honest to not make document saving slower and potentially blocking on a call to Bitly's API.

jezdez:
@robhudson looking good for the most part, comments and questions inline :)

Also, I'd like to suggest to push the branch to the main repo, as it has positive side effects on shared responsibility of the branch and allows easy collaboration in the code. It prevents having to close and reopen PRs when a handover is done over the course of the life of a branch.

darkwing:
w00t!  YES!

kuma/wiki/models.py
robhudson: I never understood when to use a single backtick and when to use a double backticks. It's not so much inline code as a reference to a variable.

kuma/wiki/templates/wiki/includes/document_macros.html
robhudson: I don't think so, I think that would break the URL.

kuma/core/utils.py
robhudson: :+1:

kuma/wiki/models.py
robhudson: We have 2 places where we're calling this. On save (here) and on document display if it's not set already. Would we want tasks for both?

On save it seems like we aren't too concerned with speed but we're also not concerned with a delay in getting the bitly URL.

On display we don't want to block but also what do we do assuming the task doesn't get the bitly URL in time... I could trigger the task and return the full URL. The next time it is loaded the document would have the bitly URL.

kuma/wiki/models.py
jezdez: The best way I think as it would allow autodoc sections is this style: http://sphinx-doc.org/domains.html#info-field-lists

kuma/wiki/models.py
jezdez: Yeah, that would work for me, as it's the same pattern we use already for the tidied content (see `get_tidied_content`)

kuma/wiki/templates/wiki/includes/document_macros.html
jezdez: Ah, oki

robhudson:
Updated based on feedback.

What's missing is adding a management command to process all (or most?) documents to generate the bitly URLs as celery tasks.

kuma/wiki/models.py
```diff
@@ -925,6 +929,10 @@ def save(self, *args, **kwargs):
         # Delete any cached last-modified timestamp.
         self.fill_last_modified_cache()
 
+        # Set the shortened URL (e.g. bit.ly).
+        # TODO: Should we do this on a subset of documents? E.g. not redirects?
+        tasks.set_share_url.delay(self.pk)
```
groovecoder: We shouldn't do this on every call to `Document.save()` right? That means every time we save a new `Revision` and set it to the document `current_revision`, we will generate a brand new share url for the document?

groovecoder:
I suggest we skip a management command to generate share urls for every doc. There's already a TODO noting that redirects do not need share urls; and neither do templates, etc.

So, I suggest we let ordinary site traffic trigger the background tasks to generate share urls as documents are visited. Though we should set a rate limit on the celery task so bitly doesn't shut us down again. :neutral_face:

kuma/wiki/models.py
robhudson: True. It wouldn't get a new share URL. Bitly is smart enough to return the same one. But you are right, we can make this smarter I believe. It's not as simple as not setting if it one already exists b/c we need to handle the case of tree moving, where docs may have a `share_url` already but the URL is changing so it needs to be updated.

robhudson:
Updated:
* don't call bitly for templates, redirects, and existing documents.
* call bitly for documents during `_move_tree`.
* add celery rate limiting.
* calls to `get_share_url` with no short_url returns full document url as fallback.

kuma/wiki/migrations/0008_add_share_url.py
```diff
@@ -0,0 +1,20 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='document',
+            name='share_url',
+            field=models.URLField(null=True, blank=True),
+            preserve_default=True,
+        ),
+    ]
```
jezdez: Oh, no new migration for the locale fields needed?

kuma/wiki/models.py
```diff
@@ -925,6 +931,11 @@ def save(self, *args, **kwargs):
         # Delete any cached last-modified timestamp.
         self.fill_last_modified_cache()
 
+        # Set the shortened URL (e.g. bit.ly). Note: We do this only for new
+        # documents that aren't templates or redirects.
+        if is_new and not self.is_template and not self.is_redirect:
```
jezdez: Please put this into the `on_document_save` signal handler in the `wiki/apps.py` and check for the `created` flag passed into the signal handler. That's the canonical place for such things now

kuma/wiki/tasks.py
```diff
@@ -460,3 +461,31 @@ def tidy_revision_content(pk):
             )
         # return the errors so we can look them up in the Celery task result
         return errors
+
+
+@task(rate_limit='10/m')
+def set_share_url(pk):
```
jezdez: please use a more descriptive name for the task such as `update_document_share_url` or something. the name of the ask will show up in the celery admin and it's then easier to distinguish from other the tasks

kuma/wiki/models.py
```diff
@@ -1467,6 +1484,21 @@ def contributors(self):
     def zone_stack(self):
         return DocumentZoneStackJob().get(self.pk)
 
+    def get_share_url(self):
+        """
+        Returns the `share_url`.
```
jezdez: Double backticks here

kuma/wiki/tests/test_models.py
```diff
@@ -333,6 +332,39 @@ def test_redirect_url_works_for_home_path(self):
         d = document(is_redirect=True, html=html)
         eq_(href, d.get_redirect_url())
 
+    @mock.patch('kuma.wiki.tasks.set_share_url')
+    def test_get_share_url_empty(self, _mock):
+        doc = document(title='test', share_url=None)
```
jezdez: that's odd, the document helper doesn't save the Document by default, so I'm not sure why the mock is being called. also, mind not using a underscored variable names here?

kuma/wiki/tests/test_tasks.py
```diff
@@ -76,3 +81,31 @@ def test_sitemaps_files(self):
...
         for loc in expected_sitemap_locs:
             ok_(loc in index_xml)
+
+
+@override_settings(BITLY_API_KEY='test', BITLY_USERNAME='test')
+class BitlyTestCase(UserTestCase):
+    fixtures = UserTestCase.fixtures + ['wiki/documents.json']
+
+    def setUp(self):
+        self.long_url = 'http://example.com/long-url'
```
jezdez: please call the super setUp

settings.py
```diff
@@ -836,8 +835,8 @@ def JINJA_CONFIG():
 FLAG_REASONS = DEMO_FLAG_REASONS + WIKI_FLAG_REASONS
 
 # bit.ly
-BITLY_API_KEY = "SET ME IN SETTINGS_LOCAL"
-BITLY_USERNAME = "SET ME IN SETTINGS_LOCAL"
+BITLY_API_KEY = None  # Set in settings_local.py.
```
jezdez: "Set me in settings_local.py"

kuma/wiki/tests/test_models.py
```diff
@@ -113,7 +112,7 @@ def test_document_is_template(self):
 
         assert not d.is_template
 
-        d.slug = 'Template:test'
+        d.slug = '%stest' % TEMPLATE_TITLE_PREFIX
```
jezdez: nice <3

kuma/wiki/models.py
robhudson: Yes, that would be much nicer. Thanks.

kuma/wiki/migrations/0008_add_share_url.py
robhudson: they are needed but I was going to make a separate PR for those.

kuma/wiki/tests/test_models.py
robhudson: It's called b/c the `get_share_url()` will trigger it if not already populated.

robhudson:
Updated based on feedback.

jezdez:
r+

PR #3506, opened 2015-09-22
Remove contributor bar

darkwing:
In a word:  controversy.  But it's here if we want to pull the trigger.


jezdez:
FTR this isn't enough to remove it from the backend

darkwing:
I looked at the model code and it appeared to just store contributors, and we still need that list.

groovecoder:
Correct - we use the same list to draw the contributor links along the bottom of the article(s), which we want to keep. This change is simply to cut out the <= 13 gravatar img requests.

jezdez:
Ah, no tests to be removed as well?

groovecoder:
Up-lifting from IRC: Can we waffle this feature (again :cry:) and create a user segment on it. That way we can let it run for a few days and compare GA site speed report between visitors with this feature and those without so we know exactly how much we're adding to average page load time with the feature.

PR #3509, opened 2015-09-22
Bug 957802: Make DB ignore more generic

openjck:
The .gitignore file is updated to also match files named
devmo_sanitized.sql

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,5 @@ celeryev.pid
 media/js/libs/ckeditor/source/ckbuilder
 .tox
 .env
-*_devmo.sql*
+*devmo*.sql*
 htmlcov/


robhudson:
Yes! Thank you.

.gitignore
```diff
@@ -38,5 +38,5 @@ celeryev.pid
 media/js/libs/ckeditor/source/ckbuilder
 .tox
 .env
-*_devmo.sql*
+*devmo*.sql*
```
groovecoder: Should we not just ignore *sql* ? :smile: No point in having SQL in our repo at all.

robhudson: There does exist the ability to have initial SQL files that Django runs to do database prep like stuff.

PR #3513, opened 2015-09-23
bug 1206309 - waffled api driven compat tables

stephaniehobson:
Display api-driven compat tables to users with a waffle enabled on pages that have been identified as ready, hide old style comapt table.

Testing:
1) download PR
2) `./manage.py compress_assets`
3) create waffle flag named `compat-api` and enable it for yourself
4) make/update a local copy of: https://developer.allizom.org/en-US/docs/Template%3AEmbedCompatTable
5) make a local copy of: https://developer.allizom.org/en-US/docs/Web/CSS/background
5) embed the macro in the browser compatibility section as the last thing in the section `{{EmbedCompatTable('web-css-background')}}`
6) verify that new table appears and the old table does not
7) disable waffle flag
8) verify that new table does not appear and the old table does

Preserving skateboard gif from #3504, because gif:

### DIFF
--- a/media/js/wiki-compat-tables.js
+++ b/media/js/wiki-compat-tables.js
@@ -1,5 +1,21 @@
 (function(win, doc, $) {
 
+    var $compatWrapper = $('.bc-api');
+    // show tables
+    $compatWrapper.removeClass('hidden');
+
+    // hide old style compat table and any footnotes
+    // go back up DOM to find section heading
+    var $sectionHead = $compatWrapper.prevAll('h2').first();
+    var sectionHeadIsMatch = $sectionHead.filter(function() {
+            return $(this).text().match(/(browser )?compat[i|a]bility/i);
+        });
+    // is it a browser compat section?
+    if(sectionHeadIsMatch) {
+        // come back down DOM hiding things until we reach the comapt table.
+        $sectionHead.nextUntil('.bc-api').hide();
+    }
+
     // Private var to assign IDs to history for accessibility purposes
     var historyCount = 0;
 
--- a/media/js/wiki.js
+++ b/media/js/wiki.js
@@ -304,8 +304,12 @@
         Load and build compat tables if present in page
     */
     (function() {
-        var $compatTables = $('.bc-table');
+        // don't run if no compat table on page with min 1 row
+        var $compatTables = $('.bc-api table tbody tr');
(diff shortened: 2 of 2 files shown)


media/js/wiki-compat-tables.js
```diff
@@ -1,5 +1,30 @@
 (function(win, doc, $) {
 
+    var $compatWrapper = $('.bc-api');
+    // check there's a table with at least one row
+    if(!$compatWrapper.length) {
```
darkwing: IMO this block would be better reversed; i.e. `if($compatWrapper.length)`

media/js/wiki-compat-tables.js
```diff
@@ -1,5 +1,30 @@
...
+                $thisWrapper.removeClass('hidden');
+            }
+        });
+    }
+
+    // hide old style compat table and any footnotes
+    // go back up DOM to find section heading
+    var $sectionHead = $compatWrapper.prevAll('h2').first();
+    var sectionHeadIsMatch = $sectionHead.filter(function() {
+            return $(this).text().match(/([B|b]rowser )?[C|c]ompat[i|a]bility/);
```
darkwing: Using the regex's `/i` modifier could save you the B|b C|c mess

darkwing:
Kicked travis too

stephaniehobson:
Updated!

stephaniehobson:
Stupid button.

stephaniehobson:
Nope, it's removing something it shouldn't now.

media/js/wiki-compat-tables.js
```diff
@@ -1,5 +1,30 @@
...
+    // check there's a table with at least one row
+    if($compatWrapper.length) {
+        $compatWrapper.each(function() {
+            var $thisWrapper = $(this);
+            if($thisWrapper.find('tbody tr').length) {
+                $thisWrapper.removeClass('hidden');
+            }
+        });
+    } else {
+        return;
```
stephaniehobson: @darkwing This is supposed to stop the function defined on line 1 from executing? Right?

stephaniehobson: Nevermind, found the problem (I think).

stephaniehobson:
Okay, really ready again.

darkwing:
R+, works great!

PR #3514, opened 2015-09-24
Fix bug 1160178 - Remove the ridiculous localization tags machinery

robhudson:


robhudson:
@jezdez Just wondering if I'm on the right track here. I have a few `TODO`s to complete before I'd consider this ready.

jezdez:
@robhudson Yep, exactly what I had in mind

jezdez:
@robhudson I had to re-add the `LocalizationTaggedRevision` and `LocalizationTag` models to the code base to be able to get the migrations running (I think). That's because taggit's magic model manager lookup registration doesn't work with frozen migration models. I guess we could also just do a manual SQL query for the `content_object_id` column of the `wiki_localizationtaggedrevision` table to get a list of Revision IDs and use that for updating the `localization_in_progress` field to `True`. What do you think?

robhudson:
Hmm, yeah. I'd imagine you'd need 3 migrations: (1) Add the new field, (2) Update the new field from the old field(s), (3) Remove the old field(s). Is that what you're thinking?

jezdez:
@robhudson That's what I did but still the manager is not available inside the migrations for doing the data migration since the migrations run with semi fake models.

willkg:
Where's this PR at? Are we planning to finish this up in the next month? Is it blocked on other work?

I think if we're not planning to finish this up soon, we should close it out.

robhudson:
I think we should close it. It's a legit bug that still needs work but I've got other things on my plate at the moment. I'll make sure I keep the branch around.

jezdez:
Leaving the branch open so we can pick it up later?

PR #3516, opened 2015-09-24
bug 1184662 - Update Intern whitespace and better code coverage

darkwing:
Updated whitespace from first call, also got rid of _cli code coverage issue.


tests/ui/intern-browserstack.js
```diff
@@ -3,6 +3,7 @@
 // packages, suites, excludeInstrumentation, and (if you want functional tests) functionalSuites.
 define(['./_base', './_cli', 'intern'], function(config, cli, intern) {
 
+    // Browsers to run integration testing against.
```
groovecoder: We should also add IE ... maybe Safari? Maybe mobile platforms?

groovecoder: Also, if we leave off the version here, BrowserStack says it will use the latest stable version of the browser available on the platform. Should we just leave it off?

darkwing: We can leave off the version numbers for Safari and Firefox -- they're updated so often that we don't need multiples.

darkwing: Let's add different browsers in another PR, please.

groovecoder:
Looks good and works well!

PR #3520, opened 2015-09-28
fix bug 1209045 - Update contributors.json to reflect current team

darkwing:
Noticed this file needed some updating.

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -16,10 +16,10 @@
             "groovecoder",
             "hoosteeno",
             "jezdez",
-            "mars",
+            "jwhitlock",
             "openjck",
-            "shobson",
-            "ubernostrum"
+            "robhudson",
+            "shobson"
         ]
     },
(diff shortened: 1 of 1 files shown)


contribute.json
```diff
@@ -16,10 +16,10 @@
             "groovecoder",
             "hoosteeno",
             "jezdez",
-            "mars",
             "openjck",
             "shobson",
-            "ubernostrum"
+            "robhudson",
+            "jwhitlock"
```
jezdez: Seems like the list of sorted alphabetically before, can we do that here, too?

darkwing:
Good catch @jezdez .  Updated.

groovecoder:
r+

nit: no need to file https://bugzilla.mozilla.org/show_bug.cgi?id=1209045 separately. Save yourself some paperwork and re-use bug 957802 for this kind of stuff if you like.

PR #3521, opened 2015-09-28
Bug 1197444: Compile Stylus before collectstatic

openjck:


groovecoder:
skipping Travis since it’s the chief_deploy script - un-used and un-affected by anything Travis can catch

PR #3522, opened 2015-09-28
Fix bug 1206429 - Fix pagination query string args in revision dashboard

robhudson:
There were 2 issues here:

1. The 'page=' parsing was picking up the page number plus the rest of
   the query string resulting in an invalid page number on the server.

2. The query string args were sent twice to the server because the form
   action was not set. It happened to work because the second set of
   query string args overwrote the first set.


darkwing:
R+, worked!  Nice work @robhudson !

PR #3523, opened 2015-09-28
Static CSS and JavaScript cleanup

openjck:


darkwing:
What's the best way to test this?

openjck:
I would test that `compile-stylesheets` still works, that `compress_assets` still runs, and that files in *media/(redesign/)js* and *media/(redesign/)css* redirect to their static counterparts assuming a provision was done recently.

darkwing:
R+, worked great for me!

PR #3525, opened 2015-09-28
Upgrade to Django 1.8

robhudson:
@jezdez / @groovecoder : This is now using django-jinja and I open it up for help to finish off the remaining Django 1.8 failures.

Namely:
- [x] `ValueError: save() prohibited to prevent data loss due to unsaved related object 'based_on'.`
- [x] `ValueError: save() prohibited to prevent data loss due to unsaved related object 'current_revision'.`
- [x] `ValueError: save() prohibited to prevent data loss due to unsaved related object 'parent_topic'.`
- [x] `ValueError: dictionary update sequence element #0 has length 0; 2 is required`

@robhudson will take:
- [x] `TemplateDoesNotExist: users/email/welcome/plain.ltxt`
- [x] `TypeError: __init__() takes exactly 2 arguments (1 given)`

I also see the following deprecation warnings in the console:
- [x] `/home/vagrant/src/vendor/src/django-cacheback/cacheback/tasks.py:5: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.`
- [ ] `/home/vagrant/src/vendor/src/django-soapbox/soapbox/models.py:42: RemovedInDjango19Warning: Model class soapbox.models.Message doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.`


groovecoder:
Travis makes it look like django-nose is a library that needs update for 1.8? :wink:

groovecoder:
Rebased, updated Makefile, and updated PR description to list the most common errors in the test output thus far.

groovecoder:
After poking into `DjangoTranslation` and tower, I'm starting to think it may be easier to completely re-do our translation code. [`DjangoTranslation` got a big refactor in 1.8](https://github.com/django/django/commit/a5f6cbce07b5f3ab48d931e3fd1883c757fb9b45#diff-5682903b7aed1649cdb7095331e90d0d).

groovecoder:
After today's backend meeting, I wonder if we could merge & ship these commits with Django 1.7?

groovecoder:
Re: `ValueError: save() prohibited to prevent data loss due to unsaved related object` ... it's [this change in 1.8.5](https://docs.djangoproject.com/en/1.8/releases/1.8/#assigning-unsaved-objects-to-relations-raises-an-error) to raise this error when saving a model instance with a related un-saved model instance.

We do this in a few places, especially where we wrap many db operations in `transaction.set_autocommit(False)` ...

* `current_revision` (8 errors) - When we move pages, [we create an un-saved `Revision`](https://github.com/mozilla/kuma/blob/dded246640d6dd7361fa6b2743b21f54a08c1cab/kuma/wiki/models.py#L1079-L1082) and save the `Document` model before we save the moved `Revision` model.
* `based_on` (8 errors): When we [revert default language documents](https://github.com/mozilla/kuma/blob/dded246640d6dd7361fa6b2743b21f54a08c1cab/kuma/wiki/models.py#L802-L818), we assign the `Revision.based_on` field to itself before calling `Revision.save()`.
* `parent_topic` (1 error): Just bad test code. [Easy fix](https://github.com/mozilla/kuma/commit/96f7808ed14049a16660ce87f20dabac063fc62c).

For the others, I'm going to try simply moving the related model (i.e., `Revision`/`Document`) operations to occur **after** the the `save()`.

groovecoder:
I've got a commit with re-ordered save operations that fixes the `ValueError`s. I [opened a pull request](https://github.com/mozilla/kuma/pull/3660) to this branch to get feedback on the approach before committing it here.

groovecoder:
Pushed a couple commits from https://github.com/mozilla/kuma/pull/3668 here to fix some more of the `ValueError`s.

robhudson:
@groovecoder Branch is updated and rebased with your work which fixed the `save()` `ValueError`s. :tada: 

@jezdez The dict `ValueError`s seem to be related to how we're subclassing django-rest-framework's `TemplateHTMLRenderer`? Since you built that do you think you could take a look at determine what it is expecting there? At first I thought it had something to do with our jinja2 instrumented renderer in `kuma/core/tests/__init__.py` but when I removed that completely the error was still producing.

Tomorrow I'll dig into that `TemplateDoesNotExist` error.

jezdez:
@robhudson Sure, I'll take a look

stephaniehobson:
A warning in foreman start I don't see recorded here yet:
`/home/vagrant/src/vendor/src/django/django/core/management/base.py:260: RemovedInDjango19Warning: "requires_model_validation" is deprecated in favor of "requires_system_checks".`

stephaniehobson:
I don't know if this is specific to this branch but the revision history is out of order:

screen shot 2015-12-14 at 13 26 01

stephaniehobson:
Caching seems a bit stickier? I needed to manually force-refresh after editing but I think we have open bugs for this on prod so I'm not sure it's a new problem?

stephaniehobson:
I can't seem to move pages around :( here's the gist from one attempt: https://gist.github.com/stephaniehobson/50c5c306a221025ada99

robhudson:
> I can't seem to move pages around :( here's the gist from one attempt

It seems that you may have 2 users in your database with the same email.

stephaniehobson:
Hm. That's possible.

stephaniehobson:
The page move problem was indeed the duplicated email addresses.

robhudson:
@stephaniehobson wrt revision history being out of order... Looking at the code it looks like it always shows the current revision first and the historical revisions after. Is that different than what you're seeing?

robhudson:
I compared the revision history page's SQL on master and this branch and the SQL was identical (for the queries that build the revision history list).

robhudson:
r+ :boom:

jezdez:
26tpdwmm4jyclgxtq

PR #3526, opened 2015-09-29
Bug 1182210 - Waffle share_links

robhudson:


groovecoder:
r+ before Travis to stop more celery tasks generating.

PR #3528, opened 2015-09-29
Fix Bug 1209667: Remove header_tools waffle

stephaniehobson:


stephaniehobson:
I know how much @groovecoder likes removing waffle flags.

groovecoder:
Removed from prod & stage.

PR #3529, opened 2015-09-29
Bug 1182210 - Send non-locale URL to bitly

robhudson:


groovecoder:
Code looks good.

Should we add a data migration to empty out any `Document.share_url` values that were generated and saved before this change?

robhudson:
Yep, a data migration is a good idea. I'll add that soon.

jezdez:
Hm, can't slugs change between translations?

groovecoder:
They can, but I spot-checked and found that, e.g., https://developer.mozilla.org/docs/Web/HTTP/Access_control_CORS with `Accept-Language: fr` correctly redirects to https://developer.mozilla.org/fr/docs/HTTP/Access_control_CORS while `Accept-Language: en` correctly redirects to https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS.

So, the urls without locales will still work even if the slugs don't match between English and non-English pages.

jezdez:
Hm, that doesn't seem like a sensible idea still though. If we want to hand out URLs to be shared and live forever, I don't understand why we'd remove the specifics of the language from it. The content is different between languages so it would muddy out statistics about page requests.

Additionally language selection of an article should not depend on the Accept-Locale header of the person clicking on the shared URL but only happen when the person getting the URL to share has selected a language and therefor the content -- either automatically via the Accept-Locale header or manually via the language selector.

I'm r- on this PR and don't see this topic discussed in bug 1182210 either.

groovecoder:
Re stats: the visits coming back to MDN will report correctly in GA which is where we're tracking the stats and effectiveness of the share links.

I'm convinced by the 2nd UX argument though - if someone shares en-US/docs/Web/HTTP/CORS, we don't want to intercept their intention and have them share fr/docs/HTTP/CORS instead. They may not be endorsing the 'fr' version of the doc; indeed - many translations can be far out of sync with their 'en-US' origins.

So, I agree with @jezdez - we should generate distinct share urls per locales.

groovecoder:
Closing ... re-open if you can make a compelling counter-argument? :wink:

jezdez:
Ah, good point about the location of the stats collection, for some reason I assumed we'd have some things happening there on bitly's side of things or that it reports to GA or something.

robhudson:
That changes how this is architected drastically. Instead of a single field on the Document model we will need a separate table for (doc_id, locale, short_url). And I'd argue that we shouldn't pre-compute that and slam bit.ly but do it at "share time" ("Just In Share-Time"?). I'm not even sure we should do it on document save.

groovecoder:
Why the extra table? `wiki_document` already has a `locale` field. So unique `short_url` fields in `wiki_document` are already associated with a locale.

groovecoder:
+1 to do Just In Share-Time™ though it probably needs some AJAX trickery for a nice UX?

robhudson:
> Why the extra table? wiki_document already has a locale field. So unique short_url fields in wiki_document are already associated with a locale.

Oh hey, look at that. Thanks for pointing that out.

So all that's remaining is something along the lines of:
1. An ajax URL to request the short URL at share time. This will either:
1a. return the short URL immediately if it is already stored
1b. request a short URL from bit.ly and return that. If for some reason the bitly request fails we can fall back to the long URL.
2. Connect the share widget to call this ajax URL to get the short url.

Does that sound right?

groovecoder:
Yup. Sounds right.

PR #3530, opened 2015-09-29
bug 1197444 - Load just-in-time assets in development environment

jwhitlock:
Add jingo helpers to get paths to bundled CSS and JS assets, and add to the in-page global mdn.assets. Use these paths rather than hard-coded values when dynamically loading JS and CSS.

In a development environment, style-prism.js is split into 5 files, and the first one has to be fully loaded before the rest will load correctly.

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -191,11 +191,30 @@
         Syntax highlighting scripts
     */
     if($('article pre').length && ('querySelectorAll' in doc)) (function() {
-        var syntaxScript = doc.createElement('script');
-        syntaxScript.setAttribute('data-manual', '');
-        syntaxScript.async = 'true';
-        syntaxScript.src = mdn.staticPath + 'js/syntax-prism-min.js?build=' + mdn.build;
-        doc.body.appendChild(syntaxScript);
+        if (mdn.assets && mdn.assets.js.hasOwnProperty('syntax-prism')) {
+            mdn.assets.js['syntax-prism'].forEach(function(url, index, array) {
+                /*
+                   Note: In development, five scripts are loaded, and the
+                   later scripts use Prism, which is declared in the first
+                   script.  This means syntax highlighting often doesn't work
+                   on the first page load. Refresh, and syntax highlighing
+                   should work.
+
+                   To fix this, we'd have to use something like require.js or
+                   in-client merging of JS files.  Maybe django-pipeline will
+                   help as well.
+                 */
+                var syntaxScript = doc.createElement('script');
+                syntaxScript.setAttribute('data-manual', '');
+                syntaxScript.src = url;
+                if (array.length === 1) {
+                    syntaxScript.async = 'true';
+                } else {
+                    syntaxScript.defer = 'true';
+                }
+                doc.body.appendChild(syntaxScript);
+            });
+        }
     })();
 
     /*
@@ -306,19 +325,27 @@
     (function() {
         // don't run if no compat table on page with min 1 row
         var $compatTables = $('.bc-api table tbody tr');
+        var compatCSS, compatJS;
         if(!$compatTables.length) return;
 
         // don't run if waffle not active
         if(!win.waffle || !win.waffle.flag_is_active('compat_api')) return;
 
+        if(!win.mdn || !mdn.assets) return;
+        /*
+           This chaining logic will break in development if these assets are
+           split into multiple files.
+         */
+        compatCSS = mdn.assets.css['wiki-compat-tables'][0];
+        compatJS = mdn.assets.js['wiki-compat-tables'][0];
         $('<link />').attr({
-                href: mdn.staticPath + 'css/wiki-compat-tables-min.css',
+                href: compatCSS,
                 type: 'text/css',
                 rel: 'stylesheet'
             }).on('load', function() {
 
                 $.ajax({
-                    url: mdn.staticPath + 'js/wiki-compat-tables-min.js',
+                    url: compatJS,
                     dataType: 'script',
                     cache: true
                 }).then(function() {
--- a/kuma/wiki/helpers.py
+++ b/kuma/wiki/helpers.py
@@ -14,6 +14,7 @@ from django.utils.html import conditional_escape
 
 from constance import config
 from jingo import register
+from jingo_minify.helpers import get_js_urls, get_css_urls
 
 from kuma.core.urlresolvers import reverse
 from .constants import DIFF_WRAP_COLUMN
@@ -232,3 +233,57 @@ def wiki_url(path):
     Look for a wiki page in the current locale, or default to given path
     """
     return reverse('wiki.document', args=[path])
+
+
+def asset_url_array(asset_urls):
+    """
+    Return a JS array of asset URLs.
+    """
+    url_array = '[' + ', '.join('"%s"' % url for url in asset_urls) + ']'
+    return jinja2.Markup(url_array)
+
+
+@register.function
+def js_url_array(bundle, debug=None):
+    """
+    Return a JS array of the URLS for a minified and bundled JS file.
+
+    In production, the array will have one item, the URL of the minified,
+    bundled JS file.  In development, each bundled item will be present, and
+    it will be the unminified versions.
+
+    For example, in production, 'main' returns CDN paths like:
+    ["https://developer.cdn.mozilla.net/static/js/main-min.js?build=97c12a9"]
+
+    In development, 'main' returns local paths like:
+    [
+      "/static/js/libs/jquery-2.1.0.js?build=1443476862,
+      "/static/js/components.js?build=1443476862",
+      "/static/js/analytics.js?build=1443476862",
+      "/static/js/main.js?build=1443476862",
+      "/static/js/auth.js?build=1443476862",
+      "/static/js/libs/fontfaceobserver/" +
+        "fontfaceobserver-standalone.js?build=1443476862",
+      "/static/js/fonts.js?build=1443476862"
+    ]
+    """
+    return asset_url_array(get_js_urls(bundle, debug))
+
+
+@register.function
+def css_url_array(bundle, debug=None):
+    """
+    Return a JS array of the URLS for a minified and bundled CSS file.
+
+    Similar to js_url_array, the array will have one item in production, and
+    potentially many items in development.  For example, in production 'mdn'
+    returns something like:
+    ["https://developer.cdn.mozilla.net/static/js/main-min.js?build=97c12a9"]
+
+    In development, 'mdn' returns something like:
+    [
+      "/static/css/font-awesome.css?build=1443558726",
+      "/static/css/main.css?build=1443563109"
+    ]
+    """
+    return asset_url_array(get_css_urls(bundle, debug))
--- a/templates/includes/config.html
+++ b/templates/includes/config.html
@@ -30,6 +30,16 @@
             // Wiki-specific settings will be placed here
             wiki: {
                 autosuggestTitleUrl: '{{ url('wiki.autosuggest_documents') }}'
+            },
+            // Just-in-time asset paths
+            assets: {
+                css: {
+                    'wiki-compat-tables': {{ css_url_array('wiki-compat-tables') }}
+                },
+                js: {
+                    'syntax-prism': {{ js_url_array('syntax-prism') }},
+                    'wiki-compat-tables': {{ js_url_array('wiki-compat-tables') }}
+                }
             }
         };
     })(this);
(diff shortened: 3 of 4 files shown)


kuma/static/js/wiki.js
```diff
@@ -306,19 +316,24 @@
...
+        var $compatCSS, $compatJS;
         if(!$compatTables.length) return;
 
         // don't run if waffle not active
         if(!win.waffle || !win.waffle.flag_is_active('compat_api')) return;
 
+        if(!win.mdn || !mdn.assets) return;
+        /* This will break if these assets are split into multiple files */
+        $compatCSS = mdn.assets.css['wiki-compat-tables'][0];
+        $compatJS = mdn.assets.js['wiki-compat-tables'][0];
```
openjck: In JavaScript, the `$` character conventionally used to indicate that the variable represents a JQuery object.

jwhitlock: OK, I'll remove it for these items

jwhitlock:
Force pushed to change JS variables to remove ``$`` (i.e. ``$compatCSS`` to ``compatCSS``) when the value is not a jQuery object.

openjck:
> Force pushed to change JS variables to remove $ (i.e. $compatCSS to compatCSS) when the value is not a jQuery object.

Awesome! Looks good. I'll test functionality and dive deeper into the code tomorrow.

This is great, by the way. I know it'll be a big improvement. :smile:

openjck:
This works beautifully. It might take me a short while to review all the code. Mostly I just want to be sure I understand everything before the migration to django-pipeline.

templates/includes/config.html
```diff
@@ -30,6 +30,16 @@
...
+            },
+            // Just-in-time asset paths
+            assets: {
+                css: {
+                    "wiki-compat-tables": {{ css_bundle_urls("wiki-compat-tables") }}
+                },
+                js: {
+                    "syntax-prism": {{ js_bundle_urls("syntax-prism") }},
+                    "wiki-compat-tables": {{ js_bundle_urls("wiki-compat-tables") }}
+                }
```
openjck: Just a small nitpick. We don't have strict rules about JavaScript coding conventions, but tend to use single quotes around strings.

kuma/wiki/helpers.py
```diff
@@ -232,3 +233,24 @@ def wiki_url(path):
     Look for a wiki page in the current locale, or default to given path
     """
     return reverse('wiki.document', args=[path])
+
+
+def url_list(urls):
+    js_list = '[' + ', '.join('"%s"' % url for url in urls) + ']'
+    return jinja2.Markup(js_list)
```
openjck: Would `asset_list` be a better name? Assuming I understand correctly that it could represent a list of CSS *or* a list of JS assets.

openjck: Ah. Makes sense now. This is building a JavaScript array for use in `config.html`. That makes sense. I think the names tripped me up a bit, partly because the term array is more conventional in JS-land. Maybe `def javascript_asset_array` and `js_array` or similar?

kuma/wiki/helpers.py
```diff
@@ -232,3 +233,24 @@ def wiki_url(path):
...
+
+
+@register.function
+def js_bundle_urls(bundle, debug=None):
+    """
+    Return a list of the URLS for a minified and bundled JS file
```
openjck: Maybe we can use the term *JavaScript array* instead of *list* here as well. Also, my initial reading of the comment led me to think that only minified URLs would be returned.

kuma/static/js/wiki.js
```diff
@@ -306,19 +316,24 @@
...
+        var compatCSS, compatJS;
         if(!$compatTables.length) return;
 
         // don't run if waffle not active
         if(!win.waffle || !win.waffle.flag_is_active('compat_api')) return;
 
+        if(!win.mdn || !mdn.assets) return;
+        /* This will break if these assets are split into multiple files */
+        compatCSS = mdn.assets.css['wiki-compat-tables'][0];
+        compatJS = mdn.assets.js['wiki-compat-tables'][0];
```
openjck: Awesome! Thanks for being clear about this. We could even make the `forEach` loop (above) more general to handle both the `syntax-prism` bundle and the `wiki-compat-tables` bundle, and any other variable-length bundles we want to dynamically inject in the future.

kuma/static/js/wiki.js
```diff
@@ -191,11 +191,21 @@
         Syntax highlighting scripts
     */
     if($('article pre').length && ('querySelectorAll' in doc)) (function() {
-        var syntaxScript = doc.createElement('script');
-        syntaxScript.setAttribute('data-manual', '');
-        syntaxScript.async = 'true';
-        syntaxScript.src = mdn.staticPath + 'js/syntax-prism-min.js?build=' + mdn.build;
-        doc.body.appendChild(syntaxScript);
+        var bundle = 'syntax-prism';
+        if (window.mdn && mdn.assets.js && mdn.assets.js.hasOwnProperty(bundle)) {
```
openjck: We assume the existence of `mdn` elsewhere, so we can probably do it here too by omitting the `window.mdn` check.

openjck: The hasOwnProperty check should go inside the `forEach` loop.

kuma/static/js/wiki.js
```diff
@@ -191,11 +191,21 @@
...
+            mdn.assets.js[bundle].forEach(function(bundle_url, index, array) {
+                var syntaxScript = doc.createElement('script');
+                syntaxScript.setAttribute('data-manual', '');
+                syntaxScript.src = bundle_url;
+                if (index === 0 && array.length > 1) {
+                    /* In development, slow load first script */
```
openjck: Thanks for adding this comment. Would it be possible to extend it further to explain why?

templates/includes/config.html
jwhitlock: OK, will change to single quotes

kuma/wiki/helpers.py
jwhitlock: how about ``asset_url_array``?

kuma/wiki/helpers.py
jwhitlock: I'm expanding the doc strings with some examples to make it clearer, and changing the names to ``js_url_array`` and ``css_url_array``.

kuma/static/js/wiki.js
jwhitlock: I thought about that, but I can't see a clean way to do the chaining (load CSS, then load JS, then run some JS) in the case of multiple files in a bundle.  The specific code seemed cleaner to me.

kuma/static/js/wiki.js
jwhitlock: I'm iterating over the list returned from ``mdn.assets.js['syntax-prism']``, after ``mdn.assets.js.hasOwnProperty('syntax-prism')`` checks that the key exists.  That check doesn't make sense inside the loop.

kuma/wiki/helpers.py
openjck: Makes sense to me :smile:

kuma/static/js/wiki.js
jwhitlock: Yes, I'll add more to the comment.  In production, ``syntax-prism-min.js`` contains all five JS files together, evaluated in order.  In development, they are loaded as five separate files.  The later four refer to ``Prism``, which is defined in the first JS file.  We need to slow down an let the first JS file load completely before loading the follow-on files, or get errors about Prism being undefined.

jwhitlock: On further testing, ``async = 'false'`` is a truthy value too, so this code does nothing.  Unless we use something stronger (like [require.js](http://www.requirejs.org)), we can't guarentee load order.  Luckily, a refresh usually fixes it.  Changing the comment to an apology.  See [SO 9711160](http://stackoverflow.com/questions/9711160/jquery-load-scripts-in-order) for more ideas.

jwhitlock:
Force pushed:

* Change helpers ``css_bundle_urls`` to ``css_url_array``, ``js_bundle_urls`` to ``js_url_array``, to match that they generate JavaScript arrays.
* Better docstrings for helpers, with output examples
* Fix helper tests in TravisCI
* Simplify belt-and-braces checks for loading syntax-prism
* Add note that syntax-prism loading is unreliable in dev mode, will require a lot of work to fix
* Use single-quotes in JS code, matching current convention

darkwing:
Not to be a nit-picker but in JS we should camelCase instead of using "_" and "-" when naming things.

jwhitlock:
@darkwing - helper.py is Python, ``css_url_array`` and ``js_url_array`` are Jinja2 context functions

darkwing:
I was referring to `bundle_url`

jwhitlock:
Force push: ``bundle_url`` -> ``url`` in /kuma/static/js/wiki.js

kuma/static/js/wiki.js
```diff
@@ -191,11 +191,26 @@
...
+                /*
+                   Note: In development, five scripts are loaded, and the
+                   later scripts use Prism, which is declared in the first
+                   script.  This means syntax highlighting often doesn't work
+                   on the first page load. Refresh, and syntax highlighing
+                   should work.
+
+                   To fix this, we'd have to use something like require.js or
+                   in-client merging of JS files.  Maybe django-pipeline will
+                   help as well.
```
openjck: `async` is a boolean attribute, so any value (even `async="false"`) will enable asynchronous loading. If the `async` attribute is omitted when injecting multiple scripts, the scripts will execute sequentially.

openjck:
I'm afraid this branch needs a rebase. It also looks like tests are failing.

Thanks for your patience as we go through this. It works beautifully and is going to be a huge improvement over what we have now.

kuma/static/js/wiki.js
jwhitlock: yes, I tried omitting async as well.  It's up to the browser though, and my test browser continued to load the async scripts first, because they were much smaller and downloaded faster.  I decided to leave out async / no async, to prevent people from thinking it was a solved problem.

jwhitlock:
Rebased, tests pass

openjck:
I was hoping to give this one last review before PTO tomorrow but got caught up in Firefox Growth work. If someone wants to grab in the meantime please feel free, or I'll be able to give it more attention next week.

openjck:
Anything more to say on the JavaScript front, @darkwing? Otherwise this is r+ from me.

darkwing:
When I set `DEV` to `False`, I get a `Prism is not defined` error.  It's the `async` we're setting.  We probably want to sniff the file name for `syntax-prism` and only set `async` for that.  Let me know if I misread any of the conversation around this though.

jwhitlock:
*TL;DR - used ``defer``, rebased on master.  @darkwing, please try again.*

To summarize the conversation:

If ``wiki.js`` detects that syntax highlighting is needed, it injects this HTML in development:

```html
<script data-manual defer src="/static/js/libs/prism/prism.js?build=1443476862"></script>
<script data-manual defer src="/static/js/prism-mdn/components/prism-json.js?build=1443476862"></script>
<script data-manual defer src="/static/js/prism-mdn/plugins/line-numbering/prism-line-numbering.js?build=1443476862"></script>
<script data-manual defer src="/static/js/libs/prism/plugins/line-highlight/prism-line-highlight.js?build=1443476862"></script>
<script data-manual defer src="/static/js/syntax-prism.js?build=1443476862"></script>
```

The ``defer`` option (which I found on [this blog article](http://davidwalsh.name/html5-async)) tells [compliant browsers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Browser_compatibility) (Chrome, Firefox, IE, Safari) to load scripts in the declared order.  This is needed, because later scripts (``prism-json.js``) use ``Prism``, which is declared in the first script (``prism.js``).  Otherwise, the browser tends to execute the smallest scripts first, and you get ``Prism is not defined`` errors.  On browsers that don't support ``defer``, a refresh usually works, since the browser finds the scripts in cache and executes them in order.

In production, it injects what you'd expect:
```html
<script data-manual async src="https://developer.cdn.mozilla.net/static/js/syntax-prism-min.js?build=2aab9d6"></script>
```
This is the same code as the 5-file version, except 1) they are combined into a single script, so no loading order issues, 2) the combined contents are minified, and 3) it is server from the CDN rather than Django.  Because they are combined, we use ``async`` so the page stays as responsive as possible.

jwhitlock:
force push - *really* rebased on master

darkwing:
Thanks for the update @jwhitlock ; `defer` is known to be flaky but since it's only for dev I think we can leave it unless we experience problems.  I'll have a fresh look on Monday and we'll roll out after merge.

jwhitlock:
Rebased on master

darkwing:
R+, worked well for me.  Nice work @jwhitlock !

PR #3535, opened 2015-09-30
bug 1184670 - add intern to chief_deploy

groovecoder:


groovecoder:
Because the test suite is less reliable when running with `maxConcurrency` > 1, I'm going to close this. We don't want to run Intern in the chief deployment process unless/until the run-time is under 10m. (I clocked the latest BrowserStack run of Mac Fx, Win Fx, Mac Chr, Win Chr, and IE at ~20m)

PR #3536, opened 2015-10-01
Bug 1197444: Move fonts to kuma/static

openjck:
## Setup

`vagrant provision`

## Testing

1. Verify that fonts look good around the site (good pages to test are the [Demo Studio](https://developer-local.allizom.org/en-US/demos), the [Dev Derby](https://developer-local.allizom.org/en-US/demos/devderby), the [Promote page](https://developer-local.allizom.org/en-US/promote), and other pages that use uncommon fonts)
2. Enter [production mode](https://kuma.readthedocs.org/en/latest/development.html#production-assets)
3. Repeat step 1


openjck:
Come to think of it, this will also fix the font 404s we have been seeing on production. We were using the URL *mozorg.cdn.mozilla.net* in *opensans.styl*, but our CDN is *developer.cdn.mozilla.net*.

Thankfully, this pull request removes all remaining absolute URLs in stylesheets.

cc @groovecoder

stephaniehobson:
The mozorg CDN reference is probably an artifact of tabzilla. Yay for new tabzilla!

stephaniehobson:
Works in production mode, all the fonts 404 in dev mode.

openjck:
Having trouble figuring out what might be causing that. @jwhitlock can you reproduce?

jwhitlock:
Provisioned, rebooted VM to force Apache restart.  I also get 404s when running in dev mode.  The requests don't make it to the Django server (for example, https://developer-local.allizom.org/static/fonts/OpenSans-Italic-webfont.woff), so I suspect Apache is intercepting the calls.  Digging in...

openjck:
So strange. I'm still not able to reproduce this.

What really confuses me is that moving assets to `kuma/static` has become pretty routine for me now. I can't figure out what would be missing. Maybe I should come back to it after the long weekend.

openjck:
My only guess is that the change to *all-servers.conf* is causing unexpected behavior. Does anything different happen if *static* is changed back to *media* in that file?

A `vagrant halt && vagrant up && vagrant provision` would be needed for the changes to take effect. Alternatively, the file could be edited directly in the VM followed by an Apache restart.

jwhitlock:
If I comment out the ``Rewrite*`` lines in ``all-servers.conf``, then I get fonts in development mode:

```
    # block hotlinking by referer to .woff, .woff2, .eof, .ttf files except mozilla domains
    #RewriteCond "%{HTTP_REFERER}" "https?://.*\.(mozilla|allizom)\.(com|org|net)/.*$"
    #RewriteRule \.(woff|woff2|eot|ttf)$ - [F,NC,L,E=!CORS]
```

I'm not sure why it works for you in development mode, though.  Maybe because you have the font files in ``/home/vagrant/src/static/fonts``?  Maybe you aren't restarting apache after modifying ``/etc/apache2/sites-enabled/kuma.conf`` for dev mode?  We can try to reproduce after the weekend.

One more thing - please remove exec bits from fonts:
```
cd /home/vagrant/src/kuma/static/fonts
find . -type f -exec chmod 664 {} \;
```

openjck:
Updated!

We're really, really stuck on the 404 issue so we decided to remove hotlinking protection for fonts instead. I also moved the *.htaccess* file directly to *static/fonts* because collectstatic doesn't pick it up.

Both feel like compromises. I'm open to other approaches but can't figure any out.

groovecoder:
+1 IIRC the blocking of hot-linked fonts originally came about when we had bought copyright fonts and needed to make sure other sites couldn't use MDN to hot-link the fonts for free. Now that we're using OpenSans that's a small concern. If we see some sites abusing us we can re-visit.

openjck:
Rebased :smile:

provisioning/roles/kuma/files/apache2/all-servers.conf
```diff
@@ -3,16 +3,12 @@
...
+<Location /static/fonts>
     # Only enable CORS for fonts when the request is coming from a Mozilla domain
     Header unset Access-Control-Allow-Origin
     SetEnvIf Origin "https?://(.*\.(mozilla|allizom)\.(com|org|net))" CORS=$0
     SetEnvIf Origin "https?://(mdn\.mozillademos\.org)" CORS=$0
     Header set Access-Control-Allow-Origin %{CORS}e env=CORS
-
-    # block hotlinking by referer to .woff, .woff2, .eof, .ttf files except mozilla domains
-    RewriteCond "%{HTTP_REFERER}" "!https?://.*\.(mozilla|allizom)\.(com|org|net)/.*$"
-    RewriteRule \.(woff|woff2|eot|ttf)$ - [F,NC,L,E=!CORS]
```
groovecoder: +1 for this. I think it was the last thing blocking merge? (Though now this needs another rebase)

openjck:
Rebased :smile:

darkwing:
R+, all works for me in both dev and prod modes!

PR #3538, opened 2015-10-01
Bug 1206820: Make .errorlist class available on all pages.

stephaniehobson:


robhudson:
Would I then add that require line to `kuma/static/stylus/wiki-edit.styl` when I submit my form changes?

stephaniehobson:
Rebased.

PR #3539, opened 2015-10-01
Fix bug 1206820 - Don't allow case similar tags

robhudson:


kuma/wiki/forms.py
```diff
@@ -295,6 +305,34 @@ def clean_slug(self):
...
+                if len(doc_tag) > 1:
+                    log.warn('Found duplicate document tags: %s' % doc_tag)
+
+                if doc_tag:
+                    if doc_tag[0] != tag and doc_tag[0].lower() == tag.lower():
+                        dupe_tags.append(doc_tag[0])
+
+            if dupe_tags:
+                raise forms.ValidationError(
+                    TAG_DUPE % {'tag': ', '.join(dupe_tags)})
```
jezdez: For the join of a list that is used in translated strings it would be good to use `django.utils.text.get_text_list` since commas aren't universal signs of combination

kuma/wiki/forms.py
```diff
@@ -295,6 +305,34 @@ def clean_slug(self):
...
+
+        if tags:
+            for tag in parse_tags(tags):
+                # Note: The exact match query doesn't work correctly with
+                # MySQL with regards to case-sensitivity. If we move to
+                # Postgresql in the future this code may need to change.
+                doc_tag = DocumentTag.objects.filter(
+                    name__exact=tag).values_list('name', flat=True)
+                # Write a log we can grep to help find duplicate document tags.
+                if len(doc_tag) > 1:
```
jezdez: Could we move that log into behind the condition a few lines later?

kuma/wiki/forms.py
```diff
@@ -295,6 +305,34 @@ def clean_slug(self):
...
+                    log.warn('Found duplicate document tags: %s' % doc_tag)
+
+                if doc_tag:
```
jezdez: For this I like to use the [first](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#first) queryset method nowadays since it sets the result to `None` if nothing is found automatically. So this could be written like this:

    doc_tag = (DocumentTag.objects.filter(name__exact=tag)
                                  .values_list('name', flat=True)
                                  .first())
    if doc_tag is not None:
       # ..

I hope this works with ValuesQuerySets as well at least..

kuma/wiki/tests/test_forms.py
```diff
@@ -14,10 +15,10 @@ class FormEditorSafetyFilterTests(UserTestCase):
     def test_form_onload_attr_filter(self):
         """RevisionForm should strip out any harmful onload attributes from
         input markup"""
-        d, r = doc_rev("""
+        rev = revision(save=True, is_approved=True, content="""
```
jezdez: :heart_eyes:

jezdez:
rwc+

kuma/wiki/forms.py
robhudson: Sure. This is for the case of MySQL/Django finding more than one tag. E.g. if `tag='Javascript'` and `doc_tag == ['JavaScript', 'Javascript']`. I thought it would be helpful to be able to grep the logs like `grep 'duplicate document tags' <logfile> | uniq` to help us track down which ones we can remove so they don't get added back.

kuma/wiki/forms.py
robhudson: The thing is, I want them all for the logging. Then only the first one (arbitrarily) as the one I'm comparing against. To make it less arbitrary I probably should order by ID and pick the one with the lowest ID as it is likely the one that came first. In the case of `['JavaScript', 'Javascript']` that's true -- `'JavaScript'` is the one we want.

kuma/wiki/forms.py
robhudson: Nice. It's funny to see `_(', ')` in the Django code. But sad that this doesn't support oxford commas.

kuma/wiki/tests/test_forms.py
robhudson: I saw the note that someone wants to get rid of `doc_rev` so I figured I'd help while in this file.

kuma/wiki/forms.py
jezdez: Yeah, if you look in some of the translations of that you'll see some funny variants of separating stuff ;)

groovecoder:
I rebased this so it should be ready to merge.

PR #3543, opened 2015-10-04
Fix Bug 1207335 - Standardize summary-box code

stephaniehobson:
Added new general class for the writers to transition all other instances too.


darkwing:
R+ when mergeable

PR #3545, opened 2015-10-05
1202115 .index bottom margin

stephaniehobson:
Test page: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Animations (more details in bug).

The bottom margin on the ul/ol was being applied sporadically by Firefox and not at all by Chrome, by adding it to .index it should always be applied now, and collapse with the margins on the lists if/when they come into play.


darkwing:
R+, nice!

PR #3546, opened 2015-10-05
bug 1209185 - django 1.8: authkeys migrations; reset wiki migrations; update django-constance

groovecoder:


groovecoder:
Note: these are the changes I had to make in order to make the test suite even run.

robhudson:
Yeah, getting the test suite to run should be the first priority to help us find the next priorities. Thanks.

groovecoder:
I created migrations for the `authkeys` app - it didn't have any before.

I tried to `squashmigrations` of the `wiki` app, but the `move_to_user` function is incompatible with the current table layout. (Kept getting errors that `userprofile` did not exist when it tried to run.) So, I just deleted the existing migrations and did a new `makemigrations` to make a new initial wiki migration. It means vms will need to be re-built, and we will need to reset the migration history on stage & prod during deploy.

jezdez:
Yeah, squashing those migrations is a great move +1

robhudson:
Should we do this on all apps to make it easy to update the migration meta tables (and also reset the migrations)?

jezdez:
@robhudson good question, I think the wiki app may be a good candidate as well. Not sure about the others

PR #3548, opened 2015-10-05
fix bug 1210394 - only diff_table with tidied_content

groovecoder:
@robhudson or @jwhitlock - can you help me figure out why this new test fails? For some reason, the `tidied_content = ''` I'm saving in the test isn't persisted in the code-under-test. I.e., when I debug `revision_from.tidied_content` and `revision_to.tidied_content` inside `compare` view, it shows the actual tidied content and not the `''` I expect. :confused:

### DIFF
--- a/kuma/wiki/templates/wiki/includes/revision_diff_table.html
+++ b/kuma/wiki/templates/wiki/includes/revision_diff_table.html
@@ -3,4 +3,9 @@
 {% else %}
 	{% set from_content = '' %}
 {% endif %}
-{{ diff_table(from_content, revision_to.get_tidied_content(allow_none=True), revision_from.id, revision_to.id) }}
+{% set to_content = revision_to.get_tidied_content(allow_none=True) %}
+{% if from_content == None or to_content == None %}
+    {% trans %}The server is rendering the content comparison. Please refresh this page in a few minutes.{% endtrans %}
+{% else %}
+    {{ diff_table(from_content, revision_to.get_tidied_content(allow_none=True), revision_from.id, revision_to.id) }}
+{% endif %}
--- a/kuma/wiki/tests/test_templates.py
+++ b/kuma/wiki/tests/test_templates.py
@@ -741,6 +741,30 @@ class CompareRevisionTests(UserTestCase, WikiTestCase):
         response = self.client.get(url)
         eq_(404, response.status_code)
 
+    def test_no_tidied_content(self):
+        """
+        Verify revisions without tidied content show appropriate message.
+        """
+
+        # update() to skip the tidy_revision_content post_save signal handler
+        Revision.objects.filter(
+            id__in=[self.revision1.id, self.revision2.id]
+        ).update(
+            tidied_content=''
+        )
+
+        url = reverse('wiki.compare_revisions', args=[self.document.slug])
+        query = {'from': self.revision1.id, 'to': self.revision2.id}
+        url = urlparams(url, **query)
+        response = self.client.get(url)
+        eq_(200, response.status_code)
+        ok_('Please refresh this page in a few minutes.' in response.content)
+
+        url = url + '&raw=1'
+        response = self.client.get(url)
+        eq_(200, response.status_code)
+        ok_('Please refresh this page in a few minutes.' in response.content)
+
     def test_compare_revisions(self):
         """Compare two revisions"""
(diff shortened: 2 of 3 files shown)


kuma/wiki/templates/wiki/includes/revision_diff_table.html
```diff
@@ -3,4 +3,9 @@
 {% else %}
 	{% set from_content = '' %}
 {% endif %}
-{{ diff_table(from_content, revision_to.get_tidied_content(allow_none=True), revision_from.id, revision_to.id) }}
+{% set to_content = revision_to.get_tidied_content(allow_none=True) %}
+{% if from_content == None or to_content == None %}
+    {% trans %}Server is rendering the content comparison. Please refresh this page in a few minutes.{% endtrans %}
```
robhudson: "The server is rendering ..."?

robhudson:
It looks like there is an on save signal that populates the tidied content via `tidy_revision_content`. You could mock that so it doesn't actually do what it is supposed to do?

groovecoder:
Every time I have to use `mock.patch` I want to pull my brain out of my nose with a hook. WTF am I supposed to `mock.patch` again? None of these seem to work:
```
@mock.patch('kuma.wiki.tasks.tidy_revision_content')
```
```
@mock.patch('tidy_revision_content')
```
```
@mock.patch('kuma.wiki.apps.tidy_revision_content')
```
```
@mock.patch.object(WikiConfig, 'on_revision_save')
```

groovecoder:
Or is it because the `tidy_revision_content` task is executed by a signal handler during `setUp` ... maybe the mock isn't applied to `setUp`?

groovecoder:
`mock.patch` can DIAF. I tried to mock the `on_revision_save` handler from `WikiConfig` and create new revisions inside the test and it still calls the real `on_revision_save` handler instead of just returning from my mock.
```
    @mock.patch.object(WikiConfig, 'on_revision_save')
    def test_no_tidied_content(self, tidy_revision_content):
        """
        Verify revisions without tidied content show appropriate message.
        """
        tidy_revision_content.return_value = None

        user = self.user_model.objects.get(username='testuser')
        doc = _create_document(title='Test no tidied content compare')
        rev1 = doc.current_revision
        rev2 = Revision(summary="test no tidied content",
                        content='mocked on_revision_save',
                        document=doc, creator=user)
        rev2.save()

        url = reverse('wiki.compare_revisions', args=[doc.slug])
        query = {'from': rev1.id, 'to': rev2.id}
        url = urlparams(url, **query)
        eq_(200, response.status_code)
        ok_('Please refresh this page in a few minutes.' in response.content)

        url = url + '&raw=1'
        eq_(200, response.status_code)
        ok_('Please refresh this page in a few minutes.' in response.content)
```
I intend to commit without a test unless someone can explain what I'm doing wrong with `mock`. :cry:

groovecoder:
Rebased an updated with my latest test code attempting to `mock.patch.object`.

robhudson:
Ah, `on_revision_save` is doing an inline import so it's too late before it can be mocked, I believe. Mocking `on_revision_save` seems like the right thing to do in that case.

groovecoder:
But even that isn't working. :cry:

groovecoder:
@jwhitlock wins the :star: for teaching me to mock the `delay` method. ( `@mock.patch('kuma.wiki.tasks.tidy_revision_content.delay')`)

groovecoder:
After all that, I just resorted to using QuerySet `update()` to empty out the tidied content fields without triggering the signal handlers. Travis should be happy now.

kuma/wiki/tests/test_templates.py
```diff
@@ -741,6 +741,28 @@ def test_bad_parameters(self):
         response = self.client.get(url)
         eq_(404, response.status_code)
 
+    def test_no_tidied_content(self):
+        """
+        Verify revisions without tidied content show appropriate message.
+        """
+        Revision.objects.filter(
```
robhudson: maybe a comment that you're intentionally bypassing the save signal handler here? In case someone tries to update it later.

robhudson:
r+wc

groovecoder:
Updated with comment. Now this PR and https://github.com/mozilla/kuma/pull/3550 are in a race for merging & deploying first ...

PR #3549, opened 2015-10-05
bug 1184670 - run_ui_tests during chief_deploy

groovecoder:
This needs WebOps to add 4 settings values to `commander_settings` module on stage & prod.

### DIFF
--- a/tests/ui/intern-local.js
+++ b/tests/ui/intern-local.js
@@ -15,5 +15,8 @@ define(['./_base', './_cli', 'intern'], function(config, cli, intern) {
         port: 4444
     };
 
+    // Format for outputting test results
+    config.reporters = ['Pretty'];
+
     return cli.mixinArgs(intern.args, config);
 });
--- a/tests/ui/intern-local.js
+++ b/tests/ui/intern-local.js
@@ -15,8 +15,5 @@ define(['./_base', './_cli', 'intern'], function(config, cli, intern) {
         port: 4444
     };
 
-    // Format for outputting test results
-    config.reporters = ['Pretty'];
-
     return cli.mixinArgs(intern.args, config);
 });
--- a/scripts/chief_deploy.py
+++ b/scripts/chief_deploy.py
@@ -18,6 +18,7 @@ import commander_settings as settings
 venv_bin_path = os.path.join(settings.VENV_DIR, 'bin')
 os.environ['PATH'] = venv_bin_path + os.pathsep + os.environ['PATH']
 
+
 @task
 def update_code(ctx, tag):
     with ctx.lcd(settings.SRC_DIR):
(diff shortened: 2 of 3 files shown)


groovecoder:
Note: I changed the code here so that it will skip `run_ui_tests` if required settings are missing. So, this should be safe to deploy to stage without having to block on putting the settings values into `commander_settings.py`.

groovecoder:
[Chief was able to run the intern tests on this branch](http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/11031009e0bba6c5dd2ee8e4213fa6c40b438699.1444337507)!! We still got the 2 super-user related failures, so we need to debug that. But this is looking pretty good now.

groovecoder:
Rebased on master with merged https://github.com/mozilla/kuma/pull/3558 which should fix the super-user test failures here.

groovecoder:
Bah, reminder: change the log output during chief to be the ugly output.

tests/ui/intern-local.js
```diff
@@ -15,5 +15,8 @@ define(['./_base', './_cli', 'intern'], function(config, cli, intern) {
         port: 4444
     };
 
+    // Format for outputting test results
+    config.reporters = ['Pretty'];
```
darkwing: We shouldn't use "Pretty" locally.  It squashes `console.log` messages and makes debugging tests difficult.

darkwing:
How are you testing this? Simply by pushing it to stage and checking that BrowserStack kicks off the tests?

groovecoder:
Yup, pushing the branch to stage. It works; it's just still ugly log output, and the dashboard admin tests still fail on stage. :(

darkwing:
Removed the "Pretty" reporter from local

groovecoder:
How did you add that commit? It brought in a few more.

darkwing:
What in the heck... grrr

darkwing:
Fixed, sorry!

groovecoder:
Note: @jezdez pointed out we shouldn't merge this until we have a way to make chief ignore the Intern tests - e.g., if BrowserStack is unavailable.

jezdez:
To clarify, I think I'd like to prevent that we're adding a potential blocker into our deployment workflow first and foremost.

Additionally I think the problem we discuss here is a symptom of a bigger problem that Chief isn't made for being used as a piece of a system that does ad-hoc continuous delivery. In that sense I think we're missing an overarching tool that is able to integrate with GitHub PRs, is able to deploy and can have various preconditions before it does that, e.g. passing tests in the various test environment, Python and Intern/Browserstack. I think looking at [circleci.com](https://circleci.com) or [Heroku's Flow pipelines](https://blog.heroku.com/archives/2015/9/3/heroku_flow_pipelines_review_apps_and_github_sync) would make sense to solve that bigger problem.

darkwing:
@jezdez Would you suggest we hold off of integration until the AWS move?  Or would this be  a separate process?

openjck:
Looks like this needs a rebase :cry:

darkwing:
Should we just close this?

groovecoder:
Yeah. Happy to close unless/until we decide to re-open this line of work.

PR #3551, opened 2015-10-07
bug 1197858 - Set larger test timeouts for remote runners

darkwing:
Pulling this from my Safari PR.  Important we set a large timeout in case the site and/or BrowserStack are slow.


groovecoder:
Looks good; helps without hurting. Will merge when Travis lets me.

PR #3553, opened 2015-10-08
Fix bug 1212844 - Properties lists

stephaniehobson:
Lists inside of the new table world order need to be properly styled, for now we just override the rules from the old list world order but eventually we can decrease the specificity of this.

Also added display improvements for small mobile screens.

Test page:
https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip
(easier to view source and copy the table markup to local dev than to try to migrate the macros)


darkwing:
Looks awesome!  R+

PR #3555, opened 2015-10-08
Fix Bug 1203301: Standardize language in dropdown

stephaniehobson:
We currently have both pages and articles in this dropdown, needed to be standardized. Writers would like us to refer to articles instead of pages.


openjck:
Looks great! :dog:

PR #3515, opened 2015-09-24
Bug 1032455: Tweaks to helpfulness

stephaniehobson:
2 new options for "What would make it better?".
Do not ask more than once per day.

To test:
1) drop helpfulness.js line 15 to 600 for a shorter wait
2) turn `helpfulness` waffle flag on
2) clear local storage for your local MDN
3) visit page on local MDN, should be prompted (don't answer survey)
4) visit another page, should not be prompted
5) clear local storage
6) visit another page, should be prompted
7) fill out helpfulness survey
8) visit another page, should not be prompted

Not sure about testing the 'never ask again' functionality or testing if it will ask again the next day. I didn't touch the 'never ask again' mechanics so it should still be working. Maybe this is just a PR that takes 25 hours to test?

PR #3556, opened 2015-10-08
Bug 1032455: Tweaks to helpfulness

stephaniehobson:
2 new options for "What would make it better?".
Do not ask more than once per day.

Should address all concerns from #3515.


darkwing:
R+, excellent!

PR #3558, opened 2015-10-08
bug 1184670 - Don't allow admin tests on prod

darkwing:
Got more complicated than I wanted but got there in the end.

### DIFF
--- a/tests/ui/_cli.js
+++ b/tests/ui/_cli.js
@@ -1,58 +1,70 @@
 // This file allows for overriding of certain config vars via the command line
 // The local default config is assumed
-define({
-    mixinArgs: function(args, config) {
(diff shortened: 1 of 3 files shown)


tests/ui/_cli.js
```diff
@@ -1,58 +1,70 @@
...
+            // Set a document for wiki testing
+            if(args.wd == undefined) {
+                greps.push('requires-doc');
+                console.log('No wiki document (-wd) provided.  Most wiki tests will be skipped.');
+            }
 
-        return config;
+            // Don't allow some test types on prod
+            if(libConfig.productionDomain == libConfig.domain) {
+                greps.push('requires-destructive', 'requires-admin');
```
groovecoder: Does this make it impossible to to run any destructive tests on prod?

PR #3560, opened 2015-10-08
Bug 1182210 - Use full URLs for sharing

robhudson:


kuma/wiki/models.py
```diff
@@ -1479,18 +1471,9 @@ def zone_stack(self):
 
     def get_share_url(self):
```
jezdez: Do we even need that method in that case? Could we instead call it `get_full_url` so we can use it for other places?

robhudson: I was keeping it a method in case we ever need it later and it is one place to update. But I agree, I like your method name better.

kuma/wiki/tests/test_tasks.py
```diff
@@ -5,17 +5,14 @@
...
-from nose.tools import eq_, ok_
+from nose.tools import ok_
 
 from kuma.core.cache import memcache
 from kuma.users.tests import UserTestCase, user
 
-from . import revision, document
-from ..tasks import (build_sitemaps, update_community_stats,
-                     update_document_share_url)
+from . import document, revision
```
groovecoder: heh, nice :abc:

robhudson: my vim isort plugin does that for me

PR #3561, opened 2015-10-08
Fix bug 1197444: Remove media/fonts symlink

openjck:
A permanent redirect from /media/fonts to /static/fonts is now active on
stage and prod, so we don't need the symlink any more.

### DIFF
--- a/media/fonts
+++ b/media/fonts
@@ -1,1 +1,0 @@
-../kuma/static/fonts


openjck:
To test, I recommend browsing around and verifying that fonts don't 404.

groovecoder:
Looks good.

PR #3563, opened 2015-10-09
Bug 1206309: Add a feedback mechanism for compat tables

stephaniehobson:
Add a menu with options: more about beta, take survey, see old tables, report an error.

### DIFF
--- a/kuma/static/js/wiki-compat-tables.js
+++ b/kuma/static/js/wiki-compat-tables.js
@@ -9,11 +9,11 @@
     var $sectionHead = $compatWrapper.prevAll('h2').first();
     var sectionHeadIsMatch = $sectionHead.filter(function() {
             return $(this).text().match(/(browser )?compat[i|a]bility/i);
-        });
+    });
     // is it a browser compat section?
     if(sectionHeadIsMatch) {
         // come back down DOM hiding things until we reach the comapt table.
-        $sectionHead.nextUntil('.bc-api').hide();
+        $sectionHead.nextUntil('.bc-api').hide().addClass('bc-old-hide');
     }
 
     // Private var to assign IDs to history for accessibility purposes
@@ -34,6 +34,9 @@
         return $(this).each(function() {
 
             var $table = $(this);
+
+            // add beta notice & menu
+            addBetaNotice($table);
 
             // Keep track of what may be open within this table
             var $openCell;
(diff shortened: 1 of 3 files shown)


darkwing:
We need to wrap English text in `gettext`

kuma/static/js/wiki-compat-tables.js
```diff
@@ -13,9 +13,50 @@
...
+    // add beta notice & associated functonality
+    var $betaMenuWrapper = $('<div>', {class: 'bc-beta-menu'});
+    var $betaMenuTrigger = $('<a>', {text: 'New compatibility tables are in beta ', href: '/en-US/docs/New_Compatability_Tables_Beta'}).append($('<i>', {class: 'icon-caret-down', 'aria-hidden': 'true'}));
+    var $betaSubmenu = $('<ul>', {class: 'submenu js-submenu'});
+    var betaSubmenuItems = [];
+
+    var $betaLink = $('<a>', {text: 'More about the beta.', href: '/en-US/docs/New_Compatability_Tables_Beta'});
+    betaSubmenuItems.push($betaLink);
+
+    var $betaSurvey = $('<a>', {text: 'Take the survey', href: 'http://www.surveygizmo.com/s3/2342437/0b5ff6b6b8f6', class: 'external external-icon'});
```
darkwing: Should we consider opening this in a new window?  New site, and so they can flip back and look at it when they're answering?

darkwing:
Overall this looks awesome!

kuma/static/js/wiki-compat-tables.js
```diff
@@ -13,9 +13,50 @@
     // is it a browser compat section?
     if(sectionHeadIsMatch) {
         // come back down DOM hiding things until we reach the comapt table.
-        $sectionHead.nextUntil('.bc-api').hide();
+        $sectionHead.nextUntil('.bc-api').hide().addClass('bc-old-hide');
     }
 
+    // add beta notice & associated functonality
+    var $betaMenuWrapper = $('<div>', {class: 'bc-beta-menu'});
```
darkwing: We should have a space at the beginning and end of an object

stephaniehobson:
Gettexted and spaced. Couldn't get target: '_blank' to work for some reason but it's a link so if they want to open it in a new window they can CMD+click.

darkwing:
This doesn't work when there are multiple compat tables on a page.  It does put an "... is in beta" message above each table but the dropdown only displays on the second (last?) table.

I think we want to place this within the `this.each` at the top of the plugin definition.

stephaniehobson:
Updated.

brhom300:
thanks your

darkwing:
Two things:

*  `class` has been a bit dangerous as an object key, so I usually put it in quotes: `'class': ....`
*  Since what you're adding has very little to do specifically with each item, we could put all of what you're adding into a new function inside the main closure, then simply call it from within `each`.  That compartmentalizes the code better.

stephaniehobson:
Updated :)

darkwing:
At the moment I'm seeing two links; hovering over the first does nothing, hovering over the second shows two dropdowns.  I see two links even when there's only one table on the page.

twolinks

stephaniehobson:
That seems like an old bug, I'll make sure I actually pushed my updates >_<

stephaniehobson:
I'm not able to reproduce locally with 1 or 2 tables on page. Can you make sure you have the most recent version of the PR?

darkwing:
R+

PR #3564, opened 2015-10-09
Bug 1206309: Remove CSS transitions

stephaniehobson:
Remove CSS transitions from elements, set speed of javascript transition to something less stressful.


darkwing:
Way to steal my glory @stephaniehobson :D.  Nice, thank you!  R+

PR #3565, opened 2015-10-12
bug 1212767 - Update test fixture CSS class

darkwing:
Per @Elchi3 's request:  https://bugzilla.mozilla.org/show_bug.cgi?id=1212767


darkwing:
This update should have 0 impact on the tests since we don't have any selector/content checking.

PR #3566, opened 2015-10-12
Some cleanup commits from #3482.

jezdez:
This contains the first original cleanup commits from #3482


jwhitlock:
+r cleaner, closer to standard Django, and lean appropriately on 3rd-party libraries.

PR #3567, opened 2015-10-12
fix bug 1212518 - Don't forward users when control-clicking a link

darkwing:
To test:

1.  Click any breadcrumb link, should go to the page
2.  Command-click any breadcrumb link, original tab should go nowhere, new tab should open

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -375,16 +375,24 @@
     /*
         Track clicks on Crumb links
     */
-    $('.crumbs').on('click', 'a', function() {
+    $('.crumbs').on('click', 'a', function(e) {
         var url = this.href;
-
-        mdn.analytics.trackEvent({
+        var newTab = (e.metaKey || e.ctrlKey);
+        var data = {
             category: 'Wiki',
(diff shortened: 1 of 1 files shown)


stephaniehobson:
Works but... maybe a silly question... why are we preventing default on these at all? Can we just track the click and let default behaviour happen as usual?

kuma/static/js/wiki.js
```diff
@@ -375,16 +375,24 @@
...
             label: url
-        }, function() {
-            window.location = url;
-        });
+        };
+
+        if(newTab) {
+            mdn.analytics.trackEvent(data);
+        }
+        else {
```
stephaniehobson: `else` doesn't need to be on a new line.

darkwing: The majority of our site js puts else on a new line

darkwing:
We're preventing default because we want to ensure the Google Analytics call is sent and "confirmed" by Google before the user moves forward. Without `preventDefault`, the analytics tracking may get "cut off" by a fast page load

stephaniehobson:
R+

PR #3568, opened 2015-10-12
Bug 1209185 - Pre django 1.8 updates

robhudson:


robhudson:
Reassigning since Luke is deploying other things. :)

Makefile
```diff
@@ -17,7 +17,7 @@ browser-tests: on_host
 	pushd tests/ui ; ./node_modules/.bin/intern-runner config=intern-local d=developer.allizom.org b=firefox; popd
 
 clean:
-	find kuma -name '*.pyc' -exec rm {} \;
+	find . -name '*.pyc' -exec rm {} \;
```
jezdez: That'll take a lot longer since it'll also go through all of `vendor/`

jezdez:
The faking of the authkeys migration need to happen in the chief deploy script

Makefile
jwhitlock: You'll need to exclude or restore ``vendor/src/python-magic/magic.pyc``, which is (probably incorrectly) checked into the the python-magic repo.

robhudson: I can back this out.  I don't see how it relates to django 1.8.

robhudson:
@jezdez Has that happened before? On Marketplace we would put in notes for the deploy to do this manually.

groovecoder:
Oops, yeah the `Makefile` change was just something I changed while switching branches. "Take a lot longer" means a few seconds here, and it saves minutes-hours of frustration when switching between branches with different `vendor/src` submodule files.

brhom300:
+1805663177

brhom300:
3568

robhudson:
@groovecoder Would you be interested in coordinating this push? Both look like they will need manual SQL run. (1) for the "fake" authkeys migration which can be inserted via SQL shell prior to push, and (2) for migration reset which I'm not sure when it would be updated (before or after?).

jezdez:
@groovecoder true indeed!

groovecoder:
I can coordinate this merge & deploy with WebOps. It may take me a few days at least though.

groovecoder:
We should be able to do both SQL updates before the merge & deploy:
```
INSERT INTO django_migrations (app, name, applied) values ('authkeys', '0001_initial', NOW());
DELETE FROM django_migrations where app = 'users';
INSERT INTO django_migrations (app, name, applied) values ('users', '0001_initial', NOW());
```
After that I see the same warning from `migrate` that we see on our deploy logs:
```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```
The `makemigrations` creates some migrations in `wiki`, `demos`, and `users` tables:
```
Migrations for 'wiki':
  0010_auto_20151022_0900.py:
    - Alter field locale on document
    - Alter field locale on documentdeletionlog
Migrations for 'demos':
  0003_auto_20151022_0900.py:
    - Alter field demo_package on submission
    - Alter field screenshot_1 on submission
    - Alter field screenshot_2 on submission
    - Alter field screenshot_3 on submission
    - Alter field screenshot_4 on submission
    - Alter field screenshot_5 on submission
Migrations for 'users':
  0002_auto_20151022_0900.py:
    - Alter field timezone on user
```
We should be able to merge & deploy without blocking on fixing those.

PR #3569, opened 2015-10-12
Fix bug 1213229 - Remove duplicate tag warning

robhudson:


jezdez:
I left a comment in the ticket that this would (I think) get rid of the ability to detect wrong casings for new tags once discovered, but if we have other means to fix them via the Django admin we should be good. 

The code change itself looks good. r+

kuma/wiki/forms.py
```diff
@@ -342,20 +339,21 @@ def clean_tags(self):
...
-
                 # Write a log we can grep to help find pre-existing duplicate
                 # document tags for cleanup.
                 if len(doc_tag) > 1:
                     log.warn('Found duplicate document tags: %s' % doc_tag)
 
-            if dupe_tags:
-                raise forms.ValidationError(
-                    TAG_DUPE % {'tag': get_text_list(dupe_tags)})
+                if doc_tag:
```
jezdez: Oh, one thing, that'd be a great fit for `first()` as it returns `None`

robhudson: Would `first()` trigger another query with a `LIMIT 1`? I'm trying to avoid 2 queries here. I could drop the logging and use `first()`, though. Either we haven't hit the case of more than one tag being found or the logging isn't set up at the right log level to see the logs.

robhudson: I'm going to leave this as-is to avoid a 2nd query. When we decide we can remove the logging we can update this to use `first()` instead. Or even `get()` for that matter.

PR #3570, opened 2015-10-13
fix bug 1177797 - Only focus on next avatar if enter key pressed

darkwing:
https://bugzilla.mozilla.org/show_bug.cgi?id=1177797

To test:

1.  Click on the "Show all" button....see no "larger" avatar
2.  Tab to "Show all" button, press `ENTER` key, see larger avatar


stephaniehobson:
Pressing tab while on "show all" now expands the full list. We should be checking `e.which === 13` before preventing default.

darkwing:
Great catch @stephaniehobson .  Been distracted by Jack's illness today, this should be set now.

stephaniehobson:
R+

PR #3571, opened 2015-10-13
Fix Bug 1213548: Min-width on .properties headings.

stephaniehobson:
Also moved some definitions around to decrease the number of lines in the complied Stylus files.

Test:
- using dev tools DOM inspector copy the HTML from this page into an article on your local instance: https://developer.mozilla.org/ja/docs/Web/CSS/flex-basis (There are several kumascript macros that need updating if you want to try to duplicate the output by actually editing the article.)
- check https://developer.mozilla.org/en-US/docs/User:teoli/In_content to make sure the other summery boxes are still displaying properly.


darkwing:
R+, they look good to me!

PR #3575, opened 2015-10-15
Bug 1209185 - Move locale/ folder from Subversion to Git

robhudson:
Before merging we need to let `#mathjazz` know to coordinate with a change in verbatim and pontoon configuration.


groovecoder:
OMG :heart_eyes: Yay!

.gitignore
```diff
@@ -37,3 +36,5 @@ kuma/static/js/libs/ckeditor/source/ckbuilder
 .env
 *devmo*.sql*
 htmlcov/
+locale/**/z-*
```
jezdez: What are those two entries for?

jezdez:
When I checked out the branch I have lots of `*.mo` files in the locale dir that git didn't know about. Assuming we don't want to commit the binary `mo` files into the repo we should add them to the .gitignore as well.

Could you split the adding of the files and the other modifications into separate commits so it's easier to review here?

jezdez:
This needs to be updated: https://github.com/mozilla/kuma/blob/d4f91f726007e65777917943bac8e4c6c4e387bf/kuma/humans/models.py#L63-L68

jezdez:
rwc+

.gitignore
robhudson: They were there before so I chose to leave them. I can look into removing them.

robhudson:
> Could you split the adding of the files and the other modifications into separate commits so it's easier to review here?

I thought I did. The addition of the locale files and updating the .pot files are together b/c the .pot files were apparently pretty far out of date and were large but I could separate those?

jezdez:
@robhudson Ah, that explains it. Fair enough. r+

PR #3579, opened 2015-10-16
[fix bug 1119551] Change fields on admin userprofile list

hoosteeno:


kuma/users/admin.py
```diff
@@ -23,20 +24,29 @@ class UserAdmin(BaseUserAdmin):
...
+            ('?user=%s' % obj.username)
+        count = obj.wiki_revisions().count()
+        return ('<a href="%(link)s"><strong>%(count)s</strong></a>' %
+            {'link': link, 'count': count})
 
+    revisions.allow_tags = True
+
+    def website(self, obj):
+        """HTML link to user's website""" 
+        return ('<a href="%(link)s"><strong>%(url)s</strong></a>' %
```
hoosteeno: If this is valuable, I need to verify whether the content of the field is sanitized. Otherwise, this presents an attack vector.

jwhitlock: We do some [basic validation](https://github.com/mozilla/kuma/blob/master/kuma/users/models.py#L99-L104) of website URLs, so I think we're safe.

Since it is the same value, this could be written:

```python
       return ('<a href="%(url)s"><strong>%(url)s</strong></a>' %
               {'url': obj.website_url})
```

hoosteeno: OK. The vector I was imagining was a django admin URL (which would pass through that validation), but I can only find one admin URL that changes state without POSTed confirmation: https://developer.allizom.org/admin/logout/.

Unless you know other admin URLs that change state without POSTed confirmation, this is probably not a risk.

kuma/users/admin.py
```diff
@@ -1,3 +1,4 @@
+from django.core.urlresolvers import reverse
```
robhudson: We should be using `from kuma.core.urlresolvers import reverse`.

kuma/users/admin.py
```diff
@@ -23,20 +24,29 @@ class UserAdmin(BaseUserAdmin):
...
 
-    formfield_overrides = {
-        NamespacedTaggableManager: {
-            'widget': TagWidget(attrs={'size': 45})
-        }
-    }
+    def revisions(self, obj):
+        """HTML link to user's revisions with count""" 
+        link = reverse('dashboards.revisions') +\
+            ('?user=%s' % obj.username)
```
robhudson: You could use:
```
from jingo.helpers import urlparams

link = urlparams(reverse('dashboards.revisions'),
                user=obj.username)
```

hoosteeno: Nice. I was looking for something like that but didn't know where to look.

kuma/users/admin.py
```diff
@@ -23,20 +22,29 @@ class UserAdmin(BaseUserAdmin):
...
-    formfield_overrides = {
-        NamespacedTaggableManager: {
-            'widget': TagWidget(attrs={'size': 45})
-        }
-    }
+    def revisions(self, obj):
+        """HTML link to user's revisions with count"""
+        link = urlparams(reverse('dashboards.revisions'),
+                         user=obj.username)
+        count = obj.wiki_revisions().count()
```
robhudson: I just noticed that the `wiki_revisions` is slicing the query so you'll always get 5 here. You probably want to duplicated some of what that method is doing here to get the full count. Also, experiment with the query b/c this will get called once for every user in the table so it could cause some database read usage we should be careful about.

robhudson: Would you ever want to sort by this count? Or is it just an informational field only?

hoosteeno: I don't think sorting by count is necessary at this point. It's not a terribly valuable metric.

hoosteeno: Thanks for the tip re: wiki_revisions. Will build my own and try to make it good.

kuma/users/admin.py
```diff
@@ -23,20 +22,29 @@ class UserAdmin(BaseUserAdmin):
...
+                         user=obj.username)
+        count = obj.created_revisions.count()
+        return ('<a href="%(link)s"><strong>%(count)s</strong></a>' %
+                {'link': link, 'count': count})
+
+    revisions.allow_tags = True
+
+    def website(self, obj):
+        """HTML link to user's website"""
+        return ('<a href="%(url)s"><strong>%(url)s</strong></a>' %
```
jezdez: If the user doesn't have a website URL that would create an empty link  (with `None`) I think. Could you check for the emptiness of the field before and return an empty strings if empty?

kuma/users/admin.py
```diff
@@ -23,20 +22,29 @@ class UserAdmin(BaseUserAdmin):
...
-    formfield_overrides = {
-        NamespacedTaggableManager: {
-            'widget': TagWidget(attrs={'size': 45})
-        }
-    }
+    def revisions(self, obj):
+        """HTML link to user's revisions with count"""
+        link = urlparams(reverse('dashboards.revisions'),
+                         user=obj.username)
+        count = obj.created_revisions.count()
```
jezdez: Depending on the number of users to display this may end up slowing down the admin page as counts are slow. Is the count really needed?

hoosteeno: The count is definitely needed, at least for counts < 25 or so. If you know off the top of your head a way to do this more quickly I'm all ears; otherwise I can dig around for alternatives.

kuma/users/admin.py
hoosteeno: Yes!

kuma/users/admin.py
jezdez: Not really, let's go with this for now and see how it impacts real life performance

groovecoder:
@hoosteeno, @jezdez - Does this need anything else or can it be merged?

hoosteeno:
I think it is ready.

kuma/users/admin.py
```diff
@@ -23,20 +22,31 @@ class UserAdmin(BaseUserAdmin):
...
+        return ('<a href="%(link)s"><strong>%(count)s</strong></a>' %
+                {'link': link, 'count': count})
+
+    revisions.allow_tags = True
+
+    def website(self, obj):
+        """HTML link to user's website"""
+        if (obj.website_url):
```
jezdez: no braces needed here

jezdez:
rwc+

PR #3580, opened 2015-10-16

openjck:
## Setup

1. Pull down this branch
2. `rm -rf kuma/static/css` (this directory isn't used any more, and will cause problems if present)

## Testing

We'll want to be sure everything looks the same on this branch. Good things to test include CKEditor, template pages, and syntax highlighting.

[Production mode](https://github.com/openjck/kuma/blob/bug-1139490-django-pipeline-working-travis/docs/development.rst#production-assets) works just like before, minus the `compile-stylesheets` and `compress_assets` steps.


openjck:
cc @robhudson for the back-end bits. Sounds like you gave that part an r+ earlier?

openjck:
I'm having some trouble getting CKEditor to load reliably in production mode. I'll need to figure that out on Monday. Aside from that, my testing hasn't turned up anything else.

openjck:
> If a stylesheet is edited, how is it updated now? Or in developer mode this doesn't matter any longer? (Which would be great).

In development mode, django-pipeline watches Stylus files for changes and compiles them on the fly during page load. In production mode, Stylus is compiled and compressed when `collectstatic` is run.

robhudson:
I looked on Friday, backend looked good to me. (Forgot to comment)

openjck:
The CKEDITOR issue has been fixed, so this is ready for final review. See pull request description for some tips.

openjck:
Updated to add the *compat-icons.woff2* file

stephaniehobson:
From my testing in dev mode:

- edit profile page throws console error `TypeError: $(...).tagit is not a function`
- dev mode is really hard on my machine. I opened 11 pages at once and 3 of them arrived with no CSS (worked fine on a refresh).
- I couldn't edit some pages. Example: [progress](https://developer-local.allizom.org/en-US/docs/Web/HTML/Element/progress$edit). Browser said: `A server error occurred.  Please contact the administrator.` [Terminal said...](https://gist.github.com/stephaniehobson/b0122778f7bc1948ee54) 

(Testing continues tomorrow in prod mode)

Also (not a problem with your code): the MDN 10 zone links to a CSS file directly, that will need to be updated by hand or just copied into the zone, there are not likely to be further changes to it.

openjck:
Thanks for the report. I'll dig into them tomorrow.

> I couldn't edit some pages. Example: progress. Browser said: A server error occurred. Please contact the administrator. Terminal said...

It looks like that might have something to do with attachments not being available. Is that page able to be edited on master?

openjck:
> edit profile page throws console error TypeError: $(...).tagit is not a function

I'm also seeing this on master. Definitely something we should follow up with though.

> dev mode is really hard on my machine. I opened 11 pages at once and 3 of them arrived with no CSS (worked fine on a refresh).

Did those pages also load simultaneously? On this branch, Stylus is compiled on-the-fly, so 11 pages loading simultaneously could be a bit much.

Once the stylesheets are compiled, they won't be compiled again until they change. So after an initial period of worse-than-average performance the site should be back to normal. Are subsequent page loads equally slow?

openjck:
@jezdez: Can you think of a reason these changes would hurt performance locally? Some people online seem to say that a different `STATICFILES_FINDERS` entry may be necessary.

stephaniehobson:
@openjck yes, every page load takes between 11 and 15 seconds.

stephaniehobson:
(in dev mode, prod mode is super quick after the initial load)

stephaniehobson:
This is looking good in prod mode! Nothing more to add to yesterday's testing :)

stephaniehobson:
@openjck Yes, also have trouble editing that page on master, attachments seem to be the problem.

stephaniehobson:
R+ with the backend review from @robhudson this is good to go!

PR #3581, opened 2015-10-16
Fix bug 1215670 - Correct multiword tag form processing

robhudson:


brhom300:
http://www.microsoft.com/october2015event/ar-us

kuma/wiki/tests/test_forms.py
```diff
@@ -105,6 +105,17 @@ def test_form_rejects_empty_slugs_with_parent(self):
         rev_form = RevisionForm(data, parent_slug='User:groovecoder')
         ok_(not rev_form.is_valid())
 
+    def test_multiword_tags(self):
+        rev = revision(save=True)
+        data = {
+            'content': 'Content',
+            'toc_depth': 1,
+            'tags': '"MDN Meta"',  # Note the lower-case "S".
```
jezdez: Code comment isn't accurate

jezdez:
Hm, will this work with the current set of tags?

kuma/wiki/tests/test_forms.py
robhudson: Oh, woops :)

jezdez:
It's odd, that we have to do this parsing on our own, for some reason. Is there some features in django-taggit that we can re-use?

robhudson:
> Hm, will this work with the current set of tags?

I'm not exactly sure what you mean? I tested a few commonly used tags like "MDN Meta". But once they page is saved and they are split apart b/c of this bug, this code will not fix them. It would require the writers to clean this up, which was mentioned in bugzilla. Which is why I would like to merge this soon.

> It's odd, that we have to do this parsing on our own, for some reason. Is there some features in django-taggit that we can re-use?

I haven't seen any code to help with this, and this is code that could be later removed with a move to Postgresql which is a code comment.

django-taggit has a setting, `TAGGIT_CASE_INSENSITIVE` which defaults to `False` but we don't want case insensitive tags. It basically switches to an `__iexact` instead of an `__in` query. See https://github.com/alex/django-taggit/blob/develop/taggit/managers.py#L163

This is basically working around the bug in django/mysql that `__in` and `__exact` are case insensitive to begin with.

jezdez:
Ok, thanks for the explanation.

jezdez:
r+

PR #3586, opened 2015-10-21
Bug 1139490: Remove cache-busting params in styles

openjck:
collectstatic automatically updates stylesheets to cache-bust images and
other assets when their contents change, so these manual cache-busting
parameters are no longer needed.


stephaniehobson:
R+

PR #3587, opened 2015-10-21
Bug 1206309: Don't display if no actual content in row

stephaniehobson:


darkwing:
In what case is this popping up?  It seems like a problem elsewhere...

stephaniehobson:
It's a side effect of how I'm handling the new Basic Support world order. I'll check there to see if I can handle it better.

darkwing:
Cool.  In theory there should never be a tabble without a row.

PR #3593, opened 2015-10-22
fix bug 1215422 - added space and used class to fix width

shiningfm:


stephaniehobson:
Perfect, thanks for rebasing, sorry for the trouble. And thank you for the patch!

I don't know if you have time for another bug... but if you do there's another bug on the same few lines of code you were working on for this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1216873

PR #3594, opened 2015-10-22
Bug 1206309: Listen for touch event on cells.

stephaniehobson:


darkwing:
Wont this trigger two events?  What isn't `click` catching?

stephaniehobson:
Because `<td>` isn't normally clickable and there is a hover effect on the `<td>` the first tap is being treated as a hover and the second tap is being treated as a click :/ I want the first tap to be treated as a click. 

I added the preventDefault to stop a click from being triggered as well, it works in the iOS simulator.

darkwing:
Looking

darkwing:
R+!

PR #3595, opened 2015-10-23
Bug 1217513: Made .callout-box more RTL and mobile friendly

stephaniehobson:
Test pages: https://developer.mozilla.org/en-US/search?locale=en-US&css_classnames=callout-box


openjck:
This is awesome. :raised_hands:

PR #3596, opened 2015-10-23
Bug 1139490: Use sourcemaps in DEBUG mode

openjck:
Big red button labeled "Easy"

### DIFF
--- a/provisioning/roles/kuma/files/vagrant/settings_local.py
+++ b/provisioning/roles/kuma/files/vagrant/settings_local.py
@@ -108,3 +108,6 @@ if SENTRY_DSN:
     )
 
 SOCIALACCOUNT_PROVIDERS['persona']['AUDIENCE'] = 'https://developer-local.allizom.org'
+
+if DEBUG:
+    PIPELINE_STYLUS_ARGUMENTS = '--sourcemap'
--- a/settings.py
+++ b/settings.py
@@ -564,6 +564,8 @@ PIPELINE_COMPILERS = (
     'pipeline.compilers.stylus.StylusCompiler',
 )
 
+PIPELINE_STYLUS_ARGUMENTS = ''
+
 PIPELINE_CSS_COMPRESSOR = 'kuma.core.pipeline.cleancss.CleanCSSCompressor'
 PIPELINE_JS_COMPRESSOR = 'pipeline.compressors.uglifyjs.UglifyJSCompressor'
 


darkwing:
Mozilla doesn't do corporate endorsements, please reconsider

openjck:
To test:

1. `vagrant provision`
2. Load the homepage (may or may not need a hard refresh)
3. Inspect an element and confirm that Stylus file names and Stylus line numbers are shown in the *Rules* panel

robhudson:
I would include the default setting (empty string?) in `/settings.py` with the other pipeline settings (and a comment to set it to true in debug mode for sourcemaps?).

openjck:
> I would include the default setting (empty string?) in /settings.py with the other pipeline settings (and a comment to set it to true in debug mode for sourcemaps?).

I like the idea of keeping this near other Pipeline settings. One challenge with putting it in *settings.py* is that the file is version-controlled, so someone could accidentally commit their modification to the setting. Would moving these two lines (including the `DEBUG` check) to *settings.py* make sense?

robhudson:
I think production-like settings should go in `settings.py` which to me means this should be in `settings.py` but set to empty string if we don't want sourcemaps there yet. But in `local_settings.py` we can add the `if DEBUG` check and enable it, so at least it is in the main settings file with a reasonable default.

openjck:
Okay cool. Updated. Does that look about right?

robhudson:
Yep, :+1:

stephaniehobson:
Is there a reason to not leave sourcemaps on for production?

openjck:
> Is there a reason to not leave sourcemaps on for production?

I'd love to enable sourcemaps on production. That would be really handy. Unfortunately they don't work at the moment. The compression step that happens after compilation breaks the sourcemap.

Newer versions of *clean-css* can generate sourcemaps, but since django-pipeline runs compressors after compilers that would give us a map from the compressed CSS to the non-compressed CSS, not to the original Stylus. Once django-pipeline [officially supports sourcemaps](https://github.com/cyberdelia/django-pipeline/issues/207) it might do something more intelligent to get sourcemaps to work from the source Stylus all the way to compiled and compressed CSS.

stephaniehobson:
R+ works like it says on the tin.

Anyone else need to R this before it goes?

robhudson:
:shipit:

PR #3597, opened 2015-10-23
Bug 1139490: Improve local Pipeline performance

openjck:


openjck:
This is similar to something we tried last week, with one change that could make a big difference. If this doesn't do the trick, we can at least use this code as a reference point to figure out what's going wrong.

openjck:
I can reproduce the problem now.

@jezdez This is what the Pipeline documentation [recommends](https://django-pipeline.readthedocs.org/en/latest/storages.html) but pages still load slowly for me locally. Is there anything else we should be considering?

jezdez:
In addition I would add a custom manifest storage backend (e.g. `kuma.core.storage`) and set `STATICFILES_STORAGE = 'kuma.core.storage.PipelineManifestStorage'`:

```python
from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
from pipeline.storage import PipelineMixin


class PipelineManifestStorage(PipelineMixin, ManifestStaticFilesStorage):
    pass
```

That makes use of the faster manifest driven approach (e.g. https://www.djangoproject.com/s/staticfiles.json) instead of having to call the cache per file.

openjck:
Okay, sounds great.

I made the updates, but performance is still pretty bad locally. @jezdez does the `PipelineManifestStorage` addition look good at least? If that's fine I can split it into a separate PR.

settings.py
```diff
@@ -415,8 +415,9 @@ def get_locales():
 )
 
 STATICFILES_FINDERS = (
-    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
-    'django.contrib.staticfiles.finders.FileSystemFinder',
+    'pipeline.finders.AppDirectoriesFinder',
+    'pipeline.finders.FileSystemFinder',
```
jezdez: It turns out this finder tries ignores files in a `build` folder, effectively making our system moot since we use "build" as the name of the folder we generate the built files into

jezdez: https://github.com/cyberdelia/django-pipeline/blob/7bac3c60979bbbe926c6a99842b0325d95c2377d/pipeline/finders.py#L113

jezdez:
So I'm having trouble getting this improved since I think we're seeing a design issue with the way we've set up pipeline. It needs to compile those files in the packs (`PIPELINE_CSS` and `PIPELINE_JS`) on every request that match the configured compiler's matching file extension (`.styl`) since that's the only way we'll be able to see its output correctly. Pipeline simply doesn't contain the ability to not compile if there are any found. That fact has an even bigger impact on speed since we have simply *a lot* of them (208 at the moment, although 37 are only referenced in the `PIPELINE_CSS` setting). So on each and every request we compile all those files since otherwise pipeline wouldn't be able to keep its promise to serve ready to consume files.

I think this is a bummer and I admit I didn't see this coming. I would suggest to take one step back and let stylus do the compilation to CSS and then use the rest of the pipeline method.

- re-add `stylus --watch kuma/static/styles --out <STATIC_ROOT>/styles` to the `Procfile` (``--out`` option is used to store resulting css files outside the committed code and next to the rest of the files. This needs `pipeline.finders.PipelineFinder` in the `STATICFILES_FINDERS` though so pipeline can find the generated files there)
- re-add `stylus` call to the chief deploy script before the collectstatic call

Sorry for the quick message, I'm afraid this needs more work and I'd be interested in helping you in whatever way I can.

groovecoder:
Dang, yeah this is pretty painful now. :cry:

openjck:
For the time being we can use production mode as a workaround. Pages load in no-time with it enabled.

openjck:
Superseded by #3614

PR #3599, opened 2015-10-26
Bug 1119397: Use Bower to manage jQuery dependency

openjck:
Testing this change:

1. `vagrant provision`
2. `vagrant ssh`
3. Verify that `bower-installer` works
4. Verify that `./manage.py collectstatic --noinput` works
5. Browse around the site and verify that jQuery still works as expected


stephaniehobson:
R+

PR #3600, opened 2015-10-26
Fix Bug 1218466: Update javascript and ajax trackError calls.

stephaniehobson:


stephaniehobson:
I think I'm going to take this opportunity to change how we handle this a bit.

PR #3601, opened 2015-10-26
Fix bug 1119397: Manage remaining straightforward dependencies with Bower

openjck:
Testing this change:

1. `vagrant provision`
2. `vagrant ssh`
3. Verify that `./manage.py collectstatic --noinput` works
4. Browse around the site and verify that things still behave as expected (I don't expect anything too weird to happen here since the files were only renamed)


darkwing:
It would be aces to get Prism into this eventually, if possible.  Would be easier to update from time to time, get the languages compiled that we need, etc.

openjck:
> It would be aces to get Prism into this eventually, if possible. Would be easier to update from time to time, get the languages compiled that we need, etc.

Definitely. I want to include Prism and other libraries in the future. For now, this pull request just migrates simple dependencies that don't have any build steps.

What does the build step for Prism look like? Unfortunately Bower doesn't offer much support with custom builds (https://github.com/bower/bower/issues/369) but maybe we can think of something.

darkwing:
We wouldn't need a custom build from Prism; we'd use bower to grab the repo and then use `PIPELINE_CSS` and `PIPELINES_JS` to pick the language support we want.

openjck:
> We wouldn't need a custom build from Prism; we'd use bower to grab the repo and then use PIPELINE_CSS and PIPELINES_JS to pick the language support we want.

Ah, okay. That makes sense. I saw [this file](https://github.com/mozilla/kuma/blob/master/kuma/static/js/libs/prism/prism.js) and thought we needed to do something special to generate it. But it sounds like we could just cherry-pick individual components of that file and have Pipeline compress them into something equivalent? That would work. If you could review that I could tackle it after this PR lands.

stephaniehobson:
+18 −159. Like a boss. R+1

PR #3602, opened 2015-10-26
Watch tree 842340

groovecoder:
Spot-check instructions:

1. Sign in
2. Go to a doc with a sub-article
3. Under the advanced menu, click the "Subscribe to this article and all its sub-articles"
    * [x] Verify the subscription notification UI is correct
4. Sign in as a different user
5. Go to the sub-article
6. Edit the sub-article
7. Check the `./manage.py worker` console output.
    * [x] Verify our notification email is sent to the user who subscribed to the article + sub-articles. E.g.:
```
...
Subject: [MDN] Page "groovechild" changed by DavidWalshOS
From: notifications@developer.mozilla.org
...
--
You are subscribed to edits on: User:groovecoder and all its sub-articles.
Unsubscribe from these emails:
https://example.com/unsubscribe/14?s=dCvbUfbpab
```
BONUS spot-check waffle flag:
8. [Add a waffle flag](https://developer-local.allizom.org/admin/waffle/flag/add/) called `watch_menu` and activate it for a user
9. Visit a doc with a sub-article as the user
   * [x] Verify there's a new UI "eye" element with the Subscribe links inside

* [x] Tests
* [x] Analytics tracking for split-testing
* [x] [Bug 1218448](https://bugzilla.mozilla.org/show_bug.cgi?id=1218448) - Subscribe to article: notification never dismisses
* [x] Feedback from @stephaniehobson on this UI approach: moving all of the "subscribe" page items into their own page nav menu:
doc-watch-menu

On non-English pages, there are 3 items in the menu:

doc-watch-menu-fr


stephaniehobson:
**TL;DR** I like this direction I want to A/B test to make sure we don't lose edits to clutter.

I don't know if it's a good idea to be adding buttons to the interface. Another display option here would be to separate them within the menu using a dotted line.

screen shot 2015-10-26 at 21 00 43

But I know that you know that subscriptions increase return visits so I'm not sure how to weigh more subscriptions vs clutter. Maybe we could get @hoosteeno to weigh in?

If we're going to increase the prominence of these options now's a good time to play around with them. There are so many good opportunities for A/B testing here:
- "subscribe to this article" vs "email me when this article updates"
- eye icon vs feed icon vs email icon (the eye icon is occasionally used for accessibility options *probably* not often enough that it's a problem but worth testing)

jezdez:
Minor nit, I think "children" is a bit of a misleading term for describing subpages as it implies that users know our data tree model. Could we call it "section"? Or maybe "subpages"?

groovecoder:
Great comments.

I'll re-use the term "sub-articles" that we use in the advanced menu. (i.e., "New sub-article")

I'll start by split-testing between:

* Separate "Subscribe" menu
* "Subscribe" menu options in the Advanced menu, separated by a dotted line

Let's test that for a week or more, then look at testing verbage and iconography.

@stephaniehobson - do we already have an element (`<hr/>`?) or class for an `<li>` separator?

groovecoder:
Updated with a `watch_menu` flag to allow us to split-test ...

mdn-watch-tree-split

Need to add tests now.

groovecoder:
Okay, this is ready for final review & spot-checking. Who wants it? @stephaniehobson for front-end & spot-check? @robhudson, @jwhitlock, or @jezdez for back-end code review?

groovecoder:
Also rebased this on latest master.

stephaniehobson:
If you put two lists next to each other in a navigation drop down the border should appear automatically, sorry, should have said.

stephaniehobson:
<progress> includes the link to subscribe to sub-articles but it [has no sub-articles](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress$children.) 

Some <bdi> tags around the language will save us some RTL headaches.
```
Subscribe to <bdi>English (US)</bdi> version
```

After subscribing to a tree, when visiting a child article there is no indication for the user that they are already subscribed to that article. Is it even possible to subscribe to a tree and than unsubscribe from a child? If the user is subscribed to both the tree and the child will they get duplicate email notifications? Is this a toggle - if I subscribe to the tree and then subscribe to the article have I actually unsubscribed?

I think we can answer some of these questions by including some text in the menu if they are subscribed to the tree already. Something like "You are subscribed to this article's parent and all of its children", only more clear and friendly than what I scribble in a github comment box 8 minutes before I have to leave to catch a flight.

Wasn't able to test the emails actually sending, happy to do it on Friday if it still needs doing but I'll need more detailed steps. Anyway, going to go get on a plane now.

groovecoder:
Great points. Some responses ...

1. Users have [asked to receive notifications when new sub-articles are created](https://bugzilla.mozilla.org/show_bug.cgi?id=857775#c1) under an article, so we need to give them the ability to watch pages that don't have sub-articles (yet).

2. Commit incoming for `<bdi>`

3. The type of subscriptions are separate. If you subscribe to the tree and then subscribe to a sub-article, you are subscribed to both. So it **is** possible to subscribe to a tree **and** and then separately (un)subscribe from/to a child. See also ...

4. If a user is subscribed to a document via both a tree and direct subscription, they will only get 1 email notification, thanks to [`django-tiding`'s `EventUnion` class and container pattern](https://django-tidings.readthedocs.org/en/latest/introduction.html#the-container-pattern) The email notification includes a direct unsubscribe link. So, if the user is subscribed to a tree and to a sub-article in the tree, they will have to un-subscribe from both via the email notification links. The footer text in the email tells them which subscription type email they're receiving.

I can get any parent trees to which the user is subscribed; presenting them in the menu will be a challenge, but I'll try a few things ...

jwhitlock:
Is there a view where a user can see their subscriptions? Something like https://en.wikipedia.org/wiki/Special:EditWatchlist ?

groovecoder:
@jwhitlock - good point on the view names. Should that change block merge?

There's no user/profile page currently listing a user's subscriptions (along with unsubscribe links), but it should be trivial to add it. Hardest part will be finding a good spot/layout in the UI for it. @stephaniehobson  - Maybe on the right side of the "Edit Profile" page - under Firefox Apps & Hacks Newsletter block?

jwhitlock:
No, I don't think view names or subscription views should block. We've gotten along in the current state for long enough.  We can handle these issues in follow-up PRs, while we get feedback on the per-page UI.

I can't see any code issues, starting practical testing

kuma/wiki/templates/wiki/email/edited.ltxt
```diff
@@ -32,4 +32,8 @@ Edit Article:
 Article History:
 {% endtrans %}
  {{ history_url|absolutify }}
-{% if watch %}{{ unsubscribe_text(watch) }}{% endif %}{% endautoescape %}
+--
+{% if watch %}
+{% trans %}You are subscribed to edits on:{% endtrans %} {{ watch.content_object.title }}{% if watch.event_type == 'wiki edit document in tree' %}{% trans %} and all its sub-articles{% endtrans %}{% endif %}.
```
robhudson: I don't think it is good to break apart a sentence into multiple chunks like this for l10n. I imagine RTL languages would have a difficult time with this.

Would something like this be easier for localizers?
```
{% if watch %}
  {% set title = watch.content_object.title %}
  {% if watch.event_type == 'wiki edit document in tree' %}
    {{ _('You are subscribed to edits on: {title} and all its sub-articles.')|f(title=title) }}
  {% else %}
    {{ _('You are subscribed to edits on: {title}.')|f(title=title) }}
  {% endif %}
{% endif %}
```

kuma/wiki/templates/wiki/includes/document_macros.html
```diff
@@ -113,3 +113,44 @@
...
+                        {{ csrf() }}
+                        <a href="#" data-subscribe-status="{{ subscribe_status }}" data-subscribe-text="{{ subscribe_text }}" data-unsubscribe-text="{{ unsubscribe_text }}" data-subscribe-message="{{ _('You are now subscribed to this article.') }}" data-unsubscribe-message="{{ _('You have been unsubscribed from this article.') }}">
+                        {% if document.is_watched_by(user) %}{{ unsubscribe_text }}{% else %}{{ subscribe_text }}{% endif %}
+                        </a>
+                    </form>
+                  </li>
+
+                  <li class="page-watch">
+                      {% set subscribe_tree_text = _('Subscribe to this article and all its sub-articles') %}
+                      {% set unsubscribe_tree_text = _('Unsubscribe from this article and all its sub-articles') %}
```
robhudson: Indenting seems off here

kuma/wiki/templates/wiki/includes/document_macros.html
```diff
@@ -113,3 +113,44 @@
...
+                    <form action="{{ url('wiki.subscribe_to_tree', document.slug, locale=document.locale) }}" method="post">
+                        {{ csrf() }}
+                        <a href="#" data-subscribe-status="{{ subscribe_status }}" data-subscribe-text="{{ subscribe_tree_text }}" data-unsubscribe-text="{{ unsubscribe_tree_text }}" data-subscribe-message="{{ _('You are now subscribed to this article and all its sub-articles.') }}" data-unsubscribe-message="{{ _('You have been unsubscribed from this article and all its sub-articles.') }}">
+                        {% if document.tree_is_watched_by(user) %}{{ unsubscribe_tree_text }}{% else %}{{ subscribe_tree_text }}{% endif %}
+                        </a>
+                    </form>
+                  </li>
+
+                    {% if document.parent %}
```
robhudson: and here?

kuma/wiki/tests/test_views.py
```diff
@@ -2549,6 +2550,137 @@ def _check_message_for_headers(message, username):
...
+        EditDocumentEvent.notify(testuser01, rev.document)
+
+        data.update({'form': 'rev',
+                     'slug': rev.document.slug,
+                     'content': 'This edit should send 2 emails',
+                     'comment': 'This edit should send 2 emails'})
+        self.client.post(reverse('wiki.edit',
+                                 args=[rev.document.slug]),
+                         data)
+        self.assertEquals(3, len(mail.outbox))
```
robhudson: The comment says 2 emails but this is 3?

robhudson: Oh is it 2 plus the 1 prior?

kuma/wiki/tests/test_views.py
```diff
@@ -2549,6 +2550,137 @@ def _check_message_for_headers(message, username):
...
+        self.client.login(username='testuser', password='testpass')
+        data = new_document_data()
+        data.update({'form': 'rev',
+                     'slug': child_doc.slug,
+                     'content': 'This edit should send an email',
+                     'comment': 'This edit should send an email'})
+        self.client.post(reverse('wiki.edit',
+                                 args=[child_doc.slug]),
+                         data)
+        eq_(1, len(mail.outbox))
```
robhudson: I think it would be good to check something basic in the emails (the subject?) to also verify it's the email you are expecting.

kuma/wiki/templates/wiki/email/edited.ltxt
jwhitlock: A space is needed between the title and the phrase "all its sub-articles."  It may also make sense to display the URL instead of or in addition to the the title when the email is due to a tree subscription.

  From notification email:

```
--
You are subscribed to edits on: Web technology for developersand all its sub-articles.
Unsubscribe from these emails:
https://developer.mozilla.org/unsubscribe/9470?s=mkrtZZGAnr
```

jwhitlock:
My practical testing worked:

* Able to subscribe / unsubscribe to pages
* Able to subscribe / unsubscribe to trees
* Get page subscription emails
* Get tree subscription emails (needs a space)
* When subscribed to page and tree, page wins
* ``watch_menu`` flag moves subscription to a new button.

Two things:
* [ ] Add a space to the tree subscription email
* [ ] Make watch menu button active if a user is subscribed to page or tree (optional)

Reassigning to robhudson, to complete his review

kuma/wiki/tests/test_views.py
groovecoder: Yeah; bad comment.

jwhitlock: If it makes the code clearer, you can [clear the test outbox](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.core.mail.django.core.mail.outbox) by assigning an empty list:

```
mail.outbox = []
```

kuma/wiki/templates/wiki/email/edited.ltxt
```diff
@@ -32,4 +32,14 @@ Edit Article:
...
  {{ history_url|absolutify }}
-{% if watch %}{{ unsubscribe_text(watch) }}{% endif %}{% endautoescape %}
+--
+{% if watch %}
+  {% set title = watch.content_object.title %}
+    {% if watch.event_type == 'wiki edit document in tree' %}
+      {{ _('You are subscribed to edits on: {title} and all its sub-articles.')|f(title=title) }}
+    {% else %}
+      {{ _('You are subscribed to edits on: {title}.')|f(title=title) }}
+    {% endif %}
```
robhudson: Since this is an email, do you want to use jinja2 whitespace eating tags? Or do you want those strings indented?

kuma/wiki/templates/wiki/includes/buttons.html
```diff
@@ -30,34 +34,33 @@
...
+        </div>
+        </li>
+        {% endif %}
 
+        <li><button id="advanced-menu" class="only-icon" aria-haspopup="true" aria-owns="advanced-menu-submenu" aria-expanded="false"><span>{{ _('Advanced') }}</span><i aria-hidden="true" class="icon-cog"></i></button>
         <div class="submenu js-submenu" id="advanced-menu-submenu">
           <!-- this page -->
           <div class="submenu-column">
             <div class="title">{{ _('Advanced') }}</div>
+                {% if user.is_authenticated() and document and not waffle.flag('watch_menu')%}
```
robhudson: space before `%}`

groovecoder:
So, the first implementation I have for showing the parent tree subscription(s) to the user is simply putting an "Unsubscribe" link for each parent tree the user is watching in the menu. It's not pretty, and it could get fugly if users watch multiple trees above a doc. (Though I'm not sure why they would?)
mdn-watch-menu-tree-unsubscribes
Would probably like some better UX/UI direction before merging this. :neutral_face: @stephaniehobson - more suggestions?

kuma/wiki/tests/test_views.py
```diff
@@ -2549,6 +2550,140 @@ def _check_message_for_headers(message, username):
...
+        data.update({'form': 'rev',
+                     'slug': rev.document.slug,
+                     'content': 'This edit should send 2 emails',
+                     'comment': 'This edit should send 2 emails'})
+        self.client.post(reverse('wiki.edit',
+                                 args=[rev.document.slug]),
+                         data)
+        self.assertEquals(2, len(mail.outbox))
+        message = mail.outbox[0]
+        assert testuser2.email in message.to
```
jezdez: any reason to not use `self.assertIn` here? I'm all for using native assert statements in tests, but that'd be a third option of conventions in our tests. I'd rather move us to pure Django tests that offer stuff like `self.assertIn` :-/

groovecoder:
Assigning to @stephaniehobson to review (and hopefully not puke from) my proposed just-stick-it-in-the-menu UX approach.

kuma/wiki/tests/test_views.py
groovecoder: In today's backend meeting we agreed to start removing nose assertion helpers to cut down on the different number of conventions in our tests. Going to leave this as-is unless/until we start removing native asserts.

stephaniehobson:
Needs a rebase :/

stephaniehobson:
Small problem with the HTML escaping:

screen shot 2015-11-09 at 15 36 27

Suggestions (not blocking):
- put the article title in italics in the menu
- re-order the menus so that events relating to the page the user is on are grouped together (ie, #1 is always (un)subscribe to this article and #2 is always (un)subscribe to this article and sub-articles)

R+ after rebase, changes are optional.

stephaniehobson:
Oh, except fixing the problem with the HTML escaping. That's not optional.

groovecoder:
Merge was easier than rebase. Looking into the rest ...

groovecoder:
* [x] Escaped the doc title in tree menu items.
* [x] Moved tree items below page items.
Italicizing the page title and escaping it at the same time will be tricky, so I'm leaving that for a follow-up enhancement.

groovecoder:
Assigning to @stephaniehobson to double-check my updates. Then this is good to merge.

stephaniehobson:
Quotes around the article name instead of italics would also be an improvement, but not one to block on ;)

R+

groovecoder:
After the first week of ~50/50 split testing ...

The separate watch menu doesn't seem to have much effect on users clicking the "Edit" button. Conversion rate with the separate watch menu is 0.04% vs. 0.03% with the watch items included in the "Advanced" menu. (Note: I haven't done a full statistical analysis on the level of Optimizely testing.)

watch-menu-week-edit-goal

In addition, the extra menu doesn't seem to have much effect on users going to the "Translate" screen either. Conversion rate with the separate watch menu is 0.03% vs. 0.02% with the watch items included in the "Advanced" menu. (Again, I haven't done a full statistical analysis on the level of Optimizely testing.)

watch-menu-week-translate-goal

There even less data for the new "Watch" goal, but there are 2x as many conversions with the separate watch menu so far.

watch-menu-week-watch-goal

With all that said, I'm leaning towards keeping the separate watch menu and removing the items from the advanced menu. But, let's give it another week to collect some more data.

PR #3604, opened 2015-10-27
Fix broken .po files

robhudson:
This fixes some broken .po files found using dennis linting. There is one that is still broke that revealed a bug in dennis that is already fixed on master.

### DIFF
--- a/locale/bn_BD/LC_MESSAGES/django.po
+++ b/locale/bn_BD/LC_MESSAGES/django.po
@@ -1217,11 +1217,11 @@ msgstr "http://creativecommons.org/publicdomain/zero/1.0/"
 
 #: kuma/demos/admin.py:38
 msgid "Censored %(item)s"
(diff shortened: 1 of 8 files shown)


jezdez:
Hm, I don't think that is enough, since the next time we see pushes from Verbatim or Pontoon this will be overwritten again. We need to fix the strings upstream so to say. Maybe @mathjazz may want to weigh in though.

mathjazz:
Actually it's ok, Pontoon will import translation from VCS. If you're interested in the details, this is how it works:
http://mozilla-pontoon.readthedocs.org/en/latest/dev/sync.html

Same goes for Verbatim, which I'll need to kick manually to get updated once this is merged.

Also - sl should be removed completely, because we're not localizing MDN. But you're probably already planning to remove inactive locales in a separate request.

jezdez:
Aha, thanks for the info! In that case r+++ merge away

PR #3605, opened 2015-10-27
Bug 1139490: Manage callout datauri with Pipeline

openjck:
### Testing

1. Enter production mode
2. Verify that the homepage still uses a datauri for the callout box sprite


groovecoder:
Looks good.

stephaniehobson:
Woah, we can do that with pipeline? Awesome.

Does it just handle the callout box sprite or does it do all the images in the css file?

openjck:
Just the callout box sprite for now. It embeds images in an *embed* folder that are under a certain configurable file size (currently 32KB). At the moment I'm looking at where it makes sense to use embedded images given that we also have the option of using far-future expires headers.

PR #3606, opened 2015-10-27
Bug 1164502: Line-highlighting shouldn't inherit padding.

stephaniehobson:


openjck:
What would be a good thing to test? I saw mention of [test 8](https://developer.allizom.org/en-US/docs/User:stephaniehobson:code#Test8_Javascript_content_line_highlighting) in IRC but it looks the same before and after the change to me.

MikkCZ:
Seems you have mistyped the bug number. [1164502](https://bugzilla.mozilla.org/show_bug.cgi?id=1164502) is something completely different. ;)

PR #3607, opened 2015-10-27
Use isort for improved import order consistency

jezdez:
https://bugzilla.mozilla.org/show_bug.cgi?id=1218957

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,11 @@ services:
     - elasticsearch
 env:
     matrix:
-        - TOXENV=py27
-        - TOXENV=flake8
-        - TOXENV=docs
-        - TOXENV=dennis
+        - TOXENV=py27-tests
+        - TOXENV=py27-flake8
+        - TOXENV=py27-docs
+        - TOXENV=py27-isort
+        - TOXENV=py27-dennis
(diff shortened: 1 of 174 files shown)


robhudson:
In general I :heart: this. I've used isort while editing individual files for awhile and am in agreement with this.

The errors appear to be missed application of isort. Perhaps something with isort running from a virtualenv? The suggested changes from the output look correct and are what I see locally as well.

Once those changes are applied, r+.

Personally I'm going to add the git hook after this lands to catch errors before waiting on Travis to find them.

jezdez:
Hmm, not sure what's wrong.

jezdez:
Can someone else try fixing this, I've been stabbing at this for too long

robhudson:
@jezdez I got the isort fixed by manually fixing the few files that were missed by running `isort --apply`, but no idea what's up with the tests.

jezdez:
@robhudson Yeah, I don't know man. I can try to pick it up again

robhudson:
I tried and failed also. I think we should just close it and perhaps work towards this as we make edits over time. Whatever is causing this problem will likely rear its head somewhere along the way and be easier to find.

PR #3608, opened 2015-10-28
Fix Bug 1218466: Update javascript, ajax, and youtube trackError calls.

stephaniehobson:
Separating the data to allow more sort options.

### DIFF
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -88,25 +88,49 @@
         },
 
         /*
-            Track specific clientside errors create by our code
+            Track specific clientside errors created by our code
+            this article was a lot of help: http://blog.gospodarets.com/track_javascript_angularjs_and_jquery_errors_with_google_analytics/
         */
         trackClientErrors: function() {
+
+            // javascript & jQuery errors
             $(win).on('error', function(e) {
-                var originalEvent = e.originalEvent;
-                analytics.trackError(' JavaScript Error: ' + originalEvent.message + ' ; ' + originalEvent.filename + ':' + originalEvent.lineno);
+                // probably javascript
+                if(e.originalEvent) {
+                    var originalEvent = e.originalEvent;
+                    var lineAndColumnInfo = originalEvent.colno ? ' line:' + originalEvent.lineno +', column:'+ originalEvent.colno : ' line:' + originalEvent.lineno;
+                    analytics.trackError('JavaScript Error', originalEvent.message , originalEvent.filename + ':' + lineAndColumnInfo);
+                }
+                // no originalEvent means probably jQuery
+                else {
+                    var message = e.message ? e.message : '';
+                    analytics.trackError('jQuery Error', message);
+                }
             });
+
+            // jQuery ajax errors
             $(doc).ajaxError(function(e, request, settings) {
-                analytics.trackError('AJAX Error: ' +  settings.url + ' : ' + e.result);
+                analytics.trackError('AJAX Error', settings.url , JSON.stringify({
+                    result: e.result,
+                    status: request.status,
+                    statusText: request.statusText,
+                    crossDomain: settings.crossDomain,
+                    dataType: settings.dataType
+                    })
+                );
             });
         },
 
         /*
             Sends universal analytics client side error
         */
-        trackError: function(category, action) {
+        trackError: function(category, action, label) {
+            // label is optional, give it a default value if it's not passed
+            label = typeof label !== 'undefined' ? label : '';
             return analytics.trackEvent({
-                category: category,
-                action: action
+                'category': category,
+                'action': action,
+                'label': label
             });
         }
     };
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -823,7 +823,7 @@
                     });
                 });
                 players[i].addEventListener('onError', function(event) {
-                    mdn.analytics.trackError('YouTube Error: ' + event.data, win.location.href);
+                    mdn.analytics.trackError('YouTube Error', event.data);
                 });
             });
         };


stephaniehobson:
Testing:
- make sure you have a google account set up in the admin config settings (you can use the prod one, there are filters to remove dev traffic)
- be in dev mode
- break some js somewhere that is loaded asynchronously (it needs to load after analytics.js) - I remove the last line in `static/js/syntax-prism.js` 
- visit a page that calls the broken js
- in the console (you might have to use Chrome if it doesn't show up in FF) you should see some diagnostic Google Analytics logging, one blob to initialize Google Analytics and a second transmitting the event for the error (you can inspect category, label, and action there).

groovecoder:
I can take this.

kuma/static/js/analytics.js
```diff
@@ -93,20 +93,21 @@
         trackClientErrors: function() {
             $(win).on('error', function(e) {
                 var originalEvent = e.originalEvent;
-                analytics.trackError(' JavaScript Error: ' + originalEvent.message + ' ; ' + originalEvent.filename + ':' + originalEvent.lineno);
+                analytics.trackError('JavaScript Error', win.location.href, originalEvent.message + ' ; ' + originalEvent.filename + ':' + originalEvent.lineno);
```
groovecoder: Sorry I didn't notice this in the email proposal until I saw it running in my console ...

Storing the `win.location.href` is duplicitous - GA already stores the `Page URL` dimension for every event. So, we could put something like `originalEvent.mesage`, `originalEvent.filename`, or `originalEvent.lineno` as the action if we think we may want to report on the data by one of those dimensions in addition to category and page.

kuma/static/js/analytics.js
```diff
@@ -93,20 +93,21 @@
...
             });
             $(doc).ajaxError(function(e, request, settings) {
-                analytics.trackError('AJAX Error: ' +  settings.url + ' : ' + e.result);
+                analytics.trackError('AJAX Error', win.location.href, settings.url + ' : ' + e.result);
```
groovecoder: Same comment here - we could put the `settings.url` value into the `action` argument if we want to report by that dimension.

kuma/static/js/wiki.js
```diff
@@ -825,7 +825,7 @@
                     });
                 });
                 players[i].addEventListener('onError', function(event) {
-                    mdn.analytics.trackError('YouTube Error: ' + event.data, win.location.href);
+                    mdn.analytics.trackError('YouTube Error', win.location.href, event.data);
```
groovecoder: Same. If there's another interesting or helpful property of the `event` that we want to store as the event `action` we can do that.

groovecoder:
All comments are the same and totally up to you if you want to change the `action` of the events to something else or not.

r+

(Feel free to merge yourself. Just add `r=groovecoder` to the merge commit message, which you can do in the GitHub "Confirm merge" UI)

stephaniehobson:
Made some substantial changes, should get a good review again.

Here are some different ways to trigger errors:
https://gist.github.com/stephaniehobson/0e8b6b06314f7dd2fba5

I wasn't able to trigger a youTube error.

kuma/static/js/analytics.js
```diff
@@ -88,25 +88,49 @@
...
+            Track specific clientside errors created by our code
+            this article was a lot of help: http://blog.gospodarets.com/track_javascript_angularjs_and_jquery_errors_with_google_analytics/
         */
         trackClientErrors: function() {
+
+            // javascript & jQuery errors
             $(win).on('error', function(e) {
-                var originalEvent = e.originalEvent;
-                analytics.trackError(' JavaScript Error: ' + originalEvent.message + ' ; ' + originalEvent.filename + ':' + originalEvent.lineno);
+                // probably javascript
```
groovecoder: :laughing:

kuma/static/js/analytics.js
```diff
@@ -88,25 +88,49 @@
...
+                if(e.originalEvent) {
+                    var originalEvent = e.originalEvent;
+                    var lineAndColumnInfo = originalEvent.colno ? ' line:' + originalEvent.lineno +', column:'+ originalEvent.colno : ' line:' + originalEvent.lineno;
+                    analytics.trackError('JavaScript Error', originalEvent.message , originalEvent.filename + ':' + lineAndColumnInfo);
+                }
+                // no originalEvent means probably jQuery
```
groovecoder: Ah, now it makes sense. Not as funny.

kuma/static/js/analytics.js
```diff
@@ -88,25 +88,49 @@
...
+            // jQuery ajax errors
             $(doc).ajaxError(function(e, request, settings) {
-                analytics.trackError('AJAX Error: ' +  settings.url + ' : ' + e.result);
+                analytics.trackError('AJAX Error', settings.url , JSON.stringify({
+                    result: e.result,
+                    status: request.status,
+                    statusText: request.statusText,
+                    crossDomain: settings.crossDomain,
+                    dataType: settings.dataType
+                    })
```
groovecoder: Will be interesting to see how these show up in GA and how we'll report across these dimensions of the error.

groovecoder:
Code looks good. Spot-checking ...

groovecoder:
Spot-check looks good, except [one question](https://gist.github.com/stephaniehobson/0e8b6b06314f7dd2fba5#gistcomment-1616631) (uplifting here from gist):

For the test jQuery (non-Ajax) error, eventAction and eventLabel were both empty? Is that expected since the error() call is empty? When I tried to fill it with a message, it didn't seem to trigger the analytics track error.

stephaniehobson:
Empty action and label were what I saw for the jQuery errors too. I was copying someone else's code and not 100% sure the method I picked for triggering those errors was correct. I was sort of hoping that after it was in production for a week we'd know if there was a better way of doing it.

There are other things we could try putting in there - the triggering element for example. If we're worried about it I'll just remove the jQuery error detection so we can get the other updates into production.

kuma/static/js/analytics.js
stephaniehobson: Yeah, no idea what kind of info we're going to get here.

groovecoder:
Not worried; looks good the way it is.

PR #3610, opened 2015-10-28
Bug 1139490: Manage all data URIs with Pipeline

openjck:
### Testing

1. Enter production mode
2. Verify that Tabzilla still looks good
3. Visit [this page](https://developer-local.allizom.org/en-US/users/signin?next=/en-US/docs/new) while logged out and verify that the Persona icon still looks good


openjck:
Updated to add a `$path-to-embedded-images` variable

stephaniehobson:
R+

PR #3611, opened 2015-10-29
Dockerize Kuma Bug 1218563

mostlygeek:
Docker'ize Kuma for production deployment

- major work in progress
- a `docker-build.sh` script creates containers, one to build static assets and the other for production deployment

TODO: 

- ~~get travis.ci build working~~
- get tests working in travis.ci w/ docker container

### DIFF
--- a/Dockerfile-builder
+++ b/Dockerfile-builder
@@ -1,0 +1,29 @@
+# This is a work in progress.
+#
+# The preferred way to set up development is still using Vagrant as documented here:
+# http://kuma.readthedocs.org/en/latest/installation.html
+
+FROM python:2.7.10
+
+WORKDIR /app
+RUN apt-get update && \
+  apt-get install -y \
+    nodejs \
+    nodejs-legacy \
+    npm \
+    libtidy-0.99-0
+
+RUN npm install -g \
+    fibers@1.0.1 \
+    clean-css@2.2.16 \
+    csslint@0.10.0 \
+    jshint@2.7.0 \
+    stylus@0.49.2 \
+    uglify-js@2.4.13
+
+COPY ./requirements/compiled.txt /app/requirements/compiled.txt
+
+RUN pip install --build ./build --no-cache-dir --no-deps \
+    -r requirements/compiled.txt
+
+COPY . /app
--- a/Dockerfile-builder
+++ b/Dockerfile-builder
@@ -25,5 +25,3 @@ COPY ./requirements/compiled.txt /app/requirements/compiled.txt
 
 RUN pip install --build ./build --no-cache-dir --no-deps \
     -r requirements/compiled.txt
-
-COPY . /app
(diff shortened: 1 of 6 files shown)


docker-build.sh
```diff
@@ -0,0 +1,34 @@
...
+   --exclude=./media \
+   --exclude=./kumascript \
+   --exclude=./kuma/static \
+   --exclude=./vendor/*/docs \
+   --exclude=./provisioning \
+   --exclude=./etc \
+   --exclude=./docker \
+   --exclude=./configs \
+   --to-stdout . | \
+   sudo docker build -t kuma:latest -f Dockerfile-production -
```
robhudson: I was wondering how this would work. Pretty neat.

Dockerfile-production
```diff
@@ -0,0 +1,19 @@
+FROM python:2.7.10
+
+# This is a work in progress.
+
+RUN useradd --create-home --home-dir /app --shell /bin/bash app
+WORKDIR /app
+
+RUN apt-get update && \
+  apt-get install -y \
+    libtidy-0.99-0
```
robhudson: would we want:
` && apt-get clean`?

Dockerfile-builder
```diff
@@ -28,3 +20,8 @@ RUN npm install -g \
     jshint@2.7.0 \
     stylus@0.49.2 \
     uglify-js@2.4.13
+
+COPY ./requirements/compiled.txt /app/requirements/compiled.txt
+
+RUN pip install --build ./build --no-cache-dir --no-deps \
+    -r requirements/compiled.txt
```
robhudson: This seems like it would still be suitable for using as a local-dev Dockerfile. We could just point the docker-compose `dockerfile` to this. Is that correct? Or would we have another Dockerfile for local dev?

Dockerfile-production
mostlygeek: yes :)

Dockerfile-builder
mostlygeek: Another dockerfile for local dev. Almost certainly the requirements for dev will be more than what we need for prod deployment.

docker-build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# make sure the build container is updated
+# the build container only includes the necessary 
+# system dependencies to build out the static assets
+sudo docker build -t kuma:builder -f Dockerfile-builder .
+
+# Build the static assets by volume mounting the 
+# current source code
+docker run -it --rm=true -v "$PWD:/app" kuma:builder \
+    sh -c './manage.py collectstatic --noinput && ./manage.py compilejsi18n'
```
robhudson: These need to be reversed now since `compilejsi18n` now creates files that `collectstatic` processes.

robhudson:
@mostlygeek I'm going to close this. A lot has changed and it sounds like you've got a good default Dockerfile we can mostly copy. Re-open if you'd like to continue with this.

PR #3613, opened 2015-10-30
Fixes bug 1217208 - Remove tower dependency

robhudson:
The puente project is a new project that should act as a bridge ("puente" is Spanish for "bridge") from tower to a completely vanilla Django/Jinja2 project. The goal of puente is to not exist but for now it provides some of the same utilities that tower provided in a nicer package.

We can spot-check the .pot files changes if you check out this branch and run `./manage.py extract`. The changes should be minimal but there may be some text wrapping differences (since babel wraps slightly differently than translate-toolkit) and some new strings (since this fixes some wrongly ignored paths and broken templates).

Todo:
- [ ] Re-run extract/merge before merging. This causes a lot of rebasing as new l10n commits come in so I'm not including it here for now.

### DIFF
--- a/kuma/core/email_utils.py
+++ b/kuma/core/email_utils.py
@@ -1,43 +1,14 @@
-import contextlib
 import logging
 from functools import wraps
 
+import jingo
 from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
 from django.test import RequestFactory
 from django.utils import translation
 
-import jingo
-import tower
-
 
 log = logging.getLogger('kuma.core.email')
-
-
-@contextlib.contextmanager
-def uselocale(locale):
-    """Context manager for setting locale and returning
-    to previous locale.
-
-    This is useful for when doing translations for things run by
-    celery workers or out of the HTTP request handling path.
-
-    >>> with uselocale('xx'):
-    ...     subj = _('Subject of my email')
-    ...     msg = render_email(email_template, email_kwargs)
-    ...     mail.send_mail(subj, msg, ...)
-    ...
-
-    In kuma, you can get the right locale from User.locale and
-    also request.LANGUAGE_CODE.
-
-    If kuma is handling an HTTP request already, you don't have to
-    run uselocale---the locale will already be set correctly.
-    """
-    currlocale = translation.get_language()
-    tower.activate(locale)
-    yield
-    tower.activate(currlocale)
 
 
 def safe_translation(f):
@@ -50,7 +21,7 @@ def safe_translation(f):
     @wraps(f)
     def wrapper(locale, *args, **kwargs):
         try:
-            with uselocale(locale):
+            with translation.override(locale):
                 return f(locale, *args, **kwargs)
         except (TypeError, KeyError, ValueError, IndexError) as e:
             # Types of errors, and examples.
@@ -65,7 +36,7 @@ def safe_translation(f):
             #    '{0} {1}'.format(42)
             log.error('Bad translation in locale "%s": %s', locale, e)
 
-            with uselocale(settings.WIKI_DEFAULT_LANGUAGE):
+            with translation.override(settings.WIKI_DEFAULT_LANGUAGE):
                 return f(settings.WIKI_DEFAULT_LANGUAGE, *args, **kwargs)
 
     return wrapper
--- a/settings.py
+++ b/settings.py
@@ -463,7 +463,7 @@ INSTALLED_APPS = (
     # util
     'pipeline',
     'product_details',
-    'tower',
+    'puente',
     'smuggler',
     'constance.backends.database',
     'constance',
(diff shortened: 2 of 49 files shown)


settings.py
```diff
@@ -504,9 +504,8 @@ def JINJA_CONFIG():
     from django.core.cache.backends.memcached import MemcachedCache
     from django.core.cache import caches
     cache = caches['memcache']
-    config = {'extensions': ['jinja2.ext.i18n', 'tower.template.i18n',
-                             'jinja2.ext.with_', 'jinja2.ext.loopcontrols',
-                             'jinja2.ext.autoescape',
+    config = {'extensions': ['jinja2.ext.i18n', 'jinja2.ext.with_',
```
willkg: You want to use `puente.ext.PuenteI18nExtension' and not `jinja2.ext.i18n`. Otherwise when rendering, trans blocks won't collapse whitespace and then none of your msgids will match and then THE MOON WILL TURN BLOOD AND VAMPIRE MONKEYS WILL RAIN FROM THE HEAVENS AND DESTROY THE CITIES OF THE EARTH WITH THEIR EVIL CHITTERING LAUGHTER AND MANIACAL BANANAS AND WE WILL ALL HAVE TO RUN UNDERGROUND AND USE EMACS AND OMG THE DESTRUCTION THE HUMANITY!!!! WHAT HAVE WE WROUGHT!?!?!

robhudson: Updated, thanks!

robhudson: Somehow that broke everything b/c `gettext` isn't in the context? Also, the example jingo test project has both - do you need both? I tried it and it failed. I'll poke with this on Monday.

willkg: You shouldn't need both. The puente one subclasses the other one. So this
situation is very fishy....

kuma/core/email_utils.py
```diff
@@ -35,9 +33,9 @@ def uselocale(locale):
     run uselocale---the locale will already be set correctly.
```
jezdez: Now that we use Django's i18n tools we can just remove this function and use `django.utils.translation.override`: https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#using-translations-outside-views-and-templates

jezdez:
One thing I noticed is that we have a few translation strings with "modern" string formatting `_('my {string}).format(string='my')', which until gettext 0.18.3 wasn't supported as a format: http://git.savannah.gnu.org/cgit/gettext.git/commit/?id=6543f644bf11c00a19843971e1858edb3d870442

Since we're running old RedHat machines we should check if we're not accidentally dropping strings or missing wrong variable substitution. One thing that could happen for example is msgfmt not showing errors since it just assumes the curly braces are not technical aspects of the strings.

Also, for consistency's sake we should replace all modern string formatting with old style (since it's totally fine).

jezdez:
This is an epic PR, a few issues here and there but in general a huge improvement.

giphy

robhudson:
> Also, for consistency's sake we should replace all modern string formatting with old style (since it's totally fine).

Thanks for the note about gettext supporting these. I kind of prefer the new format if gettext supports it and if so, wouldn't mind everyone moving towards it. Shouldn't that help any future python 3 conversion?

willkg:
The wry twist here is that Jinja2 template extraction extracts variables and puts them in the `%(foo)s` format. Thus I'm pretty sure we can't move forward to consistently use `{foo}` format everywhere. Maybe that can be changed, but I don't know how to do it.

I know with SUMO and Input we've had a lot of problems with translators and variable formats and I wonder if that's partially because we have two styles in our strings. I'm writing up a bug for Input and SUMO right now to switch all the formatting to use `%` style.

jezdez:
No no, that's a misconception, Python 3 just happily takes "old" style string substitutions. New style string formatting is entirely unrelated from that and was backported to Python 2.6: https://docs.python.org/2/whatsnew/2.6.html#pep-3101-advanced-string-formatting

jezdez:
Two funny side things:
- https://pyformat.info is useful for some examples
- https://www.python.org/dev/peps/pep-0498/ A new – *third* (!!) – way of doing string substitution coming in Python 3.x, recently accepted

settings.py
willkg: gettext doesn't get added to the context by the `jinja2.ext.i18n` extension or the `puente.ext.PuenteI18nExtension'. It gets added by calling `.install_gettext_translations()` on the environment.

Tower did that here:

https://github.com/clouserw/tower/blob/master/tower/__init__.py#L90

It had to use its own gettext/ngettext.

Jingo does it here:

https://github.com/jbalogh/jingo/blob/v0.7.1/jingo/__init__.py#L69

and specifically looks for `jinja2.ext.i18n` because awesome.

So it looks like if we take out Tower and you're using Jingo, then you do need to have `jinja2.ext.i18n`. I'd make it the first extension with `puente.ext.PuenteI18nExtension` right after (like you had). Hopefully that stomps on the right things.

See if that works. I think you said it did before. If it does, then I'll update the Puente docs. Sorry about that.

robhudson:
All extraneous commits that could be merged to master are now merged to master. This has been simplified now.

I'd like to save the `ugettext` vs `ugettext_lazy` for post merge. Likewise with the new style string formatting. I'll file a bug for both. If we can get this merged we can start testing what breaks next on the Django 1.8 branch. (A little preview -- it looks like it'll be django-celery).

jezdez:
@robhudson The `ugettext` vs `ugettext_lazy` issues in `kuma/wiki/models.py` in class attributes (`Meta`) are a regression in this PR and not an existing issue. If you could please fix those then I'm r+ on this one. Sounds like a good idea to leave the new style string fixes to another PR.

jezdez:
@robhudson Nevermind what I said, it's not a blocker since it was in master before. A fix after this PR would be appreciated -- or a fix here. But it's not a regression.

robhudson:
Updated to include some minor ugettext/ugettext_lazy cleanup.

jezdez:
Needs a rebase :(

jezdez:
r+ otherwise

willkg:
We did it! Hooray! Favorite drinks for everyone!

PR #3614, opened 2015-11-02
Bug 1139490: Improve local Pipeline performance

openjck:
Kuma pages load very slowly locally when Pipeline kicks off compilation.
This change restores the old workflow, where the `stylus` executable
compiles individual stylesheets as needed.

More info:
https://github.com/mozilla/kuma/pull/3597#issuecomment-151984236

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ kuma/static/js/libs/ckeditor/source/ckbuilder
 htmlcov/
 locale/**/*.mo
 /bower_components
+build/assets/css/*
+!build/assets/css/README.md
--- a/Procfile
+++ b/Procfile
@@ -2,3 +2,4 @@ web: make clean && python2.7 manage.py runserver 0.0.0.0:8000
 worker: python2.7 manage.py celery worker --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
 camera: python2.7 manage.py celerycam --freq=2.0
 kumascript: node kumascript/run.js
+stylus: scripts/compile-stylesheets --watch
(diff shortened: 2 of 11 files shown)


openjck:
It looks like the Travis error is about something unrelated involving locales. cc @robhudson

jezdez:
@openjck See my mail to the dev-mdn list.

.gitignore
```diff
@@ -37,3 +37,5 @@ kuma/static/js/libs/ckeditor/source/ckbuilder
 htmlcov/
 locale/**/*.mo
 /bower_components
+kuma/static/compiled-css/*
```
jezdez: The compiled CSS files should not be compiled into the kuma folder since that's where the assets are. Please create them inside a separate folder that is outside the kuma/ directory (imagine the kuma directory as a non-writable file system). Very soon we'll need to push those files to S3 automatically, so it'd be best to have that build folder in the root directory that we configure with the `STATICFILES_DIRS` setting to be used as a source of files to copy.

robhudson:
I updated the travis file to not fail on broken l10n. It was also fixed. So a rebase should help the tests here.

.gitignore
openjck: Okay, sure thing. Is there a different top-level directory you would recommend, or would we prefer to create something like a */compiled-assets* directory?

stephaniehobson:
Faster! :dancer:

R+ from me. Has @jezdez's concern been addressed?

.gitignore
jezdez: Sure, I'd also be okay with `build/assets` but I leave it up to you.

jezdez:
@stephaniehobson Yay!

openjck:
Updated

groovecoder:
Much faster. Could you could push this branch up to the mozilla repo and deploy it straight to the stage server? That way you can test the changes to the chief deploy script before merging to master.

openjck:
Sure thing. Is this r+ otherwise?

groovecoder:
Yeah, looks good and works well.

groovecoder:
Stage deploy and test of this branch looks good. Merging ...

PR #3618, opened 2015-11-03
Bug 1221061 - Use lower case for adding and querying tag names.

jezdez:
This makes use of Django's new first class support of database functions to query the tags. In our case by using `LOWER` to annotate the tag querysets with a lowercase name and then querying that.

https://bugzilla.mozilla.org/show_bug.cgi?id=1221061

### DIFF
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -8,6 +8,7 @@ from tower import ugettext_lazy as _lazy
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db.models.functions import Lower
 from django.forms.widgets import CheckboxSelectMultiple
 from django.template.loader import render_to_string
 from django.utils import translation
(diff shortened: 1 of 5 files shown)


jezdez:
Those tests will fail hard right now

robhudson:
Now that django 1.8 is merged we can continue this effort, right?

robhudson:
Can we also remove all the custom collation stuff from the vagrant install also? I can update the docker to use regular mysql instead of the custom one. I'm excited for this change to re-test pgloader.

jezdez:
This code alone won't be enough since the tag querying happens in django-taggit. Until we're on Postgres we'll need the collation.

jezdez:
https://github.com/alex/django-taggit/blob/8b23b6f19aac1558a01e54e5b69a9c38f9d9cdfe/taggit/managers.py#L169-L188 is the code responsible for that

jwhitlock:
Closing this PR as out of date.

https://github.com/mozilla/kuma/pull/4103 gets us on ``TAGGIT_CASE_INSENSITIVE``, and leans on the database to enforce case-insensitive tags.  This still relies on the custom collation to determine what case-insensitive means...

[bug 1221061](https://bugzilla.mozilla.org/show_bug.cgi?id=1221061) was closed as invalid, since we plan to move to PostgreSQL rather than keep MySQL and switch to ``utf8_bin``.  We're currently planning on a database switch after the AWS migration.

In the long run, I think we should not rely on the database for canonicalizing tags, and pull this functionality into Kuma. That may mean moving away from ``django-taggit``. However, we may need to do that anyway, to start supporting translated tags and other features.

PR #3619, opened 2015-11-03
bug 1219813, 1219818 - use djcelery_transactions for emails

groovecoder:


kuma/users/tasks.py
```diff
@@ -8,6 +7,8 @@
 from django.core.mail import EmailMultiAlternatives
 from django.template.loader import render_to_string
 
+from djcelery_transactions import task
```
robhudson: How about `from djcelery_transactions import task as transaction_task` instead?

groovecoder:
Updated to make both usages `from djcelery_transactions import task as transaction_task`.

jwhitlock:
Need to add ``CELERY_EAGER_TRANSACTION = True`` to settings (see https://github.com/fellowshipofone/django-celery-transactions#compatibility-with-celery_always_eager), which then makes tests fail :cry:

jwhitlock:
+r.  On local dev box, both emails are sent async, without the errors in the bug report.  ``CELERY_EAGER_TRANSACTION = True`` was **not** used.

PR #3621, opened 2015-11-04
Bug 1217208 - Fix completely broken email template

robhudson:


groovecoder:
I can haz tests? There's a small `tests.py` file in the app that could grow a test class like the others we have to test emails?

robhudson:
Good idea. I plucked this out of a bigger PR. I'll add some tests that would have caught this prior to this change.

groovecoder:
Spot-check works at least though.

r+

robhudson:
Updated. It was actually even more broken. Now all should be good.

PR #3622, opened 2015-11-04
fix bug 1201958 - Revert commit eef61169effdf51ec799e116735d3058110d104c

groovecoder:


jezdez:
There is no need to extract the summary on every request, please move this into a new denormalized field as we have on the `Document` model. That probably means adjusting the naming of the fields involved since the other "summary" is really something about SEO and this one about "text" summary (?). Or maybe call the other one "SEO text" and this "summary"?

groovecoder:
I can work on that and update the PR. I suggest we don't block on it though, since this is a regression for minimal performance gain.

jezdez:
@groovecoder Sure, if you can make the blank exception more specific I'd be okay letting this be optimized later.

stephaniehobson:
Summary and zones are behaving as expected again!! 

R+ from me. Assigning to @jezdez to check that the blank exception is being handled properly.

kuma/wiki/helpers.py
```diff
@@ -139,6 +141,25 @@ def wiki_bleach(val):
...
+@register.filter
+def selector_content_find(document, selector):
+    """
+    Provided a selector, returns the relevant content from the document
+    """
+    content = ''
+    try:
+        page = pq(document.rendered_html)
+    except ValueError:
+        # pass errors during constuction
```
jezdez: minor typo

jezdez:
rwc+

PR #3623, opened 2015-11-04
Bug 1217208 - Remove python template string from feed

robhudson:


kuma/wiki/tests/test_feeds.py
```diff
@@ -49,10 +51,10 @@ def test_updated_translation_parent_feed_mod_link(self):
...
         for i, item in enumerate(feed.find('item')):
             desc_text = pq(item).find('description').text()
-            ok_("%s$compare?to=%s&from=%s" % (d1.slug,
-                                              d1.current_revision.id,
-                                              first_rev_id)
-                in desc_text)
+            compare_url = '%s$compare?to=%s&from=%s' % (d1.slug,
+                                                        d1.current_revision.id,
+                                                        first_rev_id)
+            ok_(escape(compare_url) in desc_text)
```
groovecoder: drop the `ok_` nose helper(s) while we're in here?

kuma/wiki/templates/wiki/feed_docs_updated.html
```diff
@@ -0,0 +1,12 @@
+<p>
+  <a href="{{ obj.parent.get_absolute_url()|absolutify }}" title="{{ obj.parent.title }}">
+    {% trans locale=obj.parent.locale %}View {locale} parent{% endtrans %}
+  </a>
+  (<a href="{{ mod_url|absolutify }}">{% trans modified=obj.parent.modified %}last modified at {modified}{% endtrans %}</a>)
+</p>
+<p>
+  <a href="{{ obj.get_edit_url()|absolutify }}" title="{{ obj.title }}">
```
groovecoder: non-blocking scope-creep: Can we use the `add_utm` helper on the urls in this template? ([Like we do in the `plain.ltxt` template for welcome emails.](https://github.com/mozilla/kuma/blob/22390c1deec5ea40043871d8f20683cb642b5b2b/kuma/users/templates/users/email/welcome/plain.ltxt#L20-L39))

So, something like:
```
{{ add_utm(obj.get_edit_url()|absolutify, 'feed', medium=format) }}
```
Though I'm not sure you can call jinja output filters inside function arguments. :question: 

That way, GA can tell us how effective the feed links are at driving sessions that go on to translate docs, and we can compare that with emails to see which feature is more effective.

groovecoder:
r+ with nose removal(s)

kuma/wiki/tests/test_feeds.py
robhudson: Oh yeah, forgot we decided to do that. I'll update to use `assertTrue` instead of `ok_` and `assertEqual` instead of `eq_`.

kuma/wiki/templates/wiki/feed_docs_updated.html
robhudson: Looks like the filter works just fine inside the `add_utm`. Impressive.

robhudson: I also added a test for this URL.

kuma/wiki/templates/wiki/feed_docs_updated.html
```diff
@@ -0,0 +1,12 @@
+<p>
+  <a href="{{ obj.parent.get_absolute_url()|absolutify }}" title="{{ obj.parent.title }}">
+    {% trans locale=obj.parent.locale %}View {locale} parent{% endtrans %}
+  </a>
+  (<a href="{{ mod_url|absolutify }}">{% trans modified=obj.parent.modified %}last modified at {modified}{% endtrans %}</a>)
```
groovecoder: Sorry, can we add the `add_utm` helper to all the urls? That way we count every feed-driven session in the feed campaign.

robhudson: Same arguments?

groovecoder: Yes please.

kuma/wiki/templates/wiki/feed_docs_updated.html
```diff
@@ -0,0 +1,12 @@
...
+</p>
+<p>
+  <a href="{{ add_utm(obj.get_edit_url()|absolutify, 'feed', medium='format') }}" title="{{ obj.title }}">
```
groovecoder: `medium='format'` marks every pageview with the literal string `format`; we should change it to be either `rss` or `json`

robhudson: ah ok. When I used `medium=format` nothing ended up in the URL for the utm_medium query string. I'll use `rss` since that's what this used in.

groovecoder: Is it always 'rss'? I thought this was used by `json` feeds too

groovecoder: `rss` is fine though; this is enough scope-creep.

robhudson:
Updated

groovecoder:
Hmm ... not sure why I'm seeing this now?

l10n-feed-error

robhudson:
> Hmm ... not sure why I'm seeing this now?

Because I didn't use `{% trans %}` correct. :hankey:   Updated.

groovecoder:
Sorry, I'm going to live into my analytics-all-the-things reputation ...

Need to add the `add_utm` call to the `DocumentsFeed.item_link`, and `DocumentsFeed.item_author_link` methods too, so the feed item title links have it also.

robhudson:
> Need to add the add_utm call to the DocumentsFeed.item_link, and DocumentsFeed.item_author_link methods too, so the feed item title links have it also.

D00d.

robhudson:
Moved utm stuff to https://github.com/mozilla/kuma/pull/3625

PR #3624, opened 2015-11-05
fix bug 1221110 - private, no-cache ban responses

groovecoder:
This adds `private, no-cache` to the response when a banned user's request is caught by the `BanMiddleware`.

Assuming our load-balancer follows standard `Cache-Control` implementation, this should stop it from caching these responses.

In theory, `no-cache` alone can do this; `private` is an extra indicator to the cache that the response is meant for a single user.

We can deploy and test this ourselves to see if it fixes the problem before getting WebOps involved debugging the load balancer.

### DIFF
--- a/kuma/users/middleware.py
+++ b/kuma/users/middleware.py
@@ -1,5 +1,6 @@
 from django.contrib.auth import logout
 from django.shortcuts import render
+from django.utils.cache import add_never_cache_headers
 
 from .models import UserBan
 
@@ -17,8 +18,10 @@ class BanMiddleware(object):
             if not bans:
                 return None
             logout(request)
-            return render(request,
-                          'users/user_banned.html',
-                          {'bans': bans,
-                           'path': request.path})
+            banned_response = render(request, 'users/user_banned.html', {
+                'bans': bans,
+                'path': request.path
+            })
+            add_never_cache_headers(banned_response)
+            return banned_response
         return None
--- a/kuma/users/tests/test_middleware.py
+++ b/kuma/users/tests/test_middleware.py
@@ -1,4 +1,4 @@
-from nose.plugins.attrib import attr
+import mock
 
 from . import UserTestCase
 from ..models import UserBan
@@ -7,8 +7,8 @@ from ..models import UserBan
 class BanTestCase(UserTestCase):
     localizing_client = True
 
-    @attr('bans')
-    def test_ban_middleware(self):
+    @mock.patch('django.utils.cache.add_never_cache_headers')
+    def test_ban_middleware(self, mock_add_nch):
         """Ban middleware functions correctly."""
         self.client.login(username='testuser', password='testpass')
 
@@ -24,3 +24,4 @@ class BanTestCase(UserTestCase):
 
         resp = self.client.get('/')
         self.assertTemplateUsed(resp, 'users/user_banned.html')
+        self.assertTrue(mock_add_nch.called)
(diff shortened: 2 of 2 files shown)


groovecoder:
Changing assignee to @robhudson

kuma/users/middleware.py
```diff
@@ -17,8 +17,10 @@ def process_request(self, request):
...
-            return render(request,
-                          'users/user_banned.html',
-                          {'bans': bans,
-                           'path': request.path})
+            banned_response = render(request, 'users/user_banned.html', {
+                'bans': bans,
+                'path': request.path
+            })
+            banned_response['Cache-Control'] = 'private, no-cache'
+            return banned_response
```
robhudson: I'm not sure I like `private`. It suggests it's ok to cache this response in some situations (if the cache is private to this user).

Instead I would follow Django's `@never_cache` decorator and do what it does:
https://github.com/django/django/blob/master/django/views/decorators/cache.py#L57-L59

robhudson:
r+ with the addition of a test for the cache header.

jwhitlock:
I disagree with caching for the user either. They should be able to refresh the page and read the content, rather than the "you are a spammer" message.

groovecoder:
Added test line for the response header, which I changed to just `no-cache`.

groovecoder:
`never_cache` sets `Expires: <now>` ... I don't care which method we use. :neutral_face:

robhudson:
I imagine Django has discovered some edge cases with caching headers over their years, which is why I like the idea of using `add_never_cache_headers` if the intention is to never cache this response.

groovecoder:
Agree; updated to use `add_never_cache_headers` and to assert it was called in the test.

groovecoder:
I don't understand why Travis is failing on this. The test passes on my local dev vm. @robhudson ?

robhudson:
Tests pass for me locally also. I don't understand why either.

groovecoder:
Updated test to assert `Expires` time is less than or equal to current time. (Also, timezones can go DIA:fire:)

kuma/users/tests/test_middleware.py
```diff
@@ -24,3 +24,6 @@ def test_ban_middleware(self):
 
         resp = self.client.get('/')
         self.assertTemplateUsed(resp, 'users/user_banned.html')
+        expires_datetime = datetime.strptime(resp['Expires'],
+                                             '%a, %d %b %Y %H:%M:%S %Z')
+        assert expires_datetime.timetuple() <= gmtime()
```
robhudson: `email.utils.parsedate` can parse these RFC 2822 dates giving something you can pass to `time.mktime` and compare it from there? Might save this ugly strptime format string?

robhudson: I'm fine with this too if you'd prefer to run with it.

robhudson:
r+wc

kuma/users/tests/test_middleware.py
```diff
@@ -24,3 +24,4 @@ def test_ban_middleware(self):
 
         resp = self.client.get('/')
         self.assertTemplateUsed(resp, 'users/user_banned.html')
+        assert parsedate(resp['Expires']) <= gmtime()
```
robhudson: We can also add `self.assertEqual(resp['cache-control'], 'max-age=0')`

groovecoder: I thought about it, but it seems like django is changing `Cache-Control` behavior more readily than `Expires`, so I feel like this makes the test less susceptible to upstream logic changes. And the only reason I'm testing this is to make Travis pass ...

PR #3626, opened 2015-11-06
Allow removal of all *.pyc files from root.

robhudson:
The only vendor module that seems to have a problem with this (because
they have .pyc files checked into their repo) is python-magic.

According to the git submodule docs:
"dirty" will ignore all changes to the submodules work tree and takes
only differences between the HEAD of the submodule and the commit
recorded in the superproject into account.


groovecoder:
w00t! Looks good.

PR #3628, opened 2015-11-06
Fix Bug 1222631: Need Promise Polyfill

stephaniehobson:
Updated version of FFO to one that includes the promise polyfill bundled in the ffo.js file.

This is the last version of ffo.js that bundles the polyfill, after this version it is handled as a dependancy. After this upgrade we should see if bower is as smart about dependancies as we think it is.

Also, should the actual file be included in the PR? I'm not sure I did this right.


openjck:
> Also, should the actual file be included in the PR? I'm not sure I did this right.

It's perfect! :smile:

We include the files in the repository. We could alternatively run *bower-installer* on the stage and prod servers, but then our pushes would fail whenever any of the Bower sources were experiencing downtime. So this looks great: the *fontfaceobserver.js* file is updated, and the corresponding line in *bower.json* is updated accordingly. So this gives us the new version of the library and an up-to-date record of which version we're using. Nice!

PR #3629, opened 2015-11-07
Bug 1218466: Queue errors before GA loads.

stephaniehobson:
If we wait for ga to initialize we miss anything that executes before it.

### DIFF
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -38,8 +38,7 @@
             };
 
             // If Analytics has loaded, go ahead with tracking
-            // Checking for ".create" due to Ghostery mocking of ga
-            if(ga && ga.create) {
+            if(ga) {
                 // Send event to GA
                 ga('send', data);
             }


stephaniehobson:
Assigning to @darkwing since I assume he discovered the Ghostery mocking of ga to begin with and knows what bug the ga.create was added to prevent.

If the data just goes in a black hole either way I'd rather we attempt to add it to ga.

If trying to call the send method on the Ghostery surrogate causes explosions we should find another way of doing this.

kuma/static/js/analytics.js
```diff
@@ -38,8 +38,7 @@
             };
 
             // If Analytics has loaded, go ahead with tracking
-            // Checking for ".create" due to Ghostery mocking of ga
-            if(ga && ga.create) {
```
darkwing: `ga.create` needs to be checked.  If the user has Ghostery, and we don't check for `ga.create`, the user just bricks when they click a link.

PR #3631, opened 2015-11-08
Fix Bug 1222648: Track Kumascript Errors in GA

stephaniehobson:
Add hooks to markup to identify errors and javascript to parse the errors and send to google analytics.

Depends on #3629 and #3608.

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -539,13 +539,29 @@
     });
 
     /*
-        Toggle kumascript error detail pane
+        Kumascript error detected
     */
-    $('.kserrors-details-toggle').toggleMessage({
-        toggleCallback: function() {
-            $('.kserrors-details').toggleClass('hidden');
-        }
-    });
+    // is there an error?
+    var $kserrors = $('#kserrors');
+    if($kserrors.length){
+        // enable the details toggle
+        var $kserrorsToggle = $kserrors.find('.kserrors-details-toggle');
+        var $kserrorsDetails = $kserrors.find('.kserrors-details');
+        $kserrorsToggle.toggleMessage({
+            toggleCallback: function() {
+                $kserrorsDetails.toggleClass('hidden');
+            }
+        });
+        // loop through error list and log errors
+        var $kserrorsList = $('#kserrors-list');
+        $kserrorsList.each(function(){
+            var $thisError = $(this);
+            var errorType = $thisError.find('.kserror-type').text().trim();
+            var errorMacro = $thisError.find('.kserror-macro').text().trim();
+            var errorParse = $thisError.find('.kserror-parse').text().trim().replace(/\s\s+/g, ' ');
+            mdn.analytics.trackError('Kumascript Error', errorType, 'in: ' + errorMacro + '; parsing: ' + errorParse );
+        });
+    }
 
 
     /*
--- a/kuma/wiki/templates/wiki/includes/kumascript_errors.html
+++ b/kuma/wiki/templates/wiki/includes/kumascript_errors.html
@@ -20,39 +20,51 @@
         Anyone with an MDN profile can edit a document to fix an error. Use the <a href="/docs/MDN/Kuma/Troubleshooting_KumaScript_errors">KumaScript troubleshooting guide</a> as a starting point.
       {% endtrans %}</p>
 
-      <ul>
+      <ul id="kserrors-list">
       {% for error in kumascript_errors %}
         <li>
+          <dl class="compact">
             {% if error.args %}
-                {% set error_type = error.args[0] %}
-                {{ _('Error type:') }} <strong class="type">{{ error_type }}</strong><br>
-
-                {% set error_meta = error.args[2] %}
-                {% if error_meta %}
-                    {% set template_name = error_meta.name %}
-
-                    {% if '$' not in template_name %}
-                        {% set template_slug = 'Template: %s' % template_name %}
-                        {% set edit_url = url('wiki.edit', template_slug) %}
-                        {% set new_url = url('wiki.create') + '?slug=Template:%s' % template_name %}
-                    {% endif %}
-
-                    {{ _('In macro:') }} <code>{{ template_name }}</code>
-
-                    {% if 'status 404' in error.message %}
-                        {% if new_url %}
-                            (<a href="{{ new_url }}">{{ _('Create Template') }}</a>)
-                        {% endif %}
-                    {% elif edit_url  %}
-                        (<a href="{{ edit_url }}">{{ _('Edit Template') }}</a>)
-                    {% endif %}<br>
-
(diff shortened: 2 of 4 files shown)


stephaniehobson:
Had a merge conflict with the PR that changed this line:
` {% set template_slug = 'Template:{name}' | f(name=template_name) %}`

I think that was @robhudson so I'd appreciate a check that I didn't screw that up.

robhudson:
> I think that was @robhudson so I'd appreciate a check that I didn't screw that up.

What's in this branch looks good to me.

kuma/static/js/wiki.js
```diff
@@ -539,13 +539,29 @@
...
+        Kumascript error detected
     */
-    $('.kserrors-details-toggle').toggleMessage({
-        toggleCallback: function() {
-            $('.kserrors-details').toggleClass('hidden');
-        }
-    });
+    // is there an error?
+    var $kserrors = $('#kserrors');
+    if($kserrors.length > 0){
```
openjck: Alternatively:

```javascript
if($kserrors.length) {
```

kuma/static/js/wiki.js
```diff
@@ -539,13 +539,29 @@
...
+        // enable the details toggle
+        var $kserrorsToggle = $kserrors.find('.kserrors-details-toggle');
+        var $kserrorsDetails = $kserrors.find('.kserrors-details');
+        $kserrorsToggle.toggleMessage({
+            toggleCallback: function() {
+                $kserrorsDetails.toggleClass('hidden');
+            }
+        });
+        // loop through error list and log errors
+        var $kserrorsList = $kserrors.find('#kserrors-list');
```
openjck: We can probably get by with a single top-level selector here. `.find()` is sort of like a descendant selector.

```javascript
var $kserrorsList = $('#kserrors-list');
```

kuma/static/js/wiki.js
```diff
@@ -539,13 +539,29 @@
...
+        $kserrorsList.each(function(){
+            var $thisError = $(this);
+            var errorType = $thisError.find('.kserror-type').text().trim();
+            var errorMacro = $thisError.find('.kserror-macro').text().trim();
+            var errorParse = $thisError.find('.kserror-parse').text().trim().replace(/\s\s+/g, ' ');
```
openjck: I think we can omit the `.trim()` here since the replacement will remove all spaces.

openjck: Nevermind, I read that wrong. This looks good.

kuma/static/js/wiki.js
```diff
@@ -539,13 +539,29 @@
...
+            mdn.analytics.trackError('Kumascript Error', errorType, 'in: ' + errorMacro + '; parsing: ' + errorParse );
+        })
```
openjck: JSHint warns about a missing semicolon on this line

kuma/static/js/wiki.js
```diff
@@ -539,13 +539,29 @@
...
+    };
```
openjck: JSHint warns about an unnecessary semicolon on this line

openjck:
I'll have to defer to @jezdez on the template changes. I think we try to keep substantial logic in Python. Is that correct?

kuma/wiki/templates/wiki/includes/kumascript_errors.html
```diff
@@ -20,39 +20,51 @@
...
-                {{ _('Error type:') }} <strong class="type">{{ error_type }}</strong><br>
-
-                {% set error_meta = error.args[2] %}
-                {% if error_meta %}
-                    {% set template_name = error_meta.name %}
-
-                    {% if '$' not in template_name %}
-                        {% set template_slug = 'Template: %s' % template_name %}
-                        {% set edit_url = url('wiki.edit', template_slug) %}
-                        {% set new_url = url('wiki.create') + '?slug=Template:%s' % template_name %}
```
robhudson: This would be better written using `urlparams` instead of string concatenation for URL query strings:
`{% set new_url = url('wiki.create')|urlparams(slug='Template:%s' % template_name %}`

kuma/wiki/templates/wiki/includes/kumascript_errors.html
```diff
@@ -20,39 +20,51 @@
...
+              <dd class="kserror-type">
+                {{ error_type }}
+              </dd>
+              {% set error_meta = error.args[2] %}
+              {% if error_meta %}
+                {% set template_name = error_meta.name %}
+                {% if '$' not in template_name %}
+                  {% set template_slug = 'Template: %s' % template_name %}
+                  {% set edit_url = url('wiki.edit', template_slug) %}
+                  {% set new_url = url('wiki.create') + '?slug=Template:%s' % template_name %}
```
robhudson: This would be better written using `urlparams` instead of string concatenation for URL query strings:
`{% set new_url = url('wiki.create')|urlparams(slug='Template:%s' % template_name %}`

stephaniehobson: I just copied the old line, I don't understand what it's doing so I don't feel comfortable implementing these changes.

stephaniehobson:
It was not my intent to change any of the python, just to add HTML markup around it. If I haven't broken accidentally broken anything I'd rather not start making changes to the python now. 

JS changes made.

kuma/wiki/templates/wiki/includes/kumascript_errors.html
robhudson: That's fine too. The `urlparams` is just smarter about adding query string args to urls that already have them, and also urlencodes the arguments. But, this can be done another time, perhaps as a pass over all code to move to urlparams.

openjck:
> It was not my intent to change any of the python, just to add HTML markup around it. If I haven't broken accidentally broken anything I'd rather not start making changes to the python now.

Ah, my apologies. I didn't realize that the template code stayed the same. Organization is something we can think about in the longer term. I agree it doesn't make sense to make any major changes here.

openjck:
KumaScript isn't parsing macros in Vagrant for me for some reason, both on this branch and on master. I'll have to think about it. In the meantime, if anyone can run KumaScript successfully and feels the urge, feel free to grab this one from me.

openjck:
Maybe @groovecoder can grab this one? I'm still having some mysterious issues with macros.

groovecoder:
Works! I saw the GA error when there was a KS error on the page, and it went away when I fixed the error.

IIRC, we only show KS errors to super-users? So this won't necessarily show **all** KumaScript errors ... although we should have New Relic for that.

stephaniehobson:
Nope, we show the error to all users but only some subset sees the details. I was not considering that when I wrote the tracking code :( Might needs some adjustments, I'll see if it's useful and then decide what to do.

PR #3633, opened 2015-11-09
Fix bugs 1195351 and 1198301: (Live Sample "open in" buttons)

stephaniehobson:
Fix bug 1195351:

Re-jigged when the sample sections are downloaded so they are parsed by the script to make sure there's matching sample code found before we add the buttons to the page, log errors with GA if something goes wrong there.

Fix bug 1198301:

Added user to a ga segment if they click on button.

Some pages with embedLiveSample for testing purposes:
https://developer.mozilla.org/en-US/docs/User%3Astephaniehobson%3Acode
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress
https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Drawing_DOM_objects_into_a_canvas

### DIFF
--- a/kuma/static/js/wiki-samples.js
+++ b/kuma/static/js/wiki-samples.js
@@ -22,38 +22,8 @@
         var source = $this.attr('src').replace(/^https?:\/\//,'');
         source = source.slice(source.indexOf('/'), source.indexOf('$'));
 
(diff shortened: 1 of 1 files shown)


darkwing:
R+, works for me.

PR #3636, opened 2015-11-10
fix bug 1221750 - Simplify dirtiness checking

darkwing:
The checking now is prone to race conditions so we need to just simplify it.


openjck:
Fixes the issue described. R+

PR #3640, opened 2015-11-11
Fix bug 1223560: Never expire versioned assets

openjck:
#### Testing

1. Enable [production mode](https://kuma.readthedocs.org/en/latest/development.html#production-assets)
2. Verify that versioned assets (assets with an MD5 hash in the filename) are served up with an *Expires* header with a value of one year into the future

### DIFF
--- a/static/.htaccess
+++ b/static/.htaccess
@@ -1,16 +1,29 @@
-# Set far-future Expires headers for static media
 ExpiresActive on
-
-ExpiresDefault "access plus 1 week"
-ExpiresByType text/css "access plus 1 year"
-ExpiresByType text/javascript "access plus 1 year"
-ExpiresByType image/png "access plus 1 month"
-ExpiresByType image/jpeg "access plus 1 month"
-ExpiresByType image/gif "access plus 1 month"
-ExpiresByType image/vnd.microsoft.icon "access plus 1 month"
-ExpiresByType video/webm "access plus 1 week"
-ExpiresByType video/ogg "access plus 1 week"
-ExpiresByType video/x-flv "access plus 1 week"
-ExpiresByType application/x-shockwave-flash "access plus 1 week"
-
 FileETag MTime
+
+# Versioned assets (assets that contain an MD5 hash in the filename) never need
+# to expire. The longest valid Expires header is "access + 1 year".
+<If "%{REQUEST_FILENAME} =~ /\.[a-f0-9]{12}\./">
+    ExpiresDefault "access plus 1 year"
+</If>
+
+# Non-versioned assets should expire sooner
+<Else>
+    ExpiresDefault "access plus 1 week"
+
+    ExpiresByType text/css "access plus 1 year"
+    ExpiresByType text/javascript "access plus 1 year"
+    ExpiresByType image/png "access plus 1 month"
+    ExpiresByType image/jpeg "access plus 1 month"
+    ExpiresByType image/gif "access plus 1 month"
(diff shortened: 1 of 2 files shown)


static/.htaccess
```diff
@@ -1,16 +1,29 @@
-# Set far-future Expires headers for static media
 ExpiresActive on
+FileETag MTime
 
-ExpiresDefault "access plus 1 week"
-ExpiresByType text/css "access plus 1 year"
```
groovecoder: This line seems to work already to set the `.css` files to expire in a year.

static/.htaccess
```diff
@@ -1,16 +1,29 @@
...
-ExpiresByType text/javascript "access plus 1 year"
```
groovecoder: This line doesn't seem to work for the `.js` files.

groovecoder:
Sorry, deleted previous comments ...

The existing `ExpiresByType` config logic seems to already be working for assets in production?

In addition, when I applied this change, I got identical `Etag` values for the 2 separate `.css` assets on the home page. Unless it was a bug in fx dev tools.

Do we need this extra apache logic, or can we go without it?

openjck:
> The existing ExpiresByType config logic seems to already be working for assets in production?

Currently, we don't differentiate between versioned and non-versioned assets. We give all CSS and JavaScript (versioned or not) an Expires header of +1 year. Images get +1 month and fonts get +2 weeks.

After these changes, versioned and non-versioned assets will be treated differently.  Non-versioned assets (like [/static/img/blueprint.png](https://developer.mozilla.org/static/img/blueprint.png)) will continue to use the Expires headers outlined in the previous paragraph. Versioned assets (like [/static/img/blueprint.87f886510d2c.png](https://developer.cdn.mozilla.net/static/img/blueprint.87f886510d2c.png)) will Expires headers of +1 year.

So the values happen to be the same for CSS and JavaScript (we used +1 year for all versioned and non-versioned CSS/JS before these changes and will continue to do so) but versioned images and fonts get much better headers.

> In addition, when I applied this change, I got identical Etag values for the 2 separate .css assets on the home page. Unless it was a bug in fx dev tools.

Thanks for paying attention to this. I want to be sure we get this right. The second line of the *.htaccess* is `FileETag MTime`. It sounds like that causes the Etag value to reflect the modification time of the file, which would explain why some assets have identical values. I verified that the Etag values of an asset only changes when it is re-built by `collectstatic`, which in turn only happens when the contents change. So it seems to be working correctly, but I wouldn't mind a second set of eyes.

jezdez:
I would strongly recommend (especially in preparation for the AWS move that will get rid of Apache) to move us to [whitenoise](http://whitenose.readthedocs.org/) as the means to serve static files in production.

It's smart enough to know when a file that is requested contains a hash in the format as Django/pipeline uses and adds those far future headers automatically. Here's the docs for Django integration: http://whitenoise.readthedocs.org/en/latest/django.html

Note, this is a tool that works in conjunction with pipeline and is only targetted at production, not for development.

groovecoder:
Ah, thanks for the clarification @openjck. I over-looked that this adds the "never expires" headers to versioned images and fonts. That'll be a good improvement.

groovecoder:
:+1: for whitenoise ... as a follow-up, I presume?

openjck:
I would prefer to do it as a follow-up. I totally agree with using it, but we can start taking advantage of far-future headers in the meantime with these small changes.

jezdez:
Yep, totally not needed to be done in this step! Sorry, should have said that :)

PR #3641, opened 2015-11-11
Bug 1217208 - Update path to locale .po files

robhudson:


kuma/core/apps.py
```diff
@@ -12,6 +15,10 @@ class CoreConfig(AppConfig):
     name = 'kuma.core'
     verbose_name = _('Core')
 
+    def ready(self):
+        jingo.monkey.patch()
```
jezdez: Please call `super` here

robhudson: Ooh yeah, thanks.

kuma/users/templates/users/email/welcome/html.ltxt
```diff
@@ -56,6 +56,7 @@
             </li>
         </ul>
         <p>
+          {# Note: If you change this string also change ``WELCOME_EMAIL_STRINGS`` in kuma/users/tasks.py #}
```
groovecoder: I think we need to remove all the `{% trans %}` blocks in this template too, like we did in `plain.ltxt`

robhudson: No, I think `{% trans %}` is correct here and should work. I removed it from the plain text since `{% trans %}` collapses whitespace which isn't what we wanted in a plaintext email -- we want the whitespace.

groovecoder: Yup, red herring. I took the `{% trans mdn_link=... %}` block out and the Spanish email worked. So I re-compiled the `es/LC_MESSAGES/django.po` file itself and that worked too.

groovecoder:
Okay, this looks good and works well for me.

PR #3642, opened 2015-11-11
Bug 1223560: Use WhiteNoise to serve static assets

openjck:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/kuma/3642)


provisioning/roles/kuma/files/apache2/all-servers.conf
```diff
@@ -3,15 +3,6 @@
 # Our servers symlink to this file. Do not move it without first coordinating
 # with WebOps.
 
-<Location /static/fonts>
-    # Only enable CORS for fonts when the request is coming from an approved domain
-    Header unset Access-Control-Allow-Origin
-    SetEnvIf Origin "https?://(.*\.(mozilla|allizom)\.(com|org|net))" CORS=$0
-    SetEnvIf Origin "https?://(mdn\.mozillademos\.org)" CORS=$0
-    SetEnvIf Origin "https?://(kuma\.readthedocs\.org)" CORS=$0
-    Header set Access-Control-Allow-Origin %{CORS}e env=CORS
```
openjck: @groovecoder Do I remember correctly that we decided not to worry about restricting font CORS for approved any more? I'm pretty sure we did, but somehow these lines stuck around for some reason.

static/.htaccess
```diff
@@ -1,29 +1 @@
-ExpiresActive on
 FileETag MTime
```
openjck: @jezdez Do we still need this or will Whitenoise and/or Django take care of it from now on?

jezdez: Whitenoise doesn't provide Etag support out of the box since it's by design something that is left to the server to implement. The format of Etag heavily differ between server software, given our jump away from Apache soon as part of the AWS move (going to Nginx) we'll need to reevaluate the use of Etag's anyway. I'd say we remove them and only live with the ones that Whitenoise sets: `Last-Modified`, `Content-Length`, `Content-Type`, `Content-Encoding`, `Cache-Control`. That should give our CDN enough info to decide whether to fetch the files or not.

wsgi/kuma.wsgi
```diff
@@ -29,6 +29,10 @@ import manage
 from django.core.wsgi import get_wsgi_application
 django_app = get_wsgi_application()
 
+# Enable WhiteNoise
+from whitenoise.django import DjangoWhiteNoise
+django_app = DjangoWhiteNoise(django_app)
```
jezdez: There are  a few options we can set here, e.g. subclassing the class and modifying the `EXTRA_MIMETYPES` value may make sense to support more of those we have in the Apache config: https://github.com/evansd/whitenoise/blob/b3974c2ea80e93ba086331cff447a09a427a1a9f/whitenoise/base.py#L72-L73

openjck:
Updated!

I couldn't find an easy way to test this locally. Is there something you recommend?

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,12 @@
+from whitenoise.django import DjangoWhiteNoise
+
+
+class KumaDjangoWhiteNoise(DjangowhiteNoise):
+
+    def __init__(self, app):
+        DjangoWhiteNoise.__init__(self, app)
```
robhudson: What about `super(DjangoWhiteNoise, self).__init__(app)`? But also if you're not adding anything here there's no need, Python will do this for you.

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,12 @@
...
+
+    # Correct some MIME types
+    EXTRA_MIMETYPES = (
+            ('application/font-woff', '.woff'),
+            ('application/font-woff2', '.woff2'),)
```
robhudson: Should we create this like how it is done in the tests?
https://github.com/evansd/whitenoise/blob/bdeea1a18b134ba59a9f6a31dd2f1839836b4b14/tests/test_whitenoise.py#L42-L43

openjck: In terms of adding to the existing tuple, rather than replacing it? I started down that path, but then realized that the default `woff` MIME type that WhiteNoise uses is wrong. I'll have to open an issue with them about it.

https://github.com/evansd/whitenoise/blob/b3974c2ea80e93ba086331cff447a09a427a1a9f/whitenoise/base.py#L72-L73

openjck: But then, maybe a better approach would be to replace the inner tuple relating to `woff2`, so that we're completely overriding `EXTRA_MIMETYPES` given that it might contain some useful defaults later on. I'll give it a go.

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
openjck: Oh okay, great. I'll remove this then.

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
robhudson: Ah ok, I didn't see that they were also setting this and that this value was so sparse. I imagined it having a lot more. This seems fine but perhaps with a caveat if whitenoise adds more in the future?

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,16 @@
...
+    EXTRA_MIMETYPES_KUMA = (
+            ('application/font-woff', '.woff'),
+            ('application/font-woff2', '.woff2'),)
+
+    # Use our custom MIME type settings, together with the default settings, in
+    # WhiteNoise. In the case of a tie (our settings conflicting with the
+    # defaults), use ours.
+    EXTRA_MIMETYPES = tuple({e[1]: e for e in reversed(
+            EXTRA_MIMETYPES_KUMA + DjangoWhiteNoise.EXTRA_MIMETYPES
+        )}.values())
```
openjck: I'm sure there there's a better way to do this. But this works, and it gives a sense of what I'm going for.

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,15 @@
...
+    # Custom MIME types
+    KUMA_MIMETYPES = (
+            ('application/font-woff', '.woff'),
+            ('application/font-woff2', '.woff2'),)
+
+    # Use our custom MIME types, together with the default MIME types, in
+    # WhiteNoise. When our MIME types conflict with the defaults, use ours.
+    EXTRA_MIMETYPES = tuple({e[1]: e for e in reversed(
+            KUMA_MIMETYPES + DjangoWhiteNoise.EXTRA_MIMETYPES
+        )}.values())
```
openjck: I'm sure there there's a better way to do this. But it works, and it illustrates what I'm going for.

openjck:
Tests are passing. This is ready for review. :smile:

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,15 @@
+from whitenoise.django import DjangoWhiteNoise
```
jezdez: Please rename the file to `kuma/core/whitenoise.py` and then add the following as the first line to that file so it doesn't end up with a circular import error due to the equally named local module `whitenoise.py` and the package called `whitenoise`:

```python
from __future__ import absolute_import
```

That makes the feature that is in Python 3.x work on Python 2.x that we run under. More info here: https://www.python.org/dev/peps/pep-0328/

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class KumaDjangoWhiteNoise(DjangoWhiteNoise):
+
+    # Custom MIME types
+    KUMA_MIMETYPES = (
+        ('application/font-woff', '.woff'),
+        ('application/font-woff2', '.woff2'),)
```
jezdez: Whitenoise uses `font/woff2` as the mimetype for that file extension, that's no good?

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
```diff
@@ -0,0 +1,15 @@
...
+
+    # Use our custom MIME types, together with the default MIME types, in
+    # WhiteNoise. When our MIME types conflict with the defaults, use ours.
+    EXTRA_MIMETYPES = tuple({e[1]: e for e in reversed(
+        KUMA_MIMETYPES + DjangoWhiteNoise.EXTRA_MIMETYPES
+    )}.values())
```
jezdez: Ough, yeah, I'd rather just redefine the tuple completely in this case. This magic is not worth the effort given the fact that the upstream whitenoise class currently only has one item: https://github.com/evansd/whitenoise/blob/bdeea1a18b134ba59a9f6a31dd2f1839836b4b14/whitenoise/base.py#L72-L73

static/.htaccess
```diff
@@ -1,16 +0,0 @@
...
-ExpiresByType text/css "access plus 1 year"
-ExpiresByType text/javascript "access plus 1 year"
-ExpiresByType image/png "access plus 1 month"
-ExpiresByType image/jpeg "access plus 1 month"
-ExpiresByType image/gif "access plus 1 month"
-ExpiresByType image/vnd.microsoft.icon "access plus 1 month"
-ExpiresByType video/webm "access plus 1 week"
-ExpiresByType video/ogg "access plus 1 week"
-ExpiresByType video/x-flv "access plus 1 week"
-ExpiresByType application/x-shockwave-flash "access plus 1 week"
```
jezdez: I was wondering if we should work on making the expiration differ per file type as well, I think that'd be possible with the hooks in whitenoise. What do you think?

wsgi/kuma.wsgi
```diff
@@ -29,6 +29,10 @@ import manage
 from django.core.wsgi import get_wsgi_application
 django_app = get_wsgi_application()
 
+# Enable WhiteNoise
+from kuma.core.whitenoise import KumaDjangoWhiteNoise
```
jezdez: I think this should have been `from kuma.core.whitenoise.KumaDjangoWhiteNoise import KumaDjangoWhiteNoise` for the case in which the file is named `KumaDjangoWhiteNoise.py`. But if you change the `KumaDjangoWhiteNoise` class to be contained in `kuma.core.whitenoise` then this is correct.

jezdez:
Only spotchecked so far, haven't tried in a VM yet since that's broken for me right now :(

kuma/core/whitenoise/KumaDjangoWhiteNoise.py
openjck: Yeah, there's only one item in the parent class at the moment. Would this have a noticeable performance impact given that we're using a CDN? WhiteNoise could add more items in the future and this would be easier than watching their code for important changes.

static/.htaccess
openjck: Glad you brought this up. I wasn't sure about this. The WhiteNoise docs say the following about expiration values:

> The default is chosen to be short enough not to cause problems with stale versions but long enough that, if you’re running WhiteNoise behind a CDN, the CDN will still take the majority of the strain during times of heavy load.

Does that mean WhiteNoise will choose some reasonable defaults for us?

wsgi/kuma.wsgi
openjck: Great! That was an oversight alright, but it works with the new path.

provisioning/roles/kuma/files/apache2/all-servers.conf
openjck: Hm... I guess I'm not sure about whether we want to remove this. I can restore this locally and verify that it still works once I'm able to test.

openjck:
Updated. I haven't been able to spot-check this yet. Is there any easy way to run Kuma locally with wsgi?

openjck:
Hm... here's another idea. We can use Pipeline to set MIME types and avoid subclassing `DjangoWhiteNoise` altogether.

https://django-pipeline.readthedocs.org/en/latest/configuration.html#pipeline-mimetypes

I'll move ahead with that plan unless there are any objections.

openjck:
Oh nevermind, I guess that won't work because Pipeline isn't serving the assets on prod.

static/.htaccess
jezdez: Whitenoise will by default use an expiration time of 60 seconds for non-hashed files: https://github.com/evansd/whitenoise/blob/0657e7aea986fa652bd04d6ae8468a0e1129d634/whitenoise/base.py#L89

For hashed files ("immutable files") it'll use a far future value: https://github.com/evansd/whitenoise/blob/0657e7aea986fa652bd04d6ae8468a0e1129d634/whitenoise/base.py#L229-L230

So I think we should be good since in production we'll ever only use hashed values and for the non-hashed files we can allow the CDN to fetch the files every 60 seconds IMO.

kuma/core/whitenoise.py
```diff
@@ -0,0 +1,15 @@
...
+
+    # Custom MIME types
+    KUMA_MIMETYPES = (
+        ('application/font-woff', '.woff'),)
+
+    # Use our custom MIME types, together with the default MIME types, in
+    # WhiteNoise. When our MIME types conflict with the defaults, use ours.
+    EXTRA_MIMETYPES = tuple({e[1]: e for e in reversed(
+        KUMA_MIMETYPES + DjangoWhiteNoise.EXTRA_MIMETYPES
+    )}.values())
```
jezdez: This class attribute will only be evaluated once, but it's kind of smelly code and to be honest I think we should not accidentally change the mimetypes we support when we update whitenoise and it adds a new entry. In other words, let's make this a static list of mimetypes we know we want to support and ignore the upstream mimetypes.

provisioning/roles/kuma/files/apache2/all-servers.conf
jezdez: Note, Whitnoise will add `Access-Control-Allow-Origin=*` to all files it serves by default. If you think we should disable it, you can use the `WHITENOISE_ALLOW_ALL_ORIGINS` setting (http://whitenoise.readthedocs.org/en/latest/django.html?highlight=WHITENOISE_ALLOW_ALL_ORIGINS#WHITENOISE_ALLOW_ALL_ORIGINS) or pass `allow_all_origins=False` to the instantiation of `KumaDjangoWhiteNoise` in kuma.wsgi.

jezdez: Here's some more explanation why it does that: https://github.com/evansd/whitenoise/blob/0657e7aea986fa652bd04d6ae8468a0e1129d634/whitenoise/base.py#L90-L95

jezdez:
@openjck Since we currently still haven't updated key aspects of the way how we run the Python code we can't just point a Python server to pick up the wsgi file. Sorry.

provisioning/roles/kuma/files/apache2/all-servers.conf
openjck: Great. Thanks for pointing me to that documentation.

My understanding is that these lines were added to prevent others from hotlinking licensed fonts. There doesn't seem to be any documentation supporting that, though, so I'll assume it's safe to disable that protection unless someone knows otherwise.

static/.htaccess
openjck: Agreed. To my knowledge, the only pages referencing non-hashed assets are the pages of our Read the Docs theme. We can even bump the default value to "access + 1 week" since that's the lowest value we use now.

openjck:
Updated!

I added the *kuma.wsgi* file you sent me yesterday and updated *Procfile* to use it. I know that's not what we talked about earlier, but it looks like we need that to properly test assets moving forward.

kuma/vagrant-wsgi.py
```diff
@@ -0,0 +1,14 @@
+import os
```
jezdez: Sorry, this file should be called `kuma/wsgi.py`.

provisioning/roles.txt
```diff
@@ -5,3 +5,4 @@ geerlingguy.mysql,1.5.0
 Mayeu.RabbitMQ,1.4.0
 nodesource.node
 Stouts.elasticsearch,2.5.0
+gunicorn,19.3
```
jezdez: This isn't the right file for Python dependencies. Add this to the file `/provisioning/roles/kuma/tasks/python.yml` after the installation of `performance.txt` requirements file:

```yml
- name: Install some more deps in the virtualenv
  pip:
    virtualenv: "/home/vagrant/env"
    virtualenv_site_packages: yes
    name: "{{ item }}"
  with_items:
    - gunicorn
```

Procfile
```diff
@@ -1,4 +1,4 @@
-web: make clean && python2.7 manage.py runserver 0.0.0.0:8000
+web: make clean && gunicorn kuma.vagrant-wsgi -b 0.0.0.0:8000
```
jezdez: That means we'll stop using the django dev server which serves static files using the staticfiles system and instead uses whitenoise in development. I wonder if that means we'll have to run collectstatic before we run it now. Can you double check if a changed file shows up on the site without restarting Foreman?

provisioning/roles/kuma/files/vagrant/settings_local.py
```diff
@@ -108,3 +108,7 @@
     )
 
 SOCIALACCOUNT_PROVIDERS['persona']['AUDIENCE'] = 'https://developer-local.allizom.org'
+
+# Don't cache non-versioned static files in DEBUG mode
+if DEBUG:
+    WHITENOISE_MAX_AGE = 0
```
jezdez: I think that's the default, https://github.com/evansd/whitenoise/blob/0657e7aea986fa652bd04d6ae8468a0e1129d634/whitenoise/django.py#L58-L59 Unless you want to allow people to change it?

settings.py
```diff
@@ -553,6 +553,9 @@ def JINJA_CONFIG():
 STANDALONE_DOMAINS = ['django', 'javascript']
 STATICI18N_DOMAIN = 'javascript'
 
+# Cache non-versioned static files for one week
+WHITENOISE_MAX_AGE = 604800
```
jezdez: Would you mind writing that in the form of `60 * 60 * 24 * 7`? That has in my experience removed the magic behind such numbers and prevented confusion when others start working on the project.

kuma/vagrant-wsgi.py
openjck: Sure, can do. Thanks for the feedback.

Is there a name we could use that would differentiate this from *wsgi/kuma.py*? The names are close enough to be confusing. Or is the goal to incorporate *wsgi/kuma.py* into this file over time?

jezdez: The goal (once we're on AWS) is to not have a `wsgi/kuma.wsgi` anymore as that's an artifact of the way we currently run MDN via Apache w/ mod_wsgi. On AWS we'll be using Nginx and uwsgi.

Procfile
openjck: Good catch. It works the same way as `runserver` for me&mdash;collectstatic only needs to be run when `DEBUG = True`.

provisioning/roles/kuma/files/vagrant/settings_local.py
openjck: We [override the default](https://github.com/mozilla/kuma/pull/3642/files#diff-bdf3ecebd8379ca98cc89e545fc90899R557) in *settings.py*. These lines override that value when `DEBUG = True`.

settings.py
openjck: Great, completely agree. Will do.

openjck:
All good suggestions. Updated! :frog:

requirements/source.txt
```diff
@@ -115,3 +115,5 @@ django-sundial==1.0.1
 django-pipeline==1.5.4
 
 django-celery-transactions==0.3.6
+
+whitenoise==2.0.4
```
jezdez: There is a newer 2.0.6 release, sorry :(

settings.py
```diff
@@ -559,6 +559,9 @@ def JINJA_CONFIG():
 STATICI18N_ROOT = 'build/locale'
 STATICI18N_DOMAIN = 'javascript'
 
+# Cache non-versioned static files for one week
+WHITENOISE_MAX_AGE = 60 * 60 * 24 * 7
+
```
jezdez: \o/

jezdez:
LGTM r+
fpjcjvuv1h1pa

openjck:
Updated! The latest version of WhiteNoise uses exactly the `EXTRA_MIMETYPES` we need, so I removed our subclass. :raised_hands:

jezdez:
@openjck Nice!

jezdez:
r+

openjck:
Great!

This needs some coordination with WebOps. Basically, we'll need to have them stop hosting the *static* directory with Apache so that Python can do the work instead.

Let's merge this early next week and take it from there.

openjck:
Tagging as not ready (to be merged) because we want to coordinate this merge/push with WebOps tomorrow.

PR #3643, opened 2015-11-11
Fix bug 1218902 - Fix javascript l10n

robhudson:


kuma/core/helpers.py
```diff
@@ -150,8 +151,11 @@ def entity_decode(str):
 
 
 @register.function
-def inlinei18n(locale):
-    return StaticI18nJob().get(locale)
+def statici18n(locale):
+    if not locale:
+        locale = settings.LANGUAGE_CODE
+    filename = get_filename(locale, settings.STATICI18N_DOMAIN)
+    return static(os.path.join(settings.STATICI18N_OUTPUT_DIR, filename))
```
openjck: On second thought, we may not need to use Pipeline at all. If we use `staticfiles_storage.url` here and add *locale* to `STATICFILES_DIRS`, collectstatic will version of the filename and the asset will be served up with a far-future Expires header. I haven't worked out all the kinks yet, but that should work.

openjck: Can we customize where the generated `.po` files are placed? If we can, it would be best to put them into *build/locale*. That way, `collectstatic` will ultimately move them into */static/locale* rather than moving each sub-directory into */static* directly.

robhudson: Docs say, "By default, the generated catalogs are stored to STATIC_ROOT/jsi18n. You can modify the output path and more options by tweaking django-statici18n settings." So, I'd say yes.

Technically we don't want to move the `.po` files, but the generated `.js` files that are output from `compilejsi18n`.

Is `build/locale` at our root directory? E.g. you want these output as `build/es/javascript.js`?

openjck: I was thinking *build/locale* so that the files would be placed like *build/locale/es/javascript.js*. The build directory is being created and added to `STATICFILES_DIRS` as part of #3614, so by putting the *locale* directory in there `collectstatic` will see it and copy it to `/static/locale` with a versioned filename.

robhudson: I see, nice. I can probably make that happen by changing the statici18n settings.

So the build step would be first to `compilejsi18n` and then to `collectstatic`?

openjck: Yep, exactly. We would need to run `compilejsi18n` before `collectstatic`. Then we could use `staticfiles_storage.url` on this line to use the staticfiles-managed asset.

kuma/core/helpers.py
```diff
@@ -150,8 +151,11 @@ def entity_decode(str):
 
 
 @register.function
-def inlinei18n(locale):
-    return StaticI18nJob().get(locale)
+def statici18n(locale):
+    if not locale:
```
jezdez: Please just use the template tag function from statici18n here (`statici18n.templatetags.statici18n.statici18n` no kidding) to reduce the code duplication. `request.locale` is guaranteed to be set so no need to fall back to `settings.LANGUAGE_CODE` (that's already done in the locale middleware) either.

robhudson: I don't think we can if we're integrating this with django-pipeline and need to use staticfiles_storages? But, I can use `get_path` which `statici18n` just wraps.

jezdez: It uses the `static` template tag behind the scene (https://github.com/zyegfryed/django-statici18n/blob/87f70fa732b7a671660703e7096fead0f7a592dd/src/statici18n/templatetags/statici18n.py#L6): https://github.com/django/django/blob/9a2aca60304dc2e98f9ef45636e129d225cb981f/django/contrib/staticfiles/templatetags/staticfiles.py

robhudson: Ah, so it should "_just work_". :+1:

robhudson:
Updated for pipeline and all that.

settings.py
```diff
@@ -323,6 +323,7 @@ def get_locales():
     '.well-known',
     'users/persona/',
     'users/github/login/callback/',
+    'jsi18n',
```
robhudson: This is old and could be removed now that we are serving this via staticfiles.

robhudson:
I'll rebase...

kuma/core/helpers.py
```diff
@@ -150,8 +150,8 @@ def entity_decode(str):
 
 
 @register.function
-def inlinei18n(locale):
-    return StaticI18nJob().get(locale)
+def statici18n(locale):
+    return statici18n_orig(locale)
```
jezdez: You can just do `register.function(statici18n)` here and import the original without a separate name

robhudson: Less code FTW. Thanks.

settings.py
```diff
@@ -551,6 +555,7 @@ def JINJA_CONFIG():
 # in DOMAIN_METHODS.
 # http://github.com/jbalogh/zamboni/blob/d4c64239c24aa2f1e91276909823d1d1b290f0ee/settings.py#L254
 STANDALONE_DOMAINS = ['django', 'javascript']
+STATICI18N_ROOT = 'build/locale'
 STATICI18N_DOMAIN = 'javascript'
```
jezdez: Shouldn't we use the default here? `djangojs`?

settings.py
```diff
@@ -551,6 +555,7 @@ def JINJA_CONFIG():
 # in DOMAIN_METHODS.
 # http://github.com/jbalogh/zamboni/blob/d4c64239c24aa2f1e91276909823d1d1b290f0ee/settings.py#L254
 STANDALONE_DOMAINS = ['django', 'javascript']
+STATICI18N_ROOT = 'build/locale'
```
jezdez: That setting falls back to `STATIC_ROOT` so we may need to be using a full path instead of a relative one here.

jezdez:
I'll let you merge #3647 first if you want, rwc+

settings.py
robhudson: It's more than Django's javascript, it's l10n strings for all of MDN's javascript also. I find this name suitable.

settings.py
robhudson: It defaults to `STATIC_ROOT` which is the `kuma/` folder. We want to write it to `build/locale` where pipeline will pick it up and put it in `static/`. I tested and this works in my VM. If you're fine with it I'll leave it as-is.

kuma/core/middleware.py
```diff
@@ -53,7 +54,7 @@ def process_request(self, request):
             return response
 
         request.path_info = '/' + prefixer.shortened_path
-        request.locale = prefixer.locale
+        request.locale = prefixer.locale or settings.LANGUAGE_CODE
```
robhudson: @jezdez I wanted to call your attention to this. It fixes URLs w/o a locale in them (those in `LANGUAGE_URL_IGNORED_PATHS`) that were highlighted by the test failures by defaulting to `en-US`. We can keep or remove this as part of the `request.locale` -> `request.LANGUAGE_CODE` changes.

settings.py
jezdez: True :+1:

settings.py
jezdez: Ok

kuma/core/middleware.py
jezdez: I'm not sure about this, to be honest. We shouldn't break URLs or force a rendering of URLs with locale that are supposed to ignore the locale in their path. What's the worst that could happen with this change?

jezdez:
@robhudson Back to you

kuma/core/middleware.py
robhudson: > What's the worst that could happen with this change?

I was hoping you had some ideas there. :grin: 

Paths that don't have a locale presumably don't need any locale-specific content. Things like links to attachments or `robots.txt`. But, I see `/admin` in that list also and I imagine Django probably looks at the `Accept-Language` header instead. But I'd have to dig if this affects that or bypasses it since it's in the ignore list.

I see 3 options:
1. Roll with this but dig deeper on the implications.
2. Merge the `request.LANGUAGE_CODE` change first assuming it will fix this.
3. Fix the tests since there are only really 3 tests that fail w/o this change.

jezdez: That's really the issue of the custom URL prefixer and middleware we use. It confuses URL handling with language detection and completely works around Django's accepted behavior of taking `Accept-Locale` into account. It's not right to say that any non-locale specific URL doesn't require a locale attached to the request, since *any* request happens under some kind of environment that has a language. There is simply not a "language free" environment from where requests are made from. Consequently we must make sure `request.LANGUAGE_CODE` points at what we consider to be the correct language for the request, falling back to the default language if we can't deduce it from the URL.

So I think you're right we should do 1) *and* 2). The first to cover all angles and follow the idea of any request having some language and the latter to match Django's behavior of using "LANGUAGE_CODE" to define the request's language code.

robhudson:
I merged #3647 and rebased. I also removed the previous change in the middleware for now.

robhudson:
After removing the middleware changes the same attachment file tests fail because `request.LANGUAGE_CODE` is an empty string.

It looks like this originates from `split_path` returning an empty string if it doesn't find a locale prefix in the url:
https://github.com/mozilla/kuma/blob/master/kuma/core/urlresolvers.py#L107

Which then gets carried over to the `Prefixer` class:
https://github.com/mozilla/kuma/blob/master/kuma/core/urlresolvers.py#L114

And later set in the middleware:
https://github.com/mozilla/kuma/blob/master/kuma/core/middleware.py#L56

I imagine we can set fix this anywhere along that path, probably preferably at the origin in `split_path`.

robhudson:
Last night I tried patching `split_path` to return the `settings.LANGUAGE_CODE` instead of empty string and all tests pretty much broke. I think it's b/c the prefixer is also used for URL resolving. So it seems safest to patch this in the middleware.

robhudson:
@jezdez Curious of your thoughts here and what we should do next?

jezdez:
@robhudson Yeah, patching the middleware makes sense.

robhudson:
Ok, middleware patch added back in. Back to you for one last look.

jezdez:
r+

PR #3645, opened 2015-11-12
Bug 1137020: Remove unused Waffle flag

openjck:


jezdez:
r+

robhudson:
We need to remember to remove the row from database also.

openjck:
> We need to remember to remove the row from database also.

Definitely. I went ahead and removed the `hide_homepage_heading` flag from production.

PR #3646, opened 2015-11-12
Revert 3546e3550f3b0e2d5dd241087cb1934b4578d414

openjck:
Our stage and production servers use Apache 2.2, which does not support
the &lt;If&gt; directive used in this work. Let's forget about doing this with
Apache and move straight on to doing it at the application level with
WhiteNoise.

### DIFF
--- a/static/.htaccess
+++ b/static/.htaccess
@@ -1,29 +1,16 @@
+# Set far-future Expires headers for static media
 ExpiresActive on
+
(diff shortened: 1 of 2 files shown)


groovecoder:
While we wait for Travis, do you mind pushing this branch up to Mozilla repo so we can push it to stage to test there before merge?

PR #3649, opened 2015-11-12
Fix bug 1224366 - Sync up MDN_LANGUAGES with locale/ folder

robhudson:


jezdez:
I just looked at the en_US translations and I think they need to be replaced what we have in templates/ and there are some strings in the base translations that are rewritten there, which doesn’t make sense since en_US is our base translation. We should basically get rid of the templates folder and use en_US in the call to msginit. Ough, en_US is also full of fuzzy strings as well, which often don’t make any sense :(

locale/ga_IE
```diff
@@ -0,0 +1 @@
+ga
```
groovecoder: new empty directory? Will verbatim and/or pontoon populate this?

robhudson: It's a symlink. We're doing the same for `fy-NL` -> `fy` so I thought it'd be acceptable here.

groovecoder:
I got this when I tried to do a `manage.py extract` on this branch?
```
Extracting all strings in domain javascript...
Extracting all strings in domain django...
  .tox/docs/lib/python2.7/site-packages/jinja2/testsuite/res/templates/broken.html
  .tox/docs/lib/python2.7/site-packages/jinja2/testsuite/res/templates/syntaxerror.html
...
jinja2.exceptions.TemplateSyntaxError: Encountered unknown tag 'endif'. Jinja was looking for the following tags: 'endfor' or 'else'. The innermost block that needs to be closed is 'for'.
```

groovecoder:
Hmm ... but I get the same error on master too. :cry:

robhudson:
Removed `en_US` locale folder.

settings.py
```diff
@@ -134,6 +134,7 @@
     'pt-PT',
     'ro',
     'ru',
+    'sk',
```
robhudson: matjaz says to remove this so I'm working on that.

robhudson:
@groovecoder was there more you were going to look at here?

locale/ga_IE
groovecoder: Ah, didn't realize this is how GitHub displays symlinks in change diffs.

PR #3650, opened 2015-11-12
Bug 1218466: Queue errors before GA loads.

stephaniehobson:
Try to avoid missing anything that happens before GA initializes.

### DIFF
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -43,6 +43,17 @@
                 // Send event to GA
                 ga('send', data);
             }
+            else if(ga && !ga.create) {
+                // GA blocked or not yet initialized
+                // strip callback from data
+                data.hitCallback = null;
+                // add to queue without callback
+                ga('send', data);
+                // execute callback now
+                if(callback) {
+                    callback();
+                }
+            }
             else if(callback) {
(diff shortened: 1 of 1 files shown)


groovecoder:
A 9-line variation for the same 22 lines of logic?
```
            if (ga) {
                // GA blocked or not yet initialized - strip callback from data
                data.hitCallback = !ga.create ? null : data.hitCallback;
                ga('send', data);
            }
            // Always execute the callback
            if (callback) {
                callback();
            }
```
Eliminates the dupe code calls for `ga('send', data)` and `callback()`.

stephaniehobson:
The ga send function will log and execute the callback if we leave it in the data.

darkwing:
```
data.hitCallback = (ga.create ? data.hitCallback: null);
```

I'd roll with positive first.

groovecoder:
@stephaniehobson I think my ternary logic there (or @darkwing's positive logic) should remove `hitCallback` from `data` the same as the existing logic.

stephaniehobson:
But we're always executing the fallback in your proposal, correct? In cases where we pass the callback to ga sucessfully the callback will be executed twice.

groovecoder:
Aha, yeah I overlooked that.

r+ :smile:

PR #3652, opened 2015-11-13
bug 1177264 - measure dashboard pagination usage

groovecoder:
Per New Relic, many of our transaction response time spikes are coming from `kuma.dashboards.views:revisions` (green in the graph):

mdn-nr-transactions

I profiled and debugged it down to extremely slow queries against the `wiki_revision` table.

mdn-nr-revision-select

I [filed a bug with DBA to help fix it](https://bugzilla.mozilla.org/show_bug.cgi?id=1210830). The problem seems to be poor performance of DB/MySQL `OFFSET` pagination on large tables. (Note: the same problem seems to affect `kuma.users.views:user_detail` [orange in the graph] too.)

I have [a branch ready to PR](https://github.com/mozilla/kuma/compare/revision-dashboard-pagination-1177264?expand=1) that removes the default Django `OFFSET`-based pagination with [an optimized pagination approach based on "last seen ID" of the auto-increment primary index field](http://www.xarg.org/2011/10/optimized-pagination-using-mysql/). But, I've only implemented the "Next" link - not the specific page number links, because id-derived page number links requires us to store/cache a new estimated count value. I'd like to avoid it if we can.

So, before I do it, I want to measure how much users are actually clicking the page number links, vs. clicking the "Next" link. That way we can decide if I need to finish out the page number links, or if we can run with just the "Next" links.

### DIFF
--- a/kuma/dashboards/templates/dashboards/revisions.html
+++ b/kuma/dashboards/templates/dashboards/revisions.html
@@ -153,7 +153,14 @@
     // AJAX loads for pagination
     $revisionReplaceBlock.on('click', '.pagination a', function (e) {
         e.preventDefault();
-        $pageInput.val(/page=([^&#]*)/.exec(this.href)[1]);
+        var pageNum = /page=([^&#]*)/.exec(this.href)[1];
+        var linkText = this.text.trim();
+        mdn.analytics.trackEvent({
+            category: 'Dashboard Pagination',
+            action: pageNum,
+            label: linkText
+        });
+        $pageInput.val(pageNum);
         $filterForm.submit();
     });
 


groovecoder:
@darkwing or @stephaniehobson want to merge this?

groovecoder:
@stephaniehobson - can you look at this real quick? (long description in the PR, but very small change and spot-check). @darkwing's vm hates him right now. :cry:

stephaniehobson:
Filters and records R+

groovecoder:
Preliminary results from GA on which dashboard pagination links are clicked ...

ga-dash-pagination

So, if I add a "First" link to my branch, I'll have effectively covered 68% of the current clicks with just a "First" and a "Next" link on the page. Maybe more if I add a "Previous" link.

Otherwise, we have to implement the rest of the [optimized MySQL pagination](http://www.xarg.org/2011/10/optimized-pagination-using-mysql/) mechanism which is much more work. (~1 week)

@stephaniehobson - what do you think? Is "First" "Previous" and "Next" enough here?

groovecoder:
Oh, I should also note that the benefit in terms of UX should be a drastic reduction in the revision dashboard response times for all users, which are currently at 5.5s average.

stephaniehobson:
I think there's also some evidence to suggest that people are clicking on the last page too (ie, 261 and 159). Can you add a "Last" link to that list without too much trouble?

groovecoder:
I can try it, yeah.

groovecoder:
Dang. "Next" and "First" are easy to do. "Prev" is essentially impossible without implementing the full optimized MySQL pagination. @jezdez mentioned https://django-mysql.readthedocs.org/en/latest/queryset_extensions.html#approximate-counting is a good library to use for it, but that sounds more like a week-long project on something that isn't hurting much right now.

Let's keep this optimization in mind for if/when the `kuma.dashboards.views:revisions` and `kuma.users.views:user_detail` are really painful.

stephaniehobson:
Darn. First, Previous, and Next were probably enough without Last but without Previous :'(

jwhitlock:
@groovecoder just implement infinite scrolling and previous goes away

stephaniehobson:
Infinite scrolling is a UX anti-pattern but I acknowledge that this would be a compromise.

(It's an anti-pattern because it removes precisely all the useful features we want to remove)

jwhitlock:
@stephaniehobson I'm not sure I parse your last sentence.

Other options that are easy to describe but non-trivial to implement:
* "Load More" that adds more rows instead of going to the next page (infinite without the scrolling, drops "Next" and "Prev")
* Store page history in a cookie (page 1 = ID 2000-1990, page 2=1990-1980, etc.), build previous pages as you go - Allows "Next" and "Prev" on a per-user basis, challenging to reset.

groovecoder:
We started with infinite scrolling UX but yeah - it wasn't very good. :cry: 

Quick estimate of the other options:
* "Load More" - backend is done in my branch already; front-end would be 1 week
* Use cookie - could actually work really well. Worth an hour of checking on it

stephaniehobson:
Talking to the writers we've solved some mysteries:

They only use "last" and the pagination options in combination with filters most often to:
- see what the first contribution a user made is
- catch up on a week of missed opportunities

For the unfiltered view, loading the first 20 and then adding a "load more" button would satisfy the use cases.

I don't know if it makes sense to split the functionality behaviour based on if there are filters?

groovecoder:
Sounds like the next thing to learn is how often the dashboard is used with filters vs. no filters?

PR #3655, opened 2015-11-17
Bug 842340 - Fix template parse error for some document titles

robhudson:


robhudson:
This should fix these errors:
https://errormill.mozilla.org/mdn/mdn/group/410142/

groovecoder:
Thanks for the fix.

PR #3657, opened 2015-11-17
Bug 1139490: Don't use Pipeline to wrap JavaScript

openjck:
When PIPELINE_DISABLE_WRAPPER is False (the default), Pipeline wraps all
JavaScript in an anonymous function:

    (function(){
      //JS output...
    })();

We already wrap JavaScript manually, so this extra wrap is not needed.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -554,6 +554,8 @@ PUENTE = {
 STANDALONE_DOMAINS = ['django', 'javascript']
 STATICI18N_DOMAIN = 'javascript'
 
+PIPELINE_DISABLE_WRAPPER = True
+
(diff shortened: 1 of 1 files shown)


stephaniehobson:
Are there any specific symptoms I should be looking for?

stephaniehobson:
No double wrapping. Spot checked JS in production mode. :+1: Feel free to merge if there isn't something specific I should be testing.

openjck:
Nothing really in terms of behavior. But on this branch, there will be one less wrapper around JavaScript built by `collectstatic`.

PR #3659, opened 2015-11-19
fix bug 1225159 - add is_localizable to admin site

groovecoder:
When a site admin sees a page that can't be translated, they need to update the is_localizable field on the page, so translators can translate it.

Also includes UI enhancements for the doc admin page.

### DIFF
--- a/kuma/wiki/admin.py
+++ b/kuma/wiki/admin.py
@@ -268,8 +268,23 @@ class DocumentAdmin(admin.ModelAdmin):
                purge_documents,
                restore_documents)
     change_list_template = 'admin/wiki/document/change_list.html'
-    fields = ('locale', 'title', 'defer_rendering', 'render_expires',
-              'render_max_age', 'parent', 'parent_topic', 'category',)
+    fieldsets = (
+        (None, {
+            'fields': ('locale', 'title')
+        }),
+        ('Rendering', {
+            'fields': ('defer_rendering', 'render_expires', 'render_max_age')
+        }),
+        ('Topical Hierarchy', {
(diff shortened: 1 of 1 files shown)


kuma/wiki/admin.py
```diff
@@ -268,8 +268,23 @@ class Media:
...
+        ('Rendering', {
+            'fields': ('defer_rendering', 'render_expires', 'render_max_age')
+        }),
+        ('Topical Hierarchy', {
+            'fields': ('parent_topic',)
+        }),
+        ('Localization', {
+            'description': "The document should be <strong>either</strong> "
+                           "localizable, <strong>or</strong> have a parent - "
+                           "never both.",
```
robhudson: The checks in `_clean_is_localizable` seem to have more restrictions than mentioned in this description. But it seems like it would be difficult to summarize it.

groovecoder: Yeah, this description is aimed at the site admin who is changing `is_localizable` and/or `parent` values thru the admin UI. Not meant to be a comprehensive description, just the description of how the admin should use the admin UI fields.

robhudson:
r+

kuma/wiki/admin.py
jezdez: Does the included html show up correctly?

groovecoder: Yup. Works fine locally, as [the docs indicate](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets).

PR #3660, opened 2015-11-19
bug 1209185 - ro-order related model saves

groovecoder:
Django 1.8.5 raises ValueError when calling .save() on a model
object with un-saved related objects. When we move or revert a
document, we need to re-order our related model save() operations
so we keep data integrity per Django 1.8.5 practice.

Using this PR to solicit feedback from @jwhitlock, @robhudson, and/or @jezdez on this approach. If this looks okay (all test errors go away), then I can commit this straight to the mozilla/django1.8 branch.


kuma/wiki/models.py
```diff
@@ -811,6 +809,9 @@ def revert(self, revision, user, comment=None):
         revision.created = datetime.now()
         revision.creator = user
         revision.save()
+        if revision.document.original == self:
```
jezdez: I'd do checks on the pk instead of object comparisons here

kuma/wiki/models.py
```diff
@@ -811,6 +809,9 @@ def revert(self, revision, user, comment=None):
...
+            revision.based_on = revision
+        revision.save()
```
jezdez: This will trigger the `post_save` signal callbacks, are we should we should be doing that here yet?

kuma/wiki/models.py
groovecoder: This is the same comparison as exists in the code now. Can we preserve it to minimize changes since this will already be a pretty low-level refactor? I can add a `TODO` line above to change though.

jezdez: Yep, +1 `TODO` line

kuma/wiki/models.py
groovecoder: Hmm ... line 811 above it calls `save()` and triggers the `post_save` callbacks already, so it's the same order of operations we have now. But a double-call to `save()` and double `post_save` signals does seem risky. Hmm ...

PR #3661, opened 2015-11-19
Fix Bug 1226034: Console Error from Live Samples

stephaniehobson:
samples.js was raising console errors that looked like this: "Uncaught Error: Syntax error, unrecognized expression: h2[name=Usage example]" when used on pages where the code sample title was in an h3.

I tried to switch to just using $("#"+section) but it turns out we have characters in our IDs that, while legal in HTML5, throw jQuery for a loop. Plain JS FTW.

### DIFF
--- a/kuma/static/js/wiki-samples.js
+++ b/kuma/static/js/wiki-samples.js
@@ -78,8 +78,10 @@
             var htmlCode = $sample.find('pre[class*=html]').text();
             var cssCode = $sample.find('pre[class*=css]').text();
             var jsCode = $sample.find('pre[class*=js]').text();
-            var title = $sample.find('h2[name=' + section + ']').text();
-
+            // jquery can't handled all our IDs, need to use vanailla js for this
+            // http://stackoverflow.com/questions/7695898/uncaught-exception-syntax-error-unrecognized-expression-jquery
+            var title = doc.getElementById(section);
+            title = title ? title.textContent : '';
             // check that there is enough data to add buttons
             if(htmlCode.length || cssCode.length || jsCode.length){
                 // add buttons if we have good data


stephaniehobson:
Fun with testing! I have yet to find a page that had this error that also had a working live sample (investigating this problem next...).

Here is a page that was causing the error:
https://developer.mozilla.org/en-US/docs/Web/CSS/:not

openjck:
Nice! Good catch.

This would definitely be an improvement and would reduce the number of errors we see. Having these invalid IDs in markup is definitely a problem. I notice `getElementById` doesn't work for invalid IDs either unfortunately ([demo](https://jsfiddle.net/dV7xM/46/)) so we would get some empty titles.

I wonder if the invalid IDs are something we could correct on the back-end.

stephaniehobson:
They're not invalid IDs. jQuery is mis-handling perfectly valid ids :(

openjck:
> They're not invalid IDs. jQuery is mis-handling perfectly valid ids :(

Ah, gotcha. Are there characters that throw jQuery for a loop but not vanilla JS? I was having a hard time finding them in testing.

stephaniehobson:
https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Getting_started/JavaScript

The ID for "Action: A JavaScript demonstration" is valid but confuses jQuery.

jwhitlock:
It sounds like jQuery's CSS selector magic is getting in the way.  Escaping, or Vanilla JS to the rescue:

http://stackoverflow.com/questions/5552462/handling-colon-in-element-id-with-jquery

stephaniehobson:
@jwhitlock Yes, that is what this PR is for ;)

jwhitlock:
@jwhitlock fails at drive-by code review

drive by failed gif

+1 on this 2-line change, but waiting for @openjck

openjck:
Thanks for the clarification. I see the error on that page on master but not on this branch. r+

PR #3662, opened 2015-11-19
Fix Bug 1226283: Local Storage feature test

stephaniehobson:
Making local storage feature test more robust to solve problem of iOS identifying that is supports localStorage and then preventing usage in private browsing mode. (note that I was not able to duplicate this problem in my testing environment but I'm 90% sure this is the source of the problems identified by New Relic - details in bug)

Testing:
- save and recover draft
- helpfulness
- login with Persona
- check that GA event fires when login completes
- try to create a new account with a email address that matches an existing account, associate the accounts. Check you get confirmation notification. (I disconnected my git hub log in and then re-logged in with it).

### DIFF
--- a/kuma/static/js/auth.js
+++ b/kuma/static/js/auth.js
@@ -144,7 +144,7 @@
         Show notifications about account association status as part of the
         registration process.
     */
-    if('localStorage' in win) (function() {
+    if(win.mdn.features.localStorage) (function() {
         try {
             var $browserRegister = $('#browser_register');
             var matchKey = 'account-match-for';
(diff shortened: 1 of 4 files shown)


groovecoder:
Error is gone in desktop Safari private browsing window! :clap: 

All the other localStorage features still work, except ...

non-blocking: I didn't see the confirmation message when I associated my GitHub account with existing Persona account?

r+ besides

stephaniehobson:
@groovecoder It's github confirmation message is gone from prod too, I noticed it checking your subscribe to page PR. Not sure when it went or why. I guess we should file a bug.

groovecoder:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1229906

PR #3665, opened 2015-11-20
Fix bug 1222336 - Remove python braced format strings in l10n

robhudson:


kuma/wiki/events.py
```diff
@@ -70,7 +70,7 @@ def _mails(self, users_and_watches):
         log.debug('Sending edited notification email for document (id=%s)' %
                   document.id)
         subject = ugettext(
-            u'[MDN] Page "{document_title}" changed by {creator}')
+            u'[MDN] Page "%(document_title)s" changed by %(creator)s')
```
jannisleidel: That won't be enough since the `emails_with_users_and_watches` function uses new style string formatting locally. We need to update it to use `%` instead.

kuma/wiki/content.py
```diff
@@ -594,7 +594,7 @@ def process_header(self, token, buffer):
             start_inc = 2
             slug_base = slug
             while slug in self.known_ids:
-                slug = u'{0}_{1}'.format(slug_base, start_inc)
+                slug = u'%s_%s' % (slug_base, start_inc)
```
jannisleidel: I'd use `%d` for `start_inc` as it's an integer and some translation tooling knows how that and shows it to translators

jezdez: I'd use %d for start_inc as it's an integer and some translation tooling knows how that and shows it to translators

jwhitlock: yes, but the slug isn't a translated string, right?

docs/localization.rst
```diff
@@ -239,8 +35,10 @@ locale files without updating Verbatim, localizers will have merge head-aches.
 #.  Run the following in the virtual machine (see :doc:`installation`)::
 
         $ python manage.py extract
+        $ python manage.py merge
+        $ rm locale/**/*.po~  # Remove .po backup files created by merge.
```
jezdez: I've filed https://github.com/mozilla/puente/issues/49 to not having to do that

kuma/wiki/content.py
jezdez: :sigh:

jezdez:
r+

docs/localization.rst
robhudson: :+1: I was curious about that, thanks.

kuma/wiki/content.py
robhudson: I was grepping for `.format` and decided to change this anyway. Sorry for the churn.

PR #3668, opened 2015-11-23
bug 1209185 - ro-order related model saves

groovecoder:
Re-based these commits on mozilla/django-1.8.

@robhudson if these look good to you too (@jezdez already gave a review in https://github.com/mozilla/kuma/pull/3660) feel free to merge or let me know and I'll push them up to the django-1.8 branch myself.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -801,8 +801,6 @@ class Document(NotificationsMixin, models.Model):
 
     def revert(self, revision, user, comment=None):
         old_review_tags = list(revision.review_tags.names())
-        if revision.document.original == self:
-            revision.based_on = revision
         revision.id = None
         revision.comment = ("Revert to revision of %s by %s" %
                             (revision.created, revision.creator))
@@ -811,6 +809,9 @@ class Document(NotificationsMixin, models.Model):
         revision.created = datetime.now()
         revision.creator = user
         revision.save()
+        # TODO: change to primary key check instead of object comparison
+        if revision.document.original == self:
+            revision.save(update_fields=['based_on'])
         if old_review_tags:
             revision.review_tags.set(*old_review_tags)
         revision.make_current()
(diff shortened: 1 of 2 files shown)


robhudson:
What's a "ro-order"?

kuma/wiki/models.py
```diff
@@ -811,6 +809,10 @@ def revert(self, revision, user, comment=None):
         revision.created = datetime.now()
         revision.creator = user
         revision.save()
+        # TODO: change to primary key check instead of object comparison
```
robhudson: Seems like that would have been easier to add a `.pk` here than write the comment. :grin:

groovecoder: Maybe. But we're already changing so much with the update branch that I'm trying to minimize changes for now.

kuma/wiki/models.py
```diff
@@ -811,6 +809,10 @@ def revert(self, revision, user, comment=None):
...
+        if revision.document.original == self:
+            revision.based_on = revision
+        revision.save()
```
robhudson: If the only thing changing between this save and the last one is the `based_on` FK, can we use `revision.save(update_fields=['based_on'])`?

kuma/wiki/tests/test_models.py
```diff
@@ -1270,7 +1270,8 @@ def test_circular_dependency(self):
         """Make sure we can detect potential circular dependencies in
         parent/child relationships."""
         # Test detection at one level removed.
-        parent = document(title='Parent of circular-dependency document')
+        parent = document(title='Parent of circular-dependency document',
+                         save=True)
```
robhudson: off by one indent

groovecoder:
Just re-ordering the sequence of operations during these 2 workflows - i.e., _move_tree and revert. In particular, moving related model assignments to happen after the original models are saved, since saving models with un-saved related models throws the `ValueError` in django 1.8.

kuma/wiki/models.py
robhudson: I would also move this inside the `if` because when this is not True we'd be calling save 2 times right after the other with no changes. Unless I'm reading this wrong.

robhudson:
> Just re-ordering the sequence of operations during these 2 workflows

I know, I was just poking fun at "ro-order" vs "re-order".

robhudson:
After fixing the couple things in `revert` I think it'd be r+. Back to you.

groovecoder:
Hah, didn't even notice the typo the 2nd time. But I fixed it now, and made the other updates to `revert` and the flake8 fix too. Pushing to mozilla/django-1.8 branch ...

groovecoder:
Interesting and nice surprise: GitHub marked this PR merged & closed as soon as I pushed its commits into the branch. So no merge commit, but GitHub UI "just knows" what I did. :heart: :octocat:

jezdez:
@groovecoder :heart_eyes:

PR #3672, opened 2015-11-24
Bug 1070001: move addpage to component

stephaniehobson:
Originally I was just going to move `addpage-button` to its own file but then I got thinking there should be a better way of doing this so I want to replace [template:MDNButton](https://developer.mozilla.org/en-US/docs/Template:MDNButton) line 16 with this:

```
var html = "<div class='align-center'><a class='ignore-external button mega positive' href='" + $1 + "'>" + $0 + smallText + "</a></div>";
```

This pull request should allow both the old and new macro code to exist side by side until all the pages have been refreshed.

The `right-align` class does not appear to be in use so I removed it.

Note to self: create documentation bug for writers once this is in place.


kuma/static/styles/components/wiki/mega.styl
```diff
@@ -0,0 +1,41 @@
...
+    margin-bottom: $grid-spacing;
+    border-radius: 6px;
+    padding: $grid-spacing ($grid-spacing *2);
+    font-size: 28px;
+    line-height: 1.5;
+    text-align: center;
+    text-transform: none;
+
+    div {
+        font-size: $smaller-font-size;
```
openjck: We could take advantage of `set-smaller-font-size()` to remify the font size

kuma/static/styles/components/wiki/mega.styl
```diff
@@ -0,0 +1,41 @@
+/*
+    Mega, modifier class for buttons, makes them super big
+    Generated by: https://developer.mozilla.org/en-US/docs/Template:MDNButton
+*/
+
+.mega {
+    margin-bottom: $grid-spacing;
+    border-radius: 6px;
+    padding: $grid-spacing ($grid-spacing *2);
+    font-size: 28px;
```
openjck: `set-font-size(28px)`

kuma/static/styles/components/wiki/mega.styl
```diff
@@ -0,0 +1,41 @@
...
+    width: 100%;
+    text-align: center;
+}
+
+a.addpage-button {
+    margin-bottom: 20px;
+    padding: 18px 40px;
+    text-align: center;
+    border-radius: 6px;
+    font-size: 28px;
```
openjck: `set-font-size(28px)`

kuma/static/styles/components/wiki/mega.styl
```diff
@@ -0,0 +1,41 @@
...
+    display: inline-block;
+    background-color: $green;
+    white-space: nowrap;
+    color: #fff;
+    box-shadow: inset 0 -1px #486900;
+
+    div {
+        font-size: 11px;
```
openjck: `set-font-size(11px)`

stephaniehobson:
Updated :dancer:

PR #3673, opened 2015-11-24
Fix bug 1226784 - Upgrade Django

robhudson:


groovecoder:
Got a couple Persona timeouts during selenium tests, but I got them on master too. I think Persona is just having a slow day and this is good to go.

PR #3679, opened 2015-11-25
Fix Bug 1227960 - Add missing top "Previous" link to Revision History

anaran:
Use same proven logic as "to" radio button.

### DIFF
--- a/kuma/wiki/templates/wiki/list/revisions.html
+++ b/kuma/wiki/templates/wiki/list/revisions.html
@@ -52,7 +52,7 @@
                 {% if not loop.last %}<input type="radio" name="to" value="{{ rev.id }}" {% if loop.first %}checked="checked"{% endif %} />{% endif %}
               </div>
                <div class="revision-list-cell revision-list-prev">
-              {% if rev.previous_revision %}
+              {% if not loop.last %}
                 <a href="{{ url('wiki.compare_revisions', document.slug) }}?to={{ rev.id }}&amp;from={{ rev.previous_revision.id }}" rel="nofollow, noindex">{{ _('Previous') }}</a>
               {% endif %}
               </div>


anaran:
Evidently things aren't as simple as they could be:

https://travis-ci.org/mozilla/kuma/jobs/93356471#L3303

[get_previous](https://github.com/mozilla/kuma/blob/3355425fc2cfe157394e0cd9e9e266eae1eab132/kuma/wiki/views/list.py#L186) is conditional on a mysterious `pk` property.

Any advice?

anaran:
The link into the log file does not work for me, I have to scroll or search to line 3303 manually,

groovecoder:
Yeah, this is trickier than it should be ...

* The list view uses a custom `get_previous` transform on the QuerySet to only get approved revisions. (This is old code from SUMO where revisions have to be approved before they are published. On MDN, we auto-approve every revision and rely on post-publish editorial control to catch bad content & actors.) It also skips `current_revision.pk == previous_revision.pk` which is probably the actual cause of this bug.

* We have a [`Revision.previous`](https://github.com/mozilla/kuma/blob/8ef038905a99adfc3dd7b736a87ecb3ad3e0d1c8/kuma/wiki/models.py#L1759-L1770) property that we use, but it's expensive because it runs an additional query on the revisions table and sorts it by date to find the latest revision.

I would suggest trying to remove the `current_revision.pk == previous_revision.pk` condition from the `get_previous` queryset transform function. Do you have kuma running locally?

anaran:
No don't have kuma running locally
I'm trying to do small contributions via github and CI.

groovecoder:
Yeah, that won't work for this one. You'll need to verify it's all working locally too.

groovecoder:
Closing this unless/until we can verify the `get_previous` change on a local instance.

If you'd like to try running a local instance, here are the installation docs:

https://kuma.readthedocs.org/en/latest/installation.html

PR #3682, opened 2015-11-27
Fix Bug 1210471: Switch to em based media queries

stephaniehobson:


kuma/static/styles/includes/mixins.styl
```diff
@@ -50,6 +50,13 @@ set-larger-font-size() {
     set-font-size($base-bump-font-size);
 }
 
+media-query-emify(value) {
+    /* media queries work of browser base font size (16px),
+       even if we change html font size */
+    v = unit(value / 16px, 'em');
```
openjck: Would it make sense to use `$base-font-size` here instead of `16px`? Or is *our* base font size irrelevant?

stephaniehobson: I tried it with our base font size and it put all the break points in the wrong places. The calculations work as expected with the browser's base font size of 16px.

openjck: Gotcha, thanks!

openjck:
Hm... I'm seeing the same behavior on this branch. I must be doing something wrong. What steps should I be taking?

openjck:
It's strange... even after changing the default font size in Firefox and choosing to ignore font settings that websites define, the font size on MDN stays the same. All of our relative fonts continue to be relative to 16px. The same happens in Chrome. Other sites respect the browser font settings in both browsers.

I must be missing something obvious. Hmm...

openjck:
Hm... it looks like browser font size settings don't have an effect on MDN, for me at least. The font size is always the same regardless of browser settings. But if I remove the `font-size: 14px` rule from `html`, MDN *does* use browser-defined font sizes.

So while breakpoints are different for me on this branch, the site doesn't look broken for me on master when a different browser font size is used. I must be missing something...

stephaniehobson:
I fear this is a deep deep rabbit hole.

groovecoder:
bunnyman-underground

stephaniehobson:
So. Many. Ways. To. Change. Font. Size. 0.0

stephaniehobson:
Okay, there's the deal. Between Firefox, Chrome, Safari, and IE, and between iOS and Windows there are 18 different ways to increase the font size of a web page.

Of these, 4 combinations benifit from the move to em based media queries (including the one I tested) and 2 actually suffer (including the one @openjck tested). This is tied to the fact that we have set a default font value on the `html` element in pixels.

I'm going to withdraw this PR and amend the bug to also require us to move off a pixel declaration on the `html` element, which should improve 9 of the text-sizing options.

PR #3683, opened 2015-11-28
Bug 1100643: Review card-grid

stephaniehobson:
Moved card-grid declaration out of customcss file, simplified (moved it to flex box), and variablized it.


openjck:
Love it.

I notice that these changes remove vertical spacing between boxes on mobile. I could go either way. Was that intended?

stephaniehobson:
That was not my intent, there must be something wrong with the stylus declaration. Good catch, thanks. :bread:

stephaniehobson:
Updated :dancer:

PR #3691, opened 2015-12-03
Bug 1206309: Heading in drop down.

stephaniehobson:
- Style changes to accommodate heading in bc-history drop down
- Variablize $bc-grid-spacing
- Add "implementation notes" to title for button
- Fix problem with height of bc-history drop down in desktop display

Need to update embedCompatTable template for testing: https://gist.github.com/stephaniehobson/2b69ff6bb3be1a39b23a

Headsup: update staging when merged

### DIFF
--- a/kuma/static/js/wiki-compat-tables.js
+++ b/kuma/static/js/wiki-compat-tables.js
@@ -20,7 +20,7 @@
     var historyCount = 0;
 
     // The open button template
-    var $historyLink = $('<button title="' + gettext('Open') + '" class="bc-history-link only-icon" tabindex="-1"><span>' + gettext('Open') + '</span><i class="ic-history" aria-hidden="true"></i></button>');
+    var $historyLink = $('<button title="' + gettext('Open implementation notes') + '" class="bc-history-link only-icon" tabindex="-1"><span>' + gettext('Open') + '</span><i class="ic-history" aria-hidden="true"></i></button>');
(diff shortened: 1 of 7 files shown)


openjck:
@groovecoder would you mind grabbing this one? I would appreciate it since it would allow me to keep doing `vagrant destroy` while testing #3689.

openjck:
r+ for code, haven't spot-checked

groovecoder:
Yeah, but I've lost the instructions for testing browsercompat tables locally ... :neutral_face:

stephaniehobson:
Sample page here: https://developer.allizom.org/en-US/docs/User:stephaniehobson:compat 
You'll also need the macro from here: https://developer.allizom.org/en-US/docs/Template%3AEmbedCompatTable
And the translation macro:
https://developer.allizom.org/en-US/docs/Template:L10n:CompatTable

Waffle on the `compat-api` flag for yourself.

The links at the top of the sample page are other pages that have the table embeded on them. Some work and some exist to test conditions where they don't work but you should see the same behaviour between stage and test.

stephaniehobson:
*re reads her own comments*

Grab the gist from  https://gist.github.com/stephaniehobson/2b69ff6bb3be1a39b23a instead of copying EmbedCompatTable from staging.

openjck:
Looks great! r+

PR #3692, opened 2015-12-04
Fix bug 1223109 - decomission demo studio

robhudson:
Flagging as "not-ready" because there's things that need to happen before this is merged. See the bug.

I could use some front-end help to verify that the stuff I removed is ok and that there aren't more things that could be removed.


darkwing:
22976418d36d47555d33475b83033600 800x626x1

openjck:
Superseded by #3698

PR #3693, opened 2015-12-04
Fix Bug 1228442 - Report errors from CDN served JS files.

stephaniehobson:
Two pieces to this:
1) add `crossorigin="anonymous"` to our `<script>` tags
2) enable CORS for our domain on the CDN

https://raygun.io/blog/2015/05/fix-script-error-and-get-the-most-data-possible-from-cross-domain-js-errors/

### DIFF
--- a/provisioning/roles/kuma/files/apache2/all-servers.conf
+++ b/provisioning/roles/kuma/files/apache2/all-servers.conf
@@ -12,6 +12,14 @@
     Header set Access-Control-Allow-Origin %{CORS}e env=CORS
 </Location>
 
+<Location /static/build/js>
+    # Only enable CORS for JavaScript when the request is coming from an approved domain
+    Header unset Access-Control-Allow-Origin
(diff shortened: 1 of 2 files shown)


stephaniehobson:
@groovecoder Is this the correct place to be enabling CORS for the CDN? Is it okay that my editor added a line of whitespace to the end of the .conf file? Does this require any help from ops to publish?

groovecoder:
It looks right. IIRC the CDN will pass on whatever headers are served in the responses from Apache.

I'm trying to spot-check this, but even though the elements contain `crossorigin=anonymous`, the requests don't include an `Origin` header. :cry: Do I need any other trick besides [serving production assets](https://kuma.readthedocs.org/en/latest/development.html#production-assets)?

groovecoder:
Hmm ... though it looks like Chrome is sending the `Origin` header properly. But it's also throwing a `net::ERR_INSECURE_RESPONSE` error. :cry:

stephaniehobson:
Need to configure this in whitenoise now or something.

PR #3694, opened 2015-12-08
Fix Bug 1229425: Helpfulness only ask once

stephaniehobson:
Added a check to make sure we didn't ask in another tabs since they opened the article.

### DIFF
--- a/kuma/static/js/helpfulness.js
+++ b/kuma/static/js/helpfulness.js
@@ -16,6 +16,12 @@
 
     // create a notification
     function inquire() {
+        // check we haven't asked on another page since we set timeout to trigger this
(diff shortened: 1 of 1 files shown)


kuma/static/js/helpfulness.js
```diff
@@ -16,6 +16,12 @@
 
     // create a notification
     function inquire() {
+        // double chck we havn't asked on another page in the last minute
```
robhudson: The `helpfulnessTracker` is set to `now()` plus 7 days. This comment seems incorrect in that case. (Also, multiple typos)

robhudson:
r+wc

stephaniehobson:
Comment was factually correct but unclear, updated.

PR #3697, opened 2015-12-08
Bug 875062: Move share links to end of article

stephaniehobson:
Fix bug 1229884


stephaniehobson:
It would be great if we could merge and push this Monday.

PR #3698, opened 2015-12-09
Fix bug 1223109 - decomission demo studio

openjck:


stephaniehobson:
I think there are some intern tests to be removed as well. I found some references to scrollTo in the intern tests and we're removing that library.

stephaniehobson:
modal-control.js is used by popup.js which is included in base_popup.html 

I'm 80% sure this is only used by demos and popup.js and base_popup.html can be removed as well. If I'm wrong modal-control.js should not be removed.

stephaniehobson:
The devderby tombstone page is wrapped up in this. I think we need to disentangle and keep it.

openjck:
> I think there are some intern tests to be removed as well. I found some references to scrollTo in the intern tests and we're removing that library.

If I'm not mistaken, it looks like those are uses of [window.scrollTo](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo) rather than the jQuery plugin.

openjck:
> modal-control.js is used by popup.js which is included in base_popup.html
> 
> I'm 80% sure this is only used by demos and popup.js and base_popup.html can be removed as well. If I'm wrong modal-control.js should not be removed.

Spot on. We don't need *base_popup.html* any more, so I'll go ahead and remove that and the *popup* bundle. Nice catch!

openjck:
> The devderby tombstone page is wrapped up in this. I think we need to disentangle and keep it.
> 
> Or make a new page with fewer graphic assets.

Good point. We'll have to do something with the Dev Derby page as part of [bug 1234914](https://bugzilla.mozilla.org/show_bug.cgi?id=1234914).

openjck:
@robhudson: Could I bother you to help merge master into this branch? There's a conflict with *kuma/wiki/views/misc.py* that I'm having trouble understanding myself.

robhudson:
Oh yeah, this will be an ugly rebase. I'll work on it, @openjck. If you don't hear from me in a couple days send help.

robhudson:
I rebased with master. I also updated the PR comment with some "FIXME"s that I know of, feel free to add your own.

openjck:
I created a bug tree for the work that remains to be done:

https://bugzilla.mozilla.org/showdependencytree.cgi?id=1223109&hide_resolved=0

openjck:
Just a heads up: I did a force-push to update the bug numbers in the front-end commit messages. Something to keep in mind when working with this branch.

openjck:
@jezdez @stephaniehobson: Ready for review! :smile:

There's only one outstanding piece of work related to database migrations. Aside from that this should be good to go.

jezdez:
@stephaniehobson Should https://github.com/mozilla/kuma/blob/master/kuma/static/styles/components/doc-title.styl#L1-L3 be changed here?

jezdez:
@robhudson `tagging_tag` and `tagging_taggeditem` tables can be removed from `anonymize.sql`

jezdez:
@robhudson Should we update the translation catalogs as well, to remove the strings removed from the project?

robhudson:
> Should we update the translation catalogs as well, to remove the strings removed from the project?

Yes, but I don't see a big rush to do that. If we don't do it here it would get picked up in less than a week by our normal schedule. But to keep things tidy I could add a commit here to do that.

jezdez:
@robhudson Makes sense +1

jezdez:
I forgot to say, r+wc

stephaniehobson:
@jezdez Changes to https://github.com/mozilla/kuma/blob/master/kuma/static/styles/components/doc-title.styl#L1-L3 would be some nice code clean up but are not in any way a blocker.

stephaniehobson:
R+ from a front-end perspective.

PR #3703, opened 2015-12-14
Fix bug 1002282 - Add Serbian locale

robhudson:


jezdez:
Are we going to add `sr_Latn` as well? If not why? (At Django I remember this being an area of conflict where when one form was added, the translators of the other felt ignored. I would strongly recommend adding both at the same time).

rwc+

robhudson:
> Are we going to add sr_Latn as well? If not why?

It was mentioned in the bug to only add this one and not duplicate efforts.

robhudson:
Going to merge this. If we decided we want `sr_Latn` also that's simple enough to add also.

jezdez:
I know it was mentioned in the bug and used my experience as the Django translation manager to say that `sr_Latn` should be added at the same time. I'm not okay with just merging this and you ignoring my `rwc+`. Please open a new PR adding `sr_Latn`.

robhudson:
> I'm not okay with just merging this and you ignoring my rwc+.

It certainly wasn't ignored. The original comment was posed as a question, which I answered to the best of my knowledge. The bug comment I just read was more clear of your intentions, so thanks for that. I'll certainly open a new PR now. Apologies for the misunderstanding.

PR #3705, opened 2015-12-14
Fix Bug 1232404: Contributor bar visual design

stephaniehobson:
Testing: 
- wiki page
- zone page
- zone landing
- RTL
- less than 6 avatars
- more than 6 avatars (show all button should appear)
- arriving on article from search (should not display)
- smaller screen sizes (should not display)
- long page title (title should wrap to avoid contributors)

Make sure it expands and collapses when "show all" clicked/tapped/keyed.

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -583,59 +583,46 @@
     */
     (function (){
         var hiddenClass = 'hidden';
+        var shownClass = 'shown';
         var $contributors = $('.contributor-avatars');
-        var $hiddenContributors;
         var $showAllContributors;
 
         function loadImages(selector) {
             return $contributors.find(selector).mozLazyloadImage();
         }
 
-        // Don't bother with contributor bar if it starts hidden
-        if($contributors.css('display') === 'none') return;
+        function initToggle() {
+            $showAllContributors = $('<button id="contributors-toggle" type="button" class="transparent" data-alternate-message="' + $contributors.data('alternate-message') + '">' + $contributors.data('all-text') + '</button>');
 
-        // Start displaying first contributors in list
-        loadImages('li.shown noscript');
-
-        // Setup "Show all Contributors block"
-        if ($contributors.data('has-hidden')) {
-            $showAllContributors = $('<button type="button" class="transparent">' + $contributors.data('all-text') + '</button>');
-
-            $showAllContributors.on('keypress click', function(e) {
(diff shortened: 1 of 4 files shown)


stephaniehobson:
Difficult to test with the scrubbed DB dump since avatars don't load. I edited the source using dev tools to replace the URLs with the URL to my avatar (which does show up).

kuma/static/styles/components/wiki/contributor-avatars.styl
```diff
@@ -1,18 +1,27 @@
...
     color: #777;
-    padding: 10px 10px $content-vertical-spacing;
-    border-bottom: 1px solid rgb(241, 241, 241);
-    border-top: 1px solid rgb(241, 241, 241);
-    background: none repeat scroll 0 0 rgba-fallback(rgba(234, 239, 243, 0.4));
-    margin-top: -20px;
-    margin-bottom: 20px;
+    set-smaller-font-size();
+
+    .from-search &,
```
openjck: Just curious. Why not from search?

kuma/static/js/wiki.js
```diff
@@ -583,59 +583,48 @@
...
 
-            $showAllContributors.on('keypress click', function(e) {
-                var isKeyPress = e.type === 'keypress';
-                var isEnterKey = isKeyPress && e.which === 13;
+            // toggle message and state of hidden
+            $showAllContributors.toggleMessage({
+                toggleCallback: function() {
+                    toggleImages();
+                }
+            });
```
openjck: I haven't tested it, but it looks like this should work, too:

```javascript
$showAllContributors.toggleMessage({
    toggleCallback: toggleImages
});
```

kuma/static/styles/components/wiki/contributor-avatars.styl
stephaniehobson: It over-laps with the button on the right and it was complicated to fix because the <h1> starts floating. I can fix it if we think it's worth our time but I made a value judgement that is is not, with the search interface showing it makes some amount of sense to de-clutter other parts of the interface.

openjck: Ah, yes. The from-search UI is the reason I gave up on this a while ago. Good idea! :smile:

stephaniehobson:
Updated with simpler callback syntax and rebased for good measure.

openjck:
Looks great!

PR #3707, opened 2015-12-14
Bug 1217208 - Update puente version

robhudson:


settings.py
```diff
@@ -530,11 +530,6 @@ def JINJA_CONFIG():
 PUENTE = {
     'BASE_DIR': ROOT,
     'TEXT_DOMAIN': 'django',
```
willkg: Mmm... I think `TEXT_DOMAIN` should have gone away in Puente 0.4, too. I'm not sure why I left it.

willkg:
This looks fine to me. One thing 0.4.1 added was support for the rest of the Django-related translation functions. It'd be good to do a message extraction to see if any msgids changed or new ones got added. Other than that, r+.

settings.py
robhudson: Nope, it's still there and used in the `extract` command.

willkg: Right... it's still there, but I'm wondering if I should have removed it just like we did with STANDALONE_DOMAINS. It's not documented in the docs. I forget what I was thinking.

robhudson:
> One thing 0.4.1 added was support for the rest of the Django-related translation functions. It'd be good to do a message extraction to see if any msgids changed or new ones got added. 

Ran it. Saw 2 strings that were wrapped with `ugettext_lazy` but nothing else.

Thanks @willkg!

PR #3713, opened 2015-12-16
fix bug 1172807 - alter the highlighted language used in the revision…

dirtycoder:
… page to match the behavior from document page.

The document page also only highlight JS code, [here](https://github.com/mozilla/kuma/blob/master/kuma/wiki/templates/wiki/document.html#L321).


dirtycoder:
Review @stephaniehobson?

stephaniehobson:
R+

Thanks for the contribution :)

There's another bug open for the same template if you'd like to tackle a second one: https://bugzilla.mozilla.org/show_bug.cgi?id=1172803

PR #3715, opened 2015-12-17
fix bug 1233441 - code tag overflow on mobile

braden337:
This makes text in a `<code>` tag break so it doesn't overflow and create a huge amount of whitespace when on a mobile device (or small browser window)

image


stephaniehobson:
Thanks for bug AND the fix. :+1:

PR #3719, opened 2015-12-18
Bug 1159930 - MDN on Postgresql

robhudson:


robhudson:
This is down to 3 test failures all relating to adding `class="new"` to links. I looked at the `annotateLinks` code in `kuma/wiki/content.py` but don't understand what changed here and why they are no longer getting added. I can dig deeper but if @jezdez or @groovecoder have any hints it'd be much appreciated.

Also, a preliminary review of my changes here would be useful to know if this is on a good direction or not.

willkg:
I'd move the nginx-specific changes out of this PR and into another one.

It looks like you deleted all the existing migrations and created a new initial set. Is that right? I can't tell from the commit comments what happened.

For this test, we should ditch the msg in the assert:

```
_______________ DocumentParsingTests.test_cached_content_fields ________________
  File "/home/travis/build/mozilla/kuma/kuma/wiki/tests/test_models.py", line 1757, in test_cached_content_fields
    normalize_html(d.get_body_html()))
  File "/home/travis/build/mozilla/kuma/kuma/core/tests/__init__.py", line 23, in eq_
    assert first == second, msg
AssertionError: u'<h2 id="First">First</h2><p>This is a document</p><!-- --><!-- --><h2 id="Second">Second</h2><p>Another section</p><a href="/en-US/docs/document-with-sections">Existing link</a><a class="new" href="/en-US/docs/does-not-exist">New link</a>' != u'<h2 id="First">First</h2><p>This is a document</p><!-- --><!-- --><h2 id="Second">Second</h2><p>Another section</p><a class="new" href="/en-US/docs/document-with-sections">Existing link</a><a class="new" href="/en-US/docs/does-not-exist">New link</a>'
```

Then py.test can do its fancy assert stuff.

If @jezdez or @groovecoder don't help today, I can tinker with the failing tests on monday.

robhudson:
Yep, just ignore the docker related stuff for now. I can re-organize all this once the build passes.

robhudson:
> It looks like you deleted all the existing migrations and created a new initial set. Is that right? I can't tell from the commit comments what happened.

Yes. And that's only for this branch. We'll decide more on what to do with migrations as we get closer to moving to AWS.

robhudson:
Test pass! Thanks @jwhitlock for the pointer on those last 3 test failures.

robhudson:
This branch has been updated and the docker/nginx changes have been removed. It should be mostly clean now.

jwhitlock:
This PR, along with https://github.com/robhudson/mdn-pgloader, is the closest we got so far to kuma on PostgreSQL.

There's a lot of work needed on the first commit, since we've made a ton of changes to TravisCI and added a Docker development environment of our own.  I think we could combine it with a container that does the ``pgloader`` work, and get a system that converts from MySQL to PostgreSQL _and_ can verify that it works with tests and interactive work.

There are excellent changes in the SQL and querysets to identify MySQL vs PostgreSQL changes.  I think we could potentially get some of these merged early, with engine-based switches for the code.  This would allow us to incrementally support both until we could switch entirely to pgsql.

The many migration changes worry me.  I wonder if they are completely necessary.  It would be nice to squash migrations, to minimize the changes.  For the remaining ones, we may need to use [MIGRATION_MODULES](https://docs.djangoproject.com/en/1.11/ref/settings/#migration-modules) to store parallel migrations for each engine during the transition.

@robhudson is no longer on the team, and I don't expect him to do this work.  We're planning on self-hosting MySQL with custom collation for the first migration to AWS, so this is post-AWS work, late 2017 at the earliest.  Regretfully closing the PR for now.

PR #3720, opened 2015-12-19
Fix bug 1172803: Remove revision content section and revision source …

dirtycoder:
…section's title from revision template. I also removed the test for revision content, as it's not being showed anymore.

I choose to remove the section title, as suggested in the bug report, because this way the revision view better resemble the document view.

I'm not completely familiarized with semantic markup, is the new version ok?

### DIFF
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -349,6 +349,29 @@ class ViewTests(UserTestCase, WikiTestCase):
         ok_('<svg>' not in ct)
         ok_('<a href="#">Hahaha</a>' in ct)
 
+    def test_template_revision_content(self):
+        doc = document(title='Testing Template', slug='Template:Testing', save=True)
+        r = revision(save=True, document=doc, is_approved=True)
+
+        resp = self.client.get(r.get_absolute_url())
+        page = pq(resp.content)
+
+        ok_('Revision Source' in resp.content)
+        ok_('Revision Content' not in resp.content)
+        eq_(page.find('#doc-source').parent().attr('open'), 'open')
+
+    def test_article_revision_content(self):
+        doc = document(title='Testing Article', slug='Article', save=True)
+        r = revision(save=True, document=doc, is_approved=True)
+
+        resp = self.client.get(r.get_absolute_url())
+        page = pq(resp.content)
+
+        ok_('Revision Source' in resp.content)
+        ok_('Revision Content' in resp.content)
+        eq_(page.find('#wikiArticle').parent().attr('open'), 'open')
+        eq_(page.find('#doc-source').parent().attr('open'), None)
+
     def test_raw_css_view(self):
         """The raw source for a document can be requested"""
         self.client.login(username='admin', password='testpass')
--- a/kuma/wiki/jinja2/wiki/revision.html
+++ b/kuma/wiki/jinja2/wiki/revision.html
@@ -56,14 +56,16 @@
       {% include "wiki/includes/document_tag.html" %}
     </div>
 
-    <details open="open">
-      <summary><h2>{{ _('Revision Content') }}</h2></summary>
-      <article id="wikiArticle" class="text-content">
-        {{ revision.content_cleaned|safe }}
-      </article>
-    </details>
+    {% if not document.is_template %}
+      <details open="open">
+        <summary><h2>{{ _('Revision Content') }}</h2></summary>
+        <article id="wikiArticle" class="text-content">
+          {{ revision.content_cleaned|safe }}
+        </article>
+      </details>
+    {% endif %}
 
-    <details>
+    <details {% if document.is_template %}open="open"{% endif %}>
       <summary><h2>{{ _('Revision Source') }}</h2></summary>
       <div id="doc-source">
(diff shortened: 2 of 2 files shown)


dirtycoder:
Review @stephaniehobson ?

stephaniehobson:
Hm. This is more complicated than I thought at first. We need to preserve the Revision Content for wiki articles but not for wiki templates.

I think we can borrow the test from https://github.com/mozilla/kuma/blob/master/kuma/wiki/templates/wiki/document.html#L18 to change the display logic.

dirtycoder:
Hmm, that makes sense, I'll take another look at it soon.

stephaniehobson:
This is R+ from a functionality stand point. 

Tagging @robhudson for a review of the python (even includes a test!).

stephaniehobson:
Oh, I should add. These commits will need to be squashed before we can merge them. Its up to you @dirtycoder whether or not to wait for the :+1: or do it now.

kuma/wiki/tests/test_views.py
```diff
@@ -339,17 +339,23 @@ def test_footer_contributors(self, get_contributors, flag_is_active):
...
-        """)
+    def test_template_revision_content(self):
+        doc = document(title='Testing Template', slug='Template:Testing', save=True)
+        r = revision(save=True, document=doc, is_approved=True)
+
         resp = self.client.get(r.get_absolute_url())
-        page = pq(resp.content)
-        ct = page.find('#wikiArticle').html()
-        ok_('<svg>' not in ct)
-        ok_('<a href="#">Hahaha</a>' in ct)
```
robhudson: This seems like an important test. Is there a reason why it is being removed here?

kuma/wiki/tests/test_views.py
```diff
@@ -339,17 +339,23 @@ def test_footer_contributors(self, get_contributors, flag_is_active):
...
+        ok_('Revision Content' not in resp.content)
+
+    def test_article_revision_content(self):
+        doc = document(title='Testing Article', slug='Article', save=True)
+        r = revision(save=True, document=doc, is_approved=True)
+
+        resp = self.client.get(r.get_absolute_url())
+
+        ok_('Revision Source' in resp.content)
+        ok_('Revision Content' in resp.content)
```
robhudson: I think it might be nice to include a test for whether the content has `open="open"` as well. Generally I look for things that are conditional for testing.

kuma/wiki/tests/test_views.py
dirtycoder: Bleach is not being used for quite some time, [here](https://github.com/mozilla/kuma/commit/635b2bf59b2b6ecb93990006546261d530c58aed).

I assumed since bleach was replaced for "content_cleaned" it already had a test.
After a quick look at test_models.py I didn't found any reference for "content_cleaned".

Maybe I should create a new test for this at the test_models package or just restore the view test and reword it to make clear that it's testing that the "content_cleaned" is being used. What do you think?

kuma/wiki/tests/test_views.py
dirtycoder: That's in fact a good idea. :+1: 
I'll take a look at it and also squash the commits.

kuma/wiki/tests/test_views.py
robhudson: I followed the code path. The `content_cleaned` is still processed with bleach a few levels into this call:
https://github.com/mozilla/kuma/blob/master/kuma/wiki/managers.py#L23-L42

I think the test and comment are still relevant. Since you're not changing this in the commit I'd leave the test as-is and keep these 2 new ones you added.

dirtycoder: That makes perfect sense. I'll restore it, thanks.

kuma/wiki/tests/test_views.py
dirtycoder: Do you think an assertContains counting the occurrences of 'open="open"' will be enough?

There will be more than one occurrence in this page (e.g. details for 'Revision Information' will always be open), so just counting the occurrences looks like a good balance between asserting the exhibition of the information and the isolation between the test and the markup hierarchy.

robhudson: I don't think `assertContains` is enough b/c in either case there will always be 1 open. What we want to test is which one is open and `assertContains` won't be able to test that. A quick `pyQuery` filter should do the trick. Something like `eq_(pq(resp.content)('#doc-source').parents().attr('open'), 'open')` or `eq_(pq(resp.content)('#doc-source').parents().attr('open'), None)` for the non-open case.

dirtycoder: I tried to avoid that to be less dependent from the element hierarchy but since it's ok to check it this way, I'll do it. Thanks. :)

robhudson:
The latest changes look great, thanks. We merged a major refactor in the last day so this now has merge conflicts. If you could rebase master and force push back to this branch I'll merge it. If you need help I can do this. I believe the big change is that these files now live under the jinja2/ folder.

dirtycoder:
@robhudson it's done now. Thank you.
Really big refactor btw. :tada:

PR #3721, opened 2015-12-21
Bug 957802 - Update local SSL cert to use SHA256.

jezdez:
This should prevent Chrome from showing a broken cert all the time starting 1/1/16.


jezdez:
@stephaniehobson r?

stephaniehobson:
\o/

R+

PR #3722, opened 2015-12-21
Fix bug 1126746 - Attach a event handler instead of binding to it to fix IP toggle button.

jezdez:


jezdez:
@stephaniehobson r?

stephaniehobson:
R+

PR #3726, opened 2015-12-22
Bug 1070001: Remove addpage-button

stephaniehobson:
The macro has been transitioned to use the new classes outlined in https://github.com/mozilla/kuma/pull/3672 and there are no references* to the old class left on production. We can safely remove this.

* the search is turning up 3 references for some reason but I have viewed source on those pages and confirmed it is not in use. Caching?


openjck:
r+

PR #3728, opened 2015-12-24
Fix Bug 1232379 - Make .quicklinks workable on zone landings

stephaniehobson:
Writers want to move to quick links menus as part of preparation to migrate off of zones, right now at tablet and smaller that means grey links end up on the blue background and are un-readable. 

Added a white background and a fake left border to account for the negative margin on the see-also heading. 

Test page:
https://developer.allizom.org/en-US/Add-ons


stephaniehobson:
This has more urgency than my other PRs (though, it still isn't urgent).

openjck:
Nice! r+

PR #3729, opened 2015-12-24
Fix Bug 1217085: remove the class tip

jigs1212:
# Fix Bug 1217085: remove the class tip

This commit removes the tip class from the https://github.com/mozilla/kuma/blob/36220986307fd18eeac275b9c3058d7ccae40565/kuma/static/styles/components/content.styl#L257


stephaniehobson:
R+

PR #3731, opened 2015-12-29
Update .editorconfig

woliveiras:
Edited .editorconfig to not track minified JS files (Eg.: https://github.com/django/django/blob/master/.editorconfig#L26)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,8 @@ trim_trailing_whitespace = true
 [*.{css,html,js,json,py,styl}]
 indent_style = space
 indent_size = 4
+
+# Minified JavaScript files shouldn't be changed
+[**.min.js]
+indent_style = ignore
+insert_final_newline = ignore


openjck:
Thanks for the contribution, @woliveiras. The minified JavaScript files in our repository come from third-party libraries, so we don't usually edit them. Did something come up where this was needed?

woliveiras:
Hello @openjck! 
I added because a few of editors add a new lines to the end of file every time we save.
Sometimes someone can save a file even without editing and make a commit accidentally.
This is it to protect the project of little newbies (like me). :grin:

willkg:
I think in those cases, the reviewer will notice and comment and the changes can be undone. In theory, it's great to protect all things from all possible problem scenarios, but in practice I don't think this particular issue has ever come up and thus I don't think it needs to be handled.

If it hasn't come up, I'm -1 on this.

woliveiras:
Hmmm, really @willkg

willkg:
It's been 10 days and I don't see any signs that people are interested in this, so I'm going to close it out.

PR #3732, opened 2016-01-06
Fix bug 1237072 - Fix issue with whitenoise and settings

willkg:
whitenoise needs DJANGO_SETTINGS_MODULE to be set in the environment
otherwise it raises an ImproperlyConfigured exception.

This adds that variable to the environment for the vagrant vm. Doing
that causes the DJANGO_SETTINGS_MODULE hackery in manage.py to do the
wrong thing for the "./manage.py test", so this also updates the
Makefile rule and docs to specify
DJANGO_SETTINGS_MODULE=settings_test.

Feedback? This fixes one issue and creates new ones. I'm game for better
suggestions.

### DIFF
--- a/kuma/wsgi.py
+++ b/kuma/wsgi.py
@@ -1,7 +1,6 @@
 # Fix the settings so that Whitenoise is happy
 import os
-if 'DJANGO_SETTINGS_MODULE' not in os.environ:
-    os.environ['DJANGO_SETTINGS_MODULE'] = 'settings_local'
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'settings_local')
 
 # Enable WhiteNoise
 from django.core.wsgi import get_wsgi_application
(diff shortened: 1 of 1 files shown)


willkg:
I'm check to see if this is moot given pr #3736.

willkg:
It's not moot. PR #3736 changed the other wsgi file, so this is still relevant.

willkg:
Mmmm... The more I think about this, the more I think it's a terrible change. We should instead just add:

```
import os
if 'DJANGO_MODULE_SETTINGS' not in os.environ:
    os.environ['DJANGO_MODULE_SETTINGS'] = 'settings_local'
```

to the top of `kuma/wsgi.py`.

Why? Because it only affects development when running the site with gunicorn, and it should work the same as `manage.py` which is what we use when debugging things.

jwhitlock:
Let's use the same pattern as PR #3736:

```
os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'settings_local')
```

willkg:
^^^ Switches to `.setdefault()`.

Travis doesn't touch this code, so green or not green is moot.

willkg:
@jwhitlock Thank you!

PR #3738, opened 2016-01-07
Bug 1209185 - Recreate virtualenv on every deploy and install compiled dependencies as well.

jezdez:


scripts/chief_deploy.py
```diff
@@ -109,19 +109,19 @@ def setup_dependencies(ctx):
...
         activate_env = os.path.join(VENV_BIN, 'activate_this.py')
         execfile(activate_env, dict(__file__=activate_env))
 
+        ctx.local('%s install --upgrade pip' % os.path.join(VENV_BIN, 'pip'))
         ctx.local('pip --version')
-        ctx.local('%s scripts/peep.py install -r requirements/default.txt' %
-                  python)
+        python = os.path.join(VENV_BIN, 'python')
+        ctx.local('%s scripts/peep.py install '
+                  '-r requirements/default.txt'
```
willkg: You're concatenating strnigs here, so I think you need a space after `default.txt`.

scripts/chief_deploy.py
```diff
@@ -109,19 +109,19 @@ def setup_dependencies(ctx):
         # Creating a virtualenv tries to open virtualenv/bin/python for
         # writing, but because virtualenv is using it, it fails.
         # So we delete it and let virtualenv create a new one.
```
willkg: I think this comment needs to be updated. Maybe something like:

```
# Dearly beloved. We gather here to destroy this virtualenv in the
# hopes that out of the ashes will rise another new and beautiful
# virtualenv with no mistakes in it.
```

willkg:
I don't know enough about what chief is doing to know whether this will work or not, but otherwise it looks good to me as far as I can tell except for the missing space.

scripts/chief_deploy.py
jezdez: Hah, nicely put!

jezdez:
FWIW this is blocked until webops can :+1: or :-1: this

jezdez:
@willkg correctly asserted yesterday that deleting a virtualenv during deploy may have a negative impact on running cronjobs. Since we've yet to completely migrate away from cron I would say this complicates matters enough that it's not worth doing before the move to AWS.

jezdez:
@robhudson @willkg Do you concur?

willkg:
Just to clarify, @mythmon reminded me. I totally forgot we had this issue with SUMO. SUMO has a handful of cronjobs that run very frequently throughout the day.

I want to see the list of currently running cron jobs before I weigh in. I'm working on that now.

willkg:
I've seen the crontab. I think the only one that concerns me is `update_feeds` which runs every 10 minutes.

Is it ok for that to fail when we're doing a deploy? Does it fail well? Does it send error email to developers or other annoying things?

jezdez:
@willkg It should be okay failing, it just goes through all news feeds, fetches then and creates items in our db. If it fails it'll just refill it next time it runs.

willkg:
Ok. I think this is fine, then. The worst case scenario is that we push this out, some cron jobs fail in spectacular and unforeseen ways that don't involve data loss or mysterious murders and so we revert it.

jezdez:
I opened https://bugzilla.mozilla.org/show_bug.cgi?id=1238740 to track the response from webops

jezdez:
First I want to make sure all the deps in compiled are the ones we have in prod right now.

jezdez:
Ok, only one dependency didn't match the version in prod/stage: Pillow. I updated the requirements file

I also merged `compiled.txt` into `defaults.txt` as there is no need for a distinction anymore.

@willkg r?

scripts/chief_deploy.py
```diff
@@ -106,20 +106,19 @@ def update_info(ctx):
...
+        # hopes that out of the ashes will rise another new and beautiful
+        # virtualenv with no mistakes in it.
+        ctx.local('rm -rf %s' % settings.VENV_DIR)
         ctx.local('virtualenv-2.7 --no-site-packages %s' % settings.VENV_DIR)
 
         # Activate virtualenv to append to the correct path to $PATH.
         activate_env = os.path.join(VENV_BIN, 'activate_this.py')
         execfile(activate_env, dict(__file__=activate_env))
 
+        ctx.local('%s install --upgrade pip' % os.path.join(VENV_BIN, 'pip'))
```
willkg: I think I would lock this down to a specific version of pip to prevent possible apocalyptic futures where we go to deploy only to discover some new and exciting t-retro virus created by some specific version of pip/peep and then we all die.

willkg: Maybe surface the version as a global "constant" at the top of the file, too.

settings_test.py
```diff
@@ -1,4 +1,4 @@
-from settings import *
+from settings import *  # noqa
```
willkg: Between this and @robhudson, I declare the Age of Peppering the Codebase With "noqa" has begun. Long live NOQA!

(I don't have any better ideas. I'm just kvetching.)

requirements/default.txt
```diff
@@ -166,6 +181,9 @@ ordereddict==1.1
 # sha256: Rsjbda51oFa9HMB_ohc0_i5gPRGgeDPswe63TDXHLgw
 pbr==1.8.1
 
+# sha256: DxedfnXnyDtjQblZXKHzlN5wgUhKnjUq1m1VOhw9qik
+Pillow==2.8.2
```
willkg: According to peep, this has the wrong hash:

```
THE FOLLOWING PACKAGES DIDN'T MATCH THE HASHES SPECIFIED IN THE REQUIREMENTS
FILE. If you have updated the package versions, update the hashes. If not,
freak out, because someone has tampered with the packages.

    Pillow: expected DxedfnXnyDtjQblZXKHzlN5wgUhKnjUq1m1VOhw9qik
                 got LK9bmSvb5n4FCEZYt4PU3LnPe-zaN5TZXIngV1xFPDo

-------------------------------
Not proceeding to installation.
```

I think the version got downgraded (2.9 -> 2.8.2), but the hash didn't get updated.

willkg:
That's everything I can think of. I'm not sure offhand how this will do in a deploy, but ... that's what stage is for!

scripts/chief_deploy.py
jezdez: Great idea!

requirements/default.txt
jezdez: D'oh forgot to copy in the new hash, thanks for checking.

willkg:
The changes ^^^ look good to me. r+

jezdez:
niybylu0b7zei

jezdez:
This is sadly not working as expected. I ran a deploy to stage and consecutive runs did not take pip's automatic download caching into account.

That's because peep apparently implements its own downloading code and doesn't reuse pip's code for that.

This is effectively stuck until pip8 is out with official peep-like hashing support.

jezdez:
Closing in favor of #3760

PR #3739, opened 2016-01-07
Fix bug 1237661 - Correct order of middleware for locales and sessions

robhudson:


robhudson:
I could spend a lot of time testing this but I don't think this change will affect much. Can we push to stage, verify, and try to test there? r? @jezdez

jezdez:
r+

PR #3746, opened 2016-01-09
Fix datetime format string

robhudson:
This is currently used in only a single place and is overridden. Here
I've fixed the default value to remove it from the l10n strings and
provide a better default.

### DIFF
--- a/kuma/core/templatetags/jinja_helpers.py
+++ b/kuma/core/templatetags/jinja_helpers.py
@@ -14,8 +14,7 @@ from django.contrib.messages.storage.base import LEVEL_TAGS
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.template import defaultfilters
 from django.template.loader import get_template
-from django.utils import six
-from django.utils.encoding import force_text, smart_unicode
+from django.utils.encoding import force_text
 from django.utils.html import strip_tags
 from django.utils.timezone import get_default_timezone
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


robhudson:
I didn't even see `%e` on http://strftime.org/ lulz

willkg:
`%e` isn't in any of the Python docs, but it does work:

```
Python 2.7.10 (default, Sep  8 2015, 17:20:17) 
[GCC 5.1.1 20150618 (Red Hat 5.1.1-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> datetime.datetime.now().strftime('%e')
'11'
>>> 
```

That's curious.

Anyhow, this changes the shape from something like "January, 11 2016" to something like "01/11/16". That's got a two-digit year in whatever locale I have enabled. I'm a huge non-fan of two-digit years.

Is there a reason for making this change? Did one of the L10N people ask?

kuma/core/templatetags/jinja_helpers.py
```diff
@@ -341,9 +341,5 @@ def static(path):
 def as_datetime(t, fmt=None):
     """Call ``datetime.strftime`` with the given format string."""
     if fmt is None:
-        fmt = _(u'%B %e, %Y')
-    if not six.PY3:
-        # The datetime.strftime function strictly does not
-        # support Unicode in Python 2 but is Unicode only in 3.x.
-        fmt = fmt.encode('utf-8')
+        fmt = '%x'  # %x is the locale's appropriate date representation.
```
jezdez: we're not using system locales, so this won't work

jezdez:
Also, what's the bug number for this one?

robhudson:
There's no bug number. One of the localizers asked over the weekend about this string and the lack of context so I looked it up. This template filter is only used in a single template and `fmt` is passed in so this default isn't even used. I thought it would be better to (1) remove the gettext call around this and (2) make it a more sensible default.

robhudson:
Looking a little closer I'm considering 2 options:

1. Get rid of the template tag entirely. It's only used on a single template. We can simply replace it with a call to `strftime` directly.

2. Keep it if we decided it's a useful jinja2 filter but add an L10N comment so localizers know what to do with it.

jezdez:
option 1 +1

robhudson:
Updated to remove the template helper entirely.

robhudson:
I'll merge this myself manually and also remove the string from the .po files since they are currently a source of dennis errors.

jezdez:
r+

PR #3747, opened 2016-01-11
Fix bug 1238175: Upgrade gunicorn to 19.4.5

willkg:
This fixes the bug with sendfile.

Without this, doing "foreman start web" fails.

Quick r?


jezdez:
r+

PR #3748, opened 2016-01-12
Bug 1236912: Remove mdn10

stephaniehobson:
Remove the mdn10 styles and some assets, some assets still in use by the zone styles.

Headsup: remove waffle flags from admin console after deployed.


openjck:
Are the assets in *kuma/static/img/mdn10* the ones still used by Zones? What do you think about adding a README to document that?

settings.py
```diff
@@ -716,12 +716,6 @@ def get_locales():
         ),
         'output_filename': 'build/styles/fellowship.css',
     },
-    'mdn10': {
-        'source_filenames': (
-            'css/mdn10.css',
-        ),
-        'output_filename': 'build/styles/mdn10.css',
-    },
```
openjck: Where was this bundle used? I don't notice any usage on master. Was the compressed file referenced directly somewhere?

stephaniehobson: It was being included in the zone CSS as an @import but I copy and pasted it into the zone instead so we can remove it from here. I thought it was a risk to keep it in the code base without it being linked to from anywhere in the code base.

stephaniehobson:
Now with README :tada:

willkg:
I looked through this and tested it out locally and it seems fine. I don't know anything about zones, so I'm not sure I can speak to that.

Ramifications-wise:

1. if this is broken, we should find out quickly
2. if this orphans some stuff, it'll hang out until someone discovers it later and that's not great, but it's not the end of the world (I did do a grep for "mdn10" and saw things I assume are zone things, but not much else)

Does that sound right?

stephaniehobson:
Sounds right.

If you want to see the zone changes before merging you can copy this an its child pages to your local dev environment: https://developer.mozilla.org/en-US/docs/mdn_at_ten

You also have to convert the section to a zone which can be done from the gear menu, and copy the zone settings over from prod.

willkg:
I tested the zone thing. Everything looks good to me here. r+!

PR #3750, opened 2016-01-12
Set a maximum number of retries for the chord unlock task.

jezdez:


jezdez:
@jwhitlock r?

jwhitlock:
On my local VM, I added a log line to ``unlock_chord`` (``vi /home/vagrant/env/local/lib/python2.7/site-packages/celery/app/builtins.py``):

```
logger.info("in chord_unlock, max_retries is %s", max_retries)
```

I then ran the following in a Django shell:

```
>>> from django.conf import settings
>>> settings.CELERY_ANNOTATIONS
{'cacheback.tasks.refresh_cache': {'rate_limit': '120/m'},
 'celery.chord_unlock': {'default_retry_delay': 1, 'max_retries': 300}}
>>> from kuma.wiki.tasks import build_sitemaps
>>> build_sitemaps()
```

The locale sitemaps built, and the log line was ``in chord_unlock, max_retries is None``. So, the CELERY_ANNOTATION doesn't work for me.

If I change the line in ``kuma/wiki/tasks.py``:

```
chord(header=tasks, body=post_task).apply_async(max_retries=300)
```

then the locale sitemaps build, and the log line is ``in chord_unlock, max_retries is 300``.

If I drop to ``max_retries=1``, the second retry is called before all the sitemaps are built, and an exception is raised (``Task celery.chord_unlock with id d8c075d4-0d18-4348-81b7-303e27dfc054 raised exception: MaxRetriesExceededError``), and (on my VM) an email is sent to admins.

settings.py
```diff
@@ -1008,7 +1008,7 @@ def get_locales():
 CELERY_ANNOTATIONS = {
     'cacheback.tasks.refresh_cache': {
         'rate_limit': '120/m',
-    }
+    },
```
jwhitlock: Can you remove this change from the PR? Doesn't hurt, but isn't needed either.

jwhitlock:
Can you add a bug number to the commit message?

jezdez:
@jwhitlock Done.

PR #3751, opened 2016-01-12
Fix Bug 1235075: Remove show/hide toggle

stephaniehobson:
Also fix bug 938830.


openjck:
Looks great! :balloon:

PR #3753, opened 2016-01-13
Docs updates

willkg:
Handful of minor docs updates based on my experiences down the mine.

I tried to do minor tweaks rather than rewriting. I think the docs could use some more intensive brushing up, but I didn't want to deal with that right now.

r?

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -14,7 +14,6 @@ We have a script that attempts to fix some of the most common problems with
 Kuma that are described below::
 
   pushd .
-  # open https://www.youtube.com/watch?v=ckIMuvumYrg
   cd /path/to/kuma
   vagrant halt
   make clean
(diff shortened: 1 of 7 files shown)


docs/troubleshooting.rst
```diff
@@ -14,7 +14,6 @@ We have a script that attempts to fix some of the most common problems with
 Kuma that are described below::
 
   pushd .
-  # open https://www.youtube.com/watch?v=ckIMuvumYrg
```
jezdez: :panda_face:

jezdez:
r+

docs/troubleshooting.rst
willkg: I'm sorry--I should have added comments to the PR.

I was bummed about this, but the video at that location has been removed. I thought about replacing it with something else, but generally I think we've got too much documentation and it's probably worth while removing things.

Anyhow, I'm game for replacing it with something else. Maybe this?:

https://www.youtube.com/watch?v=DtRNg5uSKQ0

PR #3754, opened 2016-01-13
fix bug 1239430: Add ihavepower command

willkg:
This adds an ihavepower Django management command for elevating user
accounts to superstaff status to ease development work.

I HAVE POWER!

r?

### DIFF
--- a/kuma/core/management/commands/ihavepower.py
+++ b/kuma/core/management/commands/ihavepower.py
@@ -1,6 +1,5 @@
+from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
-
-from kuma.users.models import User
 
 
 class Command(BaseCommand):
@@ -12,6 +11,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         username = options['username'][0]
+        User = get_user_model()
 
         try:
             user = User.objects.get(username=username)
--- a/kuma/core/management/commands/ihavepower.py
+++ b/kuma/core/management/commands/ihavepower.py
@@ -1,0 +1,28 @@
+from django.core.management.base import BaseCommand, CommandError
+
+from kuma.users.models import User
+
+
+class Command(BaseCommand):
+    help = 'Converts specified user into a superuser'
+
+    def add_arguments(self, parser):
+        parser.add_argument('username', nargs=1,
+                            help='Username address for account to admin-ize.')
+
+    def handle(self, *args, **options):
+        username = options['username'][0]
+
+        try:
+            user = User.objects.get(username=username)
+
+        except User.DoesNotExist:
+            raise CommandError('User %s does not exist.' % username)
+
+        if user.is_superuser and user.is_staff:
+            raise CommandError('User already has the power!')
+
+        user.is_superuser = True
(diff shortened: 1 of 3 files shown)


kuma/core/tests/test_commands.py
```diff
@@ -0,0 +1,35 @@
...
+
+from kuma.core.tests import TestCase
+from kuma.users.models import User
+from kuma.users.tests import user
+
+
+class TestIHavePowerCommand(TestCase):
+    def test_help(self):
+        out = StringIO()
+        with self.assertRaises(CommandError) as cm:
```
jezdez: I'm not super eager for `cm` as the variable name here despite the fact that it's in the Python docs. Maybe call it `exception_check` or something?

kuma/core/tests/test_commands.py
```diff
@@ -0,0 +1,35 @@
...
+
+        exc = cm.exception
+        eq_(exc.message, 'User fordprefect does not exist.')
+
+    def test_user_exists(self):
+        out = StringIO()
+        user(username='fordprefect', save=True)
+        call_command('ihavepower', 'fordprefect', stdout=out)
+
+        ford = User.objects.get(username='fordprefect')
```
jezdez: There is a `kuma.users.tests.UserTestCase` that provides a `self.user_model` test instance attribute that uses `django.contrib.auth.get_user_model` instead of hardcoding the use of our `User` class. I know that we won't likely change the User class, but since the swappable user model is a config value it would make the code brittle to not use the officially sanctioned way to fetch the user model.

kuma/core/management/commands/ihavepower.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management.base import BaseCommand, CommandError
+
+from kuma.users.models import User
```
jezdez: Please use `django.contrib.auth.get_user_model` here instead to fetch the User model.

jezdez:
Excellent, r+wc!

kuma/core/tests/test_commands.py
willkg: It's a context manager. Generally, like `i` for index, I see `cm` for context manager a lot. Is "exc_cm" ok?

kuma/core/tests/test_commands.py
willkg: I hadn't seen `UserTestCase`. I'll switch it!

willkg:
^^^ That covers the comments. Also, I decided to change `cm` to `commanderror_cm` which I think should be clearer.

@jezdez Can you take another gander?

kuma/core/tests/test_commands.py
jezdez: I was just looking for something more descriptive for people that are not you :)

kuma/core/tests/test_commands.py
jezdez: Woohoo!

jezdez:
@willkg The management command should also use `django.contrib.auth.get_user_model`, not only the tests.

willkg:
Bah--I totally forgot. I'll fix that now.

jezdez:
r+

PR #3755, opened 2016-01-14
Fix bug 1239244 - Indicate current page in paginator

robhudson:


kuma/core/tests/test_pagination.py
```diff
@@ -43,3 +44,13 @@ def test_paginator_filter():
     html = paginator(pager)
     doc = pyquery.PyQuery(html)
     eq_(13, len(doc('li')))
+
+
+def test_paginator_filter_current_selected():
+    # Ensure the current page has 'class="selected"'.
+    url = '%s?%s' % (reverse('search'), 'page=10')
```
jezdez: Let's use the `urlparams` utility here (and fix the above tests to use it as well).

jezdez:
Nice, r+wc

PR #3757, opened 2016-01-15
Fix bug 1239380: Switch to py.test

willkg:
This is the bare minimum amount of work required to switch from
nose/django-nose to py.test/pytest-django.

Several things to know:

1. I changed the SkipTest things to xfail since that's closer to what
   they are (i.e. "This thing fails and we don't know why--look at it
   later")

2. I removed the attr() things because they seemed arbitrary,
   inconsistent and not documented anywhere I could find. I'll make sure
   to note this in the email I send after this lands.

3. I reimplemented eq_ and ok_ rather than go through and change these
   to their py.test equivalents. We can change them over as a code
   cleanup step done on a future rainy day.

Outstanding things to do:

1. [x] test with travis (which this PR will do)
2. [x] update documentation
3. [x] update Makefile
4. [x] fix arguments in pytest.ini
5. [x] find missing 2 tests
6. [x] re-add marks
7. [ ] write up bugs for ditching eq_/ok_

### DIFF
--- a/kuma/humans/tests.py
+++ b/kuma/humans/tests.py
@@ -3,10 +3,10 @@ from os import makedirs
 from os.path import dirname, exists, isdir
 
 import fileinput
-from nose.tools import assert_equal, ok_
 
 from django.test import TestCase
 
+from kuma.core.tests import ok_
 from .models import HumansTXT, Human
 
 APP_DIR = dirname(__file__)
@@ -18,7 +18,7 @@ class HumansTest(TestCase):
         ht = HumansTXT()
 
         name = 'buddyl@example.org'
-        assert_equal('buddyl', ht.split_name(name))
+        assert 'buddyl' == ht.split_name(name)
 
     def test_basic_get_github(self):
         """
@@ -27,7 +27,7 @@ class HumansTest(TestCase):
         data = json.load(open(CONTRIBUTORS_JSON, 'rb'))
         ht = HumansTXT()
         humans = ht.get_github(data)
-        assert_equal(len(humans), 19)
+        assert len(humans) == 19
 
     def test_for_login_name_when_no_name(self):
         """
@@ -41,7 +41,7 @@ class HumansTest(TestCase):
             if h.name == "chengwang":
                 human = h
 
-        assert_equal(human.name, "chengwang")
+        assert human.name == "chengwang"
 
     def test_write_to_file(self):
         if not isdir("%s/tmp/" % APP_DIR):
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+addopts = -rsxX --reuse-db --tb=native
+testpaths = kuma
+python_files = test*.py
+DJANGO_SETTINGS_MODULE = settings_test
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,3 +3,17 @@ addopts = -rsxX --tb=native
 testpaths = kuma
 python_files = test*.py
 DJANGO_SETTINGS_MODULE = settings_test
+markers =
+    bans: (kuma)
+    current: (kuma)
+    dashboards: (kuma) 
+    edit_emails: (kuma)
+    midair: (kuma)
+    move: (kuma)
+    redirect: (kuma)
+    retitle: (kuma)
+    review_tags: (kuma)
+    security: (kuma)
+    spam: (kuma)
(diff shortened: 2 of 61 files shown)


willkg:
@jezdez Any feedback on this?

Outstanding things to do:

1. test with travis (which this PR will do)
2. update documentation
3. update Makefile
4. address any comments

kuma/humans/tests.py
```diff
@@ -41,7 +41,7 @@ def test_for_login_name_when_no_name(self):
             if h.name == "chengwang":
                 human = h
 
-        assert_equal(human.name, "chengwang")
+        assert human.name == "chengwang"
```
willkg: I miss Cheng.

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+addopts = -rsxX --reuse-db --tb=native
```
willkg: I copied these from fjord. I don't know what `X` does and I don't think `x` is actually causing it to exit fast. So it's possible this has no effect on things.

I'll look that unless someone knows and can suggest better things here.

jezdez: can you expand the parameters here with their full name so they are less cryptic for future maintainers?

willkg:
Also, with the nose system:

```
    Ran 684 tests in 161.814s

    OK (SKIP=4)

    real	3m5.080s
    user	1m29.463s
    sys	0m19.707s
```

With the py.test system:

```
677 passed, 3 xfailed, 1 xpassed, 1 pytest-warnings in 194.17 seconds
```

677 + 3 + 1 + 1 = 682.

682 != 684, so I think 2 tests are missing. I'll see if I can track those down.

jezdez:
I'm not in favor of removing the attr() as they've been used by the devs when worked on tests. You can have the same by marking them `@pytest.mark.<attr-name>` and then run the marked tests with just `py.test -m <attr-name>`.

jezdez:
Attributes to restore:

- spam
- security
- current
- dashboards
- bans
- toc
- redirect
- move
- tags
- retitle
- midair
- edit_mails

jezdez:
Let's skip the make rewrite for this PR as I'm working on it i the setting refactor #3262 as I'd rather prevent work duplication

pytest.ini
willkg: Weird how github shows these comments out of order. I copied these from fjord. I don't know if these are the set we prefer. If someone has a set they prefer, I'm all ears. Otherwise I'll figure it out.

pytest.ini
```diff
@@ -0,0 +1,5 @@
...
+testpaths = kuma
+python_files = test*.py
```
jezdez: I think we want `test_*.py` here

pytest.ini
willkg: I had forgotten what `-r` does. It show's some extra test summary for the specified characters. I find listing the skip and xfail issues at the end super helpful.

All the args are expanded here.

I think I want to remove `--reuse-db` and leave that up to devs to do manually. Otherwise, I think this is a good default list.

willkg:
I still have to figure out the missing test issue.

I also have to remove things that will cause @jezdez pain when he merges this with his settings refactor.

Otherwise, I think this is good to go at this point.

pytest.ini
willkg: Ugh. Github keeps hiding things on me.

There are a handful of files named `tests.py`. So we can change this, but we'd have to rename files.

jezdez: Ah, I didn't see that particular thing coming, okay nevermind.

robhudson:
Beautiful

willkg:
nose collected `kuma.attachments.tests.test_templates.make_test_file` as a test (it passes, btw) 3 times since it's imported and used in three places. That accounts for the missing tests.

willkg:
@jezdez Is this good to go? If not, let me know what I need to do. If it is good to go, feel free to merge it.

jezdez:
Yep, looking good!

willkg:
^^^ That rebases on top of master with the demo studio removal.

jezdez:
There is one leftover nose import in https://travis-ci.org/mozilla/kuma/jobs/102693271#L2128

Other than that r+

gwccup7xizofe

willkg:
^^^ Travis is green, so this is good to go. I'm going to mergify it now.

PR #3762, opened 2016-01-21
Bug 1189515: Details and summary

stephaniehobson:
Added a bottom margin to the `<details>` element to preserve horizontal spacing (especially before headings).

Moved the CSS generated elements added in browsers that don't support `<details>` to the right.

Testing:
https://developer.mozilla.org/en-US/docs/User:stephaniehobson:details

### DIFF
--- a/kuma/static/styles/base/elements/sectioning.styl
+++ b/kuma/static/styles/base/elements/sectioning.styl
@@ -55,6 +55,11 @@ footer {
 details {
     clear: both;
     display: block;
+    margin-bottom: $grid-spacing;
+}
+
+details + details {
+    margin-top: $grid-spacing * -1;
 }
 
 summary  {
--- a/kuma/static/styles/components/wiki/no-details.styl
+++ b/kuma/static/styles/components/wiki/no-details.styl
@@ -1,20 +1,21 @@
 /* The .no-details styles will get applied only if JavaScript is enabled and
-<details> is not natively supported. Add focus styles (for keyboard
-accessibility) */
+<details> is not natively supported. */
 .no-details {
     summary {
         cursor: pointer;
+        bidi-style(padding-right, -15px, padding-left, 0);
     }
-    summary:before {
-        set-font-size(1.25em);
-        bidi-value(float, left, right);
+    summary:after {
         content: '+';
+        bidi-value(float, right, left);
         margin-top: -.15em;
-        bidi-style(margin-left, -15px, margin-right, 0);
+        bidi-style(margin-right, -15px, margin-left, 0);
+        set-font-size(1.25em);
+        font-weight: bold;
     }
-    &.open summary:before {
+    &.open summary:after {
         content: '–';
         margin-top: -.2em;
-        bidi-style(margin-left, -13px, margin-right, 0);
+        bidi-style(margin-right, -13px, margin-left, 0);
     }
 }


willkg:
Is this really bug 1189515? That bug is closed and it doesn't look like there's outstanding work to be done.

stephaniehobson:
Bug 1189515 is the original bug for these styles. I didn't want to open a new one just to adjust them.

willkg:
So what am I looking at here? Is it "switching from LTR to RTL should have appropriate spacing around details elements"?

stephaniehobson:
Right, sorry. More verbose explanation:

There didn't used to be space on the page after the `details` elements. Details are the elements that expand/contract when clicked on. There should be space now. The test page has a details element followed by a heading, you should be able to compare the test page between prod to local and see more space. There should not be any additional space between the `details` elements when they're next to each other.

`details` also have a `summary` element that has a visual indication that it is clickable. On Chrome, which supports `details` it's a right pointing triangle. On FireFox, which does not support `details` we have a javascript shim that kicks in and adds some classes and stuff. One of the things added is a '+' to indicate that the item will expand. I moved it from the left to the right in this PR.

Short cut for testing RTL is to open devtools and change the value of the `dir` attribute. This isn't always enough for testing, but in this case it is.

willkg:
Awesome! So if I understand correctly:

1. review the code changes
2. figure out how to get that page on my local machine
3. compare prod and local with chrome with both LTR and RTL
4. compare prod and local with firefox with both LTR and RTL

Does that sound right?

stephaniehobson:
Yes, that sounds good. 

If it helps, you can copy that page wherever on your local machine, it doesn't need to be at the same URL.

kuma/static/styles/base/elements/sectioning.styl
```diff
@@ -55,6 +55,11 @@ footer {
 details {
     clear: both;
     display: block;
+    margin-bottom: $grid-spacing;
+}
+
+details + details {
```
willkg: Wow--I didn't know about this (+ is sibling of). Super handy!

willkg:
I got things working enough to test this out. The + signs are right-aligned--that looks fine. The one thing that looks odd is that in LTR on Firefox, the headings don't line up with the detail parts. It looks one or two pixels off.

* Firefox LTR screenshot: http://bluesock.org/~willkg/images/link/1453486941.png
* Chrome LTR screenshot: http://bluesock.org/~willkg/images/link/1453487644.png

willkg:
Let me inline those images:

Chrome:


Firefox:

kuma/static/styles/base/elements/sectioning.styl
stephaniehobson: Reviewing front-end code is already paying dividends!

stephaniehobson:
OMG this is a problem with all headings and blobs of text. As far as I can tell it's how it's always been. The site looked fine to me last week but now I can't unsee this 0.0

If you want to confirm the problem is not in my code you can load the page in Firefox with JavaScript disabled. The code I wrote won't take effect but the spacing problem still exists.

willkg:
I see the same-ish thing with JavaScript disabled. I say "same-ish" since all the detail sections are expanded:

screenshot from 2016-02-02 09-29-39

Aesthetically speaking, I don't think it looks terrible--just curious.

stephaniehobson:
So the problem isn't with my code. R+?

willkg:
Sure! 
r_plus

PR #3763, opened 2016-01-21
Fix Bug 1228442 - Report errors from CDN served JS files.

stephaniehobson:
Updates pipeline template to add crossorigin property as recommended by:
https://raygun.io/blog/2015/05/fix-script-error-and-get-the-most-data-possible-from-cross-domain-js-errors/


willkg:
I looked at the pipeline template and verify that the only thing we're adding is this: `crossorigin="anonymous"`. I skimmed the article for the details and it looks ok to me.

Do we already have this header?:

```
Access-Control-Allow-Origin: *
```

PR #3764, opened 2016-01-21
Fix bug 1241554: Fixes rabbitmq task to work with ansible 2

willkg:
This tweaks bits of the rabbitmq task using the changes in this PR:

https://github.com/Mayeu/ansible-playbook-rabbitmq/pull/56/

That allows the task to work with both ansible 1.9 and 2.


willkg:
To test this, you can:

1. install ansible 1.9 and create a fresh vagrant vm
2. install ansible 2 and create a fresh vagrant vm

However, building a fresh vagrant vm is currently broken since pip 8 released and vagrant isn't pinned to pip 7.

I tested both scenarios manually.

Since this only affects building fresh vagrant vms and it takes a long time to test, my vote is someone looks over the changes, mulls them over in their heads and then we just land this and fix any issues that arise in a later PR.

jezdez:
@willkg Hmm, the vendored ansible roles shouldn't be modified on their own in theory. I'm afraid that we'll never get out of that rabbit hole. Now that we're almost using Docker, can't we live with Ansible 1.9.x for now?

willkg:
I definitely see where you're coming from and think that makes sense. However, this ansible role hasn't been updated since April 2015 and there are a bunch of outstanding issues and PRs that are sitting around. It sure feels abandoned.

This particular change is small in scope. Mostly it moves a couple of files around.

I think this is unfortunate, but good enough for now.

willkg:
To answer your question "can't we live with ansible 1.9.x for now?" requires that we tweak the docs to explicitly state we only support ansible <2 and fix the installation instructions to explicitly install 1.9.x. Yeah, we could do that. I think it's more hassle and I wouldn't want to do that work or test it, but beyond that I don't have a preference.

jezdez:
@willkg FTR we already say to install 1.9.x in the docs

willkg:
None of these specify versions:

```
brew install ansible
```

```
sudo pip install ansible
```

```
$ sudo apt-get install software-properties-common
$ sudo apt-add-repository ppa:ansible/ansible
$ sudo apt-get update
$ sudo apt-get install ansible
```

```
$ sudo dnf install ansible.noarch
```

```
$ sudo yum install ansible.noarch
```

or the instructions in this blog post:

http://www.azavea.com/blogs/labs/2014/10/running-vagrant-with-ansible-provisioning-on-windows/

Does that clarify my issue?

jezdez:
@willkg Aha, I updated the prose statement in 8dda414294ea543b60819ef2796002f1652358d7 only when 2.0 wasn't out yet or in beta I don't remember.

r+

willkg:
I look forward to docker, but ... I cynically suspect it'll just be trading one set of icky things like this with another set. :(

PR #3769, opened 2016-01-28
Attachments cleanup/refactor for AWS

jezdez:


jezdez:
This is WIP, more updates coming, no need for review yet

jezdez:
@robhudson @willkg r? This needs some more tests I think, what do you think would be a good thing to test?

robhudson:
What are we doing with existing attachments? Will we have a migration to find them and put them in the database?

robhudson:
I read all this. I think next step might be to test locally if that's helpful?

jezdez:
@robhudson Yep, the checklist to test is:

- add an attachment to a article, save
- add the attachment to the content of the article with the ckeditor plugin, save
- check if the table changes the "Linked in article" column from "no" to "yes"
- reverse check
- add attachment to another article
- copy URL of attachment and add to the source of the first article
- see the table add another item that is only a link to the other attachment's article
- add some more attachments
- go to the admin and inspect the attachments and attachmentrevision UI
- see if the attachment revision inline UI in the attachment admin UI works as expected
- check if deleting a single revisions from both the revison and the attachment UI works (works = deletes attachment revision, but creates a trashed attachment item before with the same file)
- check if the file still exists on the file system
- check if deleting the created trashed attachment item actually deletes the file on the file system
- delete another attachment with multiple revisions and see if it creates multiple trashed attachment items
- run management command to delete trashed attachment items (may have to update the created datetime field to make sure the management command taks the WIKI_ATTACHMENTS_KEEP_TRASHED_DAYS constance value into account (default to only deleting trashed items that were trashed 14 days ago))

robhudson:
> check if deleting a single revisions from both the revison and the attachment UI works

From the attachments admin with the revisions inline, I don't see a way to delete a revision. But, if I go to the attachments revision admin I can delete one. When I did this it removed the revision but did not add it to the trashed attachments. I see the latest revision in the attachments table of the $edit page and it says it is linked even though revision 1 is what I see in the page.

I was expecting the page that referenced the attachment to automatically switch from revision 1 to revision 2 after deleting revision 1. But the attachment in the page still shows revision 1 of the attachment. Is it expected that you update the page if you delete an attachment?

robhudson:
I'm going to stop here. I don't think I understand what the expectation is wrt the attachment tables and the document source so I'm a bit lost on whether what I'm seeing is expected or not.

robhudson:
One thing I hit locally is it seems some of my attachments don't have a current revision:
`AttributeError: 'NoneType' object has no attribute 'filename'` at
`kuma/attachments/models.pyL48`.

jezdez:
@robhudson Ok, I've fixed your findings:

- added a way to automatically fetch missing files from the main site in development
- make secure HTTP requests by installing pyopenssl and dependencies for Python < 2.7.9
- improve the attachment admin UI
- make deletion suck less
 - make previous revision the current one, when deleting a revision
 - prevent deleting the last revision by exception when calling the orm and disabling deletion in the admin
 - disable admin delete action since they eluded modification

robhudson:
Hmm, something weird happened here:

1. I created an attachment in the `$edit` page and all was good. It showed up.
2. I added a 2nd revision of the attachment in the admin and all was good. The new image showed up.
3. I deleted the 2nd revision of the attachment in the attachment revision admin. Which, I believe, should have made the 1st revision the current revision. But it seems to have deleted the attachment completely. It's still referenced in the `$edit` page HTML but shows a broken image and there is no longer the attachment list at the bottom where it once was.

robhudson:
> add an attachment to a article, save

:+1: 

> add the attachment to the content of the article with the ckeditor plugin, save

:+1: 

> check if the table changes the "Linked in article" column from "no" to "yes"

it shows as not linked. Reloading the page fixes it and shows it as linked.

> reverse check

linked changes to no. :+1: 

> add attachment to another article

:+1: 

> copy URL of attachment and add to the source of the first article

:+1: 

> see the table add another item that is only a link to the other attachment's article

Again took another page refresh to show the linked attachment in the table

> add some more attachments
> go to the admin and inspect the attachments and attachmentrevision UI

:+1: 

> see if the attachment revision inline UI in the attachment admin UI works as expected

:+1: 

> check if deleting a single revisions from both the revison and the attachment UI works (works = deletes attachment revision, but creates a trashed attachment item before with the same file)

I deleted the whole attachment. The wiki document edit page showed the attachment removed but still referenced in the editor. Should these be removed automatically?

Trashed attachment created :+1: 

> check if the file still exists on the file system

File exists on fs

> check if deleting the created trashed attachment item actually deletes the file on the file system

Deleting trashed attachment deletes file from fs

> delete another attachment with multiple revisions and see if it creates multiple trashed attachment items

2 trashed attachments created. The first revision correctly flagged as !was_current

> run management command to delete trashed attachment items (may have to update the created datetime field to make sure the management command taks the WIKI_ATTACHMENTS_KEEP_TRASHED_DAYS constance value into account (default to only deleting trashed items that were trashed 14 days ago))

I tried changing the value to zero but this didn't delete the attachments from today. Perhaps because `trashed_at` is a datetime and the query is using a date so it floors it to 00:00:00?

Running with `-f` finds them and deleted them successfully.

Deleting the attachments created broken links in the articles that still had them referenced. I do still see the alternate text displayed so this seems a bit weird to me.

> Test that deleting the current revision of the attachment reverts to using the prior revision

This worked for me this time. I think I didn't have a completely updated branch.

> Test that there is no delete option if the attachment only has 1 revision

Yep, this works. :+1:

jezdez:
FTR, this contains migrations that conflict with migrations in #3793 and may need to be updated in case the other lands first.

jezdez:
@robhudson I've updated the deletion note template a bit in f49ea0e3a6ba010a264eb651efaf48991975f2bf to make it more clear that deleting attachments won't touch article content automatically. I think this is out of scope of this PR.

robhudson:
With code reviews and local testing I think this is good enough to merge. If we find minor issues we can iterate further. r+

PR #3718, opened 2015-12-18
Bug 1189520 - Update Raven to 5.9.2 in preparation of new Sentry version.

jezdez:
Needs work on cloudops side first.

PR #3770, opened 2016-01-28
Bug 1189520 - Update raven for new Sentry server

robhudson:
This replaces PR #3718


requirements/default.txt
```diff
@@ -31,6 +32,7 @@ django-mozilla-product-details==0.7.1 --hash=sha256:c50bffb9828a9139156904a11c50
 django-picklefield==0.3.2 --hash=sha256:5489fef164de43725242d56e65e016137d3df0d1a00672bda72d807f5b2b0d99
 django-pipeline==1.5.4 --hash=sha256:a5a909f403d0ea56f211618949ff4e378f16e5a996bde819bd2b5b0e7a9bb8fe
 django-ratelimit==0.6.0 --hash=sha256:f73b53b2c4fd342cbd4b983ffc33fdcf8f8a6f595d548865ea63c839ef38293c
+djangorestframework==3.3.2 --hash=sha256:5634b1ff56581bf0fe4075e86227fc9693c1ca031c7213c9ae942c445c24817b
```
jezdez: FTR this was output from pip freeze, hence being after all the `django-*` packages

requirements/default.txt
```diff
@@ -1,8 +1,8 @@
 # This is the default requirements list to be used in production and development
 # environments
 amqp==1.4.8 --hash=sha256:94d2a68227e7984bd40f5f71c6839730622aff8a4633f29ede1ffdd0d240795e
-argparse==1.4.0 --hash=sha256:c31647edb69fd3d465a847ea3157d37bed1f95f19760b11a47aa91c04b666314
 anyjson==0.3.3 --hash=sha256:37812d863c9ad3e35c0734c42e0bf0320ce8c3bed82cd20ad54cb34d158157ba
+argparse==1.4.0 --hash=sha256:c31647edb69fd3d465a847ea3157d37bed1f95f19760b11a47aa91c04b666314
```
jezdez: D'oh!

jezdez:
r+ assuming we update the settings as well. As long as we can't use env variables for settings this is a bit harder to do sadly.

requirements/default.txt
robhudson: Ah. I used `!sort -df` from vim so it must be slightly different. I can put this back if it makes more sense.

robhudson:
If by settings you mean `SENTRY_DSN` then yeah, I'll work with ops to do this.

robhudson:
I'll leave this branch/PR open to use as a test on stage next week. If the test goes well I will merge. Thanks.

PR #3777, opened 2016-02-09
Fix bug 1244308: Return 404 content if user can't restore the document

willkg:
Previously, MDN would return an HTTP 404 with an HTML document that
has some stuff on it and lets people restore the document. That's not
great because somehow those pages are getting indexed or otherwise
making search bots grumpy.

Instead of doing that, this changes MDN to return a straight-up
regular 404 page unless the user can restore documents in which case
they get the HTML page that lets them restore documents.

@jezdez r?

### DIFF
--- a/kuma/wiki/tests/__init__.py
+++ b/kuma/wiki/tests/__init__.py
@@ -149,6 +149,30 @@ def normalize_html(input):
             .serialize(alphabetical_attributes=True))
 
 
+def create_document_editor_user():
+    """Creates a user empowered with document editing"""
+    perms = []
+    for action in ('add', 'change', 'delete', 'view', 'restore'):
+        perms.append(Permission.objects.get(codename='%s_document' % action))
+
+    group, group_created = Group.objects.get_or_create(name='editor')
+    if group_created:
+        group.permissions = perms
+        group.save()
+
+    User = get_user_model()
+    user, user_created = User.objects.get_or_create(
+        username='conantheeditor',
+        defaults=dict(email='user_%s@example.com',
+                      is_active=True, is_staff=False, is_superuser=False))
+    if user_created:
+        user.set_password('testpass')
+        user.groups = [group]
+        user.save()
+
+    return user
+
+
 def create_template_test_users():
     perms = dict(
         (x, [Permission.objects.get(codename='%s_template_document' % x)])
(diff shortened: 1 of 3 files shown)


willkg:
This fixes the existing "deleted document returns a 404" test so that it checks to see what the content of the HTML page is, too. We're keying off of the "Reason for Deletion" text because this is a Jinja template and thus we can't assert whether the template was used to render. :(

This adds a new test verifying users with the proper permission get the HTML page with the "restore the document" stuff on it just like before.

So, if the tests look ok and Travis is cool, then I think we're good.

To test it by hand, you have to:

1. delete an existing document
2. using a user that has `wiki.restore_document` permission, view the document--you should get a 404 with the "RESTORE THE DOCUMENT" button
3. using a user that is not logged in or doesn't have the `wiki.restore_document` permission, view the document--you should get just a 404

jwhitlock:
+r tests pass, practical tests pass in VM (404 with form as admin user, regular 404 for anon user).

willkg:
Awesome. Thank you!

kuma/wiki/tests/__init__.py
```diff
@@ -149,6 +149,31 @@ def __iter__(self):
...
+
+    group, group_created = Group.objects.get_or_create(name='editor')
+    if group_created:
+        group.permissions = perms
+        group.save()
+
+    User = get_user_model()
+    user, user_created = User.objects.get_or_create(
+        username='conantheeditor',
+        defaults=dict(email='user_%s@example.com',
```
jezdez: what that `%s` intentional?

kuma/wiki/tests/__init__.py
```diff
@@ -149,6 +149,31 @@ def __iter__(self):
...
+                      is_active=True, is_staff=False, is_superuser=False))
+    if user_created:
+        user.set_password('testpass')
+        user.groups = [group]
+        user.save()
+
+    user = User.objects.get(username=user.username)
```
jezdez: just use user.refresh_from_db()

willkg: Oh--I didn't know about that. Neato!

kuma/wiki/tests/__init__.py
willkg: I mimicked `create_template_test_users` which had this as well. So, yes, it was intentional, but I have no idea why `create_template_test_users` does it.

kuma/wiki/tests/__init__.py
willkg: I nixed the line altogether since it's not needed.

jezdez:
r+

willkg:
Ok--merging this in!

jwhitlock:
I'm sad this code isn't in production yet

jwhitlock:
As of yesterday it is :smiley:

PR #3778, opened 2016-02-10
Bug 1218563 - Complete docker set up.

robhudson:
This adds celery support (with Redis broker) and nginx.


docker/images/nginx/nginx.conf
```diff
@@ -0,0 +1,34 @@
...
+
+    upstream django {
+        server web:8000;
+    }
+
+    server {
+        listen 80 default;
+        charset utf-8;
+
+        root /app/static;
```
jezdez: We don't need to serve the static files since wheelhouse can take care of that

docker/docker-compose.yml
```diff
@@ -1,22 +1,47 @@
...
+  command: supervisord -n -c /app/docker/supervisor-celery.conf
   volumes:
     - ../:/app
-  ports:
-    - "8000:8000"
   links:
     - memcached
     - mysql
     - elasticsearch
     - redis
```
andymckay: celery needs access to all those things?

andymckay:
don't remember much about kuma, but it looks ok to me r+

docker/images/nginx/nginx.conf
robhudson: I'll have to look up how wheelhouse works. This was simple to set up but I can iterate on it.

docker/docker-compose.yml
robhudson: The workers do yeah for db queries, search indexing, cached things. Redis is the broker so that's needed here too.

robhudson:
Thanks for the spot check @andymckay.

docker/images/nginx/nginx.conf
robhudson: I think you mean whitenoise. I've removed the static serving from nginx and whitenoise is handling it now.

jezdez: Oh, odd, yeah, I meant whitenoise. Sorry for the confusion.

PR #3779, opened 2016-02-11
Fix bug 1247447: Fix bootstrapping problem with product-details

willkg:
If you just built a dev environment and it was pristine and didn't
have any of the product-details files, then you were in a bit of a
pickle because in order to get product-details files, it'd run through
the settings code which required the product-details files which you
didn't have because that's the very thing you were trying to get. Ugh.

This fixes that by moving some code around such that LOCALES and
LANGUAGES get built together (LANGUAGES requires LOCALES anyhow) and
if the product-details files aren't there, it prints a warning and
rolls its eyes, but then returns empty structures. This is ok because
it only happens if there are no product-details files and that should
only be the case if you haven't gotten the product-details, yet.

r?

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -237,20 +237,39 @@ for requested_lang, delivered_lang in LOCALE_ALIASES.items():
         LANGUAGE_URL_MAP[requested_lang.lower()] = delivered_lang
 
 
-def get_locales():
+def _get_languages_and_locales():
+    """Generates LANGUAGES and LOCALES data
+
+    .. Note::
+
+       This requires product-details data. If product-details data hasn't been
+       retrieved, then this prints a warning and then returns empty values. We
+       do this because in the case of pristine dev environments, you can't
+       update product-details because product-details isn't there, yet.
+
+    """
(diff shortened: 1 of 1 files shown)


willkg:
I'm pretty sure @robhudson can trigger this trivially with his docker environment of doom. The rest of us have to go in and delete the existing product-details to create the doomsday scenario that this alleviates.

Regarding how I feel about this fix, I feel kind of gross. I tried turning LANGUAGES and LOCALES into lazy things, but that creates other problems in other parts of the code that aren't easily fixed. For example `kuma/dashboard/forms.py` uses `settings.LANGUAGES` and that code triggers at module load time and isn't happy about lazy things.

Anyhow, yay.

robhudson:
When I remove the directory I get:
```
Warning: Cannot open /app/../product_details_json/languages.json because it does not exist. LANGUAGES and LOCALES will be empty. Please run "./manage.py update_product_details".
[Errno 2] No such file or directory: '/app/../product_details_json/languages.json'
```
When I run the management command all is good.

r+

PR #3783, opened 2016-02-16
Bug 1218563 - Add docs for Docker setup

robhudson:


docs/installation-docker.rst
```diff
@@ -0,0 +1,81 @@
...
+        cd kuma
+
+#. Install the `Docker Toolbox`_. This will install the Docker engine,
+   along with Docker Compose and Docker Machine.
+
+   .. _Docker Toolbox: https://www.docker.com/products/docker-toolbox
+
+#. Create the virtual machine that Docker will run within::
+
+        docker-machine create --driver virtualbox --virtualbox-memory 2048 kuma
```
jezdez: 2GB seems little to me given the amount of stuff we're running, could you add a note here how to increase it?

docs/installation-docker.rst
```diff
@@ -0,0 +1,81 @@
...
+
+#. Configure the environment variables::
+
+        eval "$(docker-machine env kuma)"
+        export COMPOSE_FILE=`pwd`/docker/docker-compose.yml
+        export COMPOSE_PROJECT_NAME=kuma  # String prepended to every container.
+
+#. Add the Docker VM to your hosts::
+
+        sudo sh -c "echo $(docker-machine ip kuma 2>/dev/null)  kuma.dev >> /etc/hosts"
```
jezdez: There is no need to use `sudo sh` for this. Docker has a `--add-host` option: https://docs.docker.com/engine/reference/run/#managing-etc-hosts Now to find out how to make this work in our workflow.

Also, `kuma.dev` doesn't match the SSL cert we're using, which we have to while we're on Persona at least and I'd make the argument that using SSL in development prevents us from making dev/prod parity mistakes. So let's keep using `developer-local.allizom.org` for it and make sure the embedded attachment files are correctly served like in stage/prod by additionally using `mdn-local.mozillademos.org`, too.

docs/installation-docker.rst
```diff
@@ -0,0 +1,81 @@
...
+create the database and tables.
+
+If you have a data dump you can connect to the "web" container and import
+the data dump::
+
+    mysql --default-character-set=utf8 -h mysql -ukuma -pkuma kuma < kuma.sql
+
+Or run the migrations::
+
+    docker exec -ti kuma_web_1 /bin/bash
```
jezdez: Can we add that tot he `Makefile` as a `make shell`?

docs/installation-docker.rst
```diff
@@ -0,0 +1,81 @@
...
+
+#. Create the virtual machine that Docker will run within::
+
+        docker-machine create --driver virtualbox --virtualbox-memory 2048 kuma
+
+#. Configure the environment variables::
+
+        eval "$(docker-machine env kuma)"
+        export COMPOSE_FILE=`pwd`/docker/docker-compose.yml
+        export COMPOSE_PROJECT_NAME=kuma  # String prepended to every container.
```
jezdez: Can we somehow hardcode that? Or add to a .env file so we don't have to run these commands all the time?

docs/installation-docker.rst
robhudson: I could bump the default up. I think Vagrant we use 4096 so I can use that. I don't think you can resize it after the fact.

docs/installation-docker.rst
robhudson: It looks like the `--add-host` option adds the host to the container, not the host machine, which is what I'm doing here.

I chose `kuma.dev` to not stomp on the hostname that Vagrant is using, at least the Docker set up is the preferred way to go.  Also, you've reminded me that I haven't set up SSL yet.

docs/installation-docker.rst
robhudson: Yeah I could tweak the Makefile in another pass. There are a few things we could add to make things easier.

docs/installation-docker.rst
robhudson: I can look into this yeah.

robhudson: Are you thinking something like `cdenv` (I think that used to be a thing)? These are standard docker-compose env variables. If you run different containers in different shells you don't want these to be global or hardcoded.

However, the `COMPOSE_FILE` defaults to looking in the CWD for a file named `docker-compose.yml`. I can move this file and this env variable is no longer required.

The `COMPOSE_PROJECT_NAME` also defaults to the CWD directory name which would probably be fine in most cases, so this isn't really required either.

With those 2 changes all we need is to set up the docker-machine env variables via `eval "$(docker-machine env kuma)"`.

robhudson:
Updated. I'd like to fix the other items pointed out in another PR. Namely... SSL and fixing the hostname.

PR #3784, opened 2016-02-17
bug 1201709 - /renamed$children returns error JSON

jwhitlock:
When the $children API view is requested for a redirected page, return this error JSON:

```json
{"error": "Document has moved."}
```

Previously, a 500 error occurred due to ``JsonResponse`` refusing to encode a ``null``.


jezdez:
r+

PR #3785, opened 2016-02-17
Fix bug 1237661: Fix wonky admin

willkg:
This fixes the wonky admin problem where the language of the admin would
change between refreshes.

What's going on is that Django 1.8 introduced a fix to
trans_real.activate where if you pass in '', it does nothing and
returns. The prefixer will set the locale to '' for /admin/, so then
the locale-aware middleware calls translation.activate('') and
suddenly we're all standing in a pool of goo in the middle ages with
flaming arrows aimed at us.

This fixes that.

r?

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -55,7 +55,10 @@ class LocaleURLMiddleware(object):
 
         request.path_info = '/' + prefixer.shortened_path
         request.LANGUAGE_CODE = prefixer.locale or settings.LANGUAGE_CODE
-        translation.activate(prefixer.locale)
+        # prefixer.locale can be '', but we need a real locale code to activate
+        # otherwise the request uses the previously handled request's
+        # translations.
+        translation.activate(prefixer.locale or settings.LANGUAGE_CODE)
 
     def process_response(self, request, response):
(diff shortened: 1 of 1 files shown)


willkg:
The issue is explained better here: https://bugzilla.mozilla.org/show_bug.cgi?id=1215125#c4

We can trivially test this on -stage by:

1. in a browser, log into -stage and go to the admin
2. in a different browser, go to https://developer.allizom.org/fr/
3. refresh the admin

That causes the admin to show up as French for me. We can do any language--it's not specific to French. This might not always always work depending on whether -stage is handling other requests. But -stage is quiet enough that it works most/all of the time for me.

We can test it locally, but you have to use gunicorn and can't use `./manage.py runserver`. I'm not entirely sure why, but suspect it has to do with the WSGI paths vs. the runserver machinery.

robhudson:
r+ :boom: :open_mouth: :tada: :balloon: :cake: :guitar: :beers:

willkg:
Thank you!

stephaniehobson:
:fireworks:

jezdez:
♪┏(・o･)┛♪┗ ( ･o･) ┓♪

PR #3786, opened 2016-02-20
Bug 1238933 - Check database sanitization

herinin-palominodb:
Add a script check_anonymize.sql that verifies that the data doesn't contain PII. This script needs to be started after anonymize.sql has been applied to the database.


jwhitlock:
r+ Works in practical test on my local VM. non zero responses returned on non-sanitized database, and zeros returned (quickly) on a recent sanitized database dump. :+1:

PR #3788, opened 2016-02-23
Fix bug 1239141 - Fix "Compare on MDN" link with correct revisions

robhudson:


kuma/wiki/models.py
```diff
@@ -1743,11 +1743,8 @@ def previous(self):
...
         """
         try:
-            return self.document.revisions.filter(
-                is_approved=True,
-                created__lt=self.created,
-            ).order_by('-created')[0]
-        except IndexError:
+            return self.get_previous_by_created(is_approved=True)
+        except self.DoesNotExist:
             return None
```
jezdez: I've tried that before in c715cac1bde384f3452633c97ede9725abf0a7a1 and it didn't work well, see https://bugzilla.mozilla.org/show_bug.cgi?id=1174820. Could you make sure that this doesn't happen again?

robhudson: Hmm, ok. I wonder if we can add a test that exposes this so if anyone tries this again the test fails?

jezdez: Yeah, I think the comments were wrong or something. fscholz may remember better what https://bugzilla.mozilla.org/show_bug.cgi?id=1174820 included

robhudson: Removed this change.

robhudson:
Since I don't know how to trigger this bug the changes here are a total guess. I would prefer to have proper STR to trigger the bug. The test I added passes both with and without the changes here. :disappointed:

jwhitlock:
I was able to trigger the bug in my VM on the master branch (edit a page, see matching revisions in /tmp/kuma-messages/<date>.log), and not when switching to this branch.

It does make me wonder what the mechanism is for ``.previous`` getting a cached value. I was able to confirm that it is only called once on page edits, so it must be some other code calling previous. Maybe ``RevisionsFeed``? But does celery cache models between tasks? It is all very bizarre, and makes me question if we want to use ``@cached_property`` at all.

But, works for me, so r+ :+1:

kuma/wiki/models.py
```diff
@@ -1739,6 +1739,9 @@ def content_cleaned(self):
 
     @cached_property
     def previous(self):
+        return self._previous()
+
+    def _previous(self):
```
jezdez: Meh, no private methods please. You can use this:
```python
previous = cached_property(get_previous, name='previous')
```
where `get_previous` would be the new name of the existing method

robhudson: I removed the private method but kept the regular syntax of the cached_property b/c it's more "scannable" to me.

PR #3789, opened 2016-02-23
Soapbox and other dep updates.

jezdez:


jezdez:
@robhudson r?

willkg:
I verified I could reproduce the problem in the soapbox bug in my development environment. Then I switched to this branch and updated requirements and can no longer reproduce the issue.

Looks good to me. r+

PR #3790, opened 2016-02-23
Bug 1248476 - Add management command to submit articles that were deleted due to spamming to Akismet for training.

jezdez:


jezdez:
@jwhitlock @willkg r?

jwhitlock:
Initial r+ on reading code, but it's going to be a day before I can do a practical test in my developer VM.

kuma/wiki/management/commands/submit_deleted_documents.py
```diff
@@ -0,0 +1,96 @@
...
+    help = "Seed Akismet spam with documents that were deleted because they were spam"
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('username', type=str)
+        parser.add_argument('-n', '--dry-run',
+                            action='store_true', dest='dry_run', default=False,
+                            help="Do everything except actually submit.")
+        parser.add_argument('-d', '--days',
+                            action='store_true', dest='days', default=365,
```
willkg: I think a better default should relate to how often we might run this. If we're thinking we might run this once a week via a cron job, then we'd want this to default to 7--not 365.

willkg: Also, I'm pretty sure the action here shouldn't be `'store_true'`. I think it should be `'store'` and maybe also have `type=int`? I'm pretty sure that's right, but I often get my argument parsing libraries all confused.

kuma/wiki/management/commands/submit_deleted_documents.py
```diff
@@ -0,0 +1,96 @@
...
+                            help="Number of days of deletion logs to inspect.")
+
+    def handle(self, *args, **options):
+        dry_run = options['dry_run']
+
+        ttl = timezone.now() - timedelta(days=options['days'])
+        # first get the last year's deleted document logs
```
willkg: Nit: This actually gets the last n days of document logs where n = `options['days']`.

kuma/wiki/management/commands/submit_deleted_documents.py
```diff
@@ -0,0 +1,96 @@
...
+        logged_deletions = DocumentDeletionLog.objects.filter(
+            timestamp__lte=ttl,
```
willkg: My brain really thinks this should be gte and not lte, but I screw up datemath all the time so I'm not confident in that. Seems like you want this:

```
TIME!
<- Past                                                          Future! ->
------X--------------------------------------------------------- now()
      |||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
      ^ ttl
```

Does that sound right or am I confused?

kuma/wiki/management/commands/submit_deleted_documents.py
```diff
@@ -0,0 +1,96 @@
...
+        ).filter(
+            models.Q(reason__icontains='spam') |
+            models.Q(reason__icontains='junk')
```
willkg: Might want a comment here saying something like "The writers always use the word 'spam' or 'junk' when deleting revisions because of spam. If that ever changes, we'll need to update this." Maybe do it as a haiku?

```
They use "spam" or "junk"
deleting spam revisions;
the spam makes me cry.
```

kuma/wiki/management/commands/submit_deleted_documents.py
```diff
@@ -0,0 +1,96 @@
...
+        sender = get_user_model().objects.get(username=options['username'])
+        self.stdout.write(u'Submitting spam to Akismet as user %s' % sender)
+
+        akismet = Akismet()
+
+        if not akismet.ready:
+            raise CommandError('Akismet client is not ready')
+
+        for i, logged_deletion in enumerate(logged_deletions.iterator(), 1):
+            self.stdout.write('%d/%d: ' % (i, count), ending='')
```
willkg: Seems silly to add a requirement just for this bug tqdm or something similar is pretty super. That's something I meant to redo with Input, but never got around to it.

kuma/wiki/management/commands/submit_deleted_documents.py
```diff
@@ -0,0 +1,96 @@
...
+                # no current revision found, which means something is fishy
+                # but we can't submit it as spam since we don't have a revision
+                self.stderr.write(u'Ignoring document %s without current '
+                                  u'revision' % document.pk)
+                continue
+
+            params = revision_akismet_parameters(document.current_revision)
+            try:
+                if not dry_run:
+                    akismet.submit_spam(**params)
```
willkg: You could probably do:

```
if dry_run:
    continue

try:
    akismet.submit_spam(**params)
...
```

and skip some of the if blocks here.

willkg:
I didn't run it or run the tests. I'm lacking Akismet configuration, so I'd have to set that up first.

I think there are some issues, but I'm not positive. Feel free to reply to my comments solely with emoji. :8ball:

kuma/wiki/management/commands/submit_deleted_documents.py
robhudson: I also think this should be `gte`.

jezdez: Bah, I always confuse those. Good catch, all.

kuma/wiki/management/commands/submit_deleted_documents.py
jezdez: Sure

kuma/wiki/management/commands/submit_deleted_documents.py
jezdez: Once the attachments PR is merged we can make this thing use the django-mysql smart iterator which has a built-in stdout output

kuma/wiki/management/commands/submit_deleted_documents.py
jezdez: Sure

kuma/wiki/management/commands/submit_deleted_documents.py
jezdez: right, leftover from old code

jezdez:
Ok, all requested changes done

willkg:
Looks good to me as far as I can tell without being able to run it. r+

This is something we're using, so pushing it to -prod, having it "make a mistake" and then pushing another version is probably fine.

Out of curiosity, if it does "make a mistake" and we have to undo a bunch of Akismet things--Is that hard to do?

jezdez:
If we really have to we can delete the site in Akismet's account UI and recreate it again to blow away the index.

willkg:
Sounds good to me. Let's do this!

jwhitlock:
BTW, some people use "Junk" to tag docs created by intern tests:

https://groups.google.com/forum/#!topic/mozilla.dev.mdc/JV_pMIx__aY

jezdez:
@jwhitlock ah, right but checking the deletion log seems ok too

jezdez:
oh and also we want to only train akismet on spam, not deleted pages that come from intern

jezdez:
I asked in #mdn for help from the writers

PR #3794, opened 2016-02-29
bug 1246967 - Add UUIDs to $json page API

jwhitlock:
Add uuid to document model, and expose it in the ``$json`` page API (like [Web/CSS/display$json](https://developer.mozilla.org/en-US/docs/Web/CSS/display$json)). Populate null Document.uuid columns with new [Version 4 (random) UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29).

This is part 2 of 3 of adding Document.UUID. Each part needs to be pushed to production, and the included migrations run, before the next portion is merged and pushed.

1. PR #3793 - Add empty Document.uuid to database
2. PR #3794 (this PR) - Populate Document.uuid and use in $json API
3. PR #3795 - Require non-null Document.uuid and refresh $json API

PR #3795, opened 2016-02-29
bug 1246967 - Add Document.uuid

jwhitlock:
Add a UUID to the Document model, which can first be used as an external ID for identifying features in the https://github.com/mdn/browsercompat project.

Because this PR includes database migrations, it needs to be manually merged and deployed in four parts:

1. [x] Update Kuma migrations to match the state of the models. Then add Document.uuid to the database, nullable with a default of null. *(first commit)*
2. [x] Add uuid to document model, and expose it in the ``$json`` page API (like [Web/CSS/display$json](https://developer.mozilla.org/en-US/docs/Web/CSS/display$json)). *(middle commits)*
3. [x] Manually populate null Document.uuid columns with new [Version 4 (random) UUIDs] (https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29). 
4. [ ] Run an automated migration to ensure UUIDs are populated. Update the schema so that Document.uuid is not allowed to be null. *(final 2 commits)*

### DIFF
--- a/kuma/wiki/migrations/0027_populate_document_uuid.py
+++ b/kuma/wiki/migrations/0027_populate_document_uuid.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from uuid import uuid4
+
+from django.db import migrations
+
+
+def populate_uuids(apps, schema_editor):
+    """Populate Document.uuid, without bumping last modified."""
+    Document = apps.get_model('wiki', 'Document')
+    docs = Document.objects.filter(uuid__isnull=True)
+    for document_id in docs.values_list('id', flat=True).iterator():
+        Document.objects.filter(id=document_id).update(uuid=uuid4())
+
+
+def clear_uuids(apps, schema_editor):
+    """Clear Document.uuid."""
+    Document = apps.get_model('wiki', 'Document')
+    Document.objects.exclude(uuid__isnull=True).update(uuid=None)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wiki', '0026_document_uuid_default'),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_uuids, clear_uuids)
+    ]
--- a/kuma/wiki/migrations/0028_require_document_uuid.py
+++ b/kuma/wiki/migrations/0028_require_document_uuid.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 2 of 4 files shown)


jwhitlock:
Force push to rebase on current master

jwhitlock:
Rebased for current master. Migrations take 193 seconds in my development VM, and shouldn't cause production downtime.

jezdez:
Clearing the JSON field for all documents will have a serious performance impact on the site's performance once this migration has been executed. I strongly suggest to split this into two parts:

- once #3794 is merged and deployed, run the population of the `uuid` field in a management command or even a Python shell for each wiki document individually
- remove the conditions in the $json view and require the `uuid` field but don't reset the `json` field

That way you reduce the real risk of bringing down the site due to the thundering herd problem

r+wc

kuma/wiki/migrations/0025_populate_document_uuid.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations
+
+
+def populate_uuids(apps, schema_editor):
+    """Populate Document.uuid, without bumping last modified."""
+    Document = apps.get_model('wiki', 'Document')
+    docs = Document.objects.filter(uuid__isnull=True)
+    for document_id in docs.values_list('id', flat=True).iterator():
+        Document.objects.filter(id=document_id).update(uuid=uuid4())
```
jwhitlock: Added ``.iterator()`` to reduce memory retained on admin node.

In reference to comment  https://github.com/mozilla/kuma/pull/3795#issuecomment-196392622 . Are you suggesting this migration should be removed from the "UUID step 2" PR #3794, and done manually in the shell instead?  If so, then it seems to make sense in this "UUID step 3" PR #3795, for developers etc. who didn't run the manual command.  If not, then I think it is in the right place.

jwhitlock:
The migrations that clears cached `$json` has been removed, instead manual updating will be used in production when needed. Waiting on PR #3816, which includes migrations in the wiki app.

jwhitlock:
Rebased. I'm assuming @jezdez also wants the UUIDs to be populated manually, so adding as another step (step 3).  I'm leaving the data migration, so that developers can just run that rather than manually updating their systems (step 4).

I'm saying that satisfies the "with comment", leaving just a +r from @jezdez, so I don't need to rebase this anymore.

jwhitlock:
Now that the manual steps are done, merging the normal way. I've manually add UUIDs in staging and production, so wiki migration 27 will be very fast. For dev boxes, 27 may be slow based on the size of your development database.

PR #3796, opened 2016-03-01
Bug 1247086 - Wrap exception when formatting date/time with missing format string.

jezdez:
This basically works around an issue with Babel's locale data that misses formatting
strings sometimes and raises a KeyError. We fall back to English that such cases.

https://bugzilla.mozilla.org/show_bug.cgi?id=1247086

### DIFF
--- a/kuma/core/tests/test_helpers.py
+++ b/kuma/core/tests/test_helpers.py
@@ -2,12 +2,12 @@
 from collections import namedtuple
 from datetime import datetime
 
+import mock
 import pytest
 import pytz
 from babel.dates import format_date, format_datetime, format_time
 from django.conf import settings
 from django.test import RequestFactory
-from pyquery import PyQuery as pq
 from soapbox.models import Message
 
 from kuma.core.tests import KumaTestCase, eq_, ok_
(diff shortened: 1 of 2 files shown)


jezdez:
@robhudson r?

kuma/core/tests/test_helpers.py
```diff
@@ -167,6 +168,21 @@ def test_unknown_format(self):
...
+        value_english = format_datetime(value_test, locale=u'en_US')
+        self.context['request'].LANGUAGE_CODE = u'fr'
+        mocked_format_datetime.side_effect = [
+            # first call is returning a KeyError as if the format is broken
+            KeyError,
+            # second call returns the English fallback version as expected
+            value_english,
+        ]
+        value_returned = datetimeformat(self.context, value_test,
+                                        format='datetime')
```
willkg: Why not have it output json and then you get a more "raw" value that you can check more easily?

robhudson:
I think Will's comment is good. r+wc.

kuma/core/tests/test_helpers.py
jezdez: because I don't care

jezdez: ok, taking care of this. fixing the other tests as well

PR #3797, opened 2016-03-01
Fix Bug 1109941 - Remove SVG fonts

stephaniehobson:
SVG used to be in there for Opera but has a bug in Blackberry. Opera has improved their WOFF support and no longer supports SVG, for the sake of the Blackberry users we should remove them.

Testing:
Opens Sans here: https://developer.mozilla.org/en-US/docs/User:stephaniehobson:fonts
Milk here: https://developer.mozilla.org/en-US/promote
MozTT is used in Gaia and included here: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox_OS/Firefox_OS_apps/Building_blocks/1.x/Button/Coding


willkg:
I made a thing to make these easier to test out:

http://people.mozilla.org/~wkahngreene/mdn/

I have no idea if something like that exists already, but I didn't see it anywhere.

willkg:
I compared the first two pages and they look identical. The third page isn't in the db dump from January and I can't figure out how to set it up right locally. I looked at everything with Firefox Dev Edition 46.0a2.

Travis is happy with this.

I think we should just land it and if there are issues we can figure those out in a future bug.

stephaniehobson:
I have one button each for switching to prod/stage/dev. I did not, however, go to the trouble of sharing my code >_<

stephaniehobson:
+1 to landing and sorting later. If I haven't left a comma out of place there should not be any problems in any of the browsers we have access to test in anyway.

willkg:
Merging!

PR #3798, opened 2016-03-01
Bug 994430 - Improve directory tree style

stephaniehobson:
Moved from image solution to CSS only solution and added some left padding.

Testing:
https://developer.mozilla.org/en-US/docs/User:wbamberg/directory-tree (and in RTL)


kuma/static/styles/components/content.styl
```diff
@@ -154,24 +154,46 @@ $table-blue = #d4dde4;
...
-        }
-    }
+            position: relative;
+
+            &:before,
+            &:after {
+                content: '';
+                position: absolute;
+                bidi-style(left, -15px, right, auto);
+                display: block;
```
willkg: Is there a style rule in MDN for having CSS properties be alphabetical?

willkg:
Other than that question, I think this looks great! r+

stephaniehobson:
Alphabetical properties are the only part of the mozorg coding guidelines we don't follow.

willkg:
Good to know. Merging!

PR #3799, opened 2016-03-02
Bug 1239757: Only ask helpful on wiki docs

stephaniehobson:
Separated the helpfulness file back out of the wiki.js file and added it only to the document pages so we won't get asked on editing pages, etc anymore.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -842,7 +842,6 @@ PIPELINE_JS = {
             'js/wiki.js',
             'js/wiki-samples.js',
             'js/social.js',
-            'js/helpfulness.js',
         ),
         'output_filename': 'build/js/wiki.js',
         'extra_context': {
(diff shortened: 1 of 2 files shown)


willkg:
I'm not sure how to verify this. Is it sufficient to see if helpfulness doesn't show up on edit pages? Pretend I don't know MDN very well... how many other kinds of non-document document pages are there?

stephaniehobson:
The biggest problem was that they were showing up on $edit and $move pages where they're super out of place. Other non-document pages include: revision dashboard, tag listings, search results, user profiles, probably more.

Rather than waiting for the helpfulness widget to pop open at 60 seconds you can watch the network panel for helpfulness.js to be loaded. It shouldn't be loaded on non document pages anymore.

On document pages if you're not seeing the pop up after 60 seconds you probably need to delete the associated local storage entries.

willkg:
Ok--that helps a lot. I verified that `helpfulness.js` doesn't load on `$edit` and `$history` pages, but does on the document pages. Looks good to me. r+!

PR #3800, opened 2016-03-02
Bug 1243741: Expanded contributor bar

stephaniehobson:
This is a quick fix that gets us away from "that looks broken" but could use more finesse.

Page with lots of users for testing:
https://developer-local.allizom.org/en-US/docs/Mozilla/Projects/Midas/Security_preferences
Page with fewer users:
https://developer.mozilla.org/en-US/docs/Glossary
Page with very few users:
https://developer-local.allizom.org/en-US/docs/Glossary/I18N
Page with fewer than minimum for button:
https://developer-local.allizom.org/en-US/docs/Glossary/ARIA

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -605,6 +605,7 @@
         function toggleImages() {
             var $hiddenContributors;
             $hiddenContributors = $contributors.find('li.' + hiddenClass);
+            $contributors.toggleClass('contributor-avatars-open');
             if($hiddenContributors.length) {
                 mdn.analytics.trackEvent({
                     category: 'Top Contributors',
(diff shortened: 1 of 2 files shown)


willkg:
I'm looking at this now. I restarted the test that failed in travis--kind of looked like a setup issue.

willkg:
In my local environment, when Firefox requests the gravatars, it gets back an HTTP 302 and then requests that url but that returns an HTTP 504. Thus I end up with no gravatars and it's hard to see what changed. Is there any way to fix that? Alternatively, can you attach screenshots to this PR?

willkg:
Scratch that--I see grey boxes now. I have no idea why I didn't see those before. Weirdness.

I think this looks great! r+

willkg:
Travis is happy--merging!

PR #3801, opened 2016-03-02
Fix bug 1252697 - Add CAPTCHA to user registration flow

robhudson:
This adds a CAPTCHA to the user registration flow below the "Choose a username" section:
screenshot 2016-03-02 09 55 26

### DIFF
--- a/kuma/users/signup.py
+++ b/kuma/users/signup.py
@@ -1,4 +1,6 @@
 from allauth.socialaccount.forms import SignupForm as BaseSignupForm
+from captcha.fields import ReCaptchaField
+from constance import config
 from django import forms
 from django.core import validators
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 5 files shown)


kuma/users/tests/test_views.py
```diff
@@ -586,6 +612,9 @@ def test_persona_signup_create_django_user(self):
         persona_signup_email = 'views_persona_django_user@example.com'
         persona_signup_username = 'views_persona_django_user'
 
+        # Put recaptcha into testing mode.
```
jezdez: You can use the test decorator `@mock.patch.dict(os.environ, {'RECAPTCHA_TESTING': 'true'})` for this

kuma/users/signup.py
```diff
@@ -43,6 +46,9 @@ def __init__(self, *args, **kwargs):
             'min_length': USERNAME_SHORT,
             'max_length': USERNAME_LONG,
         }
+        # Add recaptcha if configured in settings.
+        if getattr(settings, 'RECAPTCHA_PRIVATE_KEY', None):
```
jezdez: It would be better to make this a constance config to be able to deploy this without webops help similarly how we did with the Akismet API key

jezdez:
nicely done r+wc

robhudson:
Updated after feedback.

PR #3803, opened 2016-03-03
Fix bug 1227193 - show actual attachment file size instead of '{} bytes'

Krenair:


jezdez:
This is a duplicate of #3769. Thanks for the contribution though

PR #3804, opened 2016-03-03
Fix bug 1253143 - don't try to call preventDefault on non-events

Krenair:


stephaniehobson:
I'll review this on Monday if it doesn't get reviewed today. Sorry for the delay, thanks for identifying and fixing this problem @Krenair :)

stephaniehobson:
:+1:  Thank you for finding and fixing this!

PR #3805, opened 2016-03-03
Fix bug 1253269: Add fixaccount command

willkg:
This creates a django management command to fix MDN user accounts in
an anonymized database so you can log in with them in your dev
environment.

This only works with Persona accounts, but I'm pretty sure that's ok
since there's all this extra work to set up GitHub auth in a dev
environment. I think if that's important, someone can create a bug for
it and do it in a future PR.

To test, load an anonymized db, then run:

    $ ./manage.py fixaccount <username> <email>

r?

### DIFF
--- a/kuma/core/management/commands/fixaccount.py
+++ b/kuma/core/management/commands/fixaccount.py
@@ -32,7 +32,16 @@ class Command(BaseCommand):
             user.save()
 
         # Fix the django-allauth EmailAddress record
-        emailaddress = EmailAddress.objects.get(user=user)
+        try:
+            emailaddress = EmailAddress.objects.get(user=user)
+        except EmailAddress.DoesNotExist:
+            raise CommandError(
+                'There is no account_emailaddress record for this account '
+                'which suggests it was created with createsuperuser. Please '
+                'log into the account with Persona and go through the signup '
+                'process.'
+            )
+
         if emailaddress.email != email:
             self.stdout.write('Fixing email address in account_emailaddress '
                               'record.')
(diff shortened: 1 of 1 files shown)


kuma/core/management/commands/fixaccount.py
```diff
@@ -0,0 +1,50 @@
...
+        except User.DoesNotExist:
+            raise CommandError('User %s does not exist.' % username)
+
+        if user.email != email:
+            self.stdout.write('Fixing email address in auth_user record.')
+            user.email = email
+            user.save()
+
+        # Fix the django-allauth EmailAddress record
+        emailaddress = EmailAddress.objects.get(user=user)
```
jezdez: Assuming someone used the `createsuperuser` command, there may not be an email address. Should create one?

kuma/core/management/commands/fixaccount.py
```diff
@@ -0,0 +1,50 @@
...
+                              'record.')
+            emailaddress.change(None, email, confirm=False)
+            emailaddress.verified = True
+            emailaddress.save()
+
+        try:
+            sa = SocialAccount.objects.get(user=user)
+        except SocialAccount.DoesNotExist:
+            self.stdout.write('Creating a socialaccount_socialaccount record.')
+            sa = SocialAccount.objects.create(user=user, provider='Persona', uid=email)
```
jezdez: If you've used `create` you don't need to call `save` afterwards

jezdez:
r+wc

kuma/core/management/commands/fixaccount.py
willkg: I think if they used `createsuperuser`, then we shouldn't be doing anything in this command since that's outside the scope. The code should handle that case and provide appropriate instructions. I'll fix that. Good catch!

kuma/core/management/commands/fixaccount.py
willkg: Oh, right--my bad.

willkg:
^^^ Fixes the issues brought up in comments. I also waited for Travis to go green. It had problems with the tests, but it was a setup issue and rerunning fixed it.

willkg:
Also, I'm planning to document use of this in the internal docs. That hasn't been done, yet.

willkg:
I documented it. Even though it hasn't landed yet. So I'm landing it now!

PR #3806, opened 2016-03-03
Fix Build Path In Docker Compose File

bensternthal:
build path was missing ./docker resulting in error


bensternthal:
eh, there are other issues in the docker-compose file causing this not to work. I'll re-submit when i get something running.

PR #3808, opened 2016-03-04
Bug 1077172 - Improve managment command to fix historical attachments.

jezdez:
This tries to guess better which document an attachment was originally uploaded to.
If an attachment is included in multiple documents it:
- favors documents with smaller primary keys (denoting earlier creation)
- favors English documents as the orignal upload

### DIFF
--- a/kuma/attachments/jinja2/attachments/includes/attachment_list.html
+++ b/kuma/attachments/jinja2/attachments/includes/attachment_list.html
@@ -105,6 +105,8 @@
               <a href="{{ document_attachment.document.get_absolute_url() }}">
                 {{ document_attachment.document.title }} ({{ document_attachment.document.locale}})
               </a>{% if not loop.last %}, {% endif %}
+            {% else %}
+              {{ _('n/a') }}
             {% endfor %}
           {% endif %}
           </td>
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -318,8 +318,6 @@ class Document(NotificationsMixin, models.Model):
 
     summary_text = models.TextField(editable=False, blank=True, null=True)
 
-    attachments_populated = models.BooleanField(default=False)
-
     class Meta(object):
         unique_together = (
             ('parent', 'locale'),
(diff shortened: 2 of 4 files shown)


jezdez:
@robhudson @jwhitlock r?

kuma/wiki/models.py
```diff
@@ -1327,15 +1325,11 @@ def populate_attachments(self, update_populated_field=False):
                 defaults={
                     'attached_by': revision.creator,
                     'name': revision.filename,
+                    'is_original': False,
```
jwhitlock: Should ``is_original`` be true if this is a new attachment? This code seems to clear that flag for all attachments.

jezdez: You're right, this shouldn't be here. I confused the semantics of `get_or_create` with that of `update_or_create`

kuma/attachments/jinja2/attachments/includes/attachment_list.html
```diff
@@ -105,6 +105,8 @@
               <a href="{{ document_attachment.document.get_absolute_url() }}">
                 {{ document_attachment.document.title }} ({{ document_attachment.document.locale}})
               </a>{% if not loop.last %}, {% endif %}
+            {% else %}
+              {{ _('n/a') }}
```
jwhitlock: I wasn't able to test this branch, because I can't find any examples in my database. As far as I can tell, the case to trigger it would be something like:

1) Create the sole translation of an English document, with an attachment
2) Delete the English version of the document
3) Edit the translated version

So, saying "looks good" without practical testing.

jwhitlock:
Ran on VM before and after running the migration. The command ran successfully both times.  It's a long migration, lots of documents.

Is this a one-time use management command? If we're going to run it periodically, I'd eventually want it to notice what is new, what changed, and what was the same.  If it's a one-time initialization script, then it is fine.

+r on the code

jezdez:
@jwhitlock yeah, this is a one time thing that I didn't want to write as a data migrations since it would take too long to run during deployment and the system as it is right now won't stop working just because the data is missing

kuma/attachments/jinja2/attachments/includes/attachment_list.html
jezdez: Yeah, exactly this use case, or for those handful of attachments the script wasn't able to find any documents in question (either because the documents were deleted, or some other data inconsistency). So this is both for historical attachments as well as new attachments in edge cases

jwhitlock:
+r on PR, merge away when ready

PR #3810, opened 2016-03-08
bug 1187967 - Extract code samples via cacheback

jwhitlock:
This is a rebase of PR #3423, with an additional commit to address review comments.

Pages using ``{{EmbedLiveSample}}`` load sections from other pages that include HTML, CSS, and JS in an iframe, so it is loaded on every page load.  For example, https://developer.mozilla.org/en-US/docs/MDN/Contribute/Howto/Create_an_interactive_exercise_to_help_learning_the_web has an iframe with src=https://mdn.mozillademos.org/en-US/docs/MDN/Contribute/Howto/Create_an_interactive_exercise_to_help_learning_the_web/distant_example$samples/The_example?revision=955071  This request to the ``$sample`` API uses uncached database reads to construct an HTML page based on a section in the document.  This work may be mitigated by caching in the load balancer.

This PR moves section extraction from functions to a class, eliminating some repeated work. It also adds a caching layer, so that work is not repeated. The cache is invalidated when the document holding the code sample is updated. It also:
* Cleans up code style
* Uses ``revision(is_approved=True, save=True, content=content)`` instead of ``doc_rev(content)``
* Avoids single-character variable names
* Move ``@newrelic.agent.function_trace`` decorator to the top, so that decorator impact will also be measured.


stephaniehobson:
:heart_eyes:

jezdez:
It's not a good idea to review my own code, but I can at least r+ what @jwhitlock did fix about it. @robhudson do you have a bit of time still to review this?

kuma/core/jobs.py
```diff
@@ -10,9 +11,63 @@ class KumaJob(Job):
...
+    def __init__(self, age, for_class, generation_args, *args, **kwargs):
+        """Initialize but do not create the generation."""
+        super(GenerationKeyJob, self).__init__(*args, **kwargs)
+        self.age = age
+        self.for_class = for_class
+        self.generation_args = generation_args
+
+    def key(self, *args, **kwargs):
+        """Return a key that is derived only from the initial args."""
+        generation_args = ':'.join([str(key) for key in self.generation_args])
```
robhudson: Is it possible that the `generation_args` contains unicode that will trip up the `str`?

kuma/core/tests/test_jobs.py
```diff
@@ -1,18 +1,41 @@
 from django.utils import six
 from kuma.core.tests import KumaTestCase
 
-from ..jobs import KumaJob
+from ..jobs import KumaJob, GenerationJob
```
robhudson: alphabetical imports are nice (nitpick)

kuma/core/jobs.py
robhudson: Later I noticed you're sending primary keys, so the answer would be no.

kuma/core/tests/test_jobs.py
jwhitlock: nit picked in force-push

kuma/core/jobs.py
jwhitlock: yeah, there are things that could go wrong, but for now the only user is sending a list of one integer.

jwhitlock:
Force-push to rebase on master and fix an import that wasn't alpha-sorted. Because nothing beats alphabetical order.

kuma/wiki/content.py
```diff
@@ -150,98 +261,6 @@ def filter_out_noinclude(src):
...
-        for token in parse(content).stream:
-            if token['type'] in ('Characters', 'SpaceCharacters'):
-                txt.append(token['data'])
-        txt = ''.join(txt)
-        names.update(MACRO_RE.findall(txt))
-    except:
-        pass
-    return list(names)
-
-
```
robhudson: I was trying to understand all the new code in `Extractor` then saw that it's mostly just moving it to a class. So I'm going to ignore it. :)

kuma/wiki/models.py
```diff
@@ -1501,9 +1479,6 @@ def parent_trees_watched_by(self, user):
         """
         return [doc for doc in self.parents if doc.tree_is_watched_by(user)]
 
-    def get_document_type(self):
-        return WikiDocumentType
```
robhudson: Hmm, guess that's not used anymore? Might have been leftovers from elasticutils.

kuma/wiki/tests/__init__.py
```diff
@@ -105,15 +105,6 @@ def wait_add_rev(document):
...
 
 
-# I don't like this thing. revision() is more flexible. All this adds is
-# is_approved=True, but it doesn't even mention approval in its name.
-# TODO: Remove.
-def doc_rev(content=''):
-    """Save a document and an approved revision with the given content."""
-    r = revision(content=content, is_approved=True)
-    r.save()
-    return r.document, r
```
robhudson: omg yes! :bomb:

robhudson:
Apologies for the delay. Just a few comments otherwise r+.

jwhitlock:
@jezdez started this back in August 18, 2015 - I think he should merge.

jwhitlock:
Good work @jezdez 

code_sample

stephaniehobson:
😍

PR #3811, opened 2016-03-09
Remove openjck from cc: list of account help bugs

openjck:


kuma/users/jinja2/socialaccount/signup.html
```diff
@@ -79,7 +79,7 @@
               {% endif %}
               </p>
 
-              <p>{% trans bug_href='https://bugzilla.mozilla.org/enter_bug.cgi?assigned_to=nobody%40mozilla.org&bug_file_loc=http%3A%2F%2F&bug_ignored=0&bug_severity=normal&bug_status=NEW&cc=jkarahalis%40mozilla.com&cf_fx_iteration=---&cf_fx_points=---&component=Account%20Help&contenttypemethod=autodetect&contenttypeselection=text%2Fplain&defined_groups=1&flag_type-4=X&flag_type-607=X&flag_type-791=X&flag_type-800=X&flag_type-803=X&form_name=enter_bug&maketemplate=Remember%20values%20as%20bookmarkable%20template&op_sys=All&priority=--&product=Mozilla%20Developer%20Network&rep_platform=All&status_whiteboard=%5Btrouble-logging-in%5D&target_milestone=---&version=unspecified&format=__standard__' %}
+              <p>{% trans bug_href='https://bugzilla.mozilla.org/enter_bug.cgi?product=Mozilla%20Developer%20Network&component=Account%20Help&format=__standard__' %}
```
jwhitlock: This URL has some very different options. Did you test this? If you don't have time, can you reduce the change to just removing you from CC?

openjck: Yeah, it was tested. Most of the options aren't needed, but the *Remember bug template* button generates a URL that specifies each field with a URL parameter, even if that field should use its default value. The last time we updated this, we just copied the URL generated by the *Remember bug template* button, which included parameters we didn't really need.

On second look, though, there is one field we might want to keep: the whiteboard. I'll update the URL to keep that.

kuma/users/jinja2/socialaccount/signup.html
```diff
@@ -79,7 +79,7 @@
               {% endif %}
               </p>
 
-              <p>{% trans bug_href='https://bugzilla.mozilla.org/enter_bug.cgi?assigned_to=nobody%40mozilla.org&bug_file_loc=http%3A%2F%2F&bug_ignored=0&bug_severity=normal&bug_status=NEW&cc=jkarahalis%40mozilla.com&cf_fx_iteration=---&cf_fx_points=---&component=Account%20Help&contenttypemethod=autodetect&contenttypeselection=text%2Fplain&defined_groups=1&flag_type-4=X&flag_type-607=X&flag_type-791=X&flag_type-800=X&flag_type-803=X&form_name=enter_bug&maketemplate=Remember%20values%20as%20bookmarkable%20template&op_sys=All&priority=--&product=Mozilla%20Developer%20Network&rep_platform=All&status_whiteboard=%5Btrouble-logging-in%5D&target_milestone=---&version=unspecified&format=__standard__' %}
+              <p>{% trans bug_href='https://bugzilla.mozilla.org/enter_bug.cgi?format=__standard__&component=Account%20Help&product=Mozilla%20Developer%20Network&status_whiteboard=%5Btrouble-logging-in%5D' %}
```
openjck: Updated. Besides the change to the *cc:* field, the only difference between the two forms is the *Platform* field. The new Bugzilla default is *Unspecified*, which I think makes more sense than *All*. Most login issues do affect all platforms, but we probably shouldn't assume.

jezdez:
r+

PR #3812, opened 2016-03-10
lithuanian translation added

zygimantus:


jwhitlock:
Thanks for the PR!

We'll need approval from the product manager to add a new language. Can you [request a new feature](https://bugzilla.mozilla.org/enter_bug.cgi?product=Mozilla%20Developer%20Network&component=Localization) in Bugzilla to start that process?  You can use https://bugzil.la/1002282 (which adds Serbian) as a guide for what to include.

jezdez:
Adding to what @jwhitlock said, we're using [Pontoon](https://pontoon.mozilla.org/) to maintain languages, so this PR isn't the right way to submit a new one. As developers we can't review the changes to translations, but Pontoon offers a community of translators able to see if the translation are correct. Once the language has been added you can upload the file on Pontoon to kickstart the language from the translation community side.

zygimantus:
So the Pontoon is the way to go with translations?

jwhitlock:
The process is something like:

1. Request the language as a new feature [Request the language as a new feature](https://bugzilla.mozilla.org/enter_bug.cgi?
2. After some discussion on the bug, the product manager (hopefully) approves it
3. We add a new blank translation for Lithuanian
4. It becomes available on [Pontoon](https://pontoon.mozilla.org/projects/mdn/)
5. You make your additions there, and they are merged back into the project by Pontoon

It's not a certainty that a new language will be allowed.  You can point to the work of the [Lithuanian translation community](https://pontoon.mozilla.org/lt/) (which seems to be mostly you) to show that there is enough support to maintain the translation.

PR #3814, opened 2016-03-11
Fix bug 1246459 - Prioritize web by default

robhudson:
This adds a default boolean field to the search filter model. Upon deploy no default will be set and search will be as it is now. Once we are ready to add defaults we can do so in the admin and enable them.

### DIFF
--- a/kuma/search/managers.py
+++ b/kuma/search/managers.py
@@ -1,6 +1,5 @@
 from django.conf import settings
 from django.db import models
-
 from elasticsearch.exceptions import RequestError
 
 from kuma.wiki.search import WikiDocumentType
@@ -11,6 +10,19 @@ class FilterManager(models.Manager):
 
     def visible_only(self):
         return self.filter(visible=True)
+
+    def default_filters(self):
+        """
+        Return default filters as a list of lists of the form::
+
+            [[<group_slug>, <filter_slug>], ...]
+
+        Converting to lists of lists so we can json encode it.
+
+        """
+        return [list(f) for f in
+                self.filter(default=True).values_list('group__slug', 'slug',
+                                                      'shortcut')]
 
 
 class IndexManager(models.Manager):
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1209,21 +1209,6 @@ CONSTANCE_CONFIG = dict(
         ]),
         "JSON array listing tag suggestions for documents"
     ),
-    SEARCH_FILTER_TAG_OPTIONS=(
-        json.dumps([
-            "Accessibility", "AJAX", "API", "Apps",
-            "Canvas", "CSS", "Device", "DOM", "Events",
-            "Extensions", "Firefox", "Firefox OS", "Games",
-            "Gecko", "Graphics", "Internationalization", "History", "HTML", "HTTP", "JavaScript", "Layout",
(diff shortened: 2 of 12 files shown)


kuma/search/managers.py
```diff
@@ -12,6 +12,10 @@ class FilterManager(models.Manager):
     def visible_only(self):
         return self.filter(visible=True)
 
+    @lru_cache(maxsize=None)
```
jezdez: If using the LRU cache is wanted without maximum size I'd recommend using the `cached_property` decorator instead since it reduces the amount of logic to process per call

kuma/settings/common.py
```diff
@@ -1238,21 +1238,6 @@ def _get_languages_and_locales():
...
-            "Gecko", "Graphics", "Internationalization", "History", "HTML", "HTTP", "JavaScript", "Layout",
-            "Localization", "MDN", "Mobile", "Mozilla",
-            "Networking", "Persona", "Places", "Plugins", "Protocols",
-
-            "Reference", "Tutorial", "Landing",
-
-            "junk", "NeedsMarkupWork", "NeedsContent", "NeedsExample",
-        ]),
-        "JSON array of tags that are enabled for search faceting"
-    ),
```
jezdez: :boom:

jezdez:
looks like a good strategy to go IMO

kuma/search/managers.py
robhudson: I started there but I didn't like the `property`ness of a manager call. I didn't look into the `maxsize` thing much but did see we are using `lru_cache` in other places. I did want this cached since it'll likely rarely change and it's a small set of data. Something like `[('topic', 'css'), ('topic', 'html'), ('topic', 'js')]`.

kuma/settings/common.py
robhudson: Discovered this isn't used anywhere so I removed it.

kuma/search/managers.py
jezdez: Makes sense, go for it

kuma/settings/common.py
jezdez: Yep, totally a leftover from the first iterations of the search feature

kuma/search/managers.py
robhudson: The "size" in `maxsize` is a bit of a misnomer. It's actually number of executions before it refreshes the cache. I'll update it to use something like 1024 or something.

kuma/static/js/search-suggestions.js
```diff
@@ -210,7 +210,13 @@
                         }
                     });
                 });
-                $searchInput.focus();
```
jwhitlock: Should ``$searchInput.focus()`` be dropped?

jwhitlock:
+r with one comment.  I didn't test in my dev environment - I have a fresh database with no content.

Here's the post-merge tasks (feel free to edit and add more):

- [ ] Add default search filters in staging admin
- [ ] Test default search in staging
- [ ] Test filtered search in staging
- [ ] Add default search filters in production admin

kuma/static/js/search-suggestions.js
robhudson: I intentionally left it out since focusing the search box is a change from the current behavior.

PR #3815, opened 2016-03-14
Bug 957802 - Stop some Django deprecation warnings to show up in the console.

jezdez:


jwhitlock:
Current work in #3950

PR #3817, opened 2016-03-18
Fix bug 1257622 - Refactor search-suggestions.js

robhudson:
This removes the unicode characters in place of their escape sequences
as the unicode was affecting some editors. This also fixes all jshint
errors and applies general js syntax reformatting.


stephaniehobson:
R+

PR #3818, opened 2016-03-21
Fix Bug 1258493: Account creation disabled notice

stephaniehobson:
- add fixed width to login widget to prevent notice from pushing it too wide
- add notice to appear when registration_disabled waffle enabled
- switch to always using "sign in" label for sign in (simplifies code and makes label shorter to try to keep localization from running out of space in the widget)

Testing:
- home page and a wiki page (with a grey header)
- multiple locales
- RTL
- tablet and mobile views
- can still login
- can still 

screen shot 2016-03-21 at 14 05 13

### DIFF
--- a/jinja2/includes/login.html
+++ b/jinja2/includes/login.html
@@ -14,17 +14,19 @@
       <div class="oauth-login-container">
           <div class="oauth-login-options" tabIndex="0">
               <span class="oauth-login-options-text">
-                <span class="oauth-login-options-text-icons">{{ _('Sign in with') }}</span>
-                <span class="oauth-login-options-text-no-icons">{{ _('Sign in') }}</span>
+                {{ _('Sign in') }}
               </span>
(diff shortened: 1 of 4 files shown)


jwhitlock:
I first thought that there was a desktop Safari bug, because the icons in the "Sign In" menubar disappeared, and were there in Firefox and Chrome:

no_icons

Then I widened the window a little and realized I was triggering tablet mode.

The code looks good to my Python-trained eyes, so I focused on manual testing:
* [x] Home page, content page in English, wide and narrow
* [x] Home page, content page in English with registration disabled, wide and narrow
* [x] Home page, content page in French (fr), wide and narrow
* [x] Home page, content page in French (fr) with registration disabled, wide and narrow
* [x] Home page, content page in Farsi (fa, right-to-left language w/o translation), wide and narrow
* [x] Home page, content page in Farsi (fa, right-to-left language w/o translation) with registration disabled, wide and narrow
* [x] Sign in with Persona

The period in the appears to be in the wrong place in Farsi, but this is consistent with other non-translated strings:

farsi period

:+1: to merge and push to stage, where I'd like to test:
* [ ] View with registration disabled on an iPad
* [ ] View with registration disabled on an iPhone
* [ ] Sign in with GitHub

PR #3820, opened 2016-03-22
Bug 1258624 - Upgrade CKEditor to v4.5.7

askeing:


lonnen:
Thanks for the contribution askeing! Someone will run this locally, verify it, and come back with next steps shortly.

jwhitlock:
I re-ran build.sh and saw some different documents, that had explainable differences:

* ``kuma/static/js/libs/ckeditor/build/ckeditor/ckeditor.js`` - the timestamp changed
* ``kuma/static/js/libs/ckeditor/build/ckeditor/lang/fa.js`` - changes in translations
* ``kuma/static/js/libs/ckeditor/build/ckeditor/lang/km.js`` - changes in translations
* ``kuma/static/js/libs/ckeditor/build/ckeditor/samples/toolbarconfigurator/lib/codemirror/codemirror.js`` - Optimizer picked ``0.0010`` vs ``0.001``.

I'm sad the build isn't completely reproducible, but that's an existing problem.  Also, our [docs](https://kuma.readthedocs.org/en/latest/ckeditor.html) are out of date, but that's follow-on work.

Thanks for the contribution!

PR #3821, opened 2016-03-23
Fix bug 1257003 - Send correct list of locales for translated content

robhudson:


kuma/spam/akismet.py
```diff
@@ -25,6 +26,13 @@ def __str__(self):
...
 
 
+def akismet_locale(language):
+    """
+    Convert a Django locale to a locale akismet expects. ("en-US" to "en_us")
+    """
+    return translation.to_locale(language).lower()
+
+
 class Akismet(object):
```
jwhitlock: This seems like it should be a method on a class. ``AkismetFormMixin`` or ``Akismet`` classes would work, but wouldn't be a natural fit since it isn't used there.  ``RevisionForm`` is the place it is actually used, so that feels like the best fit, but it does feel like generic code.

``RevisionForm`` is my preferred place, but I think I have a weak argument, so I'd support keeping it here for this round.

kuma/wiki/forms.py
```diff
@@ -466,15 +467,25 @@ def akismet_parameters(self):
         for a model instance in
         ``RevisionAkismetSubmissionAdminForm.akismet_parameters`` method!
         """
-        language = self.cleaned_data.get('locale',
-                                         settings.WIKI_DEFAULT_LANGUAGE)
+        if self.instance and self.instance.pk and self.instance.document:
+            language = self.instance.document.locale
+        else:
+            language = settings.WIKI_DEFAULT_LANGUAGE
+
```
jwhitlock: What happens when a user starts a new translation of an English page?

jwhitlock:
The code looks good, and fixes a bug in non-English languages. I'd like a quick discussion on the noted lines, which may not require changes to the PR.

kuma/spam/akismet.py
robhudson: I think `AkismetFormMixin` might be a good place. The `RevisionForm` is a subclass of that where this is eventually used. Also, this form is where `akismet_parameters` is defined where this is used.

kuma/wiki/forms.py
robhudson: I'm looking into this deeper.

robhudson:
I found that there's an akismet submission in the admin UI that is linked from the revision dashboard that uses a separate form that currently doesn't allow for multiple locales. The interesting one with this one is it's not bound to the form but a separate function, which is a bit odd, and doesn't work with my current changes which puts the `akismet_locale` on the `AkismetFormMixin`...
https://github.com/mozilla/kuma/blob/master/kuma/wiki/forms.py#L578

kuma/wiki/forms.py
robhudson: This code isn't going to work in the translate view from what I'm reading. The reason being that the view doesn't pass the instance in to the form, which means there is no document to pluck the locale from. When creating a new translation the locale comes from the `tolocale` query param which gets passed into the form's POST data.

I'd propose that we do multiple checks here with the settings as the final fallback. For example: get locale from the instance provided, if no instance get locale from the form data, (maybe another check?), fallback.

jwhitlock: I'd rather have an "akismet_locale" method that is customized for each form class that does the right thing for that form's data. The priority is getting the language right for our translators, which is currently broken, so pick a solution based on what gets that fixed early next week, not what gets an A in the code design class.  I'll be revisiting this code several times in the next few weeks, and hopefully the refactors will be obvious by then.

jwhitlock:
I think the history of that code is that:

1) The first way to submit ham and spam was in the Django admin,
2) Then it got added to the [revision dashboard](kuma/wiki/management/commands/submit_deleted_documents.py|) as SPAM / HAM buttons, which open the Django admin to create the records
3) Then a management command [submit_deleted_documents.py](https://github.com/mozilla/kuma/blob/master/kuma/wiki/management/commands/submit_deleted_documents.py) was written to allow bulk submission of past spam, and needed access to ``revision_akismet_parameters``. I don't think we intend to run this again *(todo - remove this command if true)*.

I think we'll have to duplicate code in ``revision_akismet_parameters`` to make SPAM/HAM in the revision dashboard do the right thing, but we can also assume access to a document with a locale set.  In the future, I think we'll save the Akismet parameters in the database for a month, so that SPAM/HAM will send the identical parameters rather than reconstruct them. Future work, though,

kuma/wiki/forms.py
robhudson: The `akismet_locale` won't change. It is simply a helper to convert Django style locale codes to what akismet expects ("en-US" -> "en_us"). What should be unique per form is `akismet_parameters` and this is already set up and intended.

But... `RevisionForm` is used in many places and the data provided to it is different depending on where it is used (editing a document vs starting a new translation). The `akismet_parameters` will need to work in those cases. Or we need different forms. I'm going to follow the former and make `akismet_parameters` try to find the locale from multiple sources of data.

robhudson:
@jwhitlock assigning back to you. Let me know what else you think this needs.

jwhitlock:
This didn't work with practical testing in my dev box. Adding spam to a Farsi document (https://developer-local.allizom.org/fa/docs/Firefox) that I created for testing reported the language as "en_US".  My dev box is setup for Akismet:

* Signed up for [Akismet API key](https://akismet.com/plans/) (slide the "value" down to $0 for free plan).
* Added Akismet API key in [constance](https://developer-local.allizom.org/admin/constance/config/) as ``AKISMET_KEY``
* Ensure that ``wiki_spam_exempted`` waffle flag is NOT on for my test user
* Set my test user name to ``viagra_test_123`` to ensure Akismet [always judges content as spam](https://akismet.com/development/api/#comment-check)
* Edit the translated document, get blocked as spam.
* View the [DocumentSpamAttempt](https://developer-local.allizom.org/admin/wiki/documentspamattempt/), see that ``blog_lang`` is ``en_us``.

I ran ``pdb`` inside ``akismet_parameters`` to see what was happening, and it is called four times:

* ``self.instance`` = True, ``self.instance.pk`` = False, ``self.instance.document_id`` = False
* ``self.instance`` = True, ``self.instance.pk`` = False, ``self.instance.document_id`` = False
* ``self.instance`` = True, ``self.instance.pk`` = False, ``self.instance.document_id`` = True
* ``self.instance`` = True, ``self.instance.pk`` = False, ``self.instance.document_id`` = True

So, the logic needs to be adjusted for the translation form case, which has some [bizarre setup](https://github.com/mozilla/kuma/blob/master/kuma/wiki/views/translate.py#L187-L214).

robhudson:
Thanks for the high quality feedback @jwhitlock 

I updated the code to correctly grab the locale from either the `tolocale` GET parameters for new translations, or the locale in the URL for editing existing translations. I tested both scenarios locally manually with my own akismet key and verified the `"blog_lang"` value was set correctly in the admin.

jwhitlock:
+r - Tested several page types and got the expected ``blog_lang``:
* [x] New English page: ``en_us``
* [x] Edit English page: ``en_us``
* [x] New Farsi page: ``fa, en_us``
* [x] Edit Farsi page: ``fa, en_us``
* [x] New Template: ``en_us`` (planned to be removed in [bug 1259870](https://bugzilla.mozilla.org/show_bug.cgi?id=1259870))

:+1:, merge when ready

PR #3823, opened 2016-03-25
Bug 1215384 - Preview button should be available on toolbar in source mode

askeing:
- change the command's modes.
- then build the ckeditor.js file.


stephaniehobson:
R+ Thanks :)

PR #3816, opened 2016-03-15
bug 1255609 - Enable training on false positives

jwhitlock:
Add additional data to wiki's DocumentSpamAttempt:
* data - JSON-serialized data sent to Akismet
* review - Staff review status (default 'Needs Review')
* reviewer - Staff reviewer
* reviewed - Review date

Old DocumentSpamAttempts do not have this additional data, so their status is set to "Review Unavailable".

Update the Django admin for DocumentSpamAttempt so that it works better as a dashboard for second-guessing Akismet:
* Limit the length of title and slug on the list display
* Allow filtering by "Needs Review", to treat the admin as a review backlog.
* Add an editable dropdown to quickly review multiple edits from known users
* When a review of "Spam" is chosen, it confims Akismet's choice, and no futher API call is needed.
* When a review of "Ham / False Positive" is chosen, submit the original data to Akismet's submit_ham API endpoint. On failure, revert to unreviewed status so it can be tried again.

Add a management command and task to drop submission data for old records, saving database resources and removing unneeded PII.

Update the anonymization scripts to drop submission data.

It can be hard to test this locally. If you want to, it helps to create a user 'viagra-test-123', which will always be counted as spam:

spam dashboard

PR #3825, opened 2016-03-28
bug 1255609 - Fixes for spam dashboard

jwhitlock:
Fix issues with PR #3816:

* On the "dashboard" (the Django admin list view), the Document name is length-restricted, so that long names do not cause the list view to become overly long and require scrolling to get to the review dropdown.
* An email is sent to the "first edits" email list when an edit is blocked, but not again when the block is reviewed.

Also improves ``make coveragetest``, to generate the HTML coverage docs and to recreate the skeleton of the 'cleaned' ``build`` folder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ test:
 coveragetest: clean
 	py.test --cov=$(target) $(target)
 
+coveragetesthtml: coveragetest
+	coverage html
+
 locust:
 	locust -f tests/performance/smoke.py --host=https://developer.allizom.org
 
@@ -37,6 +40,8 @@ intern:
 clean:
 	rm -rf .coverage build/
 	find kuma -name '*.pyc' -exec rm {} \;
+	mkdir -p build/assets
+	mkdir -p build/locale
 
 locale:
 	@mkdir -p locale/$(LOCALE)/LC_MESSAGES && \
--- a/kuma/wiki/admin.py
+++ b/kuma/wiki/admin.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from datetime import datetime
 import json
 
@@ -10,6 +12,8 @@ from django.template.response import TemplateResponse
 from django.template.defaultfilters import truncatechars
 from django.utils import timezone
 from django.utils.html import escape
+from django.utils.safestring import mark_safe
+from django.utils.text import Truncator
 from waffle import flag_is_active
 
 from kuma.core.admin import DisabledDeletionMixin
@@ -342,7 +346,7 @@ class DocumentZoneAdmin(admin.ModelAdmin):
 @admin.register(DocumentSpamAttempt)
 class DocumentSpamAttemptAdmin(admin.ModelAdmin):
     list_display = [
-        'id', 'user', 'title_short', 'slug_short', 'document', 'review']
+        'id', 'user', 'title_short', 'slug_short', 'doc_short', 'review']
     list_display_links = ['id', 'title_short', 'slug_short']
     list_filter = [
         'created', 'review', 'document__deleted', 'document__locale']
@@ -355,13 +359,37 @@ class DocumentSpamAttemptAdmin(admin.ModelAdmin):
         'review', 'reviewed', 'reviewer']
     readonly_fields = ['created', 'submitted_data', 'reviewer', 'reviewed']
 
+    MAX_LENGTH = 25
+
     def title_short(self, obj):
-        return truncatechars(obj.title, 25)
+        return truncatechars(obj.title, self.MAX_LENGTH)
     title_short.short_description = 'Title'
 
     def slug_short(self, obj):
-        return truncatechars(obj.slug, 25)
+        return truncatechars(obj.slug, self.MAX_LENGTH)
     slug_short.short_description = 'Slug'
+
+    def doc_short(self, obj):
+        u"""
+        Shorten document 'path (name)' representation in list view.
+
+        The important part is having an HTML break character such as a space,
+        so truncating paths as well as long titles, to look like:
+
+        /en-US/docs/Start/Of/Slug… (The start of the title…)
+        """
+        doc = obj.document
+        if doc:
+            full_path = u'/%s/docs/%s' % (doc.locale, doc.slug)
+            if len(full_path) <= self.MAX_LENGTH:
+                path = full_path
+            else:
+                path = Truncator(full_path).chars(self.MAX_LENGTH, u'…')
+            title = Truncator(doc.title).chars(self.MAX_LENGTH, u'…')
+            return u'%s (%s)' % (path, title)
+        else:
+            return mark_safe('<em>new document</em>')
+    doc_short.short_description = 'Document (if edit)'
 
     class NotEnabled(Exception):
         """Akismet is not enabled"""
--- a/kuma/wiki/apps.py
+++ b/kuma/wiki/apps.py
@@ -152,4 +152,4 @@ class WikiConfig(AppConfig):
         body = render_to_string('wiki/email/spam.ltxt',
                                 {'spam_attempt': instance})
         send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
-                  [config.EMAIL_LIST_FOR_FIRST_EDITS])
+                  [config.EMAIL_LIST_SPAM_WATCH])
--- a/kuma/wiki/apps.py
+++ b/kuma/wiki/apps.py
@@ -1,3 +1,5 @@
+from constance import config
+
 from django.apps import AppConfig
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
@@ -137,7 +139,11 @@ class WikiConfig(AppConfig):
         from .tasks import tidy_revision_content
         tidy_revision_content.delay(instance.pk)
 
-    def on_document_spam_attempt_save(self, sender, instance, **kwargs):
+    def on_document_spam_attempt_save(
+            self, sender, instance, created, raw, **kwargs):
+        if raw or not created:
+            # Only send for new instances, not fixtures or edits
+            return
         subject = u'[MDN] Wiki spam attempt recorded'
         if instance.document:
             subject = u'%s for document %s' % (subject, instance.document)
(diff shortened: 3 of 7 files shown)


jwhitlock:
Pre-emptive force-push to change double quotes to single quotes.

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
...
+        """
+        Shorten document 'path (name)' representation in list view.
+
+        The important part is having an HTML break character such as a space,
+        so truncating paths to 25 characters after the locale, and long titles
+        at 25 characters.
+        """
+        doc = obj.document
+        if doc:
+            full_path = '/%s/docs/%s' % (doc.locale, doc.slug)
```
jezdez: The slug may contain non-ASCII characters so this needs to be a Unicode string

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
...
+            if len(full_path) <= 25:
```
jezdez: If you want to have consistent length between non-ASCII characters using `len(unicodedata.normalize(full_path, 'NFC'))` would help, assuming `full_path` is a Unicode string

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
         return truncatechars(obj.slug, 25)
     slug_short.short_description = 'Slug'
 
+    def doc_short(self, obj):
+        """
+        Shorten document 'path (name)' representation in list view.
+
+        The important part is having an HTML break character such as a space,
+        so truncating paths to 25 characters after the locale, and long titles
+        at 25 characters.
```
jezdez: The 25 characters would make sense to store in a parameter or variable so it's easier to see what's going on.

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
...
+        """
+        doc = obj.document
+        if doc:
+            full_path = '/%s/docs/%s' % (doc.locale, doc.slug)
+            if len(full_path) <= 25:
+                path = full_path
+            else:
+                # 25 characters like /en_US/...End/Of/Slug
+                slash2_pos = full_path.index('/', 1) + 1
+                path = '%s...%s' % (full_path[:slash2_pos],
```
jezdez: Since we can use Unicode here, I'd guess it would help to use an Ellipsis here, `…`.

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
...
+                                    full_path[(slash2_pos - 22):])
+            title = truncatechars(doc.title, 25)
+            return '%s (%s)' % (path, title)
```
jezdez: Needs to be Unicode.

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
...
+        else:
+            return mark_safe('<em>new document</em>')
+    doc_short.short_description = 'Document (if edit)'
```
jezdez: Not sure if needed, but marking this for translation is not a bad idea

kuma/wiki/apps.py
```diff
@@ -137,7 +139,11 @@ def on_revision_save(self, sender, instance, **kwargs):
         from .tasks import tidy_revision_content
         tidy_revision_content.delay(instance.pk)
 
-    def on_document_spam_attempt_save(self, sender, instance, **kwargs):
+    def on_document_spam_attempt_save(
+            self, sender, instance, created, raw, **kwargs):
+        if raw or not created:
+            # Only send for new instances, not fixtures or edits
+            return
```
jezdez: Ah, good idea

kuma/wiki/apps.py
```diff
@@ -146,4 +152,4 @@ def on_document_spam_attempt_save(self, sender, instance, **kwargs):
         body = render_to_string('wiki/email/spam.ltxt',
                                 {'spam_attempt': instance})
         send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
-                  ['mdn-spam-watch@mozilla.com'])
+                  [config.EMAIL_LIST_FOR_FIRST_EDITS])
```
jezdez: Huh, didn't see that, I wonder if it would make sense to rename the config variable to better express what this is used for (e.g. `EMAIL_LIST_SPAM_WATCH`).

kuma/wiki/admin.py
```diff
@@ -363,6 +364,30 @@ def slug_short(self, obj):
...
+        so truncating paths to 25 characters after the locale, and long titles
+        at 25 characters.
+        """
+        doc = obj.document
+        if doc:
+            full_path = '/%s/docs/%s' % (doc.locale, doc.slug)
+            if len(full_path) <= 25:
+                path = full_path
+            else:
+                # 25 characters like /en_US/...End/Of/Slug
```
jezdez: Locales are written as `en-US` in URLs

kuma/wiki/tests/test_admin.py
```diff
@@ -54,6 +55,30 @@ def test_slug_short(self):
...
 
+    def test_doc_short_without_document(self):
+        dsa = DocumentSpamAttempt(slug='Slug')
+        assert self.admin.doc_short(dsa) == '<em>new document</em>'
+
+    def test_doc_short_with_document(self):
+        slug = 'NotSpam'
+        html = '<p>This page is not spam.</p>'
+        doc = document(title='blah', slug=slug, html=html, save=True,
+                       locale='en_US')
```
jezdez: I know the test will pass nevertheless, but to prevent red herrings in locale parsing, I'd use `en-US` here

Makefile
```diff
@@ -8,7 +8,7 @@ test:
 	py.test $(target)
 
 coveragetest: clean
-	py.test --cov=$(target) $(target)
+	py.test --cov=$(target) --cov-report term --cov-report html $(target)
```
jezdez: Is generating the html report needed here?

jezdez:
r+wc

kuma/wiki/admin.py
jwhitlock: This would be the first translated string (from kuma) in the Django admin in the wiki app, if not over the whole project. Is now the time to cross that line?

jezdez: Good point, I think there is a ticket about that IIRC. Carry on :+1:

jwhitlock: I can't find an existing ticket. I also think we shouldn't localize the admin, so I'm not opening one.

Makefile
jwhitlock: "term" is the default, which I always follow with "coverage html" to generate the HTML report, which include annotations of the missed lines and branches. It is unneeded in TravisCI, though.

kuma/wiki/apps.py
jwhitlock: I like the name ``EMAIL_LIST_SPAM_WATCH`` and using it for both. How do we ship that change without breaking production?

kuma/wiki/tests/test_admin.py
jwhitlock: Did you know I have trouble with "-" vs "_"? It's true!

Changing to not set the locale, so we get the default ``en-US``, and changing the tests accordingly.

kuma/wiki/admin.py
jwhitlock: I think I'm going to lean on ``Document.get_absolute_url()`` to do the work. I'm having trouble finding good unicode URLs for testing - https://developer.mozilla.org/fr/docs/Accessibilité/ARIA/Techniques_ARIA/Utiliser_le_role_progressbar is closest.

jwhitlock: Nope, ``Document.get_absolute_url()`` appears to be broken on Bengali, like on https://developer.mozilla.org/bn-BD/docs/Web/Guide/HTML/HTML5_ডকুমেন্টের_সেকশন_এবং_আউটলাইন.  Manually constructing the path from the slug.

jwhitlock:
Force-push to rebase on master, and:

* Instead of changing ``coveragetest`` target to generate HTML, added a ``coveragetesthtml`` target
* Add a class constant ``DocumentSpamAttemptAdmin.MAX_LENGTH`` instead of using 25 everywhere
* Change to standard end-truncation of slugs in ``doc_short`` instead of funky middle truncation, which still didn't really work for long paths and titles.
* Better unicode handling with explicit ``u'Unicode format %s'``, ``…`` ellipsis character instead of three periods, and Django's unicode-aware ``Truncator``
* Split ``doc_short`` tests instead of serially testing different slugs and titles.

kuma/wiki/apps.py
jezdez: Check in the prod constance admin if the default of the config value has been changed since it was deployed and change the value of the renamed config variable once deployed.

kuma/wiki/tests/test_admin.py
```diff
@@ -54,6 +56,41 @@ def test_slug_short(self):
...
+
+    def test_doc_short_long_unicode(self):
+        slug = u'Web/Guide/HTML/HTML5_ডকুমেন্টের_সেকশন_এবং_আউটলাইন'
+        title = u'HTML5 ডকুমেন্টের সেকশন এবং আউটলাইন'
+        html = '<p>This Bengali page is not spam.</p>'
+        doc = document(title=title, slug=slug, html=html, save=True,
+                       locale='bn-BD')
+        revision(document=doc, content=html, is_approved=True, save=True)
+        dsa = DocumentSpamAttempt(slug=slug, document=doc)
+        expected = u'/bn-BD/docs/Web/Guide/HT… (HTML5 ডকুমেন্টের সেকশন এব…)'
```
jezdez: Nice!

jezdez:
@jwhitlock Awesome changes!

kuma/wiki/apps.py
jwhitlock: The default and current value of ``EMAIL_LIST_FOR_FIRST_EDITS`` is ``mdn-spam-watch@mozilla.com``. Adding another commit to make the switch to ``EMAIL_LIST_SPAM_WATCH``...

jwhitlock:
One more commit to standardize on ``EMAIL_LIST_SPAM_WATCH``. @jezdez, if this looks good feel free to merge.

jezdez:
:boom:

PR #3826, opened 2016-03-31
bug 1188029 - Capture Akismet error data

jwhitlock:
If Akismet returns an error from check-comment, continue to block the edit, but save the details of the error for debugging, and adjust the notification sent to the spam watchers list.

### DIFF
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -15,6 +15,7 @@ from taggit.utils import parse_tags
 
 import kuma.wiki.content
 from kuma.core.form_fields import StrippedCharField
+from kuma.spam.akismet import AkismetError
 from kuma.spam.forms import AkismetCheckFormMixin, AkismetSubmissionFormMixin
 
 from .constants import (DOCUMENT_PATH_RE, INVALID_DOC_SLUG_CHARS_RE,
@@ -432,7 +433,7 @@ class RevisionForm(AkismetCheckFormMixin, forms.ModelForm):
     def akismet_error_message(self):
         return mark_safe(render_to_string('wiki/includes/spam_error.html', {}))
 
-    def akismet_error(self):
+    def akismet_error(self, parameters, exception=None):
         """
         Upon errors from the Akismet API records the user, document
         and date of the attempt for further analysis. Then call the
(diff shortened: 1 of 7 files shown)


kuma/wiki/apps.py
```diff
@@ -144,12 +144,14 @@ def on_document_spam_attempt_save(
...
+        DocumentSpamAttempt = self.get_model('DocumentSpamAttempt')
         subject = u'[MDN] Wiki spam attempt recorded'
         if instance.document:
             subject = u'%s for document %s' % (subject, instance.document)
         elif instance.title:
             subject = u'%s with title %s' % (subject, instance.title)
-        body = render_to_string('wiki/email/spam.ltxt',
-                                {'spam_attempt': instance})
+        is_error = instance.review == DocumentSpamAttempt.AKISMET_ERROR
+        context = {'spam_attempt': instance, 'is_error': is_error}
```
jezdez: just for the record, you could do `is_error` on template level as `{% set is_error=instance.review == instance.AKISMET_ERROR %}` or some similar condition

kuma/wiki/forms.py
```diff
@@ -442,7 +443,20 @@ def akismet_error(self):
             document = self.instance.document
         except ObjectDoesNotExist:
             document = None
-        # wrapping this in a try/finally to make sure that even if
+
+        if exception and isinstance(exception, AkismetError):
+            # For Akismet errors, save the submission and exception details
+            dsa_params = parameters.copy()
+            dsa_params['_akismet_status_code'] = exception.status_code
```
jezdez: does this need to be arbitrarily prefixed with `_` here? I mean, what does that mean, is this "private" data?

jezdez:
LGTM r+wc

kuma/wiki/apps.py
jwhitlock: Good point. I was avoiding it because I thought I had to do ``is_error == DocumentSpamAttempt.AKISMET_ERROR``, and didn't want to pass the model. My C++ brain is showing again. Will make the change.

kuma/wiki/forms.py
jwhitlock: No, I guess not. I was trying to distinguish between the data that was sent and what was returned, but ``akismet_`` is as good of a prefix as ``_akismet_`` for that.  Will change.

jezdez: Yep, thank you!

jwhitlock:
Force push to:

* Drop leading underscore in ``_akismet_status_code``, ``_akisment_response``, etc.
* Move ``is_spam`` logic to the email template ``spam.ltxt``, removing changes to kuma/wiki/apps.py

PR #3827, opened 2016-03-31
bug 1259233 - Add more waffle flags for Akismet tuning

jwhitlock:
Updates the tests to:
* Use py.test's ``assert`` style instead of unittest's ``self.assertX`` style.
* Add ``_setup_akismet_post`` method to help setup a RevisionForm for Akismet testing

Adds some additional waffle flags to allow per-user alternate Akismet payloads:
* ``spam_admin_override`` - Sets ``"role=administrator"`` so Akismet never identifies changes as spam
* ``spam_spammer_override`` - Sets ``"username=viagra-test-123"`` so Akismet always identifies changes as spam
* ``spam_testing_mode`` - Sets ``"is_test=True"`` so Akismet doesn't train on the content

Adds one per-user flag to control the response:
* ``spam_training_mode`` - Don't block saving content if Akismet identifies as spam or returns an error (but still record as a ``DocumetSpamAttempt``).

See the [bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1259233#c2) for the planned scenarios for these new flags.

### DIFF
--- a/kuma/wiki/tests/test_forms.py
+++ b/kuma/wiki/tests/test_forms.py
@@ -179,21 +179,35 @@ class RevisionFormTests(UserTransactionTestCase):
         self.assertTrue(rev_form.is_valid())
         self.assertEqual(rev_form.cleaned_data['tags'], '"JavaScript"')
 
+    def setup_akismet_post(
+            self, mock_requests, extra_data=None, revision=None, is_spam='false'):
+        """
+        Setup a RevisionForm for content checking with Akismet.
+
+        mock_requests - provided by @requests_mock.mock() decorator
+        extra_data - Additional data to add to the form post
+        revision - The current revision, if post is an edit
+        is_spam - Akismet's response to checking this for spam
+        """
+        mock_requests.post(VERIFY_URL, content='valid')
+        mock_requests.post(CHECK_URL, content=is_spam)
+        request = self.rf.post('/')
+        request.user = self.testuser
+        data = {
+            'title': 'Title',
+            'slug': 'Slug',
+            'content': 'Content',
+            'toc_depth': str(Revision.TOC_DEPTH_ALL),
+        }
+        if extra_data:
+            data.update(extra_data)
+        rev_form = RevisionForm(data=data, request=request, instance=revision)
+        return rev_form
+
     @pytest.mark.spam
     @requests_mock.mock()
     def test_akismet_enabled(self, mock_requests):
-        mock_requests.post(VERIFY_URL, content='valid')
-        request = self.rf.get('/')
-        # using a non-admin user here to make sure we can test the
-        # exemption rule below
-        request.user = self.testuser
-        data = {
-            'slug': 'Title',
-            'title': 'Title',
-            'content': 'Content',
-        }
-        rev_form = RevisionForm(data=data, request=request)
-
+        rev_form = self.setup_akismet_post(mock_requests)
         assert rev_form.akismet_enabled()
 
         # create the waffle flag and add the test user to it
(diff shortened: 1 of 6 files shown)


jezdez:
@jwhitlock The `SPAM_EXEMPTED_FLAG` flag was prefixed `wiki_` by design to separate wiki specific spam flags from general kuma anti-spam features. In case the new flags act only for wiki code, the new flags should follow that convention.

kuma/wiki/tests/test_forms.py
```diff
@@ -179,51 +181,51 @@ def test_case_sensitive_tags(self, mock_requests):
         self.assertTrue(rev_form.is_valid())
         self.assertEqual(rev_form.cleaned_data['tags'], '"JavaScript"')
 
-    @pytest.mark.spam
-    @requests_mock.mock()
-    def test_akismet_enabled(self, mock_requests):
+    def _setup_akismet_post(
```
jezdez: I don't know why this method has an underscore in front of it, who are you hiding it from?

kuma/wiki/tests/test_forms.py
```diff
@@ -179,51 +181,51 @@ def test_case_sensitive_tags(self, mock_requests):
...
+            self, mock_requests, extra_data=None, revision=None, is_spam='false'):
+        """Setup a RevisionForm for content checking with Akismet.
```
jezdez: I'm not a fan of triple quotes without a linebreak since `"""Something something"""` is less readable than

```
"""
Something something
"""
```

jezdez:
This looks good in general, but some things are missing:

- [ ] a migration to add the new waffle flags during deploy including a verbose (read: non-technical) description for the writers
- [ ] documentation for the waffle flags

Also, in case you want to make these flags generally available, I suggest to try to add them to the `kuma.spam` app if possible, since the `kuma.wiki` classes should be considered an implementation detail for the wiki spam. Assuming that profile spam will be the next target that would be a sensible separation of concern that was the original source of the separation of the apps.

kuma/wiki/tests/test_forms.py
jwhitlock: I was following the convention of the existing ``_test_akismet_error`` method. I assumed that was named that way to say "this is a helper method, not a test method", but I guess it could have been "hey py.test, pay no attention to this method".

How about I rename ``setup_akismet_post`` and move it under ``tearDown`` ?

kuma/wiki/tests/test_forms.py
jwhitlock: Me neither! Who wrote this code?

jwhitlock:
I think ``spam_admin_override``, ``spam_spammer_override``, and ``spam_testing_mode`` are widely useful.  When we start passing profile data to Akismet, I think they'd have the same purpose.  I can move them up to the base classes in ``kuma/spam``. 

In a similar way, ``spam_training_mode`` should be ``wiki_spam_training``, with a separate ``user_spam_training`` flag, for when we have mature wiki edit checking but want to quietly start profile checking.

Where do we document waffle flags?

jermainehamilton802:
#3827

jwhitlock:
Force push to:
* Move ``spam_admin_override``, ``spam_spammer_override``, and ``spam_testing_mode`` to ``kuma/spam``
* Apply changes in a new ``akismet_parameter_overrides`` method
* Fix ``spam_admin_override`` to set ``user_role``, not ``role``
* Remove leading underscore from new test method to ``setup_kismet_post``, fix docstring (but don't move it)
* Update features documentation to sync with code and add new flags.

There are no existing data migrations that setup waffle flags or switches. I'd like to continue the current process of adding them manually in the production admin.

Also, there's a lot of waffles and switches that should probably be removed. Other bugs...

jwhitlock:
There's too much work even to write a bug per deprecated waffle flag and switch. Added a [tracking bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1262183) to come back to this after spam is in place.

Also, for future @jwhitlock - yes, it was ironic that you went ahead and made the problem worse by adding 3 new flags an a switch. Enjoy your code cleanup!

stephaniehobson:
When we were doing performance work @groovecoder discovered waffle flags had a big impact on how quickly the pages were served. Do we know if that impact is limited to pages that have the waffle flags on them or does it slow the whole site down?

jwhitlock:
I would probably have to redo groovecoder's investigation to say for sure.  I believe the performance hit is when you try to determine if a flag or switch is enabled for a particular user, and because we include the waffle JS in the pages, this hit happens at least once per session.

groovecoder:
It happens when the wafflejs middleware loops over all waffle flags & switches, so it happens on every page request.

PR #3828, opened 2016-03-31
Fix Bug 1260197 - "Save and keep editing" workflow and errors

stephaniehobson:
Change from iframe submission to ajax submission to have better control over response.

Changed name of form field called "form" to be more descriptive.


vamepirelinjung:
จบ

vamepirelinjung:
จบ

vamepirelinjung:
จบ

PR #3831, opened 2016-04-08
bug 1263157, 1263163 - Async caching and task fixes

jwhitlock:
* Adds ``GenerationKeyJob.get_constructor_args``, so that the async refresh task can recreate the generation key

To do:

- [x] Make sure cacheback jobs have reasonable lifetimes and refresh timeouts
- [x] Investigate [bug 1263163](https://bugzilla.mozilla.org/show_bug.cgi?id=1263163), see if there is a similar code fix needed.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -1760,20 +1760,17 @@ class Revision(models.Model):
         if self.tidied_content:
             tidied_content = self.tidied_content
         else:
-            from .tasks import tidy_revision_content
-            tidying_scheduled_cache_key = 'kuma:tidying_scheduled:%s' % self.pk
-            # if there isn't already a task scheduled for the revision
-            tidying_already_scheduled = memcache.get(tidying_scheduled_cache_key)
-            if not tidying_already_scheduled:
-                tidy_revision_content.delay(self.pk)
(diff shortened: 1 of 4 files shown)


jwhitlock:
After push to staging, will:

- [ ] Create a new revision, compare differences
- [ ] In Django shell, call tidy_revision_content with refresh = True
- [ ] In Django shell, call tidy_revision_content with refresh = False
- [ ] In Django shell, call tidy_revision_content with bad PK

kuma/wiki/models.py
```diff
@@ -1760,20 +1760,17 @@ def get_tidied_content(self, allow_none=False):
...
+                if self.pk:
+                    from .tasks import tidy_revision_content
+                    tidy_revision_content.delay(self.pk, refresh=False)
                 tidied_content = None
             else:
                 tidied_content, errors = tidy_content(self.content)
+                if self.pk:
+                    Revision.objects.filter(pk=self.pk).update(
+                        tidied_content=tidied_content)
+        self.tidied_content = tidied_content
```
jwhitlock: ``tidied_content`` doesn't allow ``None``, so this should be:

    self.tidied_content = tidied_content or ''

PR #3832, opened 2016-04-08
Bug 1259151: Restore draft regression fix

stephaniehobson:
When a draft was detected the code was still autosaving over top of it. This fixes that and exposes the autosave process to the user in the UI.

Testing:
no recovered draft, save draft button present but disabled
type and wait 3 seconds for auto save
local storage has correct content
type and click save button (in less than 3 seconds, got to be quick)
local storage has correct content
autosave disabled notification present
type and wait for 3 seconds, draft not autosaved
enable auto save
type and wait for 3 seconds, draft autosaved
hard refresh page
report draft to restore, no save button, no autosave
type and wait 3 seconds, no autosave
restore draft
restored correctly, autosave enabled
type and wait 3 seconds, autosave
hard refresh
draft to restore
discard draft
hard refresh
no draft
make new draft
press discard button next to preview button
return to edit page
no draft
create spam edit
submit spam and be rejected
no draft
click draft save
refresh
restore draft sucessfully
ditch your spam content
create legit edit
click publish and keep editing button
page disabled
draft saved
page renabled
draft published and autosave enabled

### DIFF
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -187,6 +187,7 @@
     var DRAFT_NAME;
     var DRAFT_TIMEOUT_ID;
     var draftingEnabled = false;
+    var draftAutoSaveEnabled = true;
 
     var supportsLocalStorage = win.mdn.features.localStorage;
     var $form = $('#wiki-page-edit');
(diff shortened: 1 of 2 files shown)


jwhitlock:
Some strangeness in Desktop Safari, but maybe ok:

* Edit, autosave a draft
* Discard. Local Cache is cleared
* Edit, manually save a draft
* Discard, Local Cache is retained, and key shows up twice (same content)

kuma/static/js/wiki-edit.js
```diff
@@ -544,62 +554,83 @@
...
+        // delate old status
         $draftStatus.empty();
+
         if(action){
+            // clone elements and populate
             var $updateAction = $draftAction.clone().text(action);
             var $updateTime = $draftTime.clone().attr('title', now.toISOString()).text(nowString);
-            $draftStatus.append(gettext('draft') + ' ').append($updateAction).append(' ').append($updateTime);
+            // append elements
+            $draftStatus.append(gettext('Draft') + ' ').append($updateAction).append(' ').append($updateTime);
```
jwhitlock: Creating the string by combining elements will probably break in some languages.  I'd need to research the gettext() variants in JS to know what to recommend though, could be covered by follow-on work,

stephaniehobson:
The bug (and that is some weird behaviour, WTF Safari) causes content to be saved accidentally rather than discarded accidentally. It is problematic but I think it is okay to go forward with this and address this bug separately.

kuma/static/js/wiki-edit.js
stephaniehobson: Yeah, I'm sure it does and I'm not happy with it but this is how it was done before and I didn't want to re-think it as part of this.

PR #3835, opened 2016-04-12
Bug 1259173 - Adjust Akismet payload 

jwhitlock:
Rewrite the RevisionForm tests to setup forms for validation in the same way the views set them up (create, edit, and translate). Then:

Update the Akismet payload:
* Add HTTP headers
* Add ``blog`` parameter (front page of site), and fix default if omitted
* Add ``permalink`` parameter (page that is being edited
* Minimize ``comment_content`` to the changed and new content

Further improvements:
* Move payload generation to new ``AkismetRevisionData`` class, so the same logic can be used in the  ``RevisionForm`` as other places
* Save Akismet data in ``RevisionIP`` instance, and replay for ``submit-ham`` and ``submit-spam``.
* Add read-only displays of Akismet data to admin detail views.

This PR should be manually merged:
* [x] The first commit rewrites the tests, and the PR will be easier to understand after merging (optional)
* [x] The second-to-last commit includes a database migration that should be applied in production before the last commit
* [ ] The last commit adds the code that uses the new database column.

### DIFF
--- a/kuma/wiki/admin.py
+++ b/kuma/wiki/admin.py
@@ -9,8 +9,9 @@ from django.conf import settings
 from django.contrib.admin.views.decorators import staff_member_required
 from django.http import HttpResponse, HttpResponseRedirect
 from django.template.response import TemplateResponse
-from django.template.defaultfilters import truncatechars
+from django.template.defaultfilters import linebreaksbr, truncatechars
 from django.utils import timezone
+from django.utils.encoding import smart_text
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
 from django.utils.text import Truncator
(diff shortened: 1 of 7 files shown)


kuma/wiki/admin.py
```diff
@@ -476,17 +478,21 @@ class RevisionAkismetSubmissionAdmin(DisabledDeletionMixin, admin.ModelAdmin):
 
     def get_fields(self, request, obj=None):
         if obj is None:
-            return ['type', 'revision']
+            self.revision_id = request.GET.get('revision')
```
jezdez: admin instances are global, that will lead to threading issues

jwhitlock: ``AdminFormWithRequest`` makes more sense now. Is that the pattern to use here as well, or do you have a better solution?

jwhitlock:
Force push to change the strategy for displaying captured data from ``RevisionIP`` when creating a ``RevisionAkismetSubmission`` from the revision dashboard. Previously, the request query parameter was captured in ``get_fields``, which was not thread safe.  Now, ``add_view`` and ``change_view`` are adjusted to inject the data into the template context, and the template displays it (almost) like it was a field.

vamepirelinjung:
แก้ไข

vamepirelinjung:
แก้ไข

jwhitlock:
Force push to rebase and renumber the migration.

PR #3837, opened 2016-04-15
bug 1264361 - Require wiki.add_document permission

jwhitlock:
A user must have the ``wiki.add_document permission`` in order to add a page.

If this strategy works, then there is some follow-on tasks, including:
* Decide criteria for adding new users to Page Creators
* Automatically adding new users to Page Creators
* Adding an exception middleware to make it easier to return permission-based templates.

_Update 1_: Added a migration, manual steps are no longer needed

### DIFF
--- a/kuma/wiki/migrations/0030_add_page_creators_group.py
+++ b/kuma/wiki/migrations/0030_add_page_creators_group.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.db.models import Count
+from django.core.management.sql import emit_post_migrate_signal
+
+
+def create_page_creators_group(apps, schema_editor):
+    # Ensure wiki.add_document permission is created in a fresh database
+    # https://code.djangoproject.com/ticket/23422#comment:20
+    db_alias = schema_editor.connection.alias
+    emit_post_migrate_signal(2, False, 'default', db_alias)
+
+    # Grab the wiki.add_document permission
+    Permission = apps.get_model('auth', 'Permission')
+    add_perm = Permission.objects.get(codename='add_document',
+                                      content_type__app_label='wiki')
+
+    # Create the Page Creators group, if needed
+    Group = apps.get_model('auth', 'Group')
+    group, created = Group.objects.get_or_create(name='Page Creators')
+    group.permissions.add(add_perm)
+
+    # Add existing unbanned users to Page Creators
+    User = apps.get_model('users', 'User')
+    unbanned = (User.objects.annotate(ban_count=Count('bans'))
+                            .exclude(ban_count__gt=0)
+                            .values_list('id', flat=True))
+    group.user_set.add(*unbanned)
+
+
+def delete_page_creators_group(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    Group.objects.filter(name='Page Creators').delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wiki', '0029_add_dsa_review_type_error'),
+        ('users', '0005_update_tz'),
+        ('contenttypes', '__latest__'),
+        ('sites', '__latest__'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_page_creators_group,
+                             delete_page_creators_group),
+    ]
--- a/kuma/wiki/views/create.py
+++ b/kuma/wiki/views/create.py
@@ -34,18 +34,10 @@ def create(request):
         raise PermissionDenied
     # TODO: Integrate this into a new exception-handling middleware
     if not request.user.has_perm('wiki.add_document'):
-        email_url = (
-            'mailto:%s?'
-            'subject=Page Creation permission&'
-            'body=My username is %s.'
-            ' I want the page creation permission on %s because...'
-        ) % (config.EMAIL_LIST_MDN_ADMINS,
-             request.user.username,
-             request.get_host())
         context = {
             'reason': 'create-page',
             'request_page_url': DEV_DOC_REQUEST_FORM,
-            'request_access_url': email_url
+            'email_address': config.EMAIL_LIST_MDN_ADMINS
         }
         return render(request, '403-create-page.html', context=context,
                       status=403)
--- a/kuma/wiki/views/create.py
+++ b/kuma/wiki/views/create.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import newrelic.agent
 
+from constance import config
 from django.core.exceptions import PermissionDenied
 from django.shortcuts import render, redirect
 
@@ -8,7 +9,8 @@ from kuma.attachments.forms import AttachmentRevisionForm
 from kuma.core.decorators import never_cache, login_required, block_user_agents
 from kuma.core.urlresolvers import reverse
 
-from ..constants import (TEMPLATE_TITLE_PREFIX,
+from ..constants import (DEV_DOC_REQUEST_FORM,
+                         TEMPLATE_TITLE_PREFIX,
                          REVIEW_FLAG_TAGS_DEFAULT)
 from ..decorators import check_readonly, prevent_indexing
 from ..forms import DocumentForm, RevisionForm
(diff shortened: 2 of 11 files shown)


jwhitlock:
Force push to update copy to current version

safwanrahman:
> 
    Add a group called "Page Creators" that has the permission
    Assign all existing, unbanned users to that group

I think, Its better to do it automatically with data migration. what do you think is it possible?

pmclanahan:
Code looks good, but I've not setup kuma and tested it locally. I agree with @safwanrahman that it'd be nice to have the permission added and applied in a data migration, but if you feel that's dangerous then at least the permission should be automatic to avoid anyone having to remember in future.

jwhitlock:
I can see it both ways, but that's two votes for it, so I'll add the data migration.

kuma/wiki/views/create.py
```diff
@@ -30,6 +30,15 @@ def create(request):
     # Try to head off disallowed Template:* creation, right off the bat
     if not Document.objects.allows_add_by(request.user, initial_slug):
         raise PermissionDenied
+    # TODO: Integrate this into a new exception-handling middleware
+    if not request.user.has_perm('wiki.add_document'):
+        context = {
+            'reason': 'create-page',
+            'request_page_url': 'https://bugzilla.mozilla.org/form.doc',
```
safwanrahman: @jwhitlock Is it better to make it configurable by importing it from ``constants`` ? So its possible to change it in future as well as reuse the url

kuma/wiki/views/create.py
```diff
@@ -30,6 +30,15 @@ def create(request):
...
+            'request_access_url': 'mailto:mdn-admins@mozilla.org',
```
safwanrahman: same for it

stephaniehobson:
I'd rather we didn't re-use the permission denied image as the red stop sign could be interpreted as unfriendly. Not a blocker though.

kuma/wiki/migrations/0030_add_page_creators_group.py
```diff
@@ -0,0 +1,44 @@
...
+from django.core.management.sql import emit_post_migrate_signal
+
+
+def create_page_creators_group(apps, schema_editor):
+    # Ensure wiki.add_document permission is created in a fresh database
+    # https://code.djangoproject.com/ticket/23422#comment:20
+    db_alias = schema_editor.connection.alias
+    try:
+        emit_post_migrate_signal(2, False, 'default', db_alias)
+    except TypeError:  # Django < 1.8
```
safwanrahman: Do Kuma still use ``Django < 1.8`` ?

jwhitlock: Good point. Removing the exception handler.

kuma/wiki/migrations/0030_add_page_creators_group.py
```diff
@@ -0,0 +1,41 @@
...
+
+from django.db import migrations
+from django.core.management.sql import emit_post_migrate_signal
+
+
+def create_page_creators_group(apps, schema_editor):
+    # Ensure wiki.add_document permission is created in a fresh database
+    # https://code.djangoproject.com/ticket/23422#comment:20
+    db_alias = schema_editor.connection.alias
+    emit_post_migrate_signal(2, False, 'default', db_alias)
```
safwanrahman: Is there any other way except this approach?

jwhitlock: yes, do it manually in the Django admin.

jwhitlock: I'm sorry, that wasn't a useful answer.  Here's a better one:

https://code.djangoproject.com/ticket/23422 is the Django bug that reports using ``Permission`` objects in data migrations is difficult because they are not created until a ``post_migrate`` signal is emitted, at the end of migrations.  This isn't a problem if you've already ran the other 29 migrations, and are just running this one.  You run into problems when running migrations on an empty database (like we do in TravisCI), when the migration that refers to permissions will fail.

I ran into this [14 months ago](https://code.djangoproject.com/ticket/23422#comment:6), and [shipped a solution](https://github.com/mdn/browsercompat/blob/master/mdn/migrations/0002_data_add_group.py) based on manually creating the permission objects. However, the [later solution](https://code.djangoproject.com/ticket/23422#comment:20) in the bug report is cleaner, and is used here.

The good news is that [someone is working on a solution](https://code.djangoproject.com/ticket/23422#comment:26) that may be included in Django 1.10. 

The alternate to all this is my previous strategy: don't create groups and assign permissions in a data migration. Instead, do it in the Django admin on the running server.  The MDN development instructions include creating a superuser for development, which will get all the permissions by default, so it's not a requirement for 95% of developers either.

But, we have the migration, so I think we should use it, even if it is ugly.

kuma/wiki/views/create.py
```diff
@@ -32,10 +34,18 @@ def create(request):
...
     if not request.user.has_perm('wiki.add_document'):
+        email_url = (
+            'mailto:%s?'
+            'subject=Page Creation permission&'
+            'body=My username is %s.'
+            ' I want the page creation permission on %s because...'
+        ) % (config.EMAIL_LIST_MDN_ADMINS,
+             request.user.username,
+             request.get_host())
         context = {
```
jwhitlock: @safwanrahman asked in IRC if this would be better in the template.

The other existing code that uses an extended ``mailto:`` URL is https://github.com/mozilla/kuma/blob/master/kuma/users/jinja2/users/user_banned.html, setting the subject line.  This example isn't very good, because it sends the entire HTML section, including the URL, to the translators, when instead it should use a placeholder for the URL.  I don't consider this a good example for how to construct a ``mailto:`` link.

A better option would be to assign the URL to a variable, and to use a placeholder in the translation:

```jinja2
{% set request_access_subject="Page Creation permission" %}
{% set request_access_body="My username is %s. I want the page creation permission on %s because..." | format(request.user.username, request.get_host()) %}
{% set request_access_url="mailto:%s?subject=%s&body=%s" | format(request_email_address, request_access_subject, request_access_body) %}
<p class="notice">
  {% trans %}
  Creating a new page is a restricted action.
  To request a new page, please <a href="{{ request_page_url }}">file a bug</a>.
  If you need permission to create new pages, please <a href="{{ request_access_url }}">contact us</a>.
  {% endtrans %}
</p>
```

I prefer the version that constructs the ``mailto:`` string in the view.

This reminds me that I need to regenerate translation strings for this PR, and others, like the new draft functionality, so that the strings are available for translation in Pontoon.  I think this should be done post-merge, as this branch is behind master.

kuma/wiki/views/create.py
```diff
@@ -30,6 +32,23 @@ def create(request):
...
+    if not request.user.has_perm('wiki.add_document'):
+        email_url = (
+            'mailto:%s?'
+            'subject=Page Creation permission&'
+            'body=My username is %s.'
+            ' I want the page creation permission on %s because...'
+        ) % (config.EMAIL_LIST_MDN_ADMINS,
+             request.user.username,
+             request.get_host())
+        context = {
```
jwhitlock: @safwanrahman asked in IRC if this would be better in the template.

The other existing code that uses an extended ``mailto:`` URL is https://github.com/mozilla/kuma/blob/master/kuma/users/jinja2/users/user_banned.html, setting the subject line.  This example isn't very good, because it sends the entire HTML section, including the URL, to the translators, when instead it should use a placeholder for the URL.  I don't consider this a good example for how to construct a ``mailto:`` link.

A better option would be to assign the URL to a variable, and to use a placeholder in the translation:

```jinja2
{% set request_access_subject="Page Creation permission" %}
{% set request_access_body="My username is %s. I want the page creation permission on %s because..." | format(request.user.username, request.get_host()) %}
{% set request_access_url="mailto:%s?subject=%s&body=%s" | format(request_email_address, request_access_subject, request_access_body) %}
<p class="notice">
  {% trans %}
  Creating a new page is a restricted action.
  To request a new page, please <a href="{{ request_page_url }}">file a bug</a>.
  If you need permission to create new pages, please <a href="{{ request_access_url }}">contact us</a>.
  {% endtrans %}
</p>
```

This looks a lot like the previous version, with some additional variable setting in the template. I prefer the version that constructs the ``mailto:`` string in the view.

This reminds me that I need to regenerate translation strings for this PR, and others, like the new draft functionality, so that the strings are available for translation in Pontoon.  I think this should be done post-merge, as this branch is behind master.

jwhitlock:
Force push to:
* Rebase on master
* Populate the "Page Creator" group with unbanned users

jwhitlock:
Force push to squash to one commit

kuma/wiki/views/create.py
safwanrahman: Thanks for the clarification. I have worked on the past project where we made some different approach to insert the ``mailto`` link. From that experience, I have made the correction and make a commit
https://github.com/mozilla/kuma/commit/1a02e121f6c401481f7324df91527f10034bfffe

@jwhitlock If you like the approach i have made, you can cherry-pick it in. Let me know your observation. As I dont have development environment running, I can not have a test, but theoretically, it should work.

jwhitlock: I'm still not sure this is better way to do this, but it's a feeling, not reasons. The new code works in my dev environment, added as a follow-on commit with a new commit message.

PR #3838, opened 2016-04-16
Bug 1265151 - Missing translation of some strings

bychek-ru:
Localisation of table title


jwhitlock:
Works in local environment, and some strings ("Ревизия", "Название", "Комментарий") are already in the corpus. Thanks for the PR!

PR #3839, opened 2016-04-16
Bug 1265151 - Missing translation of some strings

bychek-ru:


jwhitlock:
This works in my local VM, but translations are not in the corpus (yet).

If you submit future PRs, please:
* Include the bug in the first line of the commit as well, such as "bug 1265151 - Update guide_links.html"
* Make your changes in a feature branch instead of master.

Thank you for improving MDN for translators!

PR #3841, opened 2016-04-19
Fix Bug 1260832: No first edit notification when spam detected.

stephaniehobson:


jwhitlock:
In development VM, I created a fresh user, edited a page. I get the "Thank you for taking the time to improve MDN!" banner below the action buttons. I added the flag "spam_spammer_override" to the user, made an edit, and clicked "Publish". I get the "Your submission includes content that may not be suitable for MDN" warning, and not the new user banner.

As usual, my comment is longer than the PR. +r :+1:

PR #3843, opened 2016-04-21
bug 1259173 - Update anonymize scripts for Akismet, other changes

jwhitlock:
Update the anonymize scripts:

* Anonymize ``UserBans`` instead of truncating the table, so that spam scripts can be developed on anon databases
* Anonymize Akismet data in ``RevisionIP``
* Ensure that ``DocumentSpamAttempt`` is cloned (already anonymized)

Also:
* Explicitly drop legacy tables. ``auth_message`` and ``django_cache`` are in production but not newly generated databases. ``tagging_tag`` and ``tagging_taggeditem`` are no longer in production either.
* Previously, command errors would cause further unrelated script errors. Now, report the original command error and some more of the command error (for example, what SQL statement failed)
* Make the script PEP8 compliant.

This can be tested in a development VM:

```
scripts/clone_db.py kuma -u kuma -p kuma --debug
```


scripts/anonymize.sql
```diff
@@ -186,8 +187,12 @@ UPDATE wiki_revisionip SET
 UPDATE wiki_revisionip SET
     referrer = CONCAT("https://example.com/", MD5(CONCAT(referrer, @common_hash_secret)))
     WHERE referrer != "";
+UPDATE wiki_revisionip SET data = null;
```
jpetto: Forgive my lack of context - just verifying the intent is to wipe the `data` field completely.

jpetto:
Don't see any obvious syntax or logic errors, though this codebase is all new to me, so take that with a grain of salt. 😁

scripts/anonymize.sql
jwhitlock: Yes. ``data`` contains a JSON-serialized simple dictionary of request details, including sensitive information like IP and email address. ``RevisionIP`` objects are deleted after 30 days, this neuters the younger ones.

PR #3851, opened 2016-04-29
bug 1268511 and bug 1259870 - Akismet integration fixes

jwhitlock:
Two fixes:
* [bug 1268511](https://bugzilla.mozilla.org/show_bug.cgi?id=1268511) - Sometime tags have quotes, sometimes they don't.
* [bug 1259870](https://bugzilla.mozilla.org/show_bug.cgi?id=1259870) - Templates are stored like pages, but shouldn't be checked by Akismet


jwhitlock:
Force push to change ``akismet_enabled`` method so coverage won't change.

PR #3852, opened 2016-05-03
Add and use new kuma_base docker image

jgmize:
@jwhitlock this is sort of working for me locally in that it builds (with `make build-base`) and boots (with `docker-compose up`), but I don't have a local copy of an MDN dev DB to go any further.


jwhitlock:
@jgmize is the anonymized production database suitable for testing? Or are you blocked by [bug 1271509](https://bugzilla.mozilla.org/show_bug.cgi?id=1271509)?

jgmize:
> @jgmize is the anonymized production database suitable for testing? Or are you blocked by bug 1271509?

The anonymized prod db should be sufficient, thanks.  I'm very much in favor of your proposal in bug 1271509, but it's not a blocker for this.

jwhitlock:
I ran ``make build-base`` with Docker for Mac v1.11.1, output in [my Dropbox](https://www.dropbox.com/s/gkcxg81n6yntt6y/pr3852-docker.txt?dl=0).  It appears to run correctly, then shuts down the container. I don't know enough about Docker to go any farther than this.

jwhitlock:
If you are doing the rebase anyway, please follow [CONTRIBUTING.md](https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md) and prefix the commit subject, like ``bug 1218563 - new kuma_base docker image``.

jgmize:
My latest commit upgrades django-pipeline to 1.6.8, heavily inspired by https://github.com/mozilla/bedrock/pull/4151 and with hopes of similar performance improvements. Note that as part of this I had to switch from cleancss to cssmin, which is a non-issue for docker deployments but means there will be an extra step for the legacy deployments. As @jwhitlock and I discussed the django-pipeline upgrade should probably be refactored into a separate PR which can land ahead of this one, at which point I'll rebase this PR. This latest update was built by https://ci.us-west.moz.works/view/MDN/job/mdn_base_image/2/console and you can pull it down for testing locally with `docker pull quay.io/mozmar/kuma_base`.

stephaniehobson:
You said there's an "extra step for the legacy deployments" for the CSS? Is that documented anywhere? (Maybe it belongs [here](kuma.readthedocs.io/en/latest/development.html)).

Can you please diff the CSS created by cleancss to the CSS created by cssmin? If there are differences I'd like to know what they are.

jgmize:
The extra step I mentioned would be something along the lines of `sudo npm install -g cssmin` on the admin node prior to the chief deploy, but I'll need to research further and see if that should be handled by the webops-managed puppet manifests.

To see the differences in css, you can `git checkout master` and then run `./manage.py collectstatic` in your native install or your vagrant VM, and then make a copy of the css folder outside of the static dir (for example,  `cp -r static/build/styles /tmp/cleancss`). Then you can checkout this PR locally (I have an [alias](https://github.com/jgmize/dotfiles/blob/master/.gitconfig#L15) that allows me to `git pr 3852; git co pr/3852` for this, but you may of course have another method that works better for you). Then you can run collectstatic in a docker container built off this brach with `docker-compose run web ./manage.py collectstatic --noinput` (I had to `mkdir -p build/assets build/locale` the first time I ran this before it worked, but you likely already have those directories). If you get a complaint about a missing languages.json then `docker-compose run web ./manage.py update_product_details` should resolve it. Finally, you can look at the diff with `diff -r static/build/styles /tmp/cleancss`.

jwhitlock:
I was able to run ``make build-base``, and it ran without errors, with Docker version 1.12.0-rc2-beta16 (the beta).

I ran ``./manage.py collectstatic`` against commit d0ca8a0, the base commit for this PR, using Vagrant, and saved the static folder. I then ran ``docker-compose run web ./manage.py collectstatic --noinput`` to generate the new static files.  All the files were identical except for:

* ``static/build/js/syntax-prism.js``
* ``static/build/styles/editor-content.css``
* ``static/build/styles/jquery-ui.css``
* ``static/build/styles/tagit.css``
* ``static/build/styles/wiki.css``

These are compressed versions (the uncompressed are identical), so the "hashed" versions differ as well (d0ca8a0 generates ``syntax-prism.7a66ddfa68bf.js``, this PR generates ``syntax-prism.bf7f664affd6.js``).

I tried to enlist GitHub to display the diffs, but it struggles as well. I opened [a gist](https://gist.github.com/jwhitlock/95c526511ff897d8d4a745e72eafec43). The first revision is the files in d0ca8a0, and the second revision is from the PR. There are differences, but highlighting is no help in the [revisions view](https://gist.github.com/jwhitlock/95c526511ff897d8d4a745e72eafec43/revisions)

I think a CSS pretty-printer would help, but I leave that as an exercise for another reviewer..

On to testing ``docker-compose up``...

jwhitlock:
``docker-compose up`` does... something. It appears to mostly work, but I'm not familiar enough with Docker to get past the issues.

* The ``worker`` container exits almost immediately, refusing to run as root
* The ``web`` worker appears to work and connect to the other containers. A homepage request results in an Internal Server Error, due to missing database tables and settings (``settings.MDN_LANGUAGES``, maybe a side effect of the missing database)
* The ``nginx`` container appears to work, with a request to http://localhost
* The ``memcached`` container appears to work, did not verify
* The ``mysql-data`` container exits immediately
* The ``mysql`` container appears to go up and down as it goes through an initialization routine, and then runs without a database.
* The ``redis`` container appears to work, did not verify
* The ``kumascript`` container is silent but stays up, did not verify

If this isn't the Linux experience, then we may need a few rounds of troubleshooting to get the OS X build up to speed. Or, a new dev laptop for me.

On my side, I need to put together a PR for updating ``django-pipeline``, and clean up the requirements files so you don't have to disable so many of the dependencies.  For example, the missing ``argparse`` dependency is causing the TravisCI build to fail. I've already moved it to constraints in PR #3883 (which is an experiment, _not_ to be merged as is).

jgmize:
Before it will work for you, you'll need data in the database. Start a mysql container with `docker-compose up mysql` in one shell, then once it's up and running go to another shell, change to the directory where you keep your database backups, and restore an anonymized sql.gz fille with something like the following, substituting the appropriate name for the backup file if you have something more recent:

```sh
docker exec -i kuma_mysql_1 bash -c "zcat | mysql -u root --password=docker" < developer_mozilla_org-anon-20160407.sql.gz
```

jgmize:
The celery worker has never been functional for me either; as you know I'm building on top of previous incomplete work. So far locally I've been able to get it to render all of the pages in read-only mode from a restored db, but I have not yet tackled any of the editing functionality.

jwhitlock:
I had to wipe out my Docker disk, due to a [known issue](https://forums.docker.com/t/no-space-left-on-device-error/10894) with reclaiming space on the OS X beta.

After rebuilding, the process to load a database worked for me, but with a slightly different command:

```
docker exec -i kuma_mysql_1 bash -c "mysql -u root --password=kuma developer_mozilla_org" < ../kuma_attic/developer_mozilla_org-anon-20160707.sql
```

I did not time it, but it seemed to be on the order of 20-30 minutes to load the database.  This was with a full ``docker-compose up`` running, not a targeted ``docker-compose up mysql``.

The web server worked at http://localhost:8000, and there was content.

The Django-only cacheback task ``kuma.wiki.jobs.DocumentContributorsJob`` appeared to run correctly, so some of Celery is working.  There was an initial error because the celery task started before the database was available. I'm not sure if there is a way with ``docker-compose`` to sequence the container startup order.

It's good to go from n=1 to n=2 working deploys.

We can't merge yet, because of known issues, some requiring a rebase:
* [ ] flake8 issues
* [ ] needs to build and run tests in TravisCI (requirements issue around ``argparse``)
* [ ] pull pipeline changes into a separate PR, update existing static assets workflows as needed, deploy to SCL3 environments
* [ ] Minimize other requirements changes, deploy as separate PR

If you want to get the Docker changes merged, you may be able to split the Docker-only bits into a new PR, maybe with an update to the [Docker docs](https://kuma.readthedocs.io/en/latest/installation-docker.html), and we may be able to merge that faster.

stephaniehobson:
@jwhitlock Thank you for extracting and creating the CSS file diffs. I am a reasonable facsimile of a CSS pretty-print and compared them. There are some... interesting differences but nothing that breaks anything.

If there happens to be a configuration option to get cssmin to stop changing hex values like `#fff` to key word values like `white` we should tell it to stop. Not a big deal for a two character difference but a little silly when the keyword is `slategrey`.

jwhitlock:
I read the whole source of [cssmin](https://github.com/jbleuzen/node-cssmin/blob/master/cssmin.js), and can't find where it would be swapping hex colors for color names.

I then looked at [clean-css](https://github.com/jakubpawlowicz/clean-css), the old project, and it has a [config option](https://github.com/jakubpawlowicz/clean-css/blob/master/lib/utils/compatibility.js#L12) for replacing colors with names, and [the database](https://github.com/jakubpawlowicz/clean-css/blob/master/lib/colors/hex-name-shortener.js) to do it. Maybe I got the before and after backwards?

stephaniehobson:
Well then, this is an improvement, excellent :)

jwhitlock:
bedrock changed processors again:

* https://github.com/mozilla/bedrock/pull/2805 - @pmclanahan replaces yuglify with cssmin and uglify-js, "speed up local dev and speed up deployment" (March 2015)
* https://github.com/mozilla/bedrock/pull/4221 - @pmclanahan replaces cssmin and uglify-js with yuglify, no in-PR justification (July 2016)

Should we incorporate this change as well, and re-generate the output CSS / JS?

pmac:
Basically the reason is that cssmin is mostly abandoned now and yuglify is a good wrapper around the better things. I needed to upgrade libs when we upgraded to version 4.x of Node and couldn't stay with what we had.

jgmize:
will rebase and switch to yuglify.

kuma/settings/common.py
```diff
@@ -931,7 +926,19 @@ def _get_languages_and_locales():
         'output_filename': 'build/js/ace.js',
     },
 }
-
+PIPELINE = {
+    'STYLESHEETS': PIPELINE_CSS,
+    'JAVASCRIPT': PIPELINE_JS,
+    'DISABLE_WRAPPER': True,
+    'JS_COMPRESSOR': 'pipeline.compressors.uglifyjs.UglifyJSCompressor',
```
jwhitlock: If we load this from the environment as well, we might be able to merge before rolling uglify / yuglify / whatever to Vagrant and production.

jgmize:
@jwhitlock I've rebased this branch on master, but pushed it to a new, separate branch as we discussed: https://github.com/jgmize/kuma/tree/docker-rebase

jwhitlock:
This works for me: https://github.com/jwhitlock/kuma/tree/docker_1218563

stephaniehobson:
@jgmize @jwhitlock are either of you still using https://github.com/mozilla/kuma/tree/cloudops-docker or https://github.com/mozilla/kuma/tree/docker-cloudops ? If not, can we delete the branches?

jgmize:
closing in favor of #3959

PR #3853, opened 2016-05-03
Fix cs (Czech) locale

jwhitlock:
dennis is detecting an issue with the cs locale translation. This PR adds dennis to the local requirements, and fixes the translation issue.


locale/cs/LC_MESSAGES/django.po
```diff
@@ -3206,7 +3206,7 @@ msgid ""
 "Our volunteers haven't translated this article into <bdi>%(locale)s</bdi> "
 "yet. <a href=\"%(help_link)s\">Join us and help get the job done!</a>"
 msgstr ""
-"Naši dobrovolníci ještě tento článek do jazyka <bdi>%s (locale)</bdi> "
+"Naši dobrovolníci ještě tento článek do jazyka <bdi>%(locale)s</bdi> "
```
safwanrahman: This Should be fixed from the Poonton end. So @MikkCZ can fix it from there as he is the commiter there

jwhitlock: That makes sense. We have some time before the next production push, we'll see if @MikkCZ fixes it in the next 24 hours or so. Broken string:

https://pontoon.mozilla.org/cs/mdn/LC_MESSAGES/django.po/?search=our+volunteers&string=141905

MikkCZ: Thank you for letting me know. The malformed format string is even my fault, I am very sorry for that. I have fixed it in Pontoon.

Btw. great to see the new Pontoon direct links in action. ;)

safwanrahman:
So the fact is, how we determine which locale causes error? In kitsune, we save the error log in poststatus.txt(1) and serve it over media. then with a app(2), we can check the status of all locale. Can we do same with kuma?

1. https://github.com/mozilla/kitsune/blob/master/scripts/update/deploy.py#L45
2. https://postatus.herokuapp.com/p/SUMO/postatus

jwhitlock:
Does that give any advantage over TravisCI checking the .po files?

safwanrahman:
So The thing will be checked by travis ci? that there is a error with the string push?

In that case, a deploy will be blocked by bad strings.

jwhitlock:
No, it's a non-blocking build failure, like this: https://travis-ci.org/mozilla/kuma/builds/127525109

safwanrahman:
How it will be possible to understand which locale has issue with their string?

safwanrahman:
Sorry, I meant about Deploy. So while deploy it will check that if any string has error. and if there is error it will fail. right?

So should a string error be cause a deploy blockage?

jwhitlock:
Here's my process:

1. [See the red X next to ``TOXENV=dennis``](https://travis-ci.org/mozilla/kuma/builds/127525109)
2. [Click the line to view the build](https://travis-ci.org/mozilla/kuma/jobs/127525115)
3. [Scroll down to the output of the build](https://travis-ci.org/mozilla/kuma/jobs/127525115#L242)
4. [Read the output](https://travis-ci.org/mozilla/kuma/jobs/127525115#L243-L252) and see what file and what string has the error.

jwhitlock:
The ``.po`` files are compiled as part of the deploy script. Any errors like this will be found during the push to the staging server.

locale/cs/LC_MESSAGES/django.po
jwhitlock: @MikkCZ thanks for the translations, and thanks for the quick fix!

Because of the size of the MDN translation files, Pontoon only loads the first few lines. If you want to use Pontoon links in the future, I think a search term is needed to ensure the string you want is loaded when you click the page.

jwhitlock:
cs translation file fixed via Pontoon.

locale/cs/LC_MESSAGES/django.po
mathjazz: @jwhitlock Thanks for spotting the issue with referencing strings directly!

It should be fixed now:
https://pontoon.mozilla.org/cs/mdn/LC_MESSAGES/django.po/?string=141905

jwhitlock: :+1: works for me - thanks for the quick fix!

mathjazz: And thanks to @MikkCZ for linking to this thread while talking about something completely unrelated. :-)

PR #3854, opened 2016-05-03
Fix Bug 1269452: Track page create 403

stephaniehobson:
Log when a user runs up against the page creation permission denied page.

### DIFF
--- a/jinja2/403.html
+++ b/jinja2/403.html
@@ -75,3 +75,9 @@
     {{ super() }}
     {% javascript 'framebuster' %}
 {% endblock %}
+
+{% block js %}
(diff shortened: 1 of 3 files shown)


jinja2/403.html
```diff
@@ -75,3 +75,9 @@ <h1 class="page-title">{% block page_title %}{{ _('Permission Denied') }}{% endb
     {{ super() }}
     {% javascript 'framebuster' %}
 {% endblock %}
+
+{% block js %}
+<script type="text/javascript">
```
safwanrahman: Should have indent

PR #3855, opened 2016-05-03
bug 957802 - Vagrant install fixes

jwhitlock:
Tested on OS X 10.11.4, Vagrant 1.8.1, VirtualBox 5.0.16r105871, and Ansible 2.0.2.0.

* Add an environment variable ``VAGRANT_CACHIER=false`` to disable the cachier plugin config. Currently still enabled, but may fix some Linux issues.
* Bump tested version numbers
* Use Ansible output in docs
* Fix ``./manage.py import_kumascript_modules``
* Run two web workers, so that KumaScript can make Kuma API requests at the same time Kuma is asking to render a document.
* Allow Kuma to defer async tasks (``CELERY_ALWAYS_EAGER=False``)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
-web: gunicorn kuma.wsgi -b 0.0.0.0:8000
+web: gunicorn kuma.wsgi -w 2 -b 0.0.0.0:8000
 worker: python2.7 manage.py celery worker --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
 camera: python2.7 manage.py celerycam --freq=2.0
 kumascript: node kumascript/run.js
--- a/kuma/wiki/management/commands/import_kumascript_modules.py
+++ b/kuma/wiki/management/commands/import_kumascript_modules.py
@@ -35,8 +35,7 @@ class Command(BaseCommand):
 
         for slug in KS_AUTOLOAD_MODULES:
             template_response = requests.get(RAW_TEMPLATE_URL % slug)
-            doc = Document(title=slug, slug=slug,
-                           category=Document.CATEGORIES[0][0])
+            doc = Document(title=slug, slug=slug)
             try:
                 doc.save()
                 loaded_docs.append(slug)
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -26,6 +26,7 @@ Vagrant.configure('2') do |config|
     IP = ENV['VAGRANT_IP'] || '192.168.10.55'
     GUI = (ENV['VAGRANT_GUI'] || 'false') == 'true'
     ANSIBLE_VERBOSE = (ENV['VAGRANT_ANSIBLE_VERBOSE'] || 'false') == 'true'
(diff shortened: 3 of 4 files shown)


Procfile
```diff
@@ -1,4 +1,4 @@
-web: gunicorn kuma.wsgi -b 0.0.0.0:8000
+web: gunicorn kuma.wsgi -w 2 -b 0.0.0.0:8000
```
jezdez: Interesting, are you saying Gunicorn was blocking on requests when it requested things from kumascript?

kuma/settings/local.py
```diff
@@ -29,3 +29,6 @@
 # Don't cache non-versioned static files in DEBUG mode
 if DEBUG:
     WHITENOISE_MAX_AGE = 0
+
+# Send tasks to celery workers
+CELERY_ALWAYS_EAGER = False
```
jezdez: Hmm, FWIW, this may be a historical thing but in the past it was often asked to be able to run MDN without needing Celery. I guess under the 12factor app/docker pattern that would not be the case as Celery would always run, so this may totally be something that needed to happen anyway.

kuma/wiki/management/commands/import_kumascript_modules.py
```diff
@@ -35,8 +35,7 @@ def handle(self, *args, **options):
 
         for slug in KS_AUTOLOAD_MODULES:
             template_response = requests.get(RAW_TEMPLATE_URL % slug)
-            doc = Document(title=slug, slug=slug,
-                           category=Document.CATEGORIES[0][0])
+            doc = Document(title=slug, slug=slug)
```
jezdez: good catch!

Procfile
jwhitlock: Yes, it appears the user's request "GET page" (actually POST new page) was blocking KumaScript's "Fetch Raw Source" request:

KumaScript overview

jezdez: So this must have been an issue since we moved to gunicorn, not long ago. The runserver we used before is threaded and just spins up another thread to process incoming requests.

groovecoder:
I was able to get all the way thru https://kuma.readthedocs.io/en/latest/installation.html without any errors or issues.

I was able to import the kumascript modules, create a kumascript "hello" template, and use it in a page.

mdn-and-kumascript

groovecoder:
OS X 10.10.5, Vagrant 1.7.4 driving VirtualBox 5.0.10 r104061

jwhitlock:
Kumascript works without errors, but the rendered page is the N-1 version of the page.  For example, if the current version says "Edit 3", and I change it to "Edit 4", the rendered page after saving contains "Edit 3". A force-refresh fixes it.

This might be a different problem, and not a merge blocker.

kuma/settings/local.py
jwhitlock: I've tested with both ``CELERY_ALWAYS_EQUAL=True`` and ``=False``. KumaScript works with both, and in both KumaScript seems to render the n-1 version. I'll take this out so that it can be controlled from the environment like 12factor intended.

jwhitlock:
Force-push to remove the ``CELERY_ALWAYS_EAGER=False`` line, can be set or cleared from environment instead.

PR #3857, opened 2016-05-04
bug 1270032 Change Firefox OS to B2G OS

MatonAnthony:
Change the label under "Mozilla Docs" menu to match the current name of Firefox OS (B2G OS).


jwhitlock:
Merged change with PR #3859 - Thanks!

PR #3858, opened 2016-05-05
Linux statd doc update 957802

jwhitlock:
Same as PR #3856, but with an additional commit for some styling fixes.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -211,7 +211,7 @@ Now you can sign in with GitHub at https://developer-local.allizom.org/
 
 .. _Errors:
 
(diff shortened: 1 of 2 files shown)


safwanrahman:
As Its a Ubuntu related issue, it should go under the Error Section of Ubuntu in the installation docs
https://kuma.readthedocs.io/en/latest/installation.html

jwhitlock:
That makes sense to me @safwanrahman - moving to install doc.

jwhitlock:
Force push so that whitespace deletion in troubleshooting doc goes away

PR #3860, opened 2016-05-05
bug 957802 - Add dennis, make targets for locale, local linting

jwhitlock:
dennis is used to check locale files for common errors.

Added some make targets for locale work:

* make localetest - Run dennis on locales
* make localeextract - Extract translatable strings
* make localecompile - Compile translations for local use
* make localerefresh - Run all the steps to update locales, except for commit

Updated the localization documentation for these new targets.

The only string update was the addition of "B2G OS", which I don't believe should be localized, so I'm not including a string update this week.


jwhitlock:
Force push to move ``polib`` back to ``requirements/default.txt``, since it is used in the deploy script to compile ``.po`` files to ``.mo`` files, via ``locale/compile-mo.sh``.

Makefile
```diff
@@ -49,5 +49,20 @@ locale:
...
+	dennis-cmd lint --errorsonly locale/
+
+localeextract:
+	python manage.py extract
+	python manage.py merge
+
+localecompile:
+	cd locale; ./compile-mo.sh . ; cd --
+
+localerefresh: localeextract localetest localecompile compilejsi18n collectstatic
```
willkg: This suggests that you're compiling to `.mo` files and then checking those in. Typically we compile `.po` -> `.mo` files during deploy.

How does this work in MDN?

docs/localization.rst
```diff
@@ -67,25 +66,25 @@ Adding a new Locale
...
     page. E.g., https://developer-local.allizom.org/ml/
 
 #.  BONUS: Use `podebug` to test a fake translation of the locale::
 
-        $ cd locale
-        $ podebug --rewrite=bracket templates/LC_MESSAGES/django.pot ml/LC_MESSAGES/django.po
-        $ ./compile-mo.sh .
+        cd locale
+        podebug --rewrite=bracket templates/LC_MESSAGES/django.pot ml/LC_MESSAGES/django.po
+        ./compile-mo.sh .
```
willkg: Dennis can do fake translations, too. It has some of the same transforms that podebug has, but also has some super ones.

willkg:
The `chief_deploy.sh` script compiles `.mo` files, but doesn't do a dennis lint pass.

I think there are a few scenarios which don't work well.

1. You go to extract and merge strings, but running `make localerefresh` fails on the linting step because existing translated strings have errors. What happens here? Are you blocked from merging new strings?
2. One of the locales has strings which have errors. The `chief_delpoy.sh` script compiles those to `.mo` files during deploy. That means that bad strings can get to production and cause HTTP 500 errors. Does that sound right or am I reading things wrong?
3. We had a case where the `.po` file had a `Plural-Forms:` line that was malformed, so dennis didn't raise any lint errors (it doesn't look at the header), but using the `.mo` file in production raised HTTP 500 errors. It sucked. We changed how we lint things so that we raised errors during the compile step, too. We used this script: https://github.com/mozilla/fjord/blob/master/bin/compile-linted-mo.sh

Of those, I think the first and second are problematic. I'm not sure about the third, though. Maybe that was just a once-in-a-lifetime thing.

docs/localization.rst
willkg: http://dennis.readthedocs.io/en/latest/translating.html

As I recall, there's a bug in some of the translation transforms where they don't take variables into account. I started fixing that, but I don't think I ever finished. :(

Makefile
jwhitlock: [During the deploy](https://github.com/mozilla/kuma/blob/master/scripts/chief_deploy.py#L51-L54), ``locale/compile-mo.sh`` is called to compile the .mo files. The .mo files are [ignored](https://github.com/mozilla/kuma/blob/master/.gitignore#L31), so are not checked in.

``make localextract`` will extract any string data to ``.po`` files, which are checked in.  This needs to be done when strings are updated in code etc.

``make localerefresh`` is a new thing, to collect all the thing you need to do in development to extract strings, lint translations, and get the new translations available in your dev environment.

willkg: Ahh... ok. I didn't catch that `make localerefresh` was intended for the dev environment. My bad.

willkg:
I didn't catch that this is predominantly for dev environments.

I still think MDN has some of the issues I described in the previous comment. Maybe those should get spun out as separate bugs or maybe write up a bug when you hit the problems?

Anyhow, everything else here looks good.

docs/localization.rst
jwhitlock: It looks like there's a bug in plural forms:

```
#: kuma/search/jinja2/search/results.html:88
#, python-format
msgid "%(count)s document found in %(locale_native)s."
msgid_plural "%(count)s documents found in %(locale_native)s."
msgstr[0] ""
msgstr[1] ""
msgstr[0] "%(count)s document found in %(locale_native)s ye scalleywag❗."
msgstr[1] "%(count)s documents found in %(locale_native)s cap'n❗."
```

This causes an error when compiling to .mo files. If you remove the blank lines, it works:

```
#: kuma/search/jinja2/search/results.html:88
#, python-format
msgid "%(count)s document found in %(locale_native)s."
msgid_plural "%(count)s documents found in %(locale_native)s."
msgstr[0] "%(count)s document found in %(locale_native)s ye scalleywag❗."
msgstr[1] "%(count)s documents found in %(locale_native)s cap'n❗."
```

Still, fun.

jwhitlock:
Thanks for the review, and the great ideas! Added two commits based on review:

* The instructions for adding a locale needed updating, and still referred to ``settings.py`` rather than ``kuma/settings/common.py``.
* Added ``dennis`` to the deployment process, to detect locale file issues on deploy. These are also checked in TravisCI, but do not fail the build. However, it could cause issues if faulty locales are pushed to production.

In the failure cases:
1. __`make localerefresh` fails on the linting step because existing translated strings have errors.__ In this case, `make localeextract` is sufficient to extract the strings, which can be committed. `make localerefresh` is a convenience method for viewing updated translations in a development environment. String extraction doesn't require all the steps, and it alerts the developer that bad localizations were pushed from Pontoon. So, not a blocker, but does shout that work needs to be done before the deploy.
2. __One of the locales has strings which have errors.__ With this change, the lint step will detect the error in staging and halt the deploy. The localization can be fixed in Pontoon, the new master deployed to staging, and the same commit deployed to production after testing.
3. __The `.po` file has a bad header.__ I'm adding ``--check-header`` to ``locale/compile-mo.sh`` so this will be caught as well. It's outputting a bunch of warnings, though, so I need to make some changes:

```
./af/LC_MESSAGES/django.po:2: header field 'Project-Id-Version' still has the initial default value
./af/LC_MESSAGES/javascript.po:2: header field 'Project-Id-Version' still has the initial default value
./af/LC_MESSAGES/promote-mdn.po:4: header field 'Project-Id-Version' still has the initial default value
```
I'll get through these and add a commit updating the headers.

jwhitlock:
Modified the headers so that ``msgfmt --check-header`` doesn't complain. There were a lot of headers to change, so I [scripted it](https://gist.github.com/jwhitlock/238de49e43c896cc9d36745306d02dd0).  Sorry that the PR is exploding.

scripts/chief_deploy.py
```diff
@@ -51,6 +51,7 @@ def update_product_details(ctx):
 @task
 def update_locales(ctx):
     with ctx.lcd(os.path.join(settings.SRC_DIR, 'locale')):
+        ctx.local("dennis-cmd lint --errorsonly .")
```
willkg: If `dennis-cmd` returns a non-zero status code, does that stop the deploy?

jwhitlock: I think so. Here's [a recent stage build log](http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/3b6adc1d9e543d9aa575db8cb6dd6d64cedf47d0.1462472097), where the build stopped because ``polib`` wasn't installed. I think a similar build stop would happen if a non-zero exit code was returned.

willkg: Awesome! That's a little different than what we did with SUMO and Input, but I think that's fine.

Does it make sense to change `locale/compile.sh` to die with a non-zero status code if compiling ever failed.

jwhitlock: The ``set -e`` line means that if any command returns non-zero, the script exits immediately with non-zero. I'm not sure when ``msgfmt`` returns non-zero for a header issue, but it does for bad strings.

willkg: My bad--I missed that file change. It's buried in all the `.po` file changes.

willkg:
This looks good to me as far as I can tell! r+

PR #3861, opened 2016-05-09
1268040 dashboard filter known authors

dchukhin:
https://bugzilla.mozilla.org/show_bug.cgi?id=1268040

This pull request creates an extra filter for authors on the revisions dashboard, allowing authors to be filtered into "All Authors", "Known Authors" (part of the Known Authors group), and "Unknown Authors" (excluding everyone in the Known Authors group).

### DIFF
--- a/kuma/dashboards/forms.py
+++ b/kuma/dashboards/forms.py
@@ -13,6 +13,11 @@ PERIOD_CHOICES = [
     ('week', _('Week')),
     ('month', _('30 days')),
 ]
+AUTHOR_CHOICES = [
+    ('All Authors', _('All Authors')),
+    ('Known Authors', _('Known Authors')),
+    ('Unknown Authors', _('Unknown Authors')),
+]
 
 
 class RevisionDashboardForm(forms.Form):
@@ -42,3 +47,7 @@ class RevisionDashboardForm(forms.Form):
         choices=PERIOD_CHOICES,
         required=False,
         label=_(u'Preceding Period:'))
+    authors = forms.ChoiceField(
+        choices=AUTHOR_CHOICES,
+        required=False,
+        label=_(u'Authors'))
--- a/kuma/dashboards/forms.py
+++ b/kuma/dashboards/forms.py
@@ -13,10 +13,13 @@ PERIOD_CHOICES = [
     ('week', _('Week')),
     ('month', _('30 days')),
 ]
+ALL_AUTHORS = 0
+KNOWN_AUTHORS = 1
+UNKNOWN_AUTHORS = 2
 AUTHOR_CHOICES = [
-    ('All Authors', _('All Authors')),
-    ('Known Authors', _('Known Authors')),
-    ('Unknown Authors', _('Unknown Authors')),
+    (ALL_AUTHORS, _('All Authors')),
+    (KNOWN_AUTHORS, _('Known Authors')),
+    (UNKNOWN_AUTHORS, _('Unknown Authors')),
 ]
 
 
@@ -51,3 +54,12 @@ class RevisionDashboardForm(forms.Form):
         choices=AUTHOR_CHOICES,
         required=False,
         label=_(u'Authors'))
+
+    ALL_AUTHORS = 0
+    KNOWN_AUTHORS = 1
+    UNKNOWN_AUTHORS = 2
+    AUTHOR_CHOICES = [
+        (ALL_AUTHORS, _('All Authors')),
+        (KNOWN_AUTHORS, _('Known Authors')),
+        (UNKNOWN_AUTHORS, _('Unknown Authors')),
+    ]
--- a/kuma/dashboards/forms.py
+++ b/kuma/dashboards/forms.py
@@ -13,17 +13,18 @@ PERIOD_CHOICES = [
     ('week', _('Week')),
     ('month', _('30 days')),
 ]
-ALL_AUTHORS = 0
-KNOWN_AUTHORS = 1
-UNKNOWN_AUTHORS = 2
-AUTHOR_CHOICES = [
-    (ALL_AUTHORS, _('All Authors')),
-    (KNOWN_AUTHORS, _('Known Authors')),
-    (UNKNOWN_AUTHORS, _('Unknown Authors')),
-]
 
 
 class RevisionDashboardForm(forms.Form):
+    ALL_AUTHORS = 0
+    KNOWN_AUTHORS = 1
+    UNKNOWN_AUTHORS = 2
+    AUTHOR_CHOICES = [
+        (ALL_AUTHORS, _('All Authors')),
+        (KNOWN_AUTHORS, _('Known Authors')),
+        (UNKNOWN_AUTHORS, _('Unknown Authors')),
+    ]
+
     locale = forms.ChoiceField(
         choices=LANG_CHOICES,
         # Required for non-translations, which is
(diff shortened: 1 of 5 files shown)


safwanrahman:
I think a migration is needed for creating the group automatically while deploy.

dchukhin:
Sure, I can add a "Known Authors" group. Which users should be in that group? Or should I leave it empty and let someone else fill it with users?

safwanrahman:
So I think, its better to keep the past code structure so that we can add more functionality in future.

1. As you have added a ``exclude_kwargs`` for excluding, you should add a map for that like the [``query_kwargs_map``](https://github.com/mozilla/kuma/pull/3861/files#diff-b7bfb3dea52286266cf01a6b43acb37dR39), So in the future we can re use the map for other purposes also

2. It should maintain like previous style so that we can re use the functionality like [this](https://github.com/mozilla/kuma/pull/3861/files#diff-b7bfb3dea52286266cf01a6b43acb37dR47)

kuma/dashboards/forms.py
```diff
@@ -13,6 +13,11 @@
     ('week', _('Week')),
     ('month', _('30 days')),
 ]
+AUTHOR_CHOICES = [
+    ('All Authors', _('All Authors')),
```
safwanrahman: better to have the choice with underscore like ``('all_authors', _('All Authors')),``

safwanrahman:
> should I leave it empty and let someone else fill it with users?
:+1:

kuma/dashboards/forms.py
jwhitlock: Django will handle the space or the underscore, although I do like avoiding the automatic space conversion to ``All+Authors``.

I would prefer class-level constants, such as:

```python
ALL_AUTHORS=0
KNOWN_AUTHORS=1
UNKNOWN_AUTHORS=2
AUTHOR_CHOICES = [
    (ALL_AUTHORS, _('All Authors')),
    (KNOWN_AUTHORS, _('Known Authors')),
    (UNKNOWN_AUTHORS, _('Unknown Authors')),
]
```
and the usage in the view becomes ``RevisionDashboardForm.ALL_AUTHORS``.

kuma/dashboards/views.py
```diff
@@ -71,8 +73,19 @@ def revisions(request):
             start_date = end_date - datetime.timedelta(seconds=seconds)
             query_kwargs['created__range'] = [start_date, end_date]
 
-    if query_kwargs:
-        revisions = revisions.filter(**query_kwargs)
+        authors_filter = filter_form.cleaned_data['authors']
+        if authors_filter not in ['', 'All Authors']:
```
jwhitlock: If the requester asks for a username, I think they should get it, and skip the known/unknown authors filter

jwhitlock:
For future commits (or these if you rebase), use "bug 1268040 - short description" for the first line.  See [CONTRIBUTING.md](https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md) for other conventions.

I'm neutral on adding the "Known Authors" group. We don't usually add groups in migrations unless we also populate them, like [0030_add_page_creators.py](https://github.com/mozilla/kuma/blob/master/kuma/wiki/migrations/0030_add_page_creators_group.py#L23-L26). If you add the group, don't populate it.

The known/unknown author filter doesn't fit the pattern of ``query_kwargs_map``, don't try to make it fit or pre-design for future use cases.

In my VM I timed the SQL requests using the anonymized database. With no "Known Authors" group:

* Without a "Known Authors" group:
  * All Authors - 1291 ms, 1203 ms in the revision count query
  * Known Authors - 53 ms, 23 ms in the revision count query
  * Unknown Authors - 1391 ms total, 1275 in the revision count query
* With a "Known Authors" group with one staff member:
  * All Authors - 1302 ms, 1188 for revision count query
  * Known Authors - 2999 ms, 2841 ms in the revision count query
  * Unknown Authors - 1342 ms, 1188 for the revision count query

This is just about the same for the most common queries (All Authors, Unknown Authors), and significantly slower for the unlikely query (Known Authors).  A possible improvement would be to query for the group ID, and filter on that instead of the name, but since the common case is OK, it isn't required.

kuma/users/fixtures/test_users.json
```diff
@@ -32,6 +32,14 @@
     },
     {
         "fields": {
+            "name": "Trusted writers",
+            "permissions": []
+        },
+        "model": "auth.group",
+        "pk": 3
```
jwhitlock: ``"pk": 3`` is used by the ``Beta Testers`` group earlier in this file, which is probably why the test is failing in TravisCI and in the VM (``py.test kuma/users/tests/test_views.py::UserViewsTest::test_user_edit_beta``)

jwhitlock:
@dchukhin, back to you:

* [ ] Fix the test fixtures so all tests pass
* [ ] Filtering by username implies "All Authors"
* [ ] (Optional) Rename old commits to ``bug 1268040 - (short description)`` style
* [ ] (Optional) Add a data migration to create an empty "Known Authors" group
* [ ] (Optional) Convert choices to remove space and/or treat as constants
* [ ] (Very Optional) Filter by group ID instead of group name

dchukhin:
Ok, should be ready for re-review @jwhitlock 

- [x] Fix the test fixtures so all tests pass
- [x] Filtering by username implies "All Authors"
- [ ] Rename old commits to bug 1268040 - (short description) style (Optional)
- [ ] Add a data migration to create an empty "Known Authors" group (Optional)
- [x] Convert choices to remove space and/or treat as constants (Optional)
- [x] Filter by group ID instead of group name (Very Optional)

kuma/dashboards/forms.py
```diff
@@ -42,3 +50,16 @@ class RevisionDashboardForm(forms.Form):
         choices=PERIOD_CHOICES,
         required=False,
         label=_(u'Preceding Period:'))
+    authors = forms.ChoiceField(
+        choices=AUTHOR_CHOICES,
+        required=False,
+        label=_(u'Authors'))
+
+    ALL_AUTHORS = 0
```
safwanrahman: Duplication maybe

kuma/dashboards/forms.py
```diff
@@ -13,6 +13,14 @@
...
     ('month', _('30 days')),
 ]
+ALL_AUTHORS = 0
+KNOWN_AUTHORS = 1
+UNKNOWN_AUTHORS = 2
+AUTHOR_CHOICES = [
+    (ALL_AUTHORS, _('All Authors')),
+    (KNOWN_AUTHORS, _('Known Authors')),
+    (UNKNOWN_AUTHORS, _('Unknown Authors')),
+]
```
jwhitlock: This seems to be a duplicate of the same data in RevisionDashboardForm. I think this one is the one to remove.

kuma/dashboards/views.py
```diff
@@ -71,8 +74,23 @@ def revisions(request):
             start_date = end_date - datetime.timedelta(seconds=seconds)
             query_kwargs['created__range'] = [start_date, end_date]
 
-    if query_kwargs:
-        revisions = revisions.filter(**query_kwargs)
+        authors_filter = filter_form.cleaned_data['authors']
+        if (not filter_form.cleaned_data['user'] != '' and
+           authors_filter not in ['', str(RevisionDashboardForm.ALL_AUTHORS)]):
+
```
jwhitlock: This feels awkward, but I don't have strong suggestions to improve it. Filters are just awkward.

jwhitlock:
New timing with one Known Author:

* All Authors: 1290ms, 1194ms in count (same)
* Known Authors: 766ms, 634ms in count (3x faster)
* Unknown Authors: 1401ms, 1315ms in count (slightly slower)

So, I think this is good enough until we restructure this page for performance.  Thanks for the change.

Searching for my known author with the Unknown Author filter works.

The only remaining required change is:
* [ ] Remove the duplicate ``AUTHOR_CHOICES``

emullaney:
@jwhitlock thanks for the feedback. I've removed the duplicated choices code. 

[x] Remove the duplicate AUTHOR_CHOICES

jwhitlock:
Thank you @dchukhin and @emullaney, and thanks @safwanrahman for taking a look.

PR #3863, opened 2016-05-11
1268642 mark spam/ham with one click (Akismet submission)

emullaney:
https://bugzilla.mozilla.org/show_bug.cgi?id=1268642

- Replaced Django admin links for Spam/Ham with form buttons that call a new view **submit_akismet_spam**
- New post view **submit_akismet_spam** that creates the spam/ham RevisionAkismetSubmission record with one click from the dashboard page. This new view checks that the user has the correct Akismet permissions before creating this record and redirecting back to the same dashboard page that they were on.
- Tests for this new view
- New model method on kuma.models.users.models.User, checks for Akismet permission: has_akismet_submission_permission
  - This new method used in dashboard view to determine show_spam_submission

Possible TODO's:
- Switch Spam/Ham submission to AJAX view. Right now it's a post with a redirect.
- Look into adding messages to the user for errors (ie. no Akismet permission) and success (ie. Spam submission sent). Might need more guidance on how & where to add these.

### DIFF
--- a/kuma/dashboards/tests/test_views.py
+++ b/kuma/dashboards/tests/test_views.py
@@ -152,10 +152,7 @@ class RevisionsDashTest(UserTestCase):
             'revision': revision.pk,
             'submit': u'spam',
         }
-        p1 = Permission.objects.get(codename='add_revisionakismetsubmission')
-        testuser = User.objects.get(username='testuser')
-        testuser.user_permissions.add(p1)
-        self.client.login(username='testuser', password='testpass')
+        self.client.login(username='admin', password='testpass')
 
         # Response should redirect back to the revisions dash
         urlpost = reverse('dashboards.submit_akismet_spam', locale='en-US')
@@ -166,6 +163,25 @@ class RevisionsDashTest(UserTestCase):
         # 1 RevisionAkismetSubmission record should exist for this revision
         ras = RevisionAkismetSubmission.objects.get(revision=revision)
         eq_(ras.type, u'spam')
+
+    def test_submit_akismet_spam_revision_dne(self):
+        urlnext = reverse('dashboards.revisions')
+        revision_dne = '9999999'
+        data = {
+            'revision': revision_dne,
+            'submit': u'spam',
+        }
+        self.client.login(username='admin', password='testpass')
+
+        # Response should redirect back to the revisions dash
+        urlpost = reverse('dashboards.submit_akismet_spam', locale='en-US')
+        response = self.client.post(urlpost, data=data)
+        eq_(response.status_code, 302)
+        eq_(response.url, 'http://testserver' + urlnext)
+
+        # Zero RevisionAkismetSubmission records should exist for this nonexistent revision
+        ras = RevisionAkismetSubmission.objects.filter(revision__pk=revision_dne)
+        eq_(ras.count(), 0)
 
     def test_locale_filter(self):
         url = urlparams(reverse('dashboards.revisions', locale='fr'),
--- a/kuma/dashboards/tests/test_views.py
+++ b/kuma/dashboards/tests/test_views.py
@@ -94,7 +94,8 @@ class RevisionsDashTest(UserTestCase):
 
         page = pq(response.content)
         ip_button = page.find('td.dashboard-spam')
-        ok_(len(ip_button) > 0)
+        # User must be logged for spam button column to be visible
+        eq_(len(ip_button), 0)
 
     def test_submit_akismet_spam_post_required(self):
         url = reverse('dashboards.submit_akismet_spam', locale='en-US')
--- a/kuma/dashboards/tests/test_views.py
+++ b/kuma/dashboards/tests/test_views.py
@@ -123,9 +123,8 @@ class RevisionsDashTest(UserTestCase):
         eq_(response.url, 'http://testserver' + urlnext)
 
         # 1 RevisionAkismetSubmission record should exist for this revision
-        ras = RevisionAkismetSubmission.objects.filter(revision=revision)
-        eq_(ras.count(), 1)
-        eq_(ras[0].type, u'spam')
+        ras = RevisionAkismetSubmission.objects.get(revision=revision)
+        eq_(ras.type, u'spam')
 
     def test_submit_akismet_spam_no_permission(self):
         urlnext = reverse('dashboards.revisions', locale='en-US')
@@ -145,6 +144,28 @@ class RevisionsDashTest(UserTestCase):
         # No RevisionAkismetSubmission record should exist, user does not have permission
         ras = RevisionAkismetSubmission.objects.filter(revision=revision)
         eq_(ras.count(), 0)
+
+    def test_submit_akismet_spam_no_url_in_next_variable(self):
+        urlnext = reverse('dashboards.revisions')
+        revision = Revision.objects.first()
+        data = {
+            'revision': revision.pk,
+            'submit': u'spam',
+        }
+        p1 = Permission.objects.get(codename='add_revisionakismetsubmission')
+        testuser = User.objects.get(username='testuser')
+        testuser.user_permissions.add(p1)
+        self.client.login(username='testuser', password='testpass')
+
+        # Response should redirect back to the revisions dash
+        urlpost = reverse('dashboards.submit_akismet_spam', locale='en-US')
+        response = self.client.post(urlpost, data=data)
+        eq_(response.status_code, 302)
+        eq_(response.url, 'http://testserver' + urlnext)
+
+        # 1 RevisionAkismetSubmission record should exist for this revision
+        ras = RevisionAkismetSubmission.objects.get(revision=revision)
+        eq_(ras.type, u'spam')
 
     def test_locale_filter(self):
         url = urlparams(reverse('dashboards.revisions', locale='fr'),
--- a/kuma/dashboards/tests/test_views.py
+++ b/kuma/dashboards/tests/test_views.py
@@ -1,3 +1,4 @@
+import urllib
 import pytest
 from pyquery import PyQuery as pq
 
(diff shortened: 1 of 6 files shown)


kuma/users/models.py
```diff
@@ -194,3 +194,7 @@ def wiki_revisions(self, count=5):
 
     def allows_editing_by(self, user):
         return user.is_staff or user.is_superuser or user.pk == self.pk
+
+    def has_akismet_submission_permission(self):
+        return (self.groups.filter(permissions__codename=u'add_revisionakismetsubmission').count() or
```
safwanrahman: I think this should be ``exists()`` instead of ``count()``

emullaney:
I just realized that the dashboards view doesn't require login. That's what's causing my test errors, and it's going to require a little bit of a refactor around checking to see if the user is logged in before checking permissions for Akismet.

kuma/users/models.py
emullaney: Makes sense. Updated.

kuma/users/models.py
```diff
@@ -194,3 +194,7 @@ def wiki_revisions(self, count=5):
 
     def allows_editing_by(self, user):
         return user.is_staff or user.is_superuser or user.pk == self.pk
+
+    def has_akismet_submission_permission(self):
+        return (self.groups.filter(permissions__codename=u'add_revisionakismetsubmission').exists() or
+                 self.user_permissions.filter(codename=u'add_revisionakismetsubmission').exists())
```
jwhitlock: Please use this instead: ``has_perm("wiki.add_revisionakismetsubmission")``, which does the equivalent and more.

kuma/dashboards/tests/test_views.py
```diff
@@ -89,7 +93,39 @@ def test_spam_submission_buttons(self):
...
+
+    def test_submit_akismet_spam_valid_response(self):
+        url = reverse('dashboards.submit_akismet_spam', locale='en-US')
+        revision = Revision.objects.first()
+        data = {
+            'revision': revision.pk,
+            'submit': u'spam',
+        }
+        ct = ContentType.objects.get(app_label='attachments',
+                                     model='attachment')
```
jwhitlock: The app_label and model should be ``'wiki'``.

kuma/dashboards/views.py
```diff
@@ -130,3 +131,19 @@ def topic_lookup(request):
     data = json.dumps(topiclist)
     return HttpResponse(data,
                         content_type='application/json; charset=utf-8')
+
+
+@require_POST
+def submit_akismet_spam(request):
```
jwhitlock: The permissions checks could be replaced with a [@permission_required() decorator](https://docs.djangoproject.com/en/1.9/topics/auth/default/#the-permission-required-decorator):

```
@permission_required('wiki.add_revisionakismetsubmission')
```

kuma/dashboards/tests/test_views.py
```diff
@@ -89,7 +93,39 @@ def test_spam_submission_buttons(self):
...
+        p1 = Permission.objects.create(
+            name='Can add Akismet submission',
+            codename='add_revisionakismetsubmission',
+            content_type=ct)
+        admin = User.objects.get(username='admin')
```
jwhitlock: The permission was already created by the test setup, and can be replaced by:

```p1 = Permission.objects.get(codename='add_revisionakismetsubmission')```

Also, ``admin`` is a superuser, so when you change to ``has_perm`` or similar, the user will have the permission, and no adding of the permission is needed.

**Comment by jwhitlock** [comment]
I like this approach. It is simpler than the AJAX-based version I was envisioning, but works well and doesn't block turning it into an AJAX request later if needed.

A few things to do:
* [ ] Use the ``@permission_required()`` method or equivalent ``has_perm`` changes
* [ ] Update the test to avoid creating or manually adding the permission to the admin
* [ ] Add tests to exercise the view branches. After the permission change, it may just be including a ``next`` parameter, and checking the redirect.
* [ ] Fix flake8. You can run ``flake8 kuma`` in your VM to check locally.

**Comment by emullaney** [comment]
@jwhitlock, thanks very much for the very helpful feedback. Let me know if any further changes are needed.

[x] Use the @permission_required() method or equivalent has_perm changes
[x] Update the test to avoid creating or manually adding the permission to the admin (I used the testuser account in one test, to ensure that the permission was checked correctly).
[x] Add tests to exercise the view branches. After the permission change, it may just be including a next parameter, and checking the redirect.
[x] Fix flake8. You can run flake8 kuma in your VM to check locally.

File: kuma/dashboards/tests/test_views.py [hunk 6]
```diff
@@ -89,8 +93,58 @@ def test_spam_submission_buttons(self):
...
+        self.client.login(username='testuser', password='testpass')
+
+        # Response should redirect back to the revisions dash
+        urlpost = reverse('dashboards.submit_akismet_spam', locale='en-US')
+        response = self.client.post(urlpost, data=data)
+        eq_(response.status_code, 302)
+        eq_(response.url, 'http://testserver' + urlnext)
+
+        # 1 RevisionAkismetSubmission record should exist for this revision
+        ras = RevisionAkismetSubmission.objects.filter(revision=revision)
```
safwanrahman:
```python
ras = RevisionAkismetSubmission.objects.get(revision=revision)
eq_(ras.type, u'spam')
```

As ``get()`` will always raise issue if there are more than one object

safwanrahman:
So Submitting via Ajax is needed. Currently, it reloads the page and makes it hard to submit many spam revision.

kuma/dashboards/tests/test_views.py
jwhitlock: @safwanrahman's right, and now I can't un-see it.  If you have time Monday morning, please fix, but I'll merge and deploy before the standup on Monday with or without it.  There's worse than this in the existing tests...

kuma/dashboards/views.py
```diff
@@ -130,3 +132,19 @@ def topic_lookup(request):
...
                         content_type='application/json; charset=utf-8')
+
+
+@require_POST
+@permission_required('wiki.add_revisionakismetsubmission')
+def submit_akismet_spam(request):
+    """Creates SPAM or HAM Akismet record for revision"""
+    url = request.POST.get('next')
+    if url is None or not is_safe_url(url, request.get_host()):
+        url = reverse('dashboards.revisions')
```
jwhitlock: One more test is needed to exercise this line: the user has the permission, but hasn't set the next URL.

jwhitlock:
The non-test code is ready, but there's a missing test to exercise the case where the next URL is None or invalid.

jwhitlock:
@safwanrahman, AJAX is an obvious improvement for next iteration. However, I'm not requiring it for this PR, because the current spam vs good edits ratio is tiny.

safwanrahman:
Ok Then.

kuma/dashboards/views.py
```diff
@@ -130,3 +132,19 @@ def topic_lookup(request):
...
+    revision = request.POST.get('revision', 0)
```
safwanrahman: Its not a good idea to return ``0`` as the revision id if there is not any ``revision`` object present.

kuma/dashboards/views.py
```diff
@@ -130,3 +132,19 @@ def topic_lookup(request):
...
+    revision = Revision.objects.get(pk=revision)
```
safwanrahman: In the case a revision is not present, it will raise Error

kuma/dashboards/views.py
emullaney: I'll add this test now. If the URL is empty or invalid, the user will still be able to submit spam successfully, it will just go back to the dashboard with no filters/page set.

kuma/dashboards/views.py
emullaney: Ok, I can remove the 0 and leave no default unless you have a different suggestion.

kuma/dashboards/views.py
emullaney: Got it. I'm adding a try/except here and a test case for a nonexistent Revision.

kuma/dashboards/views.py
```diff
@@ -148,3 +150,23 @@ def topic_lookup(request):
...
+
+
+@require_POST
+@permission_required('wiki.add_revisionakismetsubmission')
+def submit_akismet_spam(request):
+    """Creates SPAM or HAM Akismet record for revision"""
+    url = request.POST.get('next')
+    if url is None or not is_safe_url(url, request.get_host()):
+        url = reverse('dashboards.revisions')
+    revision = request.POST.get('revision')
```
safwanrahman: I think, this should also validate that the input is an integer.

kuma/dashboards/views.py
```diff
@@ -148,3 +150,23 @@ def topic_lookup(request):
...
+    try:
```
safwanrahman: I think, the entire section rewrite can be like this

```python
try:
    revision_id = int(request.POST.get('revision'))
except (ValueError, TypeError) as e:
    return redirect(url)

revision = Revision.objects.filter(id=revision_id)
submission_type = request.POST.get('submit')
if revision.exists() and submission_type in ['spam', 'ham']:
    RevisionAkismetSubmission.objects.create(
        sender=request.user, revision_id=revision_id, type=submission_type)

return redirect(url)

jwhitlock:
@safwanrahman this is a good catch. However, to get to this point you would need:

1) To have been vetted and placed in the Spam Admins group (permissions check)
2) Be manually POSTing to the endpoint (instead of following the link on the revision dashboard, and 
3) Decided to use a string like "THIS" instead of a revision ID

It is slightly different than the "unknown revision" problem, because the database may be in an inconsistent state due to transactions. Even that is very unlikely - once a revision is visible on the revision dashboard, it's very unlikely it is visible to some but not all viewers.

This does identify a problem with writing request validation logic - it's easy to miss corner cases. For similar future changes, we should use a [ModelForm](https://docs.djangoproject.com/en/1.8/topics/forms/modelforms/#modelform) and lean on Django's deep experience of validating user input.  This would take care of validating IDs, submission types, etc. etc.

PR #3864, opened 2016-05-16
[Bug 1268642] Mark spam/ham with one click with ajax

safwanrahman:
This will make the spam submission through ajax request. It also fixes the Resubmit Button.

I will fix the test failors and add more tests. till then anyone can test it and let me know if more improvement is needed.

### DIFF
--- a/kuma/core/utils.py
+++ b/kuma/core/utils.py
@@ -1,3 +1,4 @@
+import datetime
 try:
     import urlparse
 except ImportError:
@@ -11,6 +12,7 @@ import random
 import time
 from itertools import islice
 
+from babel import dates, localedata
 import bitly_api
 from celery import chain, chord
 from django.conf import settings
@@ -19,10 +21,13 @@ from django.http import QueryDict
 from django.shortcuts import _get_queryset
 from django.utils.encoding import force_unicode, smart_str
 from django.utils.http import urlencode
+from django.utils.translation import ugettext_lazy as _
 from polib import pofile
+from pytz import timezone
 from taggit.utils import split_strip
 
 from .cache import memcache
+from .exceptions import DateTimeFormatError
 from .jobs import IPBanJob
 
 
@@ -366,3 +371,80 @@ def urlparams(url_, fragment=None, query_dict=None, **query):
     new = urlparse.ParseResult(url_.scheme, url_.netloc, url_.path,
                                url_.params, query_string, fragment)
     return new.geturl()
+
+
+def format_date_time(request, value, format='shortdatetime'):
+    """
+    Returns date/time formatted using babel's locale settings. Uses the
+    timezone from settings.py
+    """
+    if not isinstance(value, datetime.datetime):
+        if isinstance(value, datetime.date):
+            # Turn a date into a datetime
+            value = datetime.datetime.combine(value,
+                                              datetime.datetime.min.time())
+        else:
+            # Expecting datetime value
+            raise ValueError
+
+    default_tz = timezone(settings.TIME_ZONE)
+    tzvalue = default_tz.localize(value)
+
+    user = request.user
+    try:
+        if user.is_authenticated() and user.timezone:
+            user_tz = timezone(user.timezone)
+            tzvalue = user_tz.normalize(tzvalue.astimezone(user_tz))
+    except AttributeError:
+        pass
+
+    locale = _babel_locale(_get_request_locale(request))
+
+    try:
+        formatted = format_date_value(value, tzvalue, locale, format)
+    except KeyError:
+        # Babel sometimes stumbles over missing formatters in some locales
+        # e.g. bug #1247086
+        # we fall back formatting the value with the default language code
+        formatted = format_date_value(value, tzvalue,
+                                      _babel_locale(settings.LANGUAGE_CODE),
+                                      format)
+
+    return formatted, tzvalue
+
+
+def _get_request_locale(request):
+    """Return locale from the request, falling back to a default if invalid."""
+    locale = request.LANGUAGE_CODE
+    if not localedata.exists(locale):
+        locale = settings.LANGUAGE_CODE
+    return locale
+
+
+def format_date_value(value, tzvalue, locale, format):
+    if format == 'shortdatetime':
+        # Check if the date is today
+        if value.toordinal() == datetime.date.today().toordinal():
+            formatted = dates.format_time(tzvalue, format='short',
+                                          locale=locale)
+            return _(u'Today at %s') % formatted
+        else:
+            return dates.format_datetime(tzvalue, format='short',
+                                         locale=locale)
+    elif format == 'longdatetime':
+        return dates.format_datetime(tzvalue, format='long', locale=locale)
+    elif format == 'date':
+        return dates.format_date(tzvalue, locale=locale)
+    elif format == 'time':
+        return dates.format_time(tzvalue, locale=locale)
+    elif format == 'datetime':
+        return dates.format_datetime(tzvalue, locale=locale)
+    else:
+        # Unknown format
+        raise DateTimeFormatError
+
+
+def _babel_locale(locale):
+    """Return the Babel locale code, given a normal one."""
+    # Babel uses underscore as separator.
+    return locale.replace('-', '_')
--- a/kuma/dashboards/jinja2/dashboards/revisions.html
+++ b/kuma/dashboards/jinja2/dashboards/revisions.html
@@ -133,7 +133,7 @@
         var $detail;
 
         // Don't interrupt links or spam buttons, and stop if a request is already running
-        if (e.target.tagName == 'A' || e.target.name == 'submit' || $this.attr('data-running')) return;
+        if (e.target.tagName == 'A' || e.target.id == 'spam-ham-button' || $this.attr('data-running')) return;
 
         if ($this.attr('data-loaded')) {
             $this.next('.dashboard-detail').find('.dashboard-detail-details').slideToggle();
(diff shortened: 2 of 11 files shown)


safwanrahman:
Fixed the tests. Will add some more tests. till then @jwhitlock review?

jwhitlock:
I haven't had time to dig into this yet, just gave it a quick look.  It may be another few days before I can run some practical tests in my VM.

Moving ``format_date_value`` to ``core/utils.py`` _might_ be a good move, but I'd prefer it was in it's own commit, and ensured there was sufficient test coverage. I'll have to spend some time making sure you did the move correctly.

I'd be more comfortable using a Django ``ModelForm`` for validation, than continuing to add corner case checks for revision IDs and submission types.

jwhitlock:
By the way, my comments are from before you added a new test.

safwanrahman:
Oh! I will take a look on this!

safwanrahman:
@jwhitlock Without the test of utils, I think this patch is ready for review.
I am writing test for the utils in the meantime.

I have decided to move the API from ``dashboard`` to ``wiki`` because I think ``wiki`` is the better place for the API. In future, we can use the API outside of Dashboard.

kuma/core/utils.py
```diff
@@ -366,3 +371,80 @@ def urlparams(url_, fragment=None, query_dict=None, **query):
...
+        return dates.format_datetime(tzvalue, locale=locale)
+    else:
+        # Unknown format
+        raise DateTimeFormatError
+
+
+def _babel_locale(locale):
+    """Return the Babel locale code, given a normal one."""
+    # Babel uses underscore as separator.
+    return locale.replace('-', '_')
```
jwhitlock: ``_babel_locale`` should be removed from jinja_helpers.py if it is being moved here

jwhitlock:
For future commits, please include the bug number in the first line of the commit message, like:

``bug 1268642 - Move datetimeformat jinja helpers to utils``

See https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md for other rules I've forgotten.

While reviewing your change, I realized the new non-admin code is missing the important step of submitting the data to Akismet, which was done in the form (see [AkismetSubmissionFormMixin](https://github.com/mozilla/kuma/blob/master/kuma/spam/forms.py#L102-L136)). I'm going to code a fix to add that, but it will probably collide with this change.

I don't think it is a good idea to allow multiple spam submissions in a single request. If one of ten fails, I'm not sure how you would report the failure and display it to the user. Maybe it is in the PR, but I had to stop the code review to fix this bug.

safwanrahman:
@jwhitlock So #3874 fixed the submission to Akismet and also add a form there. So do you think it should be better to rabse upon master and make the PR based on that form?

> I don't think it is a good idea to allow multiple spam submissions in a single request. 
It submits every spam submission with different ajax request and update the html according to it. So if anyone of them fails, it shows error message in the table.

jwhitlock:
I have not dug into the PR, and it will probably be next week until I do. I haven't had a chance to try out the AJAX for myself, just looking at the code.  So, all this feedback is not from a full code reading and exercising of the code.

Yes, a rebase is needed, and removal of the now duplicate ``_babel_locale``.  I'll leave it up to you if that is in this PR or a new one.

I have mixed feelings on moving this to the ``wiki`` app. On the one hand, it is only called from the dashboard, so it makes sense for it to live there, and minimizes the change set. On the other hand, the model and form live in ``wiki``. I will probably come around to your way of thinking by Monday.

We are designing a new process for moderating published spam, which will include removing this functionality from the revision dashboard. All of this code will probably be replaced in about a month. If you are OK with that, we can continue, otherwise we can find some non-spam bugs that will have a longer lifespan.

kuma/wiki/tests/test_views.py
```diff
@@ -4269,3 +4271,136 @@ def test_case_insensitive_tags(self):
...
+        eq_(response.status_code, 405)
+
+        # No RevisionAkismetSubmission record should exist, user does not have permission
+        ras = RevisionAkismetSubmission.objects.filter(revision=r)
+        eq_(ras.count(), 0)
+
+    def test_submit_akismet_spam_revision_dne(self):
+        r = self.revision
+        revision_dne = r.id
+        # Delete the revision so it does not exists anymore
```
jwhitlock: "does not exist" - drop the "s" from exists

kuma/dashboards/jinja2/dashboards/revisions.html
```diff
@@ -193,6 +193,46 @@
...
     });
 
+    // Send revision to Akismat for Spam or Ham
+    $(document).on('click', '#spam-ham-button', function() {
+        var $this = $(this),
+          $tdObject = $(this).parent(),
+          $trObject = $tdObject.parent(),
+          revisionId = $trObject.data("revisionId"),
+          type = this.value,
+          url = $trObject.data("spamUrl");
```
jwhitlock: Other variable declarations in this file use multiple ``var``, which I don't like either, but it would be consistent. If you are going to use a different variable assignment pattern, I'd like a 4-char indent so the start of the variable names line up.  This is my amateur JS knowledge though, so feel free to call a nitpick.

kuma/dashboards/jinja2/dashboards/revisions.html
```diff
@@ -193,6 +193,46 @@
         });
     });
 
+    // Send revision to Akismat for Spam or Ham
```
jwhitlock: "Akismet" not "Akismat"

jwhitlock:
Initial code reading done, will move on to testing in VM next. If you've decided to abandon this PR, please close it.

In future commits or in a rebase, please follow the [CONTRIBUTING](https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md) guidelines and start the summary line with "bug 1268642 -".

In the commit message for commit 3, it's "Template".

Also, I'm pretty impressed with the JavaScript from someone that says they are a backend dev.  Of course I'm barely a JS dev, so I may not have to eyes to see what's wrong with it.

jwhitlock:
It works very nicely, and the code coverage is good. Unfortunately, it also exposes a problem with how we're doing Akismet submissions.

As you've seen in PR #3870, we want to only have the "SPAM" option on the dashboard. The dashboard only contains published edits, that Akismet did not mark as spam. The only legitimate action from the dashboard is to tell Akismet it should have blocked it as "SPAM" - submitting "HAM" says "no, you did the right things", and is meaningless.  This code should do the same.

Once a edit has been submitted as SPAM, what are the possible next actions? We've already trained as spam, so "RESUBMIT AS SPAM" should go away.

That just leaves "RESUBMIT AS HAM". But, is this a valid action? I'm not sure what Akismet will do if something is submitted both as ham and as spam.  In addition, this leaves two objects in the database, one that claims a revision is spam, the other that it is ham.  I'm already writing code that only looks at the spam Akismet submissions - this code is wrong.

At this point, I think the right thing to do would be something like:

1. Convert the type of the original submission from Spam to "Replaced", or something similar. This new value shows that someone submitted the data to Akismet, and then picked something different.
2. Submit as ham to Akismet
3. Create a new ``AkismetRevisionSubmission`` to record the new action
3. Only display the most recent decision on the dashboard.

This is too complex for a very rare situation, one that I've only seen once in the last three months, when moderators were panicking due to the February spam attack and incorrectly submitted a single good edit as spam.  Manual cleanup was sufficient for that case.

My preference would be, for this PR:

1. Show the most recent ``AkismetRevisionSubmission`` on the dashboard
2. Remove all the "RESUBMIT" buttons
3. Forbid adding a second ``AkismetRevisionSubmission`` from the dashboard

As a secondary action (new work, new PR)
1. Find all ``AkismetRevisionSubmission`` rows that refer to the same revision
2. Delete all but the last
3. Add a constraint of a single ``AkismetRevisionSubmission`` per revision
4. Require developer intervention to undo a bad Akismet submission.

To continue, please fix the identified issues:
* [ ] Use the new [``RevisionAkismetSubmissionSpamForm``](https://github.com/mozilla/kuma/blob/master/kuma/wiki/forms.py#L843-L856)
* [ ] Remove the duplicate ``_babel_locale``
* [ ] Remove the "HAM" button
* [ ] Remove the "RESUBMIT AS ..." buttons
* [ ] Use ``bug 1268642 -`` in the commit subject
* [ ] Fix "does not exists" to "does not exist"
* [ ] Fix indenting or use multiple ``var`` statements in JavaScript
* [ ] Fix the third commit message, "Template" rather than "Tamplate"
* [ ] Fix the typo "Akismat" as "Akismet"

I'm OK with the view moving to the ``wiki`` app.  In the far future (2017 at least) I want to start splitting the code between "human" views and API views, but that's long after Docker, AWS, etc.

safwanrahman:
@jwhitlock Updated the patch as you have mentioned in last comment also update the commit messages.
r?

kuma/dashboards/views.py
```diff
@@ -2,19 +2,21 @@
...
 import waffle
 
 from kuma.core.utils import paginate
+<<<<<<< HEAD
 from kuma.wiki.forms import RevisionAkismetSubmissionSpamForm
 from kuma.wiki.models import Document, Revision, RevisionAkismetSubmission
+=======
+from kuma.wiki.models import Document, Revision
+
+>>>>>>> Create API and move the dashboard to wiki views
```
jwhitlock: This represents a bad merge. Please rebase and fix this.

kuma/dashboards/jinja2/dashboards/revisions.html
```diff
@@ -220,7 +218,7 @@
                 $dl.append(subMessage);
             })
 
-            $tdObject.html($dl).append(button);
+            $tdObject.html($dl);
```
jwhitlock: Is the empty ``<dl></dl>`` needed?

jwhitlock: Ah, I see, in the full file the ``$.each`` populates it

jwhitlock:
Thanks, it works for me. In the future, please make sure each commit also has good code, not just the final commit.

PR #3869, opened 2016-05-18
1246459 search

beanoagin:
`Fix bug 1239244 - Indicate current page in paginator

r=jezdez``This is the bare minimum amount of work required to switch from
nose/django-nose to py.test/pytest-django.

Several things to know:

1. I changed the SkipTest things to xfail since that's closer to what
   they are (i.e. "This thing fails and we don't know why--look at it
   later")

2. I removed the attr() things because they seemed arbitrary,
   inconsistent and not documented anywhere I could find. I'll make sure
   to note this in the email I send after this lands.

3. I reimplemented eq_ and ok_ rather than go through and change these
   to their py.test equivalents. We can change them over as a code
   cleanup step done on a future rainy day.`


jwhitlock:
Reported

PR #3870, opened 2016-05-18
bug 1272796 - Remove HAM option from dashboard

emullaney:
Refers to bug https://bugzilla.mozilla.org/show_bug.cgi?id=1272796

- Remove HAM option from form. 
- Don't check for the spam/ham option in the submit_akismet_spam view, default to spam. 
- Remove spam submit value from tests.


safwanrahman:
I think, its better to have #3864 get reviewed. The button can be changed as per needed.

jwhitlock:
While reviewing PR #3864, I realized that spam / ham was not actually being submitted to Akismet, because that functionality was done in the form's clean method in the admin.  I'm going to add that bit, but it will be easier to do when only "spam" is an option, so merging this.

PR #3872, opened 2016-05-18
Fix Bug 875062 - remove social sharing

stephaniehobson:
Remove social sharing buttons and related assets as criteria to keep were not met.


schalkneethling:
r+ 🍧

PR #3876, opened 2016-05-20
bug 1272790 - akismet submission in admin make revision link to public site

dchukhin:
This pull request changes the link in the Akimset submissions django admin view under “REVISION” from the revision’s admin page to the revision’s public page and updates the related test.


jwhitlock:
Thanks, that fixes it.

PR #3877, opened 2016-05-20
bug 1272793 - spam admins can ban users without django admin

dchukhin:
This pull request uses the previously-created-but-unused `users.ban_user` view. A GET to this url will display the UserBanForm (with only a 'reason' field) and several common reasons for banning users.
A POST to this url will attempt to save the `UserBan` and return the user to the profile page of the user being banned. Banning requires the `users.add_userban` permission.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1209,6 +1209,12 @@ CONSTANCE_CONFIG = dict(
         ]),
         "JSON array listing tag suggestions for documents"
     ),
+    COMMON_REASONS_TO_BAN_USERS=(
+        json.dumps([
+            'Spam', 'Profile Spam ', 'Sandboxing', 'Incorrect Translation', 'Penetration Testing',
+        ]),
+        "JSON array listing some common reasons to ban users",
+    ),
     SESSION_CLEANUP_CHUNK_SIZE=(
         1000,
         'Number of expired sessions to cleanup up in one go.',
--- a/kuma/users/jinja2/users/ban_user.html
+++ b/kuma/users/jinja2/users/ban_user.html
@@ -9,6 +9,7 @@
 {% block site_css %}
   {{ super() }}
   {% stylesheet 'users' %}
+  {% stylesheet 'user-banned' %}
 {% endblock %}
 
 {% block content %}
@@ -23,7 +24,7 @@
         <div class="column-4">
           <h5>{{ _('Choose a reason') }}:</h5>
           {% for reason in common_reasons %}
-            <div><a class="ban-common-reason">{{ _(reason) }}</a></div>
+            <div><button class="ban-common-reason">{{ _(reason) }}</button></div>
           {% endfor %}
         </div>
 
--- a/kuma/users/jinja2/users/ban_user.html
+++ b/kuma/users/jinja2/users/ban_user.html
@@ -18,7 +18,7 @@
     <section id="user-head" class="column-container column-container-reverse vcard user-head">
       <div class="column-10">
         <h1 class="user-title">
-          {{ _('Why are you banning') }} {{user_to_ban }}?
+          {{ _('Why are you banning %(user)s', user=user_to_ban) }}?
         </h1>
         <div class="column-4">
           <h5>{{ _('Choose a reason') }}:</h5>
@@ -30,6 +30,7 @@
         <h5>{{ _('or enter a reason in the text box below') }}:</h5>
         <form method="post" id="ban-user-form" action="{{ url('users.ban_user',user_to_ban.id) }}">
           {% csrf_token %}
+          {{ form.reason.errors }}
           <div>{{ form.reason }}</div>
           <input type="hidden" name="previous-url" value={{ previous_url }}></input>
           <button type="submit">{{ _('Ban user') }}</button>
--- a/kuma/users/jinja2/users/ban_user.html
+++ b/kuma/users/jinja2/users/ban_user.html
@@ -1,0 +1,53 @@
+{% extends "base.html" %}
+
+{% block body_attributes %}{% endblock %}
+
+{% block bodyclass %}user ban-user{% endblock %}
+
+{% block title %}Ban {{ user_to_ban }}{% endblock %}
+
+{% block site_css %}
+  {{ super() }}
+  {% stylesheet 'users' %}
+{% endblock %}
+
+{% block content %}
+<section id="content">
+<div class="wrap">
+  <section id="content-main" class="full">
+    <section id="user-head" class="column-container column-container-reverse vcard user-head">
+      <div class="column-10">
+        <h1 class="user-title">
+          {{ _('Why are you banning') }} {{user_to_ban }}?
+        </h1>
+        <div class="column-4">
+          <h5>{{ _('Choose a reason') }}:</h5>
+          {% for reason in common_reasons %}
+            <div><a class="ban-common-reason">{{ _(reason) }}</a></div>
+          {% endfor %}
+        </div>
+
+        <h5>{{ _('or enter a reason in the text box below') }}:</h5>
+        <form method="post" id="ban-user-form" action="{{ url('users.ban_user',user_to_ban.id) }}">
+          {% csrf_token %}
+          <div>{{ form.reason }}</div>
+          <input type="hidden" name="previous-url" value={{ previous_url }}></input>
+          <button type="submit">{{ _('Ban user') }}</button>
+        </form>
+      </div>
+    </section>
+  </section>
+</div>
+</section>
+{% endblock %}
+
+{% block js %}
+<script>
+(function (jQuery) {
+    'use strict';
+    $('.ban-common-reason').on('click', function(){
+      $('#id_reason').val($(this).html())
+    });
+})(jQuery);
+</script>
+{% endblock %}
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -8,6 +8,7 @@ from allauth.account.models import EmailAddress
 from allauth.socialaccount.models import SocialAccount, SocialApp
 from allauth.socialaccount.providers import registry
 from allauth.tests import MockedResponse, mocked_response
+from constance import config as constance_config
 from constance.test.utils import override_config
 from django.conf import settings
 from django.contrib.auth.hashers import UNUSABLE_PASSWORD_PREFIX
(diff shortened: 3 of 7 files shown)


kuma/users/views.py
```diff
@@ -64,11 +64,17 @@ def ban_user(request, user_id):
                           is_active=True)
             ban.save()
             return redirect(user)
+    if user.active_ban:
+        return redirect(user)
     form = UserBanForm()
+    # A list of common reasons for banning a user
+    common_reasons = ['Spam', 'Profile Spam ', 'Sandboxing',
```
safwanrahman: This Should be configurable through ``constants``

safwanrahman:
There should be a test for the template also.

kuma/users/views.py
safwanrahman: Is it possible to pass it through the ``forms``

safwanrahman:
Another Test should be added that "The user is banned after making a http request to the views"

kuma/users/jinja2/users/ban_user.html
```diff
@@ -0,0 +1,53 @@
...
+{% endblock %}
+
+{% block content %}
+<section id="content">
+<div class="wrap">
+  <section id="content-main" class="full">
+    <section id="user-head" class="column-container column-container-reverse vcard user-head">
+      <div class="column-10">
+        <h1 class="user-title">
+          {{ _('Why are you banning') }} {{user_to_ban }}?
```
jwhitlock: Please use [newstyle gettext](http://jinja.pocoo.org/docs/dev/extensions/#newstyle-gettext) on the complete sentence, to handle languages with different word order, leading '¿' in Spanish, etc.

```
{{ _('Why are you banning %(user)s?', user=user_to_ban) }}
```

jwhitlock:
@safwanrahman I'm not sure what you mean by "There should be a test for the template also."  The status code tests seem sufficient.

This builds on an existing, unused view, which does include a test that [a POST creates a user ban](https://github.com/mozilla/kuma/blob/master/kuma/users/tests/test_views.py#L65-L85)

safwanrahman:
>  "There should be a test for the template also."

It should have a test that checks that the template and the common reasons are rendered correctly. As a new template is added.

kuma/users/views.py
jwhitlock: I think the mechanism (a ``click`` event on an ``<a>`` tag) works, and adding this through the form would be a complication. It works like I'd expect in VM testing.

It is a good point that this value should be customizable. This is the part of this that I most expect to be adjusted once more people see it in production (no one comments on bugs, everyone has an opinion once they see it). Let's save the future PR and move the reasons to constance. [WIKI_DOCUMENT_TAG_SUGGESTIONS](https://github.com/mozilla/kuma/blob/master/kuma/settings/common.py#L1197-L1210) is a good model (``USER_BAN_REASONS`` for the name?).  You'll need to decode the JSON to turn it into a Python list again, and default to ['Spam'] if it is empty or has a decoding error (to avoid the alternate logic of no suggestions).

kuma/users/jinja2/users/ban_user.html
```diff
@@ -0,0 +1,53 @@
...
+          <h5>{{ _('Choose a reason') }}:</h5>
+          {% for reason in common_reasons %}
+            <div><a class="ban-common-reason">{{ _(reason) }}</a></div>
+          {% endfor %}
+        </div>
+
+        <h5>{{ _('or enter a reason in the text box below') }}:</h5>
+        <form method="post" id="ban-user-form" action="{{ url('users.ban_user',user_to_ban.id) }}">
+          {% csrf_token %}
+          <div>{{ form.reason }}</div>
```
jwhitlock: Add any form errors are well:

```
{% csrf_token %}
{{ form.reason.errors }}
<div>{{ form.reason }}</div>
```
errors before the ``<div>`` line with:

jwhitlock:
OK, that makes sense, @safwanrahman, there should be at least one content test.

The code worked mostly like I expected:
* Selecting "BAN USER" opens the "Why are you banning (x)?" page
* Choosing a reason fills the reason box
* I can edit or replace the reason
* If I leave the reason blank, I come back (but without the error!)
* If a user is banned, clicking "BANNED" goes to the Django admin

Here's my "do it or argue about it" list:
* [ ] Move ``common_reasons`` default value to Constance, and handle bad JSON or empty lists with a very small default (with test).
* [ ] Add a test that confirms at least one of the reasons is in the output HTML. I'd use PyQuery to find the `<a>` tags, check contents, like some of the [wiki tests](https://github.com/mozilla/kuma/blob/master/kuma/wiki/tests/test_views.py#L358-L368), but I'm open to other methods.
* [ ] Update the translation of "Why are you banning..." so it's a sentence.
* [ ] Add the form error display to the template
* [ ] Adjust the ``ban_user`` view so that the invalid form is the one used when POSTing an empty reason (move the GET logic into an ``else`` block)

Also, I've been assuming these checklists are clickable. Does that only work for Mozilla staff?

kuma/users/jinja2/users/ban_user.html
```diff
@@ -0,0 +1,53 @@
...
+  <section id="content-main" class="full">
+    <section id="user-head" class="column-container column-container-reverse vcard user-head">
+      <div class="column-10">
+        <h1 class="user-title">
+          {{ _('Why are you banning') }} {{user_to_ban }}?
+        </h1>
+        <div class="column-4">
+          <h5>{{ _('Choose a reason') }}:</h5>
+          {% for reason in common_reasons %}
+            <div><a class="ban-common-reason">{{ _(reason) }}</a></div>
```
emullaney: ban-common-reason: Can we add a style for this class to make the mouse-over cursor a pointing hand?

dchukhin:
Thanks for the review @jwhitlock and @safwanrahman.
Should be updated for re-review, and yes, John, it appears the checklists are not clickable for me

kuma/users/jinja2/users/ban_user.html
stephaniehobson: Since this isn't a link that takes the user off the page an `a` element is not appropriate (and the browser responds to the missing href by not changing the cursor when the element is hovered). I'm not familiar with the purpose of the PR but since you are looping through a list of options for the user to pick between I would suggest either `select` or `button`.

kuma/users/tests/test_views.py
```diff
@@ -109,6 +110,89 @@ def test_bug_811751_banned_user(self):
...
+
+        # If there is an error in getting the common reasons from constance,
+        # then 'Spam' should still show up in the template as the default
+        testuser = self.user_model.objects.get(username='testuser')
+
+        self.client.login(username='admin', password='testpass')
+        ban_url = reverse('users.ban_user',
+                          kwargs={'user_id': testuser.id})
+
+        constance_config.COMMON_REASONS_TO_BAN_USERS = 'not valid JSON'
```
jwhitlock: Instead of setting the config value (which may persist past the end of this test), split into a new test and use the ``@override_config`` decorator, like [other tests](https://github.com/mozilla/kuma/blob/master/kuma/users/tests/test_views.py#L529-L531)

kuma/users/views.py
```diff
@@ -65,9 +65,12 @@ def ban_user(request, user_id):
                           is_active=True)
             ban.save()
             return redirect(user)
-    if user.active_ban:
-        return redirect(user)
-    form = UserBanForm()
+        else:
+            form = UserBanForm(request.POST)
```
jwhitlock: The form doesn't need to be reinitialized. The one that just got validated above (``form.is_valid()``) is fine. The important part is the next few lines, that don't overwrite it with an unbound form.

jwhitlock:
VM tests worked well, with a form error on a blank reason.  There's a few changes needed:

* Consider something other than an ``<a>`` element, to get the desired hover behavior.
* Add a test for attempting to ban a nonexistent user (for code coverage)
* Add a test that POSTs with a blank reason (for code coverage)
* Drop the extra form assignment
* Use ``@override_config`` for overriding the config

kuma/settings/common.py
```diff
@@ -1209,6 +1209,12 @@ def _get_languages_and_locales():
         ]),
         "JSON array listing tag suggestions for documents"
     ),
+    COMMON_REASONS_TO_BAN_USERS=(
```
safwanrahman: I strongly believe this should live into [``user contants``](https://github.com/mozilla/kuma/blob/master/kuma/users/constants.py)

kuma/users/tests/test_views.py
```diff
@@ -109,6 +110,89 @@ def test_bug_811751_banned_user(self):
...
+
+        # For a banned user redirect to user detail page
+        UserBan.objects.create(user=testuser, by=admin,
+                               reason='Banned by unit test.',
+                               is_active=True)
+        resp = self.client.get(ban_url)
+        eq_(302, resp.status_code)
+        ok_(testuser.get_absolute_url() in resp['Location'])
+
+    def test_common_reasons_in_template(self):
```
safwanrahman: This should live in ``test_tamplates.py``

safwanrahman:
>  Move common_reasons default value to Constance, and handle bad JSON or empty lists with a very small default (with test).

@jwhitlock Does not its a better idea to save the value in a python list? If json is made, it need to be decoded into a list again that may cause a performance issue.

kuma/settings/common.py
jwhitlock: @safwanrahman why?

jwhitlock:
This list of reasons is one that I came up with from moderating edits for a few months, and analyzing past user bans.  Like [I said](https://github.com/mozilla/kuma/pull/3877#discussion_r64284799), there is a very good chance the team will want to change the values after deployment. This is why [django-constance](http://django-constance.readthedocs.io/en/latest/index.html) is the right place for this instead of in constants in the code.  When it is stored in constance, it can be updated by going to the Django admin and changing the value, and takes effect instantly. If it was in the code, it would require a pull request and deployment, with a delay of days. Flexibility is required now. Later, when it hasn't changed in weeks, it can be switched to code, or even a choice field in the model.

``django-constance`` doesn't support Python lists, so some encoding is needed. I'm following the established convention for storing lists in , such as [WIKI_DOCUMENT_TAG_SUGGESTIONS](https://github.com/mozilla/kuma/blob/master/kuma/settings/common.py#L1197-L1211). The result looks like a Python list in the Django admin:

pr 3877 - constance

Further, we're using [the memcache backend](https://github.com/mozilla/kuma/blob/master/kuma/settings/common.py#L1100) so that values are read once from the database and then used. There is still serialization / deserialization to get the value in and out of memcache, but no database hit.

Most of our users are good actors. On a normal day, less than 1% of the edits are spam. JSON serialization is not a significant performance hit for a view only accessible to spam moderators.

kuma/users/tests/test_views.py
jwhitlock: Maybe. A lot of the ``test_templates.py`` code looks like view tests to me, and won't survive the Great Test Cleanup.  I'll leave this up to @dchukhin's judgement.

kuma/settings/common.py
safwanrahman: Because it does only applicable to a single app. Not globally.

kuma/users/tests/test_views.py
safwanrahman: I dont know the convention. Its not about clean up. It should be done according to it.

safwanrahman:
> When it is stored in constance, it can be updated by going to the Django admin and changing the value, and takes effect instantly. If it was in the code, it would require a pull request and deployment, with a delay of days. Flexibility is required now. Later, when it hasn't changed in weeks, it can be switched to code, or even a choice field in the model.

Great. Accepted.

kuma/users/tests/test_views.py
safwanrahman: So it can go as it is because of Flexibility.

jwhitlock:
:clap: :clap:

PR #3878, opened 2016-05-21
Fix bug 1265151 - Missing translation of some strings

bychek-ru:
I added localization for strings 'Thanks for your feedback!' and "We
won't bug you again."

### DIFF
--- a/kuma/static/js/helpfulness.js
+++ b/kuma/static/js/helpfulness.js
@@ -92,7 +92,7 @@
                     action: 'Clicked',
                     label: label
                 });
-                notification[type](gettext(msg), 2000);
+                notification[type](msg, 2000);
             }
         }
     }
--- a/kuma/static/js/helpfulness.js
+++ b/kuma/static/js/helpfulness.js
@@ -42,7 +42,7 @@
         // answers to the simple question include...
         $('#helpful-yes').on('click', function(e) {
             e.preventDefault();
-            confirm('Thanks for your feedback!', 'success', 'Yes');
+            confirm(gettext('Thanks for your feedback!'), 'success', 'Yes');
         });
         $('#helpful-no').on('click', function(e) {
             e.preventDefault();
@@ -51,7 +51,7 @@
         $('#helpful-ignore').on('click', function(e) {
             e.preventDefault();
             localStorage.setItem('helpful-ignore', 'true');
-            confirm("We won't bug you again.", 'info', 'Ignore');
+            confirm(gettext("We won't bug you again."), 'info', 'Ignore');
         });
 
         // create a dropdown in case the page is unhelpful
--- a/kuma/static/js/wiki-samples.js
+++ b/kuma/static/js/wiki-samples.js
@@ -94,7 +94,6 @@
                     // create icon
                     var $icon = $('<i />', { 'class': 'icon-' + sampleCodeHost, 'aria-hidden': 'true' });
                     // create text
-					var $text = interpolate(gettext('Open in %(site)s '), {site: this});
 
                     // add button icon and text to DOM
                     $button.append($text).append($icon).appendTo($buttonContainer);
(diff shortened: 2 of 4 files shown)


kuma/static/js/helpfulness.js
```diff
@@ -42,7 +42,7 @@
         // answers to the simple question include...
         $('#helpful-yes').on('click', function(e) {
             e.preventDefault();
-            confirm('Thanks for your feedback!', 'success', 'Yes');
+            confirm(gettext('Thanks for your feedback!'), 'success', 'Yes');
```
jwhitlock: The [confirm function](https://github.com/mozilla/kuma/blob/master/kuma/static/js/helpfulness.js#L95) is currently trying to translate these strings, and will need to be modified so that double-translation is not attempted.

safwanrahman: Yeah. Get text should be here and not in the ``confirm`` function.

kuma/static/js/wiki-samples.js
```diff
@@ -94,7 +94,7 @@
                     // create icon
                     var $icon = $('<i />', { 'class': 'icon-' + sampleCodeHost, 'aria-hidden': 'true' });
                     // create text
-                    var $text = 'Open in ' + this + ' ';
+                    var $text = gettext('Open in ') + this + ' ';
```
jwhitlock: This should instead use [ngettext](https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#interpolate), like ``var $text = ngettext('Open in %(site)s ', this);``, so that languages where the words come in a different order will be supported.

kuma/wiki/jinja2/wiki/list/needs_review.html
```diff
@@ -23,13 +23,13 @@
         {% if tag|string=='editorial' %}
           <p>
             {% trans url=wiki_url('MDN/Contribute/Howto/Do_an_editorial_review') + '#core-steps' %}
-            Please review these articles for spelling, grammar, formatting, layout, etc., and then clear the Editorial review flag. <a href="{{ url }}">See more detailed instructions</a>.
+            {{ _('Please review these articles for spelling, grammar, formatting, layout, etc., and then clear the Editorial review flag. <a href="%(url)s">See more detailed instructions</a>.', url=url) }}
```
jwhitlock: ``{% trans %}`` already translates this block of text.  The string is missing from the translation interface because I haven't done string extraction since merging this code. Remove changes to this file from the PR.

jwhitlock:
@bychek-ru done with my read-through, and it looks like you are also changing the code. Mention me when you're done, and I'll get you back in the code review line.

bychek-ru:
I've done and fixed first two problems

bychek-ru:
I've done and fixed all problems.

kuma/static/js/wiki-samples.js
```diff
@@ -94,7 +94,7 @@
                     // create icon
                     var $icon = $('<i />', { 'class': 'icon-' + sampleCodeHost, 'aria-hidden': 'true' });
                     // create text
-                    var $text = 'Open in ' + this + ' ';
+                    var $text = ngettext('Open in %(site)s ', this);
```
jwhitlock: I'm sorry, I'm new to the JS translation functions, and gave you bad advice.  ``ngettext`` is for picking between a singular and a plural text (like ``ngettext('View the record', 'View the records', record_count)``), but ``interpolate`` is used for populating a format string. I think this will work:

```
var $text = interpolate(gettext('Open in %(site)s '), {site: this});
```

jwhitlock:
@bychek-ru - one more change, to fix my ``ngettext`` suggestion with ``interpolate`` /  ``gettext``.

kuma/static/js/wiki-samples.js
```diff
@@ -94,7 +94,7 @@
                     // create icon
                     var $icon = $('<i />', { 'class': 'icon-' + sampleCodeHost, 'aria-hidden': 'true' });
                     // create text
-                    var $text = ngettext('Open in %(site)s ', this);
+					var $text = interpolate(gettext('Open in %(site)s '), {site: this});
```
jwhitlock: Please replace the tabs with 4 spaces

jwhitlock:
Good work, @bychek-ru :+1:

PR #3879, opened 2016-05-22
fix bug 1012106: use language icon for switching languages

singingwolfboy:
Fix for [bug 1012106](https://bugzilla.mozilla.org/show_bug.cgi?id=1012106) -- seems very straightforward.


jwhitlock:
Works like it says:
language icon

PR #3880, opened 2016-05-26
Bug 1265151 - interpolate(gettext('Open in %(site)s'), {site: this}) doesn't work

bychek-ru:
OMG! There's a bug: interpolate(gettext('Open in %(site)s'), {site: this}); doesn't put a site in the string. It returns 'Open in %(site)s'. I don't know better solution than gettext('Open in ')+this;


jwhitlock:
Use:
```
var $text = interpolate(gettext('Open in %(site)s '), {site: this}, true);
```

PR #3882, opened 2016-05-26
bug 1265151 - Add space back to sample buttons

jwhitlock:
While marking the text "Open in Codepen" for translation, the space between the text and the icon was lost. It doesn't really belong in the translated string, adding it as a literal string to the button.

[Live on MDN](https://developer.mozilla.org/en-US/docs/User:stephaniehobson:code#Code_sample_template):

without_space

With this change:

with_space


stephaniehobson:
R+

PR #3884, opened 2016-06-06
robots.txt edits

dchukhin:
https://tree.taiga.io/project/viya-mdn-durable-team/task/59

This pull request follows up on the request make in https://docs.google.com/spreadsheets/d/1X-YLmIg8vVvDWlShLF-361Cz2KcUBnlaPXBHLMaYEaQ/edit#gid=0 to add several more lines to robots.txt.
@stephaniehobson

### DIFF
--- a/media/robots.txt
+++ b/media/robots.txt
@@ -43,8 +43,7 @@ Disallow: /*docs*$subscribe_to_tree
 Disallow: /*docs*$vote
 Disallow: /*docs.json
 Disallow: /*docs/ckeditor_config.js
-Disallow: /*feed*
-Disallow: /*files*
+Disallow: /*files/
 Disallow: /media
 Disallow: /*move-requested
 Disallow: /*preview-wiki-content
--- a/media/robots.txt
+++ b/media/robots.txt
@@ -51,4 +51,4 @@ Disallow: /*profiles*edit
 Disallow: /*search*
 Disallow: /skins
 Disallow: /*type=feed
-Disallow: /*users*signin
+Disallow: /*users/
--- a/media/robots.txt
+++ b/media/robots.txt
@@ -3,13 +3,16 @@ Crawl-delay: 5
 Sitemap: https://developer.mozilla.org/sitemap.xml
 Request-rate: 1/5
 
-Disallow: /*feed=rss
+Disallow: /*admin*
+Disallow: /*feed*
 Disallow: /*type=feed
 Disallow: /skins
+Disallow: /*files*
 Disallow: /*docs/feeds
 Disallow: /*docs/templates
 Disallow: /*docs*Template:
 Disallow: /media
+Disallow: /*dashboards*
 Disallow: /*docs/all
 Disallow: /*docs/tag*
 Disallow: /*docs/needs-review*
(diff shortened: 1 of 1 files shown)


media/robots.txt
```diff
@@ -32,13 +32,24 @@ Disallow: /*docs*$styles
...
 Disallow: /*docs*$delete
 Disallow: /*docs*$restore
 Disallow: /*docs*$purge
 Disallow: /*docs*$subscribe
+Disallow: /*docs*$subscribe_to_tree
 Disallow: /*docs*$vote
 Disallow: /*docs.json
-Disallow: /*preview-wiki-content
+Disallow: /*docs/ckeditor_config.js
+Disallow: /*feed*
```
stephaniehobson: This rule needs to be more precise as it will also block articles like https://developer.mozilla.org/en-US/Firefox/Releases/2/Adding_feed_readers_to_Firefox

media/robots.txt
```diff
@@ -32,13 +32,24 @@ Disallow: /*docs*$styles
...
+Disallow: /*files*
```
stephaniehobson: This rule needs to be more precise at it will also block articles like https://developer.mozilla.org/en-US/docs/Using_files_from_web_applications

stephaniehobson:
Thanks for putting it in alphabetical order! A few rules need to be made more precise.

media/robots.txt
```diff
@@ -32,13 +32,24 @@ Disallow: /*docs*$styles
...
+Disallow: /media
 Disallow: /*move-requested
+Disallow: /*preview-wiki-content
+Disallow: /*profiles*edit
+Disallow: /*search*
+Disallow: /skins
+Disallow: /*type=feed
+Disallow: /*users*signin
```
stephaniehobson: Need to block everything under users according to the spreadsheet.

media/robots.txt
dchukhin: good :eyes:

media/robots.txt
dchukhin: good :eyes:

dchukhin:
@stephaniehobson thanks for the review; should be ready for re-review

stephaniehobson:
This works well for our regular documents and is an improvement over what we have so I'm going to merge. 👍 

When I compared some of the test URLs to the file though I discovered that all the `$ending` urls are not blocked if they're under a zone (yay, another reason for us to stop using zones). 

@dchukhin can you do a bit of research to see if the end of the URL matching syntax I saw mentioned [on one blog](http://tools.seobook.com/robots-txt/) is well supported and, if so, submit another PR using it with the `$endings` where appropriate (It doesn't look appropriate for $revisions for example). If it's not well supported and you don't recommend switching just comment here.

Example:
```
Disallow: /*$history$
```

Thanks.

dchukhin:
I have looked around and based on [here] (https://developers.google.com/webmasters/control-crawl-index/docs/robots_txt#url-matching-based-on-path-values) and [here] (https://yoast.com/ultimate-guide-robots-txt/) it seems that some crawlers (Google, and a few other major ones) support using a `$` for the end of the url. I'm not sure if that's enough for what we want or not, since Google is the most popular, but certainly not the only one. Thoughts? @stephaniehobson

stephaniehobson:
@dchukhin Thanks for doing that research. Google is what we're most concerned about. 

I've done a little more thinking on this and I think we can block zones by making the current rules less specific. Could you please change the rules looking for `/*docs*$ending` to not have the `docs` directory in them? I'll send an email about this too because git notifications are easy to miss.

PR #3885, opened 2016-06-07
New issue 1275322 spam admin info on submission

emullaney:
This addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1275322 and https://tree.taiga.io/project/viya-mdn-durable-team/task/137.

Add a paragraph with a link to the admin to allow spam moderators to mark submissions as ham and try to submit again.

cc @stephaniehobson

### DIFF
--- a/kuma/wiki/jinja2/wiki/includes/spam_error.html
+++ b/kuma/wiki/jinja2/wiki/includes/spam_error.html
@@ -14,3 +14,10 @@
     contact the site administrators</a>.
     {% endtrans %}
 </p>
+{% if perms.wiki.add_revisionakismetsubmission %}
+    <p>
+    {% trans akismet_submission_url = url('admin:wiki_revisionakismetsubmission') %}
+    As a spam moderator you can <a href="{{ akismet_submission_url }}" target="_blank">mark this submission as ham</a> and try submitting it again.
+    {% endtrans %}
+    </p>
+{% endif %}
--- a/kuma/wiki/jinja2/wiki/includes/spam_error.html
+++ b/kuma/wiki/jinja2/wiki/includes/spam_error.html
@@ -14,10 +14,8 @@
     contact the site administrators</a>.
     {% endtrans %}
 </p>
-{% if perms.wiki.add_revisionakismetsubmission %}
-    <p>
+<p>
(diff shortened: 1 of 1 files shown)


kuma/wiki/jinja2/wiki/includes/spam_error.html
```diff
@@ -14,3 +14,10 @@
     contact the site administrators</a>.
     {% endtrans %}
 </p>
+{% if perms.wiki.add_revisionakismetsubmission %}
```
stephaniehobson: Getting the error "perms is undefined" when this tires to display and the tests are failing :/

emullaney: Thanks for testing! I wasn't sure how to call this up locally to test. Can you give me some pointers on how to see this page so I can test it?

stephaniehobson: I added myself to all the spam admin groups and then tried to submit some spam, mostly by copying other stuff that has been marked as spam, that takes some time. It's possible that what I'm really triggering is Akismet's sense that I'm making too many edits too quickly. Either way, that usually works for me.

A faster way to test that this error has been caught would be to create the test user account with the username `viagra-test-123`. All submissions from that account should be marked as spam. I got the error submitting content with that account as well even though the message shouldn't display to them since it has to evaluate the file either way.

emullaney: I'm wondering if I don't have Akismet running locally.... I can't get the spam to get caught although I've created this new test account locally. I'll try to figure out a way around it and keep working!

emullaney:
I altered the text from the original specs a bit so I'm waiting on that to get approval, https://bugzilla.mozilla.org/show_bug.cgi?id=1275322#c2

I also still cannot test this locally for some reason although Akismet is running locally for me just fine.

cc @stephaniehobson

stephaniehobson:
👍

jwhitlock:
I was reviewing commits, not the full PR, so I missed that the ``_changelist`` issue was already caught.  It is sad that there is not enough context to restrict the message to spam moderators.

PR #3889, opened 2016-06-12
Fix bug 1279839: remove Recent Docs Activity 

jgmize:


jgmize:
@jwhitlock r?

kuma/users/views.py
```diff
@@ -94,13 +94,7 @@ def user_detail(request, username):
...
         return render(request, '403.html',
                       {'reason': 'bannedprofile'}, status=403)
-
-    docs_feed_items = None
-
-    context = {
-        'detail_user': detail_user,
-        'docs_feed_items': docs_feed_items,
-    }
+    context = {'detail_user': detail_user}
```
jwhitlock: 👍

kuma/wiki/forms.py
```diff
@@ -759,7 +759,7 @@ def save(self, document, **kwargs):
         Does some specific things when the revision is fully saved.
         """
         # have to check for first edit before we save
-        is_first_edit = self.request.user.wiki_revisions().count() == 0
+        is_first_edit = not self.request.user.created_revisions.exists()
```
jwhitlock: 👍

kuma/wiki/jinja2/wiki/edit.html
```diff
@@ -156,7 +156,7 @@
 
         </header>
 
-        {% if request.user.wiki_revisions().count() == 0 and not revision_form.errors %}
+        {% if not revision_form.errors and not request.user.created_revisions.exists() %}
```
jwhitlock: 👍

jwhitlock:
This code includes some needed cleanup, and removes the recent docs section. I'm worried that we'll later revert the removal and lose the cleanup as well, but hopefully we'll remember.

I want to do some more investigation, but +r on a quick merge and deploy of this PR when needed.

stephaniehobson:
I don't think we can outright remove this section without talking to the writers as it is a tool for cleaning up after spammers. In it's place can we add a spam-admin only link to the revision dashboard filtered by this user name? Is that quick and easy?

jwhitlock:
Going with PR #3890. Thanks for coding the nuclear option, and finding those ``count() == 0`` queries.

PR #3890, opened 2016-06-15
Bug 1279839 - Profile page cleanup, use IDs instead of created date for recent activity 

jwhitlock:
Bring over the profile page cleanup from #3889, but retain the recent docs activity.

Switch to sorting by revision IDs instead of the ``created`` date for identifying recent documents. When tested against a local anonymized database, this gives the same result for every profile, but is much faster. In a worse-case scenario, sorting by ``created`` took 33 seconds, and by ``id`` 0.4 milliseconds.

Use this new sorting strategy on:
* The Recent Docs Activity section on the [user profile](https://developer.mozilla.org/en-US/profiles/jwhitlock)
* [The Revision Dashboard](https://developer.mozilla.org/en-US/dashboards/revisions)
* [Revision RSS feeds](https://developer.mozilla.org/en-US/feeds/rss/revisions)

There are additional places where sorting is done by ``created``, for example when viewing the revisions for a document, that are saved for future investigation, to see if the results are the same and if there is a speed improvement.

### DIFF
--- a/kuma/dashboards/views.py
+++ b/kuma/dashboards/views.py
@@ -26,7 +26,7 @@ def revisions(request):
     revisions = (Revision.objects.prefetch_related('creator__bans',
                                                    'document',
                                                    'akismet_submissions')
-                                 .order_by('-created')
+                                 .order_by('-id')
                                  .defer('content'))
 
     query_kwargs = False
--- a/kuma/users/models.py
+++ b/kuma/users/models.py
@@ -190,7 +190,7 @@ class User(AbstractUser):
     def wiki_revisions(self, count=5):
         return (self.created_revisions.prefetch_related('document')
                                       .defer('content', 'summary')
-                                      .order_by('-created')[:count])
+                                      .order_by('-id')[:count])
 
     def allows_editing_by(self, user):
         return user.is_staff or user.is_superuser or user.pk == self.pk
--- a/kuma/wiki/jinja2/wiki/edit.html
+++ b/kuma/wiki/jinja2/wiki/edit.html
@@ -156,7 +156,7 @@
 
         </header>
 
-        {% if request.user.wiki_revisions().count() == 0 and not revision_form.errors %}
+        {% if not revision_form.errors and not request.user.created_revisions.exists() %}
           <section class="first-contrib-welcome">
             <h2>{{ _('Thank you for taking the time to improve MDN!') }}</h2>
             <p>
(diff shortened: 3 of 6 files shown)


kuma/wiki/jinja2/wiki/edit.html
```diff
@@ -156,7 +156,7 @@
 
         </header>
 
-        {% if request.user.wiki_revisions().count() == 0 and not revision_form.errors %}
+        {% if not revision_form.errors and not request.user.created_revisions.exists() %}
```
groovecoder: This got changed from `user.wiki_revisions()` to `user.created_revisions` ? Is that on purpose?

kuma/dashboards/views.py
```diff
@@ -26,7 +26,7 @@ def revisions(request):
     revisions = (Revision.objects.prefetch_related('creator__bans',
                                                    'document',
                                                    'akismet_submissions')
-                                 .order_by('-created')
+                                 .order_by('-id')
```
safwanrahman: ``order_by('id').reverse()`` is a faster method.

kuma/users/models.py
```diff
@@ -190,7 +190,7 @@ def active_ban(self):
     def wiki_revisions(self, count=5):
         return (self.created_revisions.prefetch_related('document')
                                       .defer('content', 'summary')
-                                      .order_by('-created')[:count])
+                                      .order_by('-id')[:count])
```
safwanrahman: same here

kuma/wiki/jinja2/wiki/edit.html
jwhitlock: I think the intent (if there was one) of using ``user.wiki_revisions()`` was to use the Django query cache to use the previous value.  Because there is an index on the ``user`` foreign key, ``user.created_revisions.exists()`` should be very fast, so there's no need to reuse the complex ``wiki_revisions`` query.

kuma/dashboards/views.py
jwhitlock: It results in the same query (``ORDER BY `wiki_revision`.`id` DESC``) either way, so I'm not sure how it would be faster.

safwanrahman: Oops!

jwhitlock:
@jgmize review?

jgmize:
thanks @jwhitlock

PR #3891, opened 2016-06-17
robots.txt edits: Part II

dchukhin:
remove docs directory from robots rules with `$` endings


jwhitlock:
Some of the rules added in PR #3884 are too broad and need to be fixed.  For example, ``Disallow: /*admin*`` will block MDN pages with "admin" in the slug, such as [es/docs/El_administrador_de_hilos](https://developer.mozilla.org/es/docs/El_administrador_de_hilos).  This could be narrowed to ``Disallow: /admin/`` to only block the Django admin.

Other too-broad patterns:

* ``Disallow: /*dashboards*`` (no current docs blocked)
* ``Disallow: /*files/`` blocks [/en-US/docs/Multiple_Firefox_profiles/Multiple_Firefox_Profiles](https://developer.mozilla.org/en-US/Firefox/Multiple_profiles)
* ``Disallow: /*profiles*edit`` (no current docs blocked)
* ``Disallow: /*search*`` blocks [/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search](https://developer.mozilla.org//en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search)

Can you add slashes to these patterns so they will be more focused on the views you are attempting to block?

media/robots.txt
```diff
@@ -3,11 +3,11 @@ Crawl-delay: 5
 Sitemap: https://developer.mozilla.org/sitemap.xml
 Request-rate: 1/5
 
-Disallow: /*admin*
+Disallow: /*/admin/*
```
jwhitlock: This should be ``Disallow: /admin/``, which will catch all Django admin URLs.  They don't get the funky MDN locale prefix.

jwhitlock:
Looks good, except for the admin pattern. Back to you, @dchukhin.

dchukhin:
re re review, @jwhitlock?

PR #3893, opened 2016-06-21
bug 1177264 - remove pagination from revision dashboard

8v060htwyc:
Replace Django pagination with custom "last_id"-based links for
seeing the next page of results.

See http://www.xarg.org/2011/10/optimized-pagination-using-mysql/


jwhitlock:
Smells like a bot

PR #3894, opened 2016-06-21
bug 1275322 - Adjust message show to spam moderators when revision is blocked

jwhitlock:
When Akismet blocks an edit, show the admin link to the filtered ``DocumentSpamAttempt`` changelist, but only to spam moderators.


jgmize:
r+ nice work, @jwhitlock

stephaniehobson:
😍

PR #3895, opened 2016-06-23
bug 1277577 - Rearrange and document requirements

jwhitlock:
Add a comment summarizing how each requirement is used in Kuma. For requirements that are unused, or have a more commonly used variant, add a note to remove it. Move requirements of requirements to ``constraints.txt``. Add a ``README.rst`` for future devs.

The versions have not been updated. That will be done slowly, in separate PRs.

### DIFF
--- a/requirements/README.rst
+++ b/requirements/README.rst
@@ -1,0 +1,99 @@
+Kuma Requirements
+=================
+
+The files define the third-party libraries needed for running and developing
+Kuma.  The files are:
+
+* ``constraints.txt`` - Requirements for our requirements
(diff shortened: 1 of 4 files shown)


jwhitlock:
Force-push to update hashes for ``logilab-common==1.1.0``. There was a network error and I had an empty package, so the sha256 hash was for an empty file (``echo -n "" | shasum -a 256`` is ``e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855``, a new string to memorize)

jwhitlock:
@jgmize review?

PR #3898, opened 2016-06-29
fix bug 1276929 - Unset config.vm.hostname in the Vagrantfile

jbradberry:
Unset config.vm.hostname in the Vagrantfile.

Having this option set (currently) causes Vagrant to destructively
edit the guest's /etc/hosts file, removing the localhost alias for
127.0.0.1.

This option shouldn't be needed, since the kuma role is copying over
/etc/hostname anyway.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -32,8 +32,9 @@ Vagrant.configure('2') do |config|
     config.vm.box = 'ubuntu/trusty64'
     config.vm.define 'developer-local'
     config.vm.network :private_network, ip: IP
-    config.vm.hostname = 'developer-local.allizom.org'
-    config.hostsupdater.aliases = ['mdn-local.mozillademos.org']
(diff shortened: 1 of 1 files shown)


jwhitlock:
This works for me, OS X 10.11.15, Vagrant 1.8.1, Ansible 2.0.1.0.  The old in-VM ``/etc/hosts`` line was:
```
127.0.0.1 developer-local.allizom.org developer-local
```

The new ``/etc/hosts`` line is:
```
127.0.0.1		localhost kuma developer-local.allizom.org developer-local
```

In either case, both ``developer-local.allizom.org`` and ``mdn-local.mozillademos.org`` are added to the host machine's ``/etc/hosts``.

@emullaney - have you had a chance to try this change?

emullaney:
This works well on my local Linux environment +1 @jwhitlock @jbradberry

PR #3896, opened 2016-06-24
bug 1259776 - One click ban hammer: Phase II

dchukhin:
This pull request adds the features described for Phase II of the one-click-ban feature: "Have checkbox to submit revisions to Akismet as spam."

One question that still needs to be answered is: how should we handle invalid data for the `RevisionAkismetSubmissionSpamForm` that we use in [`kuma/users/views.py`](https://github.com/caktus/kuma/blob/7f31d9edaa9dba644939fcece38e49a366c8b3c7/kuma/users/views.py#L155)?
@jwhitlock

PR #3899, opened 2016-07-01
bug 1259776 - One-click ban hammer: Phase III

emullaney:
This is a WIP for Phase 3 work described [in this doc](https://docs.google.com/document/d/1xewZXZlz-VpUFXff3GQ2QElxgqxFxldN751boGkIBPI/edit)

- [x] New diff on each revision displayed on the one-click page between English and a new translation
- [x] Back-end functions that revert a revision (if it is the current revision for a doc) and delete a revision/doc (if it is the first and only version of a new doc/translation).
- [x] Merged in updates from Phase 2 branch #3896 
- [x] Updated language on one-click page depending on revision type ie. 
    - Report as spam. Revert to last edit
    - Report as spam. Delete this new translation
    - Report as spam. Cannot revert: not the most recent revision

TODO:
- [x] Add tests for new functionality
- [ ] Make sure new functionality is working as intended 
- [x] Fix broken tests
- [x] Merge in latest updates from #3896 
- [x] Adjust the summary page per @stephaniehobson's new wireframe from 7/5


kuma/users/views.py
```diff
@@ -138,27 +145,132 @@ def ban_user_and_cleanup_summary(request, user_id):
...
-    # and either revert them or not. For now list all of the revisions for the past 3 days
     date_three_days_ago = datetime.now().date() - timedelta(days=3)
-    revisions_reverted = user.created_revisions.prefetch_related('document')
-    revisions_reverted = revisions_reverted.defer('content', 'summary').order_by('-id')
-    revisions_reverted = revisions_reverted.filter(created__gte=date_three_days_ago)
-
-    # TODO: revisions_needing_follow_up will be revisions that have not been reverted
-    revisions_needing_follow_up = revisions_reverted
-    revisions_reverted = revisions_reverted
+    revisions_from_last_three_days = user.created_revisions.prefetch_related('document')
```
safwanrahman: Its better to write it in one variable like
```python
revisions_from_last_three_days = (user.created_revisions.prefetch_related('document')
                                                        .defer('content', 'summary')
                                                        .order_by('-id')
                                                        .filter(created__gte=date_three_days_ago))
```

emullaney:
This branch can be reviewed for functionality at any time. However, it might make sense to wait for Phase 2 changes to be completed and merged in first to make comparing the code more simple. Also, the branch has conflicts which need to be resolved, but that will happen when Phase 2 branch is merged in as well.  cc @jwhitlock @stephaniehobson

emullaney:
This works well on my local Linux environment :+1: @jwhitlock @jbradberry

jwhitlock:
PR #3896 is merged, so ready for rebase.

kuma/users/views.py
```diff
@@ -138,27 +147,192 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+    else:
+        list_of_revisions_distinct_doc = []
+
+    return list_of_revisions_distinct_doc
+
+
+@permission_required('users.add_userban')
+def revert_document(request, document_path, revision_id):
+    """
+    Revert document to a specific revision.
```
jwhitlock: I see that this is the logic from ``kuma/wiki/views/delete.py``, modified for this context. The idea is sound, but I think the interface is wrong. You already have the ``revision`` and the ``document``, but by using ``locale_and_slug_from_path`` you bring in a lot of logic that could pick a different document based on the spam moderator's preferred locale.  Instead, pass in the ``revision`` and go from there. When you are done, there may or may not be enough to justify a second function.

kuma/users/views.py
```diff
@@ -138,27 +147,192 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+    except IntegrityError:
+        return False
+    return True
+
+
+@permission_required('wiki.delete_document')
+def delete_document(request, document_slug, document_locale):
+    """
+    Delete a Document.
+    """
```
jwhitlock: Similar to the comment on ``revert_document`` - the interface should be ``revision.document``, not the ``document``'s slug.

jwhitlock:
Continuing fresh in #3903

PR #3900, opened 2016-07-05
Fix Bug 1284584: Standardize learn-box

stephaniehobson:
Made .learn-box use the .properties styles but made it orange.

Removed old learn-box styles from customcss.styl

Updated .standard-table definition to exclude learn-boxes because .learn-box.standard-table is a very common combination.

Testing pages:
https://developer-local.allizom.org/en-US/Learn/HTML/Howto/Use_JavaScript_within_a_webpage
https://developer-local.allizom.org/en-US/Learn/HTML/Howto/Add_a_hit_map_on_top_of_an_image
https://developer-local.allizom.org/en-US/Learn/CSS/Using_CSS_in_a_web_page
https://developer-local.allizom.org/en-US/Learn/CSS/Basic_text_styling_in_CSS

When testing also edit a table to remove the standard-table class and make sure it displays properly.

### DIFF
--- a/kuma/static/styles/components/wiki/properties.styl
+++ b/kuma/static/styles/components/wiki/properties.styl
@@ -7,6 +7,7 @@ $properties-item-spacing = $content-horizontal-spacing;
 $properties-spacing = ($grid-spacing) - $properties-item-spacing;
 $properties-indent = ($grid-spacing * 2) - $properties-item-spacing;
 
+.learn-box, /* not part of .properties so not being phased out */
 .properties, /* writing team is transitioning to use only this class - bug 1207344 */
 .cssprop,
 .htmlelt,
@@ -72,9 +73,15 @@ $properties-indent = ($grid-spacing * 2) - $properties-item-spacing;
     }
 }
 
-/* small screen improvments to table display, only need to apply to .properties
+.text-content .learn-box {
+    border-color: #f69855;
+    background-color: #fce1ce;
+}
+
+/* small screen improvements to table display, only need to apply to .properties and .learn-box
 because we can be sure that is implemented with a table element */
 @media $media-query-small-mobile {
+    .learn-box,
     .properties {
         th,
         td {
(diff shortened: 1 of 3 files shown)


kuma/static/styles/components/wiki/properties.styl
```diff
@@ -72,9 +73,15 @@ $properties-indent = ($grid-spacing * 2) - $properties-item-spacing;
     }
 }
 
-/* small screen improvments to table display, only need to apply to .properties
+.text-content .learn-box {
+    border-color: #f69855;
+    background-color: #fce1ce;
+}
+
+/* small screen improvments to table display, only need to apply to .properties and .learn-box
```
schalkneethling: "improvments" > "improvements"

schalkneethling:
@stephaniehobson So, I did:
```
(env) vagrant@developer-local:~/src$ sudo mysql -uroot -pkuma kuma < sanitized_devmo.sql
Warning: Using a password on the command line interface can be insecure.
(env) vagrant@developer-local:~/src$ foreman start
```
But when I try to load the pages you mentioned I get a 404. Is there something else I am missing?

stephaniehobson:
Hm, that should be it and you and I should have the same data as I do :/ Can you follow any links off the home page?

kuma/static/styles/components/wiki/properties.styl
stephaniehobson: The only things I can spell correctly reliably are CSS properties ;)

stephaniehobson:
If you don't have those pages locally you can copy them over from prod. Or find other pages to copy over from prod. [All pages on prod using the learn box class](https://developer.mozilla.org/en-US/search?locale=*&css_classnames=learn-box).

You can create new pages by visiting the [new page page](https://developer-local.allizom.org/docs/new).

Updated my spelling error.

schalkneethling:
Works as expected. The only concern I have is with the contrast, especially the link color on the peach like background color:

screen shot 2016-07-12 at 10 31 57

@stephaniehobson Thoughts?

stephaniehobson:
I really want to fix the contrast problems on MDN but I think that's a separate PR. This is a combination that is already in production in other places and we have [bug 924583](https://bugzilla.mozilla.org/show_bug.cgi?id=924583) for the contrast problems.

Merging since it works as expected.

PR #3901, opened 2016-07-06
Bug 1273028: Persona shutdown banner

stephaniehobson:
Adds banner across top of all pages when a user has logged in with persona "You logged in with Persona. [Persona is being turned off. Read more.](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Persona_Sign_In_on_MDN)"

Changes sign in drop down and sign in page to say account creation can only happen with Github (but there are no back end changes to enforce this).

Some things to test:
- Banner shows up when signed in with Persona and not when signed in with GitHub or signed out.
- I used "sign in" and "MDN profile" in all copy (not log in and MDN account).
- Create account with GitHub button works (probably good enough to test that you can sign in using that button)
- No spelling mistakes in copy.
- Sign in with Persona has not been broken.
- Other stuff I haven't thought of.

### DIFF
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -70,6 +70,17 @@
 
   {{ soapbox_messages(get_soapbox_messages(request.path)) }}
 
+  {% if user.is_authenticated() %}
+      {% if request.session.sociallogin_provider == 'persona' %}
+          <div class="notification warning" data-level="warning">
+              {% trans persona_info='/docs/MDN/Contribute/Persona_Sign_In_on_MDN' %}
+                You signed in with Persona.
+                <a href="{{ persona_info }}">Persona is being turned off. Read more.</a>
+              {% endtrans %}
+          </div>
+      {% endif %}
+  {% endif %}
+
   <ul id="nav-access">
     <li><a href="#{% block main_content_id %}content{% endblock %}" id="skip-main">{{ _('Skip to main content') }}</a></li>
     <li><a id="skip-language" href="#language">{{ _('Select language') }}</a></li>
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -72,7 +72,7 @@
 
   {% if user.is_authenticated() and request.session.sociallogin_provider == 'persona' %}
       <div class="notification warning" data-level="warning">
-          {% trans persona_info='/docs/MDN/Contribute/Persona_Sign_In_on_MDN' %}
+          {% trans persona_info='/docs/MDN/Contribute/Persona_sign-in' %}
             You signed in with Persona.
             <a href="{{ persona_info }}">Persona is being turned off. Read more.</a>
           {% endtrans %}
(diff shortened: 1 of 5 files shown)


stephaniehobson:
@jwhitlock for the django and and an understanding of the code base
@schalkneethling for the CSS, UI, and a set of fresh eyes.

Some background: we're trying to push people toward GitHub since Persona is being turned off. We don't have the dev time to make back-end changes to support this but there's some quick stuff we can do on the front-end so I did it. The back-end work will follow and hopefully be a bit more through (for example: we need to remove the Persona option from the "connect accounts page").

schalkneethling:
Code looks good, will pull it down and do some clicking around.

jinja2/base.html
```diff
@@ -70,6 +70,17 @@
 
   {{ soapbox_messages(get_soapbox_messages(request.path)) }}
 
+  {% if user.is_authenticated() %}
+      {% if request.session.sociallogin_provider == 'persona' %}
+          <div class="notification warning" data-level="warning">
+              {% trans persona_info='/docs/MDN/Contribute/Persona_Sign_In_on_MDN' %}
```
jwhitlock: The URL changed to `/docs/MDN/Contribute/Persona_sign-in`. This might be a one-time change, or they may keep adjusting. It may be appropriate as a Constance variable, but we'd need a backend change to pass the variable to the template. Redirects are... OK.

jwhitlock:
Manual tests worked for me.  This required:

* Recompiing CSS (``make compilecss compilejsi18n collectstatic`` in VM)
* [Adding my GitHub OAuth app](https://kuma.readthedocs.io/en/latest/installation.html#advanced-enable-github-auth)
* Logging in with Persona, getting the banner
* Adding GitHub login, using link in profile editing (banner goes away)
* Log out, log in with Persona - banner is back
* Log out, log in with GitHub - no banner
* Remove Persona, using link in profile editing

Account creation with Persona still works, but that's a different task (mine, probably).  There's some other nice-to-haves that might make sense while doing that work, like changing the explainer URL to Constance and adding a Jinja2 helper, but I think this is a good PR as is, 👍

jinja2/base.html
```diff
@@ -70,6 +70,17 @@
 
   {{ soapbox_messages(get_soapbox_messages(request.path)) }}
 
+  {% if user.is_authenticated() %}
+      {% if request.session.sociallogin_provider == 'persona' %}
```
safwanrahman: This logic can be re written like 
```python
{% if user.is_authenticated() and request.session.sociallogin_provider == 'persona' %}
..........
{% endif %}
```

schalkneethling:
Other than the one comment by @safwanrahman r+w/c

jwhitlock:
Thanks for the review, @schalkneethling. Added commits to cover the URL change and combining the template logic.

PR #3902, opened 2016-07-06
Fix Bug 1274734: 'None' placeholder/default value

stephaniehobson:
If there's no pre-existing query we should provide a default value.


jwhitlock:
Went to https://developer-local.allizom.org/en-US/search , no "None" term, has focus. 👍

PR #3903, opened 2016-07-11
One click ban hammer phase 3 fresh

emullaney:
Phase 3 work described in [this doc](https://docs.google.com/document/d/1xewZXZlz-VpUFXff3GQ2QElxgqxFxldN751boGkIBPI/edit). I've created a new branch for this work because it became unmanageable to do the rebase. This PR replaces #3899 

This is ready for review by @jwhitlock and @stephaniehobson.

Please test out functionality by spamming and banning and make sure it is working as  expected. 
- [x] New diff on each revision displayed on the one-click page between English and a new translation
- [x] Back-end functions that revert a revision (if it is the current revision for a doc)* and delete a revision/doc (if it is the first and only version of a new doc/translation).
- [x] Add tests and adjust existing tests for new revert/delete functionality.

* Please note that if there is a case when a spammy user makes some spam edits, then a second user comes and makes a newer revision, the spammy edits will not be reverted because they are not the most recent, when we ban the spammer.


kuma/users/jinja2/users/includes/user_revision_history_form.html
```diff
@@ -19,7 +19,7 @@
                     tabindex="0"
                     class="dashboard-row {% if not previous_id %}dashboard-row-new{% endif %}"
                     >
-                    <td></td>
+                    <td></td> <!--- Empty cell -->
```
jwhitlock: Would also accept:

```
{# Empty cell (template-only comment, not in rendered page) #}
<!-- This cell intentionally left blank -->
```

kuma/users/jinja2/users/ban_user_and_cleanup_summary.html
```diff
@@ -41,7 +41,37 @@
...
+
+    <h5>{{ _('Deleted:')}}</h5>
+    <div id="revisions-deleted">
+      {% if actions_taken.revisions_deleted_list %}
+        <ul>
+          {% for revision in actions_taken.revisions_deleted_list %}
+            <li>
+              {{ revision.title }}
+            </li>
+          {% endfor %}
```
jwhitlock: I think it makes more sense to use the document title, and to link to it as well (with ``revision.document.get_absolute_url() ``), because admins have the option to revert an accidentally deleted page by following the link.

kuma/users/jinja2/users/ban_user_and_cleanup_summary.html
```diff
@@ -41,7 +41,37 @@
...
+
+    <h5>{{ _('Reverted:')}}</h5>
+    <div id="revisions-reverted">
+      {% if actions_taken.revisions_reverted_list %}
+        <ul>
+          {% for revision in actions_taken.revisions_reverted_list %}
+            <li>
+              {{ revision.title }}
+            </li>
+          {% endfor %}
```
jwhitlock: For similar reasons, I think this should be a link, either to the revision or the document history.

kuma/users/views.py
```diff
@@ -179,74 +184,152 @@ def ban_user_and_cleanup_summary(request, user_id):
...
-               'no_actions_taken': no_actions_taken}
+               'no_actions_taken': no_actions_taken
+               }
 
     return render(request,
                   'users/ban_user_and_cleanup_summary.html',
                   context)
 
 
+def revision_by_distinct_doc(list_of_revisions):
```
jwhitlock: 😍 love that this because a separate function

jwhitlock:
I'm at the end of my day, and won't have time for a full first review.  Made some initial comments, I'll pick it up in the morning again.

Also, anything I say about the front end should be taken as uniformed advice. @stephaniehobson owns the front-end code.

kuma/users/views.py
```diff
@@ -179,74 +183,147 @@ def ban_user_and_cleanup_summary(request, user_id):
...
     revisions_already_spam = revisions_from_last_three_days.filter(
         id__in=request.POST.getlist('revision-already-spam')
     )
-    distinct_already_spam_rev_ids = list(
-        Document.objects
-        .filter(revisions__in=revisions_already_spam)
-        .annotate(latest_rev=Max('revisions'))
-        .values_list('latest_rev', flat=True))
-    revisions_already_spam_by_distinct_doc = revisions_already_spam.filter(id__in=distinct_already_spam_rev_ids)
+    revisions_already_spam = [rev for rev in revisions_already_spam]
```
safwanrahman: An iteration is needed? I think ``[revisions_already_spam]`` should work

kuma/users/views.py
```diff
@@ -179,74 +183,147 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+    revisions_already_spam_by_distinct_doc = revision_by_distinct_doc(revisions_already_spam)
+
+    revisions_already_spam_ids = [rev.id for rev in revisions_already_spam]
```
safwanrahman: Its possible to get the id by using django quaryset API 
``revisions_already_spam_ids = [revisions_already_spam.values_list("id", flat=True)]``

kuma/users/views.py
jwhitlock: no, that would wrap the queryset in a list. I prefer ``list(revisions_already_spam)`` to the list comprehension.

kuma/users/views.py
jwhitlock: ``revisions_already_spam_ids = list(revisions_already_spam.values_list("id", flat=True))`` would work, before the queryset is turned into a list.  However, it doesn't save any database traffic, since you're going to request the full records anyway. I'm OK with the list comprehension here.

jwhitlock:
One of my manual tests was to create two spam edits on a document.  After reverting, the history looked something like this:

* Revision 100 - Reverted to Revision 96 (Good Edit)
* Revision 99 - Reverted to Revision 97 (Spam 1)
* Revision 98 - Spam 2
* Revision 97 - Spam 1
* Revision 96 - Good Edit

This was an unexpected surprise. Reading the logic, I would have expected Spam 2 to be reverted to Spam 1 (Revision 99), and then Spam 1 would not have been attempted to be reverted, because the most recent revision (99) was no longer in the list of revisions to be reverted.  I think when revision 97 is processed, the ``revisions.document.current_revision`` is still Revision 98, because when it was loaded from the database that was the current revision.

For my second test, I did this sequence:

1. Add some spam
2. Revert it using the page history, like a normal user would
3. Add some more content as a normal user
4. Add some more spam
5. Ban the user and revert.

The history was:

* Revision 200 - Revert to 194 (Good Content)
* Revision 199 - Revert to 197 (Additional good content)
* Revision 198 - Second spam
* Revision 197 - Additional good content
* Revision 196 - Manual Revert to 194 (Good content)
* Revision 195 - First Spam
* Revision 194 - Good content

I think this is an unwanted scenario. There should just be one reversion per document.  I hope that was clear enough to follow.

One way to get around the Django cache is to reload the document:

```
document = Document.objects.get(id=revision.document_id)
if document.current_revision not in revisions_to_mark_as_spam_and_revert:
```

I'm not used to Django 1.8 yet, which added a refresh method:
```
revision.document.refresh_from_db(fields=['current_revision'])
if revision.document.current_revision not in revisions_to_mark_as_spam_and_revert:
```

I think something is needed to make sure the ``current_revision`` is updated when the document changes in the loop.

Since that will mean just on reversion per document, it would be nice to revert a series of revisions.  I think you could use something like this around line 192:

```
later_revisions = []
while revision.previous in revisions_mark_as_spam_and_revert:
  later_revisions.add(revision)
  revision = revision.previous
```

This would walk the revision chain until it finds the first revision by the spammer.  You then process it the same way (if ``revision.previous``, revert, otherwise delete the document).  When recording the action, you extend ``later_revisions`` to the list, like:

```
if reverted:
    revisions_reverted_list.extend(later_revisions)
    revisions_reverted_list.append(revision)
```

This implies a few additional test cases:
* A spammer adds two revisions to an existing English document. The document it reverted to the good revision.
* A spammer creates a translation, and adds a further edit. The translation is deleted.
* A spammer adds spam to an existing document, then another user edits the document, then the spammer adds more spam.  The document is reverted to the other users' edit, leaving the first spam.

These "multi-edit" scenarios might be out of scope for phase 3. In that case, I'd want to explicitly skip automated reverts and deletes if a spammer has multiple edits on a document.

kuma/users/jinja2/users/ban_user_and_cleanup_summary.html
```diff
@@ -41,7 +41,41 @@
...
+      {% endif %}
+    </div><!-- /#revisions-deleted -->
+
+    <h5>{{ _('Reverted:')}}</h5>
+    <div id="revisions-reverted">
+      {% if actions_taken.revisions_reverted_list %}
+        <ul>
+          {% for revision in actions_taken.revisions_reverted_list %}
+            <li>
+              <a href="{{ revision.get_absolute_url() }}">
```
jwhitlock: Thank you for adding the link. In manual testing, I found that sometimes I wanted to see the document as it is now, and other times I wanted to see the document history.  I'd want the "View Page" and "History" buttons from the "Revisions already identified as spam" section. This can wait until @stephaniehobson does the front-end review, though, to see if she agrees.

kuma/users/jinja2/users/includes/user_revision_history_form.html
```diff
@@ -42,25 +42,36 @@
...
                             {% with revision_from=revision.previous, revision_to=revision %}
                               <div class="dashboard-compare">
                                 {% include "wiki/includes/revision_diff_table.html" %}
                               </div>
                             {% endwith %}
-                        {% endif %}
+
+                        {# If this is a new translation, compare it with English translation #}
+                        {% elif revision.document.locale != 'en-US' %}
+                            {% with revision_from=revision.document.parent.current_revision, revision_to=revision %}
```
jwhitlock: Instead of ``revision_from=revision.document.parent.current_revision``, use ``revision_fromrevision.based_on``, which is the English's documents revision at the time of translation.

kuma/users/jinja2/users/includes/user_revision_history_form.html
```diff
@@ -42,25 +42,36 @@
...
-                          <strong>Spam</strong>
+                          <strong>{{ _('Spam') }}</strong>
                         </label>
                       {% else %}
-                        <label><input type="checkbox" name="revision-already-spam" value="{{ revision.id }}" checked style="visibility:hidden;"></input></label>
-                        <p>Already reported as spam.</p>
+                        <label>
+                          <input type="checkbox" name="revision-already-spam" value="{{ revision.id }}" checked style="visibility:hidden;"></input>
+                        </label>
+                        <p>{{ _('Already reported as spam.') }}</p>
```
jwhitlock: 😄 Good catch of translatable string.

kuma/users/views.py
```diff
@@ -179,74 +183,143 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+        else:
+            # If this is a new revision on an existing document, revert it
+            if revision.previous:
+                reverted = revert_document(request=request,
+                                           revision_id=revision.previous.id)
+                if reverted:
+                    revisions_reverted_list.append(revision)
+                else:
+                    # If the revert was unsuccessful, include this in the follow-up list
+                    revisions_not_reverted_list.append(revision)
```
jwhitlock: Missing coverage for this path.  It may require mocking ``revert_document``, which could be challenging - I'm not 100% sure how to mock a function in the same module.

kuma/users/views.py
```diff
@@ -179,74 +183,143 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+
+            # If this is a new document/translation, delete it
+            else:
+                deleted = delete_document(request=request,
+                                          document=revision.document)
+                if deleted:
+                    revisions_deleted_list.append(revision)
+                else:
+                    # If the delete was unsuccessful, include this in the follow-up list
+                    revisions_not_deleted_list.append(revision)
```
jwhitlock: Also missing coverage here.  Mocking ``delete_document`` would work, if it is possible.

kuma/users/views.py
```diff
@@ -179,74 +183,143 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+    comment = "spam"
+    document = revision.document
+    old_revision_pk = revision.pk
+    try:
+        new_revision = document.revert(revision, request.user, comment)
+        # schedule a rendering of the new revision if it really was saved
+        if new_revision.pk != old_revision_pk:
+            document.schedule_rendering('max-age=0')
+    except IntegrityError:
+        return False
```
jwhitlock: Missing coverage on this branch. Mocking a ``IntegrityError`` would fix the missing coverage above as well, but it might be easier to have a test that calls ``revert_document`` directly.

kuma/users/views.py
```diff
@@ -179,74 +183,143 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+    return True
+
+
+@permission_required('wiki.delete_document')
+def delete_document(request, document):
+    """
+    Delete a Document.
+    """
+    if not document:
+        return False
```
jwhitlock: Missing coverage on this branch.  Again, calling ``delete_document`` directly is the easiest, but something else will catch coverage above as well.

kuma/users/tests/test_templates.py
```diff
@@ -642,8 +642,6 @@ class BanUserAndCleanupSummaryTestCase(SampleRevisionsMixin, UserTestCase):
     def test_no_revisions_posted(self):
         """If user has no revisions, it should be stated in summary template."""
         # The expected text
-        # TODO: Add in Phase III
```
jwhitlock: Remove the comment ``# The expected text`` as well

kuma/users/tests/test_templates.py
```diff
@@ -751,25 +743,21 @@ def test_revisions_posted_different_docs(self, mock_form):
...
         # The "No actions taken" section
         already_spam = page.find('#already-spam li')
         not_spam = page.find('#not-spam li')
-        no_delete_no_revert = page.find('#no-delete-no-revert li')
         eq_(len(already_spam), 0)
         eq_(len(not_spam), 0)
-        eq_(len(no_delete_no_revert), 3)
 
     @patch('kuma.wiki.forms.RevisionAkismetSubmissionSpamForm.is_valid')
     def test_revisions_posted_same_doc(self, mock_form):
```
jwhitlock: I believe after the changes discussed in the long, long comment, this test ``test_revisions_posted_same_doc`` will need to be adjusted.  Either it refused to revert any of the commits (detect multiple edits and give up), or it ends up deleting the document (all the edits are by the spammer).

jwhitlock:
I need to take a break on the code review, and will get back to it soon. I have the tests left to review.

Despite all the notes, this looks like a good change. I'll write more when I have time.

kuma/users/tests/test_templates.py
```diff
@@ -1296,30 +1255,98 @@ def test_delete_link_does_not_appear_summary_page_other_revisions(self):
...
         # eq_(len(delete_link_new_action_section), 0)
         eq_(len(delete_link_already_spam_section), 0)
 
+    def test_newest_revision_is_not_spam(self):
+        """
+        Test with a spam user who has made revisions to a single document,
+        but another user has made a more recent revision.
+        The newest revision was created by a non-spam user,
+        so none of the revisions actually need to be reverted.
+        """
```
jwhitlock: I agree that if there are later commits, then automatic reverting is not possible. However, it still may need follow up by a moderator, maybe appearing in the "needs followup" section. Maybe. I'd need some practice with real spam to see how it works. No change for now.

kuma/users/tests/test_templates.py
```diff
@@ -1296,30 +1255,98 @@ def test_delete_link_does_not_appear_summary_page_other_revisions(self):
...
+        eq_(len(revisions_reported_as_spam), 1)
+
+    def test_multiple_revisions_are_spam(self):
+        """
+        Test with a spam user who has made multiple revisions to a single document.
+        This document should be reverted to the last version that was created
+        by a non-spam user (self.admin).
+        The newest revision was created by a non-spam user,
+        so none of the revisions actually need to be reverted.
+        """
```
jwhitlock: I don't think this comment is correct for the test ``test_multiple_revisions_are_spam``. The admin created the first revision, and the next three by the spammer, and then no more.  I think the comment needs to change, and an additional check is needed that there _were_ reversions on this page, like:

```
revisions_reverted = page.find('#revisions-reverted li')
eq_(len(revisions_reverted), 1) 
```

kuma/users/tests/test_views.py
```diff
@@ -388,10 +390,119 @@ def test_post_submits_no_revisions_to_akismet_when_wrong_revisions_in_request(se
...
+
+        new_document = Document.objects.filter(id=new_document.id).first()
+        # Make sure that the current revision is not the spam revision
+        for revision in spam_revisions:
+            ok_(revision.id != new_document.current_revision.id)
+        # The most recent Revision object should be the document's current revision
+        latest_revision = Revision.objects.order_by('-id').first()
+        eq_(new_document.current_revision.id, latest_revision.id)
+        # Admin is the creator of this current revision
+        eq_(new_document.current_revision.creator, self.admin)
```
jwhitlock: Can you add a check that ``new_document.current_revision.content`` is the same as the original content?

kuma/users/tests/test_views.py
```diff
@@ -388,10 +390,119 @@ def test_post_submits_no_revisions_to_akismet_when_wrong_revisions_in_request(se
...
+        # The most recent Revision for this document
+        # should be the document's current revision
+        latest_revision_b = Revision.objects.filter(
+            document=new_document_b).order_by('-id').first()
+        eq_(new_document_b.current_revision.id, latest_revision_b.id)
+        # Admin is the creator of this current revision
+        eq_(new_document_b.current_revision.creator, self.admin)
+        revisions_b = Revision.objects.filter(document=new_document_b)
+        # 7 revisions on B = 1 initial + 3 spam revisions + 3 new revert revisions
+        eq_(revisions_b.count(), 7)
```
jwhitlock: Please add a check that ``new_document_b.current_revision.content`` is the same as the initial revision's content, to verify the revert was correct.

kuma/users/tests/test_templates.py
```diff
@@ -656,53 +654,49 @@ def test_no_revisions_posted(self):
 
         # The "Actions taken" section
         banned_user = page.find('#banned-user li').text()
-        # TODO: Add in Phase III
-        # revisions_reverted = page.find('#revisions-reverted li')
-        # revisions_deleted= page.find('#revisions-deleted li')
+        revisions_reverted = page.find('#revisions-reverted li')
+        revisions_deleted = page.find('#revisions-deleted li')
         # TODO: Add in Phase IV
```
jwhitlock: It's so good to see these ``TODO:``s get resolved. 🎆

jwhitlock:
I'm done with my first review of the code.  I'm happy with the results of practical tests - this code reverts spam edits, deletes spam translations, and gets it off the site as quickly as possible. It will handle the most common cases.  I'm excited to get it to production.

I think there's work to do around a series of spam edits to the same page.  It works in the tested cases (a series of spam followed by a good edit, a series of spam edits at the end), but not if there is a break in the series (spam, good edit, last edit is spam).  I think if there are multiple spam edits to the same document, we should either:

* Revert nothing and flag it for human followup, with a link to the document history page, or
* Revert only the most recent series of spam edits with a single revision

I like the second option, but not if it adds a week to the task.

I made additional comments on the templates (adding some links) and the tests (nitpicks, added things).  I think adjusting the logic for a series of edits will have effects on the other code, and should be done first.

kuma/users/tests/test_templates.py
jbradberry: The document in this test actually has one good commit, the original.  The case where all revisions are spam is covered as part `test_some_revision_ids_posted`, which correctly deletes the document in question.

jbradberry:
@jwhitlock I think with this last commit all of your review comments were addressed.

kuma/users/views.py
```diff
@@ -179,74 +183,151 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+                                 pk=revision_id)
+
+    comment = "spam"
+    document = revision.document
+    old_revision_pk = revision.pk
+    try:
+        new_revision = document.revert(revision, request.user, comment)
+        # schedule a rendering of the new revision if it really was saved
+        if new_revision.pk != old_revision_pk:
+            document.schedule_rendering('max-age=0')
```
jwhitlock: This branch is always hit, so the case of ``new_revision.pk == old_revision_pk`` is never hit. I'm not sure when it wouldn't be though, and isn't covered in the original code in ``kuma/wiki/views/delete.py`` either.

Can you add a comment ``#  pragma: no branch`` to explicitly mark it as not covered, like this:

```
        if new_revision.pk != old_revision_pk: #  pragma: no branch
            document.schedule_rendering('max-age=0')
```
The two spaces after the ``#`` are important to ``flake8``.

kuma/users/views.py
```diff
@@ -179,74 +183,151 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+            revision = revision.previous
+        # If this is a new revision on an existing document, revert it
+        if revision.previous:
+            reverted = revert_document(request=request,
+                                       revision_id=revision.previous.id)
+            if reverted:
+                revisions_reverted_list.append(revision)
+            else:
+                # If the revert was unsuccessful, include this in the follow-up list
+                revisions_not_reverted_list.append(revision)
```
jwhitlock: This branch is uncovered in tests

kuma/users/views.py
```diff
@@ -179,74 +183,151 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+
+        # If this is a new document/translation, delete it
+        else:
+            deleted = delete_document(request=request,
+                                      document=revision.document)
+            if deleted:
+                revisions_deleted_list.append(revision)
+            else:
+                # If the delete was unsuccessful, include this in the follow-up list
+                revisions_not_deleted_list.append(revision)
```
jwhitlock: This branch is uncovered in tests

jwhitlock:
Practical tests went well. I skipped the easy cases and tried:

* Series of spam edits, last is spam - One revert to the previous good edit 😄 
* Spam, manually revert, more spam - Revert to the "manual revert" version 😄 
* Translate as spam - translation deleted 😄 
* Spam on translation - revert to before spam 😄 
* Translation, then spam, didn't mark initial translation as spam - Revert to initial translation 😄 

I'm still not sure what the right thing to do is if there is a gap in the spam chain, like this sequence (which actually happened in February 2016, too many times):

* Rev 1 - Page has good content
* Rev 2 - Spammer 1 spams page
* Rev 3 - Spammer 2 spams page
* Rev 4 - Spammer 1 spams page _again_

If you ban Spammer 1, the two revisions are submitted as spam, and the page reverts to Spammer 2's Rev 3, which still contains spam. I think this is the right thing for auto-reverting, but I think that the page should be called out as needing manual cleanup. I'm also not sure if this happens enough now to care about in this PR, or save it until the next attack. @stephaniehobson - do you have a strong opinion?

I've added some notes on test coverage, which will require adding or modifying some test code, but the functionality is good from a backend perspective. @stephaniehobson, I think it is safe to start a front-end review.

stephaniehobson:
> If you ban Spammer 1, the two revisions are submitted as spam, and the page reverts to Spammer 2's Rev 3, which still contains spam. I think this is the right thing for auto-reverting, but I think that the page should be called out as needing manual cleanup. 

In this senario does the script know that Spammer 2 is a spammer or that Rev 3 is spam, yet?

jwhitlock:
It could go either way, it depends which user is banned first.  The "middle" edit doesn't need to be spam either.  It could be this:

* Rev 100 - SPAM from spammer 1
* Rev 99 - A spelling fix from user 1
* Rev 98 - SPAM from spammer 1
* Rev 97 - Original good content

The best the algorithm can do in this case is revert to Rev 99, which still has the spam from Rev 98.  It will get removed manually eventually.

stephaniehobson:
The thing to do in this senario is Rev 99. If we know the "middle" edit is spam we can just throw our hands up and leave it to the moderator to sort it out.

stephaniehobson:
- [ ] Remove `<p>` around "You cannot ban yourself for spam.", it's already in a `<p>` from further up in the same block.
- [ ] Unchecking a checkbox should put that page in a "does not need follow up" section.
- [ ] Remove `<ul>` from around `none` when it appears under "New action by user:"
- [ ] A page that is not reverted because there is a newer edit by a legitimate user should appear under the "No actions taken" section.
- [ ] Remove `tabindex="0"` from the table rows.
- [ ] Change page title to link to the page, like the "view page" button does
- [ ] Add revision number in brackets after page title, link that to revision page
- [ ] Pages that had edits so they were are not reverted or deleted should be listed under "could not delete" or "could not revert" in "needs followup".
- [ ] Is it to late to put user name instead of number in URL?
- [ ] Reverted pages are not reliably being republished
- [ ] New pages should show a diff with an empty left side

I'm not totally clear on what goes on in the background but I thought I could replicate a failure to submit to Aksimet by disconnecting my laptop from the internet. So I disconnected it from the internet and did a one-click ban. The next page reported that it had successfully reported it to Akismet, which I find unlikely but the admin panel also lists it as reported so maybe this is not a problem with the one-click code?

It's my job to point out what needs to be tweaked but this is looking really great. I'm looking forward to getting this into production.

kuma/users/views.py
jbradberry: Shouldn't that be two spaces before the `#`?

jbradberry:
@jwhitlock @stephaniehobson Digging in a bit, I see that we aren't doing anything aside from sending the revisions to akismet to explicitly mark revisions as spam.  Perhaps we should also have something on the model (`revision.is_accepted = False`?) that we can use in the algorithm to decide which rev to rewind back to?

jbradberry:
To-do:

- [ ] add a test covering the case where there is a previous revision, but revert_document() returned a false value (L208, kuma/users/views.py)
- [ ] add a test covering the case where this is a new document, but delete_document() returned a false value (L218, kuma/users/views.py)
- [x] Remove \<p\> around "You cannot ban yourself for spam.", it's already in a \<p\> from further up in the same block.
- [ ] Unchecking a checkbox should put that page in a "does not need follow up" section.
- [x] Remove \<ul\> from around none when it appears under "New action by user:"
- [ ] A page that is not reverted because there is a newer edit by a legitimate user should appear under the "No actions taken" section.
- [x] Remove tabindex="0" from the table rows.
- [x] Change page title to link to the page, like the "view page" button does
- [x] Add revision number in brackets after page title, link that to revision page
- [ ] Pages that had edits so they were are not reverted or deleted should be listed under "could not delete" or "could not revert" in "needs followup".
- [ ] Is it to late to put user name instead of number in URL?
- [ ] Reverted pages are not reliably being republished
- [ ] New pages should show a diff with an empty left side

jbradberry:
@jwhitlock @stephaniehobson I would be in favor of changing the urls for these views to use the username instead of the id, but for consistency I feel that the previously existing `ban_user` view would have to be changed as well.  Would people be opposed to this?

stephaniehobson:
There should not be any permanent links (like bookmarks or anything) to the ban pages so I don't think that changing the URL will cause problems for anyone as long as we update all the places where we link to it in kuma. I think it's a better user experience over all.

jbradberry:
@stephaniehobson For the items that get unchecked, these already get put in the "Not identified as spam" subsection under "No actions taken".  Do you want the wording changed or did you really want these broken out into their own section?

Also, I think calling it "Does not need follow up" may be both too strongly worded and too vague, since we might have two (or more) revisions by the spammer on the same document, one which gets marked as spam and the other which does not.  Presumably the revision that was unchecked will wind up being the revision that the document gets reverted to, and may still contain spam as a whole, even if the edited portion does not.

stephaniehobson:
Hm. I was thinking of "not identified as spam" was a mirror to "submitted as spam" but reading the titles again maybe the fact that the others mention Akismet is distinction enough. You're right that the wording I proposed is misleading. Let's not change the wording right now.

This does not change the fact that the item I unchecked did not appear in that section on the follow up page. I don't have a record of what kind of change it was and I've been told another project is higher priority so I can't help diagnose, sorry. If you can't reproduce it yourself after a few times assume it was user error :P

jbradberry:
@stephaniehobson I kept meaning to ask you, and then I kept forgetting...  what does it look like when a reverted page fails to be republished?  I have no idea how to debug that at the moment.

jwhitlock:
Discussed "reverted page fails to be republished" on IRC.  The expected behavior is that when a page with spam is reverted to a previous version, the page does not display the spam any more.  In development VMs, reverting is seen in the page history, but the page itself sometimes (always?) still shows the spam revision.  More investigation is needed to see if this is another issue that only impacts developer VMs, or if more work is needed to avoid this issue in production.  I'll take on that task when I do the code review.

kuma/users/views.py
```diff
@@ -51,15 +51,13 @@
...
+def ban_user(request, username):
     """
     Ban a user.
     """
     User = get_user_model()
-    try:
-        user = User.objects.get(pk=user_id)
-    except User.DoesNotExist:
-        raise Http404
+    user = get_object_or_404(User, username=username)
```
jwhitlock: 😍 love switching to this pattern

kuma/users/views.py
```diff
@@ -179,61 +181,91 @@ def ban_user_and_cleanup_summary(request, user_id):
...
+        # to be reverted, do not revert any revisions
+        try:
+            revision.document.refresh_from_db(fields=['current_revision'])
+        except Document.DoesNotExist:
+            continue  # This document was previously deleted in this loop, continue
+        if revision.document.current_revision not in revisions_to_mark_as_spam_and_revert:
+            if revision.document_id not in previous_good_rev:
+                previous_good_rev[revision.document_id] = revision.document.current_revision
+
+            continue  # This document has a more current revision, no need to revert
```
jwhitlock: This line is [not covered](https://codecov.io/gh/mozilla/kuma/src/1203d19f6d3fe1b66e609a1e62b74525c7816744/kuma/users/views.py#L195), which confuses me greatly. This may be a bug in coverage, but I'll think more about it.

kuma/users/views.py
```diff
@@ -247,6 +279,55 @@ def ban_user_and_cleanup_summary(request, user_id):
...
 
+def revision_by_distinct_doc(list_of_revisions):
+    documents = {}
+    for rev in list_of_revisions:
+        documents.setdefault(rev.document_id, rev)
+        if documents[rev.document_id].id < rev.id:
+            documents[rev.document_id] = rev
+
+    return [documents[doc_id] for doc_id in sorted(documents)]
+
```
jwhitlock: I like avoiding the extra SQL calls, and this appears like it will work. Since queries are already sorted by ``-id``, the first revision for a document is always the highest ID, so the ``if documents[rev.document_id].id < rev.id`` branch is [never taken](https://codecov.io/gh/mozilla/kuma/src/1203d19f6d3fe1b66e609a1e62b74525c7816744/kuma/users/views.py#L286).  I don't think we should replace it with an assert though, so no change.

jwhitlock:
I was worried about the switch from user IDs to usernames, but I tested it with some of our legacy users with non-ascii names ([Stanisław](https://developer.mozilla.org/en-US/profiles/Stanisław), [행복한고니](https://developer.mozilla.org/en-US/profiles/행복한고니), [@Perlyshh_76](https://developer.mozilla.org/en-US/profiles/@Perlyshh_76)).  The only one that caused trouble was a user with a percent sign in their name (``Nelson%bolyard.com``, with many revisions like [this one](https://developer.mozilla.org/en-US/docs/HTML$revision/34870)), but his user profile page is unreachable anyway.  I will probably change his username manually (sorry Nelson).

jwhitlock:
There's some messiness in code coverage, and some potential test duplication to refactor. However, the ``user`` app tests were in bad shape at the start of this project, and the new code tests are at least twice as good as the old code. I'm putting "clean up user tests" on my TODO list for the future, when we're done adding functionality.

Practical tests work just as expected, especially the new sections for revisions that could not be reverted. I'm very happy this made it into the PR.

@stephaniehobson gave her OK to merge, but may have future changes for the front-end code, in the hypothetical Phase V.

Good work @emullaney, @jbradberry, @dchukhin, and @stephaniehobson on a very long PR process. This is ready for a test drive by the spam moderators.

PR #3904, opened 2016-07-18
Fix bug 1284781 Samples opened in JSFiddle will fail if CSS includes HTML escapes like &nbsp;

MatonAnthony:
Bug 1284781

Note : I haven't been able to fully reproduce the issue, I wasn't able to make the Codepen / JSFiddle buttons.

However, I think it should work.
PS : If someone knows how to make those buttons appear, I'm interested.

### DIFF
--- a/kuma/static/js/wiki-samples.js
+++ b/kuma/static/js/wiki-samples.js
@@ -55,6 +55,8 @@
     }
 
     function openSample(sampleCodeHost, section, title, htmlCode, cssCode, jsCode) {
+        // replace &nbsp; in CSS Samples to fix bug 1284781
+        var cssCleanCode = cssCode.replace(/\xA0/g, " ");
         //track the click and sample code host as event
         mdn.analytics.trackEvent({
             category: 'Samples',
@@ -65,9 +67,9 @@
         if(win.ga) ga('set', 'dimension8', 'Yes');
 
         if(sampleCodeHost === 'jsfiddle') {
-            openJSFiddle(title, htmlCode, cssCode, jsCode);
+            openJSFiddle(title, htmlCode, cssCleanCode, jsCode);
         } else if(sampleCodeHost === 'codepen') {
-            openCodepen(title, htmlCode, cssCode, jsCode);
+            openCodepen(title, htmlCode, cssCleanCode, jsCode);
         }
     }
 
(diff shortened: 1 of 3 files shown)


stephaniehobson:
@MatonAnthony You can get the samples to appear by going into the admin panel and creating a flag called `wiki_samples` and enabling it for all users.

MatonAnthony:
Nice, thank you

stephaniehobson:
To recreate the problem you can copy the "revision source" (link down at the bottom of the page) from https://developer.mozilla.org/en-US/docs/Web/CSS/clear$revision/1005963

stephaniehobson:
It doesn't fix it in my local environment :(

Hopefully you're able to replicate it in your local dev now and can test it out.

MatonAnthony:
I reproduced the bug on master and I don't reproduce it anymore on the PR branch.

stephaniehobson:
Reassigning to @jwhitlock since I suspect the problem is in my local dev. The fix *looks* good in theory, I don't know why it won't work for me.

As he said on IRC, it might take a few days before he has time to look.

jwhitlock:
The code extractor fix doesn't fix the issue for me either.  Digging in, this code doesn't use the extracted CSS, it uses JS to extract the sections from the rendered page (which also doesn't use the extracted CSS).  Sorry if that is confusing.

The fix that worked for me was updating the function ``openSample`` in [wiki-samples.js](https://github.com/mozilla/kuma/blob/master/kuma/static/js/wiki-samples.js#L57-L72):

```
        // replace &nbsp; in CSS samples
        var cssCleanCode = cssCode.replace(/\xA0/g, " ");
        //track the click and sample code host as event
        mdn.analytics.trackEvent({
            category: 'Samples',
            action: 'open-' + sampleCodeHost,
            label: section
        });
        // add user to segement that has used samples
        if(win.ga) ga('set', 'dimension8', 'Yes');

        if(sampleCodeHost === 'jsfiddle') {
            openJSFiddle(title, htmlCode, cssCleanCode, jsCode);
        } else if(sampleCodeHost === 'codepen') {
            openCodepen(title, htmlCode, cssCleanCode, jsCode);
        }
```
@MatonAnthony, could you add this change (or one like it) to the PR?

jwhitlock:
👍 

After adding a lot of ``&nbsp;`` to a local copy of the [clear](https://developer.mozilla.org/en-US/docs/Web/CSS/clear) page, and making sure I was using the new ``wiki-sample.js`` (lots of browser refreshes) the samples transfer correctly.

mdn sample
code pen sample
js fiddle sample

PR #3905, opened 2016-07-19
Fix bug 1287644: Track revision ID as GA dimension12

stephaniehobson:


jwhitlock:
Good idea, except it sent the page revision any time a ``document`` was sent to the template for rendering, such as the history page, editing page, and the rename document view.  I added a commit to restrict it to just viewing the document.

PR #3906, opened 2016-07-19
bug 957802, others - Update DB anonymize scripts

jwhitlock:
This PR updates the database anonymization scripts to be in sync with current production.

My process:
* Setup a MySQL server with the custom collation and an expanded ``max_allowed_packet``. I used a homebred-installed MySQL server, but the Vagrant VM could be used as well.
* Get a dump of the production database, load into MySQL as ``developer_mozilla_org``
* Run ``scripts/clone_db.py developer_mozilla_org``, with options as needed for your database update
* Load the generated anonymized dump as ``kuma``, verify with ``scripts/check_anonymize.sql | mysql kuma``

Alternatively, visually inspect the changes and trust that I ran them a half a dozen times.


jwhitlock:
Non-production code, so r=self

PR #3907, opened 2016-07-20
Fix Bug 1285635: Limit revision comment.

stephaniehobson:


stephaniehobson:
Just enough to be dangerous.

jwhitlock:
👍 works, adds, ``maxlength=255`` to the input as well, so the browser has a chance to enforce the limit. No migration is needed, either, so it's just the right amount of code.

PR #3908, opened 2016-07-21
New German strings for MDN

koehlermichael:
Pontoon is not ready yet and I want to get this done today.


jwhitlock:
We do not accept pull requests for updating translated strings. Please use [Pontoon](https://pontoon.mozilla.org/de/mdn/LC_MESSAGES/django.po/) instead, even if there is a delay with the strings being updated in Pontoon.

PR #3909, opened 2016-07-22
Bug 1256189 : Add new authorized attributes for <meter>

MatonAnthony:
Add new authorized attributes for <meter> (max, min, value, low, high, optimum, form).

It should be good. However I seem to have a caching issue, preventing me to test it.

### DIFF
--- a/kuma/wiki/constants.py
+++ b/kuma/wiki/constants.py
@@ -76,6 +76,8 @@ ALLOWED_ATTRIBUTES['li'] += ['data-default-state']
 ALLOWED_ATTRIBUTES['time'] += ['datetime']
 ALLOWED_ATTRIBUTES['ins'] = ['datetime']
 ALLOWED_ATTRIBUTES['del'] = ['datetime']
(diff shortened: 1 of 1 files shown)


jwhitlock:
It's not working, because the allowed attributes for the ``meter`` element are being overwritten:

```
$ ./manage.py shell_plus
>>> from kuma.wiki.constants import ALLOWED_ATTRIBUTES
>>> ALLOWED_ATTRIBUTES['meter']
['style', 'class', 'id', 'lang', 'dir', 'title']
```

You'll need to do something like what is done for the ``li``, ``time``, and ``ins`` elements in the same file.

MatonAnthony:
```
>>> from kuma.wiki.constants import ALLOWED_ATTRIBUTES
>>> ALLOWED_ATTRIBUTES['meter']
['style', 'class', 'id', 'lang', 'dir', 'title', 'max', 'min', 'value', 'low', 'high', 'optimum', 'form']
```

After the second commit, it return this.

jwhitlock:
The shell test works, and my local copy of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter works, after I convinced my VM to render it.

PR #3910, opened 2016-07-22
Email the spam watch list upon banning a user and cleaning up their mess

jbradberry:
Phase 4 of https://bugzilla.mozilla.org/show_bug.cgi?id=1259776

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -10,13 +10,16 @@ from allauth.socialaccount.models import SocialAccount
 from allauth.socialaccount.views import SignupView as BaseSignupView
 from constance import config
 from django import forms
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import permission_required
 from django.contrib.auth.models import Group
+from django.core.mail import send_mail
 from django.db import IntegrityError, transaction
 from django.db.models import Q
 from django.http import HttpResponseBadRequest, HttpResponseForbidden
 from django.shortcuts import get_object_or_404, redirect, render
+from django.template.loader import render_to_string
 from django.views.decorators.http import require_POST
 from django.utils.translation import ugettext_lazy as _
 from honeypot.decorators import verify_honeypot_value
@@ -274,6 +277,9 @@ def ban_user_and_cleanup_summary(request, username):
                'needs_follow_up': needs_follow_up,
                'no_actions_taken': no_actions_taken}
 
+    # Send an email to the spam watch mailing list.
+    ban_and_revert_notification(user, request.user, context)
+
     return render(request,
                   'users/ban_user_and_cleanup_summary.html',
                   context)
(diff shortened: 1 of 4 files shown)


jwhitlock:
Thanks for the PR.  It appears to do the right thing, from reading the code.  It will be Monday afternoon at the earliest before I can dig into the PR and test it in my VM, but I have some high-level feedback to get started.

Because this is a plain text email and not HTML, whitespace matters. Use the [Jinja2 whitespace control features](http://jinja.pocoo.org/docs/dev/templates/#whitespace-control) to get just the right whitespace.  See the [wiki emails](https://github.com/mozilla/kuma/tree/master/kuma/wiki/jinja2/wiki/email) for examples of whitespace control.

There should be at least one test of the email content, probably in ```kuma/users/tests/test_views.py```. The test should confirm that one email was sent. I like to test the content of plain text as well (testing the full content against the expected content), but it looks like that's not standard in the kuma codebase, so only do it if it makes it easier to clean up the whitespace.

I considered asking you to change the logic to send the email asynchronously. This will both speed up the ban process, and restrict any email sending issues to the async thread. Making it async will require some manual serialization, to turn the revision and user instances into Plain Old Data so it can be JSON serialized for an async task. That's easy enough, but we should use the tool for a bit, to be sure we have the right data first. So, no change, continue sending the email in the request loop for this PR.

jbradberry:
@jwhitlock Ok, sounds good.  No hurry on a more detailed review, I should have marked this WIP.

jbradberry:
@jwhitlock This is ready for review.

kuma/users/views.py
```diff
@@ -289,6 +295,17 @@ def revision_by_distinct_doc(list_of_revisions):
     return [documents[doc_id] for doc_id in sorted(documents)]
 
 
+def ban_and_revert_notification(spammer, moderator, info):
+    subject = u'[MDN] %s has been banned by %s' % (spammer, moderator)
+    context = {'spammer': spammer,
+               'moderator': moderator}
+    context.update(info)
```
safwanrahman: Not a good Idea to change the context later on with info. In future if we can use this function for other view or change any to the passed context, so it will raise issue.
This thing can be done like
```python
def ban_and_revert_notification(spammer, moderator, context):
    subject = u'[MDN] %s has been banned by %s' % (spammer, moderator)
    context['spammer'] = spammer
    context['moderator'] = moderator
```

jbradberry: Actually, it's a fairly standard pattern to have a constructed context be updated by kwargs (or whatever) passed in, so that it can be overridden by (perhaps unanticipated) callers, viz: https://github.com/django/django/blob/1.9.8/django/views/generic/detail.py#L102-L108

jbradberry: Also, the `info` being passed in is the `context` variable of the caller, and I think it would be undesirable to mutate it.

jwhitlock:
I like the email test. I love the ``TODO`` cleanup.

I added a ``print`` step to ``ban_and_revert_notification``, and the emails look good under different conditions.  Good work on the whitespace cleanup.

I ran the "Ban User and Cleanup" one more time, and looked at the email in ``/tmp/kuma-messages/``. It works as expected.

👍 let's get this in front of the spam moderators

PR #3911, opened 2016-07-26
[Bug 1260477] Invalid authorization header raise 500 error

safwanrahman:
@jwhitlock r?


jwhitlock:
👍 Works well, good test.

Can you remove the try/except block for importing ``functools`` while we are touching this file? It's for Python 2.6, which isn't supported anymore.

safwanrahman:
@jwhitlock Removed ``Try/Except`` block.

PR #3912, opened 2016-07-26
[Bug 859499] Show locale code with locale name in language button

safwanrahman:
A very small fix. @jwhitlock r?


jwhitlock:
This works for me (language drop-down, "add a translation", right-to-left order), but it's front-end code, so assigning to @stephaniehobson

stephaniehobson:
👍

PR #3913, opened 2016-07-26
bug 1260257 - Disable signup with Persona

jwhitlock:
This PR is broken into two commits.

In the first, the social login tests have been refactored to use a a new mixin, ``SocialTestMixin``, that provides class functions ``persona_login()`` and ``github_login()`` to execute a mocked social authentication with good defaults, and the ability to override the defaults as needed.  These result in big changes to the test code, and an ugly diff.

The second commit changes the production code, in ``kuma/users/adapters.py``, which now forbids creating a new account with Persona, and in ``kuma/users/jinja2/account/signup_closed.html``, which has a custom message for Persona users:

profile creation disabled

Disabling Persona signups requires changing some tests and dropping others, but the test diff is a lot nicer due to the work in the first commit.

r @jgmize? @groovecoder?

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -455,43 +455,29 @@ class SignupView(BaseSignupView):
         self.matching_user = None
         initial_username = form.initial.get('username', None)
         # For GitHub users, see if we can find matching user by username
-        if self.sociallogin.account.provider == 'github':
-            User = get_user_model()
-            try:
-                self.matching_user = User.objects.get(username=initial_username)
-                # deleting the initial username because we found a matching user
-                del form.initial['username']
-            except User.DoesNotExist:
-                pass
+        assert self.sociallogin.account.provider == 'github'
+        User = get_user_model()
+        try:
+            self.matching_user = User.objects.get(username=initial_username)
+            # deleting the initial username because we found a matching user
+            del form.initial['username']
+        except User.DoesNotExist:
+            pass
 
         email = self.sociallogin.account.extra_data.get('email') or None
-
-        # For Persona users, see if we can find matching user by email address
-        if self.sociallogin.account.provider == 'persona':
-            try:
-                matching_addresses = EmailAddress.objects.filter(email=email,
-                                                                 verified=True)
-                matching_emailaddress = matching_addresses[0]
-                self.matching_user = matching_emailaddress.user
-                email_address = {'email': email,
-                                 'verified': matching_emailaddress.verified,
-                                 'primary': matching_emailaddress.primary}
-                self.email_addresses[email] = email_address
-            except IndexError:
-                pass
-
         extra_email_addresses = (self.sociallogin
                                      .account
                                      .extra_data
-                                     .get('email_addresses', None))
+                                     .get('email_addresses')) or []
 
         # if we didn't get any extra email addresses from the provider
         # but the default email is available, simply hide the form widget
-        if extra_email_addresses is None and email is not None:
+        if not extra_email_addresses and email is not None:
             form.fields['email'].widget = forms.HiddenInput()
 
-        # if there are extra email addresses from the provider (like GitHub)
-        elif extra_email_addresses is not None:
+        # let the user choose from provider's extra email addresses, or enter
+        # a new one.
+        else:
             # build a mapping of the email addresses to their other values
             # to be used later for resetting the social accounts email addresses
             for email_address in extra_email_addresses:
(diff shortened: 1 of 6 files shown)


jwhitlock:
I'm not happy with those coverage numbers.  I'm removing some Persona branches, adding some GitHub tests in a new commit.

safwanrahman:
I think, for dev environment, Persona Sign up should be available. Otherwise, it will be quite hard to setup a new dev environment for a new contributor.
So having a ``waffle flag`` or ``debug=True`` will be a good idea.

jwhitlock:
On the one hand, adding a ``persona_signups`` waffle switch will make the code more complicated, require additional tests, and be removed anyway in a few months when the Persona servers are decommissioned.  It is not hard to [setup GitHub](https://kuma.readthedocs.io/en/latest/installation.html#advanced-enable-github-auth), and it is possible for a developer to manually created password-backed accounts.  This makes me want to rip off the bandaid and just remove it.

However, I doubt we will have the bandwidth to implement an alternate login system for testing and development ([bug 1286289](https://bugzilla.mozilla.org/show_bug.cgi?id=1286289)), so the waffle switch (not flag) seems like a good idea for the interim.  I'll think some more.

jwhitlock:
Added a commit for the code coverage.  I'm leaning toward a waffle switch such as ``persona_signups_enabled``, default off, to allow using Persona in development environments until we have an alternate method for testing ([bug 1286289](https://bugzilla.mozilla.org/show_bug.cgi?id=1286289)), which will significantly change the PR, so leaving on the not ready tag for a while.

stephaniehobson:
Not able to create an account with Persona. 
Able to create one with GitHub.
Able to associate GitHub with existing Persona account.

\o/

We should probably take the "Connect with Persona" button off the Account Connections page. And once Persona is off entirely we'll have to do something with this page :/

jgmize:
> On the one hand, adding a persona_signups waffle switch will make the code more complicated, require additional tests, and be removed anyway in a few months when the Persona servers are decommissioned. It is not hard to setup GitHub, and it is possible for a developer to manually created password-backed accounts. This makes me want to rip off the bandaid and just remove it.

I agree, manual password-backed accounts should be sufficient for local dev, and we should remove persona completely ASAP.

stephaniehobson:
We can manually create password backed accounts?!?

jwhitlock:
Yes, but the experience isn't great:

1. As superuser, click "ADD USER" on the user changelist page (https://developer-local.allizom.org/admin/users/user/add/)
2. Enter a username and the password, save it
3. Select "Staff Status" in permissions to set ``is_staff``, save
4. In a new browser, login to the admin with the new user 
(https://developer-local.allizom.org/admin/login/). You aren't allowed to see anything, but you are now logged in, can get to the site (click VIEW SITE or update your URL), can update your profile, etc.

jwhitlock:
I've opened PR #3914, which takes the alternate approach of adding a ``enable_persona_signup`` waffle switch.  I still prefer this approach of this PR, of removing Persona signups and living with the consequences in the development environment

stephaniehobson:
I concur, remove Pesona signups and live with the consequences.

stephaniehobson:
Bah, wrong button.

jwhitlock:
Force-push to rebase on master.

@stephaniehobson do you think the "Connect with Persona" changes belong in this PR, or can we defer that for future work?

stephaniehobson:
Probably it should be done separately in a more thoughtful manner.

kuma/users/views.py
```diff
@@ -516,7 +502,10 @@ def get_form(self, form_class):
                     label = _('%(email)s Unverified')
                 next_email = email_address['email']
                 choices.append((next_email, label % {'email': next_email}))
-            choices.append((form.other_email_value, _('Other:')))
+            if extra_email_addresses:
+                choices.append((form.other_email_value, _('Other:')))
+            else:
+                choices.append((form.other_email_value, _('Email:')))
```
jgmize: I think we should require that users signing up with github have at least one verified email, and present a message to them asking that they go verify their email with github in order to sign up for MDN if there aren't any. Ideally a user signing up with github should never type an email address into an MDN form.

jwhitlock: This is a feature request, out of scope for this PR. Please open a bug:

https://bugzilla.mozilla.org/enter_bug.cgi?product=Mozilla%20Developer%20Network&component=Profiles#h=detail%7Cchange

jgmize: Agreed, will do.

jgmize:
Nice work, @jwhitlock

PR #3914, opened 2016-07-27
bug 1260257 - Disable Persona signup unless switch is active

jwhitlock:
This is an alternate for PR #3913, which removes Persona signups, related code, and tests.  However, there isn't a good documented alternate for manual testing ([bug 1286289](https://bugzilla.mozilla.org/show_bug.cgi?id=1286289)), so it may be worthwhile to keep Persona around.

In the first commit, the social login tests have been refactored to use a a new mixin, ``SocialTestMixin``, that provides class functions ``persona_login()`` and ``github_login()`` to execute a mocked social authentication with good defaults, and the ability to override the defaults as needed.  These result in big changes to the test code, and an ugly diff.

The second commit changes the production code, in ``kuma/users/adapters.py``, which now forbids creating a new account with Persona unless the waffle switch ``enable_persona_signup`` is enabled (default off), and in ``kuma/users/jinja2/account/signup_closed.html``, which has a custom message for Persona users:

profile creation disabled

Disabling Persona signups requires explicitly setting the ``enable_persona_signup`` switch for tests that exercise Persona signups.  This second commit is cleaner because of the work in the first commit.

### DIFF
--- a/docs/feature-toggles.rst
+++ b/docs/feature-toggles.rst
@@ -23,6 +23,7 @@ Switches
 * ``dumb_doc_urls`` - (deprecated) Disable the render-time changing of /docs/
   URLs to the equivalent Zone URLs (see `PR 3331`_ for reasoning).
 * ``enable_optimizely`` - Enable the Optimizely JavaScript
+* ``enable_persona_signup`` - Enable account signup with Persona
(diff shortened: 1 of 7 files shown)


jwhitlock:
Closing in favor of #3913. To the future!

PR #3915, opened 2016-07-27
bug 1262541 - Add an "All Topics" filter that will clear out checkboxes on search page

emullaney:
Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1262541
Add an 'All Topics' filter that allows user to search all topics. 
Clear out that value when other topics are selected via search.js. 
Add a filter test.

Ready for review by @jwhitlock

### DIFF
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -190,7 +190,8 @@ class DatabaseFilterBackend(BaseFilterBackend):
             else:
                 if filter_tags:
                     facet_params = F('term', tags=filter_tags[0])
-            active_facets.append((serialized_filter['slug'], facet_params))
+            if len(filter_tags):
+                active_facets.append((serialized_filter['slug'], facet_params))
 
         if active_filters:
             if len(active_filters) == 1:
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -20,7 +20,11 @@ def get_filters(getter_func):
 
     this will return `['css', 'html']`.
 
+    If the given URL contains 'none', then no filters should be applied.
+
     """
+    if getter_func("none"):
+        return [u'none']
     filters = collections.OrderedDict()
     for slug in FilterGroup.objects.values_list('slug', flat=True):
         for filters_slug in getter_func(slug, []):
@@ -184,7 +188,8 @@ class DatabaseFilterBackend(BaseFilterBackend):
             if len(filter_tags) > 1:
                 facet_params = F('terms', tags=list(filter_tags))
             else:
-                facet_params = F('term', tags=filter_tags[0])
+                if filter_tags:
+                    facet_params = F('term', tags=filter_tags[0])
             active_facets.append((serialized_filter['slug'], facet_params))
 
         if active_filters:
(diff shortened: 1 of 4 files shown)


jwhitlock:
Thanks! It will probably be next week (after the current sprint ends) when I can look at this PR. @stephaniehobson will need to review the front-end code.

emullaney:
Next week is perfect, thanks @jwhitlock and @stephaniehobson!

kuma/search/jinja2/search/results.html
```diff
@@ -171,6 +171,20 @@ <h2 class="offscreen">{{ _('Search') }}</h2>
...
+        <label>
+          <legend>
+          {% if "none" in selected_filters %}
+            <i aria-hidden="true" class="icon-check"></i>
+            <input id="no_filter" type="checkbox" class="offscreen" name="none" checked="checked" value="none">
+          {% else %}
+            <i aria-hidden="true" class="icon-check-empty"></i>
+            <input id="no_filter" type="checkbox" class="offscreen" name="none" value="none">
+          {% endif %}
+          <span class="filter-name">All Topics</span>
```
stephaniehobson: Needs localization wrapper: `{{ _('All Topics') }}`

kuma/search/jinja2/search/results.html
```diff
@@ -171,6 +171,20 @@ <h2 class="offscreen">{{ _('Search') }}</h2>
       <!-- right filters -->
       <div class="column-strip search-results-filters">
         <fieldset>
+
+        <label>
+          <legend>
```
stephaniehobson: Let's remove the legend, I'm okay with that making this smaller.

stephaniehobson:
R after changes :)

emullaney:
Thanks @stephaniehobson. These 2 small changes have been implemented.

kuma/search/filters.py
```diff
@@ -184,7 +188,8 @@ def filter_queryset(self, request, queryset, view):
             if len(filter_tags) > 1:
                 facet_params = F('terms', tags=list(filter_tags))
             else:
-                facet_params = F('term', tags=filter_tags[0])
+                if filter_tags:
+                    facet_params = F('term', tags=filter_tags[0])
```
jwhitlock: This will still break the code when ``filter_tags`` is empty, because ``facet_params`` will be undefined and the next line that uses ``facet_params`` will fail.

emullaney: You are correct. Ideas about what I should set facet_params to as default?

jwhitlock: Skip appending to ``active_facets`` if ``filter_tags`` is empty.

emullaney: Oh sure.

emullaney: Okay, I've made this change as requested. Feel free to give this a try and see if it works as intended. @jwhitlock @stephaniehobson

jwhitlock:
I'm not thrilled with the test coverage in ``kuma/search/filters.py``, but it was worse before, and MDN searchers don't care about test coverage.  Thanks for your work on this, @emullaney

PR #3917, opened 2016-07-27
[Bug 1275387] CKEditor build script uses dev version, not release

safwanrahman:
I can not understand how this script was used to build release version.
BTW, the version we use is ``4.5.8`` not ``4.5.7``

In the past, we added the submodule and we checkout according to the git commit hash, and then run the script to build the editor.
But, later we have removed the submodule but there was a little bug in the script.

@jwhitlock r?

### DIFF
--- a/kuma/static/js/libs/ckeditor/source/build.sh
+++ b/kuma/static/js/libs/ckeditor/source/build.sh
@@ -21,7 +21,7 @@ target="../build"
 echo ""
 echo "Pulling down CKEditor from GitHub..."
 rm -rf ckeditor/
-git clone https://github.com/ckeditor/ckeditor-dev.git ckeditor
+git clone -b $CKEDITOR_VERSION --single-branch https://github.com/ckeditor/ckeditor-dev.git ckeditor
 rm -rf ckeditor/.git
 
 


jwhitlock:
Thanks! The diff makes sense, but I'll need to try this out, but there's some deadlines coming up, so it will be next week.

The bug is that we were checking out the master branch, not the released version.  The CKEditor devs increment the version number in the master branch when they start development work. So, we're currently not shipping 4.5.7 or 4.5.8, but some pre-release code between the two.

It looks like this code does the right thing, explicitly checking out a branch rather than implicitly checking out the in-development master branch.  My test will be to compare the checked-out code to the release code.

stephaniehobson:
🎉 Excited to see this.

a2sheppy:
Really looking forward to this landing so I can start my gradual work on improving the editor configuration, including the toolbar, hotkeys, and plugin selection.

jwhitlock:
Verified that it installs the correct version. Good work, @safwanrahman.

jwhitlock:
@a2sheppy there's a couple of other issues we need to resolve, so you may want to hold off on your changes for a few more days:

* [bug 1291969](https://bugzilla.mozilla.org/show_bug.cgi?id=1291969) - Un-versioned plugins are installed
* [bug 1274063](https://bugzilla.mozilla.org/show_bug.cgi?id=1274063) - Upgrade to CKEditor 4.5.10
* PR #3930 - Update CKEditor docs

PR #3919, opened 2016-07-29
Fix bug 1289589 : Remove edit button from templates if the user cannot edit them

MatonAnthony:
Fix bug 1289589 : [Bug 1289589 - Edit button shows up when user doesn't have edit permission for a page ](https://bugzilla.mozilla.org/show_bug.cgi?id=1289589)

- Remove edit button from templates if the user cannot edit them.
- Remove part of code which was forgot during a clean-up

I separated this into two different commits because removing the old code didn't seem to have a reason to be in the same commit, though I might be wrong.

### DIFF
--- a/kuma/wiki/jinja2/wiki/includes/buttons.html
+++ b/kuma/wiki/jinja2/wiki/includes/buttons.html
@@ -37,7 +37,9 @@
         </div>
       </li>{% endif %}
 
+      {% if (not document.is_template) or (user.has_perm('wiki.change_template_document')) %}
       <li class="page-buttons-edit"><a href="{{ edit_link }}" class="button" data-optimizely-hook="button-edit-doc" id="edit-button">{{ _('Edit') }}<i aria-hidden="true" class="icon-pencil"></i></a></li>
+      {% endif %}
 
         {% if user.is_authenticated() and document %}
         <li><button id="watch-menu" class="only-icon" aria-haspopup="true" aria-owns="watch-menu-submenu" aria-expanded="false"><span>{{ _('Watch') }}</span><i aria-hidden="true" class="icon-eye"></i></button>
--- a/kuma/wiki/jinja2/wiki/includes/buttons.html
+++ b/kuma/wiki/jinja2/wiki/includes/buttons.html
@@ -57,8 +57,6 @@
             <div class="title">{{ _('Advanced') }}</div>
             <ul>
                 <li><a href="{{ url('wiki.document_revisions', document.slug) }}" rel="nofollow, noindex">{{_('History')}}</a></li>
-                {% if user.is_authenticated() and document %}
-                </li>{% endif %}
                 {% if user.is_authenticated() and not document.is_template %}
                   <li><a href="{{ url('wiki.create') }}?parent={{ document.id }}" rel="nofollow, noindex">{{ _('New sub-article') }}</a></li>
                   <li><a href="{{ url('wiki.create') }}?clone={{ document.id }}{% if document.parent_topic %}&amp;parent={{ document.parent_topic.id }}{% endif %}" rel="nofollow, noindex">{{ _('Clone this article') }}</a></li>


stephaniehobson:
Works as expected 👍 

Verified that lines 60 & 61 are not only unnecessary but actually broken. That was a good catch.

R+ but leaving it for @jwhitlock because Python.

kuma/wiki/jinja2/wiki/includes/buttons.html
```diff
@@ -37,7 +37,9 @@
         </div>
       </li>{% endif %}
 
+      {% if (not document.is_template) or (document.is_template and user.is_authenticated() and user.has_perm('wiki.change_template_document')) %}
```
jwhitlock: The test ``user.is_authenticated()`` is unnecessary. The [AnonymousUser](https://docs.djangoproject.com/en/1.10/ref/contrib/auth/#django.contrib.auth.models.AnonymousUser) will return ``False`` for ``user.has_perm('wiki.change_template_document')``. So, a less verbose, equivalent condition would be:

```
     {% if (not document.is_template) or user.has_perm('wiki.change_template_document') %}
```

jwhitlock:
The code works for me.  Please simplify the condition in the Jinja template, and we'll be ready.

MatonAnthony:
I simplified the condition but I don't manage to rebase them since the cleanup commit is in the middle.
If someone knows a tips, I'm interested.

jwhitlock:
I can use the "squash commits" to combine these three into one. If you want to preserve this as two commits (the necessary change and the unrelated cleanup), here's my rebase process:

```
git fetch -a             # Update with branches from all remote repos
git checkout master      # First, checkout the branch I'm rebasing on
git pull                 # Collect changes since last time I checked it out
git checkout bug-1289589 # Checkout the branch
git reset @{u} --hard    # Reset my local branch to the upstream branch
                         #  --hard means "also discard changed files"
git rebase -i master     # Start interactive rebase #1
```

This opens your text editor. I use VIM, as ``/usr/bin/vim``. You can see yours with ``git config --get core.editor``, and set it with ``git config --add core.editor /path/to/editor``.  Here's what I see:

```
pick e291a9b Remove edit button from templates if the user cannot edit them
pick c5009ff Removing part of code which was forgot during a clean-up
pick f14d1ad Simplify the jinja2 template condition

# Rebase 870a8d9..f14d1ad onto 870a8d9 (3 command(s))
#
# Commands:
# p, pick = use commit
# r, reword = use commit, but edit the commit message
# e, edit = use commit, but stop for amending
# s, squash = use commit, but meld into previous commit
# f, fixup = like "squash", but discard this commit's log message
# x, exec = run command (the rest of the line) using shell
# d, drop = remove commit
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
# Note that empty commits are commented out
```

The commands are worth reading.  As the notes say, ``pick`` means to use the commit.  I don't change anything for this first rebase, and take the default commands. This first rebase it just to move the branch work up to the current master branch. I like to do just one thing at a time with each rebase, to minimize the changes and reduce the amount of stuff that can go wrong.  I save the file and exit the editor, starting the rebase process.  In this case, no problems:

```
$ git rebase -i master    # The rebase #1 from above
Successfully rebased and updated refs/heads/bug-1289589.
$ git status
On branch bug-1289589
Your branch and 'MatonAnthony/bug-1289589' have diverged,
and have 173 and 3 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
```

This is a good result. My local checkout of the branch is 173 commits ahead of the one on GitHub, because it has moved all the way up to the master branch. But, we're not done rebasing. I run ``git rebase -i master`` again, and move the commit:


```
pick 751b682 Remove edit button from templates if the user cannot edit them
pick 01af713 Simplify the jinja2 template condition
pick f279c7f Removing part of code which was forgot during a clean-up

# ... Same instructions as above
```

The change is to move the "Simplify the jinja2 template condition" to the middle. This is easy to do in Vim (delete a line, insert the deleted line).  I save the rebase instructions, exit the editor, and again a successful rebase.

Next rebase - there's no need to keep the middle commit now. Once again, ``git rebase -i master``:

```
pick 751b682 Remove edit button from templates if the user cannot edit them
f a844298 Simplify the jinja2 template condition
pick f6dfe83 Removing part of code which was forgot during a clean-up
```

In this case, I'm using ``f`` as the shorthand for ``fixup`` - combine this commit with the previous one, and throw away the commit message, or "fix" the previous commit. I write the rebase instructions and quit the text editor, and now there are just two commits.

At this point, I may reverse the order of the commits, with the narrative being "This is what the code should have looked like when I started", followed by "here are my changes to fix the bug".  This is important for big changes (the "cleanup" commits could be merged even if the fix needs work), but not critical for this PR.

Not done yet. The commits don't follow the [contribution guidelines](https://github.com/mozilla/kuma/blob/master/CONTRIBUTING.md#conventions) for commit messages, so I fix the commit messages with another ``git rebase -i master``, this time using ``r`` for ``reword``, meaning I just want to change the commit message:

```
r f63eb2e Remove edit button from templates if the user cannot edit them
r b158e67 Removing part of code which was forgot during a clean-up
```

I write the instructions and quit the editor, which immediately opens again asking for the commit message. I add "bug 1289589:" to the front of the messages.

```
fix bug 1289589: Remove edit button from templates

If the user doesn't have permission to edit templates, don't show the
edit button.
```

and

```
bug 1289589: Cleanup template code

Remove an unnecessary </li> that must have been leftover from a previous
cleanup.
```

I use MacVim for my editor, which sees that I'm editing a git commit message, and uses highlighting to enforce some [rules for good git commit messages](http://chris.beams.io/posts/git-commit/ ):

* Limit the subject to 50 characters
* Limit the content lines to 72 characters

OK, now I have a branch with 2 commits that are focused, well commented, and will help the next developer who has to follow the history of this file over then next few years.  This is my preference for PRs, to document the changes to the codebase. Others prefer to keep the "this happened and then this happened and then I made this change due to a review" narrative of a long series of commits.  They are wrong.

If I had push permissions, my next step would be:

```
git push -f
```

and add a PR comment to say "Force push to combine and reword commits".  But, it's not my branch, so instead I throw away my changes:

```
git reset @{u} --hard
```

and post this to GitHub (my other blog).

jwhitlock:
Thank you, the rebase worked.

You'll need to update ``MatonAnthony:master`` before doing future work - it is way behind ``mozilla:master``.

PR #3920, opened 2016-07-29
Work around MySQL's case-insensitive collation logic for tag assignment

jbradberry:
Previously, if there were duplicate tags under the case-insensitive matching rules (e.g. JavaScript and javascript), both tags would get applied even when only one was intended.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1271552

### DIFF
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -586,29 +586,32 @@ class RevisionForm(AkismetCheckFormMixin, forms.ModelForm):
         tags = self.cleaned_data['tags']
         cleaned_tags = []
 
-        if tags:
-            for tag in parse_tags(tags):
-                # Note: The exact match query doesn't work correctly with
-                # MySQL with regards to case-sensitivity. If we move to
-                # Postgresql in the future this code may need to change.
-                doc_tag = (DocumentTag.objects.filter(name__exact=tag)
-                                              .values_list('name', flat=True))
+        for tag in parse_tags(tags):
+            # Note: The exact match query doesn't work correctly with MySQL
+            # with regards to case-sensitivity, due to the case-insensitive
+            # collation.  If we move to Postgresql in the future this code will
+            # need to change.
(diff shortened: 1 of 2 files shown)


jbradberry:
Note: I am a bit stuck for writing tests for this.  The issue arises on production because the production db is lacking the uniqueness constraint on `DocumentTag.name`.  However, the model does have this constraint, so the test db does, so attempts to create a set of duplicate tags fails with an `IntegrityError`.  I did briefly go down the path of dropping the constraint with some raw sql, but Django's test db machinery wasn't putting it back again afterwards.

jwhitlock:
Do you think the existing code is working, but the unique index is not applied in production? It is possible the fix is 1) manual merge duplicate tags, and 2) manually add the unique index.

jbradberry:
@jwhitlock Yes, I believe that the existing code would work (though could be simplified somewhat) if the index were put in place and the duplicate tags removed.

jwhitlock:
I've said this other places, but not here. The next action is mine, to manually clean-up the duplicate tags and apply the unique index in staging and production.

jwhitlock:
Next steps seem to be:

1. Reproduce the original bug
2. Manually remove duplicates
3. Manually apply the unique index
4. See if the bug is still present

I'm going to try it in a development VM with an anonymized production database first. If that works, it may be possible to ship this as a database migration.

Also, I'm conspiring to eliminate the custom collation, but that's out of scope.

jwhitlock:
@jbradberry thanks for starting this, but I'm going to close the PR. I've spent my day diving into the production database schema, and I've found lots of things. I think work-around code will not be an effective bandaid, and the root causes should be tackled instead.  I've put my notes in [bug 1271552](https://bugzilla.mozilla.org/show_bug.cgi?id=1271552#c8), and opened some additional bugs to cover the work.

jbradberry:
@jwhitlock yep, that makes sense, let's not band-aid the problem with code when we can do a real fix.

PR #3921, opened 2016-07-30
Fix bug 1277388 :  Missing space in French translation on edit pages

MatonAnthony:
Fix bug 1277388 : [Bug 1277388 - Missing space in French translation on edit pages ](https://bugzilla.mozilla.org/show_bug.cgi?id=1277388)

- I removed the comma.
- Translators will have to add it, I don't think we can (and if we can I don't know how) do a mass edit.

### DIFF
--- a/kuma/wiki/jinja2/wiki/translate.html
+++ b/kuma/wiki/jinja2/wiki/translate.html
@@ -80,7 +80,7 @@
             <div id="content-fields">
               <article class="approved text-content">
                 <header>
-                    <h3 class="approved-title">{{ _('Approved %(locale)s version', locale=default_locale) }}:</h3>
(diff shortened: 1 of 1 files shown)


safwanrahman:
The Punctuation mark should be in the translation string

jwhitlock:
@safwanrahman's right, the fix is to move the colon inside the translated string:

```
 <h3 class="approved-title">{{ _('Approved %(locale)s version:', locale=default_locale) }}</h3>
```

Then the French localizers can add the space before the colon, like they do for [other strings with colons](https://pontoon.mozilla.org/fr/mdn/LC_MESSAGES/django.po/?search=%3A&string=68067).

MatonAnthony:
I'm sorry, I forgot a part of the fix... 
I'll fix it when I'm home. It'll remind me to not code just after the wake up time.

jwhitlock:
Thanks for this fix!

PR #3922, opened 2016-07-31
Fix bug 1287806 - X-Frame-Options header incorrect for attachments 

MatonAnthony:
Fix bug 1287806 [Bug 1287806 - X-Frame-Options header incorrect for attachments ](https://bugzilla.mozilla.org/show_bug.cgi?id=1287806)

- Change the XFO in favor of a CSP.
- Change the unit test to reflect the changes.

### DIFF
--- a/kuma/attachments/tests/test_views.py
+++ b/kuma/attachments/tests/test_views.py
@@ -109,8 +109,8 @@ class AttachmentViewTests(UserTestCase, WikiTestCase):
 
         response = self.client.get(url, HTTP_HOST=settings.ATTACHMENT_HOST)
         self.assertTrue(response.streaming)
-        self.assertEqual(response['x-frame-options'],
-                         'ALLOW-FROM: %s' % settings.DOMAIN)
+        self.assertEqual(response['Content-Security-Policy'],
+                         'frame-ancestors %s' % settings.DOMAIN)
         self.assertEqual(response.status_code, 200)
         self.assertIn('Last-Modified', response)
         self.assertNotIn('1970', response['Last-Modified'])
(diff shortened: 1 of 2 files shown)


jwhitlock:
This will be an interesting one to test. I have not added CSP to a site before, so I'm [asking for advice](https://bugzilla.mozilla.org/show_bug.cgi?id=1287806#c2) from April King.

stephaniehobson:
Yeah, CSPs make me nervous because they're easy to mess up on and MDN is so complicated. I don't have time to look at this while I'm away this week though :/

jwhitlock:
@MatonAnthony, thanks for the PR, but I'm going to delay adding CSP headers.

Before the PR, the current VM environment puts out this header:

```
X-Frame-Options: ALLOW-FROM: developer-local.allizom.org
```

With the patch, the current VM environment puts out these headers:

```
Content-Security-Policy: frame-ancestors developer-local.allizom.org
X-Frame-Options: DENY
```

This suggests Apache or other middleware is adding a ``X-Frame-Options: DENY`` header, unless the code overwrites it.  I don't know which wins if both the ``X-Frame-Options`` and the ``Content-Security-Policy: frame-ancestors`` is set.  I suspect the ``DENY`` would win, but it's hard to tell unless we push to production.

I think the fix isn't as simple as a code patch. We need to setup a CSP reporting endpoint, survey the data being served by MDN, figure out where the headers are being set (Django Middleware, views, Apache), and come up with a plan to deploy CSP without blocking content for users.  I don't want to start down the CSP path with this one PR.

I'm closing the PR. If you are still interested, please fix the ``X-Frame-Options`` as suggested, without replacing it with a ``Content-Security-Policy`` header. Even that may break the site, we'll have to be careful in deploying it.

PR #3923, opened 2016-08-02
bug 1268609 - Do not delete draft after save click, to prevent data loss

emullaney:
https://bugzilla.mozilla.org/show_bug.cgi?id=1268609

Remove call to clearDraft() method on save click.
Disable auto save on save click (which is part of the clearDraft() method).

Line 509 clears out this draft when we return to the edit page, after saving.
https://github.com/mozilla/kuma/compare/master...caktus:new-issue-1268609-save-draft-before-publish?expand=1#diff-e318b5ff8f27b6a7605c11e49310b1aaR509 

cc @stephaniehobson @jwhitlock

### DIFF
--- a/kuma/static/js/helpfulness.js
+++ b/kuma/static/js/helpfulness.js
@@ -4,6 +4,8 @@
     var articleTracker = doc.location.pathname + '#answered-helpful';
     // this feature requires localStorage
     if (win.mdn.features.localStorage) {
+        // clear out any drafts from localStorage, now that we are not on the edit page
+        localStorage.removeItem('draft/edit' + location.pathname);
         var ignore = localStorage.getItem('helpful-ignore') === 'true'; // true if ever clicked ignore
         var articleAskedRecently = parseInt(localStorage.getItem(articleTracker), 10) > Date.now();
         var helpfulnessAskedRecently = parseInt(localStorage.getItem('helpfulnessTracker'), 10) > Date.now();
--- a/kuma/static/js/helpfulness.js
+++ b/kuma/static/js/helpfulness.js
@@ -4,8 +4,10 @@
     var articleTracker = doc.location.pathname + '#answered-helpful';
     // this feature requires localStorage
     if (win.mdn.features.localStorage) {
-        // clear out any drafts from localStorage, now that we are not on the edit page
-        localStorage.removeItem('draft/edit' + location.pathname);
+        // clear out any drafts from localStorage, now that the document has been saved
+        if (document_saved) {
+          localStorage.removeItem('draft/edit' + location.pathname);
+        }
         var ignore = localStorage.getItem('helpful-ignore') === 'true'; // true if ever clicked ignore
         var articleAskedRecently = parseInt(localStorage.getItem(articleTracker), 10) > Date.now();
         var helpfulnessAskedRecently = parseInt(localStorage.getItem('helpfulnessTracker'), 10) > Date.now();
--- a/kuma/static/js/helpfulness.js
+++ b/kuma/static/js/helpfulness.js
@@ -4,10 +4,6 @@
     var articleTracker = doc.location.pathname + '#answered-helpful';
     // this feature requires localStorage
     if (win.mdn.features.localStorage) {
-        // clear out any drafts from localStorage, now that the document has been saved
-        if (document_saved) {
-          localStorage.removeItem('draft/edit' + location.pathname);
-        }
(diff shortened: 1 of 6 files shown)


stephaniehobson:
We need to delete the draft once the content has been successfully published otherwise everyone will have out dated drafts sitting around on every page they have ever published.

stephaniehobson:
Line 509 is only triggered by the user clicking the "Publish and keep editing button" it is not triggered when they arrive on an edit page after publishing.

emullaney:
@stephaniehobson yes, you are correct. I'm taking a look at how to capture publish completion after they click the "publish" button.

stephaniehobson:
I don't think you understand the spirit of what I was hoping to achieve. I will write some requirements once I am back in the office next week so you don't have to decipher my weird ramblings on buzilla.

emullaney:
Well we can talk when you're back next week! Perfect timing for me anyway.

emullaney:
@stephaniehobson let me know when you have a chance to review this. I've made a few improvements that I thought covered the original request but do let me know if there are other requirements!

stephaniehobson:
There's what I hope are clearer requirements on bugzilla now: https://bugzilla.mozilla.org/show_bug.cgi?id=1268609#c8

emullaney:
@stephaniehobson this should be ready for re-review, once travis completes. 

I've added code that sends a variable back to the front end (document_saved=true) when the save is successful. The cached version is deleted if that variable comes back.

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -386,6 +386,16 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
...
 
+  <!-- if this document was saved, we need to clear out the cached draft in helpfulness -->
+  <script type="text/javascript">
+    {% if request.GET.document_saved %}
+      var document_saved = 1;
+    {% else %}
+      var document_saved = 0;
+    {% endif %}
+  </script>
+  {% javascript 'helpfulness' %}
```
stephaniehobson: The helpfulness widget has been removed from the site. Please do not include this library.

kuma/static/js/helpfulness.js
```diff
@@ -4,6 +4,10 @@
     var articleTracker = doc.location.pathname + '#answered-helpful';
     // this feature requires localStorage
     if (win.mdn.features.localStorage) {
+        // clear out any drafts from localStorage, now that the document has been saved
```
stephaniehobson: The helpfulness.js library is not included on the site anymore. 

This code would be better placed in wiki.js which is included on all the article pages by default.

Please put it on it's own lines with it's own check for `localStorage` rather than grouping it with unrelated actions. It can be the last thing in the file before the end of the closure statement.

kuma/static/js/wiki-edit.js
```diff
@@ -461,8 +461,8 @@
         // Save button submits to top-level
         $('.btn-save').on('click', function () {
             if (draftingEnabled) {
+                enableAutoSave(false);
                 // Clear any preserved content.
```
stephaniehobson: Please remove the comment that goes along with the removed code.

stephaniehobson:
Having some trouble testing locally since the page is not updating when I publish the content (this is a common problem with our local dev environments and not related to the PR). I'm going to have to come back and try to test the functionality on Monday :/

kuma/static/js/helpfulness.js
emullaney: I saw that this code was adding helpfulness and thought that was weird because I never added it, but I  must have been working from an older version of master. Thanks for the heads up!

emullaney:
@stephaniehobson this is ready for re-review. Thanks for the detailed feedback!

kuma/static/js/wiki.js
```diff
@@ -941,4 +941,9 @@
         initDetailsTags();
     }
 
+    // clear out any drafts from localStorage, now that the document has been saved
+    if (typeof document_saved !== 'undefined' && document_saved) {
+      localStorage.removeItem('draft/edit' + location.pathname);
```
stephaniehobson: Saving a draft also sets a variable with the save-time in it. We should remove that item as well.

stephaniehobson:
Had to jump through hoops to test it but it's working nicely 🎉 

One more small change, sorry, but ready to go after that.

emullaney:
Okay, please re-review one last time @stephaniehobson. Thanks! This was pretty straight forward. Thanks again for all the notes.

stephaniehobson:
🎉

PR #3924, opened 2016-08-02
Fix Bug 1286541: Add iPerceptions Survey

stephaniehobson:
Adds survey behind a waffle flag and segments user in GA.

Removes helpfulness, partially so they don't clash and partially because we're adding iPerceptions because helpfulness isn't collecting what we need.

😐

### DIFF
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -30,6 +30,11 @@
         ga('set', 'dimension12', {{ analytics_page_revision }});
     {%- endif %}
 
+    // dimension13 == "Saw iPerceptions Survey"
+    {% if waffle.flag('iperceptions') %}
+        ga('set', 'dimension13', 'Yes');
+    {% endif %}
+
     (function() {
         // http://cfsimplicity.com/61/removing-analytics-clutter-from-campaign-urls
         var win = window;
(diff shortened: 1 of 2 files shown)


stephaniehobson:
For staging testing purposes the survey URL will need to be updated in iPerceptions. Myself, Kadir, and Peter can do this.

stephaniehobson:
The code that iPerceptions includes is currently loading an invalid JSON file. It throws this error in Chrome: `VM413:1 Uncaught SyntaxError: Unexpected token o in JSON at position 1` 

Probably it will go away on its own but I'll email Peter to contact iPerceptions.

stephaniehobson:
Needs review from @jwhitlock to make sure I'm not messing up the django. The javascript is just copy and paste so there's no need to review it specifically.

jwhitlock:
That's a lot of JS to set a global variable and inject a script.  I wonder if we have any latitude to change it to a simple script instead of injecting it, or if that makes it easier for ad blockers to block.

I was able to see the pop-up, no JSON error. I thought there was an analytics error, but it's actually for the page revision dimension - GA wants the value to be a string.  Different PR.

I've added the [waffle flag](https://developer.mozilla.org/admin/waffle/flag/119/) in production (accidentally, but kept it), currently off for everyone. We can add it for particular users, or set a percentage to sample it. 

iperceptions

gross

PR #3925, opened 2016-08-02
Add the ability to query Google Analytics for the number of users affected by spam

jbradberry:
https://tree.taiga.io/project/viya-mdn-durable-team/task/353


jbradberry:
@jwhitlock r?

kuma/wiki/utils.py
```diff
@@ -59,3 +67,42 @@ def tidy_content(content):
...
+
+def analytics_user_counts(*revs):
+    """Given some document revision numbers, returns a dict matching those
+    with the number of users Google Analytics thinks has visited each revision.
+
+    """
+
+    scopes = ['https://www.googleapis.com/auth/analytics.readonly']
+    credentials = ServiceAccountCredentials.from_json_keyfile_dict(
+        json.loads(config.GOOGLE_ANALYTICS_CREDENTIALS), scopes=scopes)
```
jwhitlock: There are at least three things that can go wrong with this one line:

1. The user hasn't customized ``GOOGLE_ANALYTICS_CREDENTIALS``, and it deserializes as an empty dictionary
2. The user has customized it, but with invalid JSON, like ``{bad json]``.
3. The user has valid looking credentials that Google doesn't like.

The first case has a traceback that looks like:
```
>>> analytics_user_counts(rev.id)
  File "<console>", line 1, in <module>
  File "/home/vagrant/src/kuma/wiki/utils.py", line 80, in analytics_user_counts
    json.loads(config.GOOGLE_ANALYTICS_CREDENTIALS), scopes=scopes)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/oauth2client/service_account.py", line 254, in from_json_keyfile_dict
    revoke_uri=revoke_uri)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/oauth2client/service_account.py", line 173, in _from_parsed_json_keyfile
    'Expected', client.SERVICE_ACCOUNT)
ValueError: ('Unexpected credentials type', None, 'Expected', 'service_account')
```

The second case has a traceback like this:
```
>>> analytics_user_counts(100)
  File "<console>", line 1, in <module>
  File "/home/vagrant/src/kuma/wiki/utils.py", line 80, in analytics_user_counts
    json.loads(config.GOOGLE_ANALYTICS_CREDENTIALS), scopes=scopes)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
```

For the third case, I created a new private key pair, made sure it worked, then deleted it in the Google console.  Here's the error from that:

```
>>> analytics_user_counts('1096253')
  File "<console>", line 1, in <module>
  File "/home/vagrant/src/kuma/wiki/utils.py", line 105, in analytics_user_counts
    response = request.execute()
  File "/home/vagrant/env/local/lib/python2.7/site-packages/oauth2client/util.py", line 137, in positional_wrapper
  File "/home/vagrant/env/local/lib/python2.7/site-packages/googleapiclient/http.py", line 827, in execute
    method=str(self.method), body=self.body, headers=self.headers)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/googleapiclient/http.py", line 154, in _retry_request
    resp, content = http.request(uri, method, *args, **kwargs)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/oauth2client/transport.py", line 153, in new_request
    credentials._refresh(orig_request_method)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/oauth2client/client.py", line 765, in _refresh
    self._do_refresh_request(http_request)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/oauth2client/client.py", line 834, in _do_refresh_request
    raise HttpAccessTokenRefreshError(error_msg, status=resp.status)
HttpAccessTokenRefreshError: invalid_grant: Invalid JWT Signature.
```

I think the third error is OK at it is.  I'd like to distinguish between the first two.  I think a ``ValueError`` is right for badly written JSON, but I'd like a different, app-specific exception if the user hasn't configured ``GOOGLE_ANALYTICS_CREDENTIALS``.

kuma/wiki/utils.py
```diff
@@ -59,3 +67,42 @@ def tidy_content(content):
...
+                        }
+                    ],
+                    'viewId': '66726481'  # PK of the developer.mozilla.org site on GA.
+                }
+            ]
+        })
+
+    response = request.execute()
+
+    return {int(row['dimensions'][0]): int(row['metrics'][0]['values'][0])
```
jwhitlock: Here's the results for a revision ID with no results:

```
{u'reports': [{u'columnHeader': {u'dimensions': [u'ga:dimension12'],
                                 u'metricHeader': {u'metricHeaderEntries': [{u'name': u'ga:users',
                                                                             u'type': u'INTEGER'}]}},
               u'data': {u'samplesReadCounts': [u'865500'],
                         u'samplingSpaceSizes': [u'2083471'],
                         u'totals': [{u'values': [u'0']}]}}]}
```

This raises the error:

```
  File "<console>", line 1, in <module>
  File "/home/vagrant/src/kuma/wiki/utils.py", line 108, in analytics_user_counts
    for row in response['reports'][0]['data']['rows']}
KeyError: 'rows'
```

versus a result with data:

```
{u'reports': [{u'columnHeader': {u'dimensions': [u'ga:dimension12'],
                                 u'metricHeader': {u'metricHeaderEntries': [{u'name': u'ga:users',
                                                                             u'type': u'INTEGER'}]}},
               u'data': {u'maximums': [{u'values': [u'15062']}],
                         u'minimums': [{u'values': [u'15062']}],
                         u'rowCount': 1,
                         u'rows': [{u'dimensions': [u'1068728'],
                                    u'metrics': [{u'values': [u'15062']}]}],
                         u'samplesReadCounts': [u'864180'],
                         u'samplingSpaceSizes': [u'2085651'],
                         u'totals': [{u'values': [u'15062']}]}}]}
```

which results in the expected ```{1068728: 15062}```

Maybe ``totals`` is a more consistent property?

kuma/wiki/utils.py
```diff
@@ -59,3 +67,42 @@ def tidy_content(content):
...
+    http_auth = credentials.authorize(Http())
+    service = build('analyticsreporting', 'v4', http=http_auth)
+
+    request = service.reports().batchGet(
+        body={
+            'reportRequests': [
+                # `dimension12` is the custom variable containing a page's rev #.
+                {
+                    'dimensions': [{'name': 'ga:dimension12'}],
+                    'metrics': [{'expression': 'ga:users'}],
```
jwhitlock: I'm not sure [ga:users](https://developers.google.com/analytics/devguides/reporting/core/dimsmets#view=detail&group=user&jump=ga_users) is the dimension we want. I'm not sure how it is calculated, but will see what the docs say.  [ga:uniquePageviews](https://developers.google.com/analytics/devguides/reporting/core/dimsmets#view=detail&group=page_tracking&jump=ga_uniquepageviews) is the dimension they show in the site content views in GA.

kuma/wiki/tests/test_utils.py
```diff
@@ -0,0 +1,118 @@
...
+import mock
+from googleapiclient.http import HttpMockSequence
+from googleapiclient.errors import HttpError
+from kuma.core.tests import KumaTestCase
+
+from ..utils import analytics_user_counts
+
+
+class AnalyticsUserCountsTests(KumaTestCase):
+    def setUp(self):
```
jwhitlock: This might be a candidate for the [setUpClass](https://docs.python.org/2/library/unittest.html#setupclass-and-teardownclass) class method. This is also the first time I've ever said that.

jwhitlock:
Good work on the tests. This is a good foundation for the crazy results the GA API will throw our way in production.

Take a look at the verbose diff comments.  I think the actions are:

* Raise a custom exception if ``GOOGLE_ANALYTICS_CREDENTIALS`` is still the default
* Handle the cases where there are no results for a revision ID
* Switch to ``ga:uniquePageviews``, or find some docs why ``ga:users`` should be used instead
* Switch to ``setUpClass`` (maybe)

requirements/constraints.txt
```diff
@@ -320,3 +320,47 @@ argparse==1.4.0 \
     --hash=sha256:62b089a55be1d8949cd2bc7e0df0bddb9e028faefc8c32038cc84862aefdd6e4
 diff-match-patch==20121119 \
     --hash=sha256:9dba5611fbf27893347349fd51cc1911cb403682a7163373adacc565d11e2e4c
+
+# google-api-python-client
```
jwhitlock: Thank you for using ``constraints.txt`` and following the style guide.

kuma/wiki/utils.py
jbradberry: I'm inclined, now that you mention it, to make both case 1 and case 2 raise an ImproperlyConfigured error with different messages.

jbradberry:
@jwhitlock Ok, this should address your review comments.

jwhitlock:
The new code looks good, and the practical tests are going well. There appears to be an implicit filter in the API call, which is limiting the number of page views returned.  I'm digging into the docs and seeing if I can find the filter.

[Revision 1098511](https://developer.mozilla.org/en-US/docs/Web/SVG/Element$revision/1098511) is the n-1 revision of https://developer.mozilla.org/en-US/docs/Web/SVG/Element, active from July 29th to August 9th.  If I load the Google Analytics website and view July 1st to August 11th, this revision has 2,393 unique page views. If I call ``analytics_user_counts(1098511)``, I get ``{1098511: 1302}``.  

Here's the full response:

```
{u'reports': [{u'columnHeader': {u'dimensions': [u'ga:dimension12'],
                                 u'metricHeader': {u'metricHeaderEntries': [{u'name': u'ga:uniquePageviews',
                                                                             u'type': u'INTEGER'}]}},
               u'data': {u'maximums': [{u'values': [u'1302']}],
                         u'minimums': [{u'values': [u'1302']}],
                         u'rowCount': 1,
                         u'rows': [{u'dimensions': [u'1098511'],
                                    u'metrics': [{u'values': [u'1302']}]}],
                         u'samplesReadCounts': [u'999997'],
                         u'samplingSpaceSizes': [u'2076702'],
                         u'totals': [{u'values': [u'1302']}]}}]}
```

jwhitlock:
Yes, there's a [default date range](https://developers.google.com/analytics/devguides/reporting/core/v4/basics):

> If a date range is not provided, the default date range is: ``{"startDate": "7daysAgo", "endDate": "yesterday"}``.

Since the purpose is to determine how many viewers there were over the lifetime of the revision, we'll need to expand this range to include the period when all the given revisions were active.  I think the range should be a ``startDate`` of the earliest revision creation date minus a day, and a end date of today (or whatever the last valid date is).

jbradberry:
@jwhitlock back to you.

Note that based on some things mentioned in https://developers.google.com/resources/api-libraries/documentation/analyticsreporting/v4/python/latest/analyticsreporting_v4.reports.html#batchGet, I believe the date ranges to be inclusive of their endpoints.

jwhitlock:
This is now a bit harder to test.  For testing and local development, I'd like to break this into two functions, and change the function name:

* ```analytics_upageviews(revision_ids, start_date, end_date=None)``` - takes an iterable of revision IDs, a start_date, and an optional end_date (defaults to today) 
* ```analytics_upageviews_by_revisions(revisions)``` - takes an iterable of revisions (queryset, list of revisions, etc.). Calls ``analytics_pageviews`` with the computed values from live data.

But, it is still possible, if you have the anonymized database and are willing to lie to MySQL:

```
me = User.objects.get(username='jwhitlock')
doc = Document.objects.get(id=3190)
rev = Revision.objects.create(id=1098511, document=doc, creator=me, created=datetime.datetime(2016, 7, 29, 16, 22, 29))
analytics_user_counts([rev])
```

This sends the query:
```
{'reportRequests': [{
    'dateRanges': [
        {'endDate': '2016-08-11',
         'startDate': '2016-07-29T16:22:29'}],
    'dimensionFilterClauses': [{
        'filters': [{
            'dimensionName': 'ga:dimension12',
            'expressions': [1098511L],
            'operator': 'IN_LIST'}]
    }],
    'dimensions': [{'name': 'ga:dimension12'}],
    'metrics': [{'expression': 'ga:uniquePageviews'}],
    'viewId': '66726481'}]
}
```

It helped to move construction of the body outside of ``batchGet`` to see what was being sent.

Google Analytics doesn't like:

* ``1098511L`` versus ``1098511`` or even ``"1098511"``
* ``2016-07-29T16:22:29`` versus ``2016-07-29``

I tried with a few different starting dates:

* Google Analytics, July 29th to August 9th, from above: 2393
* 2016-07-29: 2327
* 2016-07-28: 2298
* 2016-07-27: 2333
* 2016-07-01: 2399
* 2016-01-01: 2778

Because sampling is used, the numbers are not exact, but they are also roughly the same, so using the creation date versus creation - 1 day seems justified.  However, the numbers seem to get further away with a larger date range, making me slightly less comfortable using today as the end date.  I'll get over it.

kuma/wiki/utils.py
```diff
@@ -59,3 +69,80 @@ def tidy_content(content):
...
+    if not ga_cred_dict:
+        raise ImproperlyConfigured(
+            "An empty GOOGLE_ANALYTICS_CREDENTIALS Constance setting is not permitted.")
+
+    credentials = ServiceAccountCredentials.from_json_keyfile_dict(ga_cred_dict,
+                                                                   scopes=scopes)
+    http_auth = credentials.authorize(Http())
+    service = build('analyticsreporting', 'v4', http=http_auth)
+
+    end_date = datetime.date.today().isoformat()
```
jwhitlock: This should only set the ``end_date`` if it is ``None``.

kuma/wiki/tests/test_utils.py
```diff
@@ -0,0 +1,290 @@
...
+    def setUpClass(cls):
+        super(AnalyticsUpageviewsTests, cls).setUpClass()
+
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        with open(os.path.join(dir_path, 'analyticsreporting-discover.json')) as f:
+            cls.valid_discovery = f.read()
+
+        cls.start_date = datetime.date(2016, 1, 1)
+
+        cls.valid_response = """{"reports": [
```
jwhitlock: These are static values, and don't need to be in ``setUpClass``. They could just be declared as class-level variables.

kuma/wiki/tests/test_utils.py
```diff
@@ -0,0 +1,290 @@
...
+                    "metricHeader": {
+                        "metricHeaderEntries": [
+                            {"type": "INTEGER", "name": "ga:uniquePageviews"}
+                        ]
+                    }
+                }
+            }
+        ]}"""
+
+    @override_config(GOOGLE_ANALYTICS_CREDENTIALS=GA_TEST_CREDS)
```
jwhitlock: It seems these decorators are used on all test methods, so I think they can be moved up to class-level decorators

kuma/wiki/utils.py
```diff
@@ -59,3 +69,80 @@ def tidy_content(content):
...
+    if hasattr(start_date, 'date'):
+        start_date = start_date.date()
```
jwhitlock: 👍 on auto-fixing ``datetime`` to ``date``. The same should be used on ``end_date``.

jwhitlock:
Manual testing is a lot easier with the two functions, thanks.  Using my fake revision from before:

```
from kuma.wiki.utils import analytics_upageviews, analytics_upageviews_by_revisions
rev_id = 1098511
analytics_upageviews_by_revisions(Revision.objects.filter(id=rev_id))
# returns {1098511: 2281}
from datetime import datetime, date
ranges = [
    (date(2016,7,29), None),
    (date(2016,7,28), None),
    (date(2016,7,27), None),
    (date(2016,7,1),None),
    (date(2016,1,1),None),
    (date(2016,7,29), date(2016,7,29)),
    (datetime(2016,7,29,8), datetime(2016,7,29,9)),
]
for start, end in ranges:
    count = analytics_upageviews([rev_id], start, end)[rev_id]
    print("%s to %s: %d" % (start, end, count))

# 2016-07-29 to None: 2299
# 2016-07-28 to None: 2282
# 2016-07-27 to None: 2304
# 2016-07-01 to None: 2426
# 2016-01-01 to None: 2813
# 2016-07-29 to 2016-07-29: 2299
# 2016-07-29 08:00:00 to 2016-07-29 09:00:00: 2299
```

The ``end_date`` = always today is a bug, and there's some possible cleanup in the tests, but this is really close.

jwhitlock:
👍 manual tests pass, coverage is good, ready to merge. Feel free to make further adjustments as needed for the spam dashboard tasks.

PR #3926, opened 2016-08-03
[Bug 859499] Show localized locale name in language selector

safwanrahman:
This patch fulfill all the specification mentioned in the [Bug 859499](https://bugzilla.mozilla.org/show_bug.cgi?id=859499#c11)

- [x] Add ISO 639 codes to the front page selector
- [x] Implement Wikipedia-style hover text (display the language in the user's locale) in the language dropdown on wiki pages
- [x] Implement Wikipedia-style hover text in the language selector on the front page.


safwanrahman:
Need both front end and back end review.
@jwhitlock @stephaniehobson r?

scripts/translate_locales.py
```diff
@@ -0,0 +1,33 @@
...
+Background: https://bugzil.la/859499#c11
+
+{% for lang in languages %}
+   {{ '{{ _("' + lang + '") }}' }}
+{% endfor %}
+
+""")
+
+
+LANGUAGES = sorted([lang_info.english.encode('ascii', 'ignore')
```
jwhitlock: 👍 on sorting the languages

Makefile
```diff
@@ -44,6 +44,8 @@ clean:
 	mkdir -p build/locale
 
 locale:
+	# For generating a new file to let locales name localizable
+	python scripts/translate_locales.py
```
jwhitlock: I think this belongs in ``localeextract``, which is run periodically to generate new strings, and not ``locale``, which is used to add a new locale.

In either case, it fails in my VM with this traceback:

```
# For generating a new file to let locales name localizable
python scripts/translate_locales.py
  File "scripts/translate_locales.py", line 23, in <module>
    for lang_code, lang_info in settings.LOCALES.items()
  File "/home/vagrant/env/local/lib/python2.7/site-packages/django/conf/__init__.py", line 48, in __getattr__
    self._setup(name)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/django/conf/__init__.py", line 44, in _setup
    self._wrapped = Settings(settings_module)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/django/conf/__init__.py", line 92, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named kuma.settings.testing
make: *** [localeextract] Error 1
```

This suggests it may be better as a Django management command, like ``manage.py extract`` and ``manage.py merge``, and automatically get the settings that way.

scripts/translate_locales.py
jwhitlock: I think ``english.encode('ascii', 'ignore')`` is the wrong thing - it will throw away non-ASCII characters.  I'm more comfortable with UTF-8, like ``english.encode('utf8')``

Makefile
safwanrahman: This script should run every-time a new locale is added. so while adding a new locale and committing, the file will be regenerated.

Yap. seems like this should be a management command

scripts/translate_locales.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+template_string = ("""
+
+This template is automatically generated by scripts/translate_locales.py to make the languages name localizable. Do not edit it manually
+Background: https://bugzil.la/859499#c11
+
+{% for lang in languages %}
+   {{ '{{ _("' + lang + '") }}' }}
+{% endfor %}
```
jwhitlock: This template has too much whitespace between the locale names. I think changing to ``{%- endfor %}`` (adding a minus sign at the beginning) will get rid of the extra new line.

scripts/translate_locales.py
safwanrahman: The language name should always be in English as well as ASCII character. Does not it?

scripts/translate_locales.py
```diff
@@ -0,0 +1,33 @@
...
+
+""")
+
+
+LANGUAGES = sorted([lang_info.english.encode('ascii', 'ignore')
+                    for lang_code, lang_info in settings.LOCALES.items()
+                    if lang_code in settings.MDN_LANGUAGES])
+
+string = Environment().from_string(template_string).render(languages=LANGUAGES)
```
jwhitlock: Jinja is probably overkill for this usage - we don't need the HTML escaping, template caching, etc. Plain Python string construction should work.  Something like:

```
    string = (u'''\
This template is automatically generated by scripts/translate_locales.py to make the languages name localizable. Do not edit it manually
Background: https://bugzil.la/859499#c11

  %s
''') % '\n  '.join(LANGUAGES)
```

This assumes that ``LANGUAGES`` contains a list of unicode strings, rather than ``encode`` to switch to bytes.  Then, when you write to the file, you can encode the whole thing in utf-8.

jwhitlock:
The template code worked for me (still needs a review from @stephaniehobson ), but I was unable to regenerate the languages file using ``make locale``. I think it may need to be converted to a Django management command to read the settings correctly, in which case it can either continue writing to a file or just output the languages, and be redirected to a file.

See the line comments for other suggestions.

jinja2/includes/translate_locales.html
```diff
@@ -0,0 +1,62 @@
+This template is automatically generated byscripts/translate_locales.py to make the languages namelocalizable. Do not edit it manually
```
jwhitlock: Needs space in "namelocalizable"

kuma/core/management/commands/translate_locales_name.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# A script to generate a template which will be used to localize the supported locale name
+# For more inofrmation see https://bugzil.la/859499#c11
+
+class Command(BaseCommand):
+    help = "Generate a template to get the locales name localized"
+
+    def handle(self, *args, **options):
+        template_string = (u"This template is automatically generated by"
```
jwhitlock: This is where the strings with missing spaces are generated

kuma/core/management/commands/translate_locales_name.py
```diff
@@ -0,0 +1,30 @@
...
+                           "languages name localizable."
+                           "Do not edit it manually\n"
+                           "Background: https://bugzil.la/859499#c11\n")
+
+        LANGUAGES = sorted([lang_info.english
+                            for lang_code, lang_info in settings.LOCALES.items()
+                            if lang_code in settings.MDN_LANGUAGES])
+
+        for lang in LANGUAGES:
+            template_string += "{{ _('%s') }}\n" % lang
```
jwhitlock: The template string should also be marked as unicode, like ``u"{{ _...``

kuma/core/management/commands/translate_locales_name.py
```diff
@@ -0,0 +1,30 @@
...
+
+        jinja_path = settings.TEMPLATES[0]["DIRS"][0]
+        template_path = jinja_path + '/includes/translate_locales.html'
+
+        outfile = open(template_path, "w")
+        outfile.write(template_string)
```
jwhitlock: This should be UTF-8 encoded, like ``template_string.encode('utf8')``

jwhitlock:
@safwanrahman This is looking good. I'm much happier with it as a management command. There's some tweaks that should be easy to implement noted against the code.

I'm still not 100% sure that it should be run only when creating a new locale, but I'm having a hard time coming up with a situation where this would change outside of adding a new locale. So, we'll leave it there until an unexpected situation comes up.

I don't think a unit test makes sense, so we can remove that from the TODOs.

safwanrahman:
@jwhitlock Updated. r?

jwhitlock:
Thank you, the backend code looks good to me, and runs as expected on the page.  @stephaniehobson, does the markup look good to you?

jinja2/includes/translate_locales.html
```diff
@@ -0,0 +1,62 @@
+This template is automatically generated by scripts/translate_locales.py to make the languages name localizable. Do not edit it manually
```
stephaniehobson: I don't see a `scripts/translate_locales.py` file in this PR, is that the right file?

stephaniehobson:
The changes in this PR look well crafted but I think the original bug is not the right way to go.

R+ on this after my question is answered but I would like to continue to make changes to the language selector. @safwanrahman are you interested in working on this a little more? Can I cc you on the other changes I would like?

jinja2/includes/translate_locales.html
safwanrahman: Oh sorry. The script was replaced by ``translate_locales_name`` management command.
I will update the line soon.

safwanrahman:
> @safwanrahman are you interested in working on this a little more? Can I cc you on the other changes I would like?

Sure, Please CC me to the bugs. It will be cool to work on the features.

safwanrahman:
@stephaniehobson Answered and Updated. This can be merged now

PR #3927, opened 2016-08-03
Bug 1243693 : HTML Tags are wrongly displayed in title's bar

MatonAnthony:
Bug 1243693 : HTML Tags are wrongly displayed in title's bar

For this bug, I didn't managed to find a "real" fix.
- I think the first escape happens when the Document is rendered and a second time when `page_title(title)` is called.
- To work around this bug, I unescaped the string using jinja2.Markup unescape() function which removes the escape.

To ensure it worked properly, I tried to trigger a `<script>` from the title and it wasn't triggered.

### DIFF
--- a/kuma/core/templatetags/jinja_helpers.py
+++ b/kuma/core/templatetags/jinja_helpers.py
@@ -96,7 +96,7 @@ def entity_decode(str):
 
 @library.global_function
 def page_title(title):
-    return u'%s | MDN' % title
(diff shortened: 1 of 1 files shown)


jwhitlock:
Yep, this solves it. The ``unescape()`` is unneeded though, please remove so it's just ``return jinja2.Markup(u'%s | MDN' % title)``, like ``yesno`` and other helpers in this file.  That will signal to Jinja that the string is already escaped, and it won't double-escape it anymore.

PR #3928, opened 2016-08-03
Bug 1138173 : Escape tags in raw en-US version of an article while editing a local

MatonAnthony:
[Bug 1138173](https://bugzilla.mozilla.org/show_bug.cgi?id=1138173): Escape tags in raw en-US version of an article while editing a local

I remove the |safe because the goal here is to have this string escaped to be able to copy/paste them inside when updating or making a translation.


jwhitlock:
Thanks, works on the example given in the bug.

PR #3929, opened 2016-08-03
bug 1287644: Add quotes around GA dim12, revision

jwhitlock:
GA debug was complaining that the value for custom dimension 12 was expected to be a string, but instead was an integer. This fixes it.

### DIFF
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -27,7 +27,7 @@
 
     {%- if analytics_page_revision %}
         // dimension12 == 'Page Revision"
-        ga('set', 'dimension12', {{ analytics_page_revision }});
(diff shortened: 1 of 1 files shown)


jinja2/includes/google_analytics.html
```diff
@@ -27,7 +27,7 @@
 
     {%- if analytics_page_revision %}
         // dimension12 == 'Page Revision"
-        ga('set', 'dimension12', {{ analytics_page_revision }});
+        ga('set', 'dimension12', '{{ analytics_page_revision }}');
```
safwanrahman: I think better written as ``{{ str(analytics_page_revision) }}``.

safwanrahman:
Sorry I thought its python code. seems like we generate javascript with templating language! :crying_cat_face:

stephaniehobson:
👍

PR #3932, opened 2016-08-05
[Bug 1291969] CKEditor build script should use release versions of plugins

safwanrahman:
@jwhitlock r?


jwhitlock:
+r, works in dev environment, some small changes in the ``scayt`` and ``wsc`` plugins that will be pulled in with the next CKEditor upgrade.

PR #3935, opened 2016-08-10
bug 1260197 - Update "publish and keep editing" workflow

dchukhin:
Pull request begun by @stephaniehobson, which I continued.
https://bugzilla.mozilla.org/show_bug.cgi?id=1260197
When editing a document and clicking "Publish and Keep Editing" (ajax request) the user is no longer always told that the changes were saved successfully. Instead, upon an unsuccessful edit (spam edit, session timeout, midair collision, or something strange), an error message is shown to the user. Back-end errors (spam, midair collision) are returned as JSON and handled by front-end code. Feedback welcome from @jwhitlock and @stephaniehobson or others.

### DIFF
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -484,7 +484,7 @@
             // record event
             mdn.analytics.trackEvent({
                     category: 'Wiki',
-                    action: 'Button',
+                    action: 'Button', // now "Publish and keep editing" but keeping label for analytics continuity
                     label: 'Save and Keep Editing'
                 });
 
@@ -512,14 +512,12 @@
                     var $responseLoginRequired = $parsedData.find('#login');
                     // If there are errors, saveNotification() for each of them
                     if ($responseErrors.length) {
-                        var liErrors = $responseErrors.find('li')
-                        for (var i = 0; i < liErrors.length; i++) {
-                            saveNotification.error(liErrors[i])
-                        }
+                        var $liErrors = $responseErrors.find('li');
+                        saveNotification.error($liErrors);
                         //$form.submit();
                     // Check if the session has timed out
                     } else if ($responseLoginRequired.length) {
-                        saveNotification.error('Publishing failed. You are not currently signed in. Please use a new tab to sign in and try publishing again.')
+                        saveNotification.error(gettext('Publishing failed. You are not currently signed in. Please use a new tab to sign in and try publishing again.'));
                     } else {
                         // assume it went well
                         saveNotification.success(gettext('Changes saved.'), 2000);
@@ -527,12 +525,12 @@
                         // We also need to update the form's current_rev to
                         // avoid triggering a conflict, since we just saved in
                         // the background.
-                        var responseData = JSON.parse(data)
-                        if (responseData['error'] == true) {
-                            saveNotification.error(responseData['error_message'])
+                        var responseData = JSON.parse(data);
+                        if (responseData['error'] === true) {
+                            saveNotification.error(responseData['error_message']);
                         } else {
                             var responseRevision = JSON.parse(data)['new_revision_id'];
-                            $("input[id=id_current_rev]").val(responseRevision)
+                            $("input[id=id_current_rev]").val(responseRevision);
 
                             // Clear the review comment
                             $('#id_comment').val('');
@@ -548,14 +546,13 @@
                 error: function(jqXHR, textStatus, errorThrown) {
                     // Try to display the error that comes back from the server
                     try {
-                        var errorMessage = JSON.parse(jqXHR['responseText'])['error_message']
+                        var errorMessage = JSON.parse(jqXHR['responseText'])['error_message'];
                     } catch (err) {
-                        errorMessage = "Publishing failed. Please copy and paste your changes into a safe place and try submitting the form using the 'Publish' button.";
+                        errorMessage = gettext("Publishing failed. Please copy and paste your changes into a safe place and try submitting the form using the 'Publish' button.");
                     }
                     saveNotification.error(errorMessage);
                     // Re-enable the form; it gets disabled to prevent double-POSTs
                     $form.attr('disabled', false);
-                    // save draft
                 }
             });
 /*
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -478,8 +478,11 @@
             // disable form
             $('#wiki-page-edit').attr('disabled', true);
 
+            // Clear previous notification messages, if any
+            $('.notification button.close').click();
+
             // give user feedback
-            var saveNotification = mdn.Notifier.growl('Saving changes…', { duration: 0 });
+            var saveNotification = mdn.Notifier.growl('Saving changes…', { duration: 0 , closable: true});
 
             // record event
             mdn.analytics.trackEvent({
@@ -510,11 +513,18 @@
                     var $parsedData = $($.parseHTML(data));
                     var $responseErrors = $parsedData.find('.errorlist');
                     var $responseLoginRequired = $parsedData.find('#login');
-                    // If there are errors, saveNotification() for each of them
+                    try {
+                      var jsonErrorMessage = JSON.parse(jqXHR['responseText'])['error_message'];
+                    } catch (err) {
+                      var jsonErrorMessage = undefined;
+                    }
+                    // If there are errors, saveNotification() for them
                     if ($responseErrors.length) {
                         var $liErrors = $responseErrors.find('li');
                         saveNotification.error($liErrors);
-                        //$form.submit();
+                    // If there was an error from the json response, saveNotification() for it
+                    } else if (typeof(jsonErrorMessage) !== 'undefined') {
+                        saveNotification.error(gettext(jsonErrorMessage));
                     // Check if the session has timed out
                     } else if ($responseLoginRequired.length) {
                         saveNotification.error(gettext('Publishing failed. You are not currently signed in. Please use a new tab to sign in and try publishing again.'));
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -459,6 +459,10 @@
     //
     function initSaveAndEditButtons () {
         var $form = $('#wiki-page-edit');
+        // Handle edits on the translate page
+        if ($form.length == 0 && $('#translate-document').length == 1) {
+            $form = $('#translate-document');
+        }
 
         // Save button submits to top-level
         $form.on('submit', function () {
@@ -471,7 +475,7 @@
 
         // save and edit attempts ajax submit
         $('.btn-save-and-edit').on('click', function(event) {
-            // TODO: diable form
+            // TODO: disable form
             console.log('save and edit triggered');
 
             // give user feedback
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -496,7 +496,6 @@
             var formURL = window.location.href; // submits to self
             console.log('posting this data: ', formData)
 
-            // ajax submit
             $.ajax({
                 url : formURL + '?async',
                 type: "POST",
@@ -528,16 +527,22 @@
                         // We also need to update the form's current_rev to
                         // avoid triggering a conflict, since we just saved in
                         // the background.
-                        var responseRevision = JSON.parse(data)['new_revision_id'];
-                        console.log('responseRevision: ', responseRevision);
-                        // TODO: why are there multiple <input id="id_current_rev">s?
-                        $("input[id=id_current_rev]").val(responseRevision)
+                        console.log('data is: ', data)
+                        var responseData = JSON.parse(data)
+                        if (responseData['error'] == true) {
+                            saveNotification.error(responseData['error_message'])
+                        } else {
+                            var responseRevision = JSON.parse(data)['new_revision_id'];
+                            console.log('responseRevision: ', responseRevision);
+                            // TODO: why are there multiple <input id="id_current_rev">s?
+                            $("input[id=id_current_rev]").val(responseRevision)
 
-                        // Clear the review comment
-                        $('#id_comment').val('');
+                            // Clear the review comment
+                            $('#id_comment').val('');
 
-                        // Trigger a `mdn:save-success` event so dirtiness can be reset throughout the page
-                        $form.trigger('mdn:save-success');
+                            // Trigger a `mdn:save-success` event so dirtiness can be reset throughout the page
+                            $form.trigger('mdn:save-success');
+                        }
 
                         // Re-enable the form; it gets disabled to prevent double-POSTs
                         $form.data('disabled', false).removeClass('disabled');
@@ -549,9 +554,13 @@
                     // console.log(jqXHR.status);
                     // console.log(textStatus);
                     // console.log(errorThrown);
-                    var msg = "Publishing failed. Please copy and paste your changes into a safe place and try submitting the form using the 'Publish' button.";
-
-                    saveNotification.error(msg);
+                    // Try to display the error that comes back from the server
+                    try {
+                        var errorMessage = JSON.parse(jqXHR['responseText'])['error_message']
+                    } catch (err) {
+                        errorMessage = "Publishing failed. Please copy and paste your changes into a safe place and try submitting the form using the 'Publish' button.";
+                    }
+                    saveNotification.error(errorMessage);
 //                    saveNotification.error(msg, {closable: true, duration: 0});
 
                     // save draft
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -458,89 +458,113 @@
     // Initialize logic for save and save-and-edit buttons.
     //
     function initSaveAndEditButtons () {
+        var $form = $('#wiki-page-edit');
+
         // Save button submits to top-level
-        $('.btn-save').on('click', function () {
-            if (draftingEnabled) {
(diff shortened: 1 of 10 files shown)


kuma/wiki/views/translate.py
```diff
@@ -214,7 +217,24 @@ def translate(request, document_slug, document_locale, revision_id=None):
...
                 return redirect(doc)
+            else:
+                # If this is an Ajax POST, then return a JsonResponse with error
+                if request.is_ajax():
+                    data = {
+                        "error": True,
+                        "error_message": mark_safe(rev_form.errors['current_rev'][0]),
+                        "new_revision_id": rev_form.instance.id,
+                    }
+                    return JsonResponse(data=data, status=500)
```
safwanrahman: Error 500?
I think it should be something else. It will be hard to debug if its 500

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                    // If there are errors, saveNotification() for each of them
+                    if ($responseErrors.length) {
+                        var liErrors = $responseErrors.find('li')
+                        for (var i = 0; i < liErrors.length; i++) {
+                            saveNotification.error(liErrors[i])
+                        }
+                        //$form.submit();
+                    // Check if the session has timed out
+                    } else if ($responseLoginRequired.length) {
+                        saveNotification.error('Publishing failed. You are not currently signed in. Please use a new tab to sign in and try publishing again.')
```
stephaniehobson: Needs the gettext wrapper like line 526

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
                     }
+                    // Re-enable the form; it gets disabled to prevent double-POSTs
+                    $form.attr('disabled', false);
+                },
+                error: function(jqXHR, textStatus, errorThrown) {
+                    // Try to display the error that comes back from the server
+                    try {
+                        var errorMessage = JSON.parse(jqXHR['responseText'])['error_message']
+                    } catch (err) {
+                        errorMessage = "Publishing failed. Please copy and paste your changes into a safe place and try submitting the form using the 'Publish' button.";
```
stephaniehobson: Needs gettext wrapper.

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                    }
+                    saveNotification.error(errorMessage);
+                    // Re-enable the form; it gets disabled to prevent double-POSTs
+                    $form.attr('disabled', false);
+                    // save draft
```
stephaniehobson: Comment here about saving draft but draft is not saved here.

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
 
+            // record event
             mdn.analytics.trackEvent({
-                category: 'Wiki',
-                action: 'Button',
-                label: 'Save and Keep Editing' // now "Publish and keep editing" but keeping label for analytics continuity
-            });
+                    category: 'Wiki',
+                    action: 'Button',
+                    label: 'Save and Keep Editing'
```
stephaniehobson: Please preserve the comment `// now "Publish and keep editing" but keeping label for analytics continuity`

kuma/wiki/jinja2/wiki/includes/page_buttons.html
```diff
@@ -19,8 +19,9 @@
 
 <ul class="page-buttons">
   {% if document %}
+  {# No save-and-edit on new pages because slug changes confuse it #}
   <li>
-    <button type="submit" class="button positive btn-save-and-edit" data-optimizely-hook="button-save-and-keep-editing button-save-and-keep-editing-{{ location }}"><span>{{ _('Publish and Keep Editing') }}</span><i aria-hidden="true" class="icon-pencil"></i></button>
+    <button type="button" class="button positive btn-save-and-edit" data-optimizely-hook="button-save-and-keep-editing button-save-and-keep-editing-{{ location }}"><span>{{ _('Publish and Keep Editing') }}</span><i aria-hidden="true" class="icon-edit"></i></button>
```
stephaniehobson: I should have left this as icon-pencil :/

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+        }
+
         // Save button submits to top-level
-        $('.btn-save').on('click', function () {
-            if (draftingEnabled) {
+        $form.on('submit', function () {
+            if (supportsLocalStorage) {
                 // Clear any preserved content.
-                clearDraft('publish');
+                clearDraft();
```
stephaniehobson: Please keep ` clearDraft('publish');` 

(The clearDraft() functionality was augmented after the branch you forked this from).

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                dataType : 'html',
+                success: function(data, textStatus, jqXHR) {
+                    // server came back 200
+                    // was there an error, or did the session expire?
+                    var $parsedData = $($.parseHTML(data));
+                    var $responseErrors = $parsedData.find('.errorlist');
+                    var $responseLoginRequired = $parsedData.find('#login');
+                    // If there are errors, saveNotification() for each of them
+                    if ($responseErrors.length) {
+                        var liErrors = $responseErrors.find('li')
```
stephaniehobson: We like to use a `$` in front of variables that contain jQuery objects. Please change liErrors to $liErrors.

stephaniehobson: And end the line with a semi-colon ;)

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                        for (var i = 0; i < liErrors.length; i++) {
+                            saveNotification.error(liErrors[i])
```
stephaniehobson: Missing semi-colon.

kuma/static/js/wiki-edit.js
stephaniehobson: Also needs a semi-colon.

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                    } else if ($responseLoginRequired.length) {
+                        saveNotification.error('Publishing failed. You are not currently signed in. Please use a new tab to sign in and try publishing again.')
+                    } else {
+                        // assume it went well
+                        saveNotification.success(gettext('Changes saved.'), 2000);
+
+                        // We also need to update the form's current_rev to
+                        // avoid triggering a conflict, since we just saved in
+                        // the background.
+                        var responseData = JSON.parse(data)
```
stephaniehobson: Semi colon please.

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                        if (responseData['error'] == true) {
```
stephaniehobson: We program javascript as strict so please use `===` for comparisons.

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                            saveNotification.error(responseData['error_message'])
```
stephaniehobson: i can haz semicolon?

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
+                        } else {
+                            var responseRevision = JSON.parse(data)['new_revision_id'];
+                            $("input[id=id_current_rev]").val(responseRevision)
```
stephaniehobson: ;?

kuma/static/js/wiki-edit.js
```diff
@@ -458,90 +458,120 @@
...
-                        clearDraft('publish');
-                        enableAutoSave(true);
                     }
+                    // Re-enable the form; it gets disabled to prevent double-POSTs
+                    $form.attr('disabled', false);
+                },
+                error: function(jqXHR, textStatus, errorThrown) {
+                    // Try to display the error that comes back from the server
+                    try {
+                        var errorMessage = JSON.parse(jqXHR['responseText'])['error_message']
```
stephaniehobson: ;

kuma/wiki/views/translate.py
stephaniehobson: The convention is to use a 200 status code if the JSON is returned without a problem,  even if what the JSON says is `{"error": True}`. Non 200 status codes are reserved for when something goes wrong producing the JSON.

stephaniehobson:
It works! It's alive! I've never been so happy to get bunches of error messages.

A bit of UX tweaking:
- [ ] error messages need to be dismiss-able
- [ ] subsequent publish attempts should clear the errors from previous ones
- [ ] spam message should not disappear on its own

The notifier script is in components.js, you should be able to accomplish these without editing it but that file is the only documentation we have for the component.

kuma/static/js/wiki-edit.js
stephaniehobson: Also, I think using the function in this way will cause the first error message to be over-written by the next error message. It's totally fine to list them all together in one notifier.

kuma/wiki/views/translate.py
```diff
@@ -200,6 +202,7 @@ def translate(request, document_slug, document_locale, revision_id=None):
                                     parent_slug=slug_dict['parent'])
             rev_form.instance.document = doc  # for rev_form.clean()
 
+#            import ipdb;ipdb.set_trace()
```
jwhitlock: Remove this line

kuma/wiki/tests/test_views.py
```diff
@@ -2281,48 +2282,180 @@ def test_quick_review(self):
             eq_(data_dict['expected_tags'], review_tags)
 
     @pytest.mark.midair
-    def test_edit_midair_collision(self):
+    def test_edit_midair_collisions(self):
+
+        def midair_collision_test(is_ajax):
```
jwhitlock: The way we've done this for other cases is to define a function that takes a optional parameter, like:

```
def test_edit_midair_collision(self, is_ajax=False):
    # Code that works both ways
```

and then, have a second test that calls it with the alternate option:

```
def test_edit_midair_collision_ajax(self):
    self.test_edit_midair_collision(is_ajax=True)
```

Please replace the inner function with this alternate pattern.

kuma/wiki/tests/test_views.py
```diff
@@ -2281,48 +2282,180 @@ def test_quick_review(self):
...
+        midair_collision_test(is_ajax=False)
+        midair_collision_test(is_ajax=True)
+
+    @override_config(AKISMET_KEY='dashboard')
+    @requests_mock.mock()
+    @mock.patch('kuma.spam.akismet.Akismet.check_comment')
+    def test_edit_spam_ajax(self, mock_requests, mock_akismet_method):
+        """Tests attempted spam edits that occur on Ajax POSTs."""
+        # Enable Akismet
+        Flag.objects.create(name=SPAM_SUBMISSIONS_FLAG, everyone=True)
```
jwhitlock: Use the [@override_flag decorator](http://waffle.readthedocs.io/en/v0.11/testing/automated.html#unit-tests) instead of creating flags in tests.

kuma/wiki/views/edit.py
```diff
@@ -168,16 +168,15 @@ def edit(request, document_slug, document_locale, revision_id=None):
...
-                        response = HttpResponse(textwrap.dedent("""
-                            <span id="iframe-response"
-                                  data-status="OK"
-                                  data-current-revision="%s">OK</span>
-                        """ % doc.current_revision.id))
-                        response['X-Frame-Options'] = 'SAMEORIGIN'
-                        return response
+                    # save-and-edit button
+                    if is_async_submit:
+                        # This is the most recent revision id
```
jwhitlock: This branch (valid AJAX edit to the document) is uncovered by tests. I'm also not sure if it is possible to get here (is AJAX allowed for doc-only saves?)

kuma/wiki/views/translate.py
```diff
@@ -214,7 +217,24 @@ def translate(request, document_slug, document_locale, revision_id=None):
                         pass
 
                 rev_form.save(doc)
+                # If this is an Ajax POST, then return a JsonResponse
+                if request.is_ajax():
```
jwhitlock: This branch (successful AJAX edit of revision or combined translation form) is not covered by a test.

kuma/wiki/views/translate.py
```diff
@@ -214,7 +217,24 @@ def translate(request, document_slug, document_locale, revision_id=None):
...
+                    data = {
+                        'error': False,
+                        'new_revision_id': rev_form.instance.id,
+                    }
+
+                    return JsonResponse(data)
                 return redirect(doc)
+            else:
+                # If this is an Ajax POST, then return a JsonResponse with error
+                if request.is_ajax():
```
jwhitlock: This branch (failed AJAX edit of revision or combined translation form) is not covered by a test.

kuma/wiki/views/edit.py
```diff
@@ -210,36 +209,54 @@ def edit(request, document_slug, document_locale, revision_id=None):
                 if not rev_form.is_valid():
                     # Was there a mid-air collision?
                     if 'current_rev' in rev_form._errors:
+                        # Make the error message safe so the '<' and '>' don't
+                        # get turned into '&lt;' and '&gt;', respectively
+                        rev_form.errors['current_rev'][0] = mark_safe(rev_form.errors['current_rev'][0])
```
jwhitlock: Should this only be done for the AJAX return, or for the HTML error as well?

kuma/wiki/views/edit.py
```diff
@@ -210,36 +209,54 @@ def edit(request, document_slug, document_locale, revision_id=None):
...
-                        # revision, but the original revision differed from the
-                        # current revision at start of editing, we should tell
-                        # the client to refresh the page.
-                        response = HttpResponse('RESET')
-                        response['X-Frame-Options'] = 'SAMEORIGIN'
-                        response.status_code = 205
-                        return response
+                    # Is this an Ajax POST?
+                    if is_async_submit:
+                        if (is_raw and orig_rev is not None and
```
jwhitlock: Before, this code ran both for normal "save" and for the AJAX "save and keep editing", and now it just runs for the AJAX version. Was this intentional?

jwhitlock:
I've added a few notes where the backend logic appears to have changed, or where there is not test coverage of the new AJAX code.  This code is in bad need of a rewrite, but that's out of scope for this round.

I am also excited about a real AJAX interface, instead of the previous silent failures. I did no practical tests - I'm going to lean on @stephaniehobson to exercise this one.

kuma/wiki/views/edit.py
dchukhin: It seems like it should be fine to mark them both safe, based on the error messages I've seen. However, it's not necessary for HTML ones, so I'll make it only mark AJAX ones as safe.

kuma/wiki/views/edit.py
dchukhin: Actually, I don't think so either. These few lines are being removed.

dchukhin:
@jwhitlock @stephaniehobson ready for rereview

jwhitlock:
Tests in my VM against ``/en-US/docs/user:jwhitlock/Test2``

* Firefox 48: Publish works. Publish and Keep Editing:
   - Almost immediately says "Publishing failed. Please copy and paste your changes into a safe place and try submitting the form using the 'Publish' button."
   - After a delay, shows the regular editing screen with error message "Publishing failed. Conflicting edit attempts detected. Please copy and paste your edits to a safe place and visit the revision history page to see what was changed before making further edits."  Editing window shows the async change.
   - On the backend, there is one ``POST`` to ``/en-US/docs/user:jwhitlock/Test2$edit?async``, followed by a second ``POST`` to ``/en-US/docs/user:jwhitlock/Test2$edit``
* Chrome 52: Publish works. Publish and Keep Editing:
   - Says "Saving Changes...", disables the edit box
   - Says "Changes Saved!", but the editing box is still disabled
   - After a delay, shows the regular editing screen with error message "Publishing failed. Conflicting edit attempts detected. Please copy and paste your edits to a safe place and visit the revision history page to see what was changed before making further edits.". Editing window shows the async change.
   - On the backend, there is one ``POST`` to ``/en-US/docs/user:jwhitlock/Test2$edit?async``, followed by a second ``POST`` to ``/en-US/docs/user:jwhitlock/Test2$edit``
* Safari 9.1.2: Publish works. Publish and Keep Editing:
    - Does not show "Saving Changes..." box
    - After a delay, shows the regular editing screen with error message "Publishing failed. Conflicted edit attempts..." etc. The editing window shows the async change.
    - On the backend, there is one ``POST`` to ``/en-US/docs/user:jwhitlock/Test2$edit?async``, followed by a second ``POST`` to ``/en-US/docs/user:jwhitlock/Test2$edit``

What results do you get, @dchukhin ?

jwhitlock:
Oops, testing against the completely different branch ``1260197-save-and-edit-continue``. Missed the ``d``.

jwhitlock:
OK, with the proper code:

* Firefox 48.0: Publish ✓ Publish and Keep Editing ✓
* Chrome 52.0.2743.116: Publish ✓ Publish and Keep Editing ✓
* Safari 9.1.2: Publish ✓ Publish and Keep Editing ✓

It has the same dumb dev environment behavior, that the rendered page is the n-1 page.  But, it appears to work with "good edits".

jwhitlock:
My comments were addressed, and practical tests work for me.

stephaniehobson:
My comments were addressed. Not going to re-re-check practical tests. This is good to go.

PR #3936, opened 2016-08-11
fix bug 1287806 : Syntax correction on X-Frame-Options header

MatonAnthony:
[bug 1287806 : Syntax correction on X-Frame-Options header](https://bugzilla.mozilla.org/show_bug.cgi?id=1287806)

- Fix X-Frame-Options header syntax by removing an excedent ':'


jwhitlock:
@MatonAnthony thanks, this is the small first step I was looking for in PR #3922. I'm going to wait for a window with no other changes, so that I can merge and push this change in isolation, and revert it quickly if it breaks demos in an obvious way.

PR #3937, opened 2016-08-11
fix 1255145 :  Deprecation warning on TEMPLATE_DEBUG

MatonAnthony:
Remove TEMPLATE_DEBUG and replace it by DEBUG because the current instruction is
`TEMPLATE_DEBUG = DEBUG`


jwhitlock:
These appear to be the same issues encountered by PR #3815 (see [build output](https://travis-ci.org/mozilla/kuma/jobs/115877822)). It looks like [django-jinja](https://github.com/niwinz/django-jinja) has [several changes](https://github.com/niwinz/django-jinja/blob/master/CHANGES.adoc) around debugging. If an upgrade is required, there may be several other changes needed.

safwanrahman:
So the fact is, in the ``TEMPLATE`` Dictionary, we need to pass the Template Debugging information. The codes need to be re-factored according to it.
Its not a good idea to change ``TEMPLATE_DEBUG`` with ``DEBUG``.

MatonAnthony:
Yes, we figured that much out.
I'm not used to how django handle this, I'm not sure I will have any time to spend on it this weekend.

I have a C exam next week and a Mathematics exam the week after, I will more likely focus on those ;)

jwhitlock:
Current work in #3950, includes this commit.

PR #3930, opened 2016-08-03
bug 957802, 1286541: Documentation updates

jwhitlock:
* Update CKEditor documentation with current process
* Switch to readthedocs.io
* Document the iperceptions waffle flag

r? @stephaniehobson

PR #3938, opened 2016-08-12
Bug 1286541: Survey Gizmo task completion

stephaniehobson:
- switching survey provider from iPerceptions to Survey Gizmo
- include task-completion file when user is randomly selected by waffle
- task-completion file checks for availability of local storage before popping up invitation using notifier
- url has query string to configure Survey Gizmo
- user will not be asked again for 30 days

Testing:
- enable `sg_task_completion` for yourself
- open dev tools console
- delete any Local Storage items relating to helpfulness
- visit an article 
- check you get notifier
- click on survey link, should open in a new window (CMD and right click should work too)
- fill out survey
-- should get 5 seconds of "please fill this out after" before questions are visible
-- clicking "I'm done" should also make questions visible
-- when submitted with errors check errors are displayed properly, without making you wait the 5 second
- check thank you message is displayed
- at no point should you see javascript errors in the dev tools console

### DIFF
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -30,11 +30,6 @@
         ga('set', 'dimension12', {{ analytics_page_revision }});
     {%- endif %}
 
-    // dimension13 == "Saw iPerceptions Survey"
-    {% if waffle.flag('iperceptions') %}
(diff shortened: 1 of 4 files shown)


jwhitlock:
The steps worked for me. I've updated PR #3930 to include documentation of this flag.

PR #3940, opened 2016-08-15
bug 1295310: Improve display of Akismet data

jwhitlock:
In Django Admin, display Akismet data as a definition list (Unicode with escaped HTML) rather than as a JSON object. This will make it easier to read non-ASCII submissions. This is used in the admin for the wiki app:

* Akismet submissions
* Document spam attempts
* Revision IPs


jwhitlock:
looks good, @jwhitlock from 2 weeks ago

PR #3941, opened 2016-08-17
bug 1277577: Stop using some deprecated requirements

jwhitlock:
After these changes, these requirements will be unused and can be removed:

* ``BeautifulSoup``
* ``bitly-api``

It will be safer to remove the requirements after there is no more code using them.


jwhitlock:
Looks good, past @jwhitlock

PR #3883, opened 2016-06-02
Bug 1277577 - Upgrade dependencies, Q3 2016

jwhitlock:
Upgrade dependencies to latest versions for the quarter.

_Note: I'm opening this PR early so that TravisCI will start checking the updates, rather than waiting until the end._

PR #3943, opened 2016-08-18
bug 1296305 - Don't cache the connections page

jwhitlock:
Add a header to the [Account Connections page](https://developer.mozilla.org/en-US/users/account/connections) to prevent it from being cached:

```
Cache-Control: no-cache, no-store, must-revalidate
```

Also note that our custom ``never_cache`` decorator will not be needed after an upgrade to Django 1.9 or later.

### DIFF
--- a/kuma/core/decorators.py
+++ b/kuma/core/decorators.py
@@ -93,6 +93,8 @@ def permission_required(perm, login_url=None, redirect=REDIRECT_FIELD_NAME,
 # http://stackoverflow.com/a/2095648/571420
 # http://stackoverflow.com/a/2068407/571420
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching_FAQ
+# Fixed in Django 1.9:
+# https://docs.djangoproject.com/en/1.9/topics/http/decorators/#caching
 def never_cache(view_func):
     def _wrapped_view_func(request, *args, **kwargs):
         resp = view_func(request, *args, **kwargs)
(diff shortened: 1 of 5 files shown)


stephaniehobson:
I'm still not getting the account connected success message :(

stephaniehobson:
I also see this message in terminal. Not sure if it's relevant:

```
/home/vagrant/env/local/lib/python2.7/site-packages/allauth/socialaccount/providers/persona/provider.py:29: UserWarning: Skipping template backend backend because its render_to_string method doesn't support the context_instance argument.
```

jwhitlock:
I'll look closer at [bug 1229906](https://bugzilla.mozilla.org/show_bug.cgi?id=1229906).  I suspect the [Django messages framework](https://docs.djangoproject.com/en/1.10/ref/contrib/messages/) bits are missing from Jinja2 templates.

The warning is due to a mismatch between how Django's templates are rendered and how Jinja2 templates are rendered.  The code works with a warning, and upgrading ``django-allauth`` will make it go away. That's on my plate for the development environment updates (see PR #3883 for in-progress work).  It think it is unrelated to the messages, and can be ignored.

jwhitlock:
bug 1229906 was in our code - we threw away "account connected" messages unless the next page was the profile edit page.  I've updated the code to also accept the message on the account connections page, and rewrote the tests.  @stephaniehobson the PR has gotten hairier, but practical tests should work now.

stephaniehobson:
Forward progress! but not a fix :( I now get the notification when connecting on the account connections page but not when forwarded from the create new account page.

Merging because some progress is better than no progress and the account connections page is where we're telling everyone to go.

stephaniehobson:
R+

PR #3944, opened 2016-08-19
bug 1260264 - Spam dashboard

jbradberry:
Adds a spam dashboard, displaying the trends of spam over time, and a list of the most recent spam revisions.

bug 1260264 (https://bugzilla.mozilla.org/show_bug.cgi?id=1260264)
Taiga user story 351 (https://tree.taiga.io/project/viya-mdn-durable-team/us/351)

### DIFF
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -1,0 +1,1 @@
+{% extends 'base.html' %}
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -1,1 +1,53 @@
 {% extends 'base.html' %}
+{% set title = 'Spam Dashboard' %}
+
+{% block title %}{{ page_title(title) }}{% endblock %}
+
+{% block content %}
+<h1>{{ title }}</h1>
+{% if processing %}
+  <p>The report is being processed. Please reload in a minute.</p>
+{% else %}
+
+  {% if recent_spam %}
+  <table>
+    <caption>Published Spam</caption>
+    <colgroup>
+      <col span="1" class="data" />
+      <col span="1" class="data" />
+      <col span="1" class="data" />
+      <col span="1" class="data" />
+      <col span="1" class="name" />
+      <col span="1" class="path" />
+    </colgroup>
+    <thead>
+      <tr>
+        <th class="data">Date</th>
+        <th class="data">Time Active (seconds)</th>
+        <th class="data">Estimated Viewers</th>
+        <th class="data">Revision</th>
+        <th class="name">Change Type</th>
+        <th class="path">Path</th>
+      </tr>
+    </thead>
+    <tbody>
+      {% for rs in recent_spam -%}
+      <tr>
+        <td class="data">{{ rs.date }}</td>
+        <td class="data">{{ rs.time_active }}</td>
+        <td class="data">0</td>{# FIXME: make use of the Google Analytics query #}
+        <td class="data"><a href="{{ rs.revision_path }}">{{ rs.revision_id }}</a></td>
+        <td class="name">{{ rs.change_type }}</td>
+        <td class="path"><a href="{{ rs.document_path }}">{{ rs.document_path }}</a></td>
+      </tr>
+      {% endfor -%}
+    </tbody>
+  </table>
+  {% else %}
+  <p>No recent published spam.</p>
+  {% endif %}{# recent spam #}
+  <ul>
+    <li>Recent event data generated at {{ now }}</li>
+  </ul>
+{% endif %}{# processing #}
+{% endblock %}
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -89,7 +89,7 @@
   <p>No recent published spam.</p>
   {% endif %}{# recent spam #}
   <ul>
-    <li>Recent event data generated at {{ now }}</li>
+    <li>Recent event data generated at {{ events_generated }}</li>
   </ul>
 {% endif %}{# processing #}
 {% endblock %}
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -5,7 +5,7 @@
 
 {% set stat_ids=(('spam_viewers_change', '% Change', 'rate'),
                  ('spam_viewers', 'Spam Viewers', 'data'),
-                 ('spam_viewers_daily_average', 'Daily Average Viewers', 'data'),
+                 ('spam_viewers_daily_average', 'Daily Average Viewers', 'float'),
                  ('published_spam', 'Published Spam', 'data'),
                  ('blocked_spam', 'Blocked Spam', 'data'),
                  ('blocked_ham', 'Blocked Ham', 'data'),
@@ -36,6 +36,8 @@
         {%- for stat, stat_name, stat_class in stat_ids %}
         {%- if stat_class == 'rate' %}
         <td class="{{ stat_class }}">{{ "%0.1f%%" | format(100 * period.stats.get(stat, 0.0)) }}</td>
+        {%- elif stat_class == 'float' %}
+        <td class="{{ stat_class }}">{{ "%0.1f" | format(period.stats.get(stat, 0.0)) }}</td>
         {%- else %}
         <td class="{{ stat_class }}">{{ period.stats[stat] | default(0) }}</td>
         {%- endif %}
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -48,7 +48,7 @@
   </table>
 
   {% if recent_spam %}
-  <table>
+  <table id="recent-spam-table">
     <caption>Published Spam</caption>
     <colgroup>
       <col span="1" class="data" />
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -13,6 +13,9 @@
                  ('true_negative_rate', 'True Negative Rate', 'rate')) %}
 {% block content %}
 <h1>{{ title }}</h1>
+{% if improperly_configured %}
+  <p>{{ improperly_configured }}</p>
+{% endif %}
 {% if processing %}
   <p>The report is being processed. Please reload in a minute.</p>
 {% else %}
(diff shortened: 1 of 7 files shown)


kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+        # following, if there is one.
+        revisions = list(
+            document.revisions.filter(
+                id__gte=revision.id
+            ).only('id', 'created').order_by('id')[:2]
+        )
+
+        # How long was it active?
+        if len(revisions) == 1:
+            if document.deleted:
```
jwhitlock: There are no tests for deleted documents.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+                try:
+                    entry = DocumentDeletionLog.objects.filter(
+                        locale=document.locale, slug=document.slug
+                    ).latest('id')
+                    time_active_raw = entry.timestamp - revision.created
+                    time_active = int(time_active_raw.total_seconds())
+                except DocumentDeletionLog.DoesNotExist:
```
jwhitlock: There are no tests for deleted documents without DocumentDeletionLog entries.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+    (28, 'period_monthly', 'Monthly'),
+    (91, 'period_quarterly', 'Quarterly'),
+)
+
+
+def date_range(start, end):
+    """ Return an iterator providing the dates between `start` and `end`, inclusive.
+    """
+    if getattr(start, 'date', None) is not None:
+        start = start.date()
```
jwhitlock: There are no tests for passing a ``datetime`` as ``start``.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+    if getattr(end, 'date', None) is not None:
+        end = end.date()
```
jwhitlock: There are no tests for passing a ``datetime`` as ``end``.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+    blocked_edits = DocumentSpamAttempt.objects.filter(
+        created__range=(day, next_day)).only('review')
+    for blocked in blocked_edits:
+        # Is it a false positive?
+        if blocked.review == DocumentSpamAttempt.HAM:
+            counts['blocked_ham'] += 1
+        elif blocked.review == DocumentSpamAttempt.SPAM:
+            counts['blocked_spam'] += 1
+        else:
+            if blocked.review == DocumentSpamAttempt.NEEDS_REVIEW:
```
jwhitlock: There are no tests for blocked edits requiring review.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Keywords Arguments:
+    periods - a sequence of (days, identifier, name) tuples
+    end_date - The ending anchor date for the statistics
+    """
+    from .jobs import SpamDayStats, SpamDashboardRecentEvents
+
+    periods = periods or SPAM_PERIODS
+    if hasattr(end_date, 'date'):
+        end_date = end_date.date()
```
jwhitlock: There are no tests for passing a ``datetime`` as ``end_date``.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+    trends = defaultdict(Counter)
+    for day in dates:
+        # Gather daily raw stats
+        raw_events = job.get(day)
+        day_events = raw_events['events']
+
+        # Regenerate stats if there are change attempts with
+        # needs_review marked and the data is stale.
+        if raw_events['needs_review']:
+            generated = parser.parse(raw_events['generated'])
```
jwhitlock: There are no tests where an event needs review.

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+def spam_dashboard_recent_events(start=None, end=None):
+    """Gather data for recent spam events."""
+    now = timezone.now()
+    data = {
+        'now': now,
+        'recent_spam': [],
+    }
+
+    # Define the start and end dates/datetimes.
+    if not end:
```
jwhitlock: There are no tests with default values for ``start`` and ``end``.

kuma/dashboards/jinja2/dashboards/spam.html
```diff
@@ -0,0 +1,95 @@
...
+                 ('spam_viewers_daily_average', 'Daily Average Viewers', 'float'),
+                 ('published_spam', 'Published Spam', 'data'),
+                 ('blocked_spam', 'Blocked Spam', 'data'),
+                 ('blocked_ham', 'Blocked Ham', 'data'),
+                 ('true_positive_rate', 'True Positive Rate', 'rate'),
+                 ('true_negative_rate', 'True Negative Rate', 'rate')) %}
+{% block content %}
+<h1>{{ title }}</h1>
+{% if improperly_configured %}
+  <p>{{ improperly_configured }}</p>
```
jwhitlock: @stephaniehobson what format should be used to turn this into an error?  The content is something like:

```
<p>GOOGLE_ANALYTICS_CREDENTIALS Constance setting is badly formed.</p>
```

kuma/dashboards/jobs.py
```diff
@@ -0,0 +1,35 @@
...
+from .utils import (spam_day_stats,
+                    spam_dashboard_historical_stats,
+                    spam_dashboard_recent_events)
+
+
+class SpamDayStats(KumaJob):
+    """Cache spam stats for multiple days."""
+    lifetime = 60 * 60 * 24 * 7
+    fetch_on_miss = True
+    version = 8
```
jwhitlock: This should be version 1

kuma/dashboards/jobs.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def fetch(self, day):
+        return spam_day_stats(day)
+
+
+class SpamDashboardHistoricalStats(KumaJob):
+    """Cache historical spam stats for multiple days."""
+    lifetime = 60 * 60 * 24
+    fetch_on_miss = False
+    version = 12
```
jwhitlock: This should be version 1

kuma/dashboards/jobs.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def fetch(self, end_date):
+        return spam_dashboard_historical_stats(end_date=end_date)
+
+
+class SpamDashboardRecentEvents(KumaJob):
+    """Cache recent event data for a very short time."""
+    lifetime = 60
+    fetch_on_miss = True
+    version = 2
```
jwhitlock: This should be version 1

kuma/dashboards/utils.py
```diff
@@ -0,0 +1,281 @@
...
+    # Prepare output data
+    data = {
+        'version': 1,
+        'generated': datetime.datetime.now().isoformat(),
+        'end_date': end_date,
+        'periods': periods,
+        'trends': [],
+    }
+
+    job = SpamDashboardRecentEvents()
```
jwhitlock: The time that data is available bugs me, and it feels like the caches are wrong, but I don't have a solution yet.

* We know a published revision is spam as soon as a moderator marks it as spam
* The page views of a spam revision aren't known until 24 - 48 hours after the revision is reverted, due to data processing delays in Google Analytics.
* Once a revision is reverted, there should be no new page views, so you shouldn't have to ask again.  It is hard to determine if GA is done counting, though.
* Google Analytics uses sampling to determine the number of unique page views, and can change 10% - 20% from day to day, if you change the date range, etc.
* We need to know the page views for a revision for the long term stats, but we won't know the numbers for recent spam for 24-48 hours.

This suggests to me that we need:
* Long-term storage for revision page views that doesn't kick in until the revision is old enough
* Short-term storage or no storage (i.e. on-demand queries) for published spam
* Some storage for the collected statistics (day, week, month) for dashboard speed.

I'm inclined to drop all the job-level caching, and just use a 30-60 minute cache on the spam dashboard page itself, until we can figure out a better way to gather and store the page view data.

kuma/dashboards/tests/test_views.py
```diff
@@ -227,3 +232,346 @@ def test_known_authors_filter_ignored_with_username(self):
...
+        # The first response will say that the report is being processed
+        response = self.client.get(reverse('dashboards.spam', locale='en-US'))
+        eq_(200, response.status_code)
+
+        response2 = self.client.get(reverse('dashboards.spam', locale='en-US'))
+        page = pq(response2.content)
+        spam_trends_table = page.find('.spam-trends-table')
+        eq_(len(spam_trends_table), 1)
+
+    def test_spam_trends_stats(self, mock_analytics_upageviews):
```
jwhitlock: This is a hell of a test, but it works and exercises a lot. For branch coverage, please test the util functions directly.

kuma/dashboards/jinja2/dashboards/spam.html
```diff
@@ -0,0 +1,95 @@
...
+        <td class="path"><a href="{{ rs.document_path }}">{{ rs.document_path }}</a></td>
+      </tr>
+      {% endfor -%}
+    </tbody>
+  </table>
+  {% else %}
+  <p>No recent published spam.</p>
+  {% endif %}{# recent spam #}
+  <ul>
+    <li>Recent event data generated at {{ now }}</li>
```
jwhitlock: Should this be ``{{ generated }}``?

jwhitlock:
I tried this locally, and I was able to generate the data.  The code worked well. I'm tempted to push it as it is, and fix the issues in further PRs.

In production, when a job is used, the "generating data" response will return immediately while a background task is scheduled to grind through the data. Our development environments are weird, generating the data and then replying with the "generating data" response. Someday we should work on the dev environment.

Because of the weirdness about the lifetimes of the different data points, I'm not sure we should be caching it like we are.  The ``SpamDashboardHistoricalStats`` job is the primary job, with a lifetime of 24 hours, but uses data from the ``SpamDashboardRecentEvents`` job, with a lifetime of 60 seconds.  This looks like we'd have to wait 24 hours after a spam attack to see the list of spam in the attack, which may be required information during an attack.

I'm wondering if we should skip the jobs entirely, and instead just use a [per-view cache](https://docs.djangoproject.com/en/1.10/topics/cache/#the-per-view-cache) for the page, with a 5-minute timeout, and see how it goes.  Start simple, and add caching later if needed.

I have a lot of comments on missing coverage, but maybe we should decided on this caching issue first, before diving into these details.

kuma/dashboards/jinja2/dashboards/spam.html
jbradberry: Either works, since both utility functions involved generated their own timestamp of when they were run -- `now` for the recent spam events, `generated` for the spam stats.  Checking this did make me notice, however, that I have both timestamps being generated in different ways: one with Django's `timezone.now()` and the other one with `datetime.datetime.now()`.

jwhitlock:
We agreed to minimize the changes, and leave 1) tests and 2) unique page view storage for future PRs.

kuma/dashboards/jinja2/dashboards/spam.html
stephaniehobson: <div class="notification error" data-level="error">{{error}}</div>

PR #3945, opened 2016-08-22
[Bug 1294908] Improvement in article pages language picker

safwanrahman:
Copy the improvements from #3926

As it is front end thing, @stephaniehobson r?


stephaniehobson:
R :+1:

PR #3946, opened 2016-08-23
Fix Bug 1297484

stephaniehobson:
Move block of JavaScript into the document head.


kuma/wiki/jinja2/wiki/document.html
```diff
@@ -87,6 +78,16 @@
   <meta name="description" content="{{ seo_summary }}">
   <meta name="twitter:description" content="{{ seo_summary }}">
   {% endif %}
+
+  <!-- if this document was saved, we need to clear out the cached draft in wiki.js, need to get the value from the request if it exists -->
```
jwhitlock: Please change to ``{# if this document was saved ... #}`` so it is a template-only comment.

jwhitlock:
Works locally, 👍

PR #3947, opened 2016-08-23
[Bug 1294904] Language drop down menus should be sorted

safwanrahman:
So I have changed the way ``other_translations`` model function works. It should return a ``list``, but it was returning ``queryset``. It was done intentionally at #1953. Reading the conversation, got that it was changed because of hard to test. But, now I have seen some test for this function and I have changed the test to test the functionality with ``list`` instead of ``queryset``.

In order to change ``other_translations`` model function, I need to change the ``search/signals.py
`` so nothing will break.

Adding total 3 template tests so that we don't get regression in the future. I hope it works

Though the affect will be in the frontend, most of the changes are in the backend. so @jwhitlock r?

### DIFF
--- a/jinja2/includes/lang_switcher.html
+++ b/jinja2/includes/lang_switcher.html
@@ -1,5 +1,5 @@
 {% set default_locale = settings.LANGUAGE_CODE %}
-{% set dafault_language = settings.LOCALES[default_locale].native %}
+{% set default_language = settings.LOCALES[default_locale].native %}
 <form class="languages go" method="get" action="#">
   <label for="language">{{ _('Other languages:') }}</label>
   {% for name, value in request.GET.iteritems() %}
--- a/jinja2/includes/lang_switcher.html
+++ b/jinja2/includes/lang_switcher.html
@@ -1,3 +1,5 @@
+{% set default_locale = settings.LANGUAGE_CODE %}
+{% set dafault_language = settings.LOCALES[default_locale].native %}
 <form class="languages go" method="get" action="#">
   <label for="language">{{ _('Other languages:') }}</label>
   {% for name, value in request.GET.iteritems() %}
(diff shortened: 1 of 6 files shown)


kuma/search/signals.py
```diff
@@ -22,7 +22,7 @@ def render_done_handler(**kwargs):
         if outdated:
             log.info('Found a newer index and scheduled '
                      'indexing it after promotion.')
-        doc_pks = set(doc.other_translations.values_list('pk', flat=True))
+        doc_pks = set([item.pk for item in doc.other_translations])
```
jwhitlock: The previous version looked like it would work and be more efficient.  ``values_list`` just returns the PKs, where the second version loads then entire ``Document`` record from the database, for the same effect.  Did the previous version have the wrong result?

jinja2/includes/lang_switcher.html
```diff
@@ -1,3 +1,5 @@
+{% set default_locale = settings.LANGUAGE_CODE %}
+{% set dafault_language = settings.LOCALES[default_locale].native %}
```
jwhitlock: please change ``dafault_language`` to ``default_language`` (``a`` to ``e``), and in the code below.

jwhitlock:
Added first impressions from a quick code review. I'll read more in depth in the morning, run it locally.

kuma/search/signals.py
safwanrahman: In the previous version, ``doc.other_translations`` would return a queryset instead of a list. so we can just get the ``pk`` in list. The previous version had the correct result, but it was not sorted and document was in random order.
Moreover, if you take closer look in the ``other_translations`` model function, you will see two sql quires was being made to generate this in the past solution. this solution works with only single query

jwhitlock: Thanks, it only took you saying it twice for me to understand 😄 .

jwhitlock:
This works locally, tests pass and there's good coverage.

I am reluctant to change ``other_translations`` to return a list instead of a queryset, because 1) a queryset can be further filtered and sorted, and 2) it can be used more efficiently, such as just extracting IDs. However, I reviewed the other uses of this function in the code, and it appears that a list is a decent substitute.  I may revisit in the future, but template rendering efficiency isn't even on page 1 of my current priorities.

jwhitlock:
That just leaves the spelling fix for ``dafault_language`` to fix.

jwhitlock:
Looks good, but requires a rebase now.

stephaniehobson:
It doesn't look like they're being sorted by language code :/ This is a screen shot from a German translation:

screen shot 2016-08-26 at 16 09 48

safwanrahman:
@

jwhitlock:
@stephaniehobson It's sorted for me:

screenshot 2016-08-27 08 39 28

I'm not sure what's going on.

safwanrahman:
Maybe a Cache issue! @jwhitlock

stephaniehobson:
Working fine for me now ¯\_(ツ)_/¯ 

R+ after rebase.

safwanrahman:
rebased upon master and it can be merged now.
Both front end and back end reviewed. @jwhitlock can you please merge?

PR #3873, opened 2016-05-18
1246684 sample template

stephaniehobson:
Small change to an MDN plugin and an upgrade to CKEditor - mostly new language files.

PR #3948, opened 2016-08-24
1246684 sample template

stephaniehobson:


jwhitlock:
Replaces PR #3873 from May 2016 😮

jwhitlock:
👍 Works in dev VM

PR #3950, opened 2016-08-25
bug 1255145: Remove TEMPLATE_DEBUG, upgrade Jinja2

jwhitlock:
Upgrade ``django-jinja`` and ``Jinja2`` to latests packages, and then remove the ``TEMPLATE_DEBUG`` settings that was deprecated in Django 1.8.  Builds on PR #3937 and #3815.


requirements/default.txt
```diff
@@ -196,8 +196,9 @@ importlib==1.0.3 \
     --hash=sha256:65f342a604a2e1028707c5e055266ab2431c26e20fe10780b423320870884dac
 
 # Improved template engine
-Jinja2==2.7.3 \
-    --hash=sha256:2e24ac5d004db5714976a04ac0e80c6df6e47e98c354cb2c0d82f8879d4f8fdb
+Jinja2==2.8 \
+    --hash=sha256:1cc03ef32b64be19e0a5b54578dd790906a34943fe9102cfdae0d4495bd536b4 \
+    --hash=sha256:bc1ff2ff88dbfacefde4ddde471d1417d3b304e8df103a7a9437d47269201bf4
```
safwanrahman: Is it the hash of wheel?

jwhitlock: One of them is, yes

jwhitlock: I found this discussion, which I think is the source of the "we don't use wheels" rule for kitsune and input: https://github.com/mozilla/kitsune/pull/2482

We never used peep on MDN, preferring submodules instead 😱. With pip 8, hashes were added, and we switched from submodules to requirement files. We also use a constraints file, so anything less than pip 7.1 will fail, and we use pip 8 in the deployment pipeline.  With [hashin](https://github.com/peterbe/hashin), it's easier to include all the hashes than exclude any, so we'll keep both if available. I think kitsune should update to pip 8 and require it as well, but that's a different project.

kuma/settings/testing.py
```diff
@@ -1,6 +1,7 @@
 from .local import *  # noqa
 
 DEBUG = False
+TEMPLATES[0]['OPTIONS']['debug'] = True  # Enable recording of templates
```
safwanrahman: I think, this should be depend on debug value. Moreover template debug should on in developer settings

jwhitlock: There's a lot going on here, and this line is the minimum needed to get the 10 failing tests in  #3937 and previous tests to pass. And, template debugging is on for the "local" settings, which is what is used in development.

There's the open Django [ticket 24622](https://code.djangoproject.com/ticket/24622) that the Jinja2 backend added in Django 1.8 doesn't populate the ``context`` and ``templates`` variables of the ``Response`` class like the Django template system does.  You can read the long history, but basically the Django devs can't agree on a way to fix the problem, or even the right words to add to the documentation to point out the problem. One of the most visible consequence is that ``assertTemplateUsed`` doesn't work with Jinja2 templates, and ``response.context`` may be empty as well.

This is done in the Django template system by [overriding the Template render method](https://github.com/django/django/blob/e76981b43325da60b8a7475661df6cbfa7fda37e/django/test/utils.py#L137) with an [instrumented test renderer](https://github.com/django/django/blob/e76981b43325da60b8a7475661df6cbfa7fda37e/django/test/utils.py#L95-L101) that emits a [template_rendered signal](https://docs.djangoproject.com/en/1.8/ref/signals/#template-rendered) that is [connected to for test requests](https://github.com/django/django/blob/a02b5848ae6025ad3ef12cf6646ffd3e82a66067/django/test/client.py#L470) to call [a function](https://github.com/django/django/blob/a02b5848ae6025ad3ef12cf6646ffd3e82a66067/django/test/client.py#L164-L174) that keeps track of the templates and their context, and then [adds them to the response](https://github.com/django/django/blob/a02b5848ae6025ad3ef12cf6646ffd3e82a66067/django/test/client.py#L501-L502).  It's complicated.

The Jinja2 renderer would need similar hooks to alter the rendering code for tests, either as part of Jinja2, or in the Django code that integrates Jinja2.  See the Django ticket for the discussion and non-action for Django 1.8 - 1.10.

We're using ``django-jinja`` instead of the built-in Jinja2 backend, which can work around this issue.  Version 1.4.2 includes [debug instrumentation code](https://github.com/niwinz/django-jinja/commit/7b605259c1803dfe715cd721656d8e82d418daee#diff-50709203549beaedd6f5c6177955ef9dR65) that relies on the ``debug`` [option](https://github.com/niwinz/django-jinja/commit/7b605259c1803dfe715cd721656d8e82d418daee#diff-d19de7378cb30cef1eaa376a850ec459R91).  This code is [roughly the same today](https://github.com/niwinz/django-jinja/blob/b8073c2ae8f7c8350ab695c637b6a55e04998f69/django_jinja/backend.py#L86-L103), and the mechanism is still to only emit the ``template_rendered`` signal when the ``debug`` option is true.

So, ``TEMPLATES[0]['OPTIONS']['debug']`` must be ``True`` for tests, or any tests that looks at the templates used for rendering or the render context will fail.

The two ways to set ``debug`` to ``True`` are:

1. Implicitly set it to ``True`` by setting ``DEBUG = True``.  This works in [django-jinja](https://github.com/niwinz/django-jinja/blob/b8073c2ae8f7c8350ab695c637b6a55e04998f69/django_jinja/backend.py#L156) for Jinja2 templates as well as Django templates (See [upgrade docs for 1.8](https://docs.djangoproject.com/en/1.9/ref/templates/upgrading/#the-templates-settings)).
2. Explicitly set the ``debug`` option in ``TEMPLATES

For the case of [local development](https://github.com/mozilla/kuma/blob/master/kuma/settings/local.py#L11), we force ``DEBUG=True``, so template debugging is on as well.

For whatever reason, ``DEBUG=False`` is used in tests, so we have to use the second method and explicitly set the ``TEMPLATES`` option.  I'm doing the minimum changes for the update, so I'm not messing with the ``DEBUG=False`` in this PR.

pmac:
LGTM and the tests pass. r+ 👍

PR #3951, opened 2016-08-26
bug 1277577: Upgrade pipeline

jwhitlock:
django-pipeline 1.5.4 -> 1.6.8: Config change, bug fixes

I generated static files under old and new system, and the output is identical.

This does not include the compressor change in #3852, since those will require updating Vagrant and the deployment environments.

### DIFF
--- a/kuma/core/pipeline/cleancss.py
+++ b/kuma/core/pipeline/cleancss.py
@@ -3,4 +3,4 @@ from pipeline.compressors import SubProcessCompressor
 
 class CleanCSSCompressor(SubProcessCompressor):
     def compress_css(self, css):
-        return self.execute_command('cleancss', css)
+        return self.execute_command(['cleancss'], css)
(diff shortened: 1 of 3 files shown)


safwanrahman:
As far as I remember, we faced some trouble about some permission issue. Can you please test by having a test deploy in stage?

jwhitlock:
@safwanrahman Yes, we'd do that anyway. Can you find references to past issues on pipeline updates?

safwanrahman:
this patch can elaborate about the issue
https://github.com/jazzband/django-pipeline/pull/570

jwhitlock:
@jgmize this may a good place for your rebase or a code review

jgmize:
this lgtm, but as we discussed I think it's worth testing in the staging environment to make sure we don't run in to filesystem permissions issues like those mentioned by @safwanrahman.

jwhitlock:
OK, read the pipeline bug and https://github.com/mozilla/kitsune/pull/2871. I'm pushing the branch to stage, looking for permissions issues on deployment, before merging.

jwhitlock:
The test deploy to staging worked:

http://developeradm.private.scl3.mozilla.com/chief/developer.stage/logs/093c66300de0060a43dc79dff0e7cb62c690db4c.1472240097

The site renders with style and scripts downloading and executing, but also the CSS and JS hasn't changed, so maybe it isn't a good test. I will leave it up for the weekend in case someone has a better test to try against the running code.

PR #3952, opened 2016-08-29
bug 1277577 - Drop BeautifulSoup, bit.ly libraries

jwhitlock:
With PR #3941 merged, these are now unused requirements and can be removed.


pmac:
r+

PR #3953, opened 2016-08-29
bug 1277577 - Remove document smuggling

jwhitlock:
[django-smuggler](https://github.com/semente/django-smuggler) was used during the 2012 migration from MindTouch to Kuma, to allow admins to transfer documents between instances (see [bug 747137](https://bugzilla.mozilla.org/show_bug.cgi?id=747137)). Since that time, Kuma has added public APIs, and we've restricted access to unsanitized and sanitized production data. This functionality is no longer needed or desired.


jwhitlock:
Force-push to fix flake8

pmac:
Whole lotta red in the diffs. I like it 👍 

I haven't run this locally, but the tests pass and I don't see anything that looks like it'd be bad to remove. It's an r+ from me, but someone more familiar w/ the project might find something I didn't.

PR #3954, opened 2016-08-30
Bug 950945: Fix and improve MDN Sphinx template

jwhitlock:
* Fix the ``./manage.py generate_sphinx_template`` command
* Update the template to match the current MDN JS and CSS
* Omit some more items from the Sphinx template, such as social graph and Google Analytics
* Add generating the documentation to the documentation
* Update the docs to use the new version of the theme.

This PR is used to generate the changes in https://github.com/mdn/sphinx-theme/pull/10, which is necessary to release ``mdn-sphinx-theme`` version 2016.0 and fix some issues with the [current documentation](https://kuma.readthedocs.io/en/latest/).

@stephaniehobson I'd appreciate a code review, especially of ``base.html`` but I don't expect you to generate the docs locally unless you are looking to level up in backend development.

### DIFF
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -7,7 +7,9 @@ Generating Documentation
 This documentation is generated and published at
 `Read the Docs`_ whenever the master branch is updated.
 
-To generate locally, install the packages in ``requirements/docs.txt``, and run::
+To generate locally, install the packages in ``requirements/docs.txt``
+(inside the development VM or in a new `virtualenv`_ on the host system),
+and run::
 
     cd docs
     make html
@@ -41,3 +43,4 @@ To use the new theme in the public documentation,
 
 .. _`Read the Docs`: https://kuma.readthedocs.io/en/latest/
 .. _PyPI: https://pypi.python.org/pypi/mdn-sphinx-theme
+.. _virtualenv: https://virtualenv.pypa.io/en/stable/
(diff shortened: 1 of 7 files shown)


jwhitlock:
Force-push to remove the upgrade to unreleased ``mdn-sphinx-theme`` 2016.0.  One thing at a time...

docs/documentation.rst
```diff
@@ -0,0 +1,43 @@
+=============
+Documentation
+=============
+
+Generating Documentation
+------------------------
+This documentation is generated and published at
+`Read the Docs`_ whenever the master branch is updated.
+
+To generate locally, install the packages in ``requirements/docs.txt``, and run::
```
stephaniehobson: More direction here would be nice. Can we install these as part of the vagrant dev environment? The jinja version numbers seem to clash.

jinja2/base.html
```diff
@@ -233,11 +236,11 @@
 
     {{ providers_media_js() }}
     {% javascript 'main' %}
-
+    {% if not is_sphinx %}
```
stephaniehobson: Pointed out on IRC that we include analytics.js but not ga.js so removing this makes sense.

stephaniehobson:
Left some comments on https://github.com/mdn/sphinx-theme/pull/10

stephaniehobson:
Comments the other repo were addressed :+1:

PR #3956, opened 2016-09-01
fix bug 1254630: Open new browsing context when reporting a problem

merwan:
The options to "Report a bug" / "Report a content problem" open in a new browsing context because the content of the page you're currently on is very frequently needed in order to accomplish your goal.


stephaniehobson:
Thanks for the contribution!  This is perfect so I'm going to merge it.

I discovered bug while testing: our outbound analytics tracking script is still causing these links to open in the same window but that's not the fault of this PR. :+1:

PR #3957, opened 2016-09-02
bug 1295310: Adjust order of Akismet data in admin

jwhitlock:
When displaying Akismet data in the Django admin, display in an order that reflects the most interesting data for spam administrators:

1. The content diff
2. The page URL
3. The author's username
4. The author's email
5. Other data in the Akismet submission spec (alpha sort)
6. HTTP Headers and other metadata (alpha sort)

### DIFF
--- a/kuma/wiki/admin.py
+++ b/kuma/wiki/admin.py
@@ -260,12 +260,26 @@ SUBMISSION_NOT_AVAILABLE = mark_safe(
 
 
 def akismet_data_as_dl(akismet_data):
+    """Format Akismet data as a definition list."""
+    favorites = (
+        'comment_content',
+        'permalink',
+        'comment_author',
+        'comment_author_email')
+
+    def moderator_sort(key):
+        """Sort data by 1) favorites, 2) data values, 3) headers"""
+        try:
+            fav_order = favorites.index(key)
+        except ValueError:
+            fav_order = len(favorites)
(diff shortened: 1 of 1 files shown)


kuma/wiki/admin.py
```diff
@@ -260,12 +260,26 @@ def rendering_info(obj):
 
 
 def akismet_data_as_dl(akismet_data):
+    """Format Akismet data as a definition list."""
+    favorites = (
```
safwanrahman: I think its better to have this ordering in a dictionary rather than list. So it will be easily configurable.
Moreover, it should live in ``constants`` so we can configure it easily

jwhitlock: I do not know what you mean by ordering in a dictionary rather than a list. Sample code?

I disagree that moving it to another file makes it easier to configure. It only make it harder to understand the code, and where to make changes.

safwanrahman: like this 
```python
favorites = {'comment_content': 0,
             'permalink': 1,
             'comment_author': 2,
             'comment_author_email': 3}
try:
    fav_order = favorites[key]
except:
    .......
```

jwhitlock: Thanks for the example code. The sequence still seems like a better method to me, because it communicates the order with the least code. Encoding it as a dictionary opens questions like "do more items get added later?" and "what does it mean if the value is a non-integer?".

PR #3958, opened 2016-09-02
Bug 1254630: _blank in nav menus

stephaniehobson:
Watch for _blank on links when tracking, open in new tab.


jwhitlock:
I've confirmed that this is preventing the "open in new tab" behavior in production.  Sorry, this should have been an instant-merge.

PR #3963, opened 2016-09-09
bug 957802 - Update node provisioning

jwhitlock:
Update to the latest version of the ansible-galaxy role ``nodesource.node``, and then patch it to install the NodeSource apt GPG key from keyserver.ubuntu.com, since they switch to CloudFront and SNI for the old file server, and the old version of Python in Ubuntu 14.04 can't handle it.

### DIFF
--- a/provisioning/roles/nodesource.node/Dockerfile
+++ b/provisioning/roles/nodesource.node/Dockerfile
@@ -1,15 +1,16 @@
 FROM ubuntu:trusty
 MAINTAINER Mark Wolfe <mark@wolfe.id.au>
 
+# http://docs.ansible.com/ansible/intro_installation.html#latest-releases-via-apt-ubuntu
+RUN apt-get install software-properties-common -y --force-yes
+RUN apt-add-repository ppa:ansible/ansible
 RUN apt-get update
-RUN apt-get install -y --force-yes \
-	python-pycurl \
-	python-apt \
-	ansible \
-	lsb-release
+RUN apt-get install ansible -y --force-yes
+
 
 ENV WORKDIR /build/ansible-nodejs
 ADD . /build/ansible-nodejs
+ADD . /etc/ansible/roles/ansible-nodejs-role
 ADD ./tests/localhosts /etc/ansible/hosts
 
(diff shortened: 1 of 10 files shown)


safwanrahman:
So I have updated ``python`` version in my VM, delete the virtualenvironment and the provisioning worked. I have installed the python from [this ppa](https://launchpad.net/~fkrull/+archive/ubuntu/deadsnakes-python2.7)

So in my VM, the ``python -V`` returns ``Python 2.7.12``

So what we can do in our vagrant setup? We can either change the setup so it will install python 2.7.12 or we can change the box to ``ubuntu 16.06`` that have python 2.7.12 by default.

safwanrahman:
I forgot to add, I have also updated ansible in my host machine. downgrading ``ansible`` to 1.9.x replicate the problem

safwanrahman:
So I was investigating the issue for some hours, and it seems that, having ``python 2.7.12`` with ``ansible 2.1.1`` does not raise the problem in my local machine.

But the travis ci made me confuse. It is showing that its using ``python 2.7.12`` but I can not understand which hell build of python 2.7.12 they are using.
I have changed the script to install ``python 2.7.12`` from a PPA, and the [test](https://travis-ci.org/safwanrahman/kuma/jobs/158704861) is working fine in travis! 
This is the [patch](https://github.com/mozilla/kuma/compare/master...safwanrahman:master)

jwhitlock:
After discussion in IRC, changing the ansible rule is the fastest way to work around this problem.

Updating the system Python to 2.7.12 via the PPA works as well, in TravisCI's Ubuntu 12.04 environment, the beta 14.04 environment, and the Vagrant 14.04 environment.  It should be safe to update the system Python to a new minor version, although the official releases avoid it, requiring the PPA route.

Changing the system Python will break user's existing Vagrant environments, requiring manual steps to rebuild the virtual environment.  If we are going that route, it may make sense to update Vagrant to use 16.04, like the new Docker environment, and make people recreate their VMs.

This change has been submitted upstream as https://github.com/nodesource/ansible-nodejs-role/pull/34.  There are a lot of people impacted, so there may be an "accepted" fix in the coming weeks, maybe by switching ansible role providers.

PR #3964, opened 2016-09-09
Update travis.yaml

safwanrahman:
According to Travis CI documentation, in the newly added repository, by default ``sudo`` is false. Therefore I could not run Travis Ci build from my fork. adding ``sudo: required`` will make the build run in my fork. more information can be found [here](https://docs.travis-ci.com/user/ci-environment/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 branches:
     only:
         - master


jwhitlock:
This build was run with ``sudo: required``: https://travis-ci.org/mozilla/kuma/jobs/158697460/config

safwanrahman:
Because ``mozilla/kuma`` was added before 2015. So by default ``sudo`` is not false. But in my fork, when I add travis to run build, it has ``sudo: flase``

So adding it is necessary for local travis testing

safwanrahman:
This test fails in my fork
https://travis-ci.org/safwanrahman/kuma/jobs/158689430

jwhitlock:
I agree, it appears ``sudo: required`` is being implicitly applied to the mozilla/kuma repo, although I can't find the reference.  Explicit is better than implicit.

PR #3965, opened 2016-09-09
bug 1277577: Remove django-smuggler

jwhitlock:
After PR #3953 was deployed, this requirement is no longer used.


jwhitlock:
@jgmize r?

PR #3966, opened 2016-09-09
bug 950945: Update mdn-sphinx-theme to 2016.0

jwhitlock:
Uses the [new theme](https://pypi.python.org/pypi/mdn-sphinx-theme) tagged from https://github.com/mdn/sphinx-theme/pull/10 and generated from #3954.

r? @stephaniehobson


stephaniehobson:
:+1:

PR #3967, opened 2016-09-10
[Bug 1292624] Suggest moving translated pages after moving an English page

safwanrahman:
[Bug 1292624 comment 2](https://bugzilla.mozilla.org/show_bug.cgi?id=1292624#c2) describes to reproduce the issue and check the feature. Unfortunately, due to [Bug 1301867](https://bugzilla.mozilla.org/show_bug.cgi?id=1301867), the page move does not work and the email is not generated if ``CELERY_ALWAYS_EAGER`` is ``True``. So you should add ``CELERY_ALWAYS_EAGER = False`` in the ``.env`` file in order to get the test email in ``/tmp/kuma-messages``

@jwhitlock r?

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -220,14 +220,21 @@ def move_page(locale, slug, new_slug, email):
 
     full_url = settings.SITE_URL + '/' + locale + '/docs/' + new_slug
 
+    other_locale_urls = [settings.SITE_URL + translation.get_absolute_url()
+                         for translation in doc.translations.all().order_by('locale')]
+
     message = """
         Page move completed.
 
         The move requested for the document with slug %(slug)s in locale
         %(locale)s, and all its children, has been completed.
 
+        But following localized articles should be moved also:
+        %(locale_urls)s
+
         You can now view this document at its new location: %(full_url)s.
-    """ % {'slug': slug, 'locale': locale, 'full_url': full_url}
+    """ % {'slug': slug, 'locale': locale, 'full_url': full_url,
+           'locale_urls': '\n'.join(other_locale_urls)}
 
     send_mail(subject,
               textwrap.dedent(message),
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -220,8 +220,16 @@ def move_page(locale, slug, new_slug, email):
 
     full_url = settings.SITE_URL + '/' + locale + '/docs/' + new_slug
 
+    # Get the parent document, if parent doc is None, it means its the parent document
+    parent_doc = doc.parent or doc
+
     other_locale_urls = [settings.SITE_URL + translation.get_absolute_url()
-                         for translation in doc.translations.all().order_by('locale')]
+                         for translation in parent_doc.translations.exclude(locale=doc.locale)
+                                                                   .order_by('locale')]
+
+    # If the document is a translation we should include the parent document url to the list
+    if doc.parent:
+        other_locale_urls = [settings.SITE_URL + doc.parent.get_absolute_url()] + other_locale_urls
 
     message = """
         Page move completed.
@@ -229,7 +237,7 @@ def move_page(locale, slug, new_slug, email):
         The move requested for the document with slug %(slug)s in locale
         %(locale)s, and all its children, has been completed.
 
-        But following localized articles should be moved also:
+        But following localized articles may need to be moved also:
         %(locale_urls)s
 
         You can now view this document at its new location: %(full_url)s.
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -231,21 +231,21 @@ def move_page(locale, slug, new_slug, email):
     if doc.parent:
         other_locale_urls = [settings.SITE_URL + doc.parent.get_absolute_url()] + other_locale_urls
 
-    message = """
+    message = textwrap.dedent("""
         Page move completed.
 
         The move requested for the document with slug %(slug)s in locale
         %(locale)s, and all its children, has been completed.
 
-        But following localized articles may need to be moved also:
+        The following localized articles may need to be moved also:
         %(locale_urls)s
 
         You can now view this document at its new location: %(full_url)s.
-    """ % {'slug': slug, 'locale': locale, 'full_url': full_url,
-           'locale_urls': '\n'.join(other_locale_urls)}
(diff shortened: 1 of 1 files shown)


jwhitlock:
I will review this code later this week, after docker documentation tasks.

It appears all translations are listed. Does this include the page that was just moved?  For example, if I take the advice and move the French translation to a new slug, does it appear again in the confirmation message?

I don't think there is a simple way to make this a reducing list, where it only shows the translations you haven't moved.

safwanrahman:
As per we have discussed through IRC, from the backend we don not log the page moves in any simple way. So we have decided that, We should include all other locale link so the admin can see the links and understand which locale need moves.

Updated the patch so it will show other translations link and english article link while moving a translation.
There are no easy way to test this, So therefore I think unit test is not necessary at this moment

kuma/wiki/tasks.py
```diff
@@ -220,14 +220,29 @@ def move_page(locale, slug, new_slug, email):
...
+    if doc.parent:
+        other_locale_urls = [settings.SITE_URL + doc.parent.get_absolute_url()] + other_locale_urls
+
     message = """
         Page move completed.
 
         The move requested for the document with slug %(slug)s in locale
         %(locale)s, and all its children, has been completed.
 
+        But following localized articles may need to be moved also:
```
jwhitlock: How about "The" instead of "But", like "The following localized articles may need to be moved also:"

kuma/wiki/tasks.py
```diff
@@ -220,14 +220,29 @@ def move_page(locale, slug, new_slug, email):
...
+        %(locale_urls)s
+
         You can now view this document at its new location: %(full_url)s.
-    """ % {'slug': slug, 'locale': locale, 'full_url': full_url}
+    """ % {'slug': slug, 'locale': locale, 'full_url': full_url,
+           'locale_urls': '\n'.join(other_locale_urls)}
```
jwhitlock: Because of the indenting in the template, the join string needs to have 8 spaces, like:
```
         'locale_urls': '        \n'.join(other_locale_urls)}
```
Otherwise if there are two other translations, the message will also be indented. Alternatively, you can ``dedent`` the message template before applying the parameters.

Here's the email body for a page with two alternate translations, demonstrating the problem:

```
        Page move completed.

        The move requested for the document with slug user:jwhitlock/Test2 in locale
        es, and all its children, has been completed.

        But following localized articles may need to be moved also:
        https://developer-local.allizom.org/en-US/docs/Test3
https://developer-local.allizom.org/fr/docs/user:jwhitlock/Test2

        You can now view this document at its new location: https://developer-local.allizom.org/es/docs/Test3.
```

jwhitlock:
Good work on making this a small, focused change that works.

The page move code has a few bugs and a bad design, and needs to be rewritten. I'd include this task - the email text should be in a template file, not inline code.  Since this is a small improvement that requires manual testing anyway, I'm happy to defer tests until the rewrite.

The URLs were included as expected when moving an English or translated page.  The output email needs adjusting if there is more than one translation, as noted in the review comments.

kuma/wiki/tasks.py
safwanrahman: Cool

kuma/wiki/tasks.py
safwanrahman: So we have discussed over IRC and both decided that, using ``dedent`` is more pythonic way to fix the indention problem.
So I am doing that.

safwanrahman:
@jwhitlock r?

jwhitlock:
Long story short: 👍 

I had to rebase to run this in the new docker environment, as expected.  I set up some pages in a fresh database:

* http://localhost:8000/en-US/docs/MadTest
* http://localhost:8000/en-US/docs/MadTest/PageOne
* http://localhost:8000/en-US/docs/MadTest/PageTwo
* http://localhost:8000/de/docs/MadTest
* http://localhost:8000/fr/docs/MadTest

And then I moved http://localhost:8000/en-US/docs/MadTest. The email was added as a file in the ``worker`` container, at ``/tmp/kuma-messages/20160929-212745-139781841802704.log``, with this content:

```
MIME-Version: 1.0
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: 7bit
Subject: Page move completed: MadTest (en-US)
From: notifications@developer.mozilla.org
To: admin@example.com
Date: Thu, 29 Sep 2016 21:27:45 -0000
Message-ID: <20160929212745.12.48810@d4788729ba6a>


Page move completed.

The move requested for the document with slug MadTest in locale
en-US, and all its children, has been completed.

The following localized articles may need to be moved also:
https://developer-local.allizom.org/de/docs/MadTest
https://developer-local.allizom.org/fr/docs/MadTest

You can now view this document at its new location: https://developer-local.allizom.org/en-US/docs/MoveTest.
```

I'd prefer a space between the text and the list of articles, but my mail client will change the color of the links, which will probably be a sufficient change in syntax.

When I moved the German article, the email had this content:

```
MIME-Version: 1.0
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: 7bit
Subject: Page move completed: MadTest (de)
From: notifications@developer.mozilla.org
To: admin@example.com
Date: Thu, 29 Sep 2016 21:57:40 -0000
Message-ID: <20160929215740.15.47682@d4788729ba6a>


Page move completed.

The move requested for the document with slug MadTest in locale
de, and all its children, has been completed.

The following localized articles may need to be moved also:
https://developer-local.allizom.org/en-US/docs/MoveTest
https://developer-local.allizom.org/fr/docs/MadTest

You can now view this document at its new location: https://developer-local.allizom.org/de/docs/MoveTest.

-------------------------------------------------------------------------------
```

So, I think this is working.  There are some issues with the Docker dev environment:

* It might be be nice if all the Kuma containers (``web``, ``worker``, and ``api``) shared the same ``/tmp/kuma-messages``.
* The ``worker`` process (and the rest) continue to load ``kuma/settings/local.py``, which overrides the environment variables with other settings (``SITE_URL = https://developer-local.allizom.org``). We should make sure everything in ``local.py`` can be set in ``common.py``, and then set ``DJANGO_SETTINGS_MODULE`` to ``kuma.settings.common``.
* I'm getting a KumaScript rendering error, ``Problem fetching source document: undefined``.  I think that's because this PR is so old it has the KS that doesn't include loading settings from the environment.

PR #3960, opened 2016-09-07
bug 1301116: Upgrade setuptools

jwhitlock:
setuptools 20.7.0 -> 26.1.1: Many fixes

The necessary fix is from 20.8.1:

https://github.com/pypa/setuptools/issues/544

The latest version has been active for a month, which seems like a long time for [setuptools releases](https://setuptools.readthedocs.io/en/latest/history.html).

This will break the Docker development environment, because it can't replace the system-installed setuptools 20.7.0.

PR #3968, opened 2016-09-12
Setuptools 26.1.1

jgmize:
This builds on #3960 but installs setuptools 26.1.1 without hash verification in `Dockerfile-base` as a workaround for the `AttributeError: 'Requirement' object has no attribute 'project_name'` error we get when attempting to use hash verification.

### DIFF
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -11,7 +11,7 @@ EXPOSE 8000
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends python2.7 libpython2.7 python-dev \
-    python-pip python-setuptools gettext build-essential \
+    python-pip gettext build-essential \
     nodejs npm \
     libtidy-0.99-0 libtidy-dev \
     libxml2-dev libxslt1.1 libxslt1-dev \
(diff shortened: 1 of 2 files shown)


jgmize:
@jwhitlock r?

safwanrahman:
Just a note, this Error will only raise with ``pip 8.1.2``
so if you change to a lower version, it will not raise!

jwhitlock:
@safwanrahman the pip version installed is pip 8.1.1, also doesn't work.

This worked for me, with some trial and error. Recreating my command line (so I might have gotten something wrong):

```
$ git fetch jgmize
$ git checkout setuptools-26.1.1
$ make build-base
$ docker tag 27f7a89 quay.io/mozmar/kuma_base:latest
$ make build-kuma
$ docker tag 35a6d393fcec quay.io/mozmar/kuma:latest
$ docker-compose up
$ docker exec -it kuma_web_1 /bin/bash
# ./manage.py migrate
# ./manage.py update_product_details -f
# pip install django-babel==0.4.0
# make localeextract  # Success!
# <Ctrl-D to exit>
# docker restart kuma_web_1  # http://localhost:8000/en-US/ works w/o server error
```

I'll refine the fresh start commands as I write the docs.

jgmize:
This specific backwards incompatibility issue also affects pip 8.1.1-2ubuntu0.1, see https://github.com/peterbe/hashin/pull/23#issue-157047068 and https://github.com/pypa/pip/issues/3687 for more background.

PR #3970, opened 2016-09-12
bug 1218563: Add basic install docs for Docker

jwhitlock:
Update the Docker installation documents for the new Docker environment, and get them to the "view the homepage" step in the Vagrant documentation.

The double-installation of product details is not ideal (once in the image, once again with the ``git clone`` checkout), but documenting what it looks like now, while I work on a PR to change it ([bug 1302459](https://bugzilla.mozilla.org/show_bug.cgi?id=1302459))

@stephaniehobson can you try the instructions to make sure they are clear?  The empty database path is sufficient for the homepage test.

@jgmize can you do a sanity check?

### DIFF
--- a/docs/installation-docker.rst
+++ b/docs/installation-docker.rst
@@ -10,70 +10,119 @@ Vagrant set up.
 Docker setup
 ============
 
-#. Ensure you have cloned the kuma Git repository and it is your current
-   working directory::
+#. Install the `Docker platform`_, following Docker's instructions for your
+   operating system.
+
+   .. _Docker platform: https://www.docker.com/products/overview
+
+#. Clone the kuma Git repository, if you haven't already::
 
         git clone git@github.com:mozilla/kuma.git
+
+#. Ensure you are in the existing or newly cloned kuma working copy::
+
         cd kuma
 
-#. Install the `Docker Toolbox`_. This will install the Docker engine,
-   along with Docker Compose and Docker Machine.
-
-   .. _Docker Toolbox: https://www.docker.com/products/docker-toolbox
-
-#. Create the virtual machine that Docker will run within::
-
-        docker-machine create --driver virtualbox --virtualbox-memory 4096 --virtualbox-disk-size 50000 kuma
-
-#. Configure the environment variables::
-
-        eval "$(docker-machine env kuma)"
-
-#. Add the Docker VM to your hosts::
-
-        sudo sh -c "echo $(docker-machine ip kuma 2>/dev/null)  kuma.dev >> /etc/hosts"
-
-#. Pull and build the containers::
+#. Pull the Docker images and build the containers::
 
         docker-compose pull
         docker-compose build
 
-#. Start the containers::
+#. Start the containers in the background::
 
         docker-compose up -d
 
-At this point everything is running but we haven't set up the database
-yet. You either need to import a data dump or run the Django migrations to
-create the database and tables.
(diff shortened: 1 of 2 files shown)


jgmize:
Nice work @jwhitlock, this looks pretty good on my first pass.

docs/installation-docker.rst
```diff
@@ -16,64 +16,111 @@ Docker setup
...
 
-#. Add the Docker VM to your hosts::
+Install Product Details
+=======================
+The `product details JSON`_ includes common data for Mozilla products,
+including the list of supported languages. It needs to be installed once,
+with::
 
-        sudo sh -c "echo $(docker-machine ip kuma 2>/dev/null)  kuma.dev >> /etc/hosts"
+    docker exec -it kuma_web_1 ./manage update_product_details -f
```
safwanrahman: small nit: ``.py`` is missing. It should be ``./manage.py``

jwhitlock: Not small - wrong! Thanks, will fix.

jwhitlock:
Force-push to [fix ``manage.py``](https://github.com/mozilla/kuma/pull/3970#discussion_r78515698), merge commits.

jwhitlock:
Force-push to standardize section headers to active voice, such as "Provision the database" instead of "Provisioning the database"

docs/installation-docker.rst
```diff
@@ -16,64 +16,111 @@ Docker setup
...
-#. Install the `Docker Toolbox`_. This will install the Docker engine,
-   along with Docker Compose and Docker Machine.
+#. Install the `Docker platform`_, following Docker's instructions for your
+   operating system.
 
-   .. _Docker Toolbox: https://www.docker.com/products/docker-toolbox
+   .. _Docker platform: https://www.docker.com/products/overview
 
-#. Create the virtual machine that Docker will run within::
+#. Pull the Docker images and build the containers::
```
stephaniehobson: Where do I do this?  Kuma directory?

jwhitlock: Yes, in the folder with the ``docker-compose.yml`` file.

stephaniehobson:
As instructed, did not attempt the database import.
Ran into Bug 1302459 but otherwise got up and running without a problem.

It would be nice to add where the commands should be run for complete and utter n00bs like myself but it's not a blocker.

jwhitlock:
I think all the commands are run in the kuma checkout directory. Please tell me if you had to change directories at some point.

I think it may be clearer if the first two "Docker setup" commands are reversed.  Something like:

1. Install the [Docker platform](Docker platform: https://www.docker.com/products/overview), following Docker's instructions for your operating system.
2. Clone the kuma Git repository and make it your current working directory:

        git clone git@github.com:mozilla/kuma.git
        cd kuma

3. Pull the Docker images and build the containers::

        docker-compose pull
        docker-compose build

Would that work, or do we need to be more explicit about the directory?  Maybe split the clone / current directory into two steps, so people with a current checkout don't skip the ``cd`` step?

stephaniehobson:
Yeah, your proposed change is clear-er :) I did not have to change directories at any point.

jwhitlock:
Force-push to rearrange the "Docker setup" section.

jwhitlock:
@stephaniehobson @jgmize are we good to merge?

stephaniehobson:
:+1:

PR #3971, opened 2016-09-13
Bug 1260264: Formating Change For Spam Dashboard

stephaniehobson:
Small changes to improve display of spam dashboard.

Added 'text-content' parent class so inherits wiki table format.
Styled configuration error and processing messages.


jwhitlock:
👍 Looks good to this backend dev, and the output looks so much better.  It really makes the columns in the Published Spam table stand out (they are ``class="data"``, etc. should be ``class="stat-header"``).

jwhitlock:
😻 So much nicer.

PR #3972, opened 2016-09-13
bug 1218563: Disable pipeline in Vagrant development

jwhitlock:
Previously, the django-pipeline settings [PIPELINE_ENABLED](https://django-pipeline.readthedocs.io/en/latest/configuration.html#pipeline-enabled) and [PIPELINE_COLLECTOR_ENABLED](https://django-pipeline.readthedocs.io/en/latest/configuration.html#pipeline-collector-enabled) were unset, and were set to default values at runtime (Off for ``DEBUG=True``, on for ``DEBUG=False`` in production).

The Docker configuration sets them explictly, either from environment variables or the documented defaults.

The Vagrant dev environment starts with ``DEBUG=False``, and then resets it to ``DEBUG=True`` in ``settings/local.py``, but not until the pipeline is enabled for production, breaking static assets.

This code updates the Vagrant settings that get set to production defaults, and adds a note that ``local.py`` has a short life remaining.

### DIFF
--- a/kuma/settings/local.py
+++ b/kuma/settings/local.py
@@ -1,6 +1,9 @@
 import logging
 from .common import *  # noqa
 
+# Settings for Vagrant Development
+# TODO: Use environment to override, not settings picker
+
 DEFAULT_FILE_STORAGE = 'kuma.core.storage.KumaHttpStorage'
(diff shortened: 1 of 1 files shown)


jwhitlock:
Merged in PR #3978

PR #3973, opened 2016-09-14
bug 1302459: Remove run-time usage of product details' languages.json

jwhitlock:
Move ``MDN_LANGUAGES`` and ``LOCALES`` to ``kuma/settings/languages.py``.  ``LOCALES`` is hard-coded (from the perspective of running code).

Previously. ``LOCALES`` was derived at runtime from ``languages.json`` (from [Mozilla product details](https://product-details.mozilla.org/1.0/)).  This particular data changes rarely. Since May 2015, these changes were made upstream:

* Added locales for Bodo (brx), Kaqchikel (cak), Guarani (gn), Konkani (kok), Kashmiri (ks), Latgalian (ltg), and Purépecha (tsz)
* Updated the English name for the locale with code 'or' from Oriya to Odia
* Updated the native name for Romanian (ro) from română to Română

Only this last change affected MDN.

Changing to this new method:

1. Removes a step from installation and deployment across all environments (the initial pain point for starting this PR)
2. Avoids processing a JSON document at startup
3. Removes a 3rd-party dependency

``kuma/settings/language.py`` can be regenerated or updated from ``mdn_languages.txt`` (maintained by the MDN team) and ``languages.json`` (downloaded from [Mozilla product details](https://product-details.mozilla.org/1.0/) with ``./manage.py generate_languages_settings``.

This is a big PR, to make it easier to review the full change. I'm open to discussion of the design and breaking it out into smaller PRs.  It is not urgent to merge this code, but removing it makes everything, including installing the new Docker environment, a little smoother.


safwanrahman:
I think, its better to parse the json in the runtime rather than generating a python file with dictionary. We can maintain our own copy of ``languages.json`` because it does not change so often. Also in future we may face some locale merges that only affect MDN like ``sr-Cyrl`` to ``sr``. So I think keeping own version of languages.json and parsing in the runtime is a better approach to do this.

jwhitlock:
Force push to rebase on master, but I think I'm going to open a new PR to address @safwanrahman's concerns, with an less ambitious PR.

jwhitlock:
Closing in preference for #3979

PR #3975, opened 2016-09-15
Bug 1218563: Advanced install instructions

jwhitlock:
Continue translating the [Vagrant install instructions](https://kuma.readthedocs.io/en/latest/installation.html), after the point the home page is viewable.  

This requires more environment overrides to accommodate the development environment, which does not use SSL at this time.  It also set the default character set and collation for MySQL, which is used to create the empty database with our settings, rather than ``latin1``.

These new settings are applied at container creation, so if you have already run ``docker-compose up``, you'll need to remove the existing containers so they can be recreated:

    docker-compose stop
    docker-compose rm
    docker-compose up

You'll have to re-run the basic installation instructions to get a database setup.


jwhitlock:
These now work for me. r @jgmize?

docs/installation-docker.rst
```diff
@@ -78,21 +99,169 @@ minutes depending on the size of the data dump.
...
+
+    docker exec -it kuma_web_1 ./manage.py createsuperuser
+
+This will prompt you for a username, email address (a fake address like
+``admin@example.com`` will work), and a password.
+
+If your database has an existing account that you want to use, use the Django
+shell, similar to this::
+
+    docker exec -it kuma_web_1 ./manage.py shell_plus
```
safwanrahman: Can we create a management command for this? Just a suggestion

jgmize: I added a few make commands in a separate branch for convenience: https://github.com/jgmize/kuma/commit/69f114c5fdef234799ce748b59565e9e51a3307e, but I wasn't sure of their value vs documenting the docker commands, and neither was @jwhitlock when we discussed it briefly yesterday, so I haven't opened a PR, but I could be easily convinced to open one for further discussion.

jwhitlock: If you are talking about the commands to upgrade an existing user, you may want to add an optional password prompt to [ihavepower](https://github.com/mozilla/kuma/blob/master/kuma/core/management/commands/ihavepower.py) in a new PR, and upgrade the instructions to:

```
docker exec -it kuma_web_1 ./manage.py ihavepower admin_username -p
```

docs/installation-docker.rst
```diff
@@ -78,21 +99,169 @@ minutes depending on the size of the data dump.
...
+
+#. Log in as the admin user.
+#. Open the `Constance / Config`_ section of the admin site.
+#. Change ``KUMASCRIPT_TIMEOUT`` to 600.
+#. Click "SAVE" at the bottom of the page.
+#. Import the `KumaScript auto-loaded modules`_:
+
+::
+
+   docker exec -it kuma_web_1 ./manage.py import_kumascript_modules
```
jgmize: The first time I ran this, I got an error: `django.db.utils.OperationalError: (1366, 'Incorrect string value: \'\\xD8\\x8 C", "...\' for column \'content\' at row 1')`. However, running it again appears to have succeeded.

jgmize:
Overall this is looking pretty good once I worked around the entrypoint issue; none of my other comments are intended to block merging this, but I'm looking forward the separate PR we discussed where we combine this with the vagrant install and refactor all the `docker exec -it kuma_web_1 $COMMAND` entries into a single `docker exec -it kuma_web_1 bash` at the top, explaining how it's equivalent to a `vagrant ssh` and then we can just document the commands themselves independent of whether running in vagrant or docker.

docs/installation-docker.rst
jwhitlock: This implies that the database initialization was incorrect, and you ended up with a latin1 encoding.  It succeeded the second time because it skips download of the half-initialized macro

docs/installation-docker.rst
```diff
@@ -78,21 +99,169 @@ minutes depending on the size of the data dump.
...
+    ./af/LC_MESSAGES/django.po:2: warning: header field 'Last-Translator' still has the initial default value
+    ...
+    ./zu/LC_MESSAGES/promote-mdn.po:4: warning: header field 'PO-Revision-Date' still has the initial default value
+    ./zu/LC_MESSAGES/promote-mdn.po:4: warning: header field 'Last-Translator' still has the initial default value
+
+Warnings are OK, and will be fixed as translators update the strings on
+Pontoon_.  If there is an error, the output will end with the error, such as::
+
+    ./az/LC_MESSAGES/django.po:263: 'msgid' and 'msgstr' entries do not both end with '\n'
+    msgfmt: found 1 fatal error
```
jwhitlock: Yes, I plan to fix it before the next deploy.

docker-compose.yml
```diff
@@ -52,6 +58,10 @@ mysql-data:
 
 mysql:
   build: ./docker/images/mysql
+  entrypoint:
+    - docker-entrypoint.sh
```
jgmize: So I tested this on a different machine, and it worked without the change I suggested above. After further research I found out that the mysql:5.6 image had been cached on my laptop from 14 months ago, and the newer version with the same tag had a new file, `/usr/local/bin/docker-entrypoint.sh`, which was not present in the older version. The `/entrypoint.sh` file is still there in the newer version and there was no difference between the two files. Note that I had run `docker-compose pull` before posting this comment, but that does *not* update recursively. This is an edge case that most developers won't run into, but the change wouldn't hurt.

jwhitlock: I picked ``docker-entrypoint.sh`` because it is the ``ENTRYPOINT`` defined in the [Dockerfile](https://github.com/docker-library/mysql/blob/9fc086343ebd36af0448438622188264d1dc2e1c/5.6/Dockerfile#L59-L61), and lists ``entrypoint.sh`` as the backward compatible name.

I am disappointed in ``docker-compose pull``, but it is good to know this behavior.

safwanrahman: We can add ``--no-cache`` flag so It will not use any of the cached image.

jgmize: Let's keep this as-is, the set of developers affected by the backwards compatibility issue is very small, and likely of size 1.

jwhitlock: Docs for ``docker-compose pull --help``:

```
Pulls images for services.

Usage: pull [options] [SERVICE...]

Options:
    --ignore-pull-failures  Pull what it can and ignores images with pull failures.
```

I don't see ``--no-cache`` as an option, @safwanrahman.  I'm adding "Update stale docker images" to my future docker debugging docs.

jwhitlock:
I see some TODO items:

* [ ] Update ``./manage.py ihavepower`` to take optional password
* [ ] Fix the az locale in master
* [ ] Test localized string compiling in TravisCI
* [ ] Switch to using ``./manage.py waffle_flag kumaediting --everyone --create`` to create / update the waffle flag
* [ ] Add execution alternatives for running Docker commands
* [ ] Investigate single installation instructions for Vagrant / Docker (I'm doubtful this will be possible)

It appears the changes to the ``docker-compose.yml`` are accepted however, so I'm going ahead with merging this. and tracking these TODO items... somewhere.  Maybe [Taiga User Story 708](https://tree.taiga.io/project/viya-mdn-durable-team/us/708)?

jwhitlock:
Moved TODO list to [Google Doc](https://docs.google.com/document/d/1cnuw2F25gtUPeHCXuu8zYf5H9v_UPdgmanJ7oVM6uoc/edit#), starting to triage as must have Q3 / nice to have Q3 / defer to future

PR #3976, opened 2016-09-17
[Bug 1218563] Adding elasticsearch to docker

safwanrahman:
It enables elasticsearch in docker and All the unit tests pass after merging #3975
@jwhitlock r?

PR #3977, opened 2016-09-18
[Bug 1218563] Run test in Docker environment with Travis CI

safwanrahman:
All the test will pass after merging #3975 #3976 
I have tried and this is the [result](https://travis-ci.org/safwanrahman/kuma/jobs/160749971)

We must configure travis to build and run all the test in docker environment  so we can track if any of our patch make the docker environment broken.

Instead of pulling ``kuma_base`` image, I have configured to build the image locally so it will be possible to test against the ``Docker-base``. So in future we can avoid any regression and it will be easier to track.

Regarding travis CI, Docker is only supported in ``Ubuntu Trusty`` Environment and unfortunately, ``Elasticsearch`` service [is not available in the trusty environment](https://docs.travis-ci.com/user/trusty-ci-environment/#Data-Stores). So needed to configure the build script in order to download the ``elasticsearch`` and install it in the environment.

@jwhitlock @jgmize r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,15 @@ language: python
 python:
     - "2.7"
 services:
+    - docker
     - memcached
-    - elasticsearch
 env:
     matrix:
         - TOXENV=py27
         - TOXENV=flake8
         - TOXENV=docs
         - TOXENV=dennis
+        - TOXENV=docker DOCKER_COMPOSE_VERSION=1.8.0
     global:
         - DJANGO_SETTINGS_MODULE=kuma.settings.travis
         - DEBIAN_FRONTEND=noninteractive
@@ -26,6 +27,7 @@ env:
 matrix:
     allow_failures:
         - env: TOXENV=dennis
+        - env: TOXENV=docker DOCKER_COMPOSE_VERSION=1.8.0
 before_install:
     - scripts/travis-install
     - pip install -U pip
--- a/scripts/docker-compose.travis.yml
+++ b/scripts/docker-compose.travis.yml
@@ -1,0 +1,10 @@
+worker: &worker
+  build: .
+  dockerfile: Dockerfile-base
+  environment:
+    - DJANGO_SETTINGS_MODULE
+    - DEBIAN_FRONTEND
+    - CFLAGS
+
+web:
+  <<: *worker
--- a/scripts/docker-compose.travis.yml
+++ b/scripts/docker-compose.travis.yml
@@ -8,3 +8,6 @@ worker: &worker
 
 web:
   <<: *worker
+
+api:
+  <<: *worker
--- a/scripts/docker-compose.travis.yml
+++ b/scripts/docker-compose.travis.yml
@@ -1,13 +1,17 @@
-worker: &worker
-  build: .
-  dockerfile: Dockerfile-base
-  environment:
-    - DJANGO_SETTINGS_MODULE
-    - DEBIAN_FRONTEND
-    - CFLAGS
+version: '2'
+services:
+  worker: &worker
+    build:
+      context: .
+      dockerfile: Dockerfile-base
+    environment:
+      - DJANGO_SETTINGS_MODULE
+      - DEBIAN_FRONTEND
+      - CFLAGS
 
-web:
-  <<: *worker
+  web:
+    <<: *worker
 
-api:
-  <<: *worker
+  api:
+    <<: *worker
+
--- a/scripts/travis-install
+++ b/scripts/travis-install
@@ -31,6 +31,4 @@ then
     curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
     chmod +x docker-compose
     sudo mv docker-compose /usr/local/bin
-    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml pull
-    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml build
 fi
--- a/scripts/travis-install
+++ b/scripts/travis-install
@@ -15,10 +15,22 @@ then
     sudo apt-get autoremove
     sudo apt-get autoclean
     sudo rm -rf /var/lib/mysql
-    sudo truncate -s 0 /var/log/mysql/error.log
 
+    # Install Elasticsearch
+    wget https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.2.deb
+    sudo dpkg -i elasticsearch-1.7.2.deb
+    
     # limit elasticsearch / java memory usage to avoid OOM Killer
     sudo service elasticsearch stop;
     echo "ES_HEAP_SIZE=256m" | sudo tee --append /etc/default/elasticsearch
     sudo service elasticsearch start;
+fi
+
+if [ "$TOXENV" == "docker" ]
+then
+    curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+    chmod +x docker-compose
+    sudo mv docker-compose /usr/local/bin
+    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml pull
+    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml build
 fi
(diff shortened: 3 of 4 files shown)


jgmize:
Travis test runs are too slow already IMO, and installing elasticsearch and running integration tests against it every time a PR is opened or updated seems like a waste of time and resources. How about we pass an `--ignore=kuma/search` argument to `py.test` instead, and run those tests manually on the relatively rare occasion that a PR affects search?

jwhitlock:
So much to say.

I think the full test suite should be run somewhere, and I think TravisCI is a good place to do it.  It's TravisCI's resources that get used, not ours.

I do like the idea of a much faster test suite for local development, where the cycle time can be a burden. We've got a few strategies planned to address that, including moving some integration tests from Django to a front-end test suite, and rewriting tests (someday) to be more unit tests than integration tests.  I think skipping tests in the normal testing process should be added to the list. There's not a lot lost to break a test and not know about it until 10 - 15 minutes later in TravisCI, but before the merge, especially since PR review is normally measured in days and weeks, not minutes and hours.

Also, I'd rather we didn't have to jump through so many hoops to test our code in TravisCI. It really stinks. It's part of the medium-term plan.

Next question: Where do we test Docker dev environment changes?

The first answer is that every dev tests it when the check out the project and run it. Something's broken, one of the 5 people working on the project will notice right away.  This seems to be the right level of detection for the problem of broken configs.

There's been a few problems with the "dev test" strategy lately. jgmize found one with an [old copy of the MySQL image](https://github.com/mozilla/kuma/pull/3975#discussion_r79291400).  It's not clear how to reset a local dev environment, and we'll get into the problem of the environment being broken for long-term devs and not fresh devs (new migrations, new dependencies, updated images, etc.), and things being broken for fresh devs but not new (in Vagrant land, node provisioning in PR #3963).

TravisCI testing of the environment helps with "environment is broken for fresh devs", because TravisCI tests fresh and breaks the "works for me" argument. It does nothing to help the "broken for existing devs" problem, but that's (development) life.

So, I'd like some automated testing of the Docker environment. I'd like automatic running of the selenium integration tests as well. In both cases, I'm not sure TravisCI is the right place to run them, or when they should be run (each pull request? each merge to master?).

I don't see automated testing of the Docker dev environment as a blocker for the Dockerized development environment task for Q3. There is no good stable Docker configuration that needs to be validated across merges. It is broken and incomplete and changes daily.  I don't plan to merge this before October, and then we may pick a different option (CircleCI? Jenkins?) for exercising the docker config.

I am leaving it open, @safwanrahman, in case you want to continue rebasing and improving this as we continue refining the Docker environment.  It may help the argument if there is something this would have caught that a PR or other process did not catch.

jgmize:
see WIP #3981 for current progress on running the full suite of tests in Jenkins. Note that my intention is to continue using Travis for PRs, and despite the green check mark on this PR, the latest travis docker build failed: https://travis-ci.org/mozilla/kuma/jobs/160802190

safwanrahman:
@jgmize As I mentioned Earlier, for passing test #3975 #3976 need to get merged. As both of them got merged, I have rebased upon master and it seems now the [docker test works](https://travis-ci.org/mozilla/kuma/jobs/162443246)

safwanrahman:
@jwhitlock I have found some regression that, the mysql container needs much time to start up working. That caused Error connecting database. In the past, it did not have this problem. So Today or tomorrow, we need to run the test with docker in CI, so better we start now as its running successfully. So we can track if anytime it got broken while we are experimenting with docker.

safwanrahman:
@jwhitlock I think now we can review this and merge. What do you think?

jwhitlock:
I'll review late this week or early next week.

It looks like ``docker`` and ``docker-compose`` need to be added to [whitelist_externals](http://tox.readthedocs.io/en/latest/config.html#confval-whitelist_externals=MULTI-LINE-LIST).

.travis.yml
```diff
@@ -10,14 +10,15 @@ language: python
 python:
     - "2.7"
 services:
+    - docker
     - memcached
-    - elasticsearch
```
jwhitlock: Why drop the TravisCI elasticsearch service?

scripts/travis-install
```diff
@@ -15,10 +15,20 @@ then
     sudo apt-get autoremove
     sudo apt-get autoclean
     sudo rm -rf /var/lib/mysql
-    sudo truncate -s 0 /var/log/mysql/error.log
 
+    # Install Elasticsearch
+    wget https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.2.deb
+    sudo dpkg -i elasticsearch-1.7.2.deb
+
```
jwhitlock: I don't think it is necessary to install a new ElasticSearch service, instead of using the one provided by TravisCI.  This would only be used by the existing py27 tests, and they were happy with the TravisCI version.

scripts/travis-install
```diff
@@ -15,10 +15,20 @@ then
     sudo apt-get autoremove
     sudo apt-get autoclean
     sudo rm -rf /var/lib/mysql
-    sudo truncate -s 0 /var/log/mysql/error.log
```
jwhitlock: What's the reason for removing this line?

tox.ini
```diff
@@ -31,6 +31,16 @@ basepython = python2.7
 deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
 commands = dennis-cmd lint --errorsonly locale/
 
+[testenv:docker]
+commands = 
+    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml up -d
+    docker exec -i kuma_web_1 make compilejsi18n collectstatic
+    # TODO: mysql needs some time to set up. find a way to eliminate the waiting time
+    sleep 60
```
jwhitlock: There are several ways to do this, but I'm not sure what is available in TravisCI. ``sleep 60`` is ok for v1, we can iterate.

scripts/docker-compose.travis.yml
```diff
@@ -0,0 +1,17 @@
+version: '2'
+services:
+  worker: &worker
+    build:
+      context: .
+      dockerfile: Dockerfile-base
```
jwhitlock: The shorter way to say this is ``build: Dockerfile-base``.

scripts/travis-install
```diff
@@ -15,10 +15,20 @@ then
...
     echo "ES_HEAP_SIZE=256m" | sudo tee --append /etc/default/elasticsearch
     sudo service elasticsearch start;
 fi
+
+if [ "$TOXENV" == "docker" ]
+then
+    curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+    chmod +x docker-compose
+    sudo mv docker-compose /usr/local/bin
+fi
```
jwhitlock: I guess ``docker-compose`` v1.4.2 is too old. 😞 Version 2 was added in 1.6.0, requiring at least Docker engine 1.10.0

tox.ini
```diff
@@ -31,6 +31,16 @@ basepython = python2.7
 deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
 commands = dennis-cmd lint --errorsonly locale/
 
+[testenv:docker]
+commands =
```
jwhitlock: Add a ``whitelist_externals`` line to avoid the tox warnings:

```
whitelist_externals = docker-compose docker sleep
```

tox.ini
jwhitlock: Here's one solution, for example - http://brunorocha.org/python/dealing-with-linked-containers-dependency-in-docker-compose.html

scripts/travis-install
safwanrahman: Because, there are no ``/var/log/mysql/error.log`` file in ubuntu 14.04 trusty

scripts/travis-install
safwanrahman: As I mentioned earlier with the first comment, according to the [doc](https://docs.travis-ci.com/user/trusty-ci-environment/#Data-Stores), elasticsearch is not available in ubuntu 14.04 trusty builds. so need to install the elasticsearch.

scripts/travis-install
safwanrahman: we replace the default docker-compose of travis with ``docker-compose 1.8`` according to the [environment variable](https://github.com/mozilla/kuma/pull/3977/files#diff-354f30a63fb0907d4ad57269548329e3R21)

PR #3979, opened 2016-09-20
Bug 1302459: Load LOCALES from cached languages.json, deprecate product-details

jwhitlock:
This replaces PR #3973, which may have been a step too far at this time.

The custom setting ``LOCALES`` is generated by loading ``languages.json`` at startup time, and ``LANGUAGES`` and other derived settings are generated from ``LOCALES``.

Previously. ``LOCALES`` was derived from ``languages.json`` as downloaded by [Mozilla product details](https://product-details.mozilla.org/1.0/).  This particular data changes rarely. Since May 2015, these changes were made upstream:

* Added locales for Bodo (brx), Kaqchikel (cak), Guarani (gn), Konkani (kok), Kashmiri (ks), Latgalian (ltg), and Purépecha (tsz)
* Updated the English name for the locale with code 'or' from Oriya to Odia
* Updated the native name for Romanian (ro) from română to Română

Only this last change affected MDN.

Changing to this new method:

1. Removes a step from installation and deployment across all environments (the initial pain point for starting this PR)
2. Removes a 3rd-party dependency


kuma/core/apps.py
```diff
@@ -17,6 +17,5 @@ def language_mapping(self):
         """
         a static mapping of lower case language names and their native names
         """
-        from product_details import product_details
-        return dict([(lang.lower(), product_details.languages[lang]['native'])
+        return dict([(lang.lower(), settings.LOCALES[lang].native)
```
jgmize: Would you mind switching this to the dict literal syntax for both readability and [performance](https://doughellmann.com/blog/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/) improvements?

jgmize:
lgtm, nice work @jwhitlock

kuma/core/apps.py
jwhitlock: Micro-optimization for WTF code, but doing it anyway because can't un-see.

jgmize:
needs rebase

jwhitlock:
Force push to rebase on master

PR #3980, opened 2016-09-21
Bug 1218563: Enable page rendering in Docker

jwhitlock:
* Update KumaScript to current master, which includes setting the configuration from the command like.
* Add a new api service to handle KumaScript requests, allowing the docker-compose v1 linking to work correctly
* Allow environment override of the setting ``KUMASCRIPT_URL_TEMPLATE``
* Configure kumascript service for rendering pages and caching responses

r? @jgmize

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ worker: &worker
     - mysql
     - elasticsearch
     - redis
+    - kumascript
   environment:
     # Django settings overrides:
     - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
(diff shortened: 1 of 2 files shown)


jwhitlock:
My testing procedure:

1. Restart the docker environment with the new settings
2. Make sure [kumascript is enabled](https://kuma.readthedocs.io/en/latest/installation-docker.html#enable-kumascript), including loading the auto-loaded modules
3. Make a test page with lorem ipsum text followed by an ``h2`` section titled "The End" (I used http://localhost:8000/en-US/docs/User:Admin)
4. Copy https://developer.mozilla.org/en-US/docs/Template:anch?raw from production, create http://localhost:8000/en-US/docs/Template:anch
5. Edit the test page and add a KumaScript macro near the top: ``{{ anch("The_End", "Skip to the end") }}``

There's a bug where KumaScript will render the n-1 version of a page.  It is not introduced by this environment.  It seems more reproducible in the docker environment, which may mean we're closer to fixing it.  If you trigger the n-1 bug, then force-refresh to regenerate the page.

PR #3981, opened 2016-09-22
Jenkins pipeline

jgmize:
WIP. I have this branch running on https://ci.us-west.moz.works/view/MDN/job/mdn_pipeline/ and as of https://ci.us-west.moz.works/view/MDN/job/mdn_pipeline/23/console all existing tests but one are passing, and that one is because I've added the --nomigrations flag to py.test by default when running tests with docker-compose. I have xfailed that test for now, but given that this feature is currently broken in production I'm questioning the value of keeping that test at all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,15 @@ REGISTRY ?= quay.io/
 IMAGE_PREFIX ?= mozmar
 BASE_IMAGE ?= ${REGISTRY}${IMAGE_PREFIX}/${BASE_IMAGE_NAME}\:${VERSION}
 BASE_IMAGE_LATEST ?= ${REGISTRY}${IMAGE_PREFIX}/${BASE_IMAGE_NAME}\:latest
+IMAGE ?= $(BASE_IMAGE_LATEST)
 KUMA_IMAGE ?= ${REGISTRY}${IMAGE_PREFIX}/${KUMA_IMAGE_NAME}\:${VERSION}
 KUMA_IMAGE_LATEST ?= ${REGISTRY}${IMAGE_PREFIX}/${KUMA_IMAGE_NAME}\:latest
+TEST ?= test #other options in docker-compose.test.yml
+DEIS_PROFILE ?= dev-usw
+DEIS_APP ?= mdn-dev
+DEIS_BIN ?= deis
+WORKERS ?= 1
+DB_PASS ?= kuma # default for ephemeral demo DBs
 
 target = kuma
 requirements = -r requirements/local.txt
@@ -109,6 +116,25 @@ push-kuma:
 
 push: push-base push-kuma
 
+deis-create:
+	DEIS_PROFILE=${DEIS_PROFILE} ${DEIS_BIN} create ${DEIS_APP} --no-remote && \
+	sleep 5 && ${DEIS_BIN} config:push -p .env-dist -a ${DEIS_APP} || \
+	${DEIS_BIN} apps | grep -q ${DEIS_APP}
+
+deis-pull:
+	DEIS_PROFILE=${DEIS_PROFILE} ${DEIS_BIN} pull ${KUMA_IMAGE} -a ${DEIS_APP}
+
+deis-scale-worker:
+	DEIS_PROFILE=${DEIS_PROFILE} ${DEIS_BIN} ps:scale worker=${WORKERS} -a ${DEIS_APP}
+
+k8s-migrate:
+	kubectl --namespace ${DEIS_APP} exec \
+	$(shell kubectl --namespace ${DEIS_APP} get pods | grep ${DEIS_APP}-cmd | awk '{print $$1}') \
+	python manage.py migrate
+
+deis-migrate:
+	DEIS_PROFILE=${DEIS_PROFILE} ${DEIS_BIN} run -a ${DEIS_APP} python manage.py migrate
+
 tag-latest:
 	docker tag ${BASE_IMAGE} ${BASE_IMAGE_LATEST}
 	docker tag ${KUMA_IMAGE} ${KUMA_IMAGE_LATEST}
@@ -125,6 +151,10 @@ bash: up
 
 shell_plus: up
 	docker exec -it kuma_web_1 ./manage.py shell_plus
+
+compose-test:
+	docker-compose -f docker-compose.yml -f docker-compose.test.yml run $(TEST)
+	docker-compose -f docker-compose.yml -f docker-compose.test.yml stop
 
 # Those tasks don't have file targets
 .PHONY: test coveragetest intern locust clean locale install compilecss compilejsi18n collectstatic localetest localeextract localecompile localerefresh
--- a/kuma/wiki/tests/test_templates.py
+++ b/kuma/wiki/tests/test_templates.py
@@ -680,6 +680,7 @@ class DocumentListTests(UserTestCase, WikiTestCase):
         doc = pq(response.content)
         eq_(1, len(doc('#document-list ul.document-list li')))
 
+    @pytest.mark.xfail(reason='Requires migrations')
     @pytest.mark.tags
     def test_tag_list_duplicates(self):
         """
--- a/.env-dist
+++ b/.env-dist
@@ -1,0 +1,12 @@
+BROKER_URL=redis://redis:6379/0
+CELERY_ALWAYS_EAGER=False
+CSRF_COOKIE_SECURE=False
+DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
+DEBUG=False
+ES_URLS=elasticsearch:9200
+KUMASCRIPT_URL_TEMPLATE=http://kumascript:9080/docs/{path}
+MEMCACHE_SERVERS=memcached:11211
+PROTOCOL=http://
+SESSION_COOKIE_SECURE=False
+ALLOWED_HOSTS=*
+C_FORCE_ROOT=1
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,56 @@
+node {
+    stage 'git'
+    checkout scm
+
+    env.DEIS_BIN = 'deis2'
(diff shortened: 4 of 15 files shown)


jgmize:
I've been triggering the builds manually b/c the github webhooks aren't triggering the job. Appears to be known issue: https://issues.jenkins-ci.org/browse/JENKINS-35132

Should be able to work this by setting up a 'freestyle' job to fire on the webhook that triggers the pipeline job.

kuma/wiki/tests/test_templates.py
```diff
@@ -680,6 +680,7 @@ def test_tag_list(self):
         doc = pq(response.content)
         eq_(1, len(doc('#document-list ul.document-list li')))
 
+    @pytest.mark.xfail(reason='Requires migrations')
```
jwhitlock: I had to test this myself.  The test passes if you call tests normally, like ``py.test kuma/wiki/tests/test_templates.py::DocumentListTests``, but fails with ``--nomigrations``, with the error:

```
  File "/app/kuma/wiki/tests/test_templates.py", line 693, in test_tag_list_duplicates
    fr_tag.save()
  File "/usr/local/lib/python2.7/dist-packages/taggit/models.py", line 89, in save
    return super(TagBase, self).save(*args, **kwargs)
.. OMIT SOME LINES ... 
  File "/usr/local/lib/python2.7/dist-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorvalue
IntegrityError: (1062, "Duplicate entry 'CSS R\xc3\xa9f\xc3\xa9rence' for key 'name'")
```
This means that "CSS Reference" is counted as different than "CSS Référence" with migrations, but as the same string without migrations.  This means the custom collation is only being applied when migrations are skipped and the tables are recreated directly. This happens both in Docker and Vagrant.

The test mentions http://bugzil.la/871638, an ISE on the wiki tag list page, fixed before the custom collation.  I'm more interested in [bug 1271552](https://bugzilla.mozilla.org/show_bug.cgi?id=1271552), where duplicate tags appear spontaneously, and [bug 776048](https://bugzilla.mozilla.org/show_bug.cgi?id=776048), where the custom collation was added.

As for marking it ``xfail``, it may be possible to make it conditionally skipped, with ``@pytest.mark.skipif(config.getvalue('migrations'))``, but I don't know what the config actually is, and ``xfail`` is fine with just 4 of them, but I'm a broken person and had to mention it.

jgmize: I'm confused as to how we're just now noticing this problem. Shouldn't this test be failing when run against the DBs in SCL3 where the custom collation was applied?

jwhitlock: Test don't run in SCL3, they run in TravisCI, Vagrant, and now Docker.  But I'm not sure if it matters - there are different code paths for running tests with and without migrations, and this is the first time they have ever been run without migrations, which appears to make a difference for this table definition.  That's why it hasn't been discovered before.

jwhitlock: I inspected the database in the test by adding a ``pdb`` breakpoint and running this code, in normal test mode and with ``--nomigrations``:

```
from django.db import connection

cursor = connection.cursor()
cursor.execute("SHOW CREATE TABLE wiki_documenttag")
table_name, create_table = cursor.fetchone()
cursor.execute("SELECT @@character_set_database, @@collation_database")
charset, collation = cursor.fetchone()
cursor.close()

print(table_name)
print(create_table)
print(charset)
print(collation)
```

In both cases, the table was ``wiki_documenttag``, the default character set was ``utf8``, and the default collation was ``utf8_general_ci``, not the custom collation of ``utf8_distinct_ci``.

The create statement for a normal run is:
```
CREATE TABLE `wiki_documenttag` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(100) CHARACTER SET utf8 COLLATE utf8_distinct_ci DEFAULT NULL,
  `slug` varchar(100) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `slug` (`slug`),
  UNIQUE KEY `name` (`name`)
) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8
```

The create table for a ``--nomigrations`` run is:
```
CREATE TABLE `wiki_documenttag` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(100) NOT NULL,
  `slug` varchar(100) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `name` (`name`),
  UNIQUE KEY `slug` (`slug`)
) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=utf8
```

The only thing they have in common is the ``varchar(100)``:
```
  `name` varchar(100) CHARACTER SET utf8 COLLATE utf8_distinct_ci DEFAULT NULL,
  `name` varchar(100) NOT NULL,
```

In docker and production, the default character set is ``utf8`` and the collation ``utf8_general_ci``.

In Docker, the the create table is the same as the normal test run:
```
CREATE TABLE `wiki_documenttag` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(100) CHARACTER SET utf8 COLLATE utf8_distinct_ci DEFAULT NULL,
  `slug` varchar(100) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `slug` (`slug`),
  UNIQUE KEY `name` (`name`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

In production, the create table is similar, but is missing the unique key:
```
CREATE TABLE `wiki_documenttag` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(100) CHARACTER SET utf8 COLLATE utf8_distinct_ci DEFAULT NULL,
  `slug` varchar(100) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `slug` (`slug`)
) ENGINE=InnoDB AUTO_INCREMENT=646911 DEFAULT CHARSET=utf8
```

So, ``--nomigrations`` is the odd one out.  Searching for bugs in Django, py.test...

jwhitlock: OK, I'm less confused about collations.  ``utf8_general_ci`` (the database default) treats a bunch of characters as the same collation point, so "A" is equivalent to "a" and "À" and "ȃ" and "Ȧ".  In ``utf8_general_ci``, "CSS Reference" is sorted the same and collides with "CSS Référence", causing the test to fail with that collation, as done by `--nomigrations``. There's a helpful [chart](http://collation-charts.org/mysql60/mysql604.utf8_general_ci.european.html) that shows equivalent characters in that collation.

``utf8_distinct_ci`` is well documented in [this blog post](https://mariadb.com/blog/adding-case-insensitive-distinct-unicode-collation), which I've skimmed several times, but read carefully today. The goal is to retain case insensitive behavior ("A" == "a"), but to treat "accented" letters as distinct ("e" != "é"), but also with some case insensitivity ("À" == "à").  So "distinct" but "case insensitive" ("ci"), as opposed to the mush of "general". Names mean something.

I looked for a bit for a bug in ``--nomigrations``, but I think this is working as intended. It tricks the tests that the app has no migrations by overriding [MIGRATION_MODULES](https://docs.djangoproject.com/en/1.10/ref/settings/#migration-modules) to always return ``None``.  The custom collation is applied in [wiki migration 0002](https://github.com/mozilla/kuma/blob/master/kuma/wiki/migrations/0002_auto_20150430_0805.py) as manual SQL, and according to [Stack Overflow](http://stackoverflow.com/questions/1198486/how-to-set-the-encoding-for-the-tables-char-columns-in-django), the migration is the way to do it.

So:

1. I was confused about the collations, and now I know what's up
2. This isn't a bug in Django, or pytest-django, or taggit
3. Django doesn't support per-column custom collations, and the migration is a decent way to apply it,
4. The test will break because it depends on the custom collation on the column, which is enforced by the migration, rather than the default collation, that is enforced in the settings.
5. I've done a lot of research to just get back to, "yep ``xfail`` is probably right".
 

I have corrected a mistaken idea about our collations

jezdez: @jwhitlock 👍 on reverse engineering the custom collation 🎆

jgmize:
Ok, I've got this running successfully as a multibranch pipeline: https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/
Once this is merged, I'll delete the redundant, non-pipeline jenkins jobs.
As you'll see from the Jenkinsfile, on merge to master it runs the full suite of unit tests via docker-compose before deploying to dev, which has persistent storage for the db & es. For any other branch, the default is to build, configure, and deploy a new Deis app based on the branch name with an 'mdn-' prefix, and to use kubectl to add some dedicated ephemeral backing services (mysql, redis, elasticsearch, memcached, & kumascript). Currently we create the database schema but without any data so home page loads but not much else-- the plan is to follow up with an "MVDB" (minimum viable db) for dev/test purposes in a later PR. You can see a couple of example deployments at http://mdn-jenkins-pipeline.dev.us-west.moz.works and http://mdn-demo-deploy.dev.us-west.moz.works and you can watch this in action yourself by simply creating a new branch based on this one and pushing it to the origin-- within a minute you should see a new pipeline job running under your branch, and once it completes successfully you should be able to see the homepage on the deployed instance.

jgmize:
@jwhitlock r?

Makefile
```diff
@@ -109,6 +116,25 @@ push-kuma:
 
 push: push-base push-kuma
 
+deis-create:
+	DEIS_PROFILE=${DEIS_PROFILE} ${DEIS_BIN} create ${DEIS_APP} --no-remote && \
+	sleep 5 && ${DEIS_BIN} config:push -p .env-dist -a ${DEIS_APP} || \
+	${DEIS_BIN} apps | grep -q ${DEIS_APP}
+
+ deis-pull:
```
jwhitlock: This extra leading space seems to work, but looks broken.

PR #3982, opened 2016-09-23
bug 1218563: Upgrade to docker-compose v2 format

jwhitlock:
Switch ``docker-compose.yml`` from version 1 to version 2 format, picking up a few features:

* Version 2 constructs a network that all the containers share, so ``links`` is unneeded for communication. Switch from ``links`` to ``depends_on``, to still show dependencies between services.
* Define a named volume ``mysqlvolume`` to hold the database data, removing the data container ``mysql-data``.  This is now recommended by the [Docker community](http://stackoverflow.com/a/20652410/10612), even though the Docker docs haven't caught up yet.

You'll have to recreate ``mysql-data`` container to make this work.  I suggest:

```
docker-compose up  # Gives warnings, uses old volume
docker exec kuma_mysql_1 sh -c 'exec mysqldump --all-database -ukuma -pkuma' > dump.sql  # Save the data
docker-compose stop
docker rm kuma_mysql_1
docker-compose up --remove-orphans  # Removes mysql-data container, rebuilds mysql container on volume
docker exec -i kuma_mysql_1 sh -c 'exec mysql -ukuma -pkuma' < dump.sql  # restore the database
docker inspect kuma_mysql_1  # Admire the new named volume usage
docker volume ls -f dangling=true  # Despair at all the anonymous unused volumes created
```

Because indenting changes, the first commit looks bigger than it actually is. Sorry, reviewers. @jgmize ?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     command: ./manage.py celery worker --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
     volumes:
       - ./:/app
-    links:
+    depends_on:
       - memcached
       - mysql
       - elasticsearch
@@ -46,7 +46,7 @@ services:
   api:
     <<: *worker
     command: gunicorn -w 4 --bind 0.0.0.0:8000 --timeout=120 kuma.wsgi:application
-    links:
+    depends_on:
       - memcached
       - mysql
       - elasticsearch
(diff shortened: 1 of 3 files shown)


jwhitlock:
Force-push to include removing ``docker/images/mysql-data/Dockerfile``.

docker-compose.yml
```diff
@@ -1,116 +1,114 @@
...
+  memcached:
+    image: memcached
 
-# TODO: determine if this data container is actually useful
-mysql-data:
-  build: ./docker/images/mysql-data
-  volumes:
-    - /var/lib/mysql
+  mysql:
+    build: ./docker/images/mysql
```
jgmize: No, you only need to specify the filename if you are using something *other* than `Dockerfile`.

jgmize:
Nice work @jwhitlock.

PR #3983, opened 2016-09-23
Fix Bug 1305156: Code Samples are LTR

stephaniehobson:
Small code change to force LTR in code samples. Problem is most apparent in webkit browsers.

Useful testing pages:
https://developer.mozilla.org/en-US/docs/User:stephaniehobson:code
https://developer.mozilla.org/ar/docs/User:stephaniehobson:code (not actually translated but appears with RTL UI and styles.)

### DIFF
--- a/kuma/static/styles/components/syntax/base.styl
+++ b/kuma/static/styles/components/syntax/base.styl
@@ -30,7 +30,7 @@ pre[class*='language-'] {
 }
 
 /*
-Enhamcements!
+Enhancements!
 ====================================================================== */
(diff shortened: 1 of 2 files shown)


kuma/static/styles/components/syntax/base.styl
```diff
@@ -30,7 +30,7 @@ pre[class*='language-'] {
 }
 
 /*
-Enhamcements!
+Enhancements!
```
jwhitlock: LOL

PR #3984, opened 2016-09-23
Fix Bug 1305178: Hide empty paragraphs

stephaniehobson:
It's a hack but it's a very targeted hack.


stephaniehobson:
This is ready to merge but I'm going to leave it until Tuesday so I can be around when it is pushed.

PR #3989, opened 2016-09-26
bug 1218563: Add docker-build.sh for CKEditor

jgmize:


docs/ckeditor.rst
```diff
@@ -9,11 +9,15 @@ document details how to update CKEditor within the MDN codebase.
 
 Building CKEditor
 -----------------
-To rebuild CKEditor::
+To rebuild CKEditor (requires java)::
```
jgmize: good point, will update.

jgmize:
@safwanrahman I updated the docs to clarify the original instructions required either vagrant or java. Would you mind testing the docker instructions & script locally?

@jwhitlock r?

safwanrahman:
So there may be some problem removing the files. I get this error after building
```
==========================
Release process completed:

    Number of files: 281
    Total size.....: 3809517 bytes
    Time taken.....: 43.222seconds

Removing added plugins...
fatal: Not a git repository (or any parent up to mount point /ckeditor)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

There must be something need to fix in the script

jwhitlock:
I saw the same problem as @safwanrahman. The lines causing the problem are in [kuma/static/js/libs/ckeditor/source/build.sh](https://github.com/jgmize/kuma/blob/bc92a1b0b6c735a0dae5eb08b23a1df4d539181b/kuma/static/js/libs/ckeditor/source/build.sh#L115-L118):

```
echo "Removing added plugins..."
cd ckeditor
git clean -d -f -f
cd ..
```
This requires that the kuma ``.git`` folder is inside the container.  The result of this command is to wipe out the folder ``kuma/static/js/libs/ckeditor/source/ckeditor``, which can also be done with:

```
echo "Removing added plugins..."
rm -rf ckeditor/
```
This mirrors [the same command earlier in the script](https://github.com/jgmize/kuma/blob/bc92a1b0b6c735a0dae5eb08b23a1df4d539181b/kuma/static/js/libs/ckeditor/source/build.sh#L29), and will work in all the environments.

jgmize:
Updated to include the @jwhitlock's suggested fix to the problem he and @safwanrahman saw, and modified the script to create and use a user inside the container with a username, UID,  and GID matching the host user running the script so that the files created by the script have the same owner.

jwhitlock:
Verified instructions on host w/ java and with ``docker-build.sh``. Good work on creating the local user. I think this problem doesn't reproduce on MacOS, but it still works here as well.

PR #3991, opened 2016-09-26
bug 1218563: Update documentation doc for Docker

jwhitlock:
Re-write the documentation for updating the documentation to use Docker. A virtualenv in the host machine is still the fastest way to generate documentation, but a method is given for using Docker, which may more practical in the future.


safwanrahman:
So I get this Error while running the build command
``/app/sphinx-theme should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+``

jwhitlock:
@safwanrahman you must clone ``sphinx-theme`` first, such as: ```git clone git@github.com:mdn/sphinx-theme.git```, so that ``sphinx-theme`` appears in the root of your kuma repository.

docs/documentation.rst
```diff
@@ -1,46 +1,119 @@
...
     cd docs
     make html
+    python -m webbrowser file://${PWD}/_build/html/index.html
+
 
-This will generate the documents with the index at
-``docs/_build/html/index.html``.
+To generate the documentation with Docker::
 
+    docker-compose run --rm web sh -c "\
```
safwanrahman: This generate the files with root privilege. We can run something like this.
```
docker-compose run --rm --user $(id -u) web sh -c "\
  virtualenv .venvs/docs && \
  . .venvs/docs/bin/activate && \
  pip install -r /app/requirements/docs.txt && \
  cd docs && \
  make html"
```

jwhitlock: This creates the virtualenv inside my kuma working directory, which causes more problems (clean-up, etc.)I can't create the virtualenv in the root folder, because I'm not root anymore.

This works for me:
```
docker-compose run --rm --user $(id -u) web sh -c "\
  virtualenv /tmp/.venvs/docs && \
  . /tmp/.venvs/docs/bin/activate && \
  pip install -r /app/requirements/docs.txt && \
  pip install -e /app/sphinx-theme && \
  cd /app/docs && \
  make html"
```

Please give that a try and report back.

safwanrahman: Yap. It should work and it worked

jwhitlock:
Noted the Linux file permission issue in the troubleshooting doc, updated the one-liners to something that works on Ubuntu and MacOS host machines.  Ready for review again.

docs/documentation.rst
```diff
@@ -1,46 +1,119 @@
...
+Regenerate the MDN Sphinx Theme template
+----------------------------------------
+When regenerating the template, the Kuma version should be one deployed to
+production.  The assets are loaded from the MDN CDN, and if a non-deployed
+Kuma commit is used to generate the template, files will be unavailable, which
+will may or may not be obvious when you view the documentation locally.
 
+The MDN Sphinx theme template can be regenerated by Kuma::
+
+    docker-compose run --rm -T \
```
safwanrahman: ``--user`` parameters should be added here also

jwhitlock: What is the error when you run it as documented?

safwanrahman: no error. same problem about root privillage

safwanrahman: Sorry. My bad. It does not do it.

jwhitlock: OK, that's what I expected. It doesn't create the file in the container - it outputs the template, and then it gets redirected to a file as the host user.

jwhitlock:
Feel free to use ``bash`` and the commands given in the one-liner.  I think experienced docker devs could see clearly how to do it given the information on the page.  However, I doubt anyone but me is building docs locally, and I do it in a host-based ``virtualenv``.  I don't think improving the instructions will compel more people to follow them - there's just no need.

jgmize:
Thanks @jwhitlock, and thanks for all your testing and feedback @safwanrahman. As I mentioned in person while I think there is some room for improvement to simplify this a bit, The steps work as-is (with previously discussed caveats) and we can iterate with further improvements later.

PR #3992, opened 2016-09-27
bug 1218563: Document building/updating base img

jgmize:
@jwhitlock r?


jgmize:
updated with instructions to restart the containers to start using the newly built base image. Note that I explicitly documented `docker-compose stop` and `docker-compose up -d` as separate steps because I have seen issues with `docker-compose restart` and do not recommend it.

jwhitlock:
Thanks @jgmize this is much better than the method I was using to use a locally-built image.

PR #3997, opened 2016-09-28
[Bug 1218563] Adding a user without root privillage in docker

safwanrahman:
It will add a without root privilege user named ``kumadev`` and adding ``USER`` parameters in docker file so all the shell access will be done with ``kumadev`` user. As per mentioned in [doc](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/user)

If anyone needs root access to the container, they can add ``--user root`` parameter to either ``docker exec`` or ``docker-compose run``.
like
``docker-compose run --rm --user root web bash``
@jwhitlock r?

### DIFF
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -41,5 +41,8 @@ ENV PIPELINE_CSS_COMPRESSOR=pipeline.compressors.cssmin.CSSMinCompressor \
 COPY ./requirements /app/requirements
 RUN pip install --no-cache-dir -r requirements/dev.txt
 
+RUN useradd kuma
+USER kuma
+
 ENV WEB_CONCURRENCY=4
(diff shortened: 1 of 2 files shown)


jwhitlock:
I don't have a Linux environment (although I guess I can borrow one from AWS). @jgmize is in a better position to review this.

jgmize:
This doesn't address matching the UID of the host user, so I believe filesystem permissions issues on the shared volume may still be an issue. However, that could be addressed in a separate PR later. Additionally, I believe https://github.com/mozilla/kuma/blob/master/docker-compose.yml#L34 can be removed together with this change.

safwanrahman:
Fixed, rebased and force pushed.
r? @jgmize

jgmize:
Tested locally, worked as expected.

PR #4000, opened 2016-10-01
[Bug 1218563] Update ihavepower command to take optional password

safwanrahman:
It now takes a optional password with the management command. @jwhitlock r?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -160,16 +160,12 @@ If you want to create a new admin account, use ``createsuperuser``::
 This will prompt you for a username, email address (a fake address like
 ``admin@example.com`` will work), and a password.
 
-If your database has an existing account that you want to use, use the Django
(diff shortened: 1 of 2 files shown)


jwhitlock:
The code looks good. I'm creating and loading an anonymized database to try it out.

jwhitlock:
Fails for me with ```CommandError: User already has the power!```, without settings the password.

safwanrahman:
> Fails for me with CommandError: User already has the power!, without setting the password.

It should fail for you cause you already have superuser access.

safwanrahman:
As per IRC conversation @jbradberry mentioned that, we can use ``./manage.py changepassword <username>`` to create a password

safwanrahman:
@jwhitlock Fixed

jwhitlock:
Thanks @safwanrahman!

PR #4002, opened 2016-10-03
fix bug 1307121: Add placeholder to Czech (cs) translation 

jwhitlock:
Add the placeholder string to the Czech (cs) translation of the search string.

The Czech translation team is active, so I want to give them some time to fix this the "right" way (via [Pontoon](https://pontoon.mozilla.org/cs/mdn/LC_MESSAGES/django.po/?search=Search+again&string=67765)), and to see if this needs a fix in Pontoon itself.  I'm opening the PR to avoid someone else opening a PR, hopefully I can close without merging in a bit.


jwhitlock:
As seen in [bug 1307121](https://bugzilla.mozilla.org/show_bug.cgi?id=1307121#c1), fixed upstream in Pontoon, will be pushed to master on next sync.

PR #4004, opened 2016-10-06
Bug 1308300 - Add a "comment" field above the diff when updating a page 

MatonAnthony:
https://bugzilla.mozilla.org/show_bug.cgi?id=1308300

### DIFF
--- a/kuma/wiki/jinja2/wiki/includes/document_macros.html
+++ b/kuma/wiki/jinja2/wiki/includes/document_macros.html
@@ -77,6 +77,12 @@
         <dd class="rev-to"> {{ ', '.join(revision_to.tags.split(',')) }}</dd>
         {% endif %}
 
+        {% if revision_from.comment or revision_to.comment %}
+        <dt>{{ _('Comment:') }}</dt>
+        <dd class="rev-from"> {{ revision_from.comment }}</dd>
+        <dd class="rev-to"> {{ revision_to.comment }}</dd>
+        {% endif %}
+
         <dt>{{ _('Content:') }}</dt>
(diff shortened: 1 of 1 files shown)


kuma/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -77,6 +77,12 @@
         <dd class="rev-to"> {{ ', '.join(revision_to.tags.split(',')) }}</dd>
         {% endif %}
 
+        {% if revision_from.comment or revision_to.comment %}
+        <dt>{{ _('Comment:') }}</dt>
```
safwanrahman: This should be in proper indent

MatonAnthony: I followed the indent used at the previous block of code.
If @stephaniehobson or @jwhitlock wants this indented in an other way, it would implie to reindent the rest of the file, I suppose.

jwhitlock: @stephaniehobson owns the template formats, deferring to her.

stephaniehobson: This document has some really creative indenting.

The way @MatonAnthony has indented is consistent with what's around it and aids understanding the code. It's good.

stephaniehobson:
Renders properly locally. I don't totally understand what I saw in the `rev-from` field when I tested it but that's partially because it wasn't in English. If there is a problem (and I don't think there is) I trust our localization community will open a follow up bug.

stephaniehobson:
Also, thank you for the quick turn around time on this :+1:

PR #4006, opened 2016-10-07
bug 1218563: Add .dockerignore

jwhitlock:
Add a .dockerignore that includes:
- Everything in .gitignore
- Everything in kumascript/.gitignore
- .git folders and others not needed in containers

This reduces my build context from 2 GB to about 50 MB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,4 @@
 # From kuma .gitignore
-!static/fonts/.htaccess
-!build/assets/css/README.md
 *.pyc
 *.pyo
 *.sw?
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -0,0 +1,74 @@
+# From kuma .gitignore
+!static/fonts/.htaccess
+!build/assets/css/README.md
```
jgmize: I don't think these are necessary to include in the docker image.

jwhitlock: My reasoning was you could do a quick check: ``diff .gitignore .dockerignore``.  But that falls apart for the kumascript ``.gitignore``. And maybe we want the ``.htaccess`` file someday. Maybe.

OK, removing them.

PR #4007, opened 2016-10-07
bug 1218563: Improve TravisCI docker build and test

jwhitlock:
Improve on PR #3977 (test docker travis):

- Use environment variables to simplify ``docker-compose`` calls
- Eliminate some warnings by adding build steps, configuring ``tox``
- Start the ``mysql`` container while the ``web`` image is building, to eliminate ``sleep 60``
- _Formerly included "build and test the base image", but PR #3977 already does that_

See #4011 for an alternate implementation with less commands.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -32,8 +32,10 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
 commands = dennis-cmd lint --errorsonly locale/
 
 [testenv:docker]
-commands = 
-    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml up -d
+setenv =
+    COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
+commands =
+    docker-compose up -d
     docker exec -i kuma_web_1 make compilejsi18n collectstatic
     # TODO: mysql needs some time to set up. find a way to eliminate the waiting time
     sleep 60
--- a/tox.ini
+++ b/tox.ini
@@ -32,18 +32,15 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
 commands = dennis-cmd lint --errorsonly locale/
 
 [testenv:docker]
-whitelist_externals =
-    docker-compose
-    sleep
+whitelist_externals = docker-compose
 setenv =
     COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
     COMPOSE_PROJECT_NAME = kumatox
 commands =
-    docker-compose up -d
+    docker-compose up -d --build mysql  # Build, init DB (can be slow)
+    docker-compose build web    # Rebuild w/ new reqs, Dockerfile-base
+    docker-compose up -d        # Startup remainin services
     docker-compose exec -T web make compilejsi18n collectstatic
-    # TODO: mysql needs some time to set up. find a way to eliminate the waiting time
-    sleep 60
-    docker-compose logs mysql
     docker-compose exec web ./manage.py migrate
     docker-compose exec -T web make test
     docker-compose stop         # Useful for local development
--- a/tox.ini
+++ b/tox.ini
@@ -32,16 +32,21 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
 commands = dennis-cmd lint --errorsonly locale/
 
 [testenv:docker]
+whitelist_externals =
+    docker-compose
+    sleep
 setenv =
     COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
+    COMPOSE_PROJECT_NAME = kumatox
 commands =
     docker-compose up -d
-    docker exec -i kuma_web_1 make compilejsi18n collectstatic
+    docker-compose exec -T web make compilejsi18n collectstatic
     # TODO: mysql needs some time to set up. find a way to eliminate the waiting time
     sleep 60
-    docker logs kuma_mysql_1
-    docker exec -it kuma_web_1 ./manage.py migrate
-    docker exec -i kuma_web_1 make test
+    docker-compose logs mysql
+    docker-compose exec web ./manage.py migrate
+    docker-compose exec -T web make test
+    docker-compose stop         # Useful for local development
 
 [flake8]
 exclude = **/migrations/**,.tox,*.egg,vendor


safwanrahman:
The test fails
https://travis-ci.org/mozilla/kuma/jobs/165934926

jwhitlock:
Force push to rebase on master, and add a 5-second pause between ``docker-compose stop`` and ``docker compose rmi``, to try to avoid [an error that the image is still used](https://travis-ci.org/mozilla/kuma/jobs/165934926#L5564-L5566).

jwhitlock:
That didn't work either.  Removing the lines that attempt to untag / remove the newly created base image.

jwhitlock:
Test passes without the image remove step.

tox.ini
```diff
@@ -32,14 +32,26 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
...
+    docker
+    docker-compose
+    make
+    sleep
+setenv =
+    COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
+    COMPOSE_PROJECT_NAME = kumatox
+commands =
+    make build-base VERSION=latest
+    docker-compose build mysql
```
safwanrahman: ``mysql`` image should be build? I think ``docker-compose up`` does all the things.

tox.ini
```diff
@@ -32,14 +32,26 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
...
+    docker-compose up -d
+    sleep 5
```
safwanrahman: The build was passing smoothly without waiting 5 seconds.

tox.ini
```diff
@@ -32,14 +32,26 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
...
-commands = 
-    docker-compose -f docker-compose.yml -f scripts/docker-compose.travis.yml up -d
-    docker exec -i kuma_web_1 make compilejsi18n collectstatic
+whitelist_externals =
+    docker
+    docker-compose
+    make
+    sleep
+setenv =
+    COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
```
safwanrahman: Awesome addition. I did not know about it.
https://docs.docker.com/compose/reference/envvars/#composefile

safwanrahman:
The test time increases by about 2 minutes. (11 min 30 sec to 13 min 48 sec)
Previous: https://travis-ci.org/mozilla/kuma/jobs/166060719
This patch: https://travis-ci.org/mozilla/kuma/jobs/166071153

tox.ini
```diff
@@ -32,14 +32,26 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
...
+    docker-compose exec -T web make compilejsi18n collectstatic
     # TODO: mysql needs some time to set up. find a way to eliminate the waiting time
     sleep 60
-    docker logs kuma_mysql_1
-    docker exec -it kuma_web_1 ./manage.py migrate
-    docker exec -i kuma_web_1 make test
+    docker-compose logs mysql
+    docker-compose exec web ./manage.py migrate
+    docker-compose exec -T web make test
+    docker-compose stop
```
safwanrahman: Do we need to stop the docker containers? Does not the environment destroyed completely after the script ends?

safwanrahman:
> Test passes without the image remove step.

I dont think we should remove the image. The environment is destroyed completely after every test

safwanrahman:
So starting ``mysql`` container first and removing 60 second sleep improve the test time. It is down to ``10 min 57 sec``
https://travis-ci.org/safwanrahman/kuma/jobs/166084265

tox.ini
jwhitlock: Yes, but spits out [a warning](https://travis-ci.org/mozilla/kuma/jobs/166060719#L390):

```
WARNING: Image for service mysql was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
```

My viewpoint is "when I look at a failed build log, what will jump out at me?" This warning will, and I'll have to ignore it, and train other devs to ignore it. Or, add a build step and eliminate it, so that only the build failure stands out.

tox.ini
```diff
@@ -32,14 +32,26 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
...
+whitelist_externals =
+    docker
+    docker-compose
+    make
+    sleep
+setenv =
+    COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
+    COMPOSE_PROJECT_NAME = kumatox
+commands =
+    make build-base VERSION=latest
```
jwhitlock: This surprised me. Yes, the override configuration rebuilds it. For a similar reason to building the ``mysql`` image, I'll switch to a manual build step for ``web`` to eliminate the warning.

tox.ini
jwhitlock: It is required to run it when I run ``tox -e docker`` locally, or the  tox environment takes over the ports needed for development.  I could just do it manually, but it takes all of 5 seconds, so why not include it...

jwhitlock:
Force-push to integrate mysql startup trick in https://github.com/safwanrahman/kuma/commit/1e183af757d67fdb142f5698674db71e20e7f6d1 , other feedback.

tox.ini
safwanrahman: I think, we can just pass the ``--build`` parameter to the ``docker-compose up``.
like ``docker-compose up --build -d mysql``

tox.ini
```diff
@@ -32,14 +32,21 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
...
+    docker-compose
+    sleep
+setenv =
+    COMPOSE_FILE = docker-compose.yml:scripts/docker-compose.travis.yml
+    COMPOSE_PROJECT_NAME = kumatox
+commands =
+    docker-compose build mysql  # Add custom collation to MySQL
+    docker-compose up -d mysql  # Give MySQL time to init DB
+    docker-compose build web    # Rebuild w/ new reqs, Dockerfile-base
+    docker-compose up -d        # Startup remainin services
```
safwanrahman: So like I mentioned, we can add the ``--build`` parameter to skip the warning.
Something like  ``docker-compose up --build -d``. And remove ``docker-compose build web``

jwhitlock: I've tried this in PR #4011, and I'm not as happy with the results.

With this version, the ``mysql`` image is built, then run without building, and then then ``web`` image is built, and then everything else runs without a build step.

With ``docker-compeose up --build -d``, the ``mysql`` image is rebuilt, and there are build steps should for ``web``, ``worker``, and ``api``.  The steps are cached and the results are the same.

I prefer this version, and the explicit steps.

jwhitlock:
Force push to rebase on master, and drop ``make`` from ``whitelist_externals``.

tox.ini
safwanrahman: I agree. We should not rebuild the mysql image

safwanrahman:
This PR looks good to me. I have just added one improvement which can be done. Other than that, r+wc from me.

jwhitlock:
Force-push to rebase and use ``docker-compose up -d --build mysql``

PR #4008, opened 2016-10-07
Adjust settings to improve debugging

jbradberry:
- enable Django Debug Toolbar by adding in the container's IP address
- log gunicorn accesses to stdout, so that the docker logs catch them


jwhitlock:
I don't like hard-coding ``'172.18.0.1'`` as the IP address, but it is also my local IP address, so maybe it is common enough to Just Work™️, and maybe can be fixed as part of [bug 1308630](https://bugzilla.mozilla.org/show_bug.cgi?id=1308630). I'm also not sure if everyone wants the toolbar, but again, let's ship it and see who complains.

PR #4010, opened 2016-10-09
[Bug 1218563] Configure Docker nginx container to serve through https

safwanrahman:
This patch setups a Docker nginx container which makes the site serve through ssl with a self signed certificate. So the site can be accessed through `https://127.0.0.1`.
I have made a redirect rule in nginx so that all request coming from `http` will be redirected to `https`.

I think, some documentation change is needed. I will find out and fix them.

@jwhitlock r?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
   # Web is based on worker b/c you cannot clear the "ports" with docker-compose.
   web:
     <<: *worker
-    command: gunicorn -w 4 --bind 0.0.0.0:8000 --timeout=120 --access-logfile=- kuma.wsgi:application
+    command: gunicorn -w 4 --bind 0.0.0.0:8000 --timeout=120 kuma.wsgi:application
     ports:
       - "8000:8000"
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,7 +38,7 @@ services:
   # Web is based on worker b/c you cannot clear the "ports" with docker-compose.
   web:
     <<: *worker
-    command: gunicorn -w 4 --bind 0.0.0.0:8000 --timeout=120 kuma.wsgi:application
+    command: gunicorn -w 4 --bind 0.0.0.0:8000 --timeout=120 --access-logfile=- kuma.wsgi:application
     ports:
       - "8000:8000"
 
@@ -55,14 +55,15 @@ services:
     ports:
       - "8001:8000"
 
-  #nginx:
-  #  build: ./docker/images/nginx
-  #  volumes:
-  #    - ./static:/app/static
-  #  ports:
-  #    - "80:80"
-  #  depends_on:
-  #    - web
+  nginx:
+    build: ./docker/images/nginx
+    volumes:
+      - ./static:/app/static
+    ports:
+      - "80:80"
+      - "443:443"
+    depends_on:
+      - web
 
   memcached:
     image: memcached
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,6 +59,7 @@ services:
     build: ./docker/images/nginx
     volumes:
       - ./static:/app/static
+      - ./ssl:/etc/nginx/ssl
     ports:
       - "80:80"
       - "443:443"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,20 +15,19 @@ services:
       - kumascript
     environment:
       # Django settings overrides:
-      - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
+      - ACCOUNT_DEFAULT_HTTP_PROTOCOL=https
       - ALLOWED_HOSTS=localhost,localhost:8000
       - BROKER_URL=redis://redis:6379/0
       - CELERY_ALWAYS_EAGER=False
-      - CSRF_COOKIE_SECURE=False
+      - CSRF_COOKIE_SECURE=True
       - DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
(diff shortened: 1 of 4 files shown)


safwanrahman:
I can not understand why the unit test consuming so much of the time after adding nginx. I will debug into it to understand where the issue is.

docker/images/nginx/Dockerfile
```diff
@@ -1,3 +1,16 @@
...
+RUN mkdir /etc/nginx/ssl
+
+RUN openssl req \
+        -x509 \
+        -nodes \
+        -newkey rsa:2048 \
+        -keyout /etc/nginx/ssl/server.key \
+        -out /etc/nginx/ssl/server.crt \
+        -days 1000 \
+        -subj "/CN=127.0.0.1"
```
jgmize: I don't think we should include a private key in a  public docker image, even for 127.0.0.1. I believe a more correct approach, although less convenient, would be to document a version of this command (possibly wrapped in a "[make generate-cert](https://github.com/mdn/viewsourceconf/pull/98/files#diff-b67911656ef5d18c4ae36cb6741b7965R75)" or similar) to be run on the developer's laptop, and then [mount the cert and key](https://github.com/mdn/viewsourceconf/pull/98/files#diff-b67911656ef5d18c4ae36cb6741b7965R24) from the host at runtime.

safwanrahman: Cool. I will add it in next iteration.
Just debugging why the test taking a long time to run!

docker-compose.yml
```diff
@@ -15,21 +15,20 @@ services:
...
       - ES_URLS=elasticsearch:9200
       - KUMASCRIPT_URL_TEMPLATE=http://kumascript:9080/docs/{path}
       - MEMCACHE_SERVERS=memcached:11211
-      - PROD_DETAILS_DIR=/app/product_details_json
-      - PROTOCOL=http://
-      - SESSION_COOKIE_SECURE=False
-      - SITE_URL=http://localhost:8000
+      - PROTOCOL=https://
+      - SESSION_COOKIE_SECURE=True
+      - SITE_URL=https://localhost:8000
```
jgmize: The nginx container is listening on the standard port 443, so the `:8000` is incorrect here and may be causing timeouts during test runs.

docker/images/nginx/Dockerfile
jgmize: Also, the generated cert is only valid for `127.0.0.1`; the `subj` should be `/CN=localhost`.

docker-compose.yml
escattone: Thanks jgmize! I made this change and, as you suspected, it resolved the timeouts that were occurring and consuming so much time when running the tests on this branch.

jgmize: I'm glad that resolved the timeouts, but in hindsight I should have questioned the value of forcing tests through https by default anyway. Another option would be to use the `docker-compose.test.yml` file and continue to use http for these tests, and only use https for the integration tests.

jwhitlock:
The environment updates have eliminated the timeouts that occurred with earlier versions of this PR.

We're now stuck with the "why" of the PR, which is less clear now.  Some possible whys:

* Persona has a requirement that the "audience" URL is an HTTPS URL (but Persona is going away)
* Browser security models prefer that SSL sites load SSL assets (but I believe HTTP sites can use HTTPS assets all day)
* Running in ``DEBUG=False`` mode turns on some checks that may fail if SSL is off
* It may be nice to check CSP headers in an HTTPS environment
* It would be nice to be able to run a production-like environment locally, including SSL serving of assets (but we're not sure what the production environment looks like yet)

In production, we'll probably:

1. Deploy web containers serving content over HTTP
2. Use AWS's [Elastic Load Balancer](https://aws.amazon.com/elasticloadbalancing/), with SSL termination at the the ELB, instead of nginx
3. Server static assets from the web containers or similar using [whitenoise](http://whitenoise.evans.io)
4. Use [CloudFront](https://aws.amazon.com/cloudfront/) to cache assets

Am I missing something? Should we put this PR on hold until we have a better idea of what the production environment is, and the benefits / costs of simulating it in the dev environment?

docker-compose.yml
```diff
@@ -38,6 +37,7 @@ services:
   # Web is based on worker b/c you cannot clear the "ports" with docker-compose.
   web:
     <<: *worker
+    cpu_shares: 700
```
jwhitlock: Is this still necessary or useful after the environment file changes?

stephaniehobson:
There are some front-end stuff that behaves differently when you serve it over http vs https. I'd like to avoid getting mixed content errors in dev. It'll be hard to know what is and isn't a real security error.

Eventually we will want to add service workers to MDN, those can only be run on https.

jwhitlock:
@stephaniehobson that sounds like a good reason to run with an SSL configuration for functional tests.  When do you need SSL in the dev environment?

stephaniehobson:
@jwhitlock based on my work this sprint and our plans this quarter it can wait a few sprints. Also, I'm still developing in Vagrant (I know... I know...) so I can test https dependent things with it.

jwhitlock:
It's been many months. We haven't needed this for testing, and service workers aren't on the roadmap.

@safwanrahman are you OK closing the PR? If not, please keep the branch up to date.

escattone:
👍 I would vote to close it as well.

safwanrahman:
@jwhitlock I am closing the PR for now. I will fix the conflict issue, and reopen this PR soon. Thanks a lot @escattone for the work. I am keeping the branch in my remote and upon fixing the issues, I will reopen it

PR #4011, opened 2016-10-10
Bug 1218563: Improve TravisCI docker build and test (version 2)

jwhitlock:
Improve on PR #3977 (test docker travis):

- Use environment variables to simplify ``docker-compose`` calls
- Eliminate some warnings by adding build steps, configuring ``tox``
- Start the ``mysql`` container while the ``web`` image is building, to eliminate ``sleep 60``

This is a variant of PR #4007 that uses ``docker-compose up --build -d`` instead of ``docker-compose build web; docker-compose up -d``. Both avoid a warning about implicit builds, but I want to compare the output from the two methods.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -32,8 +32,10 @@ deps = https://github.com/willkg/dennis/archive/aca57f6d.zip
 commands = dennis-cmd lint --errorsonly locale/
 
 [testenv:docker]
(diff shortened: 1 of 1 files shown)


jwhitlock:
closing in favor of #4007

PR #4012, opened 2016-10-11
bug 1269143: escape sample name

escattone:
* In wiki.content::Extractor.code_sample, escape the name of
the code sample when falling-back to the pyquery find.
* Add test for this case.

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -4,6 +4,7 @@ import urllib
 from collections import defaultdict
 from urllib import urlencode
 from urlparse import urlparse
+from xml.sax.saxutils import quoteattr
 
 import html5lib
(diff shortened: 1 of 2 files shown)


kuma/wiki/content.py
```diff
@@ -140,7 +141,7 @@ def code_sample(self, name):
             sample = pq('<section>%s</section>' % section)
         else:
             # If no section, fall back to plain old ID lookup
-            sample = pq(src).find('[id="%s"]' % name)
+            sample = pq(src).find('[id=%s]' % quoteattr(name))
```
jwhitlock: Interesting use of ``xml.sax.saxutils.quoteattr``. That will probably catch many corner cases we. haven't thought of

PR #4016, opened 2016-10-14
Fix Bug 1149205: Break long titles intelligently.

stephaniehobson:
Previously we've used the CSS property hyphens to break up long titles on short screens but it's not cross browser and it breaks stuff that doesn't need to be broken, obscuring meaning.

This small script targets only long article titles and adds a <wbr> tag to give the browser more accurate direction for where to break titles.

Testing pages:
https://developer-local.allizom.org/en-US/docs/Web/API/MediaStreamAudioDestinationNode/stream

https://developer-local.allizom.org/en-US/Add-ons/Firefox_for_Android/API/window.NativeWindow/window.NativeWindow.contextmenus/window.NativeWindow.contextmenus.remove

https://developer-local.allizom.org/en-US/docs/Web/CSS/:-moz-system-metric%28scrollbar-thumb-proportional%29

(make sure code escaping is working properly)
https://developer-local.allizom.org/en-US/docs/Web/HTML/Element/blockquote

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -162,6 +162,24 @@
     });
 
     /*
+        Add intelligent break points to long article titles
+    */
+    $('#wiki-document-head h1').each(function() {
+        var $title = $(this);
+        var text = $title.text();
+        // split on . - : ( or capital letter, only if followed by 2 letters
+        var split = text.split(/(?=[\.:\-\(A-Z][\.:\-\(A-Z]{0,}[a-zA-Z]{3})/g);
+        // empty h1
+        $title.empty();
+        // put array back into h1 seperated by <wbr> tags
+        $.each(split, function(key, value) {
+            $title.append('<wbr>');
+            // add text back, make sure it goes back as text, not code to run
+            $title.append(doc.createTextNode(value));
+        });
+    });
+
+    /*
         Syntax highlighting scripts
     */
     if($('article pre').length && ('querySelectorAll' in doc)) (function() {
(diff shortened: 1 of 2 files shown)


stephaniehobson:
Before:
screen shot 2016-10-14 at 14 21 05

screen shot 2016-10-14 at 14 24 05

After:
screen shot 2016-10-14 at 14 22 13

screen shot 2016-10-14 at 14 24 32

stephaniehobson:
Updated to (hopefully) not break acronyms at @escattone's suggestion. :tada:

schalkneethling:
Looking really good to me r+

Before:
screen shot 2016-10-19 at 14 42 33

After:
screen shot 2016-10-19 at 14 42 48

Before:
screen shot 2016-10-19 at 14 45 16

After:
screen shot 2016-10-19 at 14 45 24

stephaniehobson:
Hm. Should be breaking before the `-`. Will take another look :(

schalkneethling:
@stephaniehobson Ha! true. It breaks much cleaner/intelligently though so I did not even notice that ;)

schalkneethling:
The array seems to be created correctly:

screen shot 2016-10-20 at 07 36 44

Perhaps the breaking is not happening exactly as expected because `wbr` is simply a suggestion and not definitive so, it might break at other places as well as the viewport is sized?

The breaking is a lot better though so, perhaps it is not worth fighting to much with the rendering engine.

stephaniehobson:
Oh! Probably what is happening is that the browser thinks the hyphens are a sensible place to break! Yes, you are right, my suggestion is being over-ridden by the browser 'helping'. This would make sense with normal words but with code it's not clear if the hyphen is part of the code or inserted to indicate the wrap.

I could try to fight this with a &nbsp; but you're right, this is an improvement and hopefully the article body will include the proper syntax.

PR #4019, opened 2016-10-21
bug 1269143: fix for null/ctl chars in sample name

escattone:
* fix case when sample name contains null or control
  characters (in general whenever a ValueError is
  raised within the PyQuery.find method).
* add test case

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -141,7 +141,10 @@ class Extractor(object):
             sample = pq('<section>%s</section>' % section)
         else:
             # If no section, fall back to plain old ID lookup
-            sample = pq(src).find('[id=%s]' % quoteattr(name))
+            try:
+                sample = pq(src).find('[id=%s]' % quoteattr(name))
+            except ValueError:
+                return data
 
         selector_templates = (
             '.%s',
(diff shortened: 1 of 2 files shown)


kuma/wiki/content.py
```diff
@@ -141,7 +141,10 @@ def code_sample(self, name):
             sample = pq('<section>%s</section>' % section)
         else:
             # If no section, fall back to plain old ID lookup
-            sample = pq(src).find('[id=%s]' % quoteattr(name))
+            try:
+                sample = pq(src).find('[id=%s]' % quoteattr(name))
+            except Exception:
```
jwhitlock: I'd like to restrict this to ``ValueError``, so that if a different exception is raised we'll still get the traceback.

kuma/wiki/tests/test_content.py
```diff
@@ -1149,8 +1149,13 @@ def test_code_samples_with_problematic_sample_name(self):
         """)
         # The real test here is to ensure that no exception is raised, but
         # might as well also check that the sample section was not found.
-        result = rev.document.extract.code_sample("""sam<'&">ple""")
-        eq_(dict(html=None, css=None, js=None), result)
+        sample_names = (
```
jwhitlock: Is this a use case for py.test's parametrization? I haven't used it myself.

jwhitlock:
Thanks for the change, and for investigating ``py.test`` parameterization in our current frameworks.  We've got a lot of cleanup for that hypothetical rainy day...

escattone:
Thanks John. I'd love to move fully into the py.test fixtures and parameterization, but that means we have to move away from using django.test.TestCase, and unfortunately that's not a quick task. Once we define the fixtures that cover the setup and teardown that django.test.TestCase and django.test.TransactionTestCase perform, we'll be a good chunk of the way there.

Aban007:
Thanks

PR #4022, opened 2016-10-27
Bug 1259776: Display ban info on user profile

stephaniehobson:
Displays ban info to users with ban permissions instead of super user permissions.

Adds ban date and moderator to user profile page.

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -358,7 +358,7 @@ def user_detail(request, username):
     """
     detail_user = get_object_or_404(User, username=username)
 
-    if (detail_user.active_ban and not request.user.is_superuser):
+    if (detail_user.active_ban and not request.user.has_perm('users.add_userban')):
         return render(request, '403.html',
                       {'reason': 'bannedprofile'}, status=403)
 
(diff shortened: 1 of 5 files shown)


escattone:
Based on this PR, my understanding is that the current logic for determining a user's "access" to banned info is by checking `user.has_perm('users.add_userban')` rather than `user.is_superuser`? If that's the case, it appears that the logic in `kuma/users/view.py::user_detail` should be changed as well?

stephaniehobson:
Ooh, yes, good catch.

stephaniehobson:
Updated :hammer:

kuma/users/views.py
```diff
@@ -352,7 +352,7 @@ def user_detail(request, username):
     """
     detail_user = get_object_or_404(User, username=username)
 
-    if (detail_user.active_ban and not request.user.is_superuser):
+    if (detail_user.active_ban and not request.user.add_userban):
```
jwhitlock: Should be ``request.user.has_perm('users.add_userban')``, not ``request.user.add_userban``.

stephaniehobson:
Updated :aerial_tramway:

PR #4023, opened 2016-10-27
Fix Bug 1276371: Hide IP address on dashboard after filter applied

stephaniehobson:
Also fix bug 1286732 - MDN revision dashboard has duplicate element ids

Hide IP address by default on revision dashboard page, on all pages, not just first page of results.
- Moved dashboard javascript to external file and included file.
- Add hidden class to revision_ip to hide by default.
- Ensure toggle button has event listener attached to it after filters updated.
- Added 'table-layout:fixed' to stop table from moving around when IPs toggled into view.
- Fixed bug with keyboard navigation.
- Add interpolate() to .jshintrc as it is part of one of our global libraries like gettext.
- Updated spam-ham-buttton to use class instead of (duplicated) ID, everywhere - including tests.
- Updated spam-ham test to look for existance of spam-ham button instead of that column in the table.

Things to test:
- user autocompletes still work
  - the topic filter is finicky because the result set is potentially huge I have mixed success getting it to work in production and it spins up the fans on my host machine when I try it in dev. Usually the worker quits and the request times out. Smaller requests are more likely to succeed so try something like `Web/HTML/Element/pr`. Or just skip testing it since no one has complained about it being broken in prod.
- IPs are hidden by default and the button 'Toggle IPs' appears in the head of the revision dashboard table
- IPs appear if button is clicked 
- IPs are hidden after a filter is applied & can still be toggled into view
  - filters can be applied by filtering using the form, or by clicking on a page in the pagination links
- clicking on a row in the table expands it
- pressing enter while focused on a row in the table expands it


stephaniehobson:
@schalkneethling 
Revision dashboard is here: https://developer-local.allizom.org/en-US/dashboards/revisions

This one requires a database of content to do testing, if you don't have that anymore let me know.

schalkneethling:
@stephaniehobson So, the code looks good to me, trying to test it locally I am getting an Internal Server Error, not sure where I will get the trace for that though.

Also, with docker mine is running via:
http://localhost:8000/en-US/dashboards/revisions

Not sure if this might be causing the problem?

schalkneethling:
@stephaniehobson Ok, found the logs, and here is the root cause:

```
OperationalError: (1054, "Unknown column 'wiki_revisionip.data' in 'field list'")
```

I did restore the DB extract you provided a while back but, perhaps it is out of date and does not contain this column?

jwhitlock:
@schalkneethling ``./manage.py migrate`` may fix that

schalkneethling:
@jwhitlock It Works! I did:

`docker exec -i kuma_web_1 bash -c "./manage.py migrate"`

schalkneethling:
@stephaniehobson Not sure if I am still missing something but, I do not see the show IP toggle anywhere but, I can see it is the code from this branch that is running because of the `table-layout: "fixed";` being there:

screen shot 2016-11-01 at 17 04 06

Also, when I click on a row to expand it, it does expand but the table breaks out of it's container

screen shot 2016-11-01 at 17 06 05

Collapsing the row after does not fix the problem with the rows breaking out of the containing table:

screen shot 2016-11-01 at 17 06 17

Perhaps it is like this in production and not an issue but, thought I would mention it in case it is not expected.

stephaniehobson:
You might not have the correct permissions to see the button, are you a super user and logged in? It should be in the table heading next to Author.

I think the wide code preview tables are a problem I introduced in this PR. Thanks for the catch :)

stephaniehobson:
Wide code previews were only a problem for users without permission to see spam reporting. Should be fixed now.

kuma/dashboards/jinja2/dashboards/includes/revision_dashboard_body.html
```diff
@@ -92,6 +94,8 @@
 
                         {% endif %}
                     </td>
+                    {% else %}
+                    <td class="dashboard-spam">&nbsp;</td>
```
schalkneethling: This as well as the entry on line 14 above is causing the automated tests to fail, as it expects and empty array to be returned:

```
page = pq(response.content)
spam_table_cell = page.find('td.dashboard-spam')
eq_(spam_table_cell, [])
```

This is in kuma/dashboards/tests/test_views.py on line 85-87

I am guessing you added this to overcome the rows breaking out of it's container so, one would either need to update the test or, explore another solution, perhaps a different class name or no class name at all? @stephaniehobson Thoughts?

schalkneethling:
Other than my last comment, this is all good so, as soon as Travis is happy, merge at will. r+w/c

kuma/dashboards/jinja2/dashboards/includes/revision_dashboard_body.html
stephaniehobson: Went to fix this and found duplicate IDs on the page >_<

PR #4024, opened 2016-11-01
bug 1314290: Update to Django 1.8.16

jwhitlock:
Django 1.8.15 -> 1.8.16 - Security release

https://www.djangoproject.com/weblog/2016/nov/01/security-releases/


safwanrahman:
the security update seems like not affecting any of the components of kuma.
As the tests pass, r+

PR #4025, opened 2016-11-02
bug 1314613: Add locale refresh as TravisCI task

jwhitlock:
Refresh the locales in TravisCI, to detect issues in templates and translations during the PR process. Run a diff of the changed strings, to make it easier to determine when to release a new string pack.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
         - TOXENV=flake8
         - TOXENV=docs
         - TOXENV=dennis
+        - TOXENV=locales DOCKER_COMPOSE_VERSION=1.8.0
         - TOXENV=docker DOCKER_COMPOSE_VERSION=1.8.0
     global:
         - DJANGO_SETTINGS_MODULE=kuma.settings.travis
@@ -27,6 +28,7 @@ env:
 matrix:
     allow_failures:
         - env: TOXENV=dennis
+        - env: TOXENV=locales DOCKER_COMPOSE_VERSION=1.8.0
         - env: TOXENV=docker DOCKER_COMPOSE_VERSION=1.8.0
 before_install:
     - scripts/travis-install
--- a/scripts/travis-install
+++ b/scripts/travis-install
@@ -19,14 +19,14 @@ then
     # Install Elasticsearch
     wget https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.2.deb
     sudo dpkg -i elasticsearch-1.7.2.deb
-    
+
     # limit elasticsearch / java memory usage to avoid OOM Killer
     sudo service elasticsearch stop;
     echo "ES_HEAP_SIZE=256m" | sudo tee --append /etc/default/elasticsearch
     sudo service elasticsearch start;
 fi
 
-if [ "$TOXENV" == "docker" ]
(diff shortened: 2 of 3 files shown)


safwanrahman:
I think, its not possible to extract locales without making a full instance. Better we run ``make localerefresh`` in the docker environment.

jwhitlock:
Agreed. Force-push to switch to using docker to build locales.

safwanrahman:
You should pass ``docker`` and ``docker-compose`` environment variable in order to install the correct version of docker-compose as mentioned [here](https://github.com/mozilla/kuma/blob/master/scripts/travis-install#L29). We have faced some issues with the mismatch of docker-compose version. so with the ``DOCKER_COMPOSE_VERSION``, we have fixed the ``docker-compose`` version.

BTW, do you think, we should run ``make localrefresh`` inside the existing docker test environment? So after all test passes, it will run the localrefresh.
Also, I think we should run ``make localecompile`` before start running the tests in docker.

jwhitlock:
> You should pass ``docker`` and ``docker-compose`` environment variable in order to install the correct version of docker-compose as mentioned [here](https://github.com/mozilla/kuma/blob/master/scripts/travis-install#L29). 

Force-push to add ``DOCKER_COMPOSE_VERSION`` and update ``travis-install``.

> BTW, do you think, we should run make localrefresh inside the existing docker test environment? So after all test passes, it will run the localrefresh.

There's a couple of reasons I'm not adding this to the docker task:

* I think docker is overkill for this task, or at least the full docker environment used for the tests. It appears ``./manage.py extract`` needs a database to get it's job done, but it shouldn't need a worker, a search engine, and API, etc. etc.  There's a lot of room for a much simpler docker environment, or maybe eliminate docker entirely (using TravisCI's MySQL, sqlite, or eliminate the database dependency)
* String compiling finds issues that should be fixed, but shouldn't block builds.  Eventually, docker build problems should block builds. So, the locale building task should remain as a CI task that is allowed to fail, but docker may eventually move to tasks that must pass.

The reason I'm doing this today is because there is a logic error in a template that is [detected by string compiling](https://travis-ci.org/mozilla/kuma/jobs/172638765#L2160).  So, I'm motivated to merge the simplest possible version that will work, and then optimize and rearrange the task later.

> Also, I think we should run ``make localecompile`` before start running the tests in docker.

Pontoon pushes new translations to master that can potentially break ``make localecompile``. My experience is that both should be tested, but translators shouldn't have to figure out breakage from code changes, and developers shouldn't have to figure out breakage from translations.

jwhitlock:
I may have talked myself out of rushing this PR 😄 . Opened PR #4026 for the template logic error.  I may play with some alternate docker-compose configs (eliminate search, worker, etc.) to make this task faster.

safwanrahman:
Agreed with you. Thanks for the explanation.

jgmize:
>  I may play with some alternate docker-compose configs (eliminate search, worker, etc.) to make this task faster.

There are some examples in [docker-compose.test.yml](https://github.com/mozilla/kuma/blob/master/docker-compose.test.yml), and the following command worked for me:

```sh
docker-compose -f docker-compose.yml -f docker-compose.test.yml run --user=$(id -u) noext make localerefresh
```

jwhitlock:
Force-push to try @jgmize's ``noext`` configuration. Can it beat 7 min 9 seconds?

PR #4028, opened 2016-11-02
Bug 1310322: Fix thank you message overlapping close button

stephaniehobson:
Minor bug fix.


stephaniehobson:
@jwhitlock this should go after lunch if it can.

PR #4030, opened 2016-11-03
Bug 1314708: Add documentation for functional tests

stephaniehobson:


m8ttyB:
r+

docs/tests-ui.rst
```diff
@@ -1,177 +1,148 @@
...
 
-On your machine
-~~~~~~~~~~~~~~~
+   $ source mdntests/bin/activate
 
-1. Install `JDK <http://www.oracle.com/technetwork/java/javase/downloads/index.html>`_
+#. Install the requirements the tests need to run::
 
-2. Download the most current `release of Selenium <http://selenium-release.storage.googleapis.com/index.html>`_ standalone server. (It's the ``.jar`` file.)
+   $ pip install -r requirements/test.txt
```
jwhitlock: I had to upgrade ``pip`` as well, since my default is 7.1.2, which is before hashes.  This may just be me and my old laptop.  To enforce it, add:

    $ pip install "pip>=8"

docs/tests-ui.rst
```diff
@@ -1,177 +1,148 @@
...
-    # Substitute your WebDriver version in the `#` chars
-    java -jar /path/to/selenium-server-standalone-#.#.#.jar
+The minimum amount of information necessary to run the tests against the staging
+server is the directory the tests are in and which driver to use.
 
-4. Go to the ``tests/ui/`` directory::
+In the virtual environment run::
 
-    cd tests/ui
+   $ py.test tests/functional/ --driver Firefox --driver-path /path/to/geckodriver
```
jwhitlock: This is the first time the driver is mentioned. It may be necessary to point people at documentation for downloading the ``geckodriver`` first. It is not obvious from an internet search how to get it, and it is also not obvious how [these documents](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver) apply.

After downloading ``geckodriver`` for OS X from the [release page](https://github.com/mozilla/geckodriver/releases), I get an error with this command:

```
py.test tests/functional/ --driver Firefox --driver-path ~/Downloads/geckodriver

======================================================= test session starts ========================================================
platform darwin -- Python 2.7.12, pytest-3.0.3, py-1.4.31, pluggy-0.4.0
driver: Firefox
sensitiveurl: .*
rootdir: /Users/john/src/kuma, inifile: pytest.ini
plugins: base-url-1.1.0, html-1.10.1, rerunfailures-1.0.2, selenium-1.4.0, variables-1.4, xdist-1.15.0
collected 29 items / 1 errors 

============================================================== ERRORS ==============================================================
________________________________________ ERROR collecting tests/functional/test_notfound.py ________________________________________
ImportError while importing test module '/Users/john/src/kuma/tests/functional/test_notfound.py'.
Original error message:
'No module named urls'
Make sure your test modules/packages have valid Python names.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
===================================================== 1 error in 0.12 seconds ======================================================

```
This is running against the master branch, not this branch.

docs/tests-ui.rst
```diff
@@ -1,177 +1,148 @@
...
-    ./node_modules/.bin/intern-runner config=intern-local b=firefox
+Only running tests in one file
+------------------------------
 
-The above tries to run the entire suite of tests. You can change the behavior with `command line arguments`_. E.g., ::
+Add the name of the file to the test location::
 
-    node_modules/.bin/intern-runner config=intern-local b=firefox,chrome t=auth,homepage d=developer-local.allizom.org u=someone@somewhere.com p=8675309 wd='Web' destructive=true
+   $ py.test tests/functional/test_search.py --driver Firefox --driver-path
+   /path/to/geckodriver
```
jwhitlock: When running this command, I'm asked if I want ``python2.7`` to accept connections (I accept), and then I'm presented with a blank Firefox window.  After a few seconds, the test fails, with the final error:

```
WebDriverException: Message: Can't load the profile. Profile Dir: /var/folders/61/s6_xxhqd3nl27_vgq9fzjmkr0000gq/T/tmp2ANHVe If you specified a log_file in the FirefoxBinary constructor, check it for details.
```

This is with Firefox 49.0.2.  Full traceback: https://pastebin.mozilla.org/8924594

jwhitlock: If it only works with Firefox 47, that's OK (although I'd appreciate a hint how to go back), but it needs to be in the docs.

docs/tests-ui.rst
stephaniehobson: Is this a separate command they should run first?

docs/tests-ui.rst
jgmize: See https://github.com/mozilla/bedrock/blob/master/docker/jenkins/run_integration_tests.sh for an example of how to run the older firefox and a selenium grid locally and in CI via docker.

docs/tests-ui.rst
stephaniehobson: Should be fixed by #4043

docs/tests-ui.rst
m8ttyB: We'll discuss more offline, however `geckodriver` is ready for primetime. If you have difficulties with it, let's do some pairing.

stephaniehobson:
Updated 🎉

docs/tests-ui.rst
```diff
@@ -1,177 +1,151 @@
...
+to emulate users interacting with a real browser. All these test suites live in
+the /tests/ directory.
 
-.. _SitePen: http://sitepen.com
-.. _Intern: https://theintern.github.io/
+The tests directory comprises of:
 
-Installing Dependencies
------------------------
+   * ``/functional`` contains pytest tests.
```
jwhitlock: Please remove the leading whitespace, or else it will be styled like a block quote.

screenshot 2016-11-22 12 15 46

docs/tests-ui.rst
```diff
@@ -1,177 +1,151 @@
...
-To run your new tests with, add the new suite path to the `tests/ui/_tests.js` file.
+   Tests are considered destructive unless otherwise indicated. Tests that
+   create, modify, or delete data are considered destructive and should not be
+   run in production.
 
-Identifying Test Failures
--------------------------
+- smoke
+
+   These tests should be the critical baseline functional tests.
```
jwhitlock: Same here - remove one space or the "definition" will be a block quote:

screenshot 2016-11-22 12 17 02

PR #4031, opened 2016-11-03
Escaping placeholder in template for bug 1315078

SphinxKnight:
I just surrounded the string with the usual pattern and checked that HTML attributes were also dealt the same way on https://github.com/mozilla/kuma/blob/752bfb9ec0e803655c88a5391bba92b86ef33bdc/jinja2/includes/login.html#L19. Hoping it's ok for such a minor matter.


jwhitlock:
What would you localize this to in French?

SphinxKnight:
It would be "adresse@example.com" (cf. https://bugzilla.mozilla.org/show_bug.cgi?id=1315078#c3)

jwhitlock:
Makes sense to me, and no objections, so merging.

PR #4033, opened 2016-11-07
bug 1308322: Upgrade ndg-httpsclient and dependencies

jwhitlock:
This PR updates ``ndg-httpsclient`` and its many dependencies, including a security fix in [cryptography 1.5.3](https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#153---2016-11-05).

Background:

Old, un-patched versions of Python 2.7 do not support [Server Name Indication](https://en.wikipedia.org/wiki/Server_Name_Indication), which allows multiple hosts to be served with different SSL certificates from a single IP address.  Our Vagrant environment includes an un-patched Python 2.7, while our production and Docker environment has a patched Python. To see what your Python does, run the following (with the [requests](http://docs.python-requests.org/en/master/) package installed):

```
python -c 'import requests; print(len(requests.get("https://sni.velox.ch").content))'
```

An un-patched system will return something like:
```
/home/vagrant/env/local/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:315: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#snimissingwarning.
  SNIMissingWarning
/home/vagrant/env/local/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:120: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
  File "<string>", line 1, in <module>
  File "/home/vagrant/env/local/lib/python2.7/site-packages/requests/api.py", line 67, in get
    return request('get', url, params=params, **kwargs)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/requests/api.py", line 53, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/home/vagrant/env/local/lib/python2.7/site-packages/requests/adapters.py", line 447, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: hostname 'sni.velox.ch' doesn't match either of 'alice.sni.velox.ch', 'carol.sni.velox.ch'
```

A patched system will return something like ``5067``.

Vagrant installs ``ndg-httpsclient`` to patch Python 2.7 with ``pyOpenSSL``, removing SNI warnings and errors, in ``requirements/local.txt``.  This package requires ``cryptography``, which has a security issue that is fixed in 1.5.3.


jwhitlock:
Force push to rebase and copy edit commit messages

jwhitlock:
Force-push for rebase, and today's ``cryptography`` and ``cffi``.

PR #4036, opened 2016-11-08
[Bug 1308168] non-translated page should have English link to button, banner

safwanrahman:
So if there are ``fallback_reason``, we should show the ``parent document`` locale as choice also. As the ``document`` is replaced by ``parent_document`` from view, we can only use the ``document`` attributes.
@jwhitlock r?

### DIFF
--- a/kuma/wiki/jinja2/wiki/document.html
+++ b/kuma/wiki/jinja2/wiki/document.html
@@ -305,7 +305,7 @@
                   <p><bdi>{% trans help_link=help_link, locale=settings.LOCALES[request.LANGUAGE_CODE].native, parent_language=document.language %}
                     Our volunteers haven't translated this article into <bdi>{{ locale }}</bdi> yet.
                     <a href="{{ help_link }}">Join us and help get the job done!</a><br>
-                    You can also read the article in <a href="{{ doc_abs_url }}"> {{ parent_language }}.</a>
+                    You can also read the article in <a href="{{ doc_abs_url }}">{{ parent_language }}</a>.
                   {% endtrans %}</bdi></p>
                 </div>
               {% endif %}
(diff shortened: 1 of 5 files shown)


kuma/wiki/jinja2/wiki/document.html
```diff
@@ -302,9 +302,10 @@
               {% endif %}
               {% if fallback_reason %}
                 <div id="doc-pending-fallback" class="warning">
-                  <p><bdi>{% trans help_link=help_link, locale=settings.LOCALES[request.LANGUAGE_CODE].native %}
+                  <p><bdi>{% trans help_link=help_link, locale=settings.LOCALES[request.LANGUAGE_CODE].native, parent_language=document.language %}
                     Our volunteers haven't translated this article into <bdi>{{ locale }}</bdi> yet.
-                    <a href="{{ help_link }}">Join us and help get the job done!</a>
+                    <a href="{{ help_link }}">Join us and help get the job done!</a><br>
+                    You can also read the article in <a href="{{ doc_abs_url }}"> {{ parent_language }}.</a>
```
jwhitlock: Can you adjust the period placement, so it is outside the link?

```You can also read the article in <a href="{{ doc_abs_url }}">{{ parent_language }}</a>.```

safwanrahman:
@jwhitlock Fixed the issue.
Regarding branch name, I will take care of it in future. Thanks

jwhitlock:
Thanks for the PR!

jwhitlock:
Hmm this ended up with a lot more code than I thought it would. I think I was hasty to merge it.

safwanrahman:
@jwhitlock Sorry, my test debugging code also come up in the commit.
I am very much sorry for that

jwhitlock:
😄 we both were too excited about getting this merged.  I've removed the code with pushes to the master branch, will be more careful next time and won't review code from my phone.

PR #4040, opened 2016-11-15
bug 1218563: Use py.test --reuse-db for speed

jwhitlock:
Use ``--reuse-db`` by default, which does not destroy the test database at the end of testing, and re-uses it on the next test run without running the long database migrations.  Update the documentation with instructions for recreating the test database on schema changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ env:
         - DEBIAN_FRONTEND=noninteractive
         - DATABASE_URL=mysql://root:kuma@localhost:3306/kuma
         - CFLAGS=-O0
+        - PYTEST_ADDOPTS=--create-db
 matrix:
     allow_failures:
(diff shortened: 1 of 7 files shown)


jwhitlock:
Well that failed hard 😀 

```
OperationalError: (1049, "Unknown database 'test_developer_mozilla_org'")
```
I'm not sure why ``py.test`` failed to create the initial test database.

jwhitlock:
I suspect that latest versions of py.test may not require explicitly adding ``--create-db`` on the first run. We're at 2.8.7, and py.test 3.0.5 was [released in December 2016](https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst).  That will probably require some other changes, and should be in a different PR.

jwhitlock:
I use ``--reuse-db`` all the time for local development, but I think we need the requirements update first.  Removing this from the PR backlog as pre-mature.

PR #4041, opened 2016-11-15
bug 1271552: Config case-insenstive tags

jwhitlock:
django-taggit 0.16.0 added config item ``TAGGIT_CASE_INSENSITIVE``.  When ``True``, an existing tag 'JavaScript' will prevent a new tag 'javascript' from being added. When ``False`` (the default), 'javascript' will be added as well.  This doesn't prevent problems if you try to add 'TAG' and 'tag' at the same time.

This does not fix the bad data already in production, but may prevent new duplicate tags from being added.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1424,3 +1424,6 @@ if SENTRY_DSN:
 # Tell django-recaptcha we want to use "No CAPTCHA".
 # Note: The API keys are located in Django constance.
 NOCAPTCHA = True  # Note: Using No Captcha implies SSL.
+
+# Prevent adding a tag like 'javascript' if 'JavaScript' exists
+TAGGIT_CASE_INSENSITIVE = True


safwanrahman:
@jwhitlock Is it possible to add a test about it? so we dont get any regression in future while updating database or taggit.

jwhitlock:
Good idea, will attempt to add a test. MySQL is a weird database, so I'm 80% sure I can write one.

jwhitlock:
I'm abandoning this PR. The [code](https://github.com/alex/django-taggit/blob/master/taggit/managers.py#L200-L213) has some assumptions that don't work for us:

1. The  assumes that ``DocumentTag.objects.filter(name__iexact)`` will return 0 or 1 results. Our database returns multiple results (this is the problem we're trying to solve). Calling ``.get`` will raise an unhandled ``MultipleObjectsReturned``.
2. The list of tags to set is not de-duplicated. So, if you try to set to "css CSS Css", the code will go ahead and create three tags, leading you to problem #1 again.

No silver bullets. ☹️

PR #4042, opened 2016-11-16
[Bug 1308222] move page fails with multiple profile same email

safwanrahman:
Use ``id`` instead of ``email``
@jwhitlock Small fix. r?


jwhitlock:
Code looks good, and tests pass in Docker environment (locally and in TravisCI).  The py27 failure is due to a [known issue](https://www.traviscistatus.com/incidents/gmd6p074w061) with apt.dockerproject.org's CDN.

Thanks for the PR!

PR #4043, opened 2016-11-17
Bug 1314708: Clean up Tests

stephaniehobson:
- Check module imports
- Fix spacing
- Add destructive test definition
- Change test names
- Combine tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,6 +11,7 @@ VIEWPORT = {
 def base_url(base_url, request):
     return base_url or 'https://developer.allizom.org'
 
+
 @pytest.fixture
 def selenium(request, selenium):
     viewport = VIEWPORT['large']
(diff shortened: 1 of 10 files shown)


stephaniehobson:
@m8ttyB R?

tests/conftest.py
```diff
@@ -11,10 +12,21 @@
...
+
+
+@pytest.fixture(scope='session')
+def sensitive_url(request):
+    # developer-local and localhost are not sensative
+    developer_local_match = str(request).startswith('https://developer-local')
+    localhost_match = str(request).startswith('localhost')
+    # treat it as sensitive if it's not identified as okay
+    #return not (developer_local_match or localhost_match)
+    return False
```
m8ttyB: Is this a todo item? This will always return `False`.

tests/conftest.py
```diff
@@ -1,3 +1,4 @@
+import re
```
m8ttyB: Is this import being used?

stephaniehobson: It was being used when I was using a regex instead of string comparisons but it's not anymore.

stephaniehobson:
Needs to be tested running on localhost. Going to install Docker and do that myself but if someone else spots this and is excited to try it let me know and I'll post steps.

PR #4047, opened 2016-11-18
[Bug 1239756] Purge taggit_tag table of dead tags

safwanrahman:
This 2 migrations will remove the un used taggs as well as taggeditem without models from database

@jwhitlock r?

### DIFF
--- a/kuma/core/migrations/0003_remove_unused_taggeditems.py
+++ b/kuma/core/migrations/0003_remove_unused_taggeditems.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from django.contrib.contenttypes.models import ContentType
+
+def remove_taggeditem_with_no_models(apps, schema_editor):
+    TaggedItem = apps.get_model('taggit', 'TaggedItem')
+
+    tags_content_type_id = TaggedItem.objects.values_list('content_type_id', flat=True).distinct()
+
+    for ct_id in tags_content_type_id:
+        ct = ContentType.objects.get(id=ct_id)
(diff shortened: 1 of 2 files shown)


jwhitlock:
This appears to work on a anonymized production database.  I created some test fixtures:

```
ct = ContentType.objects.create(app_label='legacy', model='legacy')
tag = Tag.objects.create(name='Legacy')
taggeditem = TaggedItem.objects.create(content_type=ct, object_id=666, tag=tag)
orphan = Tag.objects.create(name='An Orphan Tag')
```

and also ran the tag counts from [bug 1239756](https://bugzilla.mozilla.org/show_bug.cgi?id=1239756#c2):

```
count_data = TaggedItem.objects.values('content_type_id').annotate(Count('content_type_id'))
for data in count_data:
    ct_id = data['content_type_id']
    tag_count = data['content_type_id__count']
    ct = ContentType.objects.get(id=ct_id)
    is_model = ct.model_class() is not None
    print("%s (%s): %d" % (ct.model, "is model" if is_model else "NOT model", tag_count))
```

Before running the migrations:
```
userprofile (NOT model): 33129
filter (is model): 35
user (is model): 8087
```
``
After running the migrations, the test items were removed, and the new counts are:
```
filter (is model): 35
user (is model): 8087
```

``TaggedItem`` drops from 41252 to 8122.

It's relatively fast, but I think I am still going to apply the changes manually in production, and then use the migration as final cleanup.

jwhitlock:
I've manually run cleanup in staging and production, so the first migration should become a no-op migration.  My code is at https://gist.github.com/jwhitlock/20eb434466bab9b192a23476cf03a8ce .

The second migration will drop the unused tags.

We're due for a migration squash for the new year.

PR #4048, opened 2016-11-21
bug 1218563: Add docs for resetting database

jwhitlock:
When the database is corrupted, the named volume must be removed and recreated. Add how-to instructions to the troubleshooting docs.

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -18,14 +18,43 @@ These commands will reset your environment to a "fresh" version, while
 retaining the database::
 
   cd /path/to/kuma
-  docker-compose stop
-  docker-compose rm
+  docker-compose down
   make clean
   git submodule sync --recursive && git submodule update --init --recursive
   docker-compose pull
   docker-compose build --pull
   docker-compose up
 
+
+Reset a corrupt database
+------------------------
+The Kuma database can become corrupted if the system runs out of disk space,
+or is unexpectedly shutdown. MySQL can repair some issues, but sometimes you
+have to start from scratch. When this happens, pass an extra argument
+``--volumes`` to ``down``::
+
+  cd /path/to/kuma
+  docker-compose down --volumes
+  make clean
+  git submodule sync --recursive && git submodule update --init --recursive
+  docker-compose pull
+  docker-compose build --pull
+  docker-compose up -d mysql
+  sleep 20  # Wait for MySQL to initialize. See notes below
+  docker-compose up
+
+The ``--volumes`` flag will remove the named MySQL database volume, which will
+be recreated when you run ``docker-compose up``.
+
+The ``mysql`` container will take longer to start up as it recreates an empty
+database, and the ``kuma`` container will fail until the ``mysql`` container
+is ready for connections. The 20 second ``sleep`` should be sufficient, but
+if it is not, you may need to cancel and run ``docker-compose up`` again.
+
+Once ``mysql`` is ready for connections, follow the
+:doc:`installation instructions <installation>`, starting at
+:ref:`Provisioning a database <provision-the-database>`,
+to configure your now empty database.
 
 Run alternate services
 ----------------------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -60,6 +60,8 @@ command accordingly, or force the ``kuma`` name with::
 
         docker-compose up -d -p kuma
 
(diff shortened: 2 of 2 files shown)


jgmize:
Travis is failing due to a hash mismatch on the `click` dependency, not due to any changes in this PR:

> THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.

    click==6.6 from https://pypi.python.org/packages/1c/7c/10b4132dd952b6a04e37626258825b8aa8c1eb99545f2eb26a77c21efb55/click-6.6-py2.py3-none-any.whl#md5=ded5bf44698fd7def254d5cb9e5ad0b3 (from -c requirements/constraints.txt (line 72)):

        Expected sha256 cc6a19da8ebff6e7074f731447ef7e112bd23adf3de5c597cf9989f2fd8defe9

             Got        fcf697e1fd4b567d817c69dab10a4035937fe6af175c05fd6806b69f74cbc6c4

jwhitlock:
[click](https://pypi.python.org/pypi/click/6.6) uploaded a new wheel to PyPi 😦

docs/troubleshooting.rst
```diff
@@ -24,8 +24,37 @@ retaining the database::
   git submodule sync --recursive && git submodule update --init --recursive
   docker-compose pull
   docker-compose build --pull
-  docker-compose up
+  docker-compose up -d
```
safwanrahman: I think its better to not add ``-d`` in developer environment until its necessary. So it will be easy to see the logs

docs/troubleshooting.rst
```diff
@@ -24,8 +24,37 @@ retaining the database::
...
 
+Reset a corrupt database
+------------------------
+The Kuma database can become corrupted if the system runs out of disk space,
+or is unexpectedly shutdown, and MySQL may not be able to recover it. When this
+happens, you can recover by resetting the database as well::
+
+  cd /path/to/kuma
+  docker-compose stop
```
safwanrahman: So I have faced similar problem while I was working with the test fixing, I think we can use ``docker-compose down -v`` in order to stop the container, remove the container, volumes and network created by docker-compose.

safwanrahman: The documentation of it can be found here
https://docs.docker.com/compose/reference/down/

jwhitlock: Good catch, I thought named volumes were excluded by ``-v``. I'll test when I can stand to lose my MySQL volume.

docs/troubleshooting.rst
jwhitlock: I go back and forth. I find the full logs to be distracting, so instead I use ``make up`` which runs ``docker-compose up -d``, and then run ``docker-compose logs -f worker`` (or a similar command) when I need to watch the logs from a particular container.

This seems to be a 1 to 1 vote, so I'll go back to the previous version which was reviewed and accepted.

jwhitlock:
Rewrote instructions to use the more useful ``docker-compose down`` command, rebased on master.

docs/troubleshooting.rst
```diff
@@ -18,15 +18,44 @@ These commands will reset your environment to a "fresh" version, while
...
+have to start from scratch. When this happens, pass an extra argument
+``--volumes`` to ``down``::
+
+  cd /path/to/kuma
+  docker-compose down --volumes
+  make clean
+  git submodule sync --recursive && git submodule update --init --recursive
+  docker-compose pull
+  docker-compose build --pull
+  docker-compose up -d mysql
```
safwanrahman: So in local development I think its not needed to start the ``mysql`` container before.

docs/troubleshooting.rst
```diff
@@ -18,15 +18,44 @@ These commands will reset your environment to a "fresh" version, while
...
+  sleep 20
```
safwanrahman: Its good idea to run this in the script. but in documentation, it will just make the user confuse.

docs/troubleshooting.rst
jwhitlock: I have to disagree with you on this one.  I'm optimizing for people that are copying the commands from the documentation and pasting them into their terminal, running it like a script.  People that read (are there any left?) can try to understand the commands and adjust.

PR #4050, opened 2016-11-22
Bug 1318821 - Fix syntaxbox to use <pre>

a2sheppy:
Fix bug 1318821 — Change CKEditor config so the toolbar option to create a Syntax Box creates a ``<pre class=“syntaxbox”>`` instead of a ``<div class=“syntaxbox”>``. This makes it actually work right. While at it, I normalized the capitalization of the titles of options in the styles menu.


stephaniehobson:
R+

PR #4052, opened 2016-11-24
Punctuation of Using New Theme Steps.

Clusker:
The steps are sentences that use punctuation, so full stops must also be considered in my opinion. Pretty insignificant pull request, but worth a mention.


docs/documentation.rst
```diff
@@ -108,12 +108,12 @@ layout, and use the browser's inspector to look for 404s on static assets.
...
-To use the new theme in the public documentation,
+To use the new theme in the public documentation:
 
-1. Commit the new template to ``sphinx-theme``, open a pull request
-2. Remove the clone of the ``sphinx-theme`` from your Kuma repo (Docker method)
-3. After review, merge the PR to the master branch of ``sphinx-theme``
-4. Tag and publish a new version of ``mdn-sphinx-theme`` to PyPI_
-5. Update ``requirements/docs.txt`` in ``kuma``, merge to master
+1. Commit the new template to ``sphinx-theme``, open a pull request.
+2. Remove the clone of the ``sphinx-theme`` from your Kuma repo (Docker method.)
```
jwhitlock: Move this period after the parenthesis

Clusker: Hello,

Sorry for my rash requests. I wasn't aware of a writing style guide - my new request is also probably wrong, I am a little lost in the world of contribution. Apologies for the hand holding, but I appreciate your patience.

If you may cancel that request so I can move the period that would be great.

Sorry once again!

jwhitlock: >> If you may cancel that request so I can move the period that would be great.

I'm not sure what you are asking. You should be able to add a new commit to move the period, or close the pull request if you want to start over from scratch.

I think this is an OK change, once the period is moved.

Clusker:
I took the time to fully understand the process this time around. Here's to the start of my contribution career! Should be good to go, thanks for the patience. If you have any other requests or any other work that I could look at I'd be happy to help.

Thanks,
Sam

jwhitlock:
Thank you for your changes, and thanks for sticking with the process!

PR #4053, opened 2016-11-25
fixing possible typo

Rishiahya:
my team believe we may have fixed some possible typos, please let us know if we are correct.


docs/conf.py
```diff
@@ -3,7 +3,7 @@
 # Kuma documentation build configuration file, created by
 # sphinx-quickstart on Mon Aug  5 15:41:51 2013.
 #
-# This file is execfile()d with the current directory set to its containing dir.
+# This file is execfile'd with the current directory set to its containing dir.
```
jwhitlock: This refers to the the Python [execfile()](https://docs.python.org/2/library/functions.html#execfile) function. I don't think there's a grammatically correct way to say this. It should also be fixed [upstream](https://github.com/sphinx-doc/sphinx/blob/f3118e1cd6ecb428b0dc56300e2913fd86ccefa3/sphinx/templates/quickstart/conf.py_t#L9), not here.

docs/conf.py
```diff
@@ -101,8 +101,7 @@
 html_theme_path = [mdn_theme.get_theme_dir()]
 
 # Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
+# further. For a list of options available for each theme, see the documentation.
```
jwhitlock: Should be fixed [upstream](https://github.com/sphinx-doc/sphinx/blob/f3118e1cd6ecb428b0dc56300e2913fd86ccefa3/sphinx/templates/quickstart/conf.py_t#L100), not here.

docs/conf.py
```diff
@@ -150,7 +149,7 @@
 # If true, the index is split into individual pages for each letter.
 #html_split_index = False
 
-# If true, links to the reST sources are added to the pages.
+# If true, links to the rest sources are added to the pages.
```
jwhitlock: [reStructuredText](https://en.wikipedia.org/wiki/ReStructuredText) is sometimes abbreviated as RST, ReST or reST. "rest" is not one of the common abbreviations.

Rishiahya:
I believe we have removed the docs.py commit, so should be ok to be merged now

Rishiahya:
We have gone through the rest of the documents, and hope you find them adequate to be merged successfully.

jwhitlock:
Very nice copy editing job!

PR #4054, opened 2016-11-28
bug 1218563: Make build-base in Jenkinsfile and chown kuma

jgmize:
Remove default deis deploy


jwhitlock:
Looks good, builds locally. Let's see if Jenkins likes it.

PR #4055, opened 2016-11-29
bug 1218563: fix make compose-test

jgmize:
Map user to $UID in docker-compose
Update Jenkinsfile syntax to eliminate deprecation warnings

https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/job/fix-compose-test/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ env:
         - TOXENV=py27
         - TOXENV=flake8
         - TOXENV=docs
-        - TOXENV=locales DOCKER_COMPOSE_VERSION=1.8.0
-        - TOXENV=docker DOCKER_COMPOSE_VERSION=1.8.0
+        - TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
+        - TOXENV=docker DOCKER_COMPOSE_VERSION=1.9.0
     global:
         - DJANGO_SETTINGS_MODULE=kuma.settings.travis
         - DEBIAN_FRONTEND=noninteractive
@@ -26,8 +26,8 @@ env:
         - CFLAGS=-O0
 matrix:
     allow_failures:
-        - env: TOXENV=locales DOCKER_COMPOSE_VERSION=1.8.0
-        - env: TOXENV=docker DOCKER_COMPOSE_VERSION=1.8.0
+        - env: TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
+        - env: TOXENV=docker DOCKER_COMPOSE_VERSION=1.9.0
 before_install:
     - scripts/travis-install
     - pip install -U pip
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,32 +7,37 @@ node {
 
     switch (env.BRANCH_NAME) {
       case 'master':
-        env.DEIS_APP = 'mdn-dev'
+        stage('Build base') {
+          sh 'make build-base VERSION=latest'
+        }
 
-        stage 'Build base'
-        sh 'make build-base VERSION=latest'
+        stage('compose-test') {
+          sh 'make compose-test TEST=noext' // "smoke" tests with no external deps
+          sh 'make compose-test TEST="noext make build-static"' // required for many tests
+          sh 'docker-compose build'
+          sh 'make compose-test'
+        }
 
-        stage 'compose-test'
-        sh 'make compose-test TEST=noext' // "smoke" tests with no external deps
-        sh 'make compose-test TEST="noext make build-static"' // required for many tests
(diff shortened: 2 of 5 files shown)


jwhitlock:
On Docker for Mac, running "docker-compose up" results in ``"WARNING: The UID variable is not set. Defaulting to a blank string."``.  Should we use compose format 2.1 and add a default , like ``${UID:-0}``?

jgmize:
> On Docker for Mac, running "docker-compose up" results in "WARNING: The UID variable is not set. Defaulting to a blank string.". Should we use compose format 2.1 and add a default , like ${UID:-0}?

@jwhitlock just to be clear, `docker-compose up` still works for you despite the warning, correct? I added a default so that `make up` will set the env var for you if you want to easily eliminate the warning (and type less ;)

We can't switch to the docker-compose 2.1 format quite yet, as it requires [docker-compose 1.9](https://github.com/docker/compose/releases/tag/1.9.0) so we'll need to set that up in Jenkins first, and we might also want to give devs a little time to upgrade to that release. 

All that said, I'm glad they finally added the ability to set defaults directly in the new docker-compose format, it's a feature I've been wanting for quite a while now, and will happily switch to it as soon as is reasonable.

Makefile
```diff
@@ -19,6 +19,7 @@ target = kuma
 requirements = -r requirements/local.txt
 # set Django settings module if not already set as env var
 export DJANGO_SETTINGS_MODULE ?= kuma.settings.testing
+export UID ?= $(shell id -u)
```
safwanrahman: Only ``(id -u)`` should work?

jgmize: Not in a Makefile. See https://www.gnu.org/software/make/manual/make.html#Shell-Function for more details.

jwhitlock:
It works... but it is weird.  If I run ``docker-compose up``, the containers come up with the warning.  If I then run ``make bash`` in a second terminal, the containers are recreated (the environment variable ``UID`` changed), the containers are restarted, and my prompt is now::

    I have no name!@42c9030330f8:/app$

So, it works, but means that I need to exclusively use ``make`` commands, or ``docker-compose`` directly, but not mix the two.  I'm afraid this is going to invalidate a lot of the documentation.  I'm also not sure if ``make`` is a viable option for Windows users (we're getting interest with the move of KumaScript macros to GitHub).

jwhitlock:
@safwanrahman have you tried this PR? Does this config have any effect on your workflow?

In my mind, I'm coming up with a matrix of what each host OS needs, and I suspect a universal ``docker-compose`` configuration is not possible.

safwanrahman:
@jwhitlock So I have tried this patch and it works perfectly. Only the warning.
It does not affect my workflow because as empty string is passed in the ``user``, the default user ``kuma`` is used.

So after reading the documentation and thinking over the night, I think its high time we move to ``docker-compose`` 2.1 version where we can use ``${UID:-0}``. Because today or tomorrow we need to move there and it will be trouble if we change the compose version when many people are continuously developing kuma.

jwhitlock:
I don't think we should block on docker-compose 1.9.

I'm more convinced we should be running as a non-root user after reading a [recent blog post](http://blog.dscpl.com.au/2016/12/what-user-should-you-use-to-run-docker.html?m=1).  Our goal is to deploy these images in Kubernetes, not just develop on them, so we need to tackle these issues.  I'm not sure pulling the user ID from the environment is the right thing (a known UID and GID seems better), but I'm also not very educated on the issues. I'd like to leave our all-hands meeting with an agreement on how to move forward.

For this PR, would it make sense to move the ``USER`` statements as ``web`` etc. overrides to ``docker-compose.test.yml``?  It's worse, but it avoids breaking the docs or having to include a doc update in this PR.

jgmize:
Upgrading our jenkins to use docker-compose 1.9 isn't a big deal as we're already running docker 1.12.3 there. The bigger annoyance, I think, is communicating to all devs with an older docker-compose that they will need to upgrade, but it's probably a smaller annoyance than the issues you described with the current solution. 

We can't use a "known UID/GID" in the container b/c we need to pull the UID from the host to eliminate the permissions issues on mounted volumes in Linux.

jgmize:
I've upgraded docker-compose to 1.9 on our CI and amended the commit to use the new docker compose syntax to default to 0 (`${UID:-0}`).

jgmize:
The latest build passed in https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/job/fix-compose-test/

jwhitlock:
Thanks @jgmize and @safwanrahman for talking me through this one. One step closer to the production pipeline!

PR #4056, opened 2016-11-29
Fix expand/collapse display bugs.

stephaniehobson:
Fix bug 1319927 - Add styling to summary/details that Firefox will display
Fix bug 1259287 - Expand/collapse indicators missing from translation interface headings
Fix bug 1304283 - Change +/- to triangle

Standardize on triangles for expand/collapse, no more plus/minus.

Remove Firefox and Chrome default styling and add back custom styling.

Make custom styles better so custom styling for different situations is less necessary. Remove custom details/summary styling for translation, details-list, and no-details.

Browsers do not let us target the auto generated summary elements as they are part of the Shadow DOM.

Relies on browsers to support styling unrecognized elements for styles to apply when fallback shim in use. Will need to be tested in Edge on staging.

Test pages:
https://developer-local.allizom.org/fr/docs/User:stephaniehobson:test$edit (or any translation page) 
https://developer-local.allizom.org/en-US/fellowship
https://developer-local.allizom.org/en-US/docs/User%3Astephaniehobson%3Adetails
https://developer-local.allizom.org/en-US/docs/User:stephaniehobson:details$edit


stephaniehobson:
@schalkneethling R?

stephaniehobson:
Tested the shim in IE10 and it's working as intended 🎉

schalkneethling:
So, in Chrome I see the following:

screen shot 2016-12-01 at 12 41 04

But in Firefox (Release and Dev), I see the following:

screen shot 2016-12-01 at 12 41 52

This is after clearing cache as well as restarting the Docker containers :-/

schalkneethling:
Also, is it expected that the Django Debug Toolbar is on and visible be default when running over Docker?

schalkneethling:
Oh, I reckon I need to rebuild static assets.

schalkneethling:
When I access this page:
en-US/docs/User%3Astephaniehobson%3Adetails and then click to add translation

screen shot 2016-12-01 at 13 27 20

I then reach an intermediary page where I select a language. Once I selected a language, I then get to the edit page, which look as follows:

screen shot 2016-12-01 at 13 27 57

A you can see here, the arrows are in the expanded state even though the content it not expanded. Clicking on one of the "headings", produces the following result:

screen shot 2016-12-01 at 13 28 07

Not sure if this was supposed to be resolved by this PR or, whether something in this PR introduced the problem but, thought I would mention it.

stephaniehobson:
I think the debug tool bar was recently changed to run by default. There's a way to disable it but I don't know how to do it in Docker yet.

Did rebuilding the assets fix the missing arrows in Firefox? (that's the bug I'm trying to fix with this PR)

I'm not seeing the open arrows, was that in Firefox as well?

schalkneethling:
> Did rebuilding the assets fix the missing arrows in Firefox? (that's the bug I'm trying to fix with this PR)

Yup, that fixed it.

> I'm not seeing the open arrows, was that in Firefox as well?

Yep.

schalkneethling:
@stephaniehobson Just confirmed, I can definitely replicate in Firefox.

stephaniehobson:
/me understood the problem immediately once she followed the steps to reproduce...

One character fix!

PR #4058, opened 2016-11-30
bug 1307789: Add Kabyle language

jwhitlock:
Add [Kayble](https://wiki.mozilla.org/L10n:Teams:kab) to the accepted MDN languages.  The important change is in ``kuma/settings/common.py``, which adds it to the list in ``jinja2/includes/translate_locales.html``, which then cascades into all the localization ``django.po`` files.

Azul si MDN!


jinja2/includes/translate_locales.html
```diff
@@ -32,6 +32,7 @@
 {{ _('Irish') }}
 {{ _('Italian') }}
 {{ _('Japanese') }}
+{{ _('Kabyle') }}
```
safwanrahman: First time its being done! :+1:

safwanrahman:
does any changes needed from the admin panel?

jwhitlock:
No, no changes are needed in the admin panel.

safwanrahman:
@jwhitlock Can you separate the PR into two commits, the code changes and the locale file changes?
So it can be used as a reference for adding a locale.

jinja2/includes/translate_locales.html
jwhitlock: Well, kind of.  I had to work around the [documented process](https://kuma.readthedocs.io/en/latest/localization.html#adding-a-new-locale).

The [new "add a locale" process](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Localize/Starting_a_localization) is:

1. Add the language as a localizable UI language
2. Wait for the team to localize 50% of the UI in [Pontoon](https://pontoon.mozilla.org/projects/mdn/)
3. Recruit a MDN staff member or core content volunteer who will be the localization lead for that language
4. Enable the language in MDN (this PR)

The [documented process](https://kuma.readthedocs.io/en/latest/localization.html#adding-a-new-locale) assumes you add the language as a UI are adding a new locale, so it is out of date.  If you follow those instructions and run ``make locale LOCALE=kab``, then you get the updated ``jinja/includes/translate_locales.html``, but also wipe out already localized UI strings.

So, back when we added Kable as a translatable language for the UI, we should have also committed an updated ``jinja2/includes/translate_locales.html``.

I worked around it by calling ``./manage.py translate_locales_name``.

We'll try to do better with the next language.  But yeah, it will help to split the commit into "stuff we should have done 2 months ago" and "stuff for today".

PR #4059, opened 2016-11-30
Fix Bug 1321360: Homepage tile for foundation.

stephaniehobson:
Added homepage tile for foundation end of year campaign behind waffle switch.


escattone:
This all looks great and worked great for me!

PR #4062, opened 2016-12-01
bug 957802: Regenerate SSL certificates

jwhitlock:
Regenerate the SSL certificates used in the Vagrant development environment.

* Update `make generate` to generate certs without prompts
* Add `make print_cert` to get certificate details
* Regenerate the SSL certs to be active from Dec 1 2016 - Dec 1 2018.


jwhitlock:
Merging before we start getting rid of vagrant 😄

PR #4066, opened 2016-12-11
bug 957802: Doc - titles and lists standardized

Clusker:
Big overhaul of most lists, correcting inconsistencies with the formatting as well as old style titles and headers converting to sentence-style capatalization.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,10 @@
-Contributing Code
+Contributing code
 =================
 
 Things to know when contributing code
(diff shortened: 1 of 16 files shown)


jwhitlock:
We've made some merges that are causing conflicts that can only be resolved on the command line.  I'm guessing that this would be difficult for you, so I did the rebase and published it as pull request #4068.  There are a couple of instances where I disagreed on punctuation, but 95% of the changes are in there.  It also shows you as the author of the changes.

PR #4067, opened 2016-12-14
Bug 1318827 - Update spell check defaults for MDN editor

a2sheppy:
Fix bug 1318827 by making the following enhancements to the CKEditor configuration:

* The as-you-type spell checker (SCAYT) is now enabled by default.

* The spell checker now ignores words which are all-caps or are in CamelCase; this will prevent API terms from being flagged as errors much of the time.

* The spell checker now supports languages other than English. Not all MDN locales have dictionaries, however. See the "Dictionary" tab in the SCAYT options dialog.

* The context menu's options have been tweaked slightly to remove some unneeded SCAYT items from the list.

There's some code in here that fiddles with the locale name string since MDN uses a different format for them than the SCAYT plugin expects. We should be in good shape for the supported locales.

### DIFF
--- a/kuma/wiki/jinja2/wiki/ckeditor_config.js
+++ b/kuma/wiki/jinja2/wiki/ckeditor_config.js
@@ -60,7 +60,7 @@
     config.toolbarGroups = [
       { name: 'document', groups: [ 'mode', 'document', 'doctools' ] },
       { name: 'clipboard', groups: [ 'clipboard', 'undo' ] },
-      { name: 'editing', groups: [ 'find', 'selection', 'spellchecker'] },
+      { name: 'editing', groups: [ 'find', 'selection', 'spellchecker' ] },
       { name: 'forms' },
       { name: 'tools' },
       { name: 'others' },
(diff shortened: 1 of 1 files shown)


stephaniehobson:
Adding *manual merge* label as @jwhitlock would like to launch these changes in a controlled test because of concerns about API restrictions.

a2sheppy:
I've made the requested changes. Should be good to go assuming tests pass. But local manual testing is good.

kuma/wiki/jinja2/wiki/ckeditor_config.js
```diff
@@ -90,6 +90,41 @@
     // Don't use HTML entities in the output except basic ones (config.basicEntities).
     config.entities = false;
 
+    // Enable the spell checker by default
+    config.scayt_autoStartup = true;
```
jwhitlock: I'm considering a run-time switch to turn this on or off, because I'm worried we will hit an API limit with SKAYT.

jwhitlock:
I've been staring at this for a while, and I think we can merge everything but the ``config.scayt_autoStartup = true;``, and start a discussion on this plugin.

When you enable the plugin, it makes a request to download the full JS, and executes it:

https://svc.webspellchecker.net/spellcheck31/lf/scayt3/ckscayt/ckscayt.js

It also loads a small stylesheet, and starts POSTing data to the CGI endpoint:

https://svc.webspellchecker.net/spellcheck31/script/ssrv.cgi

The data is form-encoded, such as:

```
cmd=getbanner&customerid=1%3AWvF0D4-UtPqN1-xxxxxx&format=json
```

It uses local storage to remember some things, like misspelled words and their suggestions, to avoid traffic to the server.

Currently, this adds an ad for the SCAYT service itself when you right-click on a word, linking to the [license page](https://www.webspellchecker.net/signup/licensed-signup.html#wsc-trial). After more usage, it may begin to show other ads, including ones we don't want to present to our users.  This risk goes way up if we auto-enable the plugin.

I'll open a new rebased PR that includes everything but the auto-start, and start a discussion w/ management about getting a license.

Elchi3:
> and start a discussion w/ management about getting a license.

I guess this discussion will happen on dev-mdc or somewhere open anyway, but my two cents here are: 

Let's not invest in a licence of proprietary spellcheck software and instead remove it entirely. If we want to invest money, lets give to it CKEditor, so that many web developers and users of it can benefit from enhanced native spellcheck. Piotrek Koszuliński (lead developer CKEditor) has done a lot of work on Kuma, too. So we have a contact there.

jwhitlock:
Replaced by PR #4118.

a2sheppy:
I concur. Let’s just enact the changes in this PR without the autoStart switched on, and start looking at how we can transition to using browsers’ built-in spell checking without losing functionality in the context menu.

What I’d love to do is find a way to move CKEditor’s contextual options into the standard HTML context menu. We should “encourage” CKEditor to make that transition.

PR #4071, opened 2016-12-19
bug 957802: Update links to PRs, What's Deployed

jwhitlock:
These sites moved from paas.allizom.org to new homes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,18 +29,22 @@ Development
 :Code:          https://github.com/mozilla/kuma
 :Issues:        http://mzl.la/mdn_backlog (Product)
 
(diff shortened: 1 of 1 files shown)


stephaniehobson:
I get a certificate warning from `https://prs.mozilla.io/mozilla:kuma,kuma-lib,kumascript,mozhacks`. If that's the correct URL it shouldn't block this PR.

stephaniehobson:
Also 👍 for having text to explain what the URLs are. Much more readable than linking URLs.

jwhitlock:
prs.mozila.io is a GitHub site, and uses that certificate.  I'll switch to the non-SSL link.

PR #4076, opened 2016-12-20
Bug 1271509: Sample database and production scraper

jwhitlock:
Now that PR #4245 and PR #4248 are merged, this PR is getting a little more sane:

* Adds a management command ``scrape_links`` to find document links on an MDN page. It is designed for the homepage, but can work on other pages as needed.
* Adds a management command ``sample_mdn`` that loads fixtures and scrapes production data from a JSON spec.
* Adds a script to create an empty but migrated database, load it with fixtures and scraped data, and export a SQL dump for development and testing. Since the data is scraped from public interfaces, avoids issues with exposing personally identifiable information.

A recent generated sample database is at http://mdn-downloads.s3-website-us-west-2.amazonaws.com/mdn_sample_db.sql.gz

To Do:
* [x] Remove unneeded stuff like ``./manage.py sample_templates`` and debug print statements that are now obvious once I've opened the PR.
* [x] Split up big files for sanity and better GitHub review
* [x] Split up long ``gather`` methods into sane methods
* [x] Update documentation to use sample database
* [x] Improve S3 download site
* [x] Set valid test account passwords for integration tests
* [x] Ensure [fr/docs/User:anonymous:uitest](https://developer.mozilla.org/fr/docs/User:anonymous:uitest) is in sample DB
* [x] Include Mozilla Hacks blog scrape
* [x] *Maybe* Add en-US/docs/Apps
* [x] Add documentation for sampling data, generating the database
* [x] Allow scraping a tree of documents, like /en-US/Web/HTML
* [x] Change management command ``scrape_homepage`` to a generic ``scrape_links``.
* [x] Add "less verbose" mode for cron-based sample DB generation
* [x] Set Site 1 to localhost:8000 from example.com
* [x] Add URLs in tests/performance/smoke.py to the sample DB

Not this PR:
* Extract attachment data from documents
* Render documents and set the render date of documents to avoid rebuild of sample DB on load

### DIFF
--- a/kuma/scrape/fixture.py
+++ b/kuma/scrape/fixture.py
@@ -12,8 +12,8 @@ class FixtureLoader(object):
 
     # Needed information about the supported Django models for fixtures
     # The key is the app_label.model_name, such as wiki.revision for Revision:
-    #  Revision._meta.app_label == wiki
-    #  Revision._meta.model_name == revision
+    #  Revision._meta.app_label == 'wiki'
+    #  Revision._meta.model_name == 'revision'
     # The value is a dictionary of:
     # - natural_key: Properties used to find existing database records
     # - relations: Details of properties that are foreign keys
@@ -155,8 +155,7 @@ class FixtureLoader(object):
             natural_key_spec = metadata['natural_key']
             relations = metadata.get('relations', {})
             filters = metadata.get('filters', {})
-            app_label, model_name = model_id.split('.')
-            assert apps.get_model(app_label, model_name)
+            assert apps.get_model(model_id)
 
             parsed.setdefault(model_id, [])
             for item_num, item in enumerate(items):
@@ -227,8 +226,7 @@ class FixtureLoader(object):
         existing, loaded, pending = 0, 0, 0
         for model_id, items in self.spec.items():
             metadata = self.model_metadata[model_id]
-            app_label, model_name = model_id.split('.')
-            Model = apps.get_model(app_label, model_name)
+            Model = apps.get_model(model_id)
 
             self.instances.setdefault(model_id, {})
             for item in items:
--- a/kuma/scrape/fixture.py
+++ b/kuma/scrape/fixture.py
@@ -19,6 +19,15 @@ class FixtureLoader(object):
     # - relations: Details of properties that are foreign keys
     # - filters: Methods to run on values before saving to the database
     model_metadata = {
+        'account.emailaddress': {
+            'natural_key': ('email',),
+            'relations': {
+                'user': {
+                    'link': 'to_one',
+                    'resource': 'users.user',
+                },
+            },
+        },
         'auth.group': {
             'natural_key': ('name',),
             'relations': {
@@ -70,6 +79,27 @@ class FixtureLoader(object):
         },
         'search.filtergroup': {
             'natural_key': ('name', 'slug'),
+        },
+        'sites.site': {
+            'natural_key': ('id',),
+        },
+        'socialaccount.socialaccount': {
+            'natural_key': ('uid', 'provider'),
+            'relations': {
+                'user': {
+                    'link': 'to_one',
+                    'resource': 'users.user',
+                }
+            }
+        },
+        'socialaccount.socialapp': {
+            'natural_key': ('name', ),
+            'relations': {
+                'sites': {
+                    'link': 'to_many',
+                    'resource': 'sites.site'
+                }
+            }
         },
         'taggit.tag': {
             'natural_key': ('name',),
--- a/kuma/scrape/fixture.py
+++ b/kuma/scrape/fixture.py
@@ -250,8 +250,8 @@ class FixtureLoader(object):
         """
         Attempt to create or update an item.
 
-        Returns the instance if attempt suceeded, or None if related items are
-        needed.
+        Returns the instance if attempt suceeded.
+        Raises NeedsDependency if a dependency must be created first.
         """
         natural_key_spec = metadata['natural_key']
         relations = metadata.get('relations', {})
--- a/kuma/scrape/fixture.py
+++ b/kuma/scrape/fixture.py
@@ -1,0 +1,297 @@
+"""Load test fixtures from a specification."""
+
+from django.apps import apps
+from django.contrib.auth.hashers import make_password
+
+import logging
+logger = logging.getLogger('kuma.scraper')
+
+
+class FixtureLoader(object):
+    """Load fixtures into the current database."""
+
+    # Needed information about the supported Django models for fixtures
+    # The key is the app_label.model_name, such as wiki.revision for Revision:
+    #  Revision._meta.app_label == wiki
+    #  Revision._meta.model_name == revision
+    # The value is a dictionary of:
+    # - natural_key: Properties used to find existing database records
+    # - relations: Details of properties that are foreign keys
+    # - filters: Methods to run on values before saving to the database
+    model_metadata = {
+        'auth.group': {
+            'natural_key': ('name',),
+            'relations': {
+                'permissions': {
+                    'link': 'to_many',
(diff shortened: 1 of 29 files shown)


jwhitlock:
With #4245 merged, I've move the work-in-progress branch to https://github.com/jwhitlock/kuma/tree/sample_database_wip_long_1271509, and replaced this PR with the mergeable commits.  There's two milestone PRs:

* Adding ``./manage.py scrape_document``
* Adding ``./manage.py scrape_homepage`` and ``sample_mdn``

I'm continuing to work on the code.

jwhitlock:
I want to re-run the ``create_sample_db.sh`` and make sure everything still works, but I'll have to wait until Friday.  Still, I think the PR is ready for review.

jwhitlock:
I ran the scrape again, and hit two problems:

* A translation was deleted on the site, but the ``$json`` metadata for the English page still included it. When fetching the page, a 404 is returned.  This caused an immediate termination of scraping, rather than setting the ``document_rendered`` source to an error state.  The code needs updating.
* All pages show "This article doesn't have any content yet.".  The document's ``html`` field is an empty string.  This should be set when the current revision is saved, to the current revision's content. More debugging is needed.

jwhitlock:
I incorrectly thought that ``revision.save()`` was skipped when using ``Revision.objects.create``. I was thinking of bulk creation, like when loading a Django test fixture. Oops.

Most documents are now setup correctly.  There are exceptions (22 of 1213 documents):

* [Mozilla/Projects/Midas](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Midas) - has a revision, but not set to ``current_revision``, ``html`` is an empty string.
* [cs/docs/Tools](https://developer.mozilla.org/cs/docs/Tools) - This is a redirect to the English doc on MDN, and has no content or revisions in the scrape.

jwhitlock:
The issue with an empty ``document``.``current_revision`` is a quirk of page moves.  On [Mozilla/Projects/Midas$history](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Midas$history), the latest revision is the page move, and the second-to-latest revision is the "current revision".  If the scraper just scrapes the latest revision, it never sets ``current_revision``. The solution is a new source type ``document_current`` that checks that ``current_revision`` is set, and requests more revisions until it is.

Other issues were translations where a root document was renamed but the child documents had the old paths\.  For example [pl/docs/Narzędzia](https://developer.mozilla.org/pl/docs/Narzędzia) was the parent topic of [pl/docs/Tools/WebIDE](https://developer.mozilla.org/pl/docs/Tools/WebIDE), and the scraper had issues following the parent relations, attempting to get the history of a redirect.  None of the APIs expose the parent topic's URL, and scraper assumes you can remove everything after the last slash to get it.  In any case, I manually moved these child translations so that they matched their parent topic's slug.

I've now successfully scraped MDN with this code, and uploaded the sample DB to the downloads server.

If you want to try out the sample DB script, it takes me about 40 minutes to run, makes a little over 10,000 requests, and takes over your working copy while doing it. Plan your work day accordingly!  You can also try a more limited version by reducing the scraped items in ``etc/sample_db.json``, for example removing items, not scraping translations, etc.

jgmize:
> It's sad that more eyes won't review this, so you get the credit you deserve. This is great work with superb tests and test coverage (100% for every file under kuma/scrape). It deserves fanfare. 🎉 🍾

agreed 💯. I also hate to [break the tradition](https://mozilla.github.io/meao/2017/07/06/kuma-report/#ship-the-sample-database) but I'm looking forward to seeing this finally merge this month.

kuma/scrape/fixture.py
```diff
@@ -0,0 +1,327 @@
...
+        """Parse and validate the specification."""
+        parsed = {}
+        for model_id, items in specification.items():
+            # Parse and validate the model
+            metadata = self.model_metadata[model_id]
+            natural_key_spec = metadata['natural_key']
+            relations = metadata.get('relations', {})
+            filters = metadata.get('filters', {})
+            app_label, model_name = model_id.split('.')
+            assert apps.get_model(app_label, model_name)
```
jwhitlock: Thanks, this worked. I didn't know that ``apps.get_model`` was flexible that way.

kuma/scrape/management/commands/sample_mdn.py
```diff
@@ -0,0 +1,54 @@
...
+
+from kuma.scrape.fixture import FixtureLoader
+from . import ScrapeCommand
+
+
+class Command(ScrapeCommand):
+    help = 'Sample data from a running MDN server.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('spec',
```
jwhitlock: I had the same thought process. Default filenames with paths feels wrong as well, asking for Windows / Unix trouble. No default seemed the best course.

kuma/scrape/management/commands/sample_mdn.py
```diff
@@ -0,0 +1,54 @@
...
+                                  ' "developer.mozilla.org")'),
+                            default='developer.mozilla.org')
+        parser.add_argument('--nossl',
+                            help='Disable SSL',
+                            action='store_false',
+                            default='true',
+                            dest='ssl')
+
+    def handle(self, *arg, **options):
+        self.setup_logging(options.get('verbosity'))
```
jwhitlock: Good catch, I thought I got all of these...

kuma/scrape/management/commands/sample_mdn.py
```diff
@@ -0,0 +1,54 @@
...
+            if source.state != source.STATE_DONE:
+                if hasattr(source, 'error'):
+                    err = ' "%s"' % source.error
+                else:
+                    err = ''
+                self.stderr.write("%s: %s%s\n" % (name, source.state, err))
+                incomplete += 1
+        if incomplete:
+            raise CommandError("%d source%s incomplete." %
+                               (incomplete, 's' if incomplete != 0 else ''))
```
jwhitlock: Oops actually ``'' if incomplete == 1 else 's'``:

```
0 sources incomplete.
1 source incomplete.
2 sources incomplete.
```

kuma/scrape/management/commands/scrape_links.py
```diff
@@ -0,0 +1,40 @@
...
+        host, ssl, path = self.parse_url_or_path(options['url'])
+        scraper = self.make_scraper(host=host, ssl=ssl)
+
+        params = {}
+        for param in ('translations', 'revisions', 'depth'):
+            if options[param]:
+                params[param] = options[param]
+        scraper.add_source("links", path, **params)
+
+        scraper.scrape()
```
jwhitlock: Sounds good. I think I should standardize on ``CommandError`` in all the management commands.

kuma/scrape/management/commands/sample_mdn.py
escattone: Oh, good point!

escattone: Although I just realized that this path won't be taken when `incomplete == 0`.

kuma/scrape/sources/links.py
```diff
@@ -0,0 +1,99 @@
...
+        'users/signin',
+    ))
+
+    def __init__(self, path=None, **options):
+        """Process and validate the initial path."""
+        if (not path) or (path == '/'):
+            path = '/en-US/'  # Default to English homepage
+        path = urlparse(path).path
+        assert path.startswith('/')
+        self.locale = urlparse(path).path.split('/')[1]
```
jwhitlock: Yep, totally ~~redundent~~ redundant

escattone: 😆

jwhitlock:
Thanks! I'm proud of this code, but I don't want anyone, including me, to follow my example of working on a feature for over a year.  Small PRs, reviewed and merged often!

PR #4074, opened 2016-12-20
Bug 1314708: Add testing login credentials, add dashboard tests

stephaniehobson:
- testing usernames and passwords
- page object for logging into admin
- test "sign in" forwards to github
- tests for revision dashboard

Testing usernames will be part of sample database or can be configured manually on dev machines. They will not be on staging or production.

Use `-m "not login"` to exclude tests requiring a login.

Not all login tests are destructive.

PR #4077, opened 2016-12-20
Bug 1314708: Add functional tests for edit pages.

stephaniehobson:
Requires #4074.

### DIFF
--- a/tests/functional/test_article_edit.py
+++ b/tests/functional/test_article_edit.py
@@ -1,0 +1,47 @@
+import pytest
+
+from pages.article import ArticlePage
+from pages.admin import AdminLogin
+
+
+@pytest.mark.smoke
+@pytest.mark.nondestructive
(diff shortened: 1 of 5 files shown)


stephaniehobson:
This is ready except that it depends on another PR.

 @m8ttyB R? Also, do you have a dev environment for MDN running? That'll be necessary to run these but someone else on the team might be able to test they actually work leaving you to just review the code.

m8ttyB:
r+ with the caveat that I do not have running MDN instance locally. If no one else is available I can set this up and give the tests a quick spin next week.

stephaniehobson:
Merging this. We're not adding stuff to `pytest.ini` for now because it's strictly, um, unit(?) tests right now, not the functional tests and we don't want to mix.

There is still a test failure but it's not in a file that was edited in this PR.

PR #4079, opened 2016-12-21
1314708: Add tests for creating a page.

stephaniehobson:
Ready for testing.

Command to test on Vagrant, edit the URL to test in Docker: 

py.test tests/functional/test_article_new.py -v --driver Chrome --driver-path /Users/shobson/Downloads/chromedriver --base-url https://developer-local.allizom.org


escattone:
This adds one test (`test_new` within `tests/functional/test_article_new.py`) marked with a `login` pytest marker, so it's not included as part of the current integration test suite run on push to the "stage-integration-tests" branch. The `login` tests currently only run within a local development environment using the sample DB.

As it stands, the `test_new` test creates a new document every time it is run, and does NOT clean-up after itself (by deleting the document).

PR #4080, opened 2016-12-21
Fix bug 1325097: Make multi-column divs easier to see

a2sheppy:
Updated the twocolumns and threecolumns classes so they draw with a pale blue border around the columns when in the editor, for ease of layout work. I used a variable, $editor-box-border-style to contain the style information to apply to these column boxes, so it can be easily changed in one place, and potentially reused elsewhere for other "draw borders in the editor" situations that may arise.

### DIFF
--- a/kuma/static/styles/components/content.styl
+++ b/kuma/static/styles/components/content.styl
@@ -276,9 +276,19 @@ p.footnote {
 /* Classes created by the WYSIWYG */
 .twocolumns {
     vendorize(column-count, 2);
+
+    body[contenteditable] & {
+        border: $editor-box-border;
+        column-rule: $editor-box-border;
+    }
 }
 .threecolumns {
     vendorize(column-count, 3);
+
+    body[contenteditable] & {
+        border: $editor-box-border;
+        column-rule: $editor-box-border;
+    }
 }
 
 @media $media-query-small-mobile {
--- a/kuma/static/styles/includes/vars.styl
+++ b/kuma/static/styles/includes/vars.styl
@@ -232,3 +232,8 @@ $column-12, $column-all {
     margin: 0;
     float: none;
 }
+
+/*
+edit-mode customizations - applied to blocks only in edit mode
+====================================================================== */
+$editor-box-border = 2px dotted #ADF;


kuma/static/styles/components/content.styl
```diff
@@ -7,6 +7,13 @@
 
 $table-blue = #d4dde4;
 
+/*
+    Used to draw borders around objects which should only be bordered in
+    the editor.
+*/
+
+$editor-box-border-style = solid 1px #6699FF40;
```
stephaniehobson: Let's move this variable into kuma/styles/includes/vars.styl so we can use it everywhere!

Remove 'style' from the variable name (I don't want to imply the value of the variable is just the style).

Border short hand should be in the order width, style, colour.

Please use a hex or rgba colour format.

kuma/static/styles/components/content.styl
```diff
@@ -276,15 +283,29 @@ p.footnote {
...
+        border: $editor-box-border-style;
+        column-rule: $editor-box-border-style;
+    }
 }
 
 @media $media-query-small-mobile {
    .twocolumns,
    .threecolumns {
         vendorize(column-count, 1);
+        body[contenteditable] & {
```
stephaniehobson: You do not need to define the border styles for the media query, they will be inherited from the desktop styles.

kuma/static/styles/components/content.styl
a2sheppy: Good idea -- although I don't like the idea of changing the style used for .hidden. It needs to have a unique look because it's a special case of showing you something that normal readers will not see. I'll work on this other stuff tomorrow though.

a2sheppy: I could, however, make .hidden use its own variable in vars.styl so that it could theoretically be reused for other stuff (and to group these special bits of CSS in one place).

a2sheppy: Hm. .hidden is very large to put into a variable like that. Not sure how well it would work. Do you really think it's worth doing? I kinda don't.

a2sheppy:
I've applied most of these changes (except the .hidden stuff, which I have thoughts on above). I also changed the format of the borders on the columns to be a little nicer looking. I suggest that if we want to move .hidden's format to a variable,  or otherwise change it, it should be a separate issue.

kuma/static/styles/includes/vars.styl
```diff
@@ -232,3 +232,8 @@ $column-12, $column-all {
     margin: 0;
     float: none;
 }
+
+/*
+edit-mode customizations - applied to blocks only in edit mode
+====================================================================== */
+$editor-box-border = 2px dotted rgba(170, 221, 255, 255);
```
stephaniehobson: 255 is not a valid 4th parameter for rgba. If you intended to use `1` please convert this to a hex value.

stephaniehobson:
Thanks for making the changes. 👍 once the rgba thing is sorted.

You're right, hidden should look different than this stuff because this stuff serves a different function.

kuma/static/styles/includes/vars.styl
jwhitlock: If you meant "no translucency", the hex value would be ``#ADF``, which is a light blue.

a2sheppy:
Changed the color to #ADF as suggested (which it had been originally until I got too creative for my own good).

jwhitlock:
Waiting on Stylus -> Sass change, then we'll rebase and merge

jwhitlock:
Replaced by PR #4115, which is a rebase after the Stylus → Sass change

PR #4081, opened 2016-12-21
Fix bug 1325125: expand blocked protocols regex to include preceding spaces

claudijd:
Expand regex to match on preceding spaces

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1185,7 +1185,7 @@ CONSTANCE_CONFIG = dict(
         "are removed from the file storage"
     ),
     KUMA_WIKI_HREF_BLOCKED_PROTOCOLS=(
-        '(?i)^\s*(data\:?)'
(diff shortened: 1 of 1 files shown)


kuma/settings/common.py
```diff
@@ -1185,7 +1185,7 @@ def _get_locales():
         "are removed from the file storage"
     ),
     KUMA_WIKI_HREF_BLOCKED_PROTOCOLS=(
-        '(?i)^(data\:?)',
+        '(?i)^\s*(data\:?)'
```
jgmize: Missing `,` after regex.

claudijd: Fixed in 4b6deb3a31f1a993019eddd9fae9b099b8be000f

jwhitlock:
Replaces with PR #4083. @claudijd is still credited, but includes tests and expanded matching in the ``trackOutboundLinks`` JS

PR #4084, opened 2017-01-03
bug 1218563: Pass UID to tox docker build

jwhitlock:
This fixes the docker build in TravisCI (hopefully).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ env:
 matrix:
     allow_failures:
         - env: TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
-        - env: TOXENV=docker DOCKER_COMPOSE_VERSION=1.9.0 UID=0
 before_install:
     - scripts/travis-install
     - pip install -U pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
         - TOXENV=flake8
         - TOXENV=docs
         - TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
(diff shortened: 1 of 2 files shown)


safwanrahman:
When the docker build broken? ❗️ ⏰

jwhitlock:
PR #4078 broke the [docker build](https://travis-ci.org/mozilla/kuma/jobs/185577329). TravisCI doesn't set the UID, so it was defaulting to UID=0 before, and now is defaulting to UID=33 and can't create a needed directory.

safwanrahman:
Seems like need some further investigation!
We should not merge #4078 while the docker break!

safwanrahman:
@jwhitlock Do you think we should make the docker test mandatory right after it get fixed? currently it is under `` allow_failures``

jwhitlock:
I'm guessing that you meant "we should not have merged #4078 because it broke the docker build."  You are probably right, this change should have been part of that one.  Making the docker test mandatory would have made that clear.  The docker build feels like it has been stable for a while.

At the same time, the build worked in Jenkins (https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/) which is the environment that matters for Docker builds, and it was fine for development for the last 14 days. TravisCI seems to be the _only_ environment that was broken. This suggests it wasn't a big deal that the TravisCI docker build was broken.

jwhitlock:
@jgmize and I had a chance to talk. He also doesn't have a strong opinion if the docker build should be required or not.  I marked it as required before he merged it, so we'll see if that makes merges more reliable or adds unnecessary pain.

PR #4088, opened 2017-01-06
Make mdn-format plugin more flexible and easier to maintain

a2sheppy:
This PR updates the mdn-format plugin for CKEditor to be more flexible and
easier to maintain. The key improvements:

It now uses an array of objects to define the buttons it will add to the
toolbar. Each object in the array allows setting the tag that will be
toggled by the button, as well as a title (tooltip) to describe the button's
purpose. Also supported is specifying the keystroke that will activate the
button and the name of the icon to use in the button.

The button's command handler is still based on the tag name at this time;
this is likely to become configurable in the future as well.

The loop that adds the buttons now works by iterating over the list of
button objects, adding the buttons and the keystrokes.

Setting the keystrokes for these buttons is no longer done in
ckeditor_config.js.


jwhitlock:
Rebased as PR #4119.

PR #4091, opened 2017-01-09
Bug 1314708: Test for translation interface

stephaniehobson:


stephaniehobson:
Oops, forgot to create a PR for this branch. @m8ttyB R?

tests/functional/test_article_translate.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+
+from pages.article_translate import TranslatePage
+from pages.article import ArticlePage
```
escattone: ArticlePage is not used so can remove this line.

tests/pages/article_translate.py
```diff
@@ -0,0 +1,44 @@
+from pypom import Region
```
escattone: `Region` is unused so can remove this line.

stephaniehobson:
Updated 🐨

escattone:
This PR adds one test, and since it's marked with "login", it will not be run as part of the integration tests on staging. With the fixes I've added, it does run against a local development environment using the sample DB, as longs as the sample DB has been modified as follows:
- the passwords for the test user accounts have to be fixed (at least mine did, they were corrupted)
- the page http://localhost:8000/fr/docs/User:anonymous:uitest must be added by translating http://localhost:8000/en-US/docs/User:anonymous:uitest

jwhitlock:
Added some tasks to PR #4076 for the missing and incorrect fixtures.

PR #4096, opened 2017-01-12
bug 1330782: fix race condition in functional tests

escattone:
bug 1330782: fix race condition in functional tests
* fix BasePage.wait_for_page_to_load method
* fix EditPage.wait_for_page_to_load method
* fix BasePage.Header.open_feedback method


stephaniehobson:
Passing for me now 👍

PR #4094, opened 2017-01-11
bug 1319259: Update to node v6

jwhitlock:
* [x] Switch to node.js v6 for the KumaScript engine
* [x] Update ansible provisioning
* [x] Update docker
* [x] Update SCL3 deployment scripts
* [x] Switch to [mozmar/base](https://github.com/mozmar/docker-base)
* [ ] Update Jenkinsfile to build and upload the [kuma_base](https://quay.io/repository/mozmar/kuma_base) image
* [x] Merge KumaScript node v6 [PR 69](https://github.com/mozilla/kumascript/pull/69), switch in this PR.

PR #4097, opened 2017-01-13
Fix Bug 1330838: Switch Stylus to Sass

stephaniehobson:
**NOTES for testing and compiling:**
- we need to specify `--precision 3` in the new compile stylesheets command

**Changes in this PR:**

**Rename all files:**
- I did this as a separate git commit so the files are treated as renamed and we can preserve the git history.

**Change Stylus syntax to Sass:**
- variable declarations from `=` to `:`
- add @ in front of if, else, for, etc.
- add @include in front of all mixins
- Add # to all {$selectors}
- add $ to all variables

**Edit some mixins because:**
- The Sass units() function cannot convert units like Stylus does
- It is not possible to use [parent selector in combination with element](https://github.com/sass/sass/issues/282) in Sass
- Sass does not have property look ups (ie, using @border-width to get last defined border-width value from source order) (we are probably better off without this feature TBH)
- Can't name placeholders the same as an existing class name.
- Can't reach back and add properties to the declaration which called the mixin.

**Differences in final output:**

- All files:
    - Sass writes out hex values with all 6 characters.
    - Border short hand declarations can no longer be combined with `rgba-fallback` mixin, so those are written out differently now.
    - Switched from @extend to @mixin for column widths, so they are different in the compiled source order.
    - Decimal numbers are output without a trailing 0 now.
    - All quotes are now double quotes.
    - Changes in output source order.

- dashboards
    - color difference for highlight (updated)
    - removed duplicate word-break value
- diff
    - no file specific changes
- error-403-alternate
    - no file specific changes
- error-404
    - no Stylus auto prefix on keyframes (good!)
- error
    - no file specific changes
- fellowship
    - smart quotes replaced with HTML escaped smart quotes
- font-awesome
    - html escaped glyphs changed to glyphs
    - removed extra classes like lg, 2x, spin, &-ul, &-border
- gaia
    - no file specific changes
- home
    - column widths are mixin now
    - I added ms-input-placeholder
    - escaped chars get changed to unescaped
- main
    - escaped chars get changed to unescaped
    - source order changes
- promote
    - no file specific changes
- search-suggestions
    - no file specific changes
- search
    - no file specific changes
- sphinx
    - no file specific changes
- submission
    - no file specific changes
- user-banned
    - no file specific changes
- users
    - Removed utilities because they are in components which are in main.css
    - wi-widget-content rules had an error and were not being output in Stylus, are being output now.
- wiki-compat-tables
    - no file specific changes
- wiki-edit
    - some duplication of bidi output but we need to address the bidi system with a separate PR.
- wiki-revisions
    - no file specific changes
- wiki-syntax
    - no file specific changes
- wiki-wysiwyg
    - no file specific changes
- wiki
    - Made some changes to .htab to simplify code.
    - Indicators and spec source order change and slightly different colours.
- zones
    - no file specific changes


jwhitlock:
The rename looks good to me, and thanks for the extensive checking of the output CSS.  The two-commit strategy preserves the renames and makes the actual changes easier to review. 👍 

I was surprised that all the tests pass in TravisCI - maybe ``sass`` is installed in the base image already?

I think we should follow the strategy in PR #4094, and update the dev environment in the same commit:

* [ ] Install Sass in the Docker base image
* [ ] Install Sass via Ansible for Vagrant / TravisCI
* [ ] Replace "Stylus" with "Sass" in the documentation
* [ ] Update other references to "Stylus" if found

I think I or any staff person could add the follow-on commits to add Stylus to the dev environment.

escattone:
I think I've covered every item in John's checklist above, but I'll let John or someone else be the final judge of that.

jwhitlock:
The changes work for me, but I'm finding that I don't understand our asset process.  It appears that ``./manage.py collectstatic`` will compile the stylus / sass files to CSS, but it is unclear what the mechanism is.  It could be django-pipeline, but current docs suggest you have to configure the compiler to make that happen, and we're not configuring the compiler.

It appears @escattone is digging into this, so I'll step back and just be happy it is working for now.

escattone:
I didn't think that running `./manage.py collectstatic` compiled the sass files to CSS, but I suddenly doubted myself as well when viewing the output of that command. However, upon further investigation, I was able to see that `./manage.py collectstatic` definitively does NOT perform the .scss file compilation. I modified `base.scss`, ran the `collectstatic` command, and it did NOT pick-up the change. However, after I ran `make compilecss` (which runs the sass compiler), the `collectstatic` command did pick-up my change.

escattone:
This PR is superseded by https://github.com/mozilla/kuma/pull/4109. That PR contains everything this one does plus it folds Sass compilation into the Django pipeline, adds gulp to the workflow, and updates the documentation.

jwhitlock:
I prefer this PR to #4109, because it handles the rename and the changes as separate commits.  This makes git history useful.  PR #4109 squashes all the work into a single commit, and will make it harder to track changed lines before 2017.

jwhitlock:
PR #4109 is unsquashed, and now is a good replacement.

PR #4098, opened 2017-01-13
Fix Bug 1329743: Compat bugs.

stephaniehobson:
Can't fix the Safari problem. Filed a bug, https://bugs.webkit.org/show_bug.cgi?id=166870

The IE problem is with an indicator that was a bit of a hack to begin with - we're styling based on inline styling. IE is changing the inline style from a hex to an rgb colour. So now we have a selector for the rgb value as well.

Text styling difference has to do with differing support for text-overflow: ellipsis. I don't think we want ellipsis here, since it could hide important information, so I removed it.

IE also has no default saying for abbr, so added that.

### DIFF
--- a/kuma/static/styles/base/elements/typography.scss
+++ b/kuma/static/styles/base/elements/typography.scss
@@ -141,6 +141,7 @@ a {
 
 abbr[title] {
     cursor: help;
+    text-decoration: dotted underline;
 }
 
 sup {
--- a/kuma/static/styles/components/syntax/syntaxbox.scss
+++ b/kuma/static/styles/components/syntax/syntaxbox.scss
@@ -11,7 +11,6 @@ pre.twopartsyntaxbox {
     @extend %code-block;
     background: $code-block-background-alt-color;
     white-space: pre-wrap;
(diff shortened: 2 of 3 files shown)


kuma/static/styles/components/wiki/indicators.styl
```diff
@@ -134,7 +134,8 @@ icons for both types of indicators
 .experimental,
 .experimentalBadge,
 /* es7 and harmony macros  */
-.overheadIndicator[style='background: #9CF49C;'] {
+.overheadIndicator[style='background: #9CF49C;'],
+.overheadIndicator[style='background: rgb(156, 244, 156);'] // IE10 {
```
jpetto: Opening curly brace got caught in the comment, and commas in the `rgb` function need to be escaped for Stylus to compile properly.

jpetto:
Here's the compiled CSS before escaping the commas in `indicators.styl`:

screen shot 2017-01-17 at 4 07 25 pm

Without escaping, Stylus seems to assume the commas indicate the end of a selector (regardless of whether they are in quotes or not). Escaping the commas fixes the issue.

PR #4100, opened 2017-01-17
Bug 1331643, 1330357: De-duplicate tags, allow mixed-case interests

jwhitlock:
With [bug 1293749](https://bugzilla.mozilla.org/show_bug.cgi?id=1293749) fixed (Add missing ``UNIQUE KEY`` indexes in production), some new issues have appeared:

* [bug 1330357](https://bugzilla.mozilla.org/show_bug.cgi?id=1330357) - ISE if profile interest tag is not lowercase in the database
* [bug 1331643](https://bugzilla.mozilla.org/show_bug.cgi?id=1331643) - ISE when adding two document tags that only differ in capitalization

Both appear to be caused by an issue with [django-taggit](https://github.com/alex/django-taggit) and MySQL with a case-insensitive collation and a unique index.  Adding two "case similar" tags, like "WEB" and "web", will result in attempting to add both, and raise an IntegrityError.

The issue is handled in this PR by querying the database for existing tags with "case similar" names, and creating new ones as needed, and then passing the unique set of Tag instances to taggit to create the many-to-many relations, rather than passing a list of names.

Additionally, user profile interests and expertise were restricted to lower-case tags to avoid triggering the creation of duplicate tags.  This removes the restriction and UI changes, so that non-lowercase interests, like "HTML" and "CSS", can be added.  This change allows mixed-case profile tags, but existing tags will need to be updated in the database as desired.

### DIFF
--- a/kuma/core/managers.py
+++ b/kuma/core/managers.py
@@ -51,22 +51,9 @@ class _NamespacedTaggableManager(_TaggableManager):
 
         if namespace is not None:
             # Namespace requested, so generate filtered set
-            # TODO: Do this in the DB query? Might not be worth it.
-            #
-            # On databases with case-insensitive collation, we can end
-            # up with duplicate tags (the same tag, differing only by
-            # case, like 'javascript' and 'JavaScript') in some
-            # cases. The most common instance of this is user
-            # tags, which are coerced to lowercase on save to avoid
-            # the problem, but because there are a large number of
-            # these duplicates already existing, we do a quick filter
-            # here to ensure we don't return a bunch of dupes that
-            # differ only by case.
-            seen = []
             results = []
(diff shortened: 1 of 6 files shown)


safwanrahman:
The Tag Code! 😢 😭 🤕

jwhitlock:
Force-push to rebase and make flake8 happy.

safwanrahman:
Do you think ``TAGGIT_CASE_INSENSITIVE`` settings can do this insensitive lookup? We can now add that.

jwhitlock:
@safwanrahman it's worth a shot. I'll create an alternate branch.

jwhitlock:
Closing in preference for PR #4103

PR #4102, opened 2017-01-19
[bug 1314708] Jenkins pipeline refactor and integration tests. 

glogiotatidis:
This pull request 

1. Refactors the current pipeline
2. Fixes the integration tests step
3. Skips failing tests (as per @m8ttyb suggestion via email on Jan 18th)

## 1. Pipeline refactor

The basic idea is that every branch can have a specific Pipeline steps or  Configuration. Jenkins will auto-discover the steps and configuration but checking within the `Jenkinsfiles` directory for a `.groovy` and / or a `.yml` file with the name of the branch.

For example for the `stage-integration-tests` branch there're  `Jenkinsfiles/stage-integration-tests.groovy` and  `Jenkinsfiles/stage-integration-tests.yml` which will be execute and loaded accordingly when the Jenkins pipeline runs tests on that branch.

The `prod-integration-tests` branch does not have a `groovy` file of it's own but it does have `yml` file. The later configures `pipeline.script` which will instruction the Jenkins Pipeline to load the same `stage-integration-tests.groovy` script but with different configuration (in this case different `base_url`)

Similarly for the `master` branch which has it's own pipeline. Any other branch will get the steps defined in the `default.groovy` file.

##  2. Fix the integration tests step

This step has been re-written in groovy only and does not use any external bash scripts. For this to happen a number of Jenkins global functions were developed (see [mozmar/jenkins-pipeline](https://github.com/mozmar/jenkins-pipeline)) which make the building, running and pushing of docker containers seamless (and other cool stuff too).  This way we can take full advantage of Pipelines, build better code and handle situations that have been painful in the past, like the stopping of background docker containers. 

For Kuma the script will run a configurable (via yml) version of Selenium and a configurable number of Selenium nodes and execute the tests against an also configurable `base_url`. Will cleanup after itself before exiting.

## 3. Skip failing tests

This pull request does not fix failing integration tests. As per @m8ttyb's suggestion I just marked them for skipping for now.

## tl;dr

This follows the flow @jgmize suggested in emails and allows developers to push integration tests in `stage-integration-tests` branch and run them against `developer.allizom.org`. It also configures `prod-integration-tests` and provides a framework for clean Jenkinsfiles and hot-loading of code and configuration per branch.

The idea is to merge this to `master` and then force-push to `stage-integration-tests` to bring it in par with `master`. It does not use any code from the current `stage-integration-tests` and that's why I didn't PR against it.

For example runs you can visit my [demo multibranch pipeline](https://ci.us-west.moz.works/job/giorgos-test/) against glogiotatidis/kuma

Things to be done:
 * Global error handling
 * IRC notifications

- [ ] Delete giorgos-test Jenkins Pipeline after merging.

See also:
 * https://github.com/glogiotatidis/jenkins-pipeline

/cc @jgmize @jwhitlock @davehunt @escattone @pmclanahan


Jenkinsfile
```diff
@@ -1,44 +1,49 @@
...
+  }
+}
+
 node {
-    stage 'git'
+  stage("Prepare") {
     checkout scm
+    sh 'git submodule sync'
+    sh 'git submodule update --init --recursive'
+    setGitEnvironmentVariables()
```
pmac: What does this do exactly? Is it a global Jenkins thing?

pmac: NM. Found it in your groovy library thing in your github. Very cool.

Jenkinsfile
```diff
@@ -1,44 +1,49 @@
+@Library('github.com/glogiotatidis/jenkins-pipeline@master')
```
jgmize: This needs to be updated now that the canonical repo is in mozmar

jgmize:
@glogiotatidis this is some very nice technical work, and I've learned several things from reading through this. While I had some initial hesitations about the complexity of the implementation, I think the overall maintainability benefits are worth the tradeoff. I assume @escattone is going to need some time to digest and give additional feedback before merge, but one thing that @jwhitlock reminded me of on Friday that unlike some other projects, MDN PRs currently require a bug # in the first line of the commit message. I think we could probably just track this work under [bug 1314708](https://bugzilla.mozilla.org/show_bug.cgi?id=1314708), but if you and/or one of the other devs think this should have its own bug then please file, comment, and update the PR accordingly.

jgmize:
Another thing this PR needs is some usage docs, probably a new section under https://github.com/mozilla/kuma/blob/master/docs/tests-ui.rst adapted from this PR description.

glogiotatidis:
Added docs in `tests-ui`, docs in jenkins pipelines repository and bug number per @jgmize review comments.

glogiotatidis:
Thanks for the review @escattone . I agree with your suggestions and made the changes in separate commits so they are easy to review. Fill free to squash-merge when it's time.

escattone:
Thanks @glogiotatidis! One final request. The three skipped tests marked with `@pytest.mark.skip` comprise all of the `pytest.mark.login` tests, so now that we have the new "job.tests" setting available via the `yml` files, I'd propose:

- [ ] Remove the `@pytest.mark.skip` decorator from the three "login" tests
- [ ] Add a `job.tests` setting with a value of "not login" to both the "prod-integration-tests.yml" and "stage-integration-tests.yml" files

Then I think we should merge this either tomorrow or on Monday of next week.

glogiotatidis:
@escattone made the changes and rebased off master.

davehunt:
Instead of skipping these tests, I would suggest marking them as expected to fail. We can even give the expected exception, so if they fail for any other reason they will cause a failure. This means we don't remove coverage, and we won't forget to include the tests again once they are expected to pass. If needed, the xfail can even be conditional based on the hostname. See http://docs.pytest.org/en/latest/skipping.html#mark-a-test-function-as-expected-to-fail for documentation of marking tests as expected to fail.

m8ttyB:
We discussed skipping this for simplicity as well as a discussion point on potentially removing/drastically reworking these tests.

@glogiotatidis skip or xfail; your choice.

Jenkinsfiles/prod-integration-tests.yml
```diff
@@ -0,0 +1,9 @@
+pipeline:
+  enabled: true
+  script: "integration-tests"
+job:
+  dockerfile: "DockerfileIntegrationTests"
+  selenium: "2.48.2"
+  selenium_nodes: 5
+  base_url: 'https://developer.mozilla.org'
+  tests: "not login"
```
escattone: Sorry @glogiotatidis, I suggested the wrong value here. You'll have to use::
   `tests: '"not login"'`
or::
    `tests: "\"not login\""`
instead, since it's an expression.

Jenkinsfiles/stage-integration-tests.yml
```diff
@@ -0,0 +1,9 @@
+pipeline:
+  enabled: true
+  script: "integration-tests"
+job:
+  dockerfile: "DockerfileIntegrationTests"
+  selenium: "2.48.2"
+  selenium_nodes: 5
+  base_url: 'https://developer.allizom.org'
+  tests: "not login"
```
escattone: Again, you'll have to use::
   `tests: '"not login"'`
or::
    `tests: "\"not login\""`
instead, since it's an expression.

glogiotatidis:
> Since the value for the pytest marker (job.tests) is an expression, it's necessary that the quotes make it through to the pytest command. So you'll have to use "\"not login\"" or '"not login"' instead of "not login". See the notes I added to the lines in the YAML files that have to change. Then pytest should run all of the integration tests except those marked with login, which is what we want. Thanks @glogiotatidis!

Fixed that by [adding quotes in the Jenkinsfile](https://github.com/mozilla/kuma/pull/4102/commits/f8e78099f92ac2784a24a02bbf34222f8166c682#diff-3f07e3eda0d56afa869c0f1e788d6f1eR12) instead to make the yml more error proof. 

Yet no tests run with this marker. I run `py.test tests/functional --markers` and there's no `login` marker so it makes sense that no tests run but is this what you want @escattone ?

escattone:
Thanks @glogiotatidis!

Regarding the tests:
- Running `py.test --markers ...` only returns the markers defined in the `pytest.ini` file, and none of the markers defined in `tests/functional` have been added to that file. If you do a `grep -R "pytest.mark.login" *` you'll see the three "login" tests.
- @jgmize and I discovered that the details of the `py.test ...` command in the Jenkins logs are shown only when one or more tests fail, so when using "not login", all of the functional tests (except the "login" tests) actually run, but do not show up in the log file since there are no failures.

I'd rather not mark the "login" tests as "skip" or "xfail" since they are valid tests in other contexts (e.g., locally using the sample db).

davehunt:
> I'd rather not mark the "login" tests as "skip" or "xfail" since they are valid tests in other contexts

For what it's worth, both skip and xfail can be applied conditionally, which means tests can be expected to pass when using a sample db and expected to fail when using a live environment.

PR #4103, opened 2017-01-20
bug 1331643, 1330357: Avoid duplicate tags with TAGGIT_CASE_INSENSITIVE=True

jwhitlock:
With [bug 1293749](https://bugzilla.mozilla.org/show_bug.cgi?id=1293749) fixed (Add missing ``UNIQUE KEY`` indexes in production), some new issues have appeared:

* [bug 1330357](https://bugzilla.mozilla.org/show_bug.cgi?id=1330357) - ISE if profile interest tag is not lowercase in the database
* [bug 1331643](https://bugzilla.mozilla.org/show_bug.cgi?id=1331643) - ISE when adding two document tags that only differ in capitalization

Both appear to be caused by an issue with [django-taggit](https://github.com/alex/django-taggit) and MySQL with a case-insensitive collation and a unique index.  Adding two "case similar" tags, like "WEB" and "web", will result in attempting to add both, and raise an IntegrityError.

The issue is handled in this PR by querying the database for existing tags with "case similar" names, and creating new ones as needed, and then passing the unique set of Tag instances to taggit to create the many-to-many relations, rather than passing a list of names.

Additionally, user profile interests and expertise were restricted to lower-case tags to avoid triggering the creation of duplicate tags.  This removes the restriction and UI changes, so that non-lowercase interests, like "HTML" and "CSS", can be added.  This change allows mixed-case profile tags, but existing tags will need to be updated in the database as desired.

This is an alternate for PR #4100 that uses ``TAGGIT_CASE_INSENSITIVE=True``, and also avoids django-taggit issue https://github.com/alex/django-taggit/issues/460.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -126,9 +126,9 @@ django-sundial==1.0.2 \
     --hash=sha256:3745891337d726041b37aed95005b9c70377116d232157726e48b46aefef1961
 
 # Simple tagging support
-django-taggit==0.18.0 \
-    --hash=sha256:7010a1f7597d954aaf8b8174563f2e3455520db949d9c5d360dd9e4f906dd32a \
-    --hash=sha256:4ddd61928a4fc6ba7ccb1145d8d3a6d5c2829de8d8738eea49924ade8969f0ce
+django-taggit==0.21.6 \
+    --hash=sha256:5d0a2d7c3933badb36bcbe1cdee62514fbc8949cd2b92ed1c351b724282318fd \
+    --hash=sha256:80e22fdf41fba66a3c6f4807e283f38c4557838e35766a942370fb90f21d7435
 
 # Send async emails in response to events
 django-tidings==1.1 \
--- a/kuma/core/managers.py
+++ b/kuma/core/managers.py
@@ -51,22 +51,9 @@ class _NamespacedTaggableManager(_TaggableManager):
 
         if namespace is not None:
             # Namespace requested, so generate filtered set
-            # TODO: Do this in the DB query? Might not be worth it.
-            #
-            # On databases with case-insensitive collation, we can end
-            # up with duplicate tags (the same tag, differing only by
-            # case, like 'javascript' and 'JavaScript') in some
-            # cases. The most common instance of this is user
-            # tags, which are coerced to lowercase on save to avoid
-            # the problem, but because there are a large number of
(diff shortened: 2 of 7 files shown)


jwhitlock:
I've opened PR https://github.com/alex/django-taggit/pull/461 to fix the issue upstream, and will apply the fixes to this branch with the assumption that a future version of ``django-taggit`` will fix the issue.

safwanrahman:
Need to update ``django-taggit==0.21.5`` as the fix has been merged in the version
https://github.com/alex/django-taggit/releases

requirements/default.txt
```diff
@@ -126,9 +126,9 @@ django-sundial==1.0.2 \
     --hash=sha256:3745891337d726041b37aed95005b9c70377116d232157726e48b46aefef1961
 
 # Simple tagging support
-django-taggit==0.18.0 \
-    --hash=sha256:7010a1f7597d954aaf8b8174563f2e3455520db949d9c5d360dd9e4f906dd32a \
-    --hash=sha256:4ddd61928a4fc6ba7ccb1145d8d3a6d5c2829de8d8738eea49924ade8969f0ce
+django-taggit==0.21.3 \
```
safwanrahman: with @jwhitlock patch, ``0.21.5`` has been released

jwhitlock: Not released: https://pypi.python.org/pypi/django-taggit

jwhitlock: and now released

jwhitlock:
Unfortunately, I introduced a bug in django-taggit 0.21.5. I've opened an issue (https://github.com/alex/django-taggit/issues/463) and a PR (https://github.com/alex/django-taggit/pull/464) to fix it.  Meanwhile, the manual tag creation continues to bypass the buggy code, so leaving it in.

I'd like to merge and deploy this PR, fixing the ISE bugs, and follow up with a new PR when django-taggit 0.21.6 is released.

safwanrahman:
Its better to use your fork until the patch is merged rather than using custom code for manual tag creating

frewsxcv:
Upstream patch is merged and published

jwhitlock:
@frewsxcv thanks for the quick merge and publish!

Rebased on current master, removed manual tag creation, crossing fingers and waiting for TravisCI 🤞

PR #4105, opened 2017-01-23
Add Gulp to front-end workflow.

jpetto:
After merging, need to run `npm install`, then `gulp`. With `gulp` running, any change to a file in `/kuma/kuma/static` should be logged to the console and copied to `/static`.

Page loads are a bit slower (~3.5s), but I'm guessing that's due to the pipeline recompiling assets each time. Still a workable delay, IMO.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,10 +14,9 @@
 .tox
 .vagrant
 bower_components
-static/*/*
+static/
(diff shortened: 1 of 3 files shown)


jpetto:
@escattone - Give it a try?

PR #4107, opened 2017-01-25
bug 1218563: Update base image to v3.6

jwhitlock:
Includes security updates.


Dockerfile-base
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/mozmar/base:v0.3.5
+FROM quay.io/mozmar/base:v0.3.6
```
jgmize: If you remove the tag and use the implicit `latest`, you won't have to file a PR like this again. That's actually the main reason we have that base.

m8ttyB:
👍

PR #4108, opened 2017-01-26
bug 1332828: Remove DekiWiki:Date from autoload

jwhitlock:
It is a lesser wrapper on the native Date type, and is shadowing the native Date type after the node 6.x update.


jwhitlock:
This was verified by @Elchi3 in [bug 1332828](https://bugzilla.mozilla.org/show_bug.cgi?id=1332828#c1), and verified in production by changing the same value in the .json config file.

PR #4109, opened 2017-01-27
bug 1330838: change css pre-processor to sass

escattone:
* Update all Stylus files to Sass files
* Install Sass in the Docker base image
* Install Sass via Ansible for Vagrant/TravisCI
* Fold Sass compilation into django-pipeline
* Add Gulp to front-end workflow
* Remove compile-stylesheets script
* Update the documentation
* Update Makefile
* Update Procfile

From Stephanie's original notes:

Change Stylus syntax to Sass:

    variable declarations from = to :
    add @ in front of if, else, for, etc.
    add @include in front of all mixins
    Add # to all {$selectors}
    add $ to all variables

Edit some mixins because:

    The Sass units() function cannot convert units like Stylus does
    It is not possible to use parent selector in combination with element in Sass
    Sass does not have property look ups (ie, using @border-width to get last defined border-width value from source order) (we are probably better off without this feature TBH)
    Can't name placeholders the same as an existing class name.
    Can't reach back and add properties to the declaration which called the mixin.

Differences in final output:

    All files:
        Sass writes out hex values with all 6 characters.
        Border short hand declarations can no longer be combined with rgba-fallback mixin, so those are written out differently now.
        Switched from @extend to @mixin for column widths, so they are different in the compiled source order.
        Decimal numbers are output without a trailing 0 now.
        All quotes are now double quotes.
        Changes in output source order.

    dashboards
        color difference for highlight (updated)
        removed duplicate word-break value

    diff
        no file specific changes

    error-403-alternate
        no file specific changes

    error-404
        no Stylus auto prefix on keyframes (good!)

    error
        no file specific changes

    fellowship
        smart quotes replaced with HTML escaped smart quotes

    font-awesome
        html escaped glyphs changed to glyphs
        removed extra classes like lg, 2x, spin, &-ul, &-border

    gaia
        no file specific changes

    home
        column widths are mixin now
        I added ms-input-placeholder
        escaped chars get changed to unescaped

    main
        escaped chars get changed to unescaped
        source order changes

    promote
        no file specific changes

    search-suggestions
        no file specific changes

    search
        no file specific changes

    sphinx
        no file specific changes

    submission
        no file specific changes

    user-banned
        no file specific changes

    users
        Removed utilities because they are in components which are in main.css
        wi-widget-content rules had an error and were not being output in Stylus, are being output now.

    wiki-compat-tables
        no file specific changes

    wiki-edit
        some duplication of bidi output but we need to address the bidi system with a separate PR.

    wiki-revisions
        no file specific changes

    wiki-syntax
        no file specific changes

    wiki-wysiwyg
        no file specific changes

    wiki
        Made some changes to .htab to simplify code.
        Indicators and spec source order change and slightly different colours.

    zones
        no file specific changes


escattone:
I fixed the Travis issues by adding:
`nvm install 6`
`nvm use 6`
after the ansible step.

docs/development.rst
```diff
@@ -88,32 +88,43 @@ For more information, see the :doc:`test documentation <tests>`.
...
+Sass files need to be compiled for changes to take effect, but don't worry,
+the compilation is done automatically by Django (specifically django-pipeline).
+
+When doing front-end development on your local machine, you'll probably
+want to run (most likely in its own shell)::
 
-In the Vagrant environment, the ``foreman`` task ``stylus`` will automatically
-compile Stylus files when they change, placing the generated CSS files at
-``build/assets/css``.
+    gulp static:watch
```
jpetto: This can just be `gulp`. The default Gulp command starts `gulp static:watch`.

docs/development.rst
```diff
@@ -306,23 +317,16 @@ To connect to the database specified in ``DATABASE_URL``, use::
...
-
-#. Set the environment variables ``DEBUG=false``.
-#. Run ``make compilejsi18n collectstatic`` in the VM or container.
-#. Restart the web process by restarting ``foreman``.
+Kuma will automatically run in debug mode, with the ``DEBUG`` setting turned to
+``True``. Setting ``DEBUG=False`` will put you in production mode and
+generate/use minified (compressed) and versioned (hashed) assets. To
+emulate production, and test compressed and hashed assets locally:
+
+#. Set the environment variable ``DEBUG=false``.
```
jpetto: Does one have to change this in both `.env` and `docker-compose.yml`?

docs/installation.rst
```diff
@@ -136,14 +136,45 @@ compiled to their final form::
...
+    1687 static files copied to '/app/static', 1773 post-processed
+
+.. _frontend-development:
+
+Frontend Development
+====================
+When doing front-end development on your local machine, you'll probably
+want to run (most likely in its own shell)::
+
+     gulp static:watch
```
jpetto: Again, this can just be `gulp`.

docs/troubleshooting.rst
```diff
@@ -169,10 +168,18 @@ running it. (See `Running individual processes`_).
...
    make sure you enabled :ref:`KumaScript <enable KumaScript>`.
 
--  If changes to stylesheets do not have any effect, try compiling the Stylus
-   manually by running this command in the VM::
+-  If changes to Sass stylesheets do not have any effect and you've got
+   ``DEBUG=True`` (debug mode, which is the default), then make sure you're
+   also running (perhaps in a separate shell)::
 
-       compile-stylesheets
+       gulp static:watch
```
jpetto: Just `gulp`.

gulpfile.js
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global __dirname, require */
```
jpetto: Review for self - remove `__dirname` here.

docs/development.rst
jpetto: After a little local testing, it appears that the `DEBUG` setting in `docker-compose.yml` is respected, while the same setting in `.env` is not (at least in terms of working with assets).

escattone: The value of DEBUG in `docker-compose.yml` will override any value set in `.env`. Within `docker-compose.yml`, we could respect the environment setting if present but default to True (`DEBUG=${DEBUG:-True}`).

escattone: [This commit](https://github.com/mozilla/kuma/pull/4109/commits/8e109d2f7a4c8b52a8d9db0969ccc3d88edf3eb6) modifies `docker-compose.yml` such that a DEBUG setting in the environment will override the default specified within the file.

jwhitlock:
I'm having trouble running locally, getting an Internal Server Error.  Last few lines of the traceback;

```
web_1            |   File "/app/jinja2/base.html", line 22, in block "site_css"
web_1            |     {% stylesheet 'mdn' %}
web_1            |   File "/usr/local/lib/python2.7/dist-packages/pipeline/jinja2/__init__.py", line 37, in package_css
web_1            |     return self.render_compressed(package, package_name, 'css')
web_1            |   File "/usr/local/lib/python2.7/dist-packages/pipeline/templatetags/pipeline.py", line 72, in render_compressed
web_1            |     package_type)
web_1            |   File "/usr/local/lib/python2.7/dist-packages/pipeline/templatetags/pipeline.py", line 108, in render_compressed_sources
web_1            |     package_type))
web_1            | AttributeError: 'PipelineExtension' object has no attribute 'render_error_css'
```

I tried a ``make build-static``, but no joy.  I'll try again in the morning, using the install instructions in the PR.  I also want to look into why tests are running much slower ([45 minutes](https://travis-ci.org/mozilla/kuma/builds/196784704) vs [11 minutes](https://travis-ci.org/mozilla/kuma/builds/196765536)) and how to get the locales test passing in TravisCI again.

jwhitlock:
I needed to run ``make build-base VERSION=latest`` to install ``node-sass`` into my base image, and then ``make up`` to rebuild and run. No ISE anymore, and pages render. I can review the PR and look at speed issues.

The Jinja2 ``PipelineExtension`` doesn't correctly handle errors.  This may be fixed in a later release of ``django-pipeline``, or we may have an upstream bug.  I'll investigate later.

PR #4113, opened 2017-02-01
bug 930102: Add "Hidden When Reading" style

jwhitlock:
Add "Hidden When Reading" option to style dropdown in editor. In the menu, it has a similar appearance to when hidden content is displayed in the editor (that is, with a dotted outline around it). The small black area at the top-left is where the "hidden" tab is in the edit box; @stephaniehobson says this is not something we can easily fix and it isn't a significant issue anyway, so it should be fine as it is.

This is a rebase of PR #4057 after the Stylus → Sass change.

### DIFF
--- a/kuma/static/styles/base/utilities.scss
+++ b/kuma/static/styles/base/utilities.scss
@@ -34,6 +34,7 @@
 .hidden {
     display: none;
 
+    html.cke_panel_container &,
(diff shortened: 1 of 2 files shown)


jpetto:
Pulled and tested locally to be sure. Still an r+. 👍

PR #4115, opened 2017-02-01
bug 1325097: Make columns more visible in editor

jwhitlock:
Updated the ``twocolumns`` and ``threecolumns`` classes so they draw with a pale blue dotted border around the columns when in the editor, for ease of layout work. The format is in a variable (``$editor-box-border``) so it can be reused in other similar cases if needed.

This is a rebase of PR #4080 after the Stylus → Sass update.


jpetto:
Looks good. 👍

PR #4116, opened 2017-02-02
bug 1330838: Generate hashed static assets with DEBUG=False

jwhitlock:
Borrow the [Bedrock solution](https://github.com/mozilla/bedrock/blob/2fb16c05fbb847b57e0fbbea8a5b51d51d554e43/bedrock/base/storage.py#L15) for generating hashed static assets.

### DIFF
--- a/kuma/core/pipeline/storage.py
+++ b/kuma/core/pipeline/storage.py
@@ -1,7 +1,8 @@
+from django.conf import settings
 from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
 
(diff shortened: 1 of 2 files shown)


jwhitlock:
The issue appears to be running this way:

```
docker-compose exec web make collectstatic
```
uses the default configuration for local testing with ``DEBUG=True``, while running this way:

```
make bash
./manage.py collectstatic
```
or 
```
docker-compose exec web ./manage.py collectstatic
```
uses the default Django config and respects the environment variable ``DEBUG=True``

PR #4118, opened 2017-02-03
bug 1031910 - Multilingual spellcheck

jwhitlock:
Enable multi-language spellchecking and adjust SCAYT settings.

This is a rebase of PR #4067, but without auto-loading of the SCAYT plugin.


jwhitlock:
Confirmed that this still works locally. The feature is really improved - new languages, skipping MixedCase words, etc.

PR #4121, opened 2017-02-03
Demo deploys to k8s via Jenkinsfile + Deis workflow

jgmize:
WIP working example: https://mdn-demo.virginia.moz.works/en-US/ deployed by https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/job/demo/16/console

### DIFF
--- a/Jenkinsfiles/demo.groovy
+++ b/Jenkinsfiles/demo.groovy
@@ -1,0 +1,16 @@
+stage('Build & push images') {
+  sh 'make build-kuma push-kuma'
+}
+
+stage('Deploy') {
+  env.KUBECONFIG = "${env.HOME}/.kube/virginia.kubeconfig"
+  env.DEIS_PROFILE = 'virginia'
+  env.DEIS_BIN = 'deis2'
+  env.DEIS_APP = 'mdn-' + env.BRANCH_NAME
+
+  sh 'make deis-create-and-or-config'
+  sh "KUBECONFIG=${env.KUBECONFIG} kubectl --namespace=${env.DEIS_APP} apply -f k8s/"
+  sh 'make deis-pull'
+  sh 'make deis-migrate'
+  sh 'make deis-scale-worker'
+}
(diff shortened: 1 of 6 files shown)


jgmize:
Working with @metadave and @escattone on this, still some items to address

Jenkinsfiles/demo.groovy
```diff
@@ -0,0 +1,21 @@
...
+  env.DEIS_BIN = 'deis2'
+  env.DEIS_APP = 'mdn-' + env.BRANCH_NAME
+
+  sh 'make deis-create'
+  // sh 'make deis-config'
+  // currently broken:
+  // DEIS_PROFILE=virginia deis2 config:push -p .env-dist -a mdn-demo
+  // Creating config... ...o..Error: Unknown Error (409): {"detail":"jenkins changed nothing - release stopped"}
+  // make: *** [deis-config] Error 1
+  sh "KUBECONFIG=${env.KUBECONFIG} kubectl --namespace=${env.DEIS_APP} apply -f k8s/"
```
metadave: I could see this being a helm chart down the road, but great for now.

Makefile
```diff
@@ -18,7 +18,7 @@ DB_PASS ?= kuma # default for ephemeral demo DBs
 target = kuma
 requirements = -r requirements/local.txt
 # set Django settings module if not already set as env var
-export DJANGO_SETTINGS_MODULE ?= kuma.settings.testing
+export DJANGO_SETTINGS_MODULE ?= kuma.settings.prod
```
escattone: I think it's safest for the default value of the `DJANGO_SETTINGS_MODULE` to remain at `kuma.settings.testing` here (as this `Makefile` is used in multiple contexts), and instead do a `env.DJANGO_SETTINGS_MODULE = 'kuma.settings.prod'` in the `Jenkinsfiles/demo.groovy` file.

escattone:
It seems that [this commit](https://github.com/mozilla/kuma/pull/4121/commits/f30fad2d92fc5e8b6fdafab9c3d42510df44be11) addresses my concern of not (yet) messing with the `DJANGO_SETTINGS_MODULE` setting in the `Makefile`, while still ensuring we use the production settings when building the Kuma Docker image (for the `make build-static` step).

I'd love to have @jwhitlock take a look at this as well to get his thoughts before I merge.

escattone:
Actually, I had an idea overnight that would be better, let me try that before any further review.

escattone:
My idea didn't work, as it was based on the Dockerfile picking up the DJANGO_SETTINGS_MODULE setting from its environment (in this case, the demo.groovy file), which it doesn't. Never mind. This is good to go.

PR #4122, opened 2017-02-03
Bug 1335671. Remove background image from tag editor.

jpetto:


jwhitlock:
Works locally for me, change is laser focused. I am disappointed that there is no laser emoji, so pretend this sparkler is a laser: 🎇

PR #4123, opened 2017-02-06
bug 1031910: Fix language detection for zones

jwhitlock:
The language detection regex worked for regular URLs, but not zone URLs (URLs that omit /docs/). The new code works on both kinds:

https://developer.mozilla.org/en-US/docs/Web/HTML
https://developer.mozilla.org/en-US/Add-ons/Themes


jpetto:
Looks good to me. 👍

PR #4124, opened 2017-02-07
bug 957802: Show locale diff only on string change

jwhitlock:
In TravisCI, show the locale diff only if the localizable string changes. This makes it clearer when a string update is needed.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,9 +34,11 @@ setenv =
     COMPOSE_FILE = docker-compose.yml:docker-compose.test.yml
     COMPOSE_PROJECT_NAME = localetox
     GIT_PAGER = cat
+# Test that locales are refreshed w/o error,
+# then display diff only if messages changed
 commands =
     docker-compose run noext make localerefresh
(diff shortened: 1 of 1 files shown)


jwhitlock:
I can now show the why of this change.

In the master branch, the locale branch demonstrates that the locales still extract and compile, which can be broken by code or by a bad translation in Pontoon. However, the [output is lengthy](https://travis-ci.org/mozilla/kuma/jobs/199246739#L3110) and mostly says "nothing has changed".  The "POT-Creation-Date" always changes, and a change to base template has changed the file names, but none of the strings have changed, so a string update is not needed.

With this change, there is [no output](https://travis-ci.org/mozilla/kuma/jobs/199294839#L3120), because none of the strings changed.

I've tested the other mode locally, making a change to a translatable string in jinja/base.html, regenerating locales (``make bash; make localeextract``) and confirming that the new diff shows the full output.

jwhitlock:
This only impacts TravisCI, not production, so self-merging.

PR #4126, opened 2017-02-09
docs and script for the creation of demo instances

metadave:
depends on https://github.com/mozilla/kuma/pull/4121

note: I'd like to change the `mdn-` prefix for branchs to the `mdn-demo-` prefix.


metadave:
ping @jgmize

metadave:
will push in a different branch

PR #4127, opened 2017-02-09
makefile target + script for creating demo k8s instances

metadave:


metadave:
@escattone that is a fantastic idea, I'll add that to this PR

escattone:
@metadave 

I'm not a Groovy programmer, but could something like this be used on doing that?

`env.BRANCH_NAME.matches("^[a-z][a-z0-9-]*[a-z0-9]$")`

metadave:
@escattone yes, that's excellent. We're prepending `mdn-demo-` to the branch name, so it may not have to be so restrictive.

escattone:
@metadave The `"$"` portion of the regex within the Groovy script was causing an exception, but fortunately it turns out we didn't need the boundary "matchers" (`"^"` and `"$"`) in the regex since the `matches` method already requires that the regex has to match the entire string (sorry, my bad). I added a commit to remove them, and now the `assert` does its job.

Jenkins doesn't report (in the curated output) that the failure was due to the `assert` though, so I'm wondering if we should do something like this instead of the `assert`?

```
if (!env.BRANCH_NAME.matches("[a-z0-9-]*[a-z0-9]")) {
    # report the error and fail
}
```

metadave:
I added one more piece to this PR:

add `instance_cleanup.sh` for SRE's to remove demo instances
- if a corresponding k8s namespace -> branch does not exist, offer to:
  - delete the namespace
  - add an annotation to keep the namespace. The next run of this script
    won't offer to delete the namespace, and a command will be displayed
    (but not run) to clear the annotation.
  - ignore the namespace
- if a corresponding k8s namespace -> branch DOES exist, don't do
  anything

metadave:
any more thoughts on this?

PR #4130, opened 2017-02-10
bug 1218563: Update install docs for Linux, .env-dist.dev, and the sample DB

jwhitlock:
* Move ``.env-dist`` to ``.env-dist.deis``, and update ``make deis-config`` to use it
* Create ``.env-dist.dev`` for development, and add notes to the install instructions for Linux developers.
* Update the install documents to use the sample database, and drop bits of the install instructions that aren't needed when you use it.

(*note, @metadave gave a thumbs up when it was just the first two items, may not have reviewed the new install instructions.*)


docs/installation.rst
```diff
@@ -43,6 +47,13 @@ Docker setup
 
         cd kuma
 
+#. Initialize and customize ``.env``. Linux users should to set the ``UID``
```
metadave: typo:  Linux users **should to** set the

metadave:
lgtm 👍

docs/installation.rst
jwhitlock: Fixed, force pushed

PR #4133, opened 2017-02-14
bug 1339375: Handle new params in GitHub emails

jwhitlock:
GitHub added a new parameter "[visibility](https://developer.github.com/v3/users/emails/#toggle-primary-email-visibility)" to reflect if a GitHub user's primary email address is visible on their profile.  Update the signup form to only load the expected variables.


escattone:
Looks great, thanks @jwhitlock!

PR #4134, opened 2017-02-16
[Bug 1325898] Logout with Post method

safwanrahman:
Now it logout with POST request.
As I do not have much knowledge in javascript, so there maybe some bug in that.
@jwhitlock r?


kuma/static/js/auth.js
```diff
@@ -50,6 +50,16 @@
         });
     });
 
+    // post request to logout user
+    var $logoutButton = $('#logout-button');
+    $logoutButton.click( function (e, service) {
```
jwhitlock: what is ``service``?  It should probably be removed.

kuma/static/js/auth.js
```diff
@@ -50,6 +50,16 @@
...
+        var url = $logoutButton.data('logout-url');
+        $.ajax(url, {
+            cache: false,
+            method: 'post'
+        })
```
jwhitlock: Missing semicolon ``;``

kuma/static/js/auth.js
```diff
@@ -50,6 +50,16 @@
         });
     });
 
+    // post request to logout user
+    var $logoutButton = $('#logout-button');
+    $logoutButton.click( function (e, service) {
+        var url = $logoutButton.data('logout-url');
+        $.ajax(url, {
+            cache: false,
+            method: 'post'
```
jwhitlock: should be ``POST``

kuma/static/js/auth.js
```diff
@@ -50,6 +50,16 @@
         });
     });
 
+    // post request to logout user
+    var $logoutButton = $('#logout-button');
+    $logoutButton.click( function (e, service) {
+        var url = $logoutButton.data('logout-url');
```
jwhitlock: Missing a ``e.preventDefault();`` to avoid following the click event.

kuma/static/js/auth.js
```diff
@@ -50,6 +50,16 @@
...
+        $.ajax(url, {
```
jwhitlock: I believe that a successful form submit will need the CSRF token in the data, which isn't available in the page content.

kuma/users/jinja2/account/logout.html
```diff
@@ -8,12 +8,11 @@
...
 
   <p>{{ _("Are you sure you want to sign out?") }}</p>
-
-  <form method="post" action="{{ url('account_logout') }}">
+  <form method="post" action="{{ url('allauth_account_logout') }}">
     {% csrf_token %}
-    {% if redirect_field_value %}
-    <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}"/>
-    {% endif %}
+      {% if redirect_value %}
```
jwhitlock: ``redirect_field_value`` is the correct [context variable name](https://github.com/pennersr/django-allauth/blob/0.24.1/allauth/account/views.py#L665)

kuma/users/views.py
```diff
@@ -644,3 +644,11 @@ def recover(request, uidb64=None, token=None):
 
 recover_done = login_required(never_cache(ConnectionsView.as_view(
     template_name='users/recover_done.html')))
+
+
+def logout(request):
```
jwhitlock: I don't see why allauth's [logout view](https://github.com/pennersr/django-allauth/blob/0.24.1/allauth/account/views.py#L634) should be reimplemented.

kuma/users/urls.py
```diff
@@ -98,4 +98,6 @@
         views.my_edit_page,
         name='users.my_edit_page'),
     url(r'^users/', include(users_patterns)),
+    url(r'^signout/?$', views.logout,
```
jwhitlock: Why are you adding a second logout view?

jinja2/includes/common_macros.html
```diff
@@ -37,7 +37,10 @@
       <nav>
         <a href="{{ user.get_absolute_url() }}">{{ _('Profile') }}</a>
         |
-        <a href="{{ url('account_logout') }}">{{ _('Sign Out') }}</a>
+        <a id="logout-button" data-logout-url="{{ url('allauth_account_logout') }}"
```
jwhitlock: This may need to be a form to implement the POST correctly.  However a form may break layout, so maybe there needs to be a hidden form that gets called from JS.

jinja2/includes/login.html
```diff
@@ -1,7 +1,9 @@
 <div class="login">
     {% if user.is_authenticated() %}
         <a href="{{ user.get_absolute_url() }}">{{ user_display(user) }}</a>
-        <a href="{{ url('account_logout') }}?next={{ next_url }}">{{ _('Sign out') }}</a>
+        <a id="logout-button" data-logout-url="{{ url('allauth_account_logout') }}"
```
jwhitlock: Again, this may need to be implemented with the assistance of a hidden form.

kuma/settings/common.py
```diff
@@ -1408,7 +1408,6 @@ def get_user_url(user):
                                        default='https')
 ACCOUNT_EMAIL_VERIFICATION = 'mandatory'
 ACCOUNT_EMAIL_REQUIRED = True
-ACCOUNT_LOGOUT_ON_GET = True
```
jwhitlock: 👍

jpetto:
Outside of the context of this PR, you can use pure JS to perform a POST - no need for a `<form>`.

The logic in the PR is valid. Clicking a link with JS enabled performs an AJAX POST.

However, I'm not clear on why we need different behavior for JS/non-JS users here. Why couldn't we perform a POST for all users? As @jwhitlock alluded to, we'd need to change the `<a>` tags to a `<form>` (with a `csrf` token), which may require some CSS work, but seems like the more sound solution (unless I'm missing something).

If AJAX is the way to go, and we need to send data along with the POST, you can include URL encoded information with the AJAX POST with the `data` property, e.g.

```
$.ajax({
    url: url,
    data: 'foo=bar&csrf=8675309'
});
```

safwanrahman:
@jwhitlock Sorry, I did not read the source of ``django-allauth``. I have read that and found they already implemented the logout view with get.
As @jpetto mentioned, I have added the post method by using html ``forms`` and change in the ``CSS`` to look it like before.
So both js activate users and non js users will get same experience.
@jwhitlock @jpetto r?

Willyntobila:
Je suis francophone, je ne comprends pas l'anglais. Aidez-moi à faire fonctionner mon appareil mobile fore os.

jinja2/includes/login.html
```diff
@@ -1,12 +1,16 @@
 <div class="login">
     {% if user.is_authenticated() %}
-        <a href="{{ user.get_absolute_url() }}">{{ user_display(user) }}</a>
-        <a href="{{ url('account_logout') }}?next={{ next_url }}">{{ _('Sign out') }}</a>
+        <form action="{{ url('account_logout') }}" method="post">
+            {% csrf_token %}
+            <a class='user-url' href="{{ user.get_absolute_url() }}">{{ user_display(user) }}</a>
```
jpetto: For better semantic structure, let's put this `<a>` tag before the `<form>`.

jinja2/includes/common_macros.html
```diff
@@ -37,7 +37,10 @@
       <nav>
         <a href="{{ user.get_absolute_url() }}">{{ _('Profile') }}</a>
         |
-        <a href="{{ url('account_logout') }}">{{ _('Sign Out') }}</a>
+          <form action="{{ url('account_logout') }}" method="post">
+            {% csrf_token %}
+            <button class="logout" type="submit">{{ _('Sign Out') }}</button>
+          </form>
```
jpetto: Is this macro in use anywhere? If not, should we get rid of it?

kuma/static/styles/base/elements/forms.scss
```diff
@@ -99,3 +99,14 @@ label {
         text-decoration: none;
     }
 }
+
+button[class='logout'] {
```
jpetto: Can simplify this selector to `.logout`.

kuma/static/styles/base/elements/forms.scss
```diff
@@ -99,3 +99,14 @@ label {
...
+        background-color: transparent;
+        text-transform: none;
+        padding: 0;
+        border: 0;
+        color: white;
```
jpetto: Nit pick - use `#fff` instead of color name.

kuma/static/styles/components/structure/login.scss
```diff
@@ -16,7 +16,7 @@ Header login link
         vertical-align: top; /* needed on mobile */
     }
 
-    > a:nth-child(n+2) {
+    > a, button:nth-child(n+2) {
```
jpetto: I would suggest moving the new `.logout` rule from `forms.scss` to this file, adding the `bidi-style` rules below to it, then removing this selector entirely.

jinja2/includes/common_macros.html
jwhitlock: Interesting. It appears [the macro greeting is used nowhere](https://github.com/mozilla/kuma/search?utf8=✓&q=greeting), and the whole macro can be removed. 👍

kuma/static/styles/components/structure/login.scss
safwanrahman: As you have mentioned earlier, need to place the login ``<a>`` before the form. So should we still do that?
@jpetto Can you suggest what will be the selector? I dont have much idea about CSS.

safwanrahman:
@jwhitlock Maybe now its ready to merge!

safwanrahman:
Oh! Need to fix test.
Will lookover in to this weekend

jwhitlock:
Yes, looking good. @safwanrahman please remove the macro ``greeting`` from common_macros.html  as well.

jwhitlock:
I had a couple of false alarms while testing - a ``make build-static`` is needed to apply the styles, and a restart to apply the settings change.  Once the code it running, it works well.

The style looks fine on the homepage:

logout on homepage

but the text color is wrong on other pages, like the profile page:

logout on profile

@jpetto, do you think you can fix the style on the alternate header?

jwhitlock:
Looks good, works well. Thanks @safwanrahman and @jpetto!

PR #4135, opened 2017-02-16
Finish first-pass at soapbox message & maintenance mode page.

jpetto:
Also hide login and edit buttons side-wide.


jpetto:
@escattone - Give it a look and merge if happy? I think this gets us pretty close on the UI.

jinja2/base.html
```diff
@@ -69,11 +73,24 @@
     {% include "includes/google_analytics.html" %}
   {% endif %}
 </head>
-<body {% block body_attributes %}{% endblock %} class="{% block bodyclass %}{% endblock %}"{% if request.user and request.user.is_authenticated() %} data-login-service="{{ request.session.sociallogin_provider }}"{% endif %}>
+<body {% block body_attributes %}{% endblock %} class="{% block bodyclass %}{% endblock %}{% if settings.MAINTENANCE_MODE %} maintenance-mode{% endif %}"{% if request.user and request.user.is_authenticated() %} data-login-service="{{ request.session.sociallogin_provider }}"{% endif %}>
 
   {% include "includes/config.html" %}
 
-  {{ soapbox_messages(get_soapbox_messages(request.path)) }}
+  {%- if not settings.MAINTENANCE_MODE %}
```
safwanrahman: Is it possible to use like following? so its more readable

```python
{%- if settings.MAINTENANCE_MODE %}
  <div class="global-notice maintenance-mode-notice">
  .....
{%- else %}
  {{ soapbox_messages(get_soapbox_messages(request.path)) }}
{%- endif %}
```

jpetto: Good suggestion. Will change.

PR #4142, opened 2017-02-27
bug 1341770: Add "Experiment:" prefix

jwhitlock:
Add a new ``Experiment:`` prefix for content experiments.

- Add a warning for viewers
- Omit from internal search
- Mark nofollow, noindex for external search crawlers

This change is needed to prevent SEO problems w/ highly duplicative content.  Further PRs will add the framework for A/B content experiments.

This is already at use at [Experiment:FrameworkTest/Comparison_Operators](https://developer.mozilla.org/en-US/docs/Experiment:FrameworkTest/Comparison_Operators), and is an exact copy of [Web/JavaScript/Reference/Operators/Comparison_Operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators).


kuma/wiki/models.py
```diff
@@ -1492,6 +1493,10 @@ def zone_stack(self):
     def get_full_url(self):
         return absolutify(self.get_absolute_url())
 
+    @property
+    def is_experiment(self):
+        return self.slug.startswith(EXPERIMENT_TITLE_PREFIX)
```
jpetto: Testing locally, I cloned an article (`docs/Web/JavaScript/Reference/Functions/Arrow_functions`), which resulted in a new article with the following URL: `docs/Web/JavaScript/Reference/Functions/Experiment:_Arrow_functions`

As the slug doesn't start with `Experiment:`, this article is not flagged as such. Did I do something wrong/unanticipated, or do we need to see if the slug contains `Experiment:` (as opposed to `startswith`)?

jwhitlock: It only counts at the start of the slug, like docs/Experiment:Arrow_functions.

jpetto: Ah, okay, so I was doing it wrong. We're assuming all experiments will be at the `/docs/` root (and most likely not following the "control" article's URL structure).

jpetto:
Everything seems to be in order here. Works as expected, new tests are solid. r+ 👍

PR #4144, opened 2017-03-02
[Bug 1269104] Compare a new translation to the English document

safwanrahman:
So its possible to compare with the english document. In order to compare, I have chosen the revision that the translation based on.
Also refactored the queryset because I think its overkilling and its possible to fetch the data with simple queryset.
@jwhitlock Should we add test for this change?
r?

### DIFF
--- a/kuma/wiki/views/list.py
+++ b/kuma/wiki/views/list.py
@@ -190,14 +190,14 @@ def revisions(request, document_slug, document_locale):
     # Grab revisions, but defer summary and content because they can lead to
     # attempts to cache more than memcached allows.
     all_revisions = (document.revisions.defer('summary', 'content').order_by('created', 'id')
-                                                                   .reverse().transform(get_previous))
+                     .select_related('creator').reverse().transform(get_previous))
 
     if not all_revisions.exists():
         raise Http404
 
     if per_page == 'all':
         page = None
-        revisions = list(all_revisions)
+        all_revisions = list(all_revisions)
     else:
         try:
             per_page = int(per_page)
@@ -205,19 +205,19 @@ def revisions(request, document_slug, document_locale):
             per_page = DOCUMENTS_PER_PAGE
 
         page = paginate(request, all_revisions, per_page)
-        revisions = list(page.object_list)
+        all_revisions = list(page.object_list)
     # In order to compare the first revision of a translation, need to insert its parent revision to the list
     # The parent revision should stay at last page in order to compare. So insert only if there are no next page or
     # all revisions are showing
     if (not page or not page.has_next()) and document.parent:
         # *all_revisions are in descending order. so call last() in order to get first revision
-        first_rev_based_on = all_revisions.last().based_on
+        first_rev_based_on = all_revisions[-1].based_on
         # Translation can be orphan so that first revision does not have any english based on. So handle the situation.
         if first_rev_based_on:
-            revisions.append(first_rev_based_on)
+            all_revisions.append(first_rev_based_on)
 
     context = {
-        'revisions': revisions,
+        'revisions': all_revisions,
         'document': document,
         'page': page,
     }
--- a/kuma/wiki/views/list.py
+++ b/kuma/wiki/views/list.py
@@ -210,7 +210,11 @@ def revisions(request, document_slug, document_locale):
     # The parent revision should stay at last page in order to compare. So insert only if there are no next page or
     # all revisions are showing
     if (not page or not page.has_next()) and document.parent:
-        revisions.append(all_revisions.last().based_on)
+        # *all_revisions are in descending order. so call last() in order to get first revision
+        first_rev_based_on = all_revisions.last().based_on
+        # Translation can be orphan so that first revision does not have any english based on. So handle the situation.
+        if first_rev_based_on:
+            revisions.append(first_rev_based_on)
 
     context = {
         'revisions': revisions,
--- a/kuma/wiki/views/list.py
+++ b/kuma/wiki/views/list.py
@@ -8,9 +8,8 @@ from kuma.core.utils import paginate
 
 from ..constants import DOCUMENTS_PER_PAGE
 from ..decorators import process_document_path, prevent_indexing
-from ..models import (Document, DocumentTag, Revision, ReviewTag,
+from ..models import (Document, DocumentTag, ReviewTag,
                       LocalizationTag)
-from ..queries import MultiQuerySet
 
 
 @block_user_agents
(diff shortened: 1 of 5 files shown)


kuma/wiki/views/list.py
```diff
@@ -190,30 +189,28 @@ def get_previous(revisions):
...
 
-        page = paginate(request, revisions, per_page)
-        revisions = page.object_list
+        page = paginate(request, all_revisions, per_page)
+        revisions = list(page.object_list)
+    # In order to compare the first revision of a translation, need to insert its parent revision to the list
+    # The parent revision should stay at last page in order to compare. So insert only if there are no next page or
+    # all revisions are showing
+    if (not page or not page.has_next()) and document.parent:
+        revisions.append(all_revisions.last().based_on)
```
jwhitlock: Not all revisions have a ``based_on`` set.  123228 of the 261724 non-English translations in production (47%) do not.  It is possible none of the first translations do 😢.

safwanrahman:
@jwhitlock I have updated the patch and now it should handle the situation while there are no ``based_on`` for the first revision of a translation.
Can you please review again? r?

jwhitlock:
This takes care of the internal server error. However, I still can't compare a translation to the English document in the $history view. Does this change work for you?

safwanrahman:
> This takes care of the internal server error. However, I still can't compare a translation to the English document in the $history view. Does this change work for you?

@jwhitlock Yes. I have tested it and it works for me. Can you make a demo instance so we can check together?

jwhitlock:
I rebased your branch on master (to get demo script changes added in the last few weeks), added a commit to launch a demo based on it, and pushed as https://github.com/mozilla/kuma/tree/compare . You'll probably want to rebase your own branch as well. The demo is at https://mdn-demo-compare.virginia.moz.works .

I'd expect to see an option to compare to the first English document on:

http://mdn-demo-compare.virginia.moz.works/fr/docs/User:anonymous:uitest$history

safwanrahman:
@jwhitlock Thanks a lot for rebasing and creating a demo instance.

The compare with english locale was not showing because the scrapper was not linking the translation revision with its parent ``based_on`` revision.
I have created a new translation in bn-IN, and it seems that its possible to see the comparison in the history

http://mdn-demo-compare.virginia.moz.works/bn-IN/docs/User:anonymous:uitest$history

jwhitlock:
I get the same results.  I guess it will be useful going forward, and we back-populated ``based_on`` if there is a call for it.

I'd like some indication on the history that the first revision is the English source.  Maybe just "English (US)" at the start of the description field, localized in that language, in an <em> tag.

The drop in coverage is worrisome.  Now that it is working, I'll take a closer look at the PR.

kuma/wiki/views/list.py
```diff
@@ -190,30 +189,32 @@ def get_previous(revisions):
...
-                         .transform(get_previous)),
-        (Revision.objects.filter(document=document)
-                         .order_by('-created', '-id')
-                         .exclude(pk=document.current_revision.pk)
-                         .prefetch_related('creator', 'document')
-                         .transform(get_previous))
-    )
-
-    if not revisions.exists():
+    all_revisions = (document.revisions.defer('summary', 'content').order_by('created', 'id')
```
jwhitlock: I enabled Django Debug Toolbar locally to investigate.  This new version takes 22 queries in 19.23 ms, and the old version takes 17 queries in 10.56ms, on Mozilla/Firefox, which has 15 revisions in the sample database.  I believe the extra requests is because the ``creator`` is pre-fetched in the previous version, and in the new one it requires an additional database query.  I think a ``prefetch_related('creator')`` is needed, to better compare the two. I think ``prefetch_related('document')`` can be skipped, but I might be guessing at Django database query implementation.

kuma/wiki/views/list.py
```diff
@@ -190,30 +189,32 @@ def get_previous(revisions):
...
+                                                                   .reverse().transform(get_previous))
```
jwhitlock: I don't actually know what ``transform()`` is.  More research is needed.

kuma/wiki/views/list.py
```diff
@@ -8,9 +8,8 @@
 
 from ..constants import DOCUMENTS_PER_PAGE
 from ..decorators import process_document_path, prevent_indexing
-from ..models import (Document, DocumentTag, Revision, ReviewTag,
+from ..models import (Document, DocumentTag, ReviewTag,
                       LocalizationTag)
-from ..queries import MultiQuerySet
```
jwhitlock: This was the only user of MultiQuerySet. That code should be deleted if we end up not using it.

kuma/wiki/views/list.py
```diff
@@ -190,30 +189,32 @@ def get_previous(revisions):
...
     else:
         try:
             per_page = int(per_page)
         except ValueError:
             per_page = DOCUMENTS_PER_PAGE
 
-        page = paginate(request, revisions, per_page)
-        revisions = page.object_list
+        page = paginate(request, all_revisions, per_page)
+        revisions = list(page.object_list)
```
jwhitlock: 👍 using ``all_revisions``, so that we don't reuse the name ``revisions`` here.

kuma/wiki/views/list.py
```diff
@@ -190,30 +189,32 @@ def get_previous(revisions):
...
+    # In order to compare the first revision of a translation, need to insert its parent revision to the list
+    # The parent revision should stay at last page in order to compare. So insert only if there are no next page or
+    # all revisions are showing
+    if (not page or not page.has_next()) and document.parent:
+        # *all_revisions are in descending order. so call last() in order to get first revision
+        first_rev_based_on = all_revisions.last().based_on
```
jwhitlock: I think this could be ``revisions[-1]``, and save a trip to the database.

kuma/wiki/views/list.py
jwhitlock: Found it: PR #2546.  Continue with the suggested changes, and we'll see what the perf is like with "all revisions" and a big revision history.

safwanrahman:
@jwhitlock Rebased and updated according to your comments.
I have used ``select_related`` because I think its more optimized.
r?

kuma/wiki/views/list.py
```diff
@@ -190,33 +189,35 @@ def get_previous(revisions):
...
-        page = paginate(request, revisions, per_page)
-        revisions = page.object_list
+        page = paginate(request, all_revisions, per_page)
+        all_revisions = list(page.object_list)
+    # In order to compare the first revision of a translation, need to insert its parent revision to the list
+    # The parent revision should stay at last page in order to compare. So insert only if there are no next page or
+    # all revisions are showing
+    if (not page or not page.has_next()) and document.parent:
+        # *all_revisions are in descending order. so call last() in order to get first revision
+        first_rev_based_on = all_revisions[-1].based_on
```
jwhitlock: There needs to be a new test that exercises this line.

jwhitlock:
The full $history is down to around 60 queries, so the ``select_related`` works.  It takes 90 seconds locally on Web/HTML, so there is still some optimization. I can do this in a follow-on commit.

I think we need a test that exercises the new code before merging.

safwanrahman:
@jwhitlock I have added test for it and changed in templates as you have commented.
I think its possible to merge now.

jwhitlock:
Thanks @safwanrahman!

PR #4146, opened 2017-03-06
Add Traffic Cop JS + initial experiment config.

jpetto:


jinja2/base.html
```diff
@@ -4,6 +4,9 @@
 <html lang="{{ LANG }}" dir="{{ DIR }}" class="redesign no-js" data-ffo-opensans="false" data-ffo-opensanslight="false">
 <head {%- if not is_sphinx %} prefix="og: http://ogp.me/ns#"{% endif %}>
   <meta charset="utf-8">
+
+  {% block experiment_js %}{% endblock %}
+
```
jwhitlock: I know it gets compressed away, but

SO

MUCH

WHITESPACE

kuma/static/js/experiment-wiki-content.js
```diff
@@ -0,0 +1,16 @@
+(function() {
+    'use strict';
+
+    // cookie will expire in 24 hours
+    var lou = new Mozilla.TrafficCop({
```
jwhitlock: [I didn't get it](http://simpsons.wikia.com/wiki/Lou).

PR #4148, opened 2017-03-07
Add simple in-line editorial comments feature

a2sheppy:
Adds a new CSS class, editorial-comment, which makes a &lt;div&gt; that is presented
in a nice box but only when in the editor; this is meant to be used for leaving
notes to self and/or other writers, review comments, or to place information that will be useful
when it's time to make upcoming changes to the article without having the editorial notes displayed to most users.

The new style can be applied by selecting "Editorial Note" from the Style
drop-down in the editor's toolbar.

### DIFF
--- a/kuma/static/styles/base/utilities.scss
+++ b/kuma/static/styles/base/utilities.scss
@@ -56,6 +56,36 @@
     }
 }
 
+/* Editorial comments; don't display these unless in the editor. */
+
+.editorial-comment {
+  display: none;
+
+  html.cke_panel_container &,
+  body[contenteditable] & {
+    position: relative;
+    display: block;
+    border: 1px solid #3e3544;
+    padding: 6px 3px 3px;
+    background-color: #e9c0ff;
+    margin-bottom: 9px;
+
+    &:before {
+      @include set-font-size(16px);
+      content: "Editorial Note";
+      position: absolute;
+      top: -25px;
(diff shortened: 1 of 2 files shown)


a2sheppy:
I chose a color that's used elsewhere, but only in smaller quantities. We may want to use something else. If we didn't already have several gray backgrounds, I'd have used that. As it stands, something bold like this makes a degree of sense, because these notes should catch the eye.

stephaniehobson:
I've wrestled with this for a month because I can see how useful it would be to the writers to have this but I keep coming back to it being a bad idea to include content we don't want to be part of the article as part of the article.

My major concerns are:
1) Content being rendered without the CSS.  (I don't expect this to happen much on our site but it does happen when other people pull our content in and re use it. It also happens over time as the CSS changes. In a year or two someone other than you or me could decide to strip the class.)
2) The label is not localizable and could be confusing to localizers.
3) Search engines indexing and displaying the comments as part of search results.

Solutions I would be willing to accept:
- Using HTML comments. (I looked and sadly there is not a CKEditor plug in to expose these in the WYSIWYG.)

Solutions I would be willing to consider with a 3rd opinion:
- HTML elements that have inline display:none styles on them. (These should still be able to be made visible with an !important rule in the style sheets.)

If you find a way to make this work, you get to explain the colour choice to Jean Yves :P

PR #4150, opened 2017-03-08
Maintenance Mode (Bug 1313237)

escattone:
This is the maintenance-mode feature for Kuma. It provides the ability to run Kuma in a read-only mode, where any endpoint that would write to the database is redirected to a maintenance-mode description page, an announcement banner is displayed across the top of each page, and any buttons/links that provided the ability to edit/revert/etc. have been hidden.

Thanks to @jpetto for his great work in handling all of the front-end tasks (the maintenance-mode banner and page, the hiding of specific buttons/links, and so on).

On the backend, the core strategy has been to tread as lightly as possible:
- ensure that the user is always anonymous by ignoring the session cookie or authorization key
- redirect (to the maintenance-mode page) on all login endpoints (this in-turn automatically covers all endpoints that require login)
- explicitly redirect on all other endpoints that would write to the database
- some operations that were essentially read-only except that they would cache their ultimate result in the database (like Document.get_json_data()) have been modified so that they can be safely called in maintenance mode
- Celery is enabled but uses memcached for its result backend and the default (Python's file-based shelve) for the beat scheduler, both instead of the database
- only a small subset of Celery tasks are allowed (e.g., cacheback.tasks.refresh_cache, which is used for getting the contributors for every document request), while all others, if called, will be skipped

A set of integration tests, marked with "maintenance_mode", have been created to cover all endpoints that should redirect, as well as a few other endpoints to ensure that specific buttons/links are hidden and that the banner is displayed.

When running the maintenance-mode integration tests against a local Kuma using the latest sample database (March 9, 2017 or later) in maintenance mode, use this command (from within your virtualenv outfitted with ``pip install -r requirements/test.txt``):

``py.test -m "maintenance_mode and not search" tests/functional --driver Chrome --base-url http://localhost:8000 --driver-path path/to/chromedriver``

Before you do that, however, you will need to render the ``en-US/docs/User:anonymous:uitest`` document. Restart your Docker instance so that it's NOT in maintenance mode and is using the write-privileged MySQL user -- ``root:kuma`` rather than ``kuma_ro:kuma`` -- and then do a shift-reload within your browser of ``http://localhost:8000/en-US/docs/User:anonymous:uitest``. Then put your Docker instance back in maintenance mode, and you should get a clean run of the tests above.

Note that the "maintenance_mode and not search" tests are run against the sample database. The "search" tests are left out because they're written for the full staging/production database.


Jenkinsfiles/maint-mode.yml
```diff
@@ -0,0 +1,3 @@
+pipeline:
```
safwanrahman: This file should be removed

escattone: Hi @safwanrahman! This file is required so that when we force push this branch to the ["maint-mode" branch](https://github.com/mozilla/kuma/tree/maint-mode), we get an automatic deployment to AWS (specifically to https://mdn-demo-maint-mode.virginia.moz.works). We couldn't use this branch name because the resulting name ("mdn-demo-maintenance-mode-1313237" violated a length limit enforced by one of our deployment tools, I think it was Deis?).

safwanrahman: @escattone I believe, this code should not live in master branch.
I think you should find a way to not merge this to master!
maybe after approving the PR, you can remove the file and make a commit?

jwhitlock: @safwanrahman It is unclear why you object to this file. It seems fine to me, and I appreciate the live server.

kuma/dashboards/views.py
```diff
@@ -77,19 +78,26 @@ def revisions(request):
...
-            if authors_filter == str(RevisionDashboardForm.KNOWN_AUTHORS):
-                query_kwargs['creator__groups__pk'] = group.pk
-            # Else query must be 'Unknown Authors', so exclude the
-            # 'Known Authors' group
+            # Get the 'Known Authors' group.
+            if settings.MAINTENANCE_MODE:
+                try:
+                    group = Group.objects.get(name="Known Authors")
+                except Group.DoesNotExist:
+                    group = None
```
jwhitlock: The ``Group.objects.get_or_create(name="Known Authors")`` was a hack from when Known Authors was a new group and not in production.  I'm leaning toward retiring the auto-creation, and instead ignore the filter if the group isn't created. In other words, make the ``MAINTENANCE_MODE`` solution the read-write solution as well.  If it makes it harder (such as significant test changes), we can do it in a future PR.

kuma/landing/jinja2/landing/maintenance-mode.html
```diff
@@ -0,0 +1,61 @@
...
+
+        <div id="beastainer">
+          <div id="servers"></div>
+          <div id="beast">
+            <div id="beast-eyes"></div>
+            <div id="beast-body"></div>
+          </div>
+          <div class="cloud" id="cloud-large"></div>
+          <div class="cloud" id="cloud-small"></div>
+        </div>
```
jwhitlock: ❤️ http://mdn-demo-maint-mode.virginia.moz.works/en-US/maintenance-mode

kuma/settings/common.py
```diff
@@ -404,16 +406,24 @@ def enable_candidate_languages():
     # LocaleURLMiddleware must be before any middleware that uses
     # kuma.core.urlresolvers.reverse() to add locale prefixes to URLs:
     'kuma.core.middleware.SetRemoteAddrFromForwardedFor',
-    'django.contrib.sessions.middleware.SessionMiddleware',
+    ('kuma.core.middleware.ForceAnonymousSessionMiddleware'
+     if MAINTENANCE_MODE else
+     'django.contrib.sessions.middleware.SessionMiddleware'),
```
jwhitlock: 👍 to Python trigraphs

kuma/settings/common.py
```diff
@@ -404,16 +406,24 @@ def enable_candidate_languages():
...
     'kuma.core.middleware.RemoveSlashMiddleware',
-    'commonware.middleware.NoVarySessionMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'django.middleware.csrf.CsrfViewMiddleware',
+)
+
+if not MAINTENANCE_MODE:
+    # We don't want this in maintence mode, as it adds "Cookie"
+    # to the Vary header, which in turn, kills caching.
+    MIDDLEWARE_CLASSES += ('django.middleware.csrf.CsrfViewMiddleware',)
```
jwhitlock: 😦 couldn't do it with a trigraph. Oh well.

kuma/urls.py
```diff
@@ -22,20 +22,41 @@
...
-    # Django admin:
-    url(r'^admin/wiki/document/purge/',
-        purge_view,
-        name='wiki.admin_bulk_purge'),
-    url(r'^admin/', include(admin.site.urls)),
+]
+
+if settings.MAINTENANCE_MODE:
+    urlpatterns.append(
+        url(r'^admin/.*', lambda x: redirect('maintenance_mode'))
```
jwhitlock: 👍 for redirecting the admin, but I'd rather use ``RedirectView`` than a ``lambda``, such as http://stackoverflow.com/questions/15706489/redirect-to-named-url-pattern-directly-from-urls-py-in-django

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -243,6 +243,16 @@
...
 
+              {% if settings.MAINTENANCE_MODE and render_raw_fallback %}
+                <div id="doc-render-raw-fallback" class="warning">
+                {% trans url='/maintenance-mode/' %}
+                  This document has not yet been rendered, and cannot be
+                  rendered while MDN is in <a href="{{ url }}">maintenance
+                  mode</a>. Please check back when maintenance has concluded.
+                {% endtrans %}
+                </div>
+              {% endif %}
```
jwhitlock: 👍 I think this is the right solution, even if it looks really bad when you use the sample DB, which currently has no rendered docs ☹️

requirements/test.txt
```diff
@@ -14,6 +14,28 @@ requests==2.11.1 \
...
+    --hash=sha256:1c9c26bb6c31c3d5b3c104e843211d9c105db60b4df6770ac42673263d55d494 \
+    --hash=sha256:793e7cfe96042fa0a11a99a70dbe61ab9275f268ea813606f63dccb288f3a1ee \
+    --hash=sha256:bb3264f80d31eb991cfe427cc20c83508eb326454c9a01a006829fa2f541be72 \
+    --hash=sha256:f396094325c1ade0c1aa69100c7e760faefafb362f17e858a531f903aff29937 \
+    --hash=sha256:01e54511034333f18772c335ec0b33a76bba988135eaf727a075897866d19604 \
+    --hash=sha256:2abf6cac9b7952047d8b7265384a9565e419a727dba675e83e4b7f5b7892b6bb \
+    --hash=sha256:6dff909020d0c030fb26004626c8f87f9116e0381702fed415caf94f5a9b9493 \
+    --hash=sha256:192fec5a960abf261bd3733a33c10c359ebd2f9a5b714a8191f382e8fa127d42 \
+    --hash=sha256:78a1ec1e674ab6dc8a0644bb4401f3fb5787b636b0aaf0c9a720923491547b1b
+pyquery==1.2.11 \
```
jwhitlock: If there is time, I'd like this to user [constraints.txt](https://github.com/mozilla/kuma/blob/master/requirements/constraints.txt) like the [default.txt](https://github.com/mozilla/kuma/blob/master/requirements/default.txt) requirements file.

tests/conftest.py
```diff
@@ -27,7 +28,7 @@ def sensitive_url(request, base_url):
     if not base_url:
         return False
     else:
-        # developer-local and localhost are not sensative
+        # developer-local and localhost are not sensitive
```
jwhitlock: 👍speeling

kuma/urls.py
escattone: Thanks @jwhitlock , see https://github.com/mozilla/kuma/commit/02a93d58f16f527993dce68600136ed6369b29cc

tests/conftest.py
escattone: 😄

kuma/dashboards/views.py
escattone: Sounds good, thanks @jwhitlock (see commit ed70e1e289fef1fd64a6d1c1a4bcff08ab0bc291)

requirements/test.txt
escattone: I work on this tomorrow, thanks @jwhitlock !

escattone: @jwhitlock On second thought, can I do this work in a separate PR?

jwhitlock: a separate PR works for me

tests/functional/test_home.py
```diff
@@ -41,15 +45,16 @@ def test_callouts(base_url, selenium):
 @pytest.mark.smoke
 @pytest.mark.nodata
 @pytest.mark.nondestructive
+@pytest.mark.maintenance_mode
 def test_header_displays(base_url, selenium):
     page = HomePage(selenium, base_url).open()
-    assert page.Header.is_displayed
-    assert page.Header.is_menu_displayed
```
jwhitlock: Specifically, ``assert page.Header.is_menu_displayed`` was removed.

Jenkinsfiles/maint-mode.yml
safwanrahman: @jwhitlock I object because I think it should not be in master. After merging, this file will also be in master that It should not!

docs/development.rst
```diff
@@ -354,3 +354,108 @@ You can deploy a hosted demo instance of Kuma by following these steps:
...
+
+       docker-compose down
+       docker-compose up -d
+       docker-compose exec web ./manage.py compilejsi18n
+       docker-compose exec web ./manage.py collectstatic
+       docker-compose restart web
+
+Now you should be ready for a successful test run::
+
+    py.test -m "maintenance_mode and not search" tests/functional --base-url http://localhost:8000 --driver Chrome --driver-path /path/to/chromedriver
```
safwanrahman: Do there any reason we should use ``chromedriver`` instead of our ``geckodriver``?

Jenkinsfiles/maint-mode.yml
escattone: @safwanrahman Why do you think it should not be in the master branch? Is it because it only applies to the "maint-mode" branch and not the "master" branch? If that is the case, we already have other files in that category (see https://github.com/mozilla/kuma/blob/master/Jenkinsfiles/stage-integration-tests.yml for example). We need to keep it in the master branch because it allows any branch based on master to be pushed to the "maint-mode" branch and deployed as an AWS demo instance.

kuma/dashboards/views.py
```diff
@@ -77,19 +77,22 @@ def revisions(request):
...
-            # The 'Known Authors' group
-            group, created = Group.objects.get_or_create(name="Known Authors")
-            # If the filter is 'Known Authors', then query for the
-            # 'Known Authors' group
-            if authors_filter == str(RevisionDashboardForm.KNOWN_AUTHORS):
-                query_kwargs['creator__groups__pk'] = group.pk
-            # Else query must be 'Unknown Authors', so exclude the
-            # 'Known Authors' group
+            # Get the 'Known Authors' group.
+            try:
```
safwanrahman: Cool approach! 👍

tests/functional/test_home.py
escattone: My original intention was simply to fix the bug, i.e. change "page.Header.is_menu_displayed" to "page.header.is_menu_displayed", but after making that change I discovered that the "is_menu_displayed" property of the header simply doesn't work (it searches for an element with an id of "nav-main-menu", but there is no such element on any of our pages), so I removed all calls to "page.header.is_menu_displayed" in all of our tests and removed the property.

docs/development.rst
jwhitlock: My memory is that geckodriver and current Firefox requires Selenium 3, and we're using Selenium 2, here and maybe in Jenkins as well.  We should update, but not in this PR.  Maybe a contributor could give it a try...

docs/development.rst
```diff
@@ -354,3 +354,108 @@ You can deploy a hosted demo instance of Kuma by following these steps:
...
+    MAINTENANCE_MODE=True
+
+   and going with a database user that has write privileges. The read-only database
+   user simply provides a level of safety as well as notification (for example, an
+   exception will be raised if an attempt to write the database slips through).
+
+#. If you were already running your local Docker instance, you need to bring it
+   down hard and start again::
+
+    docker-compose down
```
escattone: @jwhitlock I didn't even know that ``docker-compose up`` was smart enough to recreate containers as needed (while preserving mounted volumes). I will update the docs accordingly. Thank you!

jinja2/base.html
```diff
@@ -161,12 +172,11 @@
...
         </div>
       </li><li><a href="{{ wiki_url('Tools') }}">{{ _('Developer Tools') }}</a></li><li><a href="{{ wiki_url('MDN/Feedback') }}">{{ _('Feedback') }}<i aria-hidden="true" class="icon-caret-down"></i></a>
-
         <div class="submenu js-submenu" id="nav-contact-submenu">
           <div class="submenu-column">
             <ul>
-                <li><a href="https://support.mozilla.org/">{{ _('Get Firefox help') }}<i aria-hidden="true" class="icon-external-link"></i></a></li>
-                <li><a href="http://stackoverflow.com/">{{ _('Get web development help') }}<i aria-hidden="true" class="icon-external-link"></i></a></li>
+              <li><a href="https://support.mozilla.org/">{{ _('Get Firefox help') }}<i aria-hidden="true" class="icon-external-link"></i></a></li>
+              <li><a href="http://stackoverflow.com/">{{ _('Get web development help') }}<i aria-hidden="true" class="icon-external-link"></i></a></li>
```
escattone: 👍  Let's do it!

escattone:
Ok, with @jpetto 's last commit addressing @jwhitlock 's suggestions for kuma/wiki/jinja2/wiki/document.html, we're good to merge!

jgmize:
Nice work, @escattone. I went ahead and deleted the branch now that it's merged, and I'd like to delete the demo deployment as well if you're done with it. Let me know if you're interested in working on automating the deployment deletion based on git branch deletion as the trigger.

escattone:
Thanks @jgmize. Yes, you're welcome to delete the demo deployment at will, and I am interested in working with you on that.

PR #4151, opened 2017-03-10
Bug 1341770: Add content experiments

jwhitlock:
Content experiments are A/B tests where some visitors see the Document content from a different page, based on a URL parameter. See [A/B Testing of wiki page content](https://docs.google.com/document/d/1CUe6-EqNbpSkLMLxfd0RWW6-TbPZFRtLc0rMh2Uu3G0/edit#) for the full spec.

This PR doesn't include functional tests, but should.  I plan to add those in a future PR, after PR #4150 is merged.

The first experiment will test out the framework. Anonymous users visiting [Web/JavaScript/Reference/Operators/Comparison_Operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators) will have a 50% / 50% chance of viewing the content from [Experiment:FrameworkTest/Comparison_Operators](https://developer.mozilla.org/en-US/docs/Experiment:FrameworkTest/Comparison_Operators), which is currently the same.

The backend:
* Selects pages and variants for the experiment with the ``CONTENT_EXPERIMENTS`` setting
* Injects [Traffic Cop](https://mozilla.github.io/meao/2017/01/16/traffic-cop/) when a page in the experiment is requested by an anonymous user
* Switches to variant content when a variant is requested
* Add the Google Analytics custom dimensions and disables editing when an anon user requests a variant

The front-end:
* Picks a variant for the experiment, storing the choice in a cookie
* Redirects the user from the original to the variant with a query string parameter

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -953,6 +953,15 @@ PIPELINE_JS = {
         ),
         'output_filename': 'build/js/ace.js',
     },
+    'experiment-wiki-content': {
+        'source_filenames': (
+            'js/libs/mozilla.dnthelper.js',
+            'js/libs/mozilla.cookiehelper.js',
+            'js/libs/mozilla.trafficcop.js',
+            'js/experiment-wiki-content.js',
+        ),
+        'output_filename': 'build/js/experiment-wiki-content',
+    },
 }
 
 PIPELINE = {
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -960,7 +960,16 @@ PIPELINE_JS = {
             'js/libs/mozilla.trafficcop.js',
             'js/experiment-wiki-content.js',
         ),
-        'output_filename': 'build/js/experiment-wiki-content',
+        'output_filename': 'build/js/experiment-wiki-content.js',
+    },
+    'experiment-framework-test': {
+        'source_filenames': (
+            'js/libs/mozilla.dnthelper.js',
+            'js/libs/mozilla.cookiehelper.js',
+            'js/libs/mozilla.trafficcop.js',
+            'js/experiment-framework-test.js',
+        ),
+        'output_filename': 'build/js/experiment-framework-test.js',
     },
 }
 
@@ -1463,6 +1472,11 @@ NOCAPTCHA = True  # Note: Using No Captcha implies SSL.
 TAGGIT_CASE_INSENSITIVE = True
 
 # Content Experiments
+# Must be kept up to date with PIPELINE_JS setting and the JS client-side
+#  configuration. The 'id' should be a key in PIPELINE_JS, that loads
+#  Traffic Cop and a client-side configuration like
+#  kuma/static/js/experiment-wiki-content.js
+#
 CONTENT_EXPERIMENTS = [{
     'id': 'experiment-framework-test',
     'ga_name': 'framework-test',
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1452,3 +1452,18 @@ NOCAPTCHA = True  # Note: Using No Captcha implies SSL.
 
 # Tell django-taggit to use case-insensitive search for existing tags
 TAGGIT_CASE_INSENSITIVE = True
+
+# Content Experiments
+CONTENT_EXPERIMENTS = [{
+    'id': 'experiment-framework-test',
+    'ga_name': 'framework-test',
+    'param': 'v',
+    'pages': [{
+        'locale': 'en-US',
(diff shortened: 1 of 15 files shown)


kuma/settings/common.py
```diff
@@ -1452,3 +1470,23 @@ def get_user_url(user):
 
 # Tell django-taggit to use case-insensitive search for existing tags
 TAGGIT_CASE_INSENSITIVE = True
+
+# Content Experiments
+# Must be kept up to date with PIPELINE_JS setting and the JS client-side
+#  configuration. The 'id' should be a key in PIPELINE_JS, that loads
+#  Traffic Cop and a client-side configuration like
+#  kuma/static/js/experiment-wiki-content.js
+#
```
jpetto: For safety's sake, may be worth adding a comment here that only one experiment should be active for a given `locale` + `slug`.

kuma/settings/common.py
```diff
@@ -1452,3 +1470,23 @@ def get_user_url(user):
...
+CONTENT_EXPERIMENTS = [{
+    'id': 'experiment-framework-test',
+    'ga_name': 'framework-test',
+    'param': 'v',
+    'pages': [{
+        'locale': 'en-US',
+        'slug': 'Web/JavaScript/Reference/Operators/Comparison_Operators',
+        'variants': [
+            ['control', 50, 'Web/JavaScript/Reference/Operators/Comparison_Operators'],
+            ['test', 50, 'Experiment:FrameworkTest/Comparison_Operators'],
```
jpetto: What's the reason behind storing the percentage here?

kuma/settings/common.py
```diff
@@ -953,6 +953,24 @@ def enable_candidate_languages():
...
+        'output_filename': 'build/js/experiment-wiki-content.js',
+    },
+    'experiment-framework-test': {
+        'source_filenames': (
+            'js/libs/mozilla.dnthelper.js',
+            'js/libs/mozilla.cookiehelper.js',
+            'js/libs/mozilla.trafficcop.js',
+            'js/experiment-framework-test.js',
+        ),
+        'output_filename': 'build/js/experiment-framework-test.js',
```
jpetto: This experiment is to be removed prior to merge, right?

jwhitlock: @atopal asked for this as the first experiment - see [Phase 1](https://docs.google.com/document/d/1CUe6-EqNbpSkLMLxfd0RWW6-TbPZFRtLc0rMh2Uu3G0/edit#heading=h.l8azyzynxmrc)

kuma/settings/common.py
jwhitlock: I alluded to this back in [PR 4146](https://github.com/mozilla/kuma/pull/4146#pullrequestreview-25550272). As a backend dev, it made sense to store all the data in the setting, and then generate some JSON to define the experiment for the front-end.  Instead, you defined the experiment in the .js file, so there's some duplication.  I can remove the percent from here as potentially confusing.

kuma/settings/common.py
jpetto: Oh yes, I remember reading that doc a while back. 👍

kuma/settings/common.py
jpetto: Gotcha. Agree we should only keep what's needed on the backend.

jwhitlock:
Pushed two new commits to address @jpetto's review comments.  I also launched a demo instance at http://mdn-demo-content-exp.virginia.moz.works/en-US/, however page rendering doesn't work in demo instances (no ``api`` service), so it's not a great demo.  Rendering works on staging, so it can be tested there, or in a local development environment.

* [Web/JavaScript/Reference/Operators/Comparison_Operators](http://mdn-demo-content-exp.virginia.moz.works/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators) - test page, will redirect you to the [control](http://mdn-demo-content-exp.virginia.moz.works/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators?v=control) or the [test variant](http://mdn-demo-content-exp.virginia.moz.works/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators?v=test)
* [Experiment:FrameworkTest](http://mdn-demo-content-exp.virginia.moz.works/en-US/docs/Experiment:FrameworkTest) - base Experiment page
* [Experiment:FrameworkTest/Comparison_Operators](http://mdn-demo-content-exp.virginia.moz.works/en-US/docs/Experiment:FrameworkTest/Comparison_Operators) - the variant source page

jpetto:
Tested locally, created `Experiment:FrameworkTest` pages and views render as expected. Tests pass - I think we're good here. ✅

PR #4152, opened 2017-03-10
Use gulp to process .scss files for local dev.

jpetto:
For local testing, `docker-compose stop`, then `npm install`, then `docker-compose up`. Start `gulp` in a new terminal (in your `kuma` root dir) and that should be that.

Also, ensure `DEBUG=True` in your `.env` file.

### DIFF
--- a/kuma/core/pipeline/sass.py
+++ b/kuma/core/pipeline/sass.py
@@ -1,0 +1,12 @@
+from pipeline.compilers.sass import SASSCompiler
+
+
+class DebugSassCompiler(SASSCompiler):
+    """
+    DEBUG=True replacement for standard SASSCompiler. Same behavior as
+    SASSCompiler (e.g. file extension match), but no compilation takes place
+    (handled by gulp).
+    """
+
+    def compile_file(self, *args, **kwargs):
+        return
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -98,27 +98,37 @@ See https://github.com/mozilla/kumascript.
 
 Front-end Development and Compiling Sass files
 ==============================================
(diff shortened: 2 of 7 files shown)


kuma/core/pipeline/sass.py
```diff
@@ -0,0 +1,13 @@
...
+class DebugSassCompiler(SubProcessCompiler):
+    output_extension = 'css'
+
+    def match_file(self, filename):
+        return filename.endswith(('.scss', '.sass'))
+
+    def compile_file(self, infile, outfile, outdated=False, force=False):
+        # as gulp is compiling .scss files when DEBUG=True, all we need to do
+        # here is give Pipeline the location of the pre-compiled .css file
+        return outfile
```
escattone: The output of the ``compile_file`` method is not used, so there is no need to return the ``outfile`` here. In fact, since we want ``DebugSassCompiler`` to act the same as ``pipeline.compilers.sass.SASSCompiler`` except that we don't want it to compile anything, I'd recommend this:
```
from pipeline.compilers.sass import SASSCompiler


class DebugSassCompiler(SASSCompiler):
    """
    This class is a drop-in replacement for SASSCompiler, and
    behaves like that class in all aspects except compilation.
    """

    def compile_file(self, *args, **kwargs):
        return
```
so we track any potential future changes to ``pipeline.compilers.sass.SASSCompiler``.

jpetto: Ah, good to know. I was assuming `outfile` was used to instruct Pipeline on file name/path for template injection, but apparently not. 😸 

Excellent call to piggy-back on existing `SASSCompiler`.

PR #4154, opened 2017-03-14
Add API service to demo deploys

jgmize:


jgmize:
https://ci.us-west.moz.works/blue/organizations/jenkins/mdn_multibranch_pipeline/detail/api/3/pipeline/

Makefile
```diff
@@ -132,6 +132,10 @@ k8s-migrate:
 	$(shell kubectl --namespace ${DEIS_APP} get pods | grep ${DEIS_APP}-cmd | awk '{print $$1}') \
 	python manage.py migrate
 
+render-k8s-templates:
+	k8s/api-svc.yaml.template.sh > k8s/api-svc.yaml
+	rm k8s/api-svc.yaml.template.sh
```
metadave: any reason for removing this script after rendering yaml?

k8s/api-svc.yaml.template.sh
```diff
@@ -0,0 +1,22 @@
...
+  namespace: ${APP}
+spec:
+  ports:
+  - name: http
+    port: 80
+    protocol: TCP
+    targetPort: 8000
+  selector:
+    app: ${APP}
+    heritage: deis
```
metadave: do we need `heritage` here?

metadave:
lgtm 👍

k8s/api-svc.yaml.template.sh
escattone: I don't think it's necessary, but perhaps safer in that we're more specific in targeting the "app: api" container created by Deis via the Procfile? Very thin reason though given that we don't have any pods labelled with "app: api" other than the one created via the Deis Procfile! :)

I'll leave it in for now.

Makefile
escattone: I'll try it without removing the script to see what happens.

escattone: It didn't seem to make any difference whether or not it was removed. Maybe it's safer to keep the removal just so there's no possibility of confusing ``kubectl``? I'll leave it in for now.

k8s/api-svc.yaml.template.sh
jgmize: Good catch, it's not necessary, just something I forgot to strip out when copying from another service.

Makefile
jgmize: I wasn't sure if kubectl would ignore unsupported filenames off the top of my head so I added the `rm` for the initial iteration with the plan of removing it later if it turned out not to be necessary. Thanks for testing @escattone; I'll go ahead and remove it now.

metadave: cool, we can also render text and pass it to stdin of `cat foo.json | kubectl create -f -` if needed.

PR #4158, opened 2017-03-21
[Bug 1052510] Should not possible to create page as child of redirect

safwanrahman:
This is code is actually based on kitsune.
I could not find any better way to handle this other than this way.
Copied and refactored and ported for kuma
@jwhitlock @escattone r?


jwhitlock:
This works well in my local development environment. I created some pages, and then used the "move page" feature to make them into redirects:

* /en-US/docs/user:jwhitlock/redirect1 to /en-US/docs/redirect1
* /en-US/docs/user:jwhitlock/redirect2 to /en-US/docs/Mozilla/Add-ons/redirect2 (zone URL)

I then attempted to create two new pages, and they were converted on the move page view.

* /en-US/docs/user:jwhitlock/redirect1/subpage converted to /en-US/docs/redirect1/subpage
* /en-US/docs/user:jwhitlock/redirect2/subpage converted to /en-US/docs/Mozilla/Add-ons/redirect2/subpage

So, we're 1/3 of the way there 😄.  We need a view test of a zoned URL, and unit tests w/ coverage for the new methods (``Document.from_url()``, ``Document.get_redirect_document()``, ``get_doc_components_from_url``.  There may be existing tests in kitsune.  Bonus points for converting them to py.test native tests (``assert x==y`` instead of ``self.assertEqual(x, y)`` or ``_eq(x, y)``.

Detailed code review to follow.

kuma/wiki/models.py
```diff
@@ -1359,6 +1400,14 @@ def get_redirect_url(self):
                 elif len(url) == 1 and url[0] == '/':
                     return url
 
+    def get_redirect_document(self, id_only=False):
```
jwhitlock: ``id_only`` is ``True`` for the only caller, and should be the default.

kuma/wiki/models.py
```diff
@@ -380,6 +380,47 @@ def get_summary_html(self, *args, **kwargs):
     def get_summary_text(self, *args, **kwargs):
         return self.get_summary(strip_markup=True)
 
+    @classmethod
+    def from_url(cls, url, required_locale=None, id_only=False,
```
jwhitlock: ``id_only`` is ``True`` for the only caller, and should be the default.

kuma/wiki/models.py
```diff
@@ -380,6 +380,47 @@ def get_summary_html(self, *args, **kwargs):
...
+                 check_host=True):
```
jwhitlock: I can see how ``required_locale`` and ``check_host`` might be useful in the future, but not in the context of this bug.  I'd lean toward removing them and the alternate code paths, and maybe linking to the same function in the kitsune repo, so they can be added back if the need arises.

kuma/wiki/tests/test_views.py
```diff
@@ -1190,6 +1190,22 @@ def test_create_on_404(self):
...
         eq_(302, response.status_code)
 
+    def test_creating_child_of_redirect(self):
+        """While try to create a child of a redirect,
+        the parent of the child should be redirect's parent"""
+        self.client.login(username='admin', password='testpass')
+        rev = revision(is_approved=True, save=True)
+        doc = rev.document
+        child_full_slug = doc.slug + "/" + "children_document"
+        doc._move_tree(new_slug="moved_doc")
```
jwhitlock: ``_move_tree`` is such a complex way to create the redirect.  Manually creating the two pages would be faster.  A redirect is just a page where the content looks like ``REDIRECT_CONTENT``.

kuma/wiki/tests/test_views.py
```diff
@@ -1190,6 +1190,22 @@ def test_create_on_404(self):
...
+        doc.refresh_from_db()
+        url = reverse('wiki.document', args=[child_full_slug])
+        response = self.client.get(url, follow=True)
+
+        # The parent id of the query should be same because while moving, a new document is created with old slug
+        # and make redirect to the old document
+        eq_(response.request['QUERY_STRING'], 'slug=children_document&parent={}'.format(doc.id))
```
jwhitlock: This test breaks if the query string is reversed. The ``parent`` query string parameter should be tested directly.

kuma/wiki/tests/test_views.py
```diff
@@ -1190,6 +1190,22 @@ def test_create_on_404(self):
                                            locale=locale))
         eq_(302, response.status_code)
 
+    def test_creating_child_of_redirect(self):
```
jwhitlock: A second test is needed for a zoned URL, which often breaks.

kuma/wiki/models.py
jwhitlock: This function requires unit tests.

kuma/wiki/utils.py
```diff
@@ -53,6 +58,33 @@ def locale_and_slug_from_path(path, request=None, path_locale=None):
     return (locale, slug, needs_redirect)
 
 
+def get_doc_components_from_url(url, required_locale=None, check_host=True):
```
jwhitlock: This function requires tests.  ``required_locale`` and ``check_host`` should be removed or fully tested.

kuma/wiki/tests/test_views.py
safwanrahman: I am planning to change the functionality of moving mage aka redirect page in future. The way its implemented seems to be very strange both kuma and kitsune (Actually, its implemented badly in kitsune, that did not change in kuma era)
So therefore I am thinking to make the redirect with _move_tree
So in future if we change the functionality, the test will not break

safwanrahman:
@jwhitlock Added the tests and updated as you have commented.
r?

safwanrahman:
@jwhitlock Added new test. Now it should not decrease the coverage

kuma/wiki/tests/test_views.py
```diff
@@ -1208,6 +1208,48 @@ def test_create_on_404(self):
...
+        root_doc._move_tree(new_slug="moved_doc")
+
+        zoned_child_full_slug = zoned_doc.url_root + "/" + "children_document"
+        response = self.client.get(zoned_child_full_slug, follow=True)
+        print response.status_code
+
+        # The parent id of the query should be same because while moving, a new document is created with old slug
+        # and make redirect to the old document
+        parameters = parse_qs(response.request['QUERY_STRING'])
+        assert parameters['parent'][0] == str(root_doc.id)
```
jwhitlock: This works, and tests that ``/en-US/zoned_url/children_doc`` finds the right parent. A second test is needed though for the document that now follows a redirect chain into a zone, something like:

```
root_doc = rev.document
original_slug = root_doc.slug
# zone it, move it around, test new page, etc., then:
# test /en-US/docs/2017-04-21-12203442670/child2
unzoned_child_full_slug = original_slug + "/child2"  
response = self.client.get(unzoned_child_full_slug, follow=True)
# ...check status code, parse parameters, and...
assert parameters['parent'][0] == str(root_doc.id)
```
I manually tested, so this should work.

kuma/wiki/tests/test_views.py
```diff
@@ -1208,6 +1208,48 @@ def test_create_on_404(self):
...
+        # Create a zone of the document
+        zoned_doc = DocumentZone(document=root_doc, url_root="zoned_url")
+        zoned_doc.save()
+
+        # Move the document to new slug
+        root_doc._move_tree(new_slug="moved_doc")
+
+        zoned_child_full_slug = zoned_doc.url_root + "/" + "children_document"
+        response = self.client.get(zoned_child_full_slug, follow=True)
+        print response.status_code
```
jwhitlock: Remove this print statement, or turn it into an assert

kuma/wiki/utils.py
jwhitlock: Now tested in ``test_utils.py``

kuma/wiki/tests/test_views.py
jwhitlock: Now tested, but one more zoned case needed for completeness.

safwanrahman:
@jwhitlock Added one more test as you suggested. r?

jwhitlock:
@safwanrahman thanks for keeping at it though several rounds of reviews!

PR #4161, opened 2017-03-30
bug 1351403: fix functional tests for maintenance mode

escattone:
This corrects a problem I introduced with the work I did that was added via https://github.com/mozilla/kuma/pull/4150. Specifically, I used the "maintenance_mode" pytest marker to mark not only tests that should only be run in maintenance mode, but also tests that should be run in both normal mode as well as maintenance mode. This made it impossible to specify the correct set of tests to run in normal mode, since we had to use ``-m "not maintenance_mode"`` to exclude the tests that ran only in maintenance mode, but that in turn excluded the set of tests that we want to run in both modes.

This fix adds a [new command-line switch](https://github.com/escattone/kuma/blob/fix-mm-functional-tests-1351403/tests/conftest.py#L13) (``--maintenance-mode``) to the ``py.test`` command and then uses [two new decorators](https://github.com/escattone/kuma/blob/fix-mm-functional-tests-1351403/tests/utils/decorators.py) that use that switch to either exclude the maintenance-mode-only tests in normal mode, or exclude the normal-mode-only tests in maintenance mode.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -366,7 +366,7 @@ You can deploy a hosted demo instance of Kuma by following these steps:
 
 #. Connecting to the demo database instance
 
-If you have access to Kubernetes, you can run the following command to connect 
+If you have access to Kubernetes, you can run the following command to connect
 to the MySQL instance::
 
     MY_GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
@@ -429,7 +429,7 @@ For local Docker-based development in maintenance mode:
 You should be good to go!
 
 There is a set of integration tests for maintenance mode. If you'd like to run
-them against your local Docker instance, you must first have done the
+them against your local Docker instance, you will first have to do the
 following:
 
 #. Loaded the latest sample database (see :ref:`provision-the-database`).
@@ -474,7 +474,7 @@ following:
 
 Now you should be ready for a successful test run::
 
-    py.test -m "maintenance_mode and not search" tests/functional --base-url http://localhost:8000 --driver Chrome --driver-path /path/to/chromedriver
+    py.test --maintenance-mode -m "not search" tests/functional --base-url http://localhost:8000 --driver Chrome --driver-path /path/to/chromedriver
 
 Note that the "search" tests are excluded. This is because the tests marked
 "search" are not currently designed to run against the sample database.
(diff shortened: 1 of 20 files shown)


docs/development.rst
```diff
@@ -429,7 +429,7 @@ For local Docker-based development in maintenance mode:
 You should be good to go!
 
 There is a set of integration tests for maintenance mode. If you'd like to run
-them against your local Docker instance, you must first have done the
+them against your local Docker instance, you will first have to do the
```
jwhitlock: 👍 change to active case. The bullet points may need to be updated now, "load" instead of "loaded", etc. "You will first have to load the sample database, ensure the document has been rendered", etc.

tests/conftest.py
```diff
@@ -9,6 +9,14 @@
     'small': {'width': 320, 'height': 480}}
 
 
+def pytest_addoption(parser):
+    parser.addoption(
+        "--maintenance-mode",
+        action="store_true",
+        help="run tests assuming target server is in maintenance mode",
```
jwhitlock: maybe "run tests against a server in maintenance mode"?

tests/utils/decorators.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+
+skip_if_maintenance_mode = pytest.mark.skipif(
+    pytest.config.getoption('--maintenance-mode'),
+    reason='because the target server is in maintenance mode'
```
jwhitlock: Can we drop "because"?  It looks like this in the output, would be a little shorter:

```
====================================================== short test summary info ======================================================
SKIP [1] tests/functional/test_notfound.py:20: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_article_revision.py:28: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_article_revision.py:7: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_search.py:117: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_article.py:46: because the target server is not in maintenance mode
SKIP [65] tests/functional/test_maintenance_mode_redirects.py:11: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_home.py:76: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_dashboard.py:42: because the target server is not in maintenance mode
SKIP [1] tests/functional/test_profiles.py:7: because the target server is not in maintenance mode
```

escattone:
Thanks for the review @jwhitlock ! I will make those documentation changes as suggested (good points all!)

PR #4167, opened 2017-04-03
Add stylelint to dockerfile

stephaniehobson:
Adding this ahead of the other changes so the environment has it available when the PR that uses it comes.

### DIFF
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -39,7 +39,8 @@ RUN npm install -g \
     jshint@2.7.0 \
     node-sass@4.3.0 \
     uglify-js@2.4.13 \
-    clean-css@3.4.23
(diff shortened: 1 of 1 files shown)


Dockerfile-base
```diff
@@ -39,7 +39,8 @@ RUN npm install -g \
     jshint@2.7.0 \
     node-sass@4.3.0 \
     uglify-js@2.4.13 \
-    clean-css@3.4.23
+    clean-css@3.4.23 \
+    stylelint@7.9.0
```
escattone: @stephaniehobson I noticed that ``stylelint`` version 7.10.1 is the latest available. Can we use that version instead?

stephaniehobson:
🎁

PR #4170, opened 2017-04-04
Copy .stylelintrc from bedrock & update, make .scss files conform

stephaniehobson:
Mostly white space, quotes, comments, and removing trailing zeros. 

Some not whitespace changes:
- font-face declarations use different format now.
- removed old `filter` declarations in favour of `-ms-filter` which is only IE8 & 9 but is valid CSS (sorry IE7 users)


stephaniehobson:
@jpetto Can I assign you this? 🎁

jpetto:
Did a good once over of code changes (mostly looking for missed dots) and all looks good. Ran `build-static` locally and all bundles were generated without error. r+ 📐

PR #4172, opened 2017-04-06
Automatically lint .scss files at compilation and in PRs

stephaniehobson:
Should now run stylelint on all pull requests and on changed files when developing locally using gulp. It's also possible to run on all files locally with `gulp css:lint` command.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,9 +9,21 @@
       "dev": true
     },
     "ajv": {
-      "version": "4.11.4",
+      "version": "4.11.5",
       "from": "ajv@>=4.9.1 <5.0.0",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.4.tgz",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.5.tgz",
+      "dev": true
+    },
+    "ajv-keywords": {
+      "version": "1.5.1",
+      "from": "ajv-keywords@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-1.5.1.tgz",
+      "dev": true
+    },
+    "amdefine": {
+      "version": "1.0.1",
+      "from": "amdefine@>=0.0.4",
+      "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.1.tgz",
       "dev": true
     },
     "ansi-regex": {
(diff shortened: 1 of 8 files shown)


stephaniehobson:
@escattone Could I have your opinion on the test failure please?

safwanrahman:
Can we test ``stylint`` in ``docker`` environemnt test of travis?
Like before all unittest, we can just run ``stylint`` test?
In this way we can reduce the resource usage of travis and simplify it

safwanrahman:
You can just add ``docker-compose exec web stylelint "kuma/static/styles/**/*.scss"`` after this [line](https://github.com/mozilla/kuma/blob/master/tox.ini#L52).
Creating a separate environment and maintaining that seems overkilling to me.
The ``docker`` environment build a new fresh docker image by installing the requirements (which is necessary for this use case as it has its own javascript requirements). So without building a new image everytime this test will not pass.

jwhitlock:
OK or you can add another case to [travis-install](https://github.com/mozilla/kuma/blob/master/scripts/travis-install#L29-L34) to install the newer docker-compose.

escattone:
Thanks @safwanrahman for enlightening me about the dynamic build done within ``scripts/docker-compose.travis.yml`` for the ``docker`` tox env, and @jwhitlock for the ``docker-compose`` installation done within ``scripts/travis-install``.

I still think we want a separate tox env for ``stylelint``, like we do for ``flake8``, since we want to see the ``stylelint`` results no matter the results of the other runs. If we put it within the ``docker`` tox env, it would abort the run (if we put it before the ``make test`` and it failed) or it might never be run (if placed after ``make test`` and ``make test`` failed).

My last force push made another attempt to fix this, realizing we didn't need ``docker-compose`` just ``docker``, but for some as yet unknown reason it couldn't resolve ``make stylelint-direct``. I've got to call it quits for tonight, but I'll address this again on Monday.

npm-shrinkwrap.json
```diff
@@ -2661,6 +3212,12 @@
       "resolved": "https://registry.npmjs.org/slash/-/slash-1.0.0.tgz",
       "dev": true
     },
+    "slice-ansi": {
+      "version": "0.0.4",
+      "from": "slice-ansi@0.0.4",
+      "resolved": "http://127.0.0.1:50473/slice-ansi/-/slice-ansi-0.0.4.tgz",
```
safwanrahman: Why it is pointing to ``http://127.0.0.1:50473``?
Raising issue while I run ``npm install``

escattone: Good catch @safwanrahman! I'll commit a change to fix this.

escattone:
Regarding the ``stylelint`` tox env for Travis, I decided against using ``Dockerfile-base``. It's way too much (and therefore too slow), so I added a new minimal [Docker file](https://github.com/mozilla/kuma/pull/4172/files#diff-4e3fe2759370a4c54288537caf6bdbd0) just for running ``stylelint`` in Travis.

safwanrahman:
Why do we need a docker image to run the stylelint test? cant we do like the flake8 test? just install the dependency and run the test?

escattone:
Good point @safwanrahman (given that we only care about ``stylelint`` and not the specific version of ``node``, etc.). I've made an update to reflect this.

escattone:
I updated the "Development" doc. If the checks pass, I think this is ready to go.

PR #4173, opened 2017-04-06
Bug 1328989: Remove ACE editor

stephaniehobson:
Remove ACE editor, the editor we used for template files.


kuma/static/styles/components/wiki/edit/ace-content.scss
```diff
@@ -1,10 +0,0 @@
-#ace_content {
-    height: 500px;
-    position: relative;
-    width: 100%;
-
-    .ace_warning,
-    .ace_error {
-        background-image: none;
-    }
-}
```
jwhitlock: This is still @imported in ``wiki-edit.scss``

stephaniehobson: That explains that error...

stephaniehobson:
Strings will be untranslated for the small amount of time they remain.

Fixed error.

PR #4178, opened 2017-04-10
bug 1246164: Increase timeout for dashboard AJAX

jwhitlock:
The AJAX load for page=2 often takes more than 10 seconds ([4 out of 11 recent tests](https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/job/stage-integration-tests/)). Rather than increase the timeout across the tests, selectively use a timeout of ~~20~~ 30 seconds for the AJAX load by diving down to the [Python Selenium](http://selenium-python.readthedocs.io/waits.html#explicit-waits) library.

### DIFF
--- a/tests/pages/dashboard.py
+++ b/tests/pages/dashboard.py
@@ -1,5 +1,6 @@
 from pypom import Region
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait
 
 from pages.base import BasePage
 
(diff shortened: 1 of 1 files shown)


safwanrahman:
20 seconds for loading dashboard seems too hight to me.
@jwhitlock can you please open a bug about it so that we can investigate and reduce the loading time in future?

jwhitlock:
Thanks for your concern @safwanrahman, but I don't think I need to open a bug to investigate this. I believe it is a result of running these tests against the staging server, and not a code issue.

The staging server's purpose is to test code changes before they go to the production servers.  They get used a few times a week by one or two people, as opposed to the production servers, which have 1000+ simultaneous users during peak hours.  Because the need to serve traffic is much less, the hardware is less capable, and is shared between projects, which rarely are using staging servers at the same time.

The stage ElasticSearch service covers 3 nodes, behind a load balancer.  My guess is that the first request goes to one node, and the second request to another that isn't awake yet.  "Awake" may be Java moving data from memory to disk, or even a VM on a multi-VM system that has been put to sleep.  This would make the second request slower, but follow-on requests would be to a ready system, and would be responsive.  This is what happens when I manually try the test after a failure - there is almost no delay between making the request and getting results.

If my theory is correct, then manually requesting some dashboard pages before running the tests would make the tests pass every time. This is worth trying, and I'll try to remember to do this on the next production push. It would be good to see the delay in my own browser.

Further investigation would require WebOps, who are time constrained. I would rather they work on my other requests, related to getting MDN into AWS.  If the problem is a resource issue, I can't justify allocating more resources to a staging instance. If the problem is an ElasticSearch configuration, we can detect and fix in AWS using those resources.

jwhitlock:
@safwanrahman contacted me on IRC for more details.  He is right, the revision dashboard has nothing to do with ElasticSearch (I've been concentrating on ES work lately), and the database is the primary culprit for slowness on this page.  

Safwan recorded a run of 25 seconds for the page=2 query, so this isn't just a staging issue:

screen shot 2017-04-12 at 12 25 00 am

Given that timing, I'm increasing the test timeout to 30 seconds.

We are tracking pagination clicks as a category="Dashboard Pagination" event, and this gets clicked about 200 times a month.  Is it unused because it is slow? Or is it used as much as it would be if it was fast? It is hard to say.  Interesting, but out of scope for fixing this intermittently failing test.

My research at the time showed a few solutions that involved using database-specific features to speed up the full table count.  If we're going that route, I'd rather switch to PostgreSQL first, and then use its DB-specific features.  See [bug 1246164](https://bugzilla.mozilla.org/show_bug.cgi?id=1246164) for details.

jwhitlock:
Force-push to add [bug  1246164](https://bugzilla.mozilla.org/show_bug.cgi?id= 1246164) as a comment, and change the bug number in the commit message.

PR #4179, opened 2017-04-11
bug 1328989: Add /en-US/docs/macros for searching

jwhitlock:
Display a table of macro names and page counts in English and all documents, and include a form for manual macro searches for macros not known to KumaScript.


jwhitlock:
I could use other's opinions on some follow-on items:

* Does this belong at /docs/macros, or /dashboards/macros?  I have it at /docs/macros because that's where docs/templates are, but "special" doc pages seem out of place to me.
* Should I drop /docs/templates in this PR as well? If so, should I use a standard wiki redirect, or a URL rule to redirect?
* Should I cache the macro and usage lists for the view, or wait until we see the performance in prod?

stephaniehobson:
> Does this belong at /docs/macros, or /dashboards/macros? I have it at /docs/macros because that's where docs/templates are, but "special" doc pages seem out of place to me.

dashboards/macros

> Should I drop /docs/templates in this PR as well? If so, should I use a standard wiki redirect, or a URL rule to redirect?

yes, ummmm, standard wiki redirect?

> Should I cache the macro and usage lists for the view, or wait until we see the performance in prod?

dunno

escattone:
> Does this belong at /docs/macros, or /dashboards/macros? I have it at /docs/macros because that's where docs/templates are, but "special" doc pages seem out of place to me.

dashboards/macros

> Should I drop /docs/templates in this PR as well? If so, should I use a standard wiki redirect, or a URL rule to redirect?

Yes. I'm not clear about the pros/cons of each choice, but might we do a 301 via the Django view?

> Should I cache the macro and usage lists for the view, or wait until we see the performance in prod?

I'm loading the anonymized DB to check with more data, but my current feeling is that we can wait to see the performance in production.

stephaniehobson:
I was thinking standard wiki redirect because it would be less code in kuma. But now I'm realizing that we won't be able to redirect the entire sub tree, just the landing page. So mostly I stand by my 'ummm'

kuma/wiki/jinja2/wiki/list/macros.html
```diff
@@ -0,0 +1,71 @@
...
+        {% endfor %}
+      </tbody>
+    </table>
+  {% endif %}
+  <form id="macro-search" method="GET" action="{{url("search")}}">
+    <input type="hidden" name="topic" value="none" />
+    <label for="name">{{ _("Search for macro usage:") }}</label>
+    <input id="name" name="kumascript_macros" type="text" placeholder="{{_("Macro Name") }}" />
+    <br/>
+    <button type="submit" name="locale" value="*">{{ ("Search All Pages") }}</button>
```
safwanrahman: This text should be unlocalized for any reason?

jwhitlock: good catch, should be localized

jwhitlock:
I'm asking for @stephaniehobson's review as well because she's the expert on displaying tabular data.  Follow-on commits coming to:

* [ ] Localize buttons
* [ ] Move the page to /en-US/dashboards/macros
* [ ] Drop /en-US/docs/templates

kuma/wiki/tests/test_kumascript.py
```diff
@@ -60,3 +81,166 @@ def test_get_does_not_call_format_slug_for_docs(self, mock_format_slug):
...
+            'filename': 'APIFeatureList.ejs', 'name': 'APIFeatureList'
+        }]
+    }
+    mock_requests.get(macros_url, json=response)
+    macros = kumascript.macro_sources()
+    expected = {
+        'A11yRoleQuicklinks': 'A11yRoleQuicklinks.ejs',
+        'APIFeatureList': 'APIFeatureList.ejs',
+    }
+    assert macros == expected
```
safwanrahman: I would prefer using ``eq_``
It returns clear output if fail

kuma/wiki/tests/test_templates.py
```diff
@@ -1331,3 +1331,46 @@ def _translation_data():
...
+        'A11yRoleQuicklinks': {
+            'github_subpath': 'A11yRoleQuicklinks.ejs',
+            'count': 100,
+            'en_count': 50,
+        }
+    }
+
+    response = client.get(reverse('wiki.list_macros'), follow=True)
+    assert response.status_code == 200
+    assert "Found 1 active macro." in response.content
```
safwanrahman: can use ``ok_``

kuma/wiki/tests/test_kumascript.py
jwhitlock: Please post the output of ``eq_`` vs ``assert`` in a failing test.

safwanrahman: Reading source, seems like its deprecated
https://github.com/mozilla/kuma/blob/master/kuma/core/tests/__init__.py#L17

escattone: @safwanrahman We shouldn't use the ``eq_``, ``ok_``, etc. helpers any longer, as pytest takes care of providing detailed information on failure using the standard ``assert``.

escattone:
Thanks @jwhitlock, this is an awesome PR! I'm really happy to see the docs/templates updated to fit the new reality for macros.

So far, I've found what appear to be two issues:
- If I click on any number link in the "All Page Count" column (or do a manual "Search All Pages"), the search results header always says "...documents found in English (US)" but the results often contain links to non-English documents as well. It seems it should say something like "...documents found in all locales"?
- If I do a manual search (which is inactivated once one of the search buttons is clicked), view the results, then hit the browser's back button, the manual search remains inactivated.

kuma/wiki/jinja2/wiki/list/macros.html
```diff
@@ -0,0 +1,71 @@
...
+    <button type="submit" name="locale" value="en-US">{{ ("Search English Pages") }}</button>
```
stephaniehobson: Needs localization underscore here too.

kuma/wiki/jinja2/wiki/list/macros.html
```diff
@@ -0,0 +1,71 @@
...
+                  {{ params["count"] | default(0) }}
+                </a>
+              </td>
+           {% endif %}
+          </tr>
+        {% endfor %}
+      </tbody>
+    </table>
+  {% endif %}
+  <form id="macro-search" method="GET" action="{{url("search")}}">
```
stephaniehobson: Please add `class="nodisable"`. This will stop the form from being disabled when using the back button to return to the page after a search.

tests/functional/test_lists.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+from pages.article import ArticlePage
+from selenium.webdriver.common.by import By
+
+
+@pytest.mark.nondestructive
+def test_list_macros(base_url, selenium):
```
escattone: I'm wondering if we want to more fully check the functionality of this page, like adding assertions and/or click-tests for the links (macro-name, macro-counts) and the manual search buttons?

escattone:
Thanks for the changes @jwhitlock! There's still this issue I'd like to highlight again and get everyone's feedback.

- For the case when we're searching all locales, for example if I click on a number link in the "All Page Count" column (or do a manual "Search All Pages"), the search results page always says "...documents found in English (US)" but it seems to me it should say something like "...documents found in all locales" or simply "...documents found"? (https://github.com/mozilla/kuma/blob/master/kuma/search/jinja2/search/results.html )

jwhitlock:
@escattone yes, I think the template needs to reflect when all pages are searched.  I'd like to do that in a different PR.

kuma/wiki/kumascript.py
```diff
@@ -298,7 +299,11 @@ def macro_sources():
     response = requests.get(ks_macro_url)
     if response.status_code == 200:
         macros_raw = response.json()['macros']
-        return {md['name']: md['filename'] for md in macros_raw}
+        # Ensure Normal Form C used on GitHub
+        return {
+            unicodedata.normalize('NFC', md['name']):
```
escattone: Wow, I didn't even realize there were normalization forms for unicode. Good catch and nice work @jwhitlock !

jwhitlock:
I'm done with this round of changes:

* Moved from ``docs/macros`` to ``dashboards/macros``
* Localized manual search buttons, added ``class='no disable'``, and added element IDs for testing
* Made ``docs/templates`` a permanent redirect to ``dashboards/macros``
* Fixed a unicode normalization issue when there are two ways of encoding a character.  For example, the `ó` in "traducción.ejs" can be encoded as two characters, "o", and then a "combined with a diacritical", with the invalid link https://github.com/mozilla/kumascript/blob/master/macros/traduccio%CC%81n.ejs. It can also be encoded as "O WITH ACUTE, LATIN SMALL LETTER", with the valid link https://github.com/mozilla/kumascript/blob/master/macros/traducci%C3%B3n.ejs.
* Remove a duplicate call to ``macro_sources``
* Added unit test for an unexpected ElasticSearch error on the second call
* Added functional tests that follow GitHub and search links, and exercise manual search
* Add a permanent redirect from ``docs/templates`` to ``dashboards/macros``.

I did not add a generic redirect for template source, like [docs/Template:a11yEventRef](https://developer.mozilla.org/en-US/docs/Template:a11yEventRef). Due to MDN's flexible capitalization rules, I don't think there is a reliable way to link to the new source.  I think these should be 404s when the template source is removed from MDN, and any MDN links to the files should be updated to GitHub (or maybe the MDN docs moved to the GitHub repo).

I also didn't fix the misleading text "documents found in English (US)" when searching all locales, since I feel this is an existing issue that should be addressed with a new PR.

kuma/wiki/kumascript.py
jwhitlock: I didn't either, until I hit issues exporting [historical macros to files](https://github.com/mdn/archived_kumascript/blob/master/export_templates.py#L162-L176).

escattone:
Wow, thanks again @jwhitlock, this is awesome work with really thorough well-written tests!

jwhitlock:
Found [bug 1063560](https://bugzilla.mozilla.org/show_bug.cgi?id=1063560) - "documents found in English." is a 3 year-old bug 😠

PR #4180, opened 2017-04-11
bug 1341770: Set Yes/No for "is_staff" GA dim

jwhitlock:
The "yesno" template filter also localizes the text, so visits on French pages get the values "Oui" / "Non". Use regular template logic so the values are only "Yes" / "No".  Found during manual testing of #4175.

### DIFF
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -20,7 +20,7 @@
         {% endif %}
 
         // dimension18 == "Staff"
-        ga('set', 'dimension18', '{{ request.user.is_staff | yesno }}');
+        ga('set', 'dimension18', '{% if request.user.is_staff %}Yes{% else %}No{% endif %}');
     {% endif %}
 
     // dimension9 == "Section editing"
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -19,8 +19,10 @@
             ga('set', 'dimension2', 'Yes');
         {% endif %}
 
-        // dimension18 == "Staff"
-        ga('set', 'dimension18', '{% if request.user.is_staff %}Yes{% else %}No{% endif %}');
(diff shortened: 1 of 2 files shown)


stephaniehobson:
LOL, unexpected side effect.

I've thought of some feedback I probably should have had on the first pull request - better late than never?

Rather than saying yes or no, could you put the entire dimension behind the if statement and only set it (to yes) if the user is staff? I think that will make the dimension stick if the staff member is logged out. Otherwise it'll be overridden.

jwhitlock:
Making sure I understand.

Here's how it works today:

* User is anonymous / logged out - dimension18 is not set
* User is logged in, non-staff - dimension18 is set to ``No``
* User is logged in, staff - dimension18 is set to ``Yes``

and you want it changed to:

* User is anonymous / logged out - dimension18 is not set
* User is logged in, non-staff - **dimension18 is not set**
* User is logged in, staff - dimension18 is set to ``Yes``

Is this true?

jwhitlock:
Confirmed w/ @stephaniehobson on IRC, changed behavior to only set dim18 if staff

jinja2/includes/google_analytics.html
```diff
@@ -19,8 +19,10 @@
             ga('set', 'dimension2', 'Yes');
         {% endif %}
 
-        // dimension18 == "Staff"
-        ga('set', 'dimension18', '{{ request.user.is_staff | yesno }}');
+        {%- if request.user.is_staff %}
+            // dimension18 == "Staff"
+            ga('set', 'dimension18', 'Yes');
```
safwanrahman: So If ``resuest.user`` is not ``superuser`` it should not send data to GA?

stephaniehobson: Correct.

stephaniehobson:
👍

PR #4181, opened 2017-04-15
bug 1269104: Refactor notification emails, provide diff for first translations

jwhitlock:
There are [several requested changes](https://bugzilla.mozilla.org/showdependencytree.cgi?id=1142533&hide_resolved=1) for notification emails, which should be trivial to fix, but are tricky because the tests are incomplete and the code is spread over several files.

This PR:
* Moves the logic for "first edit", "page watch", and "spam attempt" emails into kuma.wiki.events
* Rewrites the tests to py.test standards, and tests against the three main edit types: English page creation, page edits, and first translations.
* Rewrites the test for ``Revision.previous``, setting creation dates rather than sleeping.
* Fixes [bug 1142533](https://bugzilla.mozilla.org/show_bug.cgi?id=1142533), so that notification emails for first translations include a diff to the English revision they are based on.

I've fixed a few other bugs (see [redo_emails_1269104](https://github.com/jwhitlock/kuma/tree/redo_emails_1269104)), but I want to get the refactor approved and merged so that the future PRs will be limited to the feature changes.

### DIFF
--- a/kuma/wiki/tests/test_models.py
+++ b/kuma/wiki/tests/test_models.py
@@ -609,10 +609,16 @@ class RevisionTests(UserTestCase):
         last_rev = revision(document=rev.document, content="Finally",
                             is_approved=True,
                             created=datetime(2017, 4, 15, 9, 25), save=True)
+        trans = Document.objects.create(parent=rev.document, locale='fr',
+                                        title='In French')
+        trans_rev = revision(document=trans, is_approved=True,
+                             based_on=last_rev,
+                             created=datetime(2017, 4, 15, 9, 56), save=True)
 
         assert rev.previous is None
         assert next_rev.previous == rev
         assert last_rev.previous == next_rev
+        assert trans_rev.previous is None
 
     @pytest.mark.toc
     def test_show_toc(self):
--- a/kuma/wiki/tests/test_models.py
+++ b/kuma/wiki/tests/test_models.py
@@ -601,19 +601,18 @@ class RevisionTests(UserTestCase):
     def test_previous(self):
         """Revision.previous should return this revision's document's
         most recent approved revision."""
-        rev = revision(is_approved=True, save=True)
-        eq_(None, rev.previous)
-        # wait a second so next revision is a different datetime
-        time.sleep(1)
+        rev = revision(is_approved=True, created=datetime(2017, 4, 15, 9, 23),
+                       save=True)
         next_rev = revision(document=rev.document, content="Updated",
-                            is_approved=True)
-        next_rev.save()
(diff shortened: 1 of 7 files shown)


safwanrahman:
I still could not find a reason why the code of ``signals`` live in ``apps.py`` 😕

kuma/wiki/tests/test_models.py
```diff
@@ -601,19 +601,24 @@ def test_correct_based_on_to_current_revision(self):
     def test_previous(self):
         """Revision.previous should return this revision's document's
         most recent approved revision."""
-        rev = revision(is_approved=True, save=True)
-        eq_(None, rev.previous)
-        # wait a second so next revision is a different datetime
-        time.sleep(1)
+        rev = revision(is_approved=True, created=datetime(2017, 4, 15, 9, 23),
```
safwanrahman: If you rewrite the [model method](https://github.com/mozilla/kuma/blob/master/kuma/wiki/models.py#L1804) queryset like this
```python
try:
    return self.document.revisions.filter(is_approved=True, id__lt=self.id).latest()
except self. DoesNotExist:
    return None```

Its possible to remove the ``time.sleep()`` easily

File: kuma/wiki/tests/test_models.py [hunk 2]
```diff
@@ -601,19 +601,24 @@ def test_correct_based_on_to_current_revision(self):
...
+                       save=True)
         next_rev = revision(document=rev.document, content="Updated",
```
safwanrahman: Can you create more than one ``revision`` so there are another revision exist. So if our code break it can be cought?

safwanrahman:
@jwhitlock I have been testing mail functionality with the data got from ``outbox`` like mentioned here
https://docs.djangoproject.com/en/1.11/topics/testing/tools/#email-services
Is it possible to implement that also in mdn?
currently, you are testing the returned value of a ``Email`` instance. Should it be better if we use ``outbox`` mail data?

kuma/wiki/tests/test_models.py
jwhitlock: I'm trying to change as little as possible, and I consider a big re-write of``get_previous`` out of scope.  It is true that IDs increase for recent revisions, but that would need to be validated for older revisions in the database.

kuma/wiki/tests/test_models.py
jwhitlock: I don't understand your comment.  There are four revisions created for this test case.

safwanrahman: @jwhitlock Sorry. I meant more than one revision for same document

PR #4182, opened 2017-04-15
[Bug 1353138] Banned user profiles should 404

safwanrahman:
This patch will make banned user profile return 404

### DIFF
--- a/jinja2/403.html
+++ b/jinja2/403.html
@@ -23,8 +23,6 @@
       <p class="notice">{{ _('The wiki is in read-only mode.') }}</p>
     {% elif reason == 'kumabanned' %}
       <p class="notice">{{ _('Your profile has been banned from making edits.') }}</p>
-    {% elif reason == 'bannedprofile' %}
-      <p class="notice">{{ _("This user's profile has been banned.") }}</p>
     {% elif reason == 'revisions_login_required' %}
       <p class="notice">{{ _("You must sign in to view all history") }}</p>
     {% endif %}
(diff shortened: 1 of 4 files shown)


safwanrahman:
@jwhitlock I can work on the feature. tired off the slow interned today.
Will check it back tomorrow!

jwhitlock:
Thanks @safwanrahman, this week works for me. If you reuse the same phrase as was used in 403.html, the translators will appreciate it.

jinja2/403.html
```diff
@@ -23,8 +23,6 @@ <h1 class="page-title">{% block page_title %}{{ _('Permission Denied') }}{% endb
       <p class="notice">{{ _('The wiki is in read-only mode.') }}</p>
     {% elif reason == 'kumabanned' %}
       <p class="notice">{{ _('Your profile has been banned from making edits.') }}</p>
-    {% elif reason == 'bannedprofile' %}
-      <p class="notice">{{ _("This user's profile has been banned.") }}</p>
```
jwhitlock: This notice should be moved to the 404 handler, if possible with the same string so that translators don't have to re-translate.

kuma/users/views.py
```diff
@@ -361,8 +361,7 @@ def user_detail(request, username):
     detail_user = get_object_or_404(User, username=username)
 
     if (detail_user.active_ban and not request.user.has_perm('users.add_userban')):
-        return render(request, '403.html',
-                      {'reason': 'bannedprofile'}, status=403)
+        return render(request, '404.html', status=404)
```
jwhitlock: Info needs to be passed to the 404 page so that it can distinguish between a banned user, a user that never existed, and the current default (navigated to missing page).  The 403 page does this by passing a ``reason``.

safwanrahman:
@jwhitlock Updated and tested. r?

PR #4183, opened 2017-04-18
Bug 1356923: Content experiment "static-examples-on-top"

jwhitlock:
Move the configuration for content experiments to a JSON file, then add the 84 pages in the ``static-examples-on-top`` experiment.

This JSON was generated from the tables on [Experiment:StaticExamplesOnTop/CSS](https://developer.mozilla.org/en-US/docs/Experiment:StaticExamplesOnTop/CSS) and [Experiment:StaticExamplesOnTop/JavaScript](https://developer.mozilla.org/en-US/docs/Experiment:StaticExamplesOnTop/JavaScript), using a [script](https://gist.github.com/jwhitlock/ea959c4011a5d54e86a2236fb636f578).

I was hoping to include the functional tests with this PR, but I've been unable to generate a new sample database, so it wouldn't be possible to get the tests to pass locally without manually copying 168 pages from production to staging and the local database.  It will take some additional work to get the sample DB code running again.


escattone:
Now that I see the number of pages involved in the ``static-examples-on-top`` experiment, I'm concerned about the back-end performance hit we'll take with every document request (in [_apply_content_experiment](https://github.com/mozilla/kuma/blob/master/kuma/wiki/views/document.py#L261) we loop over the pages in the experiment until we find a slug/locale match or exhaust the search). Might it be worth making two adjustments to the back-end machinery that would change the O(n) loop to a O(1) dictionary lookup? Here's what I had in mind:

First change the format of the ``pages`` key within an experiment in the [``contents_experiments.json``](https://github.com/jwhitlock/kuma/blob/68b1bad3fe6dc11b5d754dcba1de36ca2e0fc66a/kuma/settings/content_experiments.json) file from a list of dictionaries to a dictionary of lists, where each key (within the ``pages`` dictionary) would be a concatenation of ``locale`` and ``slug`` and the value for each key would be the list of variants, so instead of this:

```
        "pages": [
            {
                "locale": "en-US",
                "slug": "Web/JavaScript/Reference/Global_Objects/Array",
                "variants": [
                    [
                        "control",
                        "https://developer.mozilla.org/en-US/docs/Experiment:StaticExamplesOnTop/JavaScript/Array"
                    ],
                    [
                        "example",
                        "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array"
                    ]
                ]
        }, ...
```

we do something like this:

```
        "pages": {
            "en-US:Web/JavaScript/Reference/Global_Objects/Array": [
                    [
                        "control",
                        "https://developer.mozilla.org/en-US/docs/Experiment:StaticExamplesOnTop/JavaScript/Array"
                    ],
                    [
                        "example",
                        "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array"
                    ]
             ]
        }, ...
```

and then we can change the code in [_apply_content_experiment](https://github.com/mozilla/kuma/blob/master/kuma/wiki/views/document.py#L261) from this:

```
        for pages in experiment['pages']:
            if (pages['locale'] == doc.locale and pages['slug'] == doc.slug):
                ...
```

to something like this:

```
       key = '{}:{}'.format(doc.locale, doc.slug)
       variants = experiment['pages'].get(key)
       if variants:
           ...
```

The main point is that it would be nice to make the A/B testing back-end machinery insensitive to the number of pages in any experiment. But maybe this belongs in a separate PR, if we decide something like this is worthwhile, and we just go ahead with the current back-end machinery?

kuma/static/js/experiment-static-examples-on-top.js
```diff
@@ -0,0 +1,14 @@
+(function() {
```
jpetto: We miss this all the time on bedrock, but if we're passing in `window.Mozilla` to the IIFE, it should be aliased here: `(function(Mozilla) {`.

jwhitlock: I have no idea what you are saying, but I will do it.  Does [experiment-framework-test.js](https://github.com/mozilla/kuma/blob/master/kuma/static/js/experiment-framework-test.js) and [experiment-wiki-content.js](https://github.com/mozilla/kuma/blob/master/kuma/static/js/experiment-wiki-content.js) need it too?

jpetto: Yep, those could both use it. (Told you we miss it all the time. Sigh.)

Should `experiment-wiki-content.js` still live on? It was initially a created as a stub for the experiment you're starting in this PR. It's not running or anything, so maybe it can be dropped?

jwhitlock:
I'm not sure it's an actual perf problem, but changing the keys should work. I opened a Django shell and ran ``timeit`` for a slug that is not in the experiment (so it runs over all the keys):

```
import timeit
timeit.repeat(
    "_apply_content_experiment(None, doc)",
    ("from kuma.wiki.models import Document; "
     "from kuma.wiki.views.document import _apply_content_experiment; "
     "doc = Document.objects.get(locale='en-US', slug='Web/CSS/@counter-style/additive-symbols')"),
    number=1000000)
```
That returns ``[29.471986055374146, 28.72660183906555, 28.26314091682434]``, or about 29 seconds to run it 1,000,000 times.  We'll see what the numbers are after the change.

escattone:
Thanks for the actual numbers @jwhitlock . The actual performance hit may just be in the noise, so feel free to ignore this idea.

jwhitlock:
The numbers with the change are ``[1.1767399311065674, 1.1708219051361084, 1.1785950660705566]``. It's hard to not go with a 25x improvement on a function called for every document.  I also adjusted the variant format, and noticed that my scraper used full URLs, not slugs, for the variants sources. Oops.

kuma/wiki/views/document.py
```diff
@@ -257,38 +257,39 @@ def _apply_content_experiment(request, doc):
...
-                return doc, exp_params  # No (valid) variant selected
+        if key in experiment['pages']:
+            # This page is under a content experiment
+            variants = experiment['pages'][key]
+            exp_params = {
+                'id': experiment['id'],
+                'ga_name': experiment['ga_name'],
+                'param': experiment['param'],
+                'original_path': request.path,
+                'variants': experiment['pages'][key],
```
escattone: ``experiment['pages'][key]`` can be replaced with ``variants``

escattone:
Thanks @jwhitlock , for the changes and for the actual performance numbers.

The three tests in [``kuma/wiki/tests/test_templates.py``](https://github.com/jwhitlock/kuma/blob/content_exp_static_1356923/kuma/wiki/tests/test_templates.py) currently failing will succeed once this [``pages``](https://github.com/jwhitlock/kuma/blob/content_exp_static_1356923/kuma/wiki/tests/test_templates.py#L278) value is updated to the new dictionary-style format.

PR #4184, opened 2017-04-18
fix bug 1265495: Skip spam check for some edits

jwhitlock:
If an edit does not change the content (such as settings or clearing the review requested flags), then do not submit the edit to Akismet to check for spam.

### DIFF
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -776,6 +776,17 @@ class RevisionForm(AkismetCheckFormMixin, forms.ModelForm):
         parameters.update(self.akismet_parameter_overrides())
         return parameters
 
+    def akismet_call(self, parameters):
+        """
+        Skip Akismet check if the content is the same.
+
+        This happens if the edit is to a non-content field, such as
+        setting or clearing the technical review flag.
+        """
+        if not parameters['comment_content']:
+            return False  # No content change, not spam
+        return super(RevisionForm, self).akismet_call(parameters)
+
     def save(self, document, **kwargs):
         """
         Persists the revision and returns it.
--- a/kuma/wiki/tests/test_forms.py
+++ b/kuma/wiki/tests/test_forms.py
@@ -284,7 +284,7 @@ class RevisionFormEditTests(RevisionFormViewTests):
             '<h2 id="Syntax">Syntax</h2>\r\n'
             '<pre class="brush:css">\r\n'
             'display: none;\r\n'
-            '</pre>\r\n'
+            '</pre>'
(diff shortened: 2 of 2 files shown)


jwhitlock:
@stephaniehobson I marked you for review because you submitted [bug 1265495](https://bugzilla.mozilla.org/show_bug.cgi?id=1265495), but this is a backend-only change, and a nasty one at that, so feel free to reassign to @escattone. Or just merge it 😄

kuma/wiki/forms.py
```diff
@@ -776,6 +776,17 @@ def akismet_parameters(self):
...
         return parameters
 
+    def akismet_call(self, parameters):
+        """
+        Skip Akismet check if the content is the same.
+
+        This happens if the edit is to a non-content field, such as
+        setting or clearing the technical review flag.
+        """
+        if not parameters.get('comment_content', 'omitted'):
```
safwanrahman: So if ``"comment_content"`` is in ``"parameters"`` it should return "omitted"?

jwhitlock: yes. It should always always be there, so maybe a ``.get`` is the wrong thing. But, if it was omitted, I didn't want the short-circuit to run.

jwhitlock:
I'm surprised that setting the checkbox in the edit view trigger the spam filter. I'd like to investigate that before merging, and get unsurprised.

kuma/wiki/forms.py
safwanrahman: I think it should be written like
``if not parameters.get('comment_content') == '':``
So its more readable. I get confused by reading the code! 😕

jwhitlock:
🔥 ☹️ 👿 🙍‍♂️ tabs 🙍‍♂️ 👿 ☹️ 🔥

jwhitlock:
The diff of the previous revision to the new edit does not ignore whitespace, and there doesn't appear to be a option for ignoring whitespace in the Python [difflib](https://docs.python.org/2/library/difflib.html) module.  There are options for specifying junk characters, but this doesn't do what you might think it does, and can't be used to ignore whitespace.

The solution would be to normalize whitespace, and we have a 3rd-party library for doing that, ``pytidylib``.  However, there are several problems with that which need to be addressed first, tracked in [bug 1358541](https://bugzilla.mozilla.org/show_bug.cgi?id=1358541).

So, I'm going with @stephaniehobson's original suggestion.  The function will work on pages where no monster used tabs in content.

One more force push so @safwanrahman doesn't get confuse by ``dict.get``.

PR #4189, opened 2017-04-21
Bug 1298615 - Fix output of Insert Live Sample Template button

a2sheppy:
This fixes the issues in bug 1298615 as follows:

1. The labels of the subsections are now simply "CSS", "HTML", and
"JavaScript". This affects translations although the changes should
be pretty minimal. This is per our style guide, which has always said
these sections should be headed like this.

2. Removed the "name" attribute from the headings.

3. Removed the unneeded spaces from inside the generated EmbedLiveSample
macro.


jwhitlock:
Replaced by #4190

PR #4192, opened 2017-04-22
bug 1357891: Update to ElasticSearch 2.4

jwhitlock:
We're preparing for an update to ElasticSearch 2.4 in SCL3, so I'm starting with the Docker environment.  The elasticsearch 1.x libraries are compatible with 2.x, sending deprecated but supported 1.x API commands.  Once all environments are updated, we can make the code changes for 2.x, and update to the 2.x libraries, which are not compatible with 1.x.

The changes are:
* Fix a typo in the wiki_document meta that 1.7 ignored and 2.4 doesn't.
* Update the docs to reflect the new version, to remove Vagrant references, and to refresh from a recent run.

Locally, a ``make up`` downloaded the ElasticSearch 2.4 image, and then restarted the elasticsearch and kuma containers.  ES 2.4 was able to update the 1.x search index in place.  I was able to create and promote a new index of the sample database, and confirmed that searches like "CSS" and macro usage worked.

@jgmize (or @metadave, @escattone ) - are there any concerns using ES 2.4 in AWS?

@stephaniehobson - you expressed an interest in getting your feet wet w/ ElasticSearch 😛

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,7 +76,7 @@ services:
       - mysqlvolume:/var/lib/mysql
 
   elasticsearch:
-    image: elasticsearch:1.7
+    image: elasticsearch:2.4
 
   redis:
     image: redis
--- a/docs/elasticsearch.rst
+++ b/docs/elasticsearch.rst
@@ -8,10 +8,10 @@ Kuma uses Elasticsearch_ to power its on-site search.
 Installed version
 =================
 Elasticsearch `releases new versions often`_, and Kuma is slow to upgrade. The
-Docker and Vagrant environments use 1.7, but production is older.
+Docker environment uses 2.4, and production uses 1.7, with a planned update to 2.4.
 
 The Kuma search tests use the configured Elasticsearch, and can be run inside
-the ``web`` Docker container or the Vagrant VM to confirm the engine works::
+the ``web`` Docker container to confirm the engine works::
 
     py.test kuma/search/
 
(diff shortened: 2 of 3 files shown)


metadave:
@jwhitlock I only see ES v 1.5, 2.3 and 5.1 available in AWS.

jgmize:
See also https://github.com/mozmar/infra/issues/193

jwhitlock:
The breaking change in [2.4.0](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/release-notes-2.4.0.html) is that [duplicated response headers](https://github.com/elastic/elasticsearch/pull/19872) are combined.

2.3.x:
```
Access-Control-Allow-Headers: Content-Type
Access-Control-Allow-Headers: X-Requested-With
```

2.4.0:
```
Access-Control-Allow-Headers: Content-Type,X-Requested-With
```

According the bug, IE has a "last-one-wins" algorithm for duplicated response headers.  I sincerely doubt Kuma cares.

I am OK running ES 2.4 in SCL3 and ES 2.3 or 2.4 in Docker / AWS, so I'll leave it up to @jgmize and @metadave to pick the docker / AWS version.  I have a [2.3 branch](https://github.com/jwhitlock/kuma/tree/elasticsearch_2_3) ready to go.

jwhitlock:
Also, if you review, your ES database will be updated to 2.x, and you'll have to rebuild it to go back to 1.7.  Something like this:

```
docker-compose stop
docker rm kuma_elasticsearch_1 -v
make up  # Wait a bit for ES to rebuild
make up -d  # Do it again so web, worker, etc. see a working ES server
```

metadave:
lgtm 👍

PR #4196, opened 2017-04-25
Update maintenance mode

escattone:
This PR updates maintenance mode such that it no longer requires Celery (Redis and at least one Celery worker) (per discussion in https://github.com/mozmar/infra/issues/134).

- remove need for Celery in maintenance mode (see https://github.com/mozmar/infra/issues/134#issuecomment-292581709)
    - override ``DocumentContributorsJob.get`` to return an empty response when in maintenance mode
    - add Django management command to make ``humans.txt`` (for calling as a post-deploy step for maintenance mode)
    - add Django management command to make the sitemaps (for calling as a post-deploy step for maintenance mode)

- add and refactor tests for wiki jobs
    - add high-level ``conftest.py`` that provides some fixtures helpful for refactoring
      ``KumaTestCase``-based tests
    - refactor tests for wiki jobs and add test for ``DocumentContributorsJob`` when in maintenance mode

### DIFF
--- a/kuma/conftest.py
+++ b/kuma/conftest.py
@@ -1,39 +1,8 @@
 import pytest
 from django.conf import settings
 from django.core.cache import caches
-from django.utils.translation import trans_real
-
-from kuma.core.tests import SessionAwareClient, LocalizingClient
 
 
-@pytest.fixture
-def db_and_empty_caches(db):
-    """
-    Use this fixture to emulate KumaTestCase. In addition to the
-    standard set-up and tear-down that the "db" fixture provides
-    (to emulate django.test.TestCase), it will clear the caches
-    before (but not after) the test is run.
-    """
-    for cache in caches.all():
-        cache.clear()
-
-    trans_real.deactivate()
-    trans_real._translations.clear()  # Django fails to clear this cache.
-    trans_real.activate(settings.LANGUAGE_CODE)
-
-
-@pytest.fixture
-def session_aware_client():
-    """
-    Provides a session-aware client.
-    """
-    return SessionAwareClient()
-
-
-@pytest.fixture
-def localizing_client():
-    """
-    Provides a session-aware client that localizes requests without a locale
-    in their requested URL.
-    """
-    return LocalizingClient()
+@pytest.fixture()
+def cleared_cacheback_cache():
+    caches[settings.CACHEBACK_CACHE_ALIAS].clear()
--- a/kuma/conftest.py
+++ b/kuma/conftest.py
@@ -1,0 +1,39 @@
+import pytest
+from django.conf import settings
+from django.core.cache import caches
+from django.utils.translation import trans_real
+
+from kuma.core.tests import SessionAwareClient, LocalizingClient
+
+
+@pytest.fixture
+def db_and_empty_caches(db):
+    """
+    Use this fixture to emulate KumaTestCase. In addition to the
+    standard set-up and tear-down that the "db" fixture provides
+    (to emulate django.test.TestCase), it will clear the caches
+    before (but not after) the test is run.
+    """
+    for cache in caches.all():
+        cache.clear()
+
+    trans_real.deactivate()
+    trans_real._translations.clear()  # Django fails to clear this cache.
+    trans_real.activate(settings.LANGUAGE_CODE)
+
+
+@pytest.fixture
+def session_aware_client():
+    """
+    Provides a session-aware client.
+    """
+    return SessionAwareClient()
+
+
+@pytest.fixture
+def localizing_client():
+    """
+    Provides a session-aware client that localizes requests without a locale
+    in their requested URL.
+    """
+    return LocalizingClient()
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -1,5 +1,6 @@
 import collections
 
+from django.conf import settings
 from django.contrib.auth import get_user_model
 
 from kuma.core.jobs import KumaJob, GenerationJob
@@ -66,6 +67,11 @@ class DocumentContributorsJob(KumaJob):
     version = 2
     # Don't synchronously fetch the contributor bar but schedule a fetch
     fetch_on_miss = False
+
+    def get(self, *args, **kwargs):
+        if settings.MAINTENANCE_MODE:
+            return self.empty()
+        return super(DocumentContributorsJob, self).get(*args, **kwargs)
 
     def fetch(self, pk):
         from .models import Document
--- a/kuma/wiki/tests/conftest.py
+++ b/kuma/wiki/tests/conftest.py
@@ -7,6 +7,11 @@ from datetime import datetime
 from ..models import Document, Revision
 
 
+class Object(object):
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
+
+
 @pytest.fixture
 def wiki_user(db, django_user_model):
     """A test user."""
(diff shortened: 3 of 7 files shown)


kuma/wiki/jobs.py
```diff
@@ -67,6 +68,11 @@ class DocumentContributorsJob(KumaJob):
     # Don't synchronously fetch the contributor bar but schedule a fetch
     fetch_on_miss = False
 
+    def get(self, *args, **kwargs):
+        if settings.MAINTENANCE_MODE:
+            return self.empty()
+        return super(DocumentContributorsJob, self).get(*args, **kwargs)
```
jwhitlock: 👍 works locally, good use of ``empty``.  In the sample DB, it can be hard to distinguish the empty contributor list, since all the gravatars are misses - need to find edits by me or sheppy.  We need to fix the default gravitar image in dev...

kuma/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+    for cache in caches.all():
+        cache.clear()
+
+    trans_real.deactivate()
+    trans_real._translations.clear()  # Django fails to clear this cache.
+    trans_real.activate(settings.LANGUAGE_CODE)
+
+
+@pytest.fixture
+def session_aware_client():
```
jwhitlock: This is unused in any current tests. I'm also not sure it is necessary - it may be a hack for limitations in Django before 1.8, and should be evaluated if it is actually needed.  My preference would be to leave it out until we need it.

kuma/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Provides a session-aware client.
+    """
+    return SessionAwareClient()
+
+
+@pytest.fixture
+def localizing_client():
```
jwhitlock: Again, unused in any current tests, and I'd like to wait until needed to add it.  This client works with our unique localizing scheme.  I hope to evaluate getting back to [built-in Django techniques](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#django.conf.urls.i18n.i18n_patterns) in the near future.

kuma/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+    standard set-up and tear-down that the "db" fixture provides
+    (to emulate django.test.TestCase), it will clear the caches
+    before (but not after) the test is run.
+    """
+    for cache in caches.all():
+        cache.clear()
+
+    trans_real.deactivate()
+    trans_real._translations.clear()  # Django fails to clear this cache.
+    trans_real.activate(settings.LANGUAGE_CODE)
```
jwhitlock: If possible, I'd like to get the test working without manually clearing caches and translations.  Do the tests pass without this fixture?  If not, can we replace it with a [more focused fixture](https://github.com/codeinthehole/django-cacheback/blob/master/tests/conftest.py#L28-L36) that just clears the needed cache?

kuma/wiki/tests/conftest.py
```diff
@@ -70,3 +118,107 @@ def trans_doc(create_revision, wiki_user):
...
+        content='<p>The root document re-envisioned.</p>',
+        comment='Done with the second revision.',
+        created=datetime(2017, 4, 18, 10, 15))
+    root_doc.save()
+
+    root_doc.current_revision = Revision.objects.create(
+        document=root_doc,
+        creator=banned_wiki_user.user,
+        content='<p>The root document re-envisioned with malice.</p>',
+        comment='Nuke the previous revision.',
```
jwhitlock: 😆

kuma/wiki/tests/conftest.py
```diff
@@ -70,3 +118,107 @@ def trans_doc(create_revision, wiki_user):
...
+        document=great_grandparent_doc,
+        creator=wiki_user,
+        content='<p>Great-grandparent...</p>',
+        title='Great-grandparent Document',
+        created=datetime(2017, 4, 24, 13, 49)
+    )
+
+    grandparent_doc = Document.objects.create(
+        locale='en-US',
+        slug='Grandparent',
```
jwhitlock: The document slugs should reflect the hierarchy - ``Great-grandparent/Grandparent``, ``Great-grandparent/Grandparent/Parent``.  Which may include changing the slugs, since they will quickly get ridiculous. ``DocumentZoneTests`` doesn't do this, but we should, and it's a [problem in production](https://bugzilla.mozilla.org/show_bug.cgi?id=1311142#c1) that may be forbidden by future code changes.

kuma/wiki/tests/test_jobs.py
```diff
@@ -1,141 +1,86 @@
...
+
+
+@pytest.mark.parametrize("mode", ["maintenance-mode", "normal-mode"])
+def test_contributors(db_and_empty_caches, settings, wiki_user_3,
+                      root_doc_with_mixed_contributors, mode):
+    """
+    Tests basic operation, ordering, caching, and handling of banned and
+    inactive contributors.
+    """
+    settings.MAINTENANCE_MODE = mode == "maintenance-mode"
```
jwhitlock: This confused me for 2 seconds. Need more coffee. ☕️ Or parenthesis.

kuma/wiki/tests/test_jobs.py
```diff
@@ -1,141 +1,86 @@
...
+    middle_zone = DocumentZone(document=middle_doc)
+    middle_zone.save()
+
+    assert unicode(top_zone) == u'DocumentZone {} ({})'.format(
+        top_doc.get_absolute_url(), top_doc.title)
+
+    assert get_zone_stack(top_doc) == [top_zone]
+    assert get_zone_stack(middle_doc) == [middle_zone, top_zone]
+    assert get_zone_stack(below_middle_doc) == [middle_zone, top_zone]
+    assert get_zone_stack(bottom_doc) == [middle_zone, top_zone]
```
jwhitlock: Nice rewrite, and I prefer this to how ``DocumentZoneTests.test_find_roots`` just tested the first element.

kuma/wiki/tests/conftest.py
```diff
@@ -7,6 +7,11 @@
 from ..models import Document, Revision
 
 
+class Object(object):
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
```
jwhitlock: I prefer [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) for packing, but I think others (@jgmize?) prefer this, so I can get used to it.

kuma/conftest.py
escattone: Great point, will see if it works with just the ``db`` fixture, and if not, something more focused as you suggested. Also, thanks for pointing out the ``django-cacheback`` fixtures, which I didn't realize existed.

kuma/conftest.py
escattone: Totally agree, will do!

kuma/conftest.py
escattone: Again, will do!

kuma/wiki/tests/conftest.py
escattone: I like ``namedtuple`` a lot as well, and it's lighter weight anyway (in terms of memory usage, not that it matters here), so I'll change to that.

kuma/wiki/tests/test_jobs.py
escattone: Yes, parentheses would be nice 😄

kuma/conftest.py
jwhitlock: I'm not sure if we can reuse the ``django-cacheback`` test fixtures (I think you have to add fixtures to a ``plugin.py`` and add other voodoo to reuse them), but I think we can duplicate and adjust for our needs.

escattone:
Thanks for the feedback @jwhitlock! I think this [commit](https://github.com/mozilla/kuma/pull/4196/commits/42367b492f323e536823249fd67ed072f3b1b870) addresses all of your requests?

jwhitlock:
Thanks for the updates! One more feature, one more test file refactored 😄

PR #4197, opened 2017-04-26
Upgrade gunicorn and use meinheld worker

jgmize:
See also https://github.com/mozmar/lumbergh/pull/194 and https://github.com/mozmar/snippets-service/pull/234

PR #4198, opened 2017-04-26
Rework kubectl apply to work around namespace issue

jgmize:
While trying to do a demo deploy in #4197 I ran into [namespace not found errors](https://ci.us-west.moz.works/blue/organizations/jenkins/mdn_multibranch_pipeline/detail/meinheld-demo/6/pipeline) which at first didn't make sense as I repeatedly verified that the namespace did in fact exist. The problem appears to have been some sort of shell escaping issue in the Jenkinsfile, as this [fixes the issue](https://ci.us-west.moz.works/blue/organizations/jenkins/mdn_multibranch_pipeline/detail/meinheld-demo/7/pipeline).

### DIFF
--- a/Jenkinsfiles/demo.groovy
+++ b/Jenkinsfiles/demo.groovy
@@ -12,7 +12,8 @@ stage('Deploy') {
 
   sh 'make deis-create-and-or-config'
   sh 'make render-k8s-templates'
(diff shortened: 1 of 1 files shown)


jgmize:
cc @metadave

metadave:
that's an annoying Jenkins bug!
👍

PR #4199, opened 2017-04-26
Nobug : Adds eslint and husky for linting goodness

schalkneethling:
Hey @stephaniehobson 

This adds `eslint` to the project. If you use an `eslint` linter plugin for your editor, you should get real time linting feedback. There is also a lint target with:

`npm run lint`

I have marked most things as warnings so, the pre-commit hook introduced here will pass for the moment.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,9 +1,7 @@
 module.exports = {
     "env": {
         "browser": true,
-        "es6": true,
-        "jquery": true,
-        "jasmine": true
+        "jquery": true
     },
     "extends": "eslint:recommended",
     "rules": {
@@ -61,6 +59,7 @@ module.exports = {
         "gettext": true,
         "interpolate": true,
         "mdn": true,
-        "Mozilla": true
+        "Mozilla": true,
+        "waffle": true
     }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,66 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true,
+        "jquery": true,
+        "jasmine": true
+    },
+    "extends": "eslint:recommended",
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,66 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true,
+        "jquery": true,
+        "jasmine": true
```
schalkneethling: We may not need `jasmine` here, not sure what MDN uses for JS tests.

.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+        "new-cap": 1
+    },
+    "globals": {
+        "CKEDITOR": true,
+        "FontFaceObserver": true,
+        "ga": true,
+        "gettext": true,
+        "interpolate": true,
+        "mdn": true,
+        "Mozilla": true
```
schalkneethling: Also, not sure if we need *all* of these as globals, added them as they came up but, with some of the exclusions I then introduced, perhaps they are not all needed.

.eslintrc.js
stephaniehobson: We don't have any JS tests right now. We will probably use jasmine if/when we do to be inline with bedrock but it should be removed for now.

.eslintrc.js
stephaniehobson: We also need to add `waffle`. Nothing jumps out at me from that last as something we can remove if we're not linting the libraries.

.eslintrc.js
```diff
@@ -0,0 +1,66 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true,
```
stephaniehobson: Our browser support requirements don't let us use ES6 yet.

PR #4200, opened 2017-04-26
bug 1328989: Move Template:CustomSampleCSS to CSS

jwhitlock:
Live samples load in an iframe, and include [Template:CustomSampleCSS](https://developer.mozilla.org/en-US/docs/Template:CustomSampleCSS?raw) as a stylesheet. This PR coverts the contents to a regular asset styles.css, and used that in the samples iframe source.

The macros [LiveSampleURL](https://github.com/mozilla/kumascript/blob/master/macros/LiveSampleURL.ejs) creates the iframe source href, and will change the URL from http://localhost:8000 to the SSL version at https://localhost:8000. You'll need to change this macro to get live samples in your development environment.  There will be a KumaScript PR to fix this soon.

### DIFF
--- a/kuma/static/styles/samples.scss
+++ b/kuma/static/styles/samples.scss
@@ -1,0 +1,33 @@
+/* Style for the iframe added by the macro EmbedLiveSample */
+
+body {
+    padding: 0;
+    margin: 0;
+}
+
+svg:not(:root) {
+    display: block;
+}
+
+/* Canvas docs, playable samples */
+.playable-code {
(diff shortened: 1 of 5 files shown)


kuma/static/styles/samples.scss
```diff
@@ -0,0 +1,33 @@
+/* Style for the embedded live sample iframe */
```
stephaniehobson: Please add the name of the macro that includes this file to the comment here.

PR #4202, opened 2017-04-27
Bug 1218563: Update "make clean", git and docker ignore lists

jwhitlock:
``make clean`` cleans out Python 3 compiled folders, and covers the functional test directory as well.

``.gitignore`` has been updated to model [GitHub's sample for Python projects](https://github.com/github/gitignore/blob/master/Python.gitignore), and includes a few more patterns.  I've added some files that have been created from following the development instructions, as well as ones created by the SCL3 deployment process.

``.dockerignore`` gets a copy of ``.gitignore``, so this has been updated as well.

A couple of notable changes:

* There were some "Don't ignore" patterns in ``.gitignore`` which appear left-over from previous iterations, and have been removed.
* GitHub ignores ``.pot`` files, which are locale template files.  However, we use these in our TravisCI locale build to detect if strings need to be updated.  It is possible we could switch to the English locale files for that check.

I've put the branch on this repo so that Jenkins will also try building the Docker images, and we can detect any issues from the ``.dockerignore`` earlier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ install:
 
 clean:
 	rm -rf .coverage build/
-	find . \( -name \*.pyc -o -name \*.pyo -o -name __pycache__ \) -prune -exec rm -rf {} +
+	find . \( -name \*.pyc -o -name \*.pyo -o -name __pycache__ \) -delete
 	mkdir -p build/locale
 
 locale:
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ install:
 
 clean:
 	rm -rf .coverage build/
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -51,7 +51,7 @@ install:
 
 clean:
 	rm -rf .coverage build/
-	find kuma -name '*.pyc' -exec rm {} \;
+	find . \( -name \*.pyc -o -name \*.pyo -o -name __pycache__ \) -prune -exec rm -rf {} +
```
jgmize: please `s/exec rm -f {}/delete/`

PR #4206, opened 2017-04-28
Bug 944186: Migrate all zone custom CSS to kuma, serve in zones.css

stephaniehobson:
All zones include only zone.css, they do not look to the database for anything. 

- removed zone.css from wiki.css
- edited zone logic to include zone.css but not the database field
- custom CSS copied out of django database
- broken into smaller component files
- edited to conform to code style guidelines
- duplicate code removed
- styles that are no longer in common usage removed
- added attribute selector based on data-slug to swap header images
  - also re-wrote the header image code to be more versatile
  - removed some of the custom image sizes and positions


jwhitlock:
The issue with loading subpages of zone landing pages went away on master, and were still gone when switching back to this repo.  Maybe I needed a ``docker-compose restart``.

Also, a Django test needs fixing for the new zone CSS logic.

Also, this change is awesome and I'm really excited about merging it and making zones saner.

kuma/static/styles/zones/components/image.scss
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+
+/*
+archive
+====================================================================== */
+
+body[data-slug^='Archive'] .zone-image {
+    background-image: url('https://mdn.mozillademos.org/files/1591/Nuvola_apps_important.png');
```
stephaniehobson: Yes, yes we should.

stephaniehobson:
@jwhitlock:

Changes in logo placement and size are expected and okay as long as nothing is broken. I think I have fixed the problem with the Games logo.

I brought the images into kuma and ran them through an image size optimizer for good measure.

My plan for the locales with the localized slugs:
1. wait until changes are approved and merged
1. create zone specific style sheets and PR
1. attach new zone specific styles sheets with new admin interface

We can either have this queued up and ready to go or find some other solution that does not involve a temporary regression. Something like a class name.

I don't think we should alter the slug behaviour.

jwhitlock:
Tests pass, [1 flake8 nitpick](https://travis-ci.org/mozilla/kuma/jobs/228494931#L382) and this can be merged

kuma/static/styles/zones/components/image.scss
```diff
@@ -0,0 +1,123 @@
...
+    @include bidi-value(background-position, top right, top left);
+    background-size: 100% auto;
+    bottom: 0;
+}
+
+/*
+firefox
+====================================================================== */
+body[data-slug^='Mozilla/Firefox'] .zone-image {
+    background-image: url($path-to-images + 'zones/firefox_logo-new.png');
```
jwhitlock: I missed this one - ``zones/firefox_logo-new.png`` isn't in the PR, and is causing issues in [Jenkins](https://ci.us-west.moz.works/view/MDN/job/mdn_multibranch_pipeline/job/master/888/consoleText)

PR #4207, opened 2017-04-29
"Hi I'm the Recursive Lambda God. I've refactored your codebase."

RecursiveLambdaGod:


escattone:
In my view, although this PR doesn't hurt anything, it adds no value. Old-style classes in Python 2 are totally valid, and in all of these cases where old-style classes are used, they're used as namespaces and do not benefit from any of the new-style class features. I'd recommend closing this.

jgmize:
@escattone I believe old style classes were removed in Python 3.

escattone:
@jgmize Yes, but we're still using Python 2, and in all of the instances this PR changes, the old-style classes (such as ``class Meta:``) are used as namespaces. There's no benefit in using ``class Meta(object):`` over ``class Meta:``. Django itself still uses  ``class Meta:`` in its documentation, for example [this](https://docs.djangoproject.com/en/1.11/topics/db/models/#meta-options). My only point is that this PR adds no value.

jwhitlock:
I agree it adds little to no value in these cases, so I'm not anxious to merge. At the same time, I'm a fan of small improvements by first-time contributors. I'm curious to see if the author is interested in improving the PR with a better commit message, or is just looking for self-promotion.

safwanrahman:
I believe its a bot and doing it automatically!

PR #4209, opened 2017-05-02
change styles on DocumentZone models

escattone:
This PR depends on https://github.com/mozilla/kuma/pull/4206. Perhaps the portion I'd most like reviewed is the [document template](https://github.com/escattone/kuma/blob/alt-zone-css-944186/kuma/wiki/jinja2/wiki/document.html) especially [``zone_stack``](https://github.com/escattone/kuma/blob/alt-zone-css-944186/kuma/wiki/jinja2/wiki/document.html#L28), [``is_zone_root``](https://github.com/escattone/kuma/blob/alt-zone-css-944186/kuma/wiki/jinja2/wiki/document.html#L30), and [zone CSS handling](https://github.com/escattone/kuma/blob/alt-zone-css-944186/kuma/wiki/jinja2/wiki/document.html#L63-L67).

* drop "styles" field on DocumentZone model
* add "css_slug" field on DocumentZone model
* add "doc_hierarchy_with_zones" fixture to
  kuma/wiki/tests/conftest.py
* redirect "wiki.styles" endpoint to "zones"
  CSS file (temporary perhaps? will this
  endpoint eventually go away?)
* modify Document.zone_stack method (if
  the document is for a non-English locale
  and it doesn't have its own zone stack,
  returns the zone stack of its English
  parent)
* modify the document template to reflect
  the new models
* refactor/add tests in test_views.py,
  test_jobs.py, test_models.py, and
  test_templates.py


kuma/wiki/models.py
```diff
@@ -1584,7 +1598,10 @@ class DocumentZone(models.Model):
     attributes inherited by the topic hierarchy beneath it.
     """
     document = models.OneToOneField(Document, related_name='zone')
-    styles = models.TextField(null=True, blank=True)
+    css_filename = models.CharField(
+        max_length=100, null=True, blank=True,
```
safwanrahman: ``Charfield`` Should not be ``null=True``. ``blank=True`` is enough

escattone:
Is it worth removing the idea of the zone stack (multiple possible zones per document)? In other words, a document can only have at most one zone, so we search from the document upwards (via ``parent_topic``) and stop at the first DocumentZone we find. Zones are used to specify an alternate URL path as well as a possible custom CSS file. The idea of a zone stack only really makes sense for specifying multiple layers of custom CSS files, but it doesn't seem necessary and I don;t think it's ever currently used. I don't think we have any cases of multiple zones per document except for this https://developer.mozilla.org/fr/docs/Archive/Mozilla/Persona (thanks to @jwhitlock for finding that) and maybe some others under the "Archive" zone.

jwhitlock:
Another one: https://developer.mozilla.org/en-US/docs/Archive/B2G_OS

I think this is the use case for stacked zones - you want a little of the B2G zone styling, but you also want the Archive styling to make it clear that it is deprecated.  I'm not sure what should win if we do away with the stack.  If B2G wins, then it looks like it is still active. If Archive wins, does the page break?  Any ideas @stephaniehobson ?

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -25,9 +25,9 @@
   {% set approvals_html = get_approvals_html(document, request.user) %}
 {% endif %}
 
-{% set zone_stack = document.zone_stack %}
+{% set zone_stack = document.zone_stack or document.default_zone_stack %}
 {% set is_zone = zone_stack|length %}
-{% set is_zone_root = is_zone and zone_stack[0].document == document %}
+{% set is_zone_root = is_zone and (zone_stack[-1].document in (document, document.parent)) %}
```
jwhitlock: I think the old way (index ``[0]`` versus ``[-1]``)was right.  Here's a zone stack w/ 2 zones:

* [Archive](https://developer.mozilla.org/en-US/docs/Archive) - Archive Zone
* [Archive/B2G_OS](https://developer.mozilla.org/en-US/docs/Archive/B2G_OS) - B2G zone
* [Archive/B2G_OS/Add-ons](https://developer.mozilla.org/en-US/docs/Archive/B2G_OS/Add-ons) - B2G Zone subpage

On Archive/B2G_OS/Add-ons, it detects that it is not the zone root, and doesn't apply the ``zone-landing`` class.

On Archive, it would detect that it is the zone root, and applies the ``zone-landing`` class.

On Archive/B2G_OS, the behavior would change.  Before, it would get the ``zone-landing`` class, and after this it would not.

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -60,8 +60,10 @@
 {% block site_css %}
     {{ super() }}
     {% if zone_stack %}
-      {% for zone in zone_stack|reverse() %}
-        <link rel="stylesheet" type="text/css" href="{{ url('wiki.styles', zone.document.slug) }}">
+      {% for zone in zone_stack|reverse() if zone.css_filename %}
+        {% stylesheet zone.css_filename %}
+      {% else %}
+        {% stylesheet 'zones' %}
```
jwhitlock: I'm not sure this is valid Jinja2 code.  You may need to split it into

```
{% if zone.css_filename %}
   {% for zone in zone_stack|reverse() %}
       {% stylesheet zone.css_filename %}
   {% endfor %}
{% else %}
   {% stylesheet 'zones' %}
{% endif %}
```
But, I could be wrong - we'll see when tests are working.

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -159,7 +161,7 @@
       {% else %}
         <div class="zone-article-header" id="document-main"><div class="center">
           <!-- zone title -->
-          <div class="zone-title"><a href="{{ zone_stack[0].document.get_absolute_url() }}">{{ zone_stack[0].document.title }}</a></div>
+          <div class="zone-title"><a href="{{ zone_stack[-1].document.get_absolute_url() }}">{{ zone_stack[-1].document.title }}</a></div>
```
jwhitlock: Again, I think index ``[0]`` was correct for stacked zones.

kuma/wiki/models.py
jwhitlock: Agreed, [Django docs](https://docs.djangoproject.com/en/1.8/ref/models/fields/#null) advise to avoid ``null`` on CharFields.

kuma/wiki/models.py
```diff
@@ -1584,7 +1598,10 @@ class DocumentZone(models.Model):
     attributes inherited by the topic hierarchy beneath it.
     """
     document = models.OneToOneField(Document, related_name='zone')
-    styles = models.TextField(null=True, blank=True)
+    css_filename = models.CharField(
```
jwhitlock: I'm chewing on the right name for this field.  For example, if you want ``zone-archive.css``, then you'd set ``css_filename`` to "zone-archive", which doesn't feel like a filename.  Maybe ``css_slug``? Maybe I'm being pedantic? 😕

kuma/wiki/tests/conftest.py
```diff
@@ -185,6 +185,40 @@ def doc_hierarchy(wiki_user):
...
+        css_filename='top.css',
+        url_root='fleetwood-mac'
+    )
+    DocumentZone.objects.create(
+        document=doc_hierarchy.middle_top,
+        url_root='spinners'
+    )
+    DocumentZone.objects.create(
+        document=trans_bottom_doc,
+        css_filename='paris.css',
```
jwhitlock: I think this would be ``paris``, not ``paris.css``.

kuma/wiki/models.py
```diff
@@ -1543,6 +1543,20 @@ def contributors(self):
     def zone_stack(self):
         return DocumentZoneStackJob().get(self.pk)
 
+    @cached_property
+    def default_zone_stack(self):
+        """
+        The zone stack of the default-language parent of this
+        non-default-language document, otherwise an empty zone
+        stack.
+        """
```
jwhitlock: I'm not sure if this should be a separate method, or combined with the ``zone_stack`` method.

kuma/wiki/models.py
escattone: Yeah, I think that's better, if only so in the future we don't end up with a mixture both NULL's and empty strings in that column.

kuma/wiki/models.py
```diff
@@ -1599,9 +1599,8 @@ class DocumentZone(models.Model):
     """
     document = models.OneToOneField(Document, related_name='zone')
     css_filename = models.CharField(
-        max_length=100, null=True, blank=True,
-        help_text="name of an alternative CSS file for documents under this "
-                  "zone")
+        max_length=100, blank=True, help_text="name of an alternative CSS file "
+                                              "for documents under this zone")
     url_root = models.CharField(
         max_length=255, null=True, blank=True, db_index=True,
```
jwhitlock: Yeah, they did this one wrong.  Don't trust existing code!

_checks blame to make sure he didn't do it_

Don't trust it!

kuma/wiki/jinja2/wiki/document.html
escattone: I actually got the idea from http://jinja.pocoo.org/docs/2.9/templates/#for

kuma/wiki/jinja2/wiki/document.html
escattone: Thanks @jwhitlock. I think I understand now. I was interpreting the meaning of the zone root as the document being associated with the topmost zone, rather than the nearest zone.

kuma/wiki/models.py
escattone: Yes, I think it makes more sense to fold this into the ``zone_stack`` method.

kuma/wiki/models.py
escattone: No, that makes sense.  It's assumed that this is always a CSS file, so there's no need for the extension. I like ``css_slug``.

kuma/wiki/models.py
escattone: 😃 👍

escattone:
Thanks for the initial feedback @jwhitlock !

stephaniehobson:
We don't need nested zones. Once we add back the ability to pick a custom style sheet we could handle the B2G and Persona zones by giving them custom styles that include the archive styles. I think we should simplify the code to only use the first / closest zone.

escattone:
This is ready for review again. I didn't remove nested zones. If we decide for sure we want to do that, perhaps I can do that in a separate PR (and there might be some CSS changes required as well?).

kuma/wiki/tests/conftest.py
```diff
@@ -185,6 +185,103 @@ def doc_hierarchy(wiki_user):
...
+        rendered_html='<p>Oben...</p>',
+        parent=top_doc
+    )
+    Revision.objects.create(
+        document=de_top_doc,
+        creator=wiki_user_2,
+        based_on=top_doc.current_revision,
+        content='<p>Oben...</p>',
+        title='Oben Dokument',
+        created=datetime(2017, 4, 30, 10, 03)
```
jwhitlock: I agree w/ GitHub highlighting - ``3`` not ``03``, which means '3 in octal' (08, 09 don't work)

kuma/wiki/tests/conftest.py
```diff
@@ -185,6 +185,103 @@ def doc_hierarchy(wiki_user):
 
 
 @pytest.fixture
+def doc_hierarchy_with_zones(settings, doc_hierarchy, wiki_user_2, wiki_user_3):
```
jwhitlock: For future discussion - when a fixture is just used in one test or test file, should it go in ``conftest.py`` or in that test file?  I'm leaning toward test-specific fixtures in the test files, but I'm willing to wait until we have more than 50% of tests converted to pytest to start moving things around.

kuma/wiki/tests/conftest.py
```diff
@@ -185,6 +185,103 @@ def doc_hierarchy(wiki_user):
...
+        parent=top_doc
+    )
+    Revision.objects.create(
+        document=fr_top_doc,
+        creator=wiki_user_3,
+        based_on=top_doc.current_revision,
+        content='<p>Haut...</p>',
+        title='Haut Document',
+        is_approved=True,
+        created=datetime(2017, 4, 30, 12, 01)
```
jwhitlock: ``1`` not ``01``.

kuma/wiki/tests/test_templates.py
```diff
@@ -1329,3 +1329,110 @@ def _translation_data():
...
+
+
+def test_zone_styles(client, doc_hierarchy_with_zones, root_doc):
+    """
+    Check document page for zone-style-related features.
+    """
+    top_doc = doc_hierarchy_with_zones.top
+    bottom_doc = doc_hierarchy_with_zones.bottom
+    middle_top_doc = doc_hierarchy_with_zones.middle_top
+    fr_top_doc = top_doc.translations.filter(locale='fr').first()
```
jwhitlock: You can say this shorter as ``fr_top_doc = top_doc.translations.get(locale='fr')``.

kuma/wiki/tests/test_templates.py
```diff
@@ -1329,3 +1329,110 @@ def _translation_data():
         'content': 'loremo ipsumo doloro sito ameto',
         'toc_depth': Revision.TOC_DEPTH_H4,
     }
+
+
+def test_zone_styles(client, doc_hierarchy_with_zones, root_doc):
```
jwhitlock: Covers all the test cases. I might have tried to parameterize it, but it is clear enough as one after the other.

kuma/wiki/tests/conftest.py
escattone: Yeah, I've been putting most things if not everything into the higher-level ``conftest.py`` files, thinking they would most likely be used by more than one test file, but I agree that we should eventually (or even now if you prefer) put test-specific fixtures in the test files.

In this case, ``doc_hierarchy_with_zones`` is actually used by multiple files, so I think it belongs here, but ``doc_hierarchy`` is not used (except by ``doc_hierarchy_with_zones``) and ``root_doc_with_mixed_contributors`` is only used in ``test_jobs.py``. I'm going to fold ``doc_hierarchy`` into ``doc_hierarchy_with_zones`` since it's not used by any tests.

kuma/wiki/tests/conftest.py
escattone: oops!

kuma/wiki/tests/conftest.py
escattone: Oops!

kuma/wiki/tests/test_templates.py
escattone: Good point, I'll try to parametrize it.

kuma/wiki/tests/test_templates.py
escattone: Great, thanks.

kuma/wiki/models.py
```diff
@@ -1584,7 +1594,10 @@ class DocumentZone(models.Model):
     attributes inherited by the topic hierarchy beneath it.
     """
     document = models.OneToOneField(Document, related_name='zone')
-    styles = models.TextField(null=True, blank=True)
```
jwhitlock: We should keep styles for now, so that there are both ``styles`` and ``css_slug`` in prod push 1, and then drop ``styles`` in the next production push.  Otherwise, we'll have a flood of 500s between the migration and the code deployment.

escattone: Ah, good catch, thanks!

escattone:
Thanks for the reviews @jwhitlock and @stephaniehobson . If there are no other concerns, this can wait for #4206 to merge, then be rebased and ready for a final review.

stephaniehobson:
@escattone Please remove the nested zone behaviour for style sheets. I only want one extra style sheet to be attached to a page in a zone. For nested zones that style sheet should be the one defined for the closest zone, not any parent zones.

escattone:
@stephaniehobson OK, will do.

escattone:
OK, I think this is ready for another review. The zone stack idea has been replaced with a single "nearest" zone. In the course of replacing zone stack with nearest zone, I discovered that the zone stack cache invalidation was almost completely broken. It has been fixed, thoroughly tested, and I think solidly covers all of the cases (modifying fields on the zone -- including the document ID, as well as deleting the zone).

kuma/wiki/migrations/0032_alt-zone-css.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('wiki', '0031_add_data_to_revisionip'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='documentzone',
+            name='css_slug',
+            field=models.CharField(help_text=b'name of an alternative pipeline CSS group for documents under this zone', max_length=100, blank=True),
+        ),
```
jwhitlock: 👍 keeps the in-DB CSS for now.

kuma/wiki/models.py
```diff
@@ -1540,8 +1538,24 @@ def contributors(self):
...
+        """
+        The nearest zone for this document, starting from this document and
+        moving upwards via "parent_topic". For a non-default-language document
+        that does not have its own nearest zone, returns the nearest zone of
+        its parent (the document it was translated from).
+        """
+        job = DocumentNearestZoneJob()
+        result = job.get(self.pk)
+        if not result:
+            if (self.locale != settings.WIKI_DEFAULT_LANGUAGE) and self.parent:
```
jwhitlock: nitpick: this could be one long ugly ``if`` statement instead of a nested ``if``

kuma/wiki/apps.py
```diff
@@ -1,48 +1,49 @@
...
+
+    def invalidate(pk):
+        do_invalidate(pk)
+        # Since the descendants of this document search upwards for their
+        # nearest zone, recursively invalidate their caches. Note that the
+        # branches of this tree of decendants can stop at (and exclude) any
+        # descendants that have their own zones.
+        children = (Document.objects
+                            .filter(parent_topic=pk)
+                            .values_list('pk', 'zone'))
```
jwhitlock: Nitpick: Using ``zone__id`` (double underscore) will just load the ID instead of a full ``DocumentZone`` instance.

kuma/wiki/tests/conftest.py
```diff
@@ -175,12 +218,56 @@ def doc_hierarchy(wiki_user):
...
-
+    DocumentZone.objects.create(
+        document=top_doc,
+        css_slug='lindsey',
+        url_root='fleetwood-mac'
+    )
+    DocumentZone.objects.create(
+        document=middle_top_doc,
+        css_slug='bobby',
+        url_root='spinners'
```
jwhitlock: Is this a music reference too?  I don't know this one...

kuma/wiki/tests/test_jobs.py
```diff
@@ -1,35 +1,183 @@
...
-    middle_bottom_doc = doc_hierarchy.middle_bottom
-    bottom_doc = doc_hierarchy.bottom
-
-    top_zone = DocumentZone.objects.create(document=top_doc)
-    middle_top_zone = DocumentZone.objects.create(document=middle_top_doc)
-
-    assert unicode(top_zone) == u'DocumentZone {} ({})'.format(
+def test_document_zone_unicode(doc_hierarchy_with_zones):
+    top_doc = doc_hierarchy_with_zones.top
+    assert unicode(top_doc.zone) == u'DocumentZone {} ({})'.format(
```
jwhitlock: 😄 more code coverage!

kuma/wiki/tests/test_jobs.py
```diff
@@ -1,35 +1,183 @@
...
+@pytest.mark.parametrize('doc_name,expected_zone_name', [
+    ('top', 'top'),
+    ('middle_top', 'middle_top'),
+    ('middle_bottom', 'middle_top'),
+    ('bottom', 'middle_top'),
+    ('root', None),
+])
+def test_nearest_zone(doc_hierarchy_with_zones, root_doc,
+                      cleared_cacheback_cache, doc_name, expected_zone_name):
+    """
```
jwhitlock: I like the parametrized version. 18 lines fits in my brain well. Thank you.

kuma/wiki/tests/test_views.py
```diff
@@ -4478,69 +4478,38 @@ def test_move_conflict(self):
...
+    response = client.get(url, follow=True)
 
-        url = reverse('wiki.styles', args=(self.sub_doc.slug,),
-                      locale=settings.WIKI_DEFAULT_LANGUAGE)
-        response = self.client.get(url, follow=True)
-        eq_(404, response.status_code)
+    assert response.redirect_chain
+    redirect_location, redirect_code = response.redirect_chain[-1]
+    assert redirect_location.endswith('build/styles/zones.css')
+    assert redirect_code == 302
```
jwhitlock: I thought there was an easier way to do this, but I was wrong.  I was thinking of Django's [SimpleTestCase.assertRedirects](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.SimpleTestCase.assertRedirects).  If we do this a lot, we can add a py.test port.

kuma/wiki/apps.py
escattone: In my tests, using ``zone`` was equivalent to ``zone__id``, just returning the ID, but I'm happy to change to ``zone__id`` to make it more explicit.

kuma/wiki/models.py
escattone: Will do.

kuma/wiki/tests/conftest.py
escattone: 😃 Bobby Smith was one of the lead singers for the [Spinners](https://en.wikipedia.org/wiki/The_Spinners_(American_R%26B_group)).

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -59,8 +59,12 @@
 
 {% block site_css %}
     {{ super() }}
-    {% if zone_stack %}
+    {% if nearest_zone %}
+      {% if nearest_zone.css_slug %}
+        {% stylesheet nearest_zone.css_slug %}
```
escattone: I'll change it.

kuma/wiki/apps.py
jwhitlock: nope, if it is the same, ``zone`` is clearer than ``zone__id``. I learned something new about ``values_list``.

escattone:
Thanks for the review @jwhitlock! I addressed your suggestions with another [commit](https://github.com/mozilla/kuma/pull/4209/commits/b6e019e1ab2b6bee94cef29374fc67103a1fcbbb), including prepending "zone-" to the css-slug.

PR #4212, opened 2017-05-02
Bug 1218563: Only ignore /static/ in root

stephaniehobson:
There's multiple `static/` folders in the repo, we only want to ignore the one at the root.


stephaniehobson:
Added 👍

PR #4215, opened 2017-05-04
implement SCL3 httpd redirects in Django

metadave:
This PR uses @pmac's https://github.com/pmac/django-redirect-urls lib to implement SCL3 Apache httpd redirects in Python/Django. Original rewrite rules are saved as comments to each Django redirect. Test layout is similar to the [Bedrock redirect tests](https://github.com/mozilla/bedrock/tree/master/tests/redirects). 

Some (or all) of these redirects may be unusued/leftover from [A Long Time Ago](http://www.merrimentdesign.com/images/a-long-time-ago-in-a-galaxy-far-far-away-free-star-wars-printable-sign.jpg).

The only rewrite rule that isn't implemented in this PR is as follows:

```
RewriteCond %{SERVER_NAME} ^developer\.mozilla\.com$
RewriteRule (.*) http://developer.mozilla.org$1 [R=301,L]
```

which redirects requests to `developer.mozilla.com/*` --> `developer.mozilla.org/*`

Tracked in [this infra issue](https://github.com/mozmar/infra/issues/180)

cc @jgmize @pmac @jwhitlock @escattone 

--- 

[This site](http://martinmelin.se/rewrite-rule-tester/) was a great help when working with mod_rewrite.


kuma/settings/common.py
```diff
@@ -527,6 +530,7 @@ def enable_candidate_languages():
 
     # other
     'kuma.humans',
+    'redirect_urls',
```
jgmize: Is this needed? I would think that the `kuma.redirects` app added above should be sufficient.

kuma/redirects/redirects.py
```diff
@@ -0,0 +1,51 @@
+from redirect_urls import redirect
+
+# Redirects/rewrites/aliases migrated from SCL3 httpd config
+redirectpatterns = [
+    # RewriteRule ^/media/(redesign/)?css/(.*)-min.css$
+    # /static/build/styles/$2.css [L,R=301]
+    redirect(r'^media/(?:redesign/)?css/(?P<doc>.*)-min.css$',
```
pmac: Might want the `prepend_locale=False` arg on some/most of these. Probably works without, but more regex than you need.

requirements/dev.txt
```diff
@@ -46,3 +46,6 @@ tox==2.3.1 \
 Werkzeug==0.11.4 \
     --hash=sha256:7db3cb2d4725be0680abf64a45b18229186f03ad8b9989abbe053f9357b17b37 \
  --hash=sha256:e48fb7e3f2bb5a740dd9a666624699a4d83e2e028555f9c46bcc8ecfc2cd8c32
+django-redirect-urls==0.1 \
```
jwhitlock: ``dev.txt`` includes ``default.txt``, so this should be removed

requirements/test.txt
```diff
@@ -61,3 +61,9 @@ pytest-rerunfailures==1.0.2 \
 pytest-base-url==1.1.0 \
     --hash=sha256:90a1ce6a00a558231117b5aee32300edecbc0c2fd701c13e8cec62177900d28c \
     --hash=sha256:001c1b2678dae82c2891db415251361b29ed6824cb2085e4821635119499071e
+pytest-django==3.1.2 \
```
jwhitlock: Is ``pytest-django`` actually needed in functional tests?

tests/redirects/map_301.py
```diff
@@ -0,0 +1,34 @@
+import pytest
```
jwhitlock: ``pytest`` imported but unused

tests/redirects/map_301.py
```diff
@@ -0,0 +1,34 @@
...
+from redirect_urls import redirect
```
jwhitlock: ``redirect`` imported but unused

tests/redirects/test_redirects.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import absolute_import
+from operator import itemgetter
```
jwhitlock: ``itemgetter`` imported but unused

kuma/redirects/models.py
```diff
@@ -0,0 +1 @@
+from django.db import models
```
jwhitlock: can this file be omitted?

requirements/test.txt
```diff
@@ -61,3 +61,9 @@ pytest-rerunfailures==1.0.2 \
...
+    --hash=sha256:038ccc5a9daa1b1b0eb739ab7dce54e495811eca5ea3af4815a2a3ac45152309 \
+    --hash=sha256:00995c2999b884a38ae9cd30a8c00ed32b3d38c1041250ea84caf18085589662
+django-redirect-urls==0.1 \
```
jwhitlock: I think this is unused in functional tests

jwhitlock:
Also, moving ``tests/functional/urls.py`` to ``tests/redirects/base.py`` will break functional (Selenium) tests.

kuma/settings/common.py
pmac: It is needed. This loads the machinery that then goes to find the `redirects.py` files in the other apps.

kuma/redirects/models.py
pmac: I believe the file is required for django to consider it an "app" (that could be old info though). It can be empty though. Looks like the default models.py when you use `manage.py startapp`.

requirements/test.txt
metadave: I couldn't test redirects without it.

metadave:
going to split this into 2 PRs: one for requirements so the base Docker image can be updated, and another with the redirects impl.

PR #4219, opened 2017-05-05
Add seperate zone style sheets, doesn't remove combined zone.css yet

stephaniehobson:
Adds separate CSS files for individual zones, to be pulled in using the admin interface after PR #4209.

Does not remove the individual zone styles from zones.scss yet, so there's a bunch of duplicated styles. A follow up PR will remove that duplication once we've transitioned to using the new separate files.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -684,6 +684,48 @@ PIPELINE_CSS = {
         ),
         'output_filename': 'build/styles/zones.css',
     },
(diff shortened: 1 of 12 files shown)


stephaniehobson:
Not possible to test without #4209 but it is theoretically possible to merge this now without messing up the styles on anything. @jwhitlock do you want to give it a look on an instance with the new data model?

PR #4210, opened 2017-05-02
Bug 1308322: Upgrade pytest and most plugins

jwhitlock:
Convert ``requirements/test.txt`` to include comments and use ``requirements/constraints.txt``. Then update py.test and most plugins

* pytest 2.8.7 -> 3.0.7: Breaking changes release, remove and rewrite features, better Py3 and unicode support, xfail(strict=True), deprecate yield_fixture, more assert rewriting
* pytest-rerunfailures 1.0.2 -> 2.1.0: pytest 3.x support
* pytest-xdist 1.15.0 -> 1.16.0: New scheuduler hook
* pytest-cov 2.2.1 -> 2.4.0: xdist fixes
* pytest-django 2.9.1 -> 3.1.2: pytest 3 support, expanded Django support, public database setup API, mailoutbox fixture
* coverage 4.0.3 -> 4.3.4: Improved branch analysis, improved HTML report, smarter combine, tox.ini config support.
* py 1.4.31 -> 1.4.33: Bug fixes

Django tests and coverage worked for me.  I was able to run the functional tests locally with Chrome.  This is the same as #4195, except that I've omitted the Selenium upgrades.

PR #4220, opened 2017-05-05
implement SCL3 httpd redirects in Django

metadave:
previously: https://github.com/mozilla/kuma/pull/4215

This PR uses @pmac's https://github.com/pmac/django-redirect-urls lib to implement SCL3 Apache httpd redirects in Python/Django. The complex rewrites retain original mod_rewrite rules as comments, while the remaining are uncommented simple static redirects. Test layout is similar to the [Bedrock redirect tests](https://github.com/mozilla/bedrock/tree/master/tests/redirects). 

The only rewrite rule that isn't implemented in this PR is as follows:

```
RewriteCond %{SERVER_NAME} ^developer\.mozilla\.com$
RewriteRule (.*) http://developer.mozilla.org$1 [R=301,L]
```

which redirects requests to `developer.mozilla.com/*` --> `developer.mozilla.org/*`

Tracked in [this infra issue](https://github.com/mozmar/infra/issues/180)

### DIFF
--- a/tests/redirects/map_301.py
+++ b/tests/redirects/map_301.py
@@ -25,8 +25,8 @@ URLS = flatten((
              "/docs/Web/Demos_of_open_web_technologies/",
              status_code=requests.codes.found),
 
-    url_test("/foo//bar", "/foo_bar"),
-    url_test("/foo//bar//baz", "/foo_bar_baz"),
+    url_test("/docs/Mozilla/Projects/NSPR/Reference/I//O_Functions", "/docs/Mozilla/Projects/NSPR/Reference/I_O_Functions"),
+    url_test("/docs/Mozilla/Projects/NSPR/Reference/I//O//Functions", "/docs/Mozilla/Projects/NSPR/Reference/I_O_Functions"),
 
     url_test("/samples/canvas-tutorial/2_1_canvas_rect.html" ,"/docs/Web/API/Canvas_API/Tutorial/Drawing_shapes#Rectangular_shape_example"),
     url_test("/samples/canvas-tutorial/2_2_canvas_moveto.html" ,"/docs/Web/API/Canvas_API/Tutorial/Drawing_shapes#Moving_the_pen"),
--- a/tests/redirects/map_301.py
+++ b/tests/redirects/map_301.py
@@ -1,0 +1,143 @@
+from utils.urls import flatten, url_test
+
+import requests
+
+URLS = flatten((
+    url_test("/media/redesign/css/foo-min.css", "/static/build/styles/foo.css"),
+    url_test("/media/css/foo-min.css", "/static/build/styles/foo.css"),
+
+    url_test("/media/redesign/js/foo-min.js", "/static/build/js/foo.js"),
+    url_test("/media/js/foo-min.js", "/static/build/js/foo.js"),
+
+    url_test("/media/redesign/img.foo", "/static/img.foo"),
+    url_test("/media/img.foo", "/static/img.foo"),
+
+    url_test("/media/redesign/css.foo", "/static/styles.foo"),
+    url_test("/media/css.foo", "/static/styles.foo"),
+
+    url_test("/media/redesign/js.foo", "/static/js.foo"),
+    url_test("/media/js.foo", "/static/js.foo"),
+
+    url_test("/media/redesign/fonts.foo", "/static/fonts.foo"),
+    url_test("/media/fonts.foo", "/static/fonts.foo"),
+
+    url_test("/media/uploads/demos/foobar123",
+             "/docs/Web/Demos_of_open_web_technologies/",
+             status_code=requests.codes.found),
+
+    url_test("/foo//bar", "/foo_bar"),
+    url_test("/foo//bar//baz", "/foo_bar_baz"),
+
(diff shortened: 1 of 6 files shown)


tests/redirects/test_redirects.py
```diff
@@ -0,0 +1,33 @@
...
+from .map_301 import GITHUB_IO_URLS
+from .map_301 import MOZILLADEMOS_URLS
+
+# while these test methods are similar, they're each testing a
+# subset of redirects, and it was easier to work with them separately.
+@pytest.mark.headless
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('url', REDIRECT_URLS)
+def test_redirects(url):
+    url['base_url'] = "https://developer.mozilla.org"
```
jwhitlock: This should match [bedrock](https://github.com/mozilla/bedrock/blob/master/tests/redirects/test_redirects.py#L18-L19), using the ``base_url`` py.test fixture:

```
def test_redirects(url, base_url):
    url['base_url'] = base_url
```

You can then test it against staging:

```
py.test tests/redirects -vv --base-url https://developer.allizom.org --pdb
```
or the local Docker:

```
py.test tests/redirects -vv --base-url http://localhost:8000 --pdb
```

The ``--pdb`` drops you into the [Python debugger](https://docs.python.org/2/library/pdb.html), where you can print out the expected and actual values ("q" and Enter to quit).  If we were on py.test 3.x, the assets may be more useful (see PR #4210), but w/ py.test 2.x, you have to do your own test debugging.

tests/redirects/map_301.py
```diff
@@ -0,0 +1,143 @@
...
+
+    url_test("/media/redesign/js.foo", "/static/js.foo"),
+    url_test("/media/js.foo", "/static/js.foo"),
+
+    url_test("/media/redesign/fonts.foo", "/static/fonts.foo"),
+    url_test("/media/fonts.foo", "/static/fonts.foo"),
+
+    url_test("/media/uploads/demos/foobar123",
+             "/docs/Web/Demos_of_open_web_technologies/",
+             status_code=requests.codes.found),
```
jwhitlock: This fails against ``--base_url=localhost:8000``.  It expects a 302 Found, but gets a 301 Moved Permanently.

tests/redirects/map_301.py
```diff
@@ -0,0 +1,143 @@
...
+
+    url_test("/foo//bar", "/foo_bar"),
```
jwhitlock: This fails against ``--base_url=localhost:8000``, getting ``http://localhost:8000/foo//foo_bar``.

It also fails against ``--base_url=https://developer.allizom.org``, getting ``https://en-US/foo/bar``.  I think the staging server's Apache config is wrong.

I also question the purpose of this pattern, but I guess it was a problem at some point (see [bug 779501](https://bugzilla.mozilla.org/show_bug.cgi?id=779501))

tests/redirects/map_301.py
```diff
@@ -0,0 +1,143 @@
...
+    url_test("/foo//bar//baz", "/foo_bar_baz"),
```
jwhitlock: This fails against ``--base_url=localhost:8000``, getting ``http://localhost:8000/bar_baz``

It also fails against ``--base_url=https://developer.allizom.org``, getting ``https://developer.allizom.org/f/en-US/foo/bar/baz``.

WEIRD.

jwhitlock:
There's some missing stuff:

* 3 Tests that pass in developer.mozilla.org but fail in developer.allizom.org or Docker dev
* The [.htaccess](https://github.com/mozilla/kuma/blob/master/configs/htaccess) redirects
* Tests are not integrated into acceptance tests

but, I'm happy with what is here so far. It's up to @metadave if he wants to keep working in this PR, or merge and add further refinements in a future PR.

metadave:
@jwhitlock I pushed some fixes to get all the tests to pass on `http://localhost:8000` and `https://developer.mozilla.org`, however there are 2 remaining failures on `https://developer.allizom.org` related to the following tests:

```
    url_test("/docs/Mozilla/Projects/NSPR/Reference/I//O_Functions", "/docs/Mozilla/Projects/NSPR/Reference/I_O_Functions"),
    url_test("/docs/Mozilla/Projects/NSPR/Reference/I//O//Functions", "/docs/Mozilla/Projects/NSPR/Reference/I_O_Functions"),
```

which correspond to the following rewrites:

```
RewriteRule ^(.*)//(.*)//(.*)$ $1_$2_$3 [R=301,L,NC]
RewriteRule ^(.*)//(.*)$ $1_$2 [R=301,L,NC]
```

I'd prefer to submit * the [.htaccess](https://github.com/mozilla/kuma/blob/master/configs/htaccess) redirects in a separate PR.

jwhitlock:
Nice work @metadave. I'll diff the staging and production configuration, and see why there is a difference for those URL patterns.

jwhitlock:
I've looked at the staging and production Apache configuration, and the two given rewrite rules are missing from staging.  I've opened [infra bug 1365774](https://bugzilla.mozilla.org/show_bug.cgi?id=1365774) to request syncing the two configurations.

jwhitlock:
The configurations have been synced, and the tests now pass against staging: ``py.test tests/redirects/``.  I think we're ready to add these tests to the integration test suite.

PR #4223, opened 2017-05-08
bug 1330106: Add ga to context

jwhitlock:
Change the variable 'ga' from implicit to explicit, to avoid an "Uncaught ReferenceError: ga is not defined" in local development. Tested with default of no Google Analytics account, and with a fake "UA-000000-0" settings for constance config "GOOGLE_ANALYTICS_ACCOUNT".


stephaniehobson:
👍

PR #4228, opened 2017-05-09
bug 1353079: Add meta tag to avoid search indexing

jwhitlock:
For pages that we don't want indexed (based on robots.txt), set a head tag:

```
```

Remove some Jinja "set meta=" declarations that weren't doing anything.

### DIFF
--- a/kuma/wiki/jinja2/wiki/list/documents.html
+++ b/kuma/wiki/jinja2/wiki/list/documents.html
@@ -12,6 +12,7 @@
   {% set title = _('All documents') %}
 {% endif %}
 {% block title %}{{ page_title(title) }}{% endblock %}
+{% block robots_value %}noindex, nofollow{% endblock %}
 {% set crumbs = [(None, title)] %}
 
 {% block extrahead %}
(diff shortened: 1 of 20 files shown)


kuma/wiki/jinja2/wiki/list/documents.html
```diff
@@ -12,6 +12,7 @@
   {% set title = _('All documents') %}
 {% endif %}
 {% block title %}{{ page_title(title) }}{% endblock %}
+{% block robots_value %}noindex, nofollow{% endblock %}
```
stephaniehobson: I think we want list pages followed. That's [one of these pages](https://developer.mozilla.org/en-US/docs/tag/HTML), right?

jwhitlock: This page has been blocked in robots.txt since 2016 in https://github.com/mozilla/kuma/commit/7ef8e07b2d31e2179f81206e2c26faebf8dd1570, as part of the downtime mitigation for [bug 1199166](https://bugzilla.mozilla.org/show_bug.cgi?id=1199166).  I'm not sure I agree with that, but this makes it consistent.

stephaniehobson: Part of another different issue then...

jwhitlock: Opened [bug 1364963](https://bugzilla.mozilla.org/show_bug.cgi?id=1364963). I focused on just the tag indexes. I think other indexes, such as pages with KS errors and that need review, can continue being excluded from search indexes.

PR #4233, opened 2017-05-11
Bug 1328989: Remove the remaining Template-handling logic

jwhitlock:
* Remove Jinja2 code paths for templates
* Don't set ``is_template`` when saving documents
* Don't set ``is_template`` in tests and test fixtures
* Migrate some out-of-date apps, and remove the template-specific permissions.

This does _not_ delete the ``is_template`` column. It requires a full table rebuild, and took 22 minutes in the developer environment. We'll need to schedule a DB maintenance window to drop the column.


safwanrahman:
@jwhitlock Why dont delete `is_template` from the model?

jwhitlock:
Thanks for the review, @escattone!

@safwanrahman, the reason that there is no database migration to drop ``is_template`` is the one given in the PR comment. It will be a long migration (22 minutes locally, much faster in production, but still long).  During the migration, the ``Document`` table will be locked for reads and writes, which means users will see timeouts and MDN will be unusable.  We'll need to schedule that for a maintenance window, such as a weekend. That can be done in it's own PR, and the rest of these changes are useful without it.

I also tried other things, such as:

* Making it a [NullBooleanField](https://docs.djangoproject.com/en/1.8/ref/models/fields/#nullbooleanfield) so that the database will default to ``null`` if the field is omitted from ``CREATE`` statements.  Without it, we get an error if we remove the field from the model.
* Dropping the database index

Both of these modifications also require a full table rebuild. And, over an hour of timing and testing. I decided to leave the field untouched with a comment for now.

PR #4234, opened 2017-05-11
bug 944186: drop styles column from DocumentZone

escattone:
Completes the update of the DocumentZone table by dropping the "styles" column, which is already no longer used in the deployed code. This is a follow-on to PR #4209.


escattone:
Sorry, slow to realize that @jwhitlock already took care of this in the pending PR #4233. Closing.

PR #4237, opened 2017-05-12
bug 1346462: Upgrade bleach and dependencies

jwhitlock:
* [bleach](http://bleach.readthedocs.io/en/latest/changes.html#version-2-0-march-8th-2017) 1.5.0 -> 2.0.0: Rewrite with new interfaces for speed and bug fixes
* [html5lib](http://html5lib.readthedocs.io/en/latest/changes.html) 1.0b6 -> 1.0b10: compliant with Encoding Standard, avoid XSS bug, move filters._base to filters.base
* [webencodings](https://pythonhosted.org/webencodings/) 0.5.1: New dependency of html5lib
* The bleach and html5lib APIs have changed, requiring code changes.
* The behaviour of the bleach filter for ``<script>`` and ``<style>`` has changed as well, so there is slightly different test behaviour, but the output is still safe.

@willkg does this look like you were thinking back in March?

To do:
* [x] Test bleaching, extracting content in anonymized production database

_Update_: Current html5lib is 1.0b6, not 1.0b8.  Commit comment and PR updated.


willkg:
Just as a thing to quell fears, html5lib had an xss security vulnerability where it didn't quote attributes by default. Bleach always told it to quote attributes, so Bleach 1.5 wasn't affected by that issue in html5lib.

I can look at this on Monday.

jwhitlock:
Kuma has a ``Document.objects.clean_content()`` method that uses bleach and html5lib filters to sanitize content after it is rendered with KumaScript, and the output is stored in ``Document.html``.  I found a few differences between this archived cleaned content and the new cleaned content:

* Element attribute order (can be normalized w/ html5lib)
* Removal of some invalid inline styles (``<dd style="margin-left: =;">`` becomes ``<dd style="">``
* Removal of HTML comments, like ``<!-- comment -->``, including ``<!--?xml version="1.0"?-->`` in a ``<pre>`` tag
* Removal of images with data tags, like ```` on https://developer.mozilla.org/ja/docs/Web/API/Response (which appears to be from a copy-paste from Google Translate)

``Document.html`` can be very old for pages that haven't been updated in a long time, so I'm not sure if these differences need to be addressed in code or not.  As I was typing this up, I decided a good test would be the same comparison, but using bleach 1.5.0 and htmllib 1.0b6, the current production versions, and seeing the 3-way diff.  And that is why the "not ready" tag flipped.

jwhitlock:
The ``Document.objects.clean_content()`` output is identical between bleach 1.5.0 / html5lib 1.0b6 and bleach 2.0 / html5lib 1.0b10.  Kuma needs to grow a mechanism for invalidating sanitized output when the sanitizer is updated.

johan68:
Thanks so much...

willkg:
Bleach defaults to alphabetizing attributes to keep them stable between passes. I think that's a good thing.

Removing data: things--I think you can fix that. Maybe add "data" to the list of allowed protocols so it's not nixed in that section and then add an attribute filter that checks data things and makes sure they're the right content type or something like that.

Is the "removed comments even in a pre tag" a problem? If so, we can figure out how to work around that. Maybe disable `strip_comments` works?

stephaniehobson:
I'm not sure we should be allowing images with `data` in wiki articles...

jwhitlock:
To be clear, if someone saves a page in production today:

* Comments are stripped
* Attributes are alphabetized
* Images with ``src="data:..."`` (and I believe any element with a ``data:`` protocol) has the ``src`` attribute removed, removing a possible security hole

Bleach 1.5.0 does the correct thing, and Bleach 2.0 doesn't change anything.  The issue is that we cache the output of sanitation in the database (because it is _s l o w_), and some articles have not been updated in years.  It may be prudent to re-render and re-sanitize our current content, but this PR will have no effect on that process.

willkg:
Oh, I get it! That makes sense.

I have vague memories of talking about writing a cron job that would run once a night and update the 100 stalest pages or something like that as a kind of ongoing hand-wavey maintenance thing.

jwhitlock:
I've rebased to master and fixed an instance of double-quotes instead of single-quotes.

Here's where we're at:

* bleach 1.50 / html5lib 1.0b6 provide the same cleaned content as bleach 2.0 / htmllib 1.0b10
* There's some pages that have stale cleaned content, and won't be re-rendered until someone edits them. I've opened [bug 1365987](https://bugzilla.mozilla.org/show_bug.cgi?id=1365987) for the idea of periodically re-rendering all content.

Are people happy merging this scope-limited package update, and saving the re-rendering work for future PRs?

PR #4239, opened 2017-05-16
bug 1352232: Remove Ansible from py27 build

jwhitlock:
Don't use Ansible to install MySQL and ElasticSearch.  Instead:

* Use TravisCI's MySQL, without the custom collation or migrations.  Docker continues to test w/ ``utf8_distinct_ci`` and migrations.
* Install ElasticSearch as suggested by [TravisCI docs](https://docs.travis-ci.com/user/database-setup/#Installing-ElasticSearch-on-trusty-container-based-infrastructure)
* Install node packages needed by Django pipeline, so that ``make build-static`` will work for tests.
* Rewrite ``scripts/travis_install`` to use env variables like ``INSTALL_ELASTICSEARCH=1``, instead of targeting on ``TOXENV``, to make it easier to combine and customize.

All of these changes makes the ``py27`` target faster to build (8 minutes vs 11 min 30 seconds, 30% faster), and easier to customize.  I've added the first customization for [bug 1352232](https://bugzilla.mozilla.org/show_bug.cgi?id=1352232), where tests fail if PYTHONHASHSEED is random, which affects our [readiness for Python 3](http://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed).

@jgmize this might be an easy 👍, requesting your review if you can get to it by next Monday.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,8 @@ services:
     - mysql
 env:
     global:
-        - CFLAGS=-O0
         - DATABASE_URL=mysql://root:@127.0.0.1:3306/kuma
         - DJANGO_SETTINGS_MODULE=kuma.settings.travis
-        - DEBIAN_FRONTEND=noninteractive
         - DOCKER_COMPOSE_VERSION=1.9.0
           # ES_VERSION must come before ES_DOWNLOAD_URL
         - ES_VERSION=2.4.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,9 +58,9 @@ install:
     - pip install -U pip
     - pip install -r requirements/travis.txt
     # Wait for ElasticSearch to be ready
-    - if [ ${INSTALL_ELASTICSEARCH:-0} -ne 0 ]; then wget -q --waitretry=1 --retry-connrefused -T 10 -O - http://127.0.0.1:9200; fi;
+    - if [[ "$INSTALL_ELASTICSEARCH" == "1" ]]; then wget -q --waitretry=1 --retry-connrefused -T 10 -O - http://127.0.0.1:9200; fi;
 script:
-    if [[ "${PYTHONHASHSEED:-}" == "0" ]]; then
+    if [[ "$PYTHONHASHSEED" == "0" ]]; then
         tox -v --hashseed=noset;
     else
         tox -v;
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ env:
           CREATE_DB=kuma
           INSTALL_PIPELINE=1
           INSTALL_ELASTICSEARCH=1
+          PYTHONHASHSEED=0
         - TOXENV=flake8
         - TOXENV=docs
         - TOXENV=locales
@@ -40,9 +41,16 @@ env:
           INSTALL_DOCKER_COMPOSE=1
           UID=0
         - TOXENV=stylelint
+        # TODO: make this default when http://bugzil.la/1127798 is fixed
+        - TOXENV=py27
+          CREATE_DB=kuma
+          INSTALL_PIPELINE=1
+          INSTALL_ELASTICSEARCH=1
+
 matrix:
     allow_failures:
         - env: TOXENV=locales INSTALL_DOCKER_COMPOSE=1
+        - env: TOXENV=py27 CREATE_DB=kuma INSTALL_PIPELINE=1 INSTALL_ELASTICSEARCH=1
 install:
     - nvm install 6
     - nvm use 6
@@ -52,7 +60,11 @@ install:
     # Wait for ElasticSearch to be ready
     - if [ ${INSTALL_ELASTICSEARCH:-0} -ne 0 ]; then wget -q --waitretry=1 --retry-connrefused -T 10 -O - http://127.0.0.1:9200; fi;
 script:
-    tox -v;
+    if [[ "${PYTHONHASHSEED:-}" == "0" ]]; then
+        tox -v --hashseed=noset;
+    else
+        tox -v;
+    fi
 after_failure:
     - dmesg | tail
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,38 +2,59 @@ sudo: required
 branches:
     only:
         - master
-cache: pip
+cache:
+    pip: True
+    directories:
+      - node_modules
+      - downloads
 language: python
 python:
     - "2.7"
 services:
     - docker
     - memcached
+    - mysql
 env:
-    matrix:
-        - TOXENV=py27
-        - TOXENV=flake8
-        - TOXENV=docs
-        - TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
-        - TOXENV=docker DOCKER_COMPOSE_VERSION=1.9.0 UID=0
-        - TOXENV=stylelint
     global:
+        - CFLAGS=-O0
+        - DATABASE_URL=mysql://root:@127.0.0.1:3306/kuma
         - DJANGO_SETTINGS_MODULE=kuma.settings.travis
         - DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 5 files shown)


scripts/travis-install
```diff
@@ -1,34 +1,55 @@
...
-if [ "$TOXENV" == "py27" ]
+# Download ElasticSearch
+if [ ${INSTALL_ELASTICSEARCH:-0} -ne 0 ]
 then
-    sudo apt-get update -qq
-    sudo apt-get -y install build-essential libxml2-dev libxslt-dev libjpeg8 libjpeg8-dev libfreetype6 libfreetype6-dev zlib1g-dev sqlite3 tidy libtidy-dev libtidy-0.99-0 python-dev libffi-dev libssl-dev
-    sudo ln -s /usr/lib/`uname -i`-linux-gnu/libfreetype.so /usr/lib
-    sudo ln -s /usr/lib/`uname -i`-linux-gnu/libz.so /usr/lib
+    mkdir -p downloads
+    wget -q -O downloads/elasticsearch-$ES_VERSION.tar.gz $ES_DOWNLOAD_URL
```
safwanrahman: Can you build the url here?
Rather than passing the url in environment parameters

.travis.yml
```diff
@@ -2,38 +2,69 @@ sudo: required
...
     - nvm install 6
     - nvm use 6
-script: tox -v
+    - scripts/travis-install
+    - pip install -U pip
+    - pip install -r requirements/travis.txt
+    # Wait for ElasticSearch to be ready
+    - if [ ${INSTALL_ELASTICSEARCH:-0} -ne 0 ]; then wget -q --waitretry=1 --retry-connrefused -T 10 -O - http://127.0.0.1:9200; fi;
+script:
+    if [[ "${PYTHONHASHSEED:-}" == "0" ]]; then
```
jgmize: `${PYTHONHASHSEED:-}` seems redundant, I think the simpler `${PYTHONHASHSEED}` would be equivalent.

scripts/travis-install
```diff
@@ -1,34 +1,55 @@
...
+    mkdir -p downloads
+    DOCKER_COMPOSE_FILE=downloads/docker-compose-${DOCKER_COMPOSE_VERSION}
+    if [ ! -f $DOCKER_COMPOSE_FILE ]
+    then
+        wget -q --waitretry=1 --retry-connrefused -T 10 \
+            -O downloads/docker-compose-${DOCKER_COMPOSE_VERSION} \
+            https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m`
+    fi
+    chmod +x $DOCKER_COMPOSE_FILE
+    sudo cp $DOCKER_COMPOSE_FILE /usr/local/bin/docker-compose
```
safwanrahman: Can we cache this path and check if the version matches with the provided one?
```bash
INSTALLED_DOCER_COMPOSE_VERSION=$(docker-compose -v | grep -o '[0-9]\?[0-9].[0-9]\?[0-9].\?[0-9]')
if ! [ INSTALLED_DOCER_COMPOSE_VERSION == DOCKER_COMPOSE_VERSION];
then
    # Download and install
fi
```

scripts/travis-install
jwhitlock: No, you can't. ElasticSearch keeps changing their downloads domain.  Here's a sample:

Version | URL
---|---
2.2.0|https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.2.0/elasticsearch-2.2.0.tar.gz
1.7.6|https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.6.tar.gz
5.3.0|https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.3.0.tar.gz

scripts/travis-install
jwhitlock: No, that will never ever happen.  The VM is reset each time, so the version will be 1.8.1 until TravisCI updates their VM image.  When they do that, the new default may be later than 1.9.0, and we can stop installing our own docker-compose.  Until then, the ``download`` folder is cached, so we don't download the file unless we change ``DOCKER_COMPOSE_VERSION``.

safwanrahman: @jwhitlock We can add `/usr/local/bin/docker-compose` to `cache` directory to keep the `docker-compose`

.travis.yml
```diff
@@ -2,38 +2,69 @@ sudo: required
...
-        - env: TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
-before_install:
-    - scripts/travis-install
-    - pip install -U pip
+        - env: TOXENV=locales INSTALL_DOCKER_COMPOSE=1
+        - env: TOXENV=py27 CREATE_DB=kuma INSTALL_PIPELINE=1 INSTALL_ELASTICSEARCH=1
 install:
-    - pip install -r requirements/travis.txt
-    - if [[ $TOXENV == 'py27' ]]; then ansible-playbook -vvv --tags=mysql,pipeline --connection=local --inventory-file=provisioning/inventory provisioning/travis.yml; fi
     - nvm install 6
```
safwanrahman: Can we include all the installs in ``scripts/travis-install``?
We dont need to install pip and other thing in docker testing environment. So it will faster the build

tox.ini
```diff
@@ -7,13 +7,12 @@ whitelist_externals = make
...
 commands =
-    make compilejsi18n collectstatic coveragetest target={posargs:kuma}
-setenv =
-    PYTHONPATH = .
-    CFLAGS = -O0
-    # TODO: remove once http://bugzil.la/1127798 is fixed
-    PYTHONHASHSEED = 0
-passenv=DJANGO_SETTINGS_MODULE DEBIAN_FRONTEND CFLAGS
+    make compilejsi18n collectstatic clean
+    py.test --no-migrations --cov=kuma kuma
```
safwanrahman: If new migration is added, will it fail?

.travis.yml
```diff
@@ -2,38 +2,69 @@ sudo: required
...
+          INSTALL_PIPELINE=1
+          INSTALL_ELASTICSEARCH=1
+          PYTHONHASHSEED=0
         - TOXENV=flake8
         - TOXENV=docs
-        - TOXENV=locales DOCKER_COMPOSE_VERSION=1.9.0
-        - TOXENV=docker DOCKER_COMPOSE_VERSION=1.9.0 UID=0
+        - TOXENV=locales
+          INSTALL_DOCKER_COMPOSE=1
+        - TOXENV=docker
```
safwanrahman: Keeping it in single line seems more readable to me ``TOXENV=docker INSTALL_DOCKER_COMPOSE=1 UID=0``

scripts/travis-install
jwhitlock: You can cache directories but not files. I think it would be a mistake to cache ``/usr/local/bin``, so I think we're OK caching the file in ``downloads`` and copying it each time.

.travis.yml
jgmize: I prefer the new multi-line approach.

scripts/travis-install
jgmize: the docker-compose file is already in the downloads dir, which is in the cache due to https://github.com/mozilla/kuma/pull/4239/files#diff-354f30a63fb0907d4ad57269548329e3R9

tox.ini
jgmize: Not here, since we aren't running them due to `--no-migrations`.

.travis.yml
jwhitlock: This might be a good follow-on PR.  ``requirements/travis.txt`` includes ``coveralls``, which is called at the end of each test, so you'd need to ensure ``coveralls`` is not called for the targets that generate coverage data.  Which maybe should include the docker target, but currently doesn't.  But it could be skipped for the linter tasks.

That's also a lot of work to save at most 10s per build.

PR #4240, opened 2017-05-16
bug 1363916: add liveness/readiness endpoints for k8s

escattone:
This PR implements "liveness" and "readiness" endpoints as used by Kubernetes.
- They are implemented as Middleware and placed at the beginning of the middleware stack to avoid unnecessary processing (sessions, authentication, etc.) for endpoints that will be hit frequently.
- The philosophy of the "liveness" endpoint is to simply indicate whether the service itself is up and running, while the "readiness" endpoint goes one step further to also include whether its essential supporting services are also running and usable. I've made the decision for now that for Kuma "essential supporting services" means the database only, not the other supporting services (elasticsearch, Celery, memcached, kumascript), since Kuma can remain usable without them.
- The "readiness" endpoint checks the usability of the database via a simple, indexed (fast) [query on the Document table](https://github.com/escattone/kuma/blob/k8s-health-checks-1363916/kuma/health/middleware.py#L47), but I also had briefly explored using something like this instead:
```
    cursor.execute("SELECT 1")
```
- I also tested the "readiness" endpoint locally, by stopping the "mysql" service (``docker-compose stop mysql``) and ensuring that the "readiness" endpoint returned a 503 (with a detailed reason).

### DIFF
--- a/kuma/health/tests/test_views.py
+++ b/kuma/health/tests/test_views.py
@@ -1,0 +1,31 @@
+import mock
+import pytest
+from django.db import DatabaseError
+from django.core.urlresolvers import reverse
+
+
+@pytest.mark.parametrize('http_method',
+                         ['get', 'head', 'put', 'post', 'delete', 'options'])
+def test_liveness(db, client, http_method):
+    url = reverse('health.liveness')
+    response = getattr(client, http_method)(url)
+    assert (response.status_code ==
+            204 if http_method in ('get', 'head') else 405)
+
+
+@pytest.mark.parametrize('http_method',
+                         ['get', 'head', 'put', 'post', 'delete', 'options'])
+def test_readiness(db, client, http_method):
+    url = reverse('health.readiness')
+    response = getattr(client, http_method)(url)
+    assert (response.status_code ==
+            204 if http_method in ('get', 'head') else 405)
+
+
+def test_readiness_with_db_error(db, client):
+    url = reverse('health.readiness')
+    with mock.patch('kuma.wiki.models.Document.objects') as mock_manager:
+        mock_manager.filter.side_effect = DatabaseError('fubar')
+        response = client.get(url)
+    assert response.status_code == 503
+    assert 'fubar' in response.reason_phrase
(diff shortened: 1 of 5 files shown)


kuma/health/middleware.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def process_request(self, request):
+        if request.path == '/healthz':
+            return handle_liveness(request)
+
+        if request.path == '/readiness':
+            return handle_readiness(request)
+
+
+@require_GET
```
jwhitlock: I think ``HEAD`` should be allowed for  the liveness endpoint, especially since the response has no content.  The [require_safe](https://docs.djangoproject.com/en/1.8/topics/http/decorators/#django.views.decorators.http.require_safe) decorator would do that.

kuma/health/middleware.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    A successful response from this endpoint simply proves
+    that Django is up and running. It doesn't mean that its
+    supporting services (like MySQL, memcached, Celery) can
+    be successfully used from within this service.
+    """
+    return HttpResponse(status=204)
+
+
+@require_GET
```
jwhitlock: Same here, I think ``HEAD`` should be allowed.

escattone:
Thanks for the review @jwhitlock, and per our conversation over video this morning, I'm going to convert these to views (which I had already explored in a [another branch](https://github.com/escattone/kuma/tree/k8s-probe-endpoints-1363916)).

escattone:
@jwhitlock - Would you mind giving this another look when you get a chance?

kuma/health/middleware.py
escattone: Didn't know about ``require_safe``, thanks!

kuma/health/middleware.py
escattone: Good point!

jwhitlock:
👍 looks good, works well. ``204 No Content`` is weird in the browser, but ``curl`` works just fine.

PR #4241, opened 2017-05-19
Nearest-zone-job adjustments 944186

escattone:
This PR addresses two problems introduced in #4209 (addressing bug [944186](https://bugzilla.mozilla.org/show_bug.cgi?id=944186)):
- it hopefully eases the spike in celery tasks (see bug [1366044](https://bugzilla.mozilla.org/show_bug.cgi?id=1366044))
- fixes an exception when getting the nearest-zone for a document whose parent-topic has been deleted (bug [1365960](https://bugzilla.mozilla.org/show_bug.cgi?id=1365960))

I wanted to prove (via a test case) that using ``None`` for the "empty" value for the ``DocumentNearestZoneJob`` class effectively bypasses the cache, along the lines of something like this:
```
# doc is some document that doesn't have a nearest zone.
assert not DocumentNearestZoneJob().get(doc.pk) 
with mock.patch.object(DocumentNearestZoneJob, 'fetch') as mock_fetch:
    DocumentNearestZoneJob().get(doc.pk)
    mock_fetch.assert_not_called()
```
which I thought should fail (on the mock assertion) when using ``None`` for empty and succeed when using ``False``, but for the life of me I couldn't get it to fail when using ``None``. A long, detailed look at the ``django-cacheback`` code has me convinced that it should fail, but it doesn't in my tests.

### DIFF
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -17,7 +17,12 @@ class DocumentNearestZoneJob(KumaJob):
         document and going upwards via topic parents.
         """
         from .models import Document, DocumentZone
-        get_parent_id = Document.objects.values_list('parent_topic', flat=True).get
+
+        # Using "admin_objects" here to get around the
+        # filter that excludes deleted documents.
+        get_parent_id = (Document.admin_objects
+                                 .values_list('parent_topic', flat=True)
+                                 .get)
         while pk:
             try:
                 return DocumentZone.objects.get(document=pk)
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -8,7 +8,7 @@ from kuma.users.templatetags.jinja_helpers import gravatar_url
 
 
 class DocumentNearestZoneJob(KumaJob):
-    lifetime = 60 * 60 * 3
+    lifetime = 60 * 60 * 29
     refresh_timeout = 60
 
     def fetch(self, pk):
--- a/kuma/wiki/tests/test_jobs.py
+++ b/kuma/wiki/tests/test_jobs.py
@@ -34,6 +34,20 @@ def test_nearest_zone(doc_hierarchy_with_zones, root_doc,
     assert DocumentNearestZoneJob().get(doc.pk) == zone
 
 
+def test_nearest_zone_when_deleted_parent_topic(doc_hierarchy_with_zones,
+                                                cleared_cacheback_cache):
+    """
+    Make sure we handle the case where we try to get the nearest
+    zone for a document whose parent-topic has been deleted.
+    """
+    bottom_doc = doc_hierarchy_with_zones.bottom
+    middle_top_zone = doc_hierarchy_with_zones.middle_top.zone
+    # Delete the parent-topic of the bottom doc.
+    doc_hierarchy_with_zones.middle_bottom.delete()
+    # We should still successfully get the nearest zone for the bottom doc.
+    assert DocumentNearestZoneJob().get(bottom_doc.pk) == middle_top_zone
+
+
 def test_nearest_zone_cache_invalidation_on_save_shallow(doc_hierarchy_with_zones,
                                                          cleared_cacheback_cache):
     job = DocumentNearestZoneJob()


jwhitlock:
@escattone - I've done a little more research, and I think I sent you down a dead end.

First, we're on django-cacheback 1.0, so make sure you are reading that code, not the latest.  I'd like to update that and all our dependencies, but that's another PR.

First place in the tour is ``get``, which is where all the action happens.  If the cache is unset, it returns ``None``, which triggers the [first branch](https://github.com/codeinthehole/django-cacheback/blob/1b39c173bba9b2ac50f512d40d864fb1c3d3d7c0/cacheback/base.py#L103-L129).  This does a sync or async call to get and store the value.  A sync call calls [refresh](https://github.com/codeinthehole/django-cacheback/blob/1b39c173bba9b2ac50f512d40d864fb1c3d3d7c0/cacheback/base.py#L222-L230), which calls ``fetch`` and stores the result.  An async call stores ``self.empty()``, so that future calls will not schedule, and schedules a future ``refresh``.

In both cases, [cache_set](https://github.com/codeinthehole/django-cacheback/blob/1b39c173bba9b2ac50f512d40d864fb1c3d3d7c0/cacheback/base.py#L202-L220) is called.  This stores the tuple ``(expiry, data)``.  So, if ``fetch`` returns ``None``, something like ``(1495211607.434809, None)`` is stored in the cache.  This is different than the plain ``None`` returned by a cache miss.

So, ``None`` is a valid return from ``fetch``, and we don't have to go through gymnastics to avoid ``None``.

I have other guesses, but I don't want to send you down any other dead ends, so doing more research first.

jwhitlock:
OK, further research:

* With ``fetch_on_miss=True``, I think we go down the sync path for ``get``.
* The cache is invalidated whenever the root zone changes
* The cache is refreshed as background tasks every 3 hours.
* If you fetch a deeply nested translated page, like [fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/from](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/from), it generates many cacheback jobs (12 in this case, as it walks up the tree).  This is what search engine spiders do periodically.

So, I don't think there is a error causing needless cacheback jobs. I'd support increasing the cache limit to something like 29 hours, so that the cache lasts more than a day and maybe avoids being triggered by a once-a-day crawler but instead gets distributed by organic traffic. Plus, 29 is prime.

escattone:
@jwhitlock: Oh, of course, you're absolutely right, thanks for looking into that. It's frustrating how at times something can be "staring you in the face" and I still don't see it. I'll close this PR and submit a new one just for bug [1365960](https://bugzilla.mozilla.org/show_bug.cgi?id=1365960).

escattone:
@jwhitlock: Just read your latest comment. Ah yes, the cache timeout is short, especially given that with #4209 we should be invalidating the cache properly. So how does this sound, I'll modify this PR as follows:
- increase the cache timeout to 29 hours (or maybe even more?)
- change the empty case back to ``None``

jwhitlock:
The only thing I could think of that could go wrong is if you move a page into a zone, it could take a while for the zone to be applied.  But, that could be fixed by invalidating on the Document save signal.

escattone:
@jwhitlock We are already [invalidating the nearest-zone cache on document save](https://github.com/mozilla/kuma/blob/master/kuma/wiki/apps.py#L114) so I think we're covered for that case?

escattone:
OK, I refactored the two commits:
- I backed-out the change to the "empty" value of nearest-zone jobs, and instead increased their cache lifetime to 29 hours, which should hopefully ease the spike in celery tasks (see bug [1366044](https://bugzilla.mozilla.org/show_bug.cgi?id=1366044))
- I kept the fix for the exception when getting the nearest-zone for a document whose parent-topic has been deleted (bug [1365960](https://bugzilla.mozilla.org/show_bug.cgi?id=1365960)), but slightly modified the comment on the test function.

escattone:
Ready for your review again @jwhitlock!

PR #4243, opened 2017-05-24
Check window.ga directly instead of passing it in

stephaniehobson:
Old method wasn't working reliably (race conditions?) This performs better in local testing.

### DIFF
--- a/kuma/static/js/wiki-helpful-survey.js
+++ b/kuma/static/js/wiki-helpful-survey.js
@@ -1,9 +1,8 @@
-(function(ga, $, mdn) {
+(function($, mdn) {
     'use strict';
 
     var $survey;
     var $votes;
-    var gaInterval;
     var gaChecks = 0;
 
     // ga loads async, so there's a chance it wont be available when this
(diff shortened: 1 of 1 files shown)


kuma/static/js/wiki-helpful-survey.js
```diff
@@ -1,17 +1,16 @@
...
-    var gaInterval;
     var gaChecks = 0;
 
     // ga loads async, so there's a chance it wont be available when this
     // script is parsed. we'll give it 2 seconds.
     function checkGA() {
         // survey won't do much good if we can't store the data
         // checking for ".create" due to Ghostery mocking of ga
-        if (ga && ga.create) {
+        if (window.ga && ga.create) {
```
jwhitlock: Should this be ``if (window.ga && window.ga.create)`` ?

stephaniehobson: I believe it does not go on to check the second condition if the first fails and accessing ga globally if the first passes is not a problem. I could not explain why this is.

PR #4244, opened 2017-05-25
[WIP] Bug 1366868: Add PostCSS to asset pipeline

jwhitlock:
Here's a first pass at adding PostCSS to the asset pipeline.  I'm not sure if this is anything close to what we want.

This is the first time I've looked seriously at Kuma's asset pipeline, and it is in bad shape.  There's files that are being copied to the ``static`` folder that shouldn't be, such as the ckeditor builder java .jar file, and Photoshop .psd files.  In addition, all of the ``.scss`` files are being copied.  I would expect only final files served to end users to appear in this folder.  It may be possible to work within the Django staticfiles app configuration to omit most of these files.  A better solution would be to omit all files from a ``static`` folder that shouldn't be auto-processed into end user assets.  That would mean moving some files to a new place, including all the ``.scss`` files.

Our django-pipeline implementation is a mess.  It relies on the ``.scss`` files being copied to the root``static`` folder.  The output doesn't make a lot of sense - ``Post-processed 'build/styles/wiki.css' as 'build/styles/wiki.css'`` is misleading, claiming to have processed a file onto itself.  It doesn't support our use case of processing a file with Sass and then PostCSS, although I've hacked it here.  I've dug into the project itself, and it needs a maintainer.  I'd rather not pick up another abandoned project.

We have a parallel build system written in gulp, which is the one used by the front-end devs.  As we add PostCSS plugins, we'll be installing node packages.  This will require parallel installs in the Dockerfile and in the local developer's environment.  This feels like a step backwards, and asking for pain and bugs as we move forward.

At the same time, gulp is a better fit for both Sass and PostCSS workflows.  I suspect it would be trivial to teach it to process JS with UglifyJS. I'd want to move it into a Docker container.  I may need to watch while a front end dev does development to understand the issues.

I also can't tell if the PostCSS output is correct, or an improvement on the existing CSS created by Sass.  It looks worse to me.

But, here's the PR! Don't merge, but feel free to try it out.  If it is close enough to start moving functionality to PostCSS, we can start setting up SCL3 for it.  If not, I'd rather have a few rounds of fixing our assets and workflow first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ env:
         - PIPELINE_JS_COMPRESSOR=pipeline.compressors.uglifyjs.UglifyJSCompressor
         - PIPELINE_SASS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/node-sass
         - PIPELINE_UGLIFYJS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/uglifyjs
+        - PIPELINE_POSTCSS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/postcss
+        - PIPELINE_POSTCSS_ARGUMENTS="--config $TRAVIS_BUILD_DIR/postcss.config.js"
     matrix:
         - TOXENV=py27
           CREATE_DB=kuma
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -40,11 +40,15 @@ RUN npm install -g \
     node-sass@4.3.0 \
     uglify-js@2.4.13 \
     clean-css@3.4.23 \
-    stylelint@7.10.1
+    stylelint@7.10.1 \
+    postcss-cli@4.0.0 \
+    autoprefixer@7.1.1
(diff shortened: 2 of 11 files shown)


jwhitlock:
I rebased and rearranged the commits a bit, including adding a missing ``postcss.config.js``.  If we're sure we're going down this path, we should get PostCSS and autoprefixer in the Docker images, so I've moved those to the top so we can open a new PR.

I also suspect that we should standardize on ``npm install`` and ``npm install --production``, with ``npm-shrinkwrap.json`` enforcing versions,  rather than litter the codebase with environment-specific ``npm install package@x.x.x`` calls for TravisCI, Docker, gulp-driven dev, and SCL3 deployments.

After sleeping on it, I've progressed from horror to determination.  I definitely want to improve the static asset workflow. However, I've got other stuff on my plate, and want to clear that first.

jwhitlock:
Working on improving the existing static asset pipelines before tackling this again - https://tree.taiga.io/project/viya-mdn-durable-team/us/1799

jwhitlock:
I think the other tasks will need to be done first, and not much of this PR will remain. Closing.

PR #4246, opened 2017-05-30
Remove double save

jwhitlock:
This is the change needed for document-saved.


stephaniehobson:
I think I fixed it based on your comment :)

PR #4247, opened 2017-06-01
Fix Bug 1367098, replace saucelabs promo with devrel survey on landing page

schalkneethling:
@stephaniehobson r?


kuma/landing/jinja2/landing/homepage.html
```diff
@@ -112,13 +112,13 @@ <h2 class="offscreen">{{ _('Web Technologies') }}</h2>
 <div class="home-callouts">
   <h2 class="offscreen">{{ _('Open Web and Mobile') }}</h2>
   <div class="column-container center">
-    <div class="column-callout callout-saucelabs">
-      <a href="https://www.saucelabs.com/cross-browser-testing-tutorial?utm_campaign=cbt&amp;utm_medium=banner&amp;utm_source=home-page">
-        <span>
-          <p>{{ _('Test in Firefox and over 800 browser/OS combos with Mozilla and Sauce Labs') }}</p>
+    <div class="column-callout callout-devrel">
+      <a href="https://www.surveygizmo.com/s3/3544552/d11af99eb1ed">
```
stephaniehobson: This link is a 404. (It's an accurate copy of that they asked us to include though)

schalkneethling: Maciej is looking into it

kuma/landing/jinja2/landing/homepage.html
```diff
@@ -112,13 +112,13 @@ <h2 class="offscreen">{{ _('Web Technologies') }}</h2>
 <div class="home-callouts">
   <h2 class="offscreen">{{ _('Open Web and Mobile') }}</h2>
   <div class="column-container center">
-    <div class="column-callout callout-saucelabs">
-      <a href="https://www.saucelabs.com/cross-browser-testing-tutorial?utm_campaign=cbt&amp;utm_medium=banner&amp;utm_source=home-page">
-        <span>
-          <p>{{ _('Test in Firefox and over 800 browser/OS combos with Mozilla and Sauce Labs') }}</p>
```
jwhitlock: There's something to watch out for next time (although I hope we don't do this a lot).  Removing this string from the code also removes it from the translation files, and if it comes back in a few weeks, the translators will have to translate it again.  It may be better to use HTML comments (``<!-- comment -->``) or Jinja2 comments (``{# comment #}``, which is omitted from output) to "remove" sections w/ translated text temporarily.  In any case, our translators are awesome, and will retranslate in a couple of days when we put it back.

schalkneethling: Oh crud :-/ Note taken. I will remember to do this should we ever need to again. Thanks @jwhitlock

PR #4248, opened 2017-06-02
bug 1271509: Add ./manage.py scrape_document

jwhitlock:
Fix some stuff that should have been in the last PR (less user test fixtures, improve the scraper), and then add the ``./manage.py scrape_document`` command.  Some manual tests:

    ./manage.py scrape_document https://developer.mozilla.org/en-US/docs/Archive/SAX # Will also add Archive zone
    ./manage.py scrape_document https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array --depth 1  # Will scrape child docs as well
    ./manage.py scrape_document https://developer.mozilla.org/fr/docs/Web/CSS/display  # Will scrape the English document as well

### DIFF
--- a/kuma/scrape/management/commands/scrape_document.py
+++ b/kuma/scrape/management/commands/scrape_document.py
@@ -29,7 +29,7 @@ class Command(ScrapeCommand):
                                   ' ("all" for all)'))
 
     def handle(self, *arg, **options):
-        self.setup_logging(options.get('verbosity'))
+        self.setup_logging(options['verbosity'])
         host, ssl, path = self.parse_url_or_path(options['url'])
         scraper = self.make_scraper(host=host, ssl=ssl)
 
--- a/kuma/scrape/management/commands/scrape_document.py
+++ b/kuma/scrape/management/commands/scrape_document.py
@@ -1,0 +1,48 @@
+"""Scrape a wiki document from a Kuma website."""
+
+from . import ScrapeCommand
+
+
+class Command(ScrapeCommand):
+    help = 'Scrape a wiki document.'
+
+    def add_arguments(self, parser):
+        """Add arguments for scraping an MDN wiki document."""
+        parser.add_argument('url',
+                            metavar='URL_OR_PATH',
+                            help='URL or path to a wiki page')
+        parser.add_argument('--force',
+                            help='Update existing Document record',
+                            action='store_true')
+        parser.add_argument('--revisions',
+                            dest='revisions',
+                            metavar='REVS', type=int, default=1,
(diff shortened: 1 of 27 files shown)


kuma/scrape/sources/base.py
```diff
@@ -201,3 +202,48 @@ def gather(self, requester, storage):
...
+
+    PARAM_NAME = 'path'
+
+    re_path = re.compile(r"/(?P<locale>[^/]+)/docs/(?P<slug>.*)")
+
+    # Standard options for Document-based sources
+    STANDARD_DOC_OPTIONS = {
+        'force': ('bool', False),         # Update existing Document records
+        'depth': ('int_all', 0),          # Scrape the topic tree to this depth
+        'revisions': ('int', 1),          # Scrape this many past revisions
```
jwhitlock: Yes, and I think that was done because scrapers that ignored ``robots.txt`` were following the "all" links.  One way forward would be to enable ``?limit=all`` for anon callers, just not add the link.

kuma/scrape/management/commands/scrape_document.py
```diff
@@ -0,0 +1,48 @@
...
+                            action='store_true',
+                            dest='translations')
+        parser.add_argument('--depth',
+                            dest='depth',
+                            metavar='DEPTH', type=self.int_all_type, default=0,
+                            help=('Depth in the topic tree to scrape'
+                                  ' ("all" for all)'))
+
+    def handle(self, *arg, **options):
+        self.setup_logging(options.get('verbosity'))
```
jwhitlock: 👍 I'm going with user's version, ``self.setup_logging(options['verbosity'])``

kuma/scrape/sources/document.py
```diff
@@ -0,0 +1,236 @@
...
+        elif zone_data.get('errors'):
+            raise self.SourceError(
+                'Unable to load zone root for %s', self.path)
+        else:
+            self.normalized_path = self.path.replace(
+                zone_data['zone_path'], zone_data['doc_path'])
+            self.locale, self.slug = self.locale_and_slug(
+                self.normalized_path)
+
+    def load_prereq_parent_topic(self, storage, data):
```
jwhitlock: 👍, and changing to fit the pattern I'm using elsewhere, to avoid too much nesting:

```
if this prerequisite not needed:
   return  # Short explanation

prereq = storage.get_prereq(...)
if prereq is None:
  data['needs'].append(('type_of_prereq', param, {})
else:
  data['key'] = extract_data(prereq)
```

PR #4253, opened 2017-06-06
Remove home.css from edit translation page

stephaniehobson:
Don't need homepage styles on translation edit page.

### DIFF
--- a/kuma/wiki/jinja2/wiki/translate.html
+++ b/kuma/wiki/jinja2/wiki/translate.html
@@ -10,7 +10,6 @@
 {% block bodyclass%}translate{% endblock %}
 
 {% block extrahead %}
-  {% stylesheet 'home' %}
 {% endblock %}
 
 {% block site_css %}


kuma/wiki/jinja2/wiki/translate.html
```diff
@@ -10,7 +10,6 @@
 {% block bodyclass%}translate{% endblock %}
 
 {% block extrahead %}
-  {% stylesheet 'home' %}
 {% endblock %}
```
schalkneethling: Do we need to keep the block around even though we are not adding any `extrahead` anymore?

schalkneethling:
One open question. Other than that, r+w/c

kuma/wiki/jinja2/wiki/translate.html
stephaniehobson: Something broke when I removed the whole block - I didn't dig into  it too deeply.

PR #4257, opened 2017-06-08
Bug 1370594: Add new brand fonts

stephaniehobson:
The goal of this PR is to swap in the new fonts and make a first approximation of the new typography. It's not meant to be a 100% match to the designs but nothing should be obviously broken.

- add zillaslab
- update font-check to pull in zilla
- remove font size from html element \o/
- update typography declarations

Palantino seems like a well supported fallback. All systems alias their Times New Roman variant to serif so we can leave Times and Times New Roman out of the font stack. 

Just using WOFF and WOFF2 [on this advice](http://davidensinger.com/2014/10/update-your-font-face-file-formats-to-just-use-woff-and-woff2/).

You will need to add the `redesign` waffle flag and enable it in the admin console.


kuma/static/styles/includes/_vars.scss
```diff
@@ -12,8 +12,8 @@ $VENDOR-PREFIXES : '-webkit-', '-moz-', '-ms-';
 /*
 colors
 ====================================================================== */
-$text-color : #3b3c40;
-$heading-color : #4d4e53;
+$text-color : #000000;
```
schalkneethling: Nit: only `#000` needed.

stephaniehobson:
Updated. 🎉

schalkneethling:
Looking really good. Wondering though, is Open San still included as because Zilla is not meant to be used for all forms of copy i.e. mainly headings?

schalkneethling:
Do we still need to support IE9?

stephaniehobson:
Yeah, Zilla is for headings and Open Sans is for body copy.

IE9 is out of support, they should get all content but not exact styles.

schalkneethling:
>Yeah, Zilla is for headings and Open Sans is for body copy.
> IE9 is out of support, they should get all content but not exact styles.

Gotcha, thanks

jinja2/base.html
```diff
@@ -1,7 +1,7 @@
 {% from "includes/common_macros.html" import optimizely_script with context -%}
 
 <!DOCTYPE html>
-<html lang="{{ LANG }}" dir="{{ DIR }}" class="no-js" data-ffo-opensans="false" data-ffo-opensanslight="false">
+<html lang="{{ LANG }}" dir="{{ DIR }}" class="no-js" data-ffo-opensans="false" data-ffo-opensanslight="false" data-ffo-zillaslab="false" data-ffo-zillahighlight="false">
```
stephaniehobson: Oops, I need to remove data-ffo-opensanslight

schalkneethling:
I clicked around pretty much all the pages I have in the sample DB and did not find any obvious breakage. IE9 is a little funky but, it is not in out support list so not blocking on that. r+ 🦄

PR #4258, opened 2017-06-08
Bug 1370594: Remove unused custom CSS classes

stephaniehobson:
CustomCSS used to be a file anyone could add anything to and it accumulated a lot of stuff. In preparation for the redesign I'm aggressively removing stuff. With a bunch of new attention on the site, we'll be notified if I removed something important. 

You can search MDN for existence of a class by editing this URL:
https://developer.mozilla.org/en-US/search?locale=*&css_classnames=card-grid

When I say only 1 page is using a class I mean only 1 English page. Non-English pages should eventually update to have the class removed.


schalkneethling:
@stephaniehobson Are there any specific pages I should have a look at or, just generally poke around? With the regards to the classes etc. removed, I am guessing just spot checking some of those should suffice?

So much cleanliness, so much goodness. ✨

stephaniehobson:
I do a search for the class using https://developer.mozilla.org/en-US/search?locale=*&css_classnames=card-grid and editing it as necessary. 

If you edit that to locale=en-US you can find the source of the class since locales just copy classes from the English pages. Results without any English results mean the class is being phased out.

I suppose you should see if you agree that the class can be removed without causing anything to look really broken? It's possible I did a bad copy-paste job when I was searching or something.

Oh, and search seems to match only the start of the string so the `landing` class also matched `landingPageList`. I got @jwhitlock to have a look through the database for `landing` and he confirms it is not in use.

kuma/static/styles/components/wiki/customcss.scss
```diff
@@ -19,227 +19,12 @@ eventually emptying this file.
...
-.wideColumnBox {
-    @include vendorize(column-width, 375px);
-}
-
-.wideColumnBoxList li {
-    line-height: 1.5;
-    list-style-type: none;
-    padding-left: 0;
-    page-break-inside: avoid;
-}
```
schalkneethling: These two are still used by:

* https://developer.mozilla.org/en-US/docs/MDN/Doc_status/JavaScript
* https://developer.mozilla.org/en-US/docs/MDN/Doc_status/CSS
* https://developer.mozilla.org/en-US/docs/MDN/Doc_status/SVG

`wideColumnBoxList` does not cause much breakage but, removing `wideColumnBox` does.

kuma/static/styles/components/wiki/customcss.scss
```diff
@@ -19,227 +19,12 @@ eventually emptying this file.
...
-
-/* Multi-column boxes for lists of longer links; this will adapt based on screen width */
-.multiColumnList {
-    @include vendorize(column-width, 250px);
-}
```
schalkneethling: This one is used by some pages for very old versions of Firefox such as:
https://developer.mozilla.org/en-US/Firefox/Releases/12

Removing the CSS rule also does not break the page badly as it only affects the  "See also" section at the bottom. Just wanted to make a note in here.

kuma/static/styles/components/wiki/customcss.scss
schalkneethling: Whoops, never mind. I see you added them to the `columnList` file above.

kuma/static/styles/components/wiki/customcss.scss
schalkneethling: Same here as above.

schalkneethling:
r+ 🔮

PR #4259, opened 2017-06-08
Fix Bug 1371448: Remove search navigator

stephaniehobson:
Search navigator show up around the page title after you follow a link from the search results page.

I don't have search configured locally so this is untested :(

Things to check:
- assets build
- nothing visually amiss on an article page
- nothing visually amiss after following a link to an article from the search results
- no javascript errors in console during all this.


kuma/settings/common.py
```diff
@@ -1,4 +1,4 @@
-# Django settings for kuma project.
+`# Django settings for kuma project.
```
jwhitlock: Extra quote should be removed in front of this comment.

kuma/settings/common.py
```diff
@@ -1591,3 +1589,4 @@ def get_user_url(user):
 ce_path = path('kuma', 'settings', 'content_experiments.json')
 with open(ce_path, 'r') as ce_file:
     CONTENT_EXPERIMENTS = json.load(ce_file)
+`
```
jwhitlock: And remove this one as well.

Or, triple quotes if you mean to comment out the whole file 😁

stephaniehobson: Weird. I wonder how that got in there...

stephaniehobson:
Fixed weird ticks.

stephaniehobson:
Remove waffle switch.

jwhitlock:
Dead code, JS, styles _and_ a waffle flag removed. 🥇

PR #4260, opened 2017-06-08
bug 1328989: Update URLs to macros in comments

jwhitlock:
Link to the macros in the kumascript repo, rather than editing paths on MDN. CustomCSS's URL is removed, because it was deleted and purged.


stephaniehobson:
👍

PR #4263, opened 2017-06-13
Bug 1370594: Move and style newsletter for redesign

stephaniehobson:
Things to test:
- [x] newsletter takes up full width of page immediately above footer
- [x] at 1200px the newsletter box becomes a 2 column layout
- [x] privacy checkbox (revealed when input gets focus) isn't broken
- [x] error message (revealed if you use the email address failure@example.com) isn't broken
- [x] success message (reveled if you use the email success@example.com) isn't broken 
- [x] other stuff you'd normally test to make sure I'm paying attention ;)
- [x] looks good on homepage still
- [x] look good on user profile edit page still

Notes:
- you will have to delete your local storage variable newsletterHide to see it again after submitting

Spec:
- form field and button adjustments are coming in later PR
- increased spacing is also coming in later PR
screen shot 2017-06-13 at 15 04 35

### DIFF
--- a/kuma/static/styles/components/newsletter.scss
+++ b/kuma/static/styles/components/newsletter.scss
@@ -1,25 +1,15 @@
-.newsletter-box {
-    position: relative;
-    @include notification-theme($default);
-    margin-bottom: $grid-spacing;
-    border: none;
-    padding: $grid-spacing;
-
-    .wiki-block + & {
-        margin-top: ($grid-spacing * 2);
-    }
-}
-
 .newsletter-teaser {
+    @include set-font-size(26px);
     margin-bottom: $grid-spacing;
-
-    .wiki-block-newsletter & {
-        @include bidi-style(padding-right, ($grid-spacing * 3), padding-left, 0); // leave space for close button
-    }
 }
 
 .newsletter-description {
-    margin-bottom: 0;
+    margin-bottom: $grid-spacing;
+    @include set-font-size(20px);
+    font-family: $heading-font-family;
+    #{$selector-heading-font-fallback} {
+        font-family: $heading-font-family-fallback;
+    }
 }
 
 .newsletter-errors .errorlist {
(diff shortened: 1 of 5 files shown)


schalkneethling:
I submitted the form using the fail address but got the following:

screen shot 2017-06-14 at 15 30 09

schalkneethling:
Also getting a success with the fail@example.com address using the old form but, here the form is replaced by the messaging instead of appending it below the form.

schalkneethling:
Oh ok, so this is overriding `.hidden`

```
@media (min-width: 1200px)
.newsletter-box .newsletter-form {
    display: grid;
    grid-column-gap: 20px;
    grid-template-columns: 2fr 1fr;
    grid-template-rows: auto auto;
}
```

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -383,6 +383,12 @@ <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
       </div>
     </div> <!-- ends "main-content" -->
 
+    {% if waffle.switch('newsletter') and waffle.switch('newsletter_article') and waffle.flag('redesign') %}
```
schalkneethling: Does it matter that I do not have the `newsletter` and `newsletter_article` flags? I do have `redesign` enabled however.

schalkneethling:
On the front page the display looks fine however, `fail@example.com` still passes through:

screen shot 2017-06-14 at 15 39 58 1

schalkneethling:
Hmmm, I do not see the newsletter on this page for example:
http://localhost:8000/en-US/profiles/sneethling@mozilla.com

This is a profile page so it should show right?

schalkneethling:
> at 1200px the newsletter box becomes a 2 column layout

You mean two rows? i.e. stacked?

schalkneethling:
In IE10 it defaults to the stacked layout irrespective of the viewport width correct?

Also, got this weird thing in IE10(perhaps a VM thing) where after I clicked submit, I got sent to https://www.mozilla.org/en-US/newsletter/

stephaniehobson:
Hmmm. Looks like this PR is missing one of my final commits :( I'll try pushing again.

stephaniehobson:
Nope, that wasn't it. I know I fixed the sucess message showing up with the form though. Need to dig around.

kuma/wiki/jinja2/wiki/document.html
stephaniehobson: These are switches not flags (different part of admin console) but, you should not be seeing the newsletter if you don't have these added and enabled. Maybe they came with your database dump?

schalkneethling: Gotcha! must be yes, as I do see the newsletter.

kuma/static/styles/components/newsletter.scss
```diff
@@ -39,16 +29,94 @@
...
+
+            .newsletter-form {
+                display: grid;
+                grid-column-gap: $grid-spacing;
+                grid-template-columns: 2fr 1fr;
+                grid-template-rows: auto auto;
+            }
+
+            &.hidden {
+                display: none;
```
stephaniehobson: Do you see this line in your local copy? It should have fixed the message appearing conflict.

stephaniehobson:
@schalkneethling I got the testing email wrong, it's failure@example.com >_< Also see my question in the comment above where I should have @ed you ^^

stephaniehobson:
Also, it's the profile *edit* page, not the profile page. My bad.

kuma/static/styles/components/newsletter.scss
schalkneethling: I do see it now, not sure if you pushed another commit. Will test and see if I still get the same behaviour.

schalkneethling: Ah ok, I see the problem now. You need to move the `&.hidden` into the `.newsletter-form` block:

```
.newsletter-form {
                display: grid;
                grid-column-gap: $grid-spacing;
                grid-template-columns: 2fr 1fr;
                grid-template-rows: auto auto;

                &.hidden {
                    display: none;
                }
            }
```

schalkneethling:
Other than my comment here https://github.com/mozilla/kuma/pull/4263/files#r122159951 ~ This is looking good~ r+w/c 🗞

kuma/wiki/jinja2/wiki/document.html
stephaniehobson: I think this is fixed now. Test again?

PR #4265, opened 2017-06-14
bug 957802: Set default KUMASCRIPT_TIMEOUT=30s

jwhitlock:
Set the constance variable ``KUMASCRIPT_TIMEOUT`` to 30 seconds, which enables KumaScript rendering by default. Set it to 0 (disabling KS rendering) for a handful of tests that call ``doc.render()`` for the side effects, such as populating the search index and rebuilding JSON.

### DIFF
--- a/kuma/search/tests/test_indexes.py
+++ b/kuma/search/tests/test_indexes.py
@@ -1,5 +1,5 @@
+from constance.test import override_config
 from django.conf import settings
-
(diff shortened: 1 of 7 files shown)


jwhitlock:
This works, but doesn't.  Too many tests rely on ``KUMASCRIPT_TIMEOUT=0``.  I'll need to add a Django setting to disable KumaScript rendering in tests.

PR #4268, opened 2017-06-14
bug 1356795: Use default PASSWORD_HASHERS

jwhitlock:
Use Django 1.8's default ``PASSWORD_HASHERS``, and drop support for custom ``Sha256Hasher``. When this has shipped to production and it is unused, the ``Sha256Hasher`` code can be removed in the next PR.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -440,15 +440,6 @@ AUTHENTICATION_BACKENDS = (
     'allauth.account.auth_backends.AuthenticationBackend',
 )
 AUTH_USER_MODEL = 'users.User'
-
-# Django 1.8 strong defaults, plus legacy Sha256Hasher
-PASSWORD_HASHERS = (
-    'django.contrib.auth.hashers.PBKDF2PasswordHasher',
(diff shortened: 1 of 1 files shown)


escattone:
Is it necessary to verify that none of the passwords stored within the production database are using ``kuma.users.backends.Sha256Hasher`` any longer, or maybe you've already verified that? I tried logging into the read-only prod copy to check this, but I'm unable to log in.

jwhitlock:
We should get your DB login working. Maybe we can check this next week.

The User password is either:

1) An empty string (anonymized DB)
2) A string starting with ``!`` (password login not allowed)
3) A string starting with ``pbkdf2`` (PBKDF2-hashed password)
4) A string starting with ``sha256`` (The Sha256Hasher)

The Django shell command I used is:

```
User.objects.exclude(password__startswith='!').exclude(password__startswith='pbkdf2').exclude(password='').count()
User.objects.filter(password__startswith='sha256').count()  # Belts and suspenders
```

The count is 0 in production, staging, and the latest sample database.

escattone:
Thanks @jwhitlock! I'm happy approving and merging this knowing that.

PR #4269, opened 2017-06-14
Bug 1370594: Remove old sticky page-button experiment

stephaniehobson:


kuma/static/styles/components/wiki/article-head.scss
```diff
@@ -1,8 +1,3 @@
-/* Set min height for crumbs and page buttons to avoid jump, when buttons are sticky */
-.article-meta {
-    min-height: 41px;
```
jwhitlock: Did you want to remove this from ``components-blue`` as well?

stephaniehobson: It'll get deleted when `components-blue` gets deleted :)

PR #4270, opened 2017-06-16
Bug 1370594: Redesign colours, icons and forms

stephaniehobson:
- add colour sass map
- default icons to 0 margin, add back where appropriate
- move icons to left side of buttons
- add button styles
- separate button styles from button context
- form field styling (excluding header, footer, search page)
- replaced hard coded colours with variables where possible

Notes:
- form submit button will change to blue in production, this isn't a problem
- adding `.neutral` to the edit button should not effect production
- some colours are not in split files and can't be updated yet: diffs, dashboards, search, wiki-syntax.

### DIFF
--- a/kuma/static/styles/includes/_vars.scss
+++ b/kuma/static/styles/includes/_vars.scss
@@ -12,29 +12,70 @@ $VENDOR-PREFIXES : '-webkit-', '-moz-', '-ms-';
 /*
 colors
 ====================================================================== */
-$text-color : #000;
-$heading-color : $text-color;
-$link-color : #217ac0;
-$menu-link-color : $text-color;
-$blue-background-text-color : #fff;
-$header-background-color : #eaeff2;
-$light-background-color : #f4f7f8;
-$homepage-background-color : #00539f;
-$border-width : 5px;
-$code-block-background-color : #fafbfc;
-$code-block-background-alt-color : #dde4e9;
-$code-block-border-color : #558abb;
-$code-block-padding : 15px;
-$code-block-border-width : $border-width;
-$code-block-border-style : solid;
-$table-blue : #d4dde4;
+$red : #e66465;
+$orange : #f69d3c;
+$yellow : #f6b73c;
+$green : #4d9f0c;
+$blue : #3f87a6;
+$purple : #9198e5;
+$grey : #696969;
 
-$red : #ea3b28;
-$yellow : #ffcb00;
-$green : #70a300;
-$blue : #0095dd;
-$grey : #c7ccce;
-$orange : #c55200;
+$red-light : #ffe7e8;
+$orange-light : #ffe8d4;
+$yellow-light : #fff3d4;
+$green-light : #ebf8e1;
+$blue-light : #e4f0f5;
+$purple-light: #d7d9f2;
+$grey-light : #eee;
+
+$red-vdark : #5d3739;
+$orange-vdark : #332206;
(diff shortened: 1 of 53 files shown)


stephaniehobson:
@schalkneethling This isn't quite ready I still need to deal with the link colours not being AA compliant on background colours. But I would appreciate a first pass at the review.

Some pages that can help:
https://developer.mozilla.org/en-US/docs/User%3Astephaniehobson%3Aindicators
https://developer.mozilla.org/en-US/docs/User%3Astephaniehobson%3Acolours
https://developer.mozilla.org/en-US/docs/User:teoli/In_content

kuma/static/styles/base/elements/typography.scss
```diff
@@ -3,6 +3,7 @@ Typography
 - headings, blockquotes, paragraphs, links, pre, code
 ********************************************************************** */
 
+
```
schalkneethling: Was the addition of this empty line intended?

kuma/static/styles/includes/_vars.scss
```diff
@@ -12,38 +12,97 @@ $VENDOR-PREFIXES : '-webkit-', '-moz-', '-ms-';
...
+$purple : #9198e5;
+$grey : #696969;
+
+$red-light : #ffe7e8;
+$orange-light : #ffe8d4;
+$yellow-light : #fff3d4;
+$green-light : #ebf8e1;
+$blue-light : #e4f0f5;
+$purple-light: #d7d9f2;
+$grey-light : #eeeeee;
```
schalkneethling: Nit: You can just use `#eee` here.

kuma/static/styles/includes/_vars.scss
```diff
@@ -12,38 +12,97 @@ $VENDOR-PREFIXES : '-webkit-', '-moz-', '-ms-';
...
+    ),
+);
+
+$negative : $red;
+$warning : $yellow;
+$positive : $green;
+$neutral : $blue;
+$default : $grey;
+
+$text-color : #000000;
```
schalkneethling: Nit: Can use `#000` here.

kuma/static/styles/zones-blue/components/landing.scss
```diff
@@ -63,6 +63,7 @@ Zone landing page styles
 
     a.button,
     button {
+        border-color: $blue-background-text-color;;
```
schalkneethling: Duplicate `;`

stephaniehobson:
@schalkneethling updated

schalkneethling:
The spacing below `h2` headers are very tight at the moment

screen shot 2017-06-19 at 11 18 09

especially when followed by for example a `h3`

screen shot 2017-06-19 at 11 16 54

schalkneethling:
I do not know if we have any control over this, but the back "highlighting" just does not look balanced to me. What do you think?

screen shot 2017-06-19 at 11 19 08

Feels to me like there should be extra padding on the sides is what I am trying to say ;)

schalkneethling:
Noticed this on the edit article page, thought I would call it out here:

screen shot 2017-06-19 at 11 23 51

stephaniehobson:
Basically everyone complains about the highlight but brand continues to insist that it looking like selected text is more important than it being readable. The weird spacing is part of the brand approved font and I can't fix it.

I'll address spacing in a separate typography focused PR.

schalkneethling:
r+ 🌈

PR #4272, opened 2017-06-16
bug 1373720: adjust DocumentNearestZoneJob cache staleness thresholds

escattone:
See [bugzilla 1373720](https://bugzilla.mozilla.org/show_bug.cgi?id=1373720).

* increase ``DocumentNearestZoneJob.refresh_timeout`` from 60 to 180 seconds
* change ``DocumentNearestZoneJob.lifetime`` from a constant to a property
  that returns a random number of days between 1 and 10 (in seconds)

### DIFF
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -1,3 +1,4 @@
+import time
 import random
 import collections
 
@@ -13,11 +14,12 @@ class DocumentNearestZoneJob(KumaJob):
     # the task has failed and allowing another to be enqueued.
     refresh_timeout = 180
 
-    @property
-    def lifetime(self):
+    def expiry(self, *args, **kwargs):
         # Spread the cache expiration times across a random
         # number of days from 1 to 10 (in units of seconds).
-        return random.randint(1, 10) * 24 * 60 * 60
+        seconds_per_day = 24 * 60 * 60
+        return time.time() + random.randint(1 * seconds_per_day,
+                                            10 * seconds_per_day)
 
     def fetch(self, pk):
         """
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -1,3 +1,4 @@
+import random
 import collections
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


kuma/wiki/tests/test_jobs.py
```diff
@@ -13,6 +13,14 @@ def test_document_zone_unicode(doc_hierarchy_with_zones):
         top_doc.get_absolute_url(), top_doc.title)
 
 
+def test_nearest_zone_lifetime():
+    """
+    Ensure that the lifetime is not constant.
+    """
+    job = DocumentNearestZoneJob()
+    assert len(set(job.lifetime for _ in range(0, 1000))) > 1
```
jwhitlock: This will fail in 1 * 10^999 cases. I guess that is acceptable. 😉 

It's also faster than any test that interacts with the database.

kuma/wiki/jobs.py
```diff
@@ -8,8 +9,15 @@
...
-    refresh_timeout = 60
+    # Allow up to three minutes to refresh the cache before assuming
+    # the task has failed and allowing another to be enqueued.
+    refresh_timeout = 180
+
+    @property
+    def lifetime(self):
+        # Spread the cache expiration times across a random
+        # number of days from 1 to 10 (in units of seconds).
+        return random.randint(1, 10) * 24 * 60 * 60
```
escattone: Good point, will do!

escattone: I had originally intended on overriding ``expiry``, but decided on this approach thinking that it only changes what we intend. I had originally went through the same logic you outline in deciding that, but now on second thought, if ever ``fetch_on_stale_threshold`` changes to something other than ``None``, the random value of ``lifetime`` becomes problematic. So I think it's safer to just override ``expiry`` so that if ``fetch_on_stale_threshold`` ever changes to something other than ``None``, a constant value for ``lifetime`` allows more control over [should_stale_item_be_fetched_synchronously](https://github.com/codeinthehole/django-cacheback/blob/master/cacheback/base.py#L384-L391) without overriding it as well.

escattone:
Thanks for your review @jwhitlock ! Should be ready for another look.

kuma/wiki/tests/test_jobs.py
```diff
@@ -13,6 +13,15 @@ def test_document_zone_unicode(doc_hierarchy_with_zones):
         top_doc.get_absolute_url(), top_doc.title)
 
 
+def test_nearest_zone_expiry():
+    """
+    Ensure that the expiry is not constant.
+    """
+    job = DocumentNearestZoneJob()
+    with mock.patch('time.time', return_value=0):
```
jwhitlock: I thought this had to be ``mock.patch('kuma.wiki.jobs.time')``, but it appears to be correct (all times are between 1 and 10 days). It is also still fast (``py.test --reuse-db --durations=10 kuma/wiki/tests/test_jobs.py``).

PR #4275, opened 2017-06-16
Bug 1370594: Change names of media queries

stephaniehobson:
- Make media query names more explicit
- shorten media query variable names
- Add wording to _vars to encourage mobile first coding

### DIFF
--- a/kuma/static/styles/includes/_vars.scss
+++ b/kuma/static/styles/includes/_vars.scss
@@ -105,44 +105,57 @@ $grid-width-column-main : -calc-col-width(9);
 breakpoints & media queries
 ====================================================================== */
 
-$max-width-default : 1400px;
+/*
+break points
+- developers understand pixel values better
+- ems is a better user experience, because design handles text resizing better
+- divide pixels by browser default font size (16px)
+- and multiply by 1em to convert unit to em.
+*/
+$max-width-default : (1400px / 16px) * 1em;
+$small-desktop-ends : (1200px / 16px) * 1em;
+$tablet-ends : (1024px / 16px) * 1em;
+$mobile-ends : (768px / 16px) * 1em;
+$small-mobile-ends : (480px / 16px) * 1em;
 
-$small-desktop-ends : 1200px;
-$tablet-ends : 1024px;
-$mobile-ends : 768px;
-$small-mobile-ends : 480px;
+$small-desktop-starts : $tablet-ends + .001em;
+$tablet-starts : $mobile-ends + .001em;
+$mobile-starts : $small-mobile-ends + .001em;
 
-$small-desktop-starts : $tablet-ends + 1px;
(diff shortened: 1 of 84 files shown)


kuma/static/styles/includes/_vars.scss
```diff
@@ -103,45 +103,50 @@ $grid-width-column-main : -calc-col-width(9);
...
-$navigation-block-ends : 850px;
-
-$navigation-break-starts : $navigation-block-ends + 1px;
-
-$media-query-navigation-break : 'all and (max-width: #{$navigation-break-ends})'; /* Moves nav items all the way to the left, moves logo above */
-$media-query-navigation-block : 'all and (max-width: #{$navigation-block-ends})'; /* Moves nav menu items to block, its own line */
-
-$media-query-break-and-up : 'all and (min-width: #{$navigation-break-starts})';
+/* break points */
+$max-width-default : (1400px / 16px) * 1em;
```
schalkneethling: Question: Should we put a note here to explain that what this is doing is converting pixel widths to it's equivalent em width? Made immediate sense to me but, maybe would not be clear to everyone. Teachable moment?

kuma/static/styles/includes/_vars.scss
```diff
@@ -103,45 +103,50 @@ $grid-width-column-main : -calc-col-width(9);
...
+$mq-tablet-only : 'all and (min-width: #{$tablet-starts}) and (max-width: #{$tablet-ends})';
+$mq-mobile-only : 'all and (min-width: #{$mobile-starts}) and (max-width: #{$mobile-ends})';
+$mq-small-mobile-only : 'all and (max-width: #{$small-mobile-ends})';
+
+$mq-navigation-break : 'all and (max-width: #{$navigation-break-ends})'; /* Moves nav items all the way to the left, moves logo above */
+$mq-navigation-block : 'all and (max-width: #{$navigation-block-ends})'; /* Moves nav menu items to block, its own line */
+
+$mq-break-and-up : 'all and (min-width: #{$navigation-break-starts})';
+
+/* avoid using these media queries, code mobile first */
```
schalkneethling: So I guess we are keeping these around for backwards compatibility and will be phasing it out?

schalkneethling:
Looks great! A couple of general question, needs a rebase but, other than that r+ ✨

kuma/static/styles/includes/_vars.scss
stephaniehobson: Yeah, a lot of them will go with the header and footer changes in the next few weeks. And more will go in July and Aug with the content restructuring. After that it might be possible to convert the remaining examples.

kuma/static/styles/includes/_vars.scss
stephaniehobson: 👍

PR #4276, opened 2017-06-16
Bug 1370594: Remove sticky TOC

stephaniehobson:
Easier to remove it from both the old and new design than to fork the wiki.js file.

Ok'ed with Kadir and communicated on the mdc mailing list.

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -308,14 +308,6 @@
         };
 
         mdn.analytics.trackLink(event, url, linkData);
-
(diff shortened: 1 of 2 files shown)


schalkneethling:
@stephaniehobson Just to confirm, what you mean by sticky here is that the toc remains expanded as you resize the window, correct? With this change, should the toc be expanded and you resize the window, the toc is collapsed aka not sticky.

If my understanding above is correct, r+

stephaniehobson:
What I mean by sticky is that it sticks to the top of the page when the user scrolls. The code for that was tangled up with the expand/collapse functionality but I think I removed all of it.

PR #4278, opened 2017-06-20
[DO NOT MERGE] Try Trusty Edge image

jwhitlock:
TravisCI is upgrading their images, and they have a preview to find compatibility issues. Hopefully there are none.  See the blog:

https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 sudo: required
+dist: trusty
+group: edge
 branches:
     only:
         - master


jwhitlock:
The build was successful, so we'll be OK when this goes live on June 21.  A few additional things:

- There's a bug in ``travis_install`` when ``INSTALL_DOCKER_COMPOSE`` is unset
- ElasticSearch 5.0.2 appears to be preinstalled.  We're transitioning to 2.4.5, so I think we still need to manually install for now.
- docker-compose 1.9.0 is installed, so we can stop manually installing 1.9.0

PR #4280, opened 2017-06-21
Fix display of KumaScript errors for document/preview endpoints

escattone:
See [bugzilla 1324092](https://bugzilla.mozilla.org/show_bug.cgi?id=1324092).

The fix was easy for this PR, but refactoring the tests proved painful. I added a new wiki-level test fixture ``ks_toolbox``. I placed it at the wiki level assuming it will be used by more than one file that we create as we refactor more of the tests within ``kuma/wiki/tests/test_views.py``. It's currently used by both ``test_views.py`` (imported, since the machinery for using fixtures within ``unittest.TestCase`` based classes is worse) and ``test_views_document.py`` .

### DIFF
--- a/kuma/wiki/jinja2/wiki/includes/kumascript_errors.html
+++ b/kuma/wiki/jinja2/wiki/includes/kumascript_errors.html
@@ -17,7 +17,7 @@
   </button>
   <div class="kserrors-details hidden">
       <p>{% trans %}
-        Anyone with an MDN profile can edit a document to fix an error. Use the <a href="/docs/MDN/Kuma/Troubleshooting_KumaScript_errors">KumaScript troubleshooting guide</a> as a starting point.
+        Anyone with an MDN profile can edit a document to fix an error. Use the <a href="/docs/MDN/Contribute/Tools/KumaScript/Troubleshooting">KumaScript troubleshooting guide</a> as a starting point.
       {% endtrans %}</p>
 
       <ul id="kserrors-list">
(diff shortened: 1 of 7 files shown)


escattone:
The Travis build has failed twice, but I can't reproduce the problem(s) locally. I restarted the build one more time, and I'll look into this tomorrow.

escattone:
OK, the test that was failing ``kuma/wiki/tests/test_views.py::DocumentEditingTests::test_translate_rebuilds_source_json`` was making multiple rendering calls to the KumaScript service, which in the py27 test env were getting connection errors (which were caught and turned into KumaScript errors) and in the Docker test env were completing without error. So the py27 test env ended up with KumaScript errors which in turn called my new code which calls KumaScript to get the macros (which raised an exception which was not caught). None of the test code relies on the rendered output, so the fix was simple: stop overriding ``constance.config.KUMASCRIPT_TIMEOUT`` with a non-zero value so that render requests do not call the KumaScript service. Thanks to @jwhitlock for help with this.

kuma/wiki/jinja2/wiki/includes/kumascript_errors.html
```diff
@@ -17,7 +17,7 @@
   </button>
   <div class="kserrors-details hidden">
       <p>{% trans %}
-        Anyone with an MDN profile can edit a document to fix an error. Use the <a href="/docs/MDN/Kuma/Troubleshooting_KumaScript_errors">KumaScript troubleshooting guide</a> as a starting point.
+        Anyone with an MDN profile can edit a document to fix an error. Use the <a href="/docs/MDN/Contribute/Tools/KumaScript/Troubleshooting">KumaScript troubleshooting guide</a> as a starting point.
```
jwhitlock: Good eye for detail on this link. While we're updating the string, let's lift the URL out of the translation, like we do for the [copyright URL](https://github.com/mozilla/kuma/blob/master/jinja2/base.html#L304-L307)

kuma/wiki/tests/conftest.py
```diff
@@ -324,3 +330,73 @@ def root_doc_with_mixed_contributors(root_doc, wiki_user, wiki_user_2,
...
+                    'filename': 'SomeMacro2.ejs'
+                }
+            ]
+        },
+        'headers': {
+            'Content-Type': 'application/json; charset=utf-8'
+        }
+    }
+
+    return KumaScriptToolbox(errors, errors_as_headers, macros_response)
```
jwhitlock: Oh, and it took me a few readings to figure out WTF was going on, but good refactor to a purest fixture 😄

kuma/wiki/tests/test_views.py
```diff
@@ -630,6 +631,11 @@ class KumascriptIntegrationTests(UserTestCase, WikiTestCase):
 
     Note that these tests really just check whether or not the service was
     used, and are not integration tests meant to exercise the real service.
+
+    REFACTOR NOTES:
+    * The "test_error_reporting" method was moved into "test_views_document.py"
+      as "test_kumascript_error_reporting" and expanded to cover the
+      "wiki.preview" endpoint as well. (escattone, June 20, 2017)
```
escattone: Oh good, I didn't really like it either! Will be happy to remove it!

kuma/wiki/jinja2/wiki/includes/kumascript_errors.html
escattone: Thanks, will do!

escattone:
Thanks for the super fast review @jwhitlock! I've addressed your change requests, and it should be ready for you final review.

PR #4284, opened 2017-06-21
bug 1370594: Split up translation in redesign notices

jwhitlock:
* Split text into smaller translation blocks with less inner HTML.
* Ensure redesign_beta and redesign_live notices aren't displayed at
the same time.


stephaniehobson:
R+

PR #4285, opened 2017-06-22
Fixes misspelling of relative in CSS rule

schalkneethling:
@stephaniehobson Saw the icons did not line up quite correctly on for example:
https://developer.allizom.org/en-US/docs/Web/HTTP

screen shot 2017-06-22 at 13 57 59

Tracked down the problem, and this fixes it.


stephaniehobson:
👍

PR #4286, opened 2017-06-22
bug 1370514 - fixed links in editing mode

karabellyj:
@jwhitlock r?


jwhitlock:
@karabellyj thanks for helping fix this!

There's a bug with ``wiki_url`` when passing a fragment like ``#The_attachments_box`` - it's adding it to the URL like ``MDN/Contribute/Editor/Basics%23The_attachments_box``, which leads to creating a new page.  This leads to creating a new page with the title "Basics%23The_attachments_box", not linking to that part of the page.  I believe the redirect was somehow masking this bug.

I'm going to try to add a follow-on commit, so that we can fix both of these issues at the same time.

PR #4291, opened 2017-06-27
Remove unused customCSS classes

stephaniehobson:
- Remove commented out class .comment
- Remove .withoutBorder
  - Tables should not be used for layout
  - I removed this from the one pages outside of FxOS that was using it
  - FxOS pages have been archived
- Remove horizontalLine, column, bottomPart, verticalText
  - horizontalLine is not in use
  - column, bottomPart, verticalText were used together and are being
  phased out
- Remove unused class blockTable
- Remove class HTMLElmNav, used in unused macro
- Remove .method, only used in archived Jetpack docs
- Remove bannerWrapper & mozOnlyBanner, in unused macro
- Improve comments for deprecatedElement

### DIFF
--- a/kuma/static/styles/components/wiki/customcss.scss
+++ b/kuma/static/styles/components/wiki/customcss.scss
@@ -19,7 +19,7 @@ eventually emptying this file.
     font-size: 60%;
 }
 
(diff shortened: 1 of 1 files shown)


schalkneethling:
r+ ✨

PR #4292, opened 2017-06-27
Bug 1370594: Remove Zilla Highlight

stephaniehobson:
We'll have to find another way to achieve this effect.


schalkneethling:
r+ 👍

PR #4293, opened 2017-06-27
Bug 1370594: Change hit area of header and footer links.

stephaniehobson:
Footer hit areas are narrower but taller, conforming more to the
shape of the link but being forgiving if you're not exactly on it.

Header hit areas are full width of column and taller.

Bug 1375765 and bug 1375962. Better but not sure they're "fixed".

### DIFF
--- a/kuma/static/styles/components/structure/nav-footer.scss
+++ b/kuma/static/styles/components/structure/nav-footer.scss
@@ -41,19 +41,18 @@
     margin-bottom: $mobile-center-spacing;
 
     a {
-        display: block;
-        margin: $content-vertical-spacing 0;
+        display: inline-block;
+        padding: ($grid-spacing / 2) 0;
     }
 }
 
 .footer-social {
     @include set-font-size($h4-font-size);
     display: inline-block;
-    @include bidi-style(margin-right, $grid-spacing, margin-left, 0);
+    @include bidi-style(margin-right, ($grid-spacing / 2 ), margin-left, 0);
 
     a {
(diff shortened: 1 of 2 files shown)


stephaniehobson:
@schalkneethling Please let me know if you think I've got the right/expected functionality now. One bug asks for `display:block` to be added and the other to have it removed, but one is in the header and the other the footer.

I don't want random clicks in the footer triggering actions, but I also want it be be easy to move the mouse directly down from the header nav options and click the associated links.

kuma/static/styles/components/structure/nav-footer.scss
```diff
@@ -98,22 +97,24 @@
...
     }
 
-
     .languages {
         @include clearfix();
         padding-bottom: $gutter-width;
         margin-bottom: $gutter-width;
+
+        select {
+            max-width: -calc-col-width(3)
```
schalkneethling: Missing `;`

schalkneethling:
Code is code except for one missing `;` There are however multiple test failures. I have looked at some of them, and they seem completely unrelated. Good to merge once those tests pass.

jwhitlock:
@schalkneethling the [other failing build](https://travis-ci.org/mozilla/kuma/jobs/247356859) is to test an aspect of Python 3 compatibility, and can be ignored by... everyone but me.  When the semicolon is fixed, the [stylelint](https://travis-ci.org/mozilla/kuma/jobs/247356857) test will pass, TravisCI will go green, and we can merge.

PR #4295, opened 2017-06-27
bug 1376606: Disable New Relic v1 API call

jwhitlock:
We'll need to update the puppet-controlled configuration settings to include an application ID. Disabling this feature for now so that pushes won't report errors.


indyragandy:
Get direct

PR #4297, opened 2017-06-30
bug 1269104: Refactor and optimize revision history

jwhitlock:
Optimize the database queries for the document's revision history page in a few ways:

* Only request columns from the document, revision, and user tables used to display the revision history, reducing the data transferred and making the calls faster.
* Iterate over pairs of revision IDs and the previous revision Ids, avoiding the ``revision.previous_revision`` work-around.
* Add a ``load_previous=True`` parameter to the Jinja helper ``format_comment``, which can be set ``False`` to avoid querying the database when ``previous_revision`` is None.

The effects of the changes are:
* Reduces the number of DB queries from 15 to 10 (plus some waffle queries if cache is cold)
* Speeds up the DB queries by 2.5x in the local environment,
* Makes the UI more consistent by adding a "Previous" link for the first translation revision
* Removes the need for the custom ``TransformManager`` and ``TransformQueryset`` for ``Revision``, used to populate the ``previous_revision`` element.


kuma/wiki/views/list.py
```diff
@@ -142,65 +143,73 @@ def revisions(request, document_slug, document_locale):
...
-                        current_revision.pk == previous_revision.pk):
-                    continue
-                # we stick to the first revision that we find
-                if previous_revision.created < current_revision.created:
-                    current_revision.previous_revision = previous_revision
-                    break
-        return revisions
+    # Load document with only fields for history display
+    try:
+        document = (Document.objects
```
safwanrahman: Is it possible to use `get_object_or_404` here?

safwanrahman:
@jwhitlock Thanks a lot for this PR. It surely needs a closer review!

kuma/wiki/views/list.py
jwhitlock: Good idea. I switched to ``get_object_or_404`` with a longer queryset, and now ``Http404`` is no longer imported.

kuma/wiki/views/list.py
```diff
@@ -142,65 +142,70 @@ def revisions(request, document_slug, document_locale):
...
-                if previous_revision.created < current_revision.created:
-                    current_revision.previous_revision = previous_revision
-                    break
-        return revisions
 
-    per_page = request.GET.get('limit', 10)
+    # Load document with only fields for history display
+    doc_query = (Document.objects
+                 .only('id', 'locale', 'slug', 'title',
+                       'parent__slug', 'parent__locale')
```
jwhitlock: Oops, got too excited about the changes in 960d66f68f8f1b58464084985ac7d9f3c495c560...

kuma/wiki/templatetags/jinja_helpers.py
```diff
@@ -41,12 +41,17 @@ def bugize_text(content):
...
     """
-    Massages revision comment content after the fact
+    Format comment for HTML display, with Bugzilla links and slug changes.
+
+    Keyword Arguments:
+    rev - The revision
+    previous_revision - The previous revision (default None)
+    load_previous - Try loading previous revision if None (default True)
     """
     prev_rev = getattr(rev, 'previous_revision', previous_revision)
```
jwhitlock: Good idea! More code to delete! 🎉

PR #4299, opened 2017-07-04
Bug 1370594: Typography tweaks

stephaniehobson:
- Update to version of zilla with &nbsp; character in it
- Make kbd tag fit in line better
- Lighten default text color to decrease eye strain
- Remove underline on hover from a[name] links


schalkneethling:
r+

PR #4300, opened 2017-07-07
Bug 1370594: Recreate Zilla Highlight effect

stephaniehobson:
We couldn't use Zilla Highlight because of locale and code problems. Recreating the effect using JavaScript to insert a decorative element. Fallsback to highlighting the entire line.

Fix Bug 1375871 and partially address the following bugs:
Bug 1375831
Bug 1376072
Bug 1379104
Bug 1379259


kuma/static/styles/includes/_mixins.scss
```diff
@@ -527,6 +527,33 @@ Placeholders
     @include vendorize(hyphens, none);
 }
 
+%highlight {
+
+    &.highlight-spanned { // needs to preceed .hilight-span definition or links aren't coloured
```
schalkneethling: Typo s/hilight/highlight

kuma/static/js/highlight.js
```diff
@@ -0,0 +1,16 @@
...
+        $targets.each(function() {
+            $(this).addClass('highlight-spanned').wrapInner('<span class="highlight-span"></span>');
+        });
+    }
+
+    // call on aritcle body with h3 and h4 as targets
+    var $articleSubHeads = $('#wikiArticle h3, #wikiArticle h4');
+    highlight($articleSubHeads);
+
+})(window, document, jQuery);
```
schalkneethling: I am assuming you chose the JS route because adding a class to all existing `h3` and `h4` tags site wide is going to be next to impossible? So, in this regard it is a progressive enhancement i.e. no-JS === no-Highlight

kuma/static/js/highlight.js
```diff
@@ -0,0 +1,16 @@
+(function(win, doc, $) {
```
schalkneethling: Nit: As `win` and `doc` is not used inside the code block, it is safe to only pass `jQuery` an hence, receive only `$`. Currently is is probably triggering an `eslint` warning about `no-unused-vars`

schalkneethling:
Couple of nits but other than that, looking great r+w/c 👍

kuma/static/js/highlight.js
stephaniehobson: Both js and no-js should have a black background and white text. The difference between the js and no-js experience is that once the classes have been applied the black background will not extend to the end of the line.

With JS:

screen shot 2017-07-10 at 10 27 41

Without JS:

screen shot 2017-07-10 at 10 27 53

If you're not seeing the JS version properly you probably have to do a `docker-compose restart web` since changes to `common.py` aren't handled without a restart.

PR #4302, opened 2017-07-07
Bug 1370594: Rename and optimize Web Docs Logo sprite

stephaniehobson:


stephaniehobson:
@schalkneethling Added height and width like you suggested, hopefully this fixes the IE problems.

kuma/static/img/web-docs-sprite.svg
```diff
@@ -0,0 +1,9 @@
...
+</svg>
```
schalkneethling: Missing `eol` also, worth running this through SVGOMG. We should probably make SVGO part of the build process as we have more and more SVG assets.

stephaniehobson: I did run it through SVGOMG.

stephaniehobson: I don't anticipate adding more than 3 or 4 more SVGs and we only add one or two images a year when we're not redesigning, I'm not sure it's worth updating the build process.

stephaniehobson: Added the `eol`

schalkneethling:
The SVG is all good, not sure about IE yet. The main reason that I cannot confirm whether it is fixed or not, is because the entire masthead is broken with this PR. I am guessing there is an order in which this and some of the other pull requests needs to be merged?

screen shot 2017-07-11 at 09 04 54

If that is expected r+

stephaniehobson:
Nope, that's definitely wrong :(

stephaniehobson:
Hm. I suspect you still have the CSS from `1370594-home` but the correct HTML from this branch.

schalkneethling:
> Hm. I suspect you still have the CSS from 1370594-home but the correct HTML from this branch.

Indeed, how weird ¯\_(ツ)_/¯

Fixed in IE, all good r+

PR #4303, opened 2017-07-07
bug 1379288: Include optional locale-specific CSS

jwhitlock:
Include a locale-specific CSS file if it has been defined in the pipeline. Empty examples for ru and zh-CN are included, as well as helper methods for making it easier to declare these in the settings file.

I install node 8 locally, and I'm unable to get gulp working again. I suspect it should still work with gulp, but I'm not sure.  I used ``make collectstatic`` inside the container. It may need to be run twice.


stephaniehobson:
- The `ru.css` file is being included on the home page. 
- There is no `ko.css` file and it is correctly not included.
- gulp is watching the files but not moving them to where pipeline needs them, the gulp styles task needs to be updated like this:

```
// compiles top-level .scss files
gulp.task('styles', () => {
    // only process files in /styles root
    // all other .scss files are components/includes/libs
    gulp.src('./kuma/static/styles/*.scss')
        .pipe(sass().on('error', sass.logError))
        // send compiled files to where expected by Django Pipeline
        .pipe(gulp.dest('./static/styles'));
    // except locale files, compile and move those too
    gulp.src('./kuma/static/styles/locales/*.scss')
        .pipe(sass().on('error', sass.logError))
        // send compiled files to where expected by Django Pipeline
        .pipe(gulp.dest('./static/styles/locales'));
});
```

Collect static did not work for me but build-static did.

jwhitlock:
Do you want to add the gulp commit for credit, or do you want me to add it?

I think running ``make collectstatic`` twice would have worked as well. ``make build-static`` runs ``make compilejsi18n`` then ``make collectstatic``. The first call copies the file to ``static/styles/locales/ru.scss``, and the second builds it to ``static/styles/locales/ru.css``. Or something like that. Needs fixing in a future PR.

stephaniehobson:
I ran `docker-compose exec web ./manage.py collectstatic` twice. Is that different from `make collectstatic` ? Should we be worried running it twice didn't work for me? Will needing to run it twice cause problems when this goes to stage?

I don't need the credit for the gulp change.

jwhitlock:
No, your install is not broken.  There's difference when running ``./manage.py collectstatic`` and ``make collectstatic``. It is too dumb to get into.

gulp file updated.

jwhitlock:
Rebased, removed unused future code setting ``variant``, to make code coverage happier.

stephaniehobson:
Gulp is handling changes now :)

I think this is good to go but we could also remove one of the two files since they are empty.

jwhitlock:
I dropped the Russian one, because the ``zh-CN.scss`` shows how to handle locales with hyphens.

PR #4305, opened 2017-07-10
Adds post message listener for interactive examples

schalkneethling:
Adds a `message` event listener for `postMessage` messages from the interactive code examples.

The `event` object passed to the `interactiveExamplesEvent` function will contain the following:

`event.origin` ~ Origin of the window that sent the message (here it is the iframe)
`event.data` ~ The data sent via `postMessage`. This will contain a category, action and label associated with the event that is then passed to `trackEvent`

For example:

```
{
  category: 'js-examples',
  action: 'Clicked run',
  label: 'Interaction Events'
}
```

### DIFF
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -1,7 +1,6 @@
 (function(win, doc, $) {
     'use strict';
 
-    // prettier-ignore
     // Adding to globally available mdn object
     var analytics = mdn.analytics = {
         /**
@@ -42,8 +41,8 @@
             var ga = win.ga;
             var data = {
                 hitType: 'event',
-                eventCategory: eventObject.category || '', // Required.
-                eventAction: eventObject.action || '', // Required.
+                eventCategory: eventObject.category || '',    // Required.
+                eventAction: eventObject.action || '',             // Required.
                 eventLabel: eventObject.label || '',
                 eventValue: eventObject.value || 0,
                 hitCallback: callback || null
@@ -51,20 +50,22 @@
 
             // If Analytics has loaded, go ahead with tracking
             // Checking for ".create" due to Ghostery mocking of ga
-            if (ga && ga.create) {
+            if(ga && ga.create) {
                 // Send event to GA
                 ga('send', data);
-            } else if (ga && !ga.create) {
+            }
+            else if(ga && !ga.create) {
                 // GA blocked or not yet initialized
                 // strip callback from data
                 data.hitCallback = null;
                 // add to queue without callback
                 ga('send', data);
                 // execute callback now
-                if (callback) {
+                if(callback) {
                     callback();
                 }
-            } else if (callback) {
+            }
+            else if(callback) {
                 // GA disabled or blocked or something, make sure we still
                 // call the caller's callback:
                 callback();
@@ -77,7 +78,7 @@
         trackOutboundLinks: function(target) {
             target = target || document.body;
 
-            $(target).on('click', 'a', function(e) {
+            $(target).on('click', 'a', function (e) {
                 var $this = $(this);
 
                 // bug 1222864 - prevent links to data: uris
@@ -88,14 +89,13 @@
                 }
 
                 // If we explicitly say not to track something, don't
-                if ($this.hasClass('no-track')) {
+                if($this.hasClass('no-track')) {
                     return;
                 }
 
                 var host = this.hostname;
-                if (host && host !== location.hostname) {
-                    var newTab =
-                        this.target === '_blank' || e.metaKey || e.ctrlKey;
+                if(host && host !== location.hostname) {
+                    var newTab = (this.target === '_blank' || e.metaKey || e.ctrlKey);
                     var href = this.href;
                     var callback = function() {
                         win.location = href;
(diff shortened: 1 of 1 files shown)


kuma/static/js/analytics.js
```diff
@@ -1,8 +1,20 @@
 (function(win, doc, $) {
     'use strict';
 
+    // prettier-ignore
```
stephaniehobson: What does this do?

kuma/static/js/analytics.js
```diff
@@ -1,8 +1,20 @@
...
+        /**
+         * Handles postMessage events from the interactive editor, passing of
+         * the data to `trackEvent` if the origin of the message is what we expect.
+         * @param {Object} event - The event Object received from the postMessage
+         */
+        interactiveExamplesEvent: function(event) {
+            if (event.origin !== 'https://mdn.github.io') {
+                return false;
+            }
+            this.trackEvent(event.data);
```
stephaniehobson: Please add a check to ensure that event.data does not contain a callback.

kuma/static/js/analytics.js
```diff
@@ -107,16 +118,16 @@
 
         trackLink: function(event, url, data) {
             // ctrl or cmd click or context menu
-            var newTab = (event.metaKey || event.ctrlKey || event.type == 'contextmenu');
+            var newTab =
```
stephaniehobson: It looks weird to have this on a line by itself. Can we solve the line length problem a different way?

kuma/static/js/analytics.js
```diff
@@ -129,30 +140,40 @@
...
-                    analytics.trackError('JavaScript Error', originalEvent.message , originalEvent.filename + ':' + lineAndColumnInfo);
-                }
-                // no originalEvent means probably jQuery
-                else {
+                    var lineAndColumnInfo = originalEvent.colno
+                        ? ' line:' +
+                          originalEvent.lineno +
+                          ', column:' +
+                          originalEvent.colno
+                        : ' line:' + originalEvent.lineno;
```
stephaniehobson: This is unreadable spread over multiple lines like this. I think is just too long to go into a ternary function.

kuma/static/js/analytics.js
schalkneethling: So, this stops the prettier Atom plugin from formatting that specific node of the code. Ideally we would not do `var analytics = mdn.analytics = {....` but instead, simply have

```
var analytics = {
  ...
};

// then here do
mdn.analytics = analytics;
```
Because of that, Prettier formats that line something like:

```
var analytics = (mdn.analytics = {
 ...
});
```

I however did not want to mess with the code to much as part of this PR, so I just added that line so Prettier does not format that part of the code.

kuma/static/js/analytics.js
schalkneethling: So, this is the `event` object that is send from `postMessage` and will always only have the following properties:

* data - The data I send back i.e. category, action, label
* origin - Origin of the window that sent the message i.e. the iframe
* source - A reference to the window object of the origin above

One can of course loop over the properties of the `event.data` object and ensure that none of the values are a function. What do you think?

kuma/static/js/analytics.js
schalkneethling: Yeah, it is because of the line length that prettier is doing that formatting. I will have prettier ignore this line.

kuma/static/js/analytics.js
schalkneethling: Agree, probably why the Prettier formatting makes it look extra weird. Kinda hinting that it needs a bit of a refactor. I will do that.

schalkneethling:
> Does the interactive editor send these messages yet?

The only thing I am tracking at the moment are clicks on the "Run" button for the JS editor.

kuma/static/js/analytics.js
stephaniehobson: I am not comfortable committing code that is specific to one contributor's current editor set up :/

schalkneethling: We could make it a pre-commit hook. But that is a fair comment. Let me revert those changes and simply add the `postMessage` additions.

kuma/static/js/analytics.js
stephaniehobson: I had a closer look at the MDN code and I think this is okay after all.

I though the MDN code was like this

```
var data = {
                hitType: 'event',
                eventCategory: eventObject.category || '',    // Required.
                eventAction: eventObject.action || '',             // Required.
                eventLabel: eventObject.label || '',
                eventValue: eventObject.value || 0,
                hitCallback: eventObject.callback || null
            };
```

but it's like like this:

```
var data = {
                hitType: 'event',
                eventCategory: eventObject.category || '',    // Required.
                eventAction: eventObject.action || '',             // Required.
                eventLabel: eventObject.label || '',
                eventValue: eventObject.value || 0,
                hitCallback: callback || null
            };
```

stephaniehobson:
I didn't see the `message` code in either of the live samples on MDN, are they sending the message and I just missed it?

schalkneethling:
> I didn't see the message code in either of the live samples on MDN, are they sending the message and I just missed it?

This may very well be because of the `origin` check.

This on the editor side[1], and this on the Kuma side[2]

[1] https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/analytics.js#L8
[2] https://github.com/mozilla/kuma/pull/4305/files#diff-94c143958fe4718bffbe73d2b02429a9R12

stephaniehobson:
R+

PR #4307, opened 2017-07-10
bug 1271509: Beta style in CKEditor

jwhitlock:
Use the design style in CKEditor, if the user is in the beta program.


jwhitlock:
oops - wrong bug

PR #4313, opened 2017-07-18
Bug 1362025: Add classes to resize interactive examples

stephaniehobson:
Resizes the interactive example iframe based on classes passed through to the macro.

Macro changes are in: https://github.com/mozilla/kumascript/pull/239 (but this PR is not dependent on that)


schalkneethling:
r+ 🐚

PR #4317, opened 2017-07-19
Bug 1375892: Locale font fallbacks

stephaniehobson:
**The way the locale font fallbacks work is:**
- The font stack includes made up keywords (`x-locale-body`,
  `x-locale-heading-primary`, `x-locale-heading-secondary`)
- When the browser doesn't recognize the keywords it falls back
  to the next font in the stack
- We can load CSS files specific to each locale. In those files we can
  include an `@font-face` declaration which defines the keywords. This
  causes the defined font to be loaded and displayed.
- Best practice for Asian character sets (ie Chinese) is to load a
  preferred display font and then specify a better fallback for the
  Chinese. This is because Chinese fonts have English characters but
  they look terrible.

**Changes in this PR:**
- Removed the @font-face declaration for Zilla from the main.css file.
- Added made up keyword `x-locale-heading-primary` to the front of the
  heading font stack.
- Added made up keywords `x-locale-heading-secondary`, and
  `x-locale-body` to the font stack as fallback fonts
- Added locale specific files for each locale which Zilla supports.
  These files each include the Zilla @font-face declaration which
  enables Zilla to load into its place in the font stack.
- Because font paths are relative the SCSS variables are complicated.
  To handle loading Zilla from the locale files I had to move the
  Zilla files into the locale folder and the _zilla.scss file in with
  the other locale style sheets.
- Copied locale specific fixes for Open Sans from bedrock to kuma
  using the x-locale-body keyword to load in better fallbacks.
    - Did not copy font-variant fixes since we don't use font variants.
- Declared new heading defaults for locales with Arabic alphabet by
  defining a font-face for `x-locale-heading-primary` in those locale
  specific files.

**Testing notes:**
- You'll need to do a `docker-compose down` and `docker-compose up` for kuma to pick up on the existence of the new files
- If locale specific CSS files are not being included on pages after that you can try running in "production" asset mode:
  1. Create a file called `.env` in the root kuma folder
  1. Inside it write `DEBUG=False`
  1. `docker-compose up -d`
  1. `docker-compose exec web ./manage.py collectstatic`
  1. `docker-compose exec web ./manage.py collectstatic` (yes [do it twice](https://github.com/mozilla/kuma/pull/4303#issue-241394205))
  1. `docker-compose restart web`
- I'm not really sure what to test beyond that the files load when they're present. It's hard for us to say if we've got the right Japanese fallbacks defined. I think the best thing to do is to get it to our beta testers as quickly as possible.


schalkneethling:
@stephaniehobson @escattone I have loaded a couple of the different locales, and from what I can see, the fonts load as expected.

`ar`

screen shot 2017-07-20 at 14 01 13

`fa`

screen shot 2017-07-20 at 14 06 15

`ja`

screen shot 2017-07-20 at 14 08 00

`ko`

screen shot 2017-07-20 at 14 17 38

I also clicked around of few of the key areas, and could not find anything obvious. There are some strings that are not translated but, I guess those are to be expected.

@escattone if all looks good to you as well, let's merge this. 🎏

PR #4320, opened 2017-07-21
bug 1316610: temporarily remove nofollow from user-profile pages

escattone:
This is my understanding of the first step to removing user-profile pages from search indices. Is this all there is to the first step?

### DIFF
--- a/kuma/users/jinja2/users/user_detail.html
+++ b/kuma/users/jinja2/users/user_detail.html
@@ -8,7 +8,7 @@
 {% block bodyclass %}user user-display{% endblock %}
 
 {% block title %}{{ page_title(detail_user) }}{% endblock %}
-{% block robots_value %}noindex, nofollow{% endblock %}
+{% block robots_value %}noindex{% endblock %}
 
 {% block site_css %}
   {{ super() }}


jwhitlock:
This seems wrong to me. That directive will tell Google to follow links that appear on the profile page. I thought the change would be to allow allow following links **to** profile pages, such as the contributor bar or search results.  I think this change would still prevent Google from crawling profile pages.  I've asked @a2sheppy to see if there's a more direct recommendation from the SEO expert.

escattone:
Thanks @jwhitlock, this didn't make sense to me as well.

a2sheppy:
The exact text from the consultant:

> Remove the nofollow tag from profiles pages to allow search engines to read the noindex tag in the header. After four weeks, add the nofollow tag back to the header and subsequently append the nofollow tag to each profile link. Currently, the noindex tag cannot be read by search engines because of the concurrent nofollow directive.

jwhitlock:
Ok. SEO magic.

stephaniehobson:
The text from the consultant sounds unclear to me. I think we should be removing the `nofollow` tag from *links* to the profile pages so the crawler can reach those pages and see the `noindex`.  Whether or not it can follow the links on the profile page should not affect this.

jwhitlock:
After sleeping on this, and re-reading the [google search docs](https://developers.google.com/search/reference/robots_meta_tag), the recommendation still makes no sense to me.  I think I understand the requirement, "get profile pages out of the Google index", and I can't see how this will help in any way.

My counter proposal:

1. Add a canonical tag to profile pages, since the same page appears in every locale. I'm going to pick the user's set locale as the canonical version, and not list all the translations like we do for wiki docs
2. Create a sitemap of the profile pages. This will allow us to directly tell Google "go crawl these pages please", rather than wait for following contributor links or periodic page refreshes.
3. Submit the sitemap to Google, and watch profile pages drop out of the index

I'd like to tackle the wiki pages sitemap first, so I can get familiar with the Django sitemap implementation again, before attempting this one.

jwhitlock:
Also, I can't see that we ever had ``nofollow`` on links to profile pages. We certainly don't at the moment.

a2sheppy:
That's weird. Let me check with Nilaye and see what he figures out here.

PR #4321, opened 2017-07-21
[Bug 1331729] Language preference cookie to stop site redirection

safwanrahman:
**You need to clear your browser cache or using private browsing before testing this**

As the past code was redirecting permanently with http `301` status code, It should be return with temporary redirect with http `302`

It was fixed in kitsune long ago by James, But was not ported back to kuma.

The Implementation is based on kitsune, refactored some part of the code to increase readability. 

@jwhitlock @escattone r?

Need to be done:
- [x] Pass all the unittests
- [x] Write tests for the functionality

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -4,11 +4,7 @@ from urlparse import urljoin
 
 from django.conf import settings
 from django.core import urlresolvers
-from django.http import (
-    HttpResponseRedirect,
-    HttpResponseForbidden,
-    HttpResponsePermanentRedirect,
-)
+from django.http import HttpResponseForbidden, HttpResponseRedirect, HttpResponsePermanentRedirect
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
 from django.contrib.sessions.middleware import SessionMiddleware
@@ -33,8 +29,9 @@ class LocaleURLMiddleware(object):
         prefixer = Prefixer(request)
         set_url_prefixer(prefixer)
         full_path = prefixer.fix(prefixer.shortened_path)
+        lang = request.GET.get('lang')
 
-        if 'lang' in request.GET:
+        if lang in dict(settings.LANGUAGES):
             # Blank out the locale so that we can set a new one. Remove lang
             # from the query params so we don't have an infinite loop.
             prefixer.locale = ''
--- a/kuma/core/urlresolvers.py
+++ b/kuma/core/urlresolvers.py
@@ -111,9 +111,10 @@ class Prefixer(object):
     def __init__(self, request=None, locale=None):
         """If request is omitted, fall back to a default locale."""
         self.request = request or RequestFactory(REQUEST_METHOD='bogus').request()
-        self.locale, self.shortened_path = split_path(self.request.path_info)
-        if locale:
-            self.locale = locale
+        path_locale, self.shortened_path = split_path(self.request.path_info)
+
+        # Set Self locale according to priority.
+        self.locale = locale or path_locale or self.get_chosen_language() or ""
 
     def get_language(self):
         """
@@ -134,6 +135,14 @@ class Prefixer(object):
 
         return settings.LANGUAGE_CODE
 
+    def get_chosen_language(self):
+        """If the request has a cookie set for language, return that language."""
+        language = self.request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+
+        # Need to check if his cookie language is supported by Kuma.
+        if language and language in dict(settings.LANGUAGES):
+            return language
+
     def fix(self, path):
         path = path.lstrip('/')
         url_parts = [self.request.META['SCRIPT_NAME']]
@@ -142,7 +151,8 @@ class Prefixer(object):
         else:
             check_path = path + '/'
         if not check_path.startswith(settings.LANGUAGE_URL_IGNORED_PATHS):
-            locale = self.locale if self.locale else self.get_language()
+            # Set locale according to order
+            locale = self.locale or self.get_language()
             url_parts.append(locale)
 
         url_parts.append(path)
--- a/kuma/core/views.py
+++ b/kuma/core/views.py
@@ -1,10 +1,33 @@
+from django.conf import settings
+from django.http import HttpResponse
 from django.shortcuts import render
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST
+
 from .decorators import never_cache
 
 
 def _error_page(request, status):
     """Render error pages with jinja2."""
     return render(request, '%d.html' % status, status=status)
+
+
+@csrf_exempt
+@require_POST
+def set_language(request):
+    lang_code = request.POST.get("language")
+    response = HttpResponse(status=204)
+
+    if lang_code and lang_code in dict(settings.LANGUAGES):
+
+        response.set_cookie(key=settings.LANGUAGE_COOKIE_NAME,
+                            value=lang_code,
+                            max_age=settings.LANGUAGE_COOKIE_AGE,
+                            path=settings.LANGUAGE_COOKIE_PATH,
+                            domain=settings.LANGUAGE_COOKIE_DOMAIN,
+                            )
+
+    return response
 
 
 handler403 = lambda request: _error_page(request, 403)
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -420,7 +420,8 @@ LANGUAGE_URL_IGNORED_PATHS = (
     'index.php',  # Legacy MediaWiki endpoint, return 404
     # Served in AWS
     'sitemap.xml',
-    'sitemaps/'
+    'sitemaps/',
+    'i18n'
 )
 
 # Make this unique, and don't share it with anybody.
@@ -988,6 +989,7 @@ PIPELINE_JS = {
             # FFO lib contains Promise polyfill
             'js/libs/fontfaceobserver/fontfaceobserver.2.0.7.js',
             'js/fonts.js',
+            'js/lang-switcher.js'
         ),
         'output_filename': 'build/js/main.js',
     },
--- a/kuma/static/js/lang-switcher.js
+++ b/kuma/static/js/lang-switcher.js
@@ -1,5 +1,7 @@
 (function (win, doc, $) {
     var sessionStorageKey = 'changed-locale-to';
+    var neverShowNoticeKey = 'never-show-locale-notice';
+    var neverShowNotice = getNeverShowNotice();
     function storeLocaleChange(code, name) {
         sessionStorage.setItem(sessionStorageKey, JSON.stringify({code: code, name: name}));
     }
@@ -12,7 +14,15 @@
         return sessionStorage.getItem(sessionStorageKey);
     }
 
-    if(win.sessionStorage) {
+    function storeNeverShowNotice() {
+        localStorage.setItem(neverShowNoticeKey, true);
+    }
+
+    function getNeverShowNotice() {
+        return localStorage.getItem(neverShowNoticeKey) || false;
+    }
+
+    if(win.sessionStorage && win.mdn.features.localStorage && !neverShowNotice) {
         var langSwitcherSelector = document.getElementById('language');
         var langSwitcherButton = document.getElementById('translations');
 
@@ -39,11 +49,14 @@
             var locale = JSON.parse(changedLocaleTo);
             var text = gettext('You are now viewing this site in %(localeName)s.' +
                                ' Do you always want to view this site in %(localeName)s?');
-            var html = interpolate(text +
-                    '<br><button id="locale-permanent-yes" type="button" data-locale="%(localeCode)s">' +
-                    gettext('Yes') + '</button> <button id="locale-permanent-no" type="button">' + gettext('No') +
-                    '</button></p></div>', {localeCode: locale.code, localeName: locale.name}, true);
+            var message = interpolate(text, {localeName: locale.name}, true);
+            var button = '<button id="%s" type="button" data-locale="' + locale.code + '">%s</button> ';
+            var yesButton = interpolate(button, ['locale-permanent-yes', gettext('Yes')]);
+            var noButton = interpolate(button, ['locale-permanent-no', gettext('No')]);
+            var neverButton = interpolate(button, ['locale-permanent-never', gettext('Never')]);
+            var html = message + '<br>' + yesButton + noButton + neverButton;
             var notification = mdn.Notifier.growl(html, {closable: true, duration: 0});
+
             notification.question();
             removeLocaleChange();
 
@@ -56,6 +69,11 @@
             });
 
             $('#locale-permanent-no').on('click', function() {
+                notification.close();
+            });
+
+            $('#locale-permanent-never').on('click', function() {
+                storeNeverShowNotice();
                 notification.close();
             });
         }
--- a/kuma/static/js/lang-switcher.js
+++ b/kuma/static/js/lang-switcher.js
@@ -1,0 +1,27 @@
+(function () {
+    var langSwitcherSelector = document.getElementById('language'),
+        langSwitcherButton = document.getElementById('translations');
+
+    function storeLocaleChange(code, name) {
+        localStorage.setItem('changed-locale-to', JSON.stringify({code: code, name: name}));
+    }
+
+    if (langSwitcherSelector) {
+        langSwitcherSelector.addEventListener('change', function () {
+            var element = this.options[this.options.selectedIndex];
+            storeLocaleChange(element.dataset.locale, element.label);
+        });
+    }
+
+    if (langSwitcherButton) {
+        var transChoices = langSwitcherButton.querySelectorAll('li a');
+
+        for(var i = 0; i < transChoices.length; i++) {
+
+            transChoices[i].addEventListener('click', function () {
+                storeLocaleChange(this.dataset.locale, this.text);
+            });
+        }
+    }
+
+})();
--- a/kuma/static/js/lang-switcher.js
+++ b/kuma/static/js/lang-switcher.js
@@ -1,14 +1,23 @@
 (function (win, doc, $) {
+    var sessionStorageKey = 'changed-locale-to';
     function storeLocaleChange(code, name) {
-        localStorage.setItem('changed-locale-to', JSON.stringify({code: code, name: name}));
+        sessionStorage.setItem(sessionStorageKey, JSON.stringify({code: code, name: name}));
     }
 
-    if(win.mdn.features.localStorage) {
-        var langSwitcherSelector = document.getElementById('language'),
-            langSwitcherButton = document.getElementById('translations');
+    function removeLocaleChange() {
+        sessionStorage.removeItem(sessionStorageKey);
+    }
+
+    function getLocaleChange() {
+        return sessionStorage.getItem(sessionStorageKey);
+    }
+
+    if(win.sessionStorage) {
+        var langSwitcherSelector = document.getElementById('language');
+        var langSwitcherButton = document.getElementById('translations');
 
         if (langSwitcherSelector) {
-            langSwitcherSelector.addEventListener('change', function () {
+            langSwitcherSelector.addEventListener('change', function() {
                 var element = this.options[this.options.selectedIndex];
                 storeLocaleChange(element.dataset.locale, element.label);
             });
@@ -18,41 +27,37 @@
             var transChoices = langSwitcherButton.querySelectorAll('li a');
 
             for (var i = 0; i < transChoices.length; i++) {
-
-                transChoices[i].addEventListener('click', function () {
+                transChoices[i].addEventListener('click', function() {
                     storeLocaleChange(this.dataset.locale, this.text);
                 });
             }
         }
 
         // Insert notice about permanent language switch
-        var changedLocaleTo = localStorage.getItem('changed-locale-to');
+        var changedLocaleTo = getLocaleChange();
         if (changedLocaleTo) {
-            var locale = JSON.parse(changedLocaleTo),
-                text = gettext('You are now viewing this site in %(localeName)s.' +
-                               ' Do you always want to view this site in %(localeName)s?'),
-                html = interpolate(text +
+            var locale = JSON.parse(changedLocaleTo);
+            var text = gettext('You are now viewing this site in %(localeName)s.' +
+                               ' Do you always want to view this site in %(localeName)s?');
+            var html = interpolate(text +
                     '<br><button id="locale-permanent-yes" type="button" data-locale="%(localeCode)s">' +
                     gettext('Yes') + '</button> <button id="locale-permanent-no" type="button">' + gettext('No') +
-                    '</button></p></div>', {localeCode: locale.code, localeName: locale.name}, true),
-                notification = mdn.Notifier.growl(html, {closable: true, duration: 0}).question();
+                    '</button></p></div>', {localeCode: locale.code, localeName: locale.name}, true);
+            var notification = mdn.Notifier.growl(html, {closable: true, duration: 0});
+            notification.question();
+            removeLocaleChange();
 
             // Add event listener to the buttons
-            $('#locale-permanent-yes').on('click', function () {
+            $('#locale-permanent-yes').on('click', function() {
                 $.post('/i18n/setlang/', {language: this.dataset.locale})
-                    .success(function () {
+                    .success(function() {
                         notification.close();
-                        localStorage.removeItem('changed-locale-to');
                     });
-            }
-            );
+            });
 
-            $('#locale-permanent-no').on('click', function () {
+            $('#locale-permanent-no').on('click', function() {
                 notification.close();
-                localStorage.removeItem('changed-locale-to');
-            }
-            );
+            });
         }
     }
-
 })(window, document, jQuery);
--- a/kuma/static/js/lang-switcher.js
+++ b/kuma/static/js/lang-switcher.js
@@ -1,27 +1,58 @@
-(function () {
-    var langSwitcherSelector = document.getElementById('language'),
-        langSwitcherButton = document.getElementById('translations');
-
+(function (win, doc, $) {
     function storeLocaleChange(code, name) {
         localStorage.setItem('changed-locale-to', JSON.stringify({code: code, name: name}));
     }
 
-    if (langSwitcherSelector) {
-        langSwitcherSelector.addEventListener('change', function () {
-            var element = this.options[this.options.selectedIndex];
-            storeLocaleChange(element.dataset.locale, element.label);
-        });
-    }
+    if(win.mdn.features.localStorage) {
+        var langSwitcherSelector = document.getElementById('language'),
+            langSwitcherButton = document.getElementById('translations');
 
-    if (langSwitcherButton) {
-        var transChoices = langSwitcherButton.querySelectorAll('li a');
-
-        for(var i = 0; i < transChoices.length; i++) {
-
-            transChoices[i].addEventListener('click', function () {
-                storeLocaleChange(this.dataset.locale, this.text);
+        if (langSwitcherSelector) {
+            langSwitcherSelector.addEventListener('change', function () {
+                var element = this.options[this.options.selectedIndex];
+                storeLocaleChange(element.dataset.locale, element.label);
             });
+        }
+
+        if (langSwitcherButton) {
+            var transChoices = langSwitcherButton.querySelectorAll('li a');
+
+            for (var i = 0; i < transChoices.length; i++) {
+
+                transChoices[i].addEventListener('click', function () {
+                    storeLocaleChange(this.dataset.locale, this.text);
+                });
+            }
+        }
+
+        // Insert notice about permanent language switch
+        var changedLocaleTo = localStorage.getItem('changed-locale-to');
+        if (changedLocaleTo) {
+            var locale = JSON.parse(changedLocaleTo),
+                text = gettext('You are now viewing this site in %(localeName)s.' +
+                               ' Do you always want to view this site in %(localeName)s?'),
+                html = interpolate(text +
+                    '<br><button id="locale-permanent-yes" type="button" data-locale="%(localeCode)s">' +
+                    gettext('Yes') + '</button> <button id="locale-permanent-no" type="button">' + gettext('No') +
+                    '</button></p></div>', {localeCode: locale.code, localeName: locale.name}, true),
+                notification = mdn.Notifier.growl(html, {closable: true, duration: 0}).question();
+
+            // Add event listener to the buttons
+            $('#locale-permanent-yes').on('click', function () {
+                $.post('/i18n/setlang/', {language: this.dataset.locale})
+                    .success(function () {
+                        notification.close();
+                        localStorage.removeItem('changed-locale-to');
+                    });
+            }
+            );
+
+            $('#locale-permanent-no').on('click', function () {
+                notification.close();
+                localStorage.removeItem('changed-locale-to');
+            }
+            );
         }
     }
 
-})();
+})(window, document, jQuery);
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -668,48 +668,5 @@
             win.history.replaceState({}, '', location.pathname);
         }
     }
-
-    // listens for post message from interactive editor
-    window.addEventListener('message', function(event) {
-        // get the interactive editor iframe
-        var iframe = document.querySelector('iframe.interactive');
-        var isExpectedOrigin = event.origin === 'https://interactive-examples.mdn.mozilla.net';
-
-        /* there may be other post messages so, ensure that the origin is the
-        expected and, that `event.data` contains an `iframeHeight` property */
-        if (isExpectedOrigin && event.data.iframeHeight) {
-            iframe.setAttribute('height', event.data.iframeHeight);
-        }
-    }, false);
-
-    // Insert notice about permanent language switch
-    var changedLocaleTo = localStorage.getItem('changed-locale-to');
-    if (changedLocaleTo) {
-        var locale = JSON.parse(changedLocaleTo),
-            text = gettext('You are now viewing this article in %(localeName)s.' +
-                           ' Do you always want to view MDN articles in %(locale_name)s?'),
-            html = interpolate('<div id="locale-permanent-banner" class="primary overheadIndicator"><p>' + text +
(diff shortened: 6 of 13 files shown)


kuma/core/middleware.py
```diff
@@ -27,15 +27,25 @@ def process_request(self, request):
...
-                         k, v in request.GET.iteritems() if k != 'lang')
-            return HttpResponsePermanentRedirect(urlparams(new_path, **query))
+            query = dict((smart_str(k), v) for k, v in request.GET.iteritems() if k != 'lang')
+
+            # 'lang' is only used on the language selection dropdown. If this is
+            # present it is safe to set language preference for the current
+            # user.
+            if request.user.is_anonymous():
+                cookie = settings.LANGUAGE_COOKIE_NAME
+                request.session[cookie] = request.GET['lang']
```
jwhitlock: This sets a session key, not a cookie. We don't want to create sessions for anonymous users.  It also sets the language based on the URL, which is not what we want.

kuma/core/middleware.py
```diff
@@ -44,7 +54,9 @@ def process_request(self, request):
             if query_string:
                 full_path = '%s?%s' % (full_path, query_string)
 
-            response = HttpResponsePermanentRedirect(full_path)
+            # Never use HttpResponsePermanentRedirect here.
+            # Its a temporary redirect and should return with http 302, not 301
+            response = HttpResponseRedirect(full_path)
```
jwhitlock: This change is for a separate issue, and I'm not sure I agree with your assertion that these should be temporary redirects.  Please move this to a new PR.

kuma/core/urlresolvers.py
```diff
@@ -134,6 +134,26 @@ def get_language(self):
...
+        """
+        user = getattr(self.request, 'user', None)
+
+        if user:
+            if user.is_authenticated():
+                # The user is authenticated. So return if he have any locale set up
+                return user.locale
+            elif user.is_anonymous():
+                # The user is not authenticated. So find if he have a cookie for language preference
+                language = self.request.session.get(settings.LANGUAGE_COOKIE_NAME)
```
jwhitlock: Anonymous users should not have sessions. We want to store this feature in a cookie,

kuma/core/urlresolvers.py
```diff
@@ -134,6 +134,26 @@ def get_language(self):
...
+        If the user is not logged in and have a cookie set for language, return that language
+        For logged in user, return his language pointed in his profile locale.
+        Otherwise, return None
+        """
+        user = getattr(self.request, 'user', None)
+
+        if user:
+            if user.is_authenticated():
+                # The user is authenticated. So return if he have any locale set up
+                return user.locale
```
jwhitlock: The language in the user profile shouldn't be used for this feature, and should not override the cookie selection.

kuma/settings/common.py
```diff
@@ -406,32 +406,33 @@ def enable_candidate_languages():
...
     'kuma.core.middleware.RemoveSlashMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-)
 
-if not MAINTENANCE_MODE:
     # We don't want this in maintence mode, as it adds "Cookie"
     # to the Vary header, which in turn, kills caching.
-    MIDDLEWARE_CLASSES += ('django.middleware.csrf.CsrfViewMiddleware',)
+    ('django.middleware.csrf.CsrfViewMiddleware'
+     if not MAINTENANCE_MODE else ''),
```
jwhitlock: I didn't test this yet, but I'll have to. I'm not 100% sure this will be valid when ``MAINTENANCE_MODE=True``.

kuma/settings/common.py
```diff
@@ -406,32 +406,33 @@ def enable_candidate_languages():
...
-    # kuma.core.urlresolvers.reverse() to add locale prefixes to URLs:
     'kuma.core.middleware.SetRemoteAddrFromForwardedFor',
     ('kuma.core.middleware.ForceAnonymousSessionMiddleware'
      if MAINTENANCE_MODE else
      'django.contrib.sessions.middleware.SessionMiddleware'),
+
+    # LocaleURLMiddleware requires access to request.user. These two must be
+    # loaded before the LocaleURLMiddleware
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'kuma.core.anonymous.AnonymousIdentityMiddleware',
```
jwhitlock: Rearranging middleware seems dangerous to me.  Not sure if our tests would fully expose issues.

jwhitlock:
I don't think this PR represents a good starting place for this feature.  Please take a look at [bug 1331729](https://bugzilla.mozilla.org/show_bug.cgi?id=1331729#c5).

safwanrahman:
@jwhitlock Have you cleared browser cache and cookie for localhost or use private window before testing it?
Can you please reopen the PR so I can make changes according to need and push at this branch?

kuma/core/middleware.py
safwanrahman: This change is needed. Suppose, the user choose `bn-BD` and we permanently redirect him to `bn-BD`, then the user chose to go `en-US` and as we permanently redirect them to `bn-BD`, the browser will always take them there. We can not do anything from our code end.

safwanrahman:
@jwhitlock I have fixed the issue you mentioned and written a tests. Let me know if any more tests is needed.

safwanrahman:
[Screen cast](https://www.dropbox.com/s/dkyhqbm9vxd1gns/Screencast%202017-07-28%2002%3A21%3A29.mp4?dl=0)

kuma/core/middleware.py
```diff
@@ -44,7 +54,7 @@ def process_request(self, request):
             if query_string:
                 full_path = '%s?%s' % (full_path, query_string)
 
-            response = HttpResponsePermanentRedirect(full_path)
+            response = HttpResponseRedirect(full_path)
```
jwhitlock: Please add this change to a different PR, against [bug 1385315](https://bugzilla.mozilla.org/show_bug.cgi?id=1385315). Also, it needs to apply to wiki pages and not any other page, so a middleware is probably not appropriate.

kuma/core/middleware.py
```diff
@@ -27,15 +27,25 @@ def process_request(self, request):
...
+            query = dict((smart_str(k), v) for k, v in request.GET.iteritems() if k != 'lang')
+
+            # Never use HttpResponsePermanentRedirect here.
+            # Its a temporary redirect and should return with http 302, not 301
+            response = HttpResponseRedirect(urlparams(new_path, **query))
+
+            # 'lang' is only used on the language selection dropdown. If this is
+            # present it is safe to set language preference as cookie for the current
+            # user.
+            response.set_cookie(settings.LANGUAGE_COOKIE_NAME, lang)
```
jwhitlock: We want to avoid search engine scrapers getting a language preference cookie. The use of the language switcher should prompt the user if they want us to remember their language setting. A POST is needed to create the cookie. I don't believe a middleware is the right solution for setting the cookie.

kuma/core/urlresolvers.py
```diff
@@ -134,6 +135,14 @@ def get_language(self):
 
         return settings.LANGUAGE_CODE
 
+    def get_chosen_language(self):
+        """If the request has a cookie set for language, return that language."""
+        language = self.request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+
+        # Need to check if his cookie language is supported by Kuma.
+        if language and language in dict(settings.LANGUAGES):
+            return language
```
jwhitlock: Should the default return be the empty string ``""``, to match the return of ``split_path`` for no language?

kuma/core/urlresolvers.py
```diff
@@ -111,9 +111,10 @@ class Prefixer(object):
     def __init__(self, request=None, locale=None):
         """If request is omitted, fall back to a default locale."""
         self.request = request or RequestFactory(REQUEST_METHOD='bogus').request()
-        self.locale, self.shortened_path = split_path(self.request.path_info)
-        if locale:
-            self.locale = locale
+        path_locale, self.shortened_path = split_path(self.request.path_info)
+
+        # Set Self locale according to priority.
+        self.locale = locale or path_locale or self.get_chosen_language()
```
jwhitlock: I believe the previous version always sets ``self.locale`` to a string, possibly the empty string ``""``, while this could set it to ``None``.  I'll need to read closer to determine if this breaks the code, or you could ensure it is set to a string with another ``or ""``

kuma/core/middleware.py
safwanrahman: Added in #4345

kuma/core/middleware.py
safwanrahman: Hmm. I am also thinking about this that whats the best way to implement this

kuma/core/urlresolvers.py
safwanrahman: Yap. This should be changed to `or ""`

kuma/core/urlresolvers.py
safwanrahman: I think if we add another `or ""` above, returning `None` from here should be accepted

safwanrahman:
As per conversation over IRC, the kitsune implementation can be seen [here](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/middleware.py#L49-L53)
It was implemented by https://github.com/mozilla/kitsune/pull/1382

jwhitlock:
@safwanrahman thanks for finding that code reference! I think the original code reviewers missed that a session was being set, not a cookie. I think they had problems with too many sessions when they switched Kitsune back on, so maybe they's like to fix this code.  I'd like to do better for Kuma.

I've searched for Googlebot and Cookies, and found different results:

* https://webmasters.googleblog.com/2008/10/first-click-free-for-web-search.html - Google endorses using the Referer field for "first-click-free" detection (user can see articles linked from Google search, pay for others), as opposed to a cookie. 2008-10
* https://davekz.com/googlebot-and-cookies/ - Google doesn't use cookies, using Google Web Master Tools' “Fetch as Googlebot”. 2013-02
* http://www.tomanthony.co.uk/blog/googlebot-accepting-cookies/  - Google accepts cookies when scraping with JS enabled and rendering a page, but probably doesn't keep them for future page requests. 2014-05
* https://twitter.com/methode/status/843828128711495682 - A Google employee says they can detect if they need a cookie to get to content. The method isn't spelled out, but I'd guess Google is watching what URLs Chrome users view, vs what Googlebot can get to. 2017-03

SO, I think we can use a cookie. I think it still makes sense to set the cookie when the language selector is used, rather than in the middleware. We may want to avoid setting the cookie for known bots, such as Googlebot.

The JS that handles the language selector at the bottom is in [kuma/static/js/wiki.js](https://github.com/mozilla/kuma/blob/master/kuma/static/js/wiki.js#L408-L412). If JS is disabled, it acts like a form with a GET parameter.  We could include a POST that sets the language cookie.

The top language selector is a little more expected, as a list of links. No JS seems to be involved, so there's no hook for a request to set a cookie.

This will still be a complex feature to get right.

safwanrahman:
@stephaniehobson for frontend. r?
@jwhitlock for backend. r?

jwhitlock:
@safwanrahman this is not an AWS blocker, so I'm not going to have a chance to look at it before October.

kuma/urls.py
```diff
@@ -81,6 +81,8 @@
     url(r'^miel$',
         handler500,
         name='users.honeypot'),
+    # We use our own views for setting language in cookies. But to just align with django, set it like this.
+    url(r'^i18n/setlang/', core_views.set_language),
```
jwhitlock: @This should be in the ``LANGUAGE_URL_IGNORED_PATHS`` setting. Or do you mean to serve it at ``/{locale}/i18n/setlang``? What else did you want at ``/i18n/``?

kuma/static/js/wiki.js
```diff
@@ -896,4 +896,35 @@
...
+                               '<br><button id="locale-permanent-yes" data-locale="%(locale_code)s">' +
+                               gettext('Yes') + '</button> <button id="locale-permanent-no">' +
+                               gettext('No') + '</button></p></div>',
+                {locale_code: locale.code, locale_name: locale.name}, true);
+
+        $('#wiki-content').prepend(html);
+
+        // Add event listener to the buttons
+        $('#locale-permanent-yes').on('click', function () {
+                $.post('./i18n/setlang/', {language: this.dataset.locale})
```
jwhitlock: ``./i18n/setlang/`` should be ``/i18n/setlang/``. Or maybe you mean to serve it at ``/{locale}/i18n/setlang/``?

kuma/core/views.py
```diff
@@ -1,11 +1,33 @@
...
+from django.views.decorators.http import require_POST
 
 
 def _error_page(request, status):
     """Render error pages with jinja2."""
     return render(request, '%d.html' % status, status=status)
 
 
+@csrf_exempt
+@require_POST
```
jwhitlock: A ``GET`` version of this page would be useful as well, but that can be later.

kuma/static/js/wiki.js
```diff
@@ -896,4 +896,35 @@
...
+                    .success(function () {
+                        $('#locale-permanent-banner').remove();
+                        localStorage.removeItem('changed-locale-to');
+                    })
+            }
+        );
+
+        $('#locale-permanent-no').on('click', function () {
+            $('#locale-permanent-banner').remove();
+            localStorage.removeItem('changed-locale-to');
```
jwhitlock: I'd like to store that a user said "no", so that we don't ask them over and over again every time they use the language switcher.

kuma/core/urlresolvers.py
jwhitlock: Also, not covered by tests, according to codecov

kuma/core/views.py
```diff
@@ -1,11 +1,33 @@
...
+    if lang_code and lang_code in dict(settings.LANGUAGES):
+
+        response.set_cookie(key=settings.LANGUAGE_COOKIE_NAME,
+                            value=lang_code,
+                            max_age=settings.LANGUAGE_COOKIE_AGE,
+                            path=settings.LANGUAGE_COOKIE_PATH,
+                            domain=settings.LANGUAGE_COOKIE_DOMAIN,
+                            )
+
+    return response
```
jwhitlock: This function has no tests according to codecov

jwhitlock:
After a rebase and fixing the POST URL, I'd like @stephaniehobson's opinion on the approach from a UI perspective. I'm OK with shipping if this is a step in the right direction, and we can refine in future PRs.

I can also add tests in follow-on commits, if desired.

safwanrahman:
@jwhitlock Corrected the path, added tests and done other fixes. Another round of review?

@stephaniehobson Can you please review the frontend thing?

stephaniehobson:
Yay, a rebase :) I'll have a look at this on Monday.

stephaniehobson:
Didn't have time today :( Hopefully tomorrow.

kuma/static/js/wiki.js
```diff
@@ -679,4 +679,47 @@
...
+
+        /* there may be other post messages so, ensure that the origin is the
+        expected and, that `event.data` contains an `iframeHeight` property */
+        if (isExpectedOrigin && event.data.iframeHeight) {
+            iframe.setAttribute('height', event.data.iframeHeight);
+        }
+    }, false);
+
+    // Insert notice about permanent language switch
+    var changedLocaleTo = localStorage.getItem('changed-locale-to');
```
jwhitlock: This code also needs to appear in ``wiki-skinny.js``, or be moved to a new file that is included in both ``wiki-skinny.js`` and ``wiki.js``.

stephaniehobson:
Tomorrow for sure.

kuma/static/js/wiki.js
stephaniehobson: +1 Moving this code to a separate file. It should be part of lang-switcher.js

kuma/static/js/wiki.js
```diff
@@ -679,4 +679,47 @@
...
+                {locale_code: locale.code, locale_name: locale.name}, true);
+
+        $('#wiki-content').prepend(html);
+
+        // Add event listener to the buttons
+        $('#locale-permanent-yes').on('click', function () {
+                $.post('/i18n/setlang/', {language: this.dataset.locale})
+                    .success(function () {
+                        $('#locale-permanent-banner').remove();
+                        localStorage.removeItem('changed-locale-to');
```
stephaniehobson: This code depends on the local storage API, so the entire feature needs to be wrapped in a feature test for local storage. You can see an example of that[ in auth.js](https://github.com/mozilla/kuma/blob/46e328b11ab7f509a09fef16ceee5fa81f4e2997/kuma/static/js/auth.js#L57)

kuma/static/styles/components/wiki/indicators.scss
```diff
@@ -86,6 +86,11 @@ modifications for 'system' indicators
     }
 }
 
+.primary {
```
stephaniehobson: We have an existing system for notifications as well as variables defined for colours, there's no need to add new things to the style sheets for this.

kuma/static/js/wiki.js
```diff
@@ -679,4 +679,47 @@
...
+        var locale = JSON.parse(changedLocaleTo),
+            text = gettext('You are now viewing this article in %(locale_name)s.' +
+                           ' Do you always want to view MDN articles in %(locale_name)s?'),
+            html = interpolate('<div id="locale-permanent-banner" class="primary overheadIndicator"><p>' + text +
+                               '<br><button id="locale-permanent-yes" data-locale="%(locale_code)s">' +
+                               gettext('Yes') + '</button> <button id="locale-permanent-no">' +
+                               gettext('No') + '</button></p></div>',
+                {locale_code: locale.code, locale_name: locale.name}, true);
+
+        $('#wiki-content').prepend(html);
```
stephaniehobson: Please use the notification system for this. [The function is here.](https://github.com/mozilla/kuma/blob/7a7120f12c9349415877eacb040bfb393099d6d1/kuma/static/js/components.js#L473) I'm sorry we don't have any documentation on how to use it.  [Here's an example of how it's used when asking users to participate in our task completion survey](https://github.com/mozilla/kuma/blob/ad799cb8972c6b3857dc907cede8eddfbb0d5e86/kuma/static/js/task-completion.js#L36)e.

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,27 @@
+(function () {
+    var langSwitcherSelector = document.getElementById('language'),
+        langSwitcherButton = document.getElementById('translations');
+
+    function storeLocaleChange(code, name) {
```
stephaniehobson: This functionality also needs to be wrapped in a localstorage feature test.

kuma/static/js/wiki.js
```diff
@@ -679,4 +679,47 @@
...
+    }, false);
+
+    // Insert notice about permanent language switch
+    var changedLocaleTo = localStorage.getItem('changed-locale-to');
+    if (changedLocaleTo) {
+        var locale = JSON.parse(changedLocaleTo),
+            text = gettext('You are now viewing this article in %(locale_name)s.' +
+                           ' Do you always want to view MDN articles in %(locale_name)s?'),
+            html = interpolate('<div id="locale-permanent-banner" class="primary overheadIndicator"><p>' + text +
+                               '<br><button id="locale-permanent-yes" data-locale="%(locale_code)s">' +
```
stephaniehobson: Yay `<button>`! Button is the right element for this. It also needs the attribute `type=button`.

safwanrahman:
@schalkneethling I have fixed all the issue @stephaniehobson raised.
Can you please review this long waiting PR?

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
+(function (win, doc, $) {
+    function storeLocaleChange(code, name) {
+        localStorage.setItem('changed-locale-to', JSON.stringify({code: code, name: name}));
+    }
+
+    if(win.mdn.features.localStorage) {
+        var langSwitcherSelector = document.getElementById('language'),
+            langSwitcherButton = document.getElementById('translations');
```
schalkneethling: Code style: Implicitly define each `var` on it's own line. Do not use the single `var` pattern

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+                var element = this.options[this.options.selectedIndex];
+                storeLocaleChange(element.dataset.locale, element.label);
+            });
+        }
+
+        if (langSwitcherButton) {
+            var transChoices = langSwitcherButton.querySelectorAll('li a');
+
+            for (var i = 0; i < transChoices.length; i++) {
+
```
schalkneethling: Nit: Unnecessary empty line.

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+        var changedLocaleTo = localStorage.getItem('changed-locale-to');
+        if (changedLocaleTo) {
+            var locale = JSON.parse(changedLocaleTo),
+                text = gettext('You are now viewing this site in %(localeName)s.' +
+                               ' Do you always want to view this site in %(localeName)s?'),
+                html = interpolate(text +
+                    '<br><button id="locale-permanent-yes" type="button" data-locale="%(localeCode)s">' +
+                    gettext('Yes') + '</button> <button id="locale-permanent-no" type="button">' + gettext('No') +
+                    '</button></p></div>', {localeCode: locale.code, localeName: locale.name}, true),
+                notification = mdn.Notifier.growl(html, {closable: true, duration: 0}).question();
```
schalkneethling: Code style: Do not use single `var` pattern

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+
+            // Add event listener to the buttons
+            $('#locale-permanent-yes').on('click', function () {
```
schalkneethling: Nit: No space after `function` keyword

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+                $.post('/i18n/setlang/', {language: this.dataset.locale})
+                    .success(function () {
+                        notification.close();
+                        localStorage.removeItem('changed-locale-to');
+                    });
+            }
+            );
```
schalkneethling: Nit: Move closing parentheses to same line as closing curly brace

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+
+            $('#locale-permanent-no').on('click', function () {
```
schalkneethling: Nit: No space after `function` keyword

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+                notification.close();
+                localStorage.removeItem('changed-locale-to');
+            }
+            );
```
schalkneethling: Nit: Move closing parentheses to same line as closing curly brace

kuma/static/js/lang-switcher.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
```
schalkneethling: Nit: Unnecessary empty line

kuma/static/js/wiki.js
```diff
@@ -668,5 +668,5 @@
             win.history.replaceState({}, '', location.pathname);
         }
     }
-
+
```
schalkneethling: Nit: Unnecessary empty line

schalkneethling:
Found a probable bug with regards to the flow and the dialog that is shown to users once they change their locale.

1) Using a private window, go to 127.0.0.1:8000
2) This will then redirect to 127.0.0.1:8000/en-US/

screen shot 2018-01-09 at 13 28 36

3) Using the lang switcher, switch to `bn-BD`
4) The page is reloaded and redirected to 127.0.0.1:8000/bn-BD/#
5) It also shows the dialog(growl message) asking whether the user wants to access the site using the `bn-BD` locale in future

screen shot 2018-01-09 at 13 28 51

6) Do not click on either of the options but instead, change the URL back to only 127.0.0.1:8000 and hit enter
7) The page reloads and the URL is now 127.0.0.1:8000/en-US/

This is the expected behaviour but, the growl message is still shown with the same message as before even though the locale has changed back to `en-US`:

screen shot 2018-01-09 at 13 29 08

This seems odd as the message states that the user is currently viewing the site using the `bn-BD` locale when they are in fact viewing it as `en-US`. Clicking yes on the growl message here, does indeed set the cookie and, should you change the URL back to 127.0.0.1:8000 and hit enter, it will load as 127.0.0.1:8000/bn-BD/

kuma/static/js/lang-switcher.js
safwanrahman: @schalkneethling I was wondering the reason behind not using single `var` pattern. In my early days, some people suggested me to use this pattern while writing javascript.
Searching in internet did not provide me any actual answer.
Can you please explain more about the underlying reason behind it?
I have found this on eslint documentation https://eslint.org/docs/rules/one-var

safwanrahman:
@schalkneethling Regarding the `probable bug` you have found, I think we can eliminate it by changing the word something like `You have changed the language to {{ locale_name}}, Do you always want to visit this site in {{ locale_name }}?`

Or can you please guide me what approach I can take in order to fix the probabble bug?

kuma/static/js/lang-switcher.js
schalkneethling: For the most part it is a stylistic choice. One of the other reasons is that it avoid accidental implicit globals, should one forget the `var`. Basically one is more explicit about your definitions. Now, with the additional of `let`, `const` etc. being explicit is even more important as it suggests intent.

Our `eslint` rules are now also set to fail on the single var pattern. I used to also like the single `var` pattern but, over time I have come to prefer being explicit by clearly defining each variable.

schalkneethling:
@safwanrahman Yes, changing to wording could go some way to solve the problem. Ideally we would not want to show this dialog each time a user changes their locale. So, perhaps for the current session you can have logic that stores that the user has been shown the dialog once, but took no action. You defer showing the messaging on locale changes until the users starts a completely new session. I am thinking `sessionStorage` would work for this. Thoughts?

kuma/static/js/lang-switcher.js
safwanrahman: @schalkneethling I understand and also accept the concern you have raised. Since we are using ES5 still, I thought it should be one var convention.
But as we are hoping to move writing to ES6 code, I think its better approach to not use the one var constrains.

Regarding eslint rules, we use [`one-var-declaration-per-line`](https://github.com/mozilla/kuma/blob/master/.eslintrc.js#L50) that does not fail on single var pattern and thats why my PR passes the current eslint. Do you want I add the rule to fail `one-var` pattern in `.eslintrc.js`?

safwanrahman:
@schalkneethling I think `sessionStorage` will work. But the thing is it will not work if the user dont click on the notification and go to any other page.
I think we I can write something [here](https://github.com/safwanrahman/kuma/blob/d1b34d0e46b9d77da5b4b8bc4da8c195f4015604/kuma/static/js/lang-switcher.js#L38) like following:

```javascript
var notification = mdn.Notifier.growl(html, {closable: true, duration: 0});
notification..question();
localStorage.removeItem('changed-locale-to');
```
what do you think?

safwanrahman:
@schalkneethling I have fixed the bug as you mentioned as per the approach described in the above comment. What do you think about it now?
now r?

schalkneethling:
@safwanrahman Yeah, that definitely solves that problem but... ;)

@jwhitlock mentioned that if you click `No`, then say change to `en-US` and then again change to another locale say `fr`, the dialog will show again in the same session. Thing is, it happens even in the following scenario:

1. Land on the site in en-US
2. Change to bn-BD
3. Receive message dialog and click `No`
4. Change back to en-US
5. Change back to bn-BD
6. You are prompted again with the same message even though you chose `No` previously.

Another thing I was wondering about. Let's say you clicked `Yes` but now, you have decided that you actually do no want to view the site in the specific locale anymore. How do you go about changing that preference?

It almost feels like this should be a setting under profile, that a user can change at will. Thoughts @safwanrahman @jwhitlock

jwhitlock:
It takes me a bit to remember how this works. Here's a test plan for this feature.

**Setup**
* Check out branch, make sure it is up to date
* Rebuild assets (``make build-static`` or similar)
* Set browser's preferred language to ``en-US`` (alternatively, replace ``en-US`` with your browser's preferred language, ``de`` etc with alternate languages.

Before each test, start a new private browsing window, or:

* Delete cookie ``django_language``
* Delete local storage ``changed-locale-to``

## Test 1: Home page, Other Languages footer

1. Go to http://localhost:8000/. Expected:
    - Redirected to http://localhost:8000/en-US/
2. In "Other languages" dropdown in footer, change to "Deutsch (de)". Expected:
    - "changed-locale-to" gets value ``{code: "de", name: "Deutsch (de)"}``, and is then deleted (may be too fast to see unless looking)
    - http://localhost:8000/de/ is loaded
    - Dialog in upper right, "You are now viewing this site in Deutsch (de). Do you always want to view this site in Deutsch (de)?"
3. Click "X" at upper right corner of dialog. Expected:
    - Dialog is closed
    - No session cookie ``django_language``
4. Go to http://localhost:8000/. Expected:
    - Redirected to http://localhost:8000/en-US/
5. In "Other languages" dropdown in footer, change to "Deutsch (de)". Expected:
    - "changed-locale-to" gets value ``{code: "de", name: "Deutsch (de)"}``, and is then deleted (may be too fast to see unless looking)
    - http://localhost:8000/de/ is loaded
    - Dialog in upper right, "You are now viewing this site in Deutsch (de). Do you always want to view this site in Deutsch (de)?"
6. Click "No" in dialog. Expected:
    - Dialog is closed
    - No session cookie ``django_language``
7. Go to http://localhost:8000/. Expected:
    - Redirected to http://localhost:8000/en-US/
8. In "Other languages" dropdown in footer, change to "Deutsch (de)". Expected:
    - "changed-locale-to" gets value ``{code: "de", name: "Deutsch (de)"}``, and is then deleted (may be too fast to see unless looking)
    - http://localhost:8000/de/ is loaded
    - Dialog in upper right, "You are now viewing this site in Deutsch (de). Do you always want to view this site in Deutsch (de)?"
9. Click "Yes" in dialog. Expected:
    - Dialog is closed
    - Session cookie ``django_language`` is set to ``de``
10. Go to http://localhost:8000/. Expected:
    - Redirected to http://localhost:8000/de/
11. Go to http://localhost:8000/en-US/. Expected:
   - http://localhost:8000/en-US/ loads (URL wins)

## Test 2: Wiki Page, Languages Tool

1. Go to http://localhost:8000/docs/Web/HTML. Expected:
    - Redirected to http://localhost:8000/en-US/docs/Web/HTML
2. In "Languages" toolbar, on the line with the "HTML" ``<h1>`` header, switch to "Deutsch (de)". Expected:
    - "changed-locale-to" gets value ``{code: "de", name: "Deutsch (de)"}``, and is then deleted (may be too fast to see unless looking)
    - http://localhost:8000/de/docs/Web/HTML is loaded
    - Dialog in upper right, "You are now viewing this site in Deutsch (de). Do you always want to view this site in Deutsch (de)?"
3. Click "X" at upper right corner of dialog. Expected:
    - Dialog is closed
    - No session cookie ``django_language``
4. Go to http://localhost:8000/docs/Web/HTML. Expected:
    - Redirected to http://localhost:8000/en-US/docs/Web/HTML
5. In "Languages" toolbar, on the line with the "HTML" ``<h1>`` header, switch to "Deutsch (de)". Expected:
    - "changed-locale-to" gets value ``{code: "de", name: "Deutsch (de)"}``, and is then deleted (may be too fast to see unless looking)
    - http://localhost:8000/de/docs/Web/HTML is loaded
    - Dialog in upper right, "You are now viewing this site in Deutsch (de). Do you always want to view this site in Deutsch (de)?"
6. Click "No" in dialog. Expected:
    - Dialog is closed
    - No session cookie ``django_language``
7. Go to http://localhost:8000/docs/Web/HTML. Expected:
    - Redirected to http://localhost:8000/en-US/docs/Web/HTML
8. In "Languages" toolbar, on the line with the "HTML" ``<h1>`` header, switch to "Deutsch (de)". Expected:
    - http://localhost:8000/de/docs/Web/HTML is loaded
    - Dialog in upper right, "You are now viewing this site in Deutsch (de). Do you always want to view this site in Deutsch (de)?"
9. Click "Yes" in dialog. Expected:
    - Dialog is closed
    - Cookie ``django_language`` is set to ``de``
10. Go to http://localhost:8000/docs/Web/HTML. Expected:
    - Redirected to http://localhost:8000/de/docs/Web/HTML
11. Go to http://localhost:8000/en-US/docs/Web/HTML. Expected:
    - http://localhost:8000/de/docs/Web/HTML loads (URL wins)

jwhitlock:
It's looking pretty good to me, and the Django tests cover the new code.

I noticed some weirdness in the session storage. When using the Language toolbar on a wiki page, the value of  ``changed-locale-to`` gets set to empty, rather than being deleted. This may be a bug, or weirdness with the storage tool.

How about a third option for the dialog:

* "Yes" - store the new preferred redirect locale
* "No", X to close - close the dialog, do nothing
* "Never" - store that the user doesn't want to see the dialog again

My main reason for the "Never" is personal. I have to view translations all the time when debugging page relationships, and I will be annoyed if I have to close that dialog every time.

I suspect translators are in a similar situation, and often go back and forth between two languages, and will be annoyed by the dialog, but I haven't proved it. We could delay that until others tell us they are annoyed.

I'm OK to delay "how do I undo it" until later, but if @schalkneethling thinks it is important, we can handle that before shipping.

schalkneethling:
@safwanrahman 
>  If the person click no and we dont show the message to him anytime, there are no ways for him to permanently add the locale.

I was here thinking that only for the current session the user is not prompted again.

> So if you change to bn-BD then you dont want to stay in bn-BD and instead go to en-US, you just change the locale to en-US and click on yes button to make it permanent.

Indeed! My bad. I did not test this scenario.

> I also feel like there should be a setting under profile, but I think that should be handeled by another bug. This PR atleast should improve the current problem user are facing.

I agree that this is a step in the right direction.

@jwhitlock 
> My main reason for the "Never" is personal. I have to view translations all the time when debugging page relationships, and I will be annoyed if I have to close that dialog every time.

I was definitely also thinking about this scenario when I suggested not prompting the user more than once per session.

@safwanrahman I reckon if you can implement the "Never" option as suggested by John, this would good to merge. Thanks so much for all the work you have, and are putting, into this.

safwanrahman:
@schalkneethling I have implemented the never button. Can you please review? I have refactored some for clean code.
r?

PR #4322, opened 2017-07-25
Fix Bug 1375947: Add spacing after page content

stephaniehobson:
Content was sitting too close to the footer on non-article pages (e.g. $history or profile edit page).

Moved the padding from the article wrapper to the content wrapper and increased it a bit because new design has more white space and this mirrors the space on top.

### DIFF
--- a/kuma/static/styles/base/utilities.scss
+++ b/kuma/static/styles/base/utilities.scss
@@ -23,6 +23,7 @@ Utility classes
 
     main > & {
         padding-top: $first-content-top-padding;
+        padding-bottom: $last-content-bottom-padding;
     }
 
     @media #{$mq-tablet-and-up} {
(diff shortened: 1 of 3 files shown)


kuma/static/styles/base/utilities.scss
```diff
@@ -23,6 +23,7 @@ Utility classes
 
     main > & {
         padding-top: $first-content-top-padding;
+        padding-bottom: $first-content-top-padding;
```
schalkneethling: Nit: The naming here is a little strange as, `$first-content-top-padding` is applied to `padding-bottom`

schalkneethling:
r+w/c 🍓

stephaniehobson:
Now w/c 🍓

PR #4324, opened 2017-07-25
Fix Bug 1383965: Hide submenu close button on big screens

stephaniehobson:


schalkneethling:
Not sure if this is a simulator thing or, whether this will happen on the real device as well but, testing on an iPad simulator, there is no way to close the drop down menus other than clicking into a form field. See below:

https://cl.ly/04032f073W1U

stephaniehobson:
Well, the functionality isn't a regression at least :/ 

Bugs already exist for the ideal solution:
[Bug 1218669](https://bugzilla.mozilla.org/show_bug.cgi?id=1218669) - Dropdown menus need close buttons at higher screen sizes
[Bug 1255148](https://bugzilla.mozilla.org/show_bug.cgi?id=1255148) - Add feature detection for touch

I've increased the media query to just hide it on really big monitors, should be showing up on iPads now. @schalkneethling What do you think? ...or maybe this is a question for Kadir.

kuma/static/styles/components/structure/submenu.scss
```diff
@@ -103,3 +104,12 @@ everything else */
         }
     }
 }
+
+
+@media #{$mq-desktop-and-up} {
```
schalkneethling: I am guessing you meant this to be `@media #{$mq-large-desktop-and-up}`

schalkneethling:
Small typo on the media query variable but, with that changed, this works as expected on all the things r+ 🎉

PR #4327, opened 2017-07-25
Bug 1370594: Update look of helpfulness survey

stephaniehobson:


schalkneethling:
@stephaniehobson How do I get the survey to show up locally? Looks like there is a flag `helpful-survey-2` which I did turn on, but still not seeing it locally.

stephaniehobson:
@schalkneethling I think it also needs a working GA configuration. Define `GOOGLE_ANALYTICS_ACCOUNT` in `/admin/constance/config/`. It's okay to copy the production number if you want, we have a filter on GA to keep stuff from different domains out.

schalkneethling:
> It's okay to copy the production number if you want, we have a filter on GA to keep stuff from different domains out.

Thanks @stephaniehobson can you email me the details for that? Not sure where to get hold of the proper config for that locally. Is it in the repo somewhere?

jwhitlock:
I think I use "UA-000000-1" for testing.

schalkneethling:
After many attempts at getting it to load locally, I eventually simply copied the HTML from a production page and stuck it into the source using dev tools ;) With that I could test it visually. r+ 🌻

schalkneethling:
Couple of screen shots for reference:

screen shot 2017-07-26 at 12 29 14

screen shot 2017-07-26 at 12 29 26

## Small screen

screen shot 2017-07-26 at 12 29 43

PR #4328, opened 2017-07-25
bug 1379288: Use locale-specific font in CKEditor content

jwhitlock:
For locales supported by the Zilla font family, use a single source file ``default.scss`` that generates locale-specific CSS like ``locale-en-US.css``, ``locale-fr.css``, etc.  Generate the list of pipeline files, omitting ones that aren't currently generated (af, bg, sr, etc.)

Add a new locale-specific pipeline definition, ``editor-locale-XX``, that is suitable for the CKEditor content style definition.


stephaniehobson:
I'm not sure this is the right approach. The Zilla file is not a default and we should not automatically be applying it. It should intentionally be applied to European Latin languages. Perhaps in the future this will expand to include Latin non-European languages like Vietnamese.

Rather than default I think this file is better named Zilla.

Does this approach allow us to include both "default" and a local specific style?

jwhitlock:
> I'm not sure this is the right approach. The Zilla file is not a default and we should not automatically be applying it. It should intentionally be applied to European Latin languages. Perhaps in the future this will expand to include Latin non-European languages like Vietnamese.
>
> Rather than default I think this file is better named Zilla.

That would leave us with ``_zilla.scss`` and ``Zilla.scss`` in the same folder.  Or is it ``zilla.scss`` with a lowercase ``z``? Or maybe just setting these locals to use ``_zilla.scss`` directly would work...

> Does this approach allow us to include both "default" and a local specific style?

I'm not sure what you mean.  [ro.scss](https://github.com/mozilla/kuma/blob/e8d406ef9f14e76682223fcfdcb0d2ac61e50245/kuma/static/styles/locales/ro.scss) includes ``_zilla.scss`` and extends it - is that what you mean?  Or do you plan to have composed styles by sourcing on two or more ``.scss`` files?

jwhitlock:
Rebased on master, no code changes

stephaniehobson:
Assigning back to @jwhitlock to re-think architectural approach.

jwhitlock:
After our conversation, re-wrote the change.  There are now two lists.

``LOCALE_USE_CUSTOM`` is the list of locales where the font rules are in a locale-specific file, like ``locales/ja.scss``. In ``DEBUG=True`` mode, the filename in the HTML will be ``locales/<LOCALE>.css``. In ``DEBUG=False``, it will be ``locale-<LOCALE>.css``.

``LOCALE_USE_ZILLA`` are locales that are well supported by the Zilla family, and use the rules in ``locales/en-US.scss``.  In ``DEBUG=True`` mode, the filename in the HTML will be ``locales/en-US.css``. In ``DEBUG=False``, it will be ``locale-<LOCALE>.css``.

Omitted fonts (like ``bn-BD``) have no locale in the output, and will use the browser default fonts.

stephaniehobson:
Zilla appearing on `en-US` and `de` pages 👍  Locale specific fallback appearing on `ja` and `zh-CN` 👍 No locale specific sheet on `vi` 👍

PR #4331, opened 2017-07-26
Mark docker build as optional

jwhitlock:
http://status.quay.io show degraded performance. Don't fail builds while they work on fixing the issue.


jwhitlock:
http://status.quay.io/ has recovered.

PR #4332, opened 2017-07-27
Fix Bug 1370594: Remove blue design files

stephaniehobson:
- Make `redesign` waffle changes permanent
- Remove`redesign_beta` waffle
- Remove `-blue` .scss files
- Remove non .woff versions of site fonts
- Remove unused classes


kuma/static/styles/search-suggestions.scss
```diff
@@ -76,7 +74,7 @@
                 &:hover,
                 &:focus,
                 &.selected {
-                    background: $homepage-background-color;
+                    background: $blue-dark;
```
jwhitlock: This will save a lot of WTFs for the next developer 👍

kuma/static/styles/zone-ten.scss
```diff
@@ -54,7 +53,6 @@ header modifications
         .masthead-text {
             width: -calc-col-width(6);
             margin-top: -180px;
-            margin-bottom: 110px;
```
jwhitlock: 👍 https://developer.mozilla.org/en-US/docs/MDN_at_ten looks better without it.

jinja2/base.html
```diff
@@ -231,7 +175,7 @@
             </ul>
           </div>
         </div>
-      </li><li class="nav-search-link nav-main-search-link"><a href="{{ url('search') }}" title="Search"><i aria-hidden="true" class="icon-search"></i></a></li>{% if not is_sphinx %}<li class="main-nav-search nav-main-search"><form action="{{ url('search') }}" method="get" role="search">
+      </li>{% if not is_sphinx %}<li class="nav-main-search"><form action="{{ url('search') }}" method="get" role="search">
```
stephaniehobson: Yup, that was intentional. I'm trying to make the class names all`nav-main-x`

PR #4333, opened 2017-07-27
Bug 1370594: Smaller typography

stephaniehobson:
New typography approved by brand.


stephaniehobson:
@schalkneethling Can you R+/- this but not merge it? I'm not 100% sold on the changes. I'm going to see if I can talk people into a hierarchy more like bedrock uses so that everything is not just 2px different.

kuma/static/styles/components/wiki/document-head.scss
```diff
@@ -46,3 +47,5 @@
         @include large-overflow();
     }
 }
+
+*/
```
schalkneethling: Nit: I generally prefer to delete things as apposed to comment them out seeing that we are using a source control system. There is always the chance that one forgets to remove it should you never enable it again, which can lead to code bloat. I guess SASS will help here but, just a thought.

kuma/static/styles/includes/_vars.scss
```diff
@@ -122,20 +122,16 @@ $diff-font-family: $code-inline-font-family;
 /*
 sizes
 ====================================================================== */
-$h1-font-size: 50px;
-$h2-font-size: 38px;
-$h3-font-size: 24px;
+$h1-font-size: 48px;
+$h2-font-size: 24px;
```
schalkneethling: It is a seriously big jump from `48px` to `24px` with nothing in between. It kinda feels like there is going to be many situations where one is going to apply custom CSS to a lot of `h2` elements because `24px` feels to small, `48px` to big and, semantically you cannot use another `h1` anyhow.

schalkneethling:
@stephaniehobson So, the code looks good technically. Made some comment, one a nit, one a general feeling about the hierarchy of the headings. From your comment above I get the feeling that you will agree with me ;)

I would much prefer to use an actual modular scale that is based on a `16px` base body font size, scaling modularly to a `h1` of `48px` ~ http://www.modularscale.com/?16&px&48@6

Or even, 16px` base body font size modularly scaled using minor thirds ~ http://www.modularscale.com/?16&px&1.2

schalkneethling:
r+w/c

stephaniehobson:
Talked with Kadir about this and we're going to go ahead and see if people scream as loudly about the small headers as they did about the big ones.

kuma/static/styles/components/wiki/document-head.scss
stephaniehobson: Removed.

kuma/static/styles/includes/_vars.scss
stephaniehobson: I'm not too worried about the big jump, we don't use h1s for anything other than article titles. It's the closeness of the other ones that I think are a problem :( Anyway, discussed it with Kadir and we'll iterate.

PR #4334, opened 2017-07-27
bug 1370594: Update tests for redesign

jwhitlock:
Fix tests that no longer pass due to the redesign.


stephaniehobson:
Travis is happy 👍

PR #4338, opened 2017-07-28
bug 1385405: Move locale files from kuma to mdn-l10n repo

jwhitlock:
Load the locale files from the new repo, https://github.com/mozilla-l10n/mdn-l10n.

There are some additional tasks to update the document and process for:

* [x] Environment setup
* [x] Deployments
* [x] Docker image creation and automation
* [x] Locale updates

Because of the noise of this change, I'm considering making those part of a new PR.  But first, seeing if TravisCI and GitHub has issues with it.


jwhitlock:
Jenkins is now happy. I think I should get the docs up to date to see if there are any other big changes before I merge this one.

PR #4335, opened 2017-07-27
Bug 1367098: Remove DevRel Survey CSS

stephaniehobson:

PR #4340, opened 2017-07-29
Set UID=1000 for Jenkins jobs

jwhitlock:
Docker commands that generate files need to run as ``UID=1000`` (jenkins), so that Jenkins can delete them when the branch is deleted.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,6 +30,8 @@ node {
     sh 'git submodule sync'
     sh 'git submodule update --init --recursive'
     setGitEnvironmentVariables()
+    // Set UID to jenkins
+    env['UID'] = 1000
 
     // When checking in a file exists in another directory start with './' or
     // prepare to fail.


glogiotatidis:
Interesting. I searched for this recipe and i couldn't find any references of setting `uid` with this method. Can you please more information?

Also curious: with which `uid` are the files created?

jwhitlock:
This is how we set the user in our [docker-compose.yml](https://github.com/mozilla/kuma/blob/bb1353c122921efecbee133c6f21b11b42396e90/docker-compose.yml#L6).

The files are being created with uid 0. You can see the generated files with:

    find /var/lib/jenkins/jobs/mdn_multibranch_pipeline \! -user jenkins -print

metadave:
Is this change to fix files with root perms on the Jenkins server?

jwhitlock:
@metadave yes, worked for this branch (no root files created), but Jenkins builds are broken (PR #4335, #4341) so not sure if generated docker images are correct.  Open to other solutions.

jwhitlock:
Force-push to update on master, which includes locales as a submodule and fixes for the docker build in Jenkins.

jwhitlock:
Manually set ownership so Jenkins could delete some branches:

    find /var/lib/jenkins/jobs/mdn_multibranch_pipeline -depth \! -user jenkins -print -exec sudo chown jenkins:jenkins {} \;

PR #4345, opened 2017-08-01
[Bug 1385315] Use 302 not 301 for locale based redirect

safwanrahman:
We should redirect the visitor with 302 redirect rather than 301
@jwhitlock r?

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -3,8 +3,7 @@ import urllib
 
 from django.conf import settings
 from django.core import urlresolvers
-from django.http import HttpResponseForbidden, HttpResponsePermanentRedirect
-from django.http import HttpResponseRedirect
+from django.http import HttpResponseForbidden, HttpResponsePermanentRedirect, HttpResponseRedirect
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
 from django.contrib.sessions.middleware import SessionMiddleware
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -4,6 +4,7 @@ import urllib
 from django.conf import settings
 from django.core import urlresolvers
 from django.http import HttpResponseForbidden, HttpResponsePermanentRedirect
+from django.http import HttpResponseRedirect
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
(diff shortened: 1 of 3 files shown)


jwhitlock:
Please fix the ``flake8`` error.

kuma/core/middleware.py
```diff
@@ -4,6 +4,7 @@
 from django.conf import settings
 from django.core import urlresolvers
 from django.http import HttpResponseForbidden, HttpResponsePermanentRedirect
+from django.http import HttpResponseRedirect
```
jwhitlock: This could be combined with the above import statement

kuma/core/middleware.py
```diff
@@ -35,7 +36,10 @@ def process_request(self, request):
             new_path = prefixer.fix(prefixer.shortened_path)
             query = dict((smart_str(k), v) for
                          k, v in request.GET.iteritems() if k != 'lang')
-            return HttpResponsePermanentRedirect(urlparams(new_path, **query))
+
+            # Never use HttpResponsePermanentRedirect here.
+            # Its a temporary redirect and should return with http 302, not 301
```
jwhitlock: I'm not sure in this case. I could argue that:

    https://developer.mozilla.org/en-US/docs/Mozilla?lang=de

should be a permanent redirect to:

    https://developer.mozilla.org/de/docs/Mozilla

As far as I can tell from the code and the analytics, the ``?lang=foo`` feature is never used. This path and the one test that exercises it should probably be dropped. But that's a new PR.

So 👍 for now, for consistency.

jwhitlock: Update: it looks like ``?lang=de`` is used in the footer language selector when JS is disabled. And, because JS is disabled, it doesn't show up in Google Analytics 😄

safwanrahman: > `https://developer.mozilla.org/en-US/docs/Mozilla?lang=de`
> 
> should be a permanent redirect to:
> 
> `https://developer.mozilla.org/de/docs/Mozilla`

Like, If I go to `bn-BD` locale for some reason, I dont want to be there permanently.

safwanrahman:
@jwhitlock fixed the nit. r?

jwhitlock:
OK for now. Someday I'll turn on 80-column lines in ``flake8``....

PR #4346, opened 2017-08-02
Bug 1360294 - Add ability for users to hide github link from their public profile

MatonAnthony:
The change is working except for the Signup view, I don't understand how to link the Form to the model on this one.

And there might be some CSS issue with the design, as I'm far for at ease with CSS.

### DIFF
--- a/kuma/static/styles/submission.scss
+++ b/kuma/static/styles/submission.scss
@@ -143,7 +143,8 @@ p.field-note {
     clear: both;
 }
 
-#field-beta {
+#field-beta,
+#field_is_github_url_public {
     @extend %checkbox-label-container;
 }
 
--- a/kuma/users/jinja2/socialaccount/signup.html
+++ b/kuma/users/jinja2/socialaccount/signup.html
@@ -173,7 +173,13 @@
                           {% trans terms_url='https://www.mozilla.org/about/legal/terms/mozilla', privacy_url='https://www.mozilla.org/privacy/websites/' %}
                           to Mozilla's <a href="{{ terms_url }}">Terms</a> and <a href="{{ privacy_url }}">Privacy Notice</a>.
                           {% endtrans %}
-                      </label>
+                        </label>
+
+                      </li>
+                      <li id="github_url_public">
+                        <label for="{{ form.is_github_url_public.id_for_label }}">
+                          {{ form.is_github_url_public }} {{ form.is_github_url_public.label }}
+                        </label>
                       </li>
                   </ol>
                 </fieldset>
--- a/kuma/users/jinja2/users/includes/user_title_info.html
+++ b/kuma/users/jinja2/users/includes/user_title_info.html
@@ -74,7 +74,7 @@
     <li class="twitter"><i class="icon-twitter" aria-hidden="true"></i><a href="{{ detail_user.twitter_url }}" rel="me external" class="url">{{ _("Twitter") }}</a></li>
   {% endif %}
 
-  {% if accounts.github %}
+  {% if accounts.github and user.is_github_url_public %}
     {% for github_account in accounts.github %}{% if github_account.extra_data %}
     <li class="github"><i class="icon-github" aria-hidden="true"></i><a href="{{ github_account.get_profile_url() }}" rel="me external" class="url">{{ _("GitHub") }}</a></li>
     {% endif %}{% endfor %}
--- a/kuma/users/jinja2/users/includes/user_title_info.html
+++ b/kuma/users/jinja2/users/includes/user_title_info.html
@@ -74,7 +74,7 @@
     <li class="twitter"><i class="icon-twitter" aria-hidden="true"></i><a href="{{ detail_user.twitter_url }}" rel="me external" class="url">{{ _("Twitter") }}</a></li>
   {% endif %}
 
-  {% if accounts.github and user.is_github_url_public %}
+  {% if accounts.github and detail_user.is_github_url_public %}
     {% for github_account in accounts.github %}{% if github_account.extra_data %}
     <li class="github"><i class="icon-github" aria-hidden="true"></i><a href="{{ github_account.get_profile_url() }}" rel="me external" class="url">{{ _("GitHub") }}</a></li>
     {% endif %}{% endfor %}
--- a/kuma/users/jinja2/users/user_edit.html
+++ b/kuma/users/jinja2/users/user_edit.html
@@ -12,135 +12,141 @@
 {% block robots_value %}noindex, nofollow{% endblock %}
 
 {% block site_css %}
-  {{ super() }}
-  {% stylesheet 'jquery-ui' %}
-  {% stylesheet 'users' %}
-  {% stylesheet 'submission' %}
+{{ super() }}
+{% stylesheet 'jquery-ui' %}
+{% stylesheet 'users' %}
+{% stylesheet 'submission' %}
 {% endblock %}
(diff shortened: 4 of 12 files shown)


kuma/users/jinja2/socialaccount/signup.html
```diff
@@ -173,7 +173,13 @@
...
                           to Mozilla's <a href="{{ terms_url }}">Terms</a> and <a href="{{ privacy_url }}">Privacy Notice</a>.
                           {% endtrans %}
-                      </label>
+                        </label>
+
+                      </li>
+                      <li id="github_url_public">
+                        <label for="{{ form.is_github_url_public.id_for_label }}">
+                          {{ form.is_github_url_public }} {{ form.is_github_url_public.label }}
+                        </label>
```
stephaniehobson: 👍 Thank you for proper labeling,

kuma/static/styles/submission.scss
```diff
@@ -143,7 +143,8 @@ p.field-note {
     clear: both;
 }
 
-#field-beta {
+#field-beta,
+#field_is_github_url_public {
```
stephaniehobson: This is the right place to add this 👍

jwhitlock:
Please revert the changes to kumascript and locale - those should not be mixed with code pull requests.

kuma/users/models.py
```diff
@@ -152,6 +152,10 @@ class User(AbstractUser):
         blank=True,
         validators=[WEBSITE_VALIDATORS['github']],
     )
+    is_github_url_public = models.BooleanField(
+        _(u'Public Github URL'),
+        default=False,
```
jwhitlock: Is ``False`` the right default value? ``True`` matches current behavior, ``False`` changes to an opt-in. I can see either way. @stephaniehobson ?

MatonAnthony: In my opinion, an opt-in is the right way to do it because Mozilla is privacy focused and displaying the Github profile is a matter of privacy.

safwanrahman:
Is it possible to write some tests so we dont get any regression in future?

kuma/users/jinja2/users/user_edit.html
```diff
@@ -12,135 +12,141 @@
...
+            <label for="{{ user_form.github_url.id_for_label }}"><i class="icon-github" aria-hidden="true"></i>GitHub</label>
+            <div class="field-account">
+              {% for github_account in accounts.github %}{% if github_account.extra_data %}{{ github_account.get_profile_url() }}{% if not loop.last %}, {% endif %}{% endif %}{% endfor %}
+            </div>
+
+            <div id="field_is_github_url_public">
+              <label for="{{ user_form.is_github_url_public.id_for_label }}">
+                {{ user_form.is_github_url_public.label }}
+              </label> {{ user_form.is_github_url_public }}
+            </div>
```
jwhitlock: It's hard to tell because of the indenting change, but the non-indenting change is adding this new field, which looks 👍 . This is why I try to change indenting in once commit, and change code in the next.

kuma/users/jinja2/users/includes/user_title_info.html
```diff
@@ -74,7 +74,7 @@ <h1 class="user-title">
     <li class="twitter"><i class="icon-twitter" aria-hidden="true"></i><a href="{{ detail_user.twitter_url }}" rel="me external" class="url">{{ _("Twitter") }}</a></li>
   {% endif %}
 
-  {% if accounts.github %}
+  {% if accounts.github and user.is_github_url_public %}
```
jwhitlock: ``user`` is the request user. This should be ``detail_user.is_github_url_public``.

The second branch should also be updated, starting at line 82.

http://localhost:8000/en-US/profiles/Sheppy and http://localhost:8000/en-US/profiles/jwhitlock should have GitHub links, which should be not shown after the migration is applied.

kuma/users/signup.py
```diff
@@ -34,7 +34,8 @@ class SignupForm(BaseSignupForm):
     terms = forms.BooleanField(label=_(u'I agree'),
                                required=True,
                                error_messages={'required': TERMS_REQUIRED})
-
+    is_github_url_public = forms.BooleanField(label=_(u'I would like to make my github profile URL public'),
+                                              required=False)
```
jwhitlock: You'll need some more code to save this additional value on the user profile.

``kuma.users.signup.SignupForm`` is derived from django-allauth's [allauth.socialaccount.forms.SignupForm](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/forms.py#L12-L36). This calls:

    user = adapter.save_user(request, self.sociallogin, form=self)

You can see the base version in [allauth.socialaccount.adapter.DefaultSocialAccountAdapter](https://github.com/pennersr/django-allauth/blob/0367b51514e592db011511e420570789c7d1df75/allauth/socialaccount/adapter.py#L70-L82). We override it in [kuma.users.adapters.KumaSocialAccountAdapter](https://github.com/mozilla/kuma/blob/master/kuma/users/adapters.py#L179-L192), but it basically just calls the default version.

That in turn calls [allauth.account.adapter.DefaultAccountAdapter.save_user](https://github.com/pennersr/django-allauth/blob/0367b51514e592db011511e420570789c7d1df75/allauth/account/adapter.py#L221-L248) to actually save the user. We also override this in [kuma.users.adapaters.KumaAccountAdapter](https://github.com/mozilla/kuma/blob/master/kuma/users/adapters.py#L24), but don't customize the ``save_user`` method.

You'll need to add a ``save_user`` method to ``KumaAccountAdapter`` that:

1. Calls the super class's ``save_user`` method with ``commit=False``
2. Adds the ``is_github_url_public`` setting from ``form.cleaned_data``
3. Saves the user
4. Returns the user.

Plus tests.

kuma/users/models.py
jwhitlock: I believe @stephaniehobson has verbally confirmed that ``default=False`` is the right value, and requiring existing users to opt-in again is OK.

kuma/users/signup.py
MatonAnthony: I have a feeling that tests shouldn't write into the database.

Do I need to mock the save_user() to avoid writing into the Db or do we have a specific db for test which is reset everytime ?

I must admit that I don't really know how to start writing this test.

jwhitlock: If you get the code working, I'll add tests in a new commit. Or, I can add failing tests and then you can write the code to make them work 😄 

Tests run against a nearly empty copy of the database, and are wiped out between tests. Writing to the database slows down the tests, but it is safe, and we'd need more serious refactoring to move away from database writes in tests.

MatonAnthony: I finally managed to get some time to make the code work properly.

I'm sorry it took so long, I'm still trying to understand how I should do the tests.

I guess the needed tests are : 
- [ ] A test for the registration process in itself for the `save_user()` [kuma.users.adapaters.KumaAccountAdapter.save_user](https://github.com/mozilla/kuma/blob/master/kuma/users/adapters.py#L24) in [KumaSocialAccountAdapterTestCase](https://github.com/mozilla/kuma/blob/master/kuma/users/tests/test_adapters.py#L15)
- [ ] A functional test for the registration process in https://github.com/mozilla/kuma/tree/master/tests/functional 
- [ ] A functional test for the profile display in [test_profiles](https://github.com/mozilla/kuma/blob/master/tests/functional/test_profiles.py)

PR #4348, opened 2017-08-02
Bug 1370594: New page headers, new zone headers, nav tweaks

stephaniehobson:
I ended up having to make changes to the zones to support the nav and page header changes. This was planned eventually anyway, it just happened a bit sooner. 

**Nav changes:**
- Move blue border to top of global nav
  - Add box shadow under it

**Zone changes:**
- Make zone landing pages match other article landing pages (mostly)
  - Remove breadcrumbs from zone root pages
- Remove zone specific article footers
- Add new zone icons
- Remove zone coloured background
  - Update code which relied on there being a background
- Update associated tests

**Page changes:**
- Created new document_head block
  - Apply to article view/edit/move/translate/create pages
  - Add blue background to document head
  - Decreased mobile title size when inside .document-head
- Line up buttons and crumb horizontally
- Tweak mobile and tablet display
- Styles apply to non-wiki pages to moved styles to main.css

Testing notes:
- This change affects wiki pages, wiki landing pages, zone pages, zone landing pages, edit, move, translate, and create pages, and the revision dashboard (and should not affect other pages)
-  Tagging @escattone to review the test updates


stephaniehobson:
@schalkneethling I can submit just the zone commit separately for review if that would make it easier but the nav and header updates rely on that work to be done first.

schalkneethling:
There is currently a flake8 error present:
https://travis-ci.org/mozilla/kuma/jobs/260413234#L559

schalkneethling:
This might be by design but, the book icon on `/Learn` is coloured and sized differently than on `Getting_started_with_the_web`. Thought I would mention it in case it is not intentional

screen shot 2017-08-03 at 16 18 10

screen shot 2017-08-03 at 16 17 51

schalkneethling:
Also, probably by design, but the `box-shadow` below the nav bar looks a little odd on the right hand side on large viewports. Feels like the "Sign in" widget should be aligned all the way to the right of the viewport.

screen shot 2017-08-03 at 15 48 28

schalkneethling:
Curious, is the background colour for the pager numbers correct?

screen shot 2017-08-03 at 16 29 47

schalkneethling:
On the revision dashboard, when comparing to edits, the copy starts to overlap on medium(tablet) sized viewport sizes:

screen shot 2017-08-03 at 16 31 21

Perhaps they should be stacked at this width?

schalkneethling:
On small viewports (roughly large mobile), the header layout breaks a bit.

screen shot 2017-08-03 at 16 37 26

schalkneethling:
Also, at 480px - 769px the icon position if offset:

screen shot 2017-08-03 at 16 26 41

schalkneethling:
When the content is laid out rtl and on a mobile viewport, the `x` overlaps the top navigation item

screen shot 2017-08-03 at 18 40 06

schalkneethling:
Probably not related to this PR, and might be a known issue but, with rtl languages the translation edit pages kinda falls apart when the viewport is resized. Even at 1200px, especially the editor section body, and header sizes independently.

screen shot 2017-08-03 at 18 46 37

schalkneethling:
In IE11 the three grey blocks are not getting centred at smaller viewports. This causes the right side of the the last block to be cut of, and no horizontal scroll bars are added.

virtualbox_windows 8 1 - ie11_03_08_2017_19_02_27

stephaniehobson:
Large grey zone icons
The zone icon is larger and darker on the zone landing page. On the sub pages clicking the icon will take you to the landing page, on the landing page there's no need to click it. I've edited the HTML to remove the link altogether now.

Box shadow
I've updated the homepage background to span the width of the screen now, like the article header backgrounds do. I think this counters the weirdness of the black in the nav bar a bit. This let me get rid of the negative margin hack on that element, so yay!

Page numbers
Replaced another hard coded colour with a variable name 🎉 

Overlapping comparison
Probably needs to be addressed elsewhere.

Mobile layout of article header
That is displaying as intended but I'm concerned that it looks broken to you 😞 I struggled a bit with this breakpoint because the spacing looks off unless both the parent article and the current article have long titles. Would left aligning things make it look more intentional?

Icon position on tablet
Fixed by making it the correct size.

Close button in submenu in RTL
Fixed.

Edit pages
Needs to be addressed separately.

Grey boxes on HTML page
Needs to be addressed separately.  Seems like a IE/flexbox bug 😣

kuma/static/styles/includes/_mixins.scss
```diff
@@ -365,7 +364,7 @@ These are not dynamic but serve as mixins
     @include set-font-size(22px);
     border: 5px solid $accent-light;
     border-width: 5px 0 5px 0;
```
schalkneethling: Nit: Could be shortened to `border-width: 5px 0;`?

kuma/static/styles/components/structure/main-header.scss
```diff
@@ -4,13 +4,17 @@ Main header, wraps all components of logo, main, and secondary navigation
 
 #main-header {
     @include clearfix();
+    position: relative;
+    z-index: 1;
     padding-top: $mobile-center-spacing;
+    border: 5px solid $accent-light;
+    border-width: 5px 0 0 0;
```
schalkneethling: Nit: Could be shortened to `border-width: 5px 0 0;`?

schalkneethling:
> Would left aligning things make it look more intentional?

Would definitely work better I believe.

schalkneethling:
Other than the last couple of small items, this is looking great r+w/c ☕️

stephaniehobson:
Nits addressed. 

Looked at moving language switcher to left on mobile but it required too much fighting with some global styles. (This part of the code needs to be refactored, too many un-related components are using the same class). Over all this is still an improvement for mobile.

PR #4357, opened 2017-08-09
bug 1385405: Add doc for deploying to SCL3

jwhitlock:
We're adding a pre-push tasks of updating the locales submodule, which gives a good opportunity to document our push proceedures.

### DIFF
--- a/docs/deploy-scl3.rst
+++ b/docs/deploy-scl3.rst
@@ -49,7 +49,18 @@ Pre-Deployment: Update Submodules
 
   This will add references to the PRs back to the commit where there were
   deployed. You can see the commits from the `What's Deployed`_ page, or by
-  using `GitHub's compare view`_.
+  using `GitHub's compare view`_::
+
+    # This long command is needed once
+    # On OSX, change "echo" to "open" to open in browser
+    git config --add alias.submodule-compare-urls "submodule foreach --quiet 'echo \"https://github.com/\`git remote get-url origin | cut -d: -f2 | cut -d. -f1\`/compare/\$sha1...\`git rev-parse @\`\"'"
+
+    # And then can be used to print the URLs
+    git submodule-compare-urls
+
+  You can just see the merge commits with ``git log --merges``::
+
+    git submodule foreach 'git log --merges HEAD...$sha1'
 
 .. _`What's Deployed`: https://whatsdeployed.io/s-FHK
 .. _`GitHub's compare view`: https://github.com/blog/612-introducing-github-compare-view
--- a/docs/deploy-scl3.rst
+++ b/docs/deploy-scl3.rst
@@ -32,15 +32,7 @@ Pre-Deployment: Update Submodules
 
 * Update the submodules::
 
-    git submodule update --init
-    cd kumascript
-    git fetch origin
-    git checkout origin/master
-    cd ..
-    cd locale
-    git fetch origin
-    git checkout origin/master
-    cd ..
+    git submodule update --remote
     git diff
     git commit kumascript locale
 
--- a/docs/deploy-scl3.rst
+++ b/docs/deploy-scl3.rst
@@ -1,0 +1,208 @@
+=================
+Deploying to SCL3
+=================
+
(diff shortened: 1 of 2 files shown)


jwhitlock:
I think @escattone or @stephaniehobson could review this, but I didn't want to imply that both needed to.

safwanrahman:
A really good approach to have it documented in open!
Maybe, it used to be documented in mana for staff only!

docs/deploy-scl3.rst
```diff
@@ -0,0 +1,208 @@
...
+pushing to production.
+
+* Start the staging push. There are two methods for pushing, the chief website
+  (a.k.a the Big Red Button) and the command line tool james_. Both require VPN
+  access, and a commit hash you want to push.  ``james`` defaults to your current working
+  copy::
+
+    git fetch origin
+    git checkout origin/master
+    james stage
```
safwanrahman: I think, this need to be run in docker or virtualenv if `chief-james` is not installed in the local machine. Or you can specify that `pip install chief-james`

docs/deploy-scl3.rst
```diff
@@ -0,0 +1,208 @@
...
+
+    git remote -v | grep origin
+    # origin  git@github.com:mozilla/kuma.git (fetch)
+    # origin  git@github.com:mozilla/kuma.git (push)
+    git fetch origin
+    git checkout -b pre-push-2017-08-08 origin/master
+
+* Update the submodules::
+
+    git submodule update --init
```
safwanrahman: I think, if you just specify the branch in `.gitmodules`, and run `git submodule update --remote`, git will update all the submodule to that specific branch.

docs/deploy-scl3.rst
```diff
@@ -0,0 +1,208 @@
...
+    git checkout origin/master
+    cd ..
+    cd locale
+    git fetch origin
+    git checkout origin/master
+    cd ..
+    git diff
+    git commit kumascript locale
+
+  For bonus points, the commit message should mention the KumaScript PRs::
```
safwanrahman: You can also add how to check the merge log. Like `git log --merges`

docs/deploy-scl3.rst
jwhitlock: I want to leave setup details out of this document, and only have the checklist steps. I don't expect new staff members to get deployments working on their own, but have an existing staff member help with the initial setup.

docs/deploy-scl3.rst
jwhitlock: Interesting. I haven't used ``--remote`` or ``--branch`` in the past, so I'll have to play around with it and see if it does The Right Thing 👀

docs/deploy-scl3.rst
jwhitlock: nice tip, I did not know about the ``--merges`` option.

jwhitlock:
Updated docs for @safwanrahman's recommendations, and learned a little about ``git submodule foreach``.

docs/deploy-scl3.rst
```diff
@@ -0,0 +1,211 @@
...
+
+    git remote -v | grep origin
+    # origin  git@github.com:mozilla/kuma.git (fetch)
+    # origin  git@github.com:mozilla/kuma.git (push)
+    git fetch origin
+    git checkout -b pre-push-2017-08-08 origin/master
+
+* Update the submodules::
+
+    git submodule update --remote
```
safwanrahman: You need to add the branch in `.gitmodules`

jwhitlock: We want ``origin/master``, which is also [the default](https://git-scm.com/docs/git-submodule#git-submodule---remote), so we don't need it:

> --remote
> This option is only valid for the update command. Instead of using the superproject’s recorded SHA-1 to update the submodule, use the status of the submodule’s remote-tracking branch. The remote used is branch’s remote (branch.<name>.remote), defaulting to origin. The remote branch used defaults to master, but the branch name may be overridden by setting the submodule.<name>.branch option in either .gitmodules or .git/config (with .git/config taking precedence).

PR #4358, opened 2017-08-09
bug 1376606: Update to New Relic v2 API

jwhitlock:
Update ping_newrelic to use the v2 API:
https://docs.newrelic.com/docs/apm/new-relic-apm/maintenance/recording-deployments

I've manually tested the ``curl`` command, and I'd like to push to staging to confirm it work.


jwhitlock:
Tested on staging, we have NR deployments again.

PR #4359, opened 2017-08-10
bug 1388933: allow MEDIA_ROOT to be overridden from the environment

escattone:
This PR allows the MEDIA_ROOT setting to be overridden from the environment, as a convenience for the MDN->AWS work (see https://github.com/mozmeao/infra/issues/197#issuecomment-321411851). Currently it is hard-wired to the `media` folder under the application-root directory, and that directory within Kubernetes would be ephemeral since it's part of the container. This will allow us to point the MEDIA_ROOT to a location on the shared PVC (EFS).


metadave:
👍 , thanks @escattone!

PR #4360, opened 2017-08-10
Adds tmp function to hide sidebar and expand content section for user testing

schalkneethling:
@jwhitlock Here you go, this is to temporarily hide the sidebar and expand the content area for today's user testing.

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -883,4 +883,11 @@
         }
     }
 
+    function tmpHideSidebar() {
+        $('#wiki-right').hide();
+        $('#wiki-content').css('width', '70%');
+    }
+
+    tmpHideSidebar();
+
 })(window, document, jQuery);


jwhitlock:
This looks good to me, pushing it to staging for user testing.

In the long term (2-4 months), we'd like to use the demo deployments for these kinds of experiments. I think we need a mechanism for adding additional pages not in the sample DB.

stephaniehobson:
For future reference, it is possible to remove a side bar from a page by editing the page's properties.

schalkneethling:
> For future reference, it is possible to remove a side bar from a page by editing the page's properties.

That is definitely a better approach. Is that documented somewhere? Also, @jwhitlock, as this was a temporary thing, I guess you can close this and back it out from staging?

jwhitlock:
I don't think it hurts to leave it there until the next production push. I'm planning for @escattone or @stephaniehobson to do the next one, to test the instructions in PR #4357.

stephaniehobson:
It was (poorly) documented [here](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Howto/Create_and_edit_pages#Table_of_Contents). I've updated it. It's one of those features it's hard for the people who've been using it forever to remember not everyone knows about ;)

jwhitlock:
This code isn't destined for master, closing.

PR #4361, opened 2017-08-11
Set interactive editor iframe height based on postMessage from editor

schalkneethling:
This will listen for a `postMessage` event from the interactive editor with the `scrollHeight` of the CSS editor container [1]. 

To ensure arbitrary code is not run, this code checks

1. That the origin match and
2. That the message data contains the property `iFrameHeight`

If this is satisfied, the height of the `iframe` will be updated to the passed value. The editor will send a follow up `postMessage` when a specific `matchMedia` breakpoint[2] is reached.

[1] https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/mce-utils.js#L77
[2] https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/mce-utils.js#L86

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -883,4 +883,17 @@
         }
     }
 
+    // listens for post message from interactive editor
+    window.addEventListener('message', function(event) {
+        // get the interactive editor iframe
+        var iframe = document.querySelector('iframe.interactive');
+        var isExpectedOrigin = event.origin === 'https://interactive-examples.mdn.mozilla.net';
+
+        /* there may be other post messages so, ensure that the origin is the
+        expected and, that `event.data` contains an `iframeHeight` property */
+        if (isExpectedOrigin && event.data.iframeHeight) {
+            iframe.setAttribute('height', event.data.iframeHeight);
+        }
+    }, false);
+
 })(window, document, jQuery);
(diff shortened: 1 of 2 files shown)


schalkneethling:
> I'd like to see this code updated to handle multiple iframes per page but it's not a blocker.

I will make a note of this.

> so if it's blocking something we can merge and roll forward if there's a problem.

That would be great. I would like to test this in a page context to ensure it works as expected and so that we can tweak it as needed.

kuma/static/js/wiki.js
```diff
@@ -883,4 +883,17 @@
         }
     }
 
+    // listens for post message from interactive editor
+    window.addEventListener('message', function(event) {
+        // get the interactive editor iframe
+        var iframe = document.querySelector('iframe.interactive');
```
schalkneethling: Indeed. Currently we can be confident that there will only be one iframe with the class `interactive`, even if there are two somehow, the one I am targeting will most definitely be the first in the DOM so, this will grab that one.

With all of that said though, as we will currently(and for the immediate future) only have one per page, perhaps it is worth adding an id to the iframe? Something like `id="interactive-editor"`

stephaniehobson: Let's not add an ID, that will only make it harder to adapt the code to allow multiple instances of the interactive examples per page.

PR #4363, opened 2017-08-14
bug 1390267: allow ATTACHMENT_HOST to be set from env

escattone:
Allow the ATTACHMENT_HOST to be set from the environment, which helps with MDN->AWS development.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,7 @@ services:
       # Django settings overrides:
       - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
       - ALLOWED_HOSTS=*
+      - ATTACHMENT_HOST=localhost:8000
       - BROKER_URL=redis://redis:6379/0
       - CELERY_ALWAYS_EAGER=False
       - CSRF_COOKIE_SECURE=False
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1153,7 +1153,7 @@ ALLOWED_HOSTS = config('ALLOWED_HOSTS',
 MAX_FILENAME_LENGTH = 200
 MAX_FILEPATH_LENGTH = 250
 
-ATTACHMENT_HOST = 'mdn.mozillademos.org'
+ATTACHMENT_HOST = config('ATTACHMENT_HOST', default='mdn.mozillademos.org')
 
 # Video settings, hard coded here for now.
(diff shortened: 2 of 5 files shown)


metadave:
infra task [here](https://github.com/mozmeao/infra/issues/197)

safwanrahman:
Can you add this into the `docker-compose.yml` also? So we can change the thing if its needed locally!

escattone:
Thanks @metadave and good catch @safwanrahman !

docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
       # Django settings overrides:
       - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
       - ALLOWED_HOSTS=*
+      - ATTACHMENT_HOST=localhost:8000
```
jwhitlock: The same issue appears in the master branch, so this change doesn't appear to be the cause. I can't recall if I've tried uploads in the docker dev environment. Bug needed.

escattone:
Thanks for your review @jwhitlock. I'll try this again locally and file a bug.

escattone:
@jwhitlock I couldn't reproduce the bug you're seeing with uploads in my local Docker-based dev environment (set to the `master` branch). I tried uploading a JPG file on both https://localhost:8000/en-US/Add-ons/WebExtensions/Debugging and http://localhost:8000/en-US/docs/Web/HTML, and was able to successfully upload as well as add a link to the uploaded image in both docs. Perhaps, I'm missing something or doing something differently?

jwhitlock:
I'm using a recent anonymized production database. Maybe there is something in there that is interfering.  I can try again w/ the sample database, once I'm done with the work needing the full data set.

jwhitlock:
I retried this on the sample database, and had the same issue, so I looked for another source of the problem.  I'm using an alternate config for the ``web`` container, as described in the docs, [Customizing the Docker Environment](https://kuma.readthedocs.io/en/latest/development.html#customizing-the-docker-environment):

```
version: "2.1"
services:
  web:
    command: ./manage.py runserver_plus 0.0.0.0:8000
    stdin_open: true
    tty: true
```

This allows me to set breakpoints and step through the code.  However, it also restricts the web server to a single process. This may have been why the server locked up.

When I removed this and used the standard development environment, uploading an image was fast with no problems.

PR #4364, opened 2017-08-15
bug 1390381: Change default page title to "MDN Web Docs"

dylanpyle:
https://bugzilla.mozilla.org/show_bug.cgi?id=1390381

🤷‍♂️

### DIFF
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -10,7 +10,7 @@
   {% if not is_sphinx %}
     {{ optimizely_script() }}
   {% endif %}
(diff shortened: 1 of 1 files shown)


stephaniehobson:
Good eye, this definitely need changing. I think we might just want to go to "MDN" though. This is a good conversation for the bug and I added a comment there too.

stephaniehobson:
I guess we can always change it again if we decide on something else. This is an improvement and should go to production.

Thanks for you contribution @dylanpyle 

👍

PR #4365, opened 2017-08-17
bug 1391092: add ability to serve legacy files from Django (for AWS)

escattone:
This PR gives Kuma the ability to serve the legacy diagrams, presentations, and samples, and it's target use is for MDN in AWS. By default it is not enabled, so it will not affect the SCL3 deployment which already serves these legacy files via Apache. The legacy files are so rarely requested, and hopefully will be removed from MDN altogether in the future, so it doesn't seem worth the effort to serve these files in a more production-like manner.

For the AWS work, this feature can be enabled by setting two environment variables:
```
SERVE_LEGACY=True
LEGACY_ROOT=/mdn/www
```

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -367,6 +367,9 @@ STATIC_ROOT = path('static')
 
 SERVE_MEDIA = False
 
+SERVE_LEGACY = config('SERVE_LEGACY', default=False)
+LEGACY_ROOT = config('LEGACY_ROOT', default=None)
+
 # Paths that don't require a locale prefix.
 LANGUAGE_URL_IGNORED_PATHS = (
     'healthz',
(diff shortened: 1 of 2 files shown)


escattone:
See https://github.com/mozmeao/infra/pull/431, which depends on this PR.

kuma/settings/common.py
```diff
@@ -367,6 +367,9 @@ def enable_candidate_languages():
 
 SERVE_MEDIA = False
 
+SERVE_LEGACY = config('SERVE_LEGACY', default=False)
```
jwhitlock: Please add a short comment for this section, like "serve diagrams, presentations, and samples from 2005-2012".

``SERVE_LEGACY`` should include ``cast=bool``.
* Add

kuma/settings/common.py
```diff
@@ -383,6 +386,9 @@ def enable_candidate_languages():
     '__debug__',
     '.well-known',
     'users/github/login/callback/',
+    'diagrams',
+    'presentations',
+    'samples',
```
jwhitlock: Good catch!

PR #4368, opened 2017-08-18
Bug 1370594: Lint and clean files

stephaniehobson:
Going to edit these, want them nice and shiny first.


kuma/static/js/wiki.js
```diff
@@ -148,7 +154,7 @@
     /*
         Syntax highlighting scripts
     */
-    if($('article pre').length && ('querySelectorAll' in doc)) (function() {
+    if($('article pre').length && ('querySelectorAll' in doc)) {
```
stephaniehobson: I'm not sure if I did the right thing here. Should I have preserved the `(function(){})();`

kuma/static/js/wiki.js
```diff
@@ -24,7 +24,9 @@
         var $subnavList = $subnav.find(' > ol');
         var minHeightFn = $('.zone-landing-header-preview-base').length ? setMinHeight : noop;
 
-        if(!$subnavList.length) return; // Exit if the subnav isn't set up properly
+        if(!$subnavList.length) {
```
schalkneethling: Nit: Do we want to do a space between the `if` and the open `(`?

kuma/static/js/wiki.js
```diff
@@ -651,14 +466,18 @@
             });
 
             if(timeoutFlag) {
-                if(timer) clearTimeout(timer);
+                if(timer) {
+                    clearTimeout(timer);
+                }
             }else{
```
schalkneethling: Nit: Also here, spaces around the `else`

kuma/static/js/wiki.js
schalkneethling: Yes, it looks like you are going to want to keep it. But you can nest it inside the outer if statement

```
if ($('article pre').length && ('querySelectorAll' in doc)) {
    (function() {
        ...
    })();
}
```

PR #4369, opened 2017-08-18
Removes interactive editor CSS from Kuma

schalkneethling:
@stephaniehobson As the resizing of the iframe now happens via JS, between the editor and Kuma via `pushMessage`, we no longer need these styles.


stephaniehobson:
@schalkneethling We need default values to prevent the page content from jumping around as the user reads, or in case they do not have javascript enabled. Since there is a common size for most JS examples and we have a started goal of keeping the CSS ones below 5 examples I think we can come up with estimates.

We should also remove the ability to pass a class in from the macro since it will no longer have any effect.

schalkneethling:
Closing in favour of https://github.com/mozilla/kuma/pull/4379

PR #4371, opened 2017-08-18
bug 1391092: Add ability to serve legacy files from Django

jwhitlock:
Fixes nits from PR #4365.


escattone:
Thanks @jwhitlock !

PR #4373, opened 2017-08-19
Bug 1370594: Outline sample code

stephaniehobson:
Add a border around live sample code.

- This content looks different from other page content because it is in an iframe, there should be signal  for that
- Borrowed the border colour from the interative-examples
- Renamed .scss file to better indicate what it pertains to

### DIFF
--- a/kuma/static/styles/components/wiki/sample-code.scss
+++ b/kuma/static/styles/components/wiki/sample-code.scss
@@ -1,3 +1,13 @@
+.sample-code-frame {
+    border: 1px solid #9b9b9b;
+    padding: $grid-spacing;
+
+    .sample-code-table & {
+        padding: 0;
+        border: 0;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


schalkneethling:
@stephaniehobson Which page can I use to test this? There probably isn't a page available by default in the sample DB but, I can create it locally. Thanks.

stephaniehobson:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#Examples

stephaniehobson:
(it appears on the page just above the "open in CodePen" button.

schalkneethling:
I have a slight concern that the contrast of the border on the white background is very low. So much so, that it is almost not visible. I wonder if it would be worth to darken it a bit? @stephaniehobson 

Other than that r+

stephaniehobson:
Made the outline a darker grey.

PR #4374, opened 2017-08-19
bug 1391986

SphinxKnight:
Adding `:z` so that SELinux labels are properly updated and docker-compose succeeds for systems based on Fedora/RHEL.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     command: ./manage.py celery worker --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
     user: ${UID:-33}
     volumes:
-      - ./:/app
+      - ./:/app:z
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     command: ./manage.py celery worker --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
     user: ${UID:-33}
     volumes:
-      - ./:/app
+      - ./:/app:z
```
jwhitlock: From [the docs](https://docs.docker.com/engine/admin/volumes/bind-mounts/#configure-the-selinux-label):

> If you use selinux you can add the z or Z options to modify the selinux label of the host file or directory being mounted into the container
> The z option indicates that the bind mount content is shared among multiple containers.
> The Z option indicates that the bind mount content is private and unshared.

PR #4375, opened 2017-08-20
Fix bug 1389934

SphinxKnight:
The hidden class is now displayed in the "english" panel when editing a translation of an article.


stephaniehobson:
👍

PR #4376, opened 2017-08-21
bug 1391084: Switch tag names from utf8_distinct_ci to utf8_general_ci 

jwhitlock:
Drops the one test that depends on ``utf8_distinct_ci`` custom collation, and then:

1. Migrates tag names that would collide. The oldest tag retains its name, and the newer tags have  `(2)`, `(3)` appended to their name.
2. Migrates ``name`` columns to use ``utf8_general_ci`` instead of ``utf8_distinct_ci``

This migration runs in 7-10 seconds locally, so I don't think we need to schedule downtime to run it.  I've placed the output of the forwards and reverse migrations in [a gist](https://gist.github.com/jwhitlock/7b18b62fef27fc77a3b71a25d4bf3a71).

### DIFF
--- a/kuma/wiki/migrations/0034_utf8_general_ci_tags.py
+++ b/kuma/wiki/migrations/0034_utf8_general_ci_tags.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import logging
+
+logger = logging.getLogger('django.db.backends.schema')
+
+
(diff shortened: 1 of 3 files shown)


escattone:
Thanks @jwhitlock! I have never before used `COLLATE` in SQL, so that was interesting.

jwhitlock:
Me neither. The MySQL docs are great, and a big help with this task.

PR #4377, opened 2017-08-22
bug 1362026: Add interactive-examples to iframes

jwhitlock:
Add https://interactive-examples.mdn.mozilla.net to the allowed sources for iframes.  Re-arrange the long regex string so it is easier to add new items at the end.


jwhitlock:
@stephaniehobson, this is the permanent version of the change you've already made to stage / prod.  I tested with https://developer.mozilla.org/en-US/docs/Experiment%3AInteractiveEditor/Array.prototype.push%28%29.

schalkneethling:
Looks good to me. Thanks @jwhitlock

PR #4378, opened 2017-08-22
bug 1392457, 1274874: Improve document list views

jwhitlock:
Improve special views that return a list of documents:

* Reduce data returned from database to that used in the page
* Eliminate second ``COUNT(*)`` query
* Remove redirects, user pages, and archive pages from the "needs review" and "needs translation" lists.

To Do:
* [ ] Add tests for manager methods

This is not an AWS blocker, so not prioritizing getting this reviewed or merged.


jwhitlock:
Not an AWS blocker, closing for now

PR #4380, opened 2017-08-23
Fix issue #229, styling rules for interactive editor survey link

schalkneethling:
Adds a bit of styling for the survey link added in https://github.com/mdn/kumascript/pull/291

### DIFF
--- a/kuma/static/styles/components/wiki/interactive.scss
+++ b/kuma/static/styles/components/wiki/interactive.scss
@@ -1,7 +1,6 @@
 .interactive {
     height: 600px;
     width: 100%;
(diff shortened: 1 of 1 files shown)


kuma/static/styles/components/wiki/interactive.scss
```diff
@@ -12,6 +11,12 @@
     height: 665px;
 }
 
+.interactive-editor-survey {
+    margin: 0 auto;
```
stephaniehobson: We should add the `margin-bottom: $grid-spacing` here to keep space between the editor and the next page elements.

schalkneethling:
@stephaniehobson updated.

PR #4231, opened 2017-05-10
Bug 1362438: additional redirects from SCL3 Apache to Django

metadave:
source:
https://github.com/mozilla/kuma/blob/master/configs/htaccess

The remaining rules to address from [htaccess](https://github.com/mozilla/kuma/blob/master/configs/htaccess
) are in [this gist](https://gist.github.com/metadave/cea197f9333768a3df20be8af33386b7) for now.

- Rewrite rules that don't redirect (typically annotated with `[L]` in this file) will most likely need to be handled by Whitenoise or some other method:

```
# Links to FTP'ed code samples and examples
RewriteRule ^patches(.*) data/www/patches$1 [L]
RewriteRule ^presentations(.*) data/www/presentations$1 [L]
RewriteRule ^samples(.*) data/www/samples$1 [L]
RewriteRule ^diagrams(.*) data/www/diagrams$1 [L]
RewriteRule ^web-tech(.*) data/www/web-tech$1 [L]
RewriteRule ^css(.*) data/www/css$1 [L]


# Rewrites to robots & sitemaps
RewriteRule ^robots.txt$ media/robots.txt [L]
RewriteRule ^sitemap.xml$ media/sitemap.xml [L]
RewriteRule ^sitemaps/([\w\-]*)/sitemap.xml$ media/sitemaps/$1/sitemap.xml [L]
```

- These seem to be broken:

```
# Miscellaneous off-site redirects
# broken
RewriteRule ^contests/$ http://labs.mozilla.com/contests/extendfirefox/ [R=302]
# broken
RewriteRule ^contests/extendfirefox(/.*)? http://labs.mozilla.com/contests/extendfirefox$1 [R=302]
# http://wiki.ecmascript.org doesn't seem to respond.
RewriteRule ^es4(/.*)?$ http://wiki.ecmascript.org/ [R]
```

- These remain unimplemented:

```
RewriteCond %{QUERY_STRING} ^title=http [NC]
RewriteRule ^index\.php$ / [L,NC]

RewriteCond %{QUERY_STRING} ^title=(.*)$ [NC]
RewriteRule ^index\.php$ %1 [R=301,L,NC]

# HACK: Django will eventually redirect the user to the right spot, but skip a
# couple of redirects for these known legacy locales
RewriteRule ^en/(.*)$     /mwsgi/en-US/$1 [L,QSA,NE,NC]
RewriteRule ^cn/(.*)$     /mwsgi/zh-CN/$1 [L,QSA,NE,NC]
[...]

PR #4381, opened 2017-08-23
bug 1362438: Change handling of some legacy paths

jwhitlock:
Apache in SCL3 handles some legacy paths from earlier versions of MDN, and many of these are incorrect. This PR updates the Apache rules, and creates Django-based redirects for some.

A fuller analysis of these URLs is on [PR 4231](https://github.com/mozilla/kuma/pull/4231#issuecomment-324359728).

AWS blocker, tracked in https://github.com/mozmeao/infra/issues/436

### DIFF
--- a/kuma/redirects/redirects.py
+++ b/kuma/redirects/redirects.py
@@ -756,4 +756,8 @@ redirectpatterns = [
     # Legacy off-site redirects (bug 1362438)
     # RewriteRule ^contests/ http://www.mozillalabs.com/ [R=302,L]
     redirect(r'^contests', 'http://www.mozillalabs.com/', permanent=False),
+
+    # RewriteRule ^es4 http://www.ecma-international.org/memento/TC39.htm [R=302,L]
+    redirect(r'^es4', 'http://www.ecma-international.org/memento/TC39.htm',
+             permanent=False),
 ]
--- a/kuma/redirects/redirects.py
+++ b/kuma/redirects/redirects.py
@@ -752,4 +752,8 @@ redirectpatterns = [
         r'^?demos',
         '/docs/Web/Demos_of_open_web_technologies',
         permanent=True),
+
+    # Legacy off-site redirects (bug 1362438)
+    # RewriteRule ^contests/ http://www.mozillalabs.com/ [R=302,L]
+    redirect(r'^contests', 'http://www.mozillalabs.com/', permanent=False),
 ]
--- a/tests/redirects/map_301.py
+++ b/tests/redirects/map_301.py
@@ -318,4 +318,12 @@ LEGACY_URLS = flatten((
              status_code=404),
     url_test('/index.php?title=En/HTML/Canvas&revision=11',
              status_code=404),
+    url_test('/index.php?title=En/HTML/Canvas&revision=11',
+             status_code=404),
+    url_test('/patches', status_code=404),
+    url_test('/patches/foo', status_code=404),
+    url_test('/web-tech', status_code=404),
+    url_test('/web-tech/feed/atom/', status_code=404),
+    url_test('/css/wiki.css', status_code=404),
+    url_test('/css/base.css', status_code=404),
 ))
--- a/tests/redirects/map_301.py
+++ b/tests/redirects/map_301.py
@@ -2,7 +2,7 @@ from utils.urls import flatten, url_test
 
 import requests
 
-URLS = flatten((
+REDIRECT_URLS = flatten((
     url_test("/media/redesign/css/foo-min.css",
              "/static/build/styles/foo.css"),
     url_test("/media/css/foo-min.css", "/static/build/styles/foo.css"),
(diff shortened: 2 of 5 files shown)


kuma/redirects/redirects.py
```diff
@@ -752,4 +752,30 @@ def locale_redirect(pattern, to, prepend_locale=True, **kwargs):
...
+    redirect(r'^contests/', 'http://www.mozillalabs.com/', permanent=False),
+
+    # RewriteRule ^es4 http://www.ecma-international.org/memento/TC39.htm [R=302,L]
+    redirect(r'^es4', 'http://www.ecma-international.org/memento/TC39.htm',
+             permanent=False),
+
+
+    # bug 962148 - legacy locales served content instead of redirect
+    # RewriteRule ^en/(.*)$     /mwsgi/en-US/$1 [L,QSA,NE,NC]
+    redirect(r'^en/(?P<path>.*)$', '/en-US/{path}',
```
safwanrahman: We handle locale redirect already with `LocaleMiddleware`. So why should we implement this redirect here?

kuma/core/tests/test_locale_middleware.py
```diff
@@ -35,9 +35,14 @@ def test_specificity(self):
         self.assertRedirects(reponse, '/fr/', status_code=302)
 
     def test_partial_redirect(self):
-        """Ensure that /en/ gets directed to /en-US/."""
+        """
+        Ensure that /en/ gets directed to /en-US/.
+
+        This was handled in the LocaleURLMiddleware (302), but is now handled
```
safwanrahman: I do not like to handle this redirect with Redirect rule. As all the locale redirect is handled by  `LocaleURLMiddleware`, this `en` to `en-US` should also be handled by middleware.

kuma/redirects/redirects.py
jwhitlock: I think it is cleaner to handle all the legacy redirects in redirects library. I have a long-term goal of completing the migration from MindTouch to Kuma, so I'm 👎 on continuing to maintain the MT locales in their own middleware. But, retiring MindTouch is not an AWS blocker, so removing this commit from the PR.

jwhitlock:
I've moved the [bug 962148](https://bugzilla.mozilla.org/show_bug.cgi?id=962148) commit to new branch [legacy-locales-962148](https://github.com/mozilla/kuma/compare/master...jwhitlock:legacy-locales-962148?expand=1). @safwanrahman's solution using the ``LocaleMiddleware`` is being developed in PR #4387.

tests/redirects/map_301.py
```diff
@@ -310,3 +310,29 @@
...
+    url_test('/patches', status_code=404),
+    url_test('/patches/foo', status_code=404),
+    url_test('/web-tech', status_code=404),
+    url_test('/web-tech/feed/atom/', status_code=404),
+    url_test('/web-tech', status_code=404),
+    url_test('/web-tech/feed/atom/', status_code=404),
+    url_test('/css/wiki.css', status_code=404),
+    url_test('/css/base.css', status_code=404),
+    url_test('/contests/extendfirefox/faq.php', 'http://www.mozillalabs.com/',
+             status_code=302),
```
escattone: Maybe add test with no trailing slash as well?
```
url_test('contests', status_code=404)
```

tests/redirects/map_301.py
```diff
@@ -310,3 +310,29 @@
...
+    url_test('/es4/', 'http://www.ecma-international.org/memento/TC39.htm',
+             status_code=302),
```
escattone: Maybe add test with no trailing slash as well?
```
url_test('/es4', 'http://www.ecma-international.org/memento/TC39.htm', status_code=302),
```

tests/redirects/map_301.py
```diff
@@ -310,3 +310,29 @@
...
+    url_test('/index.php?title=En/HTML/Canvas&revision=11',
+             status_code=404),
+    url_test('/index.php?title=En/HTML/Canvas&revision=11',
+             status_code=404),
+    url_test('/patches', status_code=404),
+    url_test('/patches/foo', status_code=404),
+    url_test('/web-tech', status_code=404),
+    url_test('/web-tech/feed/atom/', status_code=404),
+    url_test('/web-tech', status_code=404),
+    url_test('/web-tech/feed/atom/', status_code=404),
```
escattone: It looks like lines 327 and 328 are copies of 325 and 326.

jwhitlock: 😞 yes those are dupes

tests/redirects/map_301.py
jwhitlock: 👍

jwhitlock:
Force-push to rebase and address review comments.

This changes the behavior of the ``/contests/`` endpoint - it was restricted by a trailing slash before - but I'm OK with that, I don't see us using that path ever again.

PR #4384, opened 2017-08-25
Fix issue #213, implement traffic cop for interactive examples testing

schalkneethling:
@jwhitlock I believe I have all the ingredients here but, not sure how to test this locally. Perhaps there is still something missing?

### DIFF
--- a/docs/feature-toggles.rst
+++ b/docs/feature-toggles.rst
@@ -81,36 +81,41 @@ and documented in the admin site's `constance section`_.
 Using Traffic Cop for A/B Testing
 ==================================
 
-Traffic Cop is a lightweight JavaScript library that allows conducting content experiments on MDN without the need for third party tools such as Optimizely. Traffic Cop also respects user privacy, and will not run if Do Not Track is enabled.
+Traffic Cop is a lightweight JavaScript library that allows conducting content experiments on MDN without the need for third party tools such as Optimizely. Traffic Cop also respects user privacy, and will not run if `Do Not Track <https://en.wikipedia.org/wiki/Do_Not_Track>`_ is enabled.
 
-On MDN, Traffic Cop will also not run for logged in users. Below follows the steps a developer needs to follow to implement a Traffic Cop content experiment.
+On MDN, Traffic Cop will also not run for logged in users. Here's how to implement a Traffic Cop content experiment.
 
-The first task is to choose, or create, your control, and test pages. We will here use a scenario of testing a new feature on MDN. The new feature involved replacing the static examples on top, with an interactive editor.
+The first task is to choose your control page and create your tests pages Here we will use the scenario of testing a new feature on MDN. The new feature involved replacing the static examples on top, with an interactive editor.
 
 First we choose our control page. For this example, we will use:
 
-    https://developer.mozilla.com/docs/Web/JavaScript/Reference/Global_Objects/Array/push
+`https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push() <https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()>`_
 
 Next we need our test page. The convention on MDN is to first create a base page that follows a naming convention such as:
 
-    https://developer.mozilla.org/docs/Experiment:ExperimentName/
+`https://developer.mozilla.org/docs/Experiment:ExperimentName/
+<https://developer.mozilla.org/docs/Experiment:ExperimentName/>`_
 
 With the base page created, we next create out test variation of our chosen page. To do this, navigate to:
 
-    https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()
+`https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()
+<https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()>`_
 
 You should should now be presented with an editing interface to create the new page.
 
 Switch back to the control page, and click the ``Edit`` button. Switch the editor to source mode and copy all of the contents inside the editor. Switch back to our test, ensure the editor is set to source mode, and paste the content you just copied.
 
-You can now go ahead and make the changes you wish to user test. Once you are complete, go ahead and puclish the page.
+You can now go ahead and make the changes you wish to user test. Once you are complete, go ahead and publish the page.
 
 The JavaScript
-==============
+--------------
 
 Let us first add the JavaScript code to initialise and configure Traffic Cop.
 
-Inisde ``kuma\static\js``, create a new file called some like ``experiment-experiment-name.js`` and paste the following code into the file:
+Inside ``kuma\static\js``, create a new file called some like ``experiment-experiment-name.js`` and paste the following code into the file:
+
+.. code-block:: javascript
+    :linenos:
 
     (function() {
         'use strict';
@@ -126,16 +131,19 @@ Inisde ``kuma\static\js``, create a new file called some like ``experiment-exper
         cop.init();
     })(window.Mozilla);
 
-This will initialise Traffic Cop, set up an experiment with the id ``experiment-experiment-name``, and lastely define a 50/50 split between the control, and the test page.
+This will initialise Traffic Cop, set up an experiment with the id ``experiment-experiment-name``, and lastly define a 50/50 split between the control, and the test page.
 
 Define your bundle
-==================
+------------------
 
-Next we need to add an entry into ``kuma\settings\common.py`` to identify the test, and the related JS that will be injected. Find the following line in ``common.py``:
+Next we need to add an entry into ``kuma\settings\common.py`` to identify the test, and the related JS that will be injected. Find the following line in ``common.py``::
 
         PIPELINE_JS = {
 
 Just before the closing ``}`` add a block such as the following:
+
+.. code-block:: python
+    :linenos:
 
         'experiment-experiment-name': {
             'source_filenames': (
@@ -150,9 +158,12 @@ Just before the closing ``}`` add a block such as the following:
 NOTE: The key here ``experiment-experiment-name`` needs to match the ``id`` you specified in your JS file above.
 
 Identify your A/B test pages
-============================
+----------------------------
 
 The final step is to identify the pages to the back-end so it will know where to direct traffic based on the URL parameter that will be added by Traffic Cop. Inside ``kuma\settings\content_experiments.json`` add the following:
+
+.. code-block:: json
+    :linenos:
 
         [
             {
@@ -177,11 +188,12 @@ There are a couple of important points to note here:
 5. The key for the two pages listed next, needs to match the values you used as the parameter value inside ``variations`` in your JS file earlier.
 
 Testing your experiment
-=======================
+-----------------------
 
 With you local instance of Kuma running, navigate to the page you defined as your control. In this example:
 
-    http://localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push
+`http://localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push
+<http://localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push>`_
 
 NOTE: You should not be logged in to MDN, and ensure that Do Not Track is disabled.
 
--- a/docs/feature-toggles.rst
+++ b/docs/feature-toggles.rst
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
 .. _Waffle switches: https://waffle.readthedocs.io/en/latest/types/switch.html
 .. _Waffle flags: https://waffle.readthedocs.io/en/latest/types/flag.html
 .. _PR 3331: https://github.com/mozilla/kuma/pull/3331
+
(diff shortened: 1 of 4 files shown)


kuma/settings/content_experiments.json
```diff
@@ -1,2 +1,85 @@
 [
+    {
+        "id": "experiment-interactive-editor",
+        "ga_name": "interactive-editor",
+        "param": "v",
+        "pages": [
+            {
+                "locale": "en-US",
+                "slug": "docs/Web/JavaScript/Reference/Global_Objects/Array/push",
```
jwhitlock: The ``slug`` should be ``Web/JavaScript/Reference/Global_Objects/Array/push``, the ``docs`` is implied.  Same with other ``slug`` values in the JSON.

kuma/settings/content_experiments.json
```diff
@@ -1,2 +1,85 @@
 [
+    {
+        "id": "experiment-interactive-editor",
+        "ga_name": "interactive-editor",
+        "param": "v",
+        "pages": [
```
jwhitlock: See test_views_document for the current format of this setting:

https://github.com/mozilla/kuma/blob/fc03aad8490991553c149633c2f0280693b6b9ec/kuma/wiki/tests/test_views_document.py#L18-L29

``pages`` is now a dict of ``locale:slug`` keys to a dict of ``variant``, ``slug`` pairs.

kuma/settings/common.py
```diff
@@ -1096,6 +1096,13 @@ def pipeline_one_scss(slug, **kwargs):
         ),
         'output_filename': 'build/js/selectivizr.js',
     },
+    'experiment-interactive-editor': {
+        'source_filenames': (
+            'js/libs/mozilla.trafficcop.js',
```
jwhitlock: I think you need additional libraries, unless these have been merged into ``mozilla.trafficcop.js``:

https://github.com/mozilla/kuma/blob/e780a32a1c31acbbe0d8597fea970605b0b39d05/kuma/settings/common.py#L1062-L1070

kuma/static/js/experiment-interactive-editor.js
```diff
@@ -0,0 +1,13 @@
...
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment-interactive-editor',
+        variations: {
+            'v=control': 50, // control
+            'v=test': 50 // interactive editor
+        }
+    });
+
+    cop.init();
+})();
```
jwhitlock: I think you'll need to:

1. Pass in ``window.Mozilla``
2. Set a longer cookie duration, if Kadir's metric is longer than a single visit.

https://github.com/mozilla/kuma/blob/c1d36a835fda3edb0b19f7348a77f681808a0d90/kuma/static/js/experiment-static-examples-on-top.js#L1-L14

schalkneethling:
@jwhitlock 

>If the pages are ready on production, I can get them into the next sample database, and maybe we can run some functional tests as well.

Yes they are ready, you can see the list here:
https://github.com/mdn/interactive-examples/wiki/Are-We-Beta-Ready-Yet%3F#ab-testing-pages

kuma/settings/content_experiments.json
```diff
@@ -1,2 +1,33 @@
 [
+    {
+        "id": "experiment-interactive-editor",
+        "ga_name": "interactive-editor",
+        "param": "v",
+        "pages": {
+            "en-US:Web/JavaScript/Reference/Global_Objects/Array/push": {
+                "control": "docs/Web/JavaScript/Reference/Global_Objects/Array/push",
```
jwhitlock: Remove ``docs`` from these as well: ``Web/JavaScript/Reference/Global_Objects/Array/push``.

jwhitlock:
Testing locally, I got an unexpected sidebar for the [b variant of Array.prototype.push()](http://localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push?v=b):

experiment sidebar

This may be a local issue that goes away on staging.

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
 .. _PR 3331: https://github.com/mozilla/kuma/pull/3331
+
+Using Traffic Cop for A/B Testing
+==================================
+
+Traffic Cop is a lightweight JavaScript library that allows conducting content experiments on MDN without the need for third party tools such as Optimizely. Traffic Cop also respects user privacy, and will not run if Do Not Track is enabled.
+
+On MDN, Traffic Cop will also not run for logged in users. Below follows the steps a developer needs to follow to implement a Traffic Cop content experiment.
+
+The first task is to choose, or create, your control, and test pages. We will here use a scenario of testing a new feature on MDN. The new feature involved replacing the static examples on top, with an interactive editor.
```
jwhitlock: How about "The first task is to choose your control page and create your tests pages"?, to get rid of some awkward commas?

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+
+    https://developer.mozilla.com/docs/Web/JavaScript/Reference/Global_Objects/Array/push
+
+Next we need our test page. The convention on MDN is to first create a base page that follows a naming convention such as:
+
+    https://developer.mozilla.org/docs/Experiment:ExperimentName/
+
+With the base page created, we next create out test variation of our chosen page. To do this, navigate to:
+
+    https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()
```
jwhitlock: This may need an inline link, to keep the RST parser from giving up early, like:
```
https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push() 
<https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()>`_
```

kuma/static/js/experiment-interactive-editor.js
```diff
@@ -0,0 +1,14 @@
...
+        id: 'experiment-interactive-editor',
+        cookieExpires: 24 * 365, // 1 year
+        variations: {
+            'v=a': 50, // control
+            'v=b': 50 // interactive editor
+        }
+    });
+
+    cop.init();
+})(window.Mozilla);
```
jwhitlock: Should ``Mozilla`` be an argument of this anon function? It doesn't make sense to me that ``window.Mozilla`` is passed in otherwise, but I'm a lightweight JS dev, and this version works.

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+Switch back to the control page, and click the ``Edit`` button. Switch the editor to source mode and copy all of the contents inside the editor. Switch back to our test, ensure the editor is set to source mode, and paste the content you just copied.
+
+You can now go ahead and make the changes you wish to user test. Once you are complete, go ahead and puclish the page.
+
+The JavaScript
+==============
+
+Let us first add the JavaScript code to initialise and configure Traffic Cop.
+
+Inisde ``kuma\static\js``, create a new file called some like ``experiment-experiment-name.js`` and paste the following code into the file:
```
jwhitlock: s/Inisde/Inside/

There should be a double-colon after ``file``, like ``file::``, to tell Sphinx that the following is [source code](http://www.sphinx-doc.org/en/stable/rest.html#source-code) instead of just a block quote.

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+
+        cop.init();
+    })(window.Mozilla);
+
+This will initialise Traffic Cop, set up an experiment with the id ``experiment-experiment-name``, and lastely define a 50/50 split between the control, and the test page.
+
+Define your bundle
+==================
+
+Next we need to add an entry into ``kuma\settings\common.py`` to identify the test, and the related JS that will be injected. Find the following line in ``common.py``:
```
jwhitlock: Same double colon ``::`` needed to mark following as source code.

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+            ),
+            'output_filename': 'build/js/experiment-experiment-name.js',
+        },
+
+NOTE: The key here ``experiment-experiment-name`` needs to match the ``id`` you specified in your JS file above.
+
+Identify your A/B test pages
+============================
+
+The final step is to identify the pages to the back-end so it will know where to direct traffic based on the URL parameter that will be added by Traffic Cop. Inside ``kuma\settings\content_experiments.json`` add the following:
```
jwhitlock: Another double colon ``::``

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+
+With the base page created, we next create out test variation of our chosen page. To do this, navigate to:
+
+    https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()
+
+You should should now be presented with an editing interface to create the new page.
+
+Switch back to the control page, and click the ``Edit`` button. Switch the editor to source mode and copy all of the contents inside the editor. Switch back to our test, ensure the editor is set to source mode, and paste the content you just copied.
+
+You can now go ahead and make the changes you wish to user test. Once you are complete, go ahead and puclish the page.
```
jwhitlock: s/puclish/publish

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+
+The JavaScript
+==============
+
+Let us first add the JavaScript code to initialise and configure Traffic Cop.
```
jwhitlock: We'll allow the queen's spelling just this once 🇬🇧

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+            variations: {
+                'v=a': 50, // control
+                'v=b': 50 // test
+            }
+        });
+
+        cop.init();
+    })(window.Mozilla);
+
+This will initialise Traffic Cop, set up an experiment with the id ``experiment-experiment-name``, and lastely define a 50/50 split between the control, and the test page.
```
jwhitlock: s/lastely/lastly/

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
...
+    https://developer.mozilla.org/docs/Experiment:ExperimentName/Array.prototype.push()
+
+You should should now be presented with an editing interface to create the new page.
+
+Switch back to the control page, and click the ``Edit`` button. Switch the editor to source mode and copy all of the contents inside the editor. Switch back to our test, ensure the editor is set to source mode, and paste the content you just copied.
+
+You can now go ahead and make the changes you wish to user test. Once you are complete, go ahead and puclish the page.
+
+The JavaScript
+==============
```
jwhitlock: This should use dashes (``------------------``) instead of equals signs, so that it will become an ``<h3>`` instead of another ``<h2>``.

schalkneethling:
> This may be a local issue that goes away on staging.

Yes, I am pretty confident it is just a local issue. I do not see that happening on a similar page on production: https://developer.mozilla.org/en-US/docs/Experiment:InteractiveEditor/Array.prototype.push()

docs/feature-toggles.rst
```diff
@@ -77,3 +77,116 @@ and documented in the admin site's `constance section`_.
 .. _Waffle switches: https://waffle.readthedocs.io/en/latest/types/switch.html
 .. _Waffle flags: https://waffle.readthedocs.io/en/latest/types/flag.html
 .. _PR 3331: https://github.com/mozilla/kuma/pull/3331
+
+Using Traffic Cop for A/B Testing
+==================================
+
+Traffic Cop is a lightweight JavaScript library that allows conducting content experiments on MDN without the need for third party tools such as Optimizely. Traffic Cop also respects user privacy, and will not run if Do Not Track is enabled.
```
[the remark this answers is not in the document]
schalkneethling: Good idea.Done.

docs/feature-toggles.rst
schalkneethling: 👍

docs/feature-toggles.rst
schalkneethling: :)

docs/feature-toggles.rst
schalkneethling: 👍

docs/feature-toggles.rst
schalkneethling: Meh, total spellchecker #fail

kuma/static/js/experiment-interactive-editor.js
schalkneethling: I actually do not even think we need to pass this in at all, I will test. As `Mozilla` and all objects attached to it are exposed to the global `window` object, it should just be available. I believe this was from a version I copy and pasted, and was also wondering whether it was really required.

schalkneethling:
@jwhitlock Updated. Thx

jwhitlock:
New Javascript works locally. There's still the messed up sidebar for the JS examples locally. The documentation looks good and builds without warnings. Thanks @schalkneethling!

PR #4385, opened 2017-08-25
bug 957802: Add coverage of tests

jwhitlock:
There's a lot of dead code in the tests, which is invisible because the tests are hidden from code coverage.


jwhitlock:
This commit is included in PR #4386, and will be merged if that is merged, or vice-versa.

PR #4386, opened 2017-08-26
bug 957802: Remove unused test code and branches

jwhitlock:
Refactor and remove test code to eliminate unused code and branches.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,4 +3,4 @@ source = kuma
 branch = 1
 
 [report]
-omit = *tests*,*migrations*,*/management/commands/*
+omit = *migrations*,*/management/commands/*
--- a/kuma/core/tests/__init__.py
+++ b/kuma/core/tests/__init__.py
@@ -34,12 +34,6 @@ def ok_(pred, msg=None):
     assert pred, msg
 
 
-def attrs_eq(received, **expected):
-    """Compares received's attributes with expected's kwargs."""
-    for k, v in expected.iteritems():
-        eq_(v, getattr(received, k))
-
-
 def get_user(username='testuser'):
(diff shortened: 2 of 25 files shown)


jwhitlock:
This builds on PR #4385, which shows what coverage looks like with tests included. I was surprised by the amount of dead code in tests, although it does end up being a small percent.  This PR removes the dead code, so that the new missed lines and partial branches are minimized.

This more than doubles the lines of code reported, but covering test code is controversial. The reason I think we should is:

* We can drop test utility functions as we deprecate them for purest fixtures
* We can identify when a test is testing the wrong thing, such as tests that were parsing JSON as HTML through copy and paste
* Our test suite represents a big part of our technical debt. Bad tests don't communicate how the code is supposed to work, slows down our development cycle, and takes time to run. We've committed to cleaning up tests as we change production code. Code Coverage is one way to measure if we mess up this refactoring.

jwhitlock:
It's easier for me to run coverage locally (``make coveragetesthtml`` in ``web`` container, then open ``htmlcov/index.html`` in browser) than to view uncovered lines on codecov.io.  The uncovered line is [line 1224](https://codecov.io/gh/mozilla/kuma/src/master/kuma/wiki/tests/test_templates.py#L1224):

https://github.com/mozilla/kuma/blob/c6c6bb0ae8493d71fb77dd1d1948f2c051394720/kuma/wiki/tests/test_templates.py#L1211-L1224

The test is marked ``@pytest.mark.xfail(reason='broken test')``. It fails on line 1223, which is why 1224 is not run and uncovered.

escattone:
Oh, I failed to notice that the test was marked `xfail`, that makes sense now, thanks @jwhitlock!

PR #4387, opened 2017-08-26
[Bug 962148] URL with 'en' locale don't redirect to 'en-US'

safwanrahman:
There could be more generic fix for the locale with underscore. Like `bn_BD` should redirect to `bn-BD`. But for avoiding the complexity, I think this patch is good enough to fix the burning issue.

As we handle the language redirect already from settings, I believe we do not need to add this rule to `kuma/redirects/redirects.py`

@jwhitlock r?

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -226,6 +226,7 @@ LOCALE_ALIASES = {
     'pt': 'pt-PT',
     'sr': 'sr-Cyrl',
     'zh': 'zh-CN',
+    'cn': 'zh-CN',
 
     # Create aliases for locales which do not share a prefix.
     'nb-NO': 'no',
@@ -234,6 +235,10 @@ LOCALE_ALIASES = {
     # Create aliases for locales which use region subtags to assume scripts.
     'zh-Hans': 'zh-CN',
     'zh-Hant': 'zh-TW',
+
+    # Map locale whose region subtag is separated by `_`(underscore)
+    'zh_cn': 'zh-CN',
+    'zh_tw': 'zh-TW',
 }
 
 LANGUAGE_URL_MAP = dict([(i.lower(), i) for i in MDN_LANGUAGES])
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -213,6 +213,7 @@ LOCALE_ALIASES = {
 
     # Create aliases for over-specific locales.
     'bn': 'bn-BD',
+    'cn': 'zh-CN',
     'fy': 'fy-NL',
     'ga': 'ga-IE',
     'gu': 'gu-IN',
@@ -226,7 +227,6 @@ LOCALE_ALIASES = {
     'pt': 'pt-PT',
     'sr': 'sr-Cyrl',
     'zh': 'zh-CN',
-    'cn': 'zh-CN',
 
     # Create aliases for locales which do not share a prefix.
     'nb-NO': 'no',
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -3390,7 +3390,7 @@ class MindTouchRedirectTests(UserTestCase, WikiTestCase):
          'expected': '/zh-CN/docs/XHTML'},
         {'title': 'JavaScript', 'mt_locale': 'zh_cn', 'kuma_locale': 'zh-CN',
          'expected': '/zh-CN/docs/JavaScript'},
-        {'title': 'XHTML6', 'mt_locale': 'zh_tw', 'kuma_locale': 'zh-CN',
+        {'title': 'XHTML6', 'mt_locale': 'zh_tw', 'kuma_locale': 'zh-TW',
          'expected': '/zh-TW/docs/XHTML6'},
         {'title': 'HTML7', 'mt_locale': 'fr', 'kuma_locale': 'fr',
          'expected': '/fr/docs/HTML7'},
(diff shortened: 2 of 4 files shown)


jwhitlock:
TravisCI tests are failing. Can you fix those? Did you get the TravisCI notification, or is there some configuration that we need to change?

jwhitlock:
I've moved my alternate implementation (using the redirects library to handle these URLs) to a new branch, [legacy-locales-962148](https://github.com/mozilla/kuma/compare/master...jwhitlock:legacy-locales-962148?expand=1).  The MindTouch redirects appear to be handling the ``en`` redirect correctly, so it can potentially be expanded for ``cn``, ``zh-cn``, and ``zh-tw``.

@safwanrahman, this is now an AWS blocker, so let me know if you get stuck or need to pause Kuma work.

safwanrahman:
>TravisCI tests are failing. Can you fix those? Did you get the TravisCI notification, or is there some configuration that we need to change?
@jwhitlock I am seeing the fail in my local machine also. The fail is cause of status code and redirection. It used to be redirect with 301 but now its going to be redirect with 302 and then 301 status code.
I am going to fix the test today

kuma/wiki/tests/test_views.py
```diff
@@ -3390,7 +3390,7 @@ class MindTouchRedirectTests(UserTestCase, WikiTestCase):
          'expected': '/zh-CN/docs/XHTML'},
         {'title': 'JavaScript', 'mt_locale': 'zh_cn', 'kuma_locale': 'zh-CN',
          'expected': '/zh-CN/docs/JavaScript'},
-        {'title': 'XHTML6', 'mt_locale': 'zh_tw', 'kuma_locale': 'zh-CN',
+        {'title': 'XHTML6', 'mt_locale': 'zh_tw', 'kuma_locale': 'zh-TW',
```
jwhitlock: It was very strange that this worked! However, I see that the test was just looking at the first redirect, and matching it again the ``expected`` value ``/zh-TW/docs/XHTML6``. Only took me 15 minutes to figure that out 🙁

kuma/settings/common.py
```diff
@@ -226,6 +226,7 @@ def __call__(self, value):
     'pt': 'pt-PT',
     'sr': 'sr-Cyrl',
     'zh': 'zh-CN',
+    'cn': 'zh-CN',
```
jwhitlock: nit: this should be at line 216, between ``bd`` and ``fy``, according to sorting rules

https://github.com/mozilla/kuma/blob/c6c6bb0ae8493d71fb77dd1d1948f2c051394720/kuma/settings/common.py#L214-L218

kuma/wiki/tests/test_views.py
```diff
@@ -3413,9 +3413,10 @@ def test_trailing_slash(self):
         d.title = 'FooFoo'
         d.save()
         mt_url = '/cn/%s/' % (d.slug,)
-        resp = self.client.get(mt_url)
-        eq_(301, resp.status_code)
-        eq_('http://testserver%s' % d.get_absolute_url(), resp['Location'])
+        resp = self.client.get(mt_url, follow=True)
+        eq_(200, resp.status_code)
+        # Check the last redirect chain url is the expected url
+        eq_('http://testserver%s' % d.get_absolute_url(), resp.redirect_chain[-1][0])
```
jwhitlock: I think this test should be removed, since it no longer applies.  It is testing the ``MINDTOUCH_PROBLEM_LOCALES`` branch which is no longer used, according to code coverage and manual testing.  ``MINDTOUCH_PROBLEM_LOCALES`` and the branch in ``mindtouch_to_kuma_redirect`` that deals with it should be removed as well.

https://github.com/mozilla/kuma/blob/c6c6bb0ae8493d71fb77dd1d1948f2c051394720/kuma/wiki/views/legacy.py#L23-L28

https://github.com/mozilla/kuma/blob/c6c6bb0ae8493d71fb77dd1d1948f2c051394720/kuma/wiki/views/legacy.py#L94-L104

kuma/wiki/tests/test_views.py
```diff
@@ -3425,9 +3426,11 @@ def test_document_urls(self):
...
             d.save()
             mt_url = '/%s' % '/'.join([doc['mt_locale'], doc['title']])
-            resp = self.client.get(mt_url)
-            eq_(301, resp.status_code)
-            eq_('http://testserver%s' % doc['expected'], resp['Location'])
+            resp = self.client.get(mt_url, follow=True)
+            eq_(200, resp.status_code)
+
+            # Check the last redirect chain url is the expected url
+            eq_('http://testserver%s' % doc['expected'], resp.redirect_chain[-1][0])
```
jwhitlock: I don't think the test should be altered in this way. After the changes, the request doesn't make it to the ``mindtouch_to_kuma_redirect``, but instead the ``LocaleURLMiddleware`` handles it.   I think this test and the supporting ``self.documents`` should be deleted.

If you want, you can add new test cases to ``LocalURLMiddleware``'s tests, in ``kuma/core/tests/test_locale_middleware.py``, which only tests the ``/en/`` case now.

https://github.com/mozilla/kuma/blob/c6c6bb0ae8493d71fb77dd1d1948f2c051394720/kuma/core/tests/test_locale_middleware.py#L37-L40

safwanrahman: @jwhitlock The request actually make it to `mindtouch_to_kuma_redirect`.
So it happens like this
`/cn/Web` redirect with 302 to `zh-cn/Web`
`zh-cn/Web` does not find any view, so it goes to `mindtouch_to_kuma_redirect` and redirect to `zh-cn/docs/Web`. So I think this test should be there

PR #4392, opened 2017-08-30
Disable content experiment, interactive examples

jwhitlock:
We want to advertise the new interactive examples in the next developer newsletter, so we should deactivate the experiment so we can link to it.


schalkneethling:
r+ 🦆

atopal:
Please don't stop the A/B test. The newsletter will link to the forum post. We can remove those referrers if necessary.

atopal:
Please don't stop the A/B test. The newsletter will link to the forum post and we can report on organic traffic only, but we could use some qualitative feedback, that we'll hopefully get in the forum thread.

jwhitlock:
OK

PR #4398, opened 2017-09-01
bug 1393832: Refactor Document JSON tests

jwhitlock:
Refactor ``Document.get_json_data`` and ``.build_json_data`` into ``pytest``-standard tests, and fully exercise the code branches.  Collect these and other refactored ``Document`` model tests into a new file, so there's a better separation between the old and moldy and the new and fresh.

My goal is to get the tests to where I would want them before adding the functionality for [bug 1393832](https://bugzilla.mozilla.org/show_bug.cgi?id=1393832) (detect more invalid JSON and re-render), so there's no change in the model code yes.  While writing the tests, there were a couple of branches that looked bad to me, so I marked the tests with ``TODO``s to clean up later.


jwhitlock:
TL;DR - there won't be a follow-on PR to fix [bug 1393845](https://bugzilla.mozilla.org/show_bug.cgi?id=1393845) for a while. This is still an improvement to tests, so leaving the PR open for review and merge.

Longer version:

The view tests for ``$json`` are also in bad shape and will need to be re-written to the pytest standard.  There is a second view that returns JSON, that is tested in the same test, and should be move to its own test.  There's [bug 1195411](https://bugzilla.mozilla.org/show_bug.cgi?id=1195411) that requests removing that view, but it is used by the translate view. Or, it appears to be used, but it would require reading the JS to see if it is actually used.

The tests setup the switch ``application_ACAO``, which adds a header ``AccessControlAllowOrigin: *`` to the response. This switch is set in all environments, so the switch should be removed and assumed on. The discussion around this switch is in [bug 1104260](https://bugzilla.mozilla.org/show_bug.cgi?id=1104260).

To remove the switch, we should have a headless test that ensures the header is there in all environments. Headless tests don't run during integration tests at the moment, but they could. Integration tests are broken due to the redesign ([bug 1394480](https://bugzilla.mozilla.org/show_bug.cgi?id=1394480), but are probably needed for testing AWS deployments.

So, the order is:

1. Get functional tests working
2. Add headless tests for checking this header (expanding redirect tests)
3. Remove the ``application_ACAO`` switch
4. Refactor ``$json`` view tests to pytest
5. Ensure fresher JSON at view time, update tests.

We'll need this PR merged before step 5.

jwhitlock:
Force-push to rebase on master, fix bug number in commit descriptions

kuma/wiki/tests/test_models_document.py
```diff
@@ -0,0 +1,277 @@
...
+
+def test_get_json_data_cached_parsed_json(root_doc):
+    """
+    Document.get_json_data uses cached parsed JSON from previous calls.
+
+    TODO: Is this a good idea? It seems like it should never run
+    (callers should work to run it once if efficency is needed), and
+    if it does run twice, would hide any changes made to the instance.
+    """
+    stale_data = {"stale": "like an old crouton"}
```
escattone: 😄

PR #4399, opened 2017-09-01
bug 1395684: persist revision hash in docker images

escattone:
Persist the git commit hash for HEAD within the kuma and kumascript Docker images as the environment variable `REVISION_HASH`.

For local development:
- if you use `make up` instead of doing the `docker-compose up -d` directly, the kuma-based containers (`web`, `api`, etc.) will have an accurate `REVISION_HASH`, otherwise its value will be "undefined"
- since the `kumascript` container uses the kumascript Docker image, an accurate `REVISON_HASH` will always be baked into it

Thanks to @safwanrahman  for the idea of using docker build arguments!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ target = kuma
 requirements = -r requirements/local.txt
 # set Django settings module if not already set as env var
 export DJANGO_SETTINGS_MODULE ?= kuma.settings.testing
+export KUMA_REVISION_HASH ?= $(shell git rev-parse HEAD)
+export KUMASCRIPT_REVISION_HASH ?= $(shell cd kumascript && git rev-parse HEAD)
 
 # Note: these targets should be run from the kuma vm
 test:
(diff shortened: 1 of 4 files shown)


safwanrahman:
my github is @safwanrahman actually! ;)

Makefile
```diff
@@ -24,6 +24,8 @@ target = kuma
 requirements = -r requirements/local.txt
 # set Django settings module if not already set as env var
 export DJANGO_SETTINGS_MODULE ?= kuma.settings.testing
+export KUMA_REVISION_HASH ?= $(shell git rev-parse HEAD)
+export KUMASCRIPT_REVISION_HASH ?= $(shell cd kumascript && git rev-parse HEAD)
```
jwhitlock: ``git`` is not installed inside the ``web`` container, and running ``make build-static`` or other make commands outputs:

```
make: git: Command not found
/bin/sh: 1: git: not found
```

Please avoid it by directing these messages to ``/dev/null``, like:

```
export KUMA_REVISION_HASH ?= $(shell git rev-parse HEAD 2>/dev/null)
export KUMASCRIPT_REVISION_HASH ?= $(shell cd kumascript && git rev-parse HEAD 2>/dev/null)
```

escattone: Good catch, thanks!

PR #4400, opened 2017-09-01
Typography tweaks for kbd and code in indicators

stephaniehobson:
Fix bug 1376770.

Testing notes:
`kbd` in a heading on [this page](https://developer.mozilla.org/en-US/docs/User:stephaniehobson:code#This_is_a_heading_with_a_code_tag_in_it_%3Chtml%3E._Lorem_codeipsum_kbd_dollar_sit_amet)

Code problem reported on [this page](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_prototypes#A_prototype-based_language).

Code looks a lot like the serif font in the notes now. Not sure how I feel about that. I'm wondering if we should add a background colour `like github does`.

### DIFF
--- a/kuma/static/styles/base/elements/typography.scss
+++ b/kuma/static/styles/base/elements/typography.scss
@@ -89,7 +89,12 @@ mono space elements (pre, code, kbd)
 ====================================================================== */
 
 code {
+    background-color: $code-block-background-color;
+    border-radius: 2px;
+    box-decoration-break: clone;
+    color: $text-color;
     font-family: $code-inline-font-family;
+    padding: 2px 5px;
     word-wrap: break-word;
 }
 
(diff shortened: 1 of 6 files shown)


schalkneethling:
Ok, so I am not 100% on whether I tested the correct things here. Perhaps the title on the original offending page has been changed? Also, the page on production linked to contains a Kumascript error and so, I am not 100% sure that it is rendering correctly.

With that, my understanding is that the first problem is the way `<kbd>` is displayed when used in a header. Specifically a SlabZilla type heading. So, here is the way it would display currently:

screen shot 2017-09-04 at 13 35 56

And after the change it now displays as follows:

screen shot 2017-09-04 at 13 38 56

The next issue is the way inline `<code>` is displayed. Again before:

Linked code:

screen shot 2017-09-04 at 13 49 53

And when the code is not linked:

screen shot 2017-09-04 at 13 46 26

And here is unlinked code after the change applied in this PR:

screen shot 2017-09-04 at 13 50 18

I agree @stephaniehobson that it is hard to visually see the difference between plain copy and "code". I played around with your suggestion, re: adding a background colour, and came up with this:

screen shot 2017-09-04 at 13 52 26

The code for that would be:

```
font-family: monospace;
font-size: 16px;
background-color: lightgray;
padding: 2px 5px;
```

Let me know if I overlooked something, or did not test the appropriate things ;)

stephaniehobson:
Sounds like the right things were reviewed, thanks :)

Added a background colour for inline code to this PR now.

PR #4401, opened 2017-09-02
bug 1362438: add misc views for aws

escattone:
This PR adds Django views for serving the sitemaps as well as the kuma and kumascript revision hashes. These views are for AWS. Since we're not reverse-proxying our requests through nginx or Apache, they're not served directly by Apache as they are in SCL3.

This PR depends on https://github.com/mdn/kumascript/pull/303.


escattone:
I've marked this "not ready" as I need to add tests for the views, but I wanted to make it available for comment.

kuma/misc/urls.py
```diff
@@ -0,0 +1,26 @@
+from django.conf import settings
+from django.conf.urls import url
+from django.views.static import serve
+
+from . import views
+
+
+url_patterns = [
```
jwhitlock: This should be ``urlpatterns`` (drop the underscore)

kuma/misc/urls.py
```diff
@@ -0,0 +1,26 @@
...
+    url(r'^media/kumascript-revision.txt$',
+        views.kumascript_revision_hash,
+        name='misc.kumascript_revision'),
+
+    # Serve sitemap files for AWS (these are never hit in SCL3).
+    url(r'^sitemap.xml$',
+        serve,
+        {'document_root': settings.MEDIA_ROOT, 'path': 'sitemap.xml'},
+        name='misc.sitemap'),
+    url(r'^(?P<path>sitemaps/.+)$',
```
jwhitlock: ``sitemap.xml`` and ``sitemaps/`` should be added to ``LANGUAGE_URL_IGNORED_PATHS `` in ``kuma/settings/common.py``, to avoid the locale redirect.

escattone:
Ok, I think this is ready for final review. @jwhitlock I'm not certain you'll like the organization, so let me know if you don't and I'll be happy to change it.

kuma/revision/test_views.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+import requests_mock
+
+from kuma.core.urlresolvers import reverse
+from kuma.wiki.constants import KUMASCRIPT_BASE_URL
+
+
+@pytest.mark.parametrize('method', ['get', 'head'])
+def test_revision_hash(client, db, method):
```
jwhitlock: Huh. It is sad that this requires a database. Not sure if I should blame Django or our middleware.

kuma/urls.py
```diff
@@ -74,6 +75,19 @@
 
 urlpatterns += [
     # Services and sundry.
+    url('', include('kuma.revision.urls')),
+
+    # Serve sitemap files for AWS (these are never hit in SCL3).
+    url(r'^sitemap.xml$',
```
jwhitlock: 👍 I like this much better for where the sitemap URLs live.

kuma/test_urls.py
```diff
@@ -0,0 +1,82 @@
+import pytest
```
jwhitlock: I think these tests could live in ``kuma/core/tests/test_views.py``, where the ``events`` test ended up, rather than have a "naked" test file at this level.

kuma/revision/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.conf.urls import url
+
+from . import views
+
+
+urlpatterns = [
+    # Serve the revision hashes.
+    url(r'^media/revision.txt$',
```
jwhitlock: This is a better place than "misc". I'm a little worried it will get confused with wiki revisions, but we can move code where ever. Maybe we can do a sorting exercise at an all hands to discover a better grouping. It seems to be there are some meta views like this and the health checks that could be grouped together under "endpoints that describe the current deployment". No change needed now.

kuma/revision/test_views.py
```diff
@@ -0,0 +1,52 @@
+import os
```
jwhitlock: Can this go in a ``tests/`` folder to match the other sub-apps?

kuma/revision/test_views.py
escattone: I felt the same. I had originally left the `db` fixture off thinking it didn't need it, and was surprised when it did, not realizing that our middleware forced the issue.

kuma/revision/urls.py
escattone: Oh, that's a good point. Sounds good.

escattone:
Thanks for the review @jwhitlock! I'm going to make the organizational changes you suggested and then merge after the tests pass.

kuma/urls.py
```diff
@@ -74,6 +75,19 @@
 
 urlpatterns += [
     # Services and sundry.
+    url('', include('kuma.revision.urls')),
```
jwhitlock: This should now be ``kuma.version.urls``

PR #4402, opened 2017-09-02
Bug 1370594: Sample output, border and buttons changes

stephaniehobson:
- Remove `wiki_samples` waffle flag, this feature is now permanent.
- Lint wiki-samples.js.
- Capitalize CodePen and JSFiddle names.
- Convert button-link to a mixin and apply to `.open-in` buttons.
- Added top margin to `.cta-link` mixin
    - Decrease margin on homepage search form to account for that.
- Add `sample_frame` waffle
    - Make width 100% by default
    - Make border match code examples by deafult
    - Change buttons to use `.cta-link` styles as recommended by Brigade ages ago.

Also:
- Remove outdated summary styles from wysiqyg style sheet.

Testing notes:
I waffled this feature a little differently. I didn't want to duplicate the styles sheets *again* so I added a class to the body if the `sample_frame` waffle is enabled. This means we can roll this feature into production independent of the `line_length` waffle if we want. All the code  still had to be duplicated between the two sets of .scss files though. I suggest reviewing the code for the one in /components/ and I'll copy any requested changes to the one in /components-skinny/ as well.

### DIFF
--- a/docs/feature-toggles.rst
+++ b/docs/feature-toggles.rst
@@ -57,7 +57,7 @@ other advanced criteria.
 * ``spam_testing_mode`` - Tell Akismet that edits are tests, not real content.
 * ``spam_checks_enabled`` - Toggle spam checks site wide.
 * ``spam_submissions_enabled`` - Toggle Akismet spam/spam submission ability.
-* ``wiki_samples`` - Add button to open samples in Codepen or jsFiddle.
+* ``sample_frame`` - Enable .sample-code-frame width, colour, and button styles
 * ``wiki_spam_exempted`` - Exempt users and user groups from checking.
   submissions for spam.
 * ``wiki_spam_training`` - Call Akismet to check submissions, but don't block
--- a/etc/sample_db.json
+++ b/etc/sample_db.json
@@ -324,12 +324,6 @@
         "superusers": false,
         "created": "2015-02-25"
       }, {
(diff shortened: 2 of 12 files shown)


schalkneethling:
This looks really good. But, for some reason I am running into the following issue in Firefox(both release and nightly). On the landing page I get a JS error:

`TypeError: mdn.analytics is undefined`

and then, on a page such as `/docs/Web`, I also get:

`TypeError: win.mdn.getUrlParameter is not a function`

Also, the drop down menu does not work. This happens whether I am logged in or not. This does not happen on production though. Can you replicate this behaviour?

stephaniehobson:
I can't reproduce on this branch even playing around with my GA settings.

There was a bug on master that was fixed by https://github.com/mozilla/kuma/pull/4394 that might have caused this behaviour.

Can you please  rebase on master and tell me if you're still getting the same behaviour?

schalkneethling:
> I can't reproduce on this branch even playing around with my GA settings.

Ok, it is the uBlock extension that is blocking the loading of `/js/analytics.js` - The reason is very cryptic

screen shot 2017-09-06 at 11 42 25

I am going to see if I can get an answer from the dev but, does not block this PR.

stephaniehobson:
Ah, and that problem won't happen on prod because the file is bundled into some other stuff.

PR #4404, opened 2017-09-04
Fix issue #255, exposes site url and editor url to JS and editor

schalkneethling:
@jwhitlock This should be sufficient to expose the needed URLs to the JS in kuma and, make available a `postMessage` target call for the editor to get its parent URL.

Anything here that concern you security wise? Anything you would do different? Thanks!

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -45,7 +45,9 @@ PRODUCTION_URL = SITE_URL
 STAGING_DOMAIN = 'developer.allizom.org'
 STAGING_URL = PROTOCOL + STAGING_DOMAIN
 
-INTERACTIVE_EXAMPLES_BASE = 'https://interactive-examples.mdn.mozilla.net'
+INTERACTIVE_EXAMPLES_BASE = config(
+    'https://interactive-examples.mdn.mozilla.net',
+    default='https://interactive-examples.mdn.mozilla.net')
 
 MAINTENANCE_MODE = config('MAINTENANCE_MODE', default=False, cast=bool)
 ALLOW_ROBOTS = config('ALLOW_ROBOTS', default=False, cast=bool)
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -45,6 +45,8 @@ PRODUCTION_URL = SITE_URL
 STAGING_DOMAIN = 'developer.allizom.org'
 STAGING_URL = PROTOCOL + STAGING_DOMAIN
 
+INTERACTIVE_EXAMPLES_BASE = 'https://interactive-examples.mdn.mozilla.net'
+
 MAINTENANCE_MODE = config('MAINTENANCE_MODE', default=False, cast=bool)
 ALLOW_ROBOTS = config('ALLOW_ROBOTS', default=False, cast=bool)
 
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -1,7 +1,7 @@
 {% from "includes/common_macros.html" import optimizely_script with context -%}
 
 <!DOCTYPE html>
-<html lang="{{ LANG }}" dir="{{ DIR }}" class="no-js" data-ffo-opensans="false" data-ffo-zillaslab="false">
+<html lang="{{ LANG }}" dir="{{ DIR }}" class="no-js" data-ffo-opensans="false" data-ffo-zillaslab="false" data-site-url="{{ settings.SITE_URL }}" data-editor-url="{{ settings.INTERACTIVE_EXAMPLES_BASE }}">
 <head {%- if not is_sphinx %} prefix="og: http://ogp.me/ns#"{% endif %}>
   <meta charset="utf-8">
   {%- block experiment_js %}{% endblock %}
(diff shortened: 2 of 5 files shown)


escattone:
@jwhitlock @schalkneethling 

Since @schalkneethling needs access to the editor URL (and the survey URL, which is not yet part of this PR) within kumascript macros as well, and since we'd prefer to have those settings in one place (here in kuma) rather than also having them somewhere in the kumascript settings, I'm proposing we should do the following (either in this PR or a related one):
- make the `INTERACTIVE_EXAMPLES_BASE` setting configurable from the environment, like `INTERACTIVE_EXAMPLES_BASE = config('INTERACTIVE_EXAMPLES_BASE', default='https://interactive-examples.mdn.mozilla.net')
` so we can set it from `docker-compose.yml` for local development
- add an `INTERACTIVE_EXAMPLES_SURVEY = config('INTERACTIVE_EXAMPLES_SURVEY', default='https://www.surveygizmo.com/s3/3780959/MDN-Interactive-Editor')`
- add these two key/value pairs to the env vars passed to kumascript both [here](https://github.com/mozilla/kuma/blob/master/kuma/wiki/kumascript.py#L133) and [here](https://github.com/mozilla/kuma/blob/master/kuma/wiki/kumascript.py#L49), which adds them to the `env` global variable available within all kumascript macros (which is what @schalkneethling needs)
- add the proper local development setting for `INTERACTIVE_EXAMPLES_BASE` in the `docker-compose.yml` worker [environment](https://github.com/mozilla/kuma/blob/master/docker-compose.yml#L15)

What do you guys think?

jwhitlock:
@escattone that works for me. It is a little strange to pass this value, but it is also trivial to make it a KumaScript setting if that makes more sense in the future.

For speed, I think a backend dev should make those changes. @escattone, do you want to make these changes in a follow-on commit, or should I?

I think https://interactive-examples.mdn.mozilla.net is the right default for development, until we add a container for interactive examples development to the docker dev environment.

kuma/settings/common.py
```diff
@@ -45,6 +45,8 @@ def __call__(self, value):
 STAGING_DOMAIN = 'developer.allizom.org'
 STAGING_URL = PROTOCOL + STAGING_DOMAIN
 
+INTERACTIVE_EXAMPLES_BASE = 'https://interactive-examples.mdn.mozilla.net'
```
jwhitlock: This should use ``config`` so it can be set from the environment, like:

```
INTERACTIVE_EXAMPLES_BASE = config('https://interactive-examples.mdn.mozilla.net', default='https://interactive-examples.mdn.mozilla.net')
```

escattone:
@jwhitlock It is a little strange. Maybe my desire not to repeat code is leading me down an un-natural path. I'm happy to just add it as a setting in kumascript as well. Maybe that's more natural.

escattone:
@jwhitlock I'll make the changes in a new kumascript PR, and I think I'll just add the settings to kumascript.

stephaniehobson:
Do we need to include the survey URL in this? It's going away in a few weeks and there isn't a staging version vs a production version.

stephaniehobson:
Data objects on the HTML tag feel like the wrong way to pass these to javascript. Can we make them part of our global MDN object with the other javascript config stuff in [config.html](https://github.com/mozilla/kuma/blob/418ea22282299d56d2b00ea3d8d11a88085541f3/jinja2/includes/config.html)

schalkneethling:
> Can we make them part of our global MDN object with the other javascript config stuff in config.html

Most definitely, I was not aware of this file. Let me move it there and make the relevant updates.

kuma/settings/common.py
```diff
@@ -45,7 +45,9 @@ def __call__(self, value):
 STAGING_DOMAIN = 'developer.allizom.org'
 STAGING_URL = PROTOCOL + STAGING_DOMAIN
 
-INTERACTIVE_EXAMPLES_BASE = 'https://interactive-examples.mdn.mozilla.net'
+INTERACTIVE_EXAMPLES_BASE = config(
+    'interactive-examples.mdn.mozilla.net',
```
jwhitlock: Should this be ``https://interactive-examples.mdn.mozilla.net``?

schalkneethling: Crud, indeed. Copy/paste fail. Let me update it.

schalkneethling: Fixed up.

jwhitlock:
This change worked locally, after I made some further changes to the Docker configuration. Your branch doesn't allow commits from maintainers, so I'll open a new PR for the changes:

- set ``INTERACTIVE_EXAMPLES_BASE`` in the kuma containers
- set ``interactive_examples__base_url`` in the kumascript container
- use https://github.com/mdn/kumascript/pull/306
- make some changes in interactive-examples to change domain from ``https://developer.mozilla.org`` to ``http://localhost:8000``

https://github.com/mdn/interactive-examples/blob/b05330eac3bb104ff23022220f132c477f558c1d/js/editor-libs/analytics.js#L12

https://github.com/mdn/interactive-examples/blob/b05330eac3bb104ff23022220f132c477f558c1d/js/editor-libs/events.js#L92

- Run interactive-examples locally
- ``make up``, and maybe ``docker-compose restart redis``.

PR #4419, opened 2017-09-11
bug 1397858: Increase Site Speed Sample Rate for interactive examples content experiment

jwhitlock:
Updates the content experiments functional test for the new format from PR #4183, and adds tests for the interactive examples content experiment.

 Sets the site speed sample rate to 100% when the user is viewing one of the interactive examples content variants, and verifies with functional tests.


jwhitlock:
PR #4414, fixing functional tests, so this is now ready.  No update on [bug 1397858](https://bugzilla.mozilla.org/show_bug.cgi?id=1397858) if this should be enabled across the site.

PR #4421, opened 2017-09-12
Removes siteUrl postMessage to interactive editor

schalkneethling:
I did not think this one through all the way :-/ So, there is no way to use this functionality from the interactive editor without opening it up with `*` in terms of the `targetOrigin`. We definitely do not want to do that.

Without that, it is basically a loop de loop/catch22[what have you]. The interactive editor needs to know the `targetOrigin` in order to post a message asking for the URL, and should Kuma push the URL down on load, the interactive editor will need to know the trusted origin, before accepting the message.

I am therefore removing this code as it will not be used.

In terms of a solution. I am thinking the biggest challenge is having this work on staging and production. Local development would be great, but there are most likely not going to a lot of instances where all these components are going to be tested running on localhost.

Therefore, adding a conditional inside the interactive editor that accepts either `developer.mozilla.org` or `developer.allizom.org` as the origin will resolve the above.

### DIFF
--- a/kuma/static/js/interactive.js
+++ b/kuma/static/js/interactive.js
@@ -3,8 +3,6 @@
 
     var iframe = document.querySelector('iframe.interactive');
     var mediaQuery = window.matchMedia('(min-width: 63.9385em)');
-    var siteUrl =
-        window.mdn.interactiveEditor.siteUrl || 'https://developer.mozilla.org';
     var targetOrigin =
         window.mdn.interactiveEditor.editorUrl ||
         'https://interactive-examples.mdn.mozilla.net';
@@ -24,18 +22,6 @@
         iframe.contentWindow.postMessage(message, targetOrigin);
     }
 
-    /**
-     * Posts back the current site URL.
-     * @param {Object} event - The event Object received from postMessage
-     */
-    function postSiteUrl(event) {
-        /* only post the site url if the correct property
-        exists on the message object, and its value is true */
-        if (event.data.siteUrl) {
-            postToEditor({ siteUrl: siteUrl });
-        }
-    }
-
     /* As the user sizes the browser or tilts their device,
     listen for mediaQuery events and communicate the new
     viewport state to the interactive editor */
@@ -53,8 +39,5 @@
             // add the class `small-desktop-and-below`
             postToEditor({ smallViewport: true });
         }
-
-        // add event listener for postMessages from the interactive editor
-        window.addEventListener('message', postSiteUrl, false);
     };
 })();


jwhitlock:
I think I was confused about how this feature worked. I thought the interactive examples code was passing a desired iframe size to the parent, which resized the iframe. I guess instead MDN passes the actual iframe size to the interactive example?

If we still need to pass the public MDN URL to the iframe, ``postMessage`` is probably the wrong mechanism. I think we can still have the feature by using a query parameter.

Here's the information flow as I see it:

The KumaScript macro [EmbedInteractiveExample](https://github.com/mdn/kumascript/blob/master/macros/EmbedInteractiveExample.ejs) needs to know the Interactive Examples base URL, to construct the ``iframe`` ``src``. It can get this from the environment . config.

The Kuma JavaScript needs to know the Interactive Examples domain, so it can validate the ``targetOrigin`` for the desired frame size. It can get this from the environment / config.

The Interactive Examples JavaScript needs to know the Kuma public base URL, so that it can ``postMessage`` to the correct URL. It looks like a ``postMessage`` won't work, because it would need to know the domain to safely receive the message.

Instead, let's pass the Kuma public base URL via the ``iframe`` ``src``, like:

```
<iframe src="https://interactive-examples.mdn.mozilla.net/pages/css/background-color.html?parent_base_url=https://developer.mozilla.org" class="interactive " width="100%" frameborder="0" height="250"></iframe>
```

You could then parse ``parent_base_url`` and use it in the interactive examples JS, defaulting to ``https://developer.mozilla.org`` if it is missing.  This would require a change to the KumaScript and to the Interactive Examples JS. 

I think most of this PR would still apply, but I think I'd need to read more code to understand the information flow between the pieces.

schalkneethling:
> I think I was confused about how this feature worked. I thought the interactive examples code was passing a desired iframe size to the parent, which resized the iframe.

Other way around :) Kuma passes a true/false boolean to the editor at a certain breakpoint. The editor then applies a class on the editor container. This is only done for the CSS editor. Because of the nature of the JS editor, it is now needed for it. But, the other communication that happens is from the editor, but that related to GA metrics.

> If we still need to pass the public MDN URL to the iframe

Yes, in order for the editor to send, or receive `postMessage`'s it need to know which origin/target to trust. For this to then work on the various environments, it needs to know it's parent's URL

> It looks like a postMessage won't work, because it would need to know the domain to safely receive the message.

Yup

> Instead, let's pass the Kuma public base URL via the iframe src

That is one option, the other being setting a `data-siteurl` attribute on the `iframe` and I can then read it from there.

> I think most of this PR would still apply, but I think I'd need to read more code to understand the information flow between the pieces.

Yes, the code that is removed here is only concerned with the passing of the Kuma url to the interactive editor. As we will no longer use that mechanism, it can be safely removed.

schalkneethling:
Looks like kumascript does not yet have an implementation to expose the environment URL for Kuma. @escattone am I correct? If so, would it be a lot of effort to implement and expose on the `iframe` either as a URL param or a data attribute?

I know the last part is simple, not sure about the bits leading up to that. Thanks!

jwhitlock:
@schalkneethling use something like this:

```
<% var parts = kuma.url.parse(env.url || 'https://developer.mozilla.org'),
       base_url = parts.protocol + '//' + parts.host + '/'; %>
<p>Kuma base URL is: <%= base_url %></p>
```

PR #4426, opened 2017-09-14
bug 1308630: add env overrides for aws

escattone:
This PR implements the env var overrides requested in https://github.com/mozmeao/infra/issues/477.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -34,9 +34,15 @@ DEBUG = config('DEBUG', default=False, cast=bool)
 
 ROOT = dirname(dirname(dirname(os.path.abspath(__file__))))
 
-ADMINS = config('ADMIN_EMAILS',
-                default='mdn-dev@mozilla.com',
-                cast=TupleCsv())
+ADMINS = zip(
+    config('ADMIN_NAMES', default='MDN devs', cast=Csv()),
+    config('ADMIN_EMAILS', default='mdn-dev@mozilla.com', cast=Csv())
+)
+CELERYD_MAX_TASKS_PER_CHILD = config(
+    'CELERYD_MAX_TASKS_PER_CHILD',
+    default=0,
+    cast=int
+)
 
 PROTOCOL = config('PROTOCOL', default='https://')
 DOMAIN = config('DOMAIN', default='developer.mozilla.org')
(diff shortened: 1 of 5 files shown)


kuma/settings/common.py
```diff
@@ -468,7 +480,11 @@ def enable_candidate_languages():
 )
 AUTH_USER_MODEL = 'users.User'
 USER_AVATAR_PATH = 'uploads/avatars/'
-DEFAULT_AVATAR = STATIC_URL + 'img/avatar.png'
+DEFAULT_AVATAR = config(
+    'DEFAULT_AVATAR',
+    default=STATIC_URL + 'img/avatar.png'
```
safwanrahman: I think it should be `STATIC_URL + config(......)`
The environment variable should able to override the file name only. the link should be serve as STATIC_URL.

kuma/settings/common.py
```diff
@@ -34,9 +34,15 @@ def __call__(self, value):
...
-                cast=TupleCsv())
+ADMINS = zip(
+    config('ADMIN_NAMES', default='MDN devs', cast=Csv()),
+    config('ADMIN_EMAILS', default='mdn-dev@mozilla.com', cast=Csv())
+)
+CELERYD_MAX_TASKS_PER_CHILD = config(
+    'CELERYD_MAX_TASKS_PER_CHILD',
+    default=0,
+    cast=int
+)
```
jwhitlock: The default needs to be ``None``, not 0. Unfortunately, you can't use ``default=None``. This worked for me:

```
CELERYD_MAX_TASKS_PER_CHILD = config(
    'CELERYD_MAX_TASKS_PER_CHILD',
    default=0,
    cast=int
) or None
```

escattone: Thanks @jwhitlock!

jwhitlock: Oh, while you are changing it - please move it down to https://github.com/escattone/kuma/blob/9cdce16cc5e844f0674518abc5275d312ef84f46/kuma/settings/common.py#L1274 , where the other ``CELERYD_*`` settings are.

escattone:
Thanks @jwhitlock and @safwanrahman. I've submitted changes based on your feedback, as well as updating the `OPTIONS` key in the default database to match production so we don't need to provide additional env settings for that. The production settings seem to work for Docker.

escattone:
I added back the `TEST` key with its `CHARSET` and `COLLATION` keys.

kuma/settings/common.py
```diff
@@ -34,9 +34,15 @@ def __call__(self, value):
...
 ROOT = dirname(dirname(dirname(os.path.abspath(__file__))))
 
-ADMINS = config('ADMIN_EMAILS',
-                default='mdn-dev@mozilla.com',
-                cast=TupleCsv())
+ADMINS = zip(
+    config('ADMIN_NAMES', default='MDN devs', cast=Csv()),
+    config('ADMIN_EMAILS', default='mdn-dev@mozilla.com', cast=Csv())
+)
+CELERYD_MAX_TASKS_PER_CHILD = config(
```
jwhitlock: This would be better with the other ``CELERYD_*`` settings, around line 1270

PR #4429, opened 2017-09-15
bug 1308630: Add some security headers

jwhitlock:
Enable [Django's security middleware](https://docs.djangoproject.com/en/1.11/ref/middleware/#module-django.middleware.security), and some of the headers provided:

* ``X-Content-Type-Options: nosniff ``- always on
* ``X-XSS-Protection: 1; mode=block`` - always on
* ``Strict-Transport-Security: max-age=63072000`` - configured from environment

PR #4430, opened 2017-09-15
bug 1218563: Create optional SSL docker-compose config

jwhitlock:
This rebases and resurrects PR #4010, with several changes:

* Switches from [YAML node anchors](https://github.com/docker/compose/issues/1988#issuecomment-153377186) to ``docker-compose`` 2.x's ``extends`` method for copying the environment and other config items to the ``web``, ``worker``, and ``api`` containers.  We can start teasing apart the settings per container in the future.
* Uses ``--worker-class=meinheld.gmeinheld.MeinheldWorker`` in more places
* Keeps the ``web`` container as ``http://localhost:8000``, and adds ``nginx`` and the new ``webssl`` to serve ``http://localhost`` and ``https://localhost``, in a new optional ``docker-compose.ssl.yml``. I don't think it makes sense to use SSL all the time for local development, since the experience with a self-signed certificate is pretty bad.
* Removes ``static`` serving from ``nginx``, and uses mounting to configure the default ``nginx`` image.
* Generates the self-signed certs in the ``webssl`` container in a script, since ``nginx`` doesn't have OpenSSL.
* Documented, with sample config line in ``.env-dist.dev``.

This is a big change, and low priority compared to AWS stuff.  I used it to test PR #4429 in branch https://github.com/jwhitlock/kuma/tree/security-wip-1308630.


docker-compose.ssl.yml
```diff
@@ -0,0 +1,32 @@
...
+    environment:
+      - ACCOUNT_DEFAULT_HTTP_PROTOCOL=https
+      - ALLOWED_HOSTS=localhost
+      - ATTACHMENT_HOST=localhost
+      - CSRF_COOKIE_SECURE=True
+      - PROTOCOL=https://
+      - SESSION_COOKIE_SECURE=True
+      - SITE_URL=https://localhost
+    ports:
+      - "8443:8000"
```
safwanrahman: Why `8443` port?

etc/nginx/nginx.conf
```diff
@@ -0,0 +1,52 @@
...
+    keepalive_timeout 65;
+
+    upstream django {
+        server webssl:8000;
+    }
+
+    server {
+        listen 80;
+        charset utf-8;
+        return 301 https://$host$request_uri;
```
safwanrahman: This should not be `301` in dev. Otherwise, when the user run with non ssl, the browser will redirect to `https`

docker-compose.ssl.yml
jwhitlock: 443 is the SSL port, so 8443 makes some sense. There's no rule that I know of, just what I've used on other projects.  It's for debugging, and could be removed.

safwanrahman: I think this port should not be exposed. As its already running the `nginx` container that can connect the `gunicorn` with internal network.

jwhitlock:
Rebase on master, don't export the ``webssl`` HTTP port, and fix a spelling error.

safwanrahman:
`extend` is removed in docker 3.x
https://github.com/moby/moby/issues/31101

jwhitlock:
Rebased on master, merged some commits.

@safwanrahman thanks for the heads up on removal in 3.x.  It looks like the main reason to upgrade to 3.x format is for the cloud deployment support, which I don't plan on using.  Alternatively, we could use template generation to create ``docker-compose.yml`` and the alternate environments for each service.

jwhitlock:
I've rebased, and worked the removal of ``PROD_DETAILS_DIR`` into the PR.

I think this is a candidate for closing without merging - I haven't needed local SSL since September, and it would give us issues updating to later ``docker-compose`` formats.

jwhitlock:
Rebased again, but I still think this is a dead end, so closing the PR. I'll keep the branch in my repo to use if needed.

PR #4434, opened 2017-09-19
Fix bug 1401135: Wrap task completion CTA in gettext

stephaniehobson:
And make task-completion.js file conform to coding style guide.

### DIFF
--- a/kuma/static/js/task-completion.js
+++ b/kuma/static/js/task-completion.js
@@ -17,7 +17,8 @@
     function inquire() {
 
         // dimension14 is "Saw Survey Gizmo Task Completion survey"
-        if(win.ga) ga('set', 'dimension14', 'Yes');
+        ga('set', 'dimension14', 'Yes');
+
         ga('send', 'event', 'survey', 'prompt', 'impression', {
             nonInteraction: true
         });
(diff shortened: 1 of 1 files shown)


kuma/static/js/task-completion.js
```diff
@@ -17,7 +17,9 @@
     function inquire() {
 
         // dimension14 is "Saw Survey Gizmo Task Completion survey"
-        if(win.ga) ga('set', 'dimension14', 'Yes');
+        if(win.ga) {
+            ga('set', 'dimension14', 'Yes');
+        }
```
jwhitlock: Should the next line (``ga('send'...``) also be in this block?

kuma/static/js/task-completion.js
```diff
@@ -27,7 +29,9 @@
         var clickTime = Date.now() + (1000*60)*20; // 20 min from now
         var surveyURL= 'https://www.surveygizmo.com/s3/2980494/do-or-do-not';
         var surveyLink = surveyURL + '?' + '&t='+ clickTime + '&p=' + path;
-        var ask = "Would you answer 4 questions for us? <a id='task-link' target='_blank' href='" + surveyLink + "'>Open the survey in a new tab</a> and fill it out when you're done on the site. Thanks!";
+        var surveyAttributes = 'id=\'task-link\' target=\'_blank\' href=\'' + surveyLink + '\'';
+        var ask = interpolate(gettext('Would you answer 4 questions for us? <a %(url)s>Open the survey in a new tab</a> and fill it out when you are done on the site. Thanks!'), { url : surveyAttributes }, true);
+
```
jwhitlock: This extracts, interpolates correctly 👍 ``url`` was an unexpected name for "all the anchor tag attributes", but it works.

stephaniehobson: I was thinking of what the translators will see more than what the devs see ;)

kuma/static/js/task-completion.js
stephaniehobson: Hm. It would not have been part of the same if block in the original code. So this PR does not change that functionality... but you pose a valid question.

stephaniehobson: I settled on removing the if statement altogether. It's not around the other ga() calls on the page and not necessary.

The ga object is there even if the user blocks the script from loading the rest of the google code.

jwhitlock: I think the exception is in development mode, when assets are not combined, and the developer's ad blocker blocks ``analytics.js``.  Not needed to protect production, and developers can diagnose and fix.

PR #4437, opened 2017-09-20
bug 1308630: add ROBOTS_ROOT env override

escattone:
This PR provides a way to override just the document root for serving the robots files, which in AWS we'll override to `/app/media` instead of using the default value which is the value of `MEDIA_ROOT` (`/mdn/www`). In SCL3, and every other context, we'll continue to use the default value (`MEDIA_ROOT`).

### DIFF
--- a/kuma/landing/test_views.py
+++ b/kuma/landing/test_views.py
@@ -3,6 +3,20 @@ import pytest
 from kuma.core.urlresolvers import reverse
 
 
+@pytest.fixture()
+def robots(tmpdir):
+    path = tmpdir.realpath()
+    robots_file = tmpdir.join('robots.txt')
+    robots_go_away_file = tmpdir.join('robots-go-away.txt')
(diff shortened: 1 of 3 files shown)


safwanrahman:
One thing that catch up my mind, why do we serve the `robots.txt` from media root? Is not its a static thing that change from the code? We can store it in static root or serve by template to get the same result in SCL3 as well as AWS

escattone:
On second thought, it's simplest for now to just serve these files in all contexts (SCL3, AWS, etc.) from `kuma.settings.common.path('media')`, something like:
```
from kuma.settings.common import path
...
def robots_txt(request):
    ...
    return static.serve(request, robots, document_root=path('media'))
```
I'm closing this and opening a new PR that does that.

escattone:
See https://github.com/mozilla/kuma/pull/4438

PR #4441, opened 2017-09-25
bug 1402950: Be patient for external links

jwhitlock:
* When opening links in a new tab, wait for the tab to open.
* When opening GitHub in the same window, wait for the URL to change.

This helps address some tests that are flakier in Firefox than Chrome.


jwhitlock:
Rebased on master, and replaced the premature generic ``wait_for_window`` decorator with a simpler but specific decorator.

PR #4445, opened 2017-09-28
bug 1110799: Update raven, expand configuration

jwhitlock:
``raven`` is the Python Sentry client.  This PR:

* Updates raven from 5.10.2 → 6.2.1 (no gotchas in the [changelog](https://github.com/getsentry/raven-python/blob/master/CHANGES))
* Sends the revision hash, for release tracking
* Sets a syncronous transport, which may be more compatible with meinheld
* Sets the repos name, which may make tracebacks more useful
* Ignores DisallowedHost exceptions, which happen in AWS

This extra config won't happen in AWS. It will just get ``SENTRY_DSN`` set by the local settings override file, which [continues to work as a way to specify the dsn](https://github.com/getsentry/raven-python/blob/24c0ecd44617d9fc916bf4c9c4a0ce4ef18ea05e/raven/utils/conf.py#L49).

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1681,6 +1681,25 @@ REST_FRAMEWORK = {
 SENTRY_DSN = config('SENTRY_DSN', default=None)
 
 if SENTRY_DSN:
+    from raven.transport.requests import RequestsHTTPTransport
+    RAVEN_CONFIG = {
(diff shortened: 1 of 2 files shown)


escattone:
Thanks @jwhitlock !

PR #4446, opened 2017-09-28
Restrict endpoints (bug 1402708)

escattone:
This PR address at least part of [bugzilla 1402708](https://bugzilla.mozilla.org/show_bug.cgi?id=1402708) and https://github.com/mozmeao/infra/issues/529. Currently in SCL3, there are three domains in use: **main** (developer.mozilla.org), **untrusted** (mdn.mozillademos.org), and **cdn** (developer.cdn.mozilla.net). Here's a sample of what each of the domains in SCL3  returns for specific endpoints:

### **main** (developer.mozilla.org):
- should NOT serve sample endpoints (https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color) and should redirect on file-attachment endpoints (https://developer.mozilla.org/files/12984/web-font-example.png)
- no new code needed since these endpoints already check for the proper HOST header and redirect if necessary (must be `ATTACHMENT_HOST` for file attachments and the constance setting `KUMA_WIKI_IFRAME_ALLOWED_HOSTS` for samples)
- currently does serve static/media files (via WhiteNoise)

https://developer.mozilla.org/admin/login/ --> 200
https://developer.mozilla.org/en-US/docs/Web/JavaScript --> 200
https://developer.mozilla.org/media/sitemap.xml --> 200
https://developer.mozilla.org/static/build/js/wiki-skinny.2187d581918b.js --> 200
https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color --> 403
https://developer.mozilla.org/files/12984/web-font-example.png --> 301 to https://mdn.mozillademos.org/files/12984/web-font-example.png
https://developer.mozilla.org/@api/deki/files/2926/=AudioTest_(1).ogg --> 301 to https://mdn.mozillademos.org/files/2587/AudioTest%20(1).ogg
https://developer.mozilla.org/diagrams/workflow/workflow.svg --> 200
https://developer.mozilla.org/presentations/javascript2/images/big-cake.jpg --> 200
https://developer.mozilla.org/samples/svg/swarm-of-motes.xhtml --> 200

### **untrusted** (mdn.mozillademos.org):
- should only serve the file attachment and code sample endpoints
- should not serve any other endpoints, like static, media, or document endpoints

https://mdn.mozillademos.org/admin/login/ --> 404
https://mdn.mozillademos.org/en-US/docs/Web/JavaScript --> 404
https://mdn.mozillademos.org/media/sitemap.xml --> 404
https://mdn.mozillademos.org/static/build/js/wiki-skinny.2187d581918b.js --> 404
https://mdn.mozillademos.org/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color --> 200
https://mdn.mozillademos.org/files/12984/web-font-example.png --> 200
https://mdn.mozillademos.org/@api/deki/files/2926/=AudioTest_(1).ogg --> 404
https://mdn.mozillademos.org/diagrams/workflow/workflow.svg --> 404
https://mdn.mozillademos.org/presentations/javascript2/images/big-cake.jpg --> 404
https://mdn.mozillademos.org/samples/svg/swarm-of-motes.xhtml --> 404

### **cdn** (developer.cdn.mozilla.net):
- currently in SCL3 serves all endpoints including static, media, admin, and documents

https://developer.cdn.mozilla.net/admin/login/ --> 200
https://developer.cdn.mozilla.net/en-US/docs/Web/JavaScript --> 200
https://developer.cdn.mozilla.net/media/sitemap.xml --> 200
https://developer.cdn.mozilla.net/static/build/js/wiki-skinny.2187d581918b.js --> 200
https://developer.cdn.mozilla.net/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color --> 403
https://developer.cdn.mozilla.net/files/12984/web-font-example.png --> 301 to https://mdn.mozillademos.org/files/12984/web-font-example.png
https://developer.cdn.mozilla.net/@api/deki/files/2926/=AudioTest_(1).ogg --> 301 to https://mdn.mozillademos.org/files/2587/AudioTest%20(1).ogg
https://developer.cdn.mozilla.net/diagrams/workflow/workflow.svg --> 200
https://developer.cdn.mozilla.net/presentations/javascript2/images/big-cake.jpg --> 200
https://developer.cdn.mozilla.net/samples/svg/swarm-of-motes.xhtml --> 200

So this PR effectively makes the Kuma-based code in AWS match SCL3 (the results above). In brief, it blocks the **untrusted** domain from serving anything other than code samples and file attachments. It does NOT restrict the **cdn** domain, which goes beyond what we currently do in SCL3. I can do that in this PR or another if that is what we want. If we do want to restrict the **cdn** domain, do we want to restrict it to just serve static files?

I did the following in order to test this in my local Docker development environment:
- changed `/etc/hosts` by adding a `demos` alias to `127.0.0.1`
```
##
# Host Database
#
##
127.0.0.1	localhost demos
255.255.255.255	broadcasthost
::1             localhost
```
- added `ENABLE_RESTRICTIONS_BY_HOST=True` to my `.env` file
- set `ATTACHMENT_HOST=demos:8000` in my `docker-compose.yml` file
- built a new `kuma_base` Docker image: `make build-base` (This is required due to the upgrade of the whitenoise package to version 3.3.1)
- changed line 4 of my `docker-compose.yml` to use the Docker image hash (e.g., `6c7af4df819c`) from the `make build-base` command above rather than `quay.io/mozmar/kuma_base`
- logged into  the Django admin (`localhost:8000/admin/login`) and added `demos:8000` to the constance regex setting `KUMA_WIKI_IFRAME_ALLOWED_HOSTS`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,7 @@ services:
       - DATABASE_URL=mysql://${DATABASE_USER:-root}:${DATABASE_PASSWORD:-kuma}@mysql:3306/developer_mozilla_org
       - DEBUG=${DEBUG:-True}
       - DOMAIN=localhost
+      - ENABLE_RESTRICTIONS_BY_HOST=${ENABLE_RESTRICTIONS_BY_HOST:-False}
       - ES_URLS=elasticsearch:9200
       - INTERACTIVE_EXAMPLES_BASE=${INTERACTIVE_EXAMPLES_BASE:-https://interactive-examples.mdn.mozilla.net}
       - KUMASCRIPT_URL_TEMPLATE=http://kumascript:9080/docs/{path}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       # Django settings overrides:
       - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
       - ALLOWED_HOSTS=*
-      - ATTACHMENT_HOST=localhost:8000
+      - ATTACHMENT_HOST=${ATTACHMENT_HOST:-localhost:8000}
       - BROKER_URL=redis://redis:6379/0
       - CELERY_ALWAYS_EAGER=False
       - CSRF_COOKIE_SECURE=False
@@ -32,6 +32,7 @@ services:
       - PROTOCOL=http://
       - SESSION_COOKIE_SECURE=False
       - SITE_URL=http://localhost:8000
+      - STATIC_URL=${STATIC_URL:-/static/}
       # Other environment overrides
       - PYTHONDONTWRITEBYTECODE=1
       - PYTHONUNBUFFERED=True
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -7,6 +7,7 @@ from django.http import HttpResponseForbidden, HttpResponsePermanentRedirect, Ht
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
 from django.contrib.sessions.middleware import SessionMiddleware
+from whitenoise.middleware import WhiteNoiseMiddleware
 
 from .urlresolvers import Prefixer, set_url_prefixer, split_path
 from .utils import urlparams
@@ -166,3 +167,28 @@ class ForceAnonymousSessionMiddleware(SessionMiddleware):
         Override the base-class method to ensure we do nothing.
         """
         return response
+
+
+class RestrictedEndpointsMiddleware(object):
+
+    def process_request(self, request):
+        """
+        Restricts the accessible endpoints based on the host.
+        """
+        if (settings.ENABLE_RESTRICTIONS_BY_HOST and
+                (request.get_host() == settings.ATTACHMENT_HOST)):
+            request.urlconf = 'kuma.urls_untrusted'
+
+
+class RestrictedWhiteNoiseMiddleware(WhiteNoiseMiddleware):
+
+    def process_request(self, request):
+        """
+        Restricts the use of WhiteNoiseMiddleware based on the host.
+        """
+        if (settings.ENABLE_RESTRICTIONS_BY_HOST and
+                (request.get_host() == settings.ATTACHMENT_HOST)):
+            return None
+        return super(RestrictedWhiteNoiseMiddleware, self).process_request(
+            request
+        )
--- a/kuma/core/tests/test_middleware.py
+++ b/kuma/core/tests/test_middleware.py
@@ -1,11 +1,14 @@
-from mock import MagicMock
+from mock import patch, MagicMock
 
 from django.test import RequestFactory
 
 from kuma.core.tests import KumaTestCase, eq_
 from kuma.core.middleware import (
+    WhiteNoiseMiddleware,
     SetRemoteAddrFromForwardedFor,
-    ForceAnonymousSessionMiddleware
+    ForceAnonymousSessionMiddleware,
+    RestrictedEndpointsMiddleware,
+    RestrictedWhiteNoiseMiddleware,
 )
 
 
@@ -56,3 +59,42 @@ def test_force_anonymous_session_middleware(rf, settings):
     response = middleware.process_response(request, MagicMock())
 
     assert not response.method_calls
+
+
+def test_restricted_endpoints_middleware(rf, settings):
+    settings.ATTACHMENT_HOST = 'demos'
+    settings.ENABLE_RESTRICTIONS_BY_HOST = True
+    middleware = RestrictedEndpointsMiddleware()
+
+    request = rf.get('/foo', HTTP_HOST='demos')
+    middleware.process_request(request)
+    assert request.urlconf == 'kuma.urls_untrusted'
+
+    request = rf.get('/foo', HTTP_HOST='not-demos')
+    middleware.process_request(request)
+    assert not hasattr(request, 'urlconf')
+
+    settings.ENABLE_RESTRICTIONS_BY_HOST = False
+    request = rf.get('/foo', HTTP_HOST='demos')
+    middleware.process_request(request)
+    assert not hasattr(request, 'urlconf')
+
+
+def test_restricted_whitenoise_middleware(rf, settings):
+    settings.ATTACHMENT_HOST = 'demos'
+    settings.ENABLE_RESTRICTIONS_BY_HOST = True
+    middleware = RestrictedWhiteNoiseMiddleware()
+
+    sentinel = object()
+
+    with patch.object(WhiteNoiseMiddleware, 'process_request',
+                      return_value=sentinel):
+        request = rf.get('/foo', HTTP_HOST='demos')
+        assert middleware.process_request(request) is None
+
+        request = rf.get('/foo', HTTP_HOST='not-demos')
+        assert middleware.process_request(request) is sentinel
+
+        settings.ENABLE_RESTRICTIONS_BY_HOST = False
+        request = rf.get('/foo', HTTP_HOST='demos')
+        assert middleware.process_request(request) is sentinel
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -262,7 +262,8 @@ wheel==0.29.0 \
 
 # Serve static files with compression, etc. from Django
 whitenoise==3.3.1 \
-    --hash=sha256:15f43b2e701821b95c9016cf469d29e2a546cb1c7dead584ba82c36f843995cf
+    --hash=sha256:15f43b2e701821b95c9016cf469d29e2a546cb1c7dead584ba82c36f843995cf \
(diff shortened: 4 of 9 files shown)


escattone:
I've been so caught up in the details, that I forgot to add tests. 🙁

jwhitlock:
> - built a new `kuma_base` Docker image: `make build-base` (This is required due to the upgrade of the whitenoise package to version 3.3.1)
> - changed line 4 of my `docker-compose.yml` to use the Docker image hash (e.g., `6c7af4df819c`) from the `make build-base` command above rather than `quay.io/mozmar/kuma_base`

I use ``VERSION=latest make build-base`` to build and tag as ``latest``, and skip changing ``docker-compose.yml``. ``docker-compose pull web`` will replace it with the current ``latest`` from quay.io. ``docker-compose pull`` will too, but also all the other images.

requirements/default.txt
```diff
@@ -261,9 +261,8 @@ wheel==0.29.0 \
     --hash=sha256:1ebb8ad7e26b448e9caa4773d2357849bf80ff9e313964bcaf79cbf0201a1648
 
 # Serve static files with compression, etc. from Django
-whitenoise==2.0.6 \
-    --hash=sha256:826ffe5d608c9dc8daebef1b0b43d01f7958f17c2fce36e75c80e26160172c4f \
-    --hash=sha256:5aea935dfc09ef2beeb76960b4a808b0bbe65e85fb0b0312434b9c365ca02a41
+whitenoise==3.3.1 \
+    --hash=sha256:15f43b2e701821b95c9016cf469d29e2a546cb1c7dead584ba82c36f843995cf
```
jwhitlock: This is missing a hash. It should be:

```
whitenoise==3.3.1 \
   --hash=sha256:15f43b2e701821b95c9016cf469d29e2a546cb1c7dead584ba82c36f843995cf \
   --hash=sha256:9d81515f2b5b27051910996e1e860b1332e354d9e7bcf30c98f21dcb6713e0dd
```

Please use ``hashin whitenoise -r requirements/default.txt`` to get all the hashes when [updating requirements](https://github.com/mozilla/kuma/tree/master/requirements#hash-checking-mode).

docker-compose.yml
```diff
@@ -23,6 +23,7 @@ services:
       - DATABASE_URL=mysql://${DATABASE_USER:-root}:${DATABASE_PASSWORD:-kuma}@mysql:3306/developer_mozilla_org
       - DEBUG=${DEBUG:-True}
       - DOMAIN=localhost
+      - ENABLE_RESTRICTIONS_BY_HOST=${ENABLE_RESTRICTIONS_BY_HOST:-False}
```
jwhitlock: I think it would be useful if ``ATTACHMENT_HOST`` could also be overridden by ``.env``.

requirements/default.txt
escattone: 👍 I did not know about `hashin` and that we had this documented, thanks @jwhitlock !

escattone:
@jwhitlock You were able to load a sample from `localhost:8000` because the samples endpoint is restricted by the `KUMA_WIKI_IFRAME_ALLOWED_HOSTS` constance setting, and `localhost:8000` is one of the allowed values. For local testing, we'll have to temporarily remove that value.

jwhitlock:
@escattone thanks, that makes sense now that I see the code again:

https://github.com/mozilla/kuma/blob/c7b8ea91fe09d93df4fcd8ad6192496c4b412839/kuma/wiki/views/code.py#L26-L29

``KUMA_WIKI_IFRAME_ALLOWED_HOSTS`` is my least favorite setting.  I hope to replace it soon.

escattone:
> KUMA_WIKI_IFRAME_ALLOWED_HOSTS is my least favorite setting. I hope to replace it soon.
👍 I'm in total agreement! 😄

kuma/core/tests/test_middleware.py
```diff
@@ -1,11 +1,14 @@
-from mock import MagicMock
+from mock import patch, MagicMock
 
 from django.test import RequestFactory
 
 from kuma.core.tests import KumaTestCase, eq_
 from kuma.core.middleware import (
+    WhiteNoiseMiddleware,
```
safwanrahman: As we are using `RestrictedWhiteNoiseMiddleware` which inherit `WhiteNoiseMiddleware`, is it necessary to use the `WhiteNoiseMiddleware`?

safwanrahman: If `WhiteNoiseMiddleware` should be used, it should be import from `whitenoise.middleware`

kuma/core/middleware.py
```diff
@@ -14,6 +15,11 @@
 from .views import handler403
 
 
+SITE_HOST = urlparse(settings.SITE_URL).netloc
```
escattone: I'm not sure if this approach is a good one or not. I thought it was nice that it was "automatic", in the sense that no new environment variables needed to be added, and it captures (I think) all of the current use cases (local development, testing, SCL3, AWS), but on the other hand, adding an environment variable, something like `CDN_HOST` would make this more configurable. @jwhitlock I would specifically like your opinion on this.

kuma/core/middleware.py
```diff
@@ -14,6 +15,11 @@
...
+CDN_HOST = urlparse(settings.STATIC_URL).netloc
+USING_CDN = CDN_HOST and (CDN_HOST != SITE_HOST)
```
escattone: I'm not certain about this approach of extracting a "valid" `CDN_HOST`. I like that it's "automatic", in that it requires no new environment variables while at the same time capturing (I think) our actual practice of placing the CDN host into the `STATIC_URL`. On the other hand, it's less configurable that simply adding another variable, like `CDN_HOST`, that can be set from the environment. @jwhitlock What are your thoughts?

kuma/core/tests/test_middleware.py
escattone: @safwanrahman I'm importing and using `WhiteNoiseMiddleware` only because I'm patching it's `process_request` method with a `MagicMock` instance.

escattone:
Here are my latest notes on testing locally (updated with @jwhitlock's suggestion):

I did the following in order to test this in my local Docker development environment:
- changed `/etc/hosts` by adding the `demos` and `cdn` aliases to `127.0.0.1`
```
##
# Host Database
#
##
127.0.0.1	localhost demos cdn
255.255.255.255	broadcasthost
::1             localhost
```
- added `ENABLE_RESTRICTIONS_BY_HOST=True` to my `.env` file
- added `ATTACHMENT_HOST=demos:8000` to my `.env` file
- added `STATIC_URL=http://cdn:8000/static/` to my `.env` file
- built a new `kuma_base` Docker image: `VERSION=latest make build-base` (This is required due to the upgrade of the whitenoise package to version 3.3.1)
- logged into  the Django admin (`localhost:8000/admin/login`) and added `demos:8000` as well as removed `localhost:8000` from the constance regex setting `KUMA_WIKI_IFRAME_ALLOWED_HOSTS`

kuma/core/middleware.py
jwhitlock: I think it needs to be a new setting. On SCL3, ``STATIC_URL`` is ``'https://developer.cdn.mozilla.net/static/'``, but that's the user-facing URL. I think the CDN calls ``'https://developer-origin.cdn.mozilla.net/static/'`` when it needs to request a file it hasn't seen before. I'm not sure what the incoming domain will be in AWS - we may need a list of CDN hosts...

escattone: I see that I was "having it both ways", knowing that `developer.cdn.mozilla.net` was the user-facing URL, and at the same time thinking it was how the CDN would make requests to MDN. Ugh, that's embarrassing. Yes, we'll need a list of CDN hosts, thanks @jwhitlock.

kuma/core/tests/test_middleware.py
safwanrahman: @escattone I think in that case it should be imported from `whitenoise.middleware`

escattone:
@metadave @jwhitlock and I met this morning and decided that for now we would not restrict the endpoints that the **cdn** domain(s) would serve, which has been the current practice in SCL3. This amounts to backing out https://github.com/mozilla/kuma/pull/4446/commits/d1a04dcbb1b836d2703eeb1298848cb75aacca2a and then moving forward with this PR again.

escattone:
I re-based this PR (dropping the commit that restricts the endpoints served by the **cdn** domain). @jwhitlock it's ready for your review again!

kuma/core/tests/test_middleware.py
jwhitlock: I'd normally write it the way @safwanrahman suggested, but because of the usage (check if the parent class is called), this way works too, and may be preferable. No change.

PR #4450, opened 2017-10-03
updating scss for .prevnext to restrict prevnext menu line width

chrisdavidmills:
As discussed with Stephanie.

### DIFF
--- a/kuma/static/styles/components/wiki/customcss.scss
+++ b/kuma/static/styles/components/wiki/customcss.scss
@@ -12,8 +12,8 @@ eventually emptying this file.
 ********************************************************************** */
 
 
-.prevnext p {
-    @include full-width-content();
+.prevnext {
+    @include restrict-line-length();
 }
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
@stephaniehobson

stephaniehobson:
@chrisdavidmills You managed to bundle the kumascript submodule into this PR, you'll have to extract it :/

chrisdavidmills:
Dammnit, I managed to do this last time I made a Kuma update too. How do I make sure I don't do this?

stephaniehobson:
The only way I've found is to avoid `git add *`. Since I'm always working on styles `git add kuma/*` works for me.

It's also possible to update the kumascript modules by doing `git submodule update`, but that doesn't always get me out of trouble, which is why I've been more careful about adding files intentionally.

chrisdavidmills:
Ah, that makes sense. I've become too used to just using git add * all the time. right, I'll close this request and try to fix this with a new one.

PR #4455, opened 2017-10-06
Fix issue #283, add dns-prefetch, preconnect for editor iframe

schalkneethling:
Adds `dns-prefetch` and `preconnect` for the editor `iframe` to test the performance impact.

### DIFF
--- a/kuma/wiki/jinja2/wiki/document.html
+++ b/kuma/wiki/jinja2/wiki/document.html
@@ -64,6 +64,9 @@
 {% endblock %}
 
 {% block extrahead %}
+  <link rel="dns-prefetch" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
+  <link rel="preconnect" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
+
   <link rel="alternate" type="application/json" href="{{ url('wiki.json_slug', document.slug) }}">
   <link rel="canonical" href="{{ canonical }}" >
 


kuma/wiki/jinja2/wiki/base.html
```diff
@@ -19,6 +19,11 @@
 
 {% endblock %}
 
+{% block extrahead %}
+    <link rel="dns-prefetch" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
+    <link rel="preconnect" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
+{% endblock %}
```
schalkneethling: I felt confident that this is the correct file but, now I am not so sure. When I look at a page such as `http://127.0.0.1:8000/en-US/docs/Experiment:InteractiveEditor/Array.prototype.reduce()` I do not see the `link` elements in the header :(

jwhitlock: That's because it is overridden in ``document.html``, which is a better place for this code:

https://github.com/mozilla/kuma/blob/a608faaa7c190b094fe15eda33877050e34ecb95/kuma/wiki/jinja2/wiki/document.html#L66-L86

stephaniehobson:
Do we think there are any downsides to adding it to all pages? 

I think it is possible to include the code only if the interactive examples are on the page. We do some custom configuration on the kuma side in response to the existence of the `quick-links` class for left side menus. (I'm not sure enough about how it works to give instructions though).

schalkneethling:
> I think it is possible to include the code only if the interactive examples are on the page.

@stephaniehobson That would be great, as long as it is not injected with JS. I cannot find the doc now, but I remember recently hearing something about Chrome not behaving predictably if you inject these tags with JS. I will try and find the reference to it. Perhaps it is under x condition which does not apply here.

schalkneethling:
@stephaniehobson The seems to be what you are referring to? https://github.com/mozilla/kuma/blob/master/kuma/wiki/jinja2/wiki/document.html#L316

and 

https://github.com/mozilla/kuma/blob/master/kuma/wiki/jinja2/wiki/document.html#L28

and

https://github.com/mozilla/kuma/blob/master/kuma/wiki/views/document.py#L698

and 

https://github.com/mozilla/kuma/blob/master/kuma/wiki/views/document.py#L722

stephaniehobson:
Yes! That's it exactly. Great detective work :)

PR #4461, opened 2017-10-13
bug 1407806: improve caching for file attachments

escattone:
* use Django-supplied decorators to add "Cache-Control" header as well as handle "Last-Modified" and its related request headers like "If-Modified-Since"
* allow the "max-age" caching-header value for file attachments to be configured from the environment via `ATTACHMENTS_CACHE_CONTROL_MAX_AGE`
* block Django's `SessionMiddleware` from adding the "Vary: Cookie" header
* move some generally-useful pytest fixtures from `kuma/wiki/tests/conftest.py` to the top level
  `kuma/conftest.py`
* add `file_attachment` fixture to `kuma/attachments/tests/conftest.py`
* convert/modify some of the old-style tests in `kuma/attachments/tests/test_views.py` to the
  pytest style and add one new test for the file-attachments endpoint
* add `convert_to_utc()` function to `kuma/attachments/utils.py`

### DIFF
--- a/kuma/attachments/tests/conftest.py
+++ b/kuma/attachments/tests/conftest.py
@@ -1,0 +1,31 @@
+import datetime
+
+import pytest
+from django.core.files.base import ContentFile
+
+from kuma.attachments.models import Attachment, AttachmentRevision
+
+
+@pytest.fixture
+def file_attachment(db, wiki_user):
+    file_id = 97
+    filename = 'test.txt'
+    title = 'Test text file'
+
+    attachment = Attachment(title=title, mindtouch_attachment_id=file_id)
+    attachment.save()
+    revision = AttachmentRevision(
+        title=title,
+        is_approved=True,
+        attachment=attachment,
+        mime_type='text/plain',
+        description='Initial upload',
+        created=datetime.datetime.now(),
+    )
+    revision.creator = wiki_user
+    revision.file.save(filename, ContentFile('This is only a test.'))
+    revision.make_current()
+    return dict(
+        attachment=attachment,
+        file=dict(id=file_id, name=filename),
+    )
--- a/kuma/attachments/views.py
+++ b/kuma/attachments/views.py
@@ -4,6 +4,8 @@ from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.http import Http404, StreamingHttpResponse
 from django.shortcuts import get_object_or_404, render, redirect
+from django.views.decorators.http import last_modified
+from django.views.decorators.cache import cache_control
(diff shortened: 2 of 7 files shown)


kuma/attachments/tests/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+    revision = AttachmentRevision(
+        title=title,
+        is_approved=True,
+        attachment=attachment,
+        mime_type='text/plain',
+        description='Initial upload',
+        created=datetime.datetime.now(),
+    )
+    revision.creator = wiki_user
+    revision.file.save(filename, ContentFile('This is only a test.'))
```
jwhitlock: This creates a new file "test.txt" for each test that uses this fixture. This isn't much worse than the old tests, but I suspect we could do better by making the actual file a session-scoped fixture, or even point it at an existing test file.  That may be a follow-on task.

kuma/conftest.py
```diff
@@ -1,8 +1,85 @@
...
+def wiki_user(db, django_user_model):
+    """A test user."""
+    return django_user_model.objects.create(
+        username='wiki_user',
+        email='wiki_user@example.com',
+        date_joined=datetime(2017, 4, 14, 12, 0))
+
+
+@pytest.fixture
+def wiki_user_2(db, django_user_model):
```
jwhitlock: I'm 👎 on adding all the potentially useful fixtures. I see narrowly scoped fixtures as a feature of pytest, versus the Django fixtures which tended to grow to cover many, many use cases. I'd just pull up the ``wiki_user`` and ``root_doc`` fixtures, and leave the rest until they are needed by multiple modules.  It's a weak preference, and I could be persuaded.

kuma/attachments/views.py
```diff
@@ -32,25 +34,39 @@ def raw_file(request, attachment_id, filename):
...
-        response['X-Frame-Options'] = 'ALLOW-FROM %s' % settings.DOMAIN
-        return response
+
+        def get_last_modified(*args):
+            return convert_to_utc(rev.created)
+
+        @cache_control(public=True,
+                       max_age=settings.ATTACHMENTS_CACHE_CONTROL_MAX_AGE)
+        @last_modified(get_last_modified)
+        def stream_raw_file(*args):
```
jwhitlock: What if this was a regular view instead of a nested function? ``kuma/urls_untrusted.py`` could use ``stream_raw_file`` directly, and ``kuma/urls.py`` could still use it in use ``raw_file`` for development.. Of course, that would take some rearranging of the attachment checks, so it may not be worth it.

escattone:
Thanks for the feedback @jwhitlock ! I'll address each of them in another PR so we can get this one into production.

PR #4463, opened 2017-10-13
bug 1392457, 1274874: Improve document list views

jwhitlock:
Improve special views that return a list of documents:

* Reduce data returned from database to that used in the page
* Eliminate second ``COUNT(*)`` query
* Remove redirects, user pages, and archive pages from the "needs review" and "needs translation" lists.

To Do:
* [x] Add tests for manager methods

This was originally PR #4378. These views are even slower post-AWS.

### DIFF
--- a/kuma/wiki/managers.py
+++ b/kuma/wiki/managers.py
@@ -68,34 +68,35 @@ class BaseDocumentManager(models.Manager):
             docs = docs.filter(parent_topic__isnull=True)
 
         # Only include fields needed for a list of links to docs
-        docs = docs.only('id', 'locale', 'slug', 'deleted', 'summary_text')
+        docs = docs.only('id', 'locale', 'slug', 'deleted', 'title',
+                         'summary_text')
         return docs
 
     def filter_for_review(self, locale=None, tag=None, tag_name=None):
         """Filter for documents with current revision flagged for review"""
-        query = 'current_revision__review_tags__%s'
+        docs = self.filter_for_list(locale=locale)
+        docs = docs.exclude(slug__startswith='Archive/')
+        filter_name = 'current_revision__review_tags__'
         if tag_name:
-            query = {query % 'name': tag_name}
+            docs = docs.filter(**{filter_name + 'name': tag_name})
         elif tag:
-            query = {query % 'in': [tag]}
+            docs = docs.filter(**{filter_name + 'in': [tag]})
         else:
-            query = {query % 'name__isnull': False}
-        if locale:
-            query['locale'] = locale
-        return self.filter(**query).distinct()
+            docs = docs.filter(**{filter_name + 'name__isnull': False})
+        return docs.distinct()
 
     def filter_with_localization_tag(self, locale=None, tag=None, tag_name=None):
         """Filter for documents with a localization tag on current revision"""
-        query = 'current_revision__localization_tags__%s'
+        docs = self.filter_for_list(locale=locale)
+        docs = docs.exclude(slug__startswith='Archive/')
+        filter_name = 'current_revision__localization_tags__'
         if tag_name:
-            query = {query % 'name': tag_name}
+            docs = docs.filter(**{filter_name + 'name': tag_name})
         elif tag:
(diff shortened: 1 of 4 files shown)


safwanrahman:
There should be a rate limit in this view. I see there are no rate limiting applied in this view.

kuma/wiki/managers.py
```diff
@@ -41,6 +41,10 @@ def get_by_stale_rendering(self):
 
     def filter_for_list(self, locale=None, tag=None, tag_name=None,
                         errors=None, noparent=None, toplevel=None):
+        """
+        Returns a filtered queryset for a list of names and urls.
+        """
+
         docs = (self.filter(is_redirect=False)
                     .exclude(slug__startswith='User:')
                     .exclude(slug__startswith='Talk:')
```
safwanrahman: I think `Template_talk` and `Project_talk` can be removed from here.
Did not see `User_talk` anywhere in MDN

jwhitlock: Here's the page counts for those prefixes:

|Prefix|Count|
|--|--|
|User:|3220|
|Talk:|2531|
|User_talk:|901|
|Template_talk:|17|
|Project_talk:|63|

It can be hard to find them, because we exclude them from the site search.

jwhitlock:
I've rebased, and added tests so that ``kuma/wiki/managers.py`` is covered.

My goal was to move these views to getting their data from a cacheback job. This way, a series of requests for documents by tag would get a bunch of pages saying "Please retry in a minute", and our celery workers would heat up the cluster, but the website should still be available.  That may still be useful, but I feel this is enough of a change to review.  This PR fixes [bug 1274874](https://bugzilla.mozilla.org/show_bug.cgi?id=1274874), and reduces the database load and traffic for these views.

@safwanrahman it is true the views do not have rate limiting. I'm not a fan of rate limiting for GET views, but instead think they should degrade gracefully under punishing traffic. In this case, we were seeing downtime from regular traffic to the new AWS cluster, and the most aggressive viewer was requesting a page a minute, so rate limiting would not have helped us.

The AWS cluster's performance has been better in recent weeks, but I want to lay the groundwork for better performance and for moving the work to a cacheback job, in case non-compliant scrapers are an issue. It will be easier to transition to a cacheback job after this change.

kuma/wiki/tests/test_managers.py
```diff
@@ -0,0 +1,251 @@
...
+    assert root_doc == Document.objects.get_by_natural_key(root_doc.locale,
+                                                           root_doc.slug)
+
+
+@pytest.mark.parametrize(
+    'legacy_slug',
+    ('User:ethertank',
+     'Talk:Developer_Guide/Build_Instructions/Windows_Prerequisites',
+     'User_talk:ethertank',
+     'Template_talk:anch'))
```
escattone: Nice use of `parametrize`, and seems nicer than creating a list of docs that includes all of the cases in one shot. I think a `Project_talk:` case needs to be added here as well?

kuma/wiki/managers.py
```diff
@@ -63,36 +67,36 @@ def filter_for_list(self, locale=None, tag=None, tag_name=None,
...
+            docs = docs.filter(**{filter_name + 'in': [tag]})
         else:
-            query = {query % 'name__isnull': False}
-        if locale:
-            query['locale'] = locale
-        return self.filter(**query).distinct()
+            docs = docs.filter(**{filter_name + 'name__isnull': False})
+        return docs.distinct()
 
     def filter_with_localization_tag(self, locale=None, tag=None, tag_name=None):
```
escattone: The `filter_with_localization_tag` and `filter_for_review` methods seem to be exactly the same except for the value of `filter_name`. It might be worthwhile to refactor these two methods to use the same base method under the hood?

kuma/wiki/tests/test_managers.py
jwhitlock: Oops missed that one. 👍

escattone:
Thanks @jwhitlock! I'll merge when the tests pass.

PR #4467, opened 2017-10-16
Bug 1407806 - Remove Vary: Cookie header

jwhitlock:
The ``BanMiddleware`` calls ``request.user.is_authenticated`` to check if a user is a banned user. However, this accesses the session, and the ``SessionMiddleware`` adds a ``Vary: Cookie`` header to all requests.  @escattone discovered and fixed this in PR #4461 by clearing ``request.session.accessed``.  This PR fixes the root cause, and confirms the ``/files/`` and ``@api`` methods both exclude a ``Vary: Cookie`` header.

### DIFF
--- a/kuma/attachments/tests/test_views.py
+++ b/kuma/attachments/tests/test_views.py
@@ -184,7 +184,7 @@ def test_legacy_redirect(client, file_attachment):
     assert response.status_code == 301
     assert 'Location' in response
     assert response['Location'] == file_attachment['attachment'].get_file_url()
-    assert response['Vary'] == 'Cookie'
+    assert 'Vary' not in response
 
 
 def test_edit_attachment_get(admin_client, root_doc):
--- a/kuma/attachments/tests/test_views.py
+++ b/kuma/attachments/tests/test_views.py
@@ -184,6 +184,7 @@ def test_legacy_redirect(client, file_attachment):
     assert response.status_code == 301
     assert 'Location' in response
     assert response['Location'] == file_attachment['attachment'].get_file_url()
+    assert response['Vary'] == 'Cookie'
 
 
 def test_edit_attachment_get(admin_client, root_doc):
--- a/kuma/attachments/views.py
+++ b/kuma/attachments/views.py
@@ -34,10 +34,6 @@ def raw_file(request, attachment_id, filename):
     """
     Serve up an attachment's file.
     """
-
-    # Stop SessionMiddleware from adding "Cookie" to the "Vary" header.
-    request.session.accessed = False
-
     qs = Attachment.objects.select_related('current_revision')
     attachment = get_object_or_404(qs, pk=attachment_id)
     if attachment.current_revision is None:
--- a/kuma/users/middleware.py
+++ b/kuma/users/middleware.py
@@ -12,7 +12,15 @@ class BanMiddleware(object):
     banned.
     """
     def process_request(self, request):
-        if hasattr(request, 'user') and request.user.is_authenticated():
+        # Checking request.user.is_authenticated() will access the session,
+        # and the SessionMiddleware will add Vary: Cookie. Avoid it by
+        # checking, then resetting request.session.accessed to the previous
+        # value (hopefully False).
+        old_session_accessed = request.session.accessed
+        is_auth = hasattr(request, 'user') and request.user.is_authenticated()
+        request.session.accessed = old_session_accessed
+
+        if is_auth:
             bans = UserBan.objects.filter(user=request.user,
                                           is_active=True)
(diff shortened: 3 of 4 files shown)


jwhitlock:
Manual testing shows this removes the ``Vary: Cookie`` from ``$sample`` pages as well, but I'll need to refactor the tests to prove it 🤷‍♂️

escattone:
👍 This looks good to me, thanks @jwhitlock. I do have a broader question unrelated to the intent of this PR though. HTTP caching is done per resource (per URL), so it seems to me that each resource should present a consistent set of caching headers (`Vary`, `Cache-Control`, etc.) and their values, but it seems that MDN does not (at least consistently) do this. For example, it seems that the `en-US/` URL (the landing page for the US) can sometimes respond with `Vary: Cookie` and sometimes not. That doesn't seem correct to me. The response to 'en-US/' either varies on the cookie, or it does not. It should be consistently set or not. In any case, the caching headers in general in MDN seem haphazardly applied.

jwhitlock:
I'm catching up on caching myself.

My process for looking into caching is:

1. [Setup docker for debugging](https://kuma.readthedocs.io/en/latest/development.html#customizing-the-docker-environment)
2. Add a breakpoint (``import pdb; pdb.set_trace()``) at a point before the session is accessed
3. Add a breakpoint for setting the session accessed boolean (``break /usr/local/lib/python2.7/dist-packages/django/contrib/sessions/backends/base.py:174``)

The session is accessed in the base template, to get the ``sociallogin_provider`` into the DOM. This was so we could display a Github or Persona login indicator. It's not as useful anymore.

https://github.com/mozilla/kuma/blob/409bf5b6ab5f80e5efad5b75ead8690977c689ef/jinja2/base.html#L86

I think you're right. If a view accesses ``requests.user.is_authenticated()``, the ``Vary: Cookie`` should be set.  Accessing ``request.user`` in the session loads the lazy-loaded object, and breaks the "set when accessed" feature. I'll see if I can get the homepage always setting the ``Vary`` header.

``Vary`` is the ugly old way of doing caching. The new way is the  ``Cache-Control`` header, and adjusting your application so you don't need it.  Google has a good [caching overview](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching). ``Vary`` is baked into Django, so it will take a lot of changes to get rid of it.  For example, we could make a completely anonymous base template and home page, and then load ``/user.js`` or similar, that adds per-user custom code, and then updates the page. The base page could be cached and served from edge servers, and you could go all the way back to the server for ``user.js``.  We could have ``Cache-Control: private, max-age: 300`` so that none of the intermediate cache servers caches it, a E-Tag that hashes the contents, and ``Vary: Cookie`` so that the browser re-requests when the session changes or logs out.  We'll need to look closely at each endpoint.

For CloudFront, we'll have to be careful around [caching with cookies](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Cookies.html) to avoid caching user-specific content.

jwhitlock:
Oops, wrong button. I need to adjust the PR, but I'm not finished with my blog post.  The adjustment is that a follow-on call to ``request.user`` should result in the ``Vary`` header being set. I may need to test a endpoint that doesn't access that item (not the homepage).

CloudFront will, by default, try to [cache a page even with set cookies](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Cookies.html), so we'll need to be careful or we'll see more issues like [bug 1402523](https://bugzilla.mozilla.org/show_bug.cgi?id=1402523).

CloudFront tries to standardize headers of cached pages, and documents the [current behavior](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#ResponseBehaviorCustomOrigin).  For ``Vary:`` , it says:

> If you configure your origin to include either ``Accept-Encoding`` or ``Cookie`` in the ``Vary`` header, CloudFront includes the values in the response to the viewer. 

It is unclear to me if this means that ``Accept-Encoding`` and ``Cookie`` headers will be passed to the viewer, or if ``Vary: Cookie`` will be passed to the viewer. Experimentation is needed.

jwhitlock:
I was accidentally right. There's not a good way to check a user ban in a middleware, so we shouldn't do it in a middleware. I filed [bug 1409423](https://bugzilla.mozilla.org/show_bug.cgi?id=1409423), which will require a different PR.

escattone:
Thanks for your comments @jwhitlock. I'll just add that I think we need to eventually take on the large task of carefully looking at each endpoint in terms of its caching headers, and implementing any changes.

PR #4469, opened 2017-10-18
bug 1409423: Avoid Vary: Cookie by removing ban middleware

jwhitlock:
This replaces #4467, doing it the "right" way.

The ``BanMiddleware`` looked for banned users and logged them out. This meant that the session was accessed in the middleware, and a ``Vary: Cookie`` header was added to all responses. Instead, banned user detection is split:

* A new authentication backend, ``KumaAuthBackend``, extends the default ``django-allauth`` backend to refuse to recognize the sessions of banned users, treating them as anonymous users.
* When a banned user attempts to login, they see the banned user page, which continues to tell them why they were banned, and who they can email to appeal the ban.

Changing the authentication backend required upgrading ``django-allauth``. User sessions include the authentication backend used to log them in, and our current version hard-codes this to their auth backend.  The new version allows you to set the auth backend to our extended class.

The new ``django-allauth`` includes a lot of the functionality of our customized ``GithubProvider``. It is possible we can switch to the shipped GitHub provider. It proved too challenging to remove it, and so that experiment will have to wait for a future PR.  I did remove our custom scope function, and we're now using the ``django-allauth`` method for requesting access to GitHub emails.

I've added tests for the ``Vary: Cookie`` header to ``@api`` and ``$sample`` endpoints. This PR removes them! We should be ready for CloudFront on the untrusted domain after this PR.


jwhitlock:
Force-push to fix test that relies on ``STATIC_URL`` setting.

jwhitlock:
Coverage claims that the ban path in the new auth backend is not being called. Maybe the ``pre_social_login`` hook is enough, and we can avoid the ``django_allauth`` update for a future time...

jwhitlock:
Nope, the lack of coverage just means that ``client.login`` is not the way to test the authentication endpoint. The custom ``KumaAuthBackend`` is needed to logout banned users with existing sessions.  I've updated tests to cover the new backend.

jwhitlock:
One more thing, to save the reviewer some heartache.

As I mentioned, the authentication backend is stored in the session. Use the browser's storage inspector to get your session key, and then you can look at the session in the Django shell:

```
$ ./manage.py shell_plus
>>> from pprint import pprint
>>> my_session = Session.objects.get(session_key='0e6tio4n0rv0c89epgk8hvbs4')
>>> pprint(my_session.get_decoded())
{u'_auth_user_backend': u'allauth.account.auth_backends.AuthenticationBackend',
 u'_auth_user_hash': u'7e3b4...',
 u'_auth_user_id': u'1'}
```

If your user is authenticated with the new ``KumaAuthBackend``, then the ``_auth_user_backend`` will be ``kuma.users.auth_backends.KumaAuthBackend``. 

The original backend is used to authenticate the user. This means that if you login with the ``KumaAuthBackend``, and then your user is banned, that backend will log you out. If you are logged in with allauth's ``AuthenticationBackend``, and then you are banned, you will not be logged out.  This can be pretty frustrating if you are trying to manually test the logout.

So, if you are manually testing:

1. Log out your test user
2. Log in again (you'll get the ``KumaAuthBackend``)
3. In another browser, log in as an admin, ban your test user
4. Load another page as the test user. You should be logged out
5. Attempt to log in. You should get the ban warning

kuma/wiki/tests/test_views_code.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def finalize(self):
+        from constance import config
+        for attr, value in reversed(self._original_values):
+            setattr(config, attr, value)
+        del self._original_values[:]
+
+
+@pytest.fixture
+def constance_config(db, settings):
```
escattone: ❤️  This fixture and its associated `ConstanceConfigWrapper` class are really beautifully done. I like this fixture so much better than being limited to using `override_config` as a context manager within the new-style tests.

This fixture will be generally useful (for tests within `kuma.attachments`, `kuma.landing`, `kuma.spam`, `kuma.wiki`, etc.) so could be moved now to the top-level `conftest.py` file, but if you'd rather wait until it's used in those places that's fine too.

PR #4472, opened 2017-10-18
Bug 1370594: Make sample fame changes permanent

stephaniehobson:
- Remove `sample_frame` waffle
- Integrate experimental frame styles with existing styles
- Change buttons back to buttons styles instead of links
- Make buttons match frame border & be encouraging blue colour

@jwhitlock can I edit `etc/sample_db.json` like this?

### DIFF
--- a/etc/sample_db.json
+++ b/etc/sample_db.json
@@ -394,10 +394,6 @@
         "name": "line_length",
         "note": "Full width code, restricted line length.",
         "created": "2017-08-28"
-      }, {
-        "name": "sample_frame",
-        "note": "Frame live sample output.",
-        "created": "2017-09-05"
       }
     ],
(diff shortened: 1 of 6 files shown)


etc/sample_db.json
```diff
@@ -394,10 +394,6 @@
         "name": "line_length",
         "note": "Full width code, restricted line length.",
         "created": "2017-08-28"
-      }, {
-        "name": "sample_frame",
-        "note": "Frame live sample output.",
-        "created": "2017-09-05"
```
jwhitlock: Yes, this is how you can remove a flag from ``sample_db.json``. You even got the commas right! JSON is picky.

This will be applied the next time I create a sample database.

PR #4473, opened 2017-10-19
Fix bug 1370594: Remove redesign announcement

stephaniehobson:


etc/sample_db.json
```diff
@@ -385,12 +385,6 @@
         "note": "Display a beta notice to beta users.",
         "created": "2017-06-22"
       }, {
-        "name": "redesign_live",
-        "note": "Display a notice for the 2017 redesign rollout",
-        "everyone": false,
-        "superusers": false,
-        "created": "2017-06-22"
-      }, {
```
jwhitlock: 👍 - this will take effect when we build the sample database next time

PR #4475, opened 2017-10-20
bug 1410559: prevent ISE when no file in cleaned data

escattone:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1410559


escattone:
Great change, thanks @jwhitlock! I made the change you requested and added two tests that would fail without it or something like it. Ready for your final review.

PR #4476, opened 2017-10-24
bug 1407806: add cache-control header to untrusted wiki endpoints

escattone:
This PR ensures that all of the remaining untrusted-domain endpoints appropriately set (or not) the `Cache-Control` header:

1. `attachments.raw_file` (e.g., https://mdn.mozillademos.org/files/12984/web-font-example.png)
    - The appropriate cache-control headers (public, 5-minute default max age that can be modified from settings) and handling (last-modified) have already been applied.
1. `attachments.mindtouch_file_redirect` (e.g., https://developer.mozilla.org/@api/deki/files/2926/=AudioTest_(1).ogg)
    - The response is always a permanent redirect (to the `attachments.raw_file` endpoint). There is no need for cache-control headers since the response will always be permanently cached.
1. `wiki.code_sample`(e.g., https://mdn.mozillademos.org/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color):
    - The response is dynamically-rendered HTML that depends on two things: the `wiki/code-sample.html` template, and the code sample extracted from the document.
    - We can add `public` and some `max-age` to the cache-control header, but I'm not sure what the best approach (`ETag` or `Last-Modified`), if any, is in terms of handling validation.
        - It seems to me we can't use the `Last-Modified` approach unless we cheat. We'd have to use the `max` of the two dates, one for the document and one for the template, but I don't think there's a fast and reliable way to get the last-modified date of a template. We could cheat and just use the last-modified date of the document, assuming that the template will rarely change, but we'd need a way to invalidate the cache in the rare event of a template change.
        - If we use the `ETag` approach, we'd have to select a way to generate its value. We could generate a SHA1 (or other) hash of the rendered HTML and use that, but that doesn't save much (if any) since we still must perform a complete render on validation requests and we're adding the hash computation as well.
    - This PR chooses, for now, to exclude the `ETag` and/or `Last-Modified` headers, and simply sets a max-age setting for code samples, using `Cache-Control: public, max-age=X` (where X defaults to 24 hours and can be modified from the settings).
1. `wiki.raw_code_sample_file` (e.g., `https://mdn.mozillademos.org/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color/files/<id>/<filename>`)
    - The response is a temporary redirect (to the `attachments.raw_file` endpoint), so this PR sets the cache-control header to `public` with a default `max-age` of 5 days (which can be modified from the settings).

I've submitted this PR without tests in order to start the discussion of this approach right away.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1241,7 +1241,17 @@ MAX_FILEPATH_LENGTH = 250
 ATTACHMENT_HOST = config('ATTACHMENT_HOST', default='mdn.mozillademos.org')
 ATTACHMENTS_CACHE_CONTROL_MAX_AGE = config(
     'ATTACHMENTS_CACHE_CONTROL_MAX_AGE',
-    default=300,
+    default=60 * 5,
+    cast=int
+)
+CODE_SAMPLES_CACHE_CONTROL_MAX_AGE = config(
+    'CODE_SAMPLES_CACHE_CONTROL_MAX_AGE',
+    default=60 * 60 * 24,
+    cast=int
+)
+CODE_SAMPLE_FILE_REDIRECT_CACHE_CONTROL_MAX_AGE = config(
+    'CODE_SAMPLE_FILE_REDIRECT_CACHE_CONTROL_MAX_AGE',
+    default=60 * 60 * 24 * 5,
     cast=int
 )
 
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1239,21 +1239,6 @@ MAX_FILENAME_LENGTH = 200
 MAX_FILEPATH_LENGTH = 250
 
 ATTACHMENT_HOST = config('ATTACHMENT_HOST', default='mdn.mozillademos.org')
-ATTACHMENTS_CACHE_CONTROL_MAX_AGE = config(
-    'ATTACHMENTS_CACHE_CONTROL_MAX_AGE',
-    default=60 * 5,
-    cast=int
-)
-CODE_SAMPLES_CACHE_CONTROL_MAX_AGE = config(
-    'CODE_SAMPLES_CACHE_CONTROL_MAX_AGE',
-    default=60 * 60 * 24,
-    cast=int
-)
-CODE_SAMPLE_FILE_REDIRECT_CACHE_CONTROL_MAX_AGE = config(
-    'CODE_SAMPLE_FILE_REDIRECT_CACHE_CONTROL_MAX_AGE',
-    default=60 * 60 * 24 * 5,
-    cast=int
-)
 
 # This should never be false for the production and stage deployments.
 ENABLE_RESTRICTIONS_BY_HOST = config(
--- a/kuma/wiki/views/code.py
+++ b/kuma/wiki/views/code.py
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 import re
+import hashlib
 
-from django.conf import settings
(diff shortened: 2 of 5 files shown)


kuma/settings/common.py
```diff
@@ -1241,7 +1241,17 @@ def pipeline_one_scss(slug, **kwargs):
...
+    cast=int
+)
+CODE_SAMPLES_CACHE_CONTROL_MAX_AGE = config(
+    'CODE_SAMPLES_CACHE_CONTROL_MAX_AGE',
+    default=60 * 60 * 24,
+    cast=int
+)
+CODE_SAMPLE_FILE_REDIRECT_CACHE_CONTROL_MAX_AGE = config(
+    'CODE_SAMPLE_FILE_REDIRECT_CACHE_CONTROL_MAX_AGE',
+    default=60 * 60 * 24 * 5,
```
jwhitlock: I know we already have a setting for ``ATTACHMENTS_CACHE_CONTROL_MAX_AGE``, but I'm having a hard time imagining why we'd have different values for these in different environments. Do you have a use case in mind? Otherwise, it seems simpler to have the values in the decorator.

kuma/wiki/views/code.py
```diff
@@ -18,6 +20,10 @@
 @allow_CORS_GET
 @xframe_options_exempt
 @process_document_path
+@cache_control(
+    public=True,
+    max_age=settings.CODE_SAMPLES_CACHE_CONTROL_MAX_AGE
+)
```
jwhitlock: This would work. However, I agree an ETag would be better, and would cache changes in the template as well.

I don't think I'm ready to set ``USE_TAGS`` yet, but I could be convinced. [From the docs](https://docs.djangoproject.com/en/1.11/ref/middleware/#django.middleware.common.CommonMiddleware), it looks like the processing of ETags moved from ``CommonMiddleware`` to the ``ConditionalGetMiddleware``, so we may want to wait until 1.11 to enable this across the site. Or maybe we're ready.

If we just want to use an ETag on this view, I think it would be easy to create a decorator based on Django 1.8's ``CommonMiddleware``:

https://github.com/django/django/blob/2d4bc5a60aa8a076689667c550ded96b87bc463e/django/middleware/common.py#L113-L127

I agree we do the same amount of work, however, so the ETag isn't strictly needed.  An alternate would be to hash the revision ID alone as the ETag, and add the "complete" ETag if and when we update the base template.

Or, just keep it the way you did it the first time.

kuma/settings/common.py
escattone: The only use case I had in mind is the ability to make K8s deployments with the new values (the K8s templates don't currently take these settings into account yet though) without building new Docker images, but perhaps it's better that these settings go through the PR review process. I'll put the values in the decorator.

escattone:
Thanks for the feedback @jwhitlock.
- I removed the three cache-control settings and just placed the values in the decorators.
- I also added `ETag` handling to the `wiki.code_sample` endpoint. I'm guessing that the small cost of caculating the `hashlib.md5` checksum is probably worth the reduced I/O load. It's worth a try. I nested the `get_etag` function (as I also did [here](https://github.com/mozilla/kuma/blob/17c4ca4056e9215d00cf32c641418d4931891a9d/kuma/attachments/views.py#L45)) since it avoids the duplication of work (or caching the work on the request) that would be necessary otherwise.
- I added one test and updated others to check the cache-control headers and their values.

jwhitlock:
Thanks @escattone!

PR #4480, opened 2017-10-26
Bug 1070001: Separate large file into small ones

stephaniehobson:
Separating the large content.scss into smaller files.
Re-ordered includes to be alphabetical.

Testing:
- I have not changed any of the code, please do not comment on the code.
- look for: 
  - stuff I have deleted from content.scss and not included somewhere else
  - stuff that doesn't work in its new location
- no need for cross-browser testing, it should work or be broken in all of them

You can edit this URL to search for pages that uses specific classes on MDN (example uses `callout-box` as search term):
https://developer.mozilla.org/en-US/search?locale=*&topic=none&css_classnames=callout-box

### DIFF
--- a/kuma/static/styles/components/content.scss
+++ b/kuma/static/styles/components/content.scss
@@ -2,16 +2,83 @@
 
 /*
 Additional formatting for the real page content; i.e. text in the <main> element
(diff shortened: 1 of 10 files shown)


jwhitlock:
Build is failing due to an issue with pulling images from Docker hub. The alternate build passed.

PR #4484, opened 2017-10-26
bug 1360294: Add user.is_github_url_public

jwhitlock:
Add the column ``User.is_github_url_public``. We'll need to deploy and run migrations before merging the rest of #4346.


escattone:
Which reminds me that I need to add that to the deployment documentation.

PR #4485, opened 2017-10-26
bug 1402708: add robots.txt endpoint to untrusted domains

escattone:
This PR adds a `/robots.txt` endpoint (which disallows everything) to the untrusted domains:
- `ATTACHMENT_HOST`(e.g., `mdn.mozillademos.org`)
- `ATTACHMENT_ORIGIN` (e.g., `mdn-demos-origin.moz.works`)

This is part of work involved in https://github.com/mozmeao/infra/issues/590.

### DIFF
--- a/kuma/landing/test_views.py
+++ b/kuma/landing/test_views.py
@@ -20,16 +20,22 @@ def test_promote_buttons(client, db):
 
 
 @pytest.mark.parametrize('allowed', [True, False])
-def test_robots_enabled(client, db, settings, allowed):
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


kuma/landing/test_views.py
```diff
@@ -20,16 +20,22 @@ def test_promote_buttons(client, db):
 
 
 @pytest.mark.parametrize('allowed', [True, False])
-def test_robots_enabled(client, db, settings, allowed):
+@pytest.mark.parametrize(
```
jwhitlock: Nice use of two-dimension parametrize

PR #4486, opened 2017-10-30
bug 1407806: increase caching max-age for file attachments

escattone:
This PR simply increases the `max-age` portion of the `Cache-Control` header for the file-attachment endpoints from 5 minutes to 15 minutes. The [cache-hit rate for the most popular file-attachment endpoints](https://console.aws.amazon.com/cloudfront/home?region=us-west-2#popular_urls:) is currently hovering roughly around 50%. The goal is to significantly increase that, and this should give us more data we can use to find the sweet spot.


metadave:
👍

PR #4488, opened 2017-10-31
bug 1406546: enable deploy of Kuma images to AWS

escattone:
This PR is the initial pass at the blueprint for deploying Docker images to K8s. For now, it focuses solely on pushing to stage (stage.mdn.moz.works), and borrows from the Jenkins code and bash scripts used in [bedrock](https://github.com/mozilla/bedrock). The PR https://github.com/mozmeao/infra/pull/629 is a prerequisite for successful operation.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,9 @@
+#!groovy
+
 @Library('github.com/mozmar/jenkins-pipeline@master')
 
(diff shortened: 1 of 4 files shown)


jwhitlock:
I like the idea of adding to the common library. Let's get it working for our use cases, and then extract the generic functionality.

PR #4490, opened 2017-10-31
Adds document macro to add the js interactive editor

schalkneethling:
Note: This has the caveat that we are still loading the CSS and JS from the interactive examples domain.

### DIFF
--- a/kuma/wiki/jinja2/wiki/document.html
+++ b/kuma/wiki/jinja2/wiki/document.html
@@ -1,7 +1,7 @@
 {% extends "wiki/base.html" %}
 {% block title %}{{ page_title(document.title + seo_parent_title) }}{% endblock %}
 
-{% from "wiki/includes/document_macros.html" import build_document_crumbs, get_document_quick_links, get_document_subnav, document_watch, contributor_links with context %}
+{% from "wiki/includes/document_macros.html" import build_document_crumbs, get_document_quick_links, get_document_subnav, document_watch, contributor_links, add_interactive_editor with context %}
 {% from "wiki/includes/buttons.html" import get_document_buttons with context %}
 {% from "wiki/includes/approvals.html" import get_approvals_html with context %}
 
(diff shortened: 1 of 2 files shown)


kuma/wiki/jinja2/wiki/document.html
```diff
@@ -238,6 +238,10 @@
 
               <!-- just the article content -->
               <article id="wikiArticle">
+                {% if doc_abs_url == '/en-US/docs/Experiment:InteractiveEditor/Test.Array.prototype.reduce()' %}
```
schalkneethling: Also, we need to create this page on production.

kuma/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -159,3 +159,41 @@
                   {% endif %}
 
 {%- endmacro %}
+
+{% macro add_interactive_editor() %}
+<link href="https://interactive-examples.mdn.mozilla.net/css/codemirror.css" rel="stylesheet">
```
schalkneethling: Hmmm, for some reason it is excluding the `https://`

schalkneethling: Oh, no. In this view it is there now. Ok, seems fine.

schalkneethling:
@stephaniehobson Ok, I created the testing page on production at:
https://developer.mozilla.org/en-US/docs/Experiment:InteractiveEditor/Test.Array.prototype.reduce()

At the moment there is then no interactive editor. Once this merged, it will be loaded at the top of the article content.

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -238,6 +238,10 @@
...
+                  {{ add_interactive_editor() }}
```
stephaniehobson: You need to import the macro on line 4 of document.html

schalkneethling: Ugh, had it in there and then had to revert formatting changes applied by prettier and neglected to add it back.

PR #4493, opened 2017-11-01
Tweak history drop down. Add icons for node, server, and deprecated

stephaniehobson:
Added icons for nodejs, server, and deprecated.

Tweaked history drop down to fix bugs on tablet and mobile.


schalkneethling:
@stephaniehobson This looks seriously great! My only comment is that the icons on the left feel like they are to tightly packed:

screen shot 2017-11-02 at 13 17 57

Even giving it a right margin of `.3rem;` makes a noticeable difference:

screen shot 2017-11-02 at 13 18 06

Happy to merge as is but, thought I would mention it in case you wish to add a tweak. 🦄

stephaniehobson:
Going to merge because I need to write some code that will conflict.

I'll include the space between icons in the next PR :)

stephaniehobson:
Going to make the padding a pixel value though. ems will grow with the font and there's not much space in that column. If the user bumps the font size up I want to preserve as much space as possible for the feature name.

PR #4494, opened 2017-11-02
added details about the changes in the .env

pavangudiwada:
Added a few details about what to change in the .env  file for easy understanding

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -54,9 +54,9 @@ Docker setup
 
         cd kuma
 
-#. Initialize and customize ``.env``. Linux users should set the ``UID``
-   parameter in ``.env``, to avoid issues when mixing ``docker-compose`` and
-   ``docker`` commands::
+#. Initialize and customize ``.env``. Linux users should set the ``UID`` parameter in ``.env``
+   (i.e remove the "#" before ``UDI``,example: change ``#UDI=1000`` to ``UDI=1000``) to avoid 
+   issues when mixing ``docker-compose`` and ``docker`` commands::
 
         cp .env-dist.dev .env
         vim .env  # Or your favorite editor
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -54,8 +54,9 @@ Docker setup
 
         cd kuma
 
-#. Initialize and customize ``.env``. Linux users should set the ``UID`` parameter in ``.env``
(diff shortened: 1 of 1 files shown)


pavangudiwada:
i thought that might request for a review :( but it tried to commit

docs/installation.rst
```diff
@@ -54,9 +54,9 @@ Docker setup
 
         cd kuma
 
-#. Initialize and customize ``.env``. Linux users should set the ``UID``
-   parameter in ``.env``, to avoid issues when mixing ``docker-compose`` and
-   ``docker`` commands::
+#. Initialize and customize ``.env``. Linux users should set the ``UID`` parameter in ``.env``
+   (i.e remove the "#" before ``UDI``,example: change ``#UDI=1000`` to ``UDI=1000``) to avoid
```
jwhitlock: ``UDI`` should be ``UID``.

docs/installation.rst
```diff
@@ -54,9 +54,9 @@ Docker setup
 
         cd kuma
 
-#. Initialize and customize ``.env``. Linux users should set the ``UID``
-   parameter in ``.env``, to avoid issues when mixing ``docker-compose`` and
-   ``docker`` commands::
+#. Initialize and customize ``.env``. Linux users should set the ``UID`` parameter in ``.env``
+   (i.e remove the "#" before ``UDI``,example: change ``#UDI=1000`` to ``UDI=1000``) to avoid 
+   issues when mixing ``docker-compose`` and ``docker`` commands::
```
jwhitlock: You may want to mention file permission issues as well.

safwanrahman:
The approach I always follow is
```export UID
docker-compose up
```

docs/installation.rst
pavangudiwada: I think those issues are because of the UID so i dont think there is a need to mention it and we can mention about the Troubleshooting page here in case there is a problem during instillation. We can also mention about the other troubles, not using sudo and other things in the Troubleshooting  page.

jwhitlock:
@safwanrahman ``make up`` has 7 characters, and ``export UID; docker-compose up`` has 28. You could be 400% more efficient by using ``.env``!

Also, I find it easier to have someone share their ``.env`` file (or even script creating it), then to work out the differences with environments, Linux distributions, etc., so I'd like to continue using this as the recommended method in the docs.

jwhitlock:
Thanks for updating the docs, @pavangudiwada!

PR #4495, opened 2017-11-02
Bug 1413289: Annotate links to missing documents with rel=nofollow

jwhitlock:
This addresses an SEO issue where the Google crawler is following links to known missing documents. Google [claims they do not penalize sites for 404s](https://support.google.com/webmasters/answer/35120#common_url_errors), but they do spend their "crawl budget" following these links rather than good links to other documents.  Adding a ``rel="nofollow"`` attribute to these links would suggest to the crawler that it should not follow them.

The top 404 is ``static/js/libs/ckeditor/build/ckeditor/``, which Google parses from the JavaScript ``window.CKEDITOR_BASEPATH`` included on every page. We could restrict this to wiki editing pages. This would be a different PR.

The next 404s are links from the Quick Links sidebar. For example, the page [Web/API/NavigatorOnLine/onLine](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine), there are sidebar links to the ``autocomplete`` and ``toggle`` events, but those pages do not exist.

There is an existing ``LinkAnnotationFilter`` that adds a class ``new``, so that links to missing documents are styled with a reddish-brown color.  This filter also adds a class ``external`` to external links.  This filter is only applied to the body text, not the sidebars.

The code changes are:

* Update the ``LinkAnnotationFilter`` so that links to missing wiki docs get a ``rel="nofollow"`` attribute.
* Update the section extraction tool so that Quick Links sidebar and the Zone subnav (like the sidebar on [Tools](https://developer.mozilla.org/en-US/docs/Tools)) also get the attributes ``rel="nofollow" class="new"`` on missing links.
* A big refactor of the Django tests that exercise this code. This might be best reviewed per commit.  **If this proves too difficult to review, I can split this PR.**
* Removing some code paths that are uncovered by tests because they are impossible.

For local manual testing, you may need to clear your cached HTML columns from the database.  In the Django shell:

```
$ ./manage.py shell_plus
>>> Document.objects.update(quick_links_html='', zone_subnav_local_html='', body_html='')
```

You can then force-refresh to see the effects on these pages in the sample database:
* http://localhost:8000/en-US/docs/Tools - The link color will not change, but most links will have the attributes ``rel="nofollow" class="new"``
* http://localhost:8000/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics - Sidebar links will have the new color and the new attributes.


kuma/wiki/tests/test_content.py
```diff
@@ -912,97 +836,120 @@ def test_bleach_filter_invalid_protocol(self):
...
+    else:
+        link_class = ''
+    html = normalize_html('<li><a %s href="%s"></li>' % (link_class, url))
+    actual_raw = parse(html).annotateLinks(base_url=AL_BASE_URL).serialize()
+    assert normalize_html(actual_raw) == html
+
+
+@pytest.mark.parametrize('has_class', ('hasClass', 'noClass'))
+@pytest.mark.parametrize('anchor', ('withAnchor', 'noAnchor'))
+@pytest.mark.parametrize('full_url', ('fullURL', 'pathOnly'))
```
escattone: Nice. This is when `parametrize` really shines, handling all of the combinations.

jwhitlock:
Thanks for the detailed review, @escattone! Copy-and-paste leads to mistakes every time...

PR #4496, opened 2017-11-02
Restrict code to 3 table types

stephaniehobson:
Reduce the amount of complied code we're outputting to just what we need for the 3 different kinds of tables we're publishing.

- create map with 3 table types and configuration options
- change all classes relying on number of browsers to rely on table type
- increase padding on mobile feature names
- add margin between icons

Testing notes:
- the macro branch has not been updated with these classes yet, you can test by swapping in `bc-table_web`, `bc-table_js`, and `bc-table_ext` on the web, javascript, and webextentions tables respectively using browser tools.

### DIFF
--- a/kuma/static/styles/components/compat-tables/_vars.scss
+++ b/kuma/static/styles/components/compat-tables/_vars.scss
@@ -30,4 +30,20 @@ $bc-border-width-thick: 2px;
 $bc-support: yes, no, partial, unknown;
 $bc-grid-spacing: 10px;
 
-$feature-name-width: 200px;
+$table-types: (
+	web: (
+		size: 13,
+		feature-name-width: 200px,
+		platform-border: (8)
+	),
+	js: (
+		size: 14,
+		feature-name-width: 200px,
+		platform-border: (8, 15)
+	),
+	ext: (
+		size: 5,
+		feature-name-width: 300px,
+		platform-border: (6)
+	)
+);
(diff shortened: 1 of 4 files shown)


schalkneethling:
At ~1000px each row heading, and the row just below it, does not have left hand borders. Also, the table as a whole, does not have right hand borders.

screen shot 2017-11-03 at 17 47 08

schalkneethling:
Something else I noticed at the 1000px breakpoint

screen shot 2017-11-03 at 17 51 16

schalkneethling:
On mobile/tablet, when you hover over the link "" the dialog is not shown. When you click the link, the dialog is shown but, the click event is triggered sending you to `docs/New_Compatibility_Tables_Beta`

hover-click-mobile

On desktop viewports, the dialog is shown on hover but, clicking the link has the same behavior.

schalkneethling:
Note: The behavior described here https://github.com/mozilla/kuma/pull/4496#issuecomment-341744270 does not happen in Chrome so, perhaps it is a bug with responsive design view in Nightly.

schalkneethling:
Same goes for https://github.com/mozilla/kuma/pull/4496#issuecomment-341743392

schalkneethling:
Ugh, same goes for https://github.com/mozilla/kuma/pull/4496#issuecomment-341746507 - All of the above works as expected in Chrome. I do see though that in Fx there are some JS errors thrown that I do not see in Chrome.

screen shot 2017-11-03 at 18 09 03

schalkneethling:
Something is definitely odd with devtools/responsive view in Nightly.

PR #4502, opened 2017-11-06
Implements fallback for IE9 and older

schalkneethling:
Implements fallback styles for IE9 and older. No javascript is served to these browser either, except for GA but, we can probably remove that as well, unless we create a tiny bundle of JS that collects stats.

The aim here is to make the pages as light as possible while ensuring pages and content are readable.

### DIFF
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -6,7 +6,9 @@
   <meta charset="utf-8">
   {%- block experiment_js %}{% endblock %}
   <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <!--[if lte IE 9]>
   <script>(function(d) { d.className = d.className.replace(/\bno-js/, ''); })(document.documentElement);</script>
+  <![endif]-->
   <title>{% block title %}{{ _('MDN Web Docs') }}{% endblock %}</title>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
@@ -17,19 +19,22 @@
   {% endif %}
 
   {% block site_css %}
-
     {% stylesheet 'mdn' %}
 
     {% for style in styles %}
       {% stylesheet style %}
     {% endfor %}
+
+    <!--[if lte IE 9]>
+      {% stylesheet 'fallback' %}
+    <![endif]-->
   {% endblock %}
 
   {%- if settings.MAINTENANCE_MODE %}
     {% stylesheet 'maintenance-mode-global' %}
   {%- endif %}
 
-  {#- If the stylesheet exists, include it. Otherwise, does nothing. #}
+  {#- If the stylesheet exists, include it. Otherwise, do nothing. #}
   {% stylesheet 'locale-%s' % LANG %}
 
   {% block site_analytics %}
@@ -58,11 +63,11 @@
   <link rel="apple-touch-icon-precomposed" sizes="72x72" href="{{ static('img/favicon72.png') }}">
   <!-- non-Retina iPhone, iPod Touch, and Android 2.1+ devices: -->
   <link rel="apple-touch-icon-precomposed" href="{{ static('img/favicon57.png') }}">
+
   <!-- basic favicon -->
   <link rel="shortcut icon" href="{{ static('img/favicon32.png') }}">
   <!--[if IE]>
-  <meta http-equiv="imagetoolbar" content="no">
-  {% javascript 'html5shiv' %}
+    {% javascript 'html5shiv' %}
   <![endif]-->
 
   {% javascript 'font-check' %}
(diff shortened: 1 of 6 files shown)


schalkneethling:
Until we make a final call, I am excluding GA from `lte IE9`

stephaniehobson:
It's going to take me a while to find time to review this.

At a high level I am generally in favour of doing this for the JS and not in favour of creating a separate set of styles sheets we have to maintain for IE9. IE9 should have a good user experience with the JS disabled but still seeing default CSS.

For that to work properly this line will need to be updated to reflect the fact that none of the site JS has executed: https://github.com/mozilla/kuma/blob/master/jinja2/base.html#L10 I think we should just wrap that scrip tag in the IE conditional comments.

jinja2/base.html
```diff
@@ -70,16 +84,14 @@
...
-  {% javascript 'html5shiv' %}
-  <![endif]-->
 
   {% javascript 'font-check' %}
 
   {% block extrahead %}{% endblock %}
 
-  {% include "includes/google_analytics.html" %}
+  <![if !lte IE 9]>
+    {% include "includes/google_analytics.html" %}
```
jwhitlock: I'm 👎 on removing GA from IE 9. It seems it would be useful to continue to get a browser count.

schalkneethling: Yup, that makes sense. I was thinking that as well. I am working on some tweaks on this PR, and will also create a small JS bundle to send some(the current) metrics across.

schalkneethling:
Ok so, I have tried various approaches to this but, int the end I still feel the benefit of the fallback outweigh the potential maintenance. Below are a couple of screenshots in IE9 when JS is not loaded, the line @stephaniehobson mentioned in https://github.com/mozilla/kuma/pull/4502#issuecomment-342234760 is not run, but all the standard CSS is being loaded.

ie9

As you can see, the layout is already starting to break apart here. There is also various functional elements shown which will not work without the JS. In my opinion this leads to a undesirable user experience. In IE8, things breaks apart even more.

ie8

In contrast the below are with the changes in this PR, but still loads GA. This puts content first, ensures there are no elements shown that depends on JS, ensures the content is easy to read, and loads snappy.

ie8-9

I therefore believe that the possible added maintenance of the IE specific CSS file outweighs the cons. Thoughts @stephaniehobson and @jwhitlock

schalkneethling:
Another point I failed to mention is that with JS not being loaded, the navigation also does not work so, to get to CSS, you need to click on "Technologies", and then the "CSS" page on the next page. With the fallback, the entire navigation is available.

stephaniehobson:
Still no time to give this a proper review but more high level comments:

- We know most of our users arrive directly on the page they want from a search and don't use the navigation very much. I think it's preferable to have a small global navigation and get them to the page content faster. The no-js global navigation still allows them to navigate to the CSS page, it trades one extra click in an usual situation for scrolling in a very common one.
- If there are interactive elements displaying with the javascript disabled we should hide them from all users with JS disabled

schalkneethling:
> Still no time to give this a proper review but more high level comments:

Good points @stephaniehobson, let me dig some more into this.

atopal:
Just saw this and wanted to add some data:
1.4% of MDN sessions are from IE, of those 96% are on v11 and 1% on v10. That means IE9 and older represents 0.042% of our overall audience.

stephaniehobson:
VMWare has forgotten all my machines 😫

If we have time you should rebase and demo this for me at the All-Hands next week.

jinja2/base.html
```diff
@@ -20,17 +26,30 @@
     <link rel="copyright" href="#copyright">
   {% endif %}
 
-  {% block site_css %}
+  <!--[if IE]>
+    <meta http-equiv="imagetoolbar" content="no">
```
stephaniehobson: Support for this meta tag seems to have ended in IE8, we can remove it :)

jinja2/base.html
```diff
@@ -253,20 +269,27 @@ <h2 class="footer-title">Mozilla</h2>
...
-    {{ providers_media_js() }}
-    <script type="text/javascript" src="{{ statici18n(request.LANGUAGE_CODE) }}"></script>
-    {% javascript 'main' %}
-    <script>
-      if(window.mdn && mdn.analytics) mdn.analytics.trackOutboundLinks();
-    </script>
-    {% for script in scripts %}
-      {% javascript script %}
-    {% endfor %}
+    <![if !lte IE 9]>
```
stephaniehobson: The conditional comments are inside the django block here and outside below. I think we should be consistent.

jinja2/base.html
```diff
@@ -253,20 +269,27 @@ <h2 class="footer-title">Mozilla</h2>
...
 
+        {% for script in scripts %}
+          {% javascript script %}
+        {% endfor %}
+    <![endif]>
   {% endblock %}
 
-  {% block js %}{% endblock %}
+  <![if !lte IE 9]>
+    {% block js %}{% endblock %}
```
stephaniehobson: The conditional comments are outside the django block here and inside above. I think we should be consistent.

kuma/landing/jinja2/landing/homepage.html
```diff
@@ -166,24 +167,30 @@ <h2><i aria-hidden="true" class="icon-smile-o"></i> {{ _('Help improve MDN') }}<
...
+<![if !lte IE 9]>
   {{ super() }}
   {% if settings.DEBUG %}
-      <script>
+    <script>
       jQuery('#home-search-form').on('submit', function() {
-    mdn.optimizely.push(['trackEvent', 'submit-homepage-search-form']);
-});
-</script>
+        mdn.optimizely.push(['trackEvent', 'submit-homepage-search-form']);
```
stephaniehobson: 👍 for proper formatting

kuma/landing/jinja2/landing/homepage.html
```diff
@@ -166,24 +167,30 @@ <h2><i aria-hidden="true" class="icon-smile-o"></i> {{ _('Help improve MDN') }}<
...
+      });
+    </script>
   {% endif %}
+
   {% if waffle.flag('search_suggestions') %}
     <script>
-        window.mdn.searchFilters = {{ command_search_filters|jsonencode }};
+      window.mdn.searchFilters = {{ command_search_filters|jsonencode }};
```
stephaniehobson: I think this block should disappear in a rebase.

stephaniehobson:
I still believe that the broken default styles provide a better experience than the IE9 fallback.

But I'm not the senior dev on this project anymore ;)

Can you please add a margin on the bottom of the "jump to" menu in the fallback CSS and then post some updated screen shots?

schalkneethling:
Thanks for the additional review @stephaniehobson - I reckon I am going to keep this one on ice for a while while I rethink how to approach this.

PR #4503, opened 2017-11-07
bug 1406546: refactor groovy code to handle kumascript as well as kuma

escattone:
This PR refactors [utils.groovy](https://github.com/mozilla/kuma/blob/master/Jenkinsfiles/utils.groovy) to handle kumascript as well.

### DIFF
--- a/Jenkinsfiles/utils.groovy
+++ b/Jenkinsfiles/utils.groovy
@@ -2,10 +2,43 @@
  * Define utility functions.
  */
 
-def checkout_github(repo, branch, relative_target_dir) {
+PROD_BRANCH_NAME = 'prod-push'
+STAGE_BRANCH_NAME = 'stage-push'
+KUMA_PIPELINE = 'mdn_multibranch_pipeline'
+KUMASCRIPT_PIPELINE= 'kumascript_multibranch_pipeline'
+
+def get_commit_tag() {
+    return env.GIT_COMMIT.take(7)
+}
+
+def get_target() {
+    if (env.BRANCH_NAME == PROD_BRANCH_NAME) {
+        return 'prod'
+    }
+    if (env.BRANCH_NAME == STAGE_BRANCH_NAME) {
+        return 'stage'
+    }
+    throw new Exception(
+        'Unable to determine the target from the branch name.'
+    )
+}
+
+def get_repo_name() {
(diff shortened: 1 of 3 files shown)


escattone:
I'd like to tell [the final `junit` step](https://github.com/mozilla/kuma/blob/7a7120f12c9349415877eacb040bfb393099d6d1/Jenkinsfile#L53) to ignore the case when there are no results, but I don't know how to do that. From my research, I think there's a setting on the JUnit plugin we need to enable, but I don't see that we have the JUnit plugin installed. Does anyone more familiar with Jenkins know how to do this?

Jenkinsfiles/utils.groovy
```diff
@@ -2,10 +2,38 @@
  * Define utility functions.
  */
 
-def checkout_github(repo, branch, relative_target_dir) {
+def get_commit_tag() {
+    return env.GIT_COMMIT.take(7)
```
metadave: whats does the value 7 correspond to?

Jenkinsfiles/utils.groovy
```diff
@@ -2,10 +2,38 @@
...
+    if (env.BRANCH_NAME.startsWith('stage-')) {
+        return 'stage'
+    }
+    throw new Exception(
+        'Unable to determine the target from the branch name.'
+    )
+}
+
+def get_repo_name() {
+    if (env.JOB_NAME.startsWith('mdn_multibranch')) {
```
metadave: it might be nice to see both branch names pulled out into constants

Jenkinsfiles/utils.groovy
escattone: It means "take the first 7 characters" of the git commit hash (so we match the tag we use for quay.io).

metadave: ah, ok, I'm not used to the Groovy APIs.

escattone:
Thanks for noticing that I should be using `github.com/mozmeao` instead of `github.com/mozmar` in https://github.com/mdn/kumascript/pull/395. I updated it here as well.

escattone:
It occurred to me that it may be wise to ensure that the Docker image can be pulled (that it exists on quay.io). If the image failed to be registered on quay.io, then the push will "succeed", but the pods will spin in ImagePullBackOff.

PR #4506, opened 2017-11-08
bug 1415358: replace base images

escattone:
Per https://github.com/mozmeao/infra/issues/641, this PR replaces the base images for MDN:

- Builds the `kuma_base` and `kumascript` images upon ~~`mozmeao/base:pythode-2.7-6.11`~~ `python:2.7-slim`
- Builds the `integration-tests` image upon ~~`mozmeao/base:python-2.7`~~ `python:2.7-slim`
- ~~Replaces the `kuma` and `kumascript` users with the `webdev` user provided by the new base image~~

### DIFF
--- a/docker/images/integration-tests/Dockerfile
+++ b/docker/images/integration-tests/Dockerfile
@@ -1,19 +1,23 @@
-FROM quay.io/mozmar/base
+FROM python:2.7-slim
 
 WORKDIR /app
 
-RUN apt-get update && \
+RUN set -ex && \
+    apt-get update && \
     apt-get install -y --no-install-recommends \
-                    build-essential python python-dev python-pip python-setuptools \
-                    libxml2-dev libxslt1.1 libxslt1-dev zlib1g-dev
+        mime-support \
+        build-essential \
+        libxml2-dev \
+        libxslt1.1 \
+        libxslt1-dev \
+        zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
 
 # Defaults
 ENV PYTEST_PROCESSES 5
 ENV PRIVACY "public restricted"
 ENV TESTS_PATH /app/tests
 ENV RESULTS_PATH /app/results
-ENV PYTHONDONTWRITEBYTECODE=1
-ENV PIP_DISABLE_PIP_VERSION_CHECK=1
 
 COPY ./requirements /app/requirements
 
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -1,54 +1,121 @@
-FROM quay.io/mozmar/base:latest
+FROM python:2.7-slim
 
-ENV PYTHONDONTWRITEBYTECODE=1 \
-    PIP_DISABLE_PIP_VERSION_CHECK=1
-
+# extra python env
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PYTHONUNBUFFERED=1
+ENV PIP_DISABLE_PIP_VERSION_CHECK=1
 # disable this when preparing for Django upgrade
 ENV PYTHONWARNINGS=ignore
+
+# ----------------------------------------------------------------------------
+# add node.js 6.12, copied from:
+#     https://github.com/nodejs/docker-node/blob/master/6.12/slim/Dockerfile
+# but with curl added to the list of packages installed.
+# ----------------------------------------------------------------------------
+RUN set -ex \
+  && for key in \
+    94AE36675C464D64BAFA68DD7434390BDBE9B9C5 \
+    FD3A5288F042B6850C66B31F09FE44734EB7990E \
+    71DCFD284A79C3B38668286BC97EC7A07EDE3FC1 \
+    DD8F2338BAE7501E3DD5AC78C273792F7D83545D \
+    C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
+    B9AE9905FFD7803F25714661B63B535A4C206CA9 \
+    56730D5401028683275BD23C23EFEFE93C4CFFFE \
+    77984A986EBC2AA786BC0F66B01FBB92821C587A \
+  ; do \
+    gpg --keyserver pgp.mit.edu --recv-keys "$key" || \
+    gpg --keyserver keyserver.pgp.com --recv-keys "$key" || \
+    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" ; \
+  done
+
+ENV NODE_VERSION 6.12.0
+
+RUN buildDeps='xz-utils' \
+    && ARCH= && dpkgArch="$(dpkg --print-architecture)" \
+    && case "${dpkgArch##*-}" in \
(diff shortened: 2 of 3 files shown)


jwhitlock:
The Docker build is failing - https://travis-ci.org/mozilla/kuma/jobs/298868891 . My guess is the path to the CSS compressor is no longer ``$TRAVIS_BUILD_DIR/node_modules/.bin/cleancss``, but I haven't looked closely.

docker/images/integration-tests/Dockerfile
```diff
@@ -1,19 +1,14 @@
-FROM quay.io/mozmar/base
```
pmac: The request in the infra issue was to use `FROM python:2.7-slim`. The ones in the mozmeao repo are just examples. We're moving away from all of our base images as we're not doing well enough keeping up with security updates as the official images are.

escattone: Ah, my misunderstanding, thanks @pmac, I'll update this PR accordingly.

pmac: I've added a clarifying comment with an example in the [original issue](https://github.com/mozmeao/infra/issues/641) if that helps. Thanks for jumping on this!

escattone:
The built-in Python `mimetypes` package is behaving unexpectedly using `python:2.7-slim`:
```
docker run -it python:2.7-slim python
Python 2.7.14 (default, Nov  4 2017, 14:53:30)
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import mimetypes
>>> mimetypes.guess_type('xx.json')
(None, None)
>>> mimetypes.guess_type('xx.xml')
('text/xml', None)
```
This is affecting tests that expected files with the `.json` extension to return `application/json` and `application/xml` for the `.xml` extension.

The `python:3-slim` image is better:
```
docker run -it python:3-slim python
Python 3.6.3 (default, Nov  4 2017, 14:30:48)
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mimetypes
>>> mimetypes.guess_type('xx.json')
('application/json', None)
>>> mimetypes.guess_type('xx.xml')
('text/xml', None)
>>>
```

jwhitlock:
On current images:

```
>>> import mimetypes
>>> mimetypes.guess_type('xx.json')
('application/json', None)
>>> mimetypes.guess_type('xx.xml')
('application/xml', None)
>>> mimetypes.knownfiles
['/etc/mime.types', '/etc/httpd/mime.types', '/etc/httpd/conf/mime.types', '/etc/apache/mime.types', '/etc/apache2/mime.types', '/usr/local/etc/httpd/conf/mime.types', '/usr/local/lib/netscape/mime.types', '/usr/local/etc/httpd/conf/mime.types', '/usr/local/etc/mime.types']
```

In python:2.7-slim:
```
>>> mimetypes.knownfiles
['/etc/mime.types', '/etc/httpd/mime.types', '/etc/httpd/conf/mime.types', '/etc/apache/mime.types', '/etc/apache2/mime.types', '/usr/local/etc/httpd/conf/mime.types', '/usr/local/lib/netscape/mime.types', '/usr/local/etc/httpd/conf/mime.types', '/usr/local/etc/mime.types']
```

So that appears to be a list of files it will look for MIME info, not the ones available.

I think the ``mime-support`` Debian package is needed.

escattone:
The location of node's global install changed from `/usr/bin` to `/usr/local/bin`, so within `docker/images/kuma_base/Dockerfile` I needed to update
```
PIPELINE_CLEANCSS_BINARY=/usr/local/bin/cleancss
PIPELINE_UGLIFYJS_BINARY=/usr/local/bin/uglifyjs
```
and although it probably wasn't necessary, I also added:
```
PIPELINE_SASS_BINARY=/usr/local/bin/node-sass
```

escattone:
Also, the `mime-support` suggestion did the trick, thanks @jwhitlock!

escattone:
FYI, I also manually tested the kumascript build, which worked fine, like so:
```
make build-kumascript KS_VERSION=latest
cd kumascript
make lint
make lint-macros
make test
make test-macros
```

docker/images/kumascript/Dockerfile
```diff
@@ -1,44 +1,107 @@
-FROM debian:jessie
+FROM python:2.7-slim
```
safwanrahman: kumascript also python? I think it can use `node:6.12-slim` docker image

docker/images/kuma_base/Dockerfile
```diff
@@ -1,54 +1,121 @@
-FROM quay.io/mozmar/base:latest
-
-ENV PYTHONDONTWRITEBYTECODE=1 \
-    PIP_DISABLE_PIP_VERSION_CHECK=1
+FROM python:2.7-slim
 
+# extra python env
+ENV PYTHONDONTWRITEBYTECODE=1
```
safwanrahman: It should be revoked in the kuma full image that goes to production. As we need bytecode in production.
you should add `PYTHONDONTWRITEBYTECODE=0` in `images/kuma/Dockerfile`

docker/images/kumascript/Dockerfile
```diff
@@ -1,44 +1,107 @@
...
+  && ln -s /opt/yarn/bin/yarn /usr/local/bin/yarnpkg \
+  && rm yarn-v$YARN_VERSION.tar.gz.asc yarn-v$YARN_VERSION.tar.gz
+# ----------------------------------------------------------------------------
 
-RUN useradd -m kumascript
+RUN set -ex && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+        gettext \
+        mime-support \
```
safwanrahman: `mime-support` is needed in kumascript?

docker/images/kumascript/Dockerfile
escattone: Yes, I think that's a better fit for the kumascript image, as it seems well-supported, thanks @safwanrahman. I'll change it in another PR.

docker/images/kuma_base/Dockerfile
escattone: Good catch, thanks again @safwanrahman. I'll change in another PR.

docker/images/kumascript/Dockerfile
escattone: I'm not certain it's needed. I added it just to be safe.

docker/images/kuma_base/Dockerfile
jwhitlock: I think @pmac had good reasons for ``PYTHONDONTWRITEBYTECODE=1``, but I can't find them.

pmac: Reasons were 1) it looks to be common in python images because 2) the .pyc files are useful when an app starts the 2nd time. In our containers this rarely happens since each container start is a fresh system.

jwhitlock: I suspect it could slow down Docker for Mac as well, writing (Linux only?) ``.pyc`` files to the slow shared filesystem.

PR #4511, opened 2017-11-09
Add gradient background to status colours

stephaniehobson:
- Add gradient background to colours to give them a pattern for colour
  blind users to tell them apart
- change colours to reference global variables


schalkneethling:
@stephaniehobson The X and / in the background helps a *lot* but, I wonder if we can look at bumping the contrast a bit more. Some of this still bleeds together a lot:

screen shot 2017-11-09 at 10 11 10

PR #4513, opened 2017-11-09
Bug 1414107: Type adjustments for asides

stephaniehobson:


kuma/static/styles/components/content.scss
```diff
@@ -119,14 +119,15 @@ We cannot put these styles in the base/elements/ files because they don't apply
...
+    article[id=wikiArticle] > h3,
+    article[id=wikiArticle] > div:not([class]) > h3.highlight-span {
         @include bidi-value(padding, 0 4px 0 $grid-spacing, 0 $grid-spacing 0 4px);
     }
 
     @media #{$mq-tablet-and-up} {
-        h3,
-        h3 .highlight-span {
+        article[id=wikiArticle] > h3,
+        article[id=wikiArticle] > div:not([class]) > h3.highlight-span {
```
schalkneethling: Nit: Should these be `article[id='wikiArticle']` for consistency?

schalkneethling:
@stephaniehobson Is this change intentional? I know there was a previous change that adding left padding of 40px. This however seems to behave more like a 40px margin, making the headings offset from its content.

kuma/static/styles/components/content.scss
stephaniehobson: Yes, and I'll have to change it in a few other places as well since I was copying this from somewhere else.

stephaniehobson:
I can't specificity today apparently.

stephaniehobson:
Desired behaviour:

screen shot 2017-11-13 at 17 29 42

stephaniehobson:
Okay updated.

schalkneethling:
Before:
         
screen shot 2017-11-14 at 14 50 29

After:

screen shot 2017-11-14 at 14 51 11

PR #4517, opened 2017-11-13
Remove Optimizely flag, feature, and related code

schalkneethling:
This removes the Optimizely flag as well as related code and config.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ nosetests.xml
 coverage.xml
 /test_results/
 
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -25,6 +25,9 @@ nosetests.xml
 coverage.xml
 /test_results/
 
+# Atom
+.ropeproject
```
jwhitlock: Sneaky. 👀

PR #4519, opened 2017-11-13
Removing requirejs from the mix to see how HTTP2 handles the requests

schalkneethling:
Seeing some strange TTFB so, dropping requirejs and playing to the strengths of HTTP2

### DIFF
--- a/kuma/wiki/jinja2/wiki/includes/document_macros.html
+++ b/kuma/wiki/jinja2/wiki/includes/document_macros.html
@@ -204,6 +204,7 @@ console.log(result);
     </div>
 </section>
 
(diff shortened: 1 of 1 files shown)


stephaniehobson:
We put #4521 into production. If I interpreted the intent of this PR wrong please comment and we can go forward with it. Otherwise, please close.

schalkneethling:
> We put #4521 into production. If I interpreted the intent of this PR wrong please comment and we can go forward with it. Otherwise, please close.

@stephaniehobson No, that is perfect, thanks! I was thinking of `defer` after opening the pull request and then it simply slipped my mind to update the PR :-/

PR #4520, opened 2017-11-13
Optimize prefetch_related usage in revisions dashboard

jgmize:
WIP for https://github.com/mozmeao/infra/issues/598

### DIFF
--- a/kuma/dashboards/views.py
+++ b/kuma/dashboards/views.py
@@ -4,6 +4,7 @@ import json
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import permission_required
 from django.contrib.auth.models import Group
+from django.db.models import Prefetch
 from django.http import HttpResponse
 from django.shortcuts import render
 from django.utils import timezone
@@ -95,12 +96,12 @@ def revisions(request):
         revisions = revisions.filter(**query_kwargs).exclude(**exclude_kwargs)
 
     # prefetch_related needs to come after all filters have been applied to qs
-    revisions = revisions.prefetch_related('creator__bans',
-                                           'document__deleted',
-                                           'document__locale',
-                                           'document__slug',
-                                           'akismet_submissions')
-
+    revisions = (revisions.prefetch_related('akismet_submissions',
+                                            'creator__bans')
+                          .prefetch_related(
+                              Prefetch('document',
+                                       queryset=Document.objects.only(
+                                           'deleted', 'locale', 'slug'))))
     revisions = paginate(request, revisions, per_page=PAGE_SIZE)
 
     context = {
--- a/kuma/dashboards/views.py
+++ b/kuma/dashboards/views.py
@@ -27,11 +27,7 @@ def revisions(request):
     filter_form = RevisionDashboardForm(request.GET)
     page = request.GET.get('page', 1)
 
-    revisions = (Revision.objects.prefetch_related('creator__bans',
-                                                   'document',
-                                                   'akismet_submissions')
-                                 .order_by('-id')
-                                 .defer('content'))
+    revisions = Revision.objects.order_by('-id').defer('content')
 
     query_kwargs = False
     exclude_kwargs = False
@@ -97,6 +93,13 @@ def revisions(request):
 
     if query_kwargs or exclude_kwargs:
         revisions = revisions.filter(**query_kwargs).exclude(**exclude_kwargs)
+
+    # prefetch_related needs to come after all filters have been applied to qs
+    revisions = revisions.prefetch_related('creator__bans',
+                                           'document__deleted',
+                                           'document__locale',
+                                           'document__slug',
+                                           'akismet_submissions')
 
     revisions = paginate(request, revisions, per_page=PAGE_SIZE)
 
--- a/kuma/dashboards/views.py
+++ b/kuma/dashboards/views.py
@@ -96,12 +96,16 @@ def revisions(request):
         revisions = revisions.filter(**query_kwargs).exclude(**exclude_kwargs)
 
     # prefetch_related needs to come after all filters have been applied to qs
-    revisions = (revisions.prefetch_related('akismet_submissions',
-                                            'creator__bans')
-                          .prefetch_related(
-                              Prefetch('document',
-                                       queryset=Document.objects.only(
-                                           'deleted', 'locale', 'slug'))))
+    revisions = revisions.prefetch_related('creator__bans').prefetch_related(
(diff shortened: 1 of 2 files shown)


safwanrahman:
I am getting following error
```python
ValueError: 'document__deleted' does not resolve to an item that supports prefetching - this is an invalid parameter to prefetch_related().
```

kuma/dashboards/views.py
```diff
@@ -98,6 +94,13 @@ def revisions(request):
     if query_kwargs or exclude_kwargs:
         revisions = revisions.filter(**query_kwargs).exclude(**exclude_kwargs)
 
+    # prefetch_related needs to come after all filters have been applied to qs
+    revisions = revisions.prefetch_related('creator__bans',
```
safwanrahman: I think your intention is to get the `locale`, `slug` and `deleted` only from document.
I believe this should be something like

```python
revisions = (revisions.prefetch_related('creator__bans', 'akismet_submissions').prefetch_related(
                           Prefetch('document',
                                    queryset=Document.objects.only('deleted', 'locale', 'slug')
                           )
 ))

Comment (jgmize) [reply kuma/dashboards/views.py]: Thanks @safwanrahman, I appreciate your taking a look at my untested WIP PR so quickly :)

**Comment by jgmize** [comment]
@safwanrahman I've updated the PR based on your suggestion; would you be willing to test it again?

**Comment by safwanrahman** [comment]
@jgmize While going through the code, I have seen that the `revisionip` were calling. investigating more it seems that there can be some optimization. **do not load revisionip if its not asked**.
I have make the correction and commited in my branch. can you fetch and cherry pick the commit?
https://github.com/mozilla/kuma/commit/cde03e8e6a0c254663ab5f34e0bddf55411b5c75

File: kuma/dashboards/views.py [hunk 2]
```diff
@@ -98,6 +95,13 @@ def revisions(request):
     if query_kwargs or exclude_kwargs:
         revisions = revisions.filter(**query_kwargs).exclude(**exclude_kwargs)
 
+    # prefetch_related needs to come after all filters have been applied to qs
+    revisions = (revisions.prefetch_related('akismet_submissions',
```
safwanrahman: `akismet_submissions` is only can be accessed by  the user having the proper permission. most of the user does not need it.
so only prefetch `akismet_submissions` only if `show_spam_submission` is `True

Something like
```python
show_spam_submission = (request.user.is_authenticated() and
    request.user.has_perm('wiki.add_revisionakismetsubmission'))

if show_spam_submission:
    revisions = revisions.prefetch_related('akismet_submissions')

jgmize: Good point, will update. Thanks again!

jgmize:
@safwanrahman thanks again for your suggestion; I've added https://github.com/mozilla/kuma/pull/4520/commits/991888809a555590b33a9d94377266dbd5317140

jgmize:
@safwanrahman I've also cherry-picked https://github.com/mozilla/kuma/commit/cde03e8e6a0c254663ab5f34e0bddf55411b5c75, thanks again.

jwhitlock:
@jgmize @safwanrahman thanks, this looks like it will reduce follow-on data requests, and appears to be a minimal, safe change. I'll take a look with some sample searches in the morning.

jwhitlock:
Just noticed the ``not ready`` tag. Let me know when this is ready for review @jgmize.

jwhitlock:
It looks good, no errors tested locally with the sample DB.

I tested with three user profiles:
* Anonymous - not logged in, but not private browsing
* User - logged in user, without staff permissions
* Admin - superuser

I tested four queries:
* Unfiltered ([localhost](http://localhost:8000/en-US/dashboards/revisions), [prod](https://developer.mozilla.org/en-US/dashboards/revisions)) - This gives the most recent revisions, and is used by admins.
* English-Only ([localhost](http://localhost:8000/en-US/dashboards/revisions?locale=en-US&user=&topic=&start_date=&end_date=&preceding_period=&authors=0&page=1), [prod](https://developer.mozilla.org/en-US/dashboards/revisions?locale=en-US&user=&topic=&start_date=&end_date=&preceding_period=&authors=0&page=1)) - This is used by localization leaders. There's no localization leader for English, but there's enough content to exercise the dashboard.
* Sheppy ([localhost](http://localhost:8000/en-US/dashboards/revisions?locale=&user=Sheppy&topic=&start_date=&end_date=&preceding_period=&authors=0&page=1), [prod](https://developer.mozilla.org/en-US/dashboards/revisions?locale=&user=Sheppy&topic=&start_date=&end_date=&preceding_period=&authors=0&page=1)) - This is the target of the "View all activity" link on [profile pages](https://developer.mozilla.org/en-US/profiles/Sheppy).
* Unknown Authors ([localhost](http://localhost:8000/en-US/dashboards/revisions?locale=&user=&topic=&start_date=&end_date=&preceding_period=&authors=2&page=1), [prod](https://developer.mozilla.org/en-US/dashboards/revisions?locale=&user=&topic=&start_date=&end_date=&preceding_period=&authors=2&page=1)) - Used by content managers to review just the edits of unfamiliar users for spam and other undesirable edits.

I refreshed the page a few times, because uncached waffle flags cause 40+ SQL queries.  The number given is the least queries and the fastest time (although time can be misleading for local development). The dashboard has a reduced number of queries except for admin users.

|       Test | Anon Before | Anon After | User Before | User After | Admin Before | Admin After |
| ---------- | ----------- | ---------- | ----------- | ---------- | ----------- | ---------- |
| Unfiltered | 110 Qs, 65 ms | 59 Qs, 42 ms | 117 Qs, 75 ms | 66 Qs, 44 ms | 139 Qs, 85 ms | 139 Qs, 82 ms |
|      en-US | 108 Qs, 68 ms | 57 Qs, 43 ms | 115 Qs, 81 ms | 64 Qs, 48 ms | 135 Qs, 80 ms | 135 Qs, 82 ms |
|     Sheppy | 126 Qs, 78 ms | 75 Qs, 48 ms | 133 Qs, 83 ms | 82 Qs, 52 ms | 132 Qs, 77 ms | 132 Qs, 84 ms |
|    Unknown | 111 Qs, 71 ms | 140 Qs, 84 ms | 118 Qs, 71 ms | 67 Qs, 50 ms | 140 Qs, 89 ms | 140 Qs, 84 ms |

It improves on the anonymous case, which is the one associated with downtime events, so I'm inclined to ship it.  More work is needed to fix the prefetches for the admin case, but that could be done in a future PR. First, I need to file a bug for the work.

PR #4522, opened 2017-11-14
bug 1409523: add gzip compression + BREACH-attack mitigation

escattone:
This PR adds GZIP compression handling to all requests that can accept GZIP encoding, and all endpoints with a response greater than 200 bytes that are not already content-encoded. It also adds BREACH attack mitigation via `django-debreach` and rate-limiting (the search page only). It does not add Brotli compression handling, which can be in another PR.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -430,7 +430,8 @@ _CONTEXT_PROCESSORS = (
     'django.core.context_processors.media',
     'django.core.context_processors.static',
     'django.core.context_processors.request',
-    'django.core.context_processors.csrf',
+    # todo: re-enable with Django 1.11
+    # 'django.core.context_processors.csrf',
     'django.contrib.messages.context_processors.messages',
 
     'kuma.core.context_processors.global_settings',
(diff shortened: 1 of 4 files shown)


jwhitlock:
We'll need to just merge and ship the ``django-debreach`` change to get the locales build to work.

The ``kuma/wiki/tests/test_views_code.py::test_code_sample`` test failure looks real - the response is compressed, and the test is expecting uncompressed content.

The change looks clean and minimal. I'll look closer when the tests are passing.

escattone:
Split-out the addition of the django-debreach package to https://github.com/mozilla/kuma/pull/4524. That was merged and this PR has been rebased. The failing test `kuma/wiki/tests/test_views_code.py::test_code_sample` was fixed.

kuma/settings/common.py
```diff
@@ -438,6 +438,7 @@ def enable_candidate_languages():
     'kuma.core.context_processors.next_url',
 
     'constance.context_processors.config',
+    'debreach.context_processors.csrf',
```
escattone: From reading the source code, I think this actually replaces `django.core.context_processors.csrf`, but I decided to follow instructions and just tack it at the end.

PR #4527, opened 2017-11-15
Updates filenames for non-iframe interactive editor

schalkneethling:
Change required after https://github.com/mdn/interactive-examples/issues/330

### DIFF
--- a/kuma/wiki/jinja2/wiki/includes/document_macros.html
+++ b/kuma/wiki/jinja2/wiki/includes/document_macros.html
@@ -171,7 +171,7 @@
 {%- endmacro %}
 
 {% macro add_interactive_editor() %}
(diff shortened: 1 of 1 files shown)


kuma/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -204,7 +204,6 @@
     </div>
 </section>
 
-<script src="https://interactive-examples.mdn.mozilla.net/js/lib/codemirror.js" defer></script>
-<script src="https://interactive-examples.mdn.mozilla.net/js/mode/javascript/javascript.js" defer></script>
+<script src="https://interactive-examples.mdn.mozilla.net/js/codemirror-5-31-0.js" defer></script>
```
jwhitlock: 👍to combining assets as well

PR #4528, opened 2017-11-15
Bug 1040595: Add padding to top of English

stephaniehobson:
- add padding to the top of the English display matching approximate
  height of the toolbar on the translation side
- move `.text-content` to different element so headings identifying
  English and French translations match style

### DIFF
--- a/kuma/static/styles/components/wiki/edit/translate/translate-document.scss
+++ b/kuma/static/styles/components/wiki/edit/translate/translate-document.scss
@@ -25,6 +25,21 @@
         .approved-title {
             display: inline-block;
             float: left;
+            margin-top: 0;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


SphinxKnight:
I've tested locally and the result is indeed better than the current situation. However, this doesn't really align depending on the resolution while loading the page.

I'll try to work around this in the next days. But the current state is definitely r+ to me :)

schalkneethling:
Tested it locally in Chrome and Fx, and it seems like the top padding of `223px` works consistently well for all the various viewport widths on desktop:

top-padding-translate

SphinxKnight:
@schalkneethling yes since, the toolbar is only sized while the page is loaded. Thus, resizing the page "live" doesn't alter the gap.

PR #4533, opened 2017-11-15
Fix bug 1418010: Fix lazy loading fonts

stephaniehobson:
- Copy fix from #4482 into the waffled "skinny" styles files
- Switch to mixin for all header-font-family declarations

The page title should display immediately in the fallback font. Then Zilla should either be swapped in or not in 2 seconds. On the next page load the heading should display in Zilla immediately.

The font-loading script relies on session storage to do this so a hard refresh can confuse it. Empty your session storage before testing.

Fix bug 1418010

### DIFF
--- a/kuma/static/styles/includes-skinny/_mixins.scss
+++ b/kuma/static/styles/includes-skinny/_mixins.scss
@@ -377,20 +377,14 @@ These are not dynamic but serve as mixins
 
 @mixin heading-1() {
     @include set-font-size($h1-font-size);
-    font-family: $heading-font-family;
+    @include set-heading-font-family();
     line-height: 1;
-    #{$selector-heading-font-fallback} {
-        font-family: $heading-font-family-fallback;
-    }
 }
 
 @mixin heading-2() {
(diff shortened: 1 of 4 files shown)


kuma/static/styles/includes-skinny/_mixins.scss
```diff
@@ -377,19 +377,17 @@ These are not dynamic but serve as mixins
...
-    }
 }
 
 @mixin heading-2() {
-    @include set-font-size($h2-font-size);
-    font-family: $heading-font-family;
-    line-height: 1;
-    #{$selector-heading-font-fallback} {
-        font-family: $heading-font-family-fallback;
+    @include set-font-size($mobile-h2-font-size);
```
schalkneethling: Tests are failing because `$mobile-h2-font-size: 32px;` is missing from `kuma/static/styles/includes-skinny/_vars.scss`

stephaniehobson:
Updated.

schalkneethling:
Testing using a 3G connection and it works perfectly 👍 

font-loading

PR #4534, opened 2017-11-16
Fix Bug 1136741, Removes search_suggestions waffle and related code

schalkneethling:
This removes the `search_suggestions` waffle flag and related code.


schalkneethling:
@jwhitlock I am doing an initial push to see if any tests fail as I cannot get them running locally.

schalkneethling:
Well, all the tests passed 👍

kuma/landing/views.py
```diff
@@ -32,14 +31,11 @@ def home(request):
...
         community_stats = {'contributors': 5453, 'locales': 36}
 
-    groups = FilterGroup.objects.all()
-    serializer = GroupWithFiltersSerializer(groups, many=True)
     default_filters = Filter.objects.default_filters()
 
     context = {
         'updates': updates,
         'stats': community_stats,
-        'command_search_filters': serializer.data,
```
jwhitlock: ❤️ removed the unneeded context and builder code.

kuma/landing/views.py
```diff
@@ -8,8 +8,7 @@
 from kuma.core.sections import SECTION_USAGE
 from kuma.core.cache import memcache
 from kuma.feeder.models import Bundle
-from kuma.search.models import Filter, FilterGroup
-from kuma.search.serializers import GroupWithFiltersSerializer
```
jwhitlock: The ``GroupWithFiltersSerializer`` should also be deleted from ``kuma/search/serializers.py``

safwanrahman:
Can you add the bug number in pull request?

schalkneethling:
> Can you add the bug number in pull request?

@safwanrahman Currently no bug number for these. I should probably log bugs for it though. Will do so for the follow on PRs

schalkneethling:
@jwhitlock updated - Thanks for the review.

safwanrahman:
@schalkneethling Can you please file a bug and add the bug id to the PR? so it will be future reference that why it has been removed.

jwhitlock:
You can use [bug 1136741](https://bugzilla.mozilla.org/show_bug.cgi?id=1136741). The convention is to add it to the front of the commit message, like

```
bug 1136741: Remove search-suggestions flag
```

kuma/landing/views.py
stephaniehobson: I see `FilterWithGroupSerializer` in test_serializers.py as well.

stephaniehobson:
@jwhitlock The original PR shows some changes to the models as well, not sure they're still present though: https://github.com/mozilla/kuma/commit/87a6c400e7075f79f635941d40d57cd35820b8f3

stephaniehobson:
Very through and working flawlessly locally. R+ 🔍 

Leaving in case you want to look at test_serializers.py as part of this PR but hit the merge button tomorrow if not.

PR #4536, opened 2017-11-16
bug 1406546: potential fix for kuma test reliability issue in jenkins

escattone:
This PR addresses https://github.com/mozmeao/infra/issues/650 and https://github.com/mozmeao/infra/issues/602, as well as https://github.com/mozmeao/infra/issues/595.

* use `docker-compose down` to ensure a clean context both before and after the Kuma tests (I'm hoping this addresses the symptoms of https://github.com/mozmeao/infra/issues/602)
* notify `#mdndev` in IRC on failure

I also added IRC notification on failure to the `default` case, but I would be happy to remove that if we'd rather not clutter `#mdndev` with failure notifications for random branches.

### DIFF
--- a/Jenkinsfiles/utils.groovy
+++ b/Jenkinsfiles/utils.groovy
@@ -55,32 +55,36 @@ def notify_irc(Map args) {
     sh command
 }
 
-def make(cmd, cmd_display) {
-    def target = get_target()
-    def tag = get_commit_tag()
-    def nick = "mdn${target}push"
-    def repo_upper = get_repo_name().toUpperCase()
+def sh_with_notify(cmd, display, notify_on_success=false) {
+    def nick = "mdn-${env.BRANCH_NAME}"
     try {
-        /*
-         * Run the actual make command within the proper environment.
-         */
-        sh """
-            . regions/portland/${target}.sh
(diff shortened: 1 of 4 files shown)


Jenkinsfiles/utils.groovy
```diff
@@ -55,34 +55,38 @@ def notify_irc(Map args) {
     sh command
 }
 
-def make(cmd, cmd_display) {
-    def target = get_target()
-    def tag = get_commit_tag()
-    def nick = "mdn${target}push"
-    def repo_upper = get_repo_name().toUpperCase()
+def sh_with_notify(cmd, display, notify_on_success=false) {
+    def nick = "mdn-${env.BRANCH_NAME}"
```
metadave: does the nick change with every branch on purpose?

escattone: Yes, I thought it'd be a quick indicator of where the message is coming from (although the branch name is also in the message). It's helpful for example when we're doing a push of both kuma and kumascript at the same time, since they get different nicks.

I'm totally open to a different approach though!

jwhitlock: It does make my IRC a bit wordy:

screenshot 2017-11-16 16 10 14

Let's keep this for a week or so and a few pushes. I might have a stronger opinion then.

PR #4537, opened 2017-11-17
Fix Bug 1418301, updates preload urls

schalkneethling:
Updates the `preload` urls to the new endpoints. I removed the `as="document"` as it seems this is not supported in Chrome release either so, we are not getting any benefit from this.


schalkneethling:
@stephaniehobson After reading the spec again, I would say it is safe to simply add these to the page without a conditional. The first time it is encountered the browser will fetch the resource and cache it.

As this early fetch it is not render blocking, this simply means that once the user does get to a page with the "editor", the JS will be instantly loaded, parsed and executed from browser cache. Thoughts?

PR #4539, opened 2017-11-17
Fix Bug 1418306, removes code for non-iframe interactive example test

schalkneethling:
Removes code for non-iframe interactive example experiment.


stephaniehobson:
R+ ⛺️

PR #4543, opened 2017-11-20
Fix Bug 1136741, changes registration_disabled to a switch

schalkneethling:
Changes the `registration_disabled` waffle flag to a switch.


schalkneethling:

PR #4544, opened 2017-11-20
Fix Bug 1136741, remove wiki_spam_exempted flag

schalkneethling:
Removed `wiki_spam_exempted` waffle flag

### DIFF
--- a/kuma/wiki/migrations/0011_create_spam_data.py
+++ b/kuma/wiki/migrations/0011_create_spam_data.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 from django.conf import settings
 from django.db import migrations
 from kuma.spam.constants import SPAM_CHECKS_FLAG
-from kuma.wiki.constants import SPAM_EXEMPTED_FLAG
 
 
 def create_trusted_writers(apps, schema_editor):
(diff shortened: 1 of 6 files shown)


kuma/wiki/migrations/0011_create_spam_data.py
```diff
@@ -4,24 +4,14 @@
...
-    # then add that user group to the list of people exempted from spam checks
     Flag = apps.get_model('waffle', 'Flag')
-    exempted_flag, created = Flag.objects.get_or_create(
-        name=SPAM_EXEMPTED_FLAG,
-        defaults={
-            'note': 'Used to decide which users are exempted from spam checks.'
-        },
-    )
-    exempted_flag.groups.add(group)
-
```
jwhitlock: I shouldn't have merged this... We try not to change past migrations except in extreme circumstances.

schalkneethling: @jwhitlock Ah ok, will remember for future changes. Just to double check. Did you add this code back and merge it to master?

jwhitlock: Yes, I added it in a second commit to #4553.

I hope to squash migrations in December, and we'll have less of them to cause us problems 😄

PR #4545, opened 2017-11-20
Fix Bug 1401291, fixes badges.js and component.js

schalkneethling:
Fixes eslint warnings in badges.js and component.js

### DIFF
--- a/kuma/static/js/components.js
+++ b/kuma/static/js/components.js
@@ -157,7 +157,7 @@
                                 firstLink.focus();
                             }
                             catch(e){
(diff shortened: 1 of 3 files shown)


kuma/static/js/components.js
```diff
@@ -157,7 +157,7 @@
                                 firstLink.focus();
                             }
                             catch(e){
-                                console.log('Menu focus exception! ', e);
+                                console.error('Menu focus exception! ', e);
```
stephaniehobson: 👍

PR #4547, opened 2017-11-20
bug 984149: Enable Telugu (te) as candidate locale

jwhitlock:
[bug 984149](https://bugzilla.mozilla.org/show_bug.cgi?id=984149) is an older bug, and a lot of the UI is already translated. It should be a quick turn as a candidate language.

Some additional files included from following instructions on https://kuma.readthedocs.io/en/latest/localization.html#add-a-new-locale-to-pontoon


jwhitlock:
Rebase on master, and add KumaScript and Locale submodule updates to get community string updates.

Also , it's Telugu, not Telegu like I mistyped. Code was correct, commit messages and PR title wasn't (fixed).

PR #4548, opened 2017-11-20
Fix bug 1419201: Breadcrumbs back in header for non-beta users

stephaniehobson:


stephaniehobson:
Hm. I'm not sure why `test_document_no_breadcrumbs` exists. I believe we do want breadcrumbs in that situation.

It needs to be updated or removed because all new documents will not have breadcrumbs unless they have a left column.

jwhitlock:
``test_document_no_breadcrumbs`` tests two things (which goes against the rules of good tests, but test refactoring is something else). 

The breadcrumbs don't show up for "root" documents, like https://developer.mozilla.org/en-US/docs/Mozilla - there's nothing to display. This is the same after moving them to the header, that test passes.

Currently, breadcrumbs are in the left sidebar, which is only displayed if there is a zone sub navigation, a quick links toolbar, and/or a review requested. The ``d2`` document is a child document but doesn't have anything that would add a left sidebar, so it also doesn't get breadcrumbs. Here's a child page w/ no left sidebar and thus no breadcrumbs - https://developer.mozilla.org/en-US/docs/user:jwhitlock/hidden

With the breadcrumbs in the header, ``d2`` should now get breadcrumbs as well. The functionality changed, so the test is failing.

If you want pages with no sidebar to have breadcrumbs, then I think ``test_document_no_breadcrumbs`` and ``test_document_has_breadcrumbs`` can be merged, since the second test does extra work to get a sidebar in place.

@stephaniehobson should pages with no left sidebar get breadcrumbs?

stephaniehobson:
Closing in favour of #4568

PR #4551, opened 2017-11-21
Fix Bug 1136741, change wiki_spam_training flag to switch

schalkneethling:
Changes the `wiki_spam_training` flag to a switch

### DIFF
--- a/kuma/wiki/tests/test_forms.py
+++ b/kuma/wiki/tests/test_forms.py
@@ -8,7 +8,7 @@ from django.test import RequestFactory
 import pytest
 import requests_mock
 from constance.test import override_config
-from waffle.models import Flag
+from waffle.models import Flag, Switch
 
 from kuma.core.urlresolvers import reverse
 from kuma.spam.constants import (CHECK_URL, SPAM_ADMIN_FLAG,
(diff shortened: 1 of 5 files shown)


kuma/wiki/tests/test_forms.py
```diff
@@ -537,8 +537,8 @@ def test_akismet_error(self, mock_requests):
     @pytest.mark.spam
     @requests_mock.mock()
     def test_akismet_spam_training(self, mock_requests):
-        flag, created = Flag.objects.get_or_create(name=SPAM_TRAINING_FLAG)
-        flag.users.add(self.testuser)
+        switch, created = Switch.objects.get_or_create(name=SPAM_TRAINING_SWITCH)
+        switch.users.add(self.testuser)
```
schalkneethling: @jwhitlock Not sure how to handle this as the `Switch` object does not have a user property.

schalkneethling: So, from the code I see that "Switches are active, or inactive, globally." so, perhaps I should just set active to `True` and remove line 541?

jwhitlock: This code is being overly cautious, and can be replaced with:

```
Switch.objects.create(name=SPAM_TRAINING_SWITCH, active=True)
```

Same for ``test_akismet_error_training`` below.

schalkneethling:

jwhitlock:
It looks like the rebase wasn't clean - making some small changes.

PR #4553, opened 2017-11-21
Fix Bug 1136741, change foundation_callout to setting

schalkneethling:
Changes `foundation_callout` to a Django setting

### DIFF
--- a/kuma/landing/jinja2/landing/homepage.html
+++ b/kuma/landing/jinja2/landing/homepage.html
@@ -58,7 +58,7 @@
 <div class="home-callouts">
   <h2 class="offscreen">{{ _('Featured') }}</h2>
   <div class="column-container center">
-    {% if waffle.switch('foundation_callout') %}
+    {%- if settings.FOUNDATION_CALLOUT %}
       <div class="column-callout callout-foundation">
         <a href="https://donate.mozilla.org/?utm_source=developer.mozilla.org&utm_medium=tiles&utm_campaign=EOY2016&utm_content=pixelheart">
           <span><strong>Mozilla</strong>{{ _('Proudly non-profit.') }}<br>{{ _('Donate before Dec 31.') }}<br><button>{{ _('Donate Now') }}</button></span>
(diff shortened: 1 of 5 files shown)


stephaniehobson:
@jwhitlock Are there performance benefits to making this a setting?

I'm thinking this is something it would be nice to toggle off Jan 1 without having to do a code push when there aren't many people around...

escattone:
@stephaniehobson Technically, there would be a performance improvement, but so small as to be unnoticeable (only the first request to a freshly-started Django process will hit the DB for the value of the waffle switch, while subsequent requests will just use the cached value, which is cached in local memory since).

kuma/landing/jinja2/landing/homepage.html
```diff
@@ -58,7 +58,7 @@ <h1 class="highlight highlight-spanned"><span class="highlight-span">{{ _('Resou
 <div class="home-callouts">
   <h2 class="offscreen">{{ _('Featured') }}</h2>
   <div class="column-container center">
-    {% if waffle.switch('foundation_callout') %}
+    {%- if settings.FOUNDATION_CALLOUT %}
```
escattone: It might be nice to use:
```python
FOUNDATION_CALLOUT = config('FOUNDATION_CALLOUT', False, cast=bool)
```
instead, so it can be set from the environment if desired. For local development via Docker, one could then change the value via the `.env` file, and for production/stage pushes we could change the value without creating a new Docker image for Kuma (we'd have to add some minimal code to our K8s files as well, but we can do that later if we decide we want to do that).

jwhitlock: 👍, except in ``kuma/settings/common.py``, not in the template.

escattone: Sorry, thanks @jwhitlock, yes I attached this to the wrong spot. I had meant to attach this comment to `kuma/settings/common.py`.

jwhitlock:
I was surprised that the default cache is the local memory cache. I think this means that waffle switches and flags may require server restarts to take effect (server 2 won't see that server 1 changed the waffle setting). We may want to investigate using either Memcache or Redis instead of local memory in production. And investigate using just one caching server instead of both Memcache and Redis.

escattone:
@jwhitlock Oh wow, good catch, I suspect you're right.

schalkneethling:
Thanks for the review @jwhitlock and @escattone - updated.

jwhitlock:
Hmm my attempt to use GitHub to resolve the merge conflict had a weird result. Trying again...

jwhitlock:
I had to rebase the original commit to fix the merge conflict, and then I added a commit to take care of an issue I saw in PR #4544 after merging.

PR #4559, opened 2017-11-22
Bug 1414107: Remove empty Quick_Links section

stephaniehobson:
Added a function to remove a section according to it's ID, existing function will remove the contents of a section but leave the wrapper.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -323,7 +323,8 @@ class Document(NotificationsMixin, models.Model):
         sections_to_hide = ('Quick_Links', 'Subnav')
         doc = parse_content(html)
         for sid in sections_to_hide:
-            doc = doc.replaceSection(sid, '<!-- -->')
+            doc = doc.replaceSection(sid, '')
+            doc = doc.removeSection(sid)
         doc.injectSectionIDs()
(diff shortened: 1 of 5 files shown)


stephaniehobson:
I appear to have broken this somehow. It's not working locally for me right now.

kuma/wiki/models.py
```diff
@@ -323,7 +323,8 @@ def get_body_html(self, *args, **kwargs):
         sections_to_hide = ('Quick_Links', 'Subnav')
         doc = parse_content(html)
         for sid in sections_to_hide:
-            doc = doc.replaceSection(sid, '<!-- -->')
+            doc = doc.replaceSection(sid, '')
+            doc = doc.removeSection(sid)
```
jwhitlock: The tests pass for me without the second line ``doc = doc.removeSection(sid)``. I suspect it isn't doing anything.

jwhitlock: After discussing this with @stephaniehobson, the issue is that KumaScript injects a ``<section id="Quick_Links>``, not an ``<h3 id="Quick_Links">`` like the test does. Adding a commit that fixes the test fixture to be more accurate and demonstrate the problem that the PR is trying to solve.

PR #4560, opened 2017-11-22
Fix Bug 1401291, ensure mozilla.cookiehelper and tag-it passes linting

schalkneethling:
Fixes linting error in `mozilla.cookiehelper` and `tag-it`

### DIFF
--- a/kuma/static/js/libs/mozilla.cookiehelper.js
+++ b/kuma/static/js/libs/mozilla.cookiehelper.js
@@ -41,15 +41,15 @@ Mozilla.Cookies = {
         var sExpires = '';
         if (vEnd) {
             switch (vEnd.constructor) {
-            case Number:
-                sExpires = vEnd === Infinity ? '; expires=Fri, 31 Dec 9999 23:59:59 GMT' : '; max-age=' + vEnd;
-                break;
-            case String:
-                sExpires = '; expires=' + vEnd;
-                break;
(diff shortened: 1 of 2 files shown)


kuma/static/js/libs/mozilla.cookiehelper.js
```diff
@@ -41,15 +41,15 @@ Mozilla.Cookies = {
...
-            case Number:
-                sExpires = vEnd === Infinity ? '; expires=Fri, 31 Dec 9999 23:59:59 GMT' : '; max-age=' + vEnd;
-                break;
-            case String:
-                sExpires = '; expires=' + vEnd;
-                break;
-            case Date:
-                sExpires = '; expires=' + vEnd.toUTCString();
-                break;
+                case Number:
```
jwhitlock: This should be submitted as a PR against https://github.com/mozilla/trafficcop as well.

schalkneethling: Good point, will do so.

schalkneethling:
> What do you think of excluding the /libs/ directory from linting instead of editing these files?

Yeah, after realizing that these are shared libs, I was wondering whether we should exclude them. To be honest, we use the same JS rules as bedrock so, it is odd that the `cookiehelper` slipped in here.

Let me close this PR and instead open another PR that excludes lib.

PR #4561, opened 2017-11-22
Fix Bug 1136741, remove wiki_force_immediate_rendering switch

schalkneethling:
Removes the `wiki_force_immediate_rendering` switch

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -7,7 +7,6 @@ from functools import wraps
 from uuid import uuid4
 
 import newrelic.agent
-import waffle
 from constance import config
 from django.apps import apps
 from django.conf import settings
@@ -542,8 +541,7 @@ class Document(NotificationsMixin, models.Model):
         Document.objects.filter(pk=self.pk).update(render_scheduled_at=now)
         self.render_scheduled_at = now
 
-        if (waffle.switch_is_active('wiki_force_immediate_rendering') or
-                not self.defer_rendering):
+        if (not self.defer_rendering):
             # Attempt an immediate rendering.
             self.render(cache_control, base_url)
         else:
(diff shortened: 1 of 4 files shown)


kuma/wiki/models.py
```diff
@@ -541,8 +541,7 @@ def schedule_rendering(self, cache_control=None, base_url=None):
         Document.objects.filter(pk=self.pk).update(render_scheduled_at=now)
         self.render_scheduled_at = now
 
-        if (waffle.switch_is_active('wiki_force_immediate_rendering') or
-                not self.defer_rendering):
+        if (not self.defer_rendering):
```
jwhitlock: ``wiki_force_immediate_rendering`` is not active in production. This change turns it on permanently. Only the ``else:`` branch should remain after this change.  Which may be difficult - I'm not sure why ``from . import tasks`` was a deferred import.

schalkneethling: I was wondering about that when I made the change so I decided to err on the side of caution.

schalkneethling:
@jwhitlock Ok, changed the code so only that which was in the `else` clause remains. I left the `import` as deferred. If this is not the intent, let me know and I will move it to the top with the other imports. Thanks!

schalkneethling:
@jwhitlock Seems like the latest round of changes broke more tests than it passes :-/

jwhitlock:
Yes, this one will be a hard one. The tests are relying on the switch behavior. I can take this one from here.

jwhitlock:
Tests pass after rebase. I didn't expect that. 🤔 Maybe things are OK now, or maybe this PR will be the source of flaky tests in the future. Forward! 🚀 

Adding a commit to split up the the deferred rendering test and re-arrange the ``if`` statement.

PR #4564, opened 2017-11-23
Fix Bug 1401291, ensures main and moz-jquery-plugins pass linting

schalkneethling:
Ensures `main` and `moz-jquery-plugins` pass linting. Also ignores all files in `libs` folder.

### DIFF
--- a/kuma/static/js/main.js
+++ b/kuma/static/js/main.js
@@ -1,21 +1,20 @@
 (function(win, doc, $) {
     'use strict';
 
-    /*
-        Track clientside errors
-    */
+    /**
+     * Track clientside errors
+     */
     mdn.analytics.trackClientErrors();
 
-    /*
-        Feature detection
-    */
-
+    /**
+     * Feature detection
+     */
     win.mdn.features.localStorage = (function() {
         var uid = new Date;
         var result;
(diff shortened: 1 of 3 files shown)


kuma/static/js/main.js
```diff
@@ -76,8 +75,15 @@
                     return;
                 }
 
-                if(e) e.preventDefault();
-                if(timeout) clearTimeout(timeout);
+                if(e) {
+                    e.preventDefault();
+                }
+
+
```
stephaniehobson: Extra line?

kuma/static/js/main.js
```diff
@@ -15,7 +15,7 @@
         var result;
         try {
             localStorage.setItem(uid, uid);
-            result = localStorage.getItem(uid) == uid;
+            result = localStorage.getItem(uid) === uid;
```
stephaniehobson: This test now reports false instead of true so I think `==` was the original intent here.  I guess we need to either re-write it to conform or add an exception for the linter :/

schalkneethling:
Thanks for the review @stephaniehobson - updated.

stephaniehobson:
R+ 🏪

PR #4565, opened 2017-11-23
Fix Bug 1401291, ensure promote.js and users.js pass linting

schalkneethling:
Ensure `promote.js` and `users.js` pass linting

### DIFF
--- a/kuma/static/js/promote.js
+++ b/kuma/static/js/promote.js
@@ -1,8 +1,9 @@
-/*
-    Preserving this function so that legacy users don't get errors on their site.
-*/
-function PromoteMDN() {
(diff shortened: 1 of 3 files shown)


kuma/static/js/promote.js
```diff
@@ -1,8 +1,9 @@
-/*
-    Preserving this function so that legacy users don't get errors on their site.
-*/
-function PromoteMDN() {
-    "use strict";
+/**
+ * Preserving this function so that legacy users don't get errors on their site.
+ */
+function PromoteMDN() { // eslint-disable-line no-unused-vars
```
stephaniehobson: 👍

PR #4567, opened 2017-11-23
Fix bug 1415969: Add heading section styles to editor

stephaniehobson:


schalkneethling:
Code looks good but, not sure if this is the intent, but the font size for `h2` elements inside the editor section is much larger than on the left hand side. Perhaps a base font size issue with `em`?

screen shot 2017-11-27 at 12 59 30

I do see that the `h2` font size here[1] is explicitly set different to the `h2` in the general copy so, perhaps this is intentional.

stephaniehobson:
Hm. @schalkneethling as you surmised your screenshot is definitely not the intended behaviour :/ 

Should be more like this:
screen shot 2017-11-30 at 14 33 15

The editor CSS is generated a bit differently than the other CSS on the site so while you can refresh to test regular wiki pages, it's best to give the web server a restart when working with editor CSS. `docker-compose restart web`.

If that was not the problem please comment with the URL of your test page :)

schalkneethling:
@stephaniehobson Still seeing the same behavior. For example at a URL like docs/Web/CSS/border-radius$translate?tolocale=fr

When I inspect the two headings, the heading outside the WYSIWYG editor has a set size of `1.5rem`, whilst the same heading inside the editor has a set size of `2.5rem`. This seems to be caused by:

https://github.com/mozilla/kuma/pull/4567/files#diff-d28f04e6c7fe5fdd258624947bf1b989R24

stephaniehobson:
`2.5rem` is the desired size for the headings, so something is off with your headings on the English side. There is also something funny going on because there's a block border behind both.

I'm not able to replicate.

Can you please post a screen shot of what you see on master?

schalkneethling:
git fetch mozilla 1415969-headings-editor:1415969-headings-editor
git checkout 1415969-headings-editor

docker exec -i kuma_web_1 make build-static
docker-compose restart web

Navigate to en-US/docs/Web/CSS/box-shadow
Click on Add a translation
Choose French

screen shot 2017-12-05 at 11 39 35

git checkout master

docker exec -i kuma_web_1 make build-static
docker-compose restart web

Navigate to en-US/docs/Web/CSS/box-shadow
Click on Add a translation
Choose French

screen shot 2017-12-05 at 11 52 12

schalkneethling:
Ah ok, if I rebase against master then all is good. Merging.

PR #4569, opened 2017-11-24
Fix bug 1184796: Optimize bibi mixin for multiple declarations

stephaniehobson:
**Current behaviour:**
```
.test {
    @include bidi-value(float, left, right);
    @include bidi-style(margin-right, ($grid-spacing / 2), margin-left, 0)
}
```
Outputs:
```
.test {
    float: left;
    margin-right: 10px;
}
html[dir='rtl'] .test {
    float: right;
}
html[dir='rtl'] .test {
    margin-right: 0;
    margin-left: 10px;
}
```

**Proposed behaviour:**
```
.test {
    @include bidi((
        (float, left, right),
        (margin-right, ($grid-spacing / 2), margin-left, 0)
    ));
}
```
Outputs:
```
.test {
    float: left;
    margin-right: 10px;
}
html[dir='rtl'] .test {
    float: right;
    margin-right: 0;
    margin-left: 10px;
}
```

This saves about 0.5KB in mdn.css after gzipping.

Questions:
- Is the new syntax confusing? Do we make contributions harder by changing?
- Is 0.5KB enough savings to keep working on this?

### DIFF
--- a/kuma/static/styles/components/wiki/promos.scss
+++ b/kuma/static/styles/components/wiki/promos.scss
@@ -30,8 +30,10 @@
         }
 
         &:after {
-            @include bidi-style(margin-left, 6px, margin-right, 6px);
-            @include bidi-style(content, '\f105', content, '\f104');
+            @include bidi((
+                (margin-left, 6px, margin-right, 6px),
+                (content, '\f105', content, '\f104'),
+            ));
             @include set-font-size(30px);
             font-family: FontAwesome;
             padding: 0 15px;
@@ -41,19 +43,21 @@
 
 #wiki-promo-sauce-labs-a {
     @include clearfix();
-    @include bidi-style(background-position, bottom 20px right 20px, background-position, bottom 20px left 20px);
-    @include bidi-style(padding, 20px 350px 20px 25px, padding, 20px 25px 20px 350px);
+    @include bidi((
+        (background-position, bottom 20px right 20px, background-position, bottom 20px left 20px),
+        (padding, 20px 350px 20px 25px, padding, 20px 25px 20px 350px),
+    ));
     background-image: url($path-to-embedded-images + 'promos/sauce-labs-wiki-a.png');
     min-height: 106px;
 
     .wiki-promo-copy,
     .wiki-promo-cta {
-        @include bidi-style(float, left, float, right);
+        @include bidi(((float, left, float, right),));
         box-sizing: border-box;
     }
 
     .wiki-promo-copy {
-        @include bidi-style(padding, 0 20px 0 0, padding, 0 0 0 20px);
+        @include bidi(((padding, 0 20px 0 0, padding, 0 0 0 20px),));
         width: 50%;
 
         p {
@@ -69,15 +73,19 @@
 }
 
 #wiki-promo-sauce-labs-b {
-    @include bidi-style(background-position, bottom right 20px, background-position, bottom left 20px);
-    @include bidi-style(padding, 20px 250px 20px 25px, padding, 20px 25px 20px 250px);
+    @include bidi((
+        (background-position, bottom right 20px, background-position, bottom left 20px),
+        (padding, 20px 250px 20px 25px, padding, 20px 25px 20px 250px),
+    ));
     background-image: url($path-to-embedded-images + 'promos/sauce-labs-wiki-b.png');
     min-height: 139px;
 }
 
 #wiki-promo-sauce-labs-c {
-    @include bidi-style(background-position, bottom right 20px, background-position, bottom left 20px);
-    @include bidi-style(padding, 0 230px 0 0, padding, 0 0 0 230px);
+    @include bidi((
+        (background-position, bottom right 20px, background-position, bottom left 20px),
+        (padding, 0 230px 0 0, padding, 0 0 0 230px),
+    ));
     background-color: transparent;
     background-image: url($path-to-embedded-images + 'promos/sauce-labs-wiki-c.png');
     margin-top: 20px;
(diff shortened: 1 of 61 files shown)


schalkneethling:
> Is the new syntax confusing? Do we make contributions harder by changing?

Maybe slightly but, if there is good documentation then I would say it is fair enough. I really like the cleaner grouped CSS that is output.

> Is 0.5KB enough savings to keep working on this?

I would say so. And if a small optimization such as this can yield such an impact, then I bet there are a few of these around the code base which will add up.

kuma/static/styles/components/wiki/promos.scss
```diff
@@ -150,8 +160,10 @@
 // max 768px
 @media #{$mq-mobile-and-down} {
     #wiki-promo-sauce-labs-a {
-        @include bidi-style(background-position, bottom 20px center, background-position, bottom 20px center);
-        @include bidi-style(padding, 20px 10px 140px, padding, 20px 10px 140px);
+        @include bidi((
+            (background-position, bottom 20px center, background-position, bottom 20px center),
+            (padding, 20px 10px 140px, padding, 20px 10px 140px)
```
schalkneethling: Nit: If both the LTR and RTL style rules are the same, is there a specific reason we should specify it inside `bidi`?

kuma/static/styles/components/wiki/promos.scss
```diff
@@ -168,8 +180,10 @@
     }
 
     #wiki-promo-sauce-labs-b {
-        @include bidi-style(background-position, bottom center, background-position, bottom center);
-        @include bidi-style(padding, 20px 20px 160px, padding, 20px 20px 160px);
+        @include bidi((
+            (background-position, bottom center, background-position, bottom center),
+            (padding, 20px 20px 160px, padding, 20px 20px 160px)
```
schalkneethling: Nit: Here as well, same question as above.

kuma/static/styles/components/wiki/promos.scss
```diff
@@ -201,8 +215,10 @@
     }
 
     #wiki-promo-sauce-labs-c {
-        @include bidi-style(background-position, bottom center, background-position, bottom center);
-        @include bidi-style(padding, 0 0 175px, padding, 0 0 175px);
+        @include bidi((
+            (background-position, bottom center, background-position, bottom center),
+            (padding, 0 0 175px, padding, 0 0 175px)
```
schalkneethling: Nit: Also here. This is the third instance so far so, I am starting to think there is a reason why it is done this way?

kuma/static/styles/includes-skinny/_mixins.scss
```diff
@@ -332,6 +332,27 @@ Spacing
 Bidi / l10n
 ====================================================================== */
 
+@mixin bidi($list) {
```
schalkneethling: Would be great to add some documentation here with a sample of a use case and the code that is produced.

kuma/static/styles/components/wiki/promos.scss
stephaniehobson: Yes, this is typically done when over-riding a value declared in a bidi further up the cascade, I will verify this and add notes in all the places this occurs :)

stephaniehobson:
I replaced all occurrences of bidi-style and bidi-value with this mixin and removed those mixins. Now we do it the same way everywhere. I hope that will make it easier for new contributors.

I found a problem where single line declarations don't always work right if they're not properly formatted and proper formatting is not intuitive:
`@include bidi(((float, right, left),));`

All occurrences are correct now though so if someone copies and pastes they should be able to just change the values.

Ready for second santiy-check and review.

kuma/static/styles/includes/_mixins.scss
```diff
@@ -330,26 +332,108 @@ Spacing
...
+@mixin bidi($list) {
+    & {
+        @each $property, $ltr-value, $inverse-val-or-prop, $inverse-value in $list {
+            @if type-of($ltr-value) == 'null' or type-of($inverse-val-or-prop) == 'null' {
+                @warn $list; /* I don't know why this won't output as part of the error */
+                @error 'Error processing @mixin bidi(), see previous warning for stack trace';
+                /*
+                    Trouble shooting:
+                    - Check that you have the right number of nested (((parentheses)))
+                    - Check that there is a trailing comma
```
schalkneethling: @stephaniehobson Should there always be a trailing comma? Or only in these instances:

```
@include bidi(((float, left, right),));
```

kuma/static/styles/components/wiki/promos.scss
```diff
@@ -41,19 +43,21 @@
 
 #wiki-promo-sauce-labs-a {
     @include clearfix();
-    @include bidi-style(background-position, bottom 20px right 20px, background-position, bottom 20px left 20px);
-    @include bidi-style(padding, 20px 350px 20px 25px, padding, 20px 25px 20px 350px);
+     @include bidi((
```
schalkneethling: Nit: indentation fails stylelint

kuma/static/styles/includes/_mixins.scss
stephaniehobson: Yes, there should always be a trailing comma.

Technically, it works in some situations without it but we should be consistent.

stephaniehobson: In case you're curious: it seems to work without the comma if there are no spaces or parentheses in the declaration.

stephaniehobson:
Nit fixed, squashed, double checked for trailing commas, rebased. Ready to be merged when reviews are passed!

PR #4571, opened 2017-11-28
Fix Bug 1420288, only load GA if DNT disabled

schalkneethling:
We now only load GA if DNT is disabled. Also, I moved the analytics block to before any of the CSS or JS as suggested here: https://developers.google.com/analytics/devguides/collection/analyticsjs/#the_javascript_tracking_snippet

### DIFF
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -32,6 +32,11 @@
   {#- If the stylesheet exists, include it. Otherwise, does nothing. #}
   {% stylesheet 'locale-%s' % LANG %}
 
+  {% block site_analytics %}
+    {% javascript 'dnt-helper' %}
+    {% include "includes/google_analytics.html" %}
+  {% endblock %}
+
   {%- if not untrusted %}
     <!-- common social tags -->
     {% set social_logo = request.build_absolute_uri(static('img/opengraph-logo.png')) %}
@@ -63,8 +68,6 @@
   {% javascript 'font-check' %}
 
   {% block extrahead %}{% endblock %}
-
-  {% include "includes/google_analytics.html" %}
 </head>
 <body {% block body_attributes %}{% endblock %} class="{% block bodyclass %}{% endblock %}{% if settings.MAINTENANCE_MODE %} maintenance-mode{% endif %}"{% if request.user and request.user.is_authenticated() %} data-login-service="{{ request.session.sociallogin_provider }}"{% endif %}>
 
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -1,65 +1,67 @@
 {% if config.GOOGLE_ANALYTICS_ACCOUNT != '0' %}
 {% set GA_FILE = 'analytics_debug' if settings.DEBUG else 'analytics' %}
 <script type="text/javascript">
+    // only load GA if DNT is not enabled
+    if (Mozilla && !Mozilla.dntEnabled()) {
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+        })(window,document,'script','//www.google-analytics.com/{{ GA_FILE }}.js','ga');
 
-    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-    })(window,document,'script','//www.google-analytics.com/{{ GA_FILE }}.js','ga');
+        ga('create', '{{ config.GOOGLE_ANALYTICS_ACCOUNT }}', 'mozilla.org');
+        ga('set', 'anonymizeIp', true);
 
-    ga('create', '{{ config.GOOGLE_ANALYTICS_ACCOUNT }}', 'mozilla.org');
-    ga('set', 'anonymizeIp', true);
+        {% if request.user.is_authenticated() %}
+            // dimension1 == 'Signed-In"
+            ga('set', 'dimension1', 'Yes');
 
-    {% if request.user.is_authenticated() %}
-        // dimension1 == 'Signed-In"
-        ga('set', 'dimension1', 'Yes');
+            // dimension2 == "Beta Tester"
+            {% if request.user.is_beta_tester %}
+                ga('set', 'dimension2', 'Yes');
+            {% endif %}
 
-        // dimension2 == "Beta Tester"
-        {% if request.user.is_beta_tester %}
-            ga('set', 'dimension2', 'Yes');
+            {%- if request.user.is_staff %}
+                // dimension18 == "Staff"
+                ga('set', 'dimension18', 'Yes');
+            {% endif %}
         {% endif %}
 
-        {%- if request.user.is_staff %}
-            // dimension18 == "Staff"
-            ga('set', 'dimension18', 'Yes');
+        // dimension9 == "Section editing"
+        {% if waffle.flag('section_edit') %}
+            ga('set', 'dimension9', 'Enabled');
         {% endif %}
-    {% endif %}
 
-    // dimension9 == "Section editing"
-    {% if waffle.flag('section_edit') %}
-        ga('set', 'dimension9', 'Enabled');
-    {% endif %}
+        {%- if analytics_page_revision %}
+            // dimension12 == 'Page Revision'
+            ga('set', 'dimension12', '{{ analytics_page_revision }}');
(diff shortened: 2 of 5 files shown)


kuma/settings/common.py
```diff
@@ -994,6 +994,12 @@ def pipeline_one_scss(slug, **kwargs):
 
 
 PIPELINE_JS = {
+    'dnt-helper': {
+        'source_filenames': (
+            'js/libs/mozilla.dnthelper.js',
```
stephaniehobson: I can't belive we already had it and weren't using it. *face, palm*

jwhitlock: It was added with Traffic Cop earlier this year.

jinja2/base.html
```diff
@@ -16,6 +16,11 @@
     <link rel="copyright" href="#copyright">
   {% endif %}
 
+  {% block site_analytics %}
+    {% javascript 'dnt-helper' %}
+    {% include "includes/google_analytics.html" %}
```
jwhitlock: Is there an enforced order to scripts in the header? If the GA script block executes before the dnt-helper scripts, then GA won't load, no matter what the user's DNT setting is.

schalkneethling: So, if I understand things correctly, when the browser gets to the `dnt-helper` it will pause rendering, fetch the script, parse and execute. Only then will it proceed to the next piece of the DOM which would be the GA bits. There are of course optimizations that browsers do but, inside the head I feel confident it will follow the above path.

I tested this throttling the connection all the way down the GPRS and it consistently loads GA if DNT is disabled, and does not, if DNT is enabled.

kuma/settings/common.py
stephaniehobson: Okay, I feel less bad now.

jinja2/base.html
stephaniehobson: That's my understanding as well 👍

stephaniehobson:
task-completion.js does not check for the existence of ga before doing it's thing 😢  This causes errors when the script executes. Every other place we use it is either filtered through analytics.js or checks for `win.ga` before assigning values to it. Please update task-completion.js

stephaniehobson:
What do you think of just using `navigator.doNotTrack` and making it inline? 

With ideal conditions the script downloads and executes before the CSS finishes downloading so it's not blocking the CSSOM rendering but if the JS doesn't load and the CSS does an external script early in the critical rendering path could be a problem.

schalkneethling:
Thanks for the review @stephaniehobson - updated.

> What do you think of just using navigator.doNotTrack and making it inline?

Instead of the DNT helper? If so, I would say no, just so we get the most predictable behavior across browsers and implementations of DNT

> not blocking the CSSOM rendering but if the JS doesn't load and the CSS does an external script early in the critical rendering path could be a problem.

Well, once it has `dnthelper` the rest is async right, so should not be blocking.

stephaniehobson:
The `dnthelper` is render blocking. 

Only inline scripts in the head before CSS files are non-render blocking. The only reason it is not effecting performance in my tests is that it executes before the CSS files complete downloading. I am concerned about the performance of this outside of ideal test conditions.

Looking at `dnthelper` I see support for Windows NT and Firefox 31. If we're dropping support for IE9 and browsers of that era, how much of this is relevant to our browser support requirements?

kuma/static/js/task-completion.js
```diff
@@ -9,13 +9,12 @@
         // asked task completion survey recently?
         var taskAskedRecently = parseInt(localStorage.getItem('taskTracker'), 10) > Date.now();
 
-        if (!ignore && !taskAskedRecently && !helpfulnessAskedRecently) {
+        if (ga && (!ignore && !taskAskedRecently && !helpfulnessAskedRecently)) {
```
stephaniehobson: We want the survey to be available to people who have chosen to block ga so the ga check needs to only be specifically for reporting to ga and not effect other functionality.

schalkneethling:
> The dnthelper is render blocking.

Indeed. I thought because you referred to external scripts you were specifically referencing GA.

> Only inline scripts in the head before CSS files are non-render blocking. The only reason it is not effecting performance in my tests is that it executes before the CSS files complete downloading. I am concerned about the performance of this outside of ideal test conditions.

Agreed, I was on the fence about placing this higher up. I am going to move the block down to it's previous position in the markup i.e. just above the closing `head`

kuma/static/js/task-completion.js
schalkneethling: Ah right, I completely overlooked that in my eagerness to make the smallest change possible ;p

schalkneethling:
Thanks for the review @stephaniehobson - updated. I ended up not moving the GA block to just above the `head` but, to just after the CSS blocks.

jinja2/includes/google_analytics.html
```diff
@@ -1,65 +1,67 @@
 {% if config.GOOGLE_ANALYTICS_ACCOUNT != '0' %}
 {% set GA_FILE = 'analytics_debug' if settings.DEBUG else 'analytics' %}
 <script type="text/javascript">
+    // only load GA if DNT is not enabled
+    if (Mozilla && !Mozilla.dntEnabled()) {
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+        })(window,document,'script','//www.google-analytics.com/{{ GA_FILE }}.js','ga');
```
jwhitlock: This bit of compressed JS does a couple of things. It create a global object ``ga`` that stores whatever you call on it, which then gets used in the script. It also adds a ``<script>`` tag that loads Google Analytics. When that loads, it creates the "real" ``ga`` object and sends the cached calls to Google. It's a clever bit of JS.

Here's my "uncompressed" version:

```js
(function(win, doc, tagName, src, ga, elem, elemCreated) {
    win['GoogleAnalyticsObject'] = ga;
    win[ga] = win[ga] || function() {
        (win[ga].q = win[ga].q || []).push(arguments)
    }, win[ga].l = 1 * new Date();
    elem = doc.createElement(tagName);
    elemCreated = doc.getElementsByTagName(tagName)[0];
    elem.async = 1;
    elem.src = src;
    elemCreated.parentNode.insertBefore(elem, elemCreated)
})(window, document, 'script', '//www.google-analytics.com/{{ GA_FILE }}.js', 'ga');
```

We could split this up: 

1. Create the global caching ``ga`` object, and populate it no matter what.  Something like:

```js
(function(win, ga, src) {
    win['GoogleAnalyticsObject'] = ga;
    win[ga] = win[ga] || function() {
        (win[ga].q = win[ga].q || []).push(arguments)
    }, win[ga].l = 1 * new Date();
    win['GoogleAnalyticsSrc'] = src;
})(window, '//www.google-analytics.com/{{ GA_FILE }}.js', 'ga');
```
2. Run the DNT code, and only if DNT is not enabled, add the ``<script>`` to load the real GA code. We can use pipeline to combine the upstream DNT code with the enabling code:

```
if !Mozilla.dntEnabled() {
    (function(win, srcName, doc, tagName, elem, elemCreated) {
        elem = doc.createElement(tagName),
        elemCreated = doc.getElementsByTagName(tagName)[0];
        elem.async = 1;
        elem.src = win[srcName];
        elemCreated.parentNode.insertBefore(elem, elemCreated)
    })(window, 'GoogleAnalyticsSrc', document, 'script');
}
```

I suspect that if I read the GA code, it also does DNT detection. But I'm at my limit for JavaScript for the morning.

schalkneethling: Hmmm so, I have two concerns here:

1. Do we really want to mess with Google's code? What you are suggesting seems sound but,
2. If GA then is not loaded all calls to `ga` inside the code will actually be execute and pushed onto the tmp array

Depending on the number of calls on a single page this is more than likely not a big deal but, I am not sure its required. The concern Stephanie had is resolved by moving the analytics block to below the CSS block, and moving the GA calls in `task-completion` to `analytics.js`

Thoughts @jwhitlock @stephaniehobson

jwhitlock: If you are both happy with the current solution, then we should ship it.

schalkneethling: Sure thing, lets wait and see what Stephanie thinks. Thanks!

stephaniehobson:
I'm still getting the error `ga is not defined` when the task completion survey executes but ga is blocked :(

Not sure why, the abstractions you made are 💯.

Maybe try `if(win.ga)`. If that works maybe it should also be `if(window.Mozilla && ...`

Another possible solution would be to spoof the ga object if dnt is enabled.

jinja2/includes/google_analytics.html
stephaniehobson: Yeah, I'd be nervous about messing with the ga code. Scrapers and stuff get annoyed at us if we mess with it (it can break our webmaster tools integration for example). 

I wish we could though, it would be nice to push the date stamp earlier in the rendering process.

schalkneethling:
> Maybe try if(win.ga)

Yup, it wants `win.ga` which is strange as all of the calls was just using `ga` before ¯\_(ツ)_/¯

Thanks for the review @stephaniehobson

stephaniehobson:
R+ 🔥

PR #4578, opened 2017-11-30
bug 1422009: Remove DB from readiness probe

jwhitlock:
Use the same basic smoke test (Django running, middleware working) for both the liveness and readiness probe. See https://github.com/mozmeao/infra/issues/660, [bug 1422009](https://bugzilla.mozilla.org/show_bug.cgi?id=1422009).

### DIFF
--- a/kuma/health/tests/test_views.py
+++ b/kuma/health/tests/test_views.py
@@ -1,31 +1,12 @@
-import mock
 import pytest
-from django.db import DatabaseError
 from django.core.urlresolvers import reverse
 
 
 @pytest.mark.parametrize('http_method',
                          ['get', 'head', 'put', 'post', 'delete', 'options'])
-def test_liveness(db, client, http_method):
-    url = reverse('health.liveness')
+@pytest.mark.parametrize('endpoint', ['health.liveness', 'health.readiness'])
+def test_basic_health(db, client, http_method, endpoint):
+    url = reverse(endpoint)
     response = getattr(client, http_method)(url)
     assert (response.status_code ==
             204 if http_method in ('get', 'head') else 405)
-
-
-@pytest.mark.parametrize('http_method',
-                         ['get', 'head', 'put', 'post', 'delete', 'options'])
-def test_readiness(db, client, http_method):
-    url = reverse('health.readiness')
-    response = getattr(client, http_method)(url)
-    assert (response.status_code ==
-            204 if http_method in ('get', 'head') else 405)
-
-
-def test_readiness_with_db_error(db, client):
-    url = reverse('health.readiness')
-    with mock.patch('kuma.wiki.models.Document.objects') as mock_manager:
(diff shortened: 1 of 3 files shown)


kuma/health/tests/test_views.py
```diff
@@ -1,31 +1,13 @@
 import mock
```
escattone: This is causing flake8 to complain.

jwhitlock:
I may have a basic misunderstanding. Which probe is the one that gets called as a container is booting, and which probe is called as the container is running? I think it may be OK to query the DB on startup, but not while running.

safwanrahman:
As far as I understand, `readiness` only runs at begining of starting pods to check if the container is ready. I dont know how it makes the container restart.

safwanrahman:
We can increase the `timeoutSeconds` [here](https://github.com/mozmeao/infra/blob/master/apps/mdn/mdn-aws/k8s/kuma.deploy.yaml.j2#L15) to increase the timeout. By default its 1 seconds and 1 seconds is pretty low for this. When too many request comes in, the default gunicorn server may not serve a sample request within 1 seconds.

jwhitlock:
force-push to remove ``import mock``.

@safwanrahman thanks but I think you are guessing, which is how we're in a bad situation. It would helpful to have documentation, code, or logs that says what is actually happening.

safwanrahman:
@jwhitlock Here is the documentation
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

escattone:
My understanding, from the K8s docs (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/):
- both probes are run continuously (frequency depends on the `periodSeconds` setting, ours is 3 seconds)
- if the liveness probe fails, the pod will be terminated and restarted
- if the readiness probe fails, the pod will be excluded from receiving traffic, which will be re-routed to other pods

escattone:
Sorry @safwanrahman , I didn't see your comment until after I had submitted mine. 😄

safwanrahman:
I have make some changes into the test to check whether liveliness uses any database query, and it seems like it does not need any database query.

escattone:
Yes, good point @safwanrahman, the default "timeoutSeconds" setting of 1 sec for the liveness/readiness probes may be too short. In terms of an upper bound, I don't think it makes sense for it to be greater than the "periodSeconds" setting.

jgmize:
@jwhitlock will you please add something like the following to top of https://github.com/mozilla/kuma/blob/master/kuma/wiki/middleware.py#L26 as we discussed earlier?

```python
if request.path in ('/healthz', '/readiness'):
    return None
```

jwhitlock:
After reading a few blog posts, I think the database check makes sense in the readiness probe. The Zone check does not. I'm going to start over with a new PR.

PR #4579, opened 2017-11-30
bug 1422009: Adjust liveness tests, avoid DB query in middleware

jwhitlock:
After a long day of looking into Kubernetes liveness and readiness probes, we decided to keep database connection tests in the readiness views for a bit longer (closing PR #4578), and instead adjust the probe settings in Kubernetes (https://github.com/mozmeao/infra/pull/665). There are some changes that we still want to apply. 

* Refactor the health probe tests, to show that the database is not needed for several tests
* Detect health probe URLs in the ``DocumentZoneMiddleware``, and exit early to avoid the (cached) database query.

### DIFF
--- a/kuma/health/tests/test_views.py
+++ b/kuma/health/tests/test_views.py
@@ -6,7 +6,7 @@ from django.core.urlresolvers import reverse
 
 @pytest.mark.parametrize('http_method',
                          ['get', 'head', 'put', 'post', 'delete', 'options'])
-def test_liveness(db, client, http_method):
+def test_liveness(client, http_method):
     url = reverse('health.liveness')
     response = getattr(client, http_method)(url)
     assert (response.status_code ==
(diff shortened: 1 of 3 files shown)


kuma/health/tests/test_views.py
```diff
@@ -4,22 +4,27 @@
 from django.core.urlresolvers import reverse
 
 
-@pytest.mark.parametrize('http_method',
-                         ['get', 'head', 'put', 'post', 'delete', 'options'])
-def test_liveness(db, client, http_method):
+@pytest.mark.parametrize('http_method', ['put', 'post', 'delete', 'options'])
+@pytest.mark.parametrize('endpoint', ['health.liveness', 'health.readiness'])
```
escattone: ❤️ Nice little refactor to use stacked `pytest.mark.parametrize` decorators. I've got to remember to do this more often when it makes sense!

PR #4581, opened 2017-12-01
bug 1406546: wait for mysql ready in docker-compose tests

escattone:
This is another attempt at addressing https://github.com/mozmeao/infra/issues/650. The theory here is that even a freshly created mysql container is sometimes not ready to accept connections until after all of the kuma tests have run and failed. This change only affects the tests run in the Jenkins build (the Travis build only uses the `noext` container, which doesn't depend on the new `mysql-ready` container).

Requires: https://github.com/mozilla/kuma/pull/4583


escattone:
I tested this locally by having the `mysql` container sleep for 10 seconds before starting `mysqld`. The `mysql-ready` behaved as expected (tried to connect every second until it succeeded).

docker-compose.test.yml
```diff
@@ -1,7 +1,13 @@
 version: '2.1'
 services:
+  mysql-ready:
+    image: mysql:5.6
+    depends_on:
+      - mysql
+    command: sh -c "while ! mysqladmin --host=mysql --user=root --password=kuma ping; do sleep 1; done"
```
metadave: this will run until it reaches the end of the internet, which may be ok if a dev is waiting for the build. 

there is also: https://github.com/pmclanahan/urlwait

metadave:
looks good, might be nice to have a timeout of a few minutes, but probably not a big deal either way.

docker-compose.test.yml
escattone: Oops, good catch @metadave! Thanks for the reference to @pmclanahan 's package. I wasn't aware of that.

escattone:
Thanks @metadave! I'm going to refactor this to either use @pmclanahan 's `urlwait` or add a timeout.

docker-compose.test.yml
pmac: It is not well advertised :) Feel free to ping me with questions if you decide to go that way.

escattone:
cc @safwanrahman

escattone:
I had a offline conversation with @jwhitlock who suggested that he'd prefer to see the `urlwait` command simply used inline before the `py.test` commands, like `urlwait && py.test ...`, so I'm going to modify this PR to accommodate that suggestion.

escattone:
I changed the approach to calling `urlwait` prior to the `py.test` commands. This PR now requires #4583 to be merged before it will work.

PR #4585, opened 2017-12-04
bug 1406546: ensure docker-compose up returns proper exit code

escattone:
This PR fixes a problem within the Groovy `utils.compose_test` function. Currently, when [this line](https://github.com/mozilla/kuma/blob/24744d780cb6a80f250a399feb2ed6f707cffbb5/Jenkinsfiles/utils.groovy#L138) is run, the exit code is NOT the exit code from the `test` container's command (`py.test ...`) as it should be. This resulted in Jenkins builds that failed the tests but would pass the stage, resulting in an "unstable" build (labelled with an exclamation point in yellow). This fixes that.

### DIFF
--- a/Jenkinsfiles/utils.groovy
+++ b/Jenkinsfiles/utils.groovy
@@ -135,7 +135,7 @@ def compose_test() {
     // Build the static assets required for many tests.
     sh_with_notify("${dc} run noext make build-static", 'Build static assets')
     // Run the Kuma tests, building the mysql image before starting.
-    sh_with_notify("${dc} up --build test", 'Kuma tests')
+    sh_with_notify("${dc} up --build --exit-code-from=test test", 'Kuma tests')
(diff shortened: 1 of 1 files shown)


escattone:
Oops, yep that's my fault, thanks @jwhitlock. I ran this code on Jenkins yesterday (https://ci.us-west.moz.works/blue/organizations/jenkins/mdn_multibranch_pipeline/detail/test-groovy-failures/4/pipeline) but skipped past the usage blurb to the "script returned exit code 1" and thought it had "worked" because I had purposely changed the `test` container to fail. I'll take anther crack at this.

escattone:
Upgrading `docker-compose` on the Jenkins servers is blocked by the Deis 1 decom (https://github.com/mozmeao/infra/issues/655), so I'm going to close this and try a different approach (PR coming).

PR #4589, opened 2017-12-06
bug 1110799: redirect legacy domains

escattone:
This PR provides the ability to configure (via the environment) one or more "legacy" domains. All requests from any one of these "legacy" domains will be redirected to the configured `SITE_URL` with the same path and query parameters. See its companion PR https://github.com/mozmeao/infra/pull/672.

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -1,9 +1,14 @@
 import contextlib
 import urllib
+from urlparse import urljoin
 
 from django.conf import settings
 from django.core import urlresolvers
-from django.http import HttpResponseForbidden, HttpResponsePermanentRedirect, HttpResponseRedirect
+from django.http import (
+    HttpResponseRedirect,
+    HttpResponseForbidden,
+    HttpResponsePermanentRedirect,
+)
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
 from django.contrib.sessions.middleware import SessionMiddleware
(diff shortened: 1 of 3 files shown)


kuma/settings/common.py
```diff
@@ -1187,9 +1188,13 @@ def pipeline_one_scss(slug, **kwargs):
...
-ALLOWED_HOSTS = config('ALLOWED_HOSTS',
-                       default='developer-local.allizom.org, mdn-local.mozillademos.org',
-                       cast=Csv())
+ALLOWED_HOSTS = config(
+    'ALLOWED_HOSTS',
+    default='developer-local.allizom.org, mdn-local.mozillademos.org',
+    cast=Csv()
+)
+
+LEGACY_DOMAINS = config('LEGACY_DOMAINS', default='', cast=Csv())
```
safwanrahman: I think the naming should be `LEGACY_HOSTS` for better readibility.

safwanrahman:
Is it possible to check somewhere that leagy domains dos not in `ALLOWED_DOMAINS`?

kuma/settings/common.py
escattone: Thanks @safwanrahman, I like that better as well.

kuma/core/middleware.py
```diff
@@ -190,3 +195,16 @@ def process_request(self, request):
...
+
+class LegacyDomainRedirectsMiddleware(object):
+
+    def process_request(self, request):
+        """
+        Permanently redirects all requests from legacy domains.
+        """
+        if request.get_host() in settings.LEGACY_HOSTS:
+            return HttpResponsePermanentRedirect(
+                urljoin(settings.SITE_URL, request.get_full_path())
```
jwhitlock: 👀 You did the fancy redirect, nice!

PR #4591, opened 2017-12-06
Add ratelimiting to homepage, document, and revision views

jgmize:


safwanrahman:
I think document list views also should be under ratelimitation.

metadave:
can the values `400/m`, `40/m`, `20/m`, `15/m` be constants?

jwhitlock:
I asked @jgmize not to make them constants. I find it challenging to understand what is going on when ``RATELIMIT_FOR_HOME`` is used, and ``RATELIMIT_40_REQUESTS_PER_MINUTE`` is worse than the literal string, because you have to go verify the value is what the name implies.

But, you are the second person to ask for a constant, so this may be a situation where I'm the holdout, and we should use the constant.

metadave:
I wasn't thinking of constants like `RATELIMIT_40_REQUESTS_PER_MINUTE`, but rather have them grouped according to usage. ex: `RATELIMIT_HIGH = 400/m`, `RATELIMIT_LOW = 15/m`

PR #4592, opened 2017-12-07
bug 1402950: Run tests more reliably with Firefox and Remote driver in Selenium 3.x

jwhitlock:
These changes make Firefox run more reliably with the Remote driver in Selenium 3, and should still work in Selenium 2:

* To display the on-hover submenus, first move the mouse to the logo. This more reliably triggers the hover state on the article page.
* The hover is still unreliable, especially on the homepage. If the driver is Remote for Firefox, then allow hover tests to xfail.
* On the Revision Dashboard, it can take time to expose the revision diff after the other details are displayed. Be patient.

There was also an error in ``test_dashboard_overflow``, where the wrong attributes made it always ``xfail``. Now it can ``xpass`` as well.

### DIFF
--- a/tests/functional/test_dashboard.py
+++ b/tests/functional/test_dashboard.py
@@ -50,7 +50,7 @@ def test_dashboard_overflow(base_url, selenium):
     """
     page = DashboardPage(selenium, base_url).open()
     page.open_first_details()
-    assert page.scrollWidth < page.clientWidth
+    assert page.scroll_width <= page.client_width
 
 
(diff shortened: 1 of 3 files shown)


jwhitlock:
I pushed this branch to run [on Jenkins](https://ci.us-west.moz.works/blue/organizations/jenkins/mdn_multibranch_pipeline/detail/remote-firefox-sel3-tests-1402950/1/pipeline/), and it passed.

PR #4593, opened 2017-12-07
bug 1402950: Add Chrome to integration tests

jwhitlock:
* Add Google Chrome to the integration test suite. In our selenium 2 version, this is Chrome 47, released December 2016.
* Run Chrome, Firefox, and headless (requests) tests in parallel. Add a browser prefix to distinguish between functional tests.
* Run parallel functionaltests equal to the number of selenium nodes
* Manually create and stop containers, to avoid a bug in auto-stopping containers in dockerRun
* Re-run tests on failure, to deal with flakiness of low-usage infrastructure like staging
* Be doubly verbose
* In several runs, a chrome node would become unresponsive, and the test would hang forever. Timeout the pytest run in 4 minutes, and enforce a Jenkins timeout of 5 minutes.
* Notify IRC on tests success or failure

### DIFF
--- a/Jenkinsfiles/integration-tests.groovy
+++ b/Jenkinsfiles/integration-tests.groovy
@@ -6,41 +6,122 @@ stage('Build') {
   }
 }
 
-stage('Functional') {
-  def pwd = pwd()
-  def cmd = "py.test tests/functional" +
-            " --driver Remote" +
-            " --capability browserName firefox" +
-            " --host hub" +
-            " --base-url='${config.job.base_url}'" +
-            " --junit-xml=/test_results/integration.xml"
-  if (config.job && config.job.tests) {
-    cmd += " -m \"${config.job.tests}\""
-  }
-  if (config.job && config.job.maintenance_mode) {
-    cmd += " --maintenance-mode"
-  }
+def functional_test(browser, hub_name, base_dir) {
+  // Define a parallel build that runs Selenium Node for the given browser
+  return {
+    node {
+      // Setup the pytest command
+      // Timeout after 4 minutes, due to stalled nodes
+      def cmd = "timeout --preserve-status 4m" +
+                " py.test tests/functional" +
+                " --driver Remote" +
+                " --capability browserName ${browser}" +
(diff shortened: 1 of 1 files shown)


jwhitlock:
I tried using the pytest retry failures plugin, and it looks like we don't have it installed. I'll remove.

jwhitlock:
After several rounds, I'm happy with the new integration suite. It fails more than it passes, due to many possible causes:

* One of the Chrome nodes will become unresponsive, and won't return a test failure. ``pytest`` waits forever. Mitigated by timing out ``pytest`` after 4 minutes, and the build step after 5 minutes. A successful run takes about a minute.
* stage.mdn.moz.works returns a 503 error 1% - 5% of the time. This may be due to running stage with limited resources, such as a single web pod. I've added ``--reruns=2``, so that an unlucky test will be re-run and hopefully be luckier.
* On test failures (and maybe on success!), docker nodes tend to remain running. Manually stop them by name.

Out of [16 runs](https://ci.us-west.moz.works/blue/organizations/jenkins/mdn_multibranch_pipeline/activity?branch=chrome-func-tests-1402950), only 2 passed. However, run 16 failed in a normal way (test failure), and test 15 failed with a timeout rather than being manually stopped.

Despite the issues with the test runs, I'd like to ship this, so I can move on to upgrading Selenium and the browsers.

safwanrahman:
> stage.mdn.moz.works returns a 503 error 1% - 5% of the time. This may be due to running stage with limited resources, such as a single web pod. I've added --reruns=2, so that an unlucky test will be re-run and hopefully be luckier.

Can we use the `readiness/` endpoint in test to check if the host is running? So that the test will not run if its 503

jwhitlock:
Staging is available during the tests. The problem is that the occasional request returns an error. If I'm right that this is due to the limited resources in our staging environment, then doubling requests for each test by checking the readiness endpoint would double the chances of seeing an error.

We could scale up staging so that it is as reliable as production. I think this would be a bad allocation of resources.

I believe a retry is the right solution.

PR #4597, opened 2017-12-08
Fix Bug 1401291, change warnings to error - eslint

schalkneethling:
Now attempting a commit will fail with exit code 1, if there are linting errors o/\o

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,68 +1,68 @@
 module.exports = {
-    "env": {
-        "browser": true,
-        "jquery": true
+    'env': {
+        'browser': true,
+        'jquery': true
     },
-    "extends": "eslint:recommended",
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -5,53 +5,53 @@ module.exports = {
     },
     "extends": "eslint:recommended",
     "rules": {
-        "no-global-assign": 1,
+        "no-global-assign": 2,
```
jwhitlock: According to the [online docs](https://eslint.org/docs/user-guide/configuring#configuring-rules), this can be ``2`` or ``"error"``. Can we use the text version instead?

schalkneethling:

stephaniehobson:
💃 🎉 🍾

PR #4599, opened 2017-12-08
bug 1406546: fix kuma tests by tearing-down volumes as well

escattone:
In Jenkins builds, every Kuma test that hit the database started to fail consistently with a `OperationalError: (1130, "Host '172.20.0.5' is not allowed to connect to this MySQL server")`. This was fixed by also tearing down the volumes (using `docker-compose down --volumes`) before running the tests. I also added the "--remove-orphans" option as well, so we start (and end) with a totally clean slate.

Thanks to this thread: https://github.com/docker-library/mysql/issues/275 for the idea.

### DIFF
--- a/Jenkinsfiles/utils.groovy
+++ b/Jenkinsfiles/utils.groovy
@@ -144,8 +144,9 @@ def announce_push() {
 
 def compose_test() {
     def dc = 'docker-compose -f docker-compose.yml -f docker-compose.test.yml'
+    def dc_down = "${dc} down --volumes --remove-orphans"
     // Pre-test tear down to ensure we're starting with a clean slate.
-    sh_with_notify("${dc} down", 'Pre-test tear-down')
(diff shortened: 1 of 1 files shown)


Jenkinsfiles/utils.groovy
```diff
@@ -144,8 +144,9 @@ def announce_push() {
 
 def compose_test() {
     def dc = 'docker-compose -f docker-compose.yml -f docker-compose.test.yml'
+    def dc_down = "${dc} down --volumes --remove-orphans"
```
safwanrahman: Is it possible to set environment variable `COMPOSE_FILE=docker-compose.yml:docker-compose.test.yml` above, so that you can just call `dc_down=docker-compose down --volumes --remove-orphans`

jwhitlock: That's an interesting idea, but should be in a new PR.

PR #4601, opened 2017-12-10
Use TravisCI Edge images

jwhitlock:
This is the ``travis_latest`` version that will go live on Dec 11 2017:

https://blog.travis-ci.com/2017-12-01-new-update-schedule-for-linux-build-images


jwhitlock:
Builds pass, we should be in good shape when this goes live today.

PR #4605, opened 2017-12-21
Fix Bug 1420371: Samsung, Firefox, footnote, icon update

stephaniehobson:
- Add Samsung Internet Browser Icon
- Update Firefox icon
- Change footnote icon from `[ ]` to `*`


stephaniehobson:
@Elchi3 🎁

schalkneethling:
Not sure if I missed a step but, I fetched this branch, updated the sub-modules, started docker, ran build-static, restarted web as well as kumascript, and have gulp running.

With all of the above, the new/changed icons are not showing up:

screen shot 2017-12-21 at 12 16 14

and

screen shot 2017-12-21 at 12 19 07

jwhitlock:
If you are removing ``compat-icons.eot``, etc., you'll need to remove the reference from ``kuma/static/styles/base/fonts/compat-icons.scss`` as well.

schalkneethling:
> If you are removing compat-icons.eot, etc., you'll need to remove the reference ...

Ah yes, that is why there was the message about unliked files from Gulp

stephaniehobson:
Oh no, I put the new files in `/static/` instead of `/kuma/static/`. I hope I still have them...

`/static/` is my nemesis.

stephaniehobson:
I should add, the macro doesn't add Samsung Internet Browser to the table yet, so you'll have to test by changing the class on one of the others with dev tools.

stephaniehobson:
Okay, try it again :)

PR #4614, opened 2018-01-03
bug 1423738: Add 429 "Too Many Requests" response

jwhitlock:
Return a ``429 TOO MANY REQUESTS`` and custom error page when the rate limit is exceeded.

You can view the rate limit page by adding this to a view, such as ``kuma.landing.views.home``:

```
from ratelimit.exceptions import Ratelimited; raise Ratelimited()
```

It looks a lot like the 403 page:

429-tumbeast

### DIFF
--- a/jinja2/403.html
+++ b/jinja2/403.html
@@ -23,6 +23,8 @@
       <p class="notice">{{ _('The wiki is in read-only mode.') }}</p>
     {% elif reason == 'revisions_login_required' %}
       <p class="notice">{{ _("You must sign in to view all history") }}</p>
+    {% elif reason == 'ip_banned' %}
+      <p class="notice">{{ _("Your IP address has been banned.") }}</p>
     {% endif %}
     {% endblock reason %}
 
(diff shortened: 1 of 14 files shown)


jwhitlock:
I am surprised to learn that the ``IPBan`` only applies to editing.  I'll stop using it for misbehaving scrapers, and rely on @metadave's IP ban tools instead.

I'm going to re-write this bit so that a 403 rather than a 429 is raised for IP bans, in a new commit.

safwanrahman:
I think we can include the `Retry-After` header also

jwhitlock:
I don't think we have enough information for an accurate ``Retry-After`` header, but I guess a hard-coded 60 seconds wouldn't hurt.

jwhitlock:
Lots of stuff:

* Rebased on master
* Added ``Retry-After: 60`` header
* Refactored ``IPBan`` view tests
* Rewrote ``IPBan`` logic as a new decorator ``ban_blocked_ips``, and removed it from the rate limiting decorator.
* Rewrote ``IPBan`` job to cache a list of banned IPs rather than a cache item for each banned IP. I suspect that editing is too rare to justify any caching, but I'm following the lead of past developers.

PR #4621, opened 2018-01-09
bug 1173189: Force timezone-aware times for feeds

jwhitlock:
Re-write all the feeds tests, to use ``pytest`` and fixtures, for full code coverage of ``kuma/wiki/feeds.py``, and to avoid silly things like sleeping for a second to get a new timestamp.

Then, add some additional handling of our naive timestamps, to turn them into timezone-aware timestamps needed for the feeds. This avoid errors in the second half of the year when the US moves from Daylight Savings Time to standard time, and 1 AM happens twice. When these ambiguous times are encountered, they currently cause the feed to fail with an error. The change is to assume that it is the first 1 AM, avoiding the conversion error.

### DIFF
--- a/kuma/wiki/feeds.py
+++ b/kuma/wiki/feeds.py
@@ -8,6 +8,7 @@ from django.db.models import F
 from django.utils.feedgenerator import (Atom1Feed, Rss201rev2Feed,
                                         SyndicationFeed)
 from django.utils.html import escape
+from django.utils.timezone import get_current_timezone, is_naive
 from django.utils.translation import ugettext as _
 
 from kuma.core.templatetags.jinja_helpers import add_utm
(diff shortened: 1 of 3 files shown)


kuma/wiki/tests/test_feeds.py
```diff
@@ -1,262 +1,485 @@
...
+        creator=edit_revision.creator,
+        tags='"TheTag"')
+    resp = client.get(feed_url)
+    assert resp.status_code == 200
+    data = json.loads(resp.content)
+    assert len(data) == 1
+
+
+def test_recent_documents_handles_ambiguous_time(root_doc, client):
+    """The recent_documents feed handles times during DST transition."""
```
escattone: Super nit: `recent documents` rather than `recent_documents`?

jwhitlock:
Force-push to rebase and fix doc nit.

kuma/wiki/feeds.py
```diff
@@ -52,7 +53,33 @@ def get_object(self, request, format):
...
+
+        https://stackoverflow.com/questions/21465528/resolving-ambiguoustimeerror-from-djangos-make-aware
+        TODO: In Django 1.9, can pass is_dst=False to make_aware
+        TODO: Enable timezone-aware dates in database, remove check
+        """
+        raw_pubdate = document.current_revision.created
+        if is_naive(raw_pubdate):
+            # USE_TZ is False, database has naive timestamps
+            timezone = get_current_timezone()
+            pubdate = timezone.localize(raw_pubdate, is_dst=True)
```
safwanrahman: I could not find any documentation of this `localize` in documentation.
by supplying `is_dst=True`, are we assuming that its always Daylight Saving time?

escattone: @safwanrahman Yes, we're just going to assume it's always DST until we start storing timezone-aware datetimes.

jwhitlock: @safwanrahman I had similar questions that were not answered by the docs I wrote. I assumed it was OK because other tests outside the magic hour passed. I'm investigating a little more now that you've asked it. Results to come...

jwhitlock: ``localize`` is provided by pytz, which has [pretty good narrative documentation](http://pytz.sourceforge.net), but doesn't go into details about the functions.

The details are in the source, hosted on launchpad:

https://git.launchpad.net/pytz/tree/src/pytz/tzinfo.py#n244

The code looks forward a day and back a day, to see if the time is around a transition. This can result in 0, 1, or 2 possible timezones:

- 0 cases: The time is an "impossible time", during the lost hour during the transition into daylight savings time. ``is_dst=None`` raises an error, ``is_dst=True`` uses the timezone 6 hours later, and ``is_dst=False`` uses the timezone 6 hours earlier.
- 1 case: There is no ambiguity, the one timezone is used
- 2 cases: The time is from the transition out of daylight savings time, when an hour is repeated. ``is_dst=None`` raises an error, ``is_dst=True`` uses the DST timezone, ``is_dst=False`` uses the standard timezone.

I'm not Python, so I could have read the code wrong. Here's some test code in the Django shell:

```
>>> from datetime import datetime as dt
>>> from django.utils.timezone import get_current_timezone
>>> for month in range(1,13):
...   test = dt(2017, month, 3, 1, 23)
...   print(get_current_timezone().localize(test))
...
2017-01-03 01:23:00-08:00
2017-02-03 01:23:00-08:00
2017-03-03 01:23:00-08:00
2017-04-03 01:23:00-07:00
2017-05-03 01:23:00-07:00
2017-06-03 01:23:00-07:00
2017-07-03 01:23:00-07:00
2017-08-03 01:23:00-07:00
2017-09-03 01:23:00-07:00
2017-10-03 01:23:00-07:00
2017-11-03 01:23:00-07:00
2017-12-03 01:23:00-08:00
```
Here in the Central time zone, the dates from April through November are Central Daylight Time (-07:00), and the other months are Central Standard Time (-8:00).

That was a fun question, thanks!

PR #4630, opened 2018-01-16
[Bug 1431139] Optimize Documents view of kuma

safwanrahman:
In the current platform, it seems like many more unneeded SQL queries that kills most of the time!

It can be more optimized, but for now I think small is also benefecial.

**Current Situation:**
image

**After this Patch:**
image

I was thinking about fetching all the translations data of the document using elasticsearch or cache. As making a sql queries just for showing the translation information is a bit tradeoff with the performance. What do you think @jwhitlock?

@jwhitlock @escattone Can you please review?
It partially fiexes https://github.com/mozmeao/infra/issues/677

### DIFF
--- a/kuma/wiki/jinja2/wiki/document.html
+++ b/kuma/wiki/jinja2/wiki/document.html
@@ -23,11 +23,10 @@
 {% set nearest_zone = document.nearest_zone %}
 {% set is_zone = nearest_zone %}
 {% set is_zone_root = document.is_zone_root %}
-{% set other_translations = document.other_translations %}
 
 {% set show_left = zone_subnav_html or quick_links_html or approvals_html %}
 
-{% set buttons = get_document_buttons(document, help_link) %}
+{% set buttons = get_document_buttons(document, help_link, other_translations) %}
 {% set crumbs = build_document_crumbs(document) %}
 
 
@@ -202,7 +201,7 @@
                 {% include 'wiki/includes/kumascript_errors.html' %}
               {% endif %}
 
-              {% if document.current_revision %}
+              {% if document.parent_id and document.current_revision %}
                 {% if document.current_revision.localization_in_progress %}
                   {% if not settings.MAINTENANCE_MODE and document.current_revision.translation_age >= 10 %}
                     <div class="overheadIndicator translationInProgress"><p>{% trans translate_url=document.get_edit_url() %}
@@ -245,7 +244,7 @@
               <!-- contributors -->
               <div class="wiki-block contributors">
                 <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
-                {% set tags = document.tags.all().order_by('name') %}
+                {% set tags = document.all_tags_name %}
                 {% include "wiki/includes/document_tag.html" %}
 
                 {% if has_contributors %}
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -35,7 +35,7 @@ class DocumentNearestZoneJob(KumaJob):
                                  .get)
         while pk:
             try:
-                return DocumentZone.objects.get(document=pk)
+                return DocumentZone.objects.select_related('document').get(document=pk)
             except DocumentZone.DoesNotExist:
                 pk = get_parent_id(pk=pk)
 
@@ -127,3 +127,27 @@ class DocumentCodeSampleJob(GenerationJob):
 
     def empty(self):
         return {}
+
+
+class DocumentTagsJob(KumaJob):
+    """
+    Given a wiki document returns a list of tags.
+
+    We invalidate this when a document is saved only.
+    Longer lifetime as tags are rarely modified
+    """
+    refresh_timeout = 180
+
+    @property
+    def lifetime(self):
+        # Spread the life time across a random
+        # number of days from 1 to 10 (in units of seconds).
+        # So that all the document cache do not get expired at same time
+        seconds_per_day = 24 * 60 * 60
+        return random.randint(1 * seconds_per_day, 10 * seconds_per_day)
+
+    def fetch(self, pk):
+        from .models import Document
+
+        tags = Document.objects.filter(id=pk).values_list('tags__name', flat=True).order_by('tags__name')
+        return list(tags)
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -39,7 +39,7 @@ from .content import (Extractor, H2TOCFilter, H3TOCFilter, SectionTOCFilter,
 from .exceptions import (DocumentRenderedContentNotAvailable,
                          DocumentRenderingInProgress, PageMoveError,
                          SlugCollision, UniqueCollision, NotDocumentView)
-from .jobs import DocumentContributorsJob, DocumentNearestZoneJob
+from .jobs import DocumentContributorsJob, DocumentNearestZoneJob, DocumentTagsJob
 from .managers import (DeletedDocumentManager, DocumentAdminManager,
                        DocumentManager, RevisionIPManager,
                        TaggedDocumentManager)
@@ -338,7 +338,7 @@ class Document(NotificationsMixin, models.Model):
 
     @cache_with_field('toc_html')
     def get_toc_html(self, *args, **kwargs):
-        if not self.current_revision:
+        if not self.current_revision_id:
             return ''
         if not self.current_revision.toc_depth:
             return ''
@@ -1283,7 +1283,7 @@ Full traceback:
 
     @property
     def show_toc(self):
-        return self.current_revision and self.current_revision.toc_depth
+        return self.current_revision_id and self.current_revision.toc_depth
 
     @cached_property
     def language(self):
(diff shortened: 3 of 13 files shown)


kuma/wiki/models.py
```diff
@@ -413,7 +413,9 @@ def get_zone_subnav_html(self):
         """
         src = self.get_zone_subnav_local_html()
         if (not src) and self.nearest_zone:
-            src = self.nearest_zone.document.get_zone_subnav_local_html()
+            document_id = self.nearest_zone.document_id
+            document = Document.admin_objects.only('html', 'rendered_html', 'zone_subnav_local_html').get(id=document_id)
```
jwhitlock: Why ``admin_objects`` (which returns deleted documents as well) instead of ``objects``?

kuma/wiki/jinja2/wiki/document.html
```diff
@@ -202,7 +201,7 @@ <h1>{{ document.title }}</h1>
                 {% include 'wiki/includes/kumascript_errors.html' %}
               {% endif %}
 
-              {% if document.current_revision %}
+              {% if document.parent_id and document.current_revision %}
```
jwhitlock: This is probably a good optimization. I'll need to check current docs to see if skips the section when it shouldn't.

kuma/wiki/models.py
```diff
@@ -1375,14 +1377,7 @@ def other_translations(self):
...
         """
-        if self.parent is None:
-            return list(self.translations.all().order_by('locale'))
-        else:
-            translations = list(self.parent.translations
-                                .exclude(id=self.id)
-                                .order_by('locale'))
-            # The parent doc should be at first
-            return [self.parent] + translations
+        return self.get_other_translations()
```
jwhitlock: Should ``other_translations`` be deleted instead? If there are other callers, you've made them much less efficient...

kuma/wiki/models.py
```diff
@@ -1466,7 +1474,7 @@ def nearest_zone(self):
     @cached_property
     def is_zone_root(self):
         zone = self.nearest_zone
-        return bool(zone) and (zone.document in (self, self.parent))
+        return bool(zone) and (zone.document_id in (self.id, self.parent_id))
```
jwhitlock: 👍 if this removes a query

kuma/wiki/utils.py
```diff
@@ -182,3 +183,12 @@ def analytics_upageviews_by_revisions(revisions):
     start_date = min(r.created for r in revisions)
 
     return analytics_upageviews(revision_ids, start_date)
+
+
+def get_document_tags(document_id):
+    """Get the translations of the document and return locale, slug, title"""
```
jwhitlock: The docstring says it gets translations of the document, but I don't see it in the code. Maybe I need to check the ES return.

kuma/wiki/views/document.py
```diff
@@ -117,8 +117,8 @@ def _get_seo_parent_title(slug_dict, document_locale):
     """
     if slug_dict['seo_root']:
         try:
-            seo_root_doc = Document.objects.get(locale=document_locale,
-                                                slug=slug_dict['seo_root'])
+            seo_root_doc = Document.objects.only('title').get(locale=document_locale,
```
jwhitlock: 👍 should be kinder to the database

metadave:
@jwhitlock nothing in the [pricing doc](https://www.elastic.co/cloud/as-a-service/pricing) mentions number of requests. As far as I know we pay for high-level infra costs: primary region, # of DC's to host in, CPU/disk size, Google vs AWS.

kuma/wiki/utils.py
safwanrahman: Sorry, my bad.
I wrote this docstring for other function, but it got removed later. I will fix it in next commit

kuma/wiki/models.py
safwanrahman: There are some other callers that can use `get_other_translations()` in order to be more optimized. It makes them much less efficient but changing that can make this PR more complex to review. I can work on that in different PR if you agree.

kuma/wiki/models.py
safwanrahman: For some [historical purpose](https://bugzilla.mozilla.org/show_bug.cgi?id=1365960), we have seen some error in `DocumentNearestZoneJob` about already deleted document makes `500` error
Therefore used `admin_objects` instead.

safwanrahman:
@jwhitlock Removed the elasticsearch dependency and made more optmization.
Can you give another round of review?

jwhitlock:
@safwanrahman I'll take a look when the tests are passing

jwhitlock:
Thanks @safwanrahman! I'll review on Monday.

safwanrahman:
Thanks @jwhitlock. I need to remove the 6 years old test case and write some other small test cases that cover the full tags functionality. The test case was testing multiple things and calling get request to same endpoint multiple times and because of page cache, the invalid content were served and made the test broken.
I have split up the test in different test cases and place them to the appropiate place where they should be.

jwhitlock:
My process for working with document functionality:

- Fix the code
- Identify the tests that exercise the code and are now broken
- ``git stash`` my code or put it in a local branch
- Refactor the tests to look like I wish they did when I started
- Write a new test that demonstrates the thing that I wanted to fix, to show that it is broken
- Bring in my new code to show I fixed it, without breaking any other code

This results in two commits, a test refactor plus a change in functionality.  See PR #4621 for a recent example.

jwhitlock:
Sorry @safwanrahman, I had too many urgent tasks come up today, it will be tomorrow morning (Tuesday) when I can look at this.

jwhitlock:
There's now a conflict in ``kuma/wiki/jinja2/wiki/includes/buttons.html`` - can you rebase and resolve the conflict?

safwanrahman:
@jwhitlock I have rebased and made a small fix.

safwanrahman:
Thanks @jwhitlock for the review. I will fix it tomorrow.

kuma/wiki/jobs.py
```diff
@@ -127,3 +127,20 @@ def fetch(self, pk, sample_name):
 
     def empty(self):
         return {}
+
+
+class DocumentTagsJob(KumaJob):
```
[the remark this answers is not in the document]
escattone: @jwhitlock, Great point. We did two things to the `DocumentNearestZoneJob` class to mitigate the spike in jobs:
- Increased the `refresh_timeout` to 180 seconds
- Spread the cache expiration times across a random number of days by overriding the `expiry` method.

For details/reference see https://github.com/mozilla/kuma/pull/4272 and [bugzilla 1373720](https://bugzilla.mozilla.org/show_bug.cgi?id=1373720).

I think we should do the same for `DocumentTagsJob`.

kuma/wiki/models.py
escattone: @safwanrahman Given your change to the `DocumentNearestZoneJob.fetch` above, which populates the zone's related document, it seems this change is no longer helpful to save a query?

kuma/wiki/templatetags/jinja_helpers.py
```diff
@@ -207,7 +207,7 @@ def document_zone_management_links(user, document):
 
     # Enable "add" link if there is no zone for this document, or if there's a
     # zone but the document is not itself the root (ie. to add sub-zones).
-    if ((not zone or zone.document != document) and
+    if ((not zone or zone.document_id != document.id) and
```
escattone: @safwanrahman Again, given your change to the `DocumentNearestZoneJob.fetch` above, which populates the zone's related document, it seems this change is no longer helpful to save a query?

kuma/wiki/models.py
safwanrahman: I think, `self.parent` would make another call if not compare by id!

escattone: Ah yes, thanks @safwanrahman!

kuma/wiki/templatetags/jinja_helpers.py
safwanrahman: Yes. it seems to be not saving a query anymore. but I thought abou deferred object can give wrong result in comparing, but it seems that they may make another query if value is not present.
So I think comparing by id would be feasible.

escattone: @safwanrahman Yes, thanks.

kuma/wiki/tests/test_views.py
```diff
@@ -1681,65 +1682,61 @@ def test_translation_source(self):
...
-                response = self.client.get(reverse('wiki.feeds.recent_documents',
-                                                   args=['atom', t]))
-                self.assertNotContains(response, doc.title)
-
-        # Create a new doc with tags
-        data.update({'slug': slug, 'tags': ','.join(ts1)})
-        self.client.post(reverse('wiki.create'), data)
-        assert_tag_state(ts1, ts2)
+    @pytest.mark.tags
+    def test_tags_showing_correctly_after_doc_update(self):
```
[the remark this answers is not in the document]
safwanrahman: `test_tags_show_in_document` and this are not same actually.
I wanted to test that after updating the doc with new tag from http request, the page shows the new tag.
Its a edge case that I thought while changing this.

safwanrahman:
@jwhitlock Updated as per commented.
r?

kuma/wiki/jobs.py
```diff
@@ -127,3 +127,27 @@ def fetch(self, pk, sample_name):
...
+    """
+    Given a wiki document returns a list of tags.
+
+    We invalidate this when a document is saved only.
+    Longer lifetime as tags are rarely modified
+    """
+    refresh_timeout = 180
+
+    @property
+    def lifetime(self):
```
jwhitlock: Not what we asked, but it seems to work.

jwhitlock:
Rebased to merge into a 1) test refactor, and then 2) optimization

PR #4634, opened 2018-01-24
bug 1136741: Make compat_data flag permanent

jwhitlock:
This removes the flag, but keeps the just-in-time loading of the JS and CSS. However, it may make sense to merge those into ``wiki.js`` / ``wiki.css``.

### DIFF
--- a/docs/feature-toggles.rst
+++ b/docs/feature-toggles.rst
@@ -36,7 +36,6 @@ Flags
 `Waffle flags`_ control behavior by specific users, groups, percentages, and
 other advanced criteria.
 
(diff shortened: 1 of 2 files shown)


jwhitlock:
@schalkneethling do you have an opinion on merging the styles into ``wiki.js`` / ``wiki.css``?

schalkneethling:
> do you have an opinion on merging the styles into wiki.js / wiki.css

Yeah, it might be worth doing that. I need to have a look and perhaps speak to Stephanie to understand if there was a specific reason for doing it this way. That should probably not block this from merging though?

PR #4638, opened 2018-01-27
bug 1431820: fix document-view etag header

escattone:
This PR addresses some shortcomings in the current `document` view related to caching and therefore the CDN work.

- the `ETag` and `Last-Modifed` headers do not reflect the entire response of the endpoint (they're designed to work only with `PUT`'s to prevent editing collisions) (thanks @safwanrahman!)
- conditional `GET`'s are only handled for the `Last-Modifed` header, not the `ETag` header

This PR addresses those shortcomings by:

- creating a new `document_api` view to handle editing
- removing the ability to edit from the `document` view
- removing the `Last-Modified` header from both views
- adding proper `ETag` generation and handling to both views

This is the first of several PR's that will be needed as part of the work to place a CDN in front of most of the `wiki.urls`.


kuma/wiki/views/document.py
```diff
@@ -687,68 +673,118 @@ def document(request, document_slug, document_locale):
...
+@require_http_methods(['GET', 'HEAD', 'PUT'])
+@redirect_in_maintenance_mode(methods=['PUT'])
+@allow_CORS_GET
+@accepts_auth_key
+@process_document_path
+@newrelic.agent.function_trace()
+@ratelimit(key='user_or_ip', rate='100/m', block=True)
+def document_api(request, document_slug, document_locale):
+    """
+    View/modify the content of a wiki document, or create a new wiki document.
```
safwanrahman: I think this API should not create any document!

kuma/wiki/views/document.py
```diff
@@ -687,68 +673,118 @@ def document(request, document_slug, document_locale):
...
+    """
+    if request.method == 'PUT':
+        if not (request.authkey and request.user.is_authenticated()):
+            raise PermissionDenied
+        return _document_api_PUT(request, document_slug, document_locale)
+
+    try:
```
safwanrahman: I think `get_objects_or_404` can be used here.

escattone: Good catch, thanks @safwanrahman !

escattone:
@jwhitlock I just removed the `needs rebase` label. Ready for review!

kuma/wiki/tests/test_views_document.py
```diff
@@ -3,16 +3,84 @@
...
+
+    return content_type, encoded_data
+
+
+@pytest.fixture
+def sectioned_root_doc(wiki_user):
+    """
+    A newly-created top-level English document with sections.
+    """
+    root_doc = Document.objects.create(
```
safwanrahman: There is already a fixture for `root_doc`

kuma/wiki/tests/test_views_document.py
```diff
@@ -30,6 +98,299 @@ def ce_settings(settings):
...
+    url = root_doc.get_absolute_url() + '$api'
+
+    response = client.put(url)
+    assert response.status_code == 403
+
+    # Even logged-in users without an authkey should be forbidden.
+    wiki_user.set_password('password')
+    wiki_user.save()
+
+    assert client.login(username=wiki_user.username, password='password')
```
safwanrahman: There are already another fixture name `user_client` which the client is already logged in

jwhitlock:
Also, the new tests of ``$api`` are better structured, more complete, and catch more cases than the existing tests. You went above and beyond just migrating to ``pytest``, well done.

kuma/wiki/views/document.py
```diff
@@ -709,70 +697,124 @@ def document(request, document_slug, document_locale):
...
-    response = render(request, 'wiki/document.html', context)
-    return _set_common_headers(doc, rendering_params['section'], response)
+    def get_etag(*args):
+        return calculate_etag(response.content)
+
+    # The etag decorator not only adds the ETag header to the response,
+    # but also handles conditional responses based on the incoming and
+    # outgoing ETag headers.
+    @etag(get_etag)
+    def get_response(request, *args):
```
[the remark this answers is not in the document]
escattone: I agree that it's awkward. My only excuse was that I thought I should avoid affecting all views by using `ConditionalGetMiddleware`, but maybe that's best in the end. Let me work on something better using `ConditionalGetMiddleware`.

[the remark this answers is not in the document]
jwhitlock: OK, but it looks too big, let's get this merged and work on that in a fresh PR.

kuma/wiki/views/document.py
```diff
@@ -709,70 +697,124 @@ def document(request, document_slug, document_locale):
...
+        return _document_api_PUT(request, document_slug, document_locale)
+
+    doc = get_object_or_404(
+        Document,
+        slug=document_slug,
+        locale=document_locale
+    )
+
+    section_id = request.GET.get('section', None)
+    response = HttpResponse(doc.get_html(section_id))
```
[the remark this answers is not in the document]
escattone: My intention was to avoid changing the semantics of the endpoints as much as possible, but I really like the idea of removing the "section" stuff from the plain document view, and I hate the `$` APIs!

PR #4639, opened 2018-01-29
Fix Bug 1432370, change instances of Mozilla Developer Network to MDN Web Docs

schalkneethling:
I did not make the change in the following files as I believe that it will break functionality:

```
kuma/authkeys/jinja2/authkeys/list.html
kuma/users/jinja2/users/user_delete.html
kumascript/tests/fixtures/hacks.rss
```

Also, none of the `.po` files as this will be done by localisers.

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,6 +1,6 @@
 {
     "name": "Kuma",
-    "description": "The app powering the Mozilla Developer Network (MDN).",
+    "description": "The app powering the MDN Web Docs.",
     "repository": {
         "url": "https://github.com/mozilla/kuma",
         "license": "MPL2",
--- a/contribute.json
+++ b/contribute.json
@@ -1,6 +1,6 @@
 {
     "name": "Kuma",
-    "description": "The app powering the MDN Web Docs.",
+    "description": "The app powering MDN Web Docs.",
     "repository": {
         "url": "https://github.com/mozilla/kuma",
         "license": "MPL2",
--- a/kuma/landing/jinja2/landing/fellowship.html
+++ b/kuma/landing/jinja2/landing/fellowship.html
@@ -9,7 +9,7 @@
 
 {% block extrahead %}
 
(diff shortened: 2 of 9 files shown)


contribute.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Kuma",
-    "description": "The app powering the Mozilla Developer Network (MDN).",
+    "description": "The app powering the MDN Web Docs.",
```
jwhitlock: I think ``the`` should be dropped, ``The app powering MDN Web Docs``.

kuma/landing/jinja2/landing/fellowship.html
```diff
@@ -9,7 +9,7 @@
 
 {% block extrahead %}
 
-  {% set seo_description = _('The Mozilla Developer Network (MDN) Fellowship Program invites world-class web developers to come change the world with Mozilla.') %}
+  {% set seo_description = _('The MDN Web Docs Fellowship Program invites world-class web developers to come change the world with Mozilla.') %}
```
jwhitlock: Let's not edit this one. The next action on https://developer.mozilla.org/en-US/fellowship should be to convert it to a regular wiki page, not to send new strings to translate. I can do this work in a new PR.

kuma/search/jinja2/search/plugin.html
```diff
@@ -1,8 +1,8 @@
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
                        xmlns:moz="http:/www.mozilla.org/2006/browser/search/">
-<ShortName>{{ _('Mozilla Developer Network') }}</ShortName>
-{# L10n: Search Mozilla Developer Network documentation #}
-<Description>{{ _('Search Mozilla Developer Network documentation') }}</Description>
+<ShortName>{{ _('MDN Web Docs') }}</ShortName>
+{# L10n: Search MDN Web Docs documentation #}
+<Description>{{ _('Search MDN Web Docs documentation') }}</Description>
```
jwhitlock: "documentation" is redundant, should be ``Search MDN Web Docs``

schalkneethling:
@jwhitlock updated

PR #4640, opened 2018-01-29
Adds shorter class for interactive examples

schalkneethling:


kuma/static/styles/components/wiki/interactive.scss
```diff
@@ -13,7 +13,11 @@
 }
 
 .interactive.taller {
-    height: 690px;
+    height: 692px;
```
schalkneethling: I am also bumping this one an additional 2px to avoid a vertical scrollbar

PR #4641, opened 2018-01-29
bug 1432370 - Remove 2015 MDN Fellowship call for applications 

jwhitlock:
The [2015 MDN Fellowship Program](https://developer.mozilla.org/en-US/fellowship) page is a call for applications, added in [bug 1126583](https://bugzilla.mozilla.org/show_bug.cgi?id=1126583). It was not updated with the results of the program, but instead updates were posted to blog.mozilla.org.  The page remains, and includes the old "Mozilla Developer Network" branding.

I've moved the contents to https://developer.mozilla.org/en-US/docs/Archive/2015_MDN_Fellowship_Program, along with some of the follow-in information I found by searching Google.  This PR removes the page and associated assets, and redirects the URL to the archive.


kuma/landing/jinja2/landing/fellowship.html
```diff
@@ -1,197 +0,0 @@
-{% extends "wiki/base.html" %}
```
schalkneethling: There are still references to this in the `.po` files. I am guessing this will be removed as part of future localisation updates?

PR #4644, opened 2018-01-30
bug 1434296: Use persistent DB connections

jwhitlock:
Use persistent DB connections by setting ``CONN_MAX_AGE``, defaulting to 60 seconds.

AWS RDS [limits the maximum connections](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ConnectToInstance.html) as a factor of the instance size. Our current instance allows 1320 simultaneous connections. My estimate is that we use 1010 connections, and would be below this limit:

|    Deployment | Pod Count | Per Pod | Total |
| ------------- | --------- | ------- | ----- |
| api           |         2 |       4 |     8 |
| celery-beat   |         1 |       1 |     1 |
| celery-cam    |         1 |       1 |     1 |
| celery-worker |        10 |       4 |    40 |
| kumascript    |         6 |       0 |     0 |
| web           |        20 |       8 |   960 |
| **Total**  |  |  |  1010 |

Another limiting factor is that MySQL will close idle connections, set by [wait_timeout](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_wait_timeout). It is currently set to ``28800``, or 8 hours. The ``CONN_MAX_AGE`` should be below this number, and we may want to lower MySQL's ``wait_timeout`` so that idle connections are released sooner. I believe this was much lower in SCL3, based on experience with interactive sessions, which made the default ``CONN_MAX_AGE`` a better choice for that environment.

@jgmize and @metadave may have feedback from an SRE perspective on this change.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -57,12 +57,28 @@ ALLOW_ROBOTS = config('ALLOW_ROBOTS', default=False, cast=bool)
 
 MANAGERS = ADMINS
 
+
+# CONN_MAX_AGE: 'persistent' to keep open connection, or max requests before
+# releasing. Default is 0 for a new connection per request.
+def parse_conn_max_age(value):
+    try:
+        return int(value)
+    except ValueError:
+        assert value.lower() == 'persistent', 'Must be int or "persistent"'
+        return None
+
+
+CONN_MAX_AGE = config('CONN_MAX_AGE', default=60,
+                      cast=parse_conn_max_age)
 DEFAULT_DATABASE = config('DATABASE_URL',
                           default='mysql://kuma:kuma@localhost:3306/kuma',
                           cast=dj_database_url.parse)
+
+
 if 'mysql' in DEFAULT_DATABASE['ENGINE']:
     # These are the production settings for OPTIONS.
     DEFAULT_DATABASE.update({
+        'CONN_MAX_AGE': CONN_MAX_AGE,
         'OPTIONS': {
             'charset': 'utf8',
             'use_unicode': True,


kuma/settings/common.py
```diff
@@ -57,12 +57,28 @@ def __call__(self, value):
...
+# releasing. Default is 0 for a new connection per request.
+def parse_conn_max_age(value):
+    try:
+        return int(value)
+    except ValueError:
+        assert value.lower() == 'persistent', 'Must be int or "persistent"'
+        return None
+
+
+CONN_MAX_AGE = config('CONN_MAX_AGE', default=60,
```
escattone: I'm fine with starting at 60 seconds for the default, but I lean towards going even higher (like maybe something between 5 and 15 minutes?), just so we can reduce the overhead of establishing new connections as much as possible.

I suspect that if we go beyond a certain value, we may have to contend with firewalls. In my past experience at two different employers, we had persistent DB connections that connected through a firewall, and discovered that the firewall silently dropped idle connections (wouldn't inform either end) after some period of time (it was something between 30 minutes to an hour if I remember correctly, and of course is configurable per firewall). We got around the issue by pinging the connection periodically to keep it alive.

metadave:
+1, we can try lowering `wait_timeout` for the [mdn-stage-params RDS paramter group](https://us-west-2.console.aws.amazon.com/rds/home?region=us-west-2#parameter-groups-detail:ids=mdn-stage-params;type=instance) before changing on prod.

kuma/settings/common.py
jwhitlock: We can play with the settings in prod, and adjust the default as needed. I'd like to only have a minute to shoot ourselves in the foot if, for example, a web pod doesn't properly close a the connection as a new deployment is being rolled out.  We're a little close to the max connections for my comfort...

jwhitlock:
I'd like to ship this when we're around to monitor, and then tweak ``CONN_MAX_AGE`` and ``wait_timeout`` if we have problems, especially around rolling deployments. We can experiment with a per-connection ``wait_timeout`` as well (in a new code push).

PR #4647, opened 2018-01-31
bug 1431820: use Django middleware for handling ETag/Last-Modified

escattone:
This PR does the following:

- Sets the `USE_ETAGS` setting to `True`. This enables the `ETag` functionality within Django's `CommonMiddleware`, which includes both computing/adding the `ETag` header to **all** responses (even responses to non-`GET` requests) as well as handling conditional GET requests made using the `IF-NONE-MATCH` header (the `IF-UNMODIFIED-SINCE` header is not handled here). The fact that the `ETag` header is added to **all** responses is a shortcoming of Django 1.8. Another shortcoming is that 304 ("Not Modified") responses due to an `ETag` match clear most of the headers, so headers like `Cache-Control`, `ETag`, `Last-Modified`, and `Vary` are lost. I don't think this will cause any issues, but I'm not certain. **When we move to Django 1.11, the `USE_ETAGS` setting should be deleted, and this deprecated functionality within `CommonMiddleware` will no longer be used.**

- Adds Django's `ConditionalGetMiddleware` to the list of `MIDDLEWARE_CLASSES`. In Django 1.8, this adds handling of the `IF-NONE-MATCH` and `IF-UNMODIFIED-SINCE` headers, but does not actually generate and add the `ETag` header to the response (that's done by `CommonMiddleware` as explained above). When we move to Django 1.11, this middleware will assume full responsibility for generating/handling the `ETag`, and will be free of the shortcomings mentioned in the first point above (properly preserving some headers for 304 responses, and only generating/handling `ETag` headers for GET requests).

- Removes the few `last_modified` and `etag` view decorators that were used. The `etag` decorator is fully replaced by the middleware above (but of course still valid and available for those endpoints that have a faster way to generate the `ETag` value), while the `ConditionalGetMiddleware` only replaces the conditional handling portion of the `last_modified` decorator (the `Last-Modified` header must still be generated and added to the response by an endpoint).

- Removes header checks in tests for 304 responses due to an `ETag` match (see first point above).

### DIFF
--- a/kuma/attachments/tests/test_views.py
+++ b/kuma/attachments/tests/test_views.py
@@ -273,4 +273,3 @@ def test_raw_file_if_modified_since(client, settings, file_attachment):
     assert 'Cache-Control' in response
     assert 'public' in response['Cache-Control']
     assert 'max-age=900' in response['Cache-Control']
-    assert 'Vary' not in response
--- a/kuma/attachments/views.py
+++ b/kuma/attachments/views.py
@@ -4,7 +4,6 @@ from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.http import Http404, StreamingHttpResponse
 from django.shortcuts import get_object_or_404, render, redirect
-from django.views.decorators.http import last_modified
 from django.views.decorators.cache import cache_control
 from django.views.decorators.clickjacking import xframe_options_sameorigin
 
@@ -15,7 +14,7 @@ from kuma.wiki.decorators import process_document_path
 
 from .forms import AttachmentRevisionForm
 from .models import Attachment
-from .utils import allow_add_attachment_by, convert_to_utc
+from .utils import allow_add_attachment_by, convert_to_http_date
 
 
 # Mime types used on MDN
@@ -43,11 +42,7 @@ def raw_file(request, attachment_id, filename):
     if is_untrusted(request):
         rev = attachment.current_revision
 
-        def get_last_modified(*args):
-            return convert_to_utc(rev.created)
-
         @cache_control(public=True, max_age=60 * 15)
-        @last_modified(get_last_modified)
         def stream_raw_file(*args):
             if settings.DEBUG:
                 # to work around an issue of the localdevstorage with streamed
@@ -59,6 +54,7 @@ def raw_file(request, attachment_id, filename):
                 response['Content-Length'] = rev.file.size
             except OSError:
                 pass
+            response['Last-Modified'] = convert_to_http_date(rev.created)
             response['X-Frame-Options'] = 'ALLOW-FROM %s' % settings.DOMAIN
             return response
 
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -5,11 +5,8 @@ from urlparse import urljoin
 import django.middleware.gzip
 from django.conf import settings
 from django.core import urlresolvers
-from django.http import (
-    HttpResponseRedirect,
-    HttpResponseForbidden,
-    HttpResponsePermanentRedirect,
-)
+from django.http import (HttpResponseRedirect, HttpResponseForbidden,
+                         HttpResponsePermanentRedirect)
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
 from django.contrib.sessions.middleware import SessionMiddleware
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -2,9 +2,14 @@ import contextlib
 import urllib
 from urlparse import urljoin
 
+import django.middleware.gzip
 from django.conf import settings
 from django.core import urlresolvers
-from django.http import HttpResponseForbidden, HttpResponseRedirect, HttpResponsePermanentRedirect
+from django.http import (
+    HttpResponseRedirect,
+    HttpResponseForbidden,
+    HttpResponsePermanentRedirect,
+)
 from django.utils import translation
 from django.utils.encoding import iri_to_uri, smart_str
 from django.contrib.sessions.middleware import SessionMiddleware
@@ -205,3 +210,23 @@ class LegacyDomainRedirectsMiddleware(object):
                 urljoin(settings.SITE_URL, request.get_full_path())
             )
         return None
+
+
+class GZipMiddleware(django.middleware.gzip.GZipMiddleware):
+    """
+    This is identical to Django's GZipMiddleware, except that it will not
+    modify the ETag header.
+
+    TODO: When moving to Django 1.11, this code and its tests can be deleted,
+          and django.middleware.gzip.GZipMiddleware should be used instead.
+    """
+
+    def process_response(self, request, response):
+        original_etag = response.get('etag')
+        response_out = super(GZipMiddleware, self).process_response(
+            request,
+            response
+        )
+        if (original_etag is not None) and response_out.has_header('etag'):
+            response_out['etag'] = original_etag
+        return response_out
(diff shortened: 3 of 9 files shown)


kuma/attachments/tests/test_views.py
```diff
@@ -273,4 +273,3 @@ def test_raw_file_if_modified_since(client, settings, file_attachment):
     assert 'Cache-Control' in response
     assert 'public' in response['Cache-Control']
     assert 'max-age=900' in response['Cache-Control']
-    assert 'Vary' not in response
```
escattone: I removed this because `Vary` header was now in the response, but with a value of `Accept-Encoding`.

escattone: I found that the `Vary` header was now in the response with a value of `Accept-Encoding`. I'm not sure if it's worth checking any of the headers when the response is a 304.

kuma/settings/common.py
```diff
@@ -388,6 +388,18 @@ def _get_locales():
     path('locale'),
 )
 
+# When true, Django's CommonMiddleware will compute and add an ETag header
+# to ALL responses, as well as handle conditional GET requests but based soley
+# on the ETag header (it won't handle conditional GET requests based on the
+# Last-Modified header). Django's ConditionalGetMiddleware, uses both the ETag
+# and Last-Modified headers to handle conditional GET requests.
+#
+# IMPORTANT NOTE: When we move to Django 1.11, the USE_ETAGS setting is no
```
safwanrahman: This thing can be in `TODO:`

escattone:
I feel like I stepped into a Django 1.8 minefield with this PR, but I think I've survived.

The second commit of this PR does several things:

- It fixes an issue in Django 1.8 with the way the `GZipMiddleware` and `CommonMiddleware`/`ConditionalGetMiddleware` fail to play together nicely when using the `ETag` header for conditional requests. Django's `GZipMiddleware` actually mangles the `ETag` header (inserts `;gzip` into the quoted `ETag` value) when compressing, effectively eliminating the possibility of 304 or 412 responses for subsequent conditional requests.
- It improves all of the existing `ETag`-based conditional tests by checking that the `ETag` header generated from a gzipped request (the normal case) will return a 304 or 412 for subsequent conditional `IF_NONE_MATCH` or `IF_MATCH` requests. All of these tests will now fail if the `GZipMiddleware` and `CommonMiddleware`/`ConditionalGetMiddleware` fail to play nicely together.
- It paves the way for a smooth transition to Django 1.11. When moving to Django 1.11, do the following:
    - Delete the `USE_ETAGS` setting.
    - Delete the `kuma.core.middleware.GZipMiddleware` class and its associated tests.
    - Replace `kuma.core.middleware.GZipMiddleware` with Django's `GZipMiddleware` in `settings.MIDDLEWARE_CLASSES` assuming `MIDDLEWARE_CLASSES` continues to be used, or whatever the equivalent middleware function is if `settings.MIDDLEWARE` takes its place.
    - That's it. All of the tests have been written such that they will work with Django 1.11 as well.

kuma/attachments/views.py
```diff
@@ -59,6 +54,7 @@ def stream_raw_file(*args):
                 response['Content-Length'] = rev.file.size
             except OSError:
                 pass
+            response['Last-Modified'] = convert_to_http_date(rev.created)
```
jwhitlock: 👍 conversion to ``convert_to_http_date``. I'm guessing that the ``ETag`` skips streaming responses, so this is still needed.

``convert_to_utc`` can't handle ambiguous dates:

```python
import datetime
from kuma.attachments.utils import convert_to_utc
convert_to_utc(datetime.datetime(2017, 11, 5, 1, 8, 42))
```

That feels like follow-on work on [bug 1173189](https://bugzilla.mozilla.org/show_bug.cgi?id=1173189), and moving date manipulations into ``kuma/core/utils.py``.

kuma/wiki/tests/test_views_code.py
```diff
@@ -93,14 +102,6 @@ def test_code_sample(code_sample_doc, constance_config, client, settings):
         HTTP_IF_NONE_MATCH=current_etag
     )
     assert response.status_code == 304
-    assert response['Access-Control-Allow-Origin'] == '*'
```
jwhitlock: TODO: test ``Access-Control-Allow-Origin`` in a headless test (different PR)

kuma/wiki/tests/test_views_document.py
```diff
@@ -63,6 +64,12 @@ def get_content(content_case, data):
 
 @pytest.fixture
 def section_doc(root_doc, wiki_user):
+    """
+    The content in this document's current revision contains multiple HTML
+    elements with an "id" attribute (or "sections"), and also has a length
+    greater than or equal to 200, which meets the compression threshold of
+    the GZipMiddleware.
```
jwhitlock: Is the 200-character minimum why ``SECTION5`` was added?

kuma/wiki/views/code.py
```diff
@@ -35,17 +35,7 @@ def code_sample(request, document_slug, document_locale, sample_name):
...
-    def get_etag(*args):
-        return '{}'.format(hashlib.md5(response.content).hexdigest())
-
-    @cache_control(public=True, max_age=60 * 60 * 24)
-    @etag(get_etag)
-    def render_code_sample(request, *args):
-        return response
-
-    return render_code_sample(request)
+    return render(request, 'wiki/code_sample.html', data)
```
jwhitlock: 😍 so much cleaner

kuma/attachments/views.py
escattone: Good point about streaming responses (we could have removed this code that sets the `Last-Modified` header if it wasn't a streaming response), and good catch about `content_to_utc` not handling ambiguous dates! I would never have thought of that until it failed. I guess the solution is to follow what you did for the publication date for `DocumentsFeed`, and assume DST if the timezone is naive? I'll do that in a follow-on PR.

kuma/core/middleware.py
```diff
@@ -2,9 +2,14 @@
...
 
+import django.middleware.gzip
 from django.conf import settings
 from django.core import urlresolvers
-from django.http import HttpResponseForbidden, HttpResponseRedirect, HttpResponsePermanentRedirect
+from django.http import (
+    HttpResponseRedirect,
+    HttpResponseForbidden,
+    HttpResponsePermanentRedirect,
+)
```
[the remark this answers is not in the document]
escattone: I've had always used hanging indents until my last job, where we started using aligned indents because they enabled future edits to only affect one line (but I don't always like the look of it). I'm not consistent though, since I personally didn't do it long enough to make it habit. With MDN I'm happy to settle on either style, and if you prefer the hanging indent style, let's go with that. It's a greater part of my muscle and visual memory anyway!

kuma/wiki/tests/test_views_document.py
escattone: Yes, I added `SECTION4` solely for that reason.

kuma/attachments/views.py
jwhitlock: It was on my mind because of PR #4621. I'd use the same strategy of assuming DST and being half-wrong rather than raise an ISE.

The function may be slightly different because that one converts to server timezone, while this one wants UTC. Maybe! I'd have to look at ``get_current_timezone`` vs ``get_default_timezone``.

Getting to ``USE_TZ=1`` is on my list, and I think centralizing the timezone stuff in ``kuma.core.utils`` would be a good step. This stuff may be a shared function, or two close functions in the same file

PR #4649, opened 2018-02-02
bug 801623, 1316610, 1431511: Update indexing and robots.txt

jwhitlock:
This PR lumps together a bunch of stuff, that improves guidance for search engine bots and other scrapers, sharpens the distinctions between the main website at https://developer.mozilla.org and alternate domains and deployments, and moves more to headless functional testing.

It's bigger than I'd like, but I'm limited on time for breaking this into smaller PRs, so I'm making it harder on the reviewers 😈 

Here's the changes:

* Main website at https://developer.mozilla.org:
  * ``robots.txt`` is served from a string rather than a file. The file was required for dual-serving with Apache in SCL3, and is no longer needed.
  * The ``robots.txt`` variant is now set based on the request host name and the settings ``ALLOW_ROBOTS_WEB_DOMAINS`` and ``ALLOW_ROBOTS_DOMAINS``, rather than ``ALLOW_ROBOTS``.  This allows multiple domains to serve production data from the same web deployment, but marks them so they won't be indexed.  They can be overridden in the environment, which is mostly for development testing, but could be used in the future.
  * ``robots.txt`` drops ``Crawl-delay: 5`` and ``Request-rate: 1/5``. Neither of these are used by Googlebot, might be used by Bing and others, but are well below the rate at which our content could be accessed.
  * Use ``<meta name="robots" content="noindex, nofollow">`` when displaying a "needs translation" page, where the content is English but there is a call-to-action to translate.
  * Use ``<meta name="robots" content="noindex, nofollow">`` when the HTML is unavailable, such as a partially created document.
   * The login link has ``rel=nofollow`` ([bug 1431511](https://bugzilla.mozilla.org/show_bug.cgi?id=1431511)), discouraging scraping a robots-blocked link
* Main Static Media CDN ``https://cdn.mdn.mozilla.net``, which uses ``https://developer.mozilla.org`` as the origin server:
   * For ``<link rel="alternate"...`` tags in the ``<head>`` section, use absolute URLs. The CDN serves wiki pages like https://cdn.mdn.mozilla.net/en-US/docs/Web/HTML, and includes the alternate links to translations like ``<link rel="alternate" hreflang="fr" href="/fr/docs/Web/HTML" title="HTML (HyperText Markup Language)">``. The relative link means that the CDN URL is noted as the alternate URL.  This changes to ``href="https://developer.mozilla.org/fr/docs/Web/HTML"``, signalling that the CDN variant URL should not be used.
   * Note: when https://developer.mozilla.org is served by CloudFront, this domain will go away and assets will be served from the main website. 
* Main "untrusted" domain for user uploads and live samples at https://mdn.mozillademos.org:
   * ``robots.txt`` changes from a "disallow indexing" to allow indexing of content with a blank file. In SCL3, Google and other scrapers were allowed to include images and live samples in the scraped content, but were denied in AWS. This returns to allowing indexing and scraping.
   * ``robots.txt`` is served from a string rather than a file as well.
* Alternate domain https://prod.mdn.moz.works:
  * ``robots.txt`` now forbids indexing on this domain.
* **TODO**: Allow indexing on the origin domain for https://developer.mozilla.org.

The testing for these new domain-specific changes are in the headless functional tests. The tests call the ``/_kuma_status.json`` endpoint during configuration (after parsing the command line, before collecting tests), and can be used to construct fixtures and for test selection. I haven't done all I want to do with this, but it is used to collect the domain names for a deployment, so that all their  ``robots.txt`` can be fetched and examined.


kuma/landing/views.py
```diff
@@ -51,17 +51,82 @@ def promote_buttons(request):
     return render(request, 'landing/promote_buttons.html')
 
 
+ROBOTS_ALLOWED_TXT = '''\
```
safwanrahman: I think it can be in a jinja template so that it can be generated in future

jwhitlock: What's an example of something that would be templatized?

metadave:
(apologies for the partially out-of-context comment)

> Main CDN https://cdn.mdn.mozilla.net

Can we call `cdn.mdn.mozilla.net` the "Static Media CDN", and the upcoming CDN the "Primary CDN"?

https://github.com/mozmeao/infra/pull/726/files#diff-a219fe3207264569b4a0e326d89896b3R25

kuma/landing/views.py
safwanrahman: Like the `Sitemap`, it should be dependent upon host
`Sitemap: https://developer.mozilla.org/sitemap.xml` should be `settings.HOST + /sitemap.xml`

kuma/settings/common.py
```diff
@@ -1226,6 +1224,20 @@ def pipeline_one_scss(slug, **kwargs):
...
+# Allow robots, but restrict some paths
+# If the domain is a CDN, the CDN origin should be included.
+ALLOW_ROBOTS_WEB_DOMAINS = set(
+    config('ALLOW_ROBOTS_WEB_DOMAINS',
+           default='developer.mozilla.org',
+           cast=Csv()))
+
+# Allow robots, no path restrictions
+# If the domain is a CDN, the CDN origin should be included.
+ALLOW_ROBOTS_DOMAINS = set(
```
safwanrahman: I think this name should be `ALLOW_ROBOT_MEDIA_DOMAINS`

kuma/landing/views.py
jwhitlock: It's an error if we're telling search engines that the sitemap for indexed content is anywhere other than https://developer.mozilla.org/sitemap.xml. It seems like we're asking for trouble if we allow it to be configured by settings.

kuma/settings/common.py
jwhitlock: I've been thinking about ``ALLOW_ROBOT_MEDIA_DOMAINS``. ``MEDIA`` suggests Django's media upload directory to me, but this is the domain for:

* attachments - Files uploaded by users, close to the Django media upload. Eventually I want to move these to their own domain backed by S3 or a new file upload service, and get Django out of the loop.
* code samples - These are extracted from the content, repackaged as HTML, and served in an ``<iframe>``. Again, the implementation may change in the future, and maybe even Django won't be involved in serving the content.

Together, we call these the samples domain, or the attachments domain, or the untrusted domain. The fact that we can't settle on a name is a good indication that the design is wrong.

So, I'm not going to further confuse naming by adding ``MEDIA`` in there. It's ``ALLOW_ROBOTS_DOMAINS``, where robots are allowed without restrictions.

jwhitlock: I think the defaults should be the config defaults for these:

```
- ALLOW_ROBOTS_DOMAINS=${ALLOW_ROBOTS_DOMAINS:-mdn.mozillademos.org,mdn-demos-origin.moz.works}
- ALLOW_ROBOTS_WEB_DOMAINS=${ALLOW_ROBOTS_WEB_DOMAINS:-developer.mozilla.org}
```

That way a user can override it if they are working on ``robots.txt``, but leave it in "no indexing please" mode if they aren't. This makes be think that all the environment overrides should be in ``docker-compose.yml``, with notes for the non-default ones. Which makes me think our settings need fixing. Which makes me lazy and want to do it in a new PR.

PR #4650, opened 2018-02-05
[Bug 1434690] Use different favicon on staging and development

safwanrahman:
It seems like the most efficiant fix with current configuration.
@jwhitlock small PR, r?

### DIFF
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -60,12 +60,12 @@
   <link rel="apple-touch-icon-precomposed" href="{{ static('img/favicon57.png') }}">
   <!-- basic favicon -->
      <!-- basic favicon -->
-  {% if settings.DOMAIN == settings.STAGING_DOMAIN %}
-    {% set favicon_suffix = '-staging' %}
-  {% elif settings.DOMAIN == 'localhost' %}
-    {% set favicon_suffix = '-local' %}
-  {% else %}
+  {% if settings.DOMAIN == settings.PRODUCTION_DOMAIN %}
     {% set favicon_suffix = '' %}
+  {% elif settings.DOMAIN == settings.STAGING_DOMAIN %}
+    {% set favicon_suffix = '-staging' %}
+  {% else %}
+    {% set favicon_suffix = '-local' %}
   {% endif %}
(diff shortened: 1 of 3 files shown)


jinja2/base.html
```diff
@@ -59,7 +59,15 @@
...
   <link rel="apple-touch-icon-precomposed" href="{{ static('img/favicon57.png') }}">
   <!-- basic favicon -->
-  <link rel="shortcut icon" href="{{ static('img/favicon32.png') }}">
+     <!-- basic favicon -->
+  {% if settings.DOMAIN == settings.STAGING_DOMAIN %}
+    {% set favicon_suffix = '-staging' %}
+  {% elif settings.DOMAIN == 'localhost' %}
+    {% set favicon_suffix = '-local' %}
+  {% else %}
+    {% set favicon_suffix = '' %}
```
jwhitlock: The specific cases are production and staging, the fallback should be the `-local` - see the previous version:

https://github.com/mozilla/kuma/blob/14521cb18ad486575b8aeb1459ecb03a3ff6e9dc/jinja2/base.html#L64-L71

safwanrahman: @jwhitlock In current configuration, there are nothing specific for production. But there are specific case for `dev` and `stage`. So I thought to have specific case for `dev` and `stage` and fallback to `production`

jwhitlock: The point of the different favicon.ico is to have one that only appears in production, one that only appears in staging, and everywhere else gets the dev icon. It's not OK for the production icon to be the default.

``settings.PRODUCTION_DOMAIN`` should probably be added, and ``PRODUCTION_URL`` should be built from that rather than ``PROTOCOL`` and ``DOMAIN``. Research is needed to see why it was done this way.

safwanrahman: Agreed! 👍

safwanrahman:
@jwhitlock I have fixed the thing. It should be exclusive to staging and production, and inclusive for all other instance.
the `PRODUCTION_URL` was used for serving the local media in local environment, but I think the naming convention should not be `PRODUCTION_URL`, I have changed the use case with `SITE_URL` as both are same, and added `PRODCUTION_DOMAIN`
r?

PR #4656, opened 2018-02-08
Fix bug 1434558, swap search and toolbox components

schalkneethling:
This is to swap the toolbox and search components in the site header. This also required a number of changes to the header overall.

So far I have tested this in:

* [x] Chrome
* [x] Firefox
* [x] Safari

Using both

* [x] en-US
* [x] fa

To-Do
=====

Test in:

* [x] IE 11
* [x] Edge
* [x] IE 10 (as a basic sanity check)

Code is good to review. Thanks!


jwhitlock:
Thanks @schalkneethling. This looks like a good refactoring of the base template, although I'll need to look closer to see exactly what moved where. Will Feb 13 work for a detailed review, or do you need it sooner?

schalkneethling:
@jwhitlock That works for me. Thanks!

kuma/static/js/components/nav-main-search.js
```diff
@@ -0,0 +1,66 @@
+(function() {
+    'use strict';
+
+    var collapseSearchButton = document.getElementById('close-header-search');
+    var header = document.getElementById('main-header');
+    var isExpanded = false;
+    var navMainSearchForm = document.getElementById('nav-main-search');
+    var searchInput = document.getElementById('main-q');
+    var searchTrigger = navMainSearchForm.querySelector('.search-trigger');
+    var searchWrap = navMainSearchForm.querySelector('.search-wrap');
```
stephaniehobson: Nice modern JS.

kuma/static/styles/components/structure.scss
```diff
@@ -23,18 +23,24 @@ Pre-header
 Header
 ====================================================================== */
 
-@import 'structure/main-header';
-@import 'structure/submenu';
+@import 'structure/page-header/main-header';
+@import 'structure/page-header/submenu';
```
stephaniehobson: Good structural changes :)

stephaniehobson: If you want to preserve the git history for these moved pages you can move them in a separate PR where you don't edit the files. This might be complicated rebase dance so it's not a blocker.

kuma/static/styles/components/structure/page-header/nav-main-search.scss
```diff
@@ -0,0 +1,142 @@
...
+        position: relative;
+        top: 0;
+        @include bidi((
+            (right, auto, left, auto),
+        ));
+        vertical-align: top;
+    }
+
+    input[type='search'] {
+        color: #333;
```
stephaniehobson: `$text-color`?

kuma/static/styles/components/structure/page-header/nav-main-search.scss
```diff
@@ -0,0 +1,142 @@
...
+
+    @media #{$mq-small-desktop-and-up} {
+        position: absolute;
+        top: 5px;
+        @include bidi((
+            (left, 30%, 30%),
+        ));
+        @include bidi((
+            (right, auto, auto),
+        ));
```
stephaniehobson: These declarations should be inside the same mixin:
```
@include bidi((
    (left, 30%, 30%),
    (right, auto, auto),
));
```

kuma/static/styles/components/structure/page-header/submenu.scss
```diff
@@ -3,12 +3,17 @@
...
     padding: 15px;
     border-top: 5px solid $grey-light;
     box-shadow: 0 4px 4px rgba(0, 0, 0, .15);
     background-color: #fff;
+    @include bidi((
+        (right, 0, left, 0),
+    ));
+    @include bidi((
+        (text-align, left, right),
+    ));
```
stephaniehobson: These should be combined:

```
@include bidi((
   (right, 0, left, 0),
   (text-align, left, right),
));
```

kuma/static/styles/components/structure/page-header/submenu.scss
```diff
@@ -3,12 +3,17 @@
...
     z-index: $submenu-z;
     top: 100%;
-    @include bidi(((left, 0, right, auto),));
     width: 250px; // fallback, IE & Edge
     padding: 15px;
     border-top: 5px solid $grey-light;
     box-shadow: 0 4px 4px rgba(0, 0, 0, .15);
     background-color: #fff;
+    @include bidi((
+        (right, 0, left, 0),
```
stephaniehobson: Right needs to be set to auto when left is set.

```
(right, 0, left, auto),
```

kuma/static/styles/components/structure/page-header/toolbox.scss
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @media #{$mq-small-desktop-and-up} {
+        text-align: right;
+
+        > ul > li {
+            @include large-inline-nav-item();
+
+            > a {
+                color: #333;
```
stephaniehobson: `$text-color`?

kuma/static/styles/components/structure/page-header/toolbox.scss
```diff
@@ -0,0 +1,44 @@
...
+
+                &:hover,
+                &:focus {
+                    color: #333;
```
stephaniehobson: `$text-color`?

stephaniehobson:
This should also be tested with a touch screen, or at least a touch screen emulator :)

schalkneethling:
Thank you for the review @stephaniehobson - I still need to update the PR but, I have addressed most of the items here except:

* [ ] The entire black square around the search icon needs to be clickable. And you can fix Bug 1384653!
* [ ] "Skip to search" in the a11y nav menu should trigger the animation and put focus on the search field.
* [ ] I can't tab to the search field anymore.

I will tick these of as I address them, and update the PR.

schalkneethling:
IE 11 is turning out to be a heck of a thing to test. Seems like VirtualBox does not play nice with High Sierra :-/ From the bit I could test so far, it seems these media queries[1] are not being applied in IE11

[1] https://github.com/mozilla/kuma/pull/4656/files#diff-db0e18eb73f5b50ef87c887f6567253fR137

schalkneethling:
Ok, I know what the problem is with IE, will update the PR momentarily.

jwhitlock:
@stephaniehobson I can't reproduce the issue with locales, they work for me:

screenshot 2018-02-12 10 30 46

Some things to try:

* Is the ``locale/`` folder present? A ``git submodule update --init`` would fix it.
* Re-compile locales with ``docker exec -i kuma_web_1 make localecompile``

schalkneethling:
@jwhitlock Fixed the problem you identified. I tried my hand at running the tests locally but, I keep running into walls with various Python and virtualenv error. In the interest of time, I would appreciate it if you could commit the fix for the failing test(s) - Thanks!

jwhitlock:
No problem, happy to do it. 

Next time, you might try:

```
scripts/run_functional_tests.sh tests/functional/test_search.py 
```

This runs the tests inside Docker, avoiding local setup of Python etc.

jwhitlock:
@stephaniehobson, this looks good to me now. Do you want to have another look before merging?

jwhitlock:
Ah, there are some interesting breakpoints in German:

screenshot 2018-02-13 12 10 12

stephaniehobson:
Mobile Safari is putting the rounded corners on the search field 😖 

screen shot 2018-02-13 at 14 16 36

All the interactions work with touch ☝️

schalkneethling:
Ok, both of the items raised by @jwhitlock and @stephaniehobson has been resolved. Thanks a lot for the thorough review o/\o

jwhitlock:
New "middle" layout works for German, Russian, etc. 

screenshot 2018-02-14 09 55 58

PR #4659, opened 2018-02-09
[Bug 1437022] Tags dont get updated after document edit

safwanrahman:
There was some mistry in hashing the key of cache.
While creating the cache, the `pk` was explictly defined. So it was regarded as `kwargs`.
But while invalidating the cache, the `pk` was not explictly defined, so it was regarded as `args`

`django-cacheback` generate the cache key by hashing the parameters. But it hash differently while having `args` and `kwargs`.
As there were different hash while storing and invalidating, the old `tags` were not removed and kept serving.

Moreover, if the document has no tags, it was showing `None`. Fixed that also by excluding the `None`.

And the `refresh_timeout` need to be `0` because a document can be update continuously. So if we dont create task for every revision, updated tags will be missing.

Moreover, added some tests for this regression to not happen in future.

@jwhitlock @escattone r?

### DIFF
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -137,9 +137,7 @@ class DocumentTagsJob(KumaJob):
     Longer lifetime as tags are rarely modified
     """
 
-    # A wiki document can be updated with a short interval
-    # So task for refreshing tags should be created always
-    refresh_timeout = 0
+    refresh_timeout = 180
 
     @property
     def lifetime(self):
--- a/kuma/wiki/jobs.py
+++ b/kuma/wiki/jobs.py
@@ -136,7 +136,10 @@ class DocumentTagsJob(KumaJob):
     We invalidate this when a document is saved only.
     Longer lifetime as tags are rarely modified
     """
-    refresh_timeout = 180
+
+    # A wiki document can be updated with a short interval
+    # So task for refreshing tags should be created always
+    refresh_timeout = 0
 
     @property
     def lifetime(self):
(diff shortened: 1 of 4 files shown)


kuma/wiki/tests/test_signals.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+
+from kuma.wiki.models import Revision, Document
+
+
+@pytest.mark.tags
+def test_document_tags_cache_invalidated_correctly(root_doc, wiki_user):
```
jwhitlock: I think this could live in ``kuma/wiki/tests/test_models_document.py``, but it is OK here as well. However, you need to change the test name to say you are testing the ``on_document_save`` signal handler if you want to keep it here.

kuma/wiki/jobs.py
```diff
@@ -136,7 +136,10 @@ class DocumentTagsJob(KumaJob):
     We invalidate this when a document is saved only.
     Longer lifetime as tags are rarely modified
     """
-    refresh_timeout = 180
+
+    # A wiki document can be updated with a short interval
+    # So task for refreshing tags should be created always
+    refresh_timeout = 0
```
jwhitlock: This is not what ``refresh_timeout`` is for. It is the timeout for how long to wait for an asynchronous refresh before triggering another job (see [docs](http://django-cacheback.readthedocs.io/en/latest/advanced.html?highlight=refresh_timeout).

If you believe that tags should be cached for 0 seconds, then tags are not appropriate as a cacheback job.

safwanrahman:
Fixed the issues.
r? @jwhitlock

PR #4661, opened 2018-02-12
Updating kumascript and locale submodules

escattone:
Updating `kumascript` and `locale` submodules. There were no new localizable strings.


escattone:
😄

PR #4663, opened 2018-02-13
bug 1429933: Update app loading to remove model imports

jwhitlock:
In Django 1.11, models can not be imported inside ``apps.py``. Moved invalidation functions inside ``WikiConfig``, and changed the code to use ``self.get_model()`` instead.

**Update 1:** Importing ``kuma.wiki.jobs`` was also an issue, because it imports a module that imports a module that imports a model. I updated ``kuma.wiki.jobs``. I'll try to run the Django 1.11 tests locally, rather than wait for TravisCI to tell me the next breakage.

**Update 2:** The auto-register method for [django-dbgettext](https://django-dbgettext.readthedocs.io/en/latest/) isn't compatible, so it is migrated to ``kuma.search.apps``. This last change allows app setup to complete, so URLs are the next blocker. This is big enough that I'd like to save it for the next PR.


safwanrahman:
I dont know but It would be better if the codes of signals were written in `signals.py`
and import the signals inside the `ready` method.
Maybe this is beyond this PR, but maybe an improvement can be done.

jwhitlock:
@safwanrahman that may work, I'll take a look

jwhitlock:
This is more complex than I thought, but took me in a good direction.

Django docs recommend that you pass the path to the AppConfig in the ``INSTALLED_APPS`` setting, like ``kuma.wiki.apps.WikiConfig``.  However, this is incompatible with ``django-dbgettext``, which tries to import a module ``dbgettext_registration`` from each application at startup, which fails if the app is the config file.  There's a work-around of defining ``default_app_config``, which we use but attribute to the long-remove jingo package:

https://github.com/mozilla/kuma/blob/e6cd156ca7f3b5c68f1a5ee3dc4a88ef1e981e91/kuma/wiki/__init__.py#L1-L2

This is blocking moving to new-style ``AppConfig``, which we're using in several places. 

[django-dbgettext](https://bitbucket.org/drmeers/django-dbgettext/commits/branch/default) is barely maintained.  There was a release in 2016 for Django 1.9 compatibility, but there's not a lot of action. We barely use it as well. It is used to extract translatable strings for search filters and groups from the database, and populates the [kuma/core/translations](https://github.com/mozilla/kuma/tree/master/kuma/core/translations/search) folder. However, you'd have to run this against the production database, and we don't really have a way of doing that, and creating a new commit. I don't want to go down that path. In fact, I want to go the other way: a set of filters, filter groups, and tags that are defined in the code and translated with gettext.

So, here's my plan, over a number of new PRs:

- Remove dbgettext, convert ``kuma/core/translations``
- Move signal handlers from ``apps.py`` and ``signals.py`` to ``signal_handlers.py``
- Convert to modern ``AppConfig`` configurations

PR #4664, opened 2018-02-14
[Bug 1399639] Updating Django Rest Framework

safwanrahman:
It passes all the tests! But need some more investigation.
I will investigate more later this week.
@jwhitlock Can you take a look if I have taken proper approach?

### DIFF
--- a/kuma/search/renderers.py
+++ b/kuma/search/renderers.py
@@ -9,40 +9,14 @@ class ExtendedTemplateHTMLRenderer(TemplateHTMLRenderer):
     template_name = 'search/results.html'
     exception_template_names = ['search/down.html']
 
-    def resolve_context(self, data, request, response):
-        """
-        Adds some more data to the template context.
-        """
-        data['selected_filters'] = get_filters(request.query_params.getlist)
-        data['search_ref'] = ref_from_request(request)
-        data['index'] = Index.objects.get_current()
-        return super(ExtendedTemplateHTMLRenderer,
-                     self).resolve_context(data, request, response)
-
-    def render(self, data, accepted_media_type=None, renderer_context=None):
-        """
-        Renders data to HTML, using Django's standard template rendering.
-
-        The template name is determined by (in order of preference):
-
-        1. An explicit .template_name set on the response.
-        2. An explicit .template_name set on this class.
-        3. The return result of calling view.get_template_names().
-
-        FIXME: This is a copy of the render method from upstream to support
-        passing the request to the render method of the template backend.
-        This should be removed/fixed when moving to DRF 3.x.
-        """
-        renderer_context = renderer_context or {}
-        view = renderer_context['view']
+    def get_template_context(self, data, renderer_context):
         request = renderer_context['request']
-        response = renderer_context['response']
+        new_data = {
+            'selected_filters': get_filters(request.query_params.getlist),
+            'search_ref': ref_from_request(request),
+            'index': Index.objects.get_current()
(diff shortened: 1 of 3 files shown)


requirements/default.txt
```diff
@@ -113,9 +113,9 @@ django-recaptcha==1.0.5 \
     --hash=sha256:d4d896dd399a1a0810af8e59a646bb5a1586b2cc8b8f32d16766ade5fba79475
 
 # RESTful API framework
-djangorestframework==3.3.2 \
-    --hash=sha256:5634b1ff56581bf0fe4075e86227fc9693c1ca031c7213c9ae942c445c24817b \
-    --hash=sha256:4962418a57804f8323282728a4f9b9496e78caec1adda352170697752eff01bf
+djangorestframework==3.5.4 \
+    --hash=sha256:110afa12784ceadfb50808882689302d266785b51e3d13286744333ff6d78e60 \
+    --hash=sha256:f995a35ae22f354d2a9a42ee6d2c059c101f826b1485ed46781677895ad25ee5
```
jwhitlock: I'm adding more links to the requirements files, see https://github.com/mozilla/kuma/blob/master/requirements/README.rst#requirements-format.  Here's what I have:

```
# RESTful API framework, used by search
# Code: https://github.com/encode/django-rest-framework
# Changes: http://www.django-rest-framework.org/topics/release-notes/
# Docs: http://www.django-rest-framework.org
djangorestframework==3.5.4 \
    --hash=sha256:110afa12784ceadfb50808882689302d266785b51e3d13286744333ff6d78e60 \
    --hash=sha256:f995a35ae22f354d2a9a42ee6d2c059c101f826b1485ed46781677895ad25ee5
```

jwhitlock:
Thanks for tackling this @safwanrahman! I appreciate that you are limiting changes to those needed for the upgrade. This is quick feedback, I'll get into a detailed review later today.

Tests passing is a good sign, although I really don't trust the tests around search 😬 . I'll read the release notes and poke around to see if these look good.

I'm taking an incremental approach with these updates, shipping one or two packages at a time. ``django-tidings`` is the next one (PR #4660), going out today, so I'll wait to merge until that is deployed. [bug 1399639](https://bugzilla.mozilla.org/show_bug.cgi?id=1399639) is the bug to use for package upgrades.

I think DRF 3.6 supports Django 1.8 still, and if so, we should get to that eventually. I'm happy to focus on DRF 3.5, then do DRF 3.6 in the next PR, if there are a lot of changes for DRF 3.6.

requirements/default.txt
jwhitlock: I've tested DRF 3.6.4, and that works with these changes as well. Please switch to that version.

kuma/search/renderers.py
```diff
@@ -9,40 +9,14 @@ class ExtendedTemplateHTMLRenderer(TemplateHTMLRenderer):
...
-        if response.exception:
-            template = self.get_exception_template(response)
-        else:
-            template_names = self.get_template_names(response, view)
-            template = self.resolve_template(template_names)
+        data.update(new_data)
 
-        context = self.resolve_context(data, request, response)
-        return template.render(context, request=request)
+        return super(ExtendedTemplateHTMLRenderer, self).get_template_context(data, renderer_context)
```
jwhitlock: 😄 good job resolving this TODO, the code is much cleaner.

safwanrahman:
@jwhitlock Updated with the comments.

PR #4668, opened 2018-02-15
bug 1406546: Override kubectl for standby cluster

jwhitlock:
The standby cluster runs Kubernetes 1.9, and kubectl 1.6 is having issues talking to it. This uses kubectl 1.8 for pushes to that cluster.

### DIFF
--- a/Jenkinsfiles/push.groovy
+++ b/Jenkinsfiles/push.groovy
@@ -14,12 +14,18 @@ stage("Prepare Infra") {
 }
 
 stage('Push') {
+    def kubectl = "kubectl"
+    if (config.env && config.env.kubectl) {
+        kubectl = config.env.kubectl
+    }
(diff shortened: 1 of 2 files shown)


Jenkinsfiles/push.groovy
```diff
@@ -14,12 +14,18 @@ stage("Prepare Infra") {
 }
 
 stage('Push') {
+    def kubectl = "kubectl"
+    if (config.env && config.env.kubectl) {
+        kubectl = confing.env.kubectl
```
escattone: Typo: `confing` should be `config`

jwhitlock:
Force-push to fix the typo.

Since this didn't fix the issue, there's a good chance that the kubectl version is not the root cause of the issue, and I don't think we should merge this PR that suggests it is an issue.

jwhitlock:
Let's revisit if we actually need a different kubectl version.

PR #4676, opened 2018-02-22
bug 1431259: caching headers/tests for dashboards views

escattone:
This is the first of a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds caching headers and tests for the `kuma.dashboards.urls`.

* add `shared_cache_control` decorator as a thin wrapper around Django's `cache_control` decorator to encapsulate default settings for shared caching
* add tests for `shared_cache_control` decorator
* delete custom `never_cache` decorator, and replace with Django's
* delete tests for custom `never_cache` decorator
* decorate the views for `kuma.dashboards.urls` and add tests

### DIFF
--- a/kuma/core/decorators.py
+++ b/kuma/core/decorators.py
@@ -1,17 +1,39 @@
-import re
 from functools import wraps, partial
+import re
 
 from django.conf import settings
-from django.shortcuts import redirect, render
 from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.contrib.auth.decorators import user_passes_test
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseForbidden, HttpResponseRedirect
-from django.utils.decorators import available_attrs
+from django.shortcuts import redirect, render
 from django.utils.http import urlquote
+from django.utils.decorators import available_attrs
+from django.views.decorators.cache import cache_control
 
 from .jobs import BannedIPsJob
 from .urlresolvers import reverse
+
+
+def shared_cache_control(func=None, **kwargs):
+    """
+    Decorator for view functions that defines the "Cache-Control" header
+    for shared caches like CDN's. It's simply a thin wrapper around
+    Django's cache-control that sets some defaults. By default, it does not
+    permit use of the browser's cache without validation ("max-age=0"), and
+    sets the caching period for shared caches ("s-maxage") based on the
+    CACHE_CONTROL_DEFAULT_SHARED_MAX_AGE setting. All of the defaults can be
+    overridden or extended via keyword arguments.
+    """
+    # Set the default values.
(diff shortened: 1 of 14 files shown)


escattone:
I rebased with fixes for import ordering.

escattone:
I forgot to add in the introduction, that this PR only affects public caching (e.g., CDN's) not browser caching since by default it uses `s-maxage` instead of `max-age`.

kuma/core/tests/test_decorators.py
```diff
@@ -5,13 +5,13 @@
...
-                                  login_required, never_cache,
-                                  skip_in_maintenance_mode,
+from kuma.core.decorators import (block_user_agents, login_required,
+                                  logout_required, permission_required,
                                   redirect_in_maintenance_mode,
-                                  permission_required)
+                                  shared_cache_control,
+                                  skip_in_maintenance_mode)
 
-from kuma.core.tests import KumaTestCase, eq_, ok_
```
jwhitlock: Another ``ok_`` gone! 🎉

kuma/core/decorators.py
```diff
@@ -1,19 +1,41 @@
...
+from django.shortcuts import redirect, render
 from django.utils.http import urlquote
+from django.utils.decorators import available_attrs
+from django.views.decorators.cache import cache_control
 
 from .jobs import BannedIPsJob
 from .urlresolvers import reverse
 
 
+def shared_cache_control(func=None, **kwargs):
```
jwhitlock: Nit: The name "shared cache" confused me a bit. Is the intent is to save a copy in the proxy, but to tell the browser to not cache the results? If so, a better name doesn't come to mind yet...

escattone: Yes, that was the intent. I'm happy to use whatever name you'd prefer, as I had a difficult time coming up with something better.

kuma/core/tests/test_views.py
```diff
@@ -211,4 +211,8 @@ def test_ratelimit_429(client, db):
     assert response.status_code == 429
     assert '429.html' in [t.name for t in response.templates]
     assert response['Retry-After'] == '60'
-    assert response['Cache-Control'] == 'no-cache, no-store, must-revalidate'
+    assert 'Cache-Control' in response
```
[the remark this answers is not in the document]
escattone: Yeah, I did this everywhere thinking it would make for a nicer error message than `KeyError`, but I'm happy to remove it as well.

kuma/core/decorators.py
jwhitlock: ``proxy_cache_yes_browser_cache_no_with_overrides`` is my current best choice, which is so awful. Let's merge with this name and revisit in a few weeks when we're smarter.

PR #4677, opened 2018-02-22
bug 1431259: caching headers/tests for health views

escattone:
This is another in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds caching headers and tests for the `kuma.health.urls`.

### DIFF
--- a/kuma/health/tests/test_views.py
+++ b/kuma/health/tests/test_views.py
@@ -12,35 +12,43 @@ from kuma.users.models import User
 
 
 @pytest.mark.parametrize('http_method', ['put', 'post', 'delete', 'options'])
-@pytest.mark.parametrize('endpoint', ['health.liveness', 'health.readiness'])
+@pytest.mark.parametrize('endpoint', ['liveness', 'readiness', 'status'])
 def test_disallowed_methods(client, http_method, endpoint):
     """Alternate HTTP methods are not allowed."""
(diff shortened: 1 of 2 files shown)


kuma/health/tests/test_views.py
```diff
@@ -12,35 +12,43 @@
...
-@pytest.mark.parametrize('endpoint', ['health.liveness', 'health.readiness'])
+@pytest.mark.parametrize('endpoint', ['liveness', 'readiness', 'status'])
 def test_disallowed_methods(client, http_method, endpoint):
     """Alternate HTTP methods are not allowed."""
-    url = reverse(endpoint)
+    url = reverse('health.{}'.format(endpoint))
     response = getattr(client, http_method)(url)
     assert response.status_code == 405
+    assert 'Cache-Control' in response
+    assert 'max-age=0' in response['Cache-Control']
```
jwhitlock: Nit: If the ``Cache-Control`` header is missing, then checking for ``max-age=0`` will also fail, so checking for ``Cache-Control`` is redundant.

jwhitlock:
This looks good and works well on my development machine, using ``curl -v http://localhost:8000/healthz``, etc.

The same nit on a redundant check for the ``Cache-Control`` header, which is a really weak preference, and could be merged as-is.

PR #4678, opened 2018-02-22
bug 1399639: Update pyquery and constraints

jwhitlock:
There's a simple update to pyquery 1.2.15 and its dependencies, which require no code change:

* pyquery 1.2.11 → 1.2.15: Support :has, fix .val() for textarea and select
* lxml 3.5.0 → 4.1.1: Manual charset detection for HTML5, bug fixes
* cssselect 0.9.1 → 1.0.3: Fix :nth-*(an+b) pseudo selectors

Then, to update to the current [pyquery 1.4](https://github.com/gawel/pyquery/blob/master/CHANGES.rst#140-2018-01-11), things get ugly:

* pyquery 1.2.16 changes input processing to act more like jQuery. This includes returning ``'on'`` when calling ``elem.val()`` for a checkbox, which is what a browser would send if the checkbox were selected. This required updating some test code for the user profile form, which was wrong in other ways as well.
* pyquery 1.3.0 cleaned up some code while shifting Python support (dropping 2.6 and 3.2, and adding 3.6). An error is now raised in certain cases when the content is ``None`` or an empty string. This required updating ``Extractor.css_classnames()`` to exit early in these cases. It is possible this was a source of errors when bulk indexing documents for search. The tests for these functions were also refactored.
* pyquery 1.4.0 says it matches Firefox's behavior for ``.text()``, and it appears the maintainer is adding Selenium tests to verify that pyquery matches browser and jQuery behavior. The main change appears to be that whitespace is normalized to a newline ``\n`` instead of a space. In most cases, the tests are updated to expect the different string, but in one case (``Extractor.code_sample()``) the new parameter ``.text(squash_space=False)`` is used to preserve whitespace.

pyquery is used extensively in tests, including some we haven't touched before, so there was a lot of refactoring:
* Move ``constance_config`` and ``editor_client`` into the global ``conftest.py``.
* Refactor ``@requests_mock.mock()`` decorator into similar ``mock_requests`` pytest fixture, in global ``conftest.py``.
* In ``kuma/attachments/tests/test_templates.py``, refactor ``AttachmentTemplatesTests`` to pytest style.
* In ``kuma/spam/tests/test_forms.py``, refactor ``AkistmetFormTests`` into pytest style, and to avoid using ``Exception.message``, which was deprecated in Python 2.6.
* In ``kuma/wiki/tests/test_content.py``, refactor ``ExtractCodeSampleTests`` and combine tests with existing pytest-style tests


jwhitlock:
I may have some more time to work on this, so I'm going to try to make the code changes as well to get up to the current pyquery 1.4.

PR #4679, opened 2018-02-22
bug 1431259: caching headers/tests for landing views

escattone:
This is another in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds caching headers and tests for the `kuma.landing.urls`.

The `maintenance_mode` endpoint is excluded from caching (it's decorated with `never_cache`). Whenever MDN is taken into or out-of maintenance mode, it seemed best for that endpoint to reflect the current reality immediately.

When #4676 is merged and it is re-based on master, the `not ready` label can be removed.

### DIFF
--- a/kuma/landing/urls.py
+++ b/kuma/landing/urls.py
@@ -1,6 +1,7 @@
 from django.conf.urls import url
 
 from . import views
+from kuma.core.decorators import shared_cache_control
 
 
 urlpatterns = [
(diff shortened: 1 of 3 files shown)


kuma/landing/urls.py
```diff
@@ -23,6 +24,7 @@
         views.robots_txt,
         name='robots_txt'),
     url(r'^favicon.ico$',
-        views.FaviconRedirect.as_view(icon='favicon.ico'),
+        shared_cache_control(
+            views.FaviconRedirect.as_view(icon='favicon.ico')),
```
jwhitlock: Eventually, we'll want to cache endpoints that change rarely differently than endpoints that change often - this redirect would be one of them. It may make sense to have one decorator for rarely changing endpoints, and another for ones that change unpredictably. For now, it is probably OK to use the same cache header.

PR #4680, opened 2018-02-23
bug 1431259: caching headers/tests for search/events views

escattone:
This is another in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds caching headers and tests for `kuma.search.urls` and the `events` endpoint.

None of the search-related endpoints are cached. In particular, I thought the `search` endpoint would have too many possible query-parameter combinations in its URL to provide useful hit ratios. In other words, CDN cache memory would be burned for little to no gain.

The `events` endpoint is cached since Cloudfront CDN's do cache HTTP 3xx responses.

When #4676 is merged and it is re-based on master, the not ready label can be removed.


kuma/events/tests/test_views.py
```diff
@@ -0,0 +1,9 @@
+from kuma.core.urlresolvers import reverse
+
+
+def test_events(client, db):
```
jwhitlock: I don't think we should bother testing this (it basically tests ``shared_cache_control`` and ``RedirectView``, which are tested elsewhere), and I really don't think we should create a new app just so we can.

kuma/search/tests/test_views.py
```diff
@@ -1,32 +1,28 @@
 # -*- coding: utf-8 -*-
 import elasticsearch
+import pytest
 
-from kuma.core.tests import eq_
```
jwhitlock: Another ``eq_`` gone! 🎉

kuma/search/tests/test_views.py
```diff
@@ -1,32 +1,28 @@
...
-        response = self.client.get('/en-US/search?q=test')
-        eq_(response.status_code, 200)
-        self.assertContains(response, 'Results for')
-
-        response = self.client.get('/en-US/search?q=article')
-        eq_(response.status_code, 200)
-        self.assertContains(response, 'an article title')
-
     def test_search_filters(self):
+        """The search view """
```
jwhitlock: 👍 combining two tests that seem to do the same thing
👎 keeping the same silly docstring. It should be removed or be upgraded to be more descriptive.

kuma/search/views.py
```diff
@@ -133,12 +132,13 @@ def get_filters(self, aggregations):
...
-search = ratelimit(key='user_or_ip', rate='25/m', block=True)(
+search = never_cache(ratelimit(key='user_or_ip', rate='25/m', block=True)(
     SearchView.as_view()
-)
+))
 
 
-@cache_page(60 * 15)  # 15 minutes.
+@never_cache
+@cache_page(60 * 15)  # Cache server-side (memcached) for 15 minutes.
```
jwhitlock: I think ``@never_cache`` is wrong here. ``@cache_page`` should be replaced with a more standard cache header.

Also ``suggestions`` is a dumb view and I feel dumber for having read it.

kuma/search/views.py
```diff
@@ -152,7 +152,8 @@ def suggestions(request):
     return HttpResponse(json.dumps(results), content_type=content_type)
 
 
-@cache_page(60 * 60 * 168)  # 1 week.
+@never_cache
+@cache_page(60 * 60 * 168)  # Cache server-side (memcached) for 1 week.
```
jwhitlock: I think ``@never_cache`` is wrong here, and ``@cache_page`` should be replaced with a standard cache control header. The OpenSearchDescription XML never appears to change.

```
<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
                       xmlns:moz="http:/www.mozilla.org/2006/browser/search/">
<ShortName>MDN Web Docs</ShortName>

<Description>Search MDN Web Docs</Description>
<InputEncoding>UTF-8</InputEncoding>
<Image width="16" height="16" type="image/png">/static/img/favicon32.png</Image>
<Url type="text/html" method="get" template="https://localhost:8000/en-US/search">
  <Param name="q" value="{searchTerms}"/>
  <Param name="w" value="3"/>
  <Param name="qs" value="plugin"/>
</Url>
<Url type="application/opensearchdescription+xml"
     rel="self"
     template="https://localhost:8000/en-US/search/xml"/>
<Url type="application/x-suggestions+json"
     template="https://localhost:8000/en-US/search/suggestions?q={searchTerms}"/>
<moz:SearchForm>https://localhost:8000/en-US/search</moz:SearchForm>
```

kuma/search/tests/test_views.py
escattone: Oops, good catch!

escattone:
Thanks @jwhitlock, you emboldened me to completely remove the search suggestions endpoint. By force of habit, I pushed the changes via a re-base instead of a separate commit. Sorry about that.

jwhitlock:
Small PRs make it easier to review the whole thing, even with a rebase. Thanks @escattone!

PR #4682, opened 2018-02-24
bug 951180: Position labels after checkboxes

GloverDonovan:
This pull request fixes [951180](https://bugzilla.mozilla.org/show_bug.cgi?id=951180).

Feel free to tell me if I missed anything. I could only find two checkboxes on the edit page.

No styling changes should be needed. I tested my changes with Developer Tools and both versions behaved the same.

### DIFF
--- a/kuma/users/jinja2/users/user_edit.html
+++ b/kuma/users/jinja2/users/user_edit.html
@@ -51,7 +51,8 @@
       <fieldset class="section notitle" id="personal">
         <ul>
           <li id="field-beta" class="field">
-            <label for="{{ user_form.beta.id_for_label }}">{{ user_form.beta.label }}</label> {{ user_form.beta }}
+            {{ user_form.beta }}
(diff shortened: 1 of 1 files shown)


kuma/users/jinja2/users/user_edit.html
```diff
@@ -51,7 +51,8 @@ <h1 class="user-title">{{ edit_user.username }} <a class="user-title-profile" hr
       <fieldset class="section notitle" id="personal">
         <ul>
           <li id="field-beta" class="field">
-            <label for="{{ user_form.beta.id_for_label }}">{{ user_form.beta.label }}</label> {{ user_form.beta }}
+            {{ user_form.beta }}
```
jwhitlock: I think this is cleaner on its own line - good change!

GloverDonovan: Glad I could help!

PR #4687, opened 2018-03-01
bug 1431259: caching headers/tests for wiki misc/list views

escattone:
This is another in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds caching headers and tests for endpoints for the wiki views in `list.py` and `misc.py`.

### DIFF
--- a/kuma/wiki/tests/conftest.py
+++ b/kuma/wiki/tests/conftest.py
@@ -99,6 +99,20 @@ def trans_edit_revision(trans_doc, edit_revision, wiki_user):
 
 
 @pytest.fixture
+def redirect_doc(wiki_user):
+    """A newly-created top-level English redirect document."""
+    redirect_doc = Document.objects.create(
+        locale='en-US', slug='Redirection', title='Redirect Document')
+    Revision.objects.create(
+        document=redirect_doc,
+        creator=wiki_user,
+        content='REDIRECT <a class="redirect" href="/blah">Some Redirect</a>',
+        title='Redirect Document',
+        created=datetime(2017, 4, 17, 12, 15))
+    return redirect_doc
+
+
+@pytest.fixture
 def doc_hierarchy_with_zones(settings, wiki_user, wiki_user_2, wiki_user_3):
     top_doc = Document.objects.create(
         locale='en-US',
--- a/kuma/wiki/tests/conftest.py
+++ b/kuma/wiki/tests/conftest.py
@@ -8,6 +8,8 @@ from collections import namedtuple
 import pytest
 
 from ..models import Document, DocumentZone, Revision
+from kuma.core.urlresolvers import reverse
+from kuma.wiki.constants import REDIRECT_CONTENT
 
 
 BannedUser = namedtuple('BannedUser', 'user ban')
@@ -99,14 +101,19 @@ def trans_edit_revision(trans_doc, edit_revision, wiki_user):
 
 
 @pytest.fixture
-def redirect_doc(wiki_user):
+def redirect_doc(wiki_user, root_doc):
     """A newly-created top-level English redirect document."""
     redirect_doc = Document.objects.create(
         locale='en-US', slug='Redirection', title='Redirect Document')
     Revision.objects.create(
         document=redirect_doc,
         creator=wiki_user,
-        content='REDIRECT <a class="redirect" href="/blah">Some Redirect</a>',
+        content=REDIRECT_CONTENT % {
+            'href': reverse('wiki.document',
+                            args=(root_doc.slug,),
+                            locale=root_doc.locale),
+            'title': root_doc.title,
+        },
         title='Redirect Document',
         created=datetime(2017, 4, 17, 12, 15))
     return redirect_doc
--- a/kuma/wiki/tests/test_views_list.py
+++ b/kuma/wiki/tests/test_views_list.py
@@ -154,8 +154,10 @@ def test_list_no_redirects(redirect_doc, doc_hierarchy_with_zones, client):
     assert 'public' in resp['Cache-Control']
     assert 's-maxage' in resp['Cache-Control']
     assert 'text/html' in resp['Content-Type']
-    # There should be 4 documents in the English locale.
-    assert len(pq(resp.content).find('.document-list li')) == 4
+    # There should be 4 documents in the 'en-US' locale from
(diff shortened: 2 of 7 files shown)


kuma/wiki/tests/conftest.py
```diff
@@ -98,6 +98,20 @@ def trans_edit_revision(trans_doc, edit_revision, wiki_user):
...
 
+@pytest.fixture
+def redirect_doc(wiki_user):
+    """A newly-created top-level English redirect document."""
+    redirect_doc = Document.objects.create(
+        locale='en-US', slug='Redirection', title='Redirect Document')
+    Revision.objects.create(
+        document=redirect_doc,
+        creator=wiki_user,
+        content='REDIRECT <a class="redirect" href="/blah">Some Redirect</a>',
```
jwhitlock: I think this will be a more interesting fixture for future work if it was a redirect to the ``root_doc``. If we do that, it could throw off the document count for ``test_list_no_redirects``.

kuma/wiki/tests/test_views_list.py
```diff
@@ -121,3 +145,201 @@ def test_revisions_request_invalid_pages(root_doc, client):
...
+    assert 'wiki/list/tags.html' in [t.name for t in resp.templates]
+    assert 'public' in resp['Cache-Control']
+    assert 's-maxage' in resp['Cache-Control']
+
+
+@pytest.mark.tags
+@pytest.mark.parametrize('tag', ['foo', 'bar'])
+@pytest.mark.parametrize('tag_case', ['lower', 'upper'])
+@pytest.mark.parametrize('locale_case', ['root', 'trans'])
+def test_tag_list(root_doc, trans_doc, client, locale_case, tag_case, tag):
```
jwhitlock: I can't tell if this many parameters is clever or offensive. Maybe I'll have it figured out by 2019...

kuma/wiki/tests/test_views_list.py
```diff
@@ -121,3 +145,201 @@ def test_revisions_request_invalid_pages(root_doc, client):
...
+    resp = client.get(url)
+    assert resp.status_code == 200
+    dom = pq(resp.content)
+    assert len(dom('#document-list ul.document-list li')) == 0
+    assert root_doc.slug not in resp.content
+    assert trans_doc.slug not in resp.content
+
+
+@pytest.mark.parametrize('locale', ['en-US', 'de', 'fr'])
+def test_list_with_errors(redirect_doc, doc_hierarchy_with_zones, client,
```
jwhitlock: Good job writing tests for this and other uncovered views!

kuma/wiki/tests/test_views_list.py
```diff
@@ -121,3 +145,201 @@ def test_revisions_request_invalid_pages(root_doc, client):
...
+    assert 'public' in resp['Cache-Control']
+    assert 's-maxage' in resp['Cache-Control']
+    assert 'text/html' in resp['Content-Type']
+    assert len(dom.find('.document-list li')) == len(exp_docs)
+    selector = 'ul.document-list li a[href="/{}/docs/{}"]'
+    for doc in exp_docs:
+        assert len(dom.find(selector.format(doc.locale, doc.slug))) == 1
+
+
+@pytest.mark.parametrize('tag', ['foo', 'bar'])
```
jwhitlock: Nit: I don't think parametrization should be used here. Localization tags are less flexible than they seem from looking at the code. There's exactly one, "inprogress". It's unclear why this was implemented as a tag instead of a boolean on Revisions.

kuma/wiki/tests/test_views_misc.py
```diff
@@ -0,0 +1,79 @@
...
+    url = reverse('wiki.{}'.format(endpoint), locale='en-US')
+    response = getattr(client, http_method)(url)
+    assert response.status_code == 405
+    assert 'Cache-Control' in response
+    assert 'public' in response['Cache-Control']
+    assert 's-maxage' in response['Cache-Control']
+
+
+def test_ckeditor_config(db, client):
+    response = client.get(reverse('wiki.ckeditor_config', locale='en-US'))
```
jwhitlock: and one more now-tested document 🎉

kuma/wiki/tests/test_views_list.py
```diff
@@ -121,3 +145,201 @@ def test_revisions_request_invalid_pages(root_doc, client):
...
+
+
+@pytest.mark.parametrize('locale', ['en-US', 'de', 'fr'])
+def test_list_with_localization_tags(redirect_doc, doc_hierarchy_with_zones,
+                                     client, locale):
+    top_doc = doc_hierarchy_with_zones.top
+    bottom_doc = doc_hierarchy_with_zones.bottom
+    de_doc = top_doc.translated_to('de')
+    for doc in (top_doc, bottom_doc, de_doc, redirect_doc):
+        doc.current_revision.localization_tags.set('foo', 'bar')
```
jwhitlock: Nit: Similar issue - the test code implies that localization tags other that "inprogress" would be used.

kuma/wiki/tests/test_views_list.py
```diff
@@ -121,3 +145,201 @@ def test_revisions_request_invalid_pages(root_doc, client):
...
+
+def test_list_no_redirects(redirect_doc, doc_hierarchy_with_zones, client):
+    url = reverse('wiki.all_documents', locale='en-US')
+    resp = client.get(url)
+    assert resp.status_code == 200
+    assert 'public' in resp['Cache-Control']
+    assert 's-maxage' in resp['Cache-Control']
+    assert 'text/html' in resp['Content-Type']
+    # There should be 4 documents in the English locale.
+    assert len(pq(resp.content).find('.document-list li')) == 4
```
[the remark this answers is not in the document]
escattone: There are two middle docs, middle-top and middle-bottom, so that's why it's 4 docs total in the `en-US` locale.

kuma/wiki/tests/test_views_list.py
escattone: Thanks, that's good to know.

kuma/wiki/tests/conftest.py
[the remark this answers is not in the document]
escattone: Great idea.

escattone:
Thanks for the review @jwhitlock! I addressed all of your nits, and will merge when the tests pass.

PR #4689, opened 2018-03-01
bug 1401246: Standardize signal handlers and apps

jwhitlock:
Django 1.11 fails on application startup if models are imported before apps are initialized and ready. The biggest offenders are signal handlers, which vary in style across Kuma.

To standardize signal handlers, this PR moves them into per-app ``signal_handlers.py``, standardizes imports and usage, and uses the ``@receiver`` decorator to register them. They are imported and registered in the ``AppConfig.ready`` function. This also switches to use the dotted paths to ``AppConfig``s in ``INSTALLED_APPS``, as [recommended by Django](https://docs.djangoproject.com/en/2.0/ref/applications/#for-application-authors).


safwanrahman:
I think its better to have the handling and registration code in `signals` submodule or `signals.py`. Its ususally in this place as per the [doc](https://docs.djangoproject.com/en/2.0/topics/signals/#django.dispatch.receiver)

jwhitlock:
@safwanrahman, I read that too, and would be curious to see a project that implements it that way. Django doesn't define signal handlers in ``signals.py``, just signals. It's also not really consistent on where handlers are defined and used. Don't look at the Django codebase for good signal practice.

I can't define the ``render_done`` signal in the same file as the signal handlers, because then when search imports it, it will also import models, and things will break.  There's ways to work around it, but the code was a lot cleaner with two different files.

This [StackOverflow answer](https://stackoverflow.com/a/22924754/10612) shows an alternate way it could work, with a ``signals`` folder, signals defined in ``signals/__init__.py``, and handlers in ``signals/handlers.py``. That would also work, I just thought it was uglier than ``signals.py`` and ``signal_handlers.py``.

kuma/users/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Signal handler to be called when a given email address was confirmed
+    by a user.
+    """
+    if switch_is_active('welcome_email'):
+        # only send if the user has exactly one verified (the given)
+        # email address, in other words if it was just confirmed
+        user = email_address.user
+        previous_emails = user.emailaddress_set.exclude(pk=email_address.pk)
+        if not previous_emails.exists():
```
escattone: This looks much better!

kuma/wiki/apps.py
```diff
@@ -55,7 +14,9 @@ class WikiConfig(AppConfig):
     verbose_name = _("Wiki")
 
     def ready(self):
-        super(WikiConfig, self).ready()
+        '''Configure kuma.wiki after models are loaded.'''
+        # Register signal handlers
+        from . import signal_handlers  # noqa
 
         # Configure Elasticsearch connections for connection pooling.
         es_connections.configure(
```
escattone: Unrelated to this PR, but I was just thinking that maybe this belongs under `search.apps.SearchConfig.ready`?

kuma/wiki/signal_handlers.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@receiver(pre_save, sender=DocumentZone, dispatch_uid='wiki.zone.pre_save')
+def on_zone_pre_save(sender, instance, **kwargs):
+    """
+    A signal handler to capture the previous state of the zone
+    (from the DB) before it is lost.
+    """
+    # Temporarily save the previous state on the instance itself.
+    dz_objs = DocumentZone.objects.values('document_id', 'document__locale')
```
escattone: Nice reorganization.

kuma/attachments/signal_handlers.py
```diff
@@ -0,0 +1,29 @@
+from django.db.models.signals import pre_delete, post_delete
+from django.dispatch import receiver
+
+from .models import AttachmentRevision, TrashedAttachment
+
+
+@receiver(pre_delete, sender=AttachmentRevision,
+          dispatch_uid='attachments.trash.delete')
```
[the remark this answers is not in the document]
jwhitlock: 🤦‍♂️ yep got confused.

kuma/wiki/apps.py
jwhitlock: I like that idea. Trying it out...

jwhitlock:
We're a little closer to Django 1.11. System check completes with several [urlpattern issues](https://travis-ci.org/mozilla/kuma/jobs/349847030#L784-L811). The next non-compliant library is [django-statici18n](https://github.com/zyegfryed/django-statici18n), which uses a [removed management class](https://travis-ci.org/mozilla/kuma/jobs/349847030#L919-L933) in version 1.1.5. Updating to the [current v1.7.0](https://github.com/zyegfryed/django-statici18n/blob/master/docs/changelog.rst) should help.

@escattone, ready for re-review.

kuma/attachments/signal_handlers.py
```diff
@@ -5,7 +5,7 @@
 
 
 @receiver(pre_delete, sender=AttachmentRevision,
```
escattone: The signals need to be swapped as well, so this should be `post_delete` instead of `pre_delete`.

kuma/attachments/signal_handlers.py
```diff
@@ -18,7 +18,7 @@ def after_revision_delete(instance, **kwargs):
 
 
 @receiver(post_delete, sender=TrashedAttachment,
```
escattone: ...and this should be `pre-delete`.

kuma/search/apps.py
```diff
@@ -1,12 +1,27 @@
...
+        # Configure Elasticsearch connections for connection pooling.
+        es_connections.configure(
+            default={
+                'hosts': settings.ES_URLS,
+            },
+            indexing={
+                'hosts': settings.ES_URLS,
+                'timeout': settings.ES_INDEXING_TIMEOUT,
+            },
+        )
```
escattone: ❤️ This feels better!

PR #4691, opened 2018-03-02
bug 1399639: Update django-constance to 2.1.0

jwhitlock:
* django-constance 1.1.2 → 2.1.0: Add compatibility with Django 1.11, custom fields, signal for updates.
* django-picklefield 0.3.2 → 1.0.0: Drop old Django support

These updates don't appear to require any code changes, but improve compatibility with Django 1.11 and open some possible features in the future.


jwhitlock:
Tests pass in TravisCI, site works locally, including the new "Reset to Default" link in the admin.

PR #4693, opened 2018-03-02
bug 1431259: caching headers/tests for wiki code/legacy/akismet views

escattone:
This is another in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds/modifies caching headers and tests for the endpoints for the kuma wiki views in `akismet_revision.py`, `code.py`, and `legacy.py`.

- This PR replaces the old-style `AkismetRevisionTests` in `kuma/wiki/tests/test_views.py` with new pytest-style tests in `kuma/wiki/tests/test_views_akismet.py`. As part of that effort, I realized that the existing `AkismetRevisionTests.test_submit_akismet_spam_revision_dne` test was flawed. It did not actually test the case it was intended to test (making an Akismet submission on a non-existent revision). Deep within the form's `clean` method, the Akismet key verification failed because it wasn't set-up properly in the test, which masked an ISE (a `KeyError` when getting the `revision` from the cleaned data). The new pytest-style `test_spam_revision_does_not_exist` actually checks that, so in order for it to pass, I made a fix to `RevisionAkismetSubmissionAdminForm` to check for the `revision` key prior to making any Akismet submissions.

- I didn't see any existing test for the `mindtouch_to_kuma_redirect` view in `legacy.py`, and I didn't add one. It didn't seem worth it, but I can add one if desired.

### DIFF
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -861,6 +861,14 @@ class RevisionAkismetSubmissionAdminForm(AkismetSubmissionFormMixin,
         akismet_data = AkismetHistoricalData(revision, self.request)
         return akismet_data.parameters
 
+    def clean(self):
+        if 'revision' not in self.cleaned_data:
+            raise forms.ValidationError(
+                _('Unable to make the Akismet submission (invalid revision).'),
+                code='invalid'
+            )
+        return super(RevisionAkismetSubmissionAdminForm, self).clean()
+
 
 class RevisionAkismetSubmissionSpamForm(RevisionAkismetSubmissionAdminForm):
     """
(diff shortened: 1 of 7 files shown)


escattone:
Rebased on master to remove conflict.

jwhitlock:
I've updated [bug 758730](https://bugzilla.mozilla.org/show_bug.cgi?id=758730#c8), which appears to be the "remove Mindtouch" bug, for ``mindtouch_to_kuma_redirect`` and related actions.

kuma/wiki/forms.py
```diff
@@ -861,6 +861,14 @@ def akismet_parameters(self):
         akismet_data = AkismetHistoricalData(revision, self.request)
         return akismet_data.parameters
 
+    def clean(self):
+        if 'revision' not in self.cleaned_data:
+            raise forms.ValidationError(
+                _('Unable to make the Akismet submission (invalid revision).'),
+                code='invalid'
```
jwhitlock: Good work finding this issue

PR #4694, opened 2018-03-03
bug 1428945: Update feeder app, allow calling as task

jwhitlock:
MDN used to republish a lot of RSS feeds and twitter streams, now there is just the Hacks Blog on the homepage. This used to be updated via a cronjob entry calling ``manage.py update_feeds`` every 10 minutes in SCL3, and hasn't been done automatically since moving to AWS.

This PR does many things:
- Refreshes the method for including the Hacks Blog on the homepage.
- Removes features from the feeder app that are no longer used, except for those that would require a database migration.
- Moves the feed updating code from the management command to a utility module.
- Fixes some bugs in the feed updating code that may have been introduced by ``feedparser`` updates, or may have always been there.
- Add complete testing code for the completely untested feeder app.
- Adds a celery task that does the same thing as the management task.

The sample database includes the ``moz-hacks`` Feed instance, so both the management command and the task should work. I've also tested against the production feeds, which includes 8 other feeds that are no longer republished on MDN.


jwhitlock:
I agree w/ @metadave on https://github.com/mozmeao/infra/issues/587, that we should add another ``celerybeat`` task rather than a kubernetes ``CronJob``.

PR #4703, opened 2018-03-08
bug 1401246: Remove django-localdevstorage

jwhitlock:
This library looked for missing local files on a remote server, downloading them as needed. This was useful when loading a production database on a development machine. However, we no longer ``SERVE_MEDIA`` in production, so a local development box can't access the raw URLs. In any
case, the sample database is preferred over a production database.
    
This removes the library and customized storage backend, basically reverting commit 5042d297.


kuma/attachments/views.py
```diff
@@ -41,15 +41,8 @@ def raw_file(request, attachment_id, filename):
...
-        if settings.DEBUG:
-            # to work around an issue of the localdevstorage with streamed
-            # files we'll have to read some of the file here first
-            rev.file.read(rev.file.DEFAULT_CHUNK_SIZE)
         response = StreamingHttpResponse(rev.file, content_type=rev.mime_type)
-        try:
-            response['Content-Length'] = rev.file.size
-        except OSError:
-            pass
+        response['Content-Length'] = rev.file.size
```
escattone: So much cleaner!

PR #4709, opened 2018-03-15
Fix Bug 1445889, preload title and body fonts for improved performance

schalkneethling:


jinja2/base.html
```diff
@@ -17,6 +17,10 @@
         "noindex, nofollow"
     {%- endif -%}
   >
+
+  <!-- because this is non-blocking, preload as early as possible -->
```
jwhitlock: This HTML comment seems more appropriate in ``preload.html``, since it goes with the markup there.  If you just want a comment for developers, I'd suggest a ``{# jinja comment #}``.

jinja2/base.html
```diff
@@ -17,6 +17,10 @@
...
+  {% include "includes/preload.html" %}
```
jwhitlock: Why is this included as a separate file rather than inline?

jinja2/base.html
schalkneethling: Makes sense.

jinja2/base.html
schalkneethling: I am a bit of a stickler for this kinda thing. I get a little startled ;) when I open an HTML file and it is this huge wall of code. This is especially true for `head` sections.

This allows me to have this isolated, it does not add any performance issue, and is simple to update in future.

With all of that said, I am not resistant to change so, if you would prefer I inline this, I am more than happy to do so.

schalkneethling:
> It's strange to me to ship a tag that FF doesn't recognize, but it looks like we did it for the interactive examples stuff, so it must be mostly working.

In this case browsers other than chrome will just ignore those. We have a large audience of Chrome users and so, it makes sense to roll this out as a progressive enhancement.

> We should find a way to preload those only on pages that include interactive examples...

I noticed that these are added on pages where it is not required, as it is just a hint to the browser and not a preload of a bunch of content it is "ok". But I agree, I would want to fix this as soon as possible.

Thanks for the review @jwhitlock

jinja2/includes/preload.html
```diff
@@ -0,0 +1,5 @@
+{# because this is non-blocking, preload as early as possible #}
+<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Bold.woff') }}" as="font" type="font/woff" crossorigin />
```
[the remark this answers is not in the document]
jwhitlock: See #4728

PR #4712, opened 2018-03-16
bug 1412891: Adjust Brotli compression level

MatonAnthony:
The current settings for Brotli compression level is too high and it's
increasing our average page load time as a side effect.

This patch put the compression level at a more reasonnable level to
lower the tradeoff.

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -263,7 +263,7 @@ class BrotliMiddleware(object):
             # ---------
             return response
 
-        compressed_content = brotli.compress(response.content)
+        compressed_content = brotli.compress(response.content, quality=5)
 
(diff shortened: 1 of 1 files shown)


kuma/core/middleware.py
```diff
@@ -263,7 +263,7 @@ def process_response(self, request, response):
             # ---------
             return response
 
-        compressed_content = brotli.compress(response.content)
+        compressed_content = brotli.compress(response.content, quality=5)
```
jwhitlock: 👍 This article, [The current state of Brotli compression](https://samsaffron.com/archive/2016/06/15/the-current-state-of-brotli-compression), suggests that level 5 is competitive with gzip in time and compression.

PR #4715, opened 2018-03-20
Bug 1447446: Increase InnoDB log file size

a2sheppy:
Updates docker-compose.yml to set innodb_log_file_size to
100 megabytes, in order to ensure that it's at least 10x
the maximum size of BLOB/TEXT data in any one transaction,
per MySQL 5.6.20 release notes [1]. See bug 1447446 for
more information.

Resolves issues importing the anonymized database, and
probably random database errors that may occur when
dealing with very large transactions.

[1]  https://is.gd/wvSZzV

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,6 +74,7 @@ services:
       - --character-set-server=utf8
       - --collation-server=utf8_general_ci
       - --innodb-flush-log-at-trx-commit=0
+      - --innodb_log_file_size=100M
       - --max-allowed-packet=100M
     volumes:
       - mysqlvolume:/var/lib/mysql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,7 +74,7 @@ services:
       - --character-set-server=utf8
       - --collation-server=utf8_general_ci
       - --innodb-flush-log-at-trx-commit=0
-      - --innodb_log_file_size=132M
+      - --innodb-log-file-size=132M
       - --max-allowed-packet=100M
     volumes:
       - mysqlvolume:/var/lib/mysql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,7 +74,7 @@ services:
       - --character-set-server=utf8
       - --collation-server=utf8_general_ci
       - --innodb-flush-log-at-trx-commit=0
-      - --innodb_log_file_size=100M
+      - --innodb_log_file_size=132M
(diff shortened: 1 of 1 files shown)


MatonAnthony:
Looks good to me.

This change on MySQL side is interesting, I am wondering if we are likely to hit a similar issue on staging or in production (maybe @escattone would have an idea).

escattone:
@MatonAnthony For both staging and production, we use a different value, about 132M.

docker-compose.yml
```diff
@@ -74,6 +74,7 @@ services:
       - --character-set-server=utf8
       - --collation-server=utf8_general_ci
       - --innodb-flush-log-at-trx-commit=0
+      - --innodb_log_file_size=100M
```
escattone: Since this is a command-line option, it should use dashes instead of underscores, so it should instead be:
```
- --innodb-log-file-size=100M
```

a2sheppy:
I'll update the PR.

escattone:
Strange. I first made [this comment](https://github.com/mozilla/kuma/pull/4715/files/2db0b97f1f33aaa6fc3f5e1b5c444ffa2bd4d18c#r176144264), but when @a2sheppy made a new commit before I could submit my review, it seemed to get lost. I made a new comment based on the new commit, and submitted the review, but GitHub used the old comment and tossed my new one.

docker-compose.yml
```diff
@@ -74,6 +74,7 @@ services:
       - --character-set-server=utf8
       - --collation-server=utf8_general_ci
       - --innodb-flush-log-at-trx-commit=0
+      - --innodb_log_file_size=132M
```
escattone: Since this is a command-line argument, it should use dashes instead of underscores, so it should be:
```
- --innodb-log-file-size=132M
```

escattone:
Thanks for the changes @a2sheppy. Just to be clear, we don't have to use the same value (~`132M`) as stage and production, just whatever value works for testing and local development.

My only remaining concern is that this may re-introduce the initialization issue in Travis CI. Probably not, since the value is much smaller than the 1G that was used before, but I'd like to wait for @jwhitlock to take a look before we merge since he has experience with the previous issue.

jwhitlock:
The [upstream bug](https://github.com/docker-library/mysql/issues/356) in the docker MySQL image was fixed, so the production value should be OK.

PR #4717, opened 2018-03-22
bug 1431259: caching headers/tests for various views

escattone:
This PR depends on #4701 for it's tests to pass.

This is another in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds/modifies caching headers and tests for various endpoints within kuma (e.g., `kuma.users.urls`).

It adds and uses a new dependency `django-decorator-include`, which provides the ability to decorate all views within an include tree. The version used is [1.3](https://pypi.python.org/pypi/django-decorator-include/1.3), which is the proper version to use with Django 1.8, but version [1.4.1](https://pypi.python.org/pypi/django-decorator-include/1.4.1) should be used when moving to Django 1.11.

### DIFF
--- a/kuma/authkeys/tests/test_views.py
+++ b/kuma/authkeys/tests/test_views.py
@@ -4,7 +4,6 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
 from django.test import TestCase
 
-from kuma.core.tests import eq_, ok_
 from kuma.core.urlresolvers import reverse
 from kuma.users.tests import user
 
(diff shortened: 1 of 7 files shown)


kuma/authkeys/tests/test_views.py
```diff
@@ -4,7 +4,6 @@
 from django.contrib.auth.models import Permission
 from django.test import TestCase
 
-from kuma.core.tests import eq_, ok_
```
jwhitlock: 🎉

kuma/users/tests/test_templates.py
```diff
@@ -642,24 +707,26 @@ def test_revisions_posted_same_doc(self, mock_form):
             creator=self.testuser)
 
         self.client.login(username='admin', password='testpass')
-        ban_url = reverse('users.ban_user_and_cleanup_summary',
+        ban_url = reverse('users.ban_user_and_cleanup_summary', locale='en-US',
```
jwhitlock: Nice, gets rid of that ``full_ban_url`` thing

PR #4718, opened 2018-03-26
bug 1443586: Make lang preference cookie persistent

MatonAnthony:
Increase the max-age of the language preference cookie to one
year instead of only the current session.

This avoids to ask if people wants to keep their current locale
at each of their visits.

### DIFF
--- a/kuma/core/tests/test_views.py
+++ b/kuma/core/tests/test_views.py
@@ -156,6 +156,8 @@ def test_setting_language_cookie_working(client):
     # Check language cookie is set
     assert lang_cookie
     assert lang_cookie.value == 'bn-BD'
+    # Check that the max-age from the cookie is the same as our settings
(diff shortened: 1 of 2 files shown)


kuma/core/tests/test_views.py
```diff
@@ -152,6 +152,15 @@ def test_not_possible_to_set_non_locale_cookie(self):
         # No language cookie should be saved as `foo` is not a supported locale
         assert not language_cookie
 
+    def test_setting_language_cookie_age(self):
```
escattone: @MatonAnthony I think you'll first have to rebase your branch on master, as the `LanguageCookieTest` class no longer exists. Once you rebase on the latest master, you could fold this check:
```
assert language_cookie['max-age'] == settings.LANGUAGE_COOKIE_AGE
```
into the existing test `test_setting_language_cookie_working`.

PR #4719, opened 2018-03-26
bug 1431259: caching headers/tests for wiki document views

escattone:
This PR depends on #4701 for it's tests to pass.

This is the last in a series of PR's that add the appropriate caching headers and related tests to all Kuma endpoints as part of the effort of placing a CDN in front of MDN. This PR adds/modifies caching headers and tests for the views within `kuma/wiki/views/document.py`.

### DIFF
--- a/kuma/wiki/tests/test_templates.py
+++ b/kuma/wiki/tests/test_templates.py
@@ -128,17 +128,20 @@ class DocumentTests(UserTestCase, WikiTestCase):
         revision(document=d2, is_approved=False, save=True)
         url = reverse('wiki.document', args=[d2.slug], locale='fr')
         response = self.client.get(url)
+        assert response.status_code == 200
+        assert 'public' in response['Cache-Control']
+        assert 's-maxage' in response['Cache-Control']
         doc = pq(response.content)
-        eq_(d2.title, doc('main#content div.document-head h1').text())
+        assert doc('main#content div.document-head h1').text() == d2.title
 
         # Fallback message is shown.
-        eq_(1, len(doc('#doc-pending-fallback')))
-        ok_('$translate' in doc('#edit-button').attr('href'))
+        assert len(doc('#doc-pending-fallback')) == 1
+        assert '$translate' in doc('#edit-button').attr('href')
         # Removing this as it shows up in text(), and we don't want to depend
         # on its localization.
         doc('#doc-pending-fallback').remove()
         # Included content is English.
-        eq_(pq(r.document.html).text(), doc('article#wikiArticle').text())
+        assert pq(r.document.html).text() == doc('article#wikiArticle').text()
 
     def test_document_fallback_no_translation(self):
         """The document template falls back to English if no translation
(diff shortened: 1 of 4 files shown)


kuma/wiki/tests/test_views_document.py
```diff
@@ -590,3 +635,177 @@ def test_tags_not_show_while_empty(root_doc, client, wiki_user):
...
+
+
+@pytest.mark.parametrize('params_case', ['with-params', 'without-params'])
+def test_fallback_to_translation(root_doc, trans_doc, client, params_case):
+    """
+    If a slug isn't found in the requested locale but is in the default
+    locale and if there is a translation of that default-locale document to
+    the requested locale, the translation should be served.
+    """
+    params = '?x=y&x=z' if (params_case == 'with-params') else ''
```
jwhitlock: 👍

kuma/wiki/tests/test_views_document.py
```diff
@@ -590,3 +635,177 @@ def test_tags_not_show_while_empty(root_doc, client, wiki_user):
...
+    assert 'max-age=0' in response['Cache-Control']
+    assert 'no-cache' in response['Cache-Control']
+    assert 'no-store' in response['Cache-Control']
+    assert 'must-revalidate' in response['Cache-Control']
+    assert response['Location'].endswith(
+        reverse('wiki.document', locale=root_doc.locale, args=[root_doc.slug]))
+    assert event.is_notifying(wiki_user, root_doc), 'Watch was not created'
+
+    # Unsubscribe
+    response = user_client.post(url)
```
jwhitlock: 😢  it's the same endpoint to subscribe and unsubscribe 🤦‍♂️

kuma/wiki/tests/test_views_document.py
```diff
@@ -590,3 +635,177 @@ def test_tags_not_show_while_empty(root_doc, client, wiki_user):
...
+    assert 'no-cache' in response['Cache-Control']
+    assert 'no-store' in response['Cache-Control']
+    assert 'must-revalidate' in response['Cache-Control']
+    assert response['Location'].endswith(
+        reverse('wiki.document', locale=root_doc.locale, args=[root_doc.slug]))
+    assert not event.is_notifying(wiki_user, root_doc), \
+        'Watch was not destroyed'
+
+
+def test_zone_styles(client, doc_hierarchy_with_zones):
```
jwhitlock: someday: move to ``legacy.py``

kuma/wiki/tests/test_views_document.py
```diff
@@ -590,3 +635,177 @@ def test_tags_not_show_while_empty(root_doc, client, wiki_user):
...
+    assert 'max-age=0' in response['Cache-Control']
+    assert 'no-cache' in response['Cache-Control']
+    assert 'no-store' in response['Cache-Control']
+    assert 'must-revalidate' in response['Cache-Control']
+
+
+@pytest.mark.parametrize(
+    'endpoint', ['wiki.subscribe', 'wiki.subscribe_to_tree'])
+def test_watch_login_required(client, root_doc, endpoint):
+    """Watch document with HTTP non-POST request results in 405."""
```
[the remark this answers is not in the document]
escattone: Oops, good catch, thanks!

kuma/wiki/tests/test_templates.py
```diff
@@ -128,35 +128,42 @@ def test_document_fallback_with_translation(self):
...
-        eq_(pq(r.document.html).text(), doc('article#wikiArticle').text())
+        assert pq(r.document.html).text() == doc('article#wikiArticle').text()
 
     def test_document_fallback_no_translation(self):
         """The document template falls back to English if no translation
         exists."""
         r = revision(save=True, content='Some text.', is_approved=True)
         url = reverse('wiki.document', args=[r.document.slug], locale='fr')
         response = self.client.get(url)
+        assert response.status_code == 200
```
safwanrahman: I am seeing this common thing is checked in many tests. Is it possible to have a generalized function for this?
Something like
```python
def check_cahce(response):
     assert 'public' in response['Cache-Control']
     assert 's-maxage' in response['Cache-Control']
```

Then call `check_cahce` from all the function where its needed.

kuma/wiki/tests/test_views_document.py
escattone: I couldn't believe when I first saw the tests. 🤦‍♂️

kuma/wiki/tests/test_templates.py
escattone: @safwanrahman Good question. I've often wondered what would be best when I'm writing tests, to make things DRY ("Don't Repeat Yourself"), as you've indicated, or more directly readable as I've done. For tests, I've tended towards trying to make them as directly readable as possible, so there's no need to look-up/understand additional functions.

PR #4720, opened 2018-03-28
Fix Bug 1440908, show error message to users on 500 internal server error

schalkneethling:
To test this, I load up the initial activity history page, I then use the network panel in devtools to emulate offline mode. Click to go to the next page of results, you should now see an error shown in the Growl message instead of 'Updating filter.....'

### DIFF
--- a/kuma/static/js/dashboard.js
+++ b/kuma/static/js/dashboard.js
@@ -124,6 +124,9 @@
             notification.success(gettext('Updated filters.'), 2000);
             // Reset the page count to 0 in case of new filter
             $pageInput.val(1);
+        }).fail(function(jqXHR, textStatus, errorThrown) {
+            notification.error(gettext('Error loading content, please refresh the page'), 5000);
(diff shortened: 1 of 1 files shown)


kuma/static/js/dashboard.js
```diff
@@ -124,6 +124,9 @@
             notification.success(gettext('Updated filters.'), 2000);
             // Reset the page count to 0 in case of new filter
             $pageInput.val(1);
+        }).fail(function(jqXHR, textStatus, errorThrown) {
+            notification.error(gettext('Error loading content, please refresh the page'), 5000);
```
jwhitlock: 👍 on using ``gettext`` to localize.

schalkneethling:
> I'm not sure that a refresh is the action we want people to take, but it will help in the case of an intermittent error

Yeah, I was not sure whether to say "try again later". But at least there is some form of feedback now.

PR #4722, opened 2018-03-28
Replaced hard-coded strings with translatable ones

asmforce:
Here's a fix for hard-coded (non-translatable via pontoon.mozilla.org) strings on article history page.
Please have a look.


asmforce:
@jwhitlock, thanks for that. I'll check contribution guidelines.

PR #4725, opened 2018-03-30
bug 1431259: change stage domain to developer.allizom.org

escattone:
This PR updates kuma to reflect the new domain name (actually a return to the old domain name!) for stage: `developer.allizom.org`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -277,6 +277,6 @@ Deployment and Closing Bugs
 ---------------------------
 MDN staff deploy new code one to three times a week. Deployments are announced
 in the #mdndev IRC channel. Currently deployed code can be seen on the
-[What's Deployed?](https://whatsdeployed.io/s-NyD) page.
+[What's Deployed?](https://whatsdeployed.io/s-HC0) page.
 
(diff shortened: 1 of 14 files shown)


tests/headless/test_cdn.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+import requests
+
+from kuma.core.urlresolvers import reverse
```
jwhitlock: Functional tests should not import from Kuma, or run under a Django environment. These URLs need to be hard-coded.

tests/headless/test_cdn.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@pytest.mark.headless
+@pytest.mark.nondestructive
+@pytest.mark.parametrize(
+    'endpoint',
+    ['home', 'wiki.'])
+def test_document(base_url, endpoint):
```
jwhitlock: "test_document" seems overly vague to me. "test_cached"?

escattone:
Oh no, this `tests/headless/test_cdn.py` file should not have been part of this PR, sorry! It was a quick shell for exploring some headless tests for the CDN work. I'll remove it.

PR #4726, opened 2018-04-02
Fix Bug 1450647, adds sonarwhal and related config

schalkneethling:
To test that this works do:

```
npm install
...
npm run test:sonarwhal
```


schalkneethling:
This was not envisioned to currently be part of a CI so, I am going to close and use it as a global.

PR #4728, opened 2018-04-03
Only preload .woff2 font files

alexgibson:
I currently have a few pages of MDN listed on the [performance dashboard](https://mozmeao-perf-dashboard.netlify.com/) I'm building for our team. Because I'm spending a fair bit of time looking at this, I was intrigued to see a noticable up-tick in both load times and bytes in, as you can see here:

screenshot-2018-4-3 web performance dashboard

After a bit of digging in the [web page test result](https://www.webpagetest.org/results.php?test=180403_16_4c356dcf4f3d6eb2c89cbb53d94a3cbf), I noticed that both `.woff` and `.woff2` files are being loaded, which seemed like a bug:

screenshot-2018-4-3 webpagetest test details - california dev mozilla org en-us - 04 03 18 03 25 41

I think this PR should fix the issue because:

- Any browser that supports `.woff2` should no longer need `.woff` files.
- Any browser that supports `rel="preload"` should also support `.woff2`.

### DIFF
--- a/jinja2/includes/preload.html
+++ b/jinja2/includes/preload.html
@@ -1,5 +1,3 @@
 {# because this is non-blocking, preload as early as possible #}
-<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Bold.woff') }}" as="font" type="font/woff" crossorigin />
 <link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Bold.woff2') }}" as="font" type="font/woff2" crossorigin />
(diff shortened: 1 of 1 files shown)


alexgibson:
Looks like it worked, thanks @schalkneethling 👍 

screenshot-2018-4-6 web performance dashboard

(my first contribution to MDN 🍻)

schalkneethling:
\o/ Thanks Alex!

Kind Regards,
Schalk Neethling
Senior Front-End Engineer
Mozilla

escattone:
Nice, thanks @alexgibson!

PR #4731, opened 2018-04-06
bug 1431259: Don't cache the "create page" redirect

jwhitlock:
Use a "no-cache" cache control header when redirecting to create a page, to avoid a loop where you have to wait for the cache to expire to see the new page. This also updates the ``shared_cache_control`` decorator to take no action if it detects the cache control header is set to no-cache.

### DIFF
--- a/kuma/core/decorators.py
+++ b/kuma/core/decorators.py
@@ -7,9 +7,9 @@ from django.contrib.auth.decorators import user_passes_test
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.shortcuts import redirect, render
+from django.utils.cache import patch_cache_control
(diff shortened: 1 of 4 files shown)


kuma/core/decorators.py
```diff
@@ -7,33 +7,48 @@
...
-    if not func:
-        return decorator
-    return decorator(func)
+
+    def _shared_cache_controller(viewfunc):
+        @wraps(viewfunc, assigned=available_attrs(viewfunc))
+        def _cache_controlled(request, *args, **kw):
+            response = viewfunc(request, *args, **kw)
+            nocache = (response.has_header('Cache-Control') and
+                       'no-cache' in response['Cache-Control'])
```
escattone: I think we should add
```python
 and 'no-store' in response['Cache-Control']
```
as well, since either could be used independent of the other.

PR #4732, opened 2018-04-06
bug 1431259: add cache control header to locale redirects

escattone:
This PR does the following:
- adds the shared cache control header to the locale redirects performed by `LocaleURLMiddleware`
- in the spirit of DRY, factors out the `add_shared_cache_control` function to use independently of the `shared_cache_control` decorator
- changes the `nocache` check to be an `or` of the no-cache and no-store cases (sorry, that was my mistake to suggest `and` when originally proposing the addition of no-store to the check)

### DIFF
--- a/kuma/core/decorators.py
+++ b/kuma/core/decorators.py
@@ -7,12 +7,12 @@ from django.contrib.auth.decorators import user_passes_test
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.shortcuts import redirect, render
-from django.utils.cache import patch_cache_control
 from django.utils.decorators import available_attrs
 from django.utils.http import urlquote
 
 from .jobs import BannedIPsJob
 from .urlresolvers import reverse
+from .utils import add_shared_cache_control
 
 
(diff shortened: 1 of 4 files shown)


kuma/core/decorators.py
```diff
@@ -15,6 +15,35 @@
 from .urlresolvers import reverse
 
 
+def add_shared_cache_control(response, **kwargs):
```
jwhitlock: How about ``kuma/core/cache.py`` or ``kuma/core/utils.py`` for this function?

escattone:
That's a great point about adding `Cookie` to the vary header in the case when the choice is based on the `django_language` cookie rather than for example the `Accept-Language` header. It won't affect our Cloudfront CDN (it ignores the `Vary` header as far as I can tell), but any other public caches could be affected.

I will add some tests as well.

Thanks for the review @jwhitlock!

kuma/core/decorators.py
escattone: I'll go with `kuma/core/utils.py` if you don't mind.

escattone:
Ready for a second look @jwhitlock!

PR #4736, opened 2018-04-12
create/use test functions to encapsulate cache-control header assertions

escattone:
This PR arose from @jwhitlock's [comment](https://github.com/mozilla/kuma/pull/4734#discussion_r180241870) in #4734 as well as @safwanrahman's earlier [comment](https://github.com/mozilla/kuma/pull/4719#discussion_r177575655). It creates two new test utility functions `assert_no_cache_header` and `assert_shared_cache_header`, and then replaces all groups of assertions that each encapsulates.

### DIFF
--- a/kuma/core/tests/__init__.py
+++ b/kuma/core/tests/__init__.py
@@ -13,6 +13,18 @@ from ..cache import memcache
 from ..urlresolvers import split_path
 
 
+def assert_no_cache_header(response):
+    assert 'max-age=0' in response['Cache-Control']
+    assert 'no-cache' in response['Cache-Control']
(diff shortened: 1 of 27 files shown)


kuma/core/tests/__init__.py
```diff
@@ -13,6 +13,18 @@
 from ..urlresolvers import split_path
 
 
+def assert_no_cache_header(response):
+    assert 'max-age=0' in response['Cache-Control']
```
jwhitlock: Nit: I think we should assert 's-maxage' is not in the header as well

escattone: I added a second commit that does this, thanks @jwhitlock!

PR #4737, opened 2018-04-17
bug 1453796: fix generation of HTML when filtering noincludes

escattone:
This PR fixes [Bugzilla #1453796](https://bugzilla.mozilla.org/show_bug.cgi?id=1453796). The bug occurs when the `include` parameter is used during document requests, for example when using the `page` macro. The `include` parameter triggers a removal of all elements with `class="noinclude"`, which is done using `pyquery`, after which it re-generates the filtered HTML. During the re-generation of the HTML, any empty HTML elements (e.g., `<iframe ...></iframe>`) are output as self-closing elements (e.g., `<iframe ... />`), which in the case of `iframe`, is illegal. When loaded, the self-closing `iframe` seems to be interpreted as simply an opening tag, and so swallows all subsequent HTML.

For example, https://developer.mozilla.org/en-US/docs/Web/CSS/perspective?raw=1&macros=1&section=Setting_perspective works fine, but when the `include=1` is added, the failure appears (https://developer.mozilla.org/en-US/docs/Web/CSS/perspective?raw=1&macros=1&include=1&section=Setting_perspective).

The fix is to call `doc.html(method='html')` instead of `doc.html()` in the `kuma.wiki.content.filter_out_noinclude` function (see http://pyquery.readthedocs.io/en/latest/api.html#pyquery.pyquery.PyQuery.html).

This does not appear to be due to a `pyquery` package version update. I'm not sure why and when this bug appeared. Could it be that the `iframe` elements used to be populated, but then changed to being empty?

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -264,7 +264,12 @@ def filter_out_noinclude(src):
         return ''
     doc = pq(src)
     doc.remove('*[class=noinclude]')
-    return doc.html()
+    # Use "method='html'" to prevent improper closure of some empty HTML
+    # elements, like the "iframe" element for example, which is not allowed
+    # to be closed in the opening tag. For example, without "method='html'"
+    # the output would be "<iframe/>" instead of the correct
+    # "<iframe></iframe>".
+    return doc.html(method='html')
 
 
 class ContentSectionTool(object):
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -15,6 +15,7 @@ from lxml import etree
 from pyquery import PyQuery as pq
 
 from kuma.core.urlresolvers import reverse
+from kuma.core.utils import to_html
 
 from .exceptions import DocumentRenderedContentNotAvailable
 from .utils import locale_and_slug_from_path
@@ -218,7 +219,8 @@ def get_seo_description(content, locale=None, strip_markup=True):
             if strip_markup:
                 seo_summary = summaryClasses.text()
             else:
-                seo_summary = summaryClasses.html()
+                seo_summary = ''.join(
+                    to_html(item) for item in summaryClasses.items())
         else:
             paragraphs = page.find('p')
             if paragraphs.length:
@@ -227,7 +229,7 @@ def get_seo_description(content, locale=None, strip_markup=True):
                     if strip_markup:
                         text = item.text()
                     else:
-                        text = item.html()
+                        text = to_html(item)
                     # Checking for a parent length of 2
                     # because we don't want p's wrapped
                     # in DIVs ("<div class='warning'>") and pyQuery adds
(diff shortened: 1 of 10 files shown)


a2sheppy:
Looking forward to this landing; it will let me finish up a few things. Great job!

jwhitlock:
I dug in to see when this might have broke, and I can't identify a library update that would have changed this behavior. I tested a few configurations going back 2-3 years ago, and the issue persists.  

``pyquery`` uses ``lxml.etree.tostring()`` to implement ``.html()``, and that outputs XHTML by default, where ``<iframe />`` is valid. However, it's not valid HTML 5, and using a self-closing tag will cause issues in Firefox and other browsers (https://stackoverflow.com/a/27545787/10612). Passing ``method='html'`` is the way to get HTML instead of XML (see http://lxml.de/tutorial.html#serialisation). The HTML option was added in lxml 2.0, released in 2008.

Every year, I find something that makes me wish XHTML won.

There's a few more instances where ``.html()`` is called in the code. Most are in tests, but one in particular is in the code that extracts SEO summaries. @escattone, should we include fixing those with this PR, or save it for a new one?

kuma/wiki/content.py
```diff
@@ -264,7 +264,12 @@ def filter_out_noinclude(src):
         return ''
     doc = pq(src)
     doc.remove('*[class=noinclude]')
-    return doc.html()
+    # Use "method='html'" to prevent improper closure of some empty HTML
+    # elements, like the "iframe" element for example, which is not allowed
+    # to be closed in the opening tag. For example, without "method='html'"
+    # the output would be "<iframe/>" instead of the correct
+    # "<iframe></iframe>".
+    return doc.html(method='html')
```
jwhitlock: If we decide to fix all the ``.html()`` calls, this might be done as a utility function such as ``to_html``, rather than adding a 5-line comment to every call.

kuma/wiki/tests/test_views.py
```diff
@@ -2710,6 +2710,7 @@ def test_raw_include_option(self):
               <dd>Type: <em>integer</em></dd>
               <dd>Przykłady 例 예제 示例</dd>
             </dl>
+            <p><iframe></iframe></p>
```
jwhitlock: I was going to give feedback that the ``<iframe>`` test should be in it's own test, but 1) I don't want to make you convert to ``pytest`` as well, and 2) pyquery will return ``None`` if you just serialize an empty ``<iframe>``.

a2sheppy:
👍

escattone:
@jwhitlock This is ready for another review. Here's what I changed:
- created a `to_html` utility function and replaced all calls to `.html()` on PyQuery instances
- modified tests for the `wiki.document_api` endpoint
- totally refactored the tests for the `kuma.wiki.content.get_seo_description` function, including testing more cases (with the expanded testing, I found/fixed what seemed like a bug in the handling of the case when not stripping markup with multiple elements with the class `seoSummary`)

jwhitlock:
The additional code looks good as well. Thanks @escattone

PR #4738, opened 2018-04-17
Fix Bug 1454576, dns prefetch and preconnect to cdn

schalkneethling:


jwhitlock:
Our plan is to deprecate ``cdn.mdn.mozilla.net`` and serve static assets from the main domain.

PR #4739, opened 2018-04-17
Fix Bug 1450646, remove type attribute from script tags

schalkneethling:


kuma/wiki/tests/test_views_code.py
```diff
@@ -53,7 +53,7 @@ def test_code_sample(code_sample_doc, constance_config, client, settings):
         ' rel="stylesheet" type="text/css">'
         '<style type="text/css">.some-css { color: red; }</style>'
         'Some HTML'
-        '<script type="text/javascript">window.alert("HI THERE")</script>'
+        '<scrip>window.alert("HI THERE")</script>'
```
jwhitlock: should be ``script`` not ``scrip``

schalkneethling:
Thanks @jwhitlock

PR #4740, opened 2018-04-19
bug 1431259: remove compression middleware and its tests

escattone:
This PR removes both the GZip and Brotli compression middleware and its tests. Since MDN is now completely behind a CDN, and since the CDN handles compression (GZip only currently), it is no longer necessary for Kuma to support compression. This will improve the response time as well.


escattone:
Thanks for the review @jwhitlock, great catches. I made the changes you requested, and it's ready for your second review.

PR #4742, opened 2018-04-20
bug 1455652: Register async search tasks

jwhitlock:
After the change to AppConfig, search tasks are no longer auto-registering. Force the issue by using ``CELERY_IMPORTS``.

Other tasks (wiki, user, etc) are being auto-registered.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1305,6 +1305,7 @@ CELERY_ACCEPT_CONTENT = ['pickle']
 
 CELERY_IMPORTS = (
     'tidings.events',
+    'kuma.search.tasks',
 )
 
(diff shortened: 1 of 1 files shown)


safwanrahman:
I wonder why the tasks are not auto registered! 😕 
Need to investigate!

jwhitlock:
@safwanrahman the ``django-celery`` auto register function doesn't correctly handle ``AppConfig`` apps. Other ``AppConfig``-app tasks are registered because their ``tasks.py`` are imported during app setup. ``django-celery`` is deprecated, and the [Django integration](http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html) now built into celery does handle them, but it will be much more than a 1-line change to switch.

PR #4744, opened 2018-04-21
bug 1401246: Don't use django-debreach in Django 1.10 and later

jwhitlock:
Update to django-debreach 1.4.2, which adds support for Django 2.0. Then parse the Django version in settings, to avoid installing debreach, using the context processor, or the middleware in Django 1.10 or later, where Django has built-in mitigation of [BREACH](https://en.wikipedia.org/wiki/BREACH) and the debreach middleware raises an ``ImproperlyConfigured`` exception.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -4,12 +4,14 @@ import logging
 import os
 import platform
 from collections import namedtuple
+from distutils.version import LooseVersion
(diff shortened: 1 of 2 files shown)


escattone:
By the way, I had never known about `distutils.version.LooseVersion` before this. Nice!

PR #4745, opened 2018-04-22
bug 1449210: Add hreflang for the current document

MatonAnthony:
Google is ignoring the hreflang if the current page is not referenced
too.

This commit add the hreflang for the current page you are in top
of the already existing hreflang for the translations.


kuma/wiki/jinja2/wiki/document.html
```diff
@@ -69,6 +69,8 @@
   <link rel="alternate" type="application/json" href="{{ url('wiki.json_slug', document.slug) | absolutify }}">
   <link rel="canonical" href="{{ canonical }}" >
 
+  {# Google requires the current page to be referenced as an alternate too for SEO purposes see bug 1449210 #}
```
jwhitlock: Nice use of Jinja comments

PR #4746, opened 2018-04-22
Do not preload fonts set as optional

schalkneethling:
Saw this article[1] by Zach Leatherman, shard by Alex Gibson, and it was a real doh! moment. After reading the spec, it makes no sense to preload the Bold font variant. The recommended block time defined in the spec[2] is ~100ms with no swap time.

As the Bold variant take over 400ms to download(see below), preloading this on the first hit is a wasted HTTP request.

screen shot 2018-04-22 at 11 52 29

[1] https://www.zachleat.com/web/preload-font-display-optional/
[2] https://drafts.csswg.org/css-fonts-4/#valdef-font-face-font-display-optional


jwhitlock:
Oops, I forgot to give feedback about a bug number in the commit message. Adding a comment to [bug 1445889](https://bugzilla.mozilla.org/show_bug.cgi?id=1445889).

PR #4747, opened 2018-04-23
[BUG: 1401246] Update jinja and django-jinja

Alig1493:
According to the Jinja2 v2.9 changelog:
The with and autoescape tags are now built-in.

Thus I have removed them from the settings file.

Link: http://jinja.pocoo.org/docs/2.10/changelog/#version-2-9

### DIFF
--- a/kuma/core/tests/test_helpers.py
+++ b/kuma/core/tests/test_helpers.py
@@ -50,7 +50,7 @@ class TestSoapbox(KumaTestCase):
         m = Message(message='Go to http://bit.ly/sample-demo', is_global=True,
                     is_active=True, url='/')
         m.save()
-        ok_('Go to <a href="http://bit.ly/sample-demo">'
+        ok_('Go to <a href="http://bit.ly/sample-demo" rel="noopener">'
             'http://bit.ly/sample-demo</a>' in
             soapbox_messages(get_soapbox_messages('/')))
 
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -619,9 +619,7 @@ TEMPLATES = [
             'extensions': [
                 'jinja2.ext.do',
                 'jinja2.ext.loopcontrols',
-                'jinja2.ext.with_',
                 'jinja2.ext.i18n',
-                'jinja2.ext.autoescape',
                 'puente.ext.i18n',
                 'django_jinja.builtins.extensions.CsrfExtension',
                 'django_jinja.builtins.extensions.CacheExtension',
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -254,9 +254,12 @@ pyparsing==2.2.0 \
     --hash=sha256:fee43f17a9c4087e7ed1605bd6df994c6173c1e977d7ade7b651292fab2bd010
 
 
-# Jinja2
-MarkupSafe==0.23 \
-    --hash=sha256:a4ec1aff59b95a14b45eb2e23761a0179e98319da5a7eb76b56ea8cdc7b871c3
+# Used for Jinja2
+# Code: https://github.com/pallets/markupsafe
+# Changes: https://github.com/pallets/markupsafe/blob/master/CHANGES
+# Docs: https://pypi.org/project/MarkupSafe/
+MarkupSafe==1.0 \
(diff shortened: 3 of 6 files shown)


kuma/core/tests/test_helpers.py
```diff
@@ -50,7 +50,7 @@ def test_message_with_url_is_link(self):
         m = Message(message='Go to http://bit.ly/sample-demo', is_global=True,
                     is_active=True, url='/')
         m.save()
-        ok_('Go to <a href="http://bit.ly/sample-demo">'
+        ok_('Go to <a href="http://bit.ly/sample-demo" rel="noopener">'
```
safwanrahman: According to `[jinja2 policy](http://jinja.pocoo.org/docs/2.10/api/#policies)`, seems like `noopener` is default

requirements/constraints.txt
```diff
@@ -254,9 +254,12 @@ pyparsing==2.2.0 \
     --hash=sha256:fee43f17a9c4087e7ed1605bd6df994c6173c1e977d7ade7b651292fab2bd010
 
 
-# Jinja2
-MarkupSafe==0.23 \
-    --hash=sha256:a4ec1aff59b95a14b45eb2e23761a0179e98319da5a7eb76b56ea8cdc7b871c3
+# Used for Jinja2
+# Code: https://github.com/pallets/markupsafe
+# Changes: https://github.com/pallets/markupsafe/blob/master/CHANGES
```
jwhitlock: 👍 thanks for adding the project details. It appears the jump to from 0.23 to 1.0 is about the project being feature-complete, and dropping Python 2.6 support.

requirements/default.txt
```diff
@@ -125,8 +125,12 @@ django-honeypot==0.5.0 \
     --hash=sha256:962304f21055f05c73c41f2506f5a94f5b941e831a203b68e54cb5016bdb8d5e
 
 # Use Jinja2 as the template engine in Django
-django-jinja==2.2.1 \
-    --hash=sha256:04c53ba90cbede97e6543b98291f1eec259e57dfa4e014a5fbf0291afcdae6a9
+# Code: https://github.com/niwinz/django-jinja
+# Changes: https://github.com/niwinz/django-jinja/blob/master/CHANGES.adoc
+# Docs: http://niwinz.github.io/django-jinja/latest/
```
jwhitlock: Again, thanks for the project details. 2.3.0 added Django 1.11 compatibility, along with a reverted experiment in loading Jinja2 templates inside Django templates.

requirements/default.txt
```diff
@@ -246,9 +250,12 @@ importlib==1.0.3 \
     --hash=sha256:65f342a604a2e1028707c5e055266ab2431c26e20fe10780b423320870884dac
 
 # Improved template engine
-Jinja2==2.8 \
-    --hash=sha256:1cc03ef32b64be19e0a5b54578dd790906a34943fe9102cfdae0d4495bd536b4 \
-    --hash=sha256:bc1ff2ff88dbfacefde4ddde471d1417d3b304e8df103a7a9437d47269201bf4
+# Code: https://github.com/pallets/jinja
+# Changes: http://jinja.pocoo.org/docs/2.10/changelog/
```
jwhitlock: Several changes that will impact Kuma
- urlize now sets “rel noopener” by default.
- with and autoescape tags are built-in
- Added a (default on?) trimmed modifier to ``{% trans %}``

kuma/settings/common.py
```diff
@@ -622,9 +622,7 @@ def _get_locales():
             'extensions': [
                 'jinja2.ext.do',
                 'jinja2.ext.loopcontrols',
-                'jinja2.ext.with_',
```
jwhitlock: In TravisCI, the locales are built with the current Docker image, which means it is breaking on this change that requires the updated requirements are used. I'm going to add a new commit to change the ``locale`` build to act more like the ``py27``, so the updated requirements will be used.

jwhitlock:
TravisCI passes, and I've verified that the extracted strings are the same, except for some line numbers. Thanks again @Alig1493!

Alig1493:
Thanks a lot for your fast and insightful review and for merging the PR.

PR #4748, opened 2018-04-23
[bug 1429933] remove depreciated CELERYD_LOG_LEVEL

mashrikt:
Removed `CELERYD_LOG_LEVEL` from `kuma/settings/common.py `
Used `--loglevel` flag in `docker-compose.yml` command instead.

This does not completely remove the `RemovedInDjangoXXXWarnings`, since I haven't added the SessionAuthenticationMiddleware. [It is no longer required in Django 1.10 and Session verification is enabled regardless.](https://docs.djangoproject.com/en/2.0/releases/1.10/)

> Session verification is enabled regardless of whether or not 'django.contrib.auth.middleware.SessionAuthenticationMiddleware' is in MIDDLEWARE_CLASSES. SessionAuthenticationMiddleware no longer has any purpose and can be removed from MIDDLEWARE_CLASSES. It’s kept as a stub until Django 2.0 as a courtesy for users who don’t read this note.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2.1'
 services:
   worker: &worker
     image: quay.io/mozmar/kuma_base
-    command: ./manage.py celery worker --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
+    command: ./manage.py celery worker --loglevel=INFO --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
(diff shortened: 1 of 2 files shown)


jwhitlock:
Thanks @mashrikt. I prefer small PRs with a single topic, so it is OK if this doesn't clear all the deprecation warnings. I believe there are more generated if you run the tests or other management commands, but it has been a while since I've done that.

Production doesn't use the development ``docker-compose.yml``, so I'll need to make a similar change to the production config before merging this.

mashrikt:
Thanks @jwhitlock 
Do let me know if there is anything I can do.

mashrikt:
Thanks!

PR #4751, opened 2018-04-24
bug 1429933: Enable session verification

jwhitlock:
Session verification, which includes invalidating sessions after password changes, is enabled by default in Django 1.10. We don't use passwords, but session verification can still used, and enabling the
middleware will let us detect issues before 1.11 is used.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -499,6 +499,7 @@ if not MAINTENANCE_MODE:
 
 MIDDLEWARE_CLASSES += (
     'django.contrib.auth.middleware.AuthenticationMiddleware',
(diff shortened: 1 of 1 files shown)


kuma/settings/common.py
```diff
@@ -499,6 +499,7 @@ def _get_locales():
 
 MIDDLEWARE_CLASSES += (
     'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
```
safwanrahman: Can you please add a todo to remove it after django 1.11 upgrade?

PR #4752, opened 2018-04-24
bug 1456165: switch to session-based CSRF tokens

escattone:
**If accepted, this PR eliminates the need for https://github.com/mozilla/kuma/pull/4744, so it should be reviewed/considered prior to that one.**

This PR provides Django-1.11-consistent session-based CSRF tokens for Django 1.8.19, and does so in a way that provides a simple transition for the Django 1.11 upgrade. The broader context of this PR is that it provides a possible solution to [bug 1456165](https://bugzilla.mozilla.org/show_bug.cgi?id=1456165) since it avoids cookie-based CSRF tokens and their implications when used with CloudFront.

* copied `django.middleware.csrf` from Django 1.11, which handles session-based CSRF tokens and token "salting", into `kuma.core.csrf` and modified it slightly to work within the context of Django 1.8 (this code should be removed when moving to Django 1.11)
* monkey-patched the Django 1.8 versions of `get_token`, `rotate_token`, and `CsrfViewMiddleware` with the updated versions from `kuma.core.csrf` so functions like `django.contrib.auth.login` will pick-up and use the updated version (this monkey-patching, which is odious, is only temporary, and should be removed when moving to Django 1.11)
* ~~created `kuma.core.middleware.CsrfViewMiddleware`, which inherits from `kuma.core.csrf.CsrfViewMiddleware`, but prevents it from creating session cookies for anonymous users (this remains after Django 1.11, but of course will inherit directly from Django's `CsrfViewMiddleware` rather than the copied version in `kuma.core.csrf` which will have been removed)~~
* created `kuma.core.middleware.SmartSessionMiddleware`, which inherits from `django.contrib.sessions.middleware.SessionMiddleware`, but prevents it from creating session cookies for anonymous users, which is essential if we want to use a CDN (this remains after Django 1.11)
* ~~created tests for `kuma.core.middleware.CsrfViewMiddleware`~~
* created tests for `kuma.core.middleware.SmartSessionMiddleware`
* added the `CSRF_USE_SESSIONS` setting to `kuma.settings.common` with a default value of `True` (this switches the storage of the CSRF token from the cookie to the session, and remains after the move to Django 1.11)
* removed `django-debreach`, since the `django.middleware.csrf` code copied from Django 1.11 takes care of "salting" the CSRF tokens
* switched back to using `django.core.context_processors.csrf` within the template context processors (it picks up the monkey-patched `get_token` from `kuma.core.csrf`)(this will remain in place after moving to Django 1.11)
* removed the check for the absence of the `Vary` header in three tests, since the ~~new `kuma.core.csrf.CsrfViewMiddleware`~~ `django.middleware.csrf.CsrfViewMiddleware` accesses the session on every request, which in turn triggers the addition of `Cookie` to the `Vary` header within Django's `django.contrib.sessions.middleware.SessionMiddleware` (this will NOT affect the Cloudfront CDN, since it ignores the `Vary` header)
* for the move to Django 1.11, some code needs to be deleted (as commented), but all of the rest (including configuration) remains the same

### DIFF
--- a/kuma/core/csrf.py
+++ b/kuma/core/csrf.py
@@ -1,0 +1,367 @@
+# TODO: Remove this file with Django 1.11
+"""
+Cross Site Request Forgery Middleware.
+
+This module provides a middleware that implements protection
+against request forgeries from other sites.
+
+NOTE: Copied from Django 1.11.12's django.middleware.csrf and slightly
+      modified to work within the context of Django 1.8.19.
+"""
+from __future__ import unicode_literals
+
+import logging
+import re
+import string
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.core.urlresolvers import get_callable
+from django.utils.cache import patch_vary_headers
+from django.utils.crypto import constant_time_compare, get_random_string
+from django.utils.encoding import force_text
+from django.utils.six.moves import zip
+from django.utils.six.moves.urllib.parse import urlparse
+
+logger = logging.getLogger('django.security.csrf')
+
+REASON_NO_REFERER = "Referer checking failed - no Referer."
+REASON_BAD_REFERER = "Referer checking failed - %s does not match any trusted origins."
+REASON_NO_CSRF_COOKIE = "CSRF cookie not set."
+REASON_BAD_TOKEN = "CSRF token missing or incorrect."
+REASON_MALFORMED_REFERER = "Referer checking failed - Referer is malformed."
+REASON_INSECURE_REFERER = "Referer checking failed - Referer is insecure while host is secure."
+
+CSRF_SECRET_LENGTH = 32
+CSRF_TOKEN_LENGTH = 2 * CSRF_SECRET_LENGTH
+CSRF_ALLOWED_CHARS = string.ascii_letters + string.digits
+CSRF_SESSION_KEY = '_csrftoken'
+
+
+def is_same_domain(host, pattern):
+    """
+    Return ``True`` if the host is either an exact match or a match
+    to the wildcard pattern.
+
+    Any pattern beginning with a period matches a domain and all of its
+    subdomains. (e.g. ``.example.com`` matches ``example.com`` and
+    ``foo.example.com``). Anything else is an exact string match.
+    """
+    if not pattern:
+        return False
+
+    pattern = pattern.lower()
+    return (
+        pattern[0] == '.' and (host.endswith(pattern) or host == pattern[1:]) or
+        pattern == host
+    )
+
+
+def _get_failure_view():
+    """
+    Returns the view to be used for CSRF rejections
+    """
+    return get_callable(settings.CSRF_FAILURE_VIEW)
+
+
+def _get_new_csrf_string():
+    return get_random_string(CSRF_SECRET_LENGTH, allowed_chars=CSRF_ALLOWED_CHARS)
+
+
+def _salt_cipher_secret(secret):
+    """
+    Given a secret (assumed to be a string of CSRF_ALLOWED_CHARS), generate a
+    token by adding a salt and using it to encrypt the secret.
+    """
+    salt = _get_new_csrf_string()
+    chars = CSRF_ALLOWED_CHARS
+    pairs = zip((chars.index(x) for x in secret), (chars.index(x) for x in salt))
+    cipher = ''.join(chars[(x + y) % len(chars)] for x, y in pairs)
+    return salt + cipher
+
+
+def _unsalt_cipher_token(token):
+    """
+    Given a token (assumed to be a string of CSRF_ALLOWED_CHARS, of length
+    CSRF_TOKEN_LENGTH, and that its first half is a salt), use it to decrypt
+    the second half to produce the original secret.
+    """
+    salt = token[:CSRF_SECRET_LENGTH]
+    token = token[CSRF_SECRET_LENGTH:]
+    chars = CSRF_ALLOWED_CHARS
+    pairs = zip((chars.index(x) for x in token), (chars.index(x) for x in salt))
+    secret = ''.join(chars[x - y] for x, y in pairs)  # Note negative values are ok
+    return secret
+
+
+def _get_new_csrf_token():
+    return _salt_cipher_secret(_get_new_csrf_string())
+
+
+def get_token(request):
+    """
+    Returns the CSRF token required for a POST form. The token is an
+    alphanumeric value. A new token is created if one is not already set.
+
+    A side effect of calling this function is to make the csrf_protect
+    decorator and the CsrfViewMiddleware add a CSRF cookie and a 'Vary: Cookie'
+    header to the outgoing response.  For this reason, you may need to use this
+    function lazily, as is done by the csrf context processor.
+    """
+    if "CSRF_COOKIE" not in request.META:
+        csrf_secret = _get_new_csrf_string()
+        request.META["CSRF_COOKIE"] = _salt_cipher_secret(csrf_secret)
+    else:
+        csrf_secret = _unsalt_cipher_token(request.META["CSRF_COOKIE"])
+    request.META["CSRF_COOKIE_USED"] = True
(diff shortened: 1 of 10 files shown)


kuma/core/csrf.py
```diff
@@ -0,0 +1,373 @@
+# TODO: Rmove this file with Django 1.11
```
stephendonner: Uber-nit: s/Rmove/Remove

escattone: Oops, thanks @stephendonner!

escattone:
Amended the commit to fix a misspelled comment.

jwhitlock:
My reading is that CloudFront [does look at the ``Vary:`` header](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#ResponseCustomRemovedHeaders):

> * ``Vary`` – Note the following:
>   * If you configure your origin to include either ``Accept-Encoding`` or ``Cookie`` in the Vary header, CloudFront includes the values in the response to the viewer.
 >   * If you configure CloudFront to forward a whitelist of headers to your origin, and if you configure your origin to return the header names to CloudFront in the ``Vary`` header (for example, ``Vary:Accept-Charset,Accept-Language``), CloudFront returns the Vary header with those value to the viewer.
>   * If you configure your origin to include any other values in the ``Vary`` header, CloudFront removes the values before returning the response to the viewer.

Which I think is what we want (return the session cookie to the user), but if we hand out a session cookie to everyone, I think that's the end of effective caching, or will mean multiple people getting the same session.

I'll continue reviewing, but we should plan on some time on staging to explore CF's behavior with this change and multiple users.

safwanrahman:
I dont know if its feasable, but is it possible to add a functional test for checking this? This is the only way we can automatically check while the site is behind CDN!

kuma/core/csrf.py
```diff
@@ -0,0 +1,373 @@
...
+from django.core.urlresolvers import get_callable
+from django.utils.cache import patch_vary_headers
+from django.utils.crypto import constant_time_compare, get_random_string
+from django.utils.encoding import force_text
+from django.utils.six.moves import zip
+from django.utils.six.moves.urllib.parse import urlparse
+
+logger = logging.getLogger('kuma.core.csrf')
+
+REASON_NO_REFERER = "Referer checking failed - no Referer."
```
safwanrahman: This should be under `ugettext` so it can be localizable!

safwanrahman:
Practically, there should be very less chance for a CSRF permission denied because browsers are cleaver enough to mitigate it with `CORS` headers.
Can we keep the log of CSRF permission denied in log? So we can investigate this problem if it keeps present!

safwanrahman:
After applying the patch, I have tried to login into the admin pannel in my local environment and got permission denied! I think this should block the PR!

escattone:
@safwanrahman Thanks, me too, I'm working on that right at this moment. I'll have a fix soon.

jwhitlock:
> I dont know if its feasable, but is it possible to add a functional test for checking this? This is the only way we can automatically check while the site is behind CDN!

The bug is due to the interaction of the CloudFront CDN, the Django backend, and multi-tab browsing and editing on MDN. I don't believe that automated functional testing will be a huge help here.

> Practically, there should be very less chance for a CSRF permission denied because browsers are clever enough to mitigate it with CORS headers.

Can you support this claim?

> Can we keep the log of CSRF permission denied in log? So we can investigate this problem if it keeps present!

Permission denied issues are logged, as you can see in the code in this PR. I wasn't able to get the log messages locally, but I think that's because the logger name changed. The reasons for the Permission Denied is a mismatch of the token, ``REASON_BAD_TOKEN``.

kuma/core/csrf.py
jwhitlock: These text strings are not localized in Django, and we shouldn't localize them here either. They are sent to the logs, which should be in English regardless of the requester's language, and they shouldn't be printed in the template, which could leak security information.

escattone:
What a rabbit hole! I submitted a new commit which improves the monkey-patching and fixes the admin login permission-denied issue (it also changes the logger name to `django.security.csrf` as requested). However, there remains a flaw in the admin login that would also exist even after the switch to Django 1.11. The flaw is that when the admin login page is simply loaded (the user doesn't have to log in), it saves a session cookie. This is because the admin application uses a `csrf_protect` decorator that's based on the vanilla `django.middleware.csrf.CsrfViewMiddleware`, which creates session cookies for anonymous users, instead of our `kuma.core.middleware.CsrfViewMiddleware` which does not. The fix is not obvious, but I don't know if that flaw is acceptable.

kuma/core/csrf.py
safwanrahman: Oh! Sorry, my bad!

escattone:
> However, there remains a flaw in the admin login that would also exist even after the switch to Django 1.11. The flaw is that when the admin login page is simply loaded (the user doesn't have to log in), it saves a session cookie. This is because the admin application uses a csrf_protect decorator that's based on the vanilla django.middleware.csrf.CsrfViewMiddleware, which creates session cookies for anonymous users, instead of our kuma.core.middleware.CsrfViewMiddleware, which does not. The fix is not obvious, but I don't know if that flaw is acceptable.

The fog cleared for me last night! I have a nice, clean solution to this that I should have seen before. I'll have a new commit later this morning.

safwanrahman:
> The bug is due to the interaction of the CloudFront CDN, the Django backend, and multi-tab browsing and editing on MDN. I don't believe that automated functional testing will be a huge help here.

The automated functional test runs after every deploy towards stage and procuction. So this must have caught the issue!

> Can you support this claim?
by using CORS header, browser block any request that come from other domain, and MDN Edit page has this CORS policy enables. So I believe that this must have mitigate the CSRF. https://mixmax.com/blog/modern-csrf

> Permission denied issues are logged, as you can see in the code in this PR. I wasn't able to get the log messages locally, but I think that's because the logger name changed. The reasons for the Permission Denied is a mismatch of the token, REASON_BAD_TOKEN.

I could not get any log in my console therefore I have asked about it. Seems like it got fixed!

escattone:
This is ready for review again. I re-based into a single commit since the changes were significant enough from the original that it would have been confusing to review as a patch. I made two key changes:
* decided not to sub-class `django.middleware.csrf.CsrfViewMiddleware` but simply use it as is (the monkey-patched version for now until we move to Django 1.11), since that's what other apps like the Django admin will use
* prevented the creation of a session cookie for anonymous users in a new custom `kuma.core.middleware.SmartSessionMiddleware` instead, since that will catch all users of `CsrfViewMiddleware`, like Django admin 

I did not address the issue that since we're using `CsrfViewMiddleware` with session-based tokens, we access the session on every request, which in turn causes Django's `SessionMiddleware` to patch the `Vary` header with `Cookie` on every request, but I think that's an issue with Django's system. I suspect it's going too far in many cases, and we may want to address it, but it will not affect the caching done by the CloudFront CDN. CloudFront only caches based on each behavior's configured options for headers, query parameters and cookies for example, and does not consider the `Vary` header.

I did test this on my local dev environment, but I will test more thoroughly again tomorrow morning.

escattone:
This is maddening, and I'm getting the feeling that this is a lost cause. Now the admin login is broken. I see now that its login page depends upon saving a CSRF token in an anonymous session, which means creating a `sessonid` cookie when loading the admin login page, so that same token can be grabbed server-side when the login POST is submitted. The problem is that the session, when loading the login page, is exactly what I am blocking in `SmartSessionMiddleware` (it contains only a single piece of data, a CSRF token, and its session key is None). This is not a problem for GitHub login, since it uses the OAuth process instead.

escattone:
For what it's worth, I submitted (in a separate commit) a patch the prevents the blanket patching of the `Vary` header with `Cookie` for anonymous users. We can use it or not, as desired. I just wanted to save the work as an option to consider if we continue down this path (which I'm starting to doubt).

escattone:
We may want to revisit this, but for I'm closing this for now.

PR #4754, opened 2018-04-26
bug 1451261: Translate action button text

jwhitlock:
The process that extracts translations doesn't know JS. Change to call ``gettext`` with the string directly, rather than by variable.  Without this change, these strings ('Revert', 'View Page', etc) would be removed from Pontoon and https://github.com/mozilla-l10n/mdn-l10n/ the next time strings are extracted.

The code could be further simplified to eliminate ``actionsArray`` and the ``for`` loop, but I'm not sure if this was the first step in a longer refactor, such as passing the ``actionsArrays`` to the function.

### DIFF
--- a/kuma/static/js/dashboard.js
+++ b/kuma/static/js/dashboard.js
@@ -47,7 +47,7 @@
                 pageButtons.appendChild(getListItem({
                     id: 'revert',
                     href: controlURLs.revertURL,
-                    text: gettext(actionsArray[i]),
+                    text: gettext('Revert'),
                     icon: 'undo'
                 }));
             }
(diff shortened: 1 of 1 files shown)


safwanrahman:
What about adding `gettext` [here](https://github.com/mozilla/kuma/pull/4754/files#diff-a92790dcc8bb968d17b56e184f5879cdR41) in the sting?

jwhitlock:
That would allow the text to be extracted for translation, but would break the function logic if the user wasn't on the English page (for example, French users would get no buttons).

schalkneethling:
@jwhitlock Aha, I had no idea. Thanks for the PR ~ r+

PR #4756, opened 2018-04-27
fix: Adds titles to social icons and reoptimise svg

schalkneethling:
As the social icons are most often used a link targets, we need these to have discernible text for accessibility.


jinja2/includes/icons/social/facebook.svg
```diff
@@ -1,3 +1,6 @@
 <svg class="icon icon-facebook" xmlns="http://www.w3.org/2000/svg" width="16" height="28" viewBox="0 0 16 28" aria-hidden="true">
+    <title>
+        {{ _(&apos;Facebook&apos;) }}
```
schalkneethling: I am assuming this is going to be a problem for gettext?

schalkneethling: Oh, but these do not need to be translated. so I do not need to wrap them right? This is done by [SVGO](https://github.com/svg/svgo/) and could present a problem down the line. I will need to raise this as an issue on the repo.

jwhitlock: Yes, it will be a problem if ``'Facebook'`` is translated to ``&apos;Facebook&apos;``. Maybe double-quotes are untranslated? Maybe we need to post-process SVGO output?

jwhitlock: In my opinion, Facebook and other brand names can go untranslated. However, we do currently mark some for translation ([Facebook in Pontoon](https://pontoon.mozilla.org/es/mdn/LC_MESSAGES/django.po/?search=facebook&string=67799)). Some locale teams (Spanish, Chinese) leave brand strings untranslated, others translated them (ফেসবুক in Bengali)

schalkneethling: Hmmm, that is a problem that will need a fix. Do you feel that for this PR it is safe to just not wrap them? As mentioned I will log an issue against SVGO and see if there is a workaround or, if a fix can be implemented.

jwhitlock: It seems easy enough to me to change the line to:

```
<title>{{ _("Facebook") }}</title>
```

It sounds like with the current toolchain the quick change is:

```
<title>Facebook</title>
```

That's a positive change that helps with accessibility, but not with localization. Further, there are cases like the GitHub icon that sometimes mean [Sign in with GitHub](https://github.com/mozilla/kuma/blob/5fb45f9ae67a4e89e0b68a9571c3c6d8b9864f94/kuma/users/jinja2/socialaccount/snippets/login_service_icon.html) and other times mean [Here's our GitHub project](https://github.com/jwhitlock/kuma/blob/e04bd03b9007b396fa0550508b9f8d2e51f4c714/jinja2/base.html#L143)

I think we need to work out a rendering that:

1) Allows SVGO to generate minimized SVG icons.
2) Allows templates to pick a localized title for [accessible SVG](https://www.sitepoint.com/tips-accessible-svg/).

I think it would be OK to split this into two PRs - add some non-localized titles today, and I can write some Jinja2 helpers to customize titles at render time.

jinja2/includes/icons/web-service/jsfiddle.svg
```diff
@@ -1,3 +1,6 @@
 <svg class="icon icon-jsfiddle" xmlns="http://www.w3.org/2000/svg" width="32" height="28" viewBox="0 0 32 28" aria-hidden="true">
+    <title>
+        {{ _(&apos;jsFiddle&apos;) }}
```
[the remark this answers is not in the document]
schalkneethling: Doh! Forgot about that one :(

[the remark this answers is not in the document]
schalkneethling: Done.

jwhitlock:
The Python tests are failing here:

https://github.com/mozilla/kuma/blob/8a7a4ee49aee84dcb9e6ff9f365ce80c0b2876f6/kuma/wiki/tests/test_views.py#L636

They fail because there are now multiple title elements on the page, and the combined title text ``One | MDN Twitter Github Github Facebook Instagram`` no longer matches. The test passes if the selector is more specific:

```
            ok_(page.find('head > title').text() in aught_titles)
```

schalkneethling:
@jwhitlock updated. Thanks!

PR #4757, opened 2018-04-27
bug 1456165: remove csrfmiddlewaretoken from wiki.document view responses

escattone:
This PR removes `csrfmiddlewaretoken` from the responses of all endpoints that require caching in the CloudFront CDN, which ends up being just the `wiki.document` endpoint. This provides a possible solution to bug 1456165, since it avoids setting a new `csrftoken` cookie on `wiki.document` page loads due to the fact that the CDN behavior can't forward the existing one (it can't forward the existing `csrftoken` cookie without negating caching altogether for that behavior, since CloudFront does not provide separate configuration for caching and forwarding).

### DIFF
--- a/jinja2/includes/login.html
+++ b/jinja2/includes/login.html
@@ -11,7 +11,6 @@
                     <li><a href="{{ user.get_absolute_url() }}">{{ _('View profile') }}</a></li>
                     <li><a href="{{ url('users.user_edit', username=user.username) }}">{{ _('Edit profile') }}</a></li>
                     <li><form class="login-form" action="{{ url('account_logout') }}" method="post">
-                        {% csrf_token %}
                         <input name="next" type="hidden" value="{{ next_url }}">
                         <button class="logout button link" type="submit">{{ _('Sign out') }}</button>
                     </form></li>
--- a/kuma/users/tests/test_templates.py
+++ b/kuma/users/tests/test_templates.py
@@ -169,9 +169,10 @@ class AllauthGitHubTestCase(UserTestCase, SocialTestMixin):
         # Check next url is provided as input field
         next_input = form.children("input[name='next']")
         assert next_input.val() == home_url
-        # Check CSRF is added
+        # Ensure CSRF protection has not been added, since it creates problems
+        # when used with a CDN like CloudFront (see bugzilla #1456165).
         csrf_input = form.children("input[name='csrfmiddlewaretoken']")
-        assert csrf_input.val() is not None
+        assert not csrf_input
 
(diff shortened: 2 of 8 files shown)


jwhitlock:
I'm trying to decide if we should always remove CSRF, or just for anonymous users.  If we try to restrict this to anonymous users:

1. Wiki doc behavior would change to cache on the CSRF token and session cookies
2. Anon requests would get neither the CSRF or the session cookie by default.
3. Logged-in users would get both a session and CSRF cookie.

This might work going forward. However, if an anon user that visited MDN in the last year shows up with a CSRF cookie, then the CDN would give them a per-user cache. We'd need to add a middleware that actively deletes CSRF cookies for anonymous users, and maybe add it to logout.

OK, now for the risks:

* There's a known attack pattern where an attacker 1) logs out a user, 2) shows them a look-alike login form, 3) logs in the user with a man-in-the-middle attack, getting the user's account credentials. The folks at https://security.stackexchange.com/q/62769/177031 are mostly concerned about this. Our POST logout mitigates this attack - the POST would have to come from developer.mozilla.org, and include the session cookie. Further, they would have to deflect login from GitHub to somewhere else, and OAuth2 gives us a few protections here.
* A user could force POSTs to logout users as they visit MDN.  Again, this requires the POST to come from developer.mozilla.org, and I'm not sure to what end they would be doing that. Some people suggest a denial-of-service attack, but for MDN it would be a denial-of-editing attack. It seems they would use the ability to do in-page POSTs for other purposes.
* A user could force POSTs to watch a page. Again, I'm not creative enough to see where that attack would be going. I do think we could present a login link to anon users and a watch form to logged in users, but it is probably OK.

OK, that's my preliminary thoughts. Now to actually try to code out, read it closely, etc.

escattone:
@jwhitlock That's a promising idea, restricting this approach to anonymous users. That would require:
- adding `csrftoken` to the list of cookies that are forwarded and used for caching to the `*/docs/*` CDN behavior
- modifying the templates used for document pages such that the `csrf_token` is only called for logged-in users (not strictly necessary)
- adding middleware that would modify the response such that the `csrftoken` cookie is always deleted for anonymous users (critical, since we're adding `csrftoken` to the list of cookies forwarded/cached-on, this ensures that the CDN remains useful, that all anonymous users share the same cache)

escattone:
Just for the record, @jwhitlock and I discussed the idea (via a Vidyo meeting) of restricting this approach to anonymous users, and realized that we'd have to handle the special case of Admin login, which depends on an anonymous user getting a CSRF token.

PR #4761, opened 2018-05-03
bug 1429933: Update get_form to match upstream class

jwhitlock:
Update django-allauth to the current version, before updating our customization of the ``SignupView``.

In Django 1.8, ``django.views.generic.edit.FormMixin``'s method
[get_form()](https://docs.djangoproject.com/en/1.8/ref/class-based-views/mixins-editing/#django.views.generic.edit.FormMixin.get_form) changed so that ``form_class`` was no longer a required argument, but instead calls ``get_form_class()`` if it has the default value of ``None``.

In Django 1.8, this removes a deprecation warning. In Django 1.10, this fixes a runtime error.

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -466,7 +466,7 @@ class SignupView(BaseSignupView):
     """
     form_class = SignupForm
 
-    def get_form(self, form_class):
+    def get_form(self, form_class=None):
         """
         Returns an instance of the form to be used in this view.
         """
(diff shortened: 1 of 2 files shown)


jwhitlock:
Failing code, needs more work ☹️

jwhitlock:
django-allauth 0.35.0 dropped support for Django 1.8, but it wasn't in the release notes. I've opened https://github.com/pennersr/django-allauth/pull/2016 and updated this PR to django-allauth 0.34.0. 🤞

PR #4766, opened 2018-05-04
bug 1429933: Rewrite locale middleware as extension of Django LocaleMiddleware

jwhitlock:
Replace ``LocaleURLMiddleware`` with one that extends Django's [LocaleMiddleware](https://docs.djangoproject.com/en/1.8/ref/middleware/#module-django.middleware.locale), and update tools to be closer to Django's default [localization framework](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#module-django.conf.urls.i18n).

One large changes is that ``urlpatterns`` are split into ones with a language prefix (``lang_urlpatterns``), and ones without. Django 1.8 still requires the ``LANGUAGE_URL_IGNORED_PATHS`` hack, but Django 1.11 can better determine if a given URL should have a locale prefix.

Django's language codes are lowercase (``en-us``), while Kuma's language codes use uppercase for the region (``en-US``). In order to keep the same URLs, several changes are required, such as replacing ``LocaleRegexURLResolver`` and ``i18n_patterns``.

Kuma's previous locale tools (``reverse``, Jinja's ``wiki_url``) defaulted to no locale prefix. With the new framework, the default is the currently active locale, and it is retained in ``request.path_info``. These changes require some adjustments, such as in the ``DocumentZoneMiddleware``.

Some Kuma behaviours are not part of the Django framework. The ``?lang`` querystring handling is moved to a new ``LangSelectorMiddleware``, and redirects to non-canonical locales (such as ``/en/`` to ``/en-US/``) would be 404s with Django, but are handled in the new ``LocaleStandardizerMiddleware``. Some other differences, such as removing the ``Vary: Accept-Language`` header, are handled in the overridden ``LocaleMiddleware`` class.

Some tests are also refactored, as a preparatory step for removing the old Kuma locale framework. There is still quite a bit to remove, especially around tests, but this PR feels big enough already.


kuma/core/middleware.py
```diff
@@ -8,82 +7,203 @@
...
-        if lang in dict(settings.LANGUAGES):
-            # Blank out the locale so that we can set a new one. Remove lang
-            # from the query params so we don't have an infinite loop.
-            prefixer.locale = ''
-            new_path = prefixer.fix(prefixer.shortened_path)
-            query = dict((smart_str(k), v) for
+        """Redirect if ?lang query parameter is valid."""
+        query_lang = request.GET.get('lang')
+        if query_lang not in dict(settings.LANGUAGES):
+            return
```
safwanrahman: Would be helpful if possible to explicitly mention the type of return.
Maybe `None`?

jwhitlock: ``return`` and ``return None`` have [the same effect](https://stackoverflow.com/a/15300671/10612), so I usually use the shorter one. If it makes the code clearer, I can use ``return None`` or add a comment.

jwhitlock:
Sorry, I was cleaning up branches, thought this one was merged.

kuma/core/middleware.py
```diff
@@ -8,82 +7,204 @@
...
+        script_prefix = urlresolvers.get_script_prefix()
+        expected_prefix = '%s%s/' % (script_prefix, language)
+        full_path = request.get_full_path()  # Includes querystring
+        old_path = urlsplit(full_path).path
+        if full_path.startswith(expected_prefix):
+            new_prefix = '%s%s/' % (script_prefix, query_lang)
+            new_path = old_path.replace(expected_prefix, new_prefix, 1)
+        else:
+            new_prefix = '%s%s/' % (script_prefix, query_lang)
+            new_path = old_path.replace(script_prefix, new_prefix, 1)
```
escattone: Nit. Can save a line since `new_prefix` is shared by both cases of the `if-else`:
```python
new_prefix = '%s%s/' % (script_prefix, query_lang)
if full_path.startswith(expected_prefix):
    new_path = old_path.replace(expected_prefix, new_prefix, 1)
else:
    new_path = old_path.replace(script_prefix, new_prefix, 1)
```

kuma/core/middleware.py
```diff
@@ -8,82 +7,204 @@
...
+        """Redirect if ?lang query parameter is valid."""
+        query_lang = request.GET.get('lang')
+        if query_lang not in dict(settings.LANGUAGES):
+            # Invalid language requested, don't redirect
+            return
+
+        # Check if different language is embedded in URL
+        dj_language = translation.get_language_from_request(
+            request, check_path=True)
+        language = django_language_code_to_kuma(dj_language)
```
escattone: From my tests and reading of the Django code, it seems to me that `translation.get_language_from_request` will always return the Kuma version of the language code, `en-US` instead of `en-us`, since the Django code uses `settings.LANGUAGES`, which contains the Kuma language codes. So the call to `django_language_code_to_kuma` doesn't seem necessary, unless it's not necessary now but will be in the context of future updates for the Django 1.11 upgrade?

kuma/core/middleware.py
```diff
@@ -8,82 +7,204 @@
...
+        # Call Django's LocaleMiddleware to detect the language, activate it,
+        # and set request.LANGUAGE_CODE, based on Django's language code
+        response = super(LocaleMiddleware, self).process_request(request)
+
+        # Replace Django's language code w/ Kuma's, if needed
+        dj_language_code = request.LANGUAGE_CODE
+        kuma_language_code = django_language_code_to_kuma(dj_language_code)
+        if dj_language_code != kuma_language_code:
+            translation.activate(kuma_language_code)
+            request.LANGUAGE_CODE = kuma_language_code
```
escattone: This whole chunk, lines 151-156, doesn't seem necessary to me since the underlying `django.utils.translation.get_language_from_request` that `DjangoLocaleMiddleware` uses will always return the Kuma language code (for example, the request path `/zh-cn/...` returns `zh-CN` as the language code). It seems to me that this is the case because the underlying `django.utils.translation.get_supported_language_variant` ends-up always called with `strict=False` by `django.utils.translation.get_language_from_path`, and that ensures that `zh-cn` picks up the Kuma `zh-CN` via the generic language code. The only odd case that I can see is that `sr-latn` becomes `sr` while `sr-Latn` remains `sr-Latn`. As before though, maybe I'm missing that this will be needed in the future.

kuma/core/tests/test_locale_middleware.py
```diff
@@ -1,74 +1,79 @@
...
-        response = self.client.get('/', follow=True,
-                                   HTTP_ACCEPT_LANGUAGE='fr-fr')
-        self.assertRedirects(response, '/fr/', status_code=302)
-        assert_shared_cache_header(response)
+@pytest.mark.parametrize('accept_language,locale', PICKER_CASES)
+def test_locale_middleware_picker(accept_language, locale, client, db):
+    '''The LocaleMiddleware picks locale from the Accept-Language header.'''
+    response = client.get('/', HTTP_ACCEPT_LANGUAGE=accept_language)
+    assert response.status_code == 302
+    assert response['Location'] == 'http://testserver/%s/' % locale or 'en_US'
```
escattone: I think this should have `'en-US'` for the case when `locale` is "falsy" instead of `en_US`?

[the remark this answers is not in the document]
escattone: The `pt` and `zh-Hant` cases make me wonder if the new `LocaleMiddleware` is going to have to handle the `settings.LOCALE_ALIASES` instead of counting on handling them in `LocaleStandardizerMiddleware`?

kuma/core/tests/test_urlresolvers.py
```diff
@@ -2,72 +2,6 @@
 
 from django.conf import settings
 
-from kuma.core.tests import eq_, KumaTestCase
```
escattone: ❤️ Feels so good!

kuma/core/urlresolvers.py
```diff
@@ -105,56 +64,3 @@ def split_path(path):
...
-        else:
-            check_path = path + '/'
-        if not check_path.startswith(settings.LANGUAGE_URL_IGNORED_PATHS):
-            # Set locale according to order
-            locale = self.locale or self.get_language()
-            url_parts.append(locale)
-
-        url_parts.append(path)
-
-        return '/'.join(url_parts)
```
escattone: ❤️ So nice to say goodbye to the prefixer code.

kuma/urls.py
```diff
@@ -1,6 +1,7 @@
 from decorator_include import decorator_include
 from django.conf import settings
 from django.conf.urls import include, url
+# from django.conf.urls.i18n import i18n_patterns
```
escattone: Can this be removed, or is it a reminder?

kuma/wiki/tests/test_middleware.py
```diff
@@ -158,10 +158,18 @@ def test_zone_url_ends_with_slash(self):
...
+    with the middleware that attempts to redirect requests to the right
+    location. See bug 1267197 for some of the past analysis and discussion.
+
+    With the Mozilla zamboni/amo-derived middleware, there were sometimes two
+    redirects:
+    * A 302 temporary redirect to add a locale to a locale-less path.
+    * A 301 permanent redirect to remap a wiki path to a zoned path.
 
+    With the Django 1.8-derived middleware, these are combined into a single
+    302 temporary redirect, which may also help with retiring DocumentZones.
```
escattone: Nice!

kuma/core/middleware.py
jwhitlock: Nice catch!

kuma/urls.py
jwhitlock: Nope, it can be removed. I tried many, many things, and didn't clean up completely 😬

kuma/core/tests/test_locale_middleware.py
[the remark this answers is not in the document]
jwhitlock: Hmm. The ``or 'en-US'`` was for the case where the client requests unknown locales:

https://github.com/mozilla/kuma/blob/25ddd313d7b0c1b3dd58947fe02cd312da94c110/kuma/core/tests/test_urlresolvers.py#L46-L49

``get_best_language`` returns ``False``, and the redirect URL is the default ``en-US``. That's what I get for trying to be clever...

jwhitlock:
The tests are failing (correctly) with these latest commits. The failures are matching ``Accept-Language`` headers to the desired locales.  The redirect is created in ``LocaleMiddleware.process_response``, but the accepted language is picked earlier in [get_supported_language_variant](https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/utils/translation/trans_real.py#L432-L464), called by [get_language_from_request](https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/utils/translation/trans_real.py#L485-L530).

``get_supported_language_variant`` searches ``settings.LANGUAGES`` with Django language codes (``en-us``, ``zh-hant``, etc). If it doesn't find a locale-specific code, it will match a generic version (``fr-fr`` will match ``fr``). If that fails as well, it will match the first locale-specific code (``pt-pt`` won't match ``pt-PT`` or ``pt``, and picks ``pt-BR`` because it is first).  This works poorly with MDN's locale codes.

The specific failures:

* ``fr-FR, de;q=0.5`` gets ``fr``, not ``de``. ``get_supported_language_tests ``fr`` as a possible code, and selects it. I think I'm OK with this changing, it seems like an improvement.
* ``pt, fr;q=0.5`` gets ``pt-BR``, not ``pt-PT``, because it comes first in the alpha-sorted ``settings.LANGUAGES``.  We'd want ``pt-PT``, like it says in ``LOCALE_ALIASES``.
* ``zh-Hant, fr;q=0.5`` gets ``zh-CN``, not ``zh-TW``, due to fall-backs and alpha-ordering. We'd prefer the ``LOCALE_ALIASES`` choice of ``zh-Hant`` mapping to ``zh-TW``.

Two options I'm considering:
* Let Django have ``settings.LANGUAGES``, add a ``settings.MDN_LANGUAGES`` for mapping to and from for URLs.
* Instead of wrapping ``LocaleMiddleware.process_requests``, bring in the Django 1.8 code and create our own versions of ``get_supported_language_variant`` and ``get_language_from_request``.

I'll try these out - it may take a while for the next change. I'm considering pulling out the first commit (refactoring tests, with fixes from review), since it seems like a standalone fix now.

jwhitlock:
I thought the headless tests were failing because the pages aren't in the sample database, but I believe a similar issue (not using ``LOCALE_ALIASES`` in determining locales) would cause the problem. I get 13 headless failures. Some are redirects to ``/en-US/`` rather than immediate 404s:

* ``/index.php``
* ``/index.php?title=Special:Recentchanges&feed=atom``
* ``/index.php?title=En/HTML/Canvas&revision=110``
* ``/index.php?title=En/HTML/Canvas&revision=111``
* ``/patches``
* ``/patches/foo``
* ``/web-tech``
* ``/web-tech/feed/atom/``
* ``/css/wiki.css``
* ``/css/base.css``

These also add the ``en-US``, but should be caught by ``LOCALE_ALIASES``:

* /cn/docs/Talk:Kakurady
* /zh_cn/docs/Web/API/RTCPeerConnection/addTrack
* /zh_tw/docs/AJAX

At first I thought it was an issue with the pages not being in the sample database, but that doesn't help either.

I think all 13 should continue passing.

Thanks for the review, @escattone!

jwhitlock:
"Let Django have ``settings.LANGUAGES``" looks like a dead-end. The Django language codes end up in the URL patterns, so it changes the in-page content.

I'll try a custom ``LocaleMiddleware.process_request`` tomorrow.

jwhitlock:
OK, that gets tests passing. There's some follow-on tasks I'd like to do in this PR:

* [ ] Get coverage of ``kuma/core/middleware.py`` and ``kuma/core/i18n.py`` to 100%, by adding tests and removing branches
* [ ] Rearrange ``kuma/urls.py`` to group and combine non-locale and locale-prefixed patterns

I think this will be ready for review again on Monday.

jwhitlock:
I want to more fully explain the regression, because I did the research.

Previous to this PR, if a user had the Accept-Language header ``'fr-FR, de;q=0.5'``, then they would get ``de``. The prose version of the algorithm is "Look for an exact match for each term, then look for generic fallbacks for each term":

* ``fr-FR`` has no exact match
* ``de`` has an exact match, and wins

This behavior was added in https://github.com/mozilla/kuma/pull/930 as a fix for [bug 647796](https://bugzilla.mozilla.org/show_bug.cgi?id=647796). This wasn't a consensus decision. @jsocol [argued](https://bugzilla.mozilla.org/show_bug.cgi?id=647796#c23):

> if you say "en-GB;q=0.8,fr;q=0.7", we, and I think most, would assume that English is higher preference than French, especially given that en-GB and en-US are, except slang, mutually intelligible.

The PR author @GPHemsley [responded](https://bugzilla.mozilla.org/show_bug.cgi?id=647796#c28):

>  But I am arguing that such an assumption would be incorrect. The user has clearly expressed the preferences they want to express. If you don't have English as spoken in the United Kingdom, they want French. If they had no preference as to the variety of English they received, they would have used 'en' instead of 'en-GB'.

> It is much more a game of whackamole to assume that languages that share a language subtag are as mutually intelligible as 'en-US' and 'en-GB' happen to be. And, in fact, it is that 'slang' (AKA localization) that you refer to that often spells the difference between such locales in the first place. If you, as an 'en-US' speaker, were forced to use a locale that had elevators labeled as lifts and curbs labeled as kerbs and jails labeled as gaols, you'd be disoriented, too.

The arguments stopped there. The PR was submitted, there was no further discussion on this corner case. Code wins!

The new behavior, from Django's [get_language_from_request](https://github.com/django/django/blob/a5a2ceeb453b86caaf01687342984f372428beb8/django/utils/translation/trans_real.py#L501-L511), is "For each term, look for an exact match and then a generic fallback":

* ``fr-FR`` has no exact match
* ``fr``, the generic variant, has a match, and wins

This is not be technically correct, and someone that has gone through the trouble of setting their language preference to ``fr-FR, de;q=0.5`` will get ``fr`` and maybe be disappointed.

While Django's algorithm may not be technically correct for this case, I'm going to say it is better to have a simple algorithm that matches Django then to be correct. If I'm wrong, I hope someone files a bug and we can connect the case to a visitor. I searched Django's issue tracker, and I can't find any complaints there.

In general, I'd like to retire a lot of these redirects and special cases. Maybe someday we can be just ``en``, or even ``en-us``...

kuma/core/i18n.py
```diff
@@ -1,4 +1,197 @@
...
+
+    lang_code = request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+
+    try:
+        return get_supported_language_variant(lang_code)
+    except LookupError:
+        pass
+
+    accept = request.META.get('HTTP_ACCEPT_LANGUAGE', '')
+    for accept_lang, unused in parse_accept_lang_header(accept):
```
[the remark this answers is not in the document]
jwhitlock: No need to rely on your memory, you can read the source code:

https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/utils/translation/trans_real.py#L738-L762

kuma/core/i18n.py
```diff
@@ -1,4 +1,197 @@
...
+        if not language_code_re.search(accept_lang):
+            continue
+
+        try:
+            return get_supported_language_variant(accept_lang)
+        except LookupError:
+            continue
+
+    try:
+        return get_supported_language_variant(settings.LANGUAGE_CODE)
```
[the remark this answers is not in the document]
jwhitlock: These lines are directly copied from the Django source:

https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/utils/translation/trans_real.py#L527-L530

I haven't investigated why they want the transforms in ``get_supported_language_variant`` applied to the default ``settings.LANGUAGE_CODE``.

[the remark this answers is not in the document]
jwhitlock: It was added in Django 1.6 in https://github.com/django/django/pull/824, to fix https://github.com/django/django/pull/824. It handles a situation that we don't have on MDN. I'll replace it with an assertion.

jwhitlock:
This now represents where I want to go with the code. However, there's two big issues:

* 21 commits is too many for a PR that's going to touch so much
* Adding this to [bug 1429933](https://bugzilla.mozilla.org/show_bug.cgi?id=1429933) with a bunch of smaller issues will confuse future developers.

I'm going to close this PR, and open new PRs based on master and against

PR #4767, opened 2018-05-06
Add title in code samples #811001

Ladeia:
I add charset and a title for code samples pages.


jwhitlock:
Thanks @Ladeia. The code looks correct, but the tests are failing. Do you want to fix the tests, or do you need some help?

kuma/wiki/jinja2/wiki/code_sample.html
```diff
@@ -7,6 +7,8 @@
             {{ css|safe }}
         </style>
         {% endif %}
+        <meta charset="utf-8">
```
jwhitlock: This should be just after the ``<head>`` element.

PR #4769, opened 2018-05-07
bug 773921: remove wiki.new_revision_based_on endpoint

escattone:
This PR removes the `wiki.new_revision_based_on` endpoint and its related tests. It was disabled almost 6 years ago, and given that it no longer works with the current CloudFront configuration (see [bugzilla #1459087](https://bugzilla.mozilla.org/show_bug.cgi?id=1459087)), it's time for its removal. It has the added benefit of removing the need for the `revision_id` parameter on the `kuma.wiki.views.translate.translate` function as well.

This PR resolves/fixes bugzilla #[1459087](https://bugzilla.mozilla.org/show_bug.cgi?id=1459087).


kuma/wiki/tests/test_templates.py
```diff
@@ -1180,27 +1160,6 @@ def test_translate_form_content_fallback(self):
         existing_rev = document.revisions.all()[0]
         assert doc('#id_content').text() == existing_rev.content
 
-    @pytest.mark.xfail(reason='Figure out wtf is going on with this test.')
```
jwhitlock: Looks like you figured it out 😆

escattone: LOL

PR #4779, opened 2018-05-14
bug 1431259: fix headless tests with localhost

escattone:
This PR fixes the headless tests such that they now successfully run against the local development instance loaded with a stock sample database.
- Add `is_local_url` and `is_searchable` fixtures.
- Use pytest's ability to [mark a test with `xfail` from within the test function](https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail). This was simpler than trying to create a dynamic fixture.
- Modify some URL's to work on both a local instance loaded with the stock sample database as well as the stage and production instances (e.g., `/en-US/docs/Web/HTML$revision/1252409` rather than `/en-US/docs/Web/HTML$revision/1293895`)

I also removed the `sensitive_url` fixture since it was unused.

### DIFF
--- a/tests/headless/test_cdn.py
+++ b/tests/headless/test_cdn.py
@@ -218,13 +218,13 @@ def test_not_cached_403(base_url, is_behind_cdn, slug):
      '/en-US/docs/Web/HTML$json',
      '/en-US/docs/Web/HTML$history',
      '/en-US/docs/Web/HTML$children',
-     '/en-US/docs/Web/HTML$revision/1293895',
-     '/en-US/docs/Web/HTML$compare?locale=en-US&to=1299417&from=1293895',
+     '/en-US/docs/Web/HTML$revision/1252409',
+     '/en-US/docs/Web/HTML$compare?locale=en-US&to=1287251&from=1252409',
      '/en-US/Firefox$json',
      '/en-US/Firefox$history',
      '/en-US/Firefox$children',
-     '/en-US/docs/Learn/CSS/Styling_text/Fundamentals#Color',
-     '/en-US/docs/Learn/CSS/Styling_text/Fundamentals$toc',
+     '/en-US/docs/Learn/CSS#Modules',
+     '/en-US/docs/Learn/CSS$toc',
      '/fr/docs/feeds/rss/l10n-updates',
      '/fr/docs/localization-tag/inprogress',
      '/en-US/docs/all',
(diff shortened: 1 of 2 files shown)


tests/headless/test_cdn.py
```diff
@@ -241,8 +241,13 @@ def test_not_cached_403(base_url, is_behind_cdn, slug):
      '/en-US/docs/top-level',
      '/en-US/docs/with-errors',
      '/en-US/docs/without-parent'])
-def test_cached(base_url, is_behind_cdn, slug):
+def test_cached(base_url, is_behind_cdn, is_local_url, is_searchable, slug):
     """Ensure that these requests that should return 200 are cached."""
+    if is_local_url:
+        if any(slug.startswith(p) for p in ('/diagrams/', '/presentations/')):
+            pytest.xfail('not yet available in the sample database')
```
jwhitlock: These are the [legacy files](https://github.com/mozilla/kuma/blob/df5bc3b1bcc2addc54915b8362e99c86fa50166f/kuma/urls.py#L128-L135), not part of the sample database. We could add ``SERVE_LEGACY`` to the information in ``_kuma_status.json``, or just assume they aren't available. But they won't be added to the sample DB.

tests/headless/test_cdn.py
```diff
@@ -251,8 +256,10 @@ def test_cached(base_url, is_behind_cdn, slug):
     'slug', ['/files/2767/hut.jpg',
              '/@api/deki/files/3613/=hut.jpg',
              '/en-US/dashboards/localization'])
-def test_cached_301(base_url, is_behind_cdn, slug):
+def test_cached_301(base_url, is_behind_cdn, is_local_url, slug):
     """Ensure that these requests that should return 301 are cached."""
+    if is_local_url and any(slug.startswith(p) for p in ('/files/', '/@api/')):
+        pytest.xfail('not yet available in the sample database')
```
jwhitlock: Files won't be part of the sample database either. Stage has these files because we restored an anonymized backup and synced attachments at one point.

tests/headless/test_cdn.py
```diff
@@ -241,8 +241,13 @@ def test_not_cached_403(base_url, is_behind_cdn, slug):
...
      '/en-US/docs/with-errors',
      '/en-US/docs/without-parent'])
-def test_cached(base_url, is_behind_cdn, slug):
+def test_cached(base_url, is_behind_cdn, is_local_url, is_searchable, slug):
     """Ensure that these requests that should return 200 are cached."""
+    if is_local_url:
+        if any(slug.startswith(p) for p in ('/diagrams/', '/presentations/')):
+            pytest.xfail('not yet available in the sample database')
+        elif (not is_searchable) and slug.endswith('/dashboards/macros'):
+            pytest.xfail('search is not available and populated')
```
jwhitlock: 👍 this gives a good diagnostic message if search isn't populated.

tests/headless/test_cdn.py
```diff
@@ -218,13 +218,13 @@ def test_not_cached_403(base_url, is_behind_cdn, slug):
      '/en-US/docs/Web/HTML$json',
      '/en-US/docs/Web/HTML$history',
      '/en-US/docs/Web/HTML$children',
-     '/en-US/docs/Web/HTML$revision/1293895',
-     '/en-US/docs/Web/HTML$compare?locale=en-US&to=1299417&from=1293895',
+     '/en-US/docs/Web/HTML$revision/1252409',
+     '/en-US/docs/Web/HTML$compare?locale=en-US&to=1287251&from=1252409',
```
jwhitlock: Thanks for finding resources already in the sample database!

jwhitlock:
Thanks for the message updates, @escattone!

PR #4781, opened 2018-05-15
deps(node): update to the latest version of dependencies

schalkneethling:
When doing a fresh install of Kuma I ran into a bunch of security warnings and  deprecations. `node-sass` also fails to install. I had a quick look at the version we have in `package.json` and realised they are way out of date.

I updated all dev dependencies to the latest stable, and locked down the versions using `--save-exact`

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4,43 +4,64 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "JSONStream": {
-      "version": "0.8.4",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-0.8.4.tgz",
-      "integrity": "sha1-kWV9/m/4V0gwZhMrRhi2Lo9Ih70=",
-      "dev": true,
-      "requires": {
-        "jsonparse": "0.0.5",
-        "through": "2.3.8"
-      }
-    },
     "abbrev": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
-      "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=",
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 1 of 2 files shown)


escattone:
Sorry for the delay @schalkneethling! I got a warning when I did an `npm install` within the Kuma container. It complains that `gulp-stylelint@7.0.0` requires a "peer" of `stylelint@^9.0.0`. Do you see that as well?

```sh
$ docker-compose exec --user root web bash
root@edb01abf18dc:/app# npm install
npm WARN gulp-stylelint@7.0.0 requires a peer of stylelint@^9.0.0 but none was installed.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
```

It made me want to upgrade our global installs of `fibers`, `csslint`, `jshint`, `node-sass`, `uglify-js`, `clean-css`, and `stylelint` within the Kuma container, as well as the version of node (which is at 6.12). However, when I tried going down that path, the first thing I hit was an error that seems to be an issue with `npm` (https://github.com/npm/npm/issues/16766), so I backed-off for now.

schalkneethling:
Sounds good to me @jwhitlock. With some these packages we are waaaay out of date and should really upgrade, and also find a way to make upgrading be more frictionless. At the moment you can have a newer version of Nodejs and NPM locally and use `npx` with the node package to overcome installation problems locally.

Basically you do:

`npx node@6 npm i` and then `npx node@6 gulp` - We should probably add this to contribution docs somewhere until the other parts have landed. Thoughts?

jwhitlock:
Thanks, I wasn't aware of [npx](https://www.npmjs.com/package/npx).

Our deployment pipeline requires that the tools we use to build assets run in docker. We are still recommending that front-end developers install the tools locally for speed and efficiency, such as in the [install docs](https://kuma.readthedocs.io/en/latest/installation.html#prepare-for-front-end-development) and the [development docs](https://kuma.readthedocs.io/en/latest/development.html#front-end-development-and-compiling-sass-files).  This was done because of performance issues with gulp in a docker environment, and we didn't have the bandwidth or the Docker experience to come up with a solution. The dual development methods are  part of the reason that front-end tools are behind while Python tools are updated often.

I think we'll be able to tackle this in the coming months, using some strategies that the bedrock team have implemented.

PR #4790, opened 2018-05-16
bug 1461350: Reimplement locale-prefixed URLs to mirror Django

jwhitlock:
This replaces the ``LocaleURLMiddleware`` ("Based on zamboni.amo.middleware") with one based on Django's ``LocaleMiddleware``, plus additional middlewares for Kuma-specific functionality.  This replaces PR #4766, which was the work-in-progress version of this code.  

The Django code is from release 1.8.19, and has several customizations to handle Kuma features like mixed-case locale codes (``en-US`` vs Django's ``en-us``). It also removes some features we don't use, such as storing a language preference in the session (adding a ``Vary: cookie`` to all responses), and some flexibility needed in the Django framework code.

The widest code change is the splitting of URL patterns into locale-prefixed patterns and no-locale patterns. In apps, these are grouped into ``lang_urlpatterns`` and ``urlpatterns``, and they are applied in ``kuma/urls.py`` with a customized ``i18n_patterns``. The patterns now include the active locale, and the active locale needs to be temporarily changed (``with language.override(locale)``) to generate URLs to other locales. The default URLs contain the locale, so Kuma is less likely to generate "generic" URLs.

Another subtle change is the handling of region-specific languages in the ``Accept-Language`` header. Previously, a header of ``fr-FR; de=0.5`` would select ``de``, the first exact match. After this change, the Django-based algorithm picks ``fr``, the general locale variant of ``fr-FR``.  More discussion is at https://github.com/mozilla/kuma/pull/4766#issuecomment-388512343.

There is some known follow-on work that is not included, because this monster is scary enough:

* Change Zone-remap redirects from permanent to temporary redirects, and refactor tests
* Simplify Kuma's ``reverse`` to drop support for legacy parameters.
* Simplify test code for the default of locale-prefixed URLs
* (Maybe) Convert the locale prefix from a special parameter of Kuma's ``reverse`` to a standard pattern argument.


kuma/wiki/tests/test_middleware.py
```diff
@@ -158,10 +158,18 @@ def test_zone_url_ends_with_slash(self):
...
+    with the middleware that attempts to redirect requests to the right
+    location. See bug 1267197 for some of the past analysis and discussion.
+
+    With the Mozilla zamboni/amo-derived middleware, there were sometimes two
+    redirects:
+    * A 302 temporary redirect to add a locale to a locale-less path.
+    * A 301 permanent redirect to remap a wiki path to a zoned path.
 
+    With the Django 1.8-derived middleware, these are combined into a single
+    302 temporary redirect, which may also help with retiring DocumentZones.
```
jwhitlock: Oops - this comment should have been removed. Rebase coming.

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+
+
+@lru_cache.lru_cache(maxsize=1000)
+def get_supported_language_variant(raw_lang_code):
+    """
+    Returns the language-code that's listed in supported languages, possibly
+    selecting a more generic variant. Raises LookupError if nothing found.
+
+    If `strict` is False (the default), the function will look for an alternative
+    country-specific variant when the currently checked is not found.
```
escattone: Super nit. This portion on `strict` seems out of place now that it has been removed from the function signature?

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+    * Language code can also be a Kuma language code
+    * Return Kuma languge codes
+    * Force strict=False to always allow fuzzy matching (zh-CHS gets zh-CN)
+    """
+    if raw_lang_code:
+        lang_code = kuma_language_code_to_django(raw_lang_code)
+
+        # Check for known override
+        if lang_code in settings.LOCALE_ALIASES:
+            return settings.LOCALE_ALIASES[lang_code]
```
escattone: ❤️ I love that the `LOCALE_ALIASES` are incorporated into this here. They can be handled in a much cleaner, more straightforward fashion.

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+
+def get_language_from_request(request):
+    """
+    Analyzes the request to find what language the user wants the system to
+    show. Only languages listed in settings.LANGUAGES are taken into account.
+    If the user requests a sublanguage where we have a main language, we send
+    out the main language.
+
+    If check_path is True, the URL path prefix will be checked for a language
+    code, otherwise this is skipped for backwards compatibility.
```
escattone: Super nit. This `check_path` portion seems out of place now that it has been removed from the function signature.

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+
+    accept = request.META.get('HTTP_ACCEPT_LANGUAGE', '')
+    for accept_lang, unused in parse_accept_lang_header(accept):
+        if accept_lang == '*':
+            break
+
+        if not language_code_re.search(accept_lang):  # pragma: no cover
+            # Check added with a security fix:
+            # https://www.djangoproject.com/weblog/2007/oct/26/security-fix/
+            # It is unclear how to trigger this branch, so skipping coverage.
```
escattone: Nice research!

kuma/core/middleware.py
```diff
@@ -1,93 +1,181 @@
...
-
-        if lang in dict(settings.LANGUAGES):
-            # Blank out the locale so that we can set a new one. Remove lang
-            # from the query params so we don't have an infinite loop.
-            prefixer.locale = ''
-            new_path = prefixer.fix(prefixer.shortened_path)
-            query = dict((smart_str(k), v) for
+        """Redirect if ?lang query parameter is valid."""
+        query_lang = request.GET.get('lang')
+        if query_lang not in dict(settings.LANGUAGES):
```
escattone: Nit. I was thinking that since this is run for almost every request, and that most requests will not have a `lang` query parameter, we could short circuit the need for a dictionary lookup with:
```python
if not (query_lang and (query_lang in dict(settings.LANGUAGES))):
```

kuma/core/i18n.py
jwhitlock: Yes, it is a much better fit than the "fix after the fact" method from the last PR.

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+        possible_lang_codes = [lang_code]
+        try:
+            possible_lang_codes.extend(LANG_INFO[lang_code]['fallback'])
+        except KeyError:
+            pass
+        generic_lang_code = lang_code.split('-')[0]
+        possible_lang_codes.append(generic_lang_code)
+        raw_supported_lang_codes = get_languages()
+        supported_lang_codes = [kuma_language_code_to_django(lang)
+                                for lang in raw_supported_lang_codes]
```
escattone: Nit. I was thinking that it's too bad we lose the `OrderedDict` (from `get_languages()`) and its efficiency (for the following  `code in supported_lang_codes` check), when we create the `supported_lang_codes` list. What do you think of either one of the following instead?

```python
raw_supported_lang_codes = get_languages()
supported_lang_codes = OrderedDict((kuma_language_code_to_django(lang), None)
                                   for lang in raw_supported_lang_codes)
```

or, skipping the call to `get_languages` altogether:

```python
supported_lang_codes = OrderedDict((kuma_language_code_to_django(lang), None)
                                   for lang, _ in settings.LANGUAGES)
```
with both options assuming, of course, that we add a `from collections import OrderedDict`.

kuma/core/i18n.py
escattone: I agree. It seems to me that Django's `parse_accept_lang_header` will only return values that would satisfy the `language_code_re.search` call.

kuma/core/middleware.py
```diff
@@ -1,93 +1,181 @@
...
+        lower_match = literal_from_path.lower() == language_from_path.lower()
+        if lower_match and (language_from_path != literal_from_path):
+            # Language code is a lower-case match for a known locale
+            fixed_locale = language_from_path
+        elif literal_from_path.lower() in settings.LOCALE_ALIASES:
+            # Language code is a known general -> specific locale
+            fixed_locale = settings.LOCALE_ALIASES[literal_from_path.lower()]
+        elif not match and literal_from_path.startswith(language_from_path):
+            # Language code is a specific locale (fr vs fr-FR)
+            fixed_locale = language_from_path
```
escattone: 👍 Nice, I verified that `kuma/core/tests/test_locale_middleware.py` covers all three of these cases.

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+    * Don't check session language
+    """
+    lang_code = get_language_from_path(request.path_info)
+    if lang_code is not None:
+        return lang_code
+
+    lang_code = request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+
+    try:
+        return get_supported_language_variant(lang_code)
```
safwanrahman: This language code should always be mdn locale other than user modify the cookie!
I think we can only check if the locale is supported

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+    except LookupError:
+        pass
+
+    accept = request.META.get('HTTP_ACCEPT_LANGUAGE', '')
```
safwanrahman: I dont know if there are any test for the ordering of locale choice! maybe worth writing one if not exists!

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+
+    * Always check the path (check_path=False)
+    * Don't check session language
+    """
+    lang_code = get_language_from_path(request.path_info)
+    if lang_code is not None:
+        return lang_code
+
+    lang_code = request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+
```
safwanrahman: This should be inside a if block.
```
if lang_code:
...
```

kuma/core/i18n.py
```diff
@@ -1,4 +1,165 @@
...
+        try:
+            return get_supported_language_variant(accept_lang)
+        except LookupError:
+            continue
+
+    # Fallback to default settings.LANGUAGE_CODE.
+    # Django supports a case when LANGUAGE_CODE is not in LANGUAGES
+    # (see https://github.com/django/django/pull/824). but our LANGUAGE_CODE is
+    # always in LANGUAGES.
+    assert settings.LANGUAGE_CODE == settings.LANGUAGES[0][0]
```
safwanrahman: I dont know if we should check it in every request. maybe add a test where its checked?

kuma/wiki/middleware.py
```diff
@@ -33,8 +33,16 @@ def process_request(self, request):
...
+        request_slug = path.lstrip('/')
+        maybe_lang = request_slug.split(u'/')[0]
+        if maybe_lang in settings.ENABLED_LOCALES:
+            path = u'/' + u'/'.join(request_slug.split(u'/')[1:])
+        else:
+            path = u'/' + request_slug
+
         # Skip slugs that don't have locales, and won't be in a zone
-        request_slug = request.path_info.lstrip('/')
         if any(request_slug.startswith(slug)
```
escattone: Nit. It seems more efficient to move lines 46-48 immediately after the definition of `request_slug`, and then start the `maybe_lang` and `path` stuff.

kuma/wiki/middleware.py
```diff
@@ -33,8 +33,16 @@ def process_request(self, request):
                 ('$subscribe' in request.path or '$files' in request.path)):
             return None
 
+        # Process the request path
+        path = request.path_info
+        request_slug = path.lstrip('/')
+        maybe_lang = request_slug.split(u'/')[0]
+        if maybe_lang in settings.ENABLED_LOCALES:
```
escattone: Nit. Since the code is run for almost every request, it'd be nice to make this more efficient by using a `set` or `frozenset` that we create once and use thereafter, so perhaps adding something like `ENABLED_LOCALES_SET = frozenset(ENABLED_LOCALES)` within `kuma/settings/common.py` and then using that here?

kuma/wiki/urls.py
```diff
@@ -29,6 +29,9 @@
     url(r'^\$files$',
         edit_attachment,
         name='attachments.edit_attachment'),
+    url(r'^\$edit/(?P<revision_id>\d+)$',
+        views.edit.edit,
+        name='wiki.new_revision_based_on'),
```
escattone: I'm not sure how this `wiki.new_revision_based_on` url made its way back into our lives, but it needs to go! 😄

kuma/core/middleware.py
```diff
@@ -1,93 +1,181 @@
...
-        # use Client instead) fail loudly and reliably. Otherwise, the set
-        # prefixer bleeds from one test to the next, making tests
-        # order-dependent and causing hard-to-track failures.
-        set_url_prefixer(None)
+        """Add Content-Language, convert some 404s to locale redirects."""
+        language = get_language()
+        language_from_path = get_language_from_path(request.path_info)
+        if response.status_code == 404 and not language_from_path:
+            language_path = '/%s%s' % (language, request.path_info)
+            path_valid = is_valid_path(request, language_path)
```
escattone: This fails to redirect locale-free zone URL's (e.g., `http://localhost:8000/Firefox` gets a 404 instead of redirecting to the active locale set from `get_language_from_request` on the way in), which causes a number of the tests within `tests/headless/test_cdn.py` to fail. It seems that `is_valid_path` has to consider zones as well? 🤦‍♂️ I'm tempted to say we should take this opportunity to dump zones now, but that probably is a bridge too far.

jwhitlock: Which tests fail?

escattone: I posted my actual results in the main thread below.

jwhitlock:
@escattone see if http://localhost:8000/es/Firefox is in your locale dev environment, and if it isn't, ``./manage.py scrape_document https://developer.mozilla.org/es/Firefox`` and see if it fixes that half of the failing tests. The new code is more likely to return a 404 rather than a 30x to a 404.

escattone:
@jwhitlock You're right on. All of the headless tests passed after I did the following:
```sh
docker-compose exec web ./manage.py scrape_document https://developer.mozilla.org/es/Firefox
docker-compose exec web ./manage.py scrape_document https://developer.mozilla.org/fr/Firefox
```

escattone:
That's good news! So it seems all that's needed is to `xfail` those headless tests.

jwhitlock:
Also: the ``de`` test passes, and I guess it shouldn't. I may save that for a future bug.

kuma/core/i18n.py
jwhitlock: It's unclear which line you are saying should be in an if block. I've tried to determine a line that should be in an if block, and none seem appropriate to me.

kuma/core/i18n.py
jwhitlock: No, I think it is a bad idea to avoid ``get_supported_language_variant`` here.

``lang_code`` is ``None`` if there is no request cookie, and ``get_supported_language_variant`` will raise a ``LookupError``.

``get_supported_language_variant`` hecks that the cookie-provided locale is supported, and other stuff. This is where we'll do any locale switching if we remove a previously supported language, or it will also raise a ``LookupError`` and we'll fallback to the ``Accept-Language`` header.

kuma/core/i18n.py
jwhitlock: There's a headless test, but probably not a unit test.

kuma/core/i18n.py
jwhitlock: I'm going to replace it with an assertion, and we'll let the internet tell us what string we haven't thought of.

kuma/core/i18n.py
jwhitlock: It's indirectly checked here:

https://github.com/mozilla/kuma/blob/cdfce6d69f6740d2061c8795cbfc7b12ba05b92d/kuma/settings/common.py#L236-L237

I think it makes sense to add a test and remove the assertion here.

kuma/core/middleware.py
jwhitlock: To summarize the results and discussion below:

* Tests against /es/Firefox and /fr/Firefox failed because only the English page is in the sample database. This code returns a 404, where the previous code returned a 302. This is probably an improvement, but means those tests should be skipped or marked ``XFAIL`` in local development, until the sample database is updated.
* The test for ``?lang=de`` succeeded with a 302, even though the page is not in the sample database.  The quick 404 would be nice, but it isn't a regression, so I'm not considering a change for this PR.

kuma/core/i18n.py
jwhitlock: I'm not sure why an ``OrderedDict`` would be more efficient to iterate over than a list.

 I'm not as sensitive to efficiency here, because of the ``lru_cache``, but it appears we could use similar code in a few places, so it makes sense to calculate it once and skip ``get_languages`` (or call it ``get_kuma_languages``).

kuma/core/middleware.py
jwhitlock: We have a few cases where we'd like a quick check of the supported Kuma language code, so I'll go a little farther down this path and avoid creating a ``dict`` every time.

kuma/wiki/urls.py
jwhitlock: Nice catch!

kuma/core/i18n.py
jwhitlock: There's a few tests of each piece of this. I couldn't find an explicit test that the cookie beat the ``Accept-Language`` header, so I updated ``test_locale_middleware_language_cookie`` to include a header.

jwhitlock:
I believe I've addressed the review comments. 

I made a significant change to ``settings.LANGUAGES``, which now contains locales in resolution order rather than "sorted" order (English, then sorted by locale code. This allowed further simplifications, and the use of an ``OrderedDict`` for efficiently validating locale codes (``locale in get_kuma_languages()``) and iterating (``for locale in get_kuma_languages()``), similar to how Django's ``get_languages`` is used. This big change is in  056769089c56abd358bbd2050e74ce511baeef55.

kuma/core/tests/test_settings.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.mark.parametrize(
+    'primary,secondary',
+    (('bn-BD', 'bn-IN'),
+     ('pt-PT', 'pt-BR'),
+     ('sr', 'sr-Latn'),
+     ('zh-CN', 'zh-TW'),
+     ))
+def test_preferred_locale_codes(primary, secondary):
```
escattone: 👍 Really nice idea to formalize in a test!

kuma/core/tests/test_settings.py
```diff
@@ -0,0 +1,35 @@
+'''Check that settings are consistent.'''
+import pytest
+from django.conf import settings
+
+
+def test_accepted_locales():
```
escattone: Another great test addition, thanks!

PR #4793, opened 2018-05-17
bug 1461350: Simplify test client

jwhitlock:
With the ``LocaleMiddleware``, the default is to include the locale prefix in URLs, so the ``LocalizingMixin`` and ``LocalizingClient`` are no longer needed.

The ``SessionAwareClient`` included code from Django 1.7, and was probably added when MDN was on Django 1.4 or earlier, and is included in the default 1.8 test client:

https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/test/client.py#L411-L426

### DIFF
--- a/docs/tests.rst
+++ b/docs/tests.rst
@@ -133,10 +133,6 @@ new functionality. All apps have a ``tests`` module where tests should go. They
 will be discovered automatically by the test runner as long as the look like a
 test.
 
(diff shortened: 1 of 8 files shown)


jezdez:
Yep I can confirm this was code added to bridge the gap until modern Django is used.

PR #4797, opened 2018-05-18
bug 1461350: Default to locale=en-US in tests

jwhitlock:
When using Kuma's ``reverse``, don't pass ``locale="en-US"`` (common in tests to avoid a redirect) or ``locale=request.LANGUAGE_CODE`` (used in app code to avoid a redirect), but instead rely on the default behavior of including the active language.
    
Continue passing the locale parameter for non-default locales, or if the test makes a request that sets the active language (usually by making a request to a non-default locale).
    
The active language is reset to en-US at the start of each test with an always-run fixture ``set_default_language``.

### DIFF
--- a/kuma/attachments/tests/test_templates.py
+++ b/kuma/attachments/tests/test_templates.py
@@ -17,8 +17,7 @@ def test_xss_file_attachment_title(admin_client, constance_config, root_doc,
     # use view to create new attachment
     file_for_upload = make_test_file()
     files_url = reverse('attachments.edit_attachment',
-                        kwargs={'document_path': root_doc.slug},
-                        locale='en-US')
+                        kwargs={'document_path': root_doc.slug})
     title = '"><img src=x onerror=prompt(navigator.userAgent);>'
     post_data = {
         'title': title,
--- a/kuma/attachments/tests/test_views.py
+++ b/kuma/attachments/tests/test_views.py
@@ -29,8 +29,7 @@ class AttachmentViewTests(UserTestCase, WikiTestCase):
         self.revision = revision(save=True)
         self.document = self.revision.document
         self.files_url = reverse('attachments.edit_attachment',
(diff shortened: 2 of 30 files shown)


jwhitlock:
This change isn't strictly needed, but I wanted to take advantage of the new middleware and stop using ``locale`` when it was no longer needed, and not confuse future devs.

I started this by searching the code for every use of ``reverse`` and dropping the locale parameter when I thought I could. I was surprised that some tests failed, and I found the always-run fixture in the [django-pytest FAQ](https://pytest-django.readthedocs.io/en/latest/faq.html).

I thought about taking this a step farther, and trying to eliminate Kuma's ``reverse`` entirely. This would be possible by customizing ``i18n_patterns`` further:

```
class AnyLocaleRegexURLResolver(DjangoLocaleRegexURLResolver):
    re_any_locale = re.compile('^(?P<locale>%s)/' % '|'.join(
        locale for locale, _ in settings.LANGUAGES), re.UNICODE)

    @property
    def regex(self):
        return self.re_any_locale

def i18n_patterns(*args):
    return [LocaleRegexURLResolver(list(args))] + [AnyLocaleRegexURLResolver(list(args))])
]
```

Then you could set the locale with Django's ``reverse('wiki.document', kwargs={'locale': 'fr', 'document_path': 'Web/CSS'})``. 

However, that would require several steps:
* ``document_path`` should probably be ``slug``, to match the model field
* It will take some changes to other code, such as ``process_document_path``, and probably all of the document views, to change the name of this field.
* It might be a good idea to get rid of zone root URLs first ([bug 1462475](https://bugzilla.mozilla.org/show_bug.cgi?id=1462475)), which could further simplify document views
* All the callers of ``reverse`` would probably need to change to ``kwargs`` style calls rather than ``args`` style.

This seemed like a good thing to delay until after the Django update.

PR #4798, opened 2018-05-18
bug 1301116, 1401246: Dockerfile updates for Django update

jwhitlock:
[bug 1301116](https://bugzilla.mozilla.org/show_bug.cgi?id=1301116) was about an issue running ``./manage.py extract`` with the Ubuntu image. Now that we're on a Python image and two years have passed, it appears it is no longer needed.

There's a gross warning about ``/usr/local/lib/python2.7/site-packages/backports/__init__.py`` that is caused by configparser 3.5.0 (see https://bitbucket.org/ambv/configparser/issues/20/importwarning-missing-__init__py and https://bitbucket.org/brandon/backports/issues/5/missing-__init__py). Until it is fixed, it is easy enough to avoid it by touching that file.

I'm often creating a Docker image with Django 1.9 and 1.11 in it, by changing the Django version in ``requirements/default.txt``, and running ``VERSION=django-1.9 make build-base``. I'm adding a nicer version here, so that you can make it with ``make build-base-1.9``, etc. You can then create environments for testing with Django versions, such as https://gist.github.com/jwhitlock/a643cf804746be7d23baba5b16d8fa53. 

I think these alternate Dockerfiles will be useful for the upgrade effort, and then should be deleted. Or, maybe replaced with one for Python 3, Django 2.0, etc.

### DIFF
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -95,9 +95,6 @@ RUN adduser --uid 1000 --disabled-password --gecos '' --no-create-home kuma
 WORKDIR /app
 EXPOSE 8000
 
-# bug 1301116
-RUN pip install setuptools==26.1.1
-
 RUN npm install -g \
         fibers@1.0.15 \
         csslint@0.10.0 \
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -113,6 +113,10 @@ ENV PIPELINE_CSS_COMPRESSOR=kuma.core.pipeline.cleancss.CleanCSSCompressor \
 COPY ./requirements /app/requirements
 RUN pip install --no-cache-dir -r requirements/dev.txt
 
+# Import issue with configparser 3.5.0
+# https://bitbucket.org/ambv/configparser/issues/20/importwarning-missing-__init__py
+RUN touch /usr/local/lib/python2.7/site-packages/backports/__init__.py
+
 USER kuma
 
 ENV WEB_CONCURRENCY=4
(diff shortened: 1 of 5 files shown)


docker/images/kuma_base/Dockerfile-1.10
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/mozmar/kuma_base:latest
+
+USER root
+RUN pip install Django>1.10,<1.11
```
escattone: Typo: Needs quotes around the versions like `Dockerfile-1.9`:
```sh
RUN pip install "Django>1.10,<1.11"
```

docker/images/kuma_base/Dockerfile-1.11
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/mozmar/kuma_base:latest
+
+USER root
+RUN pip install Django>1.11,<1.12
```
escattone: Typo: Needs quotes around the versions like `Dockerfile-1.9`:
```sh
RUN pip install "Django>1.11,<1.12"
```

docker/images/kuma_base/Dockerfile
```diff
@@ -95,9 +95,6 @@ RUN adduser --uid 1000 --disabled-password --gecos '' --no-create-home kuma
 WORKDIR /app
 EXPOSE 8000
 
-# bug 1301116
-RUN pip install setuptools==26.1.1
```
escattone: Nice to see this go away!

jwhitlock:
Fridays 🤦‍♂️. Force-push to update.

PR #4804, opened 2018-05-21
[Bug 1401246] Fixing select_related to select only foreignkey fields

safwanrahman:
Fixing all the `select_related` queryset to get compataibility with django 1.9+.
It seems to be a bug in django code that let selecting non related fields by `select_related` as it was fetching the whole object instead of the fields only.

The actual use case would be using `only` for limiting fetch fields instead of full object. In all of our use case, it seems that we are covered with `only`.

@jwhitlock @escattone r?

### DIFF
--- a/kuma/wiki/views/document.py
+++ b/kuma/wiki/views/document.py
@@ -241,16 +241,18 @@ def _get_doc_and_fallback_reason(document_locale, document_slug):
     doc = None
     fallback_reason = None
 
+    # Optimizing the queryset to fetch the required values only
+    related_fields = ['current_revision', 'current_revision__creator',
+                      'parent', 'parent__current_revision', 'parent__topic']
     current_revision_fields = ['current_revision__{}'.format(field) for field in
                                ('toc_depth', 'created', 'creator__id', 'creator__username', 'creator__is_active')]
     parent_fields = ['parent__{}'.format(field) for field in ('locale', 'slug', 'current_revision__slug')]
     parent_topic_fields = ['parent_topic__{}'.format(field) for field in ('id', 'title', 'slug')]
 
-    related_fields = current_revision_fields + parent_fields + parent_topic_fields
-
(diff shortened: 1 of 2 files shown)


kuma/wiki/views/document.py
```diff
@@ -241,16 +241,18 @@ def _get_doc_and_fallback_reason(document_locale, document_slug):
...
 
-    related_fields = current_revision_fields + parent_fields + parent_topic_fields
-
-    fields = ['html', 'rendered_html', 'zone_subnav_local_html', 'body_html',
-              'locale', 'slug', 'title', 'is_localizable', 'rendered_errors',
-              'toc_html', 'summary_html', 'summary_text', 'quick_links_html'] + related_fields
+    fields = (['html', 'rendered_html', 'zone_subnav_local_html', 'body_html',
+               'locale', 'slug', 'title', 'is_localizable', 'rendered_errors',
+               'toc_html', 'summary_html', 'summary_text', 'quick_links_html']
+              + current_revision_fields + parent_fields + parent_topic_fields)
```
jwhitlock: flake8 requests that the ``+`` sign ends the previous line rather than starts this line.

safwanrahman: having the `+` sign at previous line makes me confuse while I read the code.
So for readibility, I have made another field to make flake8 happy

kuma/wiki/views/document.py
```diff
@@ -241,16 +241,18 @@ def _get_doc_and_fallback_reason(document_locale, document_slug):
     doc = None
     fallback_reason = None
 
+    # Optimizing the queryset to fetch the required values only
+    related_fields = ['current_revision', 'current_revision__creator',
+                      'parent', 'parent__current_revision', 'parent__topic']
```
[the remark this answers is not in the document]
safwanrahman: Oh! Sorry, my bad!

safwanrahman:
@jwhitlock Test passed! r+

PR #4807, opened 2018-05-22
bug 1401246: Use relative URIs for redirects in Django 1.9 and later

jwhitlock:
This PR centralizes Django version detection in settings, because I was getting tired of duplicating the same two imports across the code. If desired, this can be pulled into a new PR, or removed.

In Django 1.9, [HTTP redirects are no longer converted to absolute URIs](https://docs.djangoproject.com/en/2.0/releases/1.9/#http-redirects-no-longer-forced-to-absolute-uris). Absolute URIs, such as ``https://developer.mozilla.org/en-US/docs/Web/JavaScript``, were a requirement with [RFC 2616](https://tools.ietf.org/html/rfc2616.html). Relative URIs, such as ``/en-US/docs/Web/JavaScript``, are allowed in the ``Location`` header with [RFC 7231](https://docs.djangoproject.com/en/2.0/releases/1.9/#http-redirects-no-longer-forced-to-absolute-uris). Django 1.8 still forces absolute URIs.

This PR changes our code that generates redirects to use relative URIs. A function ``assert_relative_uri`` is used to test the Django 1.8 and Django 1.9 behavior, and can be replaced with an equality assert after the upgrade.

### DIFF
--- a/kuma/core/tests/__init__.py
+++ b/kuma/core/tests/__init__.py
@@ -5,6 +5,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.messages.storage.fallback import FallbackStorage
 from django.core.cache import cache
 from django.test import TestCase
+from django.utils.six.moves.urllib.parse import urlsplit, urlunsplit
 from django.utils.translation import trans_real
 
 from ..cache import memcache
@@ -21,6 +22,31 @@ def assert_no_cache_header(response):
 def assert_shared_cache_header(response):
     assert 'public' in response['Cache-Control']
     assert 's-maxage' in response['Cache-Control']
+
+
+def assert_relative_uri(uri, expected_uri):
+    """
+    Assert that a URIs absolute portion matches.
+
+    Keyword Arguments:
+    uri - a URI like 'http://testserver/rel/path?foo=1'
+    relative_uri - The relative portion, like '/rel/path?foo=1'
+
+    This is a helper assertion for the transition to Django 1.9.
+    In Django 1.8 and earlier, redirects use RFC 2616 full URIs.
+    In Django 1.9 and later, redirects use RFC 7231 relative URIs.
+    After the transition to Django 1.11, these assertions can be
+    replaced with simple equality assertions.
+
+    See "HTTP redirects no longer forced to absolute URIs" in:
+    https://docs.djangoproject.com/en/2.0/releases/1.9/
+    """
+    if settings.DJANGO_1_9:
+        assert uri == expected_uri
+    else:
+        parts = urlsplit(uri)
+        relative_uri = urlunsplit(('', '') + parts[2:])
+        assert relative_uri == expected_uri
 
 
 def eq_(first, second, msg=None):
--- a/kuma/core/tests/__init__.py
+++ b/kuma/core/tests/__init__.py
@@ -24,29 +24,32 @@ def assert_shared_cache_header(response):
     assert 's-maxage' in response['Cache-Control']
 
 
-def assert_relative_uri(uri, expected_uri):
+def assert_relative_reference(uri, expected_reference):
     """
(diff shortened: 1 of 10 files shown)


kuma/settings/common.py
```diff
@@ -31,6 +31,12 @@ def __call__(self, value):
         return tuple((value, value) for value in split_values)
 
 
+# For the Django 1.11 update effort - Are we at least at this version?
+_dj_version = LooseVersion(get_version())
+DJANGO_1_9 = _dj_version >= LooseVersion('1.9')
+DJANGO_1_10 = _dj_version >= LooseVersion('1.10')
+DJANGO_1_11 = _dj_version >= LooseVersion('1.11')
```
escattone: ❤️ These are nice to have!

kuma/core/tests/__init__.py
```diff
@@ -23,6 +24,31 @@ def assert_shared_cache_header(response):
     assert 's-maxage' in response['Cache-Control']
 
 
+def assert_relative_uri(uri, expected_uri):
+    """
+    Assert that a URIs absolute portion matches.
```
escattone: I think you want `relative` rather than `absolute`?

jwhitlock: I looked around for a few minutes to try to figure out what to call the part of the URL after the domain name. I can't find a reference, so I think we'll have to come to an agreement among ourselves.

Django has ``Model.get_absolute_url()``, which returns the path, such as ``/en-US/docs/Web/JavaScript``. We've added ``get_full_url()`` in a few places, to get the URL with host name, like ``https://developer.mozilla.org/en-US/docs/Web/JavaScript``.  I'm sad that Django uses ``get_absolute_url`` to generate a path, but I think it will be a while before I can fix that.

Python calls ``/en-US/docs/Web/JavaScript`` the path in functions like ``urlunsplit``, and calls the other bits the scheme (``https``), netloc, (``developer.mozilla.org``), query (the stuff after the question mark ``?``), and the fragment (the stuff after a hash mark ``#``).

Wikipedia has no help for what to call "the slash and everything after" on the [Uniform Resource Identifier article](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier).

All of this makes me think "call it absolute", to follow Django, but then they go an use a whole new syntax when [describing the change](https://docs.djangoproject.com/en/2.0/releases/1.9/#http-redirects-no-longer-forced-to-absolute-uris) in the release notes:

> Relative redirects are no longer converted to absolute URIs. [RFC 2616](https://tools.ietf.org/html/rfc2616.html) required the Location header in redirect responses to be an absolute URI, but it has been superseded by [RFC 7231](https://tools.ietf.org/html/rfc7231.html) which allows relative URIs in ``Location``, recognizing the actual practice of user agents, almost all of which support them.

RFC 7231 [Section 7.1.2, Location](https://tools.ietf.org/html/rfc7231.html#section-7.1.2), is where the "relative reference" terminology comes from, and refers to RFC 3986 [Section 4.2 Relative Reference](https://tools.ietf.org/html/rfc3986#section-4.2), as opposed to an absolute URI. However, "relative reference" is not a healthy search term, and mostly gets you tips and tricks in Excel.

So, here's the options, and I see them

* ``assert_relative_uri`` - supported by just about no one
* ``assert_absolute_url`` - borrow Django's relative vs Kuma's full syntax
* ``assert_relative_reference`` - borrow from RFC 3986

My weak preference is the last one, with a link to RFC 3986.

escattone: Wow, that's a nice summary, thanks! I have a weak preference for the last one as well, but I can accept anything that you decide.

jwhitlock: OK, going with ``assert_relative_reference``. That will help me quickly get rid of it when we're on Django 1.11!

PR #4808, opened 2018-05-22
bug 1461350: Update URLs for attachment host

jwhitlock:
Update the URLs for the attachment host to ensure that code samples and file attachments are processed with locale prefixes.

This adds a test ``test_code_sample_host_restricted_host`` that confirms that a code sample can be viewed with ``ENABLED_RESTRICTIONS_BY_HOST``. This test fails before this change. Other tests are split up so they test a single configuration.

The urls cache is set when the ``RestrictedEndpointsMiddleware`` overrides the urlconf. Django's [SimpleTestCase](https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/test/testcases.py#L217-L228) has code to reset this cache, but pytest doesn't [unless mark.urls is used](https://github.com/pytest-dev/pytest-django/blob/a67ab3bd0eda8a2b76b4c5954b04482b4f27e20a/pytest_django/plugin.py#L447-L473). mark.urls is set to the default urlconf, the test sets it to the restricted urlconf, and pytest-django cleans it up.

### DIFF
--- a/kuma/urls_untrusted.py
+++ b/kuma/urls_untrusted.py
@@ -2,7 +2,9 @@ from django.conf import settings
 from django.conf.urls import include, url
 
 from kuma.core import views as core_views
+from kuma.core.urlresolvers import i18n_patterns
 from kuma.landing.views import robots_txt
+from kuma.wiki.urls_untrusted import lang_urlpatterns as wiki_lang_urlpatterns
 
 handler403 = core_views.handler403
 handler404 = core_views.handler404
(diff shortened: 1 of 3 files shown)


jens1o:
Thank you!

jwhitlock:
For future researchers: PR #4766 didn't include an update for the patterns for the stand-alone attachments host when ``ENABLED_RESTRICTIONS_BY_HOST`` is enabled, so any URL including a locale prefix was broken. This included the embedded samples and some redirects used for in-sample images. [Bug 1463428](https://bugzilla.mozilla.org/show_bug.cgi?id=1463428) collected the many user-submitted bugs about broken embedded samples. This PR fixed that bug.

PR #4811, opened 2018-05-23
bug 1401246: Use random seed in Django 1.9+ jobs

jwhitlock:
Use ``PYTHONHASHSEED=0`` in the Django 1.9, 1.10, and 1.11 jobs, to mirror the standard 1.8 build. This avoids test failures due to random behavior, such as key order in dictionaries, which are covered by a different optional TravisCI build.

Related to PR #4806.


jwhitlock:
Force-push to add ``PYTHONHASHSEED=0`` to the ``allow_failures`` section, to keep these builds optional.

PR #4813, opened 2018-05-23
bug 1401246: Update aksimet submission creation

jwhitlock:
In spam dashboard tests, adjust how ``RevisionAkismentSubmissions`` are created, to be compatible with Django 1.9 and later.


jwhitlock:
Force-push for flake8 fix and to convert two more test cases.

PR #4815, opened 2018-05-24
feat(callout): Fix Bug 1462835, replace saucelabs banner with new perception survey

schalkneethling:
Replace saucelabs banner with new perception survey banner. As we will not switch back to the Saucelabs banner, I took the opportunity to make this style banner standard going forward.

Also note, I am waiting on Dan and Matiej with regards to the `utm` parameters. So, good for review but, wait on merge.


kuma/static/styles/components/home/column-callout.scss
```diff
@@ -239,36 +238,51 @@ $saucelabs-callout-image-spacing: 129px + 20px;
...
+
+.callout-survey {
+    &:after {
+        background-image: url($path-to-embedded-images + 'promos/survey.svg');
+    }
+}
+
+@media (-webkit-min-device-pixel-ratio: 1.5), (min--moz-device-pixel-ratio: 1.5), (-o-min-device-pixel-ratio: 3/2), (min-resolution: 1.5dppx) {
+    .callout-survey a:after {
+        background-image: url($path-to-embedded-images + 'promos/survey.svg');
```
jwhitlock: I think I led you down the wrong path. Here's what the banner looks like now:

screenshot 2018-05-29 07 55 45

Maybe we should return to ``callout-standard`` for this rule.

schalkneethling: Interesting that you see that, I guess I did not tests again after rebasing. I had a hell of a time rebasing and some thing clearly got lost in translation. I will resolve this issue. Having been without power for the entire day :-/ I will take care of this first thing tomorrow. Thanks for highlighting this.

jwhitlock: Sorry about your rough day! If it helps, here's the branch before the latest rebase:

https://github.com/jwhitlock/kuma/tree/survey-banner-1462835

schalkneethling:
@jwhitlock Alrighty, updated.

kuma/static/styles/components/home/column-callout.scss
jwhitlock: Much better!

screenshot 2018-05-30 08 51 57

PR #4816, opened 2018-05-24
bug 1401246: Fix test for raw file with 304

jwhitlock:
When the ``ConditionalGetMiddleware`` in 1.9 and 1.10 detects that a file attachment is unchanged, it returns a ``304 Not Modified`` response that removes caching headers and the ``Last-Modified`` header. The test is xfailed for these Django versions before the header check.

Django 1.11 retains the headers, so tests will pass like 1.9. The tests are modified to add the attachment host to ``ALLOWED_HOSTS``, which is checked in tests starting in 1.10.

### DIFF
--- a/kuma/attachments/tests/test_views.py
+++ b/kuma/attachments/tests/test_views.py
@@ -229,12 +229,13 @@ def test_edit_attachment_post_with_vacant_file(admin_client, root_doc, tmpdir,
 
 def test_raw_file_requires_attachment_host(client, settings, file_attachment):
     settings.ATTACHMENT_HOST = 'demos'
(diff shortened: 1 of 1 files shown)


kuma/attachments/tests/test_views.py
```diff
@@ -229,12 +229,13 @@ def test_edit_attachment_post_with_vacant_file(admin_client, root_doc, tmpdir,
 
 def test_raw_file_requires_attachment_host(client, settings, file_attachment):
     settings.ATTACHMENT_HOST = 'demos'
+    settings.ALLOWED_HOSTS.append('demos')
```
escattone: Nice find.

PR #4819, opened 2018-05-25
bug 1401246: Add test hosts to ALLOWED_HOSTS

jwhitlock:
In Django 1.11, ``ALLOWED_HOSTS`` is checked when running with the test client. Previously, any domain was allowed.  Add test domains as needed for tests to pass in 1.11.


jwhitlock:
When I realized how close we were on 1.11 compatibility, I re-based to get PR #4820 and passing tests in 1.10, and added a commit for the remaining failing test.

``AdminEmailHandler`` now includes the current settings in the email message. When using a minimal ``urlpatterns`` (like the test, but also like the attachments host with ``ENABLE_RESTRICTIONS_BY_HOST``) this can result in a ``NoReverseMatch`` exception at runtime. Our use of ``reverse_lazy`` appears to be from an earlier version of Django (before 1.0)? as well as ``LOGOUT_URL``, which was officially removed in [Django 1.10](https://docs.djangoproject.com/en/2.0/releases/1.10/#miscellaneous), but hasn't been used for a long time.

This change allow tests to pass under Django 1.11, so I ratchet up the TravisCI builds to require those to pass as well.

kuma/core/tests/test_middleware.py
```diff
@@ -73,13 +73,14 @@ def test_force_anonymous_session_middleware(rf, settings):
 def test_restricted_endpoints_middleware(rf, settings):
     settings.ATTACHMENT_HOST = 'demos'
     settings.ENABLE_RESTRICTIONS_BY_HOST = True
+    settings.ALLOWED_HOSTS.append('demos')
```
safwanrahman: I dont know, if we can set `*` while tests!

kuma/settings/common.py
```diff
@@ -1656,9 +1655,8 @@ def get_user_url(user):
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 
 # Auth and permissions related constants
-LOGIN_URL = reverse_lazy('account_login')
-LOGOUT_URL = reverse_lazy('account_logout')
-LOGIN_REDIRECT_URL = reverse_lazy('home')
+LOGIN_URL = 'account_login'
+LOGIN_REDIRECT_URL = 'home'
```
safwanrahman: Where is the `LOGOUT_URL`?

safwanrahman: oh! seen it got removed!

PR #4821, opened 2018-05-29
bug 1189160: Update to django-waffle 0.14.0

jwhitlock:
* django-waffle 0.11 → 0.14.0: Significant caching improvements, fixes for management commands, update Django and Python support.

[Bug 1189160](https://bugzilla.mozilla.org/show_bug.cgi?id=1189160) was listed as a blocker (by me) for the Django 1.11 update, when I was hoping to get all the requirements up to date. I was expecting some breakage with this update, but tests pass, and browsing in the development environment appears to work. There doesn't appear to be any downside to this update.

### DIFF
--- a/kuma/conftest.py
+++ b/kuma/conftest.py
@@ -5,7 +5,7 @@ import requests_mock
 from django.conf import settings
 from django.core.cache import caches
 from django.utils.translation import activate
-from waffle.models import Flag
+from waffle.testutils import override_flag
 
 from kuma.wiki.models import Document, Revision
 
@@ -86,8 +86,8 @@ def user_client(client, wiki_user):
 
 @pytest.fixture
 def editor_client(user_client):
-    Flag.objects.create(name='kumaediting', everyone=True)
-    return user_client
+    with override_flag('kumaediting', True):
+        yield user_client
 
 
 @pytest.fixture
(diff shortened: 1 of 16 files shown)


jwhitlock:
@safwanrahman I agree with the django-waffle maintainers, that it is better to use the [override methods](https://waffle.readthedocs.io/en/stable/testing/automated.html#unit-tests), which they guarantee will clean up after themselves.

I've added a commit that uses ``override_flag`` and ``override_switch`` when it was straightforward to use, such as pytest fixtures and simple tests. When a ``Flag`` or ``Switch`` was setup in a ``setUp`` method for several tests, I instead ensured that the ``.delete`` method was called, which will ensure the waffle settings cache is cleared.

jwhitlock:
Force-push to fix flake8 error (didn't remove a ``Flag`` import), and fixed the commit message as well. ``@override_switch`` can be used as a class decorator, but not on a base class like ``WikiTestCase``, just on the ``TestCase`` where the tests are defined.

@safwanrahman We might be able to clear the cache once we get rid of zone URL roots. Until then, every test that makes a request will also involve a database read, which would be a much larger change.

Even with zone URL roots removed, my preference is to avoid database and cache writes, rather than to rely on clearing it after every single test.

PR #4822, opened 2018-05-29
bug 1401246: make middleware work with MIDDLEWARE setting

escattone:
Now that #4821 has been merged, this PR updates the Kuma middleware in `kuma.core.middleware` and `kuma.wiki.middleware` to handle both the [new style](https://docs.djangoproject.com/en/2.0/releases/1.10/#new-style-middleware) and original style. All of the third-party middleware we currently use is already compatible with both styles, except for [the middleware of `django-redirect-urls`](https://github.com/pmac/django-redirect-urls/blob/master/redirect_urls/middleware.py). I've created a temporary shim (`kuma.core.middleware.RedirectsMiddleware`) to handle that until a new version is available with middleware compatible with both the new and original style.

This has been tested locally with Django 1.8, 1.9, 1.10, and 1.11.

Once we're on Django 1.11, we may want to consider re-writing our Kuma middleware using [the functional style](https://docs.djangoproject.com/en/2.0/topics/http/middleware/#writing-your-own-middleware) (I already have a branch with this already done) so we can stop using `django.utils.deprecation.MiddlewareMixin`.

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -8,8 +8,15 @@ from django.http import (HttpResponseForbidden,
                          HttpResponsePermanentRedirect,
                          HttpResponseRedirect)
 from django.utils import translation
+# TODO: Remove the try-except wrapper after move to Django 1.10+.
+try:
+    from django.utils.deprecation import MiddlewareMixin
+except ImportError:
+    MiddlewareMixin = object
 from django.utils.encoding import iri_to_uri, smart_str
 from django.utils.six.moves.urllib.parse import urlsplit
+from redirect_urls.middleware import (
+    RedirectsMiddleware as OriginalRedirectsMiddleware)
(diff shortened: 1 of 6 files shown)


kuma/core/middleware.py
```diff
@@ -309,3 +316,15 @@ def process_request(self, request):
                 urljoin(settings.SITE_URL, request.get_full_path())
             )
         return None
+
+
+if settings.DJANGO_1_10:
+    class RedirectsMiddleware(MiddlewareMixin, OriginalRedirectsMiddleware):
+        """
+        Enables the redirect_urls middleware to be used with both MIDDLEWARE
+        and MIDDLEWARE_CLASSES until a newer version of django-redirect-urls
```
jwhitlock: ``django-redirect-urls`` is https://github.com/pmac/django-redirect-urls, we should submit a PR.

PR #4824, opened 2018-05-30
bug 1445641, bug 1464892: Add deployment recorder

jwhitlock:
``scripts/record_deployments.py`` takes the starting and ending commit hashes, as well as other parameters, and calls the APIs for New Relic and SpeedCurve to record a deployment.

It gets the API keys from the environment, to avoid printing them on the command line when calling the scripts. It assumes the New Relic application IDs and the SpeedCurve site ID are not sensitive information, and can be passed on the command line. 

I've tried this manually with the latest deployment. It took a couple of tries to call it correctly, and I used a bash script (I can share privately) to store the variables needed to call the script. I wasn't able to get a clean run this time, and I'd like another deployment to work out bugs before automating it.  My thought is that it would be called by Jenkins during deployments, either directly (with API keys in the Jenkins environment) or as a Kubernetes container (API keys stored in Kubernetes). I think the Jenkins way would be easier.

To do:

* [x] Read New Relic docs, see if they think application IDs are sensitive
* [x] Read SpeedCurve docs, see if they think site IDs are sensitive
* [x] Decide on Jenkins or Kubernetes for running script 
* [x] Allow setting IDs from environment
* [x] Run cleanly on a deployment

### DIFF
--- a/scripts/record_deployments.py
+++ b/scripts/record_deployments.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python
+from __future__ import print_function, unicode_literals
+
+import os
+import sys
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+from datetime import datetime
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+
+# API keys and IDs are set with environment variables rather than on the
+# command line, to avoid displaying the key in logs
+NR_API_KEY_NAME = 'NEW_RELIC_API_KEY'
+NR_APP_IDS_NAME = 'NEW_RELIC_APP_IDS'
+SC_API_KEY_NAME = 'SPEEDCURVE_API_KEY'
+SC_SIDE_ID_NAME = 'SPEEDCURVE_SITE_ID'
+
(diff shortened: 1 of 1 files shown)


jwhitlock:
The docs for New Relic and SpeedCurve do not say if they think the parameters are sensitive. It doesn't appear possible to get extra information without also having an API key. There also doesn't appear to be a way to regenerate new IDs. Let's be too careful and keep these secret as well.

Jenkins has the ability to set parameters on the server, using the web interface, and inject them into the build environment. Let's plan to run from Jenkins.

jwhitlock:
This worked for the most recent deployment, and I think it is ready for merging and integrating into Jenkins. I can make that part of this PR if desired.

You can see New Relic deployments by selecting "Deployments" from the sidebar of an APM overview.

Speedcurve deployments are under Synthetic -> Deploy

@escattone, I'll send my helper script (which includes API keys etc), so you can try it as part of QA.

escattone:
I think the integration with Jenkins (or K8s) can go into another PR.

PR #4830, opened 2018-06-01
bug 1401246: Update to Django 1.11

jwhitlock:
* Django 1.8.19 → 1.11.13: Too many changes to summarize here. See the release notes:
  - https://docs.djangoproject.com/en/2.0/releases/1.9/
  - https://docs.djangoproject.com/en/2.0/releases/1.10/
  - https://docs.djangoproject.com/en/2.0/releases/1.11/

The code currently runs in Django 1.8, 1.9, 1.10, and 1.11. This drops support for everything but 1.11. This allows removing a lot of branch code to handle differences between the versions, as well as form widget monkeypatching, and the django-debreach library.

Functional tests are updated to drop support for regex URLs for tests (unused in our code), and to expect relative references (``en-US/docs/Web``) rather than full URLs (``https://developer.mozilla.org/en-US/docs/Web``)


kuma/core/middleware.py
```diff
@@ -318,13 +318,10 @@ def process_request(self, request):
...
-        pass
-else:
-    RedirectsMiddleware = OriginalRedirectsMiddleware
+class RedirectsMiddleware(MiddlewareMixin, OriginalRedirectsMiddleware):
+    """
+    Enables the redirect_urls middleware to be used with both MIDDLEWARE
+    and MIDDLEWARE_CLASSES until a newer version of django-redirect-urls
+    is available that provides this "out of the box".
+    """
+    pass
```
escattone: I think we should also remove the `try-except` wrapper around the import of `MiddlewareMixin` at the top of this file as well as the top of `kuma/wiki/middleware.py`. So in both files replace:
```python
 # TODO: Remove the try-except wrapper after move to Django 1.10+.
 try:
     from django.utils.deprecation import MiddlewareMixin
 except ImportError:
    MiddlewareMixin = object
```
with:
```python
from django.utils.deprecation import MiddlewareMixin
```

kuma/core/monkeypatch.py
```diff
@@ -1,59 +0,0 @@
-from django.forms import fields, widgets
```
escattone: I ❤️ this PR!!!

kuma/wiki/views/document.py
```diff
@@ -891,11 +891,9 @@ def _document_api_PUT(request, document_slug, document_locale):
...
             except ValueError:
                 expected_etags = []
-            current_etag = calculate_etag(doc.get_html(section_id))
-            if settings.DJANGO_1_11:
-                # Django's parse_etags returns a list of quoted rather than
-                # un-quoted ETags starting with version 1.11.
-                current_etag = quote_etag(current_etag)
+            # Django's parse_etags returns a list of quoted rather than
+            # un-quoted ETags starting with version 1.11.
+            current_etag = quote_etag(calculate_etag(doc.get_html(section_id)))
```
escattone: I was hoping it would be resolved this way at this point, so thanks!

requirements/default.txt
```diff
@@ -33,9 +33,9 @@ dennis==0.9 \
     --hash=sha256:f6487392ac91800c5f0684a99b404b7fd0f72ceb48faeb5a0ce4e2c24fb62d3f
 
 # Web framework for Python projects of a certain age
-Django==1.8.19 \
-    --hash=sha256:674c525d3aa90ed683313b64aa27490c31874e16155e6b44772d84e76c83c46c \
-    --hash=sha256:33d44a5cf9d333247a9a374ae1478b78b83c9b78eb316fc04adde62053b4c047
+Django==1.11.13 \
+    --hash=sha256:18986bcffe69653a84eaf1faa1fa5a7eded32cee41cfecc77fdc65a3e046404d \
+    --hash=sha256:46adfe8e0abe4d1f026c1086889970b611aec492784fbdfbdaabc2457360a4a5
```
escattone: Feels so good!

requirements/default.txt
```diff
@@ -73,13 +73,6 @@ django-constance==2.2.0 \
     --hash=sha256:fe601fe384d1629f6f0460175e4ce3d66e8177654525bddd4bdb68025c816adc \
     --hash=sha256:1d74fa615ee6c69512faa214ec7b4962aa039dfe802e967972c3fbdad7852977
 
-# Basic/extra mitigation against the BREACH attack for Django 1.9 and lower
-# Code: https://github.com/lpomfrey/django-debreach
-django-debreach==1.4.2 \
-    --hash=sha256:1626677c47a6496c40d1794cd7264280dcccc63fc69893a86545bb84af3e0f4b \
-    --hash=sha256:20fef417bd12de35594ae2456c86ecf65870a24cfb604194254befb59a3dbe5c \
-    --hash=sha256:efefdb754fbce983302610ca3ec278fb29f86bb01ae6a7defcc573d4048fc43f
-
```
escattone: 🎉 Sorry, it's hard to resist cheering at so much of this PR!

kuma/settings/common.py
```diff
@@ -504,25 +478,15 @@ def _get_locales():
 if not MAINTENANCE_MODE:
     # We don't want this in maintence mode, as it adds "Cookie"
     # to the Vary header, which in turn, kills caching.
-
-    if _NEED_DEBREACH:
-        _MIDDLEWARE += ('debreach.middleware.CSRFCryptMiddleware',)
     _MIDDLEWARE += ('django.middleware.csrf.CsrfViewMiddleware',)
```
escattone: Missed nit, should be `MIDDLEWARE` rather than `_MIDDLEWARE`.

jwhitlock:
Force-push (because of busted tests) to fix the missed ``_MIDDLEWARES`` and the ``MiddlewareMixin`` import.

escattone:
Looks good except missing the removal of the `try-except` wrapper around the import of `MiddlewareMixin` in `kuma/wiki/middleware.py`.

jwhitlock:
That will teach me to force-push. Removed ``try/except`` from ``kuma/wiki/middleware.py``, and updated the ``kuma_base`` ``Dockerfile`` because they've removed ``gpg`` from ``python:2.7-slim``.

kuma/core/__init__.py
```diff
@@ -1,6 +1,2 @@
-from django.conf import settings
-
-if not settings.DJANGO_1_10:
-    from . import monkeypatch  # noqa
-
```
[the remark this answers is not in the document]
jwhitlock: yeah I remember reading monkeypatch.py for the first time a few months ago, and thinking "this must die"

docker/images/kuma_base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:2.7-slim
+FROM python:2.7-slim-jessie
```
jwhitlock: Confirmed, ``python:2.7-slim`` moved from being the same tag as ``python:2.7-slim-jessie`` to ``python:2.7-slim-stretch``:

https://github.com/docker-library/official-images/pull/4385/files#diff-051bb6df61b6c6e7f8c1868985011b07R43

PR #4831, opened 2018-06-01
bug 1401246: Use language code in statici18n

jwhitlock:
[django-statici18n v1.3.0](https://github.com/zyegfryed/django-statici18n/blob/master/docs/changelog.rst#v130-2017-jan-19) uses the language code (``en-US``) in the path for the translation catalog in Django 1.10 and later. For 1.9 and earlier, it uses the locale code (``en_us``).


jwhitlock:
Force push for flake8 error and slightly less awkward commit message.

PR #4838, opened 2018-06-05
bug 1401246: Update feature docs, sample database specification

jwhitlock:
* Sync the documentation, production database, and sample database specification with the feature flags and switches in use in the code today.
* Drop the interactive examples content experiment pages from the sample DB
* Add resources used in the headless tests, and update a redirect test to use a valid revision ID
* Bump sampling attempt count to 7, which means sampling will wait over a minute for the very slow [https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Midas/Security_preferences](/en-US/docs/Mozilla/Projects/Midas/Security_preferences) revision pages to load.

It took 162 minutes to download the sample database, mostly due to pausing for rate limiting. One of the final steps is to refresh the Hacks blog, which was broken in 1.11 until PR #4836. It took a few nights to get a successful run. I've uploaded a sample DB with Django 1.11 migrations applied to https://mdn-downloads.s3-us-west-2.amazonaws.com/index.html.


jwhitlock:
Good catch! I've added a commit to change those tests, and tested the CDN tests locally, against staging, and against production.

PR #4841, opened 2018-06-07
bug 1401246: convert to new-style middleware

escattone:
I did some work on our Kuma middleware a short while ago, but when I realized it needed Django 1.11 in place, I shelved it. I thought it was useful, as it mostly embraced the new, ~~functional~~class-style of writing middleware (except `ReadOnlyMiddleware`), but it's real goal was to remove the dependency on `django.utils.deprecation.MiddlewareMixin` and go all-in with the new-style. I didn't want to lose the work, which wasn't totally finished, so I added what was missing and decided it was now or never.

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -7,8 +7,6 @@ from django.core.urlresolvers import get_script_prefix, resolve, Resolver404
 from django.http import (HttpResponseForbidden,
                          HttpResponsePermanentRedirect,
                          HttpResponseRedirect)
-from django.utils import translation
-from django.utils.deprecation import MiddlewareMixin
 from django.utils.encoding import iri_to_uri, smart_str
 from django.utils.six.moves.urllib.parse import urlsplit
 from whitenoise.middleware import WhiteNoiseMiddleware
@@ -17,7 +15,8 @@ from kuma.wiki.views.legacy import (mindtouch_to_kuma_redirect,
                                     mindtouch_to_kuma_url)
 
 from .decorators import add_shared_cache_control
-from .i18n import (get_kuma_languages,
+from .i18n import (activate_language_from_request,
+                   get_kuma_languages,
                    get_language,
                    get_language_from_path,
                    get_language_from_request)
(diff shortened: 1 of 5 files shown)


escattone:
No problem. I personally prefer the functional style (when the there's no chance that the middleware might need to be sub-classed), but the class-style is great too. I'll update this with the class-style.

jwhitlock:
I don't have a strong preference for the class-style, I think you can pick and chose as appropriate. I was mostly thinking about a reduced diff, to figure out why tests are failing.

escattone:
The tests were failing because I had inadvertently removed a line involving `settings.ALLOWED_HOSTS` from two tests. That's been fixed and I also updated to use the class-style, which is nice too. Once the tests pass, this is ready for your review again @jwhitlock!

escattone:
FYI. The headless tests passed against my local dev (`pytest --base-url http://localhost:8000 tests/headless`).

kuma/core/middleware.py
```diff
@@ -126,16 +134,14 @@ class LocaleMiddleware(MiddlewareMixin):
...
     """
-    response_redirect_class = HttpResponseRedirect
 
-    def process_request(self, request):
-        """Activate the language, based on the request."""
-        language = get_language_from_request(request)
-        translation.activate(language)
-        request.LANGUAGE_CODE = language
+    def __call__(self, request):
+        # Activate the language, based on the request.
```
jwhitlock: You've made this comment redundant with the function name 🙂. However, it's not 100% clear from the name that ``request`` gains a ``LANGUAGE_CODE`` parameter, maybe that's worth including in the comment.

kuma/core/middleware.py
```diff
@@ -17,27 +15,35 @@
...
 from .utils import is_untrusted, urlparams
 from .views import handler403
 
 
-class LangSelectorMiddleware(MiddlewareMixin):
+class MiddlewareBase(object):
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        super(MiddlewareBase, self).__init__()
```
jwhitlock: I think the ``super````__init__`` is unneeded because the parent is ``object``

kuma/core/middleware.py
```diff
@@ -17,27 +15,35 @@
...
                    get_language_from_request)
 from .utils import is_untrusted, urlparams
 from .views import handler403
 
 
-class LangSelectorMiddleware(MiddlewareMixin):
+class MiddlewareBase(object):
+
+    def __init__(self, get_response):
+        self.get_response = get_response
```
jwhitlock: I like having this defined once.

kuma/core/middleware.py
```diff
@@ -276,14 +287,13 @@ def process_response(self, request, response):
...
-class RestrictedEndpointsMiddleware(MiddlewareMixin):
+class RestrictedEndpointsMiddleware(MiddlewareBase):
+    """Restricts the accessible endpoints based on the host."""
 
-    def process_request(self, request):
-        """
-        Restricts the accessible endpoints based on the host.
-        """
+    def __call__(self, request):
         if settings.ENABLE_RESTRICTIONS_BY_HOST and is_untrusted(request):
```
jwhitlock: It may make sense to move this to the ``__init__``:

```
def __init__(self, get_response):
    if not settings.ENABLE_RESTRICTIONS_BY_HOST:
        raise MiddlewareNotUsed
```

Or, I may be trying too hard to use every piece of the buffalo, and it would have a negative impact on tests...

escattone: LOL ("every piece of the buffalo") Good point! I remember reading about that new feature, and this may just be the case! I like it because if it's not ever going to be used (`settings.ENABLE_RESTRICTIONS_BY_HOST = False`), it doesn't remain embedded in the middleware stack.

escattone:
@jwhitlock Thanks for the review! I added some changes that you may want to review one more time.

jwhitlock:
Changes look good, thanks @escattone!

PR #4846, opened 2018-06-08
bug 1468271: Add custom performance metrics for interactive examples iframe

schalkneethling:
@jwhitlock before I change these to the `ga` calls I want to give @tkadlec a chance to review and try it out locally.

### DIFF
--- a/kuma/static/js/interactive.js
+++ b/kuma/static/js/interactive.js
@@ -7,6 +7,99 @@
         window.mdn.interactiveEditor.editorUrl ||
         'https://interactive-examples.mdn.mozilla.net';
 
+    /**
+     * Handles postMessages from the interactive editor, passing
+     * the data in the relevant format to trackEvent depending on the label.
+     * Also ensures that the origin of the message is what we expect.
+     * @param {Object} event - The event Object received from the postMessage
+     */
+    function interactiveExamplesMsgHandler(event) {
+        var allowedOrigin =
+            window.mdn.interactiveEditor.editorUrl ||
+            'https://interactive-examples.mdn.mozilla.net';
+
+        if (event.origin !== allowedOrigin) {
+            return false;
+        }
+
+        if (event.data.label === 'Performance Events') {
+            mdn.analytics.trackEvent({
+                category: event.data.category,
+                action: event.data.action,
+                label: event.data.label,
+                value: event.data.value - performance.timing.fetchStart,
+                hitType: 'timing'
+            });
+        } else {
+            mdn.analytics.trackEvent(event.data);
+        }
+    }
+
+    /**
+     * Returns the interactive examples iframe performance entry
+     * @param {Array} perfEntries - The array of performance entries
+     */
+    function getInteractiveExamplesPerfEntry(perfEntries) {
+        for (var i = 0, l = perfEntries.length; i < l; i++) {
+            var currentEntry = perfEntries[i];
+            if (
+                currentEntry.initiatorType &&
+                currentEntry.initiatorType === 'iframe' &&
+                currentEntry.name.indexOf(
+                    'interactive-examples.mdn.mozilla.net'
+                ) > -1
+            ) {
+                return perfEntries[i];
+            }
+        }
+    }
+
+    /* Ensure there is an iframe present and that performance is
+       supported before gathering performance metric */
+    if (iframe && performance !== undefined) {
+        document.addEventListener('readystatechange', function(event) {
+            if (event.target.readyState === 'complete') {
+                var interactiveExamplesPerfEntry = {};
+                var perfEntries = performance.getEntriesByType('resource');
+                var mainFetchStart = 0;
+                var iframeFetchStart = 0;
+                var iframeFetchStartSinceUnixEpoch = 0;
+                var timeToIframeFetchStart = 0;
+
+                if (perfEntries === undefined || perfEntries.length <= 0) {
+                    console.info('No performance entries was returned');
+                    return;
+                }
+
+                interactiveExamplesPerfEntry = getInteractiveExamplesPerfEntry(
+                    perfEntries
+                );
+
+                // ensure that the iframe was found in the array of performance entries
+                if (interactiveExamplesPerfEntry !== undefined) {
+                    mainFetchStart = performance.timing.fetchStart;
+                    iframeFetchStart = Math.round(
+                        interactiveExamplesPerfEntry.fetchStart
+                    );
+                    iframeFetchStartSinceUnixEpoch =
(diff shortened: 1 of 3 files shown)


kuma/static/js/interactive.js
```diff
@@ -7,6 +7,84 @@
...
+                // var iframeLoadTime = interactiveExamplesPerfEntry.responseEnd - interactiveExamplesPerfEntry.startTime;
+                // ga('send', {
+                //     'eventCategory' : 'Interactive Examples',
+                //     'eventAction'   : 'iFrame load time',
+                //     'eventLabel'    : 'Performance Events',
+                //     'eventValue'    : (iframeLoadTime / 1000),
+                //     'hitType': 'timing'
+                // });
+                console.log(
+                    (interactiveExamplesPerfEntry.responseEnd -
```
tkadlec: Can we collect the metric in milliseconds for additional precision? Maybe with `Math.round()` though to clean it up slightly.

kuma/static/js/interactive.js
```diff
@@ -7,6 +7,84 @@
...
+     * @param {Array} perfEntries - The array of performance entries
+     */
+    function getInteractiveExamplesPerfEntry(perfEntries) {
+        for (var i = 0, l = perfEntries.length; i < l; i++) {
+            var currentEntry = perfEntries[i];
+            if (
+                currentEntry.initiatorType &&
+                currentEntry.initiatorType === 'iframe' &&
+                currentEntry.name.indexOf(
+                    'interactive-examples.mdn.mozilla.net'
```
jwhitlock: I'm not thrilled with hard-coding the production URL, since it makes local testing impossible, but I'm no closer to a solution than I was in September https://github.com/mdn/interactive-examples/issues/257

kuma/static/js/interactive.js
```diff
@@ -7,6 +7,84 @@
...
+                //    new Date(mainFetchStart).getSeconds()
+                // ga('send', {
+                //     'eventCategory' : 'Interactive Examples',
+                //     'eventAction'   : 'Time to iframe fetch start',
+                //     'eventLabel'    : 'Performance Events',
+                //     'eventValue'    : timeToIframeFetchStart,
+                //     'hitType': 'timing'
+                // });
+                console.log(
+                    'Time in seconds between fetch start of main page, and fetch start of iframe',
```
tkadlec: Same thing here. Milliseconds would be nice for more precision. So something like

```js
new Date(iframeFetchStartSinceUnixEpoch) - new Date(mainFetchStart)
```

kuma/static/js/interactive.js
```diff
@@ -7,6 +7,84 @@
...
+                var mainFetchStart = 0;
+                var iframeFetchStart = 0;
+                var iframeFetchStartSinceUnixEpoch = 0;
+
+                if (perfEntries === undefined || perfEntries.length <= 0) {
+                    console.info('No performance entries was returned');
+                    return;
+                }
+
+                interactiveExamplesPerfEntry = getInteractiveExamplesPerfEntry(
```
jwhitlock: What if the entry is not found?

kuma/static/js/interactive.js
```diff
@@ -7,6 +7,84 @@
...
+                );
+
+                mainFetchStart = performance.timing.fetchStart;
+                iframeFetchStart = Math.round(
+                    interactiveExamplesPerfEntry.fetchStart
+                );
+                iframeFetchStartSinceUnixEpoch =
+                    mainFetchStart + iframeFetchStart;
+
+                // var iframeLoadTime = interactiveExamplesPerfEntry.responseEnd - interactiveExamplesPerfEntry.startTime;
```
jwhitlock: Do you want to ship the ``console.log`` version before enabling ``ga('send')``?

schalkneethling: Nope, those were just temporary.

kuma/static/js/interactive.js
schalkneethling: Good point. I mean, in the current context we will not reach this point unless there is an `iframe` with the class `.interactive` but, does not hurt to check that something was returned.

kuma/static/js/interactive.js
tkadlec: Both of these metrics are accurate for what they measure, but we're still missing the full picture of how long it takes the iframe to be usable. 

For example, I loaded https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach up on a simulated slow 3G connection.

It took 6569ms for the iframe to be fetched, and another 2435 for it's response to end. So if we looked at total time, we end up at 9004ms.

But, this is what the screen looks like at that point:

screen shot 2018-06-08 at 10 56 50 am

It takes another 10 seconds or so before it's load event has completed and we get a fully visible and usable iframe.

So I think we probably need to post iframe's `loadEventEnd` timestamp back to the parent window for this calculation to get the full picture.

schalkneethling:
@tkadlec and @jwhitlock updated. Thanks for the reviews o/\o

tkadlec:
Awesome! Just one more thing @schalkneethling that I didn't realize was "pending" instead of actually posted.

kuma/static/js/interactive.js
schalkneethling: @tkadlec So, just the time it took for the iframe to completely load correct? We are not adding it to anything, or subtracting. Just the raw number in milliseconds.

kuma/static/js/interactive.js
```diff
@@ -7,6 +7,99 @@
...
+                    console.info('No performance entries was returned');
+                    return;
+                }
+
+                interactiveExamplesPerfEntry = getInteractiveExamplesPerfEntry(
+                    perfEntries
+                );
+
+                // ensure that the iframe was found in the array of performance entries
+                if (interactiveExamplesPerfEntry !== undefined) {
```
jwhitlock: Thanks, this makes me feel safer.

jwhitlock:
I was able to see the first hit locally:

```
[Log] Running command: ga("send", {hitType: "timing", eventCategory: "Interactive Examples", eventAction: "Time to iframe fetch start", eventLabel: "Performance Events", eventValue: 4367, hitCallback: null}) (analytics_debug.js, line 10)
```

but not the second:
```
[Error] Unable to post message to https://developer.mozilla.org. Recipient has origin http://localhost:8000.

	postMessage
	trackEvent (editor-css.js:1:1694)
	trackFrameLoadTime (editor-css.js:1:2190)
	onload (editor-css.js:1:5034)
```

I think this is consistent with other event tracking that is hard-coded to the production environment. Let's deploy and see if it works 🤞

PR #4850, opened 2018-06-13
bug 1468490: Update to Debian 9 (stretch) base image

jwhitlock:
Update the Python base image from Debian 8 (jessie) to 9 (stretch), which is the current LTS release, along with some additional cleanup:

Move the `ENV` declarations to the top, to reduce the number of intermediate images.

Move all the package installation to the same section, to avoid refreshing the `apt-get` cache twice. Include the `gpg` package and dependencies, which is no longer in the default image. Update the names of the MySQL packages.

Update the `node.js` installation code to bump the minor version and use the current code for importing GPG signing keys. Drop the unused `yarn` install. The 8.x node.js update will come later.

### DIFF
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -1,17 +1,49 @@
-FROM python:2.7-slim-jessie
+FROM python:2.7-slim
 
-# extra python env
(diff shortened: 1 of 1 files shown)


MatonAnthony:
This looks cleaner 👍

PR #4851, opened 2018-06-15
Bug 1127798: Kuma test suite depends on dict order

MatonAnthony:
The kuma suite currently depends on dict order.

This is an attempt to remove this dependency, which would allow us to remove a considerable chunk of configuration and a better mirror of the real usage of the code.

### DIFF
--- a/kuma/wiki/tests/test_templates.py
+++ b/kuma/wiki/tests/test_templates.py
@@ -703,7 +703,7 @@ class NewRevisionTests(UserTestCase, WikiTestCase):
         # cause notification emails to error
         # and stop being sent.
         time.sleep(1)
-        eq_(2, len(mail.outbox))
+        assert 2 == len(mail.outbox)
         first_edit_email = mail.outbox[0]
         expected_to = [config.EMAIL_LIST_SPAM_WATCH]
         expected_subject = (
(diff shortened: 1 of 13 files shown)


MatonAnthony:
Beware, do not merge this until the commits are properly rebased to follow Kuma standard.

MatonAnthony:
After multiple tests run locally (10 times on the last iteration), I don't manage to trigger anymore errors.

Here is the test protocol I followed.

Enter the container
```bash
make bash
```
Define `PYTHONHASHSEED` to `random`
```bash
export PYTHONHASHSEED=random
```
Run the tests suite as many time as necessary
```bash
py.test kuma -vv
```

## Remaining work on this PR:
- Cleaning the commit tree by rebasing the branch into a single clean commit with proper name and message

jwhitlock:
Getting all the tests to pass is a huge step forward. Thanks @MatonAnthony! Let me know when you've rebased and cleaned up (feel free to combine my commit with yours), and are ready for review.

MatonAnthony:
I am happy with this now, the only thing I didn't managed to achieve is that I still have to setup
`PYTHONHASHSEED` to  `random` myself, I don't know where this variable is setup in the `dev` environment.

jwhitlock:
``tox`` added [special handling of PYTHONHASHSEED](https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed) to mimic Python 3.3's default behavior. The default in Python 2.7 is no hash randomization. We can get a similar effect in our tests by adding ``PYTHONHASHSEED=random`` to ``docker-compose.yml``, or even https://github.com/pytest-dev/pytest-randomly.

I think we can skip changing the defaults for this PR, and add it later if needed or just update to Python 3.

MatonAnthony:
Ok, let's skip it for this PR.

kuma/wiki/tests/test_templates.py
```diff
@@ -703,7 +703,7 @@ def test_new_revision_POST_document_with_current(self, get_current):
         # cause notification emails to error
         # and stop being sent.
         time.sleep(1)
-        eq_(2, len(mail.outbox))
+        assert 2 == len(mail.outbox)
```
jwhitlock: The Smallest of Nits: This is what some devs dismissively call Yoda order, because it sounds like Yoda from Start Wars to read it out loud ("assert 2 is the length of the mailbox I do!") rather than the "natural" order (assert the length of the outbox is 2").  There's a technical reason to prefer this order, especially in C (https://stackoverflow.com/q/9331259/10612), but the past Kuma devs have been switching to "natural" order as they touch code, ``assert actual == expected``.

I'm not expecting a change this time, just look for it in future test refactors.

PR #4854, opened 2018-06-18
bug 1468271: Change to hit events (not timing), and unique labels (timestamp+rand)

schalkneethling:
@jwhitlock r? @tkadlec r?

### DIFF
--- a/kuma/static/js/interactive.js
+++ b/kuma/static/js/interactive.js
@@ -8,6 +8,24 @@
         'https://interactive-examples.mdn.mozilla.net';
 
     /**
+     * Generate and returns a random string thanks to:
+     * https://stackoverflow.com/questions/1349404/generate-random-string-characters-in-javascript
+     */
+    function randomString() {
+        var text = '';
+        var possible =
+            'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+
+        for (var i = 0; i < 5; i++) {
+            text += possible.charAt(
+                Math.floor(Math.random() * possible.length)
+            );
+        }
+
+        return text;
+    }
+
+    /**
      * Handles postMessages from the interactive editor, passing
      * the data in the relevant format to trackEvent depending on the label.
      * Also ensures that the origin of the message is what we expect.
@@ -26,9 +44,8 @@
             mdn.analytics.trackEvent({
                 category: event.data.category,
                 action: event.data.action,
-                label: event.data.label,
-                value: event.data.value - performance.timing.fetchStart,
-                hitType: 'timing'
+                label: new Date().getTime() + '-' + randomString(),
+                value: event.data.value - performance.timing.fetchStart
             });
         } else {
             mdn.analytics.trackEvent(event.data);
@@ -91,9 +108,8 @@
                     mdn.analytics.trackEvent({
                         category: 'Interactive Examples',
                         action: 'Time to iframe fetch start',
-                        label: 'Performance Events',
-                        value: timeToIframeFetchStart,
-                        hitType: 'timing'
+                        label: new Date().getTime() + '-' + randomString(),
+                        value: timeToIframeFetchStart
                     });
                 }
             }


kuma/static/js/interactive.js
```diff
@@ -26,9 +26,8 @@
             mdn.analytics.trackEvent({
                 category: event.data.category,
                 action: event.data.action,
-                label: event.data.label,
-                value: event.data.value - performance.timing.fetchStart,
-                hitType: 'timing'
+                label: new Date().getTime(),
```
jwhitlock: This change confuses me. It appears you are attempting to register a unique event for each timing variable. This will prevent any aggregate statistics around the event, and make it so you have to post-process any data. Is this the intent? What exactly are you trying to track?

jwhitlock:
I don't miss a chance to do math on a Monday, so I tried to calculate the probability, using some guidance from https://math.stackexchange.com/questions/2460583/chance-of-overlapping-events-in-milliseconds-per-day. The math is beyond me, however. Using the Poisson-distribution assumption, I get a very low probability of collision (less than 0.0001%), but I get an almost certain collision with a birthday paradox approximation. For the approximation to be valid, the number of events (page hits per day) needs to be significantly less than the buckets (milliseconds per day), so maybe the approximation is bad in this case.

Adding a random number will make collisions much less likely, but not impossible.

I see a few options:

* When two events have the same timestamp, use the average. Advantage: you can just use the average for all events.
* When two events have the same timestamp, throw them away (unique count > 1). Advantage: you are only counting unique events, which is the whole point of this change
* Ship this and see if collisions are a big problem, and then decide to handle them or add a random number to make them less likely. It's possible we'll hit other problems with this approach long before collisions are the biggest issue.

atopal:
@tkadlec @jwhitlock since we are going to look at the 90th percentile or something similar, duplicates shouldn't matter much unless they are freakishly common, right?

schalkneethling:
Ok, @tkadlec and @jwhitlock I have added `Math.random` to be extra careful. I also rebased and fixed up the commit message. Thanks for the review.

atopal:
@schalkneethling instead of adding it to the time stamp, can you just add it as a pre/postfix? That way we can easily remove it after exporting the data and still get the exact time of the event. That will be useful when we run experiments that are started and finished during the day and not at the beginning/end of a day.

schalkneethling:
@atopal Done. Updated.

kuma/static/js/interactive.js
```diff
@@ -26,9 +26,8 @@
             mdn.analytics.trackEvent({
                 category: event.data.category,
                 action: event.data.action,
-                label: event.data.label,
-                value: event.data.value - performance.timing.fetchStart,
-                hitType: 'timing'
+                label: 'perf-events-' + new Date().getTime(),
```
jwhitlock: The string ``perf-events-`` does nothing to avoid collisions, so I think it should be dropped.

Here's some code that would generate a random string, borrowed from https://stackoverflow.com/questions/1349404/generate-random-string-characters-in-javascript :

```
function randString(len) {
  var text = "";
  var possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";

  for (var i = 0; i < min(len, 1); i++)
    text += possible.charAt(Math.floor(Math.random() * possible.length));

  return text;
}

...

label: new Date().getTime() + '-' + randString(4);
```

kuma/static/js/interactive.js
```diff
@@ -26,9 +44,8 @@
             mdn.analytics.trackEvent({
                 category: event.data.category,
                 action: event.data.action,
-                label: event.data.label,
-                value: event.data.value - performance.timing.fetchStart,
-                hitType: 'timing'
+                label: new Date().getTime() + '-' + randomString(),
```
jwhitlock: This worked for me in the browser console, generating labels like:

```
"1529419596413-Ij3Cy"
"1529419599067-p4UW5"
"1529419601135-NQtFu"
"1529419601945-tO8nQ"
"1529419602611-PuYzE"
"1529419605323-aUpuQ"
```

atopal: Perfect! That works for me 👍 :) Thanks guys!

tkadlec: 👍 Looks great!

PR #4858, opened 2018-06-18
bug 1469029: keep two zone-related css classes

escattone:
This is a companion PR to #4853. If/when that PR is merged/deployed, this PR should be merged/deployed as well.

The "death of zones" PR (#4853) removes all zone-related CSS. However, we'd actually like to keep two classes (see [bugzilla 1469029](https://bugzilla.mozilla.org/show_bug.cgi?id=1469029)). This PR adds those two classes into the mainstream CSS.

I wasn't sure of the best place for these two classes to live, so please suggest a different home if you don't like my choice.

### DIFF
--- a/kuma/static/styles/base/utilities.scss
+++ b/kuma/static/styles/base/utilities.scss
@@ -183,3 +183,11 @@ Utility classes
         @extend %offscreen-placeholder;
     }
 }
+
+.bump-left-padding {
+    padding-left: $grid-spacing * 2;
+}
+
+.light-grey-background {
+	background-color: $grey-light;
+}
--- a/kuma/static/styles/base/utilities.scss
+++ b/kuma/static/styles/base/utilities.scss
@@ -189,5 +189,5 @@ Utility classes
 }
 
 .light-grey-background {
-	background-color: $grey-light;
+    background-color: $grey-light;
 }
--- a/kuma/static/styles/components/wiki/customcss.scss
+++ b/kuma/static/styles/components/wiki/customcss.scss
@@ -578,5 +578,31 @@ pre span.comment {
     display: inline;
 }
 
+/*
+  Zone-related classes kept after the removal of zones.
+  (See https://bugzilla.mozilla.org/show_bug.cgi?id=1469029)
+ */
+
+/*
+  reference-values:
+
+  This gives an extra indent to a dl, when it's embedded inside another dl,
+  which comes up sometimes in the reference pages. The extra indent indicates
+  that the embedded dl is part of an embedded list, not the main list.
+ */
+dd > dl.reference-values {
+    padding-left: 40px;
+}
+
+/*
+  webextension-api-sidebar:
+
+  This applies a background color to the piece of a sidebar containing the
+  api items relevant to the current page. It helps people orient themselves
+  within the sidebar.
+ */
+li.webextension-api-sidebar {
+    background-color: #f4f7f8;
+}
 
 /*end!*/
--- a/kuma/static/styles/components/wiki/customcss.scss
+++ b/kuma/static/styles/components/wiki/customcss.scss
@@ -586,23 +586,32 @@ pre span.comment {
 /*
   reference-values:
 
+  NOTE: This has been DEPRECATED. New work should use the utility
+        class "bump-left-padding" instead of "reference-values",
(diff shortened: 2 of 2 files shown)


schalkneethling:
Hey @escattone 

These look like prime candidates for utility classes. I would therefore suggest two things:

1. Please them in https://github.com/mozilla/kuma/blob/master/kuma/static/styles/base/utilities.scss
2. Give the classes a more generic, but descriptive name with the comments added for clarity. This will make it’s way into the coding style guide, but that work is still going to take some time. The comments will therefore be great in the mean time.

kuma/static/styles/components/wiki/customcss.scss
```diff
@@ -578,5 +578,31 @@ pre span.comment {
...
+ */
+
+/*
+  reference-values:
+
+  This gives an extra indent to a dl, when it's embedded inside another dl,
+  which comes up sometimes in the reference pages. The extra indent indicates
+  that the embedded dl is part of an embedded list, not the main list.
+ */
+dd > dl.reference-values {
```
schalkneethling: This could for example be a general utility class called `bump-padding` and then defined as follows:

```
.bump-left-padding {
    padding-left: @grid-spacing * 2;
}
```
Ref: https://github.com/mozilla/kuma/blob/master/kuma/static/styles/includes/_vars.scss#L162

kuma/static/styles/components/wiki/customcss.scss
```diff
@@ -578,5 +578,31 @@ pre span.comment {
...
+}
+
+/*
+  webextension-api-sidebar:
+
+  This applies a background color to the piece of a sidebar containing the
+  api items relevant to the current page. It helps people orient themselves
+  within the sidebar.
+ */
+li.webextension-api-sidebar {
```
schalkneethling: This could be `light-grey-background`, and be implemented as:

```
.light-grey-background {
	background-color: $grey-light;
}
```
Ref: https://github.com/mozilla/kuma/blob/master/kuma/static/styles/includes/_vars.scss#L29

escattone: This requires that I also simultaneously update the `kumascript/macros/WebExtAPISidebar.ejs` macro, which I think has the only use of the `webextension-api-sidebar` class.

kuma/static/styles/components/wiki/customcss.scss
escattone: I like your utility class idea, but in order to do that we'd have to simultaneously change a significant number of reference-document wiki pages that are currently using the `reference-values` class. I think we should perhaps leave that for a future task?

escattone: Great idea, let's do that, thanks @schalkneethling!

kuma/static/styles/components/wiki/customcss.scss
escattone: 👍 thanks @schalkneethling!

wbamberg: Yeah, for this one renaming the class is easy. For `reference-values`, not so much :).

escattone:
Thanks for the review @schalkneethling and @wbamberg! This is ready for review again. I decided to keep the deprecated rules in `kuma/static/styles/components/wiki/customcss.scss` and, as suggested, place the new utility classes in `kuma/static/styles/base/utilities.scss`. Although not strictly DRY, I thought it would be cleaner, since we can simply delete the deprecated rules when we're done with them. Also, and perhaps more importantly, I didn't think the deprecated rules belonged in `kuma/static/styles/base/utilities.scss` nor the new utility classes in `kuma/static/styles/components/wiki/customcss.scss`.

kuma/static/styles/base/utilities.scss
```diff
@@ -183,3 +183,11 @@ Utility classes
         @extend %offscreen-placeholder;
     }
 }
+
+.bump-left-padding {
+    padding-left: $grid-spacing * 2;
+}
+
+.light-grey-background {
+	background-color: $grey-light;
```
jwhitlock: ``stylelint`` wants [4 spaces](https://travis-ci.org/mozilla/kuma/jobs/394784569#L809)

schalkneethling:
🎉 🎉

PR #4860, opened 2018-06-19
Add jinja helper for SVG. Fix issue 23 on sprints repo

schalkneethling:
This adds the Jinja helper for SVG. Also applies the new helper to currently identified use cases. Open question is whether we should make `include_svg` the standard way to include SVG files in HTML or only use it when we want to pass a custom title.

### DIFF
--- a/kuma/wiki/templatetags/jinja_helpers.py
+++ b/kuma/wiki/templatetags/jinja_helpers.py
@@ -9,10 +9,11 @@ from constance import config
 from cssselect.parser import SelectorSyntaxError
 from django.contrib.sites.models import Site
 from django.core.serializers.json import DjangoJSONEncoder
+from django.template import loader
+from django.utils import lru_cache
 from django.utils.html import conditional_escape
 from django.utils.translation import ugettext
 from django_jinja import library
-from lxml import etree
 from pyquery import PyQuery as pq
 
 from kuma.core.urlresolvers import reverse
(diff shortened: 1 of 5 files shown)


kuma/wiki/templatetags/jinja_helpers.py
```diff
@@ -263,3 +264,32 @@ def wiki_url(path):
...
+
+@library.global_function
+def include_svg(path, title=None):
+    """
+    Load SVG from file system. If a title was passed, replace the value of
+    the current title tag with the new title. Return the SVG element
+    """
+    env = jinja2.Environment(loader=jinja2.FileSystemLoader('./jinja2'))
+    tmplString = env.get_template(path).render()
+    svgMarkup = jinja2.Markup(pq(tmplString).outerHtml())
```
jwhitlock: This should move to an ``else:`` branch for when the title is unset, otherwise you doing a bit of work to create ``svgMarkup`` twice, and throwing one away.

kuma/wiki/templatetags/jinja_helpers.py
```diff
@@ -263,3 +264,32 @@ def wiki_url(path):
...
+
+def get_bare_tag(elem):
+    """
+    Returns the tag without the leading namespace
+    """
+    return elem.tag.rsplit('}', 1)[-1]
+
+
+@library.global_function
+def include_svg(path, title=None):
```
jwhitlock: I'm a little worried about the performance - this is called 5x per page, and uses some potentially expensive methods - but premature optimization is usually a bad idea. It should be easy to add Django's ``@lru_cache``, or look for in-code optimizations, if it looks problematic.

I do want some tests though, such as:

* ``include_svg('includes/icons/social/twitter.svg')``
* ``include_svg('includes/icons/social/twitter.svg', 'New Title')``
* ``include_svg('includes/icons/social/twitter.svg', 'Nuevo Título')``

Tests can be hard to write - I can add in a new commit if desired.

schalkneethling: @jwhitlock Is there more to `lru_cache` than adding the `@lru_cache.lru_cache()` decorator?

jwhitlock: It could be that easy, or you may need to do work to make the function arguments hashable. Here's the docs for [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache), which Django backports for Python 2.7.

schalkneethling:
@jwhitlock Ok, I have made updates to the PR. Two things I did not do

1. Add `lru_cache`
2. Add tests

As I need to focus on the performance beaconing work, I would really appreciate it if you could handle those two items in order to enable this PR to be merged. Thanks in advance.

schalkneethling:
@jwhitlock Thanks for helping to get this one across the finish line o/\o

PR #4862, opened 2018-06-20
Issue sprints/143, send additional timing events

schalkneethling:
@jwhitlock @tkadlec r?

### DIFF
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -18,46 +18,13 @@
                 win.ga('set', data.dimension, data.value);
             }
         },
-        /*
-            Tracks generic events passed to the method
-        */
-        trackEvent: function(eventObject, callback) {
-            // Submit eventArray to GA and call callback only after tracking has
-            // been sent, or if sending fails.
-            //
-            // callback is optional.
-
-            /*
-                Format:
-
-                    ga('send', {
-                        'eventCategory' : 'Star Trek',
-                        'eventAction'   : 'Fire',
-                        'eventLabel'    : 'Phasers',
-                        'eventValue'    : 100,
-                        'hitCallback'   : function () {
-                            document.location = href;
-                        },
-                        'hitType': 'event'
-                    });
-            */
-
+        /**
+         * Sends a statistic to GA dependant on the status of the win.ga object
+         * @param {object} data - The data to send to GA
+         * @param {function} [callback] - The function to call after sending data
+         */
+        send: function(data, callback) {
             var ga = win.ga;
-            var data = {
-                hitType: eventObject.hitType || 'event',
-                eventCategory: eventObject.category || '',    // Required.
-                eventAction: eventObject.action || '',             // Required.
-                eventLabel: eventObject.label || '',
-                eventValue: eventObject.value || 0,
-                hitCallback: callback || null
-            };
-
-            // task-completion.js sends an additional metric
-            if(eventObject.nonInteraction !== undefined) {
-                // append it to the data object sent to GA
-                data.nonInteraction = eventObject.nonInteraction;
-            }
-
             // If Analytics has loaded, go ahead with tracking
             // Checking for ".create" due to Ghostery mocking of ga
             if(ga && ga.create) {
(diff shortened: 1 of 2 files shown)


kuma/static/js/interactive.js
```diff
@@ -41,12 +41,21 @@
...
                 label: new Date().getTime() + '-' + randomString(),
                 value: event.data.value - performance.timing.fetchStart
             });
+
+            // one of hitType: timing
+            mdn.analytics.trackTiming({
+                category: event.data.category,
+                timingVar: event.data.action,
+                value: event.data.value - performance.timing.fetchStart,
+                label: new Date().getTime() + '-' + randomString()
```
tkadlec: We don't have to be as funky here. Since it's a timing event, GA won't add all the values together so a unique label is not necessary. It's an optional string used to help in reporting. I think we can probably just ditch it altogether.

@jwhitlock what say you sir?

schalkneethling: As this field is also optional, I reckon do agree that we can simply leave it. Thanks for the feedback @tkadlec 

Will wait on @jw	 feedback as well before making the change though.

kuma/static/js/analytics.js
```diff
@@ -18,6 +18,36 @@
                 win.ga('set', data.dimension, data.value);
             }
         },
+        /**
+         * Sends a statistic to GA dependant on the status of the win.ga object
+         * @param {object} data - The data to send to GA
+         * @param {function} [callback] - The function to call after sending data
+         */
+        send: function(data, callback) {
```
jwhitlock: I had lots of comments about this code, and then I saw that you just moved it, and the comment about Ghostery, and it all came back to me. Web is hard,

kuma/static/js/interactive.js
jwhitlock: Yes let's not set the label since is is optional. A similar change may be needed above to not send a label when omitted.

kuma/static/js/analytics.js
```diff
@@ -57,31 +85,36 @@
...
+                        },
+                        'hitType': 'timing'
+                    });
+            */
+            var data = {
+                hitType: 'tming',
+                timingCategory: timingObject.category || '', // Required.
+                timingVar: timingObject.timingVar || '', // Required.
+                timingValue: timingObject.value || 0,  // Required.
+                timingLabel: timingObject.label || '',
```
jwhitlock: I think something like this would be better, to make the label optional:

```
           var data = {
               hitType: 'tming',
               timingCategory: timingObject.category || '', // Required.
               timingVar: timingObject.timingVar || '', // Required.
               timingValue: timingObject.value || 0,  // Required.
           };
           if (timingObject.label) {
               data.timingLabel = timingObject.label;
           }
```

Please forgive if my JS looks like Python.

That way, if the label is omitted or falsy it is not sent. It is possible that Google treats an empty string the same as an omitted label, but it is certain they treat an omitted key as omitted.

kuma/static/js/analytics.js
```diff
@@ -57,31 +85,36 @@
...
+                        'timingValue'    : 100,
+                        'timingLabel'    : 'interactive-examples',
+                        'hitCallback'   : function () {
+                            document.location = href;
+                        },
+                        'hitType': 'timing'
+                    });
+            */
+            var data = {
+                hitType: 'tming',
```
jwhitlock: I suspect this should be ``"timing"``, not ``"tming"``.

kuma/static/js/interactive.js
```diff
@@ -105,12 +114,21 @@
...
                         label: new Date().getTime() + '-' + randomString(),
                         value: timeToIframeFetchStart
                     });
+
+                    // one of hitType: timing
+                    mdn.analytics.trackTiming({
+                        category: 'Interactive Examples',
+                        timingVar: 'Time to iframe fetch start',
+                        value: timeToIframeFetchStart,
+                        label: new Date().getTime() + '-' + randomString()
```
jwhitlock: We should drop this label as well, same reasons as above.

jwhitlock:
I think the "issue#143" in the name of the branches is now confusing the github bot:

https://bugzilla.mozilla.org/show_bug.cgi?id=143#c4

PR #4863, opened 2018-06-20
bug 1465085: Handle a child of a redirect with a full URL without raising an error

safwanrahman:
I have fixed the issue in `utils.py` as well as `get_redirect_url`. In `get_redirect_url`, it was returning full url sometimes, sometimes relative path. It has been fixed.
@jwhitlock @escattone r?

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -16,6 +16,7 @@ from django.utils.decorators import available_attrs
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext, ugettext_lazy as _
 from pyquery import PyQuery
+from six.moves.urllib.parse import urlparse
 from taggit.managers import TaggableManager
 from taggit.models import ItemBase, TagBase
 from taggit.utils import edit_string_for_tags, parse_tags
@@ -1292,7 +1293,9 @@ Full traceback:
                 # i.e allow "https://developer...." and "/en-US/docs/blah"
                 if len(url) > 1:
                     if url.startswith(settings.SITE_URL):
-                        return url
+                        parsed = urlparse(url)
+                        # Always return relative path instead of full url
+                        return parsed.path
                     elif url[0] == '/' and url[1] != '/':
                         return url
                 elif len(url) == 1 and url[0] == '/':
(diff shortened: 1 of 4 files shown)


kuma/wiki/utils.py
```diff
@@ -64,7 +64,11 @@ def get_doc_components_from_url(url, required_locale=None, check_host=True):
     # Extract locale and path from URL:
     parsed = urlparse(url)  # Never has errors AFAICT
     if check_host and parsed.netloc:
-        return False
+        # If its not redirecting to our domain,
```
jwhitlock: Nit: Change ``its`` to ``it's``. The contraction ``it's`` is used when it can be replaced by ``it is`` or ``it has``. The possessive ``its`` is used for ownership like "I tracked the horse by following its hoof prints", in a similar way to "his hoof prints" or "her hoof prints".

kuma/wiki/utils.py
```diff
@@ -64,7 +64,11 @@ def get_doc_components_from_url(url, required_locale=None, check_host=True):
...
+        # It should not be our wiki document
```
jwhitlock: Nit: "It" should not be capitalized, since it doesn't start a new sentence. This would fit on one line, avoiding the issue:

```
# Only allow redirects on our domain
```

kuma/wiki/models.py
```diff
@@ -1292,7 +1293,9 @@ def get_redirect_url(self):
                 # i.e allow "https://developer...." and "/en-US/docs/blah"
                 if len(url) > 1:
                     if url.startswith(settings.SITE_URL):
-                        return url
+                        parsed = urlparse(url)
```
jwhitlock: It seems to me this is the only change needed to fix the bug, and the change to ``get_doc_components_from_url`` is unneeded. Or, this change is unneeded if you change ``get_doc_components_from_url``...

In either case, this breaks a test, as reported by TravisCI.

jwhitlock:
Force-pushed to rebase, make several changes:

* Add commits to refactor the relevant tests to pytest style, and to expand some test cases
* Used ``from six.moves.urllib.parse import urlparse``, for Python 3 compatibility
* Fixed ``get_doc_components_from_url`` to return components for full URLs to local documents
* Other nits

safwanrahman:
Thank @jwhitlock for rebasing and upgrading the tests. I am hoping to fix the [bug 1435301](https://bugzilla.mozilla.org/show_bug.cgi?id=1435301)

PR #4866, opened 2018-06-21
bug 1467518: Use random hashes in Python 3.6 build

jwhitlock:
Code that depends on a fixed dictionary hash was removed in PR #4851. This removes it from the optional Python build, and removes the seed/no seed branch from the test script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,6 @@ matrix:
             CREATE_DB=kuma
             INSTALL_PIPELINE=1
             INSTALL_ELASTICSEARCH=1
-            PYTHONHASHSEED=0
     allow_failures:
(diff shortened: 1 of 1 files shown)


jwhitlock:
Failing test is failing 😄 

It's not this change. [django-memcached-hashring](https://github.com/jezdez/django-memcached-hashring) doesn't support Python 3.6.  It's also 5 years old. It also provides a feature we don't use - I'm pretty sure we have a single memcached server in production.

I think the next step is to switch to a recommended Django [memcached backend](https://docs.djangoproject.com/en/2.0/topics/cache/#memcached) and remove ``memcached_hashring``. @MatonAnthony might get to this before I do.

PR #4870, opened 2018-06-23
bug 1431497: Migrate from memcache to redis

MatonAnthony:
Redis is a better suited cache than memcache for our usage.

We have hit the memcache technical limitations such as:
    - 1MB values
    - Simple string storage
    - Lack of key introspections

All those features are provided by Redis while also providing the same
feature set as memcache.

Redis is already used by Kuma and removing memcache also means reducing
our infrastructure costs.

## Kumascript
To be able to entirely remove `memcache` from MDN infrastructure, we need to switch mdn/kumascript to `redis` too.

This PR is not done at this stage and I'm currently not specifically working on it.

## Python 3
Removing `memcache` is having the expected side-effects of remove `memcache_hashring` non-compatibility with Python 3.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -33,11 +33,11 @@ services:
       - ./:/app
     command: sh -c "urlwait && py.test --nomigrations kuma --ignore=kuma/search"
     depends_on:
-      - memcached
       - mysql
+      - redis
     environment:
       - DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
-      - MEMCACHE_SERVERS=memcached:11211
+      - REDIS_CACHE_SERVERS=redis://redis:6379/3
       - PYTHONDONTWRITEBYTECODE=1
       - URLWAIT_TIMEOUT=300
 
@@ -48,12 +48,12 @@ services:
       - ./:/app
     command: sh -c "urlwait && py.test --nomigrations --junit-xml=/app/test_results/django.xml kuma"
     depends_on:
-      - memcached
+      - redis
       - mysql
       - elasticsearch
     environment:
       - DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
       - ES_URLS=elasticsearch:9200
-      - MEMCACHE_SERVERS=memcached:11211
+      - REDIS_CACHE_SERVERS=redis://redis:6379/3
       - PYTHONDONTWRITEBYTECODE=1
       - URLWAIT_TIMEOUT=300
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -37,7 +37,7 @@ services:
       - redis
     environment:
       - DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
-      - REDIS_CACHE_SERVERS=redis://redis:6379/3
+      - REDIS_CACHE_SERVER=redis://redis:6379/3
       - PYTHONDONTWRITEBYTECODE=1
       - URLWAIT_TIMEOUT=300
 
@@ -54,6 +54,6 @@ services:
     environment:
       - DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
       - ES_URLS=elasticsearch:9200
-      - REDIS_CACHE_SERVERS=redis://redis:6379/3
+      - REDIS_CACHE_SERVER=redis://redis:6379/3
       - PYTHONDONTWRITEBYTECODE=1
       - URLWAIT_TIMEOUT=300
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7-slim
+FROM python:2.7-slim-stretch
 
 # Set the environment variables
 ENV NODE_VERSION=6.14.2 \
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7-slim-stretch
+FROM python:2.7-slim
 
 # Set the environment variables
 ENV NODE_VERSION=6.14.2 \
--- a/kuma/core/cache.py
+++ b/kuma/core/cache.py
@@ -1,4 +1,0 @@
-from django.core.cache import caches
-
-# just a helper to not have to redefine that all over the place
-redis = caches['redis']
--- a/kuma/core/cache.py
+++ b/kuma/core/cache.py
@@ -1,4 +1,4 @@
 from django.core.cache import caches
 
 # just a helper to not have to redefine that all over the place
-memcache = caches['memcache']
+redis = caches['redis']
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -118,9 +118,8 @@ CACHES = {
         'BACKEND': 'django_redis.cache.RedisCache',
         'TIMEOUT': CACHE_COUNT_TIMEOUT * 60,
         'KEY_PREFIX': CACHE_PREFIX,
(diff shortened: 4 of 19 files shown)


docker-compose.test.yml
```diff
@@ -38,6 +38,7 @@ services:
     environment:
       - DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
       - MEMCACHE_SERVERS=memcached:11211
```
MatonAnthony: I probably can remove this now.

MatonAnthony:
Ok, the travis build is broken, I don't exactly understand why but it's running as expected locally.

So either I'm missing something when I run it locally or I broke travis.

jwhitlock:
Thank you @MatonAnthony for starting on this. It's might be next week before I get a chance to review it.

I believe we're already using Redis, at least for Celery. I'm surprised a new requirement was needed.

We'll still need memcached in the development environment, because KumaScript uses it for caching. My preference is to remove the caching feature from KumaScript, but I haven't thought about the scope or impact of that change.

MatonAnthony:
I wouldn't say the new dependency is 100% required, we could probably deal without it but it provides the interface for Django caching which made the change minimal.

I have no issue working on the equivalent PR for Kumascript if we want to get rid of memcache quickly. I have not looked at the Kumascript code so I don't know how complicated it would be.

docker/images/kuma_base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:2.7-slim
+FROM python:2.7-slim-stretch
```
MatonAnthony: This change is no longer needed

kuma/settings/common.py
```diff
@@ -1260,15 +1260,15 @@ def pipeline_one_scss(slug, **kwargs):
 
 if MAINTENANCE_MODE:
     # In maintenance mode, we're going to avoid using the database, and
-    # use Celery's default beat-scheduler as well as memcached for storing
+    # use Celery's default beat-scheduler as well as redis for storing
```
jwhitlock: I think this should be "Redis" with a capital R

jwhitlock:
@MatonAnthony there's a lot of changes here. Alternatively, you can abandon this PR, and open a new one that just gets rid of ``django-memcached-hashring``, in favor of a library with Python 3 support.

MatonAnthony:
I would prefer to make the changes for this PR to move forward. It would be a step towards a better cache management.

kuma/core/cache.py
```diff
@@ -1,4 +1,4 @@
 from django.core.cache import caches
 
 # just a helper to not have to redefine that all over the place
-memcache = caches['memcache']
+redis = caches['redis']
```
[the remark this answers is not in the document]
MatonAnthony: I agree, I would be interested to get rid of this file too.

I was not confident to do it, I was in the mindset _"it's there for a reason"_. I think it would be a manageable change to make before merging this.

MatonAnthony:
[dcd1d9](https://github.com/mozilla/kuma/pull/4870/commits/dcd1d9a56474998880121d1d3995c00f68d8b94c) is addressing the minimum required changes for this PR to move foward.

[825f538](https://github.com/mozilla/kuma/pull/4870/commits/825f5380a860e0298f422c791bfd150b4ab11f55) is removing the caches.py utility file and remove the local memory based caching from the configuration (and the code)

MatonAnthony:
# Yet to be done:
- The refactoring of `clean_sessions` in a context manager based implementation.
- Remove the lock on `update_feeds`.
    - Is it just a matter of removing the annotation, or is there more to it?
- The refactoring of `update_community_stats()` into a django-cacheback job.
    - I have not really looked at it yet so I don't know how complicated it is.

jwhitlock:
Does this look like a good rebase?

https://github.com/mozilla/kuma/compare/master...jwhitlock:1431497-memcache-to-redis?expand=1

MatonAnthony:
It looks perfect to me

jwhitlock:
Thanks, force-pushed. Use ``git fetch; git reset @{u} --hard`` to reset to the upstream version.

jwhitlock:
Force-pushed for changes from PR #4853.

kuma/settings/common.py
```diff
@@ -1203,16 +1197,15 @@ def pipeline_one_scss(slug, **kwargs):
...
     # acquire the Celery settings from among Django's settings.
     CELERYBEAT_SCHEDULER = 'celery.beat.PersistentScheduler'
     DEFAULT_CELERY_RESULT_BACKEND = (
-        'cache+memcached://' + ';'.join(
-            config('MEMCACHE_SERVERS',
-                   default='127.0.0.1:11211',
-                   cast=Csv())
+        'redis://' + ';'.join(
+            config('REDIS_CACHE_SERVER',
+                   default='127.0.0.1:6379')
```
jwhitlock: This looks like it won't work, will return something like ``1;2;7;.;0;.0;...``. I'll need to try in ``MAINTENANCE_MODE`` to be sure. I also suspect we want to continue using the default rather than override.

kuma/wiki/kumascript.py
```diff
@@ -169,11 +168,11 @@ def get(document, cache_control, base_url, timeout=None):
             # Cache the request for conditional GET, but use the max_age for
             # the cache timeout here too.
             headers = response.headers
-            memcache.set(etag_key, headers.get('etag'), timeout=max_age)
-            memcache.set(modified_key, headers.get('last-modified'), timeout=max_age)
-            memcache.set(body_key, body.encode('utf-8'), timeout=max_age)
+            cache.set(etag_key, headers.get('etag'), timeout=max_age)
```
jwhitlock: This caching stuff works with Redis. I'm also sad we're doing this 😢

kuma/settings/common.py
jwhitlock: Confirmed that this results in ``DEFAULT_CELERY_RESULT_BACKEND=redis://r;e;d;i;s;:;/;/;r;e;d;i;s;:;6;3;7;9;/;3``. It should be:

```
      DEFAULT_CELERY_RESULT_BACKEND = REDIS_CACHE_SERVER
```

kuma/settings/common.py
```diff
@@ -1203,18 +1197,13 @@ def pipeline_one_scss(slug, **kwargs):
...
     # acquire the Celery settings from among Django's settings.
     CELERYBEAT_SCHEDULER = 'celery.beat.PersistentScheduler'
-    DEFAULT_CELERY_RESULT_BACKEND = (
-        'cache+memcached://' + ';'.join(
-            config('MEMCACHE_SERVERS',
-                   default='127.0.0.1:11211',
-                   cast=Csv())
-        )
-    )
+    DEFAULT_CELERY_RESULT_BACKEND = config('REDIS_CACHE_SERVER')
```
jwhitlock: This needs the same default as above. Maybe it would be better to write it as:

```
DEFAULT_CELERY_RESULT_BACKEND = CACHES['default']['LOCATION']
```

requirements/default.txt
```diff
@@ -106,11 +106,6 @@ django-jinja==2.4.1 \
     --hash=sha256:8a49d73de616a12075eee14c6d3bbab936261a463457d40348d8b8e2995cfbed \
     --hash=sha256:ceaa0eeebc4d91a5800967e50f4f087f0b6457503e3c2af85dc199bed8732a9a
 
-# Use consistent hashes despite adding or removing servers
-django-memcached-hashring==0.1.2 \
-    --hash=sha256:f71f99d60e704cc239bddbbe336dcdc5bb8a487b65a1b0b6c7ce0f4af5848442 \
-    --hash=sha256:14c2dc29bc693feaf114dea8c871f4d25daca72bba8f37b159d552fed4ab6337
```
jezdez: Huzzah! I can retire that now 🙌

PR #4871, opened 2018-06-25
Add handler for additional post messages from interactive examples, and a refactor

schalkneethling:
@tkadlec r? This is the other side of https://github.com/mdn/interactive-examples/pull/977 - I also did some refactoring to make the code more modular and easier to understand/read.

@jwhitlock Opening this for code review by Tim, I first need to have https://github.com/mdn/interactive-examples/pull/977 reviewed and merged so that I can test this properly. So, will mark as WIP for now.

### DIFF
--- a/kuma/static/js/utils/post-message-handler.js
+++ b/kuma/static/js/utils/post-message-handler.js
@@ -1,0 +1,82 @@
+/**
+ * Beacons events of type 'Performance Events' to GA
+ * @param {Object} data - The data associated with this event
+ */
+function handlePerformanceEvents(data) {
+    // one of hitType: event
+    mdn.analytics.trackEvent({
+        category: data.category,
+        action: data.action,
+        label: new Date().getTime() + '-' + mdn.utils.randomString(5),
+        value: data.value - performance.timing.fetchStart
+    });
+
+    // one of hitType: timing
+    mdn.analytics.trackTiming({
+        category: data.category,
+        timingVar: data.action,
+        value: data.value - performance.timing.fetchStart
+    });
+}
+
+/**
+ * Processes marks and sends beacons to GA based on mark value
+ * @param {Object} perfData - Object containing performance mark/measure information
+ */
+function handlePerfMarks(perfData) {
+    // if there is no `measureName` property, just set a mark
+    if (!perfData.measureName) {
+        window.mdn.perf.setMark(perfData.markName);
+    } else {
+        window.mdn.perf.setMark(perfData.markName);
+        window.mdn.perf.setMeasure({
+            measureName: perfData.measureName,
+            startMark: perfData.startMark,
+            endMark: perfData.endMark
+        });
+
+        mdn.analytics.trackTiming({
+            category: 'RUM - Interactive Examples',
+            timingVar: perfData.measureName,
+            value: window.mdn.perf.getDuration(perfData.measureName)
+        });
+    }
+}
+
+window.mdn.postMessageHandler = {
+    /**
+     * Handles postMessages from the interactive editor, passing
+     * the data in the relevant format to trackEvent depending on the label.
+     * Also ensures that the origin of the message is what we expect.
+     * @param {Object} event - The event Object received from the postMessage
+     */
+    interactiveExamplesMsgHandler: function(event) {
+        'use strict';
+
+        var allowedOrigin =
+            window.mdn.interactiveEditor.editorUrl ||
+            'https://interactive-examples.mdn.mozilla.net';
+        var eventData = event.data;
+
+        if (event.origin !== allowedOrigin) {
+            return false;
+        }
+
+        console.info('====== eventData.label ======', eventData.label);
+
+        if (eventData.label === 'Performance Events') {
+            handlePerformanceEvents(eventData);
+        } else if (eventData.markName) {
+            handlePerfMarks(eventData);
+        } else {
+            mdn.analytics.trackEvent(eventData);
+        }
+    }
+};
+
+// handles postMessages from the interactive editor
+window.addEventListener(
+    'message',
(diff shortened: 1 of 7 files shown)


kuma/static/js/utils/utils.js
```diff
@@ -0,0 +1,21 @@
+window.mdn.utils = {
+    /**
+     * Generate and returns a random string thanks to:
+     * https://stackoverflow.com/questions/1349404/generate-random-string-characters-in-javascript
+     * @param {Number} strLength - The length of the string to return
+     * @return {String} a randomly generated string with a chracter count of `strLength`
+     */
+    randomString: function(strLength) {
```
jwhitlock: This refactored code works in the console. ``randomString()`` now returns an empty string, but all the calls have been updated to ``mdn.utils.randomString(5)``.  I have a slight preference for a default, like ``strLength = Math.min(5, strLength)``.

kuma/static/js/utils/post-message-handler.js
```diff
@@ -0,0 +1,82 @@
...
+        if (event.origin !== allowedOrigin) {
+            return false;
+        }
+
+        console.info('====== eventData.label ======', eventData.label);
+
+        if (eventData.label === 'Performance Events') {
+            handlePerformanceEvents(eventData);
+        } else if (eventData.measureName) {
+            handlePerfMarks(eventData);
```
jwhitlock: This appears to be the new functionality.

kuma/static/js/utils/post-message-handler.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Processes marks and sends beacons to GA based on mark value
+ * @param {Object} perfData - Object containing performance mark/measure information
+ */
+function handlePerfMarks(perfData) {
+    // if there is no `measureName` property, just set a mark
+    if (!perfData.measureName) {
+        window.mdn.perf.setMark(perfData.markName);
```
jwhitlock: Based on current logic, we'll never run this branch - it is only called when ``measureName`` is defined.

kuma/static/js/utils/post-message-handler.js
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+        window.mdn.perf.setMark(perfData.markName);
+        window.mdn.perf.setMeasure(
```
jwhitlock: This doesn't match the definition of ``setMeasure``, which is expecting a single ``measureData`` value. I have a slight preference for changing ``setMeasure`` to expect these three variables.

kuma/static/js/utils/utils.js
schalkneethling: Yeah, I agree. My initial thought was to provide a default if nothing is passed, I should do that.

kuma/static/js/utils/post-message-handler.js
schalkneethling: Doh!, that should be `markName`

kuma/static/js/utils/post-message-handler.js
schalkneethling: Doh! again. This is why I specifically mentioned that this should not be merged and is largely untested code :) It should be passing an object of course. Thanks @jwhitlock

schalkneethling:
@jwhitlock This has been updated, thanks for the review. This should still not be merged until https://github.com/mdn/interactive-examples/pull/977 has been, and I have had a chance to do some testing locally.

I have also updated the commit message to include the bug number reference.

schalkneethling:
Alrighty, I have now tested this locally, and data is being sent to GA as expected for all editors, see below:

## Tabbed Editor

tabbed-editor

## CSS Editor

css-editor

## JS Editor

js-editor

schalkneethling:
Also, User Timing can be seen collected in devtools:

screen shot 2018-06-26 at 13 56 58

With that I am calling this ready to merge and push, baring any additional review comments.

kuma/static/js/utils/post-message-handler.js
```diff
@@ -0,0 +1,82 @@
...
+        window.mdn.perf.setMeasure({
+            measureName: perfData.measureName,
+            startMark: perfData.startMark,
+            endMark: perfData.endMark
+        });
+
+        mdn.analytics.trackTiming({
+            category: 'RUM - Interactive Examples',
+            timingVar: perfData.measureName,
+            value: window.mdn.perf.getDuration(perfData.measureName)
```
jwhitlock: The [type of the timing event is an integer](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#timingValue), but this is passing a float. We should truncate it on our side, like ``Math.round(window.mdn.perf.getDuration(perfData.measureName))``, or adjust ``get_duration`` to return an integer.

kuma/static/js/utils/post-message-handler.js
jwhitlock: 👍 I see ``interactiveExamplesMsgHandler`` changed to check for ``markName``.

kuma/static/js/utils/post-message-handler.js
schalkneethling: Aha, good catch. I did not know it had this restriction/requirement. We can simply ensure that we never return a floating point in a central location.

schalkneethling:
@jwhitlock So, the `ie-time-*` values are from inside the `iframe` itself. `Time to iframe fetch start` is from outside i.e. the parent page.

schalkneethling:
@jwhitlock Updated. Thanks!

jwhitlock:
I thought if I adjusted the subject line of the merge commit, then [bug 143](https://bugzilla.mozilla.org/show_bug.cgi?id=143) wouldn't be updated. I guess I needed to remove more than just the hash ☹️ . Try to omit "issue 143" from the next branch name.

PR #4873, opened 2018-06-28
add handler for interactive examples load end event, bug 1468271

schalkneethling:
@jwhitlock @tkadlec Here is the Kuma side of the final metric we want to track for now.

## GA Debug info

screen shot 2018-06-28 at 13 45 14

## User Timing in Devtools

screen shot 2018-06-28 at 13 47 57

### DIFF
--- a/kuma/static/js/utils/post-message-handler.js
+++ b/kuma/static/js/utils/post-message-handler.js
@@ -8,14 +8,14 @@ function handlePerformanceEvents(data) {
         category: data.category,
         action: data.action,
         label: new Date().getTime() + '-' + mdn.utils.randomString(5),
-        value: data.value - performance.timing.fetchStart
+        value: data.value - performance.timing.navigationStart
     });
 
     // one of hitType: timing
     mdn.analytics.trackTiming({
         category: data.category,
         timingVar: data.action,
-        value: data.value - performance.timing.fetchStart
+        value: data.value - performance.timing.navigationStart
     });
 }
 
@@ -51,7 +51,7 @@ function handlePerfMarks(perfData) {
  * Called by `handlePerfMarks` when a `markName` contains the
  * string `ie-load-event-end`. This sets a new mark, and a new
  * measure. It then uses this information to expose the total
- * duration from `fetchStart` of the parent document, until the
+ * duration from `navigationStart` of the parent document, until the
  * interactive example has reached `loadEventEnd`
  * @param {Object} perfData - Object containing performance mark/measure information
  */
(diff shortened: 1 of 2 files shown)


kuma/static/js/utils/post-message-handler.js
```diff
@@ -43,6 +47,33 @@ function handlePerfMarks(perfData) {
...
+ */
+function setLoadEventEnd(perfData) {
+    var measureName = perfData.markName + '-measure';
+    // set a mark
+    window.mdn.perf.setMark(perfData.markName);
+    /* set a performance measure that is the duration from
+       fetchStart until the interactive editor loaded */
+    window.mdn.perf.setMeasure({
+        measureName: measureName,
+        startMark: 'navigationStart',
```
jwhitlock: In the comment you say it is from ``fetchStart``, which is what I was expecting from other performance metrics, but this is using ``navigtionStart``. Also, I learned that this could be the [name of a measurement](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#Syntax).

schalkneethling: Yes, this is indeed the name of a performance measurement. So I can, and perhaps should be more verbose, as the documentation for `navigationStart` states:

"...representing the moment, in miliseconds since the UNIX epoch, right after the prompt for unload terminates on the previous document in the same browsing context. If there is no previous document, this value will be the same as PerformanceTiming.fetchStart."

Perhaps I do not need to be so verbose, maybe something like:

```
It then uses this information to expose the total
duration from `Performance.navigationStart` of the parent document, until the
interactive example has reached `Performance.loadEventEnd`.
```

Thoughts?

jwhitlock: I think the ``Performance.`` is implied, so you can just do:

```
It then uses this information to expose the total
duration from `navigationStart` of the parent document, until the
interactive example has reached `loadEventEnd`.
```
Do you intend to update the other metrics to use ``navigationStart`` rather than ``fetchStart``? I'll defer to you and @tkadlec on which marks to use for each measurement.

schalkneethling: Hmmm, good point. I reckon using `navigationStart` is the better option here, as it will fallback to `fetchStart` anyhow if there was no previous navigation context.

jwhitlock: According to this [2014 thread](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/Q3fYD9VZIdo), ``fetchStart - navigationStart`` measures the time it takes for a browser to initialize a new page. We will have difficulty optimizing that 😁

schalkneethling: Hmmmmmm, now I am falling back to thinking we should use `fetchStart` explicitly again. Let's see what @tkadlec thinks.

tkadlec:
Hmm....we probably should so we're consistent

schalkneethling:
Thanks for the review and feedback. This has been updated.

schalkneethling:
Ok, I believe this is ready. I have tested again to make sure everything is sent with values:

screen shot 2018-06-28 at 17 49 11

PR #4881, opened 2018-07-12
Inline dnt-helper

schalkneethling:
@jwhitlock This is with regards to the following `mdn-fiori` issue:
https://github.com/mdn/mdn-fiori/issues/31


schalkneethling:
@tkadlec Very good point, and I would have preferred that. With that said though:

1. This script has not, and probably will not, change in the near future so, I reckon it is safe to do it this way for now.
2. Our build system needs an overhaul that is going to take some time, so I would rather reap the benefit now, and improve it in future.

jwhitlock:
The only way we justify spending time on the build system is when it is a blocker for a change like this. What changes would you need to make it possible to inline the upstream code, @schalkneethling?

schalkneethling:
@jwhitlock I thought about this a bit, and here is what I am thinking:

1. I do not forsee having to update the `dnt-helper` within the next 6+ months so, perhaps going with this "good enough" approach is fine for now?
2. If there are strong feeling to the contrary, I can write a simple Nodejs util that we can call during build time to replace the string `{% javascript 'dnt-helper' %}` with the inlined script. Of course, that will have to run before Jinja/Django creates the various templates. How much of an issue that is, I am not sure. I guess that the CSS and JS bundling and minification happens before that as well, so maybe it is as simple as adding this util to that process?

Either way works for me.

jwhitlock:
A phased approach might be best. We can manually inline using something like this PR, to get the immediate benefit. For that, I'd prefer something like:

```
<script>{% include "kuma/static/js/libs/mozilla.dnthelper.min.js" %}</script>
```

Adding the minified JS alongside the original would allow people to audit the minimization, would make it easier to inject the text needed for the MPL license, and would make it easier to update when the DNT code is updated.

There are additional "inline this asset" recommendations that won't be as easy to implement, as the assets will change regularly.  This quick version will have to be followed by a maintainable asset pipeline, rather than hack after hack for each inlined asset.

However, this PR is probably not the place to specify the new toolchain.  Maybe we can discuss at the next performance meeting, and figure out the right forum.

schalkneethling:
@jwhitlock updated.

schalkneethling:
@jwhitlock I can totally add it but, once we automate minification it will genrally(unless you override it) remove it from the minified file. Is it enough to have it in the non-minified version? I am more than happy to add it manually to the minified source for now though. Let me know, thanks!

jwhitlock:
Yes please, let's do manual this time. YUI Compressor has a [preserve comment feature](https://www.electrictoolbox.com/preserve-comments-yui-compressor/) for this use case, we might be able to find a similar convention for our compressor:

```js
/*! some copyright information here */
```

schalkneethling:
@jwhitlock updated.

kuma/settings/common.py
```diff
@@ -589,7 +589,7 @@ def _get_locales():
     {
         'NAME': 'jinja2',
         'BACKEND': 'django_jinja.backend.Jinja2',
-        'DIRS': [path('jinja2')],
+        'DIRS': [path('jinja2'), path('static')],
```
jwhitlock: This could be ``path('kuma/static')``, and then ``make build-static`` wouldn't be required. We can tease this apart in the next asset pipeline.

jinja2/base.html
```diff
@@ -42,7 +42,9 @@
   {% stylesheet 'locale-%s' % LANG %}
 
   {% block site_analytics %}
-    {% javascript 'dnt-helper' %}
+    <script>
+      {% include "js/libs/mozilla.dnthelper.min.js" %}
```
jwhitlock: We could move this inside ``includes/google_analytics.html``, and then there would be one inline script block in the ``<head>``.

jwhitlock:
I'm going to open a bug for the work, and then squash these commits before merging.

jinja2/base.html
schalkneethling: Ah yes, that is a good idea. 👍

PR #4887, opened 2018-07-17
bug 1465085: Return 404 for the child of a 301 to a non-document URL

jwhitlock:
If [/en-US/docs/Main_Page](https://developer.mozilla.org/en-US/docs/Main_page) is a redirect to a non-Document MDN URL, then return a 404 to [/en-US/docs/MainPage/SubPage](https://developer.mozilla.org/en-US/docs/Main_page/SubPage) when logged in, rather than raise an Internal Server Error (500 ISE).

I didn't refactor the tests, which can be refactored after PR #4853 simplifies the logic.

### DIFF
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -840,6 +840,25 @@ class DocumentEditingTests(UserTestCase, WikiTestCase):
         parameters = parse_qs(response.request['QUERY_STRING'])
         assert parameters['parent'][0] == str(root_doc.id)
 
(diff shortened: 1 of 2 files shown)


jwhitlock:
@safwanrahman, this will also address [bug 1435301](https://bugzilla.mozilla.org/show_bug.cgi?id=1435301), but by turning them all into 404s, so I think there is still work to do there.

safwanrahman:
@jwhitlock Yes. I think its ok for now to have them `404`, and add a comment with `TODO:` to fix the bug. I am trying to figure out how to do that.

PR #4890, opened 2018-07-23
[Bug 1471829] Don't prompt remember language notification if already exist

safwanrahman:
I believe this fixes the [bug 1471829](https://bugzilla.mozilla.org/show_bug.cgi?id=1471829). @schalkneethling r?

### DIFF
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -59,6 +59,7 @@
     <link rel="search" type="application/opensearchdescription+xml" href="{{ settings.SITE_URL }}/{{ request.LANGUAGE_CODE }}/search/xml" title="{{ _('MDN Web Docs') }}">
   {% endif %}
 
+  <meta name="lang-cookie" content="{{ settings. LANGUAGE_COOKIE_NAME }}">
   <!-- third-generation iPad with high-resolution Retina display: -->
   <link rel="apple-touch-icon-precomposed" sizes="144x144" href="{{ static('img/favicon144.png') }}">
   <!-- iPhone with high-resolution Retina display: -->
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -59,7 +59,7 @@
     <link rel="search" type="application/opensearchdescription+xml" href="{{ settings.SITE_URL }}/{{ request.LANGUAGE_CODE }}/search/xml" title="{{ _('MDN Web Docs') }}">
   {% endif %}
 
-  <meta name="lang-cookie" content="{{ settings. LANGUAGE_COOKIE_NAME }}">
+  <meta name="lang-cookie" content="{{ settings.LANGUAGE_COOKIE_NAME }}">
   <!-- third-generation iPad with high-resolution Retina display: -->
   <link rel="apple-touch-icon-precomposed" sizes="144x144" href="{{ static('img/favicon144.png') }}">
   <!-- iPhone with high-resolution Retina display: -->
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -58,8 +58,6 @@
     <meta name="twitter:creator" content="@MozDevNet">
     <link rel="search" type="application/opensearchdescription+xml" href="{{ settings.SITE_URL }}/{{ request.LANGUAGE_CODE }}/search/xml" title="{{ _('MDN Web Docs') }}">
   {% endif %}
-
-  <meta name="lang-cookie" content="{{ settings.LANGUAGE_COOKIE_NAME }}">
   <!-- third-generation iPad with high-resolution Retina display: -->
   <link rel="apple-touch-icon-precomposed" sizes="144x144" href="{{ static('img/favicon144.png') }}">
   <!-- iPhone with high-resolution Retina display: -->
--- a/kuma/static/js/lang-switcher.js
+++ b/kuma/static/js/lang-switcher.js
@@ -33,6 +33,20 @@
         return localStorage.getItem(neverShowNoticeKey) || false;
     }
 
+    function trackGAEvent(action, locale) {
+        if (win.ga) {
+            var data = {
+                category: 'Remember Language',
+                action: action
+            };
+
+            if (locale) {
+                data.label = locale
+            }
+            mdn.analytics.trackEvent(data);
+        }
+    }
+
     if(win.sessionStorage && win.mdn.features.localStorage && !neverShowNotice) {
         var langSwitcherSelector = document.getElementById('language');
         var langSwitcherButton = document.getElementById('translations');
@@ -78,19 +92,23 @@
                     .success(function() {
                         notification.close();
                     });
+
                 // Track in GA
-                if (win.ga) {
-                    win.ga('set', 'locale-permanent-yes', locale);
-                }
+                // `locale` is actually locale code
+                trackGAEvent('yes', locale);
             });
 
             $('#locale-permanent-no').on('click', function() {
                 notification.close();
+                // locale is a object. Track the locale code only
+                trackGAEvent('no', locale.code);
             });
 
             $('#locale-permanent-never').on('click', function() {
                 storeNeverShowNotice();
                 notification.close();
+                // locale is a object. Track the locale code only
+                trackGAEvent('never', locale.code);
             });
         }
     }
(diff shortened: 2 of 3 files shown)


jinja2/base.html
```diff
@@ -59,6 +59,7 @@
     <link rel="search" type="application/opensearchdescription+xml" href="{{ settings.SITE_URL }}/{{ request.LANGUAGE_CODE }}/search/xml" title="{{ _('MDN Web Docs') }}">
   {% endif %}
 
+  <meta name="lang-cookie" content="{{ settings. LANGUAGE_COOKIE_NAME }}">
```
schalkneethling: Whoops, I believe there is an unintended space between `settings.` and `LANGUAGE_COOKIE_NAME`?

kuma/static/js/lang-switcher.js
```diff
@@ -3,7 +3,19 @@
...
     var neverShowNotice = getNeverShowNotice();
     function storeLocaleChange(code, name) {
-        sessionStorage.setItem(sessionStorageKey, JSON.stringify({code: code, name: name}));
+        if (!isExistingLocaleCookieExist(code)){
+            sessionStorage.setItem(sessionStorageKey, JSON.stringify({code: code, name: name}));
+        }
+    }
+
+    function getCookie(name) {
+	    var v = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
```
schalkneethling: Nit: I would give this a more "semantic" name. Perhaps `var match`?

kuma/static/js/lang-switcher.js
```diff
@@ -3,7 +3,19 @@
...
+	    return v ? v[2] : null;
+	}
+
+    function isExistingLocaleCookieExist(code) {
```
schalkneethling: Nit: Would this read better as `localeCookieExist`?

schalkneethling:
@safwanrahman @jwhitlock Pulled this down and tested locally. A portion of the problem described in the bug is resolved, but a portion still remains problematic.

### Testing steps

1. Set default language to JA in browser settings
2. Access ja/docs/MDN/Feedback
3. Click on the language icon and choose English
4. The English page loads and the dialog opens
5. Click on "Yes"
6. Open the JA version again [ja/docs/MDN/Feedback] ✖️(The user is not redirected to the en-US page, but instead the ja page still loads)
7. Click on the language icon and choose English
8. The dialog is not shown again ✅  

I checked, and after having clicked "Yes" the django language cookie is indeed set

screen shot 2018-07-24 at 13 48 32

schalkneethling:
One other thing I see is that nothing is set in session storage.

schalkneethling:
@safwanrahman If I use the language switcher to change back to `ja` after having switched to `en-US` and clicked "Yes", the dialog actually still shows up. It must therefore be an issue with `sessionStorage` not actually being set.

To rephrase that a bit. It is set, but it seems to get lost on the next page load.

schalkneethling:
@safwanrahman Did some debugging and testing. Looks to me like you need these additions:

```
function setDefaultLangChosen() {
    sessionStorage.setItem('default-lang-set', true);
}

function getDefaultLangChosen() {
    return sessionStorage.getItem('default-lang-set') || false;
}
...
var changedLocaleTo = getLocaleChange();
        if (changedLocaleTo && getDefaultLangChosen() === false) {
...
// Add event listener to the buttons
$('#locale-permanent-yes').on('click', function() {
    $.post('/i18n/setlang/', {language: this.dataset.locale})
        .success(function() {
            notification.close();
            setDefaultLangChosen();
        });
    });
...
```

schalkneethling:
So, if you set your lang to en-US for example and then access http://localhost:8000/docs/MDN/Feedback after a browser restart, it will redirect to `en-US`.

However, if you explicitly provide a lang as part of the path i.e. http://localhost:8000/ja/docs/MDN/Feedback

This will still load `ja` and not redirect to en-US. This might just be how Django works, or is configured. @jwhitlock Thoughts?

kuma/static/js/lang-switcher.js
jwhitlock: I agree, the name of the function doesn't seem to match the purpose. ``isLocalePreference`` seems better to me - ``true`` if ``code`` is the user's current locale preference override, ``false`` if ``code`` is not the user's locale preference or if no preference is set.

kuma/static/js/lang-switcher.js
```diff
@@ -3,7 +3,19 @@
...
+        var cookieName = document.querySelector('meta[name="lang-cookie"]').content;
```
jwhitlock: I wish we could inject this name directly into the JS, but we'll probably need to improve our asset pipeline first.

I'd prefer [jinja2/includes/config.html](https://github.com/mozilla/kuma/blob/master/jinja2/includes/config.html), where we already pass Django settings into JS land.

jwhitlock:
I don't think we need to remember that a user switched to their preferred language in session storage, and @safwanrahman's solution gets the desired result of avoiding the dialog. However, using the dialog, and switching to the already preferred language, are good candidates for Google Analytics events, so we can see how frequently this feature is used. I think we're currently just tracking locale switching.

schalkneethling:
@jwhitlock Thanks for the feedback. So, to be sure. 

1. A user switches from say `ja` to `en-US`
2. On the English page the user clicks yes in the dialog.
3. If they then at some point during the same session switch to `de` or even `ja`. Is it correct that we prompt them again with the dialog?

jwhitlock:
@schalkneethling, thanks for updating w/ my response from the stand up. Also, I noticed you are on the review chain for PR #4321, so you are well aware of that 7 month process...

safwanrahman:
@jwhitlock @schalkneethling Thanks a lot for the reivew. I am bit confused what need to be changed. Only the function/variable name changes that @schalkneethling proposed or change the functionality?

jwhitlock:
@safwanrahman I'd also like you to consider moving the cookie name from a ``<meta>`` tag to [jinja2/includes/config.html](https://github.com/mozilla/kuma/blob/master/jinja2/includes/config.html).

safwanrahman:
@schalkneethling updated. now lint is passig.

jwhitlock:
Thanks @safwanrahman, I want to test the functionality again locally, including the new GA tracking. If I find an issue with the tracking, do you want an r- so you can fix, or do you want me to add a commit that fixes and merge it?

safwanrahman:
@jwhitlock Yes. Feel free to let me know if anything is broken in GA so I can fix them. I have very less idea about GA in kuma.

kuma/static/js/lang-switcher.js
```diff
@@ -3,7 +3,18 @@
...
+        }
+    }
+
+    function getCookie(name) {
+        var match = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
+        return match ? match[2] : null;
+    }
+
+    function isLocalePreference(code) {
+        return getCookie(win.mdn.langCookieName) === code;
```
jwhitlock: Thanks, this name and function is sweet like cake 🍰

kuma/static/js/lang-switcher.js
```diff
@@ -62,10 +73,15 @@
...
             $('#locale-permanent-yes').on('click', function() {
-                $.post('/i18n/setlang/', {language: this.dataset.locale})
+                var locale = this.dataset.locale;
+                $.post('/i18n/setlang/', {language: locale})
                     .success(function() {
                         notification.close();
                     });
+                // Track in GA
+                if (win.ga) {
+                    win.ga('set', 'locale-permanent-yes', locale);
```
jwhitlock: Let's track this as an event, like [Dashboard Pagination](https://github.com/mozilla/kuma/blob/5bbf91c2b38722c534e0aafbad0232c9898bd17d/kuma/static/js/dashboard.js#L95-L100). Here's the code for the 'yes' path:

```
mdn.analytics.trackEvent({
    category: 'Remember Language',
    action: 'yes;,
    label: locale
});
```

We'll need a ``trackEvent`` for the "no" and "never" paths as well.

safwanrahman:
@jwhitlock updated. I think its good to merge now!

PR #4900, opened 2018-07-29
bug 1474435: "Publish & keep editing" button taller than the others

Mihir-Karbelkar:
The publish and keep editing button in the editing section looks taller than Publish button hence adding a min height to Publish button fixes this issue.

### DIFF
--- a/kuma/static/styles/base/elements/forms.scss
+++ b/kuma/static/styles/base/elements/forms.scss
@@ -88,6 +88,7 @@ input[type='button'] {
     }
 
     &.positive {
+        min-height: 32px;
         border-color: $green;
         color: $green-dark;
     }


kuma/static/styles/base/elements/forms.scss
```diff
@@ -88,6 +88,7 @@ input[type='button'] {
     }
 
     &.positive {
+        min-height: 33px;
```
schalkneethling: Nit: `33px` seems a little odd. I would suggest making it `32px`. Thanks @Mihir-Karbelkar

Mihir-Karbelkar: Shall I ammend or git commit?

PR #4901, opened 2018-07-30
Fix Bug 1307217, add link to section elements to improve ux

schalkneethling:
@hobinjk @jwhitlock r?

Example of what this looks like on a page:

screen shot 2018-07-30 at 15 11 26


kuma/static/js/components/local-anchor.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Find all level two and three headings, and append a link
+ * element that links directly to the heading
+ */
+function appendLocalAnchors() {
```
schalkneethling: I have no idea why this reads `unction` :) My local file definitely does not.

schalkneethling: Oh, and if I click on the view button, the resulting view does show `function`. Guess it's a Github issue.

jwhitlock: It looks like ``function`` to me. Update or uninstall those plugins!

Does ``appendLocalAnchors`` mean HTML element order? The link appears to prepend the headings for me.

schalkneethling:
> Visually, I was expecting something like the Kuma docs, where the anchor indicator appears when hovering over the title.

Agreed. It would be good to get some user feedback on it. My initial thought was to have that be the behaviour but:

1. The entire heading is not wrapped in the anchor so, I would have to do the hover in a hack like way by changing the pointer when somebody hovers over the heading. It would not have the "natural" behaviour as it is not actually a link.
2. As this is new, perhaps people will notice it more readily, and thus get the benefit quicker
3. This way it also works on mobile

We can of course change the behaviour if we get a lot of negative feedback.

kuma/static/js/components/local-anchor.js
schalkneethling: Good call, `prependLocalAnchors` makes more sense. I was initially appending, and did not think to change the name when I switched to prepending ;)

schalkneethling:
@jwhitlock update. Thanks @jwhitlock and @hobinjk for the reviews!

jezdez:
ZOMG!!1! Thank you for adding those! 😻

PR #4903, opened 2018-07-31
Update and optimise images in hero area

schalkneethling:
@jwhitlock @hobinjk r? I took the opportunity to inline the main background image as an SVG. This saves on a request, and is a smaller size in bytes overall. The dino head is down from ~33kb to ~11kb

## Production

screen shot 2018-07-31 at 14 47 48

## Local

screen shot 2018-07-31 at 14 47 39


kuma/static/styles/components/home/home-masthead.scss
```diff
@@ -5,12 +5,20 @@ homepage masthead (search, feature callouts)
 .home-masthead {
     @include clearfix();
     position: relative;
+    background-color: $accent-light;
```
jwhitlock: In the color meter, it appears the old background was ``#82CCEB`` text on ``#88D1F1``. The new background is on ``#7ECBED`` on ``#83D0F2``, at least with the color meter. Is this why the dino head appears purple, while the old one appeared grey?

schalkneethling:
@jwhitlock Thanks for the review. These are the images as received from the designer. I did notice the dino head does not match up exactly with the current one, but it seemed close enough. Let me tweak the colours and update the PR.

kuma/static/styles/components/home/home-masthead.scss
schalkneethling: No, I reckon the dino head just needs a bit of a colour adjustments.

schalkneethling:
Had a look at the masthead background. On production it is:

Background: `#92CBE8`
Foreground: `#97D0EE`

The new SVG

Background: `#8ECAEA`
Foreground: `#93CFEF`

I reckon that is fine, as it has a bit more contrast, but not to much. I am updating the dino head though.

schalkneethling:
@jwhitlock updated.

jwhitlock:
@schalkneethling that looks better to me, and doesn't look purple. All we need now is the bug number in the commit titles.

PR #4904, opened 2018-07-31
bug 1464093: Upgrade to ElasticSearch 5.6

jwhitlock:
* Update the development environment to ElasticSearch 5.6.10, using [elastic.co's docker image](https://www.docker.elastic.co). They are deprecating [the images on dockerhub](https://hub.docker.com/_/elasticsearch/).
* Update the client libraries to the latest compatible with ES 5.6.10. Unlike the 0.x to 2.x transition, the client libraries have to be in sync with the server version, and there are several breaking changes that require simultaneous code changes. This makes me less impressed with the ElasticSearch libraries.
* ES 5 splits ``String`` into ``Text`` (tokenization and full text search) and ``Keyword`` (optimized for exact matches). I took the opportunity to move case-insensitive keyword queries into Python.
* Fix [bug 1180144](https://bugzilla.mozilla.org/show_bug.cgi?id=1180144), where searching for ``csssyntax`` returned ``{{csssyntax}}`` and ``{{csssyntax2}}``. Matches for keywords are now exact, unless a wildcard ``*`` or ``?`` is added.

**Update**: PR #4905 and PR #4906 are merged, so this is ready for review.

### DIFF
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -124,11 +124,11 @@ class SearchQueryBackend(BaseFilterBackend):
         return queryset
 
 
-class AdvancedSearchQueryBackend(BaseFilterBackend):
+class KeywordQueryBackend(BaseFilterBackend):
     """
     A django-rest-framework filter backend that filters the given queryset
-    based on additional query parameters that correspond to advanced search
-    indexes.
+    based on additional query parameters that correspond to case-insensitive
+    keywords.
     """
     fields = (
         'kumascript_macros',
@@ -161,10 +161,10 @@ class AdvancedSearchQueryBackend(BaseFilterBackend):
         return queryset
 
 
-class DatabaseFilterBackend(BaseFilterBackend):
+class TagGroupFilterBackend(BaseFilterBackend):
     """
     A django-rest-framework filter backend that filters the given
-    queryset based on the filters stored in the database.
+    queryset based on named tag groups.
 
     If there are more than one tag attached to the filter it will
     use the filter's operator to determine which logical operation to
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -140,17 +140,21 @@ class AdvancedSearchQueryBackend(BaseFilterBackend):
         queries = []
 
         for field in self.fields:
-            search_param = view.query_params.get(field)
+            raw_search_param = view.query_params.get(field, '').lower()
+            wildcard = '*' in raw_search_param or '?' in raw_search_param
+            search_param = raw_search_param.replace('*', '').replace('?', '')
             if not search_param:
+                # Skip if not given, or just wildcard
                 continue
 
+            # Exact match of sanitized value
             queries.append(
-                Q('match', **{field: {'query': search_param.lower(),
-                                      'boost': 10.0}}))
-            queries.append(
-                Q('prefix', **{field: {'value': search_param.lower(),
-                                       'boost': 5.0}}))
-
+                Q('term', **{field: {'value': search_param, 'boost': 10.0}}))
+            if wildcard:
+                # Wildcard search of value as passed
+                queries.append(
+                    Q('wildcard', **{field: {'value': raw_search_param,
+                                             'boost': 5.0}}))
         if queries:
             queryset = queryset.query(query.Bool(should=queries))
 
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -96,10 +96,10 @@ class SearchQueryBackend(BaseFilterBackend):
     """
     search_operations = [
         # (<query type>, <field>, <boost factor>)
-        ('match', 'title', 6.0),
+        ('match', 'title', 7.2),
         ('match', 'summary', 2.0),
         ('match', 'content', 1.0),
-        ('match_phrase', 'title', 10.0),
+        ('match_phrase', 'title', 12.0),
         ('match_phrase', 'content', 8.0),
     ]
 
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -145,10 +145,10 @@ class AdvancedSearchQueryBackend(BaseFilterBackend):
                 continue
 
             queries.append(
-                Q('match', **{field: {'query': search_param,
+                Q('match', **{field: {'query': search_param.lower(),
                                       'boost': 10.0}}))
             queries.append(
-                Q('prefix', **{field: {'value': search_param,
+                Q('prefix', **{field: {'value': search_param.lower(),
                                        'boost': 5.0}}))
 
         if queries:
--- a/kuma/wiki/search.py
+++ b/kuma/wiki/search.py
@@ -31,25 +31,25 @@ class WikiDocumentType(document.DocType):
                      'Project_talk:', EXPERIMENT_TITLE_PREFIX]
 
     boost = field.Float(null_value=1.0)
-    content = field.String(analyzer='kuma_content',
-                           term_vector='with_positions_offsets')
-    css_classnames = field.String(analyzer='case_insensitive_keyword')
-    html_attributes = field.String(analyzer='case_insensitive_keyword')
+    content = field.Text(analyzer='kuma_content',
+                         term_vector='with_positions_offsets')
+    css_classnames = field.Keyword()
+    html_attributes = field.Keyword()
     id = field.Long()
-    kumascript_macros = field.String(analyzer='case_insensitive_keyword')
-    locale = field.String(index='not_analyzed')
+    kumascript_macros = field.Keyword()
+    locale = field.Keyword()
     modified = field.Date()
     parent = field.Nested(properties={
         'id': field.Long(),
-        'title': field.String(analyzer='kuma_title'),
-        'slug': field.String(index='not_analyzed'),
-        'locale': field.String(index='not_analyzed'),
+        'title': field.Text(analyzer='kuma_title'),
+        'slug': field.Keyword(),
+        'locale': field.Keyword(),
     })
-    slug = field.String(index='not_analyzed')
-    summary = field.String(analyzer='kuma_content',
-                           term_vector='with_positions_offsets')
-    tags = field.String(analyzer='case_sensitive')
-    title = field.String(analyzer='kuma_title', boost=1.2)
+    slug = field.Keyword()
+    summary = field.Text(analyzer='kuma_content',
+                         term_vector='with_positions_offsets')
+    tags = field.Keyword()
+    title = field.Text(analyzer='kuma_title', boost=1.2)
 
     class Meta(object):
         mapping = Mapping('wiki_document')
@@ -64,6 +64,11 @@ class WikiDocumentType(document.DocType):
         return cls._doc_type.name
 
     @classmethod
+    def case_insensitive_keywords(cls, keywords):
+        '''Create a unique list of lowercased keywords.'''
+        return sorted(set([keyword.lower() for keyword in keywords]))
+
+    @classmethod
     def from_django(cls, obj):
         doc = {
             'id': obj.id,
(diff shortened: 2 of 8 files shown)


jwhitlock:
My suggested sequence:

* [ ] Review, merge, and deploy PR #4905 (only code changes)
* [ ] Review, merge, and deploy PR #4906 (library updates)
* [ ] (Maybe) deploy PR #4870 (also has library updates, but requires updating infra config)
* [ ] Review, merge this PR (library updates, updates ES server image)
* [ ] Provision a ES 5.4.10 server, update infra config
* [ ] Deploy this PR, new infra config
* [ ] Create new ES 5 index

kuma/search/paginator.py
```diff
@@ -52,3 +58,17 @@ def page(self, number):
...
+        Returns the total number of results.
+
+        Paginator's count property will call .count() on the search object,
+        which returns results before the pre_filter. This will result in a
+        count that is too high. Instead, use 'total' from the results,
+        executing if needed.
+        """
+        if self._result_total is not None:
+            return self._result_total
+        return self.object_list.execute().hits.total
```
safwanrahman: `object_list.count()` would return same result

kuma/search/filters.py
```diff
@@ -140,17 +140,21 @@ def filter_queryset(self, request, queryset, view):
         queries = []
 
         for field in self.fields:
-            search_param = view.query_params.get(field)
+            raw_search_param = view.query_params.get(field, '').lower()
```
safwanrahman: Why do we lowering the params?

jwhitlock: Because they are lowercased when added to the index in ``kuma/wiki/search.py``.

safwanrahman: There are more than one analyzer. There are also case insensitive analyzer
https://github.com/mozilla/kuma/blob/master/kuma/wiki/search.py#L156-L163

kuma/wiki/search.py
```diff
@@ -63,6 +63,11 @@ def get_connection(cls, alias='default'):
     def get_doc_type(cls):
         return cls._doc_type.name
 
+    @classmethod
+    def case_insensitive_keywords(cls, keywords):
```
safwanrahman: Any reason to use lowercase keywords?

safwanrahman: You can use `simple` analyzer in the feild so Elasticsearch will use lowercase tokenizer.

kuma/wiki/search.py
```diff
@@ -153,15 +161,6 @@ def get_analysis(cls):
                         'snowball',
                     ],
                 },
-                'case_sensitive': {
```
safwanrahman: Any reson to remove them?

kuma/wiki/search.py
jwhitlock: An analyzer is useful for  [Text](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/text.html) datatype, but these are [Keywords](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/keyword.html) and analyzers are not allowed. Using lowercase means that the same logic is used both when populating ElasticSearch and when querying it.

kuma/wiki/search.py
jwhitlock: These we useful with ES 2.x, which had a generic ``String`` datatype for full text search and keyword (non-analyzed) search. They aren't required with the ``Keyword`` datatype, which more clearly states the purpose of these fields. ``case_sensitive`` is the default for ``Keyword`` search, and we can get the equivalent of ``case_insensitive_keyword`` by lowercasing keywords on indexing and querying.

kuma/wiki/search.py
safwanrahman: If you want to have case insensitive search, then you must you `Text` datatype with simple analyzer instead of `Keywords` datatype. Storing different values in ES and database is not a good idea. If you store in lowercase, its impossible to get actual value anymore from Elasticsearch.

kuma/wiki/search.py
safwanrahman: @jwhitlock storing lowercase data and lowercase tokenizer is not same.
If you use lowercase tokenizer, and you index `"This is test"` and search with `"test"`, you will still get `"This is test"`, not `"this is test"`. If you lowercase them while indexing and searching, you will always get `"this is text"`, which breaks the search result and others.

jwhitlock: I'm not convinced there is a problem yet. I did a lot of manual testing of the Kumascript macro dashboard, and this new mapping worked. I might be missing something, and will re-read your comments after sleeping on it. If you can think of a good example that would work on production but breaks in this branch, it would help me understand better and could be a good candidate for a test.

kuma/search/filters.py
jwhitlock: Only the ``case_insensitive_keyword`` analyzer was used on these fields (``kumascript_macros``, ``css_classnames``, ``html_attributes``), not the case-sensitive analyzer. I think this is a simplification, and the filter name should be renamed.

kuma/wiki/search.py
jwhitlock: The fields the ``case_insensitive_analyzer`` were used on are:

* ``css_classnames`` - extracted from the ``class`` elements in the content. This is a tool for front-end developers attempting to cross-reference CSS files to pages that use that CSS. I don't think our current front-end devs know this exists.
* ``html_attributes`` - extracted from the HTML elements in the content. This appears to be a tool for developers who are constructing the bleaching configuration. I haven't needed to do this kind of work in years.
* ``kumascript_macros`` - extracted from the content by looking for the ``{{macro}}`` pattern. KumaScript macros are case-insenstive at rendering time. This field gets used all the time, in the BCD migration project and macro retirement, and ambiguity in the queries has caused problems for these efforts.

It is simpler and more correct to use the ``Keyword`` datatype for these lists of keywords, where tokenization is already done in Python. The users is expecting a case-insensitive match when using these queries, so it makes sense to me to use Python lowercasing both when populating the index and querying, to make it easier to think about this parallel database.

We currently don't display what ElasticSearch stored for these queries, and I don't think we'll start, so there is no need to store exact values in ES for these keywords.

kuma/wiki/search.py
```diff
@@ -31,25 +31,25 @@ class WikiDocumentType(document.DocType):
...
-    slug = field.String(index='not_analyzed')
-    summary = field.String(analyzer='kuma_content',
-                           term_vector='with_positions_offsets')
-    tags = field.String(analyzer='case_sensitive')
-    title = field.String(analyzer='kuma_title', boost=1.2)
+    slug = field.Keyword()
+    summary = field.Text(analyzer='kuma_content',
+                         term_vector='with_positions_offsets')
+    tags = field.Keyword()
+    title = field.Text(analyzer='kuma_title', boost=1.2)
```
safwanrahman: Index time boosting is deprecated as per [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-boost.html)

kuma/wiki/search.py
```diff
@@ -31,25 +31,25 @@ class WikiDocumentType(document.DocType):
...
 
     class Meta(object):
         mapping = Mapping('wiki_document')
```
safwanrahman: We should also add the index name with `index` variable

kuma/wiki/search.py
jwhitlock: Good catch, I'll update the ``SearchQueryBackend``.

kuma/wiki/search.py
jwhitlock: The index is stored in the database, and is added at execution time.

kuma/search/filters.py
```diff
@@ -96,10 +96,10 @@ class SearchQueryBackend(BaseFilterBackend):
...
     search_operations = [
         # (<query type>, <field>, <boost factor>)
-        ('match', 'title', 6.0),
+        ('match', 'title', 7.2),
         ('match', 'summary', 2.0),
         ('match', 'content', 1.0),
-        ('match_phrase', 'title', 10.0),
+        ('match_phrase', 'title', 12.0),
         ('match_phrase', 'content', 8.0),
     ]
```
safwanrahman: In the query, you should explecitly mention `minimum_should_match=1`, otherwise, though the query does not match any, it will give other results.

jwhitlock: ``minimum_should_match=1`` is unneeded. From [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-bool-query.html) for ``should`` in a ``bool`` query:

> If the bool query is a filter context or **has neither must or filter** (*emphasis mine*) then at least one of the should queries must match a document for it to match the bool query.

kuma/wiki/search.py
```diff
@@ -31,25 +30,25 @@ class WikiDocumentType(document.DocType):
...
+                         term_vector='with_positions_offsets')
+    css_classnames = field.Keyword()
+    html_attributes = field.Keyword()
     id = field.Long()
-    kumascript_macros = field.String(analyzer='case_insensitive_keyword')
-    locale = field.String(index='not_analyzed')
+    kumascript_macros = field.Keyword()
+    locale = field.Keyword()
     modified = field.Date()
     parent = field.Nested(properties={
```
safwanrahman: I strongly believe this should be `Object` datatype instead of `Nested` datatype. `Nested` datatype is preferrable for arrays of objects.

kuma/wiki/search.py
```diff
@@ -63,36 +62,32 @@ def get_connection(cls, alias='default'):
...
+    def case_insensitive_keywords(cls, keywords):
+        '''Create a unique list of lowercased keywords.'''
+        return sorted(set([keyword.lower() for keyword in keywords]))
+
     @classmethod
     def from_django(cls, obj):
+        is_root_document = (obj.slug.count('/') == 1)
         doc = {
             'id': obj.id,
+            'boost': 4.0 if is_root_document else 1.0,
```
safwanrahman: Any specific reason to add this?

kuma/search/filters.py
safwanrahman: @jwhitlock I know about it but Its a filter backend. There can be multiple backend from a single view so other filter backend may add must and filter query.

kuma/wiki/search.py
jwhitlock: OK will try it

kuma/search/filters.py
jwhitlock: That's an interesting theory, but I don't believe it.

jwhitlock: I've read the source code for the ``SearchFilterBackend``, I don't believe it is possible to get a ``must`` or ``filter`` in the ``bool`` block.

safwanrahman: @jwhitlock If you combine 2 bool query, elasticsearch dsl makes a single bool query. You can check it in the view with `to_dict()` method. You will see all the `must`, `filter` and `should` is merged into a single bool query.

safwanrahman: I have tested with a sample query `"What"`, and elasticsearch dsl made following query. If you see, the filter, must and should all are included in same `bool` block
```javascript
{
  "query": {
    "boosting": {
      "positive": {
        "bool": {
          "filter": [
            {
              "terms": {
                "locale": [
                  "bn-BD",
                  "en-US"
                ]
              }
            }
          ],
          "must": [
            {
              "function_score": {
                "query": {
                  "bool": {
                    "should": [
                      {
                        "match": {
                          "title": {
                            "query": "what",
                            "boost": 6.0
                          }
                        }
                      },
                      {
                        "match": {
                          "summary": {
                            "query": "what",
                            "boost": 2.0
                          }
                        }
                      },
                      {
                        "match": {
                          "content": {
                            "query": "what",
                            "boost": 1.0
                          }
                        }
                      },
                      {
                        "match_phrase": {
                          "title": {
                            "query": "what",
                            "boost": 10.0
                          }
                        }
                      },
                      {
                        "match_phrase": {
                          "content": {
                            "query": "what",
                            "boost": 8.0
                          }
                        }
                      }
                    ]
                  }
                },
                "functions": [
                  {
                    "field_value_factor": {
                      "field": "boost"
                    }
                  }
                ]
              }
            }
          ]
        }
      },
      "negative": {
        "bool": {
          "must_not": [
            {
              "term": {
                "locale": "bn-BD"
              }
            }
          ]
        }
      },
      "negative_boost": 0.5
    }
  },
  "explain": true,
  "highlight": {
    "fields": {
      "content": {
        
      },
      "summary": {
        
      }
    },
    "post_tags": [
      "</mark>"
    ],
    "order": "score",
    "pre_tags": [
      "<mark>"
    ]
  }
}
```

jwhitlock: I see three ``bool`` blocks, the one we are talking about just has a ``should`` clause.

jwhitlock:
@safwanrahman sorry to cut your review short. We'd achieved the goal of getting our search working against ElasticSearch 5.6, and fixing Kuma's broken implementation was not a goal.  Feel free to file bugs for issues that you think still need to be addressed.  These will be easier to prioritize, review, and implement now that we're not under a deadline.

safwanrahman:
@jwhitlock I understand about the time constrains. It seems like the ES query needs to be optimized heavily. I hope to update the code part by part. Thanks

PR #4906, opened 2018-07-31
bug 1464093: Update to Elasticsearch 2 libraries

jwhitlock:
*Note: this builds on PR #4905, which I suggest deploying first*

Update to ES 2 libraries:

* elasticsearch 1.9.0 → 2.4.1: Last release in ES 2.x cycle
* elasticsearch-dsl 0.0.9 → 2.2.0: Sync interface with ES 2.x features, such as dropping F objects (filters are now merged with queries), merging "and", "or", etc. into 'bool' queries.
* python-dateutil 2.4.2 → 2.7.3: Zone updates and format change, expanded ambiguous date support, ISO 8601 parsing, drop Python 2.6 and 3.2, many more.

Make changes required by the 2.x libraries:
* Replace ``F()`` with ``Q()`` queries
* Replace ``and``, `or``, and ``filtered / filter`` queries with ``bool`` equivalents

Make changes that work with both 2.x and 5.x libraries:
* Update document counts in paginator for Django 1.10 / ES 5
* Switch from ``.optimize`` to ``.forcemerge``
* Replace more ``filtered`` queries
* Prepare for ``size=0`` (all results) no longer allowed in aggregations

### DIFF
--- a/kuma/search/filters.py
+++ b/kuma/search/filters.py
@@ -1,7 +1,7 @@
 import collections
 
 from django.conf import settings
-from elasticsearch_dsl import F, Q, query
+from elasticsearch_dsl import Q, query
 from rest_framework.filters import BaseFilterBackend
 
 from kuma.wiki.search import WikiDocumentType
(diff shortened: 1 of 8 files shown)


jwhitlock:
PR #4905 is merged, to be deployed. I still think it should be deployed before merging this one.

jwhitlock:
Yes, this is one of the cases where a review after the fact is inadequate. It would have been better to pair on this task, and had two people that knew how ElasticSearch works.

Here's what I see for those searches:

Search for CSS on the homepage, get 18 results. The URL does not include the full set of default topics:

http://localhost:8000/en-US/search?q=CSS&topic=api&topic=css&topic=html&topic=js&topic=apps&topic=webdev

Search for DOM on the search page, get 6 results:

http://localhost:8000/en-US/search?q=CSS&topic=api&topic=css&topic=html&topic=js&topic=apps&topic=webdev

Search for CSS, get 17 results:

http://localhost:8000/en-US/search?q=CSS&topic=css&topic=html&topic=js&topic=webdev

Add the "Open Web Apps" topic filter back, get 18 results:

http://localhost:8000/en-US/search?q=CSS&topic=css&topic=html&topic=js&topic=apps&topic=webdev

I think there's an existing bug in the topic filters. I also think the topic filters are a really bad feature. Still, time to write a new bug report...

PR #4907, opened 2018-08-01
Fix Bug 1478442, remove sprite fallback image for callouts

schalkneethling:
@jwhitlock I cannot reproduce the bug, but the image that is shown in the screenshot is from `promo-sprite.png`. This was set as the fallback image if no image is specified by the custom callout rule.

The way I recently changes this, we no longer need this fallback behaviour, and instead it will simply fallback to no image. So, should for some reason the reported scenario happen now, it will simply show no image, as apposed to the cut-off lightbulb image.

I tested all the variations for the current callouts in all conditional blocks, and all work as expected. r?


jwhitlock:
I may have found something.

Here's Chrome on my MacBook Pro native screen, and I get the right image:

pr 4907 macbook pro

If I drag the window over to my Dell U2415, which is not a high DPI display, I get the wrong image:

pr 4907 dell

@schalkneethling do you have a standard DPI display for testing?

schalkneethling:
Thanks for looking into this @jwhitlock 

It is definitely strange behaviour as there are no DPI based media queries effecting these images. Hmmm. Will dig in and see if I can replicate as well.

schalkneethling:
@jwhitlock Found the problem, fixing ;)

schalkneethling:
@jwhitlock updated.

kuma/static/styles/components/home/column-callout.scss
```diff
@@ -27,7 +27,6 @@ $callout-min-height: 170px;
             @include bidi(((right, $grid-spacing, left, auto),));
             height: $callout-icon-width;
             width: $callout-icon-width;
-            background-image: url($path-to-embedded-images + 'promo-sprite.png');
```
jwhitlock: Is this removal still needed? Should we remove promo-sprite.png as well?

schalkneethling: Yeah I reckon it makes sense to remove this now. Let me just double check on other uses of the sprite. If there is none, I will remove the image as well and update the PR.

schalkneethling:
@jwhitlock Image is not used anywhere else, so it is removed.

PR #4899, opened 2018-07-28
bug 1467518: Enable pytest to collect all test on the Python 3 image.

MatonAnthony:
Move a step closer to run the tests on Python 3.

This PR rewrite the minimal necessary code to allow `pytest` to properly collect all the tests before running them.

## Yet to be done:
<strike>- There is one error left that I don't manage to fix at this stage.</strike>

## Actions that needs to be taken before merging:
- Merge PR #4870 or cherry-pick those changes to a branch based on master.

## Notes:
- This PR is based on #4870 (which removes a blocker for the migration to Python 3).
- `make lint` returns errors due to code no longer compatible with Python 3.
- `django.utils.six` is used for the compatibility layer as suggested by @jwhitlock on IRC.

PR #4910, opened 2018-08-01
Bug 1467518: First chunk of unittest fixes for Python 3

MatonAnthony:
This PR fixes a first chunk of the unittest on the Python 3 version.

## Yet to be done:
- [x] Fix more tests
- [x] Ensure no python 2 tests are broken
- [ ] Clean the commits, some are dirty

## Actions that needs to be taken before merging:
- [x] Merge PR #4899 or cherry-pick those changes to a branch based on master.
- [x] Merge PR #4870 or cherry-pick those changes to a branch based on master.
    - Since this has been merged with squash, this branch needs to be rebased on #4899 once #4899 has been rebased on master

## Notes:
- This PR is based on #4870 (which removes a blocker for the migration to Python 3) & #4870 (which removes a blocker to run the unittests).
- `make lint` returns errors due to code no longer compatible with Python 3.
- This PR is open early in the process, so it is visible to anyone interested

### DIFF
--- a/kuma/core/tests/test_commands.py
+++ b/kuma/core/tests/test_commands.py
@@ -1,6 +1,6 @@
 import pytest
 from django.core.management import call_command, CommandError
-from django.utils.six import StringIO, PY2
+from django.utils.six import PY2, StringIO
 
 
 def test_help():
--- a/kuma/core/tests/test_commands.py
+++ b/kuma/core/tests/test_commands.py
@@ -6,7 +6,7 @@ from django.utils.six import StringIO
 def test_help():
     with pytest.raises(CommandError) as excinfo:
         call_command('ihavepower', stdout=StringIO())
-    assert str(excinfo.value) == 'Error: too few arguments'
+    assert 'Error: the following arguments are required: username' == str(excinfo.value)
 
 
 def test_user_doesnt_exist(db):
--- a/kuma/core/tests/test_commands.py
+++ b/kuma/core/tests/test_commands.py
@@ -1,17 +1,22 @@
 import pytest
 from django.core.management import call_command, CommandError
-from django.utils.six import StringIO
+from django.utils.six import StringIO, PY2
 
 
 def test_help():
     with pytest.raises(CommandError) as excinfo:
         call_command('ihavepower', stdout=StringIO())
-    assert 'Error: the following arguments are required: username' == str(excinfo.value)
+
+    if PY2:
+        assert str(excinfo.value) == 'Error: too few arguments'
+    else:
+        assert str(excinfo.value) == 'Error: the following arguments are required: username'
 
 
 def test_user_doesnt_exist(db):
     with pytest.raises(CommandError) as excinfo:
         call_command('ihavepower', 'fordprefect', stdout=StringIO())
+
     assert str(excinfo.value) == 'User fordprefect does not exist.'
 
 
--- a/kuma/feeder/utils.py
+++ b/kuma/feeder/utils.py
@@ -9,7 +9,7 @@ from time import mktime
 import feedparser
 import jsonpickle
 from django.conf import settings
-from django.utils.encoding import smart_str
+from django.utils.encoding import smart_bytes
 from django.utils.six.moves.urllib.error import URLError
 
 from .models import Entry, Feed
@@ -178,7 +178,7 @@ def save_entry(feed, entry):
     if len(entry.guid) <= max_guid_length:
         entry_guid = entry.guid
     else:
-        entry_guid = md5(smart_str(entry.guid).encode('utf-8')).hexdigest()
+        entry_guid = md5(smart_bytes(entry.guid)).hexdigest()
 
     last_published = datetime.fromtimestamp(mktime(entry.published_parsed))
 
--- a/kuma/feeder/utils.py
+++ b/kuma/feeder/utils.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import logging
 import socket
 from datetime import datetime
@@ -176,7 +178,7 @@ def save_entry(feed, entry):
     if len(entry.guid) <= max_guid_length:
         entry_guid = entry.guid
     else:
-        entry_guid = md5(smart_str(entry.guid)).hexdigest()
+        entry_guid = md5(smart_str(entry.guid).encode('utf-8')).hexdigest()
 
     last_published = datetime.fromtimestamp(mktime(entry.published_parsed))
 
--- a/kuma/landing/tests/test_views.py
+++ b/kuma/landing/tests/test_views.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import mock
 import pytest
 from django.core.cache import cache
@@ -68,9 +70,9 @@ def test_robots_not_allowed(client):
     assert_shared_cache_header(response)
     assert response['Content-Type'] == 'text/plain'
     content = response.content
-    assert 'Sitemap: ' not in content
-    assert 'Disallow: /\n' in content
-    assert 'Disallow: /admin/\n' not in content
+    assert b'Sitemap: ' not in content
+    assert b'Disallow: /\n' in content
+    assert b'Disallow: /admin/\n' not in content
 
 
 def test_robots_allowed_main_website(client, settings):
@@ -83,9 +85,9 @@ def test_robots_allowed_main_website(client, settings):
     assert_shared_cache_header(response)
     assert response['Content-Type'] == 'text/plain'
     content = response.content
-    assert 'Sitemap: ' in content
-    assert 'Disallow: /\n' not in content
-    assert 'Disallow: /admin/\n' in content
+    assert b'Sitemap: ' in content
+    assert b'Disallow: /\n' not in content
+    assert b'Disallow: /admin/\n' in content
 
 
 def test_robots_allowed_main_attachment_host(client, settings):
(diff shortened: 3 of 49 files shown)


kuma/core/tests/test_commands.py
```diff
@@ -6,7 +6,7 @@
 def test_help():
     with pytest.raises(CommandError) as excinfo:
         call_command('ihavepower', stdout=StringIO())
-    assert str(excinfo.value) == 'Error: too few arguments'
+    assert 'Error: the following arguments are required: username' == str(excinfo.value)
```
MatonAnthony: This change will cause issues in Python 2. I don't see any solutions that works for both environment.

MatonAnthony:
At this stage, this commit fix something like 800 failing tests :)

MatonAnthony:
We are down to 361 failing tests in Python 3. I'm gonna stop there for this PR, so I can fix everything that has probably already gone wrong in this PR so I don't spend time doing wrong things :)

MatonAnthony:
There is one `Python 2` failing test left, that I don't manage to fix without breaking a bunch of `Python 3` tests.

jwhitlock:
I'm not inclined to review this PR with a failing Python 2 test. Is it possible to create a new PR with just the "safe" changes?

jwhitlock:
Please take a look at the flake8 issues, and rebase to fix the conflict.

MatonAnthony:
Once this PR is ready to merge, we don't want to forget to rebase it to get a clean commit.

kuma/scrape/sources/base.py
```diff
@@ -144,9 +142,21 @@ def current_options(self):
 
     def decode_href(self, href):
         """Convert URL-escaped href attributes to unicode."""
-        decoded = unquote(binary_type(href))
-        assert isinstance(decoded, binary_type)
-        decoded = decoded.decode('utf8')
+        if PY3:  # pragma: no cover
+            # TODO: Remove the PY3 condition once we don't run Python 2 anymore
+            # In Python 3, unquote returns unicode
+            # decoded = unquote(href.decode('ascii'))
```
jwhitlock: It feels like this comment ``# decoded = unquote(href.decode('ascii'))`` is an earlier version of this code, and should be removed.

kuma/scrape/tests/test_scraper.py
```diff
@@ -247,7 +247,7 @@ def test_scrape(scraper):
     """The scraper will loop through sources until complete."""
     scraper.add_source('fake', 'loop', length=2, depth=2)
     sources = scraper.scrape()
-    assert sources.keys() == ['fake:loop', 'fake:loop2', 'fake:loop21']
+    assert set(sources.keys()) == {'fake:loop', 'fake:loop2', 'fake:loop21'}
```
jwhitlock: 👍 interesting way to work around iterator ``keys()`` an avoid assumed ordering

kuma/feeder/utils.py
```diff
@@ -176,7 +178,7 @@ def save_entry(feed, entry):
     if len(entry.guid) <= max_guid_length:
         entry_guid = entry.guid
     else:
-        entry_guid = md5(smart_str(entry.guid)).hexdigest()
+        entry_guid = md5(smart_str(entry.guid).encode('utf-8')).hexdigest()
```
jwhitlock: Would ``smart_bytes(entry.guid)`` work instead of ``smart_str(entry.guid).encode('utf-8')``?

kuma/search/store.py
```diff
@@ -63,6 +63,6 @@ def ref_from_url(url):
     locale = translation.get_language()
 
     for value in [query, page, locale] + filters:
-        md5.update(smart_str(value))
+        md5.update(smart_str(value).encode('utf-8'))
```
jwhitlock: Would ``smart_bytes`` work instead?

kuma/users/jobs.py
```diff
@@ -17,7 +17,7 @@ def fetch(self, email, secure=True, size=220, rating='pg',
         base_url = (secure and 'https://secure.gravatar.com' or
                     'http://www.gravatar.com')
         email_hash = hashlib.md5(email.lower().encode('utf8'))
-        params = urllib.urlencode({'s': size, 'd': default, 'r': rating})
+        params = urlencode({'s': size, 'd': default, 'r': rating})
```
jwhitlock: I like that the line is shorter too 😸

kuma/users/migrations/0001_squashed_0008_update_locales_tz_help.py
```diff
@@ -49,9 +49,9 @@ class Migration(migrations.Migration):
...
                 ('facebook_url', models.TextField(blank=True, verbose_name='Facebook', validators=[django.core.validators.RegexValidator(b'^https?://www\\.facebook\\.com/', 'Enter a valid Facebook URL.', b'invalid')])),
                 ('stackoverflow_url', models.TextField(blank=True, verbose_name='Stack Overflow', validators=[django.core.validators.RegexValidator(b'^https?://stackoverflow\\.com/users/', 'Enter a valid Stack Overflow URL.', b'invalid')])),
-                ('groups', models.ManyToManyField(related_query_name='user', related_name='user_set', to=b'auth.Group', blank=True, help_text='The groups this user belongs to. A user will get all permissions granted to each of his/her group.', verbose_name='groups')),
+                ('groups', models.ManyToManyField(related_query_name='user', related_name='user_set', to='auth.Group', blank=True, help_text='The groups this user belongs to. A user will get all permissions granted to each of his/her group.', verbose_name='groups')),
```
jwhitlock: I wasn't expecting the migration to change. I wonder if this should instead be a new migration. We probably have a few things that will change as bytes are turned into strings, such as choice fields.

kuma/wiki/content.py
```diff
@@ -493,7 +493,7 @@ def __iter__(self):
             if token['type'] == 'StartTag' and token['name'] == 'a':
                 attrs = dict(token['data'])
                 names = [name for (namespace, name) in attrs.keys()]
-                for (namespace, name), value in attrs.items():
+                for (namespace, name), value in attrs.copy().items():
```
jwhitlock: Is ``copy()`` needed here?

kuma/wiki/tests/conftest.py
```diff
@@ -312,7 +312,7 @@ def ks_toolbox():
     }
 
     d_json = json.dumps(errors)
-    d_b64 = base64.encodestring(d_json)
+    d_b64 = base64.encodestring(d_json.encode('utf-8'))
```
jwhitlock: In this case, ``encodestring`` is probably the right thing, since we split on newlines in the next line.

kuma/users/tests/test_templates.py
```diff
@@ -67,17 +67,17 @@ def test_account_email_page_requires_signin(db, client):
...
 
 def test_account_email_page_single_email(user_client):
     response = user_client.get(reverse('account_email'))
     assert response.status_code == 200
     assert_no_cache_header(response)
-    assert 'is your <em>primary</em> email address' in response.content
-    assert 'Make Primary' not in response.content
-    assert 'Re-send Confirmation' not in response.content
-    assert 'Remove' not in response.content
+    assert b'is your <em>primary</em> email address' in response.content
```
jwhitlock: Nit: When testing several strings, decode the content and test:

```python
content = response.content.decode(response.charset)
assert 'is your <em>primary</em> email address' in content
....
```

kuma/wiki/tests/test_views_document.py
```diff
@@ -100,7 +100,8 @@ def authkey(wiki_user):
     secret = key.generate_secret()
     key.save()
     auth = '%s:%s' % (key.key, secret)
-    header = 'Basic %s' % base64.encodestring(auth)
+    # TODO: Once Python 2/3 is gone, replace encodestring by encodebytes
+    header = 'Basic %s' % base64.encodestring(auth.encode('utf-8')).decode('utf-8')
```
jwhitlock: Same here, ``b64encode`` may be better for a header:

```python
header = 'Basic %s' % base64.b64encode(auth.encode('utf-8')).decode('utf-8') 
```

kuma/landing/tests/test_views.py
```diff
@@ -68,9 +70,9 @@ def test_robots_not_allowed(client):
     assert_shared_cache_header(response)
     assert response['Content-Type'] == 'text/plain'
     content = response.content
-    assert 'Sitemap: ' not in content
-    assert 'Disallow: /\n' in content
-    assert 'Disallow: /admin/\n' not in content
+    assert b'Sitemap: ' not in content
+    assert b'Disallow: /\n' in content
```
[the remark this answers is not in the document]
MatonAnthony: I don't think there is an escape here.
For me we are checking that `Disallow:` contains `/` and not `/admin`

kuma/wiki/content.py
MatonAnthony: Yes because the original `attrs` is changed during the iteration which is no longer allowed in Python 3.

kuma/wiki/tests/conftest.py
MatonAnthony: I don't understand if there is a change needed here?

MatonAnthony:
I pushed a new commit that should address this review.

Note: The commits still need to be rebased / squashed together before merging.

kuma/landing/tests/test_views.py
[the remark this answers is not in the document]
jwhitlock: You are right. Maybe I meant to request ``content = response.content.decode(response.charset)`` and got review blindness...

kuma/wiki/tests/conftest.py
jwhitlock: No, no change needed. I asked you to change encodestring somewhere else, and I didn't want you to change it here.

PR #4912, opened 2018-08-02
Fix Bug 1450922, improve homepage layout for hacks and contribute sections

schalkneethling:
@jwhitlock @hobinjk Updates to the homepage per bug 1450922.

## Before

screen shot 2018-08-02 at 13 07 02

## After

screen shot 2018-08-02 at 13 06 45

## Mobile Before

screen shot 2018-08-02 at 13 09 26

## Mobile After

screen shot 2018-08-02 at 13 09 03


jwhitlock:
I believe we paused this because of [bug 1482538](https://bugzilla.mozilla.org/show_bug.cgi?id=1482538), which also requires a homepage update. However, it seems like they are for different areas, and could be reviewed in parallel. A rebase is needed.

schalkneethling:
@jwhitlock Should I go ahead and look at removing the rest of the code related to the community stats, or will we leave that for another PR?

jwhitlock:
@schalkneethling yes, see if you can remove the community stats code.

schalkneethling:
@jwhitlock Travis seems to be stuck on orange. When you click through on details, it is green though.

kuma/static/styles/base/icons.scss
```diff
@@ -171,3 +171,17 @@ button[type='submit'] .icon-check-mark {
         margin-right: 5px;
     }
 }
+
+/* icons inside the hacks section on homepage */
```
jwhitlock: Nice cleanup moving these rules to icons.scss

PR #4915, opened 2018-08-02
bug 1458197: Merge DNT and GA inline scripts

jwhitlock:
Merge the minimized Mozilla DNT Helper and the Google Analytics setup into a single inline ``<head>`` script.

### DIFF
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -1,6 +1,8 @@
-{% if config.GOOGLE_ANALYTICS_ACCOUNT != '0' %}
-{% set GA_FILE = 'analytics_debug' if settings.DEBUG else 'analytics' %}
 <script>
+    // Mozilla DNT Helper
+    {% include "js/libs/mozilla.dnthelper.min.js" %}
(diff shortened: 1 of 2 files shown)


jinja2/includes/google_analytics.html
```diff
@@ -1,6 +1,8 @@
-{% if config.GOOGLE_ANALYTICS_ACCOUNT != '0' %}
-{% set GA_FILE = 'analytics_debug' if settings.DEBUG else 'analytics' %}
 <script>
+    // Mozilla DNT Helper
```
schalkneethling: This is not a standard HTML comment(i.e.  but I guess Jinja is ok with that?

jwhitlock: hmm I just copied the style of ``// only load GA if DNT is not enabled`` from lower. Do you want both to change to comment blocks like ``/* Mozilla DNT Helper */``?

schalkneethling: Oh doi ;p I was looking at the file extension and not thinking that this is a script block :) Ignore my comment

PR #4921, opened 2018-08-03
bug 1462475: Remove some more references to zones

jwhitlock:
Remove additional references to zoned documents from the code and documentation.


jwhitlock:
This is 2% of the 2,257 deleted lines from #4853. It was only possible to find them after removing zones from the scraper in #4920.

PR #4922, opened 2018-08-06
Remove reference to incorrect restart command

hobinjk:


docs/development.rst
```diff
@@ -319,8 +319,7 @@ is not compatible with production mode.
 To emulate production, and test compressed and hashed assets locally:
 
 #. Set the environment variable ``DEBUG=False``
-#. Start (``docker-compose up -d``) or restart (``docker-compose restart``)
-   your Docker services.
+#. Start (``docker-compose up -d``) your Docker services.
```
safwanrahman: It should be `docker-compose up --force-recreate`, because if existing container is running, it wont recreate them.

hobinjk: Thanks for the advice! I've updated accordingly

jwhitlock:
Please change the commit message to include a bug number, like:

``bug 957802: Remove reference to incorrect restart command``

hobinjk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1481486 filed and mentioned, does this look good now?

jwhitlock:
Yes, that works. [bug 957802](https://bugzilla.mozilla.org/show_bug.cgi?id=957802) is a bug we keep open for small documentation fixes, to keep us from having to write bugs that are longer than the fix 😁

hobinjk:
Oh, I assumed that was just a random bug number. I see! Definitely more convenient :)

PR #4924, opened 2018-08-07
bug 1462475: add db migration to cleanup after zones

escattone:
This PR should only be merged and deployed after #4923 has been merged and deployed. It is part of the cleanup after the death of zones #4853. It adds the DB migration necessary to remove the `wiki_documentzone` table and the `zone_subnav_local_html` field from the `wiki_document` table.


jwhitlock:
If you have an existing development environment, you may want to run this migration.

```
(host) $ make bash
 (web ) $ ./manage.py migrate
```

There is some leftover data after this change. A ``ContentType`` was created for ``DocumentZone``, along with some related permissions. To clean these up, use a new command added in [Django 1.11](https://docs.djangoproject.com/en/1.11/releases/1.11/#django-contrib-contenttypes)

```
(host) $ make bash
 (web ) $ ./manage.py remove_stale_contenttypes
```

This is worth the Django 1.11 for me, it used to be really hard to get rid of this data.

I'm expecting another migration in the PR for [bug 1455348](https://bugzilla.mozilla.org/show_bug.cgi?id=1455348), and then I'll start on a new sample database without zones.

PR #4926, opened 2018-08-07
bug 1423719: Remove suffix '| MDN' from page title

MatonAnthony:
We believe that adding '| MDN' at the end of each page title could be
harming our SEO.

To prove this theory we are removing it from Kuma as an experiment,
this commit should be reverted if this experience is unsuccesful.

### DIFF
--- a/jinja2/403.html
+++ b/jinja2/403.html
@@ -2,7 +2,7 @@
 
 {% block bodyclass %}error{% endblock %}
 
-{% block title %}{{ page_title(_('Permission Denied')) }}{% endblock %}
+{% block title %}{{ _('Permission Denied') }}{% endblock %}
 
 {% block site_css %}
     {{ super() }}
@@ -16,7 +16,7 @@
 
   <section id="content-main" class="full" role="main">
 
-    <h1 class="page-title">{% block page_title %}{{ _('Permission Denied') }}{% endblock %}</h1>
+    <h1 class="page-title">{{ _('Permission Denied') }}</h1>
     {% block reason -%}
 
     {% if reason == 'kumaediting' %}
--- a/kuma/core/templatetags/jinja_helpers.py
+++ b/kuma/core/templatetags/jinja_helpers.py
@@ -92,11 +92,6 @@ def entity_decode(str):
     return htmlparser.unescape(str)
 
 
-@library.global_function
-def page_title(title):
-    return jinja2.Markup(u'%s | MDN' % jinja2.escape(title))
-
-
 @library.filter
 def level_tag(message):
     return jinja2.Markup(force_text(LEVEL_TAGS.get(message.level, ''),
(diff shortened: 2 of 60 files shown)


jinja2/403.html
```diff
@@ -16,7 +16,7 @@
 
   <section id="content-main" class="full" role="main">
 
-    <h1 class="page-title">{% block page_title %}{{ _('Permission Denied') }}{% endblock %}</h1>
+    <h1 class="page-title">{{ _('Permission Denied') }}</h1>
```
jwhitlock: The ``{% block page_title %}`` should remain, so that it can be overridden in ``kuma/wiki/jinja2/403-create-page.html``:

https://github.com/mozilla/kuma/blob/94c0b7a4ffeea37f72c339e556db3d48e9ebda66/kuma/wiki/jinja2/403-create-page.html#L1-L3

kuma/core/templatetags/jinja_helpers.py
```diff
@@ -92,11 +92,6 @@ def entity_decode(str):
     return htmlparser.unescape(str)
 
 
-@library.global_function
-def page_title(title):
```
jwhitlock: Bold! 😮 

I would have changed this function to not add `` | MDN`` to titles for the experiment, but this way (removing it here and from all the templates) works as well. My best guess is that it won't come back after we remove it.

kuma/core/tests/test_helpers.py
```diff
@@ -204,12 +203,3 @@ def test_naive(self):
         dt = datetime(2016, 3, 10, hour, 8)
         out = in_utc(dt)
         assert out == datetime(2016, 3, 10, hour + 8, 8, tzinfo=pytz.utc)
-
-
-class TestPageTitle(TestCase):
```
jwhitlock: Two less tests to convert to pytest 😁

kuma/wiki/jinja2/403-create-page.html
```diff
@@ -1,6 +1,6 @@
 {% extends "403.html" %}
 
-{% block page_title %}{{ _("You're trying to create a new page.") }}{% endblock %}
+{{ _("You're trying to create a new page.") }}
```
jwhitlock: The ``{% block page_title %}`` should remain, so that the ``{% extends "403.html" %}`` will work properly.

tests/functional/test_article.py
```diff
@@ -7,16 +7,13 @@
...
 
 
 @pytest.mark.smoke
 @pytest.mark.nondestructive
 def test_title(base_url, selenium):
     page = ArticlePage(selenium, base_url).open()
-    assert (ARTICLE_NAME + ARTICLE_TITLE_SUFIX) == selenium.title
-    assert page.article_title_text == ARTICLE_NAME
-    assert page.article_title_text in selenium.title
+    assert page.article_title_text == selenium.title == ARTICLE_NAME
```
jwhitlock: 👍 good work updating functional tests as well

PR #4930, opened 2018-08-09
Bug 1455348: Adds discourse field in user edit and user title info pages

Mihir-Karbelkar:
I have not added the migrations files. The discourse icon is not visible in the user edit page while the code is the same as for Facebook, it might be a problem in the svg class..


Mihir-Karbelkar:
There's a linting error in "icons.scss" and fails with test_models.py giving `File "/app/kuma/users/tests/test_models.py", line 37, in test_websites
    assert url == getattr(user2, name)
AttributeError: 'User' object has no attribute 'dicourse_url'
` 
Not able to understand the rest error.

jwhitlock:
There's not much point in a deeper review until the robots are happy 🤖

jwhitlock:
I forced pushed to rebase on master:

* Updated ``hashin`` and ``pytz``
* Added a migration for the ``User`` changes for ``pytz`` (new timezones) and Django 1.10 (``username`` field [increased to 150 characters](https://docs.djangoproject.com/en/2.0/releases/1.10/#abstractuser-username-max-length-increased-to-150)
* Fixed the style linting for ``icons.scss`` (whitespace around commas and ``{``)
* Ran ``svgo`` on the new ``discourse.svg``, reducing size by half
* Added the migration for the ``discourse`` field. This required running ``docker-compose run web bash``, since the webserver crashed with the standard ``make up`` / ``make bash`` without the migration.
* Fix ``dicourse`` → ``discourse`` typo in tests.
* Update Discourse pattern to require ``https``, rather than allowing ``http://discourse.mozilla.org``.

@Mihir-Karbelkar, You'll need to reset to get the latest code, something like:

```
git fetch
git reset @{u}
```

Sorry for the force-push. I was wrong about this being a good first bug 😞

jwhitlock:
... and force-push again, because I needed to install ``pytz`` 2018.5 before creating the migration, and because the discourse migration didn't have the new validator pattern.

More work is needed on the edit view to convert from Font Awesome to SVG.

jwhitlock:
@schalkneethling There's a ``discourse.svg`` file, but the user edit page continues to use FontAwesome. The Discourse icon was added in FA 5, and we're on FA 4, so there is no Discourse icon. I think I missed the user edit page in my survey.  It gets included with the ``li_field`` macro (see ``show_icon``):

https://github.com/mozilla/kuma/blob/3453d1a436fa00306972a62a06049ef39ec27aaa/jinja2/includes/common_macros.html#L1-L13

I see a couple of easy solutions:
* Move all the icons to a common folder, so that the SVG path can be constructed from the name, like ``{% include "includes/icons/" + icon_name + ".svg" %}``
* Add a new parameter to ``li_field`` to take the path to an SVG icon

I'm not sure if the User edit page would be handled by your planned solution for in-content uses of FontAwesome.

schalkneethling:
> Move all the icons to a common folder, so that the SVG path can be constructed from the name, like {% include "includes/icons/" + icon_name + ".svg" %}

@jwhitlock This sounds like a sound solution to me. Would the current folder of `https://github.com/mozilla/kuma/tree/master/jinja2/includes/icons` suffice?

> I'm not sure if the User edit page would be handled by your planned solution for in-content uses of FontAwesome.

Yup, will add it to the list. Thanks!

jinja2/includes/icons/social/discourse.svg
```diff
@@ -0,0 +1 @@
+<svg class="icon icon-discourse" xmlns="http://www.w3.org/2000/svg" width="13pt" height="14pt" viewBox="0 0 13 14"><path d="M6.555.875C2.996.875 0 3.57 0 6.895l.004 6.23 6.55-.004c3.563 0 6.446-2.8 6.446-6.125C13 3.672 10.117.875 6.555.875zM6.5 10.5a3.845 3.845 0 0 1-1.578-.332l-2.356.55.668-2.05A3.33 3.33 0 0 1 2.785 7c0-1.934 1.664-3.5 3.715-3.5 2.05 0 3.715 1.566 3.715 3.5S8.55 10.5 6.5 10.5zm0 0"/></svg>
```
schalkneethling: I would suggest running this through SVGOmg [https://jakearchibald.github.io/svgomg/]. Also, I would recommend removing the `width` and `height` attributes or, replacing the `pt` unit with the equivalent `px` value. Roughly that would be:

```
...  width="17px" height="19px" 
```

One other nit is to ensure there is an empty line at the end of the file. Thanks @Mihir-Karbelkar

Mihir-Karbelkar:
Is it fine if I commit it tomorrow. Have to setup some of the things again. Thanks

jwhitlock:
> @jwhitlock This sounds like a sound solution to me. Would the current folder of https://github.com/mozilla/kuma/tree/master/jinja2/includes/icons suffice?

https://github.com/mozilla/kuma/tree/master/kuma/static/img/icons seems better to me, but that folder is already populated with PNGs...

Mihir-Karbelkar:
@jwhitlock About that "https" required in the discourse link, shouldn't with do with every link?

schalkneethling:
> @jwhitlock This sounds like a sound solution to me. Would the current folder of /jinja2/includes/icons@master suffice?

> /kuma/static/img/icons@master seems better to me, but that folder is already populated with PNGs...

@jwhitlock Do you mean we should move all of the SVG icons into that folder? i.e. everything that is currently in `/jinja2/includes/icons`?

jwhitlock:
> @jwhitlock Do you mean we should move all of the SVG icons into that folder? i.e. everything that is currently in /jinja2/includes/icons?

@schalkneethling that's one option, and what I would do if I was in a perfectionist mood. It may be easier to add yet another option to ``li_field``, to set the include path for the SVG icon, and then we can move them around later if needed.

> @jwhitlock About that "https" required in the discourse link, shouldn't we do this with every link?

If they were new links, yes. However, if we change the pattern now, people with existing ``http://`` links would get validation errors when they edit their page. We'll have to be more careful about making existing patterns more strict.

Mihir-Karbelkar:
@schalkneethling I have made the required changes if there are any other changes you want then please let me know. Thanks

jinja2/includes/icons/social/discourse.svg
```diff
@@ -0,0 +1,2 @@
+<svg class="icon icon-discourse" xmlns="http://www.w3.org/2000/svg" width="17px" height="19px" viewBox="0 0 11 11"><path d="M5.484.105C2.508.105 0 2.473 0 5.395v5.48l5.484-.004c2.98 0 5.395-2.46 5.395-5.383 0-2.922-2.418-5.383-5.395-5.383zm0 0"/><path d="M5.54 2.152c-1.18 0-2.27.61-2.872 1.602a3.229 3.229 0 0 0-.066 3.242l-.606 1.91 2.172-.484A3.39 3.39 0 0 0 7.855 7.8a3.24 3.24 0 0 0 .79-3.59 3.35 3.35 0 0 0-3.102-2.059zm0 0"/><path d="M8.191 3.438a3.23 3.23 0 0 1-.039 4.023 3.382 3.382 0 0 1-3.984.957l-2.172.488 2.211-.258c1.465.844 3.344.473 4.363-.859a3.236 3.236 0 0 0-.379-4.351zm0 0"/><path d="M7.578 2.832a3.233 3.233 0 0 1 .387 4.008c-.817 1.265-2.41 1.824-3.86 1.36l-2.109.706 2.172-.484c1.547.687 3.371.125 4.242-1.305a3.24 3.24 0 0 0-.832-4.285zm0 0"/><path d="M2.8 7.066a3.243 3.243 0 0 1 1.196-3.96 3.396 3.396 0 0 1 4.2.332c-1.063-1.368-3.012-1.7-4.481-.766-1.465.933-1.953 2.816-1.113 4.324l-.606 1.91zm0 0"/><path d="M2.602 6.996a3.232 3.232 0 0 1 .77-4.059 3.398 3.398 0 0 1 4.206-.105 3.39 3.39 0 0 0-4.527-.293 3.242 3.242 0 0 0-.656 4.406l-.399 1.961zm0 0"/></svg>
```
jwhitlock: This new SVG is a solid icon again - previously, there was a gap in the middle. Can we return to the version with the gap?

filled-in-discourse

schalkneethling: This one should do the trick:

```
<svg aria-hidden="true" data-prefix="fab" data-icon="discourse" class="svg-inline--fa fa-discourse fa-w-14" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path fill="currentColor" d="M225.9 32C103.3 32 0 130.5 0 252.1 0 256 .1 480 .1 480l225.8-.2c122.7 0 222.1-102.3 222.1-223.9C448 134.3 348.6 32 225.9 32zM224 384c-19.4 0-37.9-4.3-54.4-12.1L88.5 392l22.9-75c-9.8-18.1-15.4-38.9-15.4-61 0-70.7 57.3-128 128-128s128 57.3 128 128-57.3 128-128 128z"/></svg>
```

Mihir-Karbelkar: Shall I run this through Svgomg before committing? Thanks

schalkneethling: I already did that, you can use it as is. Thanks!

Mihir-Karbelkar:
Thanks @schalkneethling

Mihir-Karbelkar:
Thanks @jwhitlock and @schalkneethling!

PR #4931, opened 2018-08-14
bug 1483072: handle the new Jenkins pipelines

escattone:
This PR allows us to handle both the current and new Jenkins pipelines during this period of transition.

On the new Jenkins service owned by IT, it will use the new configuration scripts [within the `oregon` region](https://github.com/mdn/infra/tree/master/apps/mdn/mdn-aws/k8s/regions/oregon) rather than the current ones [within the `portland` region](https://github.com/mdn/infra/tree/master/apps/mdn/mdn-aws/k8s/regions/portland).

### DIFF
--- a/Jenkinsfiles/utils.groovy
+++ b/Jenkinsfiles/utils.groovy
@@ -5,8 +5,22 @@
 PROD_BRANCH_NAME = 'prod-push'
 STAGE_BRANCH_NAME = 'stage-push'
 STANDBY_BRANCH_NAME = 'standby-push'
-KUMA_PIPELINE = 'mdn_multibranch_pipeline'
-KUMASCRIPT_PIPELINE= 'kumascript_multibranch_pipeline'
+KUMA_PIPELINE = 'kuma'
+KUMASCRIPT_PIPELINE= 'kumascript'
+// TODO: After cutover to IT-owned services, remove these.
+MOZMEAO_KUMA_PIPELINE = 'mdn_multibranch_pipeline'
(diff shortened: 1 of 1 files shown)


Jenkinsfiles/utils.groovy
```diff
@@ -7,6 +7,18 @@ STAGE_BRANCH_NAME = 'stage-push'
 STANDBY_BRANCH_NAME = 'standby-push'
 KUMA_PIPELINE = 'mdn_multibranch_pipeline'
 KUMASCRIPT_PIPELINE= 'kumascript_multibranch_pipeline'
+FUTURE_KUMA_PIPELINE = 'kuma'
+FUTURE_KUMASCRIPT_PIPELINE= 'kumascript'
```
jwhitlock: Nits: When I'm migrating from one thing to another, I avoid words like "new" and "future" that look silly when they hang around for months and years. 

I make the future thing the standard, and the past the qualified one. So I would use:

```
MOZMEAO_KUMA_PIPELINE = 'mdn_multibranch_pipeline'
MOZMEAO_KUMASCRIPT_PIPELINE = 'kumascript_multibranch_pipeline'
KUMA_PIPELINE =  'kuma'
KUMASCRIPT_PIPELINE = 'kumascript'
```

Then, when we drop support for the MozMEAO infrastructure, we would just delete code paths rather than delete and modify.

escattone:
@jwhitlock I decided to amend the original commit and force-push since I thought the difference from the original commit would be more confusing to review again (if you wanted to).

PR #4932, opened 2018-08-14
bug 1483072: Get jenkins user ID via shell

jwhitlock:
The ``jenkins`` user's ID is different in the MozMEAO and IT Jenkins server, so don't hard-code it.


jwhitlock:
Ah, that's how to use a function-style syntax in Jenkins. Yes, the one-liner is nicer.

PR #4933, opened 2018-08-14
bug 1483072: accommodate DockerHub registry

escattone:
This PR accommodates the use of the DockerHub registry within the IT-owned Jenkins service.


Makefile
```diff
@@ -14,16 +14,15 @@ endif
 BASE_IMAGE_NAME ?= kuma_base
 KUMA_IMAGE_NAME ?= kuma
 KUMASCRIPT_IMAGE_NAME ?= kumascript
-REGISTRY ?= quay.io/
-IMAGE_PREFIX ?= mozmar
-BASE_IMAGE ?= ${REGISTRY}${IMAGE_PREFIX}/${BASE_IMAGE_NAME}\:${VERSION}
-BASE_IMAGE_PY3 ?= ${REGISTRY}${IMAGE_PREFIX}/${BASE_IMAGE_NAME}\:py3
-BASE_IMAGE_LATEST ?= ${REGISTRY}${IMAGE_PREFIX}/${BASE_IMAGE_NAME}\:latest
+IMAGE_PREFIX ?= quay.io/mozmar
```
jwhitlock: This is a nicer solution than what I was considering 👍

Makefile
```diff
@@ -14,16 +14,15 @@ endif
 BASE_IMAGE_NAME ?= kuma_base
 KUMA_IMAGE_NAME ?= kuma
 KUMASCRIPT_IMAGE_NAME ?= kumascript
-REGISTRY ?= quay.io/
```
safwanrahman: This will make it complex to make a image with `mdnwebdocs` in local build. I think it can be handeled something ike following and pass the `PIPELINE_NAME` env file from the `Jenkinsfile`
```make
ifeq ($(PIPELINE_NAME), mozilla_it)
REGISTRY ?= 
IMAGE_PREFIX ?= mdnwebdocs
else
REGISTRY ?= quay.io/
IMAGE_PREFIX ?= mozmar
endif
```

escattone:
> We may need to do something ugly for the kuma image, like a second Dockerfile

Good point. We can think about what we want to do there.

Makefile
escattone: @safwanrahman I'm not sure I understand your concern yet. For local builds, one could easily do:
```
make build-base IMAGE_PREFIX=mdnwebdocs
```
but for now, we really don't want everyone to use `mdnwebdocs`. Once we move completely to the IT-owned infrastructure, we can make `mdnwebdocs` the default value for `IMAGE_PREFIX`.

PR #4934, opened 2018-08-15
bug 1380388: Improve document purging

jwhitlock:
This fixes a few bugs and issues with document purging:

* On the deleted document page, show who deleted a document if the viewer has revert permissions.
* On the deleted document page, add a "Purge document" button if the viewer is allowed to purge the document.
* Add a purge confirmation page, so GET will work for ``$purge`` pages (fixes [bug 1197390](https://bugzilla.mozilla.org/show_bug.cgi?id=1197390)).
* If a document has been purged, treat it as a missing document (fixes [bug 1380388](https://bugzilla.mozilla.org/show_bug.cgi?id=1380388)).
* Add an async task to remove deletion logs for purged documents.

### DIFF
--- a/kuma/conftest.py
+++ b/kuma/conftest.py
@@ -78,6 +78,7 @@ def wiki_user_3(db, django_user_model):
 
 @pytest.fixture
 def user_client(client, wiki_user):
+    """A test client with wiki_user logged in."""
     wiki_user.set_password('password')
     wiki_user.save()
     client.login(username=wiki_user.username, password='password')
@@ -86,6 +87,7 @@ def user_client(client, wiki_user):
 
 @pytest.fixture
 def editor_client(user_client):
+    """A test client with wiki_user logged in for editing."""
     with override_flag('kumaediting', True):
         yield user_client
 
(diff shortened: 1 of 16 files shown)


jwhitlock:
Force-push to rebase on master

safwanrahman:
I think this view and template logic should have tests.

escattone:
There's just a simple linting issue:
```
kuma/wiki/tests/test_views_document.py:26:1: F401 '..models.DocumentDeletionLog' imported but unused
```

jwhitlock:
When I started adding tests, I realized there are some branches that can't  be reached.

The template ``deletion_log.html`` is only used when:

* There is a matching ``DocumentDeletionLog``
* There is a soft-deleted document (condition added by this PR)
* The requesting user has the ``wiki.restore_document`` permission:

https://github.com/mozilla/kuma/blob/cc0407c23532baeb2e16b70fd12cdc565155f9cf/kuma/wiki/views/document.py#L569-L575

The template also tests for the ``wiki.restore_document`` permission, which means the alternate path is never taken. If this was Python code, I'd ``assert user.has_perm('wiki.restore_document')``. I'm testing if a similar thing is possible in a Jinja2 template. More changes on the way (including fixing flake8 😅 )

jwhitlock:
Undoing the removal of ``page_title``, now that PR #4926 is closed, and rebasing for merging...

PR #4938, opened 2018-08-17
Fix Bug 1482538, replace survey banner with fx-dev banner

schalkneethling:
@jwhitlock @hobinjk This is ready for review. I am waiting for final sign off but, I do not foresee any major changes forthcoming.

@jwhitlock I was wondering, should I go ahead and remove the image and CSS for the survey banner? Perhaps we will switch back to it after this one has ran for a while though. Let me know, thanks!

https://bugzilla.mozilla.org/show_bug.cgi?id=1482538


schalkneethling:
Approval has been received https://bugzilla.mozilla.org/show_bug.cgi?id=1482538#c24

jwhitlock:
> @jwhitlock I was wondering, should I go ahead and remove the image and CSS for the survey banner? Perhaps we will switch back to it after this one has ran for a while though. Let me know, thanks!

I'm not sure what the agreement was around this new banner, so I'm not sure if we should get rid of the survey banner assets yet. @atopal may know the answer. Usually, we keep the old assets, markup, and translated strings for a while, in case we need to switch back.

schalkneethling:
> I'm not a fan of the design

Yeah, it is more on brand for Fx than MDN but, I guess that is the aim.

> bringing in PNGs

For the logo inside the button, we do not have a choice as the SVG version would be larger in bytes. The background image is going to be replaced with a SVG. My trial license of Sketch ran out though, so just waiting on my Hub request for a license.

The current two PNG files are optimised though so, it should not affect performance much.

schalkneethling:
@jwhitlock Received my license key. Background updated to be an SVG.

PR #4940, opened 2018-08-17
bug 1308630: Replace KUMA_WIKI_IFRAME_ALLOWED_HOSTS with ALLOWED_IFRAME_PATTERNS

jwhitlock:
The constance setting ``KUMA_WIKI_IFRAME_ALLOWED_HOSTS`` was a regex used to validate code sample request hosts and what ``src`` URLs were allowed for ``<iframe src="">``. It was also easy to get wrong, and needed to change in predictable ways for each deployment environment.

Code sample domain checking is updated to use ``ATTACHMENT_HOST`` and ``ATTACHMENT_ORIGIN`` for domain validation. This may require setting these correctly in stage and production.

A new setting ``ALLOWED_IFRAME_PATTERNS`` allows for more precise protocol and domain matching, optional path matching (by prefix or regex match), and can be automatically adjusted for ``ATTACHMENT_HOST`` and ``INTERACTIVE_EXAMPLE_BASE`` settings.

There are patterns that are no longer allowed by default:

* https://stage-files.mdn.moz.works/ - Stage's ``ATTACHMENT_HOST``. On Stage, this will be added automatically.
* http://localhost:8000/ - Dev's ``ATTACHMENT_HOST``. In the dev environment, this will be added automatically.
* http://localhost:8080/ - Dev's (old) ``INTERACTIVE_EXAMPLES_BASE``. The new port for [npm run start-server](https://github.com/mdn/interactive-examples/blob/bd8a91b5dd6c85c51adce007218838d3daf1d465/package.json#L44-L45) is ``9090``.
* http://testserver/ - The Django test client hostname. Tests now add this as needed.
* https://youtube.com/ - [EmbedYouTube](https://github.com/mdn/kumascript/blob/0006405761fc7ff977c7c6bdc467e90014b8a63b/macros/EmbedYouTube.ejs#L28) always uses www.youtube.com
* http:// variants - Production iframes use https://, and SSL should be the default in 2018.

``KUMA_WIKI_IFRAME_ALLOWED_HOSTS`` is retained for now, for debugging issues as this deploys.

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -6,7 +6,8 @@ from xml.sax.saxutils import quoteattr
 import html5lib
 import newrelic.agent
 from django.conf import settings
-from django.utils.six.moves.urllib.parse import unquote, urlencode, urlparse
+from django.utils.six.moves.urllib.parse import (unquote, urlencode, urlparse,
+                                                 urlsplit)
 from django.utils.translation import ugettext
 from html5lib.filters.base import Filter as html5lib_Filter
 from lxml import etree
@@ -335,8 +336,8 @@ class ContentSectionTool(object):
         return self
 
     @newrelic.agent.function_trace()
-    def filterIframeHosts(self, hosts):
-        self.stream = IframeHostFilter(self.stream, hosts)
+    def filterIframeHosts(self, patterns):
+        self.stream = IframeHostFilter(self.stream, patterns)
         return self
 
     @newrelic.agent.function_trace()
@@ -1074,12 +1075,12 @@ class IframeHostFilter(html5lib_Filter):
     """
     Filter which scans through <iframe> tags and strips the src attribute if
     it doesn't contain a URL whose host matches a given list of allowed
-    hosts. Also strips any markup found within <iframe></iframe>.
+    host patterns. Also strips any markup found within <iframe></iframe>.
     """
 
-    def __init__(self, source, hosts):
+    def __init__(self, source, patterns):
         html5lib_Filter.__init__(self, source)
-        self.hosts = hosts
+        self.allowed_src_patterns = patterns
 
     def __iter__(self):
         in_iframe = False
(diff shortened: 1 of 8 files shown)


kuma/wiki/content.py
```diff
@@ -1089,11 +1090,26 @@ def __iter__(self):
...
             if not in_iframe:
                 yield token
+
+    def validate_src(self, src):
+        """Validate an iframe src against the allowed patterns."""
+        parts = urlsplit(src)
+        for scheme, netloc, path in self.allowed_src_patterns:
+            if parts.netloc != netloc or parts.scheme != scheme:
+                continue
+            if isinstance(path, re._pattern_type):
```
escattone: As an alternative to using a non-public attribute of `re`, what do you think of something like this instead?
```python
if hasattr(path, 'match'):
```
or maybe something more strict, like this:
```python
if callable(getattr(path, 'match', None)):
```
or maybe this is more readable:
```python
if hasattr(path, 'match') and callable(path.match):
```

If you don't like any of those, that's fine. We can keep it like this.

jwhitlock: ``if hasattr(path, 'match'):`` looks good to me. I've also seen this, which seems ugly:

```
if isinstance(path, type(re.compile(''))):
```

jwhitlock:
It looks like we're setting ``ATTACHMENT_HOST`` and ``ATTACHMENT_ORIGIN``, so we should be ok:

https://github.com/mdn/infra/blob/c8f6a2567a154ca5d3092af8a2e7d367f4fd08ac/apps/mdn/mdn-aws/k8s/regions/portland/prod.sh#L106-L107

https://github.com/mdn/infra/blob/c8f6a2567a154ca5d3092af8a2e7d367f4fd08ac/apps/mdn/mdn-aws/k8s/regions/portland/stage.sh#L106-L107

Still, it would be good to manually check that attachments / live samples continue to work on deploy...

escattone:
> It looks like we're setting ATTACHMENT_HOST and ATTACHMENT_ORIGIN, so we should be ok:

Yes. I actually had been considering a change to `KUMA_WIKI_IFRAME_ALLOWED_HOSTS` prior to this PR that would allow us to include the ATTACHMENT_HOST and ATTACHMENT_ORIGIN via an environment variable to eliminate the need to do that manually, but you addressed it perfectly with this PR.

PR #4946, opened 2018-08-22
Fix Bug 1485302, use rel=noopener when using target=_blank

schalkneethling:
@jwhitlock small, but important change.


schalkneethling:
@jwhitlock I tested these two links locally and on production and the end result of following the links works the same. It seems like the only thing that is passed along is the page's URL where the link was clicked.

The component is also correctly shown as MDN Web Docs. Let me know if you see something else. Thanks!

PR #4947, opened 2018-08-23
fix(styles.wiki): Hide sidebar contents in preview mode

ExE-Boss:
This PR makes it so that the result of `{{...Sidebar}}` macros doesn’t appear in preview mode, which is implemented in some sidebar macros, but not others.

## TODO:

- [ ] Either find or open a bug for this PR.

### DIFF
--- a/kuma/static/styles/components/wiki/quick-links.scss
+++ b/kuma/static/styles/components/wiki/quick-links.scss
@@ -100,3 +100,8 @@
         }
     }
 }
+
+#wikiArticle .Quick_links,
+#wikiArticle #Quick_Links {
+    display: none;
+}


schalkneethling:
I agree that we need a bug to discuss this further. The CSS change looks fine, unfortunate that we need to use id selectors, but I guess this is a specificity issue.

ExE-Boss:
> The CSS change looks fine, unfortunate that we need to use id selectors, but I guess this is a specificity issue.

It’s not a specificity issue, `#wikiArticle` is necessary to avoid hiding the sidebar when outside of article preview.

schalkneethling:
> It’s not a specificity issue, #wikiArticle is necessary to avoid hiding the sidebar when outside of article preview.

Yup, I notice that the element does not have a `class` attribute on it. :(

ExE-Boss:
> @ExE-Boss are you still interested in this PR?

I’ve been waiting on @a2sheppy (and been more focused on #4979)

a2sheppy:
I wish we didn't have to do this but it makes sense given the problems we have in preview node and resource availability. I'm fine with it for now at least.

kuma/static/styles/components/wiki/quick-links.scss
```diff
@@ -100,3 +100,8 @@
         }
     }
 }
+
+#wikiArticle .Quick_links,
+#wikiArticle #Quick_Links {
+    display: none;
+}
```
schalkneethling: I tested this one the following pages:

* http://localhost:8000/en-US/docs/Web/HTML/Element
* http://localhost:8000/en-US/docs/Web/JavaScript/Reference
* http://localhost:8000/en-US/docs/Mozilla/Add-ons/WebExtensions
* http://localhost:8000/en-US/docs/Web/API/Canvas_API/Tutorial
* http://localhost:8000/en-US/docs/Games

This hides the sidebars as intended. We can, and should, however simplify the above CSS to only:

```
.Quick_links {
    display: none;
}
```

Thanks @ExE-Boss

schalkneethling: Outside article preview the class for quick links is `quick-links`

ExE-Boss: [The Kuma documentation says that it&nbsp;looks for an element with the&nbsp;id&nbsp;of&nbsp;`Quick_Links`.](https://developer.mozilla.org/docs/MDN/Contribute/Structures/Quicklinks#Quicklinks_syntax)

schalkneethling: If you look for example at the HTML/Element page that uses the HTMLSidebar:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element

You can see that the `id` and `class` is `quick-links`

screenshot 2018-10-03 at 20 37 19

Now, if you go to edit and then preview, you will see that the `id` and `class` has changed to `Quick-Links` which is how it is defined in the macro

screenshot 2018-10-03 at 20 36 14

You will also notice that the JS in `wiki.js` references these using the `#quick-links` id:

https://github.com/mozilla/kuma/blob/master/kuma/static/js/wiki.js#L14

I believe this document macro plays a role in the change:
https://github.com/mozilla/kuma/blob/master/kuma/wiki/jinja2/wiki/includes/document_macros.html#L21

Also please feel free to test it locally with my suggestions or, using Chrome's local overrides.

schalkneethling:
@ExE-Boss any updates since my last comment? Thanks!

ExE-Boss:
I’ve been currently focusing on https://github.com/mozilla/kuma/pull/4979.

schalkneethling:
@ExE-Boss are you still planning to update this one? Thanks!

ExE-Boss:
As I said above, I would personally prefer if the sidebar could be properly supported in preview mode by converting the first `#Quick_Links` element into an actual `#quick-links` sidebar, like in a normal page.

schalkneethling:
@jwhitlock @a2sheppy Your thoughts on @ExE-Boss comment above? Should/Could we do the above, or should we just hide it for the time being? Thanks!

a2sheppy:
The best solution from a contributor standpoint, and by extension for users, is to fix it so sidebars always work in preview mode. Or are at least displayed, even if they're not functional.

jwhitlock:
> As I said above, I would personally prefer if the sidebar could be properly supported in preview mode by converting the first #Quick_Links element into an actual #quick-links sidebar, like in a normal page, but I don’t understand Kuma’s inner workings well enough to be able to easily do that.

Here's the code that generates the preview:

https://github.com/mozilla/kuma/blob/6d15902235c4e2bd95fd8d3a29b2084d297ff346/kuma/wiki/views/revision.py#L47-L79

It ``POST``s the data to KumaScript. The difference between the ``POST`` and the ``GET`` APIs are why some data is missing in the macros during preview

When you render a page, it calls ``regenerate_cache_with_fields`` to extract data and store it in the database:

https://github.com/mozilla/kuma/blob/ffc038aee4e1517befaf70a2dc5dc22d7f646c18/kuma/wiki/models.py#L398-L406

You can't store rendered data during a preview, but you could call the ``get_section_text`` function that extracts the sidebar:

https://github.com/mozilla/kuma/blob/ffc038aee4e1517befaf70a2dc5dc22d7f646c18/kuma/wiki/models.py#L338

You'd also need to remove it from the body HTML:

https://github.com/mozilla/kuma/blob/ffc038aee4e1517befaf70a2dc5dc22d7f646c18/kuma/wiki/models.py#L326-L334

A lot of these use the ``Document`` model, and you can't use those same methods. You would need to use the tools in ``kuma/wiki/content.py`` more directly, so that you can use them on the HTML returned by KumaScript.

All the caching on the ``Document`` model was added for performance reasons. I'd move this cached data to a different table, or avoid storing it in the database at all and rely on a page-level caching layer. Maybe in 2019.

ExE-Boss:
Thank you, I’ll look into that.

jwhitlock:
Let's revisit after a bug is filed and prioritized.

PR #4950, opened 2018-08-24
bug 1218563: Add SSL development configuration

jwhitlock:
Add an optional docker-compose override for SSL:

- Generate a self-signed cert for ``developer-local.allizom.org`` and ``demos.developer-local.allizom.org`` on startup
- Add an nginx service that uses the cert
- Configure other services to expect SSL

This is yet another attempt at merging an SSL configuration. The last was #4430.  This avoids ``extends`` in docker-compose, which is removed in v3, and changes the SSL hostname to ``developer-local.allizom.org``.


safwanrahman:
Cant we use `localhost` instead of `developer-local.allizom.org`?

jwhitlock:
I need two domains for testing the live samples (the main website and the ``ATTACHMENT_HOST``, to trigger Chrome's iframe [feature policy protections](https://goo.gl/EuHzyv). This requires changing the ``/etc/hosts`` file anyway, so I leaned in with two testing domains.

@jgmize I think you had strong feelings about domains for testing, but I can't find those in a PR. What are your preferences for testing domains?

safwanrahman:
@jwhitlock In that case, I think you can use https://127.0.0.1.nip.io ;)

jwhitlock:
Thanks for the feedback @safwanrahman and @jgmize. I've switched to https://developer.127.0.0.1.nip.io and https://demos.developer.127.0.0.0.nip.io.

I'd like a squash when this is merged, this should all be one commit.

escattone:
I'd never heard of `nip.io` until now, interesting service! Thanks @safwanrahman.

docs/development.rst
```diff
@@ -468,3 +468,41 @@ Now you should be ready for a successful test run::
...
+
+   * In GitHub, generate a new GitHub OAuth app for the test SSL domain,
+     modifying the procees at :ref:`enable-github-auth`. When creating the
+     GitHub OAuth app, replace ``http://localhost:8000`` with
+     ``https://developer.127.0.0.1.nip.io`` in both URLs. When creating the
+     ``SocialApp`` in Kuma, chose the ``developer.127.0.0.1.nip.io`` site.
+
+#. Include the SSL containers by updating ``.env``::
+
+    COMPOSE_FILE=docker_compose.yml:docker-compose.ssl.yml
```
escattone: Typo. Should be `docker-compose.yml` rather than `docker_compose.yml`.

jwhitlock:
Force push to rebase on master, fix that document, and squash the commits so I'll look good. 😉

docker-compose.ssl.yml
```diff
@@ -0,0 +1,37 @@
...
+  web:
+    <<: *worker
+    command: /bin/bash -c "/etc/nginx/ssl/generate_dev_cert.sh && gunicorn -w 4 --bind 0.0.0.0:8000 --access-logfile=- --timeout=120 --worker-class=meinheld.gmeinheld.MeinheldWorker kuma.wsgi:application"
+    volumes:
+      - ./:/app:z
+      - ./etc/nginx/ssl:/etc/nginx/ssl
+  nginx:
+    image: nginx
+    volumes:
+      - ./etc/nginx/nginx.conf:/etc/nginx/nginx.conf:ro
```
safwanrahman: I suspect that this may fail in linux because of file permission. @jwhitlock do you have any linux machine to check with?

docker-compose.ssl.yml
```diff
@@ -0,0 +1,37 @@
...
+      - PROTOCOL=https://
+      - SESSION_COOKIE_SECURE=True
+      - SITE_ID=2
+      - SITE_URL=https://developer.127.0.0.1.nip.io
+      - STATIC_URL=https://developer.127.0.0.1.nip.io/static/
+  api:
+    <<: *worker
+  web:
+    <<: *worker
+    command: /bin/bash -c "/etc/nginx/ssl/generate_dev_cert.sh && gunicorn -w 4 --bind 0.0.0.0:8000 --access-logfile=- --timeout=120 --worker-class=meinheld.gmeinheld.MeinheldWorker kuma.wsgi:application"
```
safwanrahman: can we run the script in nginx container?

jwhitlock:
@safwanrahman, if you have a Linux box available, please try out this configuration and let me know if you hit permission errors or other problems. It works well enough on macOS, and we only seem to need SSL every 3 months or so, so I went ahead and merged with the open question of Linux support.

PR #4952, opened 2018-08-27
bug 1308630: Refactor clean_content

jwhitlock:
The method ``Document.objects.clean_content`` provides the default bleaching and ``iframe`` ``src`` filtering for rendered wiki content. However, it doesn't act like a manager method, and this makes it awkward to call as a Jinja2 filter or on KumaScript preview content. It also made it awkward to optionally turn off ``iframe`` ``src`` filtering via a setting.

This moves ``clean_content`` to ``kuma.wiki.content`` as a plain function, and refactors to ``AllowedHTMLTests`` tests to be standard ``pytest`` tests, and to use ``clean_content`` rather than a ``bleach.clean`` call that is almost the same.  It also reduces the test cases, since many of the existing tests were little more than ``bleach.clean`` configuration tests.

### DIFF
--- a/kuma/wiki/tests/test_content.py
+++ b/kuma/wiki/tests/test_content.py
@@ -14,7 +14,7 @@ import kuma.wiki.content
 from . import document, normalize_html
 from ..constants import (ALLOWED_ATTRIBUTES, ALLOWED_PROTOCOLS,
                          ALLOWED_STYLES, ALLOWED_TAGS)
-from ..content import (CodeSyntaxFilter, get_content_sections,
+from ..content import (clean_content, CodeSyntaxFilter, get_content_sections,
                        get_seo_description, H2TOCFilter, H3TOCFilter, parse,
                        SECTION_TAGS, SectionIDFilter, SectionTOCFilter)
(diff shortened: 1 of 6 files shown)


kuma/wiki/tests/test_content.py
```diff
@@ -988,109 +987,115 @@ def test_editor_safety_filter(self):
...
+     '<a class="foo" href="foo" id="foo" title="foo">foo</a>',
+     '<div class="foo">foo</div>',
+     '<video class="movie" controls id="some-movie" lang="en-US" src="some-movie.mpg">Fallback</video>',
+     ))
+def test_clean_content_allows_some_attributes(html):
+    """clean_content allows attributes, but doesn't preserve order."""
+    assert normalize_html(clean_content(html)) == html
+
+
+def test_clean_content_allows_some_some_styles():
```
escattone: Seems like a typo in the name? Probably wanted `test_clean_content_allows_some_styles` instead?

jwhitlock:
Force-push to rebase on master and remove some "some" from "test_clean_content_allows_some_some_styles"

PR #4953, opened 2018-08-27
bug 1308630: Allow disabling iframe filtering, or adding more patterns

jwhitlock:
*This builds on #4952, and will be easier to review after than is merged.*

Add some additional environment-based tuning of ``iframe`` ``src`` filtering:

* ``ALLOW_ALL_IFRAMES`` - If truthy, then don't filter ``iframe`` ``src`` values. Useful for local debugging of KumaScript macro changes.
* ``ALLOWED_IFRAME_PATTERNS`` - A comma-separated list of additional ``src`` patterns. This is added to the default list, rather than replacing it. It requires an update to ``python-decouple`` to handle omitted environment keys.


jwhitlock:
Rebased on master, down to the 3 commits for this change.

PR #4957, opened 2018-08-29
Front-end assets work for removing the MozToggler sidebar behavior

tkadlec:
This addresses https://app.zenhub.com/workspace/o/mdn/sprints/issues/258.

There's a PR for Kumascript as well (https://github.com/mdn/kumascript/pull/789) that will need to happen at the same time.


tkadlec:
@schalkneethling Would love your eyes on this.

kuma/static/styles/components/wiki/quick-links.scss
```diff
@@ -40,41 +40,26 @@
...
-
-        .no-js & {
-            display: block;
+    details {
+        margin-bottom: 0;
+        li {
+            padding-left: 0;
+        }
+        summary:before {
+            font-size: .6em;
```
schalkneethling: For consistency, use the `set-font-size` mixin here:

```
@include set-font-size(.6rem);
```

tkadlec: Missed that! Fixed.

kuma/static/js/wiki.js
```diff
@@ -59,20 +49,6 @@
...
 
-    // Utility method for the togglers
-    function setupTogglers($elements) {
-        $elements.each(function() {
-            var $li = $(this);
-            var $sublist = $li.find('> ul, > ol');
-
-            if ($sublist.length) {
-                $li.addClass('toggleable closed');
-                $li.find('> a').addClass('toggler').prepend('<i aria-hidden="true" class="icon-caret-up"></i>');
```
jwhitlock: Here's the before and after on [Web/Events/addstream](https://developer.mozilla.org/en-US/docs/Web/Events/addstream), with ``{{EventRef}}``:

pr 4957 - no indicator

I think the caret indicator helps distinguish between links like "Events" and menus that open like "WebRTC Events". Is there a way to indicate this without the FontAwesome ``icon-caret-*``?

jwhitlock: Hmm it may just be {{EventRef}}. Here's [Web/API/Animation/id](https://developer.mozilla.org/en-US/docs/Web/API/Animation/id) with ``{{APIRef("Web Animations")}}``, before and after.

pr 4957 - indicator

The weight of the section names are a little lighter, but it is functionally the same, with the indicator.

This includes the change from PR #4963, which allows the ``open`` attribute on ``details`` elements.  We'll need that as well, with a merge and rebase or you can cherry-pick it onto this branch.

I've opened [bug 1488884](https://bugzilla.mozilla.org/show_bug.cgi?id=1488884), since Kuma changes require a related bug. I can get the bug number into commit messages when we're ready to merge.

tkadlec: Yeah...that surprised me. The little twistie thing should be free now that we use details/summary.

tkadlec: So turns out I just missed that file. It _should_ be fixed in the Kumascript PR (https://github.com/mdn/kumascript/pull/789/commits/00fa08a97f78667126b317c47130bf40993d37e7)

Is there a secret to being able to test that page locally? Mine 404's.

schalkneethling: @tkadlec If you get a 404 it generally means you do not have that page locally and need to manually create it.

jwhitlock: The short version:

```
$ make bash # To start a shell in the web container
$ ./manage.py scrape_links https://developer.mozilla.org/en-US/docs/Web/API/Animation/id
```

See https://kuma.readthedocs.io/en/latest/data.html for all the details.

schalkneethling: Oh. my. goodness. I had no idea you could do it like this. That is going to save me a ton of time. Thanks @jwhitlock

schalkneethling:
@tkadlec I know I previously mentioned the `20px` bottom margin on the `details` element, which was changed to `0`. Looking at it now, the menu items seems to tight, and can probably do with a little breathing room ;)

screen shot 2018-09-10 at 10 52 14

Perhaps `10px`

screen shot 2018-09-10 at 10 54 37

schalkneethling:
Another layout suggestion is to give some bottom margin to each top level item, for example "CSS — Styling the Web". This is the current layout:

screen shot 2018-09-10 at 10 58 55

With the `10px` bottom margin added to `details`, and updating `.quick-links a` to:

```
...
display: inline-block;
margin-bottom: 10px;
```

The layout feels a little less cramped:

screen shot 2018-09-10 at 10 58 44

tkadlec:
@schalkneethling Made the suggested tweaks, and also played a bit with the spacing around the nest li's. Here's the final result, which I think looks pretty good.

screen shot 2018-09-26 at 3 01 00 pm

PR #4958, opened 2018-08-30
bug 1486527: Improve indicator snugging

jwhitlock:
Multiple indicator sections are separated by whitespace unless hacks and CSS classes are applied. Use the ``:matches`` selector to add some more automatic "snugging", with auto-expanded rules for browsers (like Firefox) that don't support :matches, since PostCSS will require additional work to integrate into the asset pipeline ([bug 1366868](https://bugzilla.mozilla.org/show_bug.cgi?id=1366868)).

This was cherry-picked from PR #4884, without the changes that add PostCSS and ``postcss-selector-matches``

### DIFF
--- a/kuma/static/styles/components/wiki/indicators.scss
+++ b/kuma/static/styles/components/wiki/indicators.scss
@@ -41,7 +41,8 @@ block indicators aka banners
 }
 
 /* Expand out the following for browsers which don’t support `:matches(…)`
-   TODO: Use postcss-matches-selectro once https://bugzil.la/1366868 is resolved. */
+   TODO: Use postcss-selector-matches once https://bugzil.la/1366868 is resolved.
+   See https://github.com/mozilla/kuma/pull/4884 */
 .warning + .warning,
 .overheadIndicator + .warning,
 .note + .warning,


jwhitlock:
I tested it locally with a copy of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto, with this modified markup (removed the ``overheadIndicator`` class from the second div).

```
<div>{{Deprecated_header}}
<div class="warning">
<p><strong>Warning:</strong> Changing the <code>[[Prototype]]</code> of an object is, by the nature of how modern JavaScript engines optimize property accesses, a very slow operation, in <strong><em>every</em></strong> browser and JavaScript engine. The effects on performance of altering inheritance are subtle and far-flung, and are not limited to simply the time spent in <code>obj.__proto__ = ...</code> statement, but may extend to <strong><em>any</em></strong> code that has access to <strong><em>any</em></strong> object whose <code>[[Prototype]]</code> has been altered. If you care about performance you should avoid setting the <code>[[Prototype]]</code> of an object. Instead, create a new object with the desired <code>[[Prototype]]</code> using {{jsxref("Object.create()")}}.</p>
</div>

<div class="warning">
<p><strong>Warning:</strong> While <code>Object.prototype.__proto__</code> is supported today in most browsers, its existence and exact behavior has only been standardized in the ECMAScript 2015 specification as a legacy feature to ensure compatibility for web browsers. For better support, it is recommended that only {{jsxref("Object.getPrototypeOf()")}} be used instead.</p>
</div>
</div>
```

Here's the rendering before and after:

pr 4958 before vs after

jwhitlock:
Oops, I pushed the commit to master 😕 . The only remaining change is the docstring fix.

ExE-Boss:
🤦🏻‍♂️ Well, we should still at least get this docstring change into `master`.

Luckily, this change is very low‑risk.

PR #4962, opened 2018-09-05
Bug 1492018: Use `.blockIndicator` for styling `{{SeeCompatTable}}`

ExE-Boss:
This is a follow‑up to mdn/kumascript#788, which added the standard `.overheadIndicator` class to the `{{SeeCompatTable}}` macro.

Depends on #4979


jwhitlock:
There's no mention of ``notice`` on https://github.com/mdn/kumascript/pull/788, maybe you mean https://github.com/mozilla/kuma/pull/4884#issuecomment-417121020 ? Should this PR remove some ``.notice`` rules as well then? And, do you have a page where all this matters?

ExE-Boss:
I meant specifically commit https://github.com/mdn/kumascript/commit/d0c3a47deced6d2f1cb7eada221c573c6bda7c46.

Once this is merged, I can make a follow-up to https://github.com/mdn/kumascript/pull/788, where I’d remove the `notice` class from the `{{SeeCompatTable}}` macro.

ExE-Boss:
review?(@jwhitlock): I’d say this is ready for review.

jwhitlock:
I think I'm the wrong person to review this. I can review style changes if you are trying to fix something that I can see on a particular page on MDN, because I can load the page in my browser before and after the change and see what the effect is. I'm not aware of any pages that this change would make a visual difference on.

This appears to be a refactoring of classes for some quality goal in the CSS source, which isn't clear to me. I think @schalkneethling or @hobinjk are better suited to review.

ExE-Boss:
> I think I'm the wrong person to review this. I can review style changes if you are trying to fix something that I can see on a particular page on MDN, because I can load the page in my browser before and after the change and see what the effect is. I'm not aware of any pages that this change would make a visual difference on.

This is necessary for [mdn/kumascript#830](https://github.com/mdn/kumascript/pull/830/files#diff-e515d1cf4fe601f0046012893c234bd7L15).

jwhitlock:
@schalkneethling or @hobinjk, please take a look when you have a chance. There's several PRs in the same topic, and I've [documented my understanding of the issue](https://discourse.mozilla.org/t/indicator-style-class-simplification-project-prefer-blockindicator-for-styling-block-indicators/31976/5).

I'd also support closing this and related PRs, it is hard for me to justify spending more time on what appears to be low-priority cleanup work.

schalkneethling:
As mentioned, it would be super useful if you could merge the work here into https://github.com/mozilla/kuma/pull/4979

ExE-Boss:
I now have developed a better solution locally than this now anyway.

PR #4967, opened 2018-09-11
Bug 1490392: Reduce web font usage

tkadlec:
This PR addresses https://app.zenhub.com/workspace/o/mdn/sprints/issues/255 by removing Zilla Slab Italic in favor of font synthesis and eliminating Open Sans and using Verdana instead.

This drops our web font count from 6 to 2, decreases our font weight by 39% (211.4kb) and improves reduces the amount of rendering work the browser has to do.


tkadlec:
@schalkneethling @jwhitlock Killing fonts here. Would love a review.

jinja2/includes/preload.html
```diff
@@ -1,2 +1,3 @@
 {# because this is non-blocking, preload as early as possible #}
 <link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Regular.woff2') }}" as="font" type="font/woff2" crossorigin />
+<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Bold.woff2') }}" as="font" type="font/woff2" crossorigin />
```
tkadlec: I'm sneaking another preload in here now that we have it down to just the two web fonts. Helps a bit with rendering (about 35% reduction over not preloading for the home page which doesn't have much render work to begin with, but only about 5% reduction over no preload on an interactive example page), but mostly ensures the web fonts are available right away.

kuma/static/styles/components/wiki/toc.scss
```diff
@@ -47,7 +47,6 @@ $padding-horizontal: 30px;
         color: $accent-light;
         display: inline-block;
         @include set-font-size(17px);
-        font-weight: bold;
```
tkadlec: Verdana's bold is heavier than Open Sans. I pulled the bold here (and a couple other spots) to minimize the difference.

kuma/static/styles/base/elements/document.scss
```diff
@@ -8,6 +8,7 @@ html {
 }
 
 body {
-    line-height: 1.5;
+    line-height: 1.6;
     @include set-site-font-family();
+    @include set-font-size(15px);
```
tkadlec: I don't normally like to reduce the font size below 16px, but Verdana is big. Playing with https://meowni.ca/font-style-matcher/ shows that 15px (with some tiny tweaks like the line-height and some letter-spacing) is nearly identical to 16px Open Sans.

kuma/static/styles/includes/_vars.scss
```diff
@@ -177,18 +177,18 @@ break points
 - divide pixels by browser default font size (16px)
 - and multiply by 1em to convert unit to em.
 */
-$max-width-default: (1399px / 16px) * 1em;
-$small-desktop-ends: (1199px / 16px) * 1em;
-$tablet-ends: (1023px / 16px) * 1em;
-$mobile-ends: (767px / 16px) * 1em;
-$small-mobile-ends: (479px / 16px) * 1em;
```
tkadlec: This was necessary to keep the same general width of our content. I tested the layout using devtools and we seem ok with the minor adjustments to our breakpoints.

tkadlec:
A few side by sides for comparison. (After font changes on the left, before font changes on the right)

screen shot 2018-09-11 at 9 16 04 am
screen shot 2018-09-11 at 9 15 18 am
screen shot 2018-09-11 at 9 15 01 am

schalkneethling:
@tkadlec Thanks for the PR. One thing I notice is that no actual font files are removed in this PR. As we are no longer using them, I would suggest you delete them from the codebase as well. Thanks!

tkadlec:
Heh. I was just realizing the same thing. :)

tkadlec:
There was a whole bunch of random variations of ZillaSlab that we weren't using anywhere, so I removed those as well now.

jwhitlock:
I opened [bug 1490392](https://bugzilla.mozilla.org/show_bug.cgi?id=1490392) and updated the title for this issue. We have a bugzilla bot that will see a merge with the text "Bug 255", and will update [bug 255](https://bugzilla.mozilla.org/show_bug.cgi?id=255), reported 21 years ago, instead of https://github.com/mdn/sprints/issues/255.

kuma/static/styles/includes/_vars.scss
```diff
@@ -177,18 +177,18 @@ break points
...
+$max-width-default: (1399px / 15px) * 1em;
+$small-desktop-ends: (1199px / 15px) * 1em;
+$tablet-ends: (1023px / 15px) * 1em;
+$mobile-ends: (767px / 15px) * 1em;
+$small-mobile-ends: (479px / 15px) * 1em;
```
jwhitlock: Nit: Is it possible to move ``15px`` into a SCSS variable? This change from ``16px`` to ``15px`` would have been a one-line change...

kuma/static/styles/gaia.scss
```diff
@@ -5,4 +5,3 @@ Fonts for Gaia demos
 ********************************************************************** */
 
 @import 'base/fonts/moztt';
-@import 'base/fonts/opensans';
```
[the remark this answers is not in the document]
schalkneethling: From what I know, Gaia was the codename for the front-end codebase of Firefox-OS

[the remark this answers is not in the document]
jwhitlock: @schalkneethling is right: https://developer.mozilla.org/en-US/docs/Archive/B2G_OS/Platform/Gaia

A lot of this was added in 2013, like [bug 844531](https://bugzilla.mozilla.org/show_bug.cgi?id=844531), to support documentation and samples. There's a few samples in the archive, like https://developer.mozilla.org/en-US/docs/Archive/B2G_OS/Firefox_OS_apps/Building_blocks/Slider/2.0. That one has been converted to use external assets, hosted on GitHub.

Removing these may break some samples, but keeping those samples working is a low priority. Future work.

jwhitlock:
I mostly reviewed from the backend, "does this break the build" perspective. I'm not qualified to judge how it looks, or if the new media breakpoints are appropriate. I think a positive review from @schalkneethling is needed to merge.

schalkneethling:
@tkadlec I am going to suggest that we add the following to the `body` element:

```
-webkit-font-smoothing: antialiased;
-moz-osx-font-smoothing: grayscale;
```

A large portion of our audience will get the benefit and it really makes the copy a lot more readable:

compare

kuma/static/styles/includes/_vars.scss
schalkneethling: Agreed. It is dividing by the base body font size. I would suggest updating the variable here https://github.com/mozilla/kuma/blob/master/kuma/static/styles/includes/_vars.scss#L132 to be `15px` and then use the variable for consistency.

kuma/static/styles/includes/_mixins.scss
```diff
@@ -41,6 +41,7 @@ Provides main mixins for use within the MDN theme.
 
 @mixin set-site-font-family() {
     font-family: $site-font-family;
+    letter-spacing: -.05px;
```
schalkneethling: Nit: Prefer `rem` units over `px` for typography

schalkneethling:
I notice that the newsletter section overrides `Verdana` with `x-locale-heading-primary, zillaslab, "Palatino", "Palatino Linotype", x-locale-heading-secondary, serif`. Not sure if there is a specific reason for this, but it looks strange to have a serif mixed in with a sans-serif typeface.

screen shot 2018-09-12 at 11 43 16

The same is happening on production so, was not introduced here but, thought I would mention it. @jwhitlock you know of a specific reason for this?

schalkneethling:
This is looking good. Thanks @tkadlec - just a couple of items to address, and then this should be good to go. 🎉

kuma/static/styles/includes/_mixins.scss
tkadlec: Should be set here.

kuma/static/styles/includes/_vars.scss
tkadlec: All set!

kuma/static/styles/base/elements/document.scss
schalkneethling: You can use `$body-font-size` here now.

schalkneethling:
@tkadlec What are your thoughts on https://github.com/mozilla/kuma/pull/4967#issuecomment-420575013 ?

tkadlec:
@schalkneethling I thought I had done that! I remember looking at it in the browser. :/ Anyway...added and pushed now.

kuma/static/styles/base/elements/document.scss
tkadlec: 👍

jwhitlock:
@tkadlec is a squash commit OK for these changes?

tkadlec:
@jwhitlock Yeah, I reckon so.

a2sheppy:
I love the size reduction of page loads here, but there are definitely some text legibility problems introduced by this. Letter spacing is a little too broad, especially for bold type, and is really pretty bad when looking at bold type in the editor, as seen in this screen shot:

image

Previously, this was much more legible, looking something like this (this is sort of a recreation made by tweaking the CSS to restore Open Sans as the body font):

image

tkadlec:
Believe we can tweak this to make it better. To be clear, are you saying that wherever Verdana is bold we have issues?

stephaniehobson:
I think I'd prefer arial to verdana. I think the legibility is more important than matching the old width.

schalkneethling:
> I think I'd prefer arial to verdana. I think the legibility is more important than matching the old width.

While `font-smooth` does improve it, I have to agree that Arial is more legible font-face in general. Not sure if the decision was made by brand or, based on the closest match. @tkadlec @atopal you know?

atopal:
I think it was made based on closest match. I'm fine with Arial.

tkadlec:
Yup....based on similarity. All for switching to Arial. Is there anyone else we should include in this discussion or are we good to go?

jwhitlock:
I'm OK with the change, not sure about the timing. @atopal, do you want to aim for before or after Oct 1st to ship, or does it mater?

PR #4974, opened 2018-09-13
bug 1490727: Add stripe dependency

jwhitlock:
stripe 2.8.0 is the Python library for the [Stripe Payments API](https://stripe.com/docs/api). The commits are cherry-picked from https://github.com/mozilla/kuma/tree/potato-1490727, and have two authors.


escattone:
While checking the changelog for `stripe`, I noticed that [version 2.8.1 was released today](https://github.com/stripe/stripe-python/blob/master/CHANGELOG.md#281---2018-09-13). Would you like to move to the newer version within this PR?

jwhitlock:
Looking at the [Changelog](https://github.com/stripe/stripe-python/blob/master/CHANGELOG.md), they seem to like releasing early and often, sometimes twice a day...

PR #4979, opened 2018-09-19
Bug 1492018: Prefer `.blockIndicator` for styling block indicators

ExE-Boss:
See [bug&nbsp;1492018](https://bugzil.la/1492018) for details.

## TODO
- [x] Add the&nbsp;`.properties` class to&nbsp;the&nbsp;`:matches(…)` selector (see [`ime-mode`](https://developer.mozilla.org/docs/Web/CSS/ime-mode) for a&nbsp;case where this is&nbsp;necessary)
- [x] Push remaining commits (3/3)

## Changes:
- Add support for snugging `.blockIndicator`.
- Add indicator colour override classes (obsoletes&nbsp;https://github.com/mozilla/kuma/issues/4962):
  - `.indicator-info` - grey
  - `.indicator-version` - blue
  - `.indicator-warning` - yellow
  - `.indicator-danger` - red
- Use classes instead of `style` attributes for ES7/Harmony and JSOverrides.
- Fix `<pre>` block and `.properties` table snugging.

---

review?(@schalkneethling)


schalkneethling:
@ExE-Boss please rebase this against the latest in master. There are some display problems caused by this branch being out of sync with some recent changes. Thanks!

ExE-Boss:
@schalkneethling: This is ready for a second round of review.

schalkneethling:
Thanks for the update @ExE-Boss, will review as soon as possible.

kuma/static/styles/components/wiki/indicators.scss
```diff
@@ -206,14 +269,9 @@ icons for both types of indicators
     @include indicator-icon('\f017'); /* clock */
 }
 
+.esHarmony,
```
schalkneethling: We need to standardize on how classes are written. For the vast majority of the code base, and pretty much the standard across projects, class names should follow the pattern `descriptive-word` and not the JS variable form of `descriptiveWord`

Therefore this needs to be updated to be `es-harmony`. There are a couple of classes here that does not follow that convention, we can clean this up in a follow on pull request.

kuma/static/styles/components/wiki/indicators.scss
```diff
@@ -206,14 +269,9 @@ icons for both types of indicators
...
 .experimental,
-.experimentalBadge,
-/* es7 and harmony macros  */
-.overheadIndicator[style='background: #9CF49C;'],
-/*
-IE 11 converts hex to rgb, so target that as well.
-*/
-.overheadIndicator[style='background: rgb(156, 244, 156);'] {
+.experimentalBadge {
```
schalkneethling: Same comment here as above, unless this class is already widely used in this form. If that is the case, we can clean this up later.

kuma/static/styles/components/wiki/indicators.scss
```diff
@@ -356,13 +433,15 @@ Special snowflakes
...
-    border-color: #{map-get( map-get($colors-lookup, 'positive'), 'medium')} !important; /* stylelint-disable-line declaration-no-important */
-    background: #{map-get( map-get($colors-lookup, 'positive'), 'light')} !important; /* stylelint-disable-line declaration-no-important */
+.template-jsOverrides {
+    border-color: #d1d1ff;
+    background: #f5f5ff;
 }
 
+/* es7 and harmony macros */
+.esHarmony {
+    @include box-theme('positive');
```
schalkneethling: I feel like we need to be consistent with the colors we assign to these callout widgets. Experimental feel to me more like a blue, or even orange, than a green.

Here for example it is blue:
http://localhost:8000/en-US/docs/Web/API/DeviceLightEvent/Using_light_sensors

But then on the `Object.observe` page locally it is green:
http://localhost:8000/tr/docs/Web/JavaScript/Reference/Global_Objects/Object/observe

I would suggest that we keep this consistent, and rather opt for blue in experimental cases.

ExE-Boss: I think I’ll just use the `.experimental` and `.indicator-warning` classes for `{{ES7}}` to match `{{SeeCompatTable}}`.

kuma/static/styles/components/wiki/indicators.scss
ExE-Boss: `.experimentalBadge` is, like most other legacy classes in this file, used quite a bit in this form on MDN. 😔

schalkneethling: No worries, we can clean this up later.

ExE-Boss:
ping @schalkneethling

ExE-Boss:
So, can this be merged now?

schalkneethling:
Ok, I have tested this on a couple of pages(will do some more testing) where I have the new styles in Kuma, but the old macros in kumascript. I cannot find any instances where things fall apart.

So, I believe we are safe to merge this and the KumaScript PR and push to production. For the space of time where we will have the old macros and new classes(until the pages have been regenerated) all should be fine.

With that said, we should probably hold of until Monday when @jwhitlock is back.

jwhitlock:
> With that said, we should probably hold of until Monday when @jwhitlock is back.

I don't think this needs to wait for me. It appears to me the worst that can happen is a banner is un-styled, and @ExE-Boss can then either force-refresh a page or edit the content to add the class.

ExE-Boss:
So then let’s merge this and https://github.com/mdn/kumascript/pull/830.

(also I’ve had this loaded in [Stylus](https://addons.mozilla.org/firefox/addon/styl-us/) for the last week or so and everything looks fine)

PR #4982, opened 2018-09-21
Removing preload for Zilla Bold

tkadlec:
In #4967, we slipped an extra font `preload` in and it looks as if it may be offsetting the improvements brought on by the font reduction.

This PR removes the `preload` for Zilla Slab Bold for now. We can revisit this later on after we've subsetted, etc to see if maybe it makes sense at that point.


tkadlec:
@jwhitlock @schalkneethling Quick one.

jinja2/includes/preload.html
```diff
@@ -1,3 +1,2 @@
 {# because this is non-blocking, preload as early as possible #}
-<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Regular.woff2') }}" as="font" type="font/woff2" crossorigin />
-<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Bold.woff2') }}" as="font" type="font/woff2" crossorigin />
+<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Regular.woff2') }}" as="font" type="font/woff2" crossorigin />
```
schalkneethling: Nit: Missing empty line at end of file

tkadlec: Fixed. :)

jwhitlock:
Fast work gents! We’re aiming for a prod push this weekend.

PR #4988, opened 2018-09-22
End compat table beta

stephaniehobson:
Removes beta menu and links.
Forces display of new tables during roll out (the hack for this can be removed once the tables are rolled out).

This should be merged with mdn/kumascript#860


stephaniehobson:
Tested @Elchi3's code locally. Works great. R+

PR #5000, opened 2018-09-24
Add accessible text to social icons

hidde:
the `focusable=none` attribute is used so that in IE11, the icon does not get focus

Signed-off-by: Hidde de Vries <hidde@hiddedevries.nl>


jinja2/includes/icons/social/facebook.svg
```diff
@@ -1,4 +1,4 @@
-<svg class="icon icon-facebook" xmlns="http://www.w3.org/2000/svg" width="16" height="28" viewBox="0 0 16 28">
+<svg class="icon icon-facebook" xmlns="http://www.w3.org/2000/svg" width="16" height="28" viewBox="0 0 16 28" aria-label="Facebook" role="img" focusable="false">
```
schalkneethling: Thank you for your contribution @hidde - I recently read this article to understand SVG and a11y better: https://developer.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/

Based on that, I wonder if it makes sense to instead do `aria-labelledby="title"`

For the Discourse SVG we would then need to add a `title` and then make the same change. Thoughts?

hidde: I think in terms of the accessibility tree, an `aria-label` and `aria-labelledby` with an idref would yield the same result: they give the image an accessible name, which conveys alternative text to assistive technologies. The difference should be mostly personal preference.

If you prefer going the `aria-labelledby` route I am more than happy to update my PR accordingly.

schalkneethling: True that. Seeing we already have the text there in the `title` tag, let's go with `aria-labelledby`. Thanks @hidde

hidde: Aah, @schalkneethling, apologies, when I just started to fix it, I remembered that I did try and use ID's first, but found  a problem with it: some icons are used twice. An example is the Twitter icon, which is used for MDN Twitter and Mozilla Twitter. That would mean the ID would occur twice on the page, and that's invalid HTML. 

Would there be a way to make these IDs dynamic, so that we can add unique ones? One dirty yet easy way I could think of if have mdn-twitter and mozilla-twitter svg files, with different IDs in them. But that's extra requests for the users. What do you think?

schalkneethling: If by ids you are referring to the `title` tag inside the SVG, these are/can be dynamic. There is a Jinja helper[https://github.com/mozilla/kuma/blob/master/kuma/wiki/templatetags/jinja_helpers.py#L252] that does this. So, if aria-labelledby points to the `title` tag, then there should be no problems. If a string needs to be translated, the above macro will also take care of that.

So, I would say adding `aria-labelledby="title"` to the root `svg` element, is what is needed here. Thanks!

hidde: Ah, but aria-labelledby points to an id, so it would be `<title id="twitter-title">` with `aria-labelledby="twitter-title"`. 

Would a jinja helper work inside the svg? (as it is not a jinja file?)

jwhitlock: The ``include_svg`` Jinja2 macro is used to set localized titles for inlined SVG, such as in the footer:

https://github.com/mozilla/kuma/blob/e39c2b7fd46907198dd1d052025bfb2488935ce8/jinja2/includes/footer/moz-footer.html#L8-L17

It does this by parsing the SVG as XML and modifying the title:

https://github.com/mozilla/kuma/blob/9d38be950e5202a05790af151cbc24e6ecd192f5/kuma/wiki/templatetags/jinja_helpers.py#L250-L261

Similar magic could be used to modify the ``id`` value.

schalkneethling: @hidde @jwhitlock Now this `id` bit is starting to bother me(I know, I suggested it ;)) a little. We will have to be very careful to ensure we do not get duplicate ids on a page.

If we for example have two or three instances of the same icon on the same page, it seems likely that we are going to have a problem where two have the same `id`.

Perhaps it will be better to specify the accessible title explicitly with `aria-label`, setting it to the same value as the `title` element. Seems a bit more sane, and less error prone. That way `aria-label` also inherits the benefit of being localised where appropriate.

So the tl;dr - I suggest we:

1. Stick with using `aria-label`
2. Add a `title` element to the Discourse SVG for consistency
3. Post merge, I can update the helper to also update the `aria-label`

Thoughts?

jwhitlock: ``aria-label`` makes sense to me, even with duplication. Unique ``id``s would be manageable when used for footer icons, but not for other uses, like the experimental and obsolete icons in sidebars.

hidde: Sounds good to me, I will add the title when I'm at my computer.

hidde: I've just added it.

PR #5002, opened 2018-09-24
bug 948151: Add permissive CSP

jwhitlock:
Builds on PR #5001, which can be merged first to get the libraries into the base image.

Add a CSP configuration, based on testing several pages on development. Enabling CSP and other options are configured from the environment, and off by default. The policy is configured in the Django settings, and not configured by the environment. The policy allows inline CSS and JavaScript, so that the site works without modification, but doesn't add as much security as it could.

CKEditor 4.5.10 still requires ``unsafe-eval``, but [4.7](https://ckeditor.com/blog/CKEditor-4.7-released/) claims to remove this requirement. A decorator adds ``unsafe-eval`` only on the editing pages that use CKEditor.

The reporting view is copied from [mozilla/bedrock](https://github.com/mozilla/bedrock/blob/333d3df02ad697f1694a88553119f98931519557/bedrock/base/views.py#L153-L181), and it used to forward violation reports to Sentry. There were no tests to copy.

### DIFF
--- a/kuma/health/views.py
+++ b/kuma/health/views.py
@@ -1,10 +1,15 @@
+import json
+import logging
+
 from django.conf import settings
 from django.db import DatabaseError
-from django.http import HttpResponse, JsonResponse
+from django.http import HttpResponse, HttpResponseBadRequest, JsonResponse
 from django.views.decorators.cache import never_cache
-from django.views.decorators.http import require_safe
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST, require_safe
 from elasticsearch.exceptions import (ConnectionError as ES_ConnectionError,
                                       NotFoundError)
+from raven.contrib.django.models import client
 from requests.exceptions import ConnectionError as Requests_ConnectionError
 
 from kuma.users.models import User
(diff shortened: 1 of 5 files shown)


jwhitlock:
> With everything back to normal, I logged in, and then noticed that I was getting a CSP violation for http://i2.wp.com violating the img-src directive (which was triggered by the call to https://secure.gravatar.com -- i2.wp.com is an image CDN). I discovered that this was due to the fact that I was running in non-SSL mode. When I switched to SSL mode by adding COMPOSE_FILE=docker-compose.yml:docker-compose.ssl.yml to my .env file, the CSP violation disappeared.

I had similar problems with the gravatar images, but I think that's because the ``STATIC_URL`` is set to a relative path ``/static/`` in development, and that turns into http://i2.wp.com/static/img/avatar.png and an error (or something similar). I think if ``STATIC_URL`` was ``http://localhost:8000/static/``, then both gravatar and CSP would be happy. I snuck a full ``STATIC_URL`` in ``docker-compose.yml`` into PR #5014. SSL-mode includes a full ``STATIC_URL`` to a local file, and I think I had similar good results with that.

> I assume the next step is to enable CSP but only in reporting mode, with these settings in an infra PR:
>     CSP_ENABLE_MIDDLEWARE=True
>     CSP_REPORT_ENABLE=True
>     CSP_REPORT_ONLY=True

Yes, that's my thought too, but start with reporting-only on staging, and make sure the Sentry integration works with a manageable number of reports, before we go reporting-only in production and turn on the fire hose. I've also had a decent experience running a similar configuration in local development, we may want to make that a default to catch issues before they get to production, once we have an enforced CSP.

kuma/health/views.py
```diff
@@ -156,3 +161,39 @@ def status(request):
...
+    Capture CSP violation reports, forward to Sentry.
+
+    HT @glogiotatidis https://github.com/mozmeao/lumbergh/pull/180
+    HT @pmac, @jgmize https://github.com/mozilla/bedrock/pull/4335
+    """
+    if not settings.CSP_REPORT_ENABLE:
+        # mitigation option for a flood of violation reports
+        return HttpResponse()
+
+    data = client.get_data_from_request(request)
```
jwhitlock: I added a breakpoint here. The POST data, including the full report, is captured by ``client.get_data_from_request(request)``, and sent with ``client.captureMessage`` below.

escattone: 🎉 Awesome, thanks for checking that!

escattone:
Good to know about the relative vs full STATIC_URL, thanks @jwhitlock!

kuma/health/views.py
```diff
@@ -156,3 +161,39 @@ def status(request):
...
+        # Cannot decode CSP violation data, ignore
+        return HttpResponseBadRequest('Invalid CSP Report')
+
+    try:
+        blocked_uri = csp_data['csp-report']['blocked-uri']
+    except KeyError:
+        # Incomplete CSP report
+        return HttpResponseBadRequest('Incomplete CSP Report')
+
+    client.captureMessage(message='CSP Violation: {}'.format(blocked_uri),
```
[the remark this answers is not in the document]
pmac: I feel we do, but it could be better. We don't have this enabled in production for fear of DDOSing ourselves. Mostly we see blocked things being injected by addons.

PR #5004, opened 2018-09-26
Bug 1490727, defer stripejsand load async

schalkneethling:
@jwhitlock This now defers and loads stripejs async on all pages where it is used in popover mode. On the contribute page, it is still loaded as before. r?

### DIFF
--- a/kuma/contributions/jinja2/contributions/contribute.html
+++ b/kuma/contributions/jinja2/contributions/contribute.html
@@ -4,12 +4,6 @@
 
 {% block title %}{{ page_title(_('Contribute')) }}{% endblock %}
 
-{% block site_js %}
-    {{ super() }}
-    <script src="https://checkout.stripe.com/checkout.js"></script>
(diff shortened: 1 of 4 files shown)


kuma/contributions/jinja2/contributions/contribute.html
```diff
@@ -4,12 +4,6 @@
 
 {% block title %}{{ page_title(_('Contribute')) }}{% endblock %}
 
-{% block site_js %}
-    {{ super() }}
-    <script src="https://checkout.stripe.com/checkout.js"></script>
-    {% javascript 'contribute' %}
-{% endblock %}
```
schalkneethling: @jwhitlock This is also done at the bottom of the page so, this looks like duplication to me. If not, let me know and I will revert.

joshJarr: 👍 I think you're right, this is duplicated and should be removed. Thanks!

PR #5005, opened 2018-09-27
Bug 1490727: Various blockers for release.

joshJarr:
In this PR We have addressed: 

- [x] Copy changes, email, social share, thank you page.
- [x] Flow for the stripe error pages (needs copy)
- [x] Stripe data fixes - we now send the name of the contributor in the metadata and set the receipt email. This should be tested with the live stripe account.
- [x] Fixes email images being loaded with the wrong domain on staging.
- [x] Lazy load `checkout.js` using `$.getScript()` on all pages upon interaction with the popover (except `/contribute`).
- [x] Fix banner flashing bug after it has been closed.
- [x] Fix Analytics firing when the popover has been dismissed or is not shown on mobile.

Let me know if you have any feedback or if we have missed anything, Thanks!


jwhitlock:
@joshJarr thanks for the PR. Do you want to try to fix the flake8 issue and 2 test failures?

joshJarr:
@jwhitlock taking a look now 👍

joshJarr:
Got a fix, will test everything and push in a hour or two.

kuma/contributions/jinja2/contributions/thank_you.html
```diff
@@ -5,46 +5,44 @@
...
                 {% else %}
                     <h2 class="highlight highlight-spanned">
                         <span class="highlight-span">{{_('Ooops, something went wrong')}}</span>
-                        <p>
-                            <a href="{{ url('contribute') }}">{{_('Click here to go back')}}</a>
-                        </p>
                     </h2>
+                    <p>
+                        {% trans %}
+                        Please contact <a href="mailto:support@mozilla.com">support@mozilla.com</a>  for more information on this problem.
```
jwhitlock: Future task: We'll need to move the email outside of the translated string, and use a real email address (unless support@mozilla.com is the real email address).

kuma/static/js/contribution-handler.js
```diff
@@ -399,10 +408,36 @@
         }
     }
 
+    /**
+     * Gets and executes stripe's checkout.js script to be used when submitting
+     * also handles errors when getting the resource
+     */
+    function getStripeCheckoutScript() {
+        $.getScript('https://checkout.stripe.com/checkout.js')
```
jwhitlock: I think if ``stripeHandler`` is initialized (not ``null``), then the script should not be reloaded. You can see the multiple loads by watching the network tab while collapsing and expanding the banner multiple times.

kuma/static/js/contribution-handler.js
```diff
@@ -399,10 +408,36 @@
...
+    function toggleScriptError() {
+        formButton.attr('disabled') ? formButton.removeAttr('disabled') : formButton.attr('disabled', 'true');
+        formErrorMessage.toggle();
+    }
+
     /**
      * Expands the popover to show the full contents.
      */
-    function expandCta() {
+    function expandPopover() {
```
jwhitlock: I think expanding the popover deserves to be tracked in analytics.

kuma/contributions/forms.py
```diff
@@ -118,12 +118,22 @@ def make_charge(self):
...
+                stripe.Charge.create(
+                    amount=amount,
+                    currency='usd',
+                    source=token,
+                    description='Contribute to MDN Web Docs',
+                    receipt_email=self.cleaned_data['email'],
+                    metadata={'name': self.cleaned_data['name']}
+                )
+                return True
+            except Exception as e:
```
jwhitlock: I'm not a fan of generic exception handlers, but this worked for the card error for ``4100000000000019``, so it removes the launch blocker. I can adjust the error handler later.

kuma/static/js/contribution-handler.js
schalkneethling: Also, you should add the following config:

```
$.ajaxSetup({
  cache: true
});
```

If that is not set, then the script will be loaded over the network each time it is requested, instead of loading it from browser cache.

kuma/static/js/contribution-handler.js
```diff
@@ -412,7 +447,7 @@
         // if not already initialised
         if (stripeHandler === null) {
             // initialise handler
-            stripeHandler = initStripeHandler();
+            // stripeHandler = initStripeHandler();
```
schalkneethling: @joshJarr From the diff it looks like this line is commented out. So, if `stripeHandler` is `null` nothing is going to happen. Is this intentional? If so, and this is handled somewhere else in the code, then it is probably best to completely remove this piece of code?

matthewhall: Hi Schalk. Yep, I think this `if` statement is redundant and slipped through. `stripeHandler` will be set as part of `getStripeCheckoutScript()` when the popover is expanded. We then check if it's not `null` in `onSubmit` then the form is submitted.

I think, as a further improvement to this, we might consider showing an error message if the script fails to load for whatever reason or if it hasn't loaded by the time the user clicks submit, but I think this is an adjustment we can probably follow up with next week.

kuma/contributions/forms.py
matthewhall: Please let us know if you'd like us to adjust this.

kuma/static/js/contribution-handler.js
matthewhall: Good suggestions, thank you. Please also see my comments around error handling improvements for the script loading as in the other comment as well.

Are we too consider this feedback launch blocking or is it something we can follow up with quickly next week?

kuma/contributions/jinja2/contributions/thank_you.html
matthewhall: Hi John. Yep, I believe we're still waiting on the final support email address. As of yesterday, Kadir still hadn't received this so it might fall to you guys to update once you have it.

matthewhall:
Hi @jwhitlock @schalkneethling,

Thanks for the review. Michal has pushed a fix for the tests this morning: a57c89c. Thanks for bearing with us on that.

I've added a couple of comments in to your feedback above. Please could you let me know if any of these are launch blocking and need to be addressed before Monday? If possible, and none of these are P0s, then it might be prudent to address and quickly follow up with fixes for them post-launch. Concerned of further changes today potentially causing bugs so close to launch and would be cool to give a little more time to addressing them and any other fixes needed and have a day next week to test before launching them.

Please let me know your thoughts. Thanks!

schalkneethling:
@atopal Thoughts?

kuma/contributions/jinja2/contributions/thank_you.html
atopal: The email address is mdn-support@mozilla.com

atopal:
Agreed, the performance and UX issues mentioned here should be dealt with before Wednesday, but I don't see them as blockers for Monday. I'll defer to John and Schalk for anything else.

PR #5019, opened 2018-10-02
bug 1504246: Manually inline a set of examples for testing in production

hobinjk:
In support of https://github.com/mdn/sprints/issues/467, this is only the first step on a long journey. I just want to make sure that it seems like a reasonable approach.

This introduces the process_inlined_examples function which I'll be filling in with the manually inlined interactive example sources (likely putting it in a new file). I'll also be adding the resources/assets required by the examples to kuma and fixing any visual bugs that pop up.


jwhitlock:
First of all, it works:

screenshot 2018-10-02 16 34 34

TravisCI / flake8 isn't happy with the whitespace, but that it easy enough to fix. If you are going to do Kuma work, it would be worth seeing how to integrate flake8 with your editor.

I don't think we can merge something that adds a "hello world" if ``@AlertHelloWorld@`` is in the content - there's nothing stopping a user from doing it. Adding an HTML comment would give the same proof-of-concept, but be mergable.

I think a long-term home for the code would be a filter like ``IFrameHostFilter``:

https://github.com/mozilla/kuma/blob/6d650f4a2b7090896154650d3ab3cbd3693a7fd8/kuma/wiki/content.py#L1094-L1137

This filter runs after bleaching, and is used to validate the ``iframe`` source. It could also inject HTML by yielding different elements. However, I'm not sure where you are headed, so I'm also not sure if a htmllib_filter is appropriate.

Another possibility is the code that extracts sidebars and moves them to the sideof the document. I'll have to track down that code again.

I think this is a decent place to start, I'd like to see more.

hobinjk:
Okay, I'm at the point where the inline interactive example is visually 1:1 with the embedded. To do this I had to rip out a ton of global css rules. The remaining steps I need to take are reinstating these rules and then make sure that there are no conflicts with the inlined interactive example styles and code.

screen shot 2018-10-17 at 1 49 32 pm

@atopal With this actually off the ground and working, I'm relatively confident that the remaining work will take about 5 points of time and be done within this sprint.

atopal:
cool. Thanks James! Can you walk me through how we'd measure metrics for this or should we have a quick working session to hammer that out?

Thanks!
Kadir

schalkneethling:
>  It may be useful to add comments that the code is temporary.

Agreed

> Do we want inlined versions of the other key pages (JS, HTML)?

I would highly recommend that we test at least one of each i.e. Highest trafficked page for JS, CSS, and HTML examples. This way we get an overall view.

> Is week vs week numbers good enough to test the hypothesis?

@atopal Would probably be best to answer this one but, I reckon with the amount of traffic we get an these high traffic pages, a week of data on each side should be sufficient.

schalkneethling:
@hobinjk There is some code in `editor.js` that would need to be changed as it is assuming that it is run from within an `iframe`.

The specific change is `window.parent.postMessage(...`. This can be rewritten as just `postMessage(...`

kuma/wiki/kumascript.py
```diff
@@ -214,8 +214,54 @@ def process_body(response):
...
+      <div class="buttons-container">
+          <button id="execute" class="button run" type="button">Run &rsaquo;</button>
+          <button id="reset" type="button" class="button">Reset</button>
+      </div>
+      <div id="output" class="output"><code></code></div>
+  </div>
+</section>
+
+<script src="/static/js/interactive-examples/codemirror-5-31-0.js"></script>
+<script src="/static/js/interactive-examples/editor-js.js"></script>
```
schalkneethling: Would be ideal if we could load the JS files at the bottom along with the other JS. Currently the browser will be blocked when it hits the CSS and JS assets which will not be the case if this is actually baked into Kuma.

It is probably a limitation we need to keep in mind when looking at the numbers. Perhaps we can load the scripts "dynamically" based on the presence of a specific `id` attached to https://github.com/mozilla/kuma/pull/5019/files#diff-bad48985f515ae03ad4b8ad4839f7b2dR227

Perhaps something like:

`<div id="inlined-foreach" class="iex interactive interactive-js">`

then:

```
var inlinedForeach = document.getElementById('inlined-foreach');

if (inlinedForeach) {
    // create script tags and stick into the DOM
}
```

Thoughts @hobinjk

schalkneethling:
Couple of comments on the implementation, and some require changes. Generally, this is great stuff and works well locally. Thanks @hobinjk

kuma/wiki/kumascript.py
```diff
@@ -214,8 +214,54 @@ def process_body(response):
...
+def process_inlined_examples(body):
+    """
+    Manually expand @IExMacro@ declarations
+    """
+    return body.replace("@InlineArrayForEach@", """
+<div class="iex interactive interactive-js">
+<link href="/static/styles/interactive-examples/codemirror-5-31-0.css" rel="stylesheet" />
+<link href="/static/styles/interactive-examples/editor-js.css" rel="stylesheet" />
+
+<script>"use strict";function postToKuma(e){window.parent.postMessage(e,"https://developer.mozilla.org")}postToKuma({markName:"interactive-editor-loading"}),document.addEventListener("readystatechange",function(e){switch(e.target.readyState){case"interactive":postToKuma({markName:"interactive-editor-interactive",measureName:"ie-time-to-interactive",startMark:"interactive-editor-loading",endMark:"interactive-editor-interactive"});break;case"complete":postToKuma({markName:"interactive-editor-complete",measureName:"ie-time-to-complete",startMark:"interactive-editor-loading",endMark:"interactive-editor-complete"})}});</script>
```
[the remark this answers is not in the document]
atopal: We are not collecting RUM data for whole-page-statistics. We are only doing that for the interactive-example load events that we are firing. If it's not possible to get that data, we'll have to discuss alternatives.

atopal:
> Open questions:
>    Do we want inlined versions of the other key pages (JS, HTML)?
As a first step having just the JS editor is fine. It accounts for about 70% of all interactive example page views

>   Is week vs week numbers good enough to test the hypothesis?
Yeah, we should actually be fine within days, since we expect 50k page views for the pages mentioned in [mdn/sprints#467](https://github.com/mdn/sprints/issues/467).

jwhitlock:
There's still some Python linting changes to fix.

This code only covers ``Array.forEach``. Rather than expand to the other pages mentioned on mdn/sprints#467, we should figure out the analytics question first.  There's no point in implementing this for 5 more pages if we're not collecting the data we need to decide to proceed.

hobinjk:
Yep, I'm still working on other parts of this. The main thing I figured out today is that the best metric to focus on will be js-ie-load-event-end since it's entirely the same between inline and embedded (and I'm pretty sure it's the hero metric as well). I also switched over the other performance marks but they'll be measuring the readyState of the entire document instead of only the interactive example which will cause unpredictable changes to their results.

hobinjk:
I plan to fix the python linting errors when I switch the example source to be in a separate file so that it doesn't take over the entirety of kumascript.py

kuma/wiki/kumascript.py
hobinjk: I believe my current approach is better than dynamically loading since if we're programmatically creating the script tags I think the browser will be unable to prefetch them. I did a cursory check of the waterfall view and it looks like there is no significant blocking caused by loading the JS files here since it precedes to the bottom-of-page files in parallel
editor-js js inlined doesn't block that much

hobinjk:
Okay, I think this is in a relatively good state.

@schalkneethling and/or @jwhitlock does everything look alright? The interactive-examples changes I made are on [this branch](https://github.com/hobinjk/interactive-examples/tree/manual-inline-examples) if you want to look at them

Also, @tkadlec the current plan is to reuse the js-ie-load-event-end metric as the basis for comparing between the iframe implementation and the inline implementation. Does this sound reasonable to you?

jwhitlock:
I'll look at this tomorrow. I think a "squash and merge" is appropriate, and the bug number can be added then. If there's an existing bug for this work, let me know.

hobinjk:
Thanks! I agree that squashing is correct (and changing the squashed commit to reference a bug number properly). There is no existing bug as far as I know.

jwhitlock:
Testing locally by replacing this section ([Array/forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)):

```html
<div>{{EmbedInteractiveExample("pages/js/array-foreach.html")}}</div>

<p class="hidden">The source for this interactive example is stored in a GitHub repository. If you'd like to contribute to the interactive examples project, please clone <a href="https://github.com/mdn/interactive-examples">https://github.com/mdn/interactive-examples</a> and send us a pull request.</p>
```

with:

```html
<div>@InlineArrayForEach@</div>

<p class="hidden">We're experimenting with inlining interactive examples, and measuring the performance impact. We plan to revert this change in about a week. See https://github.com/mdn/sprints/issues/467.</p>
```

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
+INLINE_EXAMPLES = {
+    "FOR_EACH": """
+<div class="interactive interactive-js">
+<link href="/static/styles/libs/interactive-examples/codemirror-5-31-0.css" rel="stylesheet" />
+<link href="/static/styles/libs/interactive-examples/editor-js.css" rel="stylesheet" />
+
+<script>"use strict";function postToKuma(e){postMessage(e,"https://developer.mozilla.org")}postToKuma({markName:"interactive-editor-loading"}),document.addEventListener("readystatechange",function(e){switch(e.target.readyState){case"interactive":postToKuma({markName:"interactive-editor-interactive",measureName:"ie-time-to-interactive",startMark:"interactive-editor-loading",endMark:"interactive-editor-interactive"});break;case"complete":postToKuma({markName:"interactive-editor-complete",measureName:"ie-time-to-complete",startMark:"interactive-editor-loading",endMark:"interactive-editor-complete"})}});</script>
```
jwhitlock: I get this error locally:

> Failed to execute ‘postMessage’ on ‘DOMWindow’: The target origin provided (‘https://developer.mozilla.org’) does not match the recipient window’s origin (‘http://localhost:8000’).

This could be changed by using ``settings.SITE_URL``. It is probably OK for this experiment.

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
+INLINE_EXAMPLES = {
+    "FOR_EACH": """
```
jwhitlock: Here's ForEach (left is inlined, right is iframed, after hitting "Run"). Only diff appears to be fonts, 👍 

screenshot 2018-11-02 10 32 59 2

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
...
+      </div>
+      <div id="output" class="output"><code></code></div>
+  </div>
+</section>
+
+<script src="/static/js/libs/interactive-examples/codemirror-5-31-0.js"></script>
+<script src="/static/js/libs/interactive-examples/editor-js.js"></script>
+</div>
+""",
+    "MAP": """
```
jwhitlock: Here's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map:

screenshot 2018-11-02 10 36 16 2

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
...
+      </div>
+      <div id="output" class="output"><code></code></div>
+  </div>
+</section>
+
+<script src="/static/js/libs/interactive-examples/codemirror-5-31-0.js"></script>
+<script src="/static/js/libs/interactive-examples/editor-js.js"></script>
+</div>
+""",
+    "FILTER": """
```
jwhitlock: Here's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

screenshot 2018-11-02 10 39 15

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
...
+      </div>
+      <div id="output" class="output"><code></code></div>
+  </div>
+</section>
+
+<script src="/static/js/libs/interactive-examples/codemirror-5-31-0.js"></script>
+<script src="/static/js/libs/interactive-examples/editor-js.js"></script>
+</div>
+""",
+    "FIND": """
```
jwhitlock: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

screenshot 2018-11-02 10 41 11 2

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
...
+      </div>
+      <div id="output" class="output"><code></code></div>
+  </div>
+</section>
+
+<script src="/static/js/libs/interactive-examples/codemirror-5-31-0.js"></script>
+<script src="/static/js/libs/interactive-examples/editor-js.js"></script>
+</div>
+""",
+    "REDUCE": """
```
jwhitlock: Here's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

screenshot 2018-11-02 10 44 32

kuma/wiki/inline_examples.py
```diff
@@ -0,0 +1,237 @@
...
+      </div>
+      <div id="output" class="output"><code></code></div>
+  </div>
+</section>
+
+<script src="/static/js/libs/interactive-examples/codemirror-5-31-0.js"></script>
+<script src="/static/js/libs/interactive-examples/editor-js.js"></script>
+</div>
+""",
+    "SPLICE": """
```
jwhitlock: Here's https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice:

screenshot 2018-11-02 10 45 59 2

jwhitlock:
Oops, I forgot to look at the performance events, back to review...

jwhitlock:
I also opened [bug 1504246](https://bugzilla.mozilla.org/show_bug.cgi?id=1504246) to use when we're ready to merge.

hobinjk:
Thanks for pointing out the secondary filter of the events! I've updated the original file accordingly but what should I do for the manually minified one? (I'm running make build-static right now in case that does it)

jwhitlock:
It's manual, @schalkneethling knows how it was done. We continue working around our asset pipeline rather than fix it.

schalkneethling:
@hobinjk Yup, it is manual. I simply use:
https://duckduckgo.com/?q=minify+javascript+online&atb=v138-1__&ia=answer

Copy & paste, Copy & paste ;)

hobinjk:
Thanks! I updated the PR accordingly although now it's failing on security/snyk. I've rebased on master and am testing locally to make sure that didn't break anything too badly

jwhitlock:
Yes, Snyk thinks there is an issue because https://github.com/mozilla/kuma/pull/5045 was merged, but this is based on a commit that includes the old node-sass version. I wish it was clever enough to see that your PR doesn't include changes to ``package.json``, so it should be OK. 

In either case, a rebase fixed it.

jwhitlock:
@hobinjk, @atopal, and I had a meeting to sync up on this. They've been working on a test plan, and we're all on the same page now.

The goal is to measure the ``js-ie-load-event-end`` as a GA event, and compare it to the same event from the iframed interactive example. I wasn't able to get this event to fire in development, because it requires the hostname to match, but I'm pretty sure it will work in production (and only production). We'll verify this on a test page, then modify the test pages. We should have sufficient data after 2-3 days of workday traffic, then disable the experiment and analyze the data.

PR #5021, opened 2018-10-03
bug 1490727: Analytic structure update and code cleanup.

joshJarr:
This PR updates the analytics structure for the contributions flow and cleans up some legacy issues.

PR is awaiting:

- [x] Full test
- [x] Confirmation on what to do with recording the customer name in Stripe's card field

### DIFF
--- a/kuma/contributions/jinja2/contributions/contribute.html
+++ b/kuma/contributions/jinja2/contributions/contribute.html
@@ -195,7 +195,7 @@
         </div>
         <div id="contribute-feedback" class="faqs clear feedback">
             <form id="faq-feedback" class="faq-feedback">
-                <textarea data-action="FAQ Any other questions" placeholder="{{_('Your feedback goes here')}}" name="faq3" id="faq3" cols="30" rows="2"></textarea>
+                <textarea data-action="FAQ Any other questions" placeholder="{{_('Your feedback goes here')}}" name="faq-feedback" id="contribution-feedback" cols="30" rows="2"></textarea>
                 <button type="submit"> {{_('Send')}} {% include 'includes/icons/arrows/arrow.svg' %}</button>
             </form>
         </div>
--- a/kuma/static/js/contribution-faq.js
+++ b/kuma/static/js/contribution-faq.js
@@ -9,7 +9,7 @@
 
     function sendAnalyticsVoteEvent(action, label ,value) {
         var event = {
-            category: 'FAQ Feedback',
+            category: 'payments',
             action: action,
             value: value,
         };
@@ -27,7 +27,7 @@
         $(event.target).next().removeClass('active');
         $(event.target).addClass('active');
 
-        sendAnalyticsVoteEvent('question_' + questionNumber, null, 1);
+        sendAnalyticsVoteEvent('FAQ vote', 'question_' + questionNumber, 1);
     }
 
     function onThumbsDown(event) {
@@ -36,20 +36,21 @@
         $(event.target).prev().removeClass('active');
         $(event.target).addClass('active');
 
-        sendAnalyticsVoteEvent('question_' + questionNumber, null, 0);
+        sendAnalyticsVoteEvent('FAQ vote', 'question_' + questionNumber, 0);
     }
 
 
     function onFeedback(event) {
         event.preventDefault();
 
-        var feedback = $(this).find('textarea').val() || '';
-        var action = $(this).find('textarea').attr('data-action') || '';
-        mdn.analytics.trackEvent({
-            category: 'Contribution feedback',
-            action: action,
-            label: feedback,
-        });
+        var feedback = $(this).find('textarea').val();
+        if (feedback) {
+            mdn.analytics.trackEvent({
+                category: 'payments',
+                action: 'FAQ - Any other questions',
+                label: feedback,
+            });
+        }
     }
 
     thumbsUp.click(onThumbsUp);
--- a/kuma/static/js/contribution-faq.js
+++ b/kuma/static/js/contribution-faq.js
@@ -43,7 +43,7 @@
     function onFeedback(event) {
         event.preventDefault();
 
-        var feedback = $(this).find('textarea').val();
+        var feedback = document.getElementById('contribution-feedback').value;
         if (feedback) {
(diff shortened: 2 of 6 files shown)


kuma/static/js/contribution-handler.js
```diff
@@ -567,8 +575,9 @@
     customAmountInput.blur(function(event) {
         // Send GA Event.
         mdn.analytics.trackEvent({
-            category: 'Contribution popover',
-            action: 'Amount manually selected',
+            category: 'payments',
+            action: 'banner',
+            label: 'custom amount',
```
jwhitlock: For this event, the user's custom value is used as the event value. However, the user is allowed to enter a decimal, such as 2.4, and an analytics event value must be an integer. I think the value should be:

```
value: Math.max(0, Math.floor(event.target.value))
````

File: kuma/static/js/contribution-handler.js [hunk 2]
```diff
@@ -359,9 +360,10 @@
 
         // Send GA Event.
         mdn.analytics.trackEvent({
-            category: 'Contribution submission',
-            action: isPopoverBanner ? 'On Popover' : 'On Page',
-            value: 1
+            category: 'payments',
+            action: 'submission',
```
jwhitlock: Note to future jwhitlock: This event is fired when the Stripe popup is shown, not when the callback is called. A user can cancel, not complete payment.

kuma/static/js/contribution-handler.js
```diff
@@ -217,8 +217,8 @@
 
     var isPopoverBanner = $('.contribution-banner').hasClass('contribution-popover');
 
-    /* If `isPopoverBanner` is false, then this is the
-       contribute page. Init the handler immediately */
+    // If `isPopoverBanner` is false then this is the contribute page.
+    // Init the handler immediately
```
schalkneethling: We follow JSDoc comment style. The standard for multi line comments that are not documenting a function is to use the following style:

```
/* If `isPopoverBanner` is false, then this is the
   contribute page. Init the handler immediately */
```

kuma/static/js/contribution-faq.js
```diff
@@ -36,20 +36,21 @@
...
         event.preventDefault();
 
-        var feedback = $(this).find('textarea').val() || '';
-        var action = $(this).find('textarea').attr('data-action') || '';
-        mdn.analytics.trackEvent({
-            category: 'Contribution feedback',
-            action: action,
-            label: feedback,
-        });
+        var feedback = $(this).find('textarea').val();
```
schalkneethling: We know the `textarea` has an id of `faq3` so, it would be more performant to do:

```
var feedback = document.getElementById('faq3').value;
```

kuma/static/js/contribution-handler.js
joshJarr: Maybe I can create an event for submission (in the token callback) and an event for closing the modal (in the close callback)?

kuma/static/js/contribution-handler.js
jwhitlock: Further playing with this, an event is fired if an invalid value is entered into the form, like ``foo``, which would result in the event ``NaN`` with my additions. I'm not sure if it is worth making the event handler bulletproof for any garbage input, but it would be fun.

kuma/static/js/contribution-handler.js
schalkneethling: I think that is a good idea. That way we can also see how many users initiate the stripe dialog but never complete payment.

kuma/static/js/contribution-handler.js
jwhitlock: ``value * 100`` works (US / UK only?), but then the radio selects should also be ``3200``, ``6400``.

kuma/static/js/contribution-handler.js
joshJarr: I did have a simple solution where I just throw an event in the checkout.js callbacks for `close` and `token` I just pushed, but I'm happy to remove it if there's too much complexity, but looking at the above solution it would be good to include that after this release?

kuma/contributions/jinja2/contributions/contribute.html
```diff
@@ -195,7 +195,7 @@ <h2>{{_('Feedback')}}</h2>
         </div>
         <div id="contribute-feedback" class="faqs clear feedback">
             <form id="faq-feedback" class="faq-feedback">
-                <textarea data-action="FAQ Any other questions" placeholder="{{_('Your feedback goes here')}}" name="faq3" id="faq3" cols="30" rows="2"></textarea>
+                <textarea data-action="FAQ Any other questions" placeholder="{{_('Your feedback goes here')}}" name="faq-feedback" id="contribution-feedback" cols="30" rows="2"></textarea>
```
schalkneethling: +1 for clearer `name` and `id` 👍

kuma/static/js/contribution-handler.js
```diff
@@ -359,9 +362,10 @@
 
         // Send GA Event.
         mdn.analytics.trackEvent({
-            category: 'Contribution submission',
-            action: isPopoverBanner ? 'On Popover' : 'On Page',
-            value: 1
+            category: 'payments',
+            action: 'submission',
+            label: isPopoverBanner ? 'On pop over' : 'On FAQ page',
+            value: selectedAmount
```
jwhitlock: This should be ``selectedAmount * 100`` for consistency with the other events with values.

kuma/static/js/contribution-handler.js
```diff
@@ -203,8 +205,15 @@
...
             description: 'One-time donation',
             token: function(token) {
+                submitted = true;
                 stripeToken.val(token.id);
                 addDisabledLocaleStorageItem();
+                // Send GA Event.
+                mdn.analytics.trackEvent({
+                    category: 'payments',
+                    action: 'submission',
+                    label: 'completed'
```
jwhitlock: Should this have a value of amount * 100?

kuma/static/js/contribution-handler.js
jwhitlock: 👍 The "completed" event worked for me

PR #5023, opened 2018-10-04
Fix Bug 1491154, switch from Verdana to Arial

schalkneethling:
This switches to Arial from Verdana. As a result I had to also make some tweaks to the typography. It is not a 1:1 match, but then it is not intended to be. Let me know your thoughts. Thanks!

### DIFF
--- a/kuma/static/styles/components/home/home-masthead.scss
+++ b/kuma/static/styles/components/home/home-masthead.scss
@@ -20,7 +20,7 @@ homepage masthead (search, feature callouts)
     }
 
     h1 {
-        @include set-font-size(36px);
+        @include set-font-size(42px);
         margin: 0 0 ($grid-spacing * 2);
         line-height: 1;
     }
--- a/kuma/static/styles/components/wiki/quick-links.scss
+++ b/kuma/static/styles/components/wiki/quick-links.scss
@@ -10,7 +10,7 @@
     margin-bottom: $grid-spacing;
     position: relative;
     overflow: hidden;
-    @include set-font-size($left-nav-font-size);
+    @include set-font-size($base-font-size);
 
     a {
         display: inline-block;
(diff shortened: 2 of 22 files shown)


kuma/static/styles/components/home/home-masthead.scss
```diff
@@ -20,7 +20,7 @@ homepage masthead (search, feature callouts)
     }
 
     h1 {
-        @include set-font-size(36px);
+        @include set-font-size(42px);
```
ExE-Boss: Shouldn’t this be using `$h2-font-size` or&nbsp;another such variable?

schalkneethling: Well, yes and no ;) I am working on the side on a more modular scale for our typography with some meaningful class/variable names. It just feels strange to me to set the `font-size` for an `h1` to `$h2-font-size`

jwhitlock:
@schalkneethling can you rebase? PR #5031 was merged first.

schalkneethling:
@jwhitlock rebased.

kuma/static/styles/components/wiki/quick-links.scss
```diff
@@ -37,7 +37,7 @@
     }
 
     li li {
-        @include bidi(((padding-left, $grid-spacing, padding-right, 0),));
+        @include bidi(((padding-left, $grid-spacing, padding-right, 0)));
```
ExE-Boss: Looks like there’s [an error in calling this](https://travis-ci.org/mozilla/kuma/jobs/442037276#L2606).

https://github.com/mozilla/kuma/blob/f088e94455be79acae6bb3a9e932c18e3338db4d/kuma/static/styles/includes/_mixins.scss#L378-L383

Based on the above, this needs the trailing comma which you removed 🤦‍🏻‍♂️

schalkneethling: Ooooh, bad rebase. Will fixup.

schalkneethling: Oh no, it was prettier that removed the `,`

PR #5024, opened 2018-10-04
Subsetting Zilla Slab

tkadlec:
This PR addresses https://github.com/mdn/sprints/issues/254.

It introduces a subsetted version of the Zilla Slab font to be used instead of the full file, cutting the font file sizes in half.


tkadlec:
@schalkneethling and @jwhitlock: This one is pretty lightweight, me thinks.

tkadlec:
Ooo....almost forgot about the file I used to subset. I mentioned it in https://github.com/mdn/sprints/issues/254:

> I created a file that contains the unicode ranges to include. There are other shortcuts to go to certain subsets, but I like the clarity this gives us in case we edit the subset in the future

I wasn't sure exactly where to put it. Right now it's in `/kuma/static/`.

kuma/static/latin-subset
```diff
@@ -0,0 +1,17 @@
...
+U+0100-017F
+
+# Latin Extended-B
+U+0180-024F
+
+# General Punctuation
+U+2000-206F
+
+# Currency Symbols
+U+20A0-20CF
```
schalkneethling: Nit: Missing empty line at end of file. This seems to be a new file? I am genuinely curious as to the purpose of this file? @tkadlec ?

schalkneethling:
@tkadlec Thanks for this, how would you suggest we go about testing this? Which pages etc. did you test? Thanks!

kuma/static/latin-subset
tkadlec: Added the new line. Also added some info to the top of the file for clarity. The file is used to generate the subset. You can see more information here: https://github.com/mdn/sprints/issues/254#issuecomment-427059741

tkadlec:
@schalkneethling Good question! I tested a handful of random pages in each locale. Here are the pages I tested:

```
/
/docs/Web/JavaScript
/docs/Mozilla/Add-ons/WebExtensions
/docs/Web/HTML
/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
```

And here are the locales:

```
Svenska (sv-SE)
Portugues (pt-PT) and (pt-BR)
Polski (pl)
Nederlands (nl)
Melayu (ms)
Taqbaylit (kab)
Italiano (it)
Bahasa Indonesia (id)
magyar (hu)
Francais (fr)
suomi (fi)
Espanol (es)
Deutsche (de)
Catala (ca)
English (en-US)
```

It's tedious, unfortunately. :/

schalkneethling:
Thanks for the additional information @tkadlec - I will take this for a test run.

schalkneethling:
@tkadlec I notice that the issue mentions subsetting ZillaSlab and OpenSans but this only address ZillaSlab. Is that intentional, perhaps to keep the work separate?

schalkneethling:
@tkadlec I see there are still opensans that looks like it is being used for https://github.com/mozilla/kuma/tree/master/kuma/static/fonts/locales/ro

tkadlec:
Ah, yeah. And that locale is being killed per
https://github.com/mdn/sprints/issues/449

jwhitlock:
No, that review works for me.

jwhitlock:
I can't find a relevant existing bug, I'll write a new one. I can add a bug # and squash on merge.

PR #5027, opened 2018-10-07
Bugfix in tag creations in the Russian language

A-Mikhail:
In Russian keyboard layout letter "б" work as a comma, that in the library used to separate tags between each other as well as tab and enter.

The bug prevents writing any words containing this letter.

### DIFF
--- a/kuma/static/js/libs/tag-it.js
+++ b/kuma/static/js/libs/tag-it.js
@@ -24,16 +24,29 @@
 *   jQuery v1.4+
 *   jQuery UI v1.8+
 */
-(function($) {
+(function ($) {
 
     $.widget('ui.tagit', {
         options: {
-            itemName          : 'item',
-            fieldName         : 'tags',
-            availableTags     : [],
-            tagSource         : null,
-            removeConfirmation: false,
-            caseSensitive     : true,
+            allowDuplicates: false,
+            itemName: 'item',
+            fieldName: 'tags',
+            placeholderText: null,   // Sets `placeholder` attr on input field.
+            readOnly: false,  // Disables editing.
+            removeConfirmation: false,  // Require confirmation to remove tags.
+            tagLimit: null,   // Max number of tags allowed (null for unlimited).
+
+            // Used for autocomplete, unless you override `autocomplete.source`.
+            availableTags: [],
+
+            // Use to override or add any options to the autocomplete widget.
+            //
+            // By default, autocomplete.source will map to availableTags,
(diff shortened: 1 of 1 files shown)


jwhitlock:
@A-Mikhail Thanks for tackling [bug 1298017](https://bugzilla.mozilla.org/show_bug.cgi?id=1298017). 

Unfortunately, this PR is a bit of a mess. It looks like your last commit did a global search and replace for "let" to "var", changing variable declarations but also "deleted" to "devared" and "autocomplete" to "autocompvare", which probably breaks it.

Our copy of `tag-it.js` was added in https://github.com/mozilla/kuma/commit/dc657b583f52267b2eb50db189614e0fbaf7130a, from the version at https://github.com/aehlke/tag-it/commit/f5fe54e4b67979da6a563ed03b9f0caae09b43c0. The tag-it maintainers continued making changes, but didn't bump the version number (this was a few changes after v2.0 was tagged).

This PR resembles the latest code from https://github.com/aehlke/tag-it, but not quite, and with a patch similar to https://github.com/aehlke/tag-it/issues/251 applied. It would be clearer if this was explicit in the commits, breaking the change into a few commits:

1. A commit from the current version in Kuma to the latest in aehlke/tag-it. It would be nice to identify the current tag-it commit in the comments, since the upstream project does a poor job of bumping version numbers.
2. A commit fixing any linting errors, to make it clear where we're deviating from upstream.
3. A commit bringing back changes from https://github.com/mozilla/kuma/commit/6280ffb0c85e2e6cd4f901ac925cde40df5ca8dd#diff-40e89d0482b1bd7797dd5f7ceecd748f and https://github.com/mozilla/kuma/commit/de10cdb4d19598dcd14f5be11dcd2078eb0bb0d5#diff-40e89d0482b1bd7797dd5f7ceecd748f, if needed, or adjusted to work with new tag-it.js features.
3. A commit fixing the issue with "б"/"," , which probably requires switching from a ``keyDown`` to ``keyPress`` handler.

I wonder if there is a more recent, maintained library that would be better for entering tags (I couldn't see a good one linked from the [django-taggit docs](https://django-taggit.readthedocs.io/en/latest/external_apps.html), or if the goal of "pretty" tags is a bad one, and we should just have a plain, comma-delimited text field.

jwhitlock:
There's still a lot of work to make this PR mergable. Let's revisit after some sidebar work in Q1 2019 - I suspect tags will be a bit different after that project.

PR #5029, opened 2018-10-09
bug 1483072: post-cutover Kuma clean-up

escattone:
This PR should not be merged until after we've cut-over to the MozIT infrastructure.
- remove the MozMEAO-related code from `Jenkinsfiles/utils.groovy`
- pushes the standard (not maintenance-mode) deployment of MDN for production
- effectively disables the MozMEAO Jenkins service from successfully pushing MDN (since `utils.get_repo_name()` will fail)
- change the default `IMAGE_PREFIX` within the `Makefile`
-  just use the default `IMAGE_PREFIX` within the `Jenkinsfile`
- update all files (docs, etc.) to reference DockerHub and the MozIT Jenkins service instead of quay.io and the MozMEAO Jenkins service
- update all relevant files to reflect that the origin of the attachments CDN has changed from `mdn-demos-origin.moz.works` to `demos-origin.mdn.mozit.cloud`
- update all relevant files to reflect that the stage attachments domain has changed from `stage-files.mdn.moz.works` to `files-stage.mdn.mozit.cloud`


jwhitlock:
@escattone would it make sense to bring in the quay.io to dockerhub changes sooner? We could merge the ``Makefile``, ``docker-compose.yml``, and documentation changes, start testing with them this week, and bring in the Jenkins changes after the cut-over and retiring the MozMEAO Jenkins server.

escattone:
@jwhitlock Thanks for taking a look in the midst of your busy schedule! I like your idea. I'll create a separate PR for that.

escattone:
I've created #5044, which does much of what this PR does, so when #5044 is merged, this PR will have to be reduced an re-based.

escattone:
I'm going to close this and create a new one.

PR #5033, opened 2018-10-10
bug 948151: Add attachment URLs to CSP img-src

jwhitlock:
Add the production attachment URL to the list of CSP allowed domains for images (``img-src``). Include the local attachment URL if it isn't production or the same as the web domain (covered by 'self').

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1143,7 +1143,9 @@ MAX_FILENAME_LENGTH = 200
 MAX_FILEPATH_LENGTH = 250
 
 _PROD_ATTACHMENT_HOST = 'mdn.mozillademos.org'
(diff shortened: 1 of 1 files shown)


escattone:
@jwhitlock I know that in IRC you mentioned that you might want to tweak this a bit more, so I'll hold off reviewing this until you do.

jwhitlock:
Force-push to rebase on master, and to avoid including the local attachment domain if it is the same as the web domain, since that would be covered by 'self'. This took more work than planned, since ``DOMAIN`` is just ``localhost``, not ``localhost:8000``, in development. I assume that's right...

In stage, this should include both https://mdn.mozillademos.org and https://stage-files.mdn.moz.works. The production attachment domain is always included, because it makes it easy as possible to copy production revisions to the local environment.

PR #5034, opened 2018-10-10
Bug 1480106: Fix mobile BCD table support background and name overflow

ExE-Boss:
See [bug&nbsp;1480106](https://bugzil.la/1480106) for details.

Follow‑up to #4948, which didn’t correctly update the **No&nbsp;support** and **Partial&nbsp;support** backgrounds.

Also fixes the <q>Samsung&nbsp;Internet&nbsp;Android</q> text overflow bug.

review?(@jwhitlock, @escattone)

## Screenshots:
<details>
<summary>Before</summary>

Kuma PR-5034 (Before)

</details>
<details>
<summary>After</summary>

Kuma PR-5034 (After)

</details>

### DIFF
--- a/kuma/static/styles/components/compat-tables/_vars.scss
+++ b/kuma/static/styles/components/compat-tables/_vars.scss
@@ -31,7 +31,7 @@ $bc-border-width-thin: 1px;
 $bc-border-width-thick: 2px;
 $bc-support: yes, no, partial, unknown;
 $bc-grid-spacing: 10px;
-$bc-mobile-icon-width: 50px;
+$bc-mobile-browser-width: 220px;
 
 $table-types: (
 	web: (
--- a/kuma/static/styles/components/compat-tables/bc-browser.scss
+++ b/kuma/static/styles/components/compat-tables/bc-browser.scss
@@ -14,7 +14,8 @@ Hide icons and show browser name in mobile view
         left: 0;
         padding-left: 6px;
         height: 25px;
-        width: 200px;
+        width: $bc-mobile-browser-width;
         color: $text-color;
+        white-space: nowrap;
     }
 }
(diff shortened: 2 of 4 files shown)


jwhitlock:
Here's what it looks like for me on a local copy of https://developer.mozilla.org/en-US/docs/Web/CSS/frequency:

screenshot 2018-10-22 17 40 59

The "X" is at least centered, but the "No" is not, so it looks bad. I think we should just drop the X when we're in the tablet / mobile layout.

It looks fine at desktop layout.

I'm also deferring to @schalkneethling or @hobinjk for front-end code.

schalkneethling:
@ExE-Boss Thank you for the PR. I agree with @jwhitlock that we should fix the misalignment between the background "X" and the text. It would be odd to have only the "X" line centered and not the rest. It is also much better from a UX perspective to have the text more closely aligned.

So, my suggestion is to remove the background "X" and then we can leave everything left aligned. Thanks!

Before:

screenshot 2018-11-05 at 13 33 41

After:

screenshot 2018-11-05 at 13 37 11

This is for mobile sized viewports only.

Elchi3:
fwiw, the feedback about the cross being unhelpful was brought up before, but only once so far: https://bugzilla.mozilla.org/show_bug.cgi?id=1452083

I would be fine with trying to remove it.

schalkneethling:
Thanks for the feedback @Elchi3

ExE-Boss:
I would personally prefer doing the removal of the × in a separate PR, as it tackles [bug&nbsp;14520843](https://bugzilla.mozilla.org/show_bug.cgi?id=1452083), whereas this fixes leftovers from [bug&nbsp;1480106](https://bugzilla.mozilla.org/show_bug.cgi?id=1480106).

schalkneethling:
Fair enough @ExE-Boss, ping me when you open that PR. Thanks!

PR #5039, opened 2018-10-17
bug 1490727: Rename /contribute to /payments

joshJarr:
Renaming the `/contribute` view to `/payments`.
Keeping file structure and paths to reflect change.
Hitting `/contribute` should redirect to `/payments`.


kuma/landing/jinja2/landing/homepage.html
```diff
@@ -142,7 +142,7 @@ <h2>{% include 'includes/icons/general/smile.svg' %} {{ _('Help improve MDN') }}
 {% block contrib_popup %}
     {%- if contribution_popup %}
         {%- with form=contribution_form, is_popover=True %}
-            {%- include "contributions/includes/contribution-form.html" %}
+            {%- include "payments/includes/payments-form.html" %}
```
jwhitlock: Another change is needed on line 156 to load the correct JavaScript.

kuma/urls.py
```diff
@@ -91,12 +93,17 @@ def serve_from_media_root(request, path):
...
 urlpatterns += [url('users/', include('kuma.users.urls'))]
 urlpatterns += i18n_patterns(
-    url(r'contribute/?$',
-        contribution_views.contribute,
-        name='contribute'),
+    url(r'^payments/$',
+        payment_views.contribute,
+        name='payments'),
+)
+urlpatterns += i18n_patterns(
```
jwhitlock: We have enough redirects that we have a whole module, ``kuma/redirects/redirects.py``, to handle them. This works, but it should eventually move there.

kuma/landing/jinja2/landing/homepage.html
joshJarr: Good spot! I've addressed this.

joshJarr:
@jwhitlock Updated the bundle names here: https://github.com/mozilla/kuma/pull/5039/commits/7b8684c191cac5c57235b2d49a23dfa8ed640563
And resolved the conflicts.

We should keep track of the other feedback here if we're not going to implement it in this PR, I'll defer to @michal-macioszczyk judgement as to if we can address it now or later in another PR.

PR #5047, opened 2018-10-24
Bug 1499455 - Hide code line numbers from screen reader.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1499455

the span element with line-numbers-rows class is added by prism, and this looks like the code.
added `aria-hidden=true`

this at least skips the line number on Safari + VoiceOver


arai-a:
to be clear, I've tested the attribute on standalone HTML page, not the JS code itself

PR #5048, opened 2018-10-24
 bug 1490727: Fix payment submission analytic event and add error event.

joshJarr:
**In this PR**
- [x] - Send the submission or error events if a user submits or encounters an error.
- [x] - The amount paid is set in pennies in the event as a value.
- [x] - The location of the submission (banner or FAQ) is also set.
- [x] - Event defaults to no data if nothing can be found.
- [x] - Event only submits once a payment has completed and not if the browser is refreshed (event should only be sent once)

**Implementation**
Payment submitted form banner:
```
eventCategory: "payments",
eventAction: "success",
eventLabel: "Banner",
eventValue: 3200,
```

Payment submitted from FAQ page
```
eventCategory: "payments",
eventAction: "success",
eventLabel: "FAQ",
eventValue: 3200,
```

Payment submission error
```
 eventCategory: 'Payment error',
 eventAction: 'Payment failed'
```

**Notes for QA**
Should be tested/monitored live to ensure events are working correctly


kuma/settings/common.py
```diff
@@ -954,8 +954,9 @@ def pipeline_one_scss(slug, **kwargs):
         'source_filenames': (
             'js/payments-handler.js',
             'js/payments-faq.js',
+            'js/payments-confirmation.js',
```
jwhitlock: Should this be part of the always-running JS, or should it be in a separate JS file that is only loaded in the ``payments/thank_you.html`` template?

kuma/static/js/payments-confirmation.js
```diff
@@ -0,0 +1,33 @@
...
+    // flow denoted by `amountSubmitted` session varible.
+    var amountSubmittedStoreKey = 'submissionDetails';
+    var submissionDetails = JSON.parse(sessionStorage.getItem(amountSubmittedStoreKey));
+
+    // Default to no data
+    var amountSubmitted = submissionDetails.amount || 0;
+    var submissionPage = submissionDetails.page || '';
+
+    var path = win.location.pathname;
+    if (path.includes('/payments/success') && submissionDetails) {
```
schalkneethling: Both the conditional statements depend on its conditions evaluating to true. Could there ever be a scenario then where the `submissionDetails` would be stored in session storage but not contain any data? i.e. line 7 above could return an empty `Object`?

I am asking because if the item is not found in `sessionStorage`, then `submissionDetails` would be null and neither of the conditionals will pass and execute. This means there is no real need for lines 9-11

@joshJarr Thoughts?

joshJarr: That's a fair point, I guess we're never really expecting a blank object, but if for some reason there is I was using the `||` to ensure `null` isn't passed to the events to prevent a runtime error or the wrong event parameters being sent.

kuma/settings/common.py
joshJarr: I think only loading this in the confirmation pages makes more sense, I'll quickly update this PR so it does that!

kuma/static/js/payments-confirmation.js
jwhitlock: Thanks @schalkneethling, I was just about to ask if you could take a look.

I'm a belt and bracers dev, so I like code that works even if someone's session is screwed up.

joshJarr: Actually, this will cause an error if there is no session storage since we'd be trying to access `null.amount`.
Looks like I should check `submissionDetails` and ensure its truthy before line 10 & 11.

joshJarr:
@jwhitlock @schalkneethling Addressed your feedback here and tested locally - all good. Let me know if there's anything else.
Thanks!

jwhitlock:
And I'm also wondering if the error state should include ``eventLabel``: ``FAQ``|``Banner`` and the amount.

joshJarr:
Good point @jwhitlock! I've updated the error event to be consistent.

kuma/static/js/payments-confirmation.js
```diff
@@ -0,0 +1,40 @@
...
+            action: 'success',
+            label: submissionPage,
+            value: amountSubmitted
+        }, function() {
+            sessionStorage.removeItem(amountSubmittedStoreKey);
+        });
+
+    } else if (path.includes('/payments/error')) {
+        mdn.analytics.trackEvent({
+            category: 'Payments',
```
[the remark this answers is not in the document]
jwhitlock: I don't think this is a nit, there's a good chance that GA will consider them distinct categories, which is a 0% change if it is changed to ``payments``.

PR #5049, opened 2018-10-24
bug 1467532: Update to flake8 3.6.0, dependencies

jwhitlock:
Our TravisCI flake8 test installs the latest, and flake8 3.6.0 was released, so flake8 builds are busted for a while. This updates the libraries and makes the code work with the latest linters.

* flake8 3.5.0 → 3.6.0: Update dependencies, small fixes
* flake8-import-order 0.17.1 → 0.18: Python 3.7 support
* pycodestyle 2.3.1 → 2.4.0: Checks break after binary operator (``W504``), invalid escape sequences in string literals (``W605``).
* pyflakes 1.6.0 → 2.0.0: Checks unused exception binding, etc.

Update the code to mark regex with ``r''``, to avoid escape sequence checks. Add ``W504`` to the ignore list, since it conflicts with ``W503`` (break before binary operator).

### DIFF
--- a/kuma/dashboards/forms.py
+++ b/kuma/dashboards/forms.py
@@ -42,11 +42,11 @@ class RevisionDashboardForm(forms.Form):
     start_date = forms.DateField(
         required=False, label=_(u'Start Date:'),
         input_formats=['%m/%d/%Y'],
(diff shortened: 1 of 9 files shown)


jwhitlock:
Force push to fix flake8 issues in the functional tests folder as well.

PR #5052, opened 2018-10-24
bug 1490727: Force en-US for some payments UI

jwhitlock:
We didn't intend to ship translations of the payments UI until next year, but it is live:

https://developer.mozilla.org/de/payments/

Rather than manually remove the translated strings, this PR adds a new tag pair ``{% translation 'en-US' %}``/``{% endtranslation %}`` that forces a locale for a block of markup. The goal is to retain translatable strings, but force English in some contexts:

* [ ] The banner (open or closed) is in English on the [homepage](http://localhost:8000/de/)
* [ ] The banner (open or closed) is in English on [wiki pages](http://localhost:8000/de/docs/Apps)
* [ ] The form is in English on the stand-alone page [/payments](http://localhost:8000/de/payments/), but the FAQ and Feedback form is localized
* [ ] Form validation errors, such as an invalid custom amount of 0.5, are displayed in English
* [ ] The [Thank You page](http://localhost:8000/de/payments/success) is in English, except for the Feedback section, which is localized
* [ ] The [Error page](http://localhost:8000/de/payments/error) is in English, except for the Feedback section, which is localized

To manually review, it is good to reproduce locally first:

* Checkout the master branch
* Checkout the current ``locale`` folder: ``git submodules update --init``
* [Build the localizations](https://kuma.readthedocs.io/en/latest/localization.html#build-the-localizations) to enable the latest translations.
* Set ``MDN_CONTRIBUTION=True``, ``STRIPE_PUBLIC_KEY``, and ``STRIPE_SECRET_KEY`` in ``.env``, ``make up`` to enable.
* If needed, log into MDN in development environment
* If needed, remove the local storage item ``contributionsPopoverDisabled``
* Confirm that the banner is in German on http://localhost:8000/de/

Once you have German locally, you can force English in the payments UI by checking out this branch, and running ``docker-compose restart web`` to force re-parsing of Jinja2 templates.

I've marked @escattone for review, but if @joshJarr wants to take a look, please do.

### DIFF
--- a/kuma/core/i18n.py
+++ b/kuma/core/i18n.py
@@ -13,6 +13,8 @@ from django.utils import lru_cache, translation
 from django.utils.translation.trans_real import (
     check_for_language, get_languages as _django_get_languages,
     language_code_prefix_re, language_code_re, parse_accept_lang_header)
+from jinja2 import nodes
+from jinja2.ext import Extension
 
 
 def django_language_code_to_kuma(lang_code):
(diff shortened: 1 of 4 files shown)


jwhitlock:
Good catch @escattone. I think links should remain in the locale. I'll think about the least tacky way to do that...

jwhitlock:
@escattone I've updated with a working solution - set the URLs outside the ``{% translation %}`` block, and use them inside. The form ``action`` URL gets the user's locale, so the follow-on pages stick in the same locale. Not pretty, but not as ugly as other solutions I considered.

PR #5053, opened 2018-10-25
Fix Bug 1496826, some icons are missing from the font-awesome bundle

schalkneethling:
This adds the missing icons mention in Bug 1496826 - https://bugzilla.mozilla.org/show_bug.cgi?id=1496826

Easiest way to test this would be to create a local copy of https://developer.mozilla.org/docs/Sandbox/Indicators

With regards to `privilegedBadge` and `translationInProgress`. You can search MDN for pages that uses these classes using:

* https://developer.mozilla.org/en-US/search?locale=*&css_classnames=translationInProgress&topic=none
* https://developer.mozilla.org/en-US/search?locale=*&css_classnames=privilegedBadge&topic=none

I did not find any pages using the latter, and the one page that does use `translationInProgress` seems to 404 when the link is clicked in the search results :/

With that said, I did make sure that the font code match what is defined in the CSS.

@hobinjk r?


ExE-Boss:
https://developer.mozilla.org/en-US/search?locale=*&css_classnames=translationInProgress&topic=none now returns 0.

This is because the page at https://developer.mozilla.org/ja/docs/Embedding_SpiderMonkey was a redirect, which sometimes causes issues with the search. This was compounded by the fact that https://developer.mozilla.org/ja/docs/SpiderMonkey/Embedding_SpiderMonkey was deleted.

I have now moved the page to https://developer.mozilla.org/ja/docs/SpiderMonkey/JSAPI_User_Guide to match https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_User_Guide, which seems to have fixed the search.

schalkneethling:
Thanks for the review @hobinjk and for the update and fix on the search @ExE-Boss - merging

PR #5059, opened 2018-10-31
bug 1490727: Add stripe customer id to MDN user

michal-macioszczyk:
As a MDN developer I want users and stripe customers to be linked so that managing user and customer becomes easier.

- [ ] MDN user was updated with new optional field called stripe_customer_id

- [ ] create two separate migrations one for timezone and second one for stripe_customer_id


kuma/users/models.py
```diff
@@ -188,6 +188,10 @@ class User(AbstractUser):
         blank=True,
         validators=[WEBSITE_VALIDATORS['discourse']],
     )
+    stripe_customer_id = models.CharField(
```
jwhitlock: Migrations are rare enough that I need to think through this, and when I think, I do it in a GitHub comment.

The deployment is done in Jenkins:

https://github.com/mozilla/kuma/blob/6472eed398cfce95be57ff18a5ce8372a8f155d1/Jenkinsfiles/push.groovy#L22-L28

1. The database is migrated while the old code is running. Django doesn't use `*`, but instead asks for each column by name. The old code won't load the new column.
2. The new code is rolled out. Users have a chance to get the new or old code. If the new code loads a user, it will include the (blank) customer ID.
3. The new code is rolled out, everything works.

PR #5062, opened 2018-10-31
Static pages for testing layout perf

tkadlec:
This adds some static pages so that we can get a really clear picture of where our heavy layout performance issues are coming from.

### DIFF
--- a/kuma/static/layout-tests/no-inline-js.html
+++ b/kuma/static/layout-tests/no-inline-js.html
@@ -10,7 +10,7 @@
   <title>Array.prototype.forEach() - JavaScript | MDN</title>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <meta name="robots" content="index, follow">
+  <meta name="robots" content="noindex, nofollow">
 
   
 <link rel="preload" href="https://developer.mozilla.org/static/fonts/locales/ZillaSlab-Regular.subset.bbc33fb47cf6.woff2" as="font" type="font/woff2" crossorigin />
--- a/kuma/static/layout-tests/no-inline-js.html
+++ b/kuma/static/layout-tests/no-inline-js.html
@@ -1,0 +1,1309 @@
+
+
+
+<!DOCTYPE html>
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
+
+  
+<link rel="preload" href="https://developer.mozilla.org/static/fonts/locales/ZillaSlab-Regular.subset.bbc33fb47cf6.woff2" as="font" type="font/woff2" crossorigin />
+    <link rel="home" href="/en-US/">
+    <link rel="license" href="#license">
+  
+
+  
+    
+    
+
+    <link href="https://developer.mozilla.org/static/build/styles/mdn.63b447789549.css" rel="stylesheet" type="text/css" />
+
+    
+  
+
+    <link href="https://developer.mozilla.org/static/build/styles/wiki.f92c28aa6cd8.css" rel="stylesheet" type="text/css" />
+
+
+    
+
+  <link href="https://developer.mozilla.org/static/build/styles/locale-en-US.d798f0544bdf.css" rel="stylesheet" type="text/css" />
+
+  
+    <!-- common social tags -->
+    
+    <meta property="og:type" content="website">
+    <meta property="og:image" content="https://developer.mozilla.org/static/img/opengraph-logo.72382e605ce3.png">
+    <meta property="og:site_name" content="MDN Web Docs">
+    <meta name="twitter:card" content="summary">
+    <meta name="twitter:image" content="https://developer.mozilla.org/static/img/opengraph-logo.72382e605ce3.png">
+    <meta name="twitter:site" content="@MozDevNet">
+    <meta name="twitter:creator" content="@MozDevNet">
+    <link rel="search" type="application/opensearchdescription+xml" href="https://developer.mozilla.org/en-US/search/xml" title="MDN Web Docs">
+    
+  
+  <!-- third-generation iPad with high-resolution Retina display: -->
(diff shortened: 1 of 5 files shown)


tkadlec:
@schalkneethling @jwhitlock I think the pages are set, but first up I need to figure out how to get those pages to be accessible when the site is build. 

It _looks_ like I need to somehow get them to be pulled in when `collectstatic` runs, but I'm not 100% sure if that's true, or how to make that happen.

tkadlec:
@schalkneethling I SWEAR I tried that URL before and it didn't work! Huh. 

Ok. I think we're all set then to merge these in.

kuma/static/layout-tests/no-inline-js.html
```diff
@@ -0,0 +1,1438 @@
...
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <script>(function(d) { d.className = d.className.replace(/\bno-js/, ''); })(document.documentElement);</script>
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
+
+  
+<link rel="preload" href="https://developer.mozilla.org/static/fonts/locales/ZillaSlab-Regular.subset.bbc33fb47cf6.woff2" as="font" type="font/woff2" crossorigin />
```
jwhitlock: There's a risk that the hashed versions of these assets will not be available. We don't keep old versions of assets, but the CDN probably does. I'd say this is a low risk, but look for it in speedcurve. The same name without the hash will give the latest asset (``https://developer.mozilla.org/static/fonts/locales/ZillaSlab-Regular.subset.woff2``) with a 7-day cache timeout.

tkadlec: Do you recommend dropping the has before deploying?

jwhitlock: I'm not sure. Will the 7-day cache cause issues in speedcurve (continue using hashed, update if it breaks), or does it always request fresh (use unhashed and get latest)?

kuma/static/layout-tests/no-inline-js.html
```diff
@@ -0,0 +1,1332 @@
...
+<!DOCTYPE html>
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
```
jwhitlock: Let's tell robots to ignore this duplicate content.

```suggestion
  <meta name="robots" content="noindex, nofollow">
```

kuma/static/layout-tests/no-js.html
```diff
@@ -0,0 +1,1309 @@
...
+<!DOCTYPE html>
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
```
jwhitlock: Let's tell robots to ignore this duplicate content.

```suggestion
  <meta name="robots" content="noindex, nofollow">
```

kuma/static/layout-tests/normal.html
```diff
@@ -0,0 +1,1460 @@
...
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <script>(function(d) { d.className = d.className.replace(/\bno-js/, ''); })(document.documentElement);</script>
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
```
jwhitlock: Let's tell robots to ignore this duplicate content.

```suggestion
  <meta name="robots" content="noindex, nofollow">
```

kuma/static/layout-tests/raw-html.html
```diff
@@ -0,0 +1,1286 @@
...
+<!DOCTYPE html>
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
```
jwhitlock: Let's tell robots to ignore this duplicate content.

```suggestion
  <meta name="robots" content="noindex, nofollow">
```

kuma/static/layout-tests/unused-css.html
```diff
@@ -0,0 +1,1452 @@
...
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <script>(function(d) { d.className = d.className.replace(/\bno-js/, ''); })(document.documentElement);</script>
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="index, follow">
```
jwhitlock: Let's tell robots to ignore this duplicate content.

```suggestion
  <meta name="robots" content="noindex, nofollow">
```

jwhitlock:
Oh, and we need a bug number as well. @tkadlec, do you know if there is an existing performance bug in Bugzilla, or do we need a new one?

tkadlec:
@jwhitlock I think we need a new one. It exists in sprints, but nowhere else that I know of.

kuma/static/layout-tests/no-inline-js.html
tkadlec: Good catch.

tkadlec:
Changes made, using GH's beta "apply suggestion" thing which was pretty slick, other than needing separate commits for everything.

PR #5069, opened 2018-11-02
Bug 1493499, add aria roles to notes/warning boxes inside ckeditor

schalkneethling:
To test this, edit a page on MDN. In CKEditor, choose the "Note" component from the `Styles` dropdown. Switch to the source output, confirm that the following is output:

```
<div class="note" role="complementary">
<p>&nbsp;</p>
</div>
```

Switch back to the WYSIWYG and add a "Warning" component. Switch to source and confirm that the output is as follows:

```
<div aria-label="Important note" class="warning" role="region">
<p>&nbsp;</p>
</div>
```

r?


ExE-Boss:
This now needs rebasing on top of&nbsp;https://github.com/mozilla/kuma/pull/5079 (https://github.com/mozilla/kuma/commit/b6baa5c4f39cc91bc87aa32e6b6d5cba08d8fb19)

PR #5071, opened 2018-11-02
bug 1467518: Migration from Python 2 to Python2/3

MatonAnthony:
Address the remaining low hanging fruits tests on the migration from Python 2.7 to a compatible Python 2.7/3.6+ compatible codebase.

### DIFF
--- a/kuma/core/utils.py
+++ b/kuma/core/utils.py
@@ -51,7 +51,7 @@ def is_untrusted(request):
 
 def paginate(request, queryset, per_page=20):
     """Get a Paginator, abstracting some common paging actions."""
-    paginator = Paginator(queryset, per_page)
+    paginator = Paginator(list(queryset), per_page)
 
     # Get the page from the request, make sure it's an int.
     try:
--- a/kuma/core/utils.py
+++ b/kuma/core/utils.py
@@ -20,6 +20,7 @@ from django.utils.http import urlencode
 from django.utils.translation import ugettext_lazy as _
 from polib import pofile
 from pytz import timezone
+from six.moves import xrange
 from six.moves.urllib.parse import parse_qsl, ParseResult, urlparse, urlsplit, urlunsplit
 from taggit.utils import split_strip
 
--- a/kuma/search/tests/test_views.py
+++ b/kuma/search/tests/test_views.py
@@ -19,9 +19,9 @@ class ViewTests(ElasticTestCase):
         response = self.client.get('/en-US/search?q=article')
         assert response.status_code == 200
         assert_no_cache_header(response)
-        assert 'Results for' in response.content.decode()
-        assert 'an article title' in response.content.decode()
-        assert '4 documents found for "article" in English' in response.content.decode()
+        assert 'Results for' in response.content.decode('utf-8')
+        assert 'an article title' in response.content.decode('utf-8')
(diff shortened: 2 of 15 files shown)


kuma/core/utils.py
```diff
@@ -20,6 +20,7 @@
 from django.utils.translation import ugettext_lazy as _
 from polib import pofile
 from pytz import timezone
+from six.moves import xrange
```
jwhitlock: Interesting fix!

kuma/wiki/tests/test_views_document.py
```diff
@@ -167,7 +167,7 @@ def test_api_safe(client, section_doc, section_case, if_none_match, method):
                 str(section_doc.current_revision_id))
 
     if method == 'GET':
-        assert response.content == exp_content
+        assert response.content == exp_content.encode('utf-8')
```
MatonAnthony:
```suggestion
        assert response.content == exp_content.decode('utf-8')
```

kuma/search/tests/test_views.py
```diff
@@ -19,9 +19,9 @@ def test_search_filters(self):
         response = self.client.get('/en-US/search?q=article')
         assert response.status_code == 200
         assert_no_cache_header(response)
-        assert 'Results for' in response.content
-        assert 'an article title' in response.content
-        assert '4 documents found for "article" in English' in response.content
+        assert 'Results for' in response.content.decode('utf-8')
```
jwhitlock: nit: I've been using this pattern other places, which feels less repetitive and less magical to me:

```
content = response.content.decode(response.charset)
assert 'Results for' in content
assert 'an article title' in content
assert '4 documents found for "article" in English' in content
```

kuma/wiki/tests/test_feeds.py
```diff
@@ -467,7 +467,7 @@ def test_feeds_update_after_doc_tag_change(client, wiki_user, root_doc):
         response = client.get(reverse('wiki.feeds.recent_documents',
                                       args=['atom', tag]), follow=True)
         assert response.status_code == 200
-        assert root_doc.title not in response.content
+        assert root_doc.title not in response.content.decode()
```
jwhitlock: nit: Don't omit the encoding in general. The docs say this uses the "default string encoding" when omitted.  ``sys.getdefaultencoding()`` returns ``ascii`` for me. Probably OK for the test, if there are no non-ASCII characters in the response, but not in general.

kuma/wiki/tests/test_views_document.py
jwhitlock: I'd expect ``response.content`` to be a UTF-8 encoded binary string, and ``exp_content`` to be a binary string in Python 2 and a unicode string in Python 3. I think you need to change line 158 to:

```
exp_content = b''
```
or, better yet, get rid of the variable and change this line to:

```suggestion
        assert response.content == b''
```

MatonAnthony:
I will fix the nits first, then we can merge so we keep a small PR.

MatonAnthony:
- The failing Python 2.7 Travis job is failing due to a GPG key related issue, it probably just needs a rerun.
- The `security/snyk` job is failing, my understanding is that snyk is only checking for dependencies in `package.json`, I didn't modified the frontend, so it is probably unrelated. (I cannot really check because I don't have access to the project on snyk)

PR #5082, opened 2018-11-08
bug 1490727: Include confirmation checkbox for recurring payments.

joshJarr:
This PR adds a checkbox to confirm recurring payment submission.

**In this PR**
- https://trello.com/c/dIsVyv2F/44-as-a-user-i-can-agree-to-the-payment-terms-so-that-i-can-confirm-my-recurring-payment

**Notes for QA**
- [ ] checkbox should be focusable
- [ ] checkbox should be toggle-able
- [ ] form should not submit until checkbox is active
- [ ] checkbox is required

**screenshots**

screenshot 2018-11-07 at 18 36 06
screenshot 2018-11-07 at 18 36 01

### DIFF
--- a/kuma/payments/forms.py
+++ b/kuma/payments/forms.py
@@ -213,20 +213,20 @@ User email: {email}""".format(**{
 
 
 class RecurringPaymentForm(ContributionForm):
+    accept_checkbox = forms.BooleanField(
+        label=u'',
+        required=True,
+        widget=forms.CheckboxInput(
+            attrs={
+                'class': 'required checkbox form-control',
+                'data-error-message': _('Must be more than $1')
+            },
+        ),
+    )
 
     def __init__(self, *args, **kwargs):
         super(RecurringPaymentForm, self).__init__(*args, **kwargs)
         self.fields['donation_choices'].choices = RECURRING_PAYMENT__CHOICES
-        self.fields['accept_checkbox'] = forms.BooleanField(
-            label=u'',
-            required=True,
-            widget=forms.CheckboxInput(
-                attrs={
-                    'class': 'required checkbox form-control',
-                    'data-error-message': _('Must be more than $1')
-                },
-            ),
-        )
 
     @staticmethod
     def create_customer(email, token, user, name):
--- a/kuma/payments/forms.py
+++ b/kuma/payments/forms.py
@@ -217,6 +217,16 @@ class RecurringPaymentForm(ContributionForm):
     def __init__(self, *args, **kwargs):
         super(RecurringPaymentForm, self).__init__(*args, **kwargs)
         self.fields['donation_choices'].choices = RECURRING_PAYMENT__CHOICES
+        self.fields['accept_checkbox'] = forms.BooleanField(
+            label=u'',
+            required=True,
+            widget=forms.CheckboxInput(
+                attrs={
+                    'class': 'required checkbox form-control',
+                    'data-error-message': _('Must be more than $1')
+                },
+            ),
+        )
 
     @staticmethod
     def create_customer(email, token, user, name):
--- a/kuma/payments/forms.py
+++ b/kuma/payments/forms.py
@@ -219,7 +219,7 @@ class RecurringPaymentForm(ContributionForm):
         widget=forms.CheckboxInput(
             attrs={
                 'class': 'required checkbox form-control',
-                'data-error-message': _('Must be more than $1')
+                'data-error-message': _('You must agree to the terms to continue')
             },
         ),
     )
--- a/kuma/payments/jinja2/payments/includes/payments-form.html
+++ b/kuma/payments/jinja2/payments/includes/payments-form.html
@@ -32,7 +32,7 @@
                 Payments are not tax deductible. Your payment details will be processed by
                 <a href="https://stripe.com/" rel="noopener" target="_blank">Stripe</a>.
                 You can read our Payment Terms
-                <a href="{{ terms_url }}" rel="noopener" target="_blank">Payment terms</a>,
+                <a href="{{ terms_url }}" target="_blank">Payment terms</a>,
                 for more information.
             {% endtrans %}
         </p>
--- a/kuma/payments/jinja2/payments/includes/payments-form.html
+++ b/kuma/payments/jinja2/payments/includes/payments-form.html
@@ -40,12 +40,13 @@
 
         {% if recurring_payment %}
             <input type="hidden" id="stripe_source_setup" value="1">
-            <div>
+            <label class="recurring-payment-confirmation">
                 {{form.accept_checkbox}}
+                <div class="control-indicator"></div>
                 <span class="legal-copy">
                     By clicking this button, I authorize Mozilla to charge this payment method $16 each month, according to the <a>Payment Terms</a>, until I cancel my subscription.
                 </span>
-            </div>
+            </label>
         {% endif %}
     </div>
     <ul class="errorlist hidden" id="contribution-error-message">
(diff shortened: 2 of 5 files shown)


kuma/payments/forms.py
```diff
@@ -213,6 +213,16 @@ def make_charge(self):
 
 
 class RecurringPaymentForm(ContributionForm):
+    accept_checkbox = forms.BooleanField(
+        label=u'',
+        required=True,
+        widget=forms.CheckboxInput(
+            attrs={
+                'class': 'required checkbox form-control',
+                'data-error-message': _('Must be more than $1')
```
jwhitlock: This message should be changed to something like ``_('You must agree to the terms to continue')``. I think it is impossible without developer shenanigans to get the message, however.

kuma/payments/jinja2/payments/includes/payments-form.html
```diff
@@ -26,20 +26,32 @@
...
+
+        {% if recurring_payment %}
+            <input type="hidden" id="stripe_source_setup" value="1">
+            <label class="recurring-payment-confirmation">
+                {{form.accept_checkbox}}
+                <div class="control-indicator"></div>
+                <span class="legal-copy">
+                    {% trans amount='16' %}
+                        By clicking this button, I authorize Mozilla to charge
+                        this payment method $<span id="amount">{{ amount }}</span>
```
jwhitlock: I like the idea of the ``<span>``, but I expected the amount to update as the buttons changed or a custom amount was entered.

kuma/payments/jinja2/payments/includes/payments-form.html
```diff
@@ -26,20 +26,32 @@
...
             {% endtrans %}
         </p>
+
+        {% if recurring_payment %}
+            <input type="hidden" id="stripe_source_setup" value="1">
+            <label class="recurring-payment-confirmation">
+                {{form.accept_checkbox}}
+                <div class="control-indicator"></div>
+                <span class="legal-copy">
+                    {% trans amount='16' %}
```
jwhitlock: According to @atopal, we shouldn't translate legal copy, even with the "force English" hack.

kuma/payments/jinja2/payments/includes/payments-form.html
```diff
@@ -26,20 +26,32 @@
             {{form.stripe_public_key}}
         {% endif %}
 
-        {% if recurring_payment %}
-            <input type="hidden" id="stripe_source_setup" value="1">
-        {% endif %}
 
         <p class="legal-copy">
-            {% trans %}
+            {% trans terms_url=url('payment_terms') %}
```
jwhitlock: This may be a good time to remove the translation block from this legal copy as well (https://trello.com/c/X78QBZWn/55-bug-legal-copy-should-be-english-only)

kuma/payments/forms.py
joshJarr: 👍 fixed, and now displaying errors on the form.

kuma/payments/jinja2/payments/includes/payments-form.html
joshJarr: Good spot! Completely overlooked this, I've fixed this now.

kuma/payments/jinja2/payments/includes/payments-form.html
```diff
@@ -1,4 +1,9 @@
-<form id="contribute-form" class="contribution-form" method="{% if recurring_payment and not request.user.is_authenticated %}get{% else %}post{% endif %}" action="{% if recurring_payment %}{{ recurring_payment_subscription_url }}{% else %}{{ payments_url }}{% endif %}" novalidate>
+<form id="contribute-form"
+    class="contribution-form"
+    method="{% if recurring_payment and not request.user.is_authenticated %}get{% else %}post{% endif %}"
+    action="{% if recurring_payment %}{{ recurring_payment_subscription_url }}{% else %}{{ payments_url }}{% endif %}"
+    data-payment-type="{% if recurring_payment %}recurring{% else %}one-time{% endif %}"
+    novalidate>
```
joshJarr: I find all this logic confusing, would like to refactor this at a later stage. I've created a task in the backlog to look at this.

kuma/static/js/payments-handler.js
```diff
@@ -231,6 +233,12 @@
                 setFieldError(customAmountInput);
             }
 
+            if (isRecurringPayment && recuringConfirmationCheckbox[0].checkValidity()) {
+                clearFieldError(recuringConfirmationCheckbox[0]);
+            } else {
+                setFieldError(recuringConfirmationCheckbox[0]);
```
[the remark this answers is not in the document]
joshJarr: Good point, I'll check with our designer here, but just some spacing make this look much better
screenshot 2018-11-09 at 12 51 51

kuma/payments/jinja2/payments/includes/payments-form.html
```diff
@@ -28,13 +33,11 @@
...
-                Payments are not tax deductible. Your payment details will be processed by
-                <a href="https://stripe.com/" rel="noopener" target="_blank">Stripe</a>.
-                You can read our Payment Terms
-                <a href="{{ terms_url }}" target="_blank">Payment terms</a>,
-                for more information.
-            {% endtrans %}
+            Payments are not tax deductible. Your payment details will be processed by
+            <a href="https://stripe.com/" rel="noopener" target="_blank">Stripe</a>.
+            You can read our Payment Terms
+            <a href="{{ url('payment_terms') }}" rel="noopener" target="_blank">Payment terms</a>,
```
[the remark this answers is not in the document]
joshJarr: Ahh! Great spot

joshJarr:
@jwhitlock @schalkneethling I've updated this PR, now the validation for the checkbox appears and the amount in the legal copy updates upon new amount being selected. 
Thanks for the review. Sorry I missed these 2 things!

Thanks!

schalkneethling:
Looking good @joshJarr. Just one more suggestion ;)

Can we give these two paragraphs a bit more breathing room between each other, and make sure they line up on the left?

screenshot 2018-11-09 at 16 07 03

I know the "footer" portion seems like it wants to be centre aligned but, at the moment the alignments looks a bit jagged. If we want to keep a centre alignment, perhaps we can set a smaller width on the second paragraph so it is clearly centred?

Let me know your thoughts. Thanks!

joshJarr:
@schalkneethling I've increased the spacing from 12px to 18px, gives it a little more room and makes allot of difference!
Centre alignment of the footer buttons are to make it easier for the footer, 
screenshot 2018-11-09 at 16 57 35
left aligned causes an issue since there would be a load of empty space to the right.
And since the form animates upwards, moving the button on expand looks a little strange.

Let me know your thoughts, I didn't quite understand the solution to reduce the size of the second paragraph, did you mean doing something like so: 
screenshot 2018-11-09 at 17 04 22

kuma/payments/jinja2/payments/includes/payments-form.html
jwhitlock: The new layout makes it less confusing than it was before, nice change.

kuma/payments/tests/test_views.py
```diff
@@ -163,3 +166,21 @@ def test_template_render_not_logged_in_recurring_payment_initial_view(mock_enabl
     assert '<input type="email"' in response.content
     assert 'id="id_donation_amount"' in response.content
     assert 'id="id_donation_choices"' in response.content
+
+
+@pytest.mark.django_db
+@mock.patch('kuma.payments.views.enabled')
+def test_template_render_not_logged_in_payment_view(mock_enabled, client, settings):
+    """If enabled, contribution error page is returned."""
```
jwhitlock: +1 for a docstring, but -2 for a mismatch between the string and the test

jwhitlock: I've addressed this and other issues in https://github.com/mozilla/kuma/pull/5084

PR #5084, opened 2018-11-09
bug 1490727: Fix payments nits

jwhitlock:
* 🇺🇸 Americanize "programme" before sending strings to translators.
* Update test docstrings, which had been duplicated.
* Use the one-line mock for enabling or disabling payments (thanks @michal-macioszczyk!)


jwhitlock:
I've got one more string change.

PR #5086, opened 2018-11-10
bug 1506363: improve query in Revision.get_previous

escattone:
This PR addresses the problem of an expensive query run by the `Revision.get_previous` method. When we looked into why we were exhausting our MySQL connections (see [bug 1506363](https://bugzilla.mozilla.org/show_bug.cgi?id=1506363)), we discovered many (on the order of hundreds) of long-lived connections stuck in the "sending data" state for many hours. The query being run by many of these problematic connections was the query within `Revision.get_previous`, which is used heavily for example by the `https://developer.mozilla.org/en-US/dashboards/revisions` endpoint.

The problem with the original query is that the MySQL query optimizer sees both the index on the `created` column as well as the index on the `document_id` column as options, and unfortunately selects the index on the `created` column. The difference between the two choices in terms of rows that need to be examined is several orders of magnitude (e.g., roughly 200,000 rows versus roughly 200 rows).

This PR changes the query such that the index on the `document_id` column is used without having to provide a specific index hint (for that we'd have to provide the specific index name, in this case `wiki_revision_document_id_3dad7483977be7a4_fk_wiki_document_id`, which would probably change if we changed databases, say to PostgreSQL). Here's the output from the `explain` for the new query approach:
```
MySQL [developer_mozilla_org]> explain SELECT * from (SELECT * FROM wiki_revision WHERE document_id = 971) revs WHERE (revs.is_approved = 1) AND (revs.created < '2013-03-22 10:29:45') ORDER BY revs.created DESC LIMIT 1;
+----+-------------+---------------+------+----------------------------------------------------------------+----------------------------------------------------------------+---------+-------+------+-----------------------------+
| id | select_type | table         | type | possible_keys                                                  | key                                                            | key_len | ref   | rows | Extra                       |
+----+-------------+---------------+------+----------------------------------------------------------------+----------------------------------------------------------------+---------+-------+------+-----------------------------+
|  1 | PRIMARY     | <derived2>    | ALL  | NULL                                                           | NULL                                                           | NULL    | NULL  |  267 | Using where; Using filesort |
|  2 | DERIVED     | wiki_revision | ref  | wiki_revision_document_id_3dad7483977be7a4_fk_wiki_document_id | wiki_revision_document_id_3dad7483977be7a4_fk_wiki_document_id | 4       | const |  267 | NULL                        |
+----+-------------+---------------+------+----------------------------------------------------------------+----------------------------------------------------------------+---------+-------+------+-----------------------------+
2 rows in set (0.00 sec)
```

versus the output from `explain` for the original approach:

```
MySQL [developer_mozilla_org]> explain SELECT `wiki_revision`.`id`, `wiki_revision`.`document_id`, `wiki_revision`.`title`, `wiki_revision`.`slug`, `wiki_revision`.`summary`, `wiki_revision`.`content`, `wiki_revision`.`tidied_content`, `wiki_revision`.`keywords`, `wiki_revision`.`tags`, `wiki_revision`.`toc_depth`, `wiki_revision`.`render_max_age`, `wiki_revision`.`created`, `wiki_revision`.`comment`, `wiki_revision`.`creator_id`, `wiki_revision`.`is_approved`, `wiki_revision`.`based_on_id`, `wiki_revision`.`is_mindtouch_migration` FROM `wiki_revision` WHERE (`wiki_revision`.`document_id` = 971 AND `wiki_revision`.`created` < '2013-03-22 10:29:45' AND `wiki_revision`.`is_approved` = 1) ORDER BY `wiki_revision`.`created` DESC LIMIT 1;
+----+-------------+---------------+-------+--------------------------------------------------------------------------------------------------------------+------------------------+---------+------+--------+------------------------------------+
| id | select_type | table         | type  | possible_keys                                                                                                | key                    | key_len | ref  | rows   | Extra                              |
+----+-------------+---------------+-------+--------------------------------------------------------------------------------------------------------------+------------------------+---------+------+--------+------------------------------------+
|  1 | SIMPLE      | wiki_revision | range | wiki_revision_e2fa5388,wiki_revision_9ab2607b,wiki_revision_document_id_3dad7483977be7a4_fk_wiki_document_id | wiki_revision_e2fa5388 | 8       | NULL | 197961 | Using index condition; Using where |
+----+-------------+---------------+-------+--------------------------------------------------------------------------------------------------------------+------------------------+---------+------+--------+------------------------------------+
1 row in set (0.00 sec)
```

This is perhaps just a stop-gap approach. A better approach may be to add an additional column to the Revision table which points directly (foreign key) to the previous revision. However, that approach is more involved and requires changes to the database, and there may be better things to spend our time on. It is also worth noting that most uses of `Revision.get _previous` are via the `Revision.previous` cached property, so that reduces the number of times this is called.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -1714,11 +1714,31 @@ class Revision(models.Model):
         """
         Returns the previous approved revision or None.
         """
+        # If we use the following more natural Django ORM usage:
+        #
+        #     self.document.revisions.filter(
+        #         is_approved=True,
+        #         created__lt=self.created,
+        #     ).order_by('-created')[0]
+        #
+        # the problem arises where the MySQL query optimizer considers
+        # using the index on the "created" column as well as the index
+        # on the "document_id" column, and unfortunately selects the
+        # index on the "created" column. The difference between the two
+        # choices is several orders of magnitude (e.g., roughly 200,000
+        # rows versus roughly 200 rows). The following raw query is a
+        # way to force the choice of the "document_id" index.
+        sql = """
+            SELECT * FROM (
+                SELECT * FROM {rev_table} WHERE document_id = %(document_id)s
+            ) rev
+            WHERE (rev.is_approved = 1) AND (rev.created < %(created)s)
+            ORDER BY rev.created DESC
+            LIMIT 1
+        """.format(rev_table=Revision._meta.db_table)
+        params = {'document_id': self.document.id, 'created': self.created}
         try:
-            return self.document.revisions.filter(
-                is_approved=True,
-                created__lt=self.created,
-            ).order_by('-created')[0]
+            return Revision.objects.raw(sql, params)[0]
         except IndexError:
             return self.based_on
 


jwhitlock:
Thanks for looking into this, @escattone. I think things are happier after the ``ANALYZE TABLE``. When we migrate data centers in 2019, we'll have to remember to do that.

PR #5089, opened 2018-11-11
bug 1467520: bug 1467532: Resolve RemovedInDjango20Warnings

MatonAnthony:
To upgrade Kuma to Django 2.0, we need to address multiple warnings.

- 15971ffd7a8915cec67b3138a1424930472dc27b fix the issues within our code.
- f5b781577b56119c434264d0415783c24e7d263f updates django-allauth to a version which is Django 2.0 ready.
- 377ab2d5a72bb2e3e43eeb6d9a5c589693d5d790 address an issue with the Docker build and introduce `PYTHONWARNINGS=all` for documentation purposes.

# Next steps:
- [ ] Update `django-ratelimit` which is responsible for most of the remaining warnings, there is currently no release compatible with Django 2.0 (see: https://github.com/jsocol/django-ratelimit/issues/146 however the master tree is compatible with Django 2.0 according to: https://github.com/jsocol/django-ratelimit/pull/136/files)
- [ ] Find a bug number for this PR or open a bug and reference it for this PR

# Remaining warnings:
After the `django-ratelimit` update, the remaining warnings are coming from `django-tidings` but will disappear automatically when we update to Django 2.0 (see: https://github.com/mozilla/django-tidings/blob/master/tidings/compat.py)

# Personal opinion:
I am not sure it is worth waiting for a `django-ratelimit` release to merge this, but I believe we should take contact with the module maintainer to see if a release can be made.

### DIFF
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -8,6 +8,8 @@ ENV NODE_VERSION=8.11.3 \
     PIP_DISABLE_PIP_VERSION_CHECK=1 \
     # disable this when preparing for Django upgrade
     PYTHONWARNINGS=ignore \
+    # enable this when preparing for Django upgrade
+    # PYTHONWARNINGS=all \
     # Kuma Pipeline definitions
     PIPELINE_CSS_COMPRESSOR=kuma.core.pipeline.cleancss.CleanCSSCompressor \
     PIPELINE_CLEANCSS_BINARY=/usr/local/bin/cleancss \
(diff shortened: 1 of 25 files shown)


jwhitlock:
Please use [bug 1467520](https://bugzilla.mozilla.org/show_bug.cgi?id=1467520).

MatonAnthony:
Sorry it took a while to fix the few nits.

I have no hard feelings about the Dockerfile change, but I didn't saw it documented anywhere and I felt like the Dockerfile itself was an appropriate place.

So if you want it gone, I can remove the commit.

PR #5093, opened 2018-11-12
Fix Bug 1505486, reduce top margin for inline secureContext icon

schalkneethling:
This reduces the top margin and overall sizing of the secure context icon when used inline.

Page to test: https://developer.mozilla.org/en-US/docs/Web/API/PaymentResponse

Before:

screenshot 2018-11-12 at 14 09 06

After:

screenshot 2018-11-12 at 14 26 45

@hobinjk r?


schalkneethling:
@jwhitlock Yup, these needs a bit of work so we can get a consistent look across these various inlined widgets. On my to-do list ;) Thanks for the review.

PR #5098, opened 2018-11-12
Add static test of inlined example plus unused css

hobinjk:
@atopal I think this is the combination we want to test for maximum impact. I can also add inlined example versions of the other static layout tests

### DIFF
--- a/kuma/static/layout-tests/iex-unused-css.html
+++ b/kuma/static/layout-tests/iex-unused-css.html
@@ -1,0 +1,1490 @@
+
+
+
+<!DOCTYPE html>
+<html lang="en-US" dir="ltr" class="no-js">
+<head prefix="og: http://ogp.me/ns#">
+  <meta charset="utf-8">
+
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <script>(function(d) { d.className = d.className.replace(/\bno-js/, ''); })(document.documentElement);</script>
+  <title>Array.prototype.forEach() - JavaScript | MDN</title>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="robots" content="noindex, nofollow">
+
+
+<link rel="preload" href="https://developer.mozilla.org/static/fonts/locales/ZillaSlab-Regular.subset.bbc33fb47cf6.woff2" as="font" type="font/woff2" crossorigin />
+    <link rel="home" href="/en-US/">
+    <link rel="license" href="#license">
+
+
+
+
+    <style>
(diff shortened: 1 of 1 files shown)


jwhitlock:
This looks decent, here's some quick feedback.

Here's this example at http://localhost:8000/static/layout-tests/iex-unused-css.html, versus @tkadlec's at http://localhost:8000/static/layout-tests/unused-css.html:

screenshot 2018-11-13 13 22 56

The code lacks syntax highlighting, the line numbering isn't as styled, and the code isn't padded but instead is printed on top of the line numbers.

In both examples, the "Section" text for the section anchor links is visible. 

Possibly related, both have a console error for ``mdn.assets.css["wiki-compat-tables"]``. This code is used to load the styles (and the JS) for the compatibility table just in time. This was a performance optimization, like syntax highlighting, that should probably get a second look. Some changes may eliminate the JS error and allow further just-in-time JS to run.

I'll need more time to look into the performance events, although I suspect they are not important for this test, since the static pages won't be shown at the same URLs.

hobinjk:
@tkadlec Which metrics are you collecting for these static tests? I'm asking because I think that if you're doing the default speedcurve ones the minor visual issues shouldn't impact the final values (which will save me from redoing the css inlining and unused filtering)

atopal:
I don't think this worked, the interactive example is not interactive. I only see the fall back text.

hobinjk:
Oh, that's unfortunate. The referenced assets were removed in my PR which removed inlined interactive example but I didn't notice that locally since their minified versions weren't cleaned out. I'll make a PR to get the assets in their in a way which fits with the optimizations currently

atopal:
hmm, now the interactive example does work, but the js-ie-load-event-end measure is missing, so we aren't getting any data for our main KPI. Also, wiki.js and some fonts are still 404ing. I guess we can live with the font issue, but the missing wiki.js means that we can't compare directly with the production version

atopal:
Never mind, the 404ing of wiki.js and the fonts was an artifact of my testing. The missing performance measure however is real.

hobinjk:
I dug into this a bit more locally and as far as I can tell this version should produce the js-ie-load-event-end through the standard postMessage mechanism

atopal:
Sorry, my fault, this isn't on production yet.

jwhitlock:
PR #5134 updates this example by updating the sample. However, these are served as static files with a 7 day cache timeout, so you'll probably continue to get the old version for a week.

Some solutions:

* Find the hashed URL for the file, and use that for testing. See https://bugzilla.mozilla.org/show_bug.cgi?id=1503923#c7 for how, and this one is https://developer.mozilla.org/static/layout-tests/iex-unused-css.3405659d6a7e.html
* Change the filename with each change, so that the cache is skipped
* Ask me or @escattone to invalidate the cache. I've done this.

atopal:
Ah, thanks for the heads up, John. And thanks for the cache invalidation. Unfortunately I still can't see the js-ie-load-event-end measure :(

PR #5100, opened 2018-11-13
Fix Bug 1504320, replace grid tools banner with future of design tools

schalkneethling:
This replaces the current grid tools banner with the future of design tools. I uploaded screen shots at various viewport sizes and asked Victoria for feedback.

https://bugzilla.mozilla.org/show_bug.cgi?id=1504320#c15

I also took the opportunity to separate the various banners' SaSS and cleaned up `home.html` as well. Looks like there definitely is some additional CSS we can get rid of for old banners we no longer have, and that will not be coming back anytime soon.

We should clean this up in a separate PR.

@hobinjk @jwhitlock r?


schalkneethling:
Ignore `dashboard.scss` it is not meant to be in this PR ;) Will update shortly.

schalkneethling:
Thanks for review @hobinjk - I addressed two small items from Victoria, so we are all good to merge this one. 🎉

PR #5101, opened 2018-11-13
bug 1490727: include recurring payments on banner

joshJarr:
This PR adds a the recurring payment form to the popover banner. This form is dynamic and should behave like a toggle switch between recurring and one-time payments.
Worth noting the FAQ page will not be dynamic, and the toggle should still link to separate views.
If recurring payments are enabled, the monthly recurring form should be active by default.

**In this PR**
- https://trello.com/c/To2Lk1I7/48-include-recurring-payment-on-the-popover

**notes for QA**
- [x] all users should be able to see and expand the banner on desktop browser
- [x] the banner should link to the payment page on smaller browsers and tablets
- [x] all users should be able to toggle between recurring payments and one-time payments
- [x] toggling between payments should be represented in the UI change
    - [x] the corresponding toggle button should change
    - [x] the payment bands should change and expand
    - [x] the recurring confirmation checkbox should show/hide
        - [x] if not checked this checkbox should show validation on recurring payments
        - [x] the checkbox should be clickable
- [x] both one-time and recurring payments form should expand correctly and show the entire form on the banner
- [x] validation should work on the banner
- [ ] form data should be preserved when toggling the payment type (recurring/one-time)
- [x] form should be pre-filled with the users data if the user is logged in 
- [x] anonymous users wishing to complete a recurring payment should be directed to login after completing the form
- [x] once logging in the user should be redirected back to the FAQ page with the form data pre-filled
- [x] the two payment forms on the FAQ page should be navigable by URL
- [x] FAQ payment form navigation may not preserve user input upon switching between form type (recurring/one-time)
- [x] banner should not be visually obstructed by the window or any other content

**Screenshots**
screenshot 2018-11-13 at 14 33 42
screenshot 2018-11-13 at 14 33 35
screenshot 2018-11-13 at 14 33 28
screenshot 2018-11-13 at 14 33 15
screenshot 2018-11-13 at 14 32 58


jwhitlock:
Thanks @joshJarr, I ran through the QA steps, and it looks good, with a few things that jumped out.

1. On Firefox, there's a waning cresent moon when the One off payment button is selected. It doesn't appear on the stand-alone FAQ page. It is a visual bug that doesn't impact usage.

  pr 5101 - firefox - waning crescent

2. On all browsers, the space between then checkbox invalid message and the checkbox text is tight. It appears to be an issue when the browser is narrow enough that the checkbox message goes to 3 lines.

  pr 5101 - tight validation

3. As we mentioned in the meeting, clicking on the "Payment Terms" link in the checkbox label toggles the checkbox rather than opens the terms page.

4. This sequence has an unexpected effect:

    - Select One-off payment
    - Select $128 then $32 plan. The selector shows "$32" and the button switches to "Pay $32"
    - Select Monthly payment. The selector is "$16/mo", the checkbox text says "$32 each month", and the action button says "Pay $32".

    pr 5101 - switching

Do you think these should be fixed in this PR or a future change?

joshJarr:
👍 Pushed a couple changes that should fix all 4 issues, let me know if there's anything else, thanks.

jwhitlock:
Yes, the four issues appear addressed. Here's some new behavior:

1. The "Payments Terms" link in the checkbox label works, but clicking the label no longer toggles the checkbox. Is this expected? Should the checklist above be changed? The QA checklist?
2. The selectors appear to be bound together between the One Off and Monthly Payment forms. When I first expand the banner, the middle item, "$32/mo" is selected. However, the checkbox text says "$64 each month", and the button "Pay $64". The checkbox text and button are only confused on the initial load. If I click "Pay $64", Stripe seems to think it is $64 not $32.

    pr 5101 - 32 or 64

    If I switch back and forth, everything syncs as expected. If I select the third option on the one-time page ($128), then the third option is selected on the Monthly page ($64/mo), and so on for the other options. I think it is OK for the selections to sync when switching, and it would be less jarring if the values were the same (16 / 32 / 64 on both one time and recurring).

    It doesn't happen on the standalone page http://localhost:8000/en-US/payments/recurring .

joshJarr:
Thanks @jwhitlock. Pushed 2 fixes: 
The first updates the button text on form load.
The seconds prevents form selection persisting between form change - now when switching between forms the selection will be lost (to force a user to explicitly re-select). Note custom amount input will be persisted.

schalkneethling:
@joshJarr - When the page initially loads the banner has the following state:

screenshot 2018-11-14 at 15 14 00

✅ 

I think click on "Contribute" and the banner expands:

screenshot 2018-11-14 at 15 14 06

✅ 

When I then click on "One of payment" it does switch, but I get an "error" message below the custom amount:

screenshot 2018-11-14 at 15 14 13

kuma/payments/jinja2/payments/includes/payments-form.html
```diff
@@ -30,29 +32,36 @@
...
+        {% set show_checkbox = recurring_payment or (is_popover and recurring_payment_enabled) %}
+        <div id="recurring-confirmation-container" {% if not show_checkbox %}class="hidden"{% endif %}>
+            <div class="recurring-payment-confirmation">
+                <label for="{{form.accept_checkbox.id_for_label}}">
+                    <span class="offscreen">
+                        By clicking this button, I authorize Mozilla to charge
+                        this payment method <span data-dynamic-amount>$64</span>
+                        each month, according to the <a href="{{ url('payment_terms') }}">Payment Terms</a>, until
+                        I cancel my subscription.
+                    </span>
```
schalkneethling: Is there a reason we are duplicating this copy? By marking this as offscreen, clicking on the "label" will also no longer cause the checkbox to be checked. It would be better for accessibility to keep this functionality.

If we cannot get this to work, we should probably mark this as `aria-hidden="true"`, which should prevent screen readers from reading the legal copy twice.

schalkneethling:
If I have expanded the banner, switched to "One off payment", which triggers the "error" message mentioned before, collapse the banner, and click refresh, I get the following in the console:

```
TypeError: event.target is undefined; can't access its "value" property
```
Which points to this line in `onAmountSelect`:

```
// Validate against minimum value.
if (parseInt(event.target.value) < 1 || isNaN(event.target.value)) {
```

This error then prevents the banner from expanding.

schalkneethling:
I am also seeing this frequently in Firefox, but it might just be because I am running on localhost:

```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”)
```

schalkneethling:
@joshJarr - At around 1023px viewport width on `/en-US/payments/recurring` the Pay button changes to "Support MDN" and clicking the button simply takes you to the same URL:

screenshot 2018-11-14 at 15 46 06

schalkneethling:
The same happens on http://localhost:8000/en-US/payments/

schalkneethling:
* [x] - all users should be able to see and expand the banner on desktop browser
* [x] - the banner should link to the payment page on smaller browsers and tablets
* [x] - all users should be able to toggle between recurring payments and one-time payments
* [x] - toggling between payments should be represented in the UI change
* [x] - validation should work on the banner (except for the one area, mentioned earlier, where validation seems to happen to soon)

schalkneethling:
* [-] - both one-time and recurring payments form should expand correctly and show the entire form on the banner

I does, but if I resize the browser down, the height is not adjusted and the banner area is not scrollable. Instead the page behind it scrolls.

banner-ex-col

kuma/payments/jinja2/payments/includes/payments-form.html
joshJarr: We had a bug: clicking the link in the label would trigger the checkbox and not follow the href.
To solve this I moved the legal copy outside the label.
In order to ensure screenreaders read the copy with the context of the checkbox I've visually hidden the copy.

I think marking the copy outside of the label with `aria-hidden="true"` makes sense! I'll do that.

schalkneethling:
* [x] - validation should work on the banner
* [x] - form data should be preserved when toggling the payment type (recurring/one-time)
* [x] - form should be pre-filled with the users data if the user is logged in
* [x] - anonymous users wishing to complete a recurring payment should be directed to login after completing the form
* [x] - once logging in the user should be redirected back to the FAQ page with the form data pre-filled
* [x] - the two payment forms on the FAQ page should be navigable by URL
* [x] - FAQ payment form navigation may not preserve user input upon switching between form type (recurring/one-time) ~ This works consistently for me when logged in
* [-] - banner should not be visually obstructed by the window or any other content

joshJarr:
@schalkneethling Thanks for the review! Great spot with these issues. I've pushed a couple of fixes to this branch, let me know if there's anything else.

I couldn't replicate this error `TypeError: event.target is undefined; can't access its "value" property` but I've pushed a fix that _should_ solve it - so let me know if thats good.

I've implemented your accessibility suggestion for the payment terms checkbox label too, hope that's all good.

Cheers.

jwhitlock:
I'm getting the ``TypeError: event.target is undefined; can't access its "value" property`` in Firefox 64 (Dev Edition) as well. I need to investigate more to see why.

Another issue appears in Firefox, Chrome, and Safari. When I first expand the banner (tablet width?), "$32/mo" is selected, everything looks good. 

screenshot 2018-11-14 11 01 38

When I switch to One off payment, nothing is selected, the button is just "Pay":

screenshot 2018-11-14 10 59 32

Switching back to Monthly, nothing is selected, and neither the checkbox or the Pay button have a value:

screenshot 2018-11-14 10 59 32

@joshJarr is this expected?

joshJarr:
Yeah that's expected.
I didn't want to pre select the option encase the user thinks clicking 'monthly' payments will use their selected option
(IE user clicks $32, clicks 'monthly' and their option has not changed to the default $64 payment)
I've decided to not pre select an option when changing tabs.
However the custom amount input persists.

I'll investigate the JS console error now 👍🏻

jwhitlock:
@joshJarr thanks, I'll test again with that behavior in mind. I'm inclined to merge with the known Firefox bug if everything else looks solid.

PR #5118, opened 2018-11-19
bug 1490727: Round floating point amounts to 2 decimal places

joshJarr:
This PR simply ensures we represent pennies to the correct place, ie 1.2 to  $1.20

**in this PR**
- https://trello.com/c/Kw80DkXl/58-bug-fractional-custom-amounts-should-be-displayed-as-000

**Notes for QA**
- [ ] A custom amount like 2.5 should show as "Pay $2.50" on the button, not "Pay $2.5".
- [ ] Ensure this is also represented on the stripe overlay and the overlay and button are the same amount
- [ ] Ensure in the legal copy for recurring payments this is displayed to 2 decimal places too.

### DIFF
--- a/kuma/static/js/payments-handler.js
+++ b/kuma/static/js/payments-handler.js
@@ -227,7 +227,7 @@
                 triggerOneTimePaymentEvent({
                     action: 'banner',
                     label: 'Amount radio selected',
-                    value: event.target.value * 100
+                    value: (selectedAmount * 100).toFixed(0)
                 });
             }
 
@@ -238,7 +238,7 @@
             form.find('input[type=\'radio\']:checked').prop('checked', false);
         }
 
-        selectedAmount = event.target.value % 1 === 0 ? event.target.value : parseFloat(event.target.value).toFixed(2);
+        selectedAmount = event.target.value % 1 === 0 ? parseInt(event.target.value) : parseFloat(event.target.value).toFixed(2);
         var newValue = (selectedAmount < 1 || isNaN(selectedAmount)) ? '' : '$' + selectedAmount;
         amountToUpdate.html(newValue);
 
@@ -326,12 +326,12 @@
         currrentPaymentForm === 'recurring'
             ? triggerRecurringPaymentEvent({
                 action: 'Form completed',
-                value: (selectedAmount * 100).toFixed(2)
+                value: (selectedAmount * 100).toFixed(0)
             }, true)
             : triggerOneTimePaymentEvent({
                 action: 'submission',
                 label: isPopoverBanner ? 'On pop over' : 'On FAQ page',
-                value: (selectedAmount * 100).toFixed(2)
+                value:  (selectedAmount * 100).toFixed(0)
             });
 
         if (requestUserLogin && currrentPaymentForm === 'recurring') {
@@ -349,7 +349,7 @@
                 zipCode: true,
                 allowRememberMe: false,
                 panelLabel: currrentPaymentForm === 'recurring' ? 'Pay {{amount}}/month' : 'Pay',
-                amount: (selectedAmount * 100).toFixed(2),
+                amount: parseInt((selectedAmount * 100).toFixed(0)),
                 email: $(emailField).val(),
                 closed: function() {
                     // Send GA Event.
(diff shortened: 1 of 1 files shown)


schalkneethling:
- [-] A custom amount like 2.5 should show as "Pay $2.50" on the button, not "Pay $2.5".

Stumbled on a strange rounding error. For some reason 9.7(and seemingly only 9.7) rounds to 9.69 :/

- [x] Ensure this is also represented on the stripe overlay and the overlay and button are the same amount
- [x] Ensure in the legal copy for recurring payments this is displayed to 2 decimal places too.

jwhitlock:
> Stumbled on a strange rounding error. For some reason 9.7(and seemingly only 9.7) rounds to 9.69 :/

That is an issue with floating point numbers, and will happen with other numbers. [What Every JavaScript Developer Should Know About Floating Points](https://modernweb.com/what-every-javascript-developer-should-know-about-floating-points/) is a good article on the issue, but maybe a bit math heavy.

I believe [toFixed(2)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) is what you want.

schalkneethling:
@jwhitlock `toFixed(2)` is what is currently there though :/

joshJarr:
Fix incoming: just found some issue with validation that I'm addressing.

jwhitlock:
> @jwhitlock toFixed(2) is what is currently there though :/

🤦‍♂️ sorry I didn't read the code in detail. Time for coffee...

joshJarr:
@jwhitlock @schalkneethling Should be fixed!

jwhitlock:
@joshJarr The button and the custom value now look really good, with the possible exception of a user adding an unnecessary 0, like ``3.0``:

pr 5118 3.0

I was staring at the analytics, and there still some rounding issues. For FF 64 (in case it is browser dependent), a custom value of '2.3' resulted in an event value of `229` rather than the expected `230`. The UI shows ``$2.30``, and the analytics is only off by a penny, and who is donating $2.30? 

Still, the analytics code appears to be doing something to change a float to an integer, which gets confused for decimals that don't map well to floats. Maybe ``(selectedAmount * 100).toFixed()`` would be better than ``toFixed(2)`` for events?

Here's my console JS to explore the numbers:

```
for (x=0; x < 100; x +=1) {
  value = 0.1 * x;
  console.log((value * 100).toFixed(),
              (value * 100).toFixed(2),
              value);
}
```

Some of the output:

```
360 360.00 3.6
370 370.00 3.7
380 380.00 3.8000000000000003
390 390.00 3.9000000000000004
400 400.00 4
410 410.00 4.1000000000000005
420 420.00 4.2
430 430.00 4.3
```

The danger seems to be when the last digit of the float representation is a 5, like ``4.1``.

joshJarr:
Should be good, check https://github.com/mozilla/kuma/pull/5118/commits/2404d5b7f8568c64f0fe99c0a95ed69f0c7711e4

I fear this is another example of 'wack-a-mole' fixes. I would've really liked to have some sort of unit test in place for the onAmountSelected function to ensure all side effects are accounted for and don't break in later iterations.

jwhitlock:
Thanks, @schalkneethling or I will take a look.

> I fear this is another example of 'wack-a-mole' fixes. I would've really liked to have some sort of unit test in place for the onAmountSelected function to ensure all side effects are accounted for and don't break in later iterations.

Currently, we just have Selenium for front-end testing, and the effort we put into keeping it working is probably not worth the benefit. Browser-based testing also seems like the wrong solution for algorithm testing. If this was Django / Python, I'd test the function in a unit test. 

Do you have a suggestion for a unit testing framework for JS code? @schalkneethling or @hobinjk or @davidflanagan may have ideas as well.

schalkneethling:
Entering 3.0 now shows "Pay $3/month" on button and $3.00 in the Stripe modal
Entering 9.7 now correctly shows $9.70 on the pay button and Stripe modal
2.9 Works as expected as does for example 4.1

I did not test the analytics, but seeing the representation in UI is correct, and I am guessing the GA bits uses the same `toFixed`, we should be good to go. @jwhitlock You happy with that, or do you want to also give this a spin?

If all is good, merge at will ;) 🎉

jwhitlock:
Analytics looks good for 4.0 to 5.5. No lost pennies!

PR #5120, opened 2018-11-19
Update locale submodule

jwhitlock:


jwhitlock:
I'm not sure why TravisCI isn't running against this PR. Running the locale refresh process locally to find issues.

PR #5125, opened 2018-11-21
bug 1490727: Fix validation for first name input, adjust custom amount input validation.

joshJarr:
Been working on solving some validation issues for the form fields. Primary issue was the users name field not being filled, but I've also made a small adjustment to the custom input field while here to enforce it to accept numerical values.

** in this PR ** 
- https://trello.com/c/rxWp9MoR/88-add-field-validation-on-your-full-name

** notes for qa **
- [ ] adding a white space should no longer be considered valid for the users name input field.
- [ ] whitespace surrounding the name should be removed. 
        ie `       josh` should change to `josh` upon focusing out of the input. 
- [ ] The custom amount input should now only accept numbers.
- [ ] custom amount input should increment by 0.01 when adjusting


joshJarr:
👍 Addressed, I'm now updating the amount on blur. Thanks!

PR #5128, opened 2018-11-22
WIP bug 1365987: Add ./manage.py rerender_docs

jwhitlock:
Convert my [Django shell script](https://gist.github.com/jwhitlock/43e34e07bef8c3f1863e91f076778ca6) to a command ``./manage.py rerender_docs``, to re-render documents that use a macro without overwhelming the celery queues. My goals are to enable running this in the cluster rather than from my laptop, and eventually to run as an async task with reporting. This may also be the seed for a periodic (nightly, weekly) rendering task.

To do:
* Add tests, documentation

Stretch goals:
* Move re-render state into a Redis key
* Switch from printing status to polling state
* Move to asynchronous execution with Celery
* Create a final report
* Optionally email the final report


jwhitlock:
De-prioritized, I'll re-open if / when it gets resurrected.

PR #5135, opened 2018-11-26
bug 1487403: upgrade cleancss to version 4.x

davidflanagan:
This patch updates package.json to upgrade from clean-css 3.x
to clean-css-cli 4.x. It makes other modifications to add and
correctly process the command-line arguments that are required
for the new version of cleancss.

Test plan:
- spot check the files in static/build/styles to make sure they
  have not changed too dramatically. In particular, pay attention
  to relative paths for images and fonts.
- run locally to verify that there is no obvious CSS breakage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
         - ES_VERSION=5.6.10
         - ES_DOWNLOAD_URL=https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.6.10.tar.gz
         - PIPELINE_CLEANCSS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/cleancss
-        - PIPELINE_CLEANCSS_ARGUMENTS="-O2 --skip-rebase"
+        - PIPELINE_CLEANCSS_ARGUMENTS="-O1 --skip-rebase"
         - PIPELINE_CSS_COMPRESSOR=kuma.core.pipeline.cleancss.CleanCSSCompressor
         - PIPELINE_JS_COMPRESSOR=pipeline.compressors.uglifyjs.UglifyJSCompressor
         - PIPELINE_SASS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/node-sass
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
         - ES_VERSION=5.6.10
         - ES_DOWNLOAD_URL=https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.6.10.tar.gz
         - PIPELINE_CLEANCSS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/cleancss
+        - PIPELINE_CLEANCSS_ARGUMENTS="-O2 --skip-rebase"
         - PIPELINE_CSS_COMPRESSOR=kuma.core.pipeline.cleancss.CleanCSSCompressor
         - PIPELINE_JS_COMPRESSOR=pipeline.compressors.uglifyjs.UglifyJSCompressor
         - PIPELINE_SASS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/node-sass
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
         - ES_VERSION=5.6.10
         - ES_DOWNLOAD_URL=https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.6.10.tar.gz
         - PIPELINE_CLEANCSS_BINARY=$TRAVIS_BUILD_DIR/node_modules/.bin/cleancss
+        - PIPELINE_CLEANCSS_ARGUMENTS="-O2 --skip-rebase"
```
davidflanagan: I don't know what uses travis.yml, so I don't know how to test that this works. In particular I don't know whether the double quotes here are the right thing.

kuma/core/pipeline/cleancss.py
```diff
@@ -6,5 +6,7 @@ class CleanCSSCompressor(SubProcessCompressor):
     def compress_css(self, css):
         binary = settings.PIPELINE.get('CLEANCSS_BINARY', 'cleancss')
         args = settings.PIPELINE.get('CLEANCSS_ARGUMENTS', '')
-        command = (binary, args) if args else (binary,)
+        # If the arguments ever include quoted arguments with spaces then
+        # the simple split() call here is not going to be good enough.
+        command = (binary,) + tuple(args.split())
```
davidflanagan: The existing code worked for a single argument but was not written to work with more than one.

.travis.yml
davidflanagan: Well, the travis ran automatically when I created the PR, so presumably that means that this worked.

davidflanagan:
@jwhitlock I think I updated the package-lock file with npmrefresh. I don't really understand those files, but I never see a small diff in a package lock file, so I gather that they have non-deterministic ordering of the entries, unfortunately.

I had noticed that the new version was removing redundant font sizes and combining styles where possible. But I had not noticed the color changes you point out. Devtools indicate that our dashboard tools are using the rgba version of that border color, so I'm surprised to see that that is the one that is being removed rather than the hex code. Maybe CSS specifies that borders are never actually translucent, and in that case maybe the two colors are actually equivalent. If not, then I'm worried that we will have visual changes...

davidflanagan:
So that is totally a bug in cleancss. I'm updating this PR to add another command-line argument to work around it.

jwhitlock:
This one looks better. cleancss 3.x looks like it was a little more efficient. Here's a rule from wiki.css:

```css
.reviews{
    background:#fff3d4;
    border-color:#f6b73c;
    color:#333
}
/* Some space */
.quick-links,.reviews{
    font-size:16px;
    overflow:hidden
}
/* Some space */
.reviews{
    box-sizing:border-box;
    max-width:42rem;
    margin-bottom:20px;
    padding:10px;
    font-size:.88889rem;
    border-left-width:5px;
    border-left-style:solid
}
```

The same rule from cleancss 4.x at -01:

```css
.reviews{
    background:#fff3d4;
    border-color:#f6b73c;
    color:#333
}
/* Some space */
.reviews{
    box-sizing:border-box;
    max-width:42rem;
    overflow:hidden;
    margin-bottom:20px;
    padding:10px;
    font-size:16px;
    font-size:.88889rem
}
.reviews{
    border-left-width:5px;
    border-left-style:solid
}
```

The rules are the same but in a different order.  Maybe that ``-O2`` can do better when it is working.

I looked at some pages locally, and they appear the same:

* https://developer.mozilla.org/en-US/
* https://developer.mozilla.org/en-US/docs/Sandbox/Indicators
* https://developer.mozilla.org/en-US/docs/Sandbox/All-Indicators
* https://developer.mozilla.org/en-US/docs/User:stephaniehobson:indicators

This looks ready to me. And our CSS could use some cleanup.

jwhitlock:
@schalkneethling said he'd like to take a look as well. I've duplicated this branch to https://github.com/mozilla/kuma/tree/cleancss-1487403 (to build a docker image) and force-pushed to ``stage-push``, so now it is deployed to https://developer.allizom.org.  Hopefully this makes the review a bit easier.

@schalkneethling, feel free to merge if you feel it is ready.

schalkneethling:
> This looks ready to me. And our CSS could use some cleanup.

Oh boy can it ever ;)

Poked around on staging, all looks good. Merging, Thx @davidflanagan 🌮

PR #5137, opened 2018-11-26
bug 1505084: update the kuma_base Docker image to use Node 10.14

davidflanagan:
Test Plan:
- compare the static/build directory before and after this change
  to ensure that none of the built assets change with the upgrade.

### DIFF
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:2.7-slim
 
 # Set the environment variables
-ENV NODE_VERSION=8.11.3 \
+ENV NODE_VERSION=10.14.2 \
     # extra python env
     PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1 \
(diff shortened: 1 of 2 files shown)


davidflanagan:
@escattone: Since I didn't actually copy updated commands from the node 10 Dockerfile (because I didn't want to risk mis-applying the changes that we applied to the node 8 Dockerfile), I updated the comment in a different way than you suggested. Are you okay with it as it stands now?

jwhitlock:
We try to keep in sync with the node Dockerfile changes, because there are many more eyes on it than our version. There's not a lot of changes, I think we should apply some of them and update the reference URL like @escattone suggested.

They added a note that we should add, ``# gpg keys listed at https://github.com/nodejs/node#release-keys``

It looks like they used ``--batch`` instead of ``--no-tty``, we should try that as well. The [docs](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html) make them look like similar options, but ``--batch`` turns off interactive stuff while ``--no-tty`` also turns off some printing.

There's some whitespace changes in the big multi-line ``RUN`` command, we can apply or skip those.

@escattone, I skip the ``yarn`` stuff because we don't use it. Its another gpg key that can fail, a slight increase to the docker image sizes, and it appears that ``npm`` is adopting most of the ``yarn`` innovations.

davidflanagan:
10.14.0 is just out today, so I should probably redo this and the other two PRs to use that instead of 10.13.0

PR #5138, opened 2018-11-27
bug 1510020: In absolutify, use settings.SITE_URL instead of Site.get_current()

jwhitlock:
``absolutify`` is a Jinja2 filter and utility function that turns paths into absolute URLs. It currently uses ``Site.objects.get_current()`` to get the domain name, and assumes ``https`` is the protocol. This means tests assume the site prefix is ``https://example.com``, and require a database or to mock the ``get_current`` method. Also, when used locally, the link is ``https://localhost:8000``, and broken due to the ``https``. It also has some unneeded flexibility, to allow using a different ``Site`` object which is unused outside of tests.

Maybe for these reasons, lots of code now has custom methods for creating absolute URLs. I'd like to get back to One Way of doing this common operation.

This PR updates ``absolutify`` to use ``settings.SITE_URL``, which is ``https://localhost:8000`` in development, ``https://developer.127.0.0.1.nip.io`` when using the SSL config, ``https://developer.allizom.org`` in staging, and ``https://developer.mozilla.org`` in production.

There's a few test changes:
* All the remaining tests in ``wiki/tests/test_helpers.py``, including the ones for ``absolutify``, are converted to pytest-style.
* Some tests could be modified to pass before the change, and continue passing after the change.
* Other tests needed to be changed at the same time as the switch to ``settings.SITE_URL``.

### DIFF
--- a/kuma/wiki/tests/test_helpers.py
+++ b/kuma/wiki/tests/test_helpers.py
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
 from datetime import datetime
 
-import mock
 import pytest
-from django.contrib.sites.models import Site
 from django.template import TemplateDoesNotExist
 from pyquery import PyQuery as pq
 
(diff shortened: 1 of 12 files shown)


kuma/wiki/tests/test_helpers.py
```diff
@@ -1,43 +1,46 @@
...
-
-    def test_tojson(self):
-        assert (tojson({'title': '<script>alert("Hi!")</script>'}) ==
-                '{"title": "&lt;script&gt;alert(&quot;Hi!&quot;)&lt;/script&gt;"}')
+def test_tojson():
+    """tojson converts dicts to JSON objects with escaping."""
+    output = tojson({'title': '<script>alert("Hi!")</script>'})
+    expected = ('{"title": "&lt;script&gt;alert(&quot;Hi!&quot;)'
+                '&lt;/script&gt;"}')
+    assert output == expected
```
escattone: Thanks for converting all of these to pytest-style! ⭐️

PR #5139, opened 2018-11-27
bug 1490727: Add cancelation ability while deleting User through Django signals

michal-macioszczyk:
This is PR for:
https://trello.com/c/eqa0Zh7e/20-as-an-mdn-admin-i-want-to-be-prevented-from-deleting-a-subscribers-mdn-account-so-that-i-can-stop-their-subscription-first

Now when User is deleted all his subscriptions will be canceled.

Just for the record, there are two functions from other work which is not merged yet but is useful for this functionality (kuma/payments/utils.py): 
- cancel_stripe_customer_subscription
- get_stripe_customer_data
Code is the same so there shouldn't be a problem with merging it.

### DIFF
--- a/kuma/payments/utils.py
+++ b/kuma/payments/utils.py
@@ -1,7 +1,13 @@
+import stripe
+import logging
 from django.conf import settings
 from waffle import flag_is_active
 
 from .constants import CONTRIBUTION_BETA_FLAG, RECURRING_PAYMENT_BETA_FLAG
+
+stripe.api_key = settings.STRIPE_SECRET_KEY
+
+log = logging.getLogger('kuma.payments.utils')
 
 
 def enabled(request):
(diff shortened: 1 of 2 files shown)


kuma/payments/utils.py
```diff
@@ -20,3 +26,53 @@ def recurring_payment_enabled(request):
...
+    except stripe.error.InvalidRequestError as e:
+        log.error(
+            'Stripe subscription cancellation: Invalid parameters were supplied to Stripe API: {} [{}] {}'.format(
+                username,
+                email,
+                e
+            )
+        )
+    except Exception as e:
+        log.error(
```
safwanrahman: I think it would be better if you log with `log.exception`

kuma/payments/utils.py
```diff
@@ -20,3 +26,53 @@ def recurring_payment_enabled(request):
     """Returns True if recurring payment is enabled for the user."""
     return (popup_enabled(request) and
             flag_is_active(request, RECURRING_PAYMENT_BETA_FLAG))
+
+
+def get_stripe_customer_data(stripe_customer_id, email, username):
```
safwanrahman: Where this function is getting used?

michal-macioszczyk: This is shared part from https://github.com/potatolondon/kuma/pull/52 which is in internal review right now.

jwhitlock:
The feedback so far is on functions that were merged as part of PR #5146. After a rebase, the code is much smaller. I'm not able to push to this repo, so I re-opened as #5199.

PR #5140, opened 2018-11-27
bug 1467518: Add --no-tty option to py3 Dockerfile

jwhitlock:
Copy the ``--no-tty`` fix from PR #5089 (81059de7f8e0f80ffe2f04c537a7a3c7c317ad92).

This image can be built with ``make build-base-py3``, and is used for local Python 3 testing.


jwhitlock:
I think we'll resync after PR #5137 instead.

PR #5141, opened 2018-11-27
Bug 1505084: update the KumaScript Dockerfile to install Node 10.13.0

davidflanagan:
https://github.com/mdn/kumascript/pull/1008 updates KumaScript to
use Node 10 instead of Node 8. This patch updates the KumaScript
Dockerfile in the Kuma repo to install Node 10 in the KumaScript image.

### DIFF
--- a/docker/images/kumascript/Dockerfile
+++ b/docker/images/kumascript/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.10.0-slim
+FROM node:10.14.2-slim
 
 RUN set -ex && \
     apt-get update && \


docker/images/kumascript/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.10.0-slim
+FROM node:10.14.2-slim
```
jwhitlock: I'd support ``node:10-slim``, to always get the latest minor version of node 10, but I'm not sure how often node breaks things in minor releases.

PR #5142, opened 2018-11-28
Bug 1423716: stop adding the SEO root title to all document titles

davidflanagan:
We currently constuct the <title> of a document like this:

  `<title> - <SEO root document title> | MDN`

So for https://developer.mozilla.org/en-US/docs/Learn/CSS we have:

>  Learn to style HTML using CSS - Learn web development | MDN

This patch shortens the title by removing the SEO root title, leaving:

> Learn to style HTML using CSS | MDN

### DIFF
--- a/kuma/wiki/tests/test_views.py
+++ b/kuma/wiki/tests/test_views.py
@@ -32,7 +32,6 @@ from ..events import EditDocumentEvent, EditDocumentInTreeEvent
 from ..forms import MIDAIR_COLLISION
 from ..models import Document, RevisionIP
 from ..templatetags.jinja_helpers import get_compare_url
-from ..views.document import _get_seo_parent_title
 
 
 class RedirectTests(UserTestCase, WikiTestCase):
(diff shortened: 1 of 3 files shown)


davidflanagan:
screen shot 2018-11-27 at 5 15 05 pm

davidflanagan:
So it looks like I need to learn to run the tests locally before I submit a PR...

kuma/wiki/tests/test_views.py
```diff
@@ -562,13 +561,11 @@ def test_dont_render_previews_for_deferred_docs(self, mock_post):
...
-        doc = document(save=True)
-        slug_dict = {'seo_root': 'Root/Does/Not/Exist'}
-        _get_seo_parent_title(doc, slug_dict, 'bn-BD')  # Should not raise Http404
-
-    def test_seo_title(self):
+    # NOTE(djf): In the past, we included the title of the "SEO Root"
+    # document in the title of every page, and this test tested for it.
+    # That feature has been removed now, and now this test verifies that
+    # the SEO root title does *not* appear in document titles.
+    def test_no_seo_title(self):
```
escattone: For me, I think something like `test_no_seo_root_in_title` would be clearer, but this is fine too.

PR #5144, opened 2018-11-28
Bug 1503923, additional page variations for performance experiment

schalkneethling:
Adds additional page variations as requested in the doc:

With inlined/unused CSS and iframed interactive examples
* Without sidebar elements
* Without contributor list
* Without compat data table (and no JS/CSS loaded conditionally)

@hobinjk r?


schalkneethling:
@hobinjk Do not review this yet. I misunderstood @atopal. Will update the PR with the correct variations and ping you when ready.

schalkneethling:
@hobinjk Ok, this is ready for review. Thanks!

schalkneethling:
Thanks @hobinjk will take a quick look.

schalkneethling:
@hobinjk updated. Prettier was not being its usual helpful self, but instead messed up the minified code ;(

schalkneethling:
> hopefully not js-ie-event-load-end

I do not believe so as some of these do not have an IEx at all. Thanks for the review @hobinjk

atopal:
> > hopefully not js-ie-event-load-end
> 
> I do not believe so as some of these do not have an IEx at all. Thanks for the review @hobinjk

For those that don't have an IEx, we don't need those numbers, but we do want them for the pages that have an IEx.

PR #5146, opened 2018-11-29
bug 1490727: Manage recurring payments - cancel button.

joshJarr:
Implementing a new feature to allow customers to cancel their recurring subscription on their profile page. This simply calls the Stripe API to cancel the current payment plan.

**in this PR**
https://trello.com/c/XuSGCflk/33-as-a-user-i-want-to-be-able-to-cancel-my-subscription-from-my-profile-page

**images**
screenshot 2018-11-27 at 10 17 15
screenshot 2018-11-27 at 18 09 02
screenshot 2018-11-27 at 18 09 14

**Notes for QA**
- [ ] Ensure subscription link appears if the recurring_payments feature is enabled.
- [ ] If the user has a subscription ensure the information (card details/amount) is correct.
- [ ] Ensure styling copy and links on the payment management page work and display correctly
- [ ] Ensure cancelling a subscription works and returns a success message


jinja2/includes/icons/general/money-bill-alt.svg
```diff
@@ -0,0 +1,3 @@
+<svg class="icon icon-money-bill" xmlns="http://www.w3.org/2000/svg" width="23px" height="16px" viewBox="0 0 23 16" aria-hidden="true">
```
schalkneethling: We should `role="img"` to all of these SVG files.

kuma/payments/jinja2/payments/management.html
```diff
@@ -0,0 +1,128 @@
...
+                </table>
+                {% if not cancel_request %}
+                    <form id="id_stripe_management_form" method="post">
+                        {% csrf_token %}
+                        <button class="button negative" type="button" toggle-confirmation>{{_('Cancel your subscription')}}</button>
+                        <div id="delete-confirmation" class="hidden" aria-hidden="true">
+                            <div class="confirmation column-8 alert danger">
+                                <h5>{{_('Are you sure?')}}</h5>
+                                <p>
+                                    {{_('You will have to set up a new subscription if you wish to resume making payments to MDN.')}}
```
schalkneethling: Just noticed this but, is the official brand name for MDN not MDN Web Docs? We should probably update all instances of MDN to MDN Web Docs to be consistent

kuma/payments/jinja2/payments/management.html
```diff
@@ -0,0 +1,128 @@
...
+                </div>
+            </section>
+        {% endif %}
+        {% if cancel_request %}
+            {% if cancel_success %}
+                <div class="confirmation column-8 alert success">
+                    <b>{{_('Your monthly subscription has been successfully cancelled')}}</b>
+                </div>
+            {% else %}
+                <div class="confirmation column-8 alert success">
```
schalkneethling: This uses the `success` class but it is an error state?

kuma/payments/jinja2/payments/management.html
```diff
@@ -0,0 +1,128 @@
...
+                    </b>
+                </div>
+            {% endif %}
+        {% endif %}
+    </div>
+{% endblock %}
+
+{% block js %}
+  {%- if contribution_enabled %}
+    <script src="https://checkout.stripe.com/checkout.js"></script>
```
schalkneethling: Could we `defer` this?

kuma/static/js/payments-manage.js
```diff
@@ -0,0 +1,27 @@
+
```
schalkneethling: Nit: remove leading empty line

kuma/static/js/payments-manage.js
```diff
@@ -0,0 +1,27 @@
...
+    var toggleConfirmationButtons = Array.from(doc.querySelectorAll('button[toggle-confirmation]'));
+    var confirmationDialog = doc.getElementById('delete-confirmation');
+
+    /**
+     * Toggle the visibility of the confirmation dialog
+     */
+    function toggleDeleteConfirmaiton() {
+        toggleConfirmationButtons[0].classList.toggle('hidden');
+        confirmationDialog.classList.toggle('hidden');
+        // var hidden = confirmationDialog.getAttribute('aria-hidden');
```
schalkneethling: Nit: remove commented out line

kuma/static/js/payments-manage.js
```diff
@@ -0,0 +1,27 @@
...
+        confirmationDialog.toggleAttribute('aria-hidden');
+    }
+
+    toggleConfirmationButtons.forEach(function(button) {
+        button.addEventListener('click', toggleDeleteConfirmaiton);
+    });
```
schalkneethling: Nit: Up to you, but I would prefer if we delegate the `eventListener` to the container instead of attaching `eventListeners` to each button. Not a huge deal here as there are only a few instances, but generally a good idea.

joshJarr: I agree, but in this case it's just one or two buttons. But this is definitely something to consider once we improve the manage payments section, as more logic would be needed.

jinja2/includes/icons/general/money-bill-alt.svg
joshJarr: Addressed.

kuma/payments/jinja2/payments/management.html
joshJarr: Good spot! I've updated a few places here, but it may be worth raising. I'll make a card for this.

kuma/payments/jinja2/payments/management.html
joshJarr: Yep! My bad for blindly duplicating code.  Addressed.

kuma/payments/jinja2/payments/management.html
joshJarr: Actually we currently don't need checkout on this page. This will change if we want to allow for more payment management options in future. But for now I'll remove it.

kuma/static/js/payments-manage.js
joshJarr: 👍

kuma/static/js/payments-manage.js
joshJarr: 👍 oops

kuma/static/styles/components/payments/form.scss
```diff
@@ -321,6 +321,9 @@
     &.is-open {
         display: block;
         visibility: visible;
+        // &.popover-fixed {
+        //     position: fixed;
+        // }
```
schalkneethling: Nit: Can remove the commented out code?

schalkneethling:
@joshJarr I reckon the link on the edit profile page to manage subscriptions is to small and gets lost. I definitely missed it.

screenshot 2018-11-30 at 12 39 23

I am thinking we should change it to be a more obvious button such as the "Back to profile" button on the "Manage subscriptions" page. Thoughts?

screenshot 2018-11-30 at 12 39 33

schalkneethling:
The color of the border and the copy of the cancel button does not match:

screenshot 2018-11-30 at 12 44 24

I would suggest using `#b65456` for both.

kuma/payments/jinja2/payments/management.html
```diff
@@ -0,0 +1,127 @@
...
+                <h1>{{_('Your MDN profile')}}</h1>
+            </header>
+            <div>
+                <a href="{{ url('users.user_edit', username=user.username) }}" class="button neutral back">{% include 'includes/icons/arrows/arrow-left.svg' %} {{_('Back to profile')}}</a>
+            </div>
+            <div class="manage-payments-help column-4">
+                <h4>{{_('Help')}}</h4>
+                <ul>
+                    <li>
+                        <a href="{{ url('recurring_payment_initial') }}" class="cta-link full-width">{{_('Monthly subscriptions FAQ')}} {% include 'includes/icons/arrows/arrow.svg' %}</a>
```
schalkneethling: This should like to the `#contribute-monthly-faqs` fragment.

schalkneethling:
The "confirm" button has the same problem as the "Delete subscription" button mentioned before:

screenshot 2018-11-30 at 12 48 03

schalkneethling:
The contrast ratio of the success message is to low:

screenshot 2018-11-30 at 12 53 03

I would suggest using `#008200`

screenshot 2018-11-30 at 12 53 20

schalkneethling:
The "hey why not set up a thing" message is really large, even trumping its headline:

screenshot 2018-11-30 at 12 54 49

I would suggest bumping this down to `1rem`

screenshot 2018-11-30 at 12 55 10

schalkneethling:
@joshJarr I am also wondering whether the text "Edit payment subscription" should instead read "Edit monthly subscription"?

schalkneethling:
Again, me wondering out loud. Should we not add a link to the subscription management view here?

screenshot 2018-11-30 at 13 01 41

schalkneethling:
Also, seems odd that you need to go to Edit profile to get to the manage subscriptions page. Should this not also be displayed on the profile view?

Not sure how often people view their own profile, so this might not be an issue.

joshJarr:
@schalkneethling Addressed this, thanks! 
Regarding the subscription link on the edit profile, I think you're right, but since we're now linking in the FAQs it's not the end of the world. We're following the current user profile layout, if we start adding buttons for links maybe it would be worth introducing this format across this page? 

Thanks!

PR #5151, opened 2018-11-29
bug 1490727: Remove clear field `x` from email input on IE on payments form.

joshJarr:
IE puts an `X` to the right of inputs in order to clear them.
This overlaps the `?` help text action on the email field, so we should disable the IE marker for this input.

**in this PR**
- https://trello.com/c/Vh52EyJj/96-internet-explorer-x-button-displayed-in-email-field-overlaps-the-button

**Screenshot of the bug:**
screenshot 2018-11-29 at 13 13 41

### DIFF
--- a/kuma/static/styles/components/payments/form.scss
+++ b/kuma/static/styles/components/payments/form.scss
@@ -10,6 +10,11 @@
         text-align: center;
     }
 
+    // Remove 'x' from email input as it overlaps the hint action
+    input[type='email']::-ms-clear {
+        display: none;
+    }
+
     .legal-copy {
         font-size: 12px;
         font-weight: normal;


kuma/static/styles/components/payments/form.scss
```diff
@@ -10,6 +10,11 @@
         text-align: center;
     }
 
+    // Remove 'x' from email input as it overlaps the hint action
+    input[type='email']::-ms-clear {
```
schalkneethling: Interesting that it does it on an input of type `email`. This[1] seems to suggest that it only happens on input of type `text`. Either way, this will resolve the problem. Thanks @joshJarr 

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-clear

PR #5164, opened 2018-12-11
bug 1462400: fix test_robots for attachments domains

escattone:
This PR fixes the `test_robots` test within `tests/headless/test_robots.py` when run against https://developer.mozilla.org (typically via the `prod-integration-tests` branch), since only then was the attachment domain (`mdn.mozillademos.org`) provided as one of the values for the dynamic pytest fixture `any_host_url`. In that case, the attachment domain case should be handled the same as a non-indexed domain.

### DIFF
--- a/tests/headless/__init__.py
+++ b/tests/headless/__init__.py
@@ -5,12 +5,6 @@ import pytest
 pytest.register_assert_rewrite('utils.urls')
 
 
-# Untrusted attachments and samples domains that are indexed
-INDEXED_ATTACHMENT_DOMAINS = set((
-    'mdn.mozillademos.org',         # Main attachments domain
-    'mdn-demos-origin.moz.works',   # Attachments origin
-))
-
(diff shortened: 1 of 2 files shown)


jwhitlock:
Looking closer, it appears that the attachment host origin needs to be updated:

https://github.com/mozilla/kuma/blob/d3e69b1cb20714becc3113ac2ebcc21e461083be/kuma/settings/common.py#L1168

I think this test will work as written when then is shipped, the CDN is cleared, etc.

escattone:
You're right on, great catch @jwhitlock! I'm embarrassed that I missed this! 🤦‍♂️

escattone:
I'll submit a PR to update the `_PROD_ATTACHMENT_ORIGIN`, and I'm closing this PR.

escattone:
See https://github.com/mozilla/kuma/pull/5169

PR #5165, opened 2018-12-11
Issue 683: Kuma-side changes for simplifying KumaScript

davidflanagan:
For https://github.com/mdn/sprints/issues/683 I'm trying to simplify
KumaScript. My first target is to remove the layers of HTTP caching
between Kuma and KumaScript so that I can try to remove the memcached
dependency (and save on hosting costs).

This patch is changes on the Kuma side to prepare for that refactoring.
First, the patch includes a couple of new scripts that I used to build
a test corpus of raw and rendered documents so I can be sure that my
KumaScript changes do not actually change how documents are rendered.

More substantially, this patch modifies kuma/wiki/kumascript.py to
change the get() function to actually make a POST request. This means
that the KumaScript no longer needs to support the GET request for
rendering

### DIFF
--- a/kuma/scrape/scraper.py
+++ b/kuma/scrape/scraper.py
@@ -36,6 +36,12 @@ class Requester(object):
         return self._session
 
     def request(self, path, raise_for_status=True, method='GET'):
+        # When doing a long-running recursive scrape, you may find
+        # it useful to uncomment the following two lines. The sleep
+        # helps to avoid rate-limiting, and the print allows you to
+        # monitor the progress of your scrape.
+        # time.sleep(1)
+        # print path
         url = self.base_url + path
         logger.debug("%s %s", method, url)
         attempts = 0
--- a/kuma/wiki/kumascript.py
+++ b/kuma/wiki/kumascript.py
@@ -1,5 +1,4 @@
 import base64
-import hashlib
 import json
 import time
 import unicodedata
@@ -10,7 +9,6 @@ import requests
 from constance import config
 from django.conf import settings
 from django.contrib.sites.models import Site
-from django.core.cache import cache
 from django.utils.six.moves.urllib.parse import urljoin
 from elasticsearch import TransportError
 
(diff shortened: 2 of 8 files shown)


davidflanagan:
John and Ryan: this is the start of my attempt to remove memcached from KumaScript, and these are changes I need on the Kuma side. Don't feel you need to review this right away... I still have plenty of work to do on the KumaScript side.  But feel free to take a look if you want to see where I'm going with this...  The main point here is that by changing wiki.kumascript.get() and .post() to both use the same POST endpoint, I'll be able to remove the GET endpoint in KumaScript and that is the one that has the complicated (and unnecessary, I think) HTTP caching.

davidflanagan:
I've verified (running locally on my laptop) that with this patch applied Shift-Cmd-R still re-renders the page (using the modified kumascript.get() method) and that editing a page (which uses the modified kumascript.post() method) also correctly previews and renders-on-publish. Not sure I've covered all possible cases, but the basic stuff is working, at least.

davidflanagan:
I've been experimenting with re-rendering the page http://localhost:8000/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters with and without this patch, using Shift-Cmd-R.

On the master branch, the GET request to kumascript takes ~2200ms to complete. But with this patch, the POST request takes ~1500ms to complete. So there is a significant time savings not having KumaScript have to handle a GET request and immediately turn around and make a GET request back to Kuma.

This also seems to demonstrate that all the HTTP caching stuff around the get endpoint wasn't doing us any good, since the only time we re-render is when we were explicitly bypassing the cache.

ExE-Boss:
You could probably use [bug&nbsp;1509430](https://bugzilla.mozilla.org/show_bug.cgi?id=1509430) for this.

davidflanagan:
PR #5182 is the updated version. Closing this one.

PR #5169, opened 2018-12-13
bug 1483072: update prod attachment origin

escattone:
Thanks to @jwhitlock, who pointed out that we're not serving the correct `robots.txt` file for the attachments domain. We should be serving an empty string for `robots.txt` for `mdn.mozillademos.org` (and its origin `demos-origin.mdn.mozit.cloud`), but instead we've been serving `ROBOTS_GO_AWAY_TXT` since we cut-over to the MozIT-based infrastructure. That was my fault for overlooking the need for this update during the cut-over. This PR corrects that oversight.


escattone:
One more thing. This oversight was actually detected by the integration test suite (specifically `tests/headless/test_robots.py::test_robots`) when run against `developer.mozilla.org`.

PR #5170, opened 2019-01-03
Fix sprints/707, always position tooltip correctly

schalkneethling:
See https://github.com/mdn/sprints/issues/707 for problem description.

Tested in Firefox, Chrome, Safari and Edge

@davidflanagan r?


schalkneethling:
Tested in IE11, and this solves the following issue as well:
https://github.com/mdn/sprints/issues/712

kuma/static/js/tooltip.js
```diff
@@ -46,11 +44,7 @@
          */
         function closeTooltip(tooltip) {
             $(tooltip).removeClass('is-open has-arrow-top')
-                .attr('aria-hidden', true)
-                .css({
-                    left: 0,
-                    top: 0
-                });
+                .attr('aria-hidden', true);
```
davidflanagan: You can remove the has-arrow-top class in the line above, too.

kuma/static/js/tooltip.js
```diff
@@ -63,42 +57,6 @@
             });
         }
 
-        /**
```
davidflanagan: Nice to be able to remove so much code. I was surprised, though. Given the name of the file, I assumed that it was intended as a fully-generic tooltip implementation, so I was suprised to realize that we're only using it for one specific pre-positioned tooltip. I'd suggest adding a comment at the top of the file with that context and a pointer to the scss file for the email tooltip.

kuma/static/styles/components/payments/_tooltip.scss
```diff
@@ -0,0 +1,89 @@
...
+        border-color: transparent $text-color;
+        right: -10px;
+    }
+
+    &.is-open {
+        display: block;
+        visibility: visible;
+    }
+
+    &.has-arrow-top {
```
davidflanagan: You can cut this entire bit, if you want, since it appears to be unused now.

kuma/static/styles/components/payments/_tooltip.scss
```diff
@@ -0,0 +1,89 @@
+.tooltip {
```
davidflanagan: It seems like a good idea to move this to a file of its own.

I suggest that you give the file a more specific name, like email-tooltip, since this includes absolute positioning that makes it specific to this one field in this one form

davidflanagan: And maybe also add a comment on this .tooltip class explaining that despite its generic name, it has hardcoded positioning and is only intended for use for one specific tooltip.

schalkneethling:
Thanks for the review @davidflanagan - I will address your comments and update the PR

kuma/static/js/tooltip.js
schalkneethling: I was thinking that we can definitely turn it into one, but for now, it makes more sense to couple it to the payments bit.

escattone:
Deployed.

PR #5171, opened 2019-01-04
Fix Bug 1509833, ensure browser text labels are consistent across all pages for BCD tables

schalkneethling:
Problem is described here[https://github.com/mdn/sprints/issues/680] and in the screenshots below:

## Before

screenshot 2019-01-04 at 16 33 41

## After

screenshot 2019-01-04 at 16 33 51

Tested in Firefox, Chrome, Safari, Edge and IE11

@davidflanagan 	r?

### DIFF
--- a/kuma/static/styles/components/compat-tables/bc-icons.scss
+++ b/kuma/static/styles/components/compat-tables/bc-icons.scss
@@ -4,18 +4,16 @@
 .bc-browsers {
     .bc-head-txt-label {
         display: inline-block;
-        margin: 0 0 0 50px;
+        bottom: 0;
+        margin: 0 0 20px 10px;
         padding: 0;
-        width: 140px;
-        height: 210px;
+        width: 20px;
+        height: 20px;
         transform: rotate(-90deg);
+        transform-origin: left;
         @include set-font-size($small-bump-font-size);
         text-align: left;
         white-space: nowrap;
-
-        @media #{$mq-tablet-only} {
-            margin-left: 42px;
-        }
 
         &:before {
             position: absolute;
--- a/kuma/static/styles/components/compat-tables/bc-table.scss
+++ b/kuma/static/styles/components/compat-tables/bc-table.scss
@@ -106,6 +106,8 @@ Table layout
 .bc-table .bc-browsers th {
     border-top: 0;
     padding: 3px 0;
+    height: 200px;
+    vertical-align: bottom;
     @include set-font-size(22px);
 }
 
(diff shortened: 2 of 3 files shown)


kuma/static/styles/components/compat-tables/bc-icons.scss
```diff
@@ -4,17 +4,26 @@
...
-            margin-left: 42px;
+        /* At this breakpoint the table uses grid layout
+           where supported. This stops `vertical-align: bottom`
+           from working so, we need to switch to this workaround
+           to ensure the layout still works. */
+        @supports (display: grid) {
+            @media #{$mq-tablet-small-desktop-only} {
+                position: relative;
+                bottom: -150px;
+            }
```
davidflanagan: Wouldn't these properties be better in the bc-table-tablet-display-grid mixin in bc-table.scss?

Thanks for the detailed comment here. But I notice that it refers to a vertical-align property that appears in bc-table.scss, which also makes me think that maybe this special case should be in that file. bc-table.scss already has one mention of the bc-browsers class in it, so it wouldn't be breaking abstractions too badly to move this there, I suspect.

Having said all that, I should confess that I'm not very familiar with scss, so don't know what best practices are.

kuma/static/styles/components/compat-tables/bc-icons.scss
```diff
@@ -4,17 +4,26 @@
 .bc-browsers {
     .bc-head-txt-label {
         display: inline-block;
-        margin: 0 0 0 50px;
+        bottom: 0;
+        margin: 0 0 20px 10px;
```
davidflanagan: The asymmetry of the margins surprises and confuses me. Assuming you want equal left and right margins, and given the rotation, I'd have thought that you'd have equal top and bottom margins declared here...

Not saying it is wrong; just mentioning the parts that are hard to understand.

kuma/static/styles/components/compat-tables/bc-icons.scss
```diff
@@ -4,17 +4,26 @@
...
         padding: 0;
-        width: 140px;
-        height: 210px;
+        width: 20px;
+        height: 20px;
         transform: rotate(-90deg);
+        transform-origin: left;
```
davidflanagan: Hmm. I'm used to rotating around the center or around a corner. Maybe this rotate around the midpoint of an edge has something to do with the asymmetrical margins...  I can't visualize what is going on here, but I can see from your screenshots that it worked.

schalkneethling: Yup, on this point and the previous. At this point rotating around the left origin resolves the positioning problem. I would like to revisit the BCD table CSS at some point and improve the overall CSS so there are fewer surprises like these.

schalkneethling:
Feedback addressed, thanks @davidflanagan

PR #5172, opened 2019-01-04
Fix "?redirect=no" for document-based redirects (bug 1404669)

escattone:
See [bug 1404669](https://bugzilla.mozilla.org/show_bug.cgi?id=1404669) for more context.

This PR:
- fixes `?redirect=no` handling for content-based redirects within the development environment (it does not fix the bug for the stage and production environments, for that and for a fuller explanation of this bug and its resolution, see https://github.com/mdn/infra/pull/168)
- since no current unit or functional tests are able to detect this bug, adds two headless functional tests, one of which specifically checks for this bug, that can be run against the sample, stage, and production databases
- refactors the existing content-based redirection tests into the `pytest` style

This PR is a companion to https://github.com/mdn/infra/pull/168, and if approved, should be merged along with that PR before making a deployment.

When testing, after powering-up your development environment with `docker-compose up -d`, the first run of:
```sh
pytest --base-url http://localhost:8000 tests/headless/test_endpoints.py::test_document_based_redirection_suppression
```
should fail, since the stale, incorrect, rendered content is still saved on the `/en-US/docs/MDN/Promote` document. Re-render that document:
```sh
docker-compose exec web ./manage.py render_document /en-US/docs/MDN/Promote
```
and then the test should run successfully.

I believe there are 12 documents within the sample database that are content-based redirects. We might want to consider creating a new sample database once this PR and https://github.com/mdn/infra/pull/168 have been deployed.

### DIFF
--- a/kuma/wiki/tests/test_views_document.py
+++ b/kuma/wiki/tests/test_views_document.py
@@ -6,6 +6,7 @@ Legacy tests are in test_views.py.
 import base64
 import json
 from collections import namedtuple
+from datetime import datetime
 
 import mock
 import pytest
@@ -22,6 +23,7 @@ from kuma.core.models import IPBan
 from kuma.core.tests import assert_no_cache_header, assert_shared_cache_header
 from kuma.core.urlresolvers import reverse
 
+from ..constants import REDIRECT_CONTENT
 from ..events import EditDocumentEvent, EditDocumentInTreeEvent
 from ..models import Document, Revision
 from ..views.document import _apply_content_experiment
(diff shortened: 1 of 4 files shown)


escattone:
Sorry, just noticed that I need to fix the two new tests I added, update coming.

kuma/wiki/tests/test_views_document.py
```diff
@@ -768,3 +770,70 @@ def test_deleted_doc_no_purge_permdeleted(deleted_doc, wiki_moderator,
     assert full_description in content
     assert 'Restore this document' in content
     assert 'Purge this document' not in content
+
+
+@mock.patch('kuma.wiki.kumascript.get')
+def test_redirect_suppression(mock_kumascript_get, constance_config, client,
```
jwhitlock: Thanks for converting this to `pytest`! 😍

tests/headless/test_endpoints.py
```diff
@@ -47,6 +49,34 @@ def test_document(base_url, is_indexed):
         assert content == 'noindex, nofollow'
 
 
+@pytest.mark.headless
+@pytest.mark.nondestructive
+def test_document_based_redirection(base_url):
+    """Ensure that content-based redirects properly redirect."""
+    url = base_url + '/en-US/docs/MDN/Promote'
```
jwhitlock: Here's my sequence of thoughts when reviewing this line:

"OK, this might work in production, but it may not be on staging and would need to be added to the sample database for development."

"Huh, it works in staging and development. Did I scrape that page?"

"We have a redirect link on the homepage 🤦‍♂️"

So good job, and we need to update that link on the homepage while also ensuring ``/en-US/docs/MDN/Promote`` remains in the sample DB. Future work...

PR #5174, opened 2019-01-07
Bug 1509430: rewrite and simplify KumaScript

davidflanagan:
This is a big patch that effectively rewrites the KumaScript server
to use async/await instead of fibers and Future and hirelings. It
also removes unused features and dependencies and modernizes the code.

It removes the need for a memcached server, and renders documents much
faster than the old version of the server does.

- environment.js, cache.js, config.js, plus the macros directory

- Start using prettier prettier

- Add await to lots of macros

- more tests

- Add a test for the template() function

- Add MacroSyntaxError class

- update pegjs version to get better token location reporting

- better error reporting and handling

- Port over the server.js file

- add server tests

- add a parallel request server test

- containerize ks2

- Kuma-side changes for simplifying KumaScript

For https://github.com/mdn/sprints/issues/683 I'm trying to simplify
KumaScript. My first target is to remove the layers of HTTP caching
between Kuma and KumaScript so that I can try to remove the memcached
dependency (and save on hosting costs).

This patch is changes on the Kuma side to prepare for that refactoring.
First, the patch includes a couple of new scripts that I used to build
a test corpus of raw and rendered documents so I can be sure that my
KumaScript changes do not actually change how documents are rendered.

More substantially, this patch modifies kuma/wiki/kumascript.py to
change the get() function to actually make a POST request. This means
that the KumaScript no longer needs to support the GET request for
rendering

- kumascript environment tweaks and macro debugging, including string.asyncReplace() function

- Change the way the macro execution envionment is bound together; add a test to verify that all macros compile; fix the macros that do not compile.

- fix a bug where the Compat macro was writing into the bdd data

- Fix final bugs to reach rendering parity

- Better error classes + simplify requestHTTPResource

- tweak the cache size


Makefile
```diff
@@ -10,10 +10,12 @@ VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || git rev-pars
 KS_VERSION ?= $(shell cd kumascript && git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD)
 export KUMA_REVISION_HASH ?= $(shell git rev-parse HEAD)
 export KUMASCRIPT_REVISION_HASH ?= $(shell cd kumascript && git rev-parse HEAD)
+export KUMASCRIPT2_REVISION_HASH ?= $(shell cd kumascript2 && git rev-parse HEAD)
```
davidflanagan: I don't intend to leave this file as it is. The KUMASCRIPT2 variables are here as they are so that I could easily run the two versions side-by-side.

Before deploying, this file will be cleaned up.

docker-compose-2.yml
```diff
@@ -0,0 +1,107 @@
+version: '2.1'
```
davidflanagan: This file is for testing with the new version of kumascript while still being able to run the old.

Before deplying, I'd just copy this file over to docker-compose.yml

docker/images/kumascript2/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM node:10.14.0-slim
```
davidflanagan: Same deal here: I've got a separate kumascript2/ directory because I wanted to be able to have old and new containers at the same time. This would be fixed before deploying.

kuma/wiki/kumascript.py
```diff
@@ -40,20 +38,18 @@ def should_use_rendered(doc, params, html=None):
             (force_macros or (not no_macros and not show_raw)))
 
 
-def post(request, content, locale=settings.LANGUAGE_CODE):
+def _post(content, env_vars, timeout=config.KUMASCRIPT_TIMEOUT):
```
davidflanagan: This file contains the main changes I made to Kuma as part of this KumaScript refactor: KumaScript no longer supports a GET endpoint: everything has to be done with POST. So this file is changed so that the get() method actualliy does a POST request. (I punted on changing the method name, for now.)

kumascript2/macros/A11yRoleQuicklinks.ejs
```diff
@@ -0,0 +1,26 @@
...
+    case "ru":
+        url = "/ru/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
+        break;
+    default:
+        url = "/en-US/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
+        break;
+}
+%>
+<section class="Quick_links" id="Quick_Links">
+<%-await wiki.tree(url, 2, 0, 0, 1)%>
```
davidflanagan: You should be able to ignore the macro files. Since this PR is against Kuma, these are copies of them not just the changed lines.

The primary change is that we now use the `await` keyword before all async (or possbily async) function calls.

davidflanagan: At this point in the review skip way down to kumascript2/src/

kumascript2/src/environment.js
```diff
@@ -0,0 +1,1107 @@
+/**
```
davidflanagan: This is the longest file in the src/ directory. Much of it is ported from the old auto-required macro files, which is why some of the code is kind of funky.

kumascript2/src/firelogger.js
```diff
@@ -0,0 +1,99 @@
+/**
```
davidflanagan: This file is pretty much unmodified from the old KumaScript

kumascript2/src/parser.pegjs
```diff
@@ -0,0 +1,135 @@
+/*
```
davidflanagan: The parser.js file is generated from this one. This is slightly changed from the old KumaScript to get better error reporting. And we now build parser.js at build time instead of at runtime.

scripts/dump_local_db.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
davidflanagan: After scraping a whole bunch of files into my local db, I wrote this script to make a backup of the database. Seems worth keeping around.

scripts/extract_documents.py
```diff
@@ -0,0 +1,66 @@
+""" A script to extract raw and rendered versions of all documents in the db
```
davidflanagan: This is the script I used to render all the documents in my local db, to verify that the new versino of KumaScript produces the same output as the old version.

davidflanagan:
Sorry this is such a giant patch to review. The main changes are in kuma/wiki/kumascript.py and in kumascript2/src/*.js and kumascript2/tests/*.js.  You can probably ignore all of kumascript2/macros and kumascript2/tests/fixtures/

ExE-Boss:
I think you should first open this against the&nbsp;KumaScript repository.

davidflanagan:
The KS PR version is now ready at https://github.com/mdn/kumascript/pull/1035

It will be easier to review the changes with that PR.

But if you want to try the new version of KS out, this PR will be easier because it has the necessary Kuma changes to make it all work.

jwhitlock:
Thanks @davidflanagan, the diff is easier to view on mdn/kumascript#1035

PR #5175, opened 2019-01-08
Bug 1490727, sprints/710, cosmetic updates to payments FAQ page

schalkneethling:
To clarify the first item in the list on https://github.com/mdn/sprints/issues/710, below is the desired design layout:

screenshot 2019-01-08 at 09 23 55

@davidflanagan r?

### DIFF
--- a/kuma/static/styles/components/payments/page.scss
+++ b/kuma/static/styles/components/payments/page.scss
@@ -2,6 +2,9 @@
     .section {
         display: flex;
         position: relative;
+        margin: 0 auto;
+        padding: 56px 18px;
+        max-width: 1170px;
 
         .sub-section {
             margin-bottom: 2em;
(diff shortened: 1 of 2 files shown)


schalkneethling:
@jwhitlock There are no Bugzilla bugs for these. We discussed this prior to All Hands, and as these were existing cards on Potato's Trello board, we decided it would be fine to just have issues, as opposed to creating a lot of Bugzilla bugs.

I can, of course, go ahead and create those if need be. Let me know, thanks!

kuma/static/styles/components/payments/page.scss
```diff
@@ -64,6 +72,13 @@
     .faqs {
         flex: 60%;
 
+        a {
+            border-bottom: 1px solid $payment-form-button-color;
```
davidflanagan: This surprises me. What's the reason for a border instead of a text-decoration? Is this related to the comment "same as legal copy links"?

schalkneethling: Yup, I found it strange as well. I did not want to mess too much with the existing code other than fixing what was required. Not sure why the Potato devs decided to go this route. Something I will look into as I clean up the SaSS

jwhitlock:
@schalkneethling we gathered previous work for payments under [bug 1490727](https://bugzilla.mozilla.org/show_bug.cgi?id=1490727). I'm OK continuing with that bug unless you think an issue is large enough to deserve a new bug.

schalkneethling:
@jwhitlock title and commit message updated to reflect Bugzilla bug.

PR #5176, opened 2019-01-08
Bug 1490727, sprints/706, payment button obscured when fields are invalid

schalkneethling:
Issue details are available in https://github.com/mdn/sprints/issues/706

## Before

screenshot 2019-01-08 at 12 56 26

## After

screenshot 2019-01-08 at 12 56 37

@davidflanagan 	r?

### DIFF
--- a/kuma/static/js/components/payments/payments-handler.js
+++ b/kuma/static/js/components/payments/payments-handler.js
@@ -386,22 +386,6 @@
         }
     }
 
-    var mediaQueryList = null;
-
-    /**
-     * Increases or decreases the height of the `popoverBanner`
-     * based on a `mediaQueryList` match
-     */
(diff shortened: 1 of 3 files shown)


schalkneethling:
@jwhitlock Same here as in https://github.com/mozilla/kuma/pull/5175#issuecomment-452384234

schalkneethling:
@jwhitlock title and commit message updated to reflect Bugzilla bug.

schalkneethling:
> I'm not thrilled with the new design: it feels cramped to me at the top and bottom now.

Totally agree, I do have some ideas to improve this.

> But that is a nit: it is obviously more important to have the button reliably displayed, so I'm okay with landing this as is.

Yup, that is why I decided to stop myself and stick with good enough ;)

PR #5177, opened 2019-01-08
Bug 1490727, sprints/704, cannot check terms checkbox with field label

schalkneethling:
Details of the problem detailed here:
https://github.com/mdn/sprints/issues/704

@davidflanagan 	r?

### DIFF
--- a/kuma/static/styles/components/payments/form.scss
+++ b/kuma/static/styles/components/payments/form.scss
@@ -16,13 +16,14 @@
     }
 
     .legal-copy {
+        display: inline-block;
         font-size: 12px;
         font-weight: normal;
         letter-spacing: .1px;
         line-height: 1.35;
         margin-bottom: 18px;
 
-        @media #{$mq-small-desktop-and-up} {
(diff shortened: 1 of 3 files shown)


schalkneethling:
@jwhitlock Same here as in https://github.com/mozilla/kuma/pull/5175#issuecomment-452384234

kuma/static/styles/components/payments/form.scss
```diff
@@ -16,13 +16,14 @@
...
     .legal-copy {
+        display: inline-block;
         font-size: 12px;
         font-weight: normal;
         letter-spacing: .1px;
         line-height: 1.35;
         margin-bottom: 18px;
 
-        @media #{$mq-small-desktop-and-up} {
+        @media #{$mq-large-desktop-and-up} {
```
davidflanagan: I don't get this change. Is the text going to be centered on small desktop or something?

schalkneethling: The alignment existed before. I simply needed to have the media query kick in a bit later due to the addition of the `max-width`

PR #5182, opened 2019-01-09
Bug 1509430: Kuma changes for the new version of KumaScript

davidflanagan:
https://github.com/mdn/kumascript/pull/1035 is a big patch that
rewrites and modernizes the KumaScript server. This is a smaller PR
that updates the Kuma repo for those changes.

The primary change is that the KumaScript server no longer supports
a GET endpoint for rendering documents, so Kuma now needs to send
document content to KumaScript for rendering with a POST request. This
change is primarily in wiki/kumascript.py, but it causes small
follow-on changes in a nubmer of other files as well.

The patch also includes some new useful scripts I wrote while testing.

And it changes docker-compose.yml to remove memcached which is no
longer required.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,9 +58,6 @@ services:
     ports:
       - "8001:8000"
 
-  memcached:
-    image: memcached
-
   mysql:
     image: mysql:5.6
     environment:
@@ -95,22 +92,13 @@ services:
     command: node run.js
     depends_on:
       - api
-      - memcached
+      - redis
     environment:
-      - interactive_examples__base_url=${INTERACTIVE_EXAMPLES_BASE:-https://interactive-examples.mdn.mozilla.net}
-      - live_samples__base_url=http://${ATTACHMENT_HOST:-localhost:8000}
-      - log__console=true
-      - log__file=
-      - server__template_root_dir=macros
-      - server__document_url_template=http://api:8000/en-US/docs/{path}?raw=1&redirect=no
-      - server__memcache__server=memcached:11211
-      - server__template_class=EJSTemplate
-      - server__autorequire__mdn=MDN:Common
-      - server__autorequire__Page=DekiScript:Page
-      - server__autorequire__String=DekiScript:String
-      - server__autorequire__Uri=DekiScript:Uri
-      - server__autorequire__Web=DekiScript:Web
-      - server__autorequire__Wiki=DekiScript:Wiki
+      - DOCUMENT_URL_TEMPLATE=http://api:8000/en-US/docs/{path}?raw=1&redirect=no
+      - DOCUMENT_URL=http://api:8000
+      - INTERACTIVE_EXAMPLES_URL=${INTERACTIVE_EXAMPLES_BASE:-https://interactive-examples.mdn.mozilla.net}
+      - LIVE_SAMPLES_URL=http://${ATTACHMENT_HOST:-localhost:8000}
+      - REDIS_URL=redis://redis:6379/2
     ports:
       - "9080:9080"
     volumes:
--- a/kuma/wiki/kumascript.py
+++ b/kuma/wiki/kumascript.py
@@ -1,5 +1,4 @@
 import base64
-import hashlib
 import json
 import time
 import unicodedata
@@ -10,7 +9,6 @@ import requests
 from constance import config
 from django.conf import settings
 from django.contrib.sites.models import Site
-from django.core.cache import cache
 from django.utils.six.moves.urllib.parse import urljoin
 from elasticsearch import TransportError
 
@@ -40,20 +38,26 @@ def should_use_rendered(doc, params, html=None):
             (force_macros or (not no_macros and not show_raw)))
 
 
-def post(request, content, locale=settings.LANGUAGE_CODE):
+def _post(content, env_vars, cache_control='',
+          timeout=config.KUMASCRIPT_TIMEOUT):
     url = settings.KUMASCRIPT_URL_TEMPLATE.format(path='')
     headers = {
         'X-FireLogger': '1.2',
     }
-    env_vars = {
-        'url': request.build_absolute_uri('/'),
-        'locale': locale,
-    }
+
+    # If the user does a hard reload we see Cache-Control:no-cache in
+    # the request header. And we pass that header on to Kumascript so
+    # that it does not use its cache when re-rendering the page.
+    if cache_control == 'no-cache':
+        headers['Cache-Control'] = cache_control
+
     add_env_headers(headers, env_vars)
+
     response = requests.post(url,
-                             timeout=config.KUMASCRIPT_TIMEOUT,
                              data=content.encode('utf8'),
-                             headers=headers)
+                             headers=headers,
+                             timeout=timeout)
+
     if response:
         body = process_body(response)
         errors = process_errors(response)
@@ -61,6 +65,13 @@ def post(request, content, locale=settings.LANGUAGE_CODE):
     else:
         errors = KUMASCRIPT_TIMEOUT_ERROR
         return content, errors
+
+
+def post(request, content, locale=settings.LANGUAGE_CODE):
+    return _post(content, {
+        'url': request.build_absolute_uri('/'),
+        'locale': locale,
+    })
 
 
 def _get_attachment_metadata_dict(attachment):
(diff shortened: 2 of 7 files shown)


docker-compose.yml
```diff
@@ -95,22 +92,11 @@ services:
     command: node run.js
```
jwhitlock: This should be
```suggestion
     command: node index.js
```

kuma/wiki/content.py
```diff
@@ -183,6 +183,13 @@ def clean_content(content):
     else:
         allowed_iframe_patterns = settings.ALLOWED_IFRAME_PATTERNS
         filtered = parsed.filterIframeHosts(allowed_iframe_patterns)
+
+    # NOTE(djf): if you ever want to compare HTML renderings of a document
+    # (before and after a KumaScript or macro change, for example) you
+    # may need to add alphabetical_attributes=True to the serialize call
+    # below so that attributes are output in a deterministic order. This was
+    # a particular problem for <iframe> tags, presumably because of the
+    # iframe filtering above.
```
jwhitlock: I think adding ``alphabetical_attributes=True`` would be a good stand-alone change, then we could get rid of this comment.

docker-compose.yml
davidflanagan: The KS patch originally had index.js, but I changed it back to run.js so that I wouldn't have to modify it here or in the Dockerfile.  So I'm pretty sure this is correct as it stands.

kuma/wiki/content.py
davidflanagan: It seems like it might only be iframes that have their attribute ordering changed? If so, then forcing alphabetical might drive writers nuts when they look at their diffs?

docker-compose.yml
jwhitlock: Yep, I was a few commits behind, on e3f80902fe52895b2463ab4e5f9ee2726b74d678.

davidflanagan:
@jwhitlock: I've rebased to fix the kumascript issue and the docker-compose.yml merge conflict. 

I'm waiting for more feedback from you on the alphabetical attributes issue before doing anything there.

jwhitlock:
It looks like there were some changes [way back in 2013](https://github.com/html5lib/html5lib-python/blob/master/CHANGES.rst#10b2) to allow html5lib to preserve source attribute order. However, it appears that none of the parsers preserve attribute order - not [etree](https://docs.python.org/2/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.items) (the default and the one we're using), not [dom](https://docs.python.org/2/library/xml.dom.html#xml.dom.NamedNodeMap.item), and probably not [lxml](https://lxml.de/tutorial.html#elements-carry-attributes-as-a-dict). So ``alphabetical_attributes=True`` would be needed to get a consistent order, and would probably cause a lot of chaos in our diffs.

I think there's no action for this PR, and the comment can stay. I'll have to sleep on if we want ``alphabetical_attributes=True`` in the near future.

kuma/scrape/scraper.py
```diff
@@ -36,6 +36,12 @@ def session(self):
         return self._session
 
     def request(self, path, raise_for_status=True, method='GET'):
+        # When doing a long-running recursive scrape, you may find
+        # it useful to uncomment the following two lines. The sleep
+        # helps to avoid rate-limiting, and the print allows you to
+        # monitor the progress of your scrape.
+        # time.sleep(1)
```
jwhitlock: Is it faster to `sleep(1)` for each request, rather than hit the rate limit and wait for it to clear? If so, we could try to add some smarter self-limiting.

scripts/dump_local_db.sh
```diff
@@ -0,0 +1,32 @@
...
+#################
+# Setup Variables
+#################
+
+if [ -z "$DATABASE_URL" ]; then
+    echo "DATABASE_URL must be a valid connection URL. Are you in the docker environment?"
+    exit 1
+fi
+
+# Parse DATABASE_URL for database connection parts
```
jwhitlock: Note: copied from `scripts/create_sample_db.sh`. I'm not sure what the function reuse story is for bash.

scripts/extract_documents.py
```diff
@@ -0,0 +1,66 @@
...
+from datetime import datetime
+
+# Adjust the path so the imports below all work.
+sys.path[0] = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+
+import django
+from django.conf import settings
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'kuma.settings.testing')
+django.setup()
```
jwhitlock: This suggests this should be a management command rather than a stand-alone script.

scripts/extract_documents.py
```diff
@@ -0,0 +1,66 @@
+""" A script to extract raw and rendered versions of all documents in the db
+
+This script extracts documents from the db into a testdocs/ directory.
+It is intended for creating a test corpus and verifying that changes
+to kuma and/or kumascript do not cause rendering changes to the documents.
+
+In order to make it work it may be necessary to modify the KUMASCRIPT_TIMEOUT
+setting in kuma/settings/common.py to be at least 60 seconds
```
jwhitlock: [django-constance](https://django-constance.readthedocs.io/en/latest/) is used to set variables from the database, with the default values in the settings file. The preferred way to adjust the setting is in the Django admin or with ``./manage.py constance set KUMASCRIPT_TIMEOUT 60``.

I find environment variables to be a better way to configure Django dynamically, and it is a personal goal to get rid of Constance or drastically reduce its usage. ``KUMASCRIPT_TIMEOUT`` is one of the harder ones to transition, because a lot of tests use ``KUMASCRIPT_TIMEOUT=0`` to avoid live calls to the kumascript service during tests.

kuma/scrape/scraper.py
davidflanagan: It seemed faster to me, but maybe just becasue of the print statement so I could see the steady progress. The 1 minute sleeps were a drag.

Not a change I was confident enough in to uncomment, and not something that seems worth spending more time on. So I could also just remove the comment entirely and leave the file unmodified.

scripts/extract_documents.py
davidflanagan: It was useful to me while developing and testing this change. But I'm not sure whether it is generally useful enough to warrant further work. If you don't want to land this as-is, I'd be inclined to just remove it rather than invest more time in it.

davidflanagan:
> This is working locally in conjunction with the kumascript changes.
> 
> I would prefer this as several PRs:
> 
>     1. Add `scripts/dump_local_db.sh` (along with extracting the `DATABASE_URL` parsing into a shared file)
> 
>     2. Add `scripts/extract_documents.py` (preferably converted to a Django management command)
> 
>     3. Remove the caching infrastructure from `kumascript.py` `get`, along with tests and utility functions
> 
>     4. Add a comment on `kuma/scrape/scraper.py` (or better yet, implement `--per-second=15`, but that could get complex)
> 
>     5. Add a comment on `kuma/wiki/content.py` (or implement `alphabetical_attributes=True`)
> 
>     6. Convert to standard environment variables instead of underscored config overrides for configuring KumaScript
> 
>     7. Convert to `POST` for all kumascript rendering calls (and maybe transitioning `get()` to `render`, but that does seem easier to do after the API change.
> 
>     8. Drop memcache
> 
> 
> Changes 1-5 are independent of the kumascript PR, and could be merged in any order. Changes 6 and 7 are required at the same time as the kumascript PR, but I suspect they would pair to more incremental changes on the kumascript side as well. Change 8 could be done well after the KS changes are merged, deployed, and de-bugged.
> 
> Bundling these as a single PR makes it easy to submit, but challenging to review other than "yep mostly works" and merge a bunch of unrelated changes.
> 
> I'm wavering between Approve, since it all works, and Request Changes, since they are things I'd request if this was broken up and wasn't blocking forward movement. So you get a Comment for now 😝. I'll return after I wrestle with the kumascript PR.

Changes 1, 2, 4 and 5 could also just be deleted from the PR entirely. They were scripts and notes that were useful to me during development, but none of them seem worth spending more time (PRs, reviews, deploys) on to me.

I agree that #8 (modifying the docker-compose.yml file to remove memcache) can be done separately. And that seems like a good idea: make sure everything still works before we start changing the infrastructure config.

Since 6 and 7 have to land at the same time, it seems wrong to have them in separate commits to me--it just increases the number of places in the commit logs where a rollback would be broken.

And I see that 3 could be done separately, but if I'm understanding you that would mean that we'd leave the get() method and all of its caching apparatus intact, but modify the rest of the code so that nothing calls it anymore? I'm not sure, but I suspect we'd also have to delete tests that test the get method as part of 6 and 7. I would have expected you'd want to delete the unused code when it became unused.

So if we're going to break this up, I'd recommend doing 3, 6, and 7 in one PR, 8 in a second PR, and just punting on the other stuff.

kuma/wiki/models.py
```diff
@@ -564,7 +564,7 @@ def render(self, cache_control=None, base_url=None, timeout=None):
             # A timeout of 0 should shortcircuit kumascript usage.
             self.rendered_html, self.rendered_errors = self.html, []
         else:
-            self.rendered_html, errors = kumascript.get(self, cache_control,
```
escattone: Just documenting here what @davidflanagan and I already discussed in an IRC channel. We need to keep the `cache_control` as part of the `kumascript.get` signature and calls.

kuma/wiki/kumascript.py
```diff
@@ -80,43 +83,21 @@ def _get_attachment_metadata_dict(attachment):
...
-    if not cache_control:
-        # Default to the configured max-age for cache control.
-        max_age = config.KUMASCRIPT_MAX_AGE
-        cache_control = 'max-age=%s' % max_age
+# TODO(djf): This get() function is actually implemented on top of
+# _post() and it performs an HTTP POST request. It should probably
+# be renamed to render_document(), and the post() method above should
+# be renamed to render_string(), maybe. For now, though, there are so
+# many tests that mock kumascript.get() that I've left the name unchanged.
+def get(document, base_url, timeout=config.KUMASCRIPT_TIMEOUT):
```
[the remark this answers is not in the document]
escattone: I would vote for adding `cache_control` only to the headers, and not including it into the `env_vars` since @davidflanagan's mdn/kumascript#1035 (which this PR requires) doesn't need it in both places.

kuma/wiki/kumascript.py
```diff
@@ -88,7 +96,9 @@ def _get_attachment_metadata_dict(attachment):
 # be renamed to render_document(), and the post() method above should
 # be renamed to render_string(), maybe. For now, though, there are so
 # many tests that mock kumascript.get() that I've left the name unchanged.
-def get(document, base_url, timeout=config.KUMASCRIPT_TIMEOUT):
+def get(document, base_url,
+        cache_control='',
```
escattone: Super minor, but it's more common practice in this case to use `cache_control=None`.

kuma/wiki/kumascript.py
```diff
@@ -38,11 +38,19 @@ def should_use_rendered(doc, params, html=None):
             (force_macros or (not no_macros and not show_raw)))
 
 
-def _post(content, env_vars, timeout=config.KUMASCRIPT_TIMEOUT):
+def _post(content, env_vars, cache_control='',
```
escattone: Super minor, but it's more common practice in this case to use `cache_control=None`.

kuma/wiki/kumascript.py
```diff
@@ -40,20 +38,26 @@ def should_use_rendered(doc, params, html=None):
             (force_macros or (not no_macros and not show_raw)))
 
 
-def post(request, content, locale=settings.LANGUAGE_CODE):
+def _post(content, env_vars, cache_control='',
```
escattone: I would have used `cache_control=None` here, but strings are also immutable and using an empty string better indicates the type of what is expected.

ExE-Boss:
```suggestion
def _post(content, env_vars, cache_control=None,
```

jwhitlock:
Thanks @davidflanagan for this work (and for removing the non-essential bits), and to @escattone for the review.

PR #5184, opened 2019-01-10
Fix the label of the seoSummary style to be just "Summary"

a2sheppy:
In PR #5154, the name of the seoSummary class was changed
in the CKEditor menu to just "Summary". This patch makes
the same change to the CSS that is used to display the
style when in the editor, so that the label drawn above
the styled content says that instead of "SEO Summary".


a2sheppy:
I expect @schalkneethling would be a good choice to review this, both because obviously, and because he reviewed the previous PR #5154.

a2sheppy:
I did test this locally, btw. :)

PR #5188, opened 2019-01-14
Bug 1518844, remove focus rings from focusable elements, except for keyboard users

schalkneethling:
Problem described in https://github.com/mdn/sprints/issues/705

We can eventually get rid of the JS once `:focus-visible` has wide enough browser support. You can read more on `:focus-visible` here:
https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo

@davidflanagan r?

### DIFF
--- a/kuma/static/js/components/payments/focus-visible.js
+++ b/kuma/static/js/components/payments/focus-visible.js
@@ -1,0 +1,29 @@
+(function() {
+    'use strict';
+    var contributeForm = document.getElementById('contribute-form');
+
+    function removeFocusVisible() {
+        this.classList.remove('focus-visible');
+        this.removeEventListener('focusout', removeFocusVisible);
+    }
+
+    // When a keyup event is triggered on the contribute form,
+    contributeForm.addEventListener('keyup', function(event) {
+        var currentTarget = event.target;
+        var targetType = currentTarget.type;
+
+        /* The key pressed was the tab key, and the target field
+           type is one of `text`, `email`, or `number */
+        if (
+            event.key === 'Tab' &&
+            (targetType === 'text' ||
+                targetType === 'email' ||
+                targetType === 'number')
+        ) {
+            // add the `focus-visible` class.
+            currentTarget.classList.add('focus-visible');
+            // listen for when the field loses focus,
+            currentTarget.addEventListener('focusout', removeFocusVisible);
+        }
+    });
+})();
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -959,6 +959,7 @@ PIPELINE_JS = {
     },
     'payments': {
         'source_filenames': (
+            'js/components/payments/focus-visible.js',
             'js/components/payments/payments-handler-utils.js',
             'js/components/payments/payments-handler.js',
             'js/components/payments/payments-faq.js',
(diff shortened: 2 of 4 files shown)


kuma/static/js/components/payments/focus-visible.js
```diff
@@ -0,0 +1,25 @@
...
+        //  and the target field type is one of `text`, `email`, or `number
+        if (
+            targetType === 'text' ||
+            targetType === 'email' ||
+            targetType === 'number'
+        ) {
+            // add the `focus-visible` class.
+            currentTarget.classList.add('focus-visible');
+            // listen for when the field loses focus,
+            currentTarget.addEventListener('focusout', function() {
```
davidflanagan: I wish the DOM had something like the jQuery 'once' event handler registration method...

This focusout handler never gets removed, but gets added multiple times, so if the user tabs around through the form a lot, I think you're going to end up calling classList.remove() multiple times.

Would it work to just have a focusout handler on contributeForm in parallel to the keyup handler?

schalkneethling: I was a bit concerned about that as well. I believe I have found a way to address it.

kuma/static/js/components/payments/focus-visible.js
```diff
@@ -0,0 +1,27 @@
+(function() {
+    'use strict';
+    var contributeForm = document.getElementById('contribute-form');
+
+    function removeFocusVisible() {
+        this.classList.remove('focus-visible');
+        this.removeEventListener('focusout', removeFocusVisible);
```
schalkneethling: @davidflanagan Removing the event listener here.

kuma/static/js/components/payments/focus-visible.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    // When a keyup event is triggered on the contribute form,
+    contributeForm.addEventListener('keyup', function(event) {
```
davidflanagan: IIUC, when the user clicks on an input field, it will not display a focus ring. But as soon as they start typing into the field, the focus ring will appear. What if you only added the focus-visible class if the key up event was for the Tab key. Would that work?

kuma/static/js/components/payments/focus-visible.js
```diff
@@ -0,0 +1,27 @@
...
+        var currentTarget = event.target;
```
davidflanagan: What happens if the user clicks on text field #1 and the types the tab key to move focus to text field #2. Which one gets the focus ring?  I'd guess that the focus and blur events happen between the keydown and keyup events, so that by the time the keyup event arrives focus has moved. But I'd also expect that event.target would have to be the same for keydown and key up, which would imply that the wrong text field would get the focus ring in that case.

So would it be better to use document.activeElement here instead of event.target?

kuma/static/js/components/payments/focus-visible.js
schalkneethling: Yup, that would definitely be an improvement. I will update the code.

schalkneethling:
Thank you for the review @davidflanagan. In general, I do not like to remove accessibility affordances provided by the user agent, only to add them back using JavaScript or similar.

In this case, it seemed ok because of the custom input fields. Generally, I would push back on this.

PR #5157, opened 2018-12-06
 Bug 1512490: Add Prism syntax definition for EJS

ExE-Boss:
See [bug&nbsp;1512490](https://bugzilla.mozilla.org/show_bug.cgi?id=1512490) for details.

Also updates PrismJS to version 1.15, which adds official support for JSON syntax highlighting.

review?(@davidflanagan, @Elchi3, @jwhitlock, @wbamberg)

PR #5189, opened 2019-01-14
Bug 1520004: Update Prism to version 1.15

ExE-Boss:
Resolves [bug&nbsp;1520004](https://bugzilla.mozilla.org/show_bug.cgi?id=1520004).

Extracted from #5157.

---

review?(@escattone, @jwhitlock)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,7 +3,7 @@
     "dependencies": {
         "html5shiv": "3.6.2",
         "jquery": "2.2.0",
-        "prism": "1.2.0",
+        "prism": "1.15.0",
         "selectivizr": "https://github.com/keithclark/selectivizr.git#1.0.2"
     },
     "install": {
@@ -19,6 +19,7 @@
                 "bower_components/prism/components/prism-css.js",
                 "bower_components/prism/components/prism-clike.js",
                 "bower_components/prism/components/prism-javascript.js",
+                "bower_components/prism/components/prism-json.js",
                 "bower_components/prism/components/prism-css-extras.js",
                 "bower_components/prism/components/prism-rust.js",
                 "bower_components/prism/plugins/line-highlight/prism-line-highlight.js",
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1007,12 +1007,12 @@ PIPELINE_JS = {
             "js/libs/prism/prism-css.js",
             "js/libs/prism/prism-clike.js",
             "js/libs/prism/prism-javascript.js",
+            "js/libs/prism/prism-json.js",
             "js/libs/prism/prism-css-extras.js",
             "js/libs/prism/prism-rust.js",
             "js/libs/prism/prism-line-highlight.js",
             "js/libs/prism/prism-line-numbers.js",
 
-            'js/prism-mdn/components/prism-json.js',
             'js/syntax-prism.js',
         ),
         'output_filename': 'build/js/syntax-prism.js',
--- a/kuma/static/js/libs/prism/prism-clike.js
+++ b/kuma/static/js/libs/prism/prism-clike.js
@@ -1,26 +1,30 @@
 Prism.languages.clike = {
 	'comment': [
 		{
-			pattern: /(^|[^\\])\/\*[\w\W]*?\*\//,
+			pattern: /(^|[^\\])\/\*[\s\S]*?(?:\*\/|$)/,
 			lookbehind: true
 		},
 		{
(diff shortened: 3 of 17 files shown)


jwhitlock:
Sorry for the noise. I had to close and open the PR because our TravisCI integration was broken when this was opened.

ExE-Boss:
Why has the flake8 job failed?
https://travis-ci.org/mozilla/kuma/jobs/481110184

jwhitlock:
TravisCI hasn't been building for a while. A PR with that error was merged to master, and your commit was based on it. We can ignore it or you can rebase on current master to get rid of it.

ExE-Boss:
Let’s ignore it and merge this.

ExE-Boss:
This has no&nbsp;conflicts with&nbsp;https://github.com/mozilla/kuma/pull/5182 if&nbsp;that’s what you’re worried about.

jpmedley:
@ExE-Boss I appreciate you thinking of me, but I have nothing to contribute here.

jwhitlock:
> This has no conflicts with #5182 if that’s what you’re worried about.

I wrote a 900 word response, but I think it belongs on Discourse instead of a PR comment.

There is no code conflict, but there are conflicts:

1. Our review and deployment focus is on the KumaScript refactor, and @escattone an @davidflanagan shouldn't take time or attention away from that.
2. We're planning on major changes to front-end development (https://github.com/mdn/sprints/issues/766), and @davidflanagan and @schalkneethling are collaborating to make those recommendations. The bias is to delay asset changes until the new pipelines are in place.
3. It's unclear what the benefit is of updating this library, versus the risk of breaking something. A reviewer will need to investigate what has updated from Prism 1.2 to 1.15, if others have issues with the change, if your changes are complete, etc.
4. Syntax highlighting is used on many pages on MDN, and the effect of this change will need to be tested against real pages, preferably during the review process. This will either be a lengthy review, or committing to a series of fixes as content issues are found.

I'm assigning @schalkneethling as reviewer, but it is up to him if reviewing this is compatible with his other priorities.

jwhitlock:
We've shipped the KumaScript refactor, but the other blockers remain. I don't think there is much that can unblock the asset pipeline work, other than not taking time for other reviews.

For the Prism changes, it would be helpful to link to the Prism changelog, if available, and see if there are any warnings for this upgrade path.

For the content changes, it would be helpful to find the URLs of pages that would be affected by this change. Screenshots of the before and after would be super helpful. If there are no visual changes, it would be good to know if that is success (nothing broke!) or evidence that we don't need the update.

ExE-Boss:
Here’s the Prism changelog:
https://github.com/PrismJS/prism/blob/master/CHANGELOG.md

schalkneethling:
@ExE-Boss is there a specific feature or improvement to Primsjs that makes you feel like this is a priority for us to update?

Might there be a performance hit if we add support for more programming languages? This touches almost all pages on MDN Web Docs, so I just want to be very cautious when upgrading.

We are pretty far behind, and these are dot releases but, pays to be on the safe side though ;) Let me know your thoughts, thanks!

ExE-Boss:
The primary reason is that JSON syntax highlighting is now built into Prism, which means that it no longer extends the JavaScript syntax and isn’t as error prone as&nbsp;the&nbsp;current JSON syntax.

As for performance, there shouldn’t be any noticeable difference as the number of Prism language syntax definitions installed on&nbsp;MDN stays the&nbsp;same. In&nbsp;fact, performance should be improved, as there have been a lot of optimisations to the regexes to make them faster.

Also, Prism 1.14 adds support for textual WebAssembly, which might be useful to add in a follow&#x2011;up.

ExE-Boss:
I’ve now added a section to my Prism issues page on MDN for a comparison of the current JSON syntax definition and the official Prism JSON syntax definition:
https://developer.mozilla.org/en-US/docs/User:ExE-Boss/Prism_issues#MDN_JSON_vs_Prism_JSON_syntax_highlighting

ExE-Boss:
I have now opened [bug&nbsp;1527533](https://bugzilla.mozilla.org/show_bug.cgi?id=1527533) to&nbsp;add&nbsp;the&nbsp;PrismJS syntax&nbsp;definition for&nbsp;the&nbsp;WebAssembly textual&nbsp;representation.

jwhitlock:
I'm going to prepare this PR for merging today.

jwhitlock:
Force-push with more packages updated from ``bower-installer``, plus add a commit comment.

PR #5192, opened 2019-01-15
Bug 1518844, UI updates to payments banner

schalkneethling:
Details of required changes can be seen here:
https://github.com/mdn/sprints/issues/709

My comments on some of the changes here:
https://github.com/mdn/sprints/issues/709#issuecomment-454416389

@davidflanagan r?

### DIFF
--- a/kuma/static/styles/components/payments/banner.scss
+++ b/kuma/static/styles/components/payments/banner.scss
@@ -139,6 +139,10 @@
             padding: 0;
             vertical-align: middle;
 
+            svg {
+                fill: #333;
+            }
+
             &:last-child {
                 margin-right: 0;
             }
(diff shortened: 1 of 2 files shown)


kuma/static/styles/components/payments/banner.scss
```diff
@@ -213,7 +217,7 @@
     }
 
     .login-prompt {
-        padding: 20px;
+        padding: 100px 20px;
```
davidflanagan: This gives you a large bottom padding, too... Would `100px 20px 20px 20px` be a better choice here?

schalkneethling: This is desired. The design calls for the container to keep a similar height when switching from the form to the sign-in prompt, so the additional space below is to ensure it is vertically centred inside the container.

PR #5196, opened 2019-01-16
Bug 1520616: require login for the revision dashboard

davidflanagan:
We believe that scrapers/crawlers hitting the revision dashboard
(while ignoring our robots.txt file) are causing significant
database load that is reducing performance of the site.

This patch changes the dashboards/revisions endpoint to require
login and to not be cached. The changes to kuma/dashboards/views.py
are trivial; most of the changes are adjustments to tests.

### DIFF
--- a/kuma/dashboards/tests/test_views.py
+++ b/kuma/dashboards/tests/test_views.py
@@ -74,23 +74,30 @@ def known_author(wiki_user):
     return wiki_user
 
 
-def test_revisions(root_doc, client):
-    """The revision dashboard works."""
-    response = client.get(reverse('dashboards.revisions'))
+def test_revisions_not_logged_in(root_doc, client):
+    """A user who is not logged in can't see the revisions dashboard."""
+    url = reverse('dashboards.revisions')
+    response = client.get(url)
+    assert response.status_code == 302
+    assert response['Location'] == '/en-US/users/signin?next={}'.format(url)
+    assert_no_cache_header(response)
+
+
+def test_revisions(root_doc, user_client):
+    """The revision dashboard works for logged in users."""
+    response = user_client.get(reverse('dashboards.revisions'))
     assert response.status_code == 200
-    assert 'Vary' in response
-    assert 'X-Requested-With' in response['Vary']
     assert 'Cache-Control' in response
-    assert_shared_cache_header(response)
+    assert_no_cache_header(response)
     assert 'text/html' in response['Content-Type']
     assert ('dashboards/revisions.html' in
             (template.name for template in response.templates))
 
 
-def test_revisions_list_via_AJAX(dashboard_revisions, client):
+def test_revisions_list_via_AJAX(dashboard_revisions, user_client):
     """The full list of revisions can be returned via AJAX."""
-    response = client.get(reverse('dashboards.revisions'),
-                          HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+    response = user_client.get(reverse('dashboards.revisions'),
+                               HTTP_X_REQUESTED_WITH='XMLHttpRequest')
     assert response.status_code == 200
     page = pq(response.content)
     rev_rows = page.find('.dashboard-row')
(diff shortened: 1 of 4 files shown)


davidflanagan:
For some reason the tests in tests/functional/dashboard_tests.py are super, super slow (more than a minute each) when we're logged in. There used to be just one of these, but now all of them require login. The tests still pass, but they just sit there for a minute or more while Chrome's loading indicator just spins and spins. I don't know why that is, but it does not look like a problem introduced by this patch.

kuma/dashboards/tests/test_views.py
```diff
@@ -74,23 +74,28 @@ def known_author(wiki_user):
     return wiki_user
 
 
-def test_revisions(root_doc, client):
-    """The revision dashboard works."""
+def test_revisions_not_logged_in(root_doc, client):
+    """A user who is not logged in can't see the revisions dashboard."""
     response = client.get(reverse('dashboards.revisions'))
+    assert response.status_code == 302
```
escattone: I'd vote for adding something like:
```python
assert 'en-US/users/signin?' in response['Location']
```
after this line, just to make it clear what's going on -- like we did [here](https://github.com/mozilla/kuma/blob/9589b848ff098cce4c4827fc15668e4bae4d3c5a/kuma/wiki/tests/test_views_admin.py#L37).

tests/functional/test_dashboard.py
```diff
@@ -11,8 +11,11 @@
 
 
 @pytest.mark.smoke
+@pytest.mark.login
```
escattone: Nice catch! It would have been easy to forget this!

tests/functional/test_dashboard.py
```diff
@@ -63,14 +73,19 @@ def test_dashboard_overflow(base_url, selenium):
...
+    admin.login_new_user()
     page = DashboardPage(selenium, base_url).open()
     page.open_first_details()
     assert page.scroll_width <= page.client_width
 
 
 @pytest.mark.nondestructive
+@pytest.mark.login
 @skip_if_not_maintenance_mode
 def test_dashboard_in_mm(base_url, selenium):
```
escattone: This test (`test_dashboard_in_mm`) you can delete now, since in maintenance mode all logins (even admin logins) are disabled and the test would fail.

escattone:
I also saw the same super slow test runs when running some of the tests in `tests/functional/dashboard_tests.py`, for both the Chrome and Firefox drivers, although the Firefox driver seemed faster. I reverted back to the master branch, re-ran the same tests, and again saw the same thing, so I'm sure it has nothing to do with your changes. It's probably worth investigating why those tests run so slowly sometime in the future.

kuma/dashboards/tests/test_views.py
ExE-Boss:
```suggestion
    assert response.status_code == 302
    assert 'en-US/users/signin?' in response['Location']
```

davidflanagan: Done

tests/functional/test_dashboard.py
davidflanagan: Thanks for noticing that. Done.

davidflanagan:
@escattone I've addressed your review comments on this.

kuma/dashboards/tests/test_views.py
```diff
@@ -74,23 +74,30 @@ def known_author(wiki_user):
...
 
-def test_revisions(root_doc, client):
-    """The revision dashboard works."""
-    response = client.get(reverse('dashboards.revisions'))
+def test_revisions_not_logged_in(root_doc, client):
+    """A user who is not logged in can't see the revisions dashboard."""
+    url = reverse('dashboards.revisions')
+    response = client.get(url)
+    assert response.status_code == 302
+    assert response['Location'] == '/en-US/users/signin?next={}'.format(url)
```
escattone: 😄 Even better!

PR #5197, opened 2019-01-17
Bug 1518844, ensure focus is clearly indicated on amount radio buttons

schalkneethling:
Problem is described here(although not very clearly ;p)
https://github.com/mdn/sprints/issues/713

In essence, the problem is that even though the radio buttons do receive focus, and is navigable using the arrow keys, it is not clear when they have focus.

Here is the current behaviour on production. As can be seen, when focus is moved from the email field to the radio buttons, nothing actually indicates that the focus has moved to the `checked` radio button. It feels like focus moves from the email field, to someplace on the form, and then to the custom amount field.

radio-fields-focus-current

Adding the `focus-visible` class with the clear focus ring solves the problem

radio-fields-focus

@davidflanagan r?

### DIFF
--- a/__tests__/sass/mdn/_a11y-helpers-mixins-test.scss
+++ b/__tests__/sass/mdn/_a11y-helpers-mixins-test.scss
@@ -1,0 +1,15 @@
+@include describe('focus-visible') {
+    @include it(
+        'Returns a box-shadow style that is applied to elements that currently has focus'
+    ) {
+        @include assert {
+            @include output {
+                @include focus-visible();
+            }
+
+            @include expect {
+                box-shadow: 1px 1px 5px #2196f3, -1px -1px 5px #2196f3;
+            }
+        }
+    }
+}
--- a/kuma/static/js/components/payments/focus-visible.js
+++ b/kuma/static/js/components/payments/focus-visible.js
@@ -2,6 +2,9 @@
     'use strict';
     var contributeForm = document.getElementById('contribute-form');
 
+    /**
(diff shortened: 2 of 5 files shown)


__tests__/sass/mdn/_a11y-helpers-mixins-test.scss
```diff
@@ -0,0 +1,15 @@
...
+    @include it(
+        'Returns a box-shadow style that is applied to elements that currently has focus'
+    ) {
+        @include assert {
+            @include output {
+                @include focus-visible();
+            }
+
+            @include expect {
+                box-shadow: 1px 1px 5px #2196f3, -1px -1px 5px #2196f3;
```
davidflanagan: Is there a way to write these tests that allows you to change the color or size of the shadow without breaking the test?

kuma/static/js/components/payments/focus-visible.js
```diff
@@ -12,6 +15,23 @@
         var currentTarget = event.target;
         var targetType = currentTarget.type;
 
+        /* The key pressed was the tab key, or the left or right
+           arrow keys(used to cycle through radio buttons), and the
+           target field type is one of `radio` */
+        if (
+            (event.key === 'Tab' ||
+                event.key === 'ArrowLeft' ||
+                event.key === 'ArrowRight') &&
```
davidflanagan: Nice catch for the arrow keys.

jwhitlock:
Sorry for the noise. I had to close and open the PR because our TravisCI integration was broken when this was opened.

PR #5198, opened 2019-01-17
bug 1518653: Update to django-ratelimit 2.0.0

jwhitlock:
* django-ratelimit 1.1.0 → 2.0.0: Fail open on missing cache, drop support for Django 1.10 and earlier, add support for Django 2.0 and 2.1. This will remove the deprecation warnings about using ``is_authenticated`` as a function rather than a property.


jwhitlock:
Sorry for the noise. I have to close and open the PR because our TravisCI integration was broken when this was opened.

PR #5199, opened 2019-01-17
Bug 1490727: Cancel subscriptions when a user is deleted

jwhitlock:
*This is an update of PR #5139 after rebasing on current master.*

When a User is deleted all subscriptions will be canceled. This uses the utility function ``cancel_stripe_customer_subscription``, which is also used when a User cancels their subscription subscription.

I changed a few other things as well, which could be broken up into more PRs if this one is too big:

* The utility functions ``get_stripe_customer_data`` and ``cancel_stripe_customer_subscription`` took a username and email, for the sole purpose of logging them if Stripe returned an error. I've removed the error handling from these functions and moved it to the ``recurring_payment_management`` view, which is the only function that uses these utilities.
* I've adjusted the error handling as well, to gracefully handle all Stripe API errors by logging them and acting like the user has no subscription. Previously any error, including syntax errors, would be treated the same way.
* Added error-handling tests for the view, and tests for the utility functions I touched.
* Fixed Django 1.11 deprecations and Python 3 issues

### DIFF
--- a/kuma/payments/constants.py
+++ b/kuma/payments/constants.py
@@ -1,2 +1,4 @@
+from __future__ import unicode_literals
+
 CONTRIBUTION_BETA_FLAG = 'contrib_beta'
 RECURRING_PAYMENT_BETA_FLAG = 'recurring_payment_beta'
--- a/kuma/payments/context_processors.py
+++ b/kuma/payments/context_processors.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import json
 
 from django.conf import settings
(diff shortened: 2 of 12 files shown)


jwhitlock:
I have to close and open the PR because our TravisCI integration was broken when this was opened. Sorry for the noise...

jwhitlock:
This is a payments PR, which @schalkneethling has been handling, but it is a lot of Python, so maybe @escattone is a better reviewer. There are some Python 3 changes needed to payments, so this may be a blocker for Python 3 work (or will need a rebase after).

schalkneethling:
@jwhitlock Yup, because of all the Python I felt that I am not the best person to review this. Would be great if @escattone has time to review this. Thanks!

jwhitlock:
OK back to you @escattone :) I still consider review and merging a low priority, since 1) user deletion is rare, and 2) we can manually cancel subscriptions if needed. @Atopal may give it a different priority.

atopal:
+1 to what John said. Obviously useful, but not urgent.

PR #5202, opened 2019-01-22
Bug 1521135, Stage A of codepen-iex implementation

schalkneethling:
A description of this can be found here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1521135

There is also another doc from @wbamberg that details the various stages. The PR addresses Stage A:
https://gist.github.com/wbamberg/ca85831ec166a0bfed4bbcc031d3bf73#stage-a-enable-mdn-codepen-integration

Here are a couple of screenshots showing the interaction:

On load:

screenshot 2019-01-22 at 12 56 41

When hovering over the button:

screenshot 2019-01-22 at 12 56 48

After clicking the button

screenshot 2019-01-22 at 12 57 00

I used `/en-US/docs/Web/API/Document/querySelector` for the example. To test, you need to add the following code to the document via CKEditor(This is also how the writers will add the examples for Stage A):

```
<div class="codepen-iex" data-default-tab="js,result" data-editable="true" data-height="400" data-prefill="">
<pre class="hidden" data-lang="html">
&lt;button class="add-para"
    type="button"&gt;
Add para
&lt;/button&gt;
  </pre>

<pre class="static-sample-code" data-lang="js">
const addParaButton = document.querySelector('.add-para');

addParaButton.addEventListener('click', () =&gt; {
  const para = document.createElement('p');
  para.textContent = 'All work and no play makes Jack a dull boy.';
  document.body.appendChild(para);
});
  </pre>
</div>
```

Because of some of the `data-` attributes required for the Codepen integrations, I also had to update the `constants.py` file with the following:

```
# attributes required for codepen-iex
ALLOWED_ATTRIBUTES['div'] += ['data-prefill', 'data-height', 'data-editable',
                              'data-default-tab']
ALLOWED_ATTRIBUTES['pre'] += ['data-lang']
```

I also included the minified source from Codepen. We could elect to instead load this from their CDN if that makes more sense.

@davidflanagan @jwhitlock @wbamberg r?

### DIFF
--- a/__tests__/sass/mdn/_button-states-mixins-test.scss
+++ b/__tests__/sass/mdn/_button-states-mixins-test.scss
@@ -1,0 +1,15 @@
+@include describe('active-button-state') {
+    @include it(
+        'Returns an active button style that is applied to elements that are interactive and has focus'
+    ) {
+        @include assert {
+            @include output {
+                @include active-button-state();
+            }
+
+            @include expect {
+                box-shadow: 1px 1px 5px #212121, -1px -1px 5px #bdbdbd;
+            }
+        }
+    }
+}
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -697,6 +697,12 @@ PIPELINE_CSS = {
         'output_filename': 'build/styles/mdn.css',
         'variant': 'datauri',
     },
+    'codepen-iex': {
+        'source_filenames': (
+            'styles/components/codepen-iex/codepen-iex.scss',
+        ),
+        'output_filename': 'build/styles/codepen-iex.css',
+    },
     'jquery-ui': {
         'source_filenames': (
             'js/libs/jquery-ui-1.10.3.custom/css/ui-lightness/jquery-ui-1.10.3.custom.min.css',
@@ -931,6 +937,12 @@ PIPELINE_JS = {
         'extra_context': {
             'async': True,
         },
+    },
+    'codepen-iex': {
+        'source_filenames': (
+            'js/components/codepen-iex/codepen-iex.js',
+        ),
+        'output_filename': 'build/js/codepen-iex.js',
     },
     'dashboard': {
         'source_filenames': (
--- a/kuma/static/js/components/codepen-iex/codepen-iex.js
+++ b/kuma/static/js/components/codepen-iex/codepen-iex.js
@@ -1,0 +1,21 @@
+(function() {
+    'use strict';
+    var sampleCodeContainer = document.querySelector('.codepen-iex');
+
+    /* only execute the rest of the code if there
+       is a `sampleCodeContainer` on the page */
+    if (sampleCodeContainer) {
+        var tryItButton = document.createElement('button');
+        tryItButton.classList.add('tryit');
+        tryItButton.textContent = gettext('Try it live!');
+        sampleCodeContainer.appendChild(tryItButton);
+
+        // when the button is clicked
+        tryItButton.addEventListener('click', function() {
+            // call codepen to load the editor
+            window.__CPEmbed('.codepen-iex');
+            // hide the button
+            tryItButton.classList.add('hidden');
+        });
+    }
+})();
--- a/kuma/static/styles/components/codepen-iex/codepen-iex.scss
+++ b/kuma/static/styles/components/codepen-iex/codepen-iex.scss
@@ -1,0 +1,23 @@
+@import '../../includes/vars';
+@import '../../includes/mixins';
+
+.codepen-iex {
+    position: relative;
+
+    .tryit {
+        position: absolute;
+        top: 10px;
+        right: 10px;
+        background-color: #fff;
+        border: 3px solid #3d7e9a;
+        height: 2rem;
+        font-size: 1rem;
+        cursor: pointer;
+
+        &:hover,
+        &:focus,
+        &:active {
+            @include active-button-state();
+        }
+    }
+}
--- a/kuma/wiki/constants.py
+++ b/kuma/wiki/constants.py
@@ -80,6 +80,10 @@ ALLOWED_ATTRIBUTES['del'] = ['datetime']
 ALLOWED_ATTRIBUTES['meter'] += ['max', 'min', 'value', 'low', 'high', 'optimum',
                                 'form']
 ALLOWED_ATTRIBUTES['details'] += ['open']
+# attributes required for codepen-iex
+ALLOWED_ATTRIBUTES['div'] += ['data-prefill', 'data-height', 'data-editable',
+                              'data-default-tab', 'data-theme-id']
+ALLOWED_ATTRIBUTES['pre'] += ['data-lang']
 # MathML
 ALLOWED_ATTRIBUTES.update(dict((x, ['encoding', 'src']) for x in (
     'annotation', 'annotation-xml')))
--- a/__tests__/sass/test.scss
+++ b/__tests__/sass/test.scss
@@ -6,3 +6,4 @@
 /* Tests */
 @import 'mdn/typography-mixin-test';
 @import 'mdn/a11y-helpers-mixins-test';
+@import 'mdn/button-states-mixins-test';
--- a/kuma/static/styles/includes/_mixins.scss
+++ b/kuma/static/styles/includes/_mixins.scss
@@ -149,6 +149,15 @@ Links
     }
 }
 
+/*
+Button states
+====================================================================== */
+/* Used for :hover, :active, :focus states
+   When changing the values of this mixin, remember to also update the
+   relevant test at __tests__/sass/mdn/_button-states-mixins-test.scss */
+@mixin active-button-state() {
+    box-shadow: 1px 1px 5px #212121, -1px -1px 5px #bdbdbd;
+}
 
 /*
 Colours
--- a/kuma/wiki/jinja2/wiki/base.html
+++ b/kuma/wiki/jinja2/wiki/base.html
@@ -16,6 +16,7 @@
     {{ super() }}
 
     {% stylesheet 'wiki' %}
+    {% stylesheet 'codepen-iex' %}
 
 {% endblock %}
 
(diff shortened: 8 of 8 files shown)


kuma/settings/common.py
```diff
@@ -932,6 +938,13 @@ def pipeline_one_scss(slug, **kwargs):
             'async': True,
         },
     },
+    'codepen-iex': {
+        'source_filenames': (
+            'js/components/codepen-iex/lib/ei.js',
```
davidflanagan: I don't feel strongly about whether we host this or pull it from a CDN. But we should ensure that it isn't part of the first load. I'm not sure how our jinja2 macros for scripts work, but hopefully we can use a script defer and/or script async for this and just make sure that the "Try it live" button is disabled by default until the script loads.

kuma/static/js/components/codepen-iex/codepen-iex.js
```diff
@@ -0,0 +1,21 @@
...
+    if (sampleCodeContainer) {
+        var tryItButton = document.createElement('button');
+        tryItButton.classList.add('tryit');
+        tryItButton.textContent = gettext('Try it live!');
+        sampleCodeContainer.appendChild(tryItButton);
+
+        // when the button is clicked
+        tryItButton.addEventListener('click', function() {
+            // call codepen to load the editor
+            window.__CPEmbed('.codepen-iex');
```
davidflanagan: I can't find any docs for this __CPEmbed() function. Do you know if you can pass the DOM element directly to it instead of passing the selector as a string? If so, then we'd be able to support more than one example on a page by doing querySelectorAll, looping through the array and setting up the event handler to embed that one particular element in the query result.  I suppose we may never need that, but it would be cool.  

Another approach would be to check for an id attribute on the container element and use that id as the selector instead of a class selector if it exists.

kuma/static/styles/components/codepen-iex/codepen-iex.scss
```diff
@@ -0,0 +1,21 @@
...
+        top: 10px;
+        right: 10px;
+        background-color: #fff;
+        border: 3px solid #3d7e9a;
+        height: 2rem;
+        font-size: 1rem;
+        cursor: pointer;
+
+        &:hover {
+            @include focus-visible();
```
davidflanagan: I'm not sure I agree with using a focus style for the hover state. Should we define a separate thing for hover, so that we can later decide to use different styling for focus and hover?

Also, if you're doing :hover, is it worth doing :active as well? Or does the <button> do something reasonable by default here?

kuma/static/styles/components/codepen-iex/codepen-iex.scss
```diff
@@ -0,0 +1,21 @@
...
+        }
+    }
+}
```
davidflanagan: Is the static-sample-code style already defined somewhere else?

schalkneethling:
> but the only one I feel strongly about is that we should be loading the iex script with defer so we don't impact page load time.

Agreed. I wanted to wait and see whether anyone has strong feelings regarding bundling this ourselves or loading from Codepen. The more I think about it, the more I am thinking I am going to simply load this from Codepen(with a defer).

My thinking is that this is not a repo type dependency where will be notified of updates. Unless the folks over at Codepen publishes a changelog that we can somehow easily follow, it is going to be hard to be sure we are not missing out on features, or even more importantly, security and general bug fixes.

kuma/static/styles/components/codepen-iex/codepen-iex.scss
schalkneethling: Yes, that is being taken care of sitewide via Prismjs

kuma/static/styles/components/codepen-iex/codepen-iex.scss
schalkneethling: Agreed. As this is an early prototype, I was a little lazy here ;p I am going to not be, and create a more effective `:hover/:focus/:active` style we can adopt for situations like these. Thanks for calling it out @davidflanagan

schalkneethling:
I was also thinking that we might `fetch` the script once we need it. We have some options here. Use the `jQuery` method of doing this while we are still in this world.

Just use `fetch` and include a polyfill. But this is something we can discuss as we move closer to the final implementation. Just wanted to make a note of it here.

kuma/static/js/components/codepen-iex/codepen-iex.js
schalkneethling: A couple of things here. First, we have agreed that there will never(I know never is a long time :)) have more than one of these on a page, so this should be fine.

At the moment the only documentation is here:
https://blog.codepen.io/documentation/prefill-embeds/

and more specifically here:
https://blog.codepen.io/documentation/prefill-embeds/#enhance-pre-blocks-into-editable-embeds-on-click-or-on-any-other-event-3

We could ask Chris directly though if this is possible, and it is worth doing so if this scenario seems likely. @wbamberg Thoughts?

schalkneethling:
Thanks for the review @davidflanagan - I am going to address the `hover` etc. state on the button, as well as the loading of the Codepen API and will update the PR.

schalkneethling:
Updated!

__tests__/sass/mdn/_button-states-mixins-test.scss
```diff
@@ -0,0 +1,16 @@
+@include describe('active-button-state') {
```
jwhitlock: Why is ``__tests__ `` (which looks super ugly but apparently is a React best practice?) a top-level folder?

davidflanagan: For KumaScript, I found that the other Jest convention was to name the tests as *.test.js, and it will find them even if they aren't in the ugly __tests__ folder. So that is something we could go with as well.

davidflanagan: Separate PR sounds right to me.

kuma/static/js/components/codepen-iex/codepen-iex.js
wbamberg: > First, we have agreed that there will never(I know never is a long time :)) have more than one of these on a page, so this should be fine.

It's not a requirement for now, but I think we might very well want to have > 1 of these on a page, at some point in the future.

__tests__/sass/mdn/_button-states-mixins-test.scss
```diff
@@ -0,0 +1,16 @@
...
+    @include it(
+        'Returns an active button style that is applied to elements that are interactive and has focus'
+    ) {
+        @include assert {
+            @include output {
+                @include active-button-state();
+            }
+
+            @include expect {
+                margin: 1px 0 0 1px;
```
wbamberg: I'm not a designer by any means, but I didn't love the little wobble it does on hover. But again, not a designer :).

kuma/wiki/constants.py
```diff
@@ -80,6 +80,10 @@
 ALLOWED_ATTRIBUTES['meter'] += ['max', 'min', 'value', 'low', 'high', 'optimum',
                                 'form']
 ALLOWED_ATTRIBUTES['details'] += ['open']
+# attributes required for codepen-iex
+ALLOWED_ATTRIBUTES['div'] += ['data-prefill', 'data-height', 'data-editable',
```
wbamberg: It would be good to allow `data-theme-id` as well. At some point we should look into different themes for this including custom ones.

wbamberg:
Thanks @schalkneethling for this PR!

It sort of works for me: I can click the button and when I do I see the CodePen embed, but it's empty (Firefox 63, clean profile):

screen shot 2019-01-23 at 2 29 46 pm

Perhaps this is just some quirk of my local setup?

Apart from that I had a couple of comments - the main one really is that we should allow the `data-theme-id` so we can experiment with themes.

wbamberg:
Yes, it does work in Chrome :/. But the demo (https://wbamberg.github.io/codepen-demo/demo.html) also works in a clean Firefox, which this does not.

schalkneethling:
> Yes, it does work in Chrome :/. But the demo ([wbamberg.github.io/codepen-demo/demo.html](https://wbamberg.github.io/codepen-demo/demo.html)) also works in a clean Firefox, which this does not.

Yeah, that is super weird. Curious to see what this does once on production.

wbamberg:
On production:
* it works in a clean Firefox and Chrome.
* it fails in my Firefox, due to Privacy Badger: if I configure PB to allow "static.codepen.io" and "codepen.io" then it works there too.

Test page: https://developer.mozilla.org/en-US/docs/User:wbamberg

Very nice!

schalkneethling:
> it fails in my Firefox, due to Privacy Badger: if I configure PB to allow "static.codepen.io" and "codepen.io" then it works there too.

I will notify Chris regarding this. Probably something they are going to want to look into. Thanks @wbamberg

kuma/static/js/components/codepen-iex/codepen-iex.js
chriscoyier: `window.__CPEmbed(".codepen-later");` is definitely a supported API, at the link @schalkneethling pointed to there. 

If you had more than 1 on a page, you could use different classes for each potentially? So they wouldn't be all triggered at once?

chriscoyier:
Both @shshaw tried fresh installs of Privacy Badger in Firefox/Chrome and it doesn't block codepen.io or static.codepen.io by default it seems like.

screen shot 2019-01-25 at 9 50 59 am

But certainly if those toggles are set to block, it'll block this.

wbamberg:
I think Privacy Badger "learns" to block trackers, so one person's badger experience may differ from anothers, and both may differ from a fresh install. I don't know how it learns or what about codepen.io makes PB think it's a potential tracker.

In my experience though it's quite common to have to teach PB to allow a domain, so I'd expect that PB users in general should be able to deal with this.

kuma/static/js/components/codepen-iex/codepen-iex.js
schalkneethling: Thanks for the feedback @chriscoyier - I agree that we should use different classes or, the clicked button needs to be context-aware so it knows which parent element to pass into the `__CPEmbed` function on click.

schalkneethling:
Thanks for all the testing and feedback regarding PB everyone. Looks like we should be good, but definitely something to keep an eye on as we user test this.

PR #5204, opened 2019-01-22
bug 1509430: Move to just-in-time KS timeout

jwhitlock:
Loading ``config.KUMASCRIPT_TIMEOUT`` as a function default requires database and cache access, which isn't available for all configurations, such as collecting localization data when building the production Docker images. Instead, use ``None`` and read the config in the function body.

Update ``cache_control`` to also use the None pattern, for consistency.

### DIFF
--- a/kuma/wiki/kumascript.py
+++ b/kuma/wiki/kumascript.py
@@ -38,8 +38,7 @@ def should_use_rendered(doc, params, html=None):
             (force_macros or (not no_macros and not show_raw)))
 
 
-def _post(content, env_vars, cache_control='',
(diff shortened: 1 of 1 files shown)


jwhitlock:
> I made some kind of considered decision when using '' as the default for cache_control, but I can't remember what it was, and I now think that I was just mistaken. cache_control has a default of None everywhere else, so changing it here is the correct thing to do.

I could see this as a way to communicate "This argument expects a string". Someday, we'll be on Python 3.5 or later, and can use [Optional](https://docs.python.org/3/library/typing.html#typing.Optional) to be clear about "This argument is a string or None".

PR #5206, opened 2019-01-22
bug 1520574: Remove file data from the KumaScript API

jwhitlock:
Remove file attachment data from the KumaScript API. This avoids creating a header that exceeds 8K bytes when a page has several attachments.

This API was used by two macros, [EmbedSVG](https://github.com/mdn/kumascript/blob/811f0f19b8acb6990ca94ee4f8e5334a65b975ba/macros/EmbedSVG.ejs#L6) and [Embed_text](https://github.com/mdn/kumascript/blob/811f0f19b8acb6990ca94ee4f8e5334a65b975ba/macros/Embed_text.ejs#L30-L35). @wbamberg has removed this macro from all pages on MDN Web Docs.

*Update*: Macros removed in https://github.com/mdn/kumascript/pull/1047

### DIFF
--- a/kuma/wiki/kumascript.py
+++ b/kuma/wiki/kumascript.py
@@ -78,14 +78,6 @@ def post(request, content, locale=settings.LANGUAGE_CODE):
     })
 
 
-def _get_attachment_metadata_dict(attachment):
-    current_revision = attachment.current_revision
-    return {
-        'filename': current_revision.filename,
-        'url': attachment.get_file_url(),
-    }
-
-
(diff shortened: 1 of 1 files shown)


jwhitlock:
> I'm happy to know that we can just remove the feature instead of trying to bump the max header size.

I agree. If this feature is needed again, I think it should either be a pull operation (add file data to the `$json` or a new API), or the `POST` to KumaScript should be changed to use `multipart/form-data` or `application/x-www-form-urlencoded`. I'm not in a rush to implement until we have a good use case.

PR #5208, opened 2019-01-24
bug 1518482: remove some docs from sitemap

escattone:
This PR represents a small part of the work related to [bug 1518482](https://bugzilla.mozilla.org/show_bug.cgi?id=1518482), but I wanted to have something to show after a frustrating day of trying to make sense of SEO and Google's new and old search consoles (e.g., the new search console doesn't seem to show the 38K+ errors that the old one does).

This PR removes 101 Google Bot indexing errors of the category `Submitted URL blocked by robots.txt`, which means the URL is included within the `sitemap` but excluded by `robots.txt`. The main lot here are the document URL's with a slug that starts with `Experiment:` (100 of them). The mysterious one is the single URL https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver/status (a redirect to a bugzilla page), which is not explicitly excluded by `robots.txt`, but the Google Search Console still complains.

### DIFF
--- a/kuma/wiki/managers.py
+++ b/kuma/wiki/managers.py
@@ -29,6 +29,7 @@ class BaseDocumentManager(models.Manager):
                     .exclude(slug__startswith='User_talk:')
                     .exclude(slug__startswith='Template_talk:')
                     .exclude(slug__startswith='Project_talk:')
+                    .exclude(slug__startswith='Experiment:')
                     .order_by('slug'))
         if locale:
             docs = docs.filter(locale=locale)


kuma/wiki/tasks.py
```diff
@@ -298,6 +298,17 @@ def build_locale_sitemap(locale):
...
 
     queryset = Document.objects.filter_for_list(locale=locale)
+    # Exclude any document whose slug starts with one of these prefixes:
+    #     - "Experiment:":
+    #         these documents are disallowed by robots.txt
+    #     - "Mozilla/QA/Marionette/WebDriver/status":
+    #         this document is a hard-wired redirect to an external domain, so
+    #         it can't be excluded as the content-based redirects are by the
+    #         "filter_for_list" method above
+    for slug_prefix in ('Experiment:',
```
jwhitlock: I think ``Experiment:`` should be moved into the default filters in ``filter_for_list``. In general, these URLs are implementation details for content swapping and shouldn't show up in list views like [document errors](https://developer.mozilla.org/en-US/docs/with-errors).

kuma/wiki/tasks.py
```diff
@@ -298,6 +298,17 @@ def build_locale_sitemap(locale):
...
+                        'Mozilla/QA/Marionette/WebDriver/status'):
```
jwhitlock: I wish there was a different way to do this. It seems ugly to special-case a single URL. Maybe we should remove the external redirect and instead have content that says "Moved to ``<bugzilla link>``"? But it works, so 🤷‍♂️

kuma/wiki/tasks.py
escattone: I had wondered about this myself, so I like it much better, thanks @jwhitlock!

kuma/wiki/tasks.py
escattone: I feel exactly the same, I don't like it. I initially thought I could change this into a content-based redirect before realizing that's not possible for external redirects.

escattone: Since the [entire set of Marionette-related documents that are now hard-redirected](https://github.com/mozilla/kuma/blob/c724ba2e3db2a47806410731639e308cd0da2636/kuma/redirects/redirects.py#L836) should probably be ignored in general, perhaps we should either add them to the default filters of `filter_for_list` or soft-delete them?

escattone:
Thanks for the review @jwhitlock. I've updated my commit for this PR, limiting it to the main lot of 100 document URL's with a slug that starts with `Experiment:`, and ignoring the one special case of `Mozilla/QA/Marionette/WebDriver/status` until we decide on a better approach for that one.

PR #5210, opened 2019-01-24
Bug 1522445, move Jest and True tests to existing tests directory

schalkneethling:
@jwhitlock r?


jwhitlock:
This is a little better. I prefer ``/tests/unit`` to ``__tests__``.

I'm deferring to @davidflanagan for a Jest sanity check. This might have an impact on integration tests, so @escattone may want to take a look.

schalkneethling:
This was just a quick move. We can definitely improve the overall structure of this once we start to add more tests. In the end, the tests should map directly to the components they are testing.

schalkneethling:
@davidflanagan @jwhitlock That *is* a better spot. Made me think about it a bit, and I thought then it would also make more sense to have the SaSS tests under `tests/sass`

Made the above changes. Let me know your thoughts. Thanks!

PR #5211, opened 2019-01-24
Bug 1522441, switch back to Github link from survey for BCD

schalkneethling:
I decided the keep the JS and SCSS files in the codebase. Seeing that it is very generic, and we are bound to run another survey for BCD at some point, this will make it much faster to get it up and running.

Also, the code is not loaded anywhere, so there is no negative performance impact to keeping it.

@davidflanagan r?

### DIFF
--- a/jinja2/includes/config.html
+++ b/jinja2/includes/config.html
@@ -31,8 +31,7 @@
             },
             js: {
                 'syntax-prism': [{% javascript 'syntax-prism' %}],
-                'wiki-compat-tables': [{% javascript 'wiki-compat-tables' %}],
-                'wiki-compat-table-survey': [{% javascript 'wiki-compat-table-survey' %}]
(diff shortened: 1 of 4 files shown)


jwhitlock:
We have very different philosophies on dead code @schalkneethling. We're currently collecting 2200 assets, then doubling that with hashed name. We end up using about 30% - 40% of assets in production. I hope that after the pipeline work we're using 95% or more of collected assets. 

Keep your eyes on these, they may end up on the chopping block in a future PR from me 😈

jwhitlock:
Approval from @atopal.

schalkneethling:
@jwhitlock Yeah, I get you ;) I was in two minds regarding this as well. It seemed innocent enough. There is always source control. I will not oppose the removal should we deem it necessary in future. 👍

PR #5215, opened 2019-01-25
Bug 1522718: render stale documents in series, not in parallel

davidflanagan:
Kuma's render_stale_documents task splits the list of stale
documents into groups of 5, then re-renders each group in
parallel. This typically causes about 12 requests to the KumaScript
server at the same time, causing load spikes that only occur
once an hour when this task runs.

This patch changes the task so that the subtasks for each group
of documents are chained and run in series instead of in parallel.
This should reduce the load on KumaScript and enable us to drop
down to a single KumaScript server.

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -6,7 +6,7 @@ import os
 import textwrap
 from datetime import datetime, timedelta
 
-from celery import chord, task
+from celery import chain, chord, task
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


davidflanagan:
You can see the sub-tasks all starting up at the same time here: https://papertrailapp.com/groups/10717942/events?q=%22document%20chunk%22

And you can see the hourly spikes on the KumaScript server here: https://rpm.newrelic.com/accounts/1807330/applications/72968481

PR #5217, opened 2019-01-25
bug 1518482: nofollow search pagination links

escattone:
This PR blocks well-behaved crawlers from following the `Previous`/`Next` pagination buttons at the bottom of the search results page. From some research into the 429 errors that the Google Bot experiences when crawling MDN (mainly from users running searches), this should significantly reduce the number of times the Google Bot is rate-limited by the `search` endpoint. Currently, it appears that the crawler follows the pagination links until the end of the pagination, and is often rate-limited.

### DIFF
--- a/kuma/search/jinja2/search/results.html
+++ b/kuma/search/jinja2/search/results.html
@@ -165,12 +165,12 @@
 
             <div class="pager">
               {% if previous %}
-                <a class="button" href="{{ previous }}" id="search-result-previous">
+                <a class="button" href="{{ previous }}" id="search-result-previous" rel="nofollow">
                 {{ _('Previous') }}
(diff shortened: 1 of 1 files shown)


jwhitlock:
I'm surprised that the search results page has ``<meta name="robots" content="noindex, follow">``. I was expecting ``nofollow``. It looks like ``noindex`` was added in [bug 1024798](https://bugzilla.mozilla.org/show_bug.cgi?id=1024798) 5 years ago, but without much SEO justification.

I'd like to get the opinion of our SEO expert before we make this change.

escattone:
Closing this PR, since our SEO expert recommended that we simply disallow the search endpoint via robots.txt.

PR #5221, opened 2019-01-29
bug 1518482: block search results from google search

escattone:
This PR disallows the search endpoint in `robots.txt` per recommendation of our SEO expert.

It has been tested via the [Google Search console's robots.txt Tester](https://www.google.com/webmasters/tools/robots-testing-tool?siteUrl=https://developer.mozilla.org/).

To test locally:
- Add `ALLOW_ROBOTS_WEB_DOMAINS=localhost:8000` to your `.env` file prior to `docker-compose up -d`
- Navigate to http://localhost:8000/robots.txt

### DIFF
--- a/kuma/landing/views.py
+++ b/kuma/landing/views.py
@@ -99,6 +99,7 @@ Disallow: /media
 Disallow: /*move-requested
 Disallow: /*preview-wiki-content
 Disallow: /*profiles*/edit
+Disallow: /*/search
 Disallow: /skins
 Disallow: /*type=feed
 Disallow: /*users/


kuma/landing/views.py
```diff
@@ -99,6 +99,7 @@ def promote_buttons(request):
 Disallow: /*move-requested
 Disallow: /*preview-wiki-content
 Disallow: /*profiles*/edit
+Disallow: /*/search
```
jwhitlock: This will block page like:

* https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/search
* https://developer.mozilla.org/en-US/docs/Web/Events/search
* https://developer.mozilla.org/en-US/docs/Web/API/URL/search

escattone: Good catch!

escattone: I think you're right @jwhitlock, the robots.txt language is too limited to single-out the search endpoint. I think we have to settle for updating the `meta` tag for the search page from what we have currently
```
<meta name="robots" content="noindex, follow">
```
to this
```
```
to get the same/similar effect.

escattone:
I can't see a way to single-out the search endpoint with the `robots.txt` language. It's too limited. I'm closing this, and I'll open a new PR (see https://github.com/mozilla/kuma/pull/5221#discussion_r251986991).

escattone:
Closed in favor of https://github.com/mozilla/kuma/pull/5223.

PR #5224, opened 2019-01-30
bug 1518482: disallow search in robots.txt

escattone:
This PR disallows the search endpoint in `robots.txt` per recommendation of our SEO expert. It is the rebirth of #5221, but with a new approach to disallowing the search endpoint in the `robots.txt` file such that valid document pages whose slug includes `/search` will not be included.

It has been tested using the [Google Search console's robots.txt Tester](https://www.google.com/webmasters/tools/robots-testing-tool?siteUrl=https://developer.mozilla.org/).

To test locally:
- Add `ALLOW_ROBOTS_WEB_DOMAINS=localhost:8000` to your `.env` file prior to `docker-compose up -d`
- Navigate to http://localhost:8000/robots.txt

### DIFF
--- a/kuma/landing/views.py
+++ b/kuma/landing/views.py
@@ -102,7 +102,8 @@ Disallow: /*profiles*/edit
 Disallow: /skins
 Disallow: /*type=feed
 Disallow: /*users/
-'''
+''' + '\n'.join('Disallow: /{locale}/search'.format(locale=locale)
+                for locale in settings.ENABLED_LOCALES)
 
 ROBOTS_GO_AWAY_TXT = '''\
 User-Agent: *


escattone:
Sorry for the new PR @jwhitlock, but GitHub wouldn't let me re-open #5221 after I force-pushed my branch. 😞

jwhitlock:
I spent some time looking for links to the search page that should get a ``rel=nofollow`` as well. The search boxes in the header and the front page are implemented as ``<form method="GET">``, and I think most search engines would ignore those.

The other interesting bit is our meta data pointing at search:

```
<link rel="search" type="application/opensearchdescription+xml" href="https://developer.mozilla.org/en-US/search/xml" title="MDN Web Docs">
```

This would block robots from accessing this URL. I think I'm OK with that? It may mean that users can still install it as a custom search provider. @a2sheppy is probably the biggest user.

jwhitlock:
Google's robots.txt tester says these are allowed:

* https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/search
* https://developer.mozilla.org/en-US/docs/Web/Events/search
* https://developer.mozilla.org/en-US/docs/Web/API/URL/search

The OpenSearchDescription XML is blocked:
* https://developer.mozilla.org/en-US/search/xml

I think we could change the URL in a new PR if this causes a problem for users.

PR #5228, opened 2019-02-01
bug 1467518: update kuma.core to support Python 3

escattone:
This PR adds support for Python 3 to the `core` app. It seems large but the vast majority of the changes are routine and easy to review. There are a few places that require explanation however:
- The three module-level functions of `kuma/core/managers.py` (`parse_tag_namespaces`, `allows_tag_namespace_for`, and `resolve_allowed_tags`) don't seem to be used anywhere (except for within themselves) and `resolve_allowed_tags` contains [this](https://github.com/mozilla/kuma/blob/ded7edc29de3f996636405156d9d5ed0b332f826/kuma/core/managers.py#L150) which needs to be modified to work with Python 3's iterable-view implementation of `keys()` (and there are no tests), so I've simply removed all of them.
- I've removed the context manager `safe_query_string` within `kuma/core/middleware.py`. It's not needed since `HttpResponsePermanentRedirect` already calls `iri_to_uri` on its path argument.
- Within `kuma/core/tests/test_validators.py`, I was careful to maintain the byte strings that `test_valid_javascript_identifier` is called with (since the function has conversion code for that case), but all of them probably don't need to be maintained that way.
- I decided to add `from __future__ import unicode_literals` to every file, even ones that really didn't need it like `kuma/core/exceptions.py` (to support the consistent addition of string literals to those files in the future). 
- @jwhitlock and I discussed the change to [this test](https://github.com/mozilla/kuma/blob/ded7edc29de3f996636405156d9d5ed0b332f826/kuma/core/tests/test_middleware.py#L40) over video.


kuma/core/managers.py
```diff
@@ -32,9 +34,10 @@ def __init__(self, *args, **kwargs):
         super(NamespacedTaggableManager, self).__init__(*args, **kwargs)
 
 
+@python_2_unicode_compatible
```
jwhitlock: Nice use of ``@python_2_unicode_compatible``

kuma/core/managers.py
```diff
@@ -112,59 +115,6 @@ def _ensure_ns(self, namespace, tags):
         return ns_tags
 
 
-def parse_tag_namespaces(tag_list):
```
jwhitlock: Kuma is weird. This appears to have been added 8 years ago, never used.

kuma/core/tests/test_validators.py
```diff
@@ -11,31 +13,31 @@ def test_valid_javascript_identifier(self):
         The function ``valid_javascript_identifier`` validates a given identifier
         according to the latest draft of the ECMAScript 5 Specification
         """
-        self.assertTrue(valid_javascript_identifier('hello'))
+        self.assertTrue(valid_javascript_identifier(b'hello'))
```
jwhitlock: It's so tempting to re-write this in the parametrized purest style. But I think this straightforward migration is best for getting Py3 done 👍

kuma/core/admin.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
```
jwhitlock: I like including this, it should help shake out any string issues before we switch.

PR #5229, opened 2019-02-01
bug 1467518: update kuma.dashboards to support Python3

escattone:
Minor changes to update `kuma.dashboards` to better support Python 3.

### DIFF
--- a/kuma/dashboards/forms.py
+++ b/kuma/dashboards/forms.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from django import forms
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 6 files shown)


jwhitlock:
@escattone the reason for the failing Python 2 tests was so dumb I had to add the commit to fix it. ``datetime.isoformat`` was throwing a ``TypeError`` when you pass it a unicode string in Python 2, or a byte string in Python 3. This was getting swallowed by the Constance machinery and returned as a "try again later please" with no data.

The call was ``now.isoformat(' ')``, to use a space rather than `T` to separate the date from the time. Luckily, ``str(now)`` does the same thing, and picks the right string type for Python 2 and 3.

escattone:
@jwhitlock Nice catch, thanks!

PR #5230, opened 2019-02-01
bug 1467518: update kuma.feeder to better support Python3

escattone:
This PR updates the `kuma.feeder` app to better support Python3. The only changes here provide a string experience that is more consistent between Python2 and Python3.

### DIFF
--- a/kuma/feeder/tests/test_utils.py
+++ b/kuma/feeder/tests/test_utils.py
@@ -25,39 +25,39 @@ HACKS_PARSED = FeedParserDict(
         # Omited attributes: author_detail, authors, content, guidislink,
         # links, summary_detail, tags, title_detail, comments, slash_comments,
         # wfw_commentrss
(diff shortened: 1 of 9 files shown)


kuma/feeder/tests/test_utils.py
```diff
@@ -143,7 +143,7 @@ def test_fetch_feed_unchanged(hacks_feed, mocked_parse, enabled):
     """If the ETag matches, the feed is enabled but not processed."""
     hacks_feed.last_modified = datetime(2018, 2, 26, 21, 23, 38)
     hacks_feed.etag = 'W/"1da1fc6a456fd49c32a9291b38ec31ee-gzip"'
```
jwhitlock: This line that stood out to me, so I double-checked. In Python 3, ``feedparser.parser`` returns with a string ``etag``, not a byte stream, so we should be OK. There may be other issues, but not this :)

PR #5219, opened 2019-01-26
Update ja.sccs to enable italic font style back

hmatrjp:
The reason for introducing font-style: normal !important is now (in 2019) obsolete, as wrote in line 48.

PR #5236, opened 2019-02-04
Update ja.sccs to enable italic font style back

hmatrjp:
The reason for introducing font-style: normal !important is now obsolete, as wrote in line 49.

Now I hope I'm doing this right, after #5219

### DIFF
--- a/kuma/static/styles/locales/ja.scss
+++ b/kuma/static/styles/locales/ja.scss
@@ -44,14 +44,8 @@ Styles for Japanese
        local('IPA P Gothic');
 }
 
-/* Bug 973171 */
-/* stylelint-disable declaration-no-important */
-* {
-    /* !important required for locale specific override */
-    font-style: normal !important;
-}
-
 /* Bug 1522385 */
+/* stylelint-disable declaration-no-important */
 .syntaxbox {
     var,
     em {


kuma/static/styles/locales/ja.scss
```diff
@@ -46,9 +46,10 @@ Styles for Japanese
 
 /* Bug 973171 */
 /* stylelint-disable declaration-no-important */
+/** Back ground of Bug 973171 is obsolete already, the fix was for Windows XP, lacks Anti-Alias function **/
 * {
     /* !important required for locale specific override */
-    font-style: normal !important;
+    /* font-style: normal !important; */  /*  !important prevents vars outside of syntaxbox to be italic  */
```
schalkneethling: Instead of commenting out the code, perhaps it is better to simply remove it from the codebase altogether?

hmatrjp: It is OK for me, revision comment in github shows the reason of the change for the later references.

schalkneethling: Ok, if you can make the change and push an update to this PR, I will test it out locally. Thanks @hmatrjp

hmatrjp: Oh!, did I replied on wrong place!? It is shown as a " ... reviewed ..."
 I modified the change to ja.sccs following the advice from @schalkneethling . Thanks.

kuma/static/styles/locales/ja.scss
```diff
@@ -46,10 +46,8 @@ Styles for Japanese
 
 /* Bug 973171 */
 /* stylelint-disable declaration-no-important */
-* {
-    /* !important required for locale specific override */
-    font-style: normal !important;
-}
+/** Back ground of Bug 973171 is obsolete already, the fix was removed. 
+    It was for Windows XP, lacks Anti-Alias function **/
```
schalkneethling: @hmatrjp Thank you for the update. I believe you can safely remove this comment as well. The bug and this pull request are there to keep the history. I will test this locally and let you know if I see any other problems. Thank you!

hmatrjp: Hi,
did you finish the test locally? @schalkneethling 
Or should I remove the comment first?

schalkneethling: You can safely remove the comment. Thx @hmatrjp

hmatrjp: I removed comment. Thanks. @schalkneethling

jwhitlock:
Force-push to rebase on current master, squash commits, expand and improve the commit message, and re-add the ``stylelint`` exception for ``!important``.

I'm not sure if the override is needed for the syntax box styles, but I'm leaving it in place for now, since @schalkneethling didn't mention it.

jwhitlock:
This works locally on a copy of https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from:

pr 5236

Thanks @hmatrjp!

hmatrjp:
Thanks a lot! @jwhitlock @schalkneethling

PR #5241, opened 2019-02-08
Bug 1525719: create a non-production /ducks/ endpoint for the React spike

davidflanagan:
This patch adds a new /ducks/ url just like our existing /docs/ url
for wiki documents, except it has a duck in the header. The URL route
is not installed if we're in production so this should be safe. We
are going to use this new url for our experimentation with React.

### DIFF
--- a/kuma/urls.py
+++ b/kuma/urls.py
@@ -26,6 +26,7 @@ from kuma.search.urls import (
 from kuma.users.urls import lang_urlpatterns as users_lang_urlpatterns
 from kuma.wiki.admin import purge_view
 from kuma.wiki.urls import lang_urlpatterns as wiki_lang_urlpatterns
+from kuma.wiki.urls import react_document_urlpatterns as wiki_react_urlpatterns
 from kuma.wiki.views.document import as_json as document_as_json
 from kuma.wiki.views.legacy import mindtouch_to_kuma_redirect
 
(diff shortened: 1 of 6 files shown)


davidflanagan:
screen shot 2019-02-08 at 12 24 01 pm

davidflanagan:
The react_*.html templates files are copies of existing templates, barely changed, except for the addition of ducks.

The react_document() method in document.py is a copy of document(), except with a different template file.

The interesting changes to be reviewed are in kuma/urls.py and kuma/wiki/urls.py

davidflanagan:
I improved the logo!

screen shot 2019-02-08 at 1 19 40 pm

kuma/urls.py
```diff
@@ -89,6 +90,12 @@ def serve_from_media_root(request, path):
 urlpatterns += i18n_patterns(url(r'^docs.json$', document_as_json,
                                  name='wiki.json'))
 urlpatterns += i18n_patterns(url(r'^docs/', include(wiki_lang_urlpatterns)))
+
+# A special endpoint for our React spike, local and staging only
+if settings.DOMAIN != settings.PRODUCTION_DOMAIN:
```
jwhitlock: I double-checked, this will work correctly - it will show up in staging and development, but will be a 404 in production.

jwhitlock:
@davidflanagan you may want to remove the CodePen stuff that was removed in https://github.com/mozilla/kuma/pull/5243

davidflanagan:
Thanks for catching that, John. Done.

PR #5245, opened 2019-02-11
bug 1527010: add home page to each locale sitemap

escattone:
This PR adds the home page to each enabled locale's sitemap (e.g., https://developer.mozilla.org/en-US/ is added to https://developer.mozilla.org/sitemaps/en-US/sitemap.xml), and refactors/improves the sitemap tests.

### DIFF
--- a/kuma/wiki/tasks.py
+++ b/kuma/wiki/tasks.py
@@ -18,6 +18,7 @@ from djcelery_transactions import task as transaction_task
 from lxml import etree
 
 from kuma.core.decorators import skip_in_maintenance_mode
+from kuma.core.urlresolvers import reverse
 from kuma.core.utils import chunked
 from kuma.search.models import Index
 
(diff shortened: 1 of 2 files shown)


jwhitlock:
@escattone I'll take a look at this tomorrow or Wednesday. 

I'm curious if it would be less changes to create a new sitemap for homepages and non-wiki pages, named something like ``sitemaps_other.xml``, rather than try to inject the homepage into the locale-based wiki sitemaps, and worry about sitemap size limits.

escattone:
@jwhitlock Yes, great point, I think it would be, and I'm embarrassed that I blindly went down the more complicated rabbit. 🤦‍♂️ I'll work-up an alternative.

escattone:
I'm closing this in favor of #5246. Thanks for the simplification @jwhitlock.

PR #5246, opened 2019-02-11
bug 1527010: add home page to each locale sitemap

escattone:
This PR replaces #5245. It adds the home page to each enabled locale's sitemap (e.g., https://developer.mozilla.org/en-US/ is added to https://developer.mozilla.org/sitemaps/en-US/sitemap.xml), and refactors/improves the sitemap tests.

### DIFF
--- a/kuma/wiki/tests/test_tasks.py
+++ b/kuma/wiki/tests/test_tasks.py
@@ -1,12 +1,13 @@
 from __future__ import with_statement
 
 import os
+import re
 from datetime import datetime
 
-from django.conf import settings
-
+from kuma.core.urlresolvers import reverse
 from kuma.users.models import User
 from kuma.users.tests import UserTestCase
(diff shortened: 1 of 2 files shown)


kuma/wiki/tests/test_tasks.py
```diff
@@ -1,49 +1,70 @@
...
-        with open(sitemap_path, 'r') as sitemap_file:
-            index_xml = sitemap_file.read()
-        for loc in expected_sitemap_locs:
-            assert loc in index_xml
+def test_sitemaps(tmpdir, settings, doc_hierarchy):
+    """
+    Test the build of the sitemaps.
+    """
+    settings.SITE_URL = 'https://example.com'
+    settings.MEDIA_ROOT = str(tmpdir.mkdir('media'))
```
jwhitlock: Nice! 😍 We've been doing it the wrong way for years...

PR #5247, opened 2019-02-11
Bug 1521693: escape angle brackets in redirect link text

davidflanagan:
Our mdn-redirect plugin for CKEditor was creating HTML links without
escaping the angle brackets and ampersands in the link text, which
caused HTML validity problems when trying to create redirect links to
MDN pages that include angle brackets in their title.

This patch replace &, < and > characters with entities.


davidflanagan:
I only changed the last file shown here. All of the other changes were made by the CKEditor build process. I was tempted to just check out the old versions of all the CSS files since they don't need to be modified, but https://kuma.readthedocs.io/en/latest/ckeditor.html says that those changes are expected, so I figured it was safer to just leave them alone.

schalkneethling:
Thanks @davidflanagan - This works for me:

```
'This title includes a & a < and a >, what is the output'.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
"This title includes a &amp; a &lt; and a &gt;, what is the output"
'This title includes a & a </ and a />, what is the output'.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
"This title includes a &amp; a &lt;/ and a /&gt;, what is the output"
```

PR #5248, opened 2019-02-12
Bug 1522998: fix the CKEditor save buttons in maximized mode

davidflanagan:
The CKEditor save buttons are enabled and disabled depending
on whether any of the content has changed. The JavaScript to do this
uses a jQuery find() function to locate a ".editor-container" element
within a form, and then dispatches dirty or clean events on that
container element.

This is fine in the normal case, but if the editor is maximized
then the DOM hierarchy changes and there is no .editor-container.
The find() call returns an empty array and the events are dispatched
on the zero elements in that array.

This patch adds a special case: when no container element is found,
we dispatch the required events directly on the form (which is where
the handlers for those events are installed.)

### DIFF
--- a/kuma/static/js/wiki-edit.js
+++ b/kuma/static/js/wiki-edit.js
@@ -565,11 +565,29 @@
         // Keep track of editor dirtiness
         function checkEditorDirtiness() {
             var editorDirty = editor.checkDirty();
+            var $container = $form.find('.editor-container');
 
-            if (editorDirty) {
-                $form.find('.editor-container').addClass('dirty').trigger('mdn:dirty');
+            if ($container.length) {
+                if (editorDirty) {
+                    $container.addClass('dirty').trigger('mdn:dirty');
+                } else {
+                    $container.removeClass('dirty').trigger('mdn:clean');
+                }
             } else {
-                $form.find('.editor-container').removeClass('dirty').trigger('mdn:clean');
+                /*
+                 * When the editor is maximized, the container doesn't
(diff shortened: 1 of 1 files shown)


kuma/static/js/wiki-edit.js
```diff
@@ -565,11 +565,27 @@
...
+                }
             } else {
-                $form.find('.editor-container').removeClass('dirty').trigger('mdn:clean');
+                // When the editor is maximized, the container doesn't
+                // exist in the form, but we still need to fire mdn:dirty
+                // and mdn:clean events to enable and disable the save
+                // buttons in the editor. Fortunately, those events are
+                // handled on the form, and we can trigger the events
+                // directly on the form in this case. In this case we don't
+                // bother with adding and removing the 'dirty' class.
```
schalkneethling: Nit: I would suggest using JSDoc multiline comment style here:

```
/* When the editor is maximized, the container doesn't
   exist in the form, but we still need to fire mdn:dirty
   and mdn:clean events to enable and disable the save
   buttons in the editor. Fortunately, those events are
   handled on the form, and we can trigger the events
   directly on the form in this case. In this case we don't
   bother with adding and removing the 'dirty' class. */
```

davidflanagan: Done

schalkneethling: Thanks!

schalkneethling:
I just need to pull this one down and test it locally, but the code looks good. Thanks @davidflanagan

PR #5249, opened 2019-02-12
bug 1468491: Update package-lock "requires" for npm 6, merge to devDependencies

jwhitlock:
npm 6.4.1, included with node 10.14.2, uses loose version identifiers in the requires list. The exact versions installed are still specified, so nothing changes. However, since it would confuse the output of real package changes, I'm adding as an independent PR.

For example, ``@babel/highlight`` used to look like this:

```json
    "@babel/highlight": {
      "version": "7.0.0",
      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0.tgz",
      "integrity": "sha512-UFMC4ZeFC48Tpvj7C8UgLvtkaUuovQX+5xNWrsIoMG8o2z+XFKjKaN9iVmS84dPwVN00W4wPmqvYoZF3EGAsfw==",
      "dev": true,
      "requires": {
        "chalk": "2.4.1",
        "esutils": "2.0.2",
        "js-tokens": "4.0.0"
},
```

but now it looks like this:

```json
    "@babel/highlight": {
      "version": "7.0.0",
      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0.tgz",
      "integrity": "sha512-UFMC4ZeFC48Tpvj7C8UgLvtkaUuovQX+5xNWrsIoMG8o2z+XFKjKaN9iVmS84dPwVN00W4wPmqvYoZF3EGAsfw==",
      "dev": true,
      "requires": {
        "chalk": "^2.0.0",
        "esutils": "^2.0.2",
        "js-tokens": "^4.0.0"
},
```

The ``"requires"`` section now uses loose version, like ``"chalk": "^2.0.0"``, instead of strict versions like ``"chalk": "2.4.1"``, to reflect how the requires are specified in the ``babel/highlight`` package.

However, chalk still has an unchanged strict version:

```json
"chalk": {
  "version": "2.4.1",
```

So, although the diff is huge, the actual installed versions of packages have not changed. See [npm issue 20434](https://github.com/npm/npm/issues/20434#issuecomment-402988364) for more information.

PR #5250, opened 2019-02-12
bug 1380419: Add react and webpack requirements

jwhitlock:
This builds on #5249, and ``package-lock.json`` may be easier to review after merging that.

This adds a Django package:

* django-webpack-loader 0.6.0 - Reads the bundles JSON and determines the URLs for Django / Jinja2 templates\ 

This adds some npm packages:

* react 16.8.1: JS library for creating user interfaces
* webpack 4.29.3: Bundle JS files, and transform other assets
* webpack-bundle-tracker 0.4.2-beta: Create JSON describing bundling, so that django-webpack-loader can determine the URLs.

It also moves all the existing npm packages from ``dependencies`` to  ``devDependencies``. These were split (by me) to separate packages required for building assets from those that happened to be used in other processes. However, this doesn't appear to be a useful distinction as we move forward. Since none will be needed for production (just creating production images), I picked moving them all to ``devDependencies``. This required some changes when building the images and in TravisCI.

These new packages aren't built yet, but I find it easier to write and review the code that uses these new libraries when they are in the master branch, images on DockerHub, etc. I'm roughly following [this blog article](https://owais.lone.pw/blog/webpack-plus-reactjs-and-django/) from 2015, but I'm not sure if we'll need Babel or not, so I left it out.

One milestone will be to replace ``gulp`` with ``webpack``, but both will be needed until that is complete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,24 +20,22 @@
         "precommit": "npm run eslint && npm run stylelint",
         "test": "jest"
     },
-    "dependencies": {
+    "devDependencies": {
         "clean-css-cli": "4.2.1",
         "eslint": "5.1.0",
         "eslint-plugin-jest": "22.1.2",
-        "node-sass": "4.9.4",
-        "stylelint": "9.3.0",
-        "uglify-js": "2.8.29"
-    },
-    "devDependencies": {
         "gulp": "3.9.1",
         "gulp-sass": "4.0.1",
         "gulp-stylelint": "7.0.0",
         "gulp-watch": "5.0.0",
         "husky": "0.14.3",
         "jest": "23.6.0",
+        "node-sass": "4.9.4",
         "react": "16.8.1",
         "sass-true": "4.0.0",
+        "stylelint": "9.3.0",
         "svgo": "1.0.5",
+        "uglify-js": "2.8.29",
         "webpack": "4.29.3",
         "webpack-bundle-tracker": "0.4.2-beta"
     }
--- a/package.json
+++ b/package.json
@@ -35,7 +35,10 @@
         "gulp-watch": "5.0.0",
         "husky": "0.14.3",
         "jest": "23.6.0",
+        "react": "16.8.1",
         "sass-true": "4.0.0",
-        "svgo": "1.0.5"
+        "svgo": "1.0.5",
+        "webpack": "4.29.3",
+        "webpack-bundle-tracker": "0.4.2-beta"
     }
 }
--- a/docker/images/kuma_base/Dockerfile
+++ b/docker/images/kuma_base/Dockerfile
@@ -114,7 +114,7 @@ WORKDIR /tools
 USER kuma
 RUN ln -s /app/package.json /tools \
     && ln -s /app/package-lock.json /tools \
-    && npm install --production
+    && npm install
 USER root
 RUN find /tools/node_modules/.bin/ -executable -type f -o -type l -exec ln -s {} /usr/local/bin/ \;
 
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -20,22 +20,23 @@
...
-    },
-    "devDependencies": {
         "gulp": "3.9.1",
         "gulp-sass": "4.0.1",
         "gulp-stylelint": "7.0.0",
         "gulp-watch": "5.0.0",
         "husky": "0.14.3",
         "jest": "23.6.0",
+        "node-sass": "4.9.4",
+        "react": "16.8.1",
```
davidflanagan: I don't think there is ever a case where you use react on the web without react-dom, so you might as well add that now, if you're adding react.  Also, you can't really use react without Babel, so consider adding that now.

Actually, why is react a dev dependency? When we start actually using it, it will be a regular dependency, and react-dom will be required too. Maybe you've got it here for some other reason, I guess.

I've got a wip branch with this set of devDependencies. I don't know if that is my final set or if it will need tweaks:

```
    "devDependencies": {
        "@babel/core": "^7.2.2",
        "@babel/plugin-proposal-class-properties": "^7.3.0",
        "@babel/preset-env": "^7.3.1",
        "@babel/preset-flow": "^7.0.0",
        "@babel/preset-react": "^7.0.0",
        "@emotion/babel-preset-css-prop": "^10.0.7",
        "babel-loader": "^8.0.5",
        "babel-plugin-emotion": "^10.0.7",
        "flow-bin": "^0.92.1",
        "gulp": "3.9.1",
        "gulp-sass": "4.0.1",
        "gulp-stylelint": "7.0.0",
        "gulp-watch": "5.0.0",
        "husky": "0.14.3",
        "jest": "23.6.0",
        "prettier": "^1.16.4",
        "sass-true": "4.0.0",
        "svgo": "1.0.5",
        "webpack": "^4.29.3",
        "webpack-cli": "^3.2.3"
    }
```

jwhitlock: This looks like a good list to me, and includes some things I'm missing like ``webpack-cli``.

What are your ``dependencies``?

jwhitlock: Also, I'm on my first day with ``react``, so that's why it is in dev dependencies.

jwhitlock:
The more I look at this, the more I think @davidflanagan should get the commit that brings in ``webpack`` and friends. We can bring in ``django-webpack-loader`` when needed.

I think it still would be useful to move ``gulp`` etc. to ``devDependencies``, and to get the npm updates in before the code that uses them.

package.json
davidflanagan: Here are my dependencies, though this is from a branch that doesn't have running code yet: 

```
    "dependencies": {
        "@emotion/core": "^10.0.7",
        "@emotion/styled": "^10.0.7",
        "clean-css-cli": "4.2.1",
        "eslint": "5.1.0",
        "eslint-plugin-jest": "22.1.2",
        "node-sass": "4.9.4",
        "react": "^16.8.1",
        "react-dom": "^16.8.1",
        "stylelint": "9.3.0",
        "uglify-js": "2.8.29"
    },
```

ExE-Boss: I&nbsp;don't really feel like react should be&nbsp;added just yet in&nbsp;this&nbsp;PR, given the&nbsp;discussion in&nbsp;https://github.com/mdn/sprints/issues/967.

ExE-Boss:
```suggestion
```

jwhitlock:
This PR won't be merged for the reasons I gave above:

> The more I look at this, the more I think @davidflanagan should get the commit that brings in `webpack` and friends. We can bring in `django-webpack-loader` when needed.

@davidflanagan's upcoming PR will actually use Webpack and React, so his package choices are more relevant than mine, which are guesses. It was useful to see what each of us are considering, but that can be done on a closed PR as easily as an open one.

I'm closing to reduce confusion, but we're still going to bring in React. Investigating React, and how it will integrate with Kuma, is the major goal of this sprint. See [ADR 001](https://github.com/mdn/mdn/blob/master/ADRs/001-use-react.md) for the reasoning.

PR #5251, opened 2019-02-12
bug 1505084: Use node 10 in TravisCI

jwhitlock:
This may require clearing the caches in TravisCI, otherwise packages compiled for node 8 may mess up the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,8 +69,8 @@ matrix:
             INSTALL_ELASTICSEARCH=1
             INSTALL_NPM_TOOLS=1
 install:
-    - nvm install 8
-    - nvm use 8
+    - nvm install 10
+    - nvm use 10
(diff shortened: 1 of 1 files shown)


jwhitlock:
This eventually worked, but it required clearing the TravisCI caches, to ensure that node 8 packages were not used. It can pick up the cache from the master branch or the PR-specific branch. If we merge this, we may need to clear the cache a few more times as PRs are updated.

The master branch cache was over 1000 MB, and this PR cache is now 72 MB. Maybe clearing caches is something to do anyway.

jwhitlock:
Too often I'm waiting for TravisCI before moving on to the next task. It would be painful for me to have to wait 11 minutes for the Docker build, versus the 3 minutes for ``assetlint`` and 4 minutes for ``locales``.

PR #5252, opened 2019-02-13
Bug 1522937, test three scenarios for hiding IEx

schalkneethling:
The testing scenarios are described here:
https://github.com/mdn/sprints/issues/921#issuecomment-460967275

Pages for testing can be copied from production. These are:

* https://developer.mozilla.org/en-US/docs/User:schalkneethling/mobile/box-shadow-no-iex
* https://developer.mozilla.org/en-US/docs/User:schalkneethling/mobile/box-shadow-display-none
* https://developer.mozilla.org/en-US/docs/User:schalkneethling/mobile/box-shadow-display-none-empty-src

This is temporary while we gather some performance metrics. Once @atopal has sufficient data, we will back-out the script.

@davidflanagan r?

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1004,6 +1004,7 @@ PIPELINE_JS = {
             'js/utils/utils.js',
             'js/utils/post-message-handler.js',
             'js/wiki.js',
+            'js/utils/bug1522937-iex-test.js',
             'js/interactive.js',
             'js/wiki-samples.js',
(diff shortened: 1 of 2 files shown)


atopal:
Test is set up on speedcurve. Will run for 3 days: https://speedcurve.com/mozilla-add-ons/mdn/favorite/?d=7&db=47831&dc=2&de=1&ds=1

PR #5254, opened 2019-02-14
Bug 1525719: react spike step 2

davidflanagan:
This PR sets up React, Babel, Webpack and Flow and adds a trivial
Logo.jsx component. It does not affect any production endpoints
but will allow us to get a baseline performance measurements on
the staging server.


davidflanagan:
This is working for me locally, but I kind of expect that there will be test failures or build failures that prevent us from deploying it to staging easily... Just seems like the kind of change that might take a couple of tries to get right.

kuma/javascript/src/logo.jsx
```diff
@@ -0,0 +1,20 @@
+//@flow
+import * as React from 'react';
```
schalkneethling: Curious, what is the benefit/difference between this and `import React from "react";`?

kuma/javascript/src/logo.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function(props : {|url: string|}) : React.Node {
+    return (
+        <a css={{margin:0}} className="logo" href={props.url}>
```
schalkneethling: This outputs as `[object Object]` in the DOM. Is there a reason we cannot:

1. not use the inline styling
2. have it present in sass (perhaps this will cause problems when deployed to production)
3. use a class

package.json
```diff
@@ -24,20 +26,38 @@
...
+        "@babel/plugin-proposal-class-properties": "^7.3.0",
+        "@babel/preset-env": "^7.3.1",
+        "@babel/preset-flow": "^7.0.0",
+        "@babel/preset-react": "^7.0.0",
+        "@emotion/babel-preset-css-prop": "^10.0.7",
+        "babel-loader": "^8.0.5",
+        "babel-plugin-emotion": "^10.0.7",
         "clean-css-cli": "4.2.1",
         "eslint": "5.1.0",
         "eslint-plugin-jest": "22.1.2",
```
schalkneethling: I believe we may need a couple of additional `eslint` plugins, see:
https://github.com/schalkneethling/mdn-feature-banner-component/blob/master/package.json#L40

jwhitlock:
The builds are failing because in the docker image the node packages are installed to ``/tools/node_modules``. This avoids issues with using a macOS ``node_modules`` inside of a Debian image.

I have a version of webpack.config.js that works, but we'll see if you like it or not... I'll submit a PR to your branch.

kuma/javascript/src/logo.jsx
davidflanagan: Doing it that way imports flow types like React.Node that I use below. The alternative is to have one import for the React module and then a separate 'import type' line (or lines) for the types.

I tend to favor separate type imports, but this seems like it might be an idiomatic thing that we could just get in the habit of doing so we always have React.Node available... 

This seems like the sort of thing we'll want to have a consistent style for eventually.

kuma/javascript/src/logo.jsx
davidflanagan: Oops! That must mean I don't have emotion hooked up right.

Note that this is not inline styling. If I get it working right, it will output a stylesheet with a class name.

It obviously isn't really necessary to use this here, but it was supposed to force me to get the Emotion plugin working, so I'd like to keep it as a proof of concept.

Thanks for catching that the output was off!

package.json
davidflanagan: Thanks for that link! I got flow working for this PR, but forgot about eslint. I'll add the extra plugins.

kuma/javascript/src/logo.jsx
```diff
@@ -0,0 +1,20 @@
...
+          MDN Web Docs
+        </a>
+    );
+
+    /*
+      Or, to get the ducks back:
```
jwhitlock: I'll miss the ducks 🦆

kuma/settings/common.py
```diff
@@ -528,6 +528,7 @@ def _get_locales():
 
 STATICFILES_DIRS = (
     path('kuma', 'static'),
+    path('kuma', 'javascript', 'dist'),
```
jwhitlock: This works for now. I'd also like to consider ``/assets/src`` and ``/assets/dist`` as we migrate away from ``django-pipeline``, and move stuff out of ``/kuma/static``.

kuma/javascript/src/logo.jsx
schalkneethling: Just pulled down the latest, but I am still seeing `object Object`

schalkneethling:
@davidflanagan It looks like we do not yet have a `.babelrc` config. I have this on the mdn-featured-banner` component:
https://github.com/schalkneethling/mdn-feature-banner-component/blob/master/.babelrc

I also notice from the Emotion docs that for the `css` prop you need the following in `.babelrc`

```
{
  "presets": ["@emotion/babel-preset-css-prop"]
}
```

The other options seems to be adding the following to the file that uses the `css` prop:

```
/** @jsx jsx */
import { jsx } from '@emotion/core'
```

kuma/javascript/src/logo.jsx
unclechu: If you have to use `style` attribute and it is static (as we see here) you could just separate this object outside a component-function to some variable and use that variable as a value for `style` attribute, in this case virtual-dom could be just checked for object-equality when some new props being received instead of producing new object every render and deeply checking all the values inside that style-object.

webpack.config.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = {
+    mode: "production",
+    entry: path.resolve(__dirname, './kuma/javascript/src/index.jsx'),
+    output: {
+        filename: 'react.js',
+        path: path.resolve(__dirname, './kuma/javascript/dist/')
+    },
+    module: {
+        rules: [{
+            "test": /\.(js|jsx)$/,
```
schalkneethling: We could change this to the following:

```suggestion
            test: /\.jsx?$/,
```

davidflanagan: Thanks, Schalk. I *thought* I'd put the babel config in package.json. I wonder if I lost it while fixing a merge confict or something?  My intent is to be using that emotion plugin. It also gives us class names that have meaninful names instead of completely random strings, which will be nice.

davidflanagan: Ah. The babel config is inside the webpack config. I've added the pretter preset, and am going to keep babel inside webpack, unless we come up with some reason to separate them.

kuma/javascript/src/logo.jsx
davidflanagan: For reasons I did not figure out, the inilne CSS styles I was using to cover the dino head with the duck do not work when I try to duplicate them in react. Since I failed to get the ducks back, I just removed the entire comment.

kuma/javascript/src/logo.jsx
davidflanagan: Schalk: that's because I'd been distracted and hadn't fixed it yet. It works now, though.

Unclechu: I'm purposely avoiding using inline styles. But that is a really good point about creating the objects outside of the render function.

package.json
```diff
@@ -39,18 +40,25 @@
...
+        "eslint-plugin-prettier": "^3.0.1",
+        "eslint-plugin-react": "^7.12.4",
         "flow-bin": "^0.92.1",
         "gulp": "3.9.1",
         "gulp-sass": "4.0.1",
         "gulp-stylelint": "7.0.0",
         "gulp-watch": "5.0.0",
         "husky": "0.14.3",
-        "jest": "23.6.0",
+        "jest": "^24.1.0",
```
schalkneethling: @davidflanagan I tend to lean towards locking dependencies dow a specific version using `--save-exact` when installing. It seems we now have a mixture of locked down versions and "variable" versions. What are your thoughts on this?

I use Renovate on most of my other repos, not sure if we have the option of using it here. https://github.com/marketplace/renovate

schalkneethling:
Thanks, @davidflanagan - One open question regarding dependency versioning. This is looking great!

Also, Travis is complaining about the following:

```
ERROR in ./kuma/javascript/src/index.jsx
Module build failed (from /tools/node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/plugin-proposal-class-properties' from '/app'
```

webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+                "loader": "babel-loader",
+                "options": {
+                    "presets": [
+                        "@babel/preset-env",
+                        "@babel/preset-react",
+                        "@babel/preset-flow",
+                        "@emotion/babel-preset-css-prop"
+                    ],
+                    "plugins": [
+                        "@babel/plugin-proposal-class-properties"
```
schalkneethling: Should this be a plugin inside `.babelrc` instead?

schalkneethling: https://github.com/schalkneethling/mdn-feature-banner-component/blob/master/.babelrc#L17

webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    module: {
+        rules: [{
+            "test": /\.jsx?$/,
+            "exclude": /node_modules/,
+            "use": {
+                "loader": "babel-loader",
+                "options": {
+                    "presets": [
+                        "@babel/preset-env",
```
jwhitlock: This and other ``presets`` (and possibly ``plugins``) need a prefix to work in the docker container:

```
nodePath + "/@babel/preset-env"
```
``nodePath`` resolves to ``./node_modules`` when run normally, and ``/tools/node_modules`` in the Docker container.

The ``tools`` directory was used to avoid an issue where node always looks in a local ``.node_modules`` folder first, and this may contain packages built on a macOS or Windows system, which are not compatible with the Debian system in the Docker container.

I wish we had something like Python's virtual environments, so we could instead tell node "don't looks at ``./node_modules``, look here", but as the [old npm FAQ](
https://web.archive.org/web/20150216184318/https://docs.npmjs.com/misc/faq#node_modules-is-the-name-of-my-deitys-arch-rival-and-a-forbidden-word-in-my-religion-can-i-configure-npm-to-use-a-different-folder), this is baked into node and the maintainers aren't planning to ever change it. ``NODE_PATH`` might be part of the solution, but I think even then ``./node_modules`` wins.

You can verify it works locally with:
```
VERSION=latest make build-base build-kuma
```

If this works, then TravisCI and production should work too.

davidflanagan: John: I took the nodePath stuff out because it was causing my local webpack runs to fail, and I couldn't see why it would be necessary since you also have it set in the resolve and resolveLoader properties of the config.

I can the local webpack again; maybe I was just doing something wrong. But I've never seen a config file that explicitly includes the node_modules/ directory when specifying a dependency, so I'm hoping we can find a different way to make the config work in both contexts.

webpack.config.js
```diff
@@ -1,9 +1,23 @@
...
+// But we also have to convince webpack and babel to use the /tools
+// directory when we're running in docker, so this nodePath constant
+// is the absolute path to the node_modules director we want to use.
+// We set it on resolve.modules below, and prepend it to the name of
+// all the babel plugins so we're sure we're gettting precisely the
+// file we want both locally (NODE_PATH not set) and in docker.
+//
+// TODO: there ought to be a better way to do this.
+//
+const nodePath = process.env.NODE_PATH || path.join(__dirname, 'node_modules');
```
jwhitlock: 👍``path.join`` feels more Pythonic. But this is node.js. Not sure if they have a word for that

webpack.config.js
```diff
@@ -0,0 +1,53 @@
+const path = require('path');
+
+// When running in docker, we install libraries into /tools/node_modules
+// so that we don't overwrite the user's local kuma/node_modules directory
+// with linux versions of modules (like the binary newrelic module).
+// To make this work, the docker container sets the NODE_PATH environment
+// variable so that code can be found in /tools/node_modules.
+// But we also have to convince webpack and babel to use the /tools
+// directory when we're running in docker, so this nodePath constant
+// is the absolute path to the node_modules director we want to use.
```
jwhitlock:
```suggestion
// is the absolute path to the node_modules directory we want to use.
```

webpack.config.js
```diff
@@ -0,0 +1,53 @@
...
+// We set it on resolve.modules below, and prepend it to the name of
+// all the babel plugins so we're sure we're gettting precisely the
+// file we want both locally (NODE_PATH not set) and in docker.
+//
+// TODO: there ought to be a better way to do this.
+//
+const nodePath = process.env.NODE_PATH || path.join(__dirname, 'node_modules');
+
+module.exports = {
+    mode: "production",  // Or switch to "development"
```
jwhitlock: This could be loaded from the ``process.env`` as well 😁 . Maybe ``webpack --mode=development`` works too, https://webpack.js.org/concepts/mode/

PR #5257, opened 2019-02-14
bug 1380419: Document static assets

jwhitlock:
This is an attempt to document how assets are processed in development and production. My goal was to understand the system enough that we can safely remove parts that are no longer needed.

I thought assets were complex before, but they are even more complex than I thought. This is reflected in a long document that jumps back and forth. Still, I think some changes, like removing ``django-pipeline`` and avoiding collecting assets not served to users, can simplify the system and increase performance.

I'm working to get this branch built on readthedocs.org, which will make it much easier to read.


jwhitlock:
Rendered version at https://kuma.readthedocs.io/en/static-asset-docs-1380419/assets.html.

I'll re-read in the morning and find some more obvious typos, but reviews are welcome.

jwhitlock:
I made some minor changes, and I'm done with my tweaks to this document.

jwhitlock:
I've marked everyone for review, but I'll probably take the first positive one as my go-ahead for merging.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+The CKEditor build process is documented on the
+:doc:`CKEditor document</ckeditor>`. The built files are checked into the
+Kuma repository.
+
+.. _include-js-libraries:
+
+Including JS libraries
+======================
+Third-party JavaScript libraries are included in the Kuma repository.
+Some were added manually, and others with Bower_. See
```
davidflanagan: But the Bower ones are still checked in, right? (I don't remember running any kind of Bower step when setting up my repo...)

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+* `kuma/static/js/analytics.js`_
+* `kuma/static/js/main.js`_
+* `kuma/static/js/components/nav-main-search.js`_
+* `kuma/static/js/auth.js`_
+* `kuma/static/js/highlight.js`_
+* `kuma/static/js/wiki-compat-trigger.js`_
+* `kuma/static/js/lang-switcher.js`_
+
+In development, the source files (10 for ``main.js``) are served, so there are
+10 ``<script>`` elements in the HTML when when ``{{javascript('main')}}`` is
```
davidflanagan: "when when"

Also, I think it would be really helpful to link to kuma/settings/common.py to specify where these bundles are defined.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+* `kuma/static/js/components.js`_
+* `kuma/static/js/analytics.js`_
+* `kuma/static/js/main.js`_
+* `kuma/static/js/components/nav-main-search.js`_
+* `kuma/static/js/auth.js`_
+* `kuma/static/js/highlight.js`_
+* `kuma/static/js/wiki-compat-trigger.js`_
+* `kuma/static/js/lang-switcher.js`_
+
+In development, the source files (10 for ``main.js``) are served, so there are
```
davidflanagan: Is it worth specifying here (perhaps with just a link) how the build process knows whether it is in development or production? As I read this that is a detail that I do not remember.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+Building pipeline CSS bundles
+=============================
+Pipeline CSS bundles are conceptually similar to
+:ref:`Pipeline JS Bundles <pipeline-js-bundles>`. Some contain multiple
+source files, such as `static/build/styles/dashboards.css`_,
+which combines:
+
+* `kuma/static/styles/dashboards.scss`_
+* `kuma/static/styles/diff.scss`_
+
```
davidflanagan: I like that you give specific bundle examples here and above. I'd suggest again that you link to the common.py file to specify where this bundle is defined.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+   For example, the :ref:`localization JavaScript <build-locale-js>`
+   ``build/locale/jsi18n/fr/javascript.js`` will be collected to
+   ``static/jsi18n/fr/javascript.js``.
+``STATICFILES_STORAGE``
+   The dotted path to a class implementing ``staticfiles``
+   :ref:`Storage <staticfiles-storage>`.  Storage determines where files
+   are stored, what URLs they have, and provides hooks for modifying files when
+   copying them.  Kuma uses three different storages, depending on the context:
+
+   * Development server (``DEBUG=False``): pipeline.storage.NonPackagingPipelineStorage_,
```
davidflanagan: I think you have the DEBUG values reversed here and below.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+   * Production server (``DEBUG=True``): kuma.core.pipeline.storage.ManifestPipelineStorage_,
+     which combines packaged files, hashes the names, and creates a manifest.
+   * Testing (``pytest``, TravisCI, etc.) and ``make`` commands: pipeline.storage.PipelineStorage_,
+     which combines packaged files but does not hash the names.
+
+.. _django.contrib.staticfiles: cs.djangoproject.com/en/1.11/ref/contrib/staticfiles/
+
+.. _staticfiles-finder:
+
+Finder classes
```
davidflanagan: I don't have any objections to the sections here and below on Finder and Storage classes, but I'm not sure why I need to understand things at this level of detail. Maybe I'm missing something. Is there a configuration setting somehwere that controls which finders and storage classes are used in which circumstances? Perhaps I'd understand this better if it was clear which parts are under our control and which are just the things that staticfiles and django-pipeline do for us by default.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+.. _SASS: https://sass-lang.com/
+
+.. _gulp-alternative:
+
+Compiling and collecting assets with Gulp
+=========================================
+An alternate way to compile and collect assets is to use Gulp, as described in
+:ref:`compiling-with-gulp`. This requires installing node and related packages
+on the "host" system, rather than relying on the Docker containers, but it
+matches the preferred workflow of some front-end developers.
```
davidflanagan: I'm not sure I every really appreciated that gulp is a front-end alternative to collectstatic. So if I'm understanding correctly, we've got different build workflows for development, testing, and production, and also, frontend and backend developers have different versions of the development flow.

There is probably some kind of diagram that could summarize this. I'm not suggesting that you take any time to create such a diagram, however.

docs/assets.rst
jwhitlock: Yes, Bower was used to download the versions specified in https://github.com/mozilla/kuma/blob/master/bower.json, which were then checked in. It takes a while for Mozilla to trust package managers.

docs/assets.rst
jwhitlock: I've tried to draw it three times 😁 Maybe the fourth will look sane...

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+Kuma uses Pontoon_ to translate strings in the user interface, in error
+messages, and in emails. These are stored in the mdn-l10n_ repository,
+and included as a `git submodule`_ at ``locale/``.  See the
+:doc:`localization document </localization>` for more details about locales.
+
+Puente_ extracts strings to the Portable Object Template (``.pot``) files. The
+file ``locale/templates/LC_MESSAGES/django.pot`` contains strings from template
+files and Python code. The file ``javascript.pot`` contains strings from
+JavaScript files. Puente looks for the string parameters of ``gettext``
+functions, such as ``gettext()``, the common alias ``_()``, and ``ngettext()``.
```
schalkneethling: Should this be the underscore alias?

schalkneethling: Oh wait nevermind, I read comma ;p

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+* **Collecting** - Built assets are collected to the ``/static`` folder.
+    - Django's :ref:`staticfiles <django-staticfiles>` provides a framework
+      of :ref:`Finder <staticfiles-finder>` and
+      :ref:`Storage <staticfiles-storage>` classes for collecting files.
+    - django-pipeline_ augments ``staticfiles``, to allow compiling, bundling,
+      and minifiying JS and CSS assets in the collection phase.
+    - :ref:`gulp <gulp-alternative>` provides alternative tools to ``staticfiles``
+      for local development of assets.
+* **Serving** - Collected assets are served to visitors in development and production
+    - The :ref:`static tag <static-tag>` returns the URL of assets
```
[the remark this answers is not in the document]
jwhitlock: "tag" is the name given to this in both Django and Jinja, not macros. I'll add the phrase "In [Django and Jinja templates](http://jinja.pocoo.org/docs/2.10/templates/#)," at the front for context.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+files and Python code. The file ``javascript.pot`` contains strings from
+JavaScript files. Puente looks for the string parameters of ``gettext``
+functions, such as ``gettext()``, the common alias ``_()``, and ``ngettext()``.
+It also parses longer strings in the template tag ``trans``.
+
+Next the changes are merged into the existing Portable Object (``.po``) files,
+such as ``locale/fr/LC_MESSAGES/django.po``, to add new strings and comment out
+removed strings.
+
+Extracting and merging is done with ``make localeextract``, usually
```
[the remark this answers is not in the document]
jwhitlock: If you use ``_()`` or ``_gettext()``, it will be trivial to [configure Puente](https://puente.readthedocs.io/en/latest/installation.html#configure) to extract those strings as well.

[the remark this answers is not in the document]
jwhitlock: I'm not sure if this belongs in the doc. I'll describe it, and maybe you'll have a suggestion how to document it?

[Puente](https://github.com/mozilla/puente) is a Python library that handles l10n things for Django projects using Jinja2 templates. It provides the ``extract`` management command. [Management commands](https://docs.djangoproject.com/en/1.11/howto/custom-management-commands/) are discovered in the ``management/commands/`` folder of Django apps, so [puente/management/commands/extract.py](https://github.com/mozilla/puente/blob/master/puente/management/commands/extract.py) is the handler for ``./manage.py extract``.

Like most well-written management commands, the code for the ``extract`` command is mostly focused on setting up command line arguments, extracting configurations, and calling another command. In this case, [extract_command](https://github.com/mozilla/puente/blob/master/puente/commands.py#L67-L129). This allows the function that does the work to be [tested](https://github.com/mozilla/puente/blob/master/tests/test_extract.py) more easily.

The code is a wrapper around the [Babel](http://babel.pocoo.org/en/latest/) library, which is an integrated collection of utilities that assist in internationalizing and localizing Python applications, with an emphasis on web-based applications. Puente uses the [Catalog](http://babel.pocoo.org/en/latest/api/messages/catalog.html#catalogs) to represent a gettext message catalog, and [extract_from_dir](http://babel.pocoo.org/en/latest/api/messages/extract.html#babel.messages.extract.extract_from_dir) to extract strings from files.

docs/assets.rst
```diff
@@ -0,0 +1,673 @@
...
+Django includes a `JavaScriptCatalog view`_ that provides JavaScript
+implementations of gettext_ functions, as well as translations for each
+locale. It is ineffecient to use this view directly, since it is generated
+on access. For efficiency, django-statici18n_ generates files for each locale
+from the ``JavaScriptCatalog`` output, so they can be served as static assets.
+
+The translation catalog files are created with ``make compilejsi18n``
+from the locale Machine Object ``.mo`` files.  Kuma sets ``STATICI18N_ROOT``
+to ``build/locale``, and the output files have names like
+``build/locale/jsi18n/de/javascript.js``.
```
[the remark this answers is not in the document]
jwhitlock: The link to [JavaScriptCatalog view](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#module-django.views.i18n) goes to the documentation. There is a [[source] link](https://docs.djangoproject.com/en/1.11/_modules/django/views/i18n/#JavaScriptCatalog) that goes to the Django source for the class.

The [compilejsi18n management command](https://github.com/zyegfryed/django-statici18n/blob/master/src/statici18n/management/commands/compilejsi18n.py), which isn't as nice as Puente, calls the Django views to get the outputs of those views, and writes them to files.

In our [current configuration](https://github.com/mozilla/kuma/blob/0405221dff45e77dfb2df6cb6081470e15be562e/kuma/settings/common.py#L651-L666), we have one ``django`` catalog for Python and templates, and one ``javascript`` catalog for all the JavaScript.

How much of that should be added to the docs?

docs/assets.rst
jwhitlock: Some of this is just "I had to understand this, so now you need to as well".

Our configuration points are what finders we use (and what order), which storage we use (just one at a time), and how we configure them. We also need to treat this as something we control, rather than something that we work around.

A lot of this is necessary because ``django-pipeline`` uses the framework to do "magic", and you have to understand the framework to know why we have at least 3 different configurations. I hope to get rid of ``django-pipeline``, and then a lot of this may become irrelevant.

docs/assets.rst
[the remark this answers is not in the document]
jwhitlock: I think writing it out made it clearer to me how to update the docs.

I've avoided linking to the configuration files, because the links would be out of date when the code is updated. After some thought, I agree links to code would be more useful than making the reader find the code, even if out of date. I'm going to try adding the links with a 🍟 emoji, to signal to the reader (and maintainers) that the link is awesome if fresh, and awful if stale.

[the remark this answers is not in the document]
jwhitlock: The french fry emoji 🍟 works, but draws too much attention to itself. I'll just have to trust that people know they might be looking at old code.

docs/assets.rst
jwhitlock: The short answer is ``DEBUG=True`` or ``=False``, but it is quite a bit more complicated, because ``django-pipeline`` is too much magic. I'll move the links higher in the paragraph.

docs/assets.rst
jwhitlock: Nope, failed the fourth time as well. I'm going to concentrate on removing ``django-pipeline`` so there is no more variety or magic.

docs/assets.rst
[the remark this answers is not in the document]
davidflanagan: LOL at the frenchfries! I was making the assumption that the localeextract command was defined in the kuma source itself, which is why I asked for a pointer. Sorry.

jwhitlock:
A interesting side effect of documentation (and testing) is that you notice bad code designs. Rather than continue polishing the docs, I'm going to switch to improving the design, which should make the documentation better as well.

PR #5258, opened 2019-02-16
Bug 1525719: Prebuild React.js components as static HTML

unclechu:
This is actually a fork of this PR: https://github.com/mozilla/kuma/pull/5254
In some issue about React I volunteered to help with server-side rendering.
Here I implemented pre-rendering React components independently for each locale and storing results as static HTML partials.


davidflanagan:
@unclechu I'm embarassed that I didn't notice that you had opened this PR. I haven't been ignoring it; I just didn't see that it was here. I'll try to take a look soon.

kuma/javascript/src/index.jsx
```diff
@@ -0,0 +1,10 @@
+// @flow
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Logo from './Logo';
+
+const container = document.getElementById('main-header-logo');
+
+if (container) {
+    ReactDOM.render(<Logo url={container.dataset.url}/>, container);
```
davidflanagan: With SSR, shouldn't this become a hydrate() call instead of a render() call? Or am I missing something?

react-prebuild.jsx
```diff
@@ -0,0 +1,25 @@
...
+import {renderToString} from 'react-dom/server';
+import Logo from './kuma/javascript/src/Logo';
+
+let stdinAccumulator = '';
+
+process.stdin.on('data', input => {
+    stdinAccumulator += input.toString();
+});
+
+process.stdin.on('end', () => {
```
davidflanagan: I like how simple you've made the Python <=> Node communication here.

react-prebuild.jsx
```diff
@@ -0,0 +1,25 @@
...
+    const locales = JSON.parse(stdinAccumulator);
+
+    const result = Object.keys(locales).reduce((acc, locale) => ({
+        [locale]: {
+            Logo: renderToString(<Logo
```
davidflanagan: I think I'd be inclined to not use JSX in this file, and use React.createComponent() (or whatever the function is called) that the JSX compiles too.

react-prebuild.jsx
```diff
@@ -0,0 +1,25 @@
...
+                url={locales[locale].Logo.props.url}
```
davidflanagan: I suspect that this could be generalized to just `{...locales[locale].Logo.props}` so we just pass in whatever props are passed to us.

react-prebuild.jsx
```diff
@@ -0,0 +1,25 @@
...
+            />),
+        },
+        ...acc
```
davidflanagan: Object interpolation in a loop is a stealth way to get O(n^2) performance, so I avoid this pattern. (I also don't like using reduce() because I find it hard to understand.) But those are my style preferences. I can see that it would be easy to rewrite this loop to avoid those things.

react-prebuild.jsx
```diff
@@ -0,0 +1,25 @@
...
+    }), {});
+
+    console.log(JSON.stringify(result, null, 2))
```
davidflanagan: I'm guessing that it would be useful to have a try/catch block above and print errors to stderr so that they can be reported by the management command. And maybe set th eprocess exit code when an error occurs? (The point is that I'd want to be sure that the build process fails if SSR fails.)

davidflanagan:
@unclechu: thank you for this! Since I've never set up SSR before, it is really helpful to have a working example to show how it can be done. I particularly like how simple you've made it to make the python build process work with Node. I'm not going to merge this as it stands... I want to get some non-trivial React code merged before thinking about SSR. But I'm certain that I'll be referring back to this PR when we are ready to incorporate SSR.  

I'm surprised that your code does not call hydrate() anywhere... I thought that was the client-side part of the SSR process. Am I misunderstanding something?

Also: the commits that you based this on have been merged now, so if you have a chance to rebase this against master it would make it easier to see which parts are specific to your SSR code.  (The logo.jsx component changed in my branch, so your changes to that will have merge conflicts.)

And here's another question for you @unclechu... Do you find that it is typically necessary to render each locale separately as is done here? Or is that just because of the way I structured the Logo component to take a URL that depends on the URL? Right now the template is calling `url('home')` to generate that URL, but it would be easy to port that from python to javascript so that we compute the home url on the client side. Then no prop would be needed, and a single pre-rendered would work for all locales.  But if it is normal for projects to have to pre-render for each locale, then it is good for us to plan on having to do that from the start, I suppose.

@jwhitlock: this PR from a new contributor shows how we could set up SSR. See in particular kuma/core/management/commands/prebuild_react.py and react-prebuild.jsx to see how he has hooked up a Django management command to a Node process that does the SSR. (I'm not sure if we want to get Django involved in that part of the build process or just run the SSR when we run webpack. Right now because of the locale dependency, I suppose it is necessary to get Django involved, but maybe we can avoid that?)

jwhitlock:
It will be next week before I can take a look. Now that PR #5254 is merged, a rebase on master would help make it clearer what the additions are.

kuma/javascript/src/index.jsx
unclechu: @davidflanagan This part isn't touched at all on SSR side, it's for frontend only. This module isn't even imported in SSR.

unclechu:
> I'm surprised that your code does not call hydrate() anywhere... I thought that was the client-side part of the SSR process. Am I misunderstanding something?

We don't need it since that module (`index.jsx`, the main entrance to the app) isn't called from SSR part, only particular components modules are (well, for now it's only one single component/module `Logo`). So, that `render` is for frontend only.

unclechu:
@davidflanagan I'm going to rebase it over `master` branch and make some changes you requested soon.

unclechu:
> Do you find that it is typically necessary to render each locale separately as is done here? Or is that just because of the way I structured the Logo component to take a URL that depends on the URL?..

> ..Right now the template is calling `url('home')` to generate that URL, but it would be easy to port that from python to javascript so that we compute the home url on the client side. Then no prop would be needed, and a single pre-rendered would work for all locales. But if it is normal for projects to have to pre-render for each locale, then it is good for us to plan on having to do that from the start, I suppose.

1. Correct, it is done like that because a component may depend on locale-dependent prop;
2. Yes, it would be better to port i18n dictionary to make localization directly from JS but separately for each language (e.g. load JSON of localization dictionary of specific locale) but we still have to produce renders for each locale since we need static HTML files, otherwise we could render it dynamically by permanently running node.js process which would handle dynamic props and dynamic locales but I've been told you prefer to avoid that;
3. We could use some redux.js store and create store also on SSR side and build a whole page with a main root-component which would render another component inside, etc.

unclechu:
For now it's only single partial component, as I understand it's just a beginning of rewriting the whole app and rendering whole app from united single entry point.

jwhitlock:
I took a look at rebasing this, and I think the project has moved on enough that it would be easier to start over with the current master. The focus of this PR was to show how to render the logo on the server, and the current master has dropped that code path. We've also split the babel configuration into babel.config.js, and it was getting painful to move the webpack.config.js changes over there.

I think we'll be looking at server-side rendering later this month, so no rush. But I'm adding a "not ready" tag, to signal that this PR isn't a candidate for review and merging.

unclechu:
I'm sorry but I'm currently extremely busy, so I don't have time to work more on this PR in next month or maybe even next few months.

davidflanagan:
We landed an SSR server and a Jinja template tag that uses it in https://github.com/mozilla/kuma/pull/5362 so I'm closing this PR because it will no longer be needed.

@unclechu thank you again for creating this PR. It was very helpful to me when I was working on PR #5362.

PR #5259, opened 2019-02-19
bug 1380419: Move to plain assets to /assets

jwhitlock:
Move assets that are served without processing (other than hashing names) to ``/assets``:

* [x] ``/assets/static`` - Files served to visitors
* [x] ``/assets/legacy`` - Files that appear unused by current code, and are candidates for deletion
* [x] ``/assets/reference`` - Files that are unused but that we want to keep for future reference or manual processing
* [x] Remove statically-served cloned pages, now that speed measurements are taken ([bug 1503923](https://bugzilla.mozilla.org/show_bug.cgi?id=1503923)).
* [x] Remove SCL3 maintenance mode files, currently in ``/media/maintenance`` and ``/kuma/static`` ([bug 1445638](https://bugzilla.mozilla.org/show_bug.cgi?id=1445638)).

I've also added a README with some planned future file moves, but I think this will be a new PR:

* ``/assets/build``: Output of ``gulp`` and ``make compilejsi18n``, copied to
  ``static/build``. As we convert from ``gulp`` to ``webpack``, this folder will just be output of ``make compilejsi18n``. Not checked into repo.
* ``/assets/ckeditor4`` The source and build folders for CKEditor. ``ckeditor/build`` is copied to ``static/js/libs/ckeditor/build``. This is a candidate for a submodule, if we liked submodules.
* ``/assets/dist``: Output of the ``webpack`` process, copied to ``static/dist``
* ``/assets/lib``: Third-party libraries that served to users. As we convert to ``webpack``, these will switch to being installed with ``npm``.
* ``/assets/pipeline``: Input of the ``gulp`` process. It should be possible to convert these to ``webpack`` soon.
* ``/assets/src``: Input for the ``webpack`` process.

I have some more work, so using the **not ready** label. It looks like [draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/) isn't available yet.

I'd like any feedback from @davidflanagan or @schalkneethling if this looks like it will destroy your workflow or plans.


assets/README.md
```diff
@@ -0,0 +1,29 @@
...
+for the details.
+
+Here are the subfolders:
+
+* **build**: (TODO) Output of ``gulp`` and ``make compilejsi18n``, copied to
+  ``static/build``.
+* **ckeditor**: (TODO) The source and build folders for CKEditor.
+  ``ckeditor/build`` is copied to ``static/js/libs/ckeditor/build``.
+* **dist**: (TODO) Output of the ``webpac``k process, copied to ``static/dist``
+* **layout-tests**: (TODO) Some statically-served pages for speed testing,
```
schalkneethling: I wonder if we still need to keep these. @atopal?

atopal: No, you can remove them. I can still access the tests I have run in in Q4, and I can't use those pages for current testing anyway, because the hash added to file names for cache busting means those files are 404ing now.

jwhitlock:
I've forced push to get a few more files that I missed yesterday. 

Instead of moving the ``layout-tests`` and SCL3 maintenance mode files, I deleted them, with references to the relevant bugs.

assets/README.md
```diff
@@ -0,0 +1,30 @@
...
+  ``ckeditor/build`` is copied to ``static/js/libs/ckeditor/build``.
+* `dist`: (TODO) Output of the ``webpack`` process, copied to ``static/dist``.
+* `legacy`: These files are no longer served to users, and are candidates
+  for deletion.
+* `lib`: (TODO) Third-party libraries that are served to users.
+* `pipeline`: (TODO) Input of the ``gulp`` process. It should be possible to
+  convert these to ``webpack`` some day.
+* `reference`: These files are related to assets, but are not part of an
+  automated build process, and are not served to users.
+* `src`: (TODO) Input for the ``webpack`` process.
```
davidflanagan: I tend not to think of JavaScript code as an "asset" until it has been packed or bundled up. So I would slightly prefer to have a toplevel /javascript directory to develop frontend code in.  Then have webpack package that code and store it in /assets/dist or /assets/webpacks

assets/README.md
```diff
@@ -0,0 +1,30 @@
...
+[Building, collecting, and serving assets](https://kuma.readthedocs.io/en/latest/assets.html)
+for the details.
+
+Here are the subfolders:
+
+* `build`: (TODO) Output of ``gulp`` and ``make compilejsi18n``, copied to
+  ``static/build``.
+* `ckeditor4`: (TODO) The source and build folders for CKEditor.
+  ``ckeditor/build`` is copied to ``static/js/libs/ckeditor/build``.
+* `dist`: (TODO) Output of the ``webpack`` process, copied to ``static/dist``.
```
davidflanagan: it seems a little confusing to have assets/build and assets/dist.  While we're working with two sets of build tools, maybe we could use more descriptive names like assets/gulp and assets/webpack?

(we're also going to have to figure out how to do the i18n step on the code in this directory, either with webpack or after webpack)

davidflanagan:
Thanks, John for taking the time to get rid of old files. This seems like a very reasonable assets scheme to me. I'd prefer writing code under /javascript tha /assets/src/, but that's a pretty minor quibble.

assets/README.md
jwhitlock: Does ``/kuma/javascript`` or ``/javascript`` make more sense to you? And what do you think of the repo split of https://github.com/mozilla/addons-frontend/ and https://github.com/mozilla/addons-server/?

jwhitlock:
It appears the image changes work, and we can tune the folder structure for gulp / pipeline / react as I work on those PRs.

PR #5261, opened 2019-02-20
bug 1441672: make locale redirects for the homepage permanent

escattone:
This PR changes locale-based redirects for the homepage from 302's to 301's for SEO reasons (see https://github.com/mdn/sprints/issues/932). All other endpoints are still redirected via 302 by `LocaleMiddleware`.

For reference, especially see:
- https://github.com/mdn/sprints/issues/932#issuecomment-466117128
- https://github.com/mdn/sprints/issues/932#issuecomment-466123774.

For the record, this was the description at the time this PR was originally submitted:

> This PR changes the redirects that `LocaleMiddleware` performs from 302's to 301's, per https://github.com/mdn/sprints/issues/932#issuecomment-462832669. My understanding is that this should be done for all paths without a locale, not just the homepage (so for example, https://developer.mozilla.org/docs/Web/Apps redirects with a 301 just like https://developer.mozilla.org/ would).
> 
> NOTE: Putting any SEO justifications aside, I have to admit that this feels wrong simply because we'd be making permanent any locale selected and stored in the language cookie (if there is one) or, failing that, the locale in the HTTP `Accept-Language` header, but perhaps that's outweighed by the SEO justifications.

### DIFF
--- a/kuma/core/tests/test_locale_middleware.py
+++ b/kuma/core/tests/test_locale_middleware.py
@@ -70,7 +70,7 @@ def test_locale_middleware_picker(accept_language, locale, client, db):
 @pytest.mark.parametrize('original,fixed', REDIRECT_CASES)
 def test_locale_middleware_fixer(original, fixed, client, db):
     '''The LocaleStandardizerMiddleware redirects non-standard locale URLs.'''
-    response = client.get('/%s/' % original)
+    response = client.get(('/%s/' % original) if original else '/')
     if original == '':
         # LocaleMiddleware handles this case, and it's a 301 instead
         # of a 302 since it's the homepage.
--- a/kuma/core/tests/test_locale_middleware.py
+++ b/kuma/core/tests/test_locale_middleware.py
@@ -49,11 +49,19 @@ REDIRECT_CASES = [
 ] + [(orig, new) for (orig, new) in SIMPLE_ACCEPT_CASES if orig != new]
 
 
+def test_locale_middleware_redirect_when_not_homepage(client, db):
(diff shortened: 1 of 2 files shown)


escattone:
In my haste, I forgot to fix the tests. Labeling this `not ready` until I fix them.

escattone:
As both @jwhitlock and @safwanrahman have also pointed out (in other places), this PR would break our cookie-based redirect mechanism. It seems worth discussing the consequences further before continuing with this PR.

kuma/core/tests/test_locale_middleware.py
```diff
@@ -49,11 +49,19 @@
 ] + [(orig, new) for (orig, new) in SIMPLE_ACCEPT_CASES if orig != new]
 
 
+def test_locale_middleware_redirect_when_not_homepage(client, db):
```
jwhitlock: I like the specific test for the new code path.

kuma/core/tests/test_locale_middleware.py
```diff
@@ -63,7 +71,13 @@ def test_locale_middleware_picker(accept_language, locale, client, db):
 def test_locale_middleware_fixer(original, fixed, client, db):
     '''The LocaleStandardizerMiddleware redirects non-standard locale URLs.'''
     response = client.get('/%s/' % original)
-    assert response.status_code == 302
+    if original == '':
```
jwhitlock: So we are testing the URL `//`? WEIRD

escattone: 😄 good catch! I added `('/%s/' % original) if original else '/'` to get rid of that weirdness.

PR #5262, opened 2019-02-21
Inline emotion styles

jesstelford:
After speaking with one of the emotion core maintainers, he said inlining the styles can give the Babel plugin a better chance to optimise, and in the end it hoists anything it can't optimise anyway.

### DIFF
--- a/kuma/javascript/src/logo.jsx
+++ b/kuma/javascript/src/logo.jsx
@@ -1,21 +1,21 @@
 //@flow
 import * as React from 'react';
 
-const styles = {
-    logo: {
-        display: 'block',
-        height: 48,
-        width: 219,
-        marginTop: 15,
-        // The sprite image includes a mask, and is not itself 48px high
-        // so we need overflow to hide the mask.
-        overflow: 'hidden'
-    }
-};
 
 export default function Logo(props: {| url: string |}): React.Node {
     return (
-        <a css={styles.logo} href={props.url}>
+        <a
+            css={{
+                display: 'block',
+                height: 48,
+                width: 219,
+                marginTop: 15,
+                // The sprite image includes a mask, and is not itself 48px high
+                // so we need overflow to hide the mask.
+                overflow: 'hidden'
+            }}
+            href={props.url}
+        >
             <img src="/static/img/web-docs-sprite.svg" alt="MDN Web Docs" />
         </a>
     );


davidflanagan:
Thanks for this, Jess. I actually had the styles inlined this way to start. But I moved so I could assign them to something with a name because the emotion babel plugin is supposed to create class names that are more human readable, and that didn't seem to be happening. (Maybe because I used a functional component instead of a class-based component.)

@jesstelford is the plugin issue that the styles need to be fully inlined, or is it just that they way I've written it, the plugin can't tell that the styles are styles? When I used Aphrodite, I liked the practice of defining one big JS object with all the styles either at the top or bottom of each file. And if I can avoid creating the style objects on each render, I'd like to do it that way.

So I wonder if I can do this:

```
import {css} from '@emotion/core';

const styles = {
    logo: css({....})
};
```

Then in my react component: `<a css={styles.logo} href={...}>`

Does that give the emotion plugin what it needs to do optimizations without having to recreate the style object on each render?

(@schalkneethling: FYI, since we need to figure out what conventions we want to use for this kind of thing)

mitchellhamilton:
👋I'm the maintainer that @jesstelford is referring to. I just want to clarify something, the way it currently is right now is totally fine, it won't really cause any performance issues. What you suggested would work fine as well and labels should work with that approach vs the current approach which they won't be generated for.

Something that I'd suggest is don't make stylistic decisions based on small performance characteristics, just do whatever you prefer and then tools like babel-plugin-emotion can be improved to optimise more code styles better if it actually becomes an issue. (which it's very very unlikely to be, it's much more likely that other things will cause performance issues)

> I moved so I could assign them to something with a name because the emotion babel plugin is supposed to create class names that are more human readable, and that didn't seem to be happening. (Maybe because I used a functional component instead of a class-based component.)

This should definitely work. I'm gonna look into it and fix it.

davidflanagan:
Thanks, Mitchell! I'm flattered to have your input on this.

davidflanagan:
@mitchellhamilton: I might well have just been doing something wrong when I wasn't getting the human-readable bits in the class names. webpack config issues are at least a likely to be the problem as a bug in your plugin. So take my description above with a grain of salt.

mitchellhamilton:
> Thanks, Mitchell! I'm flattered to have your input on this.

❤️ Thanks so much for your work on Kuma and MDN!

> @mitchellhamilton: I might well have just been doing something wrong when I wasn't getting the human-readable bits in the class names. webpack config issues are at least a likely to be the problem as a bug in your plugin. So take my description above with a grain of salt.

I've actually reproduced it and it's looks like it's an issue with babel-plugin-emotion. Thanks for finding it!

jwhitlock:
If this is still a valid change, a rebase is needed to sync with the current master.

davidflanagan:
No, I don't think this PR is needed any longer. I'm now passing the style objects to the css() function before using them on components, and am happy with that convention. Thank you @jesstelford

PR #5263, opened 2019-02-21
Bug 1523643, fixes anchor widget dropping to new line on small mobile viewports

schalkneethling:
@davidflanagan r?

### DIFF
--- a/kuma/static/styles/components/wiki/content/collapsible.scss
+++ b/kuma/static/styles/components/wiki/content/collapsible.scss
@@ -1,6 +1,6 @@
 h2 > .collapse-trigger {
     /* over-ride button styles */
-    @include set-font-size($mobile-h2-font-size);
+    @include set-font-size($collapse-trigger-mobile-font-size);
     background-color: inherit;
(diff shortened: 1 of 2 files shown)


schalkneethling:
> Its weird that the previous inline-block fix doesn't work when the line wraps. I don't understand what the underlying cause is.
> 
> Are there other reasons why reducing the font size is a good idea? If 32px is the preferred size, then I'd suggest that it would be better to leave this bug unfixed than to change the size. But if you think that 28px is a better size anyway, this this seems like a good approach. (I assume though that even with the smaller size, the bug will still exist for even longer titles...)

Yes, it definitely will. I feel like those are things we should not keep fighting, as that is just how the web works. It's all about the button and its bounds. When it wraps, the button takes up space as follows:

screenshot 2019-02-22 at 05 14 25

With the smaller font size, it then does not break and leaves enough space for the icon to fit on the right.

screenshot 2019-02-22 at 05 15 22

The new font size gives us more horizontal space, and so there should be very few, if any, of these edge cases.

PR #5267, opened 2019-02-22
bug 1380419 - Move CKEditor to /assets/ckeditor4

jwhitlock:
Move the CKEditor build from ``kuma/static/js/libs/ckeditor`` to ``assets/ckeditor4``. The name change will help prepare for CKEditor 5 development.

Use the prefix form of ``STATICFILES_DIRS`` to only copy the build files, not the source files, to ``/static/js/libs/ckeditor4/build``.


jwhitlock:
With PR #5266 merged, I've rebased on master and I'm removing the "not ready" label.

PR #5268, opened 2019-02-23
bug 1523273: use proper hreflang values

escattone:
This PR updates the way `hreflang` values are set for documents. For reference, see [bug 1523273](https://bugzilla.mozilla.org/show_bug.cgi?id=1523273). There's one special case that made this more challenging. The special case is that the `hreflang` value for any given document sometimes depends on the other translations that are available for that document. For example, if the document is available in both the `pt-BR` and `pt-PT` locales, the `hreflang` value for the `pt-BR` document would be `pt-BR`, like this:

```
<link rel="canonical" href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array">
<link rel="alternate" hreflang="en" href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array" title="Array">
<link rel="alternate" hreflang="pt-BR" href="https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Array" title="Array">
<link rel="alternate" hreflang="pt" href="https://developer.mozilla.org/pt-PT/docs/Web/JavaScript/Reference/Global_Objects/Array" title="Array">
```

but if the document is only available in the `pt-BR` locale, the `hreflang` value for the `pt-BR` document becomes `pt`, like this:

```
<link rel="canonical" href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array">
<link rel="alternate" hreflang="en" href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array" title="Array">
<link rel="alternate" hreflang="pt" href="https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Array" title="Array">
```

See the code and comments on the `get_hreflang` method I added to the `Document` class for more details.

This PR also updates how the `lang` attribute of the `html` element gets set ([the bug mentions that it should match the `hreflang` value of the document](https://bugzilla.mozilla.org/show_bug.cgi?id=1523273#c8)).

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -1358,7 +1358,7 @@ Full traceback:
             current_parent = current_parent.parent_topic
         return parents
 
-    def get_other_translations(self, fields=[]):
+    def get_other_translations(self, fields=()):
         """
         Return a list of Documents - other translations of this Document
         :param fields: Model Fields that should only be fetched
@@ -1430,6 +1430,37 @@ Full traceback:
     @property
     def is_experiment(self):
         return self.slug.startswith(EXPERIMENT_TITLE_PREFIX)
+
+    def get_hreflang(self, other_locales=None):
+        """
+        Return the hreflang value for this document, possibily depending upon
+        the other locales in which this document is available.
+        """
+        # In most cases, just return the language code, removing the country
+        # code if present (so, for example, 'en-US' becomes 'en').
+        hreflang = self.locale.split('-')[0]
+
+        # Check for the special case when we want the full locale (i.e.,
+        # including the country code). This is the case when this document's
+        # locale is not the preferred locale, and the preferred locale is
+        # among the other locales in which this document is available. For
+        # example, if this document is available in 'en-US', 'pt-PT', and
+        # 'pt-BR', the respective hreflang values would be 'en', 'pt', and
+        # 'pt-BR' (we're using the full locale for 'pt-BR'), but if it's only
+        # available in 'en-US' and 'pt-BR', we'll choose hreflang values of
+        # 'en' and 'pt' (we're not using the full locale for 'pt-BR', just the
+        # language code 'pt', so the one translation covers all countries for
+        # that language).
+        if hreflang in settings.PREFERRED_LOCALE:
+            preferred = settings.PREFERRED_LOCALE[hreflang]
+            if self.locale != preferred:
+                if other_locales is None:
+                    trans = self.get_other_translations(fields=('locale',))
+                    other_locales = (doc.locale for doc in trans)
+                if preferred in other_locales:
+                    return self.locale
+
+        return hreflang
 
 
 class DocumentDeletionLog(models.Model):
--- a/kuma/wiki/tests/test_models_document.py
+++ b/kuma/wiki/tests/test_models_document.py
@@ -12,7 +12,7 @@ from datetime import timedelta
 import mock
 import pytest
 
-from . import normalize_html
+from . import HREFLANG_TEST_CASES, normalize_html
 from ..models import Document, Revision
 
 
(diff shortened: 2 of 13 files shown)


escattone:
Pinging @rraue to get his review as well.

kuma/wiki/models.py
```diff
@@ -1358,7 +1358,7 @@ def parents(self):
             current_parent = current_parent.parent_topic
         return parents
 
-    def get_other_translations(self, fields=[]):
+    def get_other_translations(self, fields=()):
```
escattone: This is just force of habit in avoiding mutable types as keyword defaults. It's not required for this PR.

escattone:
Hm, just noticed that I'm getting slug collision errors. Strange that I didn't see these when testing locally. I'll dig into them.

escattone:
I just force-pushed an update to the commit that will fix the issue. I had made a last minute change to the tests and inadvertently broke one of them.

rraue:
Is it possible to see this running in a staging environment, so I can test it?

escattone:
@jwhitlock Now that I know what was causing the `zh-TW` and `zu` related tests to fail within `kuma/wiki/tests/test_views_document.py::test_hreflang`, I'm surprised that we hadn't hit this before. It turns out that since we weren't compiling the locales within the various `Travis` runs, Django's `check_for_language` (called from `kuma.core.i18n.get_supported_language_variant`) always returned `False` for the `zu` and `zh-TW` locales (and perhaps others as well? or maybe the others have defaults where these locales don't?), because it couldn't find an `.mo` file for those locales. The fix was to simply add  the `localecompile` Makefile target to each of the `pytest`, `pytest3`, and `docker` `tox.ini` environments. I had first thought I should make the changes to `tox.ini` a separate PR, but then thought they perhaps best belonged here since the tests for that change are shared with this PR. In the end, I made it a separate commit of this PR. Thanks again for your kind help with this @jwhitlock!

escattone:
@rraue This code has been pushed to our stage instance (https://developer.allizom.org/en-US/) for your review.  For example, here's a document URL that may find useful:

https://developer.allizom.org/en-US/docs/Web/HTTP

Cheers, thank you so much for your help with this!

rraue:
@escattone That looks all very good and should fix over time the bug. 

Just one thing: hreflang currently just supports ISO_639-1 and in the example url there is the alternate "kab" which is unfortunately right now not supported as it is in ISO 639-2. Not part of this pr, but we should only include ISO 639-1 languages in hreflang. In the language switcher in the footer of the page should be all languages available included thought.

jwhitlock:
Google also doesn't support some ISO 639-1 languages. The unsupported locales issue is tracked in [bug 1441682](https://bugzilla.mozilla.org/show_bug.cgi?id=1441682). We can address that in a future PR.

kuma/wiki/tests/test_models_document.py
```diff
@@ -43,6 +43,42 @@ def doc_with_sections(root_doc, wiki_user):
...
+    for locale, expected_result in zip(locales, expected_results):
+        doc = Document.objects.create(
+            locale=locale,
+            slug='Root',
+            title='Root Document',
+            rendered_html='<p>...</p>',
+            parent=root_doc
+        )
+        with mock.patch.object(doc, 'get_other_translations',
+                               return_value=()) as mocked_method:
```
jwhitlock: When I'm mocking to ensure a method isn't called, I use ``side_effect=Exception('with optional string')``, so that an exception would be raised if the method was called. With a ``return_value``, it implies you are expecting a call, and returning an empty sequence for test purposes.

escattone: I like that much better, thanks @jwhitlock! I will update this PR with that change and then merge.

PR #5270, opened 2019-02-25
Adds test for Logo component, and config to ensure Jest runs successfully

schalkneethling:
This adds a snapshot test for the `Logo` component. There was also some additional config required to get Jest to run successfully.

@davidflanagan r?


.babelrc
```diff
@@ -0,0 +1,15 @@
...
+        [
+            "@babel/env",
+            {
+                "targets": {
+                    "browsers": "last 2 versions"
+                }
+            }
+        ]
+    ],
+    "plugins": ["@babel/plugin-proposal-class-properties"]
```
jwhitlock: The Docker build is broken, because packages are installed to ``/tools/node_modules``. I was able to fix this in ``webpack.config.js``, because code was available, but I'm not sure how to fix this in a JSON config.

schalkneethling: Hmmmm. Let me look at this.

schalkneethling: We can swith to using `babel.config.js` and then probably use the same solution as in Webpack config. Will have a look.

schalkneethling:
@jwhitlock updated.

davidflanagan:
I just realized that by creating an independent babel config file, we can probably remove the babel config from the webpack config, so that we only have to define it once.

I can do that in a followup patch in PR #5273, since I'll be updating that PR to add snapshot tests as soon as this PR is merged.

PR #5272, opened 2019-02-26
Debug TravisCI

jwhitlock:
Same as #5268, but with some extra debugging to try to understand why the tests fail.


escattone:
Closing this now that we have a [fix](https://github.com/mozilla/kuma/pull/5268#issuecomment-467682716), thanks for the help @jwhitlock!

PR #5273, opened 2019-02-26
Bug 1525719: rewrite the MDN page header in React

davidflanagan:
This PR is an experiment with using React for the MDN page header.
This is a follow-on to PR #5254 which set up React for the MDN logo
for pages with special URLs that are never served in production. This
PR takes it further, adding the rest of the (logged-out) page header.

As part of the PR, we also:

- create a kuma/javascript/src/header directory for React components
- add a gettext.js module to (later) be used for l10n
- add a new .eslintrc.js config file for React-specific linting
- experiment with new visual design for the header


.eslintrc.js
```diff
@@ -9,19 +9,14 @@ module.exports = {
     'parser': "babel-eslint",
     'parserOptions': {
         "ecmaVersion": 6,
-        "ecmaFeatures": {
-            "jsx": true
-        }
     },
-    "plugins": ["jest", "flowtype", "react", "import", "jsx-a11y"],
+    "plugins": ["jest", "flowtype", "import", "jsx-a11y"],
```
davidflanagan: I've created another eslint config in kuma/javascript/src that is specific to react, and I've removed the react plugin from this one because it was complaining about not being configured correctly.

kuma/javascript/src/.eslintrc.js
```diff
@@ -0,0 +1,71 @@
+module.exports = {
```
davidflanagan: I created a new eslint config for this directory because this is mainly where we are using prettier and react. The eslint config above has indentation rules that are not always compatible with what prettier does, but I don't want to remove them since we still have JS code that does not use prettier. So instead, I'm proposing that we have a different eslint config for the new code we're writing here.

davidflanagan:
Schalk: this will need some snapshot tests. I'll add those (and fix merge conflicts) once your PR that adds the logo snapshot test lands. But I thought you might want to take a look even before the tests are added.

davidflanagan:
screen shot 2019-02-26 at 1 11 46 pm

davidflanagan:
I should also mention that this is desktop-only at this point. The new header does not work for mobile screen sizes.

davidflanagan:
This is weird. Git is showing Ryan as the author of the commit. And attempting to rebase against master is causing all kind of weirdness. So I've started over with a new commit in a new branch. I'm closing this and will open a new PR.

PR #5269, opened 2019-02-25
Fix bug 1529886: Make `no‑line‑numbers` always hide the line numbers

ExE-Boss:
Fixes [bug 1529886](https://bugzilla.mozilla.org/show_bug.cgi?id=1529886)

See&nbsp;also https://github.com/mozilla/kuma/pull/5189

PR #5275, opened 2019-02-27
Bug 1530782 - Skip adding line numbers if <pre> has no-line-numbers

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1530782

line-numbers plugin was checking `line-numbers` class, but the pattern also matches to `no-line-numbers`.
for simplicity (without introducing more complicated pattern), just added another check for `no-line-numbers`.

### DIFF
--- a/kuma/static/js/libs/prism/prism-line-numbers.js
+++ b/kuma/static/js/libs/prism/prism-line-numbers.js
@@ -73,12 +73,17 @@
 
 		// works only for <code> wrapped inside <pre> (not inline)
 		var pre = env.element.parentNode;
+		if (!pre || !/pre/i.test(pre.nodeName)) {
+			return;
+		}
+		// If either <pre> or <code> has no-line-numbers class, ignore it.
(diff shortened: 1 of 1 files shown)


schalkneethling:
Thanks so much for your PR @arai-a - Unfortunately I only saw your PR after merging https://github.com/mdn/sprints/issues/1089

Also, while this will work, it is preferable to not edit the library directly as it will make future updates harder. Thank you again for your contribution. ❤️

arai-a:
how is mdn/sprints#1089 related to this and bug 1530782?

schalkneethling:
> how is [mdn/sprints#1089](https://github.com/mdn/sprints/issues/1089) related to this and bug 1530782?

As discussed in IRC, it was the incorrect link. I have updated the comment. Here is the correct link to the PR:
https://github.com/mozilla/kuma/pull/5269

ExE-Boss:
It might be a good idea to open this on the [PrismJS repository](https://github.com/PrismJS/prism).

/cc @RunDevelopment @mAAdhaTTah

PR #5277, opened 2019-02-28
Bug 1525719: rewrite the MDN page header in React

davidflanagan:
This PR is an experiment with using React for the MDN page header.
This is a follow-on to PR #5254 which set up React for the MDN logo
for pages with special URLs that are never served in production. This
PR takes it further, adding the rest of the (logged-out) page header.

As part of the PR, we also:

- create a kuma/javascript/src/header directory for React components

- add a gettext.js module to (later) be used for l10n

- add a new .eslintrc.js config file for React-specific linting

- experiment with new visual design for the header

- Move the logo test to be in the same directory as the logo component.

- Add jest-emotion dev dependency for better snapshots of styles.

- Add snapshot tests for all the header-related components.


davidflanagan:
I made a number of config file changes in this PR. Here's a rundown:

- there are now two eslint config files. One at the root, and one in kuma/javascript/src, where it applies only to our react code. This was needed because the original root eslint config had indentation rules that were not working well with prettier (in the special case of indenting within template strings). I didn't want to just remove the indentation rule at the root because we have JS code that does not use prettier there. So instead, I just removed it from the new react-specific config file. Also, I removed react-related stuff from the root eslint config, since it isn't needed there.

- A previous patch moved babel config from inside the webpack config to its own file. This was needed to make jest tests properly transform react code. I've added another preset to that babel config file (to properly handle the emotion css property). And I've removed the babel config from webpack since it isn't needed there anymore.

- I've updated the Jest configuration file to add an option that makes it create more useful snapshots of emotion styles, using the jest-emotion plugin which is added as a new dev dependency.

kuma/javascript/src/header/header.jsx
```diff
@@ -0,0 +1,142 @@
...
+        return `/${locale}/docs/${url}`;
+    }
+}
+
+const HOME_URL =
+    window && window.location
+        ? `${window.location.protocol}//${window.location.host}/${
+              window.location.pathname.split('/')[1]
+          }/`
+        : 'https://developer.mozilla.org/en-US/';
```
jwhitlock: In the Django version, ``url(home)`` renders to the path ``/en-US/``, or whatever the current locale is, without the protocol and domain. I think leaving the domain out could make offline rendering for different environments easier.

kuma/javascript/src/header/__snapshots__/header.test.js.snap
```diff
@@ -0,0 +1,258 @@
...
+.emotion-28 {
+  vertical-align: -6px;
+}
+
+<div
+  className="emotion-32 emotion-33"
+>
+  <a
+    className="emotion-0"
+    href="http://localhost//"
```
jwhitlock: I didn't expect the double slash at the end of ``http://localhost//``.

kuma/javascript/src/header/dropdown.test.js
```diff
@@ -0,0 +1,63 @@
...
+    expect(openString).toContain('bar');
+    expect(openTree).toMatchSnapshot();
+
+    // Fake a click on the document body to dismiss the menu
+    act(() => {
+        document.body.dispatchEvent(new MouseEvent('click'));
+    });
+
+    // Expect the menu to be closed, checking deep equality
+    expect(JSON.stringify(dropdown.toJSON())).toEqual(closedString);
```
jwhitlock: I didn't expect that clicking on the document body would dismiss the menu, given setting ``document.documentElement.style.pointerEvents = 'none';`` in ``Dropdown``'s ``useEffect``, but I just learned about [pointer-events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) a few hours ago, so I must be misunderstanding the purpose of setting ``pointerEvents``.

kuma/javascript/src/header/login.jsx
```diff
@@ -0,0 +1,48 @@
...
...656 5.704 6.969c-.656 1.656-.25 2.875-.125 3.187-.766.844-1.234 1.906-1.234 3.219 0 4.594 2.797 5.625 5.469 5.922-.344.313-.656.844-.766 1.609-.688.313-2.438.844-3.484-1-.656-1.141-1.844-1.234-1.844-1.234-1.172-.016-.078.734-.078.734.781.359 1.328 1.75 1.328 1.75.703 2.141 4.047 1.422 4.047 1.422 0 1 .016 1.937.016 2.234 0 .313-.219.688-.828.578C3.439 23.796.002 19.296.002 13.999c0-6.625 5.375-12 12-12zM4.547 19.234c.031-.063-.016-.141-.109-.187-.094-.031-.172-.016-.203.031-.031.063.016.141.109.187.078.047.172.031.203-.031zm.484.532c.063-.047.047-.156-.031-.25-.078-.078-.187-.109-.25-.047-.063.047-.047.156.031.25.078.078.187.109.25.047zm.469.703c.078-.063.078-.187 0-.297-.063-.109-.187-.156-.266-.094-.078.047-.078.172 0 .281s.203.156.266.109zm.656.656c.063-.063.031-.203-.063-.297-.109-.109-.25-.125-.313-.047-.078.063-.047.203.063.297.109.109.25.125.313.047zm.891.391c.031-.094-.063-.203-.203-.25-.125-.031-.266.016-.297.109s.063.203.203.234c.125.047.266 0 .297-.094zm.984.078c0-.109-.125-.187-.266-.172-.141 0-.25.078-.25.172 0 .109.109.187.266.172.141 0 .25-.078.25-.172zm.906-.156c-.016-.094-.141-.156-.281-.141-.141.031-.234.125-.219.234.016.094.141.156.281.125s.234-.125.219-.219z" />
```
jwhitlock: I'd like to figure out a way to load SVG icons from files before we inline too many into ``.jsx`` code. I find an SVG file a lot easier to review than embedded SVG.

kuma/javascript/src/header/logo.test.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { create } from 'react-test-renderer';
-import Logo from '../../kuma/javascript/src/logo.jsx';
+import Logo from './logo.jsx';
```
jwhitlock: 👍 to moving tests next to the code they test. As a happy side effect, it means the two files are close to each other in the pull request 😁

kuma/javascript/src/header/search.test.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { create } from 'react-test-renderer';
+import Search from './search.jsx';
+
+test('Search snapshot', () => {
+    const search = create(<Search />);
+    expect(search.toJSON()).toMatchSnapshot();
```
jwhitlock: I like having a simple render test for components. It seems like "at least a snapshot test" would be a good standard going forward.

kuma/javascript/src/header/search.jsx
```diff
@@ -0,0 +1,66 @@
...
+        flex: '0 1 300px'
+    })
+};
+
+export default function Search() {
+    return (
+        <form
+            css={styles.container}
+            id="nav-main-search"
+            action="/en-US/search"
```
jwhitlock: We'll need to "localize" this URL eventually

kuma/javascript/src/header/dropdown.test.js
davidflanagan: The click and keydown event handlers I register for dismissing the menu are capturing handlers, which means they are triggered before the event gets dispatched anywhere (during the "capturing phase", before the event is dispatched to its target and then bubbles back up). This means my handlers get first crack at the events.

Setting style.pointerEvents to none on the html element makes everything "transparent" to events. The main point there is to make it so that things with hover behaviors don't do anything when hovered over while the menu is displayed. But it also means that I don't have to call preventDefault() on the event to stop it from actually activating links if the user clicks on one.   (I could instead call preventDefault on events, but would have to be careful to only do that when they were outside of the menu, and it is easier to just do the pointerEvents thing and get the hover disable effect as well.)

Note also that the pointer-events CSS thing is completely different from PointerEvents (which is an attempt to unify mouse and touch events into a single thing, I think.)

kuma/javascript/src/header/__snapshots__/header.test.js.snap
davidflanagan: I didn't expect it either, but it seems harmless. The href is based on window.location, and there is a fallback for when window.location is not defined. I assumed it would not be defined in this node-based testing environment, since it is not a browser. But Jest does define it, but window.location.pathname is empty I guess? Or maybe there is an extra slash after the hostname?  Anyway: I'm pretty sure this is an artifact of the testing environment, but I'll double-check the code to make sure I'm not adding an extra slash.

davidflanagan: Oh; that's right. It is because there is supposed to be a locale between those slashes, but in the test environment window.location.pathname does not include a locale

kuma/javascript/src/header/dropdown.test.js
escattone: That helps, thanks @davidflanagan!

kuma/javascript/src/header/header.jsx
davidflanagan: Good point. Fixed.

kuma/javascript/src/header/__snapshots__/header.test.js.snap
davidflanagan: Fixed as a side effect of fixing your point about about just using the path and not the protocol and host.

kuma/javascript/src/header/login.jsx
davidflanagan: Agreed. The svg-inline-loader package for webpack will probably do what we need. 

Not going to make that change now, though.

kuma/javascript/src/header/search.jsx
davidflanagan: Thanks for catching that.

kuma/javascript/src/header/dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+    // on the entire document when the menu is up, but we need to be
+    // able to click on the menu items themselves
+    pointer-events: auto;
+`;
+
+const MenuLabel = styled.div`
+    font-size: 18px;
+    font-weight: bold;
+    padding: 5px 10px;
+    :hover {
```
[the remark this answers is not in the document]
davidflanagan: Or maybe the hand cursor to match what is used for links? Actually there is probably an a11y issue here and I should be rendering the menu label as a link or a button or something more semantically correct.

I'll try switching to Button to see if that fixes the cursor issue, and will add a note to the code that there may be an a11y issue, and that maybe the menu items always need to be rendered, even when the menu is closed for the benefit of screen readers.

(I don't want to spend a lot of time on this because I'm hoping that when we do a site redesign the menus just go away!)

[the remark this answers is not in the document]
davidflanagan: Switching to a button fixes this.

PR #5279, opened 2019-03-01
Bug 1525719: configure Puente to extract strings from jsx files

davidflanagan:
This PR changes the Puente settings in kuma/settings/common.py so
that Puente can find and extract gettext() strings in the React-based
code in kuma/javascript/src/

Note that if we actually deploy this, we will cause work for localizers
for things like the React header components that we've been using as
test cases.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -672,6 +672,8 @@ PUENTE = {
             ('kuma/static/js/components/**.js', 'javascript'),
             ('kuma/static/js/libs/ckeditor/source/plugins/mdn-**/*.js',
              'javascript'),
+            ('kuma/javascript/src/**.js', 'javascript'),
+            ('kuma/javascript/src/**.jsx', 'javascript'),
         ],
     },
     'PROJECT': 'MDN',


davidflanagan:
@escattone: Thanks for the review. I've already updated #5277 to fix localization in header.jsx.

What's your feeling about whether this PR should be landed? #5277 doesn't introduce new strings, but it appears that strings from python do not automatically get used again when they appear in JavaScript, so I think that if we land this, our localizers will all of a sudden have a bunch of already-translated strings that need re-translation. And they are strings that we're not going to be user visible except on the staging server. I'm reluctant to waste people's time like that, and so am not sure we should actually merge this until we're ready to deploy React stuff to production.

escattone:
@davidflanagan Good point. Let's hold off, and we can see what @jwhitlock thinks tomorrow.

jwhitlock:
Here's a few option that come to mind:

1. Add a configuration option to optionally include ``*.jsx`` files. We similar switches around CSP (``CSP_ENABLE_MIDDLEWARE``) and locales (``ENABLE_CANDIDATE_LOCALES``). We can then develop locally and see English on staging, but avoid posting to https://github.com/mozilla-l10n/mdn-l10n/.
2. Move ``.jsx`` to a new file, like ``react.po``, for now or forever. We may be able to hide the new file from Pontoon, or do some magic to add ``django.po`` translations so they don't show up as untranslated strings. A variant would be to add these to both ``django.po`` and ``react.po`` - it would show up as a new source for the string in ``django.po`` but not as a new translation, unless there was a copy and paste error. As a side benefit, we may use a different toolchain to load ``react.po``.

jwhitlock:
Also, I think a few block diagrams in the assets doc will help with scanning, I may get those into the next version.

escattone:
Since #5302 has been approved and supersedes this PR, I'm closing this.

PR #5280, opened 2019-03-02
Bug 1525719: React spike: try simple client-side navigation

davidflanagan:
This PR builds on previous /ducks/ PRs and adds simple pushState/popstate
client side navigation for MDN wiki pages. The intent isn't to get
everything working perfectly but to get an idea for what kind of
speedup we can expect from this technique

### DIFF
--- a/kuma/javascript/src/client-side-nav.js
+++ b/kuma/javascript/src/client-side-nav.js
@@ -1,0 +1,113 @@
+const locale = window.location.pathname.split('/')[1];
+const pattern = `/${locale}/docs/Web/`;
+
+function init() {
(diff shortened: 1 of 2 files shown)


davidflanagan:
This was a fun friday experiment with client-side navigation, doing the best I can with $json to fetch the new page title and ?raw&macros to get page content. It doesn't work perfectly--no way to fetch breadcrumbs, for example. But it works better than I expected given that we haven't written any specific APIs to support this yet.

This only affects /ducks/ URLs and I think it would be interesting to land it to get an idea of how much of a speedup it gives us (over full reloads) on the staging server.

PR #5281, opened 2019-03-03
bug 1532147: Respect white spaces of localization source's code blocks

alattalatta:
[Bugzilla 1532147](https://bugzilla.mozilla.org/show_bug.cgi?id=1532147)

This PR removes `white-space: normal` from `<pre>` blocks of localization source document.

Currently all line breaks and multiple blanks are gone from source doc.

The issue itself had been gone for sometime by some reason I do not know. Now it's happening again.

The code has been there for a long time too; I wonder what was the reason for `white-space: normal`, and would removing it cause any side effects? (I don't think so...)

### DIFF
--- a/kuma/static/styles/components/wiki/edit/translate/translate-document.scss
+++ b/kuma/static/styles/components/wiki/edit/translate/translate-document.scss
@@ -1,12 +1,7 @@
 #translate-document {
-
     /* approved version of the article content */
     .translate-rendered {
         overflow-x: hidden;
-
-        pre {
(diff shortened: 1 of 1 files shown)


SphinxKnight:
Agreeing with @alattalatta for this change (as filed in dupes PR/bug) :)

jwhitlock:
I've confirmed this works locally. I agree that the cause is the Prism 1.15 update (PR #5189), but we're hopefully getting to the end of the changes.

PR #5284, opened 2019-03-05
Bug 1525719: fix a JavaScript error for the /ducks/ URLs

davidflanagan:
When I rewrote the header in React I removed the animated search box
but left the nav-main-search.js component in the bundle. In development
this caused a benign JavaScript error, but in production the error prevents
the execution of other JS in the same bundle and the header does not appear
at all. This patch just modifies bundle definition to remove that
now-unnecessary JavaScript file.


davidflanagan:
This is a followup to PR #5280, and will hopefully make it work correctly on staging.

PR #5285, opened 2019-03-05
bug 1525719: first cut of the whoami api

escattone:
This is the first cut of the `whoami` user API, which simply populates and returns a JSON object based on the presence/absence of the `sessionid` cookie.

### DIFF
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -8,7 +8,6 @@ from allauth.account.models import EmailAddress
 from allauth.socialaccount.models import SocialAccount
 from constance.test.utils import override_config
 from django.conf import settings
-from django.contrib.auth.models import Group
 from django.core import mail
 from django.db import IntegrityError
 from django.http import Http404
@@ -44,11 +43,6 @@ def wiki_user_github_account(wiki_user):
             html_url="https://github.com/{}".format(wiki_user.username)
         )
     )
-
-
-@pytest.fixture
-def beta_testers_group(db):
-    return Group.objects.create(name='Beta Testers')
 
 
 def test_old_profile_url_gone(db, client):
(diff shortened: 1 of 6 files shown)


kuma/users/tests/test_views.py
```diff
@@ -46,11 +45,6 @@ def wiki_user_github_account(wiki_user):
     )
 
 
-@pytest.fixture
-def beta_testers_group(db):
```
davidflanagan: This is still used in this file. Does pytest automatically find it in the conftest file where you've moved it to?

escattone: Yes, all of the test files within any given folder have access to the fixtures defined in `conftest.py` file within that folder as well as all folders upwards in the hierarchy.

safwanrahman:
Can we use Django Rest Framework generic view for this purpose?

safwanrahman:
I think the API design should be better constructed as
`api/v1/me`. It will return the logged in user information if the user is logged in, otherwise return `403`.

jwhitlock:
@escattone and I discussed this before this PR was written, and we discussed it some more in IRC.

We currently are planning a read-only API that is just this endpoint. In the future we may add more endpoints, but we haven't committed to any writable endpoints. Django REST Framework is a natural fit for a writable API in Django, but we don't want to commit to it until we need it. We'd also need to carefully update the DRF version, since the search feature uses it. We decided to do the simplest thing that works for this version, but use a design that could be implemented in DRF in the future.

I've used ``/api/v1/users/me`` in the past, but I think ``/api/v1/me`` or ``/api/v1/whoami`` is a better idea. I used ``/api/v1/users/<user_id>`` as a way to get user data, and  ``/api/v1/users/me`` redirected to the logged-in users record. However, that means you have to make sure that ``me`` is never a valid user ID or username. I think your idea to move it up and out of ``/api/v1/users`` will make a future user API easier to implement. I prefer ``/api/v1/whoami``.

PR #5287, opened 2019-03-08
bug 1490356: Update sample DB, improve scraper, update timezones

jwhitlock:
Update the sample DB specification:
* Drop ``Web/Apps``, now that the Firefox OS App Center is decomissioned
* Add some test pages like [Sandbox/All-Indicators](https://developer.mozilla.org/en-US/docs/Sandbox/All-Indicators) that were useful during style refactoring
* Add explicit scrape of [MDN/Promote](https://developer.mozilla.org/en-US/docs/MDN/Promote), which is a redirect on the homepage, in preparation to de-redirect the homepage
* Add ``contrib_beta`` and ``recurring_payment_beta`` waffle flags, used by payments

Scraper updates:
* When run in verbose mode (``--verbosity 2``, used by ``scripts/create_sample_db.sh``), log once rather than to both ``stderr`` and ``stdout``
* Make verbose logging more verbose, including a report on how long the scrape took

In preparation for a new sample database, update some dependencies:

* [django-sundial](https://github.com/charettes/django-sundial/) 1.0.2 → 1.0.6: Update support to Django 1.11 to 2.1, Python 2.7, 3.4 to 3.7
* [pytz](http://pytz.sourceforge.net/) 2018.5 → 2018.9: Sync with latest timezone database

These packages set the defaults and validation for the ``timezone`` field on user profiles, so a migration is needed on update.

I successfully generated a new sample database, so I don't believe there are bugs outside of Kuma and out-of-date JSON. I uploaded and tested it today, but it says "Your models have changes that are not yet reflected in a migration". I'll generate a new sample DB on this code, so that we have a clean database for a few days, until the next model change.


jwhitlock:
The migration changes the choices for the timezone field, which is a Django-side feature, and doesn't propagate to MySQL. This will be a very fast migration.

PR #5289, opened 2019-03-08
Bug 1525719: Query user data on the client side with new whoami API

davidflanagan:
This PR introduces a new CurrentUser.Provider React component
that uses fetch() to get user data from the whoami API and then
uses the React Context API to share that user data with any components
that need it. The patch also includes changes to the existing Login
component to display the avatar of logged in users.

The patch also modifies the Login component to display a dropdown menu for
logged in users to allow view profile, edit profile and sign out commands.
This also includes modifications to the Dropdown component to make
it work with non-textual menu labels and to attach the menu to the
right side of the label instead of the left side.

Finally, the patch also bumps the minor version of React and makes
some changes to ensure that we can use Flow on our Jest tests.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ kuma/static/js/utils/perf.min.js
 kuma/static/js/utils/perf-post-message-handler.min.js
 kuma/static/layout-tests/js/*.js
 kuma/javascript/dist/*.js
+kuma/javascript/src/flow-typed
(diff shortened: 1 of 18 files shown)


davidflanagan:
The primary new stuff in this patch is src/current-user.jsx and the code in src/header/login.jsx that uses the current-user context.  But login also now uses a Dropdown menu, and there are some changes in src/header/dropdown.jsx to support that.

There are quite a few other changed files: mostly tests, updated snapshots, additions of "//@flow" comments to existing tests, and config file changes to make Flow work with Jest tests. None of those things should need detailed review.

PR #5290, opened 2019-03-09
Bug 1525719: move SVGs into external files

davidflanagan:
This patch modifies our React components that use SVGs so that the
SVGs live in external files rather than being inlined directly
into the components that use them.

And it also installs a webpack loader to automatically load them
as if they were React components. And it configures Jest to just
mock them out so they don't interfere with testing and so that changes
to the svg don't cause changes to the snapshots.


davidflanagan:
There appear to be at least three different webpack loaders that people use with SVG images, and various ways of handling them. This approach (just treating svgs as if they are React components that can be imported) seems really appealing. It required some Jest configuration to make tests work, however.

@jwhitlock: you had requested something like this in a review of an eariler PR. Please see what you think. Note that webpack is inlining the svgs into the javascript bundle, so they don't end up as external assets in the dist/ directory

@schalkneethling: do you have an opinion on the best way to do this, or better ideas about how to make this work with Jest tests?

davidflanagan:
@jwhitlock: I think you're right that the svg files should be grouped together into an icons/ directory of some sort. I'll take care of that down the road when we've got some more of them.

schalkneethling:
> @jwhitlock: I think you're right that the svg files should be grouped together into an icons/ directory of some sort. I'll take care of that down the road when we've got some more of them.

Agreed. I thought about it some more yesterday and it would make sense to have the be colocated

PR #5291, opened 2019-03-10
bug 1401272: Enable MySQL Strict Mode

jwhitlock:
This will eliminate warning w001 from [django-mysql](http://django-mysql.readthedocs.org/en/latest/checks.html#django-mysql-w001-strict-mode), and the default in MySQL 5.7.


jwhitlock:
Tests passed locally and in TravisCI, so I have more confidence in this change. I'd prefer to deploy this to stage on its own, such as merging and deploying to stage after a production push, to give it some time in a production-like environment before it goes to production.

PR #5292, opened 2019-03-11
Bug 1530203, expanded arrow not visible on mobile

schalkneethling:
This resolves the bug on mobile where the expanded arrow is not shown:

## Before

Screenshot 2019-03-11 at 11 06 30

## After

Screenshot 2019-03-11 at 11 06 54

@davidflanagan r?


schalkneethling:
> I suspect that this may do the wrong thing for /ar/ and other RTL locales. If I understand correctly, I'd expect to see an up arrow for those locales when the section is expanded.

Good point. Let me test this and add another rule for RTL that rotates -90deg.

schalkneethling:
@davidflanagan Alrighty, updated with fixes for RTL locales.

schalkneethling:
Thanks @davidflanagan 

>  I'm assuming that the changes to the chevron-right.svg path were done by some kind of svg optimizer

Yup. I use SVGO for that.

PR #5293, opened 2019-03-11
Bug 1534298, ensure contributeForm exist before adding event listener

schalkneethling:
This solves the initial problem but, I ran into another problem. Once I confirmed that I want to cancel my subscription, I am sent to this URL https://developer.mozilla.org/en-US/payments/recurring/management but, here I get a permission denied error.

Going back to https://developer.mozilla.org/en-US/payments/recurring/management via Edit profile > Manage monthly subscription still shows the subscription.

That might simply be because it is still active, but maybe not? If it is because it is still active, we probably want to look at adding a note that indicates that it is in fact cancelled, and perhaps the date on which the subscription will end.

That is out of scope for this bug though, but thought I would mention it.

@jwhitlock @atopal Thoughts?

@davidflanagan r?

### DIFF
--- a/kuma/static/js/components/payments/focus-visible.js
+++ b/kuma/static/js/components/payments/focus-visible.js
@@ -10,40 +10,45 @@
         this.removeEventListener('focusout', removeFocusVisible);
     }
 
-    // When a keyup event is triggered on the contribute form,
-    contributeForm.addEventListener('keyup', function(event) {
-        var currentTarget = event.target;
-        var targetType = currentTarget.type;
+    if (contributeForm) {
+        // When a keyup event is triggered on the contribute form,
+        contributeForm.addEventListener('keyup', function(event) {
+            var currentTarget = event.target;
+            var targetType = currentTarget.type;
 
-        /* The key pressed was the tab key, or the left or right
-           arrow keys(used to cycle through radio buttons), and the
-           target field type is one of `radio` */
-        if (
-            (event.key === 'Tab' ||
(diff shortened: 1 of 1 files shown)


kuma/static/js/components/payments/focus-visible.js
```diff
@@ -10,40 +10,45 @@
         this.removeEventListener('focusout', removeFocusVisible);
     }
 
-    // When a keyup event is triggered on the contribute form,
-    contributeForm.addEventListener('keyup', function(event) {
-        var currentTarget = event.target;
-        var targetType = currentTarget.type;
+    if (contributeForm) {
```
davidflanagan: Given that the body of this if statement is basically then entire file, consider up at line 4 just adding `if (!contributeForm) return;`

schalkneethling: Doh! only saw this after the merge. That is much better. Additional PR coming your way.

atopal:
> > We need to be more attentive to JS errors in development and production, since they have unpredictable effects when assets are combined.
> 
> @jwhitlock I agree. I also believe we send these to GA? Is there a custom GA query or dashboard I can monitor?

Yes, there is: https://analytics.google.com/analytics/web/#/report/content-event-events/a36116321w64965862p66726481/_r.drilldown=analytics.eventCategory:JavaScript%20Error&explorer-table.plotKeys=%5B%5D

I don't think anyone is looking at that though. Not sure how useful it is. Probably not that useful for low frequency issues.

PR #5294, opened 2019-03-11
bug 1519582: bleach revision content on save

escattone:
The intent of this PR is to bring more unity to the rendering of documents among the various endpoints of Kuma (see https://kuma.readthedocs.io/en/latest/rendering.html#future-changes). It takes the step of cleaning the content that's saved with a revision of any document, where the cleaning does the following:
- bleaches the content
- adds section ID's
- filters the sources of `iframe` elements
- alphabetically orders the attributes of elements

That change is small. Most of the other changes in this PR are related to updating the tests as well as adding a few new tests in light of the change to the revision content.

I've also added `TODO` comments within the code to note areas that could be simplified after/if this code is deployed and all existing documents have been updated with a cleaned current revision. So this PR is the first step in the following process:

1. Review/merge/deploy this PR
1. Create a formal or informal management command that creates a new, cleaned current revision for each document
1. Create/review/deploy another PR to clean-up the code as noted in the `TODO` comments

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -172,19 +172,14 @@ class Extractor(object):
 
 def clean_content(content):
     """Clean content with standard bleaching and filtering."""
-    bleached = bleach.clean(content,
-                            attributes=ALLOWED_ATTRIBUTES,
-                            tags=ALLOWED_TAGS,
-                            styles=ALLOWED_STYLES,
+    bleached = bleach.clean(content, attributes=ALLOWED_ATTRIBUTES,
+                            tags=ALLOWED_TAGS, styles=ALLOWED_STYLES,
                             protocols=ALLOWED_PROTOCOLS)
     parsed = parse(bleached)
-    if settings.ALLOW_ALL_IFRAMES:
-        filtered = parsed
-    else:
-        allowed_iframe_patterns = settings.ALLOWED_IFRAME_PATTERNS
-        filtered = parsed.filterIframeHosts(allowed_iframe_patterns)
-    content_out = filtered.serialize()
-    return content_out
+    parsed.injectSectionIDs()
+    if not settings.ALLOW_ALL_IFRAMES:
+        parsed.filterIframeHosts(settings.ALLOWED_IFRAME_PATTERNS)
+    return parsed.serialize()
 
 
 @newrelic.agent.function_trace()
@@ -300,8 +295,11 @@ class ContentSectionTool(object):
 
         self._serializer = None
         self._default_serializer_options = {
-            'omit_optional_tags': False, 'quote_attr_values': 'always',
-            'escape_lt_in_attrs': True}
+            'omit_optional_tags': False,
+            'quote_attr_values': 'always',
+            'escape_lt_in_attrs': True,
+            'alphabetical_attributes': True,
+        }
         self._serializer_options = None
         self.walker = html5lib.treewalkers.getTreeWalker("etree")
 
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -522,12 +522,10 @@ class RevisionForm(AkismetCheckFormMixin, forms.ModelForm):
                 self.initial['slug'] = self.instance.document.slug
 
             content = self.instance.content
-            parsed_content = kuma.wiki.content.parse(content)
-            parsed_content.injectSectionIDs()
             if self.section_id:
+                parsed_content = kuma.wiki.content.parse(content)
                 parsed_content.extractSection(self.section_id)
-            parsed_content.filterEditorSafety()
-            content = parsed_content.serialize()
+                content = parsed_content.serialize()
             self.initial['content'] = content
 
             self.initial['review_tags'] = list(self.instance
--- a/kuma/wiki/management/commands/clean_document.py
+++ b/kuma/wiki/management/commands/clean_document.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+"""
+Manually schedule the cleaning of one or more documents
+"""
+from __future__ import division
+from __future__ import unicode_literals
+
+from math import ceil
+
+from celery import chain
+from django.contrib.auth.models import Group
+from django.core.management.base import BaseCommand, CommandError
+
+from kuma.core.utils import chunked
+from kuma.users.models import User
+from kuma.wiki.models import Document
+from kuma.wiki.tasks import (clean_document_chunk, email_document_progress)
+
+
+class Command(BaseCommand):
+    args = '<document_path document_path ...>'
+    help = 'Clean the current revision of one or more wiki documents'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'paths',
+            help='Path to document(s), like /en-US/docs/Web',
+            nargs='*',  # overridden by --all or --locale
+            metavar='path')
+        parser.add_argument(
+            '--all',
+            help='Clean ALL documents (rather than by path)',
+            action='store_true')
+        parser.add_argument(
+            '--locale',
+            help='Clean ALL documents in this locale (rather than by path)')
+
+    def handle(self, *args, **options):
+        user = get_or_create_known_user('mdnwebdocs-bot')
+        if options['all'] or options['locale']:
+            filters = {}
+            if options['locale'] and not options['all']:
+                locale = options['locale'].decode('utf8')
+                self.stdout.write(
+                    'Cleaning all documents in locale {}'.format(locale))
+                filters.update(locale=locale)
+            else:
+                self.stdout.write('Cleaning all documents')
+            docs = Document.objects.filter(**filters)
+            docs = docs.order_by('-modified')
+            docs = docs.values_list('id', flat=True)
+            self.stdout.write('...found {} documents.'.format(len(docs)))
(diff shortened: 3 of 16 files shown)


kuma/wiki/models.py
```diff
@@ -323,12 +323,16 @@ def __str__(self):
 
     @cache_with_field('body_html')
     def get_body_html(self, *args, **kwargs):
-        html = self.rendered_html and self.rendered_html or self.html
+        html = self.rendered_html or self.html
```
escattone: Sorry, this is not really necessary for this PR. It just bugged me! 😄

kuma/wiki/models.py
```diff
@@ -343,7 +347,11 @@ def get_toc_html(self, *args, **kwargs):
             return ''
         if not self.current_revision.toc_depth:
             return ''
-        html = self.rendered_html and self.rendered_html or self.html
+        html = self.rendered_html or self.html
```
escattone: Again, this is not really necessary for this PR. It just bugged me! 😄

kuma/wiki/models.py
```diff
@@ -612,7 +620,11 @@ def get_summary(self, strip_markup=True, use_rendered=True):
         return get_seo_description(src, self.locale, strip_markup)
 
     def build_json_data(self):
-        html = self.rendered_html and self.rendered_html or self.html
+        html = self.rendered_html or self.html
```
escattone: One more! 😄

kuma/wiki/jinja2/wiki/revision.html
```diff
@@ -56,7 +56,7 @@ <h1>{{ _('Revision %(id)s of <a href="%(url)s">%(title)s</a>'|safe, id=revision.
     <details open="open">
       <summary><h2>{{ _('Revision Content') }}</h2></summary>
       <article id="wikiArticle" class="text-content">
-        {{ revision.content_cleaned|safe }}
+        {{ revision.content|safe }}
```
jwhitlock: There are XSS attempts in historical revisions that need to be cleaned, such as the ones reported in [bug 821988](https://bugzilla.mozilla.org/show_bug.cgi?id=821988). We may be able to assume clean content going forward from when this ships, but we can't trust current content.

To emulate an old revision in the dev environment:

1. Edit a local page, get the revision ID
2. Run this in the Django shell:

```
rev_id = 123456 # Or whatever your ID is
content = '<p onclick="alert(1)">Click me!</p>'
Revision.objects.filter(id=rev_id).update(content=content)
```
3. Load the revision, click "Click me!"

kuma/wiki/models.py
```diff
@@ -1735,10 +1749,6 @@ def get_tidied_content(self, allow_none=False):
         self.tidied_content = tidied_content or ''
         return tidied_content
 
-    @property
-    def content_cleaned(self):
```
jwhitlock: I think we'll need this for historical revisions. Once this ships, we can try to distinguish between new revisions and unsafe old revisions, using a revision ID, a new boolean, or other method.

kuma/wiki/tests/test_content.py
```diff
@@ -1031,16 +1032,28 @@ def test_clean_content_preserves_whitespace():
...
+      ('<video class="movie" controls id="some-movie" lang="en-US" '
+       'src="some-movie.mpg">Fallback</video>')),
      ))
-def test_clean_content_allows_some_attributes(html):
-    """clean_content allows attributes, but doesn't preserve order."""
-    assert normalize_html(clean_content(html)) == html
+def test_clean_content_allows_some_attributes(html, expected):
+    """
+    clean_content allows attributes, orders them alphabetically, and
+    normalizes whitespace between them.
```
jwhitlock: 👍 Nice update to the test cases

kuma/wiki/jinja2/wiki/translate.html
```diff
@@ -93,10 +93,10 @@ <h3 class="approved-title">{{ _('Approved %(locale)s version:', locale=default_l
                     </div>
                 </header>
                 <div class="boxed translate-rendered text-content">
-                  {{ based_on.content_cleaned|safe }}
+                  {{ based_on.content|safe }}
```
jwhitlock: Similar to the revision view, we can't trust old content. Translations may be way out of date, so ``based_on`` could be ancient.

kuma/wiki/tests/test_content.py
```diff
@@ -964,7 +964,7 @@ def test_editor_safety_filter(self):
             <svg><circle onload=confirm(3)>
             <h1 class="header1">Header One</h1>
             <p>test</p>
-            <section>
+            <section onclick="alert('hacked!')">
```
jwhitlock: I'm surprised this was needed - ``filterEditorSafety`` is supposed to remove it, which I think it is whole purpose of this test.

kuma/wiki/forms.py
```diff
@@ -522,12 +522,10 @@ def __init__(self, *args, **kwargs):
                 self.initial['slug'] = self.instance.document.slug
 
             content = self.instance.content
-            parsed_content = kuma.wiki.content.parse(content)
-            parsed_content.injectSectionIDs()
             if self.section_id:
+                parsed_content = kuma.wiki.content.parse(content)
                 parsed_content.extractSection(self.section_id)
-            parsed_content.filterEditorSafety()
```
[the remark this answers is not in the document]
jwhitlock: To be clear, I agree with this change. I just needed to talk myself into it.

kuma/wiki/content.py
```diff
@@ -172,19 +172,14 @@ def code_sample(self, name):
...
-        filtered = parsed
-    else:
-        allowed_iframe_patterns = settings.ALLOWED_IFRAME_PATTERNS
-        filtered = parsed.filterIframeHosts(allowed_iframe_patterns)
-    content_out = filtered.serialize()
-    return content_out
+    parsed.injectSectionIDs()
+    if not settings.ALLOW_ALL_IFRAMES:
+        parsed.filterIframeHosts(settings.ALLOWED_IFRAME_PATTERNS)
+    return parsed.serialize()
```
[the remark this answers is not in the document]
escattone: I was tempted to use `tidylib.tidy_document` in this pass, but decided I didn't want to think through the consequences at the time.

kuma/wiki/forms.py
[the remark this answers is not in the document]
escattone: I thought about this change quite a bit as well. In the end, I concluded that `bleach.clean` included functionality effectively similar to `filterEditorSafety` (i.e., strips unapproved attributes like `onclick`, and although there are cases with `bleach.clean` where the "on*" attributes are not stripped, that only happens when the element they're associated with has been escaped, like for the case when `<svg><circle onload=confirm(3)>` becomes `&lt;svg&gt;&lt;circle onload="confirm(3)"&gt;&lt;/circle&gt;&lt;/svg&gt;`).

kuma/wiki/jinja2/wiki/revision.html
escattone: Oh man, good catch, sorry I missed this!

kuma/wiki/jinja2/wiki/translate.html
escattone: Again, good catch, and sorry I missed this.

kuma/wiki/tests/test_content.py
escattone: On second look, you're right, it doesn't add much new here, other than adding a different flavor of `on*` attribute to the one already there (`onload=confirm(3)`). I must have been thinking in terms of how this contrasts with `bleach.clean`, which would escape the element with the `onload=confirm(3)` but strip the `onclick`. If you don't mind, I think I'll leave it though.

kuma/wiki/forms.py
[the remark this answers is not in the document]
escattone: To be honest, I only now realize that I had not thought this through as thoroughly as you have done. Thanks for the detailed explanation!

kuma/wiki/tests/test_content.py
jwhitlock: OK, we can leave it, but this whole test should be removed when we remove ``filterEditorSafety``.

escattone:
Thanks for the great review @jwhitlock. This is ready for review again. I've made the following changes to the original commit:
- I've added back the `content_cleaned` property on the `Revision` class (with new comments)
- Removed all of the changes I made to `kuma/wiki/jinja2/wiki/revision.html` and `kuma/wiki/jinja2/wiki/translate.html` (they're back to using the `content_cleaned` property)
and I've added a new commit that provides a `clean_document` management command. I've also manually created the new `mdnwebdocs-bot` user (in both stage and production) that this new management command uses.

kuma/wiki/management/commands/clean_document.py
```diff
@@ -0,0 +1,94 @@
...
+        parser.add_argument(
+            '--all',
+            help='Clean ALL documents (rather than by path)',
+            action='store_true')
+        parser.add_argument(
+            '--locale',
+            help='Clean ALL documents in this locale (rather than by path)')
+
+    def handle(self, *args, **options):
+        user = User.objects.get(username='mdnwebdocs-bot')
```
jwhitlock: Nit: This user could be created if it doesn't exist, which would make it easier to ensure it doesn't have a password, is in known authors, etc.

PR #5295, opened 2019-03-12
Bug 970278: Add noindex, nofollow tag to deprecated MindTouch pages

davidflanagan:
We've got various pages with slugs that begin with User:, Talk:,
User_talk:, and Template_talk: in our wiki that we really don't
want to support any more. We're not quite ready to delete them
but we at least don't want them to be turn up in search results.
This PR gives these pages a <meta name="robots" content="noindex,nofollow">
tag to prevent crawlers from indexing or crawling them.

Note that (per bug 1443179) we, in general, do not want to use
nofollow on pages. But in this case, since we're almost ready to
delete these pages, I don't think we want to rely on any links
they contain.

### DIFF
--- a/kuma/wiki/constants.py
+++ b/kuma/wiki/constants.py
@@ -201,6 +201,19 @@ ALLOWED_PROTOCOLS = [
 
 DIFF_WRAP_COLUMN = 65
 EXPERIMENT_TITLE_PREFIX = 'Experiment:'
+LEGACY_MINDTOUCH_NAMESPACES = (
+    'Help',
+    'Help_talk',
+    'Project',
+    'Project_talk',
+    'Special',
+    'Talk',
+    'Template',
+    'Template_talk',
+    'User',
+    'User_talk'
+)
+
 DOCUMENTS_PER_PAGE = 100
 _ks_urlbits = urlparse(settings.KUMASCRIPT_URL_TEMPLATE)
 KUMASCRIPT_BASE_URL = urlunparse((_ks_urlbits.scheme, _ks_urlbits.netloc,
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -30,6 +30,7 @@ from kuma.spam.models import AkismetSubmission, SpamAttempt
 
 from . import kumascript
 from .constants import (DEKI_FILE_URL, EXPERIMENT_TITLE_PREFIX, KUMA_FILE_URL,
+                        LEGACY_MINDTOUCH_NAMESPACES,
                         REDIRECT_CONTENT, REDIRECT_HTML)
 from .content import (clean_content, Extractor, get_content_sections,
                       get_seo_description, H2TOCFilter, H3TOCFilter,
@@ -1430,6 +1431,11 @@ Full traceback:
     @property
     def is_experiment(self):
         return self.slug.startswith(EXPERIMENT_TITLE_PREFIX)
+
+    @property
+    def has_legacy_namespace(self):
+        namespace, separator, _ = self.slug.partition(':')
+        return namespace in LEGACY_MINDTOUCH_NAMESPACES if separator else False
 
     def get_hreflang(self, other_locales=None):
         """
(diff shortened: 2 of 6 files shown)


davidflanagan:
I'm guessing that I'm going to have test failures, and if so I'll fix those up tomorrow.

Note that even though we don't want to use nofollow generally on our noindex pages, these pages, which are candidates for deletion, seem ideal for a nofollow. If a content page is only linked to via a User: or Talk: page, then I assume we don't want google to find it right? So I *think* that nofollow is the right way to go here.

davidflanagan:
Note that testing this PR locally is difficult because jinja2/base.html puts noindex,nofollow on every page if the hostname is not developer.mozilla.org. I couldn't figure out how to change this default via an environment variable, so in order to test locally, I had to edit base.html so that it would always output the robots_value block

kuma/wiki/constants.py
```diff
@@ -201,6 +201,7 @@
 
 DIFF_WRAP_COLUMN = 65
 EXPERIMENT_TITLE_PREFIX = 'Experiment:'
+DEPRECATED_PAGE_NAMESPACES = ('User:', 'Talk:', 'User_talk:', 'Template_talk:')
```
jwhitlock: Nit: It would be nice if there was one source for all of these namespaces. Now we have a list is kuma/wiki/views/legacy.py:

https://github.com/mozilla/kuma/blob/2ee1e66dec001d4ddf1c9fa0a32929d33faca39c/kuma/wiki/views/legacy.py#L13-L23

and a slightly different list in managers.py:

https://github.com/mozilla/kuma/blob/2ee1e66dec001d4ddf1c9fa0a32929d33faca39c/kuma/wiki/managers.py#L26-L33

and in search.py:

https://github.com/mozilla/kuma/blob/2ee1e66dec001d4ddf1c9fa0a32929d33faca39c/kuma/wiki/search.py#L30-L31

kuma/wiki/models.py
```diff
@@ -1431,6 +1432,13 @@ def get_full_url(self):
     def is_experiment(self):
         return self.slug.startswith(EXPERIMENT_TITLE_PREFIX)
 
+    @property
+    def is_namespace_page(self):
```
jwhitlock: A test is needed in ``test_models.py`` for this new function. The test in ``test_search`` would be a decent template for the test:

https://github.com/mozilla/kuma/blob/2ee1e66dec001d4ddf1c9fa0a32929d33faca39c/kuma/wiki/tests/test_search.py#L28-L35

If you aren't comfortable writing a test, @escattone or I can take that on.

davidflanagan: Thanks, John. I'm happy to write the test. I just ran out of time last night and figured that Travis would show me failing tests that I could then fix. I'll get something into models.py

davidflanagan:
@jwhitlock: I've requested your review again. This new version uses the list of namespaces from legacy.py (but moves them to constants.py). And it has tests.

kuma/wiki/models.py
```diff
@@ -1431,6 +1432,11 @@ def get_full_url(self):
     def is_experiment(self):
         return self.slug.startswith(EXPERIMENT_TITLE_PREFIX)
 
+    @property
+    def has_legacy_namespace(self):
+        namespace, separator, _ = self.slug.partition(':')
```
jwhitlock: Today I learned about [partition](https://docs.python.org/2/library/stdtypes.html#str.partition). I usually use ``namespace = self.slug.split(':', 1)[0]``, but ``partition`` is a good way to both split the string and detect that the partition character is in the string.

kuma/wiki/tests/test_models.py
```diff
@@ -40,6 +40,30 @@ def test_document_is_experiment():
...
+    ('Template:domxref', True),
+    ('Template_talk:domxref', True),
+    ('User:jezdez', True),
+    ('User_talk:jezdez', True),
+    # Experiments aren't legacy yet
+    ('Experiment:Blue', False),
+    # Slugs without colons don't have namespaces
+    ('CSS', False),
+    # Slugs with colons might not be legacy
+    (':hover', False)
```
jwhitlock: Nice set of tests!

PR #5300, opened 2019-03-13
Bug 1525719: remove eyeball and gear menus, move languages and edit to header

davidflanagan:
This removes server-side user-dependent code from the jinja template
for our /ducks/ URL and simplifies the UX the user sees by removing
controls from wiki page titles and moving some of them to the main
header. The subscribe and advanced menus are removed. The language
menu and the edit button are moved to the header. The edit link is
only visible when the user is logged in. Then language menu is not
yet implemented (because it will need a document API to work with
client-side navigation).

Additionally, this patch also consolidates all of the SVG icon files
in a new icons/ directory instead of having them in the same directory
as the .jsx files.


davidflanagan:
Sorry about the big snapshot file changes; I think you can ignore those.

In addition to moving a couple of .svg images, I also moved the Row, Spring and Strut components out of header.jsx into a layout.jsx file of their own so that they can be shared.

davidflanagan:
Here's what the page header looks like with these changes:

Screen Shot 2019-03-12 at 5 39 49 PM

davidflanagan:
Thanks for the quick review, Ryan. The typo will go away in the next PR when I start using the document API, so I'm just going to merge this in its current state.

PR #5301, opened 2019-03-14
Bug 1525719: Add a document API at /api/v1/doc/<locale>/<slug>.

davidflanagan:
This PR implements a simple read-only API for querying document
content and metadata as JSON. The intent is to use this API to
enable client-side navigation for MDN documents.

### DIFF
--- a/kuma/urls.py
+++ b/kuma/urls.py
@@ -138,7 +138,10 @@ else:
 
 urlpatterns += [
     # User API service
-    url('api/v1/', include('kuma.users.api.v1.urls')),
+    url('^api/v1/', include('kuma.users.api.v1.urls')),
+
+    # Document API service
+    url('^api/v1/', include('kuma.wiki.api.v1.urls')),
 
     # Services and sundry.
     url('', include('kuma.version.urls')),
--- a/kuma/wiki/api/v1/urls.py
+++ b/kuma/wiki/api/v1/urls.py
@@ -1,0 +1,10 @@
+from django.conf.urls import url
+
+from . import views
+
+
+urlpatterns = [
+    url(r'^doc/(?P<locale>[^/]+)/(?P<slug>.*)$',
+        views.doc,
+        name='wiki.api.doc'),
+]
--- a/kuma/landing/views.py
+++ b/kuma/landing/views.py
@@ -56,6 +56,7 @@ User-agent: *
 Sitemap: https://developer.mozilla.org/sitemap.xml
 
 Disallow: /admin/
+Disallow: /api/
 Disallow: /*/dashboards/*
 Disallow: /*docs/feeds
 Disallow: /*docs/templates
--- a/kuma/wiki/api/v1/views.py
+++ b/kuma/wiki/api/v1/views.py
@@ -1,0 +1,48 @@
+from django.http import JsonResponse
+from django.shortcuts import get_object_or_404
+from django.views.decorators.cache import never_cache
+from django.views.decorators.http import require_GET
+
+from kuma.wiki.models import Document
+
+
+@never_cache
+@require_GET
+def doc(request, locale, slug):
+    """
+    Return a JSON object that includes document content and metadata
+    for the document specified by the locale and path. Raises a 404
+    error if no such document exists. This is an API with URL
+    /api/v1/doc/<locale>/<path>
+    """
+    document = get_object_or_404(Document, locale=locale, slug=slug)
+    translations = document.get_other_translations(
+        fields=('locale', 'slug', 'title'))
+
+    return JsonResponse({
+        'locale': locale,
+        'slug': slug,
+        'id': document.id,
+        'title': document.title,
+        'summary': document.get_summary_html(),
+        'language': document.language,
+        'absoluteURL': document.get_absolute_url(),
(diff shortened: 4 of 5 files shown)


davidflanagan:
I still need to add a test, but I wanted to get this first cut of an API in for review to make sure this is what you were both imagining the API and its implementation would look like.

kuma/wiki/views/document.py
```diff
@@ -1107,3 +1107,43 @@ def _document_api_PUT(request, document_slug, document_locale):
         resp.status_code = 205
 
     return resp
+
+
+@never_cache
+@require_GET
+def doc_api(request, locale, path):
```
escattone: I'm wondering if we should move this into a new file `kuma/wiki/api/v1/view.py`? Besides the fact that this function has a name similar enough to `document_api` to perhaps be a little confusing, I'm thinking that it would be nice to separate this into its own file so that we could more easily move it later (in another PR). For example, if we decide, as @jwhitlock suggested [here](https://github.com/mozilla/kuma/pull/5285#pullrequestreview-211928950), to gather all of the `/api/v1` endpoints within their own folder like `/kuma/api/v1`.

kuma/core/middleware.py
```diff
@@ -166,7 +166,9 @@ def process_response(self, request, response):
         urlconf = getattr(request, 'urlconf', settings.ROOT_URLCONF)
 
         # Kuma: assume locale-prefix patterns, including default language
-        if response.status_code == 404 and not language_from_path:
+        if (response.status_code == 404 and
+                not language_from_path and
+                not request.path_info.startswith("/api/")):
```
escattone: 👍 Good catch.

kuma/wiki/views/document.py
```diff
@@ -1107,3 +1107,43 @@ def _document_api_PUT(request, document_slug, document_locale):
...
+                'url': d.get_absolute_url(),
+                'title': d.title
+            } for d in document.parents
+        ],
+        'translations': [
+            {
+                'locale': t.locale,
+                'url': t.get_absolute_url(),
+                'title': t.title
+            } for t in document.other_translations
```
escattone: Since you're only using the `locale`, `url`, and `title` fields of each of the translations, it's more efficient to replace this:
```
document.other_translations
``` 
with this:
```
document.get_other_translations(fields=('locale', 'url', 'title'))
```
in order to avoid the cost of transferring all of the other fields (especially, the content fields like `html` and `rendered_html`) from the database (the `other_translations` property is only cached on the document instance, so doesn't help us in this context).

kuma/wiki/views/document.py
escattone: I was thinking also that perhaps `slug` would be a better name than `path`, only because it matches our current usage?

kuma/wiki/api/v1/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.conf.urls import url
+
+from kuma.wiki.views import document
+
+
+urlpatterns = [
+    url(r'^doc/(?P<locale>[^/]+)/(?P<path>.*)$',
```
escattone: Perhaps `slug` would be a better name than `path`, only because it matches our current usage?

kuma/wiki/views/document.py
davidflanagan: I'll do both of those things. I started with a views.py file, then thought I was going to end up using utility functions defined in document.py, so moved it there. But the implementation turned out to be simpler, so now there is no reason not to have it in its own file. Thanks for pointing that out.

kuma/wiki/api/v1/urls.py
davidflanagan: Done.

kuma/wiki/views/document.py
davidflanagan: Done

kuma/wiki/api/v1/urls.py
davidflanagan: Done

kuma/core/middleware.py
jwhitlock: This doesn't seem right to me. We have other 'no locale' paths that don't need a special case, and we've stopped using ``LANGUAGE_URL_IGNORED_PATHS`` in the last 12 months. I'm seeing if there is a better place for this logic, but I can fix in a later PR if I'm slow.

kuma/urls.py
```diff
@@ -140,6 +140,9 @@ def serve_from_media_root(request, path):
     # User API service
     url('api/v1/', include('kuma.users.api.v1.urls')),
 
+    # Document API service
+    url('api/v1/', include('kuma.wiki.api.v1.urls')),
```
jwhitlock: A caret ``^`` is needed to restrict this to the start of the path

```suggestion
    url('^api/v1/', include('kuma.wiki.api.v1.urls')),
```

Otherwise, URLs like ``/zzz/api/v1/doc/en-US/Web`` will match. The User API line needs a similar change, to prevent https://developer.allizom.org/garbage/api/v1/whoami from matching.

Once this is done, the middleware no longer needs a special case.

davidflanagan: I fixed both in the latest push.

PR #5302, opened 2019-03-14
bug 1525719: Extract strings from React, use existing translations

jwhitlock:
This builds on PR #5279, creating a new translations domain ``react`` that includes all of the ``javascript`` strings, as well as the React strings that are very likely to appear in ``django``. 

This requires the changes in https://github.com/mozilla-l10n/mdn-l10n/commit/bfc4155109fcb1a264454d51d3f01ef3a7d29279, which include an updated ``.gitignore``. This was merged to master with the latest push to production.

``make localerefresh`` now calls the new target ``make locale-populate-react``, which collects existing translations into [compendium](https://www.gnu.org/software/gettext/manual/html_node/Compendium.html#Compendium) files, which are used to populate ``react.po`` from existing translations, without sending them to Pontoon.

``make build-static`` now calls the new target ``make compile-react-i18n``, which works the same as ``make compilejsi18n``, but for the new ``react`` domain. A new Jinja2 command ``react_i18n`` returns the URL of these ``react``-domain JS files.

To get things started, run this in the web container:

```
make locale-populate-react localecompile compile-react-i18n collectstatic
# make localerefresh # Alternatively, run everything
```

You can then see the translated strings in the React app, such as the [French Learn page](http://localhost:8000/fr/ducks/Apprendre):

react-fr

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,11 +51,16 @@ compilejsi18n:
 	@ mkdir -p build/locale
 	@ python manage.py compilejsi18n
 
+compile-react-i18n:
+	@ echo "## Generating React translation catalogs ##"
+	@ mkdir -p build/locale
+	@ python manage.py compilejsi18n -d react
+
 collectstatic:
 	@ echo "## Compiling (Sass), collecting, and building static files ##"
 	@ python manage.py collectstatic --noinput
 
-build-static: webpack compilejsi18n collectstatic
+build-static: webpack compilejsi18n compile-react-i18n collectstatic
 
 install:
 	@ echo "## Installing $(requirements) ##"
@@ -81,10 +86,19 @@ localeextract:
 	python manage.py extract
 	python manage.py merge
 
+locale-populate-react:
+	@ mkdir -p locale/compendia
+	for localename in `find locale -mindepth 1 -maxdepth 1 -type d | cut -d/ -f2 | grep -v templates | grep -v compendia`; do \
+		rm -f locale/compendia/$$localename.compendium; \
+		msgcat --use-first -o locale/compendia/$$localename.compendium locale/$$localename/LC_MESSAGES/django.po locale/$$localename/LC_MESSAGES/javascript.po; \
+		rm -f locale/$$localename/LC_MESSAGES/react.po; \
+		msgmerge --compendium locale/compendia/$$localename.compendium -o locale/$$localename/LC_MESSAGES/react.po /dev/null locale/templates/LC_MESSAGES/react.pot; \
+	done
+
 localecompile:
 	cd locale; ../scripts/compile-mo.sh .
 
-localerefresh: localeextract localetest localecompile build-static
+localerefresh: localeextract locale-populate-react localetest localecompile build-static
 
 pull-base:
 	docker pull ${BASE_IMAGE}
@@ -157,4 +171,4 @@ npmrefresh:
 	npm install
 
 # Those tasks don't have file targets
-.PHONY: test coveragetest locust clean locale install compilejsi18n collectstatic localetest localeextract localecompile localerefresh npmrefresh webpack
+.PHONY: test coveragetest locust clean locale install compilejsi18n collectstatic localetest localeextract localecompile localerefresh npmrefresh webpack compile-react-i18n locale-populate-react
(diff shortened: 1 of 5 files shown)


jwhitlock:
I fixed a docker-only path in the Makefile, and added the missing ``make localecompile`` to the instructions.  This is now working locally, and with the latest commit, we'll get translations on staging.

Makefile
```diff
@@ -81,10 +86,19 @@ localeextract:
 	python manage.py extract
 	python manage.py merge
 
+locale-populate-react:
```
davidflanagan: It took me a while to figure out what this code is doing. Consider adding a comment here to explain the concatenation and extraction going on with msgcat and msgmerge. 

It would also be useful to explain (because I'm not entirely clear) what the bottom line is here. I'm pretty sure I understand that this means that React components can use any string that is already in use by python or the existing static javascript files, and no new work will be created for translators.  I'm not sure what happens if our React code starts using new strings... Will those end up on Pontoon for translation? Or will they just remain untranslated for now?

Makefile
```diff
@@ -81,10 +86,19 @@ localeextract:
...
+	@ mkdir -p locale/compendia
```
davidflanagan: If you add an `rmdir -rf` at the end of this command, then you won't have to add that in the Dockerfile below...  It doesn't look like the compendium files get used after they are created, so it is unclear to me that it is worth keeping them around. I suppose you could also use temp files to hold the compendia... But creating a special-purpose directory as you do here seems simpler to me.

Makefile
```diff
@@ -81,10 +86,19 @@ localeextract:
...
+	for localename in `find locale -mindepth 1 -maxdepth 1 -type d | cut -d/ -f2 | grep -v templates | grep -v compendia`; do \
+		rm -f locale/compendia/$$localename.compendium; \
+		msgcat --use-first -o locale/compendia/$$localename.compendium locale/$$localename/LC_MESSAGES/django.po locale/$$localename/LC_MESSAGES/javascript.po; \
+		rm -f locale/$$localename/LC_MESSAGES/react.po; \
```
davidflanagan: If a React component had used a new string, is it possible that this react.po file being deleted here could contain a translation of that new string that does not appear in the compendium? If so, then this deletion might be a problem...  Would it work to use this react.po (or a copy of it) instead of /dev/null in the msgmerge line below so that any React-only translations are not lost?

Makefile
```diff
@@ -81,10 +86,19 @@ localeextract:
 	python manage.py extract
 	python manage.py merge
 
+locale-populate-react:
+	@ mkdir -p locale/compendia
+	for localename in `find locale -mindepth 1 -maxdepth 1 -type d | cut -d/ -f2 | grep -v templates | grep -v compendia`; do \
```
escattone: For me, this `find` command returns the `locale/.git` directory as well, which in turn causes:
```
msgcat: error while opening "locale/.git/LC_MESSAGES/django.po" for reading: No such file or directory
msgmerge: error while opening "locale/compendia/.git.compendium" for reading: No such file or directory
```
when I run `make locale-populate-react`.

I modified the `find` command like this:
```
find locale -mindepth 1 -maxdepth 1 -type d -not -path locale/.git
```
and that got rid of the issue, but you may know a more elegant way to exclude the `locale/.git` directory.

jwhitlock:
I can add a comment explaining more about this code, and I'll try out moving the directory removal into the command. It may be next week.

I intend this to be a temporary step until the React-based code is used in production. During this beta period, ``react.po`` files are omitted in ``.gitignore``, which is how they are omitted from Pontoon. I expect new strings will not be translated, unless we duplicate them in another file used by Django or JavaScript. It shouldn't impact our development much.

When we're ready to ship this in production, I expect that strings will move from the ``django`` domain to the ``react`` domain, so they won't be duplicated anymore. We may even get to a point where webpack is processing all t.e files, so ``javascript`` goes away as well. The future is fuzzy, and I think the transition will be messy. This just gives us some translations to play with during beta,

When we're ready to go live, I expect to run this command one last time to populate ``react.po`` files, then check them into Pontoon. Translators may see that there are new strings, but many are already populated. Until then, there won't be any translations specific to the ``react`` domain, just borrowed translations.

PR #5306, opened 2019-03-18
Bug 1525719: implement a React frontend for the new document API

davidflanagan:
This PR is a followup to a previous patch that introduced the
/api/v1/doc/[locale]/[slug] API that returns document information
in a JSON blob. This patch adds frontend code that uses the new
API for our /ducks/ URLs on staging. Changes in this patch include:

- The backend API is refactored slightly to define a document_api_data
  function that can be called by other APIs.

- The react_document() function in wiki/views/document.py is modified
  to call document_api_data() and include that data in the context
  that it passes to the Jinja template. (I've also removed a lot of
  unneeded code from react_document() and no longer pass any HTML
  content to the template.)

- The react_document.html template is dramatically simplified. All the
  sections that previously held document content are replaced by a
  single react container div into which React will render the content
  on the frontend. The JSON data from the document api is base64
  encoded and included in a script inside this div so that the
  frontend code has immediate access to the data and does not need
  to fetch it.

- On the frontend, a new React Page component is responsible for
  rendering an MDN page including the header, the title bar, the
  TOC, the sidebar (breadcrumbs and quicklinks) and the main article
  content. (The footer and things like newsletter signup are still
  rendered by Jinja at this point.)

- The Page component gets the document data it needs via the React
  context mechanism. The new DocumentProvider component defines
  the context and makes document data available to Page and its
  sub-components. DocumentProvider also intercepts clicks on links to
  implement a client-side navigation mechanism were we simply fetch
  new document data from the document API instead of replacing the
  entire document.

- The old, hacky, client-side-nav.js implementation of client-side
  navigation is deleted.


kuma/wiki/views/document.py
```diff
@@ -865,106 +826,56 @@ def react_document(request, document_slug, document_locale):
...
+    other_translations = doc.get_other_translations(
+        fields=['title', 'locale', 'slug', 'parent']
+    )
+    all_locales = (set([doc.locale]) |
+                   set(trans.locale for trans in other_translations))
+
+    # Bundle it all up and, finally, return.
+    context = {
+        'document_api_data': base64.b64encode(
+            json.dumps(document_api_data(doc))),
```
davidflanagan: This is how I pass the JSON blob to the frontend when the document is first loaded, so that the frontend doesn't have to make a fetch() call to get it. The base64 encoding is because the data is going to go in an HTML <script> tag, and the data could include the string "</script>" which would screw things up.

kuma/wiki/views/document.py
```diff
@@ -799,48 +801,7 @@ def react_document(request, document_slug, document_locale):
...
-
-            # The user may be trying to create a child page; if a parent exists
-            # for this document, redirect them to the "Create" page
-            # Otherwise, they could be trying to create a main level doc.
-            create_url = _document_redirect_to_create(document_slug,
-                                                      document_locale,
-                                                      slug_dict)
-            response = redirect(create_url)
-            add_never_cache_headers(response)
-            return response
```
davidflanagan: All of this seemed like code that was specific to editing and unnecessary here.

kuma/wiki/views/document.py
```diff
@@ -865,106 +826,56 @@ def react_document(request, document_slug, document_locale):
...
     else:
-        toc_html = None
-    doc_html = _filter_doc_html(request, doc, doc_html, rendering_params)
+        en_slug = ''
 
-    if rendering_params['raw']:
-        response = _document_raw(doc_html)
-    else:
-        # Get the SEO summary
-        seo_summary = doc.get_summary_text()
```
davidflanagan: All the rendering_params code here and above seems like wiki-related functionality that we don't need for the static site.

kuma/wiki/views/document.py
```diff
@@ -865,106 +826,56 @@ def react_document(request, document_slug, document_locale):
...
+    share_text = ugettext(
+        'I learned about %(title)s on MDN.') % {"title": doc.title}
 
-        # Get the additional title information, if necessary.
-        seo_parent_title = _get_seo_parent_title(
-            original_doc, slug_dict, document_locale)
-
-        # Retrieve pre-parsed content hunks
-        quick_links_html = doc.get_quick_links_html()
-        body_html = doc.get_body_html()
```
davidflanagan: The HTML content is now in the JSON blob, so we don't need to pass it to the template this way.

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -83,264 +82,69 @@
...
-              </article>
-
-              <!-- contributors -->
-              <div class="wiki-block contributors">
-                <h2 class="offscreen">{{ _('Document Tags and Contributors') }}</h2>
-                {% set tags = document.all_tags_name %}
-                {% include "wiki/includes/document_tag.html" %}
+  <!-- Content will go here -->
+  <div id="react-container">
+    <script type="base64-encoded-json">{{ document_api_data }}</script>
```
davidflanagan: Here's where I hardcode the document API json payload so that react has all the data it needs to render without making a fetch() call.

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -1,68 +1,67 @@
-{% extends "wiki/react_base.html" %}
```
davidflanagan: This template was complicated and hard for me to understand, so I started with react_base and wiki/react_base and got rid of the extends calls and the block definitions. Then I cut things that were not needed for the static site and cut that parts of the template that were actually rendering document content, since React is taking that job.  What's left is mostly the document head, but also the footer at the bottom of every page and things like the newsletter and payments stuff that I still need to figure out. Big changes here, and probably many more changes still needed. But it seems to be in a reasonable working state for now.

kuma/wiki/api/v1/views.py
```diff
@@ -5,23 +5,13 @@
...
-@require_GET
-def doc(request, locale, slug):
-    """
-    Return a JSON object that includes document content and metadata
-    for the document specified by the locale and path. Raises a 404
-    error if no such document exists. This is an API with URL
-    /api/v1/doc/<locale>/<path>
-    """
-    document = get_object_or_404(Document, locale=locale, slug=slug)
+def document_api_data(document):
```
davidflanagan: Most of what was previously the body of the doc() request handler below is now in this utility document_api_data() function. I made this change so that I can also get the json blob in kuma/wiki/views/document.py

kuma/settings/common.py
```diff
@@ -940,15 +940,15 @@ def pipeline_one_scss(slug, **kwargs):
...
-            'js/main.js',
-            'js/auth.js',
-            'js/highlight.js',
-            'js/wiki-compat-trigger.js',
-            'js/lang-switcher.js',
+            #'js/main.js',
+            #'js/auth.js',
+            #'js/highlight.js',
+            #'js/wiki-compat-trigger.js',
+            #'js/lang-switcher.js',
```
davidflanagan: I'm experimenting here to see how much JS I can strip out. I think I've got a bit too far: I'm seeing errors in the console about missing jquery in js/analytics.js.

(And come to think of it, that will become a fatal error in production when these modules are squashed into one, so I need to fix it...)

kuma/javascript/src/page.jsx
```diff
@@ -0,0 +1,165 @@
...
+import DocumentProvider from './document-provider.jsx';
+import gettext from './gettext.js';
+import { Row, Strut } from './layout.jsx';
+import Header from './header/header.jsx';
+
+const strings = {
+    relatedTopics: gettext('Related Topics')
+};
+
+const styles = {
```
davidflanagan: These styles define the high-level structure of an MDN page, but we still rely on main.css and wiki.css to define styles for document content. At this point these styles only support a desktop layout and there is no customization for tablets and phones.

kuma/javascript/src/page.jsx
```diff
@@ -0,0 +1,165 @@
...
+function TOC() {
+    const documentData = useContext(DocumentProvider.context);
+
+    return (
+        documentData && (
+            <Row
+                css={styles.toc}
+                dangerouslySetInnerHTML={{
+                    __html: `<ol>${documentData.tocHTML}</ol>`
+                }}
```
davidflanagan: This dangerouslySetInnerHTML is how you inject HTML with React. They force you to type "dangerous" like Jinja forces you to type "safe".

kuma/javascript/src/page.jsx
```diff
@@ -0,0 +1,165 @@
...
+        <div>
+            <Header />
+            <Titlebar />
+            <TOC />
+            <Row css={styles.content}>
+                <Sidebar />
+                <Strut width="3%" />
+                <Article />
+            </Row>
+        </div>
```
davidflanagan: This is the rough structure of an MDN page, minus the Footer, which I haven't converted to React yet.

kuma/javascript/src/document-provider.jsx
```diff
@@ -0,0 +1,117 @@
...
+    const [documentData, setDocumentData] = useState(props.initialDocumentData);
+
+    if (!document.body) {
+        throw new Error('DocumentProvider rendered before a body exists.');
+    }
+    const body = document.body;
+
+    // A one-time effect that runs only on mount, to set up
+    // an event handler for client-side navigation
+    useEffect(() => {
```
davidflanagan: This is the function that sets up client-side navigation. The click event listener is what does most of the work.

kuma/wiki/views/document.py
```diff
@@ -799,48 +801,7 @@ def react_document(request, document_slug, document_locale):
...
-            if deleted_doc.exists():
-                return _document_deleted(request, deletion_log_entries)
-
-        # We can throw a 404 immediately if the request type is HEAD.
-        # TODO: take a shortcut if the document was found?
-        if request.method == 'HEAD':
-            raise Http404
-
-        # Check if we should fall back to default locale.
-        fallback_doc, fallback_reason, redirect_url = _default_locale_fallback(
```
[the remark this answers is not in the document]
davidflanagan: Thanks, Ryan. I was taking a weed wacker approach there, and just tossing out stuff wholesale. 
Eventually I think I'll also need to add back in some way of handling content experiments, which also got tossed here.

For the redirect case, I could try to convert /docs/ to /ducks/ in the URL. But I think I won't because that will just break when we switch to a beta subdomain.

davidflanagan:
@escattone I've added that code back in, rebased, squashed the commits and force-pushed the latest version.

Thanks for pointing out the redirect issue. It has made me realize that I'll need to do something like that for the document API.

davidflanagan:
And now I've made one more change....  Inspired by your comment I did some testing with french documents and ran into a case where I'd get a 404 from the document API which would cause the web page to dim, and then stay stuck in that dimmed state. Now, if the document API returns a 404 we just give up on the client-side navigation and attempt a regular page load. That should give us the locale fallback or redirect behavior or at least show the user a real 404 page.

escattone:
I restarted the docker tests (it was a temporary failure).

PR #5309, opened 2019-03-22
Bug 1525719: implement the translations menu, plus UX fixes

davidflanagan:
This patch implement the translations menu in the MDN page header.
See also https://github.com/mdn/sprints/issues/1207
In order to implement it, this patch also modifies the document API
to return the local and translated names of the languages displayed
in the menu.

In addition, it modifies the Dropdown component so that it always
renders its children so that they are visible to web crawlers and
screen readers even when the menu is closed and they are not
visually rendered on screen.

The patch also improves the appearance of menus and the header
titlebar and TOC bar at the top of the page. The MDN logo is now
a standalone SVG file (we previously used a sprite file that included
an unnecessary mask).


davidflanagan:
Here's a screenshot showing the improved (IMHO) design, and the translations menu

Screen Shot 2019-03-21 at 7 34 19 PM

There's no rush to review this. It looks like a kind of big patch, but it isn't really. The changes to the test snapshots are the bulk of it, and they shouldn't really need review. There are also a bunch of minor CSS changes, but not a lot of genuinely new code.

@escattone I've asked for your review specifically so you can look at my change to the document API in views.py. In particular, my use of `activate(locale)` to make _() work correctly for that API. (It is needed to translated english language names like "Arabic" into the locale of the document.

PR #5310, opened 2019-03-22
bug 1519582: ensure same slug & title when cleaning

escattone:
This PR adjusts the `Document.clean_current_revision` method to avoid two things:
- errors during the `Revision.make_current` call when the slug of the revision doesn't match the slug of the document (see https://github.com/mdn/sprints/issues/936#issuecomment-475694728)
- duplicate calls to `Revision.make_current` when the current revision is approved (most cases)

The tests for the `Document.clean_current_revision` method have also been updated to detect these changes.

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -1519,6 +1519,13 @@ Full traceback:
                 rev.based_on_id = prior_pk
             rev.comment = 'Clean prior revision of {} by {}'.format(
                 prior_created, prior_creator)
+            # The current revision sometimes has an old slug that's
+            # different than its document's current slug, so let's ensure
+            # that they're the same so we don't trigger unique index errors
+            # when the document is saved within the make_current() call made
+            # within the rev.save() call below, or explicitly further below.
+            rev.slug = self.slug
+            rev.title = self.title
             rev.save()
             if prior_review_tags:
                 rev.review_tags.set(*prior_review_tags)
@@ -1528,8 +1535,10 @@ Full traceback:
         # Populate the model instance with fresh data from database.
         rev.refresh_from_db()
 
-        # Make this new revision the current one for the document.
-        rev.make_current()
+        # Make this new revision the current one for the document if it hasn't
+        # been done already when saving the revision above.
+        if rev.pk != self.current_revision.pk:
+            rev.make_current()
         return rev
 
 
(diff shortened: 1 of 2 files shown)


kuma/wiki/models.py
```diff
@@ -1519,6 +1519,13 @@ def clean_current_revision(self, user):
                 rev.based_on_id = prior_pk
             rev.comment = 'Clean prior revision of {} by {}'.format(
                 prior_created, prior_creator)
+            # The current revision sometimes has an old slug that's
+            # different than its document's current slug, so let's ensure
+            # that they're the same so we don't trigger unique index errors
+            # when the document is saved within the make_current() call made
+            # within the rev.save() call below, or explicitly further below.
+            rev.slug = self.slug
+            rev.title = self.title
```
davidflanagan: I don't know enough (or really anything) about our DB schemas. Based on this code, I'm assuming that the document slug and title are the source of truth, and those fields on the revisions aren't definitive. And that there must be or must have been a bug where when a writer changes a slug or title we updated the document but didn't update (or create a new) current revision.  If that is basically correct, then this makes sense to me.

kuma/wiki/models.py
```diff
@@ -1528,8 +1535,10 @@ def clean_current_revision(self, user):
         # Populate the model instance with fresh data from database.
         rev.refresh_from_db()
 
-        # Make this new revision the current one for the document.
-        rev.make_current()
+        # Make this new revision the current one for the document if it hasn't
+        # been done already when saving the revision above.
+        if not rev.is_approved:
```
davidflanagan: Is it possible to write `if self.current_revision is not rev:` here instead?  (I'm not sure what the right way to compare is... maybe comparing the pk fields of the revisions?)

I've got no idea what the approval status of the revision has to do with the behavior of rev.save(). So if it is possible to test for this condition without requiring internal knowledge of how save() works, I'd prefer that. But if this is the best way to do it, then I'm fine with this as an optimization.

Also, I'm assuming that if the atomic transaction above fails, that an error is raised and we never get here. Right?

escattone: You're absolutely right, calling `rev.make_current()` based on `rev.is_approved` requires internal knowledge of `save()`, and something like `if self.current_revision.pk is not rev.pk` is better, thanks!

escattone: Oh, and in terms of what happens when an error occurs within the atomic transaction, yes, the error will be raised and the rest of the code will never be executed.

kuma/wiki/models.py
escattone: Yes, that's my understanding as well.

PR #5314, opened 2019-03-26
Bug 1538723, add banners django model and register with django admin

schalkneethling:
Here is the first portion of Grizzly ;p

Here is how existing banners will be listed:

Screenshot 2019-03-26 at 11 16 38

From here you can also bulk activate or deactivate banners:

Screenshot 2019-03-26 at 11 16 52

And here is the UI for adding/editing a banner:

Screenshot 2019-03-26 at 11 21 48

@jwhitlock r?


kuma/banners/migrations/0001_initial.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
schalkneethling: I will merge all of the migrations into this one and update the PR

schalkneethling: Done.

kuma/banners/admin.py
```diff
@@ -0,0 +1,33 @@
...
+class BannerAdmin(admin.ModelAdmin):
+    actions = ['activate_all', 'deactivate_all']
+    list_display = ('banner_name', 'banner_active', 'banner_priority')
+    fields = ('banner_name', 'banner_title', 'banner_copy',
+              'banner_button_copy', 'banner_theme', 'banner_active',
+              'banner_priority')
+    search_fields = ('banner_name', 'banner_active')
+    ordering = ('banner_priority',)
+
+    def activate_all(self, request, queryset):
```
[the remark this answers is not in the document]
schalkneethling: I think this might be useful to the users of this. It was simple enough to implement as well.

kuma/banners/tests/test_models.py
```diff
@@ -0,0 +1,63 @@
+from django.core.exceptions import ValidationError
+
+from . import BannerTestCase
+from ..models import Banner
+
+
+class TestBanner(BannerTestCase):
```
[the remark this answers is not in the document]
schalkneethling: Will rework these, thanks for pointing this out.

schalkneethling:
@jwhitlock Updated, thanks for the review!

kuma/banners/migrations/0001_initial.py
```diff
@@ -0,0 +1,29 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='Banner',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(max_length=50, verbose_name='Banner Name')),
+                ('title', models.CharField(max_length=100, verbose_name='Banner Title')),
+                ('main_copy', models.TextField(max_length=200, verbose_name='Main Copy')),
+                ('button_copy', models.CharField(max_length=50, verbose_name='Button Copy')),
+                ('theme', models.CharField(choices=[(b'default', b'Default'), (b'gradient', b'Gradient'), (b'dinohead', b'Dinohead')], default=b'default', max_length=20, verbose_name='Theme')),
```
jwhitlock: I'm surprised the ``choices`` are binary strings, since they are unicode strings in ``model.py``. I tried regenerating the initial migration and got the same result. This must be some constraint on values for choices. Django can still contain mysteries.

schalkneethling:
> we brainstormed on locale-based targeting and an optional waffle flag.

I reckon this is fine as a first pass. But I definitely have those in the back of my to add, as it will be a requirement.

PR #5315, opened 2019-03-26
 Bug 1525719: show the list of contributors at the bottom of each page 

davidflanagan:
This PR modifies the document API to include the list of contributors
and the last modified date for each document. On the frontend, it
modifies the React Page component to display that information near
the bottom of each page.

The patch also modifies the DocumentProvider component to add a
localeFromURL property to the document data. This new property is
intented as a single source of truth for the locale that is part
of the currently-displayed URL. Various React components that had
previously been manually extracting the locale from window.location
are modified to use this new localeFromURL property. This relatively
minor change required changes to test files and snapshots, making this
patch appear larger than it actually is.

PR #5325, opened 2019-03-30
bug 1525719: allow separate beta and wiki domains

escattone:
### Intro
The basic thrust of this PR is to add the `beta` and `wiki` sub-domains, where the `beta` sub-domain provides read-only, React-based endpoints for the homepage and the document pages, as well as the JSON API endpoints (`/api/v1/doc` and `/api/v1/whoami`), and the `wiki` sub-domain is for writers and is exactly like what we have today. This PR touched more aspects of Kuma and required more changes than I originally anticipated. If it helps to split this up, I will do that, but I wanted to submit it as a whole so it's not delayed any longer and others can take a look.

### Summary of changes

* update the `RestrictedEndpointsMiddleware` to recognize the new `beta` domain and restrict the accessible endpoints for that domain
* add tests for changes to `RestrictedEndpointsMiddleware`
* create a new `WaffleWithCookieDomainMiddleware` middleware that allows the domain of waffle cookies to be configured for use across the `beta` and `wiki` sub-domains
* add tests for `WaffleWithCookieDomainMiddleware`
* add `is_beta`, `is_untrusted`, and `url_for_wiki_site` Jinja helper functions
* add the `beta` and `wiki` site URL's to the JSON provided by the `/_kuma_status.json` endpoint to aid functional testing
* add some functional headless tests
* update some of the React code to use the `wiki` site URL
* add new `react_homepage.html` template
* add `urls_beta.py` files that provide a subset of URL's for the `beta` domain
* update the `Document.get_edit_url` and `Revision.get_absolute_url` methods to use the `kuma.urls` urlconf explicitly so they can be used within the restricted `beta` domain
* modify the `absolutify` Jinja helper to take a second argument that allows its caller to use the `WIKI_SITE_URL` rather than the `SITE_URL`
* refactor the homepage view to provide both a regular and React-based version
* add `LANGUAGE_COOKIE_DOMAIN`, `SESSION_COOKIE_DOMAIN`, `CSRF_COOKIE_DOMAIN`, and `WAFFLE_COOKIE_DOMAIN` settings in order to configure all cookies to work across the `beta` and `wiki` sub-domains
* remove the `/api/v1` endpoints and the React-based document endpoints from all domains except for the `beta` domain
* add `WAFFLE_SECURE` setting to allow for testing waffle cookies locally when not using `https`
* reorganize the user and document API code into the top-level `api` folder
* add top-level `reset_urlconf` pytest fixture to prevent cross-test contamination of the default urlconf for all tests (it's an `autouse` fixture so it's automatically included for all tests)
* move some pytest fixtures (e.g., `trans_doc`) from the wiki-level to the top-level for use by the API tests
* patch the `403.html`, `404.html`, `base.html`, `config.html`, and `react_base.html` templates to handle the `beta` domain until a better approach can be defined
* update `docker-compose.yml` to allow for easier testing via settings in the `.env` file
* add the `beta` site URL to the `connect-src` and `img-src` CSP policies
* update `contribute.json` (a "drive-by" change not necessary for this PR)

### Testing set-up
For testing, I'd recommend doing the following:
* update your `/etc/hosts` file to look something like this:
```
##
# Host Database
#
##
255.255.255.255	broadcasthost
::1             localhost
```
* and update your `.env` file to look like this:
```
DOMAIN=mdn.localhost
ATTACHMENT_HOST=demos:8000
DEBUG=True
CSP_ENABLE_MIDDLEWARE=True
```


escattone:
> Does the ALLOW_ROBOTS_WEB_DOMAINS setting keep google from indexing anything on the beta subdomain, even if we deploy to production?

@davidflanagan Yes, since we won't add the `beta` domain to the `ALLOW_ROBOTS_WEB_DOMAINS` list yet, the `robots.txt` endpoint for the beta` domain will disallow the crawling of all endpoints.

escattone:
Now that #5315 has merged, I'll re-base this on master and make sure the React-based tests pass.

tests/headless/test_endpoints.py
```diff
@@ -119,3 +119,44 @@ def test_hreflang_basic(base_url):
...
+@pytest.mark.nondestructive
+@pytest.mark.parametrize(
+    'uri,expected_keys',
+    [('/api/v1/whoami', ('username', 'is_staff', 'is_authenticated', 'timezone',
+                         'is_beta_tester', 'gravatar_url', 'is_superuser')),
+     ('/api/v1/doc/en-US/Web/CSS', ('locale', 'title', 'slug', 'tocHTML',
+                                    'bodyHTML', 'id', 'quickLinksHTML',
+                                    'parents', 'translations', 'editURL',
+                                    'summary', 'language', 'absoluteURL',
+                                    'redirectURL'))],
```
davidflanagan: I added contributors, lastModified and lastModifiedBy in my most recent update.

tests/headless/test_endpoints.py
```diff
@@ -119,3 +119,44 @@ def test_hreflang_basic(base_url):
...
+    ids=('whomai', 'doc')
+)
+def test_beta_api_basic(beta_site_url, uri, expected_keys):
+    """Basci test of beta site's api endpoints."""
```
davidflanagan: Typo on this line.

jinja2/base.html
```diff
@@ -103,6 +104,9 @@
     {% include "includes/a11y-nav.html" %}
 
     <!-- Header -->
+    {%- if beta %}
+    <header id="react-header"></header>
```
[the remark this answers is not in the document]
escattone: Yes, this was for the homepage.

tests/headless/test_endpoints.py
escattone: I just picked-up your changes from your recent PR and this is resolved now.

kuma/javascript/src/header/login.jsx
```diff
@@ -85,21 +87,27 @@ export default function Login(): React.Node {
             <>
                 <Dropdown label={label} right={true}>
                     <li>
-                        <a href={`/${LOCALE}/profiles/${userData.username}`}>
+                        <a href={`${WIKI_SITE_URL}/${LOCALE}/profiles/${
+                               userData.username
+                           }`}
```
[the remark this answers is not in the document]
escattone: Fixed by running `npm run format`. Thanks for the reminder, as I did do it by hand and I did forget to run `npm run format`!

kuma/landing/jinja2/landing/react_homepage.html
```diff
@@ -0,0 +1,160 @@
+{% extends "react_base.html" %}
```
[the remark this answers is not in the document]
escattone: Good catch, the one in wiki is unused. Fixed, thanks!

kuma/settings/common.py
```diff
@@ -1211,6 +1217,12 @@ def pipeline_one_scss(slug, **kwargs):
 _PROD_ATTACHMENT_ORIGIN = 'demos-origin.mdn.mozit.cloud'
 ATTACHMENT_ORIGIN = config('ATTACHMENT_ORIGIN', default=_PROD_ATTACHMENT_ORIGIN)
 
+BETA_HOST = config('BETA_HOST', default='beta.' + DOMAIN)
+BETA_ORIGIN = config('BETA_ORIGIN', default='beta.mdn.mozit.cloud')
+BETA_SITE_URL = PROTOCOL + BETA_HOST
+WIKI_HOST = config('WIKI_HOST', default='wiki.' + DOMAIN)
```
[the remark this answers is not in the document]
escattone: I think it's a subtle back-end thing (so I don't expect it to be obvious), but the `BETA_ORIGIN` is the domain used by the CloudFront CDN to point to the backing service (the ELB fronting the Django-based `web` pods) but using a different domain name so it can still be distinguished from standard requests even if we bypass the CDN (e.g., for basic testing). I don't think we're going to use a CDN for the `wiki` domain, so I don't see us needing a `WIKI_ORIGIN` setting that's different than the `WIKI_HOST` setting.

kuma/core/tests/test_middleware.py
```diff
@@ -71,14 +73,22 @@ def test_force_anonymous_session_middleware(rf, settings):
...
 
-def test_restricted_endpoints_middleware(rf, settings):
-    settings.ATTACHMENT_HOST = 'demos'
+@pytest.mark.parametrize(
+    'host,key,expected',
+    (('beta', 'BETA_HOST', 'kuma.urls_beta'),
+     ('beta-origin', 'BETA_ORIGIN', 'kuma.urls_beta'),
+     ('demos', 'ATTACHMENT_HOST', 'kuma.urls_untrusted'),
+     ('demos-origin', 'ATTACHMENT_ORIGIN', 'kuma.urls_untrusted')),
+    ids=('beta', 'beta-origin', 'attachment', 'attachment-origin')
```
jwhitlock: nice parametrize!

kuma/core/middleware.py
```diff
@@ -360,3 +363,24 @@ def __call__(self, request):
...
+    """
+    def process_response(self, request, response):
+        keys_before = frozenset(response.cookies.keys())
+        try:
+            response = super(WaffleWithCookieDomainMiddleware,
+                             self).process_response(request, response)
+        finally:
+            keys_after = frozenset(response.cookies.keys())
+            for key in (keys_after - keys_before):
+                response.cookies[key]['domain'] = settings.WAFFLE_COOKIE_DOMAIN
```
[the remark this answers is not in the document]
jwhitlock: I peeked at the code. I believe a "modified" waffle will still appear as a new cookie in this middleware.

https://github.com/django-waffle/django-waffle/blob/014385974d3c37ff9611db16e7325c8ef7ac1437/waffle/middleware.py#L10-L31

It is something to look for in testing.

kuma/landing/jinja2/landing/react_homepage.html
```diff
@@ -0,0 +1,160 @@
...
+{% if waffle.flag('developer_needs') %}
+    {%- block banners %}
+        {% include "banners/cta-banners.html" %}
+    {% endblock %}
+{% endif %}
+
+{% block contrib_popup %}
+    {%- if contribution_popup %}
+        {%- with form=contribution_form, is_popover=True %}
+            {%- include "payments/includes/payments-banner.html" %}
```
jwhitlock: I get a failure in the payments banner on ``reverse('payments')``

escattone: Good catch, I'll look into that.

escattone: This should be fixed.

docker-compose.yml
```diff
@@ -23,21 +23,22 @@ services:
...
       - SESSION_COOKIE_SECURE=False
-      - SITE_URL=http://localhost:8000
+      - SITE_URL=${SITE_URL:-http://localhost:8000}
       - STATIC_URL=${STATIC_URL:-http://localhost:8000/static/}
       # Other environment overrides
       - PYTHONDONTWRITEBYTECODE=1
       - PYTHONUNBUFFERED=True
       - MAINTENANCE_MODE=${MAINTENANCE_MODE:-False}
       - REVISION_HASH=${KUMA_REVISION_HASH:-undefined}
+      - WAFFLE_COOKIE_SECURE=False
```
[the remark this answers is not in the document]
escattone: Oops, sorry, didn't notice that! Fixed.

kuma/core/utils.py
```diff
@@ -41,6 +41,13 @@ def to_html(pq):
     return pq.html(method='html')
 
 
+def is_beta(request):
```
jwhitlock: Nit: Someday "beta" will be the primary domain, and "wiki" will be the secondary. Start thinking about what these tests look like then. Do all the ``{% if beta %}`` stay and the ``{% else %}`` branches go away? Do they become ``{% if main %}`` or ``{% if published %}`` or something else? I think we should optimize for making that transition in as few lines of code as possible, and I'm worried about a lot of changes around ``beta``, or worse, a 2021 document saying "``beta`` refers to the version of the site after 2019..."

PR #5327, opened 2019-04-01
Bug 1523662, Google results container wbr in displayed URL

schalkneethling:
Having read over the report in https://bugzilla.mozilla.org/show_bug.cgi?id=1523662 it seems the problem is that `JavaScript` is split into two words by the RegExp i.e. `Java<wbr />Script`

When Google then for some pages uses the text of the breadcrumbs as the text for the linking URL the raw HTML is displayed, including the `<wbr />` tag. As per @atopal's suggestion, and tweaking the RegExp, this no longer happens.

@davidflanagan r?

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -68,7 +68,6 @@
     */
     $('.document .document-head h1, .quick-links a code, .crumbs span[property=name]').each(function() {
         var $wrapper = $(this);
-        var text = $wrapper.text();
         // don't do anything if there are any child elements, they would be removed
         if ($wrapper.children().length > 0) {
             return;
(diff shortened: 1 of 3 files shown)


kuma/static/js/wiki.js
```diff
@@ -80,7 +80,7 @@
          * IF followed by [ . : ( OR capital letter
          * IF followed by 3 letters or [ @ . : (
          */
-        var split = text.split(/([a-z]{3})(?=[[.:(A-Z][@.:(A-Z]{0,}[a-zA-Z]{3})/g);
+        var split = text.split(/([a-z]{3})(?=[[.:(][@.:(A-Z]{0,}[a-z]{3})/g);
```
davidflanagan: If I'm understanding this change correctly, it:

1) means that the "OR capital letter" comment on line 80 is no longer correct.

2) means that CanvasRenderingContext2D will no longer get word breaks after "Canvas" and "Rendering" and for a genuinely long class name like that, this might actually cause layout problems on narrow screen sizes.

schalkneethling: 1. Correct, I will need to update the comment.
2. I agree and was never *really* confident that this simple change was going to work without side effects.

I need to find some additional example pages to test against and consider some additional options.

> Have we tried using U+200B ZERO-WIDTH SPACE instead of ?

Do you mean instead of the `<wbr />`?

davidflanagan: Yes, I meant instead of `<wbr/>`. The markdown editor ate my HTML tag!

schalkneethling:
@davidflanagan This has been updated. Your suggestion, along with some additional CSS changes/additions, seems to work great! Thanks!

kuma/static/styles/components/wiki/crumbs.scss
```diff
@@ -29,7 +42,18 @@ wiki article bread crumbs
...
+            border-bottom: $crumb-vertical-spacing-tablet solid transparent;
+        }
+
+        @media #{$mq-small-desktop-and-up} {
+            border-bottom: $crumb-vertical-spacing-desktop solid transparent;
+            line-height: 1;
+        }
 
         &:after {
             margin: 10px 5px 3px;
```
davidflanagan: I suspect this is the line causing the misalignment. A bottom margin of 5px looks good to me in tablet mode, though that messes things up at desktop size.

(This would be easier if we could just use text instead of an image here, like https://www.fileformat.info/info/unicode/char/203a/index.htm used with css content. Thought I suppose that having a directional character makes bidi handling harder...)

schalkneethling:
Thanks @davidflanagan - updated!

PR #5330, opened 2019-04-03
bug 1525719: fix provider_login_url for beta

escattone:
This is a potential fix for the ISE (e.g., see the error in Sentry for stage: `Reverse for 'github_login' not found. 'github_login' is not a valid view function or pattern name.`) for both the `beta` homepage as well as any `beta` doc page on stage (e.g., `https://beta.developer.allizom.org/en-US` and `https://beta.developer.allizom.org/en-US/docs/Web/CSS`).

I was unable to get the contribution banner to display for `beta.mdn.localhost:8000` for local testing (it did display for `mdn.localhost:8000`), so I'm unable to confirm if this will fix the issue, but I wanted to submit the PR for review as early as possible.

### Testing set-up
For testing, I'd recommend doing the following:
* update your `/etc/hosts` file to look something like this:
```
##
# Host Database
#
##
255.255.255.255	broadcasthost
::1             localhost
```
* and update your `.env` file to look like this:
```
DOMAIN=mdn.localhost
ATTACHMENT_HOST=demos:8000
DEBUG=True
STRIPE_PUBLIC_KEY=<test-public-key>
STRIPE_SECRET_KEY=<test-secret-key>
STRIPE_PRODUCT_ID=prod_test
MDN_CONTRIBUTION=True
CSP_ENABLE_MIDDLEWARE=True
```

### DIFF
--- a/kuma/payments/jinja2/payments/includes/payments-banner.html
+++ b/kuma/payments/jinja2/payments/includes/payments-banner.html
@@ -85,7 +85,11 @@
                         {% endwith %}
 
                         {% if not user.is_authenticated %}
+                            {% if beta %}
+                            {% set github_url = provider_login_url('github', process='login') | absolutify(for_wiki_site=True) %}
+                            {% else %}
                             {% set github_url = provider_login_url('github', process='login') %}
+                            {% endif %}
                             <div id="login-popover" class="hidden align-center login-prompt">
                                 <p>{{_('We’ve noticed you’re not logged in to MDN. To complete your monthly payment, we’ll need you to log in first.')}}</p>
                                 <a href="{{ github_url }}" id="github_redirect_payment" class="payment-form-button js-login-link" data-service="GitHub" rel="nofollow" data-next="{{ github_next_url }}">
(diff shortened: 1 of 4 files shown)


kuma/payments/jinja2/payments/includes/payments-banner.html
```diff
@@ -85,7 +85,11 @@ <h4 tabindex="-1">{{_('Support MDN, and it comes back to you.')}}</h4>
                         {% endwith %}
 
                         {% if not user.is_authenticated %}
+                            {% if beta %}
+                            {% set github_url = provider_login_url('github', process='login') | absolutify(for_wiki_site=True) %}
```
escattone: For this URL to work, I suspect we'll have to create a new OAuth App for the `wiki` domain?

jwhitlock: Yes, there are probably limitations in django-allauth and GitHub's oauth that would require a new app for the wiki domain. Maybe we can send logins through developer.allizom.org instead, using the next parameter to send to the proper domain for return.

PR #5331, opened 2019-04-04
Bug 1525719: Make the React header responsive.

davidflanagan:
This patch modifies the React header component (and it subcomponents)
so that the layout works at phone, tablet and desktop size. It converts
the Header component from flexbox layout to grid layout and uses media
queries to select different grid templates at different sizes.

Additionally, media queries are used to set a base font size, and other
elements of the header (such as the MDN logo) are sized in 'em' units.


davidflanagan:
There are a lot of changes to the test snapshots and a couple of changes to icons that make this look like a bigger patch than it actually is. Most of the interesting changes are in header.jsx.

I haven't done much responsive design before, and this is my first time using CSS Grid, but this seemed to come out okay. I'm sure a designer would have lots of nits to fix, but maybe this is good enough until we have a designer?  For the phone layout I'm thinking about a follow on patch to add a hamburger menu that hides and shows the header, so that it does not appear by default.

Iphone, logged in:
Screen Shot 2019-04-03 at 5 08 35 PM

And logged out: 
Screen Shot 2019-04-03 at 5 06 38 PM

ipad, logged in:
Screen Shot 2019-04-03 at 5 08 23 PM

logged out: 
Screen Shot 2019-04-03 at 5 07 01 PM

desktop: logged in:
Screen Shot 2019-04-03 at 5 08 12 PM

logged out:
Screen Shot 2019-04-03 at 5 07 39 PM

schalkneethling:
Thanks so much for this @davidflanagan as well as the screenshots. I really like the direction this is moving in. I have some thoughts on our overall design, as well as our approach to CSS, HTML, accessibility and our UI build(no related to tooling ;)) process in general.

This is on me though, and involves a complete, fleshed out style guide for the MDN brand, and more specifically, for building our web sites. I have had a couple of stabs at it, but it was never prioritised and so, once high priority work came around, work was stopped, and it has simply gotten stale, and very out of date now.

With this rewrite and the future work that surrounds MDN, it makes more sense now than ever to just do it, and build that style guide.

All of that said, I definitely *do not* want to hold up the work yourself and @escattone are doing. My plan is to instead only point out things that are clearly a large deviation, or something that could be problematic to change in future.

As the styleguide comes to live, I will share the work, and also start to submit "patches" as appropriate to ensure that we take full advantage of this opportunity to have an incredible code base. From both a technical, aesthetic, as well as maintainability perspective.

I hope this all makes sense, and again thanks so much for all of the work that is going into this.

schalkneethling:
@davidflanagan Just realized that I have no idea how to access the React powered pages now that the ducks have left.

jwhitlock:
@schalkneethling see the "Testing Setup" at https://github.com/mozilla/kuma/pull/5325. We discussed yesterday some option for making this a more permanent part of the development environment, but it will probably be a few more weeks while we work out the details of the new arrangement.

davidflanagan:
Thanks, Schalk. I just discoverd via a comment you made in a different PR that there is the old mdn-fiori repo... I didn't know about that.  My approach has been to just keep moving forward with this, creating something that looks and functions okay for a beta site with the expectation that eventually someone with better design skills will come along and tell me to make changes. 

It is my understanding that there is a job search for a UX person. Perhaps it would make sense to wait on the style guide work until that person is on board? Though on the other hand, that person may end up having all their time taken up with prototyping of future features.

PR #5332, opened 2019-04-04
Bug 1525719: call history.pushState() on initial page

davidflanagan:
This patch fixes a bug in client-side navigation on the beta site.
We weren't calling history.pushState() on the first page load, so
after navigating away, the back button would not correctly return
to that initial page.

As part of the fix, the localeAndSlug property of the state is
modified to include an initial slash character.


davidflanagan:
Ryan: this is a fix for the bug you found yesterday where after navigating away from the initial page loaded on beta.mdn.localhost, the back button would not return that page. Thanks for pointing that out. This should be an easy review.

PR #5335, opened 2019-04-05
bug 1475871: Fix the "Other" email field in signup, add tests

jwhitlock:
The signup page is displayed after a new user logs in with GitHub for the first time. It includes an email selector that is highly custom, like it came from 2012:

* The email includes **Verified** or Unverified, to show GitHub's verification status. We trust emails that GitHub say are verified, and send a verification email otherwise
* The last selection is "Other", and includes an ``email`` ``<input>`` that is displayed on the same line
* If an existing user has a matching email, then we display with a strikeout, to indicate that while GitHub returned it, picking that email will cause the form to fail

These special features have been broken for a while. My suspicion is that it broke during the Django 1.11 update, since we're using undocumented APIs to build the custom form. I'm not sure there is a "right" way to build this form in Django, so I focused on making it work roughly the same way, and adding tests to verify that the generated HTML works as expected.

For local testing, I already have local GitHub auth working. I added a legacy Persona account and deleted my GitHub link:

```
me = User.objects.get(username='jwhitlock')
me.socialaccount_set.all().delete()
me.socialaccount_set.create(provider='persona', uid=me.email)
```

I then get a sign-up page that looks a bit like this (emails changed to protect the guilty, including myself):

mdn signup bug 1475871 before fix 2

**After the fix, it looks like this:**

mdn signup bug 1475871

### DIFF
--- a/kuma/users/tests/test_templates.py
+++ b/kuma/users/tests/test_templates.py
@@ -192,33 +192,36 @@ class AllauthGitHubTestCase(UserTestCase, SocialTestMixin):
         parsed = pq(response.content)
         expected_emails = [
             {
-                # BUG: The class should be omitted if empty
-                'li_attrib': {'class': ''},
-                # BUG: The label 'for' should match the radio 'id'
-                'label_attrib': {'for': '_0'},
+                'li_attrib': {},
+                'label_attrib': {'for': 'email_0'},
                 'radio_attrib': {'required': '',
                                  'type': 'radio',
                                  'name': 'email',
                                  'value': 'octocat-private@example.com',
                                  'id': 'email_0'},
+                'verified': True,
             }, {
-                'li_attrib': {'class': ''},
-                'label_attrib': {'for': '_1'},
+                'li_attrib': {},
+                'label_attrib': {'for': 'email_1'},
                 'radio_attrib': {'checked': 'checked',
                                  'required': '',
                                  'type': 'radio',
                                  'name': 'email',
                                  'value': 'octocat@example.com',
                                  'id': 'email_1'},
+                'verified': False,
             }, {
-                # BUG: This should include an email input for "other"
-                'li_attrib': {'class': ''},
-                'label_attrib': {'for': '_2'},
+                'li_attrib': {},
+                'label_attrib': {'class': 'inner other-label',
+                                 'for': 'email_2'},
                 'radio_attrib': {'required': '',
                                  'type': 'radio',
                                  'name': 'email',
                                  'value': '_other',
                                  'id': 'email_2'},
+                'other_attrib': {'type': 'email',
+                                 'name': 'other_email',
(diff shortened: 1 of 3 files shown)


kuma/users/tests/test_templates.py
```diff
@@ -178,21 +178,82 @@ def test_signin_form_present(self):
 
     def test_signup(self):
         """
-        After a new user signs up with Persona, their username, an
-        indication that Persona was used to log in, and a logout link
-        appear in the auth tools section of the page.
+        After a new user logs in with GitHub, they signup to pick a username
+        and email to use on MDN. One signup is complete, the sign-in
```
peterbe:
```suggestion
        and email to use on MDN. Once signup is complete, the sign-in
```

kuma/users/tests/test_views.py
```diff
@@ -1262,6 +1262,11 @@ def test_matching_accounts(self):
         assert response.status_code == 200
         assert_no_cache_header(response)
         assert not response.context['matching_accounts']
+        # The template is tested here instead of test_templates.py because
+        # test setup is so painful for login tests.
+        parsed = pq(response.content)
+        li_exists = parsed.find('ul.choices li.exists')
+        assert len(li_exists) == 0
```
peterbe:
```suggestion
        assert not li_exists, li_exists
```

The `li_exists' (from the `.find()`) becomes an empty list if nothing was found.

kuma/users/tests/test_views.py
```diff
@@ -1271,6 +1276,14 @@ def test_matching_accounts(self):
                                                       user=octocat3)
         response = self.client.get(self.signup_url)
         assert list(response.context['matching_accounts']) == [social_account]
+        parsed = pq(response.content)
+        # li with class=exists is rendered with a strikeout, to suggest to the
+        # user that signup may fail and they should use account recovery.
+        li_exists = parsed.find('ul.choices li.exists')
```
peterbe: nit: If you use `li_exists = parsed.find('ul.choices li.exists').items()` you get a generator which you can use like jQuery. E.g.

```python
li_exists = parsed.find('ul.choices li.exists').items()
assert len(li_exists) == 1
email_input = li_exists('input[type=radio]')
...
```

kuma/users/tests/test_views.py
```diff
@@ -1262,6 +1262,11 @@ def test_matching_accounts(self):
         assert response.status_code == 200
         assert_no_cache_header(response)
         assert not response.context['matching_accounts']
+        # The template is tested here instead of test_templates.py because
+        # test setup is so painful for login tests.
```
peterbe: I don't understand this comment. 
Do you mean "The generated HTML is tested ..."?

Also, for the record, I think it's great idea to do assertions on the `response.content` instead of doing anything with `response.context` because that would be a test of implementation details.

PR #5337, opened 2019-04-05
Bug 1525719: make the React Page component responsive.

davidflanagan:
This patch modifies the page.jsx component to use CSS Grid layout
and display itself differently on narrow screens than it does on
larger screens. The main difference is that on narrow screens the
sidebar is moved to the bottom of the page, and the TOC bar is
hidden. There are also some font and margin size changes for
narrow screens as well.

The patch also includes a change to the react_document.html template
to make the MDN newsletter banner display with the correct margins.

### DIFF
--- a/kuma/javascript/src/__snapshots__/page.test.js.snap
+++ b/kuma/javascript/src/__snapshots__/page.test.js.snap
@@ -1,7 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Page snapshot 1`] = `
-.emotion-46 {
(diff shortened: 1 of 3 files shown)


davidflanagan:
Before:

Screen Shot 2019-04-05 at 1 37 21 PM

After: 
Screen Shot 2019-04-05 at 1 37 52 PM

davidflanagan:
@schalkneethling: I'll be OOO on Monday, so if this patch looks good to you, please feel free to merge it to increase the chance of it being deployed by the end of the sprint.

PR #5338, opened 2019-04-05
correct Whatsdeployed URL

peterbe:


peterbe:
Arguably, this isn't "[new] contributor setup" related. It's an op thing. 

It'll take me some getting used to re-use existing bugs. I guess this Bugzilla product has too many open bugs to afford many little small bugs for things like chores :)

PR #5339, opened 2019-04-08
Bug 1539488, add grizzly jinja custom tag

schalkneethling:
This adds one additional field to the model. Other than that, this gets us to where we are now but, the developer needs banner' content is now loaded from the DB instead of being hardcoded. Thus it can also be managed from the admin.

@jwhitlock r?


kuma/banners/jinja2/banners/banners.html
```diff
@@ -0,0 +1,52 @@
+{% set banners = get_banners() %}
+<div
+    id="developer_needs"
```
schalkneethling: I'm working on the next iteration of this which will support multiple banners. This `id` will be removed as part of that.

kuma/banners/jinja2/banners/banners.html
```diff
@@ -0,0 +1,52 @@
...
+    class="mdn-cta-container hidden"
+    aria-expanded="false"
+    aria-hidden="true"
+    data-banner="developer_needs"
+>
+    {% if banners|count > 0 %}
```
peterbe: `banners` is a QuerySet so this line will cause a `SELECT COUNT(*) FORM banners_banner`. Even though we don't need the count. 

Since this template is jinja I'd turn the QuerySet into a list with
```jinja
{% set banners = list(get_banners()) %}
```

Then you can do:

```jinja
{% if banners %}
  {% for banner in banners %}
    ...
  {% endfor %}
  <buttons>
{% endif %}
```

kuma/banners/jinja2/banners/banners.html
```diff
@@ -0,0 +1,52 @@
+{% set banners = get_banners() %}
+<div
+    id="developer_needs"
+    class="mdn-cta-container hidden"
+    aria-expanded="false"
+    aria-hidden="true"
+    data-banner="developer_needs"
+>
```
peterbe: Why display this container div at all if there are no banners?
I mean, why not do this:

```jinja
{% if banners %}
  <div id="developer_needs">
    {% for banner in banners %}
    {% endfor %}
    <button/>
  </div>
{% endif %}
```

kuma/banners/models.py
```diff
@@ -20,7 +20,8 @@ class Banner(models.Model):
     title = models.CharField('Banner Title', max_length=100)
     main_copy = models.TextField('Main Copy', max_length=200)
     button_copy = models.CharField('Button Copy', max_length=50)
-    theme = models.CharField('Theme', max_length=20,
+    button_url = models.URLField('URL')
```
peterbe: This line doesn't match the migration. Where is the `verbose_name` and the `default`?

kuma/banners/models.py
```diff
@@ -20,7 +20,8 @@ class Banner(models.Model):
...
+    theme = models.CharField('Theme', max_length=40,
                              choices=THEMES, default=THEME_DEFAULT)
```
peterbe: Be very mindful of using `choices` in model fields. Any edits of any of these choices will require a new DB migration. It's not expensive in terms of processing to execute these migrations but it will cause more noise and more files and potentially confusing merge conflicts if there are more migration files.

kuma/banners/jinja2/banners/banners.html
schalkneethling: @peterbe That container is going away in the next PR. I need to rewrite some other bits as well and did not want to add additional noise to PR.

kuma/banners/models.py
schalkneethling: Whoops, my bad for manually merging and editing the migration directly ;) Fixing.

schalkneethling:
Thanks for the review @peterbe, completely forgot I can ping you for it now ;)

kuma/banners/templatetags/jinja_helpers.py
```diff
@@ -0,0 +1,11 @@
...
+from __future__ import unicode_literals
+
+from django_jinja import library
+
+from ..models import Banner
+
+
+@library.global_function
+def get_banners():
+    return Banner.objects.all()
```
peterbe: I appreciate that I'm very very new to the kuma code base so take me with a handful of salt. But, this seems like a not-ideal pattern. Jinja helpers shouldn't be used for get more data. They should be to express/repackage stuff that can't easily be done in template language. 

This function is very powerless. It doesn't support filtering or ordering. 
In particular, the query should probably be: 
```python
Banner.objects.filter(active=True).order_by('priority')
```

A better place to gather the data before template rendering even starts is the view functions. But since, I think, this data isn't related to a specific view, but rather, to *every* view, it should live in its own context processor.

kuma/banners/models.py
peterbe: What I mean by that is; if the list is really very unlikely to change keep it in the model as class attributes. If it's possible that it will change once or thrice a year, consider moving it to some admin.py file or something.

kuma/banners/templatetags/jinja_helpers.py
schalkneethling: So, the template tag will call the function on the context_processor or, should the templates simply use the variable set into context directly?

I ask because I have seen both being used.

kuma/banners/models.py
schalkneethling: Is there another `Field` type on can use that will use `select` as its widget? From what I understand, adding `choices` to a `CharField` is generally how this is accomplished.

peterbe: It is indeed but there are other ways to set it too. For example in a form class. 
I don't think you need to change anything at this point. Just something to be mindful about if you can foresee that there *will* be a lot of flux on this `choices`.

kuma/banners/jinja2/banners/banners.html
```diff
@@ -0,0 +1,52 @@
...
+    {% if banners %}
+        {% for banner in banners %}
+            <div class="mdn-cta-content {{ banner.theme }}">
+                <div id="developer-needs" class="mdn-cta-content-container">
```
peterbe: You mentioned you were planning to edit all of this. And I'm not the reviewer. But I don't think the PR is too big or noisy to include your other changes for this. 
Right now it's dangerously wrong with the ID being repeated in a for-loop.

schalkneethling: Ah yes. I am aware of that. But as of right now, and this won't change soon, there will only be one banner.

With that said, in the follow on PR, I will add support for multiple banners as that is the long term use case for this.

kuma/banners/templatetags/jinja_helpers.py
peterbe: As far as I can remember, the way it works is something like this:

1. View Python function code does stuff and calls something like `return render(request, 'foo.html', mycontext)`
2. Context processors are looped over and for each one it gets a chance to update that `mycontext` with extra stuff. 
3. The Jinja template is rendered with the fat `mycontext`.

Generally, context processors are really useful for data stuff needed by the `base.html` template. Otherwise, if it depends on some context data, you'd have to explicitly ask every single view function to provide it.

schalkneethling: That sounds good to me. We could then definitely replace the jinja helper with a context processor, and have it stick the list of banners into the context. @jwhitlock Thoughts?

kuma/banners/jinja2/banners/banners.html
jwhitlock: I'm guessing this will be a database field or use a slugified version of the name. If you intend to add a new database field, I think we should do it with this PR, which already has a migration.

kuma/banners/templatetags/jinja_helpers.py
```diff
@@ -0,0 +1,14 @@
...
+
+from ..models import Banner
+
+
+@library.global_function
+def get_banners():
+    """
+    Get all active banners ordered by priority
+    """
+    return list(Banner.objects.filter(active=True).order_by('priority'))
```
peterbe: Actually, if `{% include "banners/banners.html" %}` is only ever called from `landing/homepage.html` then there *is* a great place to generate this context *before* rendering. 

And that's [`landing.view.home`](https://github.com/mozilla/kuma/blob/391774e15431971ea82aeac4e006f104a16f4d51/kuma/landing/views.py#L25)
Or, perhaps the `render_home` view function to benefit the React based one. 

Actually, why is the banner not displayed from `landing/react_homepage.html` also?

peterbe: Also, we'd need to conditional that on `waffle.flag('developer_needs')` in the view function and once that's in place there's no need to use `waffle.flag('developer_needs')` in the jinja template. You can just leverage the fact that `banners` is truthy (if waffled on and has a list of banners).

schalkneethling: > Actually, if `{% include "banners/banners.html" %}` is only ever called from `landing/homepage.html` then there _is_ a great place to generate this context _before_ rendering.

It is going to be added to other pages as well once the solution is more fully baked.

> And that's [`landing.view.home`](https://github.com/mozilla/kuma/blob/391774e15431971ea82aeac4e006f104a16f4d51/kuma/landing/views.py#L25)
> Or, perhaps the `render_home` view function to benefit the React based one.
> 
> Actually, why is the banner not displayed from `landing/react_homepage.html` also?

It is not yet essential to be added there, and there will be some additional JS rewriting required once it is moved to the React portion of the codebase. At the moment neither the React bits nor this banner is shown to users.

peterbe: So, it's going to be used on all pages? If so, it should move to a context processor. Happy to help if in doubt. 

If you just want to get this landed ASAP, then there should be a code comment pointing to the bug about rewriting it later.

peterbe:
Tease me for being kuma newbie but one thing that I don't see clearly is; should everyone get this? For example, is the survey available in multiple languages? If not, it might not make sense to display it for certain locales. 
What about `*$edit` pages? I could definitely see it being worth skipping if the user is logged in and in the mood to do some editing.

peterbe:
How does localizing the content work? I haven't worked on a localized Django app in over 10 years. [This PR renders the DB content through the `_()` function](https://github.com/mozilla/kuma/blob/391774e15431971ea82aeac4e006f104a16f4d51/kuma/banners/jinja2/banners/banners.html#L14). 
As far as I can see, there are no other instances of that being done in the templates at least. Since we prepare the .po files as a build step, as far as I can guess, we *don't* do this using the Production MySQL data, which is where the `banner` objects's `title`, `main_body`, and `button_copy` are stored.

schalkneethling:
> How does localizing the content work? I haven't worked on a localized Django app in over 10 years. [This PR renders the DB content through the `_()` function](https://github.com/mozilla/kuma/blob/391774e15431971ea82aeac4e006f104a16f4d51/kuma/banners/jinja2/banners/banners.html#L14).
> As far as I can see, there are no other instances of that being done in the templates at least. Since we prepare the .po files as a build step, as far as I can guess, we _don't_ do this using the Production MySQL data, which is where the `banner` objects's `title`, `main_body`, and `button_copy` are stored.

Good question. I do not know all the details here. @jwhitlock could you expand here, please?

kuma/banners/templatetags/jinja_helpers.py
jwhitlock: I agree, a context processor is probably the right tech for a site-wide banner.

jwhitlock:
The strings will need to be present in a source file as well. For example, the search topics are exported to a Python file using ``./manage.py generate_search_names``:

https://github.com/mozilla/kuma/blob/master/kuma/search/management/commands/generate_search_names.py

The output is written to a file that is checked into the repository:

https://github.com/mozilla/kuma/blob/master/kuma/search/names.py

This works decently for search topic names, because it is rare to change them. You'll have to set up some processes around generating source files for banner text.

peterbe:
@jwhitlock Is this something a developer runs, on his/her laptop, and then makes a PR. Fine by me. 

What do you think @schalkneethling ? What would you prefer? That approach, or adding a `locale` field on the `Banner` model and make it so you type in different banner titles and content, in the Django admin, **per locale**.

schalkneethling:
> @jwhitlock Is this something a developer runs, on his/her laptop, and then makes a PR. Fine by me.
> 
> What do you think @schalkneethling ? What would you prefer? That approach, or adding a `locale` field on the `Banner` model and make it so you type in different banner titles and content, in the Django admin, **per locale**.

I created the [following issue](https://github.com/mdn/sprints/issues/1294) for the work. I think sticking with a process that has already been used might be the path of least resistance. The localisers also do not have access to the Django admin, and so adding the localisations that way could be problematic.

schalkneethling:
@peterbe Updated to use context processor.

kuma/banners/jinja2/banners/banners.html
peterbe: I don't like that this div tag gets injected all and every time even though there are potentially, no banners. 
A big step forward would be to move the `{% if banners %}` statement on line 1.

kuma/banners/jinja2/banners/banners.html
peterbe: What is the follow-on PR going to be about? Is there a reason you can't do it as part of this PR. Is there benefits in landing this (with its little flaws) and having it in prod for a while? 

One concern I have is that the l10n'ness of the banner's content is far from worked out and unless there's an urgent rush I think we should nail that *and* this niggled about repeated IDs.

peterbe: By the way, why does it need an ID at all??

kuma/banners/jinja2/banners/banners.html
peterbe: Same here, why does it need an ID?

schalkneethling:
@peterbe As we are starting the next sprint on Monday, I agree that it makes sense to fold in the work for https://github.com/mdn/sprints/issues/1295 into this PR. I will start on that, and ping you here once it has been added. Thanks!

schalkneethling:
I am going to close this one for now. We can cycle back to this once we pick up the work again.

PR #5342, opened 2019-04-09
bug 1357416: Normalize slugs to NFC

jwhitlock:
This adds a NFC normalization step to the ``RevisionForm`` and ``DocumentForm``, used when editing, translating, and moving pages, so that slugs are automatically normalized before saving.

Some unicode text is displayed the same, but is composed of a different sequence of unicode codepoints. For example, the Greek small case epsilon with a tonos (έ) can be represented by a [single codepoint](https://codepoints.net/U+03AD), or by an [epsilon](https://codepoints.net/U+03b5) (ε) followed by a [combining acute accent](https://codepoints.net/U+0301) (``´``). This can mean that
two unicode strings appear the same but are different because of the code points.

Unicode defines [normalization forms](http://www.unicode.org/faq/normalization.html) that define the canonical representation of these strings. One form, Normalization Form D (NFD, or "decomposed"), uses the canonical decomposition, where έ is represented by two codepoints, and there is a defined order for multiple combining codepoints. Normalization Form C (NFC, or "composed") uses NFD followed by a canonical composition, where έ is represented by one codepoint. There are two others, NFKD and NFKC, but you don't need to know those, or [all the details of Annex #15](https://www.unicode.org/reports/tr15/#Norm_Forms), for this PR.

The W3C specifies [NFC for content](https://www.w3.org/International/questions/qa-html-css-normalization), and it appears browsers perform NFC on URLs. It is possible to enter a non-NFC slug in editing, translating, and moving pages, and because the browser performs NFC, these pages become unreachable. This is most likely to occur in the Bengali (bn-BD) domain.

I've [manually updated slugs with non-NFC forms](https://bugzilla.mozilla.org/show_bug.cgi?id=1357416#c3). There were 16 of them left. This prevents new ones from being created.

### DIFF
--- a/kuma/wiki/forms.py
+++ b/kuma/wiki/forms.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import unicodedata
 from difflib import ndiff
 
 import waffle
@@ -375,6 +376,11 @@ class DocumentForm(forms.ModelForm):
         elif self.parent_slug:
             # Prepend parent slug if given from view
             slug = self.parent_slug + '/' + slug
+
+        # Convert to NFC, required for URLs (bug 1357416)
+        # http://www.unicode.org/faq/normalization.html
+        slug = unicodedata.normalize('NFC', slug)
+
         # check both for disallowed characters and match for the allowed
         if (INVALID_DOC_SLUG_CHARS_RE.search(slug) or
                 not DOCUMENT_PATH_RE.search(slug)):
@@ -546,6 +552,10 @@ class RevisionForm(AkismetCheckFormMixin, forms.ModelForm):
 
         # Get the cleaned slug
         slug = self.cleaned_data['slug']
+
+        # Convert to NFC, required for URLs (bug 1357416)
+        # http://www.unicode.org/faq/normalization.html
+        slug = unicodedata.normalize('NFC', slug)
 
         # first check if the given slug doesn't contain slashes and other
         # characters not allowed in a revision slug component (without parent)
--- a/kuma/wiki/tests/test_forms.py
+++ b/kuma/wiki/tests/test_forms.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import json
+import unicodedata
 
 import pytest
 import requests_mock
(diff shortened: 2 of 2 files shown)


jwhitlock:
I had a brain itch, and went to see how Django does it. They use NFKC:

https://github.com/django/django/blob/ad9a28ee38e3352b16cc6c9ae7f55f90c64710cc/django/utils/text.py#L398-L411

We have a number of URLs that are different in NFKC and NFC.

I didn't find a definitive "Use NFC in URLs doc", just that old W3C document about slugs. Now I'm wondering if the W3C doc is out of date, and Django is right...

jwhitlock:
I found [RFC 3987, section 5.3.2.2](https://tools.ietf.org/html/rfc3987#section-5.3.2.2)

> Equivalence of IRIs MUST rely on the assumption that IRIs are
   appropriately pre-character-normalized rather than apply character
   normalization when comparing two IRIs.  The exceptions are conversion
   from a non-digital form, and conversion from a non-UCS-based
   character encoding to a UCS-based character encoding. In these cases,
   NFC or a normalizing transcoder using NFC MUST be used for
   interoperability.  To avoid false negatives and problems with
   transcoding, IRIs SHOULD be created by using NFC.  Using NFKC may
   avoid even more problems; for example, by choosing half-width Latin
   letters instead of full-width ones, and full-width instead of
   half-width Katakana.

So it looks like the standard is "must be normalized", not "normalized in NFC". I'm inclined to follow Django's lead and use NFKC.

One change that even I can understand is that this URL:

https://developer.mozilla.org/fr/docs/JavaScript/Reference/Instructions/for_each%E2%80%A6in

would stop using the [horizontal ellipsis](https://www.fileformat.info/info/unicode/char/2026/index.htm) ``…`` and change to this URL:

https://developer.mozilla.org/fr/docs/JavaScript/Reference/Instructions/for_each...in

In this case, this was already done, and the first URL is a Kuma redirect to the second. I can change the code, and I think I'd need a new PR for the form change going forward.

peterbe:
Should we close this one now?
https://github.com/mozilla/kuma/pull/5342

jwhitlock:
Yes, closing in favor of #5344

PR #5343, opened 2019-04-10
bug 1525719: add command for publishing to the doc API

escattone:
Note: This PR depends upon https://github.com/mdn/infra/pull/220, which has already been merged.

This PR is one step towards a complete publish command, which is one of our final goals. The complete publish command would, for each document, start from the raw HTML, render any KumaScript macros, clean the content, generate pre-rendered JSON files for the document API (used for fast, inter-page navigation after the initial page load), and also generate the deliverable HTML via React SSR (used for the initial page load). This PR is the piece that generates the pre-rendered JSON files for the document API, and depends on some AWS infrastructure that I've already created: two S3 buckets that serve the document JSON API, one each for `stage` and `prod`, and a change to the `stage` CloudFront CDN (I haven't updated the `prod` CDN yet) that changes the origin of all document API calls (`https://beta.developer.allizom.org/api/v1/doc/*`) from the ELB (in front of the Django service) to the `stage` S3 bucket. So, effectively, the `stage` S3 bucket serves the document API for `stage` via the `stage` CDN.

The main changes in this PR are:
- add a new Django management command (`publish`), along with its associated Celery task, that creates the JSON files on S3 for the document API (overwriting any existing)
- modifies `kuma.api.v1.views.document_api_data` to handle redirects, as well as bypass Celery and the cache when getting the contributors

The `publish` command relies on our current Celery set-up for distributed task processing, kicking off  tasks that can be run in parallel, each of which handles a chunks of documents, the size of which can be specified via the command line. For example, assuming we're running within one of the Kuma-based pods within the Kubernetes cluster, here's the command to publish all of the documents in tasks that each handle 1000 documents (there are roughly 130K documents):
```
./manage.py publish --all --chunk-size=1000
```

Currently, the approach to redirects is to simply handle them in `kuma.api.v1.views.document_api_data` and not via the `x-amz-website-redirect-location` property on S3 objects (or the `WebsiteRedirectLocation` argument to the `put_object` method on the `Bucket` resource when using `boto3`). This means that if `/api/v1/doc/en-US/MDN/Promote` is really a redirect to `/api/v1/doc/en-US/MDN/About/Promote`, then the JSON file for `/api/v1/doc/en-US/MDN/Promote` is a duplicate of the one for `/api/v1/doc/en-US/MDN/About/Promote` rather than a redirect. This is faster since it avoids redirects altogether (but of course at the cost of extra storage), and was simpler than trying to determine a way to do 302's on S3 (the `x-amz-website-redirect-location` property uses 301's). We can revisit this if desired.

Most content-based redirects are from one document to another document, but there are some that redirect from a document to a non-document endpoint on the site (redirects to external URL's are not allowed, although from my investigation yesterday there seems to be a good number of invalid URL's of the form `/docs/https://some.external.domain.com`). For those, I'll return a `redirectURL` that is either a relative link (so we'll stay within the `beta` domain) to the home page ('/', '/en-US', or '/en-US/') or a full URL for the `wiki` site (`https://wiki.developer.allizom.org/...`), both of which will be handled by a full page load (so we can take advantage of the hard-wired redirects as well as the rest of our middleware, e.g. `LocaleMiddleware`).

The AWS access keys are acquired from the environment which, in turn, for the `stage`, `prod`, and `standby` Kubernetes clusters gets them from secrets deployed within the cluster. For local testing, I explicitly set the keys (for `stage`) in my `docker-compose.ssl.yml` file (and I also had to rebuild the `Kuma` base image via `VERSION=latest make build-base` to ensure that `boto3` was installed).

For testing, I think the simplest thing would be to push the Kuma docker image for this to staging (where the secrets containing the AWS access keys have already been deployed). I'll work on getting that ready, but I wanted to get this out for review in the meantime.

### DIFF
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from functools import partial
 
 import mock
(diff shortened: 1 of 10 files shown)


escattone:
Hm, the docs build is failing. I suspect it may be due to the upgrade of the `docutils` package from 0.12 to 0.14. I'll check into that.

escattone:
I returned to using `docutils` `0.12` (instead of `0.14`), which should fix the docs build error (at least it worked for me locally). I'll leave the upgrade of `docutils` (and probably `Sphinx` along with it) for a separate PR sometime in the future.

kuma/api/v1/tests/test_views.py
```diff
@@ -7,12 +8,71 @@
...
+        },
+        title='Redirect to Home Page',
+        created=datetime(2015, 7, 4, 11, 15))
+    return redirect_doc
+
+
+@pytest.fixture
+def redirect_to_macros_dashboard(wiki_user):
+    """
+    A top-level English redirect document that redirects to the home page.
```
davidflanagan: s/home page/macros dashboard/

kuma/api/v1/views.py
```diff
@@ -33,10 +34,48 @@ def doc(request, locale, slug):
     return JsonResponse(document_api_data(document))
 
 
-def document_api_data(document):
+def document_api_data(document, ensure_contributors=False):
+    while document.is_redirect:
+        redirect_document = document.get_redirect_document(id_only=False)
+        if redirect_document:
+            document = redirect_document
```
davidflanagan: Would you mind putting a `continue` here? Given how long the else clause is, I was not sure whether there was any more code in the while loop after the else. It would be simpler to see an explicit continue.

davidflanagan:
I'm fine ignoring the minor drop in coverage the codecov is reporting. That tasks.py file is going to be changing so much as we extend the publish process, that any tests for it now will be out of date next week. So better to wait and add tests for that down the road.

escattone:
@peterbe suggested, in a video meeting I had with him earlier today, that we skip publishing to S3, and instead use the Django document API as the CDN origin (instead of the S3 bucket), use a large TTL for the CDN, and add code to Django that invalidates the document API data for a given document whenever that document is rendered anew (and also consider pre-warming the CDN after the invalidation, as part of the post-rendering step). I think that's a compelling idea. We're going to discuss it more tomorrow, but this PR may be closed if we decide to go down that route.

escattone:
Closing this in favor of a new PR coming soon.

PR #5348, opened 2019-04-15
fixes bug 1544084 - Sending first edit email potentially premature

peterbe:
The way I tested manually was messy but insightful. 

The code that calls `send_first_edit_email` used to look like this:

```python
if is_first_edit:
    send_first_edit_email.delay(new_rev.pk)
```

What I did was temporarily edit it to look like this:

```python
# if is_first_edit:
if 1:
    print("Trigger to send email...")
    send_first_edit_email.delay(new_rev.pk)

import time, time
time.sleep(10)
if random.random() > 0.5:
    raise Exception("oh no!")
else:
    print("i'm done sleeping")
```

Then, after editing a wiki page I watched the console (by the way, I changed my email to console printing rather than files inside the docker container). The was the out:

```
<timestamp1>: Trigger to send email...
<timestamp2>: **the whole email string output**
<timestamp3>: i'm done sleeping
```
...or...
```
<timestamp1>: Trigger to send email...
<timestamp2>: **the whole email string output**
<timestamp3>: Traceback (most recent call last):
 ...
Exception: oh no!
```

Point being, the email got sent even before all the other things happened. And had an error occurred the email would still have been sent :(

### DIFF
--- a/kuma/core/tests/__init__.py
+++ b/kuma/core/tests/__init__.py
@@ -1,7 +1,9 @@
 from __future__ import unicode_literals
 
+from functools import wraps
 import os
 
+import mock
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.messages.storage.fallback import FallbackStorage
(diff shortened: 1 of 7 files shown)


peterbe:
There's a subtle connection between this work and https://bugzilla.mozilla.org/show_bug.cgi?id=1544469
My ultimate goal was to get rid of all and any usage `djcelery_transactions` and this particular PR could have perhaps been solved without removing `djcelery_transactions` from where it was used. 

Once this lands, I intend to go after all the other places where `djcelery_transactions.task` gets used. 

Most importantly, wrapping the whole edit in a transaction definitely seems the right thing to do. If one DB write goes wrong, all others should be cancelled too otherwise it might lead to data corruption. 

The wiki document `edit` view function is a critical part of the wiki but for what it's worth, it's only the wiki and not the even more critical part which is the viewing.

kuma/core/tests/__init__.py
```diff
@@ -62,3 +64,42 @@ def assertFileNotExists(self, path):
...
+
+        # In test_something.py
+
+        class MyTests(TestCase):
+
+            @call_on_commit_immediately
+            def test_something(self):
+                self.client.get('/do/something')
+
+    In this example, without the decorator the
```
escattone: Nit: the comment is cut-off mid-sentence. Did you want to add more?

peterbe: Fixed in a followup.

PR #5350, opened 2019-04-16
rel='noopener' everywhere it's not

peterbe:
I know this PR is a beast. It's hard to test beyond CI but I would appreciate some eyes on it for sanity. 

There were very few instances that were scary but some were a little bit. 
Links like `<a href="https://bugzilla.mocilla.org/xxx" target="_blank">` are low risk because they're only dangerous if bugzilla.mozilla.org gets compromised but it's good to make a separation because if any of those URLs it links to are compromised, it compromises us. 

I also noticed instance of `target="_blank"` in the .po files so I could use some help understanding what changes there mean and how to deal with them.

### DIFF
--- a/jinja2/includes/header-main-nav.html
+++ b/jinja2/includes/header-main-nav.html
@@ -60,12 +60,12 @@
                       <a href="{{ wiki_url('MDN/Community') }}">{{ _('Join the MDN community') }}</a>
                   </li>
                   <li>
-                      <a target="_blank" rel="noopener" href="https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title={{ request.build_absolute_uri()|urlencode }}">{{ _('Report a content problem') }}
+                      <a target="_blank" rel="noopener noreferrer" href="https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title={{ request.build_absolute_uri()|urlencode }}">{{ _('Report a content problem') }}
                           {% include 'includes/icons/general/external-link.svg' %}
                       </a>
                   </li>
                   <li>
-                      <a target="_blank" rel="noopener" href="https://bugzilla.mozilla.org/form.mdn">{{ _('Report a bug') }}
+                      <a target="_blank" rel="noopener noreferrer" href="https://bugzilla.mozilla.org/form.mdn">{{ _('Report a bug') }}
                           {% include 'includes/icons/general/external-link.svg' %}
                       </a>
                   </li>
--- a/jinja2/includes/header-main-nav.html
+++ b/jinja2/includes/header-main-nav.html
@@ -60,12 +60,12 @@
                       <a href="{{ wiki_url('MDN/Community') }}">{{ _('Join the MDN community') }}</a>
                   </li>
                   <li>
-                      <a target="_blank" rel="noopener noreferrer" href="https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title={{ request.build_absolute_uri()|urlencode }}">{{ _('Report a content problem') }}
+                      <a target="_blank" rel="noopener" href="https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title={{ request.build_absolute_uri()|urlencode }}">{{ _('Report a content problem') }}
                           {% include 'includes/icons/general/external-link.svg' %}
                       </a>
                   </li>
                   <li>
-                      <a target="_blank" rel="noopener noreferrer" href="https://bugzilla.mozilla.org/form.mdn">{{ _('Report a bug') }}
+                      <a target="_blank" rel="noopener" href="https://bugzilla.mozilla.org/form.mdn">{{ _('Report a bug') }}
                           {% include 'includes/icons/general/external-link.svg' %}
                       </a>
                   </li>
--- a/kuma/attachments/admin.py
+++ b/kuma/attachments/admin.py
@@ -39,7 +39,11 @@ class AttachmentAdmin(DisabledDeleteActionMixin, admin.ModelAdmin):
 
     def full_url(self, obj):
         url = obj.get_file_url()
-        return format_html('<a href="{}" target="_blank">{}</a>', url, url)
+        return format_html(
+            '<a href="{}" target="_blank" rel="noopener noreferrer">{}</a>',
+            url,
+            url
+        )
     full_url.short_description = 'Full URL'
 
     def delete_revisions(self, request, revisions):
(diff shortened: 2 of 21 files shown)


peterbe:
@davidflanagan 
> Is 'noreferrer' a privacy thing or is there also a security issue there? 

A security thing. See https://mathiasbynens.github.io/rel-noopener/

I'm terrible at that kind of thing but I suspect that the more correct way is to include those Google Analytics query string parameters so they can at least track where the traffic came from.

schalkneethling:
@peterbe I am completely in favour of this change. The only slight concern I have is if any of these external services that we link to might depend on knowing the `Referer`. It seems unlikely, and I do not believe we have automated tests that will catch these so, we might want to push it out and keep an eye out for bugs being opened that could be traced down to a missing referer.

With regards to the translation bits. I agree this seems problematic, but I do not know enough about that process to be able to provide any good suggestions. Pinging @escattone 🦇

peterbe:
Yeah, I'm starting to rethink the use of `noreferrer`. 

The thing is, `rel="noopener"` is the minimum to prevent that new page to access the `window.opener`. 

Adding `noreferrer` does "ruin" it for that site you land on. It leaks less private information but it feels only like a privacy thing rather than a security thing and I think the two should be separate concerns. 

In [Mathias's blog post](https://mathiasbynens.github.io/rel-noopener/) he says: *"For older browsers, you could use rel=noreferrer which also disables the Referer HTTP header"* and I have no idea what that *really* means. Anybody? Could it be that someone using IE 8, clicking on `<a href="http://couldbep0wned.com" target="_blank" rel="noopener">` is not protecting their `window.opener`??

The fact that 99% of our "partners", that we link to (bugzilla, codepen, stripe), are responsible players (but could be hacked!) tells me it's more useful to share the `Referer` HTTP header than not. It's also the "Web way of doing things".

So I think I'm going to revert the changes and switch to `rel="noopener"` unless anybody knows a strong reason to *never* leak the `Referer` HTTP header.

peterbe:
For the record, and I [tested this](https://codepen.io/peterbe/pen/WWMMxz?editors=1000#0), if you use `<a href="{{url}}" target="_blank" rel="noopener">` that `{{url}}` will receive the `Referer` HTTP header which can be useful for them. If effectively works the same as `<a href="{{url}}" target="_blank">`.

peterbe:
So a snag; Attention @davidflanagan  
See https://github.com/yannickcr/eslint-plugin-react/issues/2054 
Basically, the [eslint `jsx-no-target-blank` rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md) seems to require that *both* things are there (`noopener` (no-brainer) AND `noreferer`)

Currently, in [kuma master for `header.jsx`](https://github.com/mozilla/kuma/blob/master/kuma/javascript/src/header/header.jsx#L174) we use both. Meaning, any header link with `target="_blank"` loses the referer (sorry Codepen etc.). If you try to change `header.jsx` to *only* have `noopener` you get an eslint error which prevents a git commit because that's protected with a git pre-commit hook. 

For now, perhaps best to leave `header.jsx` alone and proceed with the rest. 

I put thumbs up in various places and wrote: https://github.com/yannickcr/eslint-plugin-react/issues/2054#issuecomment-484641410

peterbe:
So did the safest thing and rewrote it so that it's always `rel="noopener"` only. That way we do "leak" the referer which is almost always ideal for MDN. 
And for the `header.jsx` I left it as is for now. 

r?

peterbe:
@davidflanagan This got merged now as I suspect @schalkneethling missed our discussion above .po files. 

> Having target="_blank" in our translations files seems like a bug to me. But I see that task-completion.js is causing at least one case of that. 

I actually feel unconfident about all of this because l10n is pretty new to me. 

Can you guide me in what else is now next to do to close out the bug?

peterbe:
Also, I am feeling somewhat unimpressed with my own work on this. It was just too large to do it all super carefully. 
What we really should do is make a distinction between our links and user content links. In particular, about the `referer` vs. `noreferer` thing.

If someone edits the wiki and tries to put in a link to their untrusted site. Malicious or not, it might be best to assure that it uses `noreferer` which we don't know it very well. However for statically typed links (like, in jinja templates) where we link to codepen or bugzilla, there we might have `target="_blank" rel="noopener"`.

davidflanagan:
@peterbe: I'm no expert on our l10n code paths. I don't really understand the format of .po files, and I haven't looked to see what you saw there. It does look like task-completion.js is putting a target="_blank" in there, and now with this merge, that will probably change to a `target="_blank" rel="noopener"`.  I think that since that is a new string, translators may find themselves prompted to translate it, which is a little bit of unnecessary work. On the other hand, I think that maybe the task-completion.js module is an experiment that may be turned off right now? This is where things get fuzzy for me... Not being a translator, I don't know what kind of UX the translators see, or whether they will know that this is a very low priority string for them to translate.

In any case, I suspect that our translators are used to this kind of churn, and it is probably not a big deal.

As far as I'm concerned, you can close this now. Maybe file a new bug to fix the gettext() call in task-completion.js to remove the html attributes from the part that gets translated? If you do that we'll at least remember to discuss the issue at the next bug triage and we can decide whether it is worth fixing. 

I actually have no idea what the policy is about external links in user-edited wiki articles... Do we allow them? Does our bleach code automatically add target and rel attributes? Or maybe they're just not allowed at all?  If you're able to edit a page to successfully add an external link, then it sounds like a good idea to me to file a followup bug to make sure that they're appropriately decorated with target and rel attributes (and I'd guess that is part of the bleaching process).

peterbe:
@davidflanagan 
First of all, studying the code on: https://github.com/mozilla/kuma/blob/091ab4cc617e774c50e5d5ac66f5ebfae1bf9ef8/kuma/static/js/task-completion.js#L37
it seems to me that what translators are confronted with is not different as of this change. They see:

```
Would you answer 4 questions for us? <a %(url)s>Open the survey in a new tab</a> and fill ...
```
and don't have to worry about the `target` and `rel` stuff at all. 
So, it's right to NOT worry about this particular thing. 

Secondly, regarding...

> Does our bleach code automatically add target and rel attributes?

I clicked to edit a page and skipped the cfkeditor went into "source mode" and typed in

```
<p>
<a href="https://www.peterbe.com">www.peterbe.com</a>
</p>
```
and after publishing and looking at the rendered HTML it became:
```html
<p>
<a class="external external-icon" href="https://www.peterbe.com" rel="noopener">www.peterbe.com</a>
</p>
```
So that's something. I think if someone makes links to not-malicious sites it's fine like that. We also have content chiefs who look after edits and check that links aren't too spammy or dangerous.

davidflanagan:
Sorry for misreading task-completion.js. I wonder where the _blank you saw in the .po file was coming from, then?

schalkneethling:
>  On the other hand, I think that maybe the task-completion.js module is an experiment that may be turned off right now? 

This is controlled via a Waffle flag called `sg_task_completion` and shown to a small percentage of users. It is indeed turned off currently.

PR #5354, opened 2019-04-18
fixes 1544482 - elasticsearch 6.x

peterbe:
I can search and I can reindex. 

There might be some stuff in `kuma/wiki/search.py` that is obsolete not with the upgrade of `elasticsearch_dsl`. For example, in other projects where I use ES 6.x and `elasticsearch_dsl` I don't specify a mapping in the sub-class `Meta` but instead you specify settings and an index name. However, kuma is odd in that you created and define indexes using the Django Admin. I didn't want to rock the boat too much at this point. In particular, I'm not fully grasped up on why we bother doing indexes through the Django Admin. I've never done it like that before and it all feels, to me, right now, like overkill.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,11 +97,13 @@ services:
       - mysqlvolume:/var/lib/mysql
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:5.6.10
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.7.1
     environment:
       - discovery.type=single-node
       - xpack.security.enabled=false
       - ES_JAVA_OPTS=-Xms512m -Xmx512m
+    ports:
+      - "9200:9200"
     volumes:
       - esdata:/usr/share/elasticsearch/data
 
--- a/kuma/search/management/commands/reindex.py
+++ b/kuma/search/management/commands/reindex.py
@@ -11,7 +11,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument(
             '-c', '--chunk',
-            help='Chunk size when reindexing (default 100). Lower is better'
+            help='Chunk size when reindexing (default 1000). Lower is better'
                  'for slow computers with little memory',
             type=int,
             dest='chunk_size',
--- a/kuma/wiki/search.py
+++ b/kuma/wiki/search.py
@@ -25,7 +25,7 @@ from .constants import EXPERIMENT_TITLE_PREFIX
 log = logging.getLogger('kuma.wiki.search')
 
 
-class WikiDocumentType(document.DocType):
+class WikiDocumentType(document.Document):
     excerpt_fields = ['summary', 'content']
     exclude_slugs = ['Talk:', 'User:', 'User_talk:', 'Template_talk:',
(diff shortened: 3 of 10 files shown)


docker-compose.yml
```diff
@@ -77,11 +77,13 @@ services:
...
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:5.6.10
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.7.1
     environment:
       - discovery.type=single-node
       - xpack.security.enabled=false
       - ES_JAVA_OPTS=-Xms512m -Xmx512m
+    ports:
+      - "9200:9200"
```
peterbe: This is a bit controversial. I first added it just so I can poke around inside ES from my browser on the host. But I think that's likely to happen again so I decided to leave it. It would cause trouble, in local dev, if you already have an ES running in another terminal. But that's no different from using `:8000` in another terminal too.

kuma/search/management/commands/reindex.py
```diff
@@ -11,7 +11,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument(
             '-c', '--chunk',
-            help='Chunk size when reindexing (default 100). Lower is better'
+            help='Chunk size when reindexing (default 1000). Lower is better'
```
peterbe: Unrelated but it couldn't *not* see it once I saw it.

kuma/wiki/search.py
```diff
@@ -70,7 +70,7 @@ def case_insensitive_keywords(cls, keywords):
 
     @classmethod
     def from_django(cls, obj):
-        is_root_document = (obj.slug.count('/') == 1)
+        is_root_document = obj.slug.count('/') == 1
```
peterbe: Unrelated but it couldn't *not* see it once I saw it.

peterbe:
Uh?! coverage went down in [test_filters.py](https://codecov.io/gh/mozilla/kuma/pull/5354/diff?src=pr&el=tree#diff-a3VtYS9zZWFyY2gvdGVzdHMvdGVzdF9maWx0ZXJzLnB5)
I should look into that.

peterbe:
I forgot to update the `.travis.yml` and I also, there's an even newer version of elasticsearch-dsl now.

peterbe:
I forgot to update the `.travis.yml` and I also, there's an even newer version of elasticsearch-dsl now.

peterbe:
@escattone I did a poor job before. I failed to grep for "elasticsearch 5" properly before. Now I've updated the `.travis.yml` and the mentions in RST files.

escattone:
#### Next steps before the next push
- [x] Create new Elasticsearch version 6.7.1 deployments (via the elastic.co console) for stage & prod
- [ ] Update the `elasticsearch-url` key within the Kubernetes `mdn-secrets` resource for stage, prod, and standby with the new URL's from step 1 above
- [ ] Manually deploy the `mdn-secrets` resource on stage, prod, and standby

#### Step after the next push
- [ ] Re-index stage
- [ ] Re-index prod
- [ ] Delete the no-longer-used Elasticsearch version 5 deployments (via the elastic.co console)

PR #5357, opened 2019-04-19
Bug 1539793, fix landing page for RTL locales

schalkneethling:
@davidflanagan r? I used `fa` for testing RTL

### DIFF
--- a/kuma/static/styles/components/home/callouts/_grid-tools.scss
+++ b/kuma/static/styles/components/home/callouts/_grid-tools.scss
@@ -67,7 +67,8 @@ Fx-dev callout
             display: inline-block;
             background: transparent url($path-to-images + 'embed/promos/fx-dev-logo.png') 0 0 no-repeat;
             background-size: 32px 33px;
-            margin-right: 5px;
+            @include bidi(((margin-right, 5px, margin-left, 5px),));
+            @include bidi(((margin-right, 5px, margin-left, 0),));
             width: 36px;
             height: 36px;
             vertical-align: middle;
--- a/kuma/static/styles/components/home/callouts/_grid-tools.scss
+++ b/kuma/static/styles/components/home/callouts/_grid-tools.scss
@@ -68,7 +68,6 @@ Fx-dev callout
             background: transparent url($path-to-images + 'embed/promos/fx-dev-logo.png') 0 0 no-repeat;
             background-size: 32px 33px;
             @include bidi(((margin-right, 5px, margin-left, 5px),));
-            @include bidi(((margin-right, 5px, margin-left, 0),));
             width: 36px;
             height: 36px;
             vertical-align: middle;
--- a/kuma/static/styles/components/newsletter.scss
+++ b/kuma/static/styles/components/newsletter.scss
@@ -68,7 +68,7 @@ input.newsletter-input-email {
         top: 30%;
         @include bidi(((right, 10px, left, 10px),));
         @include bidi(((left, auto, right, 10px),));
-        transition: right .2s;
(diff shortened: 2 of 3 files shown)


schalkneethling:
@davidflanagan Do you have a moment to have a look at this one? Thanks!

kuma/static/styles/components/home/callouts/_grid-tools.scss
```diff
@@ -67,7 +67,8 @@ Fx-dev callout
             display: inline-block;
             background: transparent url($path-to-images + 'embed/promos/fx-dev-logo.png') 0 0 no-repeat;
             background-size: 32px 33px;
-            margin-right: 5px;
+            @include bidi(((margin-right, 5px, margin-left, 5px),));
+            @include bidi(((margin-right, 5px, margin-left, 0),));
```
davidflanagan: These two lines seem contradictory. I don't understand why you would need both.

kuma/static/styles/components/newsletter.scss
```diff
@@ -61,13 +61,13 @@ input.newsletter-input-email {
     position: relative;
     width: 100%;
     padding: ($grid-spacing / 2);
-    @include bidi(((padding-right, 2em, padding-left, inherit),));
-    text-align: left;
+    @include bidi(((text-align, left, right),));
```
davidflanagan: I'm assuming that the padding you deleted has been replaced by the arrow positioning stuff below?

kuma/static/styles/components/newsletter.scss
```diff
@@ -61,13 +61,13 @@ input.newsletter-input-email {
...
 
     .icon-arrow {
         position: absolute;
         top: 30%;
-        right: 10px;
+        @include bidi(((right, 10px, left, 10px),));
+        @include bidi(((left, auto, right, 10px),));
```
davidflanagan: IIUC, in ltr, this becomes:

```
right: 10px;
left: auto;
```

And in RTL, it becomes:

```
left: 10px;
right: 10px;
right: auto:
left: 10px;
```

So it seems like the 3-argument version of the bidi macro might make more sense here? Or am I misunderstanding something?

Also, given the nature of the output produced by bidi() is it more efficient to call bidi() once and pass both lists to a single invocation? (I don't know if it makes a difference, but given that it is an option, I'm wondering about it.)

kuma/static/styles/components/payments/banner.scss
```diff
@@ -50,6 +50,11 @@
             height: $payment-banner-popover-collapsed-height;
         }
 
+        .column-container > [class^='column-'],
+        .column-container > [class^='column-']:last-child {
```
davidflanagan: I can't see why it should be necessary to explicitly override the last-child here. So consider adding a comment to clarify.

Given that we've now turned the payment banners off, this may be hard to test, and I suppose you could just omit these changes if that would be simpler.

kuma/static/styles/components/payments/banner.scss
```diff
@@ -194,6 +199,7 @@
     }
 
     .payments-form {
+        @include bidi(((float, right, left),));
```
davidflanagan: It looks like this change affects the LTR layout as well as the RTL, right? Just want to make sure it is an intentional change, not an accidental one.

schalkneethling:
@davidflanagan Thanks for the feedback. You make a good point about the payments cleanup. Let's put this one on ice until I have completed that. I will then rebase this branch and see what remains.

schalkneethling:
@davidflanagan I finally got back to this one. Please take a look when you have a moment. Thanks!

PR #5359, opened 2019-04-19
fixes bug 1114650 - stay logged in much longer

peterbe:
Easy change. Less easy in terms of security discussion. 

The way I see it, 2 weeks is far too short and it's annoying to be thrown out. 
Our site doesn't really have anything critical. At best you can edit wiki documents but the money $tuff that we get involved in is shifted onto Stripe anyway. There's no sharing of critical PII. 
In other words, I feel like we can allow for a much longer lasting session cookie. Rogueness can lead to destroying stuff but it can't really lead to revealing sensitive stuff. 

My ideal functionality is a bit like GitHub and Amazon.com. You're more or less always logged in if you've logged in once but if you haven't manually logged in recently, and you click on something "more sensitive" then it triggers a fresh check that I still know my password. 

I'm only speculating but I wonder if we could, when you click `https://developer.mozilla.org/en-US/profiles/${username}/edit`, check that current user's session cookie age as it's known in the MySQL database. If it's older than say 1 day, then redirect the user to GitHub to sign in again, just like you'd do if someone goes straight to `https://developer.mozilla.org/en-US/profiles/${username}/edit` (from a bookmark or manually typing the URL) on a browser with no cookies.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1192,6 +1192,11 @@ for override in pipeline_overrides:
 SESSION_COOKIE_DOMAIN = DOMAIN
 SESSION_COOKIE_SECURE = config('SESSION_COOKIE_SECURE', default=True, cast=bool)
 SESSION_COOKIE_HTTPONLY = True
+SESSION_COOKIE_AGE = config(
+    'SESSION_COOKIE_AGE',
+    default=60 * 60 * 24 * 365,
+    cast=int
+)
 
 WAFFLE_SECURE = config('WAFFLE_COOKIE_SECURE', default=True, cast=bool)
 # This is a setting unique to Kuma which specifies the domain


peterbe:
Actually, I just checked. I pasted `https://developer.mozilla.org/en-US/profiles/peterbe/edit` into my Chrome browser and it didn't redirect me, automatically, to GitHub.com to sign in. It took dynamically rendered the "Permission Denied" page and there it had a link to sign in. Hmm... That's kinda annoying.

peterbe:
Also, it gives me heart to see that we're doing cookies correctly to boot: https://observatory.mozilla.org/analyze/developer.mozilla.org 
I.e. `Secure` and `HttpOnly`.

peterbe:
Another reason why the current cookie expiry age is so wrong is that oftentimes, I get logged out but GitHub remembers me so all I have to do is click "Sign in" and I get immediately redirected back. 
So what was the point of "signing me out" of kuma in the first place?

callahad:
With this change, I'd still have to log in once a year.

Would it be better to refresh the cookie more often, so it only expires if you don't visit MDN for a fortnight?

peterbe:
> With this change, I'd still have to log in once a year. 

Yeah, perhaps equally "dangerous" because if you only sign in once a year, it is even more likely that you forget which authentication provider you used last time.
(easy to fix if the `localStorage` persists longer than the cookie)

> Would it be better to refresh the cookie more often, 

That might be worthwhile doing. I guess you could use the `KumaUser.last_logged_in` (or whatever it's called) and try to update the cookie that way.

peterbe:
Truth be told @callahad despite years of using Django I don't know much about this kinda stuff and what's possible. 

Perhaps we can just set the cookie to last 10 years. 

Thing is, you could set it for 10 years, and every time the user is trying to do something more sensitive (E.g. ban a spammy user) you look at the `KumaUser.last_logged_in` and if it's less than 1 day, you redirect to a page that urges you to log in again. Like GitHub does when it just asks for your password one more time.

atopal:
Let's go ahead with this

PR #5360, opened 2019-04-19
fixes bug 1544078 - rendered_html shouldn't update if rendered_errors

peterbe:
**INCOMPLETE!** This doesn't have tests yet. Or rather, all old tests passed just passed despite this rather critical change. 

Steps to reproduce are here: https://bugzilla.mozilla.org/show_bug.cgi?id=1544078#c6
However, a better thing to do is to 
1. Click the "Edit" button, type in a change, but before you hit "Publish"
2. Go to http://localhost:8000/admin/constance/config/ and set the `KUMASCRIPT_TIMEOUT` to 0.1
3. Go back to edit and now press "Publish"
What you should see is the old page *without* your change. But you shouldn't see `{{SomeMacro(...)}}` and stuff like that. 

As of this change; what you get now is this:
Screen Shot 2019-04-19 at 4 13 23 PM
And the search snippets don't show `{{SomeMacro(...)}}` stuff. 

What's not clear to me is how are these pages re-attempted?? Is there a cron job that does that (aka. django celery).

By the way, are we sure it's safe to fully display that error (when you click the "More information about this error"). What if that reveals something sensitive that we don't want to expose about our infra?!

Lastly, note that this change does NOT attempt to change how the document is **viewed**. I didn't want to rock the boat too much. Also, at the moment, I haven't had any feedback on [this needinfo](https://bugzilla.mozilla.org/show_bug.cgi?id=1544078#c7).

### DIFF
--- a/kuma/wiki/models.py
+++ b/kuma/wiki/models.py
@@ -539,6 +539,9 @@ class Document(NotificationsMixin, models.Model):
         Document.objects.filter(pk=self.pk).update(render_scheduled_at=now)
         self.render_scheduled_at = now
 
+        print("INSIDE schedule_rendering\tself.defer_rendering=", self.defer_rendering)
+        # raise Exception
+
         if self.defer_rendering:
             # Attempt to queue a rendering. If celery.conf.ALWAYS_EAGER is
             # True, this is also an immediate rendering.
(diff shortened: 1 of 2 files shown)


peterbe:
@davidflanagan wanna review this one too? It's "python-heavy" but I don't think it matters that much. The code isn't the hard part. It's the whole everything.

peterbe:
Ugh. There's something weird going on. I don't know if I corrupted my own document with debugging/testing but now it won't try to re-render after I edit it again even with the `KUMASCRIPT_TIMEOUT` raised up to 10sec.

peterbe:
Off-site discussion consensus was that it's better to display errored but up-to-date than the opposite of that.

PR #5361, opened 2019-04-19
bug 1525719: publish/unpublish to/from the doc API

escattone:
#### Description
This PR provides the foundation for connecting, via the document API, the Wiki side of MDN (creating, editing, deleting, restoring, reverting, and moving documents) with the new read-only, `beta` domain side of MDN. All operations on the Wiki side (listed above) will be automatically reflected to the document API via handlers connected to Django signals (i.e., `post_delete`, `render_done`, and `restore_done`). Those three signals, according to my testing, cover all of the operations listed above, triggering handlers that in turn either `publish` or `unpublish` to the S3 bucket that serves the document API. So for example, when a document is deleted, the `post_delete` handler will launch an `unpublish` task into Celery, which when run within a Celery worker, will delete the pre-rendered JSON file in S3 that corresponds to that document.

This PR also provides two new management commands: `publish` and `unpublish`. The purpose of both of these commands is twofold. Primarily, they provide a way to manually `publish` or `unpublish` for those times when the automatic Celery tasks fail for whatever reason. However, in the case of the `publish` command, it also exists to "prime the pump" so to speak, to initially publish all of the documents.

The idea here is that, after an initial publication of all of the documents, the document API will automatically track all changes as they're made within the Wiki domain: document re-renders, document edits, document reverts, document restores, document moves, the creation of new documents, and so on.

The `publish` and `unpublish` Celery tasks are brokered through the new `mdn_api` Celery queue. Initially, all of the Celery workers will be configured to pull from that queue (see https://github.com/mdn/infra/pull/236), but I've created that new queue so that in the near future those tasks could be handled by one or more Celery workers configured to use a read-replica of the current production MySQL RDS instance in order to avoid loading-down the production database.

#### Infrastructure Context
The infrastructure context here is that the document API will be served via CloudFront backed by an S3 origin serving pre-rendered JSON files (and handling redirects from one key in S3 to another). This provides a simple, fast, and efficient way to serve the document API, and is already in place within the staging environment (for now the TTL on the stage CloudFront instance for all `/api/v1/doc/*` paths is 12 hours). There is no need for invalidation of paths within CloudFront. When the TTL expires (after 12 hours or whatever we finally choose) for any given path, CloudFront will return to S3 to get the latest data (if needed, since it'll make a conditional request using the `ETag` and `If-None-Match` HTTP headers). Of course, this means that any document page within the read-only `beta` domain may be at most 12 hours out-of-date with the same document page in the Wiki domain.

#### Redirects
This document API as implemented within this PR also handles content-based redirects (a redirect directive embedded within a document). There are two kinds of content-based redirects (both are 301's): a redirect from one document to another, as well as a redirect from a document to a non-document page. The first kind of redirect, the redirect from one document to another, has been built upon S3's ability to handle redirects. The origin for CloudFront uses the S3 bucket's "website" domain rather than its raw domain, since the "website" domain will automatically handle the redirection locations embedded as metadata within the S3 objects. So, for example, a request via the document API for `/api/v1/doc/en-US/MDN/Promote` will be 301 re-directed to `/api/v1/doc/en-US/MDN/About/Promote` (this 301 is cached within the CDN after the first miss), so the `fetch` within the React `DocumentProvider` call must be configured to follow redirects. Since the returned JSON, in the case of redirects from one document to another, will be for a different locale and/or slug, the `DocumentProvider` code must account for that and update the `history` accordingly (which I've taken a stab at within this PR, but only as a first attempt, and I'll leave it to @davidflanagan to finalize that code).

The redirects from document pages to non-document pages are handled by passing the URL for the non-document page in the `redirectURL` attribute of the JSON returned by the document API, which is normally `null`. When it's not `null`, it contains either a relative or full URL to a non-document page.

#### Testing
I've tested this locally, and in fact populated the stage S3 bucket serving the document API from my sample database. The AWS access keys are already deployed as secrets within the stage, production, and standby Kubernetes clusters, and the stage/production S3 bucket names are already baked into the configuration settings for stage, production, and standby (for example. see https://github.com/mdn/infra/blob/ad561b1fb13852ac2564334b71f80ed5d9ff307f/apps/mdn/mdn-aws/k8s/regions/oregon/stage.sh#L106). Once this PR is merged, I'll re-publish the document API for stage from the actual stage database, but for now there are enough documents to test on stage.

Even though you won't have the AWS credentials for actually publishing/unpublishing to/from S3, i'd still recommend confirming that those tasks are launched (you'll see either a `Skipping publish for XXX: no S3 bucket configured` or `Skipping unpublish for XXX: no S3 bucket configured` message in the `worker` log -- `docker-compose logs -f worker`) when you try each of the following:
- create a new document
- edit a document
- delete a document
- restore a deleted document
- revert a document to an earlier revision
- move a document

You can also do this while testing the management commands, for example:
```sh
docker-compose exec web bash
./manage.py publish --locale en-US --chunk-size 10
./manage.py unpublish /en-US/docs/Web/CSS /en-US/docs/Web/HTML
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,6 +17,8 @@ services:
       - ALLOW_ALL_IFRAMES=${ALLOW_ALL_IFRAMES:-False}
       - ALLOWED_HOSTS=*
       - ATTACHMENT_HOST=${ATTACHMENT_HOST:-localhost:8000}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-}
       - BROKER_URL=redis://redis:6379/0
       - CELERY_ALWAYS_EAGER=False
       - CELERY_RESULT_BACKEND=redis://redis:6379/1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2.1'
 services:
   worker: &worker
     image: mdnwebdocs/kuma_base
-    command: ./manage.py celery worker --loglevel=INFO --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
+    command: ./manage.py celery worker --loglevel=INFO --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,mdn_api,celery
     user: ${UID:-1000}
     volumes:
       - ./:/app:z
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -506,14 +506,10 @@ simply be skipped (although, for verification/debugging purposes, you can see
 the detailed skip messages in the ``worker`` log (
 ``docker-compose logs -f worker``).
 
-However, if for testing purposes you'd like
-to locally configure the ``publish`` and ``unpublish`` tasks to use S3, you can
-simply add the following::
+However, if for testing purposes you'd like to locally configure the
+``publish`` and ``unpublish`` tasks to use S3, you can simply add the
+following to your ``.env`` file::
 
-    - MDN_API_S3_BUCKET_NAME=<your-s3-bucket-name>
-    - AWS_ACCESS_KEY_ID=<your-aws-access-key>
-    - AWS_SECRET_ACCESS_KEY=<your-aws-secret-key>
-
-to the ``environment`` section of the ``worker`` service within either
-``docker-compose.ssl.yml`` or ``docker-compose.yml``, depending on whether or
-not your serving over SSL/HTTPS.
+    MDN_API_S3_BUCKET_NAME=<your-s3-bucket-name>
+    AWS_ACCESS_KEY_ID=<your-aws-access-key>
+    AWS_SECRET_ACCESS_KEY=<your-aws-secret-key>
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -491,3 +491,29 @@ value you can edit your ``.env`` file. For example::
 
 The ``docker-compose.yml`` will read this and start ``gunicorn`` and the
 ``celery`` worker with this setting.
+
+Configuring AWS S3
+==================
+
+The ``publish`` and ``unpublish`` Celery tasks and Django management commands
+require AWS S3 to be configured in order for them to do any real work, that is,
+creating/updating/deleting S3 objects used by the stage/production document API.
+In stage and production, the S3 bucket name as well as the AWS credentials are
+configured via the container environment, which in turn, gets the AWS credentials
+from a Kubernetes ``secrets`` resource. For local development, there is no need
+for any of this configuration. The ``publish`` and ``unpublish`` tasks will
+simply be skipped (although, for verification/debugging purposes, you can see
+the detailed skip messages in the ``worker`` log (
+``docker-compose logs -f worker``).
+
+However, if for testing purposes you'd like
+to locally configure the ``publish`` and ``unpublish`` tasks to use S3, you can
+simply add the following::
+
+    - MDN_API_S3_BUCKET_NAME=<your-s3-bucket-name>
+    - AWS_ACCESS_KEY_ID=<your-aws-access-key>
+    - AWS_SECRET_ACCESS_KEY=<your-aws-secret-key>
+
+to the ``environment`` section of the ``worker`` service within either
+``docker-compose.ssl.yml`` or ``docker-compose.yml``, depending on whether or
+not your serving over SSL/HTTPS.
--- a/kuma/api/management/commands/publish.py
+++ b/kuma/api/management/commands/publish.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+"""
+Manually schedule the publishing of one or more documents to the document API.
+"""
+from __future__ import unicode_literals
+
+from collections import namedtuple
+from math import ceil
+
+from celery.canvas import group
+from django.core.management.base import BaseCommand, CommandError
+
+from kuma.api.tasks import publish
+from kuma.core.utils import chunked
+from kuma.wiki.models import Document
+
+
+class Command(BaseCommand):
+    args = '<document_path document_path ...>'
+    help = 'Publish one or more documents to the document API'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'paths',
+            help='Path to document(s), like /en-US/docs/Web',
+            nargs='*',  # overridden by --all or --locale
+            metavar='path')
+        parser.add_argument(
+            '--all',
+            help='Publish ALL documents (rather than by path)',
+            action='store_true')
+        parser.add_argument(
+            '--locale',
+            help='Publish ALL documents in this locale (rather than by path)')
+        parser.add_argument(
+            '--chunk-size',
+            type=int,
+            default=1000,
+            help='Partition the work into tasks, with each task handling this '
+                 'many documents (default=1000)')
+
+    def handle(self, *args, **options):
+        Logger = namedtuple('Logger', 'info, error')
+        log = Logger(info=self.stdout.write, error=self.stderr.write)
+        if options['all'] or options['locale']:
+            filters = {}
+            if options['locale'] and not options['all']:
+                locale = options['locale']
+                log.info('Publishing all documents in locale {}'.format(locale))
+                filters.update(locale=locale)
+            else:
+                locale = None
+                log.info('Publishing all documents')
+            chunk_size = max(options['chunk_size'], 1)
+            docs = Document.objects.filter(**filters)
+            doc_pks = docs.values_list('id', flat=True)
+            num_docs = len(doc_pks)
+            num_tasks = int(ceil(num_docs / float(chunk_size)))
+            log.info('...found {} documents.'.format(num_docs))
+            # Let's publish the documents in a group of chunks, where the
+            # tasks in the group can be run in parallel.
+            tasks = []
+            for i, chunk in enumerate(chunked(doc_pks, chunk_size)):
+                message = 'Published chunk #{} of {}'.format(i + 1, num_tasks)
+                tasks.append(publish.si(chunk, completion_message=message))
+            if num_tasks == 1:
+                msg = ('Launching a single task handling '
+                       'all {} documents.'.format(num_docs))
+            else:
+                msg = ('Launching {} paralellizable tasks, each handling '
+                       'at most {} documents.'.format(num_tasks, chunk_size))
+            log.info(msg)
+            group(*tasks).apply_async()
+        else:
+            paths = options['paths']
+            if not paths:
+                raise CommandError('Need at least one document path to publish')
+            doc_pks = []
+            get_doc_pk = Document.objects.values_list('id', flat=True).get
+            for path in paths:
+                if path.startswith('/'):
+                    path = path[1:]
+                locale, sep, slug = path.partition('/')
+                head, sep, tail = slug.partition('/')
+                if head == 'docs':
+                    slug = tail
+                try:
+                    doc_pk = get_doc_pk(locale=locale, slug=slug)
+                except Document.DoesNotExist:
+                    msg = 'Document with locale={} and slug={} does not exist'
+                    log.error(msg.format(locale, slug))
+                else:
+                    doc_pks.append(doc_pk)
+            publish(doc_pks, log=log)
--- a/kuma/api/management/commands/unpublish.py
+++ b/kuma/api/management/commands/unpublish.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+"""
+Manually schedule the removal of one or more documents from the document API.
+"""
+from __future__ import division
+from __future__ import unicode_literals
+
+from collections import namedtuple
+
+from django.core.management.base import BaseCommand, CommandError
+
+from kuma.api.tasks import unpublish
+
+
+class Command(BaseCommand):
+    args = '<document_path document_path ...>'
+    help = 'Remove one or more documents from the document API'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'paths',
+            help='Path to document(s), like /en-US/docs/Web',
+            nargs='*',
+            metavar='path')
+
+    def handle(self, *args, **options):
+        Logger = namedtuple('Logger', 'info, error')
+        log = Logger(info=self.stdout.write, error=self.stderr.write)
+        paths = options['paths']
+        if not paths:
+            raise CommandError('Need at least one document path to remove')
+        doc_locale_slug_pairs = []
+        for path in paths:
+            if path.startswith('/'):
+                path = path[1:]
+            locale, sep, slug = path.partition('/')
+            head, sep, tail = slug.partition('/')
+            if head == 'docs':
+                slug = tail
+            doc_locale_slug_pairs.append((locale, slug))
+        unpublish(doc_locale_slug_pairs, log=log)
--- a/kuma/api/management/commands/unpublish.py
+++ b/kuma/api/management/commands/unpublish.py
@@ -2,7 +2,6 @@
 """
 Manually schedule the removal of one or more documents from the document API.
 """
-from __future__ import division
 from __future__ import unicode_literals
 
 from collections import namedtuple
--- a/kuma/api/signal_handlers.py
+++ b/kuma/api/signal_handlers.py
@@ -1,0 +1,40 @@
+from __future__ import unicode_literals
+
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+
+from kuma.wiki.models import Document
+from kuma.wiki.signals import render_done, restore_done
+
+from .tasks import publish, unpublish
+
+
+@receiver(restore_done, sender=Document,
+          dispatch_uid='api.document.restore_done.publish')
+def on_restore_done(sender, instance, **kwargs):
+    """
+    A signal handler to publish the document to the document API after it
+    has been restored.
+    """
+    publish.delay([instance.pk])
+
+
+@receiver(render_done, sender=Document,
+          dispatch_uid='api.document.render_done.publish')
+def on_render_done(sender, instance, **kwargs):
+    """
+    A signal handler to publish the document to the document API after it
+    has been rendered.
+    """
+    if not instance.deleted:
+        publish.delay([instance.pk])
+
+
+@receiver(post_delete, sender=Document,
+          dispatch_uid='api.document.post_delete.unpublish')
+def on_post_delete(instance, **kwargs):
+    """
+    A signal handler to remove the given document from the document API after
+    it has been deleted.
+    """
+    unpublish.delay([(instance.locale, instance.slug)])
--- a/kuma/api/tasks.py
+++ b/kuma/api/tasks.py
@@ -12,14 +12,30 @@ from kuma.wiki.models import Document
 from .v1.views import document_api_data, get_content_based_redirect, get_s3_key
 
 
+_s3_resource = None
 S3_MAX_KEYS_PER_DELETE = 1000
 
 
-def get_s3_bucket():
-    """ Get the S3 bucket name from the environment, otherwise None."""
+def get_s3_resource(config=None):
+    """
+    Get or create the S3 resource. This function is not thread-safe, since it
+    uses the default session, rather than a separate session for each thread.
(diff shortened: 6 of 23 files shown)


kuma/api/v1/views.py
```diff
@@ -30,45 +32,104 @@ def doc(request, locale, slug):
...
 
-def document_api_data(document):
-    translations = document.get_other_translations(
-        fields=('locale', 'slug', 'title'))
+def get_s3_key(doc=None, locale=None, slug=None, for_redirect=False):
+    if doc:
+        locale, slug = doc.locale, doc.slug
+    key = reverse('api.v1.doc', args=(locale, slug), urlconf='kuma.urls_beta')
+    if for_redirect:
+        # Redirects within an S3 bucket must be prefixed with "/".
```
davidflanagan: Would it be simpler then to just always keep the leading / for all keys?

kuma/javascript/src/document-provider.jsx
```diff
@@ -67,7 +67,7 @@ export default function DocumentProvider(
         // This is the function that does client side navigation
         function navigate(url, localeAndSlug) {
             body.style.opacity = '0.15';
-            fetch(`/api/v1/doc${localeAndSlug}`)
+            fetch(`/api/v1/doc${localeAndSlug}`, {redirect:'follow'})
```
davidflanagan: MDN says that redirect:follow is the default, but implies that it might only be the default on Chrome? So this might not be necessary, but it seems safer to make it explicit, as you have done.

kuma/javascript/src/document-provider.jsx
```diff
@@ -80,9 +80,20 @@ export default function DocumentProvider(
                     }
                 })
                 .then(json => {
-                    window.scrollTo(0, 0);
-                    setDocumentData(json);
-                    body.style.opacity = '1';
+                    if (json.redirectURL) {
+                        window.location = json.redirectURL;
```
davidflanagan: I'd add a comment before this line:

// We got a redirect to a document that can't be handled
// via the /api/v1/doc/ API, so we just do a full page load of that document.

kuma/javascript/src/document-provider.jsx
```diff
@@ -80,9 +80,20 @@ export default function DocumentProvider(
...
+                    } else {
+                        let jsonLocaleAndSlug = `/${json.locale}/${json.slug}`;
```
davidflanagan: I'd rename this variable to receivedLocaleAndSlug

kuma/javascript/src/document-provider.jsx
```diff
@@ -80,9 +80,20 @@ export default function DocumentProvider(
...
+                        if (jsonLocaleAndSlug !== localeAndSlug) {
+                            // This was a redirect.
+                            let jsonURL = json.absoluteURL;
```
davidflanagan: I'd either rename this to receivedURL, or just drop the variable and use json.absoluteURL in the two places below.

requirements/docs.txt
```diff
@@ -3,6 +3,7 @@
 
 # These should stay synced with other requirements files
 Babel==2.2.0
+docutils==0.12
```
davidflanagan: I don't know why this is moved. Just checking that you intended to do this.

kuma/api/signal_handlers.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    A signal handler to publish the document to the document API after it
+    has been restored.
+    """
+    publish.delay([instance.pk])
+
+
+@receiver(render_done, sender=Document,
+          dispatch_uid='api.document.render_done.publish')
+def on_render_done(sender, instance, **kwargs):
```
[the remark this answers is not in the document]
escattone: When a document is edited, a new revision is created via the revision form, and when the revision form is saved, it kicks-off a document render (see https://github.com/mozilla/kuma/blob/78216c4a949fa75ff421d3b0a71bd8792c67f24d/kuma/wiki/forms.py#L838), so the writer does not have to do a Cmd-Shift-R to force a render/publish after editing a document.

requirements/docs.txt
escattone: Yes, I moved it in order to indicate that it is no longer a requirement unique to docs, since it's part of `constraints.txt` (it's needed by `boto3`), and needs to be kept in sync with that file.

kuma/api/management/commands/unpublish.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+"""
+Manually schedule the removal of one or more documents from the document API.
+"""
+from __future__ import division
```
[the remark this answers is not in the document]
escattone: Good catch, thanks! This is a leftover from earlier changes that I threw away, and is not needed either here or in the publish command.

kuma/api/management/commands/publish.py
```diff
@@ -0,0 +1,94 @@
...
+        parser.add_argument(
+            '--chunk-size',
+            type=int,
+            default=1000,
+            help='Partition the work into tasks, with each task handling this '
+                 'many documents (default=1000)')
+
+    def handle(self, *args, **options):
+        Logger = namedtuple('Logger', 'info, error')
+        log = Logger(info=self.stdout.write, error=self.stderr.write)
```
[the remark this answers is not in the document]
escattone: Thanks! I totally agree. My first thought when I saw `self.stdout.write` and `self.stderr.write` was that they didn't add newlines, but they do by default (although that can be suppressed).

kuma/api/tasks.py
```diff
@@ -0,0 +1,100 @@
...
+from kuma.core.utils import chunked
+from kuma.wiki.models import Document
+
+from .v1.views import document_api_data, get_content_based_redirect, get_s3_key
+
+
+S3_MAX_KEYS_PER_DELETE = 1000
+
+
+def get_s3_bucket():
```
peterbe: Every single call to `publish()` or `unpublish()` will trigger this function which will create a new HTTPS connection object I think. 
I haven't actually tested it yet but it should be possible to see if you set the boto3 logger to `DEBUG`.

Perhaps that's fine and might only matter those rare few times you manually run `./manage.py publish --all` which you're doing interactively anyway so you can just restart it if you get weird network connection errors.

kuma/api/tasks.py
```diff
@@ -0,0 +1,100 @@
...
+            }
+        )
+        for info in response.get('Deleted', ()):
+            log.info('Unpublished {}'.format(info['Key']))
+        for info in response.get('Errors', ()):
+            log.error('Unable to unpublish {}: ({}) {}'.format(
+                info['Key'], info['Code'], info['Message']))
+
+    if completion_message:
+        log.info(completion_message)
```
peterbe: Neat!

kuma/javascript/src/document-provider.jsx
```diff
@@ -67,7 +67,7 @@ export default function DocumentProvider(
         // This is the function that does client side navigation
         function navigate(url, localeAndSlug) {
             body.style.opacity = '0.15';
-            fetch(`/api/v1/doc${localeAndSlug}`)
+            fetch(`/api/v1/doc${localeAndSlug}`, { redirect: 'follow' })
```
peterbe: I don't think you need the `redirect: 'follow'`. It's done by default.
See https://codepen.io/peterbe/pen/pBxvYM

kuma/javascript/src/document-provider.jsx
```diff
@@ -80,9 +80,28 @@ export default function DocumentProvider(
...
+                        }`;
+                        if (receivedLocaleAndSlug !== localeAndSlug) {
+                            // This was a redirect.
+                            let receivedURL = json.absoluteURL;
+                            history.replaceState(
+                                { receivedURL, receivedLocaleAndSlug },
+                                '',
+                                receivedURL
+                            );
+                        }
```
peterbe: Neat!

peterbe:
What happens if the S3 PUT or the S3 DELETE fails for weird network reasons? Will it reattempt automatically (i.e. putting it back into the Celery queue)?

We should also consider putting an explicit timeout on the S3 boto3 commands. If S3 is acting up (or rather, our network to S3) is almost-dead, these queue tasks would build up intensively. A controlled timeout would prevent the tasks for exploding. Perhaps this is a premature worry.

docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '2.1'
 services:
   worker: &worker
     image: mdnwebdocs/kuma_base
-    command: ./manage.py celery worker --loglevel=INFO --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,celery
+    command: ./manage.py celery worker --loglevel=INFO --events --beat --autoreload --concurrency=4 -Q mdn_purgeable,mdn_search,mdn_emails,mdn_wiki,mdn_api,celery
```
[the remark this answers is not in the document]
escattone: No, the order doesn't matter.

kuma/api/management/commands/publish.py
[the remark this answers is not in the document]
escattone: The Django management commands recommend using `self.stdout.write` and `self.stderr.write` for console output, so my `log` is really just a console logger for the command.

kuma/api/signal_handlers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@receiver(restore_done, sender=Document,
+          dispatch_uid='api.document.restore_done.publish')
+def on_restore_done(sender, instance, **kwargs):
+    """
+    A signal handler to publish the document to the document API after it
+    has been restored.
+    """
+    publish.delay([instance.pk])
```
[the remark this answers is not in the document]
escattone: It wouldn't hurt, but since the `restore_document` view isn't wrapped in `transaction.atomic`, it wouldn't make any difference.

kuma/api/signal_handlers.py
```diff
@@ -0,0 +1,40 @@
...
+
+@receiver(render_done, sender=Document,
+          dispatch_uid='api.document.render_done.publish')
+def on_render_done(sender, instance, **kwargs):
+    """
+    A signal handler to publish the document to the document API after it
+    has been rendered.
+    """
+    if not instance.deleted:
+        publish.delay([instance.pk])
```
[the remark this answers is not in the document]
escattone: Same here, it wouldn't make any difference since we wouldn't have any transaction active.

kuma/api/signal_handlers.py
```diff
@@ -0,0 +1,40 @@
...
+    has been restored.
+    """
+    publish.delay([instance.pk])
+
+
+@receiver(render_done, sender=Document,
+          dispatch_uid='api.document.render_done.publish')
+def on_render_done(sender, instance, **kwargs):
+    """
+    A signal handler to publish the document to the document API after it
```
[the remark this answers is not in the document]
escattone: This is correct as is I think.

kuma/api/signal_handlers.py
```diff
@@ -0,0 +1,40 @@
...
+    has been rendered.
```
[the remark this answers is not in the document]
escattone: This is also correct as is I think.

kuma/javascript/src/document-provider.jsx
escattone: From MDN (https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) it seems like it's the default in Chrome, but maybe not in other browsers? Anyway, I thought it was safest to be explicit.

escattone:
@peterbe I like the idea of retrying the S3 `boto3` commands up to a limited number of times (maybe with exponential backoff) on connection and timeout errors. Or maybe that's something I can configure in boto3 itself? Anyway, something to consider in the near future.

escattone:
@peterbe I added some documentation on configuring S3 for local testing.

docker-compose.yml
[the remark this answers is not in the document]
peterbe: Why is it even there? Why not put these things in .py? 
Thing is, we're not running docker-compose.yml in Prod (Kubernetes) so putting it here means it needs to be put somewhere else too.

kuma/api/management/commands/publish.py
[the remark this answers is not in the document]
peterbe: Ah! Good point.

kuma/api/signal_handlers.py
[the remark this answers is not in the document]
peterbe: It should be :)
They should all be :)

kuma/api/signal_handlers.py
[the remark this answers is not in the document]
peterbe: Right you are!

escattone:
FYI, I've created https://github.com/mdn/infra/pull/236 as a companion to this PR (and edited my initial comment to include a link to it).

docker-compose.yml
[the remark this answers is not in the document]
escattone: Yes, it's been added to the Kubernetes world via https://github.com/mdn/infra/pull/236.

kuma/api/tasks.py
escattone: Sorry, I wanted to comment, but just realized that I hadn't yet! Yes, this is a great point, and worth looking into more! I'm not sure what boto3 does. I think the use of `boto3` in the line of code `boto3.resource('s3')` forces the use of the default session object. From some quick poking around on the web, that default session object may provide a connection pool of 10 connections (https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html), but I'm surprised that I can't find an obvious piece of boto3 documentation to confirm or deny that. I'm going to see if I can gather something from the `boto3` source code, but I've looked into the source code before and it's not simple to navigate and understand.

I did see that boto3 recommends a separate session for each thread (https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html#multithreading-multiprocessing), but since we're not running any threads (just processes) in our `gunicorn` set-up, I think we're OK there.

escattone: After some digging into the source code and running an experiment in the shell, I think I should be calling `boto3.resource('s3')` only once. Every time it's called, although it's reusing the default session, it's creating a new resource object with an underlying new client with its own connection pool of up to 10 connections? Frustrating how bad the documentation is on this point. I think I need to modify this code.

peterbe: Quick story; the reason this stuff makes me nervous was because when I built symbols.mozilla.org it relies on doing a LOT of HEAD requests to S3. And it's perfectly normal, and common, that the S3 key you're asking for does NOT exist. Like 9 out of 10 times, we expect it to not exist. 
So doing `mybucket.head_object({'Key': key})` would ultimately trigger a python exception in boto3. So, I did something like this:

```python
try:
    mybucket.head_object({'Key': key})
    return "Yes, it exists!"
except botocore.exceptions.ClientError as exception:
    if exception.response["Error"]["Code"] == "NoSuchKey":
        return "Bummer, it does not exist"
    raise  # Any other unexpected exceptions
```

Makes sense, right? 
But when starting all of this up in Staging for the first time I noticed the Django server performance sucked. It could only handle a couple of requests per second!
I turned over every stone and started to suspect everything and anything. 

It was not until I upped the logging level that boto3 uses to DEBUG and that I actually started studying all the verbose logging. 

What was happening was that ANY Python exception would cause the connection pool to drain. Even if I, in a higher level, handled the exception. 
So every time it happened, all the connections were lost and the whole HTTP connection (including initial SSL negotiation) had to start over. It was fast (<1 seconds) but the service had to handle ~100 requests/second and it bottlednecked on boto3 creating brand new HTTP connections way too often.

By the way, the solution was crazy but worked. Instead of `mybucket.head_object()` I use:
```
for found in mybucket.list_objects_v2({'Prefix': key}):
    return "Yes it exists!"
return "Bummer it does not exist"
```
I.e. doing a prefix search with the prefix being the whole key name.

docs/development.rst
```diff
@@ -491,3 +491,29 @@ value you can edit your ``.env`` file. For example::
...
+the detailed skip messages in the ``worker`` log (
+``docker-compose logs -f worker``).
+
+However, if for testing purposes you'd like
+to locally configure the ``publish`` and ``unpublish`` tasks to use S3, you can
+simply add the following::
+
+    - MDN_API_S3_BUCKET_NAME=<your-s3-bucket-name>
+    - AWS_ACCESS_KEY_ID=<your-aws-access-key>
+    - AWS_SECRET_ACCESS_KEY=<your-aws-secret-key>
```
peterbe: Before I forget, r- on this.

Like we discussed, it should be possible to switch S3 publishing on and off from the .env in local dev. I don't think it affects Stage and Prod. Just a matter of local dev convenience. 
I think all of us core devs should always be publishing to S3 most days to help find bugs. 

(Just being verbose about this so we don't forget)

peterbe: With that said, *I* am using my *personal* AWS account. And it makes it easy that I don't see David's or Schalks file based on my MySQL on this laptop. 

How about we create some credentials and buckets for us guys and contributors can either A) use their personal AWS accounts or B) disable S3 publishing.

docker-compose.yml
```diff
@@ -17,6 +17,8 @@ services:
       - ALLOW_ALL_IFRAMES=${ALLOW_ALL_IFRAMES:-False}
       - ALLOWED_HOSTS=*
       - ATTACHMENT_HOST=${ATTACHMENT_HOST:-localhost:8000}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-}
```
peterbe: It appears, unless I've gone crazy that `docker-compose.yml` will automagically read this file. So this line (and line above and line 33) are not needed.

docs/development.rst
```diff
@@ -491,3 +491,25 @@ value you can edit your ``.env`` file. For example::
...
+the detailed skip messages in the ``worker`` log (
+``docker-compose logs -f worker``).
+
+However, if for testing purposes you'd like to locally configure the
+``publish`` and ``unpublish`` tasks to use S3, you can simply add the
+following to your ``.env`` file::
+
+    MDN_API_S3_BUCKET_NAME=<your-s3-bucket-name>
+    AWS_ACCESS_KEY_ID=<your-aws-access-key>
+    AWS_SECRET_ACCESS_KEY=<your-aws-secret-key>
```
peterbe: 👍 

In an ideal world we would have, and keep updating, instructions for how to use AWS S3 Console to create an S3 bucket with website config so that writes are protected but things are all public by default. 

For now I'm r+ on the documentation side.

kuma/api/signal_handlers.py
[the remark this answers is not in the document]
peterbe: Actually, from what I can read if you do:

```python
@transaction.atomic
def view(request):
    transaction.on_commit(only_if_successful)
    return response
```
It *will* call `only_if_successful` only if the transaction is successful. But if you change it to this:

```python
# @transaction.atomic   COMMENTED OUT
def view(request):
    transaction.on_commit(only_if_successful)
    return response
```

then, it will act exactly the same as this:
```python
def view(request):
    only_if_successful()
    return response
```

I tested it with a Django 2.2 I had inside a side-project. One of my sample views looked like this:

```python
def log_transaction(line):
    with open('transactions.log', 'a') as f:
        f.write(line)
        f.write('\n')
    print("LOG TRANSACTION", time.time())

def view3(request):
    transaction.on_commit(lambda: log_transaction('view3'))
    sleep(2.1)
    print("In View 3")
    print(time.time())
    raise http.Http404
```
The output on stdout on this became:
```
LOG TRANSACTION 1556300548.503019
In View 3
1556300550.608419
```
which confirms what I mentioned above. Using `transaction.on_commit`, when not inside an atomic context, will work as it's not there. But once the `@transaction.atomic` is put in place, you get all sorts of juicy benefits such as NOT doing a S3 PUT if there's any Python exception in any of the other signals.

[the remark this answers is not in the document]
peterbe: But here, my suggestion does add risk.

kuma/javascript/src/document-provider.jsx
peterbe: I ran my Codepen above in Firefox Nightly and it implied that `{ redirect: 'follow' }` was *not* needed. I also opened that codepen in Safari and it behaves the same. 

But perhaps best to leave it in case there was an initial time when it was *not default* and that might get stuck in some ancient browsers.

escattone:
Thanks for the great reviews @peterbe and @davidflanagan!

docs/development.rst
escattone: That's a really good point. I think would simply be the terraform that @limed will create in mdn/infra for creating the stage and prod S3 buckets with the proper settings.

PR #5362, opened 2019-04-20
Bug 1525719: enable server-side rendering of the React UI

davidflanagan:
This PR enables server-side rendering of the React UI for pages
on the beta domain with these changes:

- In kuma/wiki/jinja2/wiki/react_document.html we use a new
  template tag {{render_react_app()}} to insert the rendered HTML

- We define that tag in kuma/wiki/templatetags/ssr.py

- The ssr.py file JSON serializes the document data and POSTs it
  to a new ssr server which returns a string of rendered HTML.
  If the HTTP connection fails or times out, ssr.py falls back
  on client side rendering.

- This new ssr server is defined in kuma/javascript/ssr-server.js
  It uses the Express package as its framework and calls
  kuma/javascript/dist/ssr.js to actually do the server-side
  rendering.

- The ssr.js file is a webpack bundle created from
  kuma/javascript/src/ssr.jsx. There are changes to the webpack
  config file to enable the creation of ssr.js

- docker-compose.yml is modified to define a new 'ssr' container
  (using the same image that the 'web' container uses, for
  simplicity). This container is what ssr-server.js runs in.

- The existing file kuma/javascript/src/index.jsx is modified to
  do a hydrate if server side rendering is done and to do a render
  if server side rendering was not done. It is also modified to get
  the necessary state data from window._document_data instead of
  doing the base64 trick that we used previously.

The PR also includes miscellaneous changes in a few existing React
components to make them compatible with SSR. There were a few places
where the components assumed that they could reference the window.mdn
or window.location when rendering themselves. That doesn't work for
SSR, since the window object is not defined in that case.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,10 @@ services:
     volumes:
       - ./:/app:z
     depends_on:
+      - mysql
+      - elasticsearch
       - redis
+      - kumascript
     environment:
       # Django settings overrides:
       - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,10 +7,7 @@ services:
     volumes:
       - ./:/app:z
     depends_on:
-      - mysql
-      - elasticsearch
       - redis
-      - kumascript
     environment:
       # Django settings overrides:
       - ACCOUNT_DEFAULT_HTTP_PROTOCOL=http
@@ -40,11 +37,21 @@ services:
       - PYTHONWARNINGS=${PYTHONWARNINGS:-default}
       - MAINTENANCE_MODE=${MAINTENANCE_MODE:-False}
       - REVISION_HASH=${KUMA_REVISION_HASH:-undefined}
+      # For the server side rendering server and the code that connects to it.
+      - SSR_PORT=8000
+      - SSR_URL=http://ssr:8000/ssr
+      - SSR_TIMEOUT=1
 
   # Web is based on worker b/c you cannot clear the "ports" with docker-compose.
   web:
     <<: *worker
     command: gunicorn -w 4 --bind 0.0.0.0:8000 --access-logfile=- --timeout=120 --worker-class=meinheld.gmeinheld.MeinheldWorker kuma.wsgi:application
+    depends_on:
+      - mysql
+      - elasticsearch
+      - redis
+      - kumascript
+      - ssr
     ports:
       - "8000:8000"
 
@@ -56,9 +63,16 @@ services:
       - mysql
       - elasticsearch
       - redis
-      # Drop kumascript link
     ports:
       - "8001:8000"
+
+  # ssr is a Node server that performs server-side rendering of our React UI
+  ssr:
+    <<: *worker
+    command: node kuma/javascript/ssr-server.js
+    depends_on: []
+    ports:
+      - "8002:8000"
 
   mysql:
     image: mysql:5.6
--- a/kuma/javascript/src/header/header.jsx
+++ b/kuma/javascript/src/header/header.jsx
@@ -132,9 +132,9 @@ const menus = [
             {
                 label: gettext('Report a content problem'),
                 external: true,
-                url: `https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title=${encodeURIComponent(
-                    window.location
-                )}`
+                // See fixurl() for code that replaces the {{SLUG}}
+                url:
+                    'https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title={{SLUG}}'
             },
             {
                 label: gettext('Report a bug'),
@@ -150,12 +150,16 @@ export default function Header(): React.Node {
     if (!documentData) {
         return null;
     }
-    const { localeFromURL } = documentData;
+    const { localeFromURL, slug } = documentData;
 
     function fixurl(url) {
-        return url.startsWith('https://')
-            ? url
-            : `/${localeFromURL}/docs/${url}`;
+        // The "Report a content issue" menu item has a link that requires
+        // the document slug, so we work that in here.
+        url = url.replace('{{SLUG}}', encodeURIComponent(slug));
+        if (!url.startsWith('https://')) {
+            url = `/${localeFromURL}/docs/${url}`;
+        }
+        return url;
     }
 
     return (
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -9,18 +9,34 @@ import Page from './page.jsx';
 let container = document.getElementById('react-container');
 
 if (container) {
-    let script = container.firstElementChild;
-    if (script && script instanceof HTMLScriptElement) {
-        // We expect the script to contain a base64-encoded JSON blob
-        // that contains all the content of this document
-        let data = JSON.parse(atob(script.text));
-        ReactDOM.render(
-            <DocumentProvider initialDocumentData={data}>
-                <CurrentUser.Provider>
-                    <Page />
-                </CurrentUser.Provider>
-            </DocumentProvider>,
-            container
-        );
+    // The HTML page that loads this code is expected to have an inline
+    // script that sets this window._document_data property to an object
+    // with all the data needed to hydrate or render the UI.
+    let data = window._document_data;
+
+    // Remove the global reference to this data object so that it can
+    // be garbage collected once it is no longer in use.
+    window._document_data = null; // eslint-disable-line camelcase
+
+    // This is the React UI for a page of documentation
+    let page = (
+        <DocumentProvider initialDocumentData={data}>
+            <CurrentUser.Provider>
+                <Page />
+            </CurrentUser.Provider>
+        </DocumentProvider>
+    );
+
+    if (container.firstElementChild) {
+        // If the container element is not empty, then it was presumably
+        // rendered on the server, and we just need to hydrate it now.
+        ReactDOM.hydrate(page, container);
+    } else {
+        // Otherwise, if the container is empty, then we need to do a full
+        // client-side render. The goal is that pages should always be
+        // server-side rendered when first loaded (for speed and SEO). But
+        // this is here for robustness in case there are errors during
+        // server side rendering.
+        ReactDOM.render(page, container);
     }
 }
--- a/kuma/javascript/src/ssr.jsx
+++ b/kuma/javascript/src/ssr.jsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+
+import CurrentUser from './current-user.jsx';
+import DocumentProvider from './document-provider.jsx';
+import Page from './page.jsx';
+
+/*
+ * This function performs server-side rendering of our UI, given
+ * a JSON object of document data. It is used by ../ssr-server.js
+ */
+export default function ssr(data) {
+    return renderToString(
+        <DocumentProvider initialDocumentData={data}>
+            <CurrentUser.Provider>
+                <Page />
+            </CurrentUser.Provider>
+        </DocumentProvider>
+    );
+}
--- a/kuma/javascript/ssr-server.js
+++ b/kuma/javascript/ssr-server.js
@@ -1,0 +1,96 @@
+/* eslint-disable no-console */
+/*
+ * This is a simple Express server for performing server side rendering
+ * of our React UI. The main endpoint is POST /ssr, and it invokes the
+ * ssr() function from dist/ssr.js (which is webpacked from src/ssr.jsx).
+ */
+const express = require('express'); // Express server framework
+const morgan = require('morgan');
+
+const ssr = require('./dist/ssr.js'); // Function to do server-side rendering
+
+// Configuration
+const PID = process.pid;
+const PORT = parseInt(process.env.SSR_PORT) || 8000;
+const MAX_BODY_SIZE = 1024 * 1024; // 1 megabyte max JSON payload
+
+// Start New Relic logging if it is configured
+if (process.env.NEW_RELIC_LICENSE_KEY && process.env.NEW_RELIC_APP_NAME) {
+    console.log('Starting New Relic logging for KumaScript.');
+    require('newrelic');
+}
+
+// We're using the Express server framework
+const app = express();
+
+// Log all requests, so we get timing data for SSR.
+app.use(morgan('tiny'));
+
+// Handle JSON payloads in POST requests, putting data in req.body
+app.use(express.json({ limit: MAX_BODY_SIZE }));
+
+// A hello world endpoint for easy verification that the service is running
+app.get('/', (req, res) => {
+    res.send('<html><body><p>SSR server ready</p></body></html>');
+});
+
+// revision, health and readiness endpoints used (I think) by Kubernetes
+app.get('/revision/?', (req, res) => {
+    res.set({ 'Content-Type': 'text/plain; charset=utf-8' }).send(
+        process.env.REVISION_HASH || 'undefined'
+    );
+});
+app.get('/healthz/?', (req, res) => {
+    res.sendStatus(204);
+});
+app.get('/readiness/?', (req, res) => {
+    res.sendStatus(204);
+});
+
+/*
+ * This is the main endpoint. It expects document API data as application/json
+ * in the POST request body and returns rendered HTML in the response body.
+ * The response is an HTML fragment, not a complete document and it is not
+ * intended for display on its own. The content type of the response is
+ * text/plain instead of text/html to make it clear that the response should
+ * not be parsed, escaped, or displayed as HTML.
+ */
+app.post('/ssr/?', (req, res) => {
+    res.set({ 'Content-Type': 'text/plain; charset=utf-8' }).send(
+        ssr(req.body)
+    );
+});
+
+if (require.main === module) {
+    // If we're actually being run directly with node, then
+    // set up signal handlers and start listening for connections
+
+    // More gracefully handle some common exit conditions...
+    const exit = function() {
+        console.log(`SSR server (PID ${PID}) exiting.`);
+        server.close();
+        process.exit(0);
+    };
+    process.on('SIGINT', function() {
+        console.log('Received SIGINT, exiting...');
+        exit();
+    });
+    process.on('SIGTERM', function() {
+        console.log('Received SIGTERM, exiting...');
+        exit();
+    });
+    process.on('uncaughtException', function(err) {
+        console.error('uncaughtException:', err.message);
+        console.error(err.stack);
+        exit();
+    });
+
+    // And finally, start listening for connections.
+    const server = app.listen(PORT, () => {
+        console.log(`SSR server (PID ${PID}) listening on port ${PORT}.`);
+    });
+} else {
+    // If we've just been required (in a test, for example) then
+    // we just export the app object and don't actually start listening.
+    module.exports = app;
+}
--- a/kuma/wiki/templatetags/ssr.py
+++ b/kuma/wiki/templatetags/ssr.py
@@ -1,0 +1,95 @@
+from __future__ import print_function
+
+import json
+
+import requests
+import requests.exceptions
+from django.conf import settings
+from django_jinja import library
+
+
+@library.global_function
+def render_react_app(data, ssr=True):
+    """
+    Render a script tag to define the data and any other HTML tags needed
+    to enable the display of a React-based UI. By default, this does
+    server side rendering, falling back to client-side rendering if
+    the SSR attempt fails. Pass False as the second argument to do
+    client-side rendering unconditionally.
+
+    Note that we are not defining a generic Jinja template tag here.
+    The code in this file is specific to Kuma's React-based UI.
+    """
+    if ssr:
+        return server_side_render(data)
+    else:
+        return client_side_render(data)
+
+
+def _render(html, state):
+    """A utility function used by both client side and server side rendering.
+    Returns a string that includes the specified HTML and a serialized
+    form of the state dict, in the format expected by the client-side code
+    in kuma/javascript/src/index.jsx.
+    """
+    # Serialize the state object to JSON and be sure the string
+    # "</script>" does not appear in it, since we are going to embed it
+    # within an HTML <script> tag.
+    serializedState = json.dumps(state).replace('</', '<\\/')
+
+    # Now return the HTML and the state as a single string
+    return (
+        u'<div id="react-container">{}</div>\n'
+        u'<script>window._document_data = {};</script>\n'
+    ).format(html, serializedState)
+
+
+def client_side_render(data):
+    """
+    Output an empty <div> and a script with complete state so that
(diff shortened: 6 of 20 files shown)


davidflanagan:
Ryan: I'm expecting that you'll probably the the primary reviewer for this.

Schalk and Peter: I thought you both might be interested in this patch, so I tagged you as reviewers. Feel free to leave review comments if you want, but don't feel that you have to do a review.

Note that I have not added any tests, and I probably should. But I haven't tried yet to figure out how to write tests for template tags or how to test the inter-process communication I've got between ssr.py and ssr.jsx. I'll try to add something before merging this, but they might not be super robust tests given the unusual nature of the code to be tested.

peterbe:
> haven't tried yet to figure out how to write tests for template tags 

More than happy to help if it helps. Or perhaps you just need some peace and quiet :)

peterbe:
Is there a reason the "Contributors to this page:" isn't working? It's empty when using the React version:
Screen Shot 2019-04-22 at 1 15 14 PM

But the JSON does contain it: http://beta.mdn.localhost:8000/api/v1/doc/en-US/Learn/CSS/Styling_text

Screen Shot 2019-04-22 at 1 16 49 PM

davidflanagan:
Hmm. Contributors is working for the SSR version of a page, but subsequent client-side navigation is not getting the contributors. I suspect that Ryan's patch (which does something with the contributors job) will fix this. But I'll look into it.

Separately, Peter has pointed out that the links in the header menus aren't working right. It seems that I broke the fixurl() method in this patch and will need to fix that.

davidflanagan:
I just pushed a commit to fix the broken menus issue. I think the contributors thing is working correctly on the front end but that there is flakiness on the backend, and I think that Ryan has a fix for that in his pending patch.

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,123 @@
...
+        # If we haven't started a renderer process, or if that process
+        # exited for some reason, then start one.
+        if not renderer or renderer.poll() is not None:
+            print "Starting renderer subprocess"
+            renderer = subprocess.Popen(['/usr/bin/env',
+                                         'node',
+                                         '/app/kuma/javascript/dist/ssr.js'],
+                                        stdin=subprocess.PIPE,
+                                        stdout=subprocess.PIPE,
+                                        stderr=sys.stderr)
```
davidflanagan: I'm working on a test for ssr.py, and realize that I'm going to move the code about this into a separate get_renderer() function, so that I can mock it in the tests.

peterbe:
I don't understand why redirects don't work. 

On `http://beta.mdn.localhost:8000/en-US/docs/Web/Reference` I have, in the rendered HTML a link like this: 
```html
<a href="/en-US/Learn/HTML/Introduction_to_HTML">Introduction to HTML</a>
```

I don't know why it's like that but it's another story for another day. By the way, it's actually the same wrong link here: https://developer.mozilla.org/en-US/docs/Web/Reference 
Clicking `https://developer.mozilla.org/en-US/Learn/HTML/Introduction_to_HTML` will redirect to `https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML`. (injecting the missing `/docs`)

However, with the SSR, opening:
`curl -v http://beta.mdn.localhost:8000/en-US/Learn/HTML/Introduction_to_HTML` will cause a 404.
But `curl -v http://wiki.mdn.localhost:8000/en-US/Learn/HTML/Introduction_to_HTML` will correctly redirect to `/en-US/docs/Learn/HTML/Introduction_to_HTML`.

I haven't yet figured out where non-`/docs` url patterns fit in but we ought to have the same functionality for beta as we do for wiki.

kuma/wiki/views/document.py
```diff
@@ -875,8 +874,8 @@ def react_document(request, document_slug, document_locale):
 
     # Bundle it all up and, finally, return.
     context = {
-        'document_api_data': base64.b64encode(
-            json.dumps(document_api_data(doc))),
+        'document_api_data': document_api_data(doc),
```
peterbe: Yay!

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,128 @@
...
+    Output an empty <div> and a script with complete state so that
+    the UI can be rendered on the client-side.
+    """
+    # Convert to JSON and be sure the string "</script>" does not appear
+    serializedState = json.dumps(data).replace('</', '<\\/')
+    return OUTPUT_TEMPLATE.format('', serializedState)
+
+
+def _get_renderer():
+    global renderer
```
peterbe: All of this is making me nervous. 

First of all, it's a pattern I'm just not very familiar with. Secondly, this pattern of using `stdin` and `stdout` is blocking and I don't know what that means if individual Gunicorn (Meinheld) worker use threads. 
Thirdly, perhaps I'm barking up the wrong tree and that all this subprocess stuff will be replaced by a Node HTTP server (and `requests.get()` on the Python side)  before any of this goes into Prod.

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,128 @@
+import json
+import subprocess
```
peterbe: The `subprocess` in stdlib Python 2.7 does not support timeouts.
Having done a lot of Python-to-Node subprocess stuff, it's only a matter of time until you get a dead process and Python ties up a blocked and neverdying process waiting for it.

Again, I don't know if the use of subprocess is a temporary hack but if we're going to land this in master we might as well switch to [`subprocess32`](https://pypi.org/project/subprocess32/) and use `import subprocess32 as subprocess`.

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,128 @@
...
+
+        return OUTPUT_TEMPLATE.format(renderedHTML, serializedState)
+
+    except Exception as e:
+        # If server side rendering fails, fall back to client-side
+        # Is there a better way to log this?
+        print "SSR failed after {}s".format(time.time() - start)
+        print e
+        print "Falling back to client side rendering..."
+        return client_side_render(data)
```
peterbe: Actually, I think we should eliminate this. `except Exception` is generally a code-smell because it's too big and too broad. Again, I don't know how "serious" this code is on a scale of prototype vs. production ready. 

*if* any exception happens on the 49 lines of code (that it covers) I'd rather we let it happen and learn from it. It should end up in Sentry and since we have weeks/months until this needs to go into Prod proper, we can fail forward by patching unexpected errors deliberately. If anything, we could refactor it to look like this:


```python
renderer = _get_renderer()
renderer.stdin.write(json.dumps(data))
...

# The reading from stdout the renderer can sometimes trigger bla bla yada yada
try:
    while True:
       ...reading from stdout...
except SomeSpecificSubprocessError as exception:
    return client_side_data(data, exception=exception)

...continue with the rest...
```

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,128 @@
...
+                break
+            renderedHTML += line
+
+        # Even though we've got fully rendered HTML now, we still need
+        # to send the document data along with it so that React can sync
+        # its state on the client side with what is in the HTML.
+        # Fortunately, however, it turns out not to be necessary to
+        # duplicate the biggest parts (the HTML strings) of that data,
+        # so we can delete those from the data now
+        partial = data.copy()
```
peterbe: The, great, code comment above is written as if it's for this `partial = data.copy()` line. I'd rather see something like this:
```python
# Even though we've got fully rendered HTML now, we still need
# to send the document data along with it so that React can sync
# its state on the client side with what is in the HTML.
# Fortunately, however, it turns out not to be necessary to
# duplicate the biggest parts (the HTML strings) of that data,
# so we can delete those from the data now

# We need to make a shallow clone of the data because we ...
partial = data.copy()
```

Also, *why* do you make a shallow clone/copy of it? I don't see it being used by anything else later.

kuma/javascript/src/header/header.jsx
```diff
@@ -150,12 +150,15 @@ export default function Header(): React.Node {
...
     function fixurl(url) {
-        return url.startsWith('https://')
-            ? url
-            : `/${localeFromURL}/docs/${url}`;
+        // The "Report a content issue" menu item has a link that requires
+        // the document slug, so we work that in here.
+        url = url.replace('{{SLUG}}', encodeURIComponent(slug));
+        if (url.startsWith('https://')) {
+            url = `/${localeFromURL}/docs/${url}`;
+        }
```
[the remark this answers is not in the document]
davidflanagan: Fixed now. The https thing is confusing because I forgot the !.  That code path is meant for for relative URLs that are not hardcoded with https!

peterbe:
One other thing that is on my mind, and again I'm outside my comfort zone here: emotion-css. 

I extract the generated HTML to a file a disk and pretty printed it: https://gist.github.com/peterbe/9d41ea10ee0ff75b81da1dd0d2b2b6f1

It's got 24 `<style>` tags. My first question is: Don't we already have all of this CSS in the compiled SCSS files?
Secondly, what are the performance (rendering speed) implications of the browser having to stop, context switch to CSS rendering, updating the CSSOM, and only then carry on with the rest of the HTML. Then rinse and repeat that 24 times. 

Usually, a browser will use a single thread to start parsing the HTML, read until the `</head>` and then it can run one thread to parse the DOM and one to parse and process the CSSOM, concurrently. As soon as the DOM parsing encounters a `<style>` tag I think it needs to pause the DOM parsing to somehow tell the CSSOM parsing thread to stop and take note.

kuma/javascript/src/index.jsx
```diff
@@ -9,18 +9,31 @@ import Page from './page.jsx';
...
-                    <Page />
-                </CurrentUser.Provider>
-            </DocumentProvider>,
-            container
-        );
+    // The HTML page that loads this code is expected to have an inline
+    // script that sets this window._document_data property to an object
+    // with all the data needed to hydrate or render the UI.
+    let data = window._document_data;
+    window._document_data = null; // eslint-disable-line camelcase
```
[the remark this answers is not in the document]
davidflanagan: This object is only used for the initial page load, and is completely useless after a client side navigation. If we don't erase it from the global scope here, it is a memory leak.

I'll add a note.

kuma/javascript/src/index.jsx
```diff
@@ -9,18 +9,31 @@ import Page from './page.jsx';
...
+    );
+
+    if (container.firstElementChild) {
+        // If the container element is not empty, then it was presumably
+        // rendered on the server, and we just need to hydrate it now.
+        ReactDOM.hydrate(page, container);
+    } else {
+        // Otherwise, if the container is empty, then we need to do a full
+        // client-side render. The goal is that pages should always be
+        // server-side rendered when first loaded (for SEO reasons). But
```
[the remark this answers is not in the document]
davidflanagan: Done.

kuma/wiki/templatetags/ssr.py
davidflanagan: I didn't know about subprocess32. If that enables me to add a timeouts for the reads and writes, that would be great.

I don't know if this is a temporary hack. After Ryan has had a chance to review, maybe we can all get together again to discuss how to move forward with this and how to integrate it with Ryan's JSON publishing work.

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,128 @@
...
+        return client_side_render(data)
+
+
+def client_side_render(data):
+    """
+    Output an empty <div> and a script with complete state so that
+    the UI can be rendered on the client-side.
+    """
+    # Convert to JSON and be sure the string "</script>" does not appear
+    serializedState = json.dumps(data).replace('</', '<\\/')
```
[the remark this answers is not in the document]
davidflanagan: I'm happy to have some backup on this. It seems hacky, but better than doing the full base64 encode like I was doing before.  I thought I could use a CDATA section, but that wasn't working for me, so maybe it was an XHTML thing that has been dropped...

kuma/wiki/templatetags/ssr.py
davidflanagan: I can't say I'm not nervous about this too.

1) Are you nervous about the global variable? Or the fact that I'm using a long-lived Node process? Would you be less nervous if I fired up the node process to render a single file and then it exited each time? There would be a lot more overhead, but maybe less risk?

2) Isn't blocking the norm in python code? Aren't we already blocking all over the place in Kuma? What am I missing here?

3) I'm not qualified to change our production infrastructure to introduce a Node-based server into the mix. If we decide to do it that way, I'd be happy to write the server, though!

kuma/wiki/templatetags/ssr.py
```diff
@@ -0,0 +1,123 @@
...
+        partial['bodyHTML'] = ''
+        partial['tocHTML'] = ''
+        partial['quickLinksHTML'] = ''
+        # Convert to JSON and be sure the string "</script>" does not appear
+        serializedState = json.dumps(partial).replace('</', '<\\/')
+
+        # Is there a better way to write log messages from a template tag?
+        # On my dev server, rendering takes 30-50ms once the server is
+        # already running and 150-200ms when the server needs to start
+        print "Server side rendered in {}s".format(time.time() - start)
```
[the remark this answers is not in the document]
davidflanagan: done

kuma/wiki/templatetags/ssr.py
davidflanagan: You make a good point. I'll wait and see what Ryan things and whether we decide to move forward with this approach before deciding whether to remove or keep  the except clause.

If I'm able to use childprocess32 to apply timeouts to the read and write calls, then I think it might be legit to catch those. And probably kill the child process if it happens.

Is there a way we can handle an exception *and* report it to Sentry?

kuma/wiki/templatetags/ssr.py
davidflanagan: The data object is also used in the exception handling code that falls back to client-side rendering, so I'm making a copy in case there is an exception below this point. (Which admittedly is not very likely).

davidflanagan: I'll update the comment

davidflanagan: Actually, a better reason for making the copy is that the data dictionary belongs to the caller of this template tag. If we change it here, that change will be seen by anything else in the template that uses it.

kuma/wiki/views/document.py
escattone: Later, when we want to pre-render the SSR files to S3, I suppose we'll get the document API data directly from the pre-rendered data on S3 (e.g., `https://<s3-website-url>/api/v1/doc/<locale>/<slug>`)?

kuma/javascript/src/index.jsx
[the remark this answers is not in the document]
peterbe: I can see the benefit of changing its value so the GC, in the browser, can remove this and run at a lesser total memory footprint. But why is it a leak? Or, are you saying it *might become* a leak.

kuma/wiki/templatetags/ssr.py
peterbe: It's a deal! I'll retract everything I said about `ssr.py` because I don't any of my concerns would remain valid. Having said that I'd rather not delete the comments. We'll just squint our eyes and ignore them being there because they shouldn't block a merge.

kuma/wiki/templatetags/ssr.py
peterbe: You're not wrong but it's a cost. A memory cost. And it costs a little in CPU too because of the actual copying of the bytes. 
I think I'm just trying to say; before you commit to this, have you asked yourself if it's worth the cost?

peterbe:
> As for pages like http://beta.mdn.localhost:8000/dashboards, they aren't currently supported in the beta domain.

Yeah, I can see that. That kinda stuff truly belongs to `wiki.developer.mozilla.org` but since our React SSR solution will take over the `developer.mozilla.org` domain we're going to need a redirect *somewhere* that takes care of non-document URLs like `https://developer.mozilla.org/dashboards`.

I think this somewhat relates to my concern that if you click on a link that doesn't *yet* exist, it shouldn't throw you a hard 404 but instead encourage you to write that new document. That's also going to need a smarter redirect from SSR document to Wiki dynamic page.

davidflanagan:
@escattone and @peterbe I've pushed a commit that changes this to use its own docker container.

To review, you'll probably want to look at:

- docker-compose.yml for the setup of the new ssr container

- kuma/wiki/templatetags/ssr.py for the code that uses the server

- kuma/javascript/ssr-server.jsx for the express server

- kuma/javascript/src/ssr.jsx for the simple ssr() function that the server uses to convert document data to html text

- tests in kuma/wiki/tests/test_ssr.py and kuma/javascript/ssr-server.test.js

There are also some updated snapshots in the javascript tests. I think those broke in some previous commit and I forgot to update them.

kuma/javascript/src/ssr.jsx
```diff
@@ -0,0 +1,20 @@
...
+ * a JSON object of document data. It is used by ../ssr-server.js
+ */
+export default function ssr(data) {
+    return renderToString(
+        <DocumentProvider initialDocumentData={data}>
+            <CurrentUser.Provider>
+                <Page />
+            </CurrentUser.Provider>
+        </DocumentProvider>
+    );
```
peterbe: Is there any chance you can merge these file and index.jsx to avoid the DRY?
Today they're both pretty simple but might outgrow itself and someone might refactor `Page` to become `DocumentPage` and get confused that they had to edit one other file too.

kuma/wiki/templatetags/ssr.py
davidflanagan: It is a trivial cost because this is a shallow copy of one dict. Not making a copy here would be software engineering malpractice because it would be a stealth change to a data structure that might be used elsewhere in the template that calls this template tag.

kuma/javascript/src/ssr.jsx
davidflanagan: Yes, I've been meaning to create a top-level App component that just renders this DocumentProvider/CurrentUser/Page thing. But I didn't want to complicate this PR by doing it now.

peterbe: Fine by me. Perhaps put one of those `// TODO: refactor this bla bla bla` :)

kuma/wiki/templatetags/ssr.py
peterbe: My job is done here I think. I just wanted to force a pause-and-think. 

When I worked on Socorro we fought this hard. A Breakpad crash is read into a Python dict and then it's passed through a long list of "processor classes" that would all get a chance to evaluate and potentially mutate the dict. Or trigger side-effects. We noticed some nasty bugs because of those processor classes needed to *delete* things from the dict, right before storing in Elasticsearch, otherwise it would be too fat on Elasticsearch's disk. So we did `copy.deepcopy(crashobj)` (or whatever it is called). Everything worked locally and in unit tests and in Stage. But then hell broke loose in Prod one day when this change deployed. It blew up because these dicts were big and the deep clone way too expensive. 

We had to go back to the drawing board quick and rewrite it into a more complicated hybrid solution. 

Yeah, deciding between mutation and "immutation" is never a slam dunk.

docker-compose.yml
```diff
@@ -55,10 +60,16 @@ services:
...
-      # Drop kumascript link
     ports:
       - "8001:8000"
 
+  # ssr is a Node server that performs server-side rendering of our React UI
+  ssr:
+    <<: *worker
+    command: node kuma/javascript/ssr-server.js
+    ports:
+      - "8002:8000"
```
escattone: If we go with my suggestion above, we'd have to explicitly state that `ssr` has no dependencies, like this:
```
depends_on: []
```

kuma/javascript/ssr-server.js
```diff
@@ -0,0 +1,90 @@
...
+// Handle JSON payloads in POST requests, putting data in req.body
+app.use(express.json({ limit: MAX_BODY_SIZE }));
+
+// A hello world endpoint for easy verification that the service is running
+app.get('/', (req, res) => {
+    res.send('<html><body><p>SSR server ready</p></body></html>');
+});
+
+// revision, health and readiness endpoints used (I think) by Kubernetes
+app.get('/revision/?', (req, res) => {
```
escattone: I don't think this endpoint is needed, but it doesn't hurt, so let's keep it, at least for now.

kuma/javascript/ssr-server.js
```diff
@@ -0,0 +1,90 @@
...
+    res.sendStatus(204);
+});
+app.get('/readiness/?', (req, res) => {
+    res.sendStatus(204);
+});
+
+// This is the main endpoint that receives document api data in the
+// request body and returns rendered HTML in the response
+app.post('/ssr/?', (req, res) => {
+    res.set({ 'Content-Type': 'text/plain; charset=utf-8' }).send(
```
escattone: It probably doesn't matter, but I'm curious why `text/plain` rather than `text/html`? Is it because it's not a full HTML document, but just a fragment?

kuma/javascript/ssr-server.js
```diff
@@ -0,0 +1,90 @@
...
+        ssr(req.body)
```
escattone: Nice and simple (when used with `app.use(express.json(...`)!

kuma/wiki/tests/test_ssr.py
```diff
@@ -3,22 +3,25 @@
 
 import mock
 import pytest
+import requests.exceptions
+from django.conf import settings
```
escattone: Just FYI, you could use the `settings` fixture provided by `pytest-django` (https://pytest-django.readthedocs.io/en/latest/helpers.html#settings) in your tests below, but since you're not modifying any settings, it's fine to use this.

docker-compose.yml
```diff
@@ -6,11 +6,6 @@ services:
     user: ${UID:-1000}
     volumes:
       - ./:/app:z
-    depends_on:
-      - mysql
-      - elasticsearch
-      - redis
-      - kumascript
```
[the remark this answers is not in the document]
davidflanagan: I forgot that there were actual worker instances that got started! I was thinking that the worker definition here was just to be overrridden below.

Fixed to add just redis. But let me know if you want me to just put it back and list them all.

docker-compose.yml
davidflanagan: Done. Thanks. (I never new how to create an empty list in yaml...)

kuma/javascript/ssr-server.js
davidflanagan: I copied it from KumaScript. Seemed like the idea of being able to query what version is running might be useful to someone at some point.

kuma/javascript/ssr-server.js
davidflanagan: I was worried that if I used text/html there might be code that would try to escape it or add a doctype or a body tag or something. If you do a client side xhr request for html, I think there is an option that causes it to be automatically parsed into a DOM tree, or something, and I wanted to be sure that nothing like that would happen.

Probably not necessary, but it seemed vaguely safer...

I'll add a comment documenting this and explaining the rationale.

kuma/wiki/tests/test_ssr.py
davidflanagan: I've changed it to use the fixture. I'm happy to learn about that!

PR #5364, opened 2019-04-22
fixes bug 1546114 - control PYTHONWARNINGS in docker-compose

peterbe:
I was tempted to add it to [`.env-dist.dev`](https://github.com/mozilla/kuma/blob/master/.env-dist.dev) but there are so many other juicy tips and tricks that is not already in there.


peterbe:
@escattone Is this still right? Now that we're using `- env_file: .env` doesn't it mean that *any* env var in the `.env` file will be available accordingly.

PR #5366, opened 2019-04-22
bug 1544084: revert atomic edits

escattone:
It appears that https://github.com/mozilla/kuma/pull/5348 has broken editing (see https://github.com/mdn/sprints/issues/1366). I'm not sure why, since I can't reproduce it locally or even on stage, so this PR reverts the change that wrapped the edit view in an atomic transaction, in order to restore functionality for the writers while this is investigated further.

### DIFF
--- a/kuma/wiki/views/edit.py
+++ b/kuma/wiki/views/edit.py
@@ -4,7 +4,6 @@ import newrelic.agent
 from csp.decorators import csp_update
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
-from django.db import transaction
 from django.http import HttpResponse, JsonResponse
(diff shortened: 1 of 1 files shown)


peterbe:
It's pretty depressing that [the tests](https://github.com/mozilla/kuma/pull/5348/files) didn't break as of this mini-custom-revert. I understand why but it's still depressing. Mocking sucks. 

By the way, as of that change, I made it so that `send_first_edit_email.delay(new_rev.pk)` gets called **at the end of the transaction**. `transaction.on_commit` still occurs when your view isn't made atomic but it's just that it gets called no matter what happens. Even if something fails later in the view code.

PR #5371, opened 2019-04-26
Ssr cleanup

davidflanagan:


davidflanagan:
This is miscellaneous cleanups to the react front-end code. I'm keeping the commits separate for now to make reviewing easier, since each commit is kind of a separate thing.

davidflanagan:
This is not ready for review yet.

davidflanagan:
Okay, this is ready for review now.  The three commits are three separate fixes, so probably easiest to read them individually.

kuma/wiki/views/document.py
```diff
@@ -872,9 +872,15 @@ def react_document(request, document_slug, document_locale):
     all_locales = (set([doc.locale]) |
                    set(trans.locale for trans in other_translations))
 
+    # Get the JSON data for this document
+    document_data = document_api_data(doc)
+    # And add another property to specify the requested locale
+    # (which may not be the same as the locale of the document)
+    document_data['requestLocale'] = document_locale
```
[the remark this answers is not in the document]
peterbe: I'm not advocating for the "_underscore prefix" but something/anything that sets it apart so that, when reading the code, you appreciate that some things come from state and some things come from runtime.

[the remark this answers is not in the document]
davidflanagan: Peter: that's an excellent point. Also, I've been talking with Ryan about handling redirects differently in kuma/api/v1/views.py, so that we just return the one redirect URL in that case and not a dict full of empty strings.

I propose to add a bit TODO comment for now, and merge this, but then go back in to handle both the redirect case and the request locale case so that the data structure we're using looks like this:

```
{
  documentData: { ... lots of stuff here, or just null for redirect ... }
  redirectURL: ... a string for redirect or null ...
  requestLocale: locale of the request as a string
}
```

davidflanagan:
@escattone: I've updated this patch to add the ensure_contributors flag at your suggestion. And I've addressed Peter's comment by adding a TODO. I'd like to fix that issue when we slightly refactor the way redirects are handled (as you can see in the comment.)

Anyway, I think this will be a pretty easy review, especially if you ignore the snapshot changes and look at the individual commits one by one.

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -93,7 +93,7 @@
 
   {#{% include "includes/a11y-nav.html" %}#}
 
-  {{ render_react_app(document_api_data)|safe }}
+  {{ render_react_app(document_api_data, request_locale)|safe }}
```
escattone: I can't find where `request_locale` is defined.

escattone: It seems to me that `request_locale` is not needed anyway since `requestLocale` is already in the `document_api_data`, so it seems this can remain:
```
{{ render_react_app(document_api_data)|safe }}
```

davidflanagan: You're right. I'll take that second argument out.  (I tried doing this two different ways... This is a leftover from the first way. Thank you for catching it.

PR #5373, opened 2019-04-29
soft page transitions

peterbe:
Tease me all you like but this demonstrates, what I think, a nice upgrade to how we navigate between pages. 

What I disliked was that the whole top nav was fading from opacity:1 to opacity:0.1. And it felt a bit jarring that it isn't fading smoothly. 

I tried to make 3 videos to demonstrate. I use the network dev tools to slow down the api XHR fetches. 

Original:

version0

Better:

version2

Smoother (this PR):

version3

**THE CODE IS TERRIBLE.**

The idea was just to make a change and hope that @davidflanagan or @schalkneethling can do it correctly. I just forced in some CSS into the DOM and I bet there are more correct ways to do it. 

**Most importantly, do you like it enough to make it worth tidying up?**

### DIFF
--- a/kuma/javascript/src/document-provider.jsx
+++ b/kuma/javascript/src/document-provider.jsx
@@ -65,7 +65,7 @@ export default function DocumentProvider(
         const body = document.body;
         // This is the function that does client side navigation
         function navigate(url, localeAndSlug) {
-            body.style.opacity = '0.15';
+            document.querySelector('#page-layout').classList.add('opaque');
             fetch(`/api/v1/doc${localeAndSlug}`, { redirect: 'follow' })
                 .then(response => {
                     if (response.ok) {
@@ -99,7 +99,7 @@ export default function DocumentProvider(
                         }
                         window.scrollTo(0, 0);
                         setDocumentData(json);
-                        body.style.opacity = '1';
+                        document.querySelector('#page-layout').classList.remove('opaque');
                     }
                 })
                 .catch(() => {
--- a/kuma/javascript/src/page.jsx
+++ b/kuma/javascript/src/page.jsx
@@ -295,7 +295,18 @@ export default function Page() {
     return (
         <>
             <Header />
-            <div css={styles.pageLayout}>
+            <style>{`
+                #page-layout {
+                    opacity: 1;
+                    transition: opacity .2s ease-in-out;
+                }
+                #page-layout.opaque {
+                    opacity: 0.0;
+                    transition: opacity .2s ease-in-out;
+                }
+                `}
+            </style>
+            <div css={styles.pageLayout} id="page-layout">
                 <Titlebar />
                 <TOC />
                 <Article />


peterbe:
If you're interested, please ignore my animated gifs and check out the code instead. 
Also, instead of trying to mess with the dev tools's network throttle, put in a `import time; time.sleep(0.5)` into `kuma/api/v1/views.py` around line 35.

davidflanagan:
Thanks for this, @peterbe!  I like it. Did you try it with no opacity effect at all? In my earliest version it felt more responsive to fade as soon as the user clicks on something. But I'm not sure if  that is still true, and it isn't worth optimizing the effect if we're better off with no effect.

And if we're keeping the effect, I like the soft fade out, but do you think it would work to do a fast (or instant) fade in when the new document is ready? If we've got the data, I don't want to make the user wait an extra 200ms to be able to read it...

My original implementation was already a hack and kind of non-Reacty, and your changes take that up a notch. I think the "right" way to do it would be to add an optional  `navigating` property to the DocumentData type, and set it to true at the beginning of navigate(). Then any component that uses the DocumentProvider context can do whatever is appropriate during this transition period.

peterbe:
> your changes take that up a notch. 

right on! I just wanted to play with the effect and didn't care to make it correct. I too think it should be something like:
```
this.setState({aboutToNavigate: true}, () => {
  this.navigate(newEndpoint)
})
```

peterbe:
> don't want to make the user wait an extra 200ms to be able to read it

I'm not sure I agree. The animation is there to indicate that the content has arrived. 

Most people will require hundreds of milliseconds to fetch the XHR. Even if the XHR is browser-cached (thanks to cache-control on the XHR endpoint hopefully) makes the request near instant, I still think it would feel better if the page deliberately demonstrates that it changes content. I don't have the research to back that up. Nick Brandt, where are you on GitHub?!

I do remember skimming some very interesting research that demonstrated the effect of using a spinner animation whilst waiting vs. not bothering with the spinner. When the spinner was displayed whilst waiting, people *perceived* the page to be slower even though they could control for that and make sure it was not. So, the best practice is to do something like:

```
let loaded = false;
window.setTimeout(() => {
   if (!loaded) {
      this.setState({displayLoadingSpinner: true})
   }
}, 500);
let res = await fetch(apiEndpoint)
loaded = true;
this.setState({content: await res.json()})
```
(I hope my pseudo code makes sense)

Anyway, more importantly, if we're going to have *any* animation (spinner or fadeout/fadein), we need to be more careful about timing it the actual delivery. It would be horrible if we A) start the fade out for T-100ms B) at T-50ms, whilst fading, we update the DOM with the new content. C) you can faintly see the new content but now have to wait the remaining 50ms for the fade out and another 100ms for the fade in. Yuck!

Also, what if we A) fade out B) wait wait wait for the XHR (slow 3G connection) C) sit there staring at a nearly all white (or 0.15 opacity) for 3 seconds. Gross!
If the XHR is going to take 3 whole seconds, I'd rather nothing happens with the DOM until the XHR response comes back and React is ready to mutate the DOM with the new stuff. Mind you, since it's XHR you can't rely on the browser chrome to animate that it's loading. What GitHub did (they use the "pjax" pattern too) is they put that tiny thing loading bar that crawls along the very very top of the page. But, unlike our current functionality, they don't use a fadein/fadeout effect once the XHR is ready. They just remove the thin top bar and change the DOM.

peterbe:
In conclusion,...

* It would be great to invite Nick to this conversation
* If we're going to keep a fadeout/fadein we should probably invest in my prototype that attempts to do it softer. Also, we should tie it to mutating the DOM, not before and after waiting for XHR. 
* Invest in some feedback that indicates that a link-click has triggered work and that the user is supposed to wait and not click again. 
* Mine or anybody's code/implementation is probably immaterial in comparison to deciding what the ideal user interface should be. Once that's settled, experts like @davidflanagan can do it the right way with emotion and React state etc.

nickbrandt:
@peterbe thanks for inviting me to the conversation.

I'll try and express my thoughts as clearly as possible, but let me know if I'm interpreting anything wrong.

> The animation is there to indicate that the content has arrived.

I agree here that animations can help to provide an affordance to users that the information on the page has changed. [Here is an article from Nielsen](https://www.nngroup.com/articles/animation-usability) that discusses this along with some other research around animations. It also goes into points around how using animations can actually help make your site appear more modern (ex: we are not reloading the whole page, just the portion that is changing)


> If we've got the data, I don't want to make the user wait an extra 200ms to be able to read it...

This is a fair point, and I do think it is important to not create obstacles in viewing the content on the site. I mean the majority of our audience is developers who care less about animations and more about speed and performance (making a generalization here ;)) However, if we keep the fade relatively quick (similar to last GIF at the top or slightly faster), I don't see this being much of a barrier, while still providing that affordance of updated content. They even discuss instances where UI changes are [too quick](https://www.nngroup.com/articles/too-fast-ux), where a user blinks and doesn't notice the change (though not as common of a problem to have ;) and with most of the content on the page changing, it would be harder to miss)


> Also, what if we A) fade out B) wait wait wait for the XHR (slow 3G connection) C) sit there staring at a nearly all white (or 0.15 opacity) for 3 seconds. Gross!

Completely agree, I don't think the animation should occur until after the response has come back and the page is ready to "transition". 


> Mind you, since it's XHR you can't rely on the browser chrome to animate that it's loading.

Oh, interesting. So this means there will be no loading indicator in the browser chrome that it is working on the request? That seems like a fairly significant problem. If that is the case, I would suggest we implement something similar to what you referenced GitHub uses, indicating it is working on the request. I don't like the idea of using a spinner on the page, so a bar at the top would be more ideal while still providing that affordance. I think we can still incorporate a quick fade if we go this route, but it would be essentially an extra affordance.


To sum up, I think the most important thing here is that there is some sort of affordance that lets the user know it is working on the request (either the browser chrome showing it's loading or us incorporating some other loading indicator). Fade in/out is a nice enhancement/affordance as well, but less of a problem than the one mentioned above. Hope that all makes sense, let me know if you have any questions!

peterbe:
>  there will be no loading indicator in the browser chrome that it is working on the request? 

correct. And that kinda sucks ("Uh?! Did I click it? Lemme click it again!"). And it means we have some choices:

1. A fadeout effect
2. A thin progress bar a la GitHub and YouTube
3. An animated spinner

schalkneethling:
> > there will be no loading indicator in the browser chrome that it is working on the request?
> 
> correct. And that kinda sucks ("Uh?! Did I click it? Lemme click it again!"). And it means we have some choices:
> 
> 1. A fadeout effect
> 2. A thin progress bar a la GitHub and YouTube

I would vote for this one as it seems pretty much the way it is done these days. Also https://github.com/schalkneethling/arewethereyet(Note: This is old and can definitely be improved on) ;p

> 3. An animated spinner

schalkneethling:
> Completely agree, I don't think the animation should occur until after the response has come back and the page is ready to "transition".

The "progress bar" can start immediately though to signal that the click/tap was received and processing of the request has started.

davidflanagan:
Fortunately, MDN already has a little (currently useless) blue bar up at the top of the page that we could use as a progress bar if we choose to go that way.

Another alternative maybe would be an immediate slight fade to .75 opacity or something (instead of the current 0.15) when the user clicks on the link, then a full fade out/fade in effect when the data arrives? Or both.

I'm happy we're having this conversation, and I look forwad to working on it. To be honest, though it feels possibly premature since I think I still have work to do to ensure the correctness of the React-based pages.

escattone:
I'm closing this now that https://github.com/mozilla/kuma/pull/5460 has been merged and deployed.

PR #5374, opened 2019-04-30
use .env in CI

peterbe:
Work in progress to reduce the environment variables in `docker-compose.yml` to an absolute minimum. Everything should either be defaults in `settings/common.py` or personal overrides in `.env`.

### DIFF
--- a/.env.travis
+++ b/.env.travis
@@ -1,7 +1,5 @@
-from .testing import *  # noqa
-
-ES_LIVE_INDEX = False
-ES_DEFAULT_NUM_REPLICAS = 0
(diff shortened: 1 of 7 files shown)


escattone:
The Travis errors are due to Travis using MySQL 5.7 on their Ubuntu Xenial boxes (see https://docs.travis-ci.com/user/database-setup/#mysql) and MySQL 5.7 no longer uses `storage_engine` but `default_storage_engine` instead. In order to fix that, we need to somehow ensure that we're using MySQL 5.6 in our Travis set-up.

PR #5375, opened 2019-04-30
bug 1548130 - index faster

peterbe:
The sad truth is that when doing this change locally, it doesn't actually make a huge difference. 
Instead of incurring 1-2 SQL queries per document, it's now 0 SQL queries (warmed up by the prefetch_related). 

The real heavy cost lies in the lines `obj.get_summary_text()` and `obj.extract.macro_names()`.

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -216,6 +216,7 @@ def get_seo_description(content, locale=None, strip_markup=True):
     if content:
         # Try constraining the search for summary to an explicit "Summary"
         # section, if any.
+        # This line is ~20x times slower than doing the PyQuery analysis.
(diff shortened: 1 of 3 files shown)


escattone:
The Travis errors are due to Travis using MySQL 5.7 on their Ubuntu Xenial boxes (see https://docs.travis-ci.com/user/database-setup/#mysql) and MySQL 5.7 no longer uses `storage_engine` but `default_storage_engine` instead. In order to fix that, we need to somehow ensure that we're using MySQL 5.6 in our Travis set-up.

escattone:
This looks good and worked for me locally. The Travis failures will have to be fixed in another PR. Thanks for looking into this @peterbe!

PR #5376, opened 2019-05-01
 bug 1548591 - use docker-compose in travis

peterbe:
- [x] Tie this to a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1548591)
- [x] Make it pass
- [x] Do docs builds
- [x] Do `localerefresh`
- [x] Check that we do everything we did in `tox.ini` here too.
- [x] Do what can be done in a matrix of parallelism
- [ ] build-static (e.g. `collectstatic`) should only run once
- [x] Coverage reporting
- [ ] In [readthedocs](https://readthedocs.org/dashboard/kuma/advanced/) we're still using Python 2 but progress in this PR proves that we can `sphinx-build` with Python 3. So let's switch to that in readthedocs when this lands.


escattone:
I really like where you're going with this! I've wanted to do this for a long time.

peterbe:
Do you know why we're not using CircleCI? For someone who has used both extensively, CircleCI is better.

.travis.yml
```diff
@@ -1,89 +1,94 @@
...
-notifications:
-    irc:
-        channels:
-            - "irc.mozilla.org#mdndev"
-        on_success: always
-        on_failure: always
-        use_notice: true
+      - docker-compose --version
+      - docker-compose pull
+      - docker-compose build
```
escattone: We don't need to do this since we don't have any images that need to be built in the `docker-compose.yml` file.  Some time ago, we needed to do it for the `mysql` image I think, but no longer.

However, we do need to build a fresh `mdnwebdocs/kuma_base` Docker image. This is so the tests pick-up any changes made within the PR that would affect that image -- e.g., changes to the `docker/images/kuma_base/Dockerfile` itself of course, but also changes to Kuma's `package.json` or changes to many of the Python requirements files).

So I think instead of the `docker-compose build` we need something like `VERSION=latest make build-base` (which must be placed after the `docker-compose pull` so it overrides the pulled `mdnwebdocs/kuma_base`).

escattone: I noticed that you're avoiding using the `make` commands, but if you do, you'll probably need to at least define `DJANGO_SETTINGS_MODULE=kuma.settings.testing` in the Travis environment.

.travis.yml
```diff
@@ -1,89 +1,94 @@
...
+      - docker-compose up
```
escattone: I think you want `docker-compose up -d` here?

.travis.yml
```diff
@@ -1,89 +1,94 @@
...
+      - docker-compose exec web npm run webpack
```
escattone: All of the `docker-compose exec ...` calls should be `docker-compose exec -T ...` so that a terminal isn't allocated (since it won't be interactive).

.travis.yml
```diff
@@ -1,89 +1,94 @@
...
+      - docker-compose exec web manage.py compilejsi18n
+      - docker-compose exec web manage.py compilejsi18n -d react
+      - docker-compose exec web manage.py collectstatic --noinput
+      - docker-compose exec web pytest kuma/search -x
```
escattone: The steps above don't fully capture what we run currently with:
```
- docker-compose exec -T web make localecompile build-static clean
- docker-compose exec -T web ./manage.py migrate
- docker-compose exec -T web make test
```

.travis.yml
```diff
@@ -1,89 +1,94 @@
...
+      - docker-compose exec web flake8 kuma docs tests
```
escattone: This run of `flake8` will need the following config that used to be provided by `tox.ini`:
```
exclude = **/migrations/**,.tox,*.egg,vendor
# E501 - line too long (82 > 79 characters)
# E731 - do not assign a lambda expression, use a def
# F405 - name may be undefined, or defined from star imports: module
# W504 - line break after binary operator
#        conflicts with W503 (line break before binary operator)
ignore = E501,E731,F405,W504
# flake8-import-order settings
import-order-style=edited
application-import-names=kuma,pages,utils
```

One option is that we could create a new `.flake8` file that has this configuration.

.travis.yml
```diff
@@ -1,89 +1,94 @@
...
 branches:
     only:
         - master
-        - potato-1490727
-cache:
-    pip: True
-    directories:
-      - node_modules
-      - downloads
 language: python
```
escattone: I think we'll need the following as well?
```
python:
  - "2.7"
```

scripts/travis-run
```diff
@@ -0,0 +1,18 @@
...
+set -e  # Exit on non-zero status
+set -x  # Print commands run
+set -u  # Treat unset variables as an error
+
+# XXX would be nice to have code comments explaining each of these
+docker-compose --version
+docker-compose pull
+docker-compose build
+docker-compose up -d
+# XXX Sleep here?!
```
escattone: You can use the Python package `urlwait` (https://pypi.org/project/urlwait/), which is baked into the `mdnwebdocs/kuma_base` image. In our Jenkins code (see the `compose_test()` function within `Jenkinsfiles/utils.groovy`), we use `docker-compose.test.yml` which waits on `mysql` to spin-up before running tests. So perhaps we could add the following to the Travis environment (by default `urlwait` waits to connect to the service defined in `DATABASE_URL`) :
```
- DATABASE_URL=mysql://root:kuma@mysql:3306/developer_mozilla_org
- URLWAIT_TIMEOUT=300
```
and then use 
```
docker-compose exec -T urlwait && npm run webpack
```

.travis.yml
```diff
@@ -1,89 +1,87 @@
...
-    - redis
+    # - mysql
+    # - redis
 env:
     global:
-        - DATABASE_URL=mysql://root:@127.0.0.1:3306/kuma
-        - DJANGO_SETTINGS_MODULE=kuma.settings.travis
+#         - DATABASE_URL=mysql://root:@127.0.0.1:3306/kuma
+#         - DJANGO_SETTINGS_MODULE=kuma.settings.travis
         - DOCKER_COMPOSE_VERSION=1.23.2
```
escattone: We just need this also within the `env: global:` section, which solves the permission issue:
```
- UID=0
```

docker/images/kuma_base/Dockerfile
```diff
@@ -123,4 +125,10 @@ ENV NODE_PATH=/tools/node_modules
 USER kuma
 WORKDIR /app
 EXPOSE 8000
+
+# XXX This is not used.
+# This is how it's done in Prod
+# https://github.com/mdn/infra/blob/585e7fea362e80b341b8e0f3d4f9070806966b4d/apps/mdn/mdn-aws/k8s/kuma.deploy.yaml.j2#L8
+# And in local dev we use the 'command:' in docker-compose.
+# Delete??
 CMD ["gunicorn", "--bind", "0.0.0.0:8000", "--timeout=120", "--worker-class=meinheld.gmeinheld.MeinheldWorker", "kuma.wsgi:application"]
```
escattone: Yes, I think we can delete the `CMD`.

docker/images/kuma_base/Dockerfile
```diff
@@ -98,7 +98,9 @@ RUN useradd --uid 1000 --shell /bin/bash --create-home kuma \
 # install Python libraries
 WORKDIR /app
 COPY --chown=kuma:kuma ./requirements /app/requirements
-RUN pip install --no-cache-dir -r requirements/dev.txt
+# RUN pip install --no-cache-dir -r requirements/dev.txt
+# Make it less noisy. Might remove this once Travis and docker-compose is working spiffily
+RUN pip install --no-cache-dir -r requirements/dev.txt --quiet
```
escattone: Once you're done debugging, I would prefer that `--quiet` was removed. I like to see the details sometimes.

.travis.yml
```diff
@@ -1,89 +1,63 @@
...
-        on_success: always
-        on_failure: always
-        use_notice: true
+        - language: python
+          python: 3.6
+          # env:
+          #   BUILD_DOCS=1
+          #   # BUILD_DOCKER=0
+          #   # JSLINT=0
+          script:
```
escattone: This was something that yesterday I was wondering if we could do (i.e., could each section of the build matrix have its own script). Nice!

Now that I know this, I would suggest we dump the bash scripts and just put the commands in their proper `script:` section in the build matrix, so that it's all in one place. I think we can do that since you can embed the `set -e` commands within the `script:` section as desired.

scripts/travis-run
```diff
@@ -0,0 +1,77 @@
...
+
+    docker-compose --version
+    docker-compose up -d
+    # Hang tight until MySQL appears to be working
+    docker-compose exec -T web urlwait mysql://root:kuma@mysql:3306/developer_mozilla_org 30
+    # Hang tight until Elasticsearch appears to be working
+    docker-compose exec -T web urlwait http://elasticsearch:9200 30
+
+    # XXX Would it help to bundle all of these make commands into one?!
+    docker-compose exec -T web make webpack
```
escattone: `make build-static` includes the `make webpack, so I don't think we need it here, or perhaps it needs to be run prior to the `make localecompile`? I don't think so, but maybe I'm missing something.

scripts/travis-run
```diff
@@ -0,0 +1,77 @@
...
+    docker-compose exec -T web make localecompile
+    docker-compose exec -T web make build-static
+    docker-compose exec -T web make pylint
+
+    #docker-compose exec -T web ./manage.py migrate ## Do we really need this?!
```
escattone: I think we need this simply to test that the migrations run without error.

scripts/travis-run
```diff
@@ -0,0 +1,77 @@
...
+    docker-compose exec -T web make test
+    # XXX Why is this necessary?
+    # docker-compose down --volumes --remove-orphans
```
escattone: It probably isn't, unless Travis caches the docker-compose volumes, which I assume it doesn't by default. I put it in there because I needed it within Jenkins since the Jenkins server is long-lived.

.travis.yml
```diff
@@ -1,89 +1,59 @@
...
+            directories:
+              - downloads
+          script:
+            - scripts/travis-python
+        - language: node_js
+          node_js: 10
+          cache: npm
+          script:
+            - scripts/travis-js
+# install:
```
ExE-Boss:
```suggestion
    allow_failures:
        - python: "3.6"
# install:
```

peterbe:
> Shouldn’t Python 3 still be allowed to fail?

In this branch, my intention is to no longer do python2 and python3. Instead, Python 2 only for all the docker stuff and the actual tests & linting. And Python 3 exclusively to build the docs.

docker/images/kuma_base/Dockerfile
escattone: On second thought, `--quiet` is fine.

.travis.yml
peterbe: I disagree. I much prefer doing the work inside a dedicated bash script. The yaml indentation is just a pain and editing a bash script gives my syntax highlighting for bash and not for yaml. 

I acknowledge that the file is named `travis-python` (for example) but as it stands, its *content* is theoretically agnostic of TravisCI. I.e. we *could* switch to CircleCI and keep this file as is.

ExE-Boss:
> > Shouldn’t Python 3 still be allowed to fail?
> 
> In this branch, my intention is to no longer do python2 and python3. Instead, Python 2 only for all the docker stuff and the actual tests & linting. And Python 3 exclusively to build the docs.

The intention however, is to migrate **Kuma** to **Python 3** (eventually…), and the **Python 3** build was added in #4848, which is part of [bug 1467518](https://bugzilla.mozilla.org/show_bug.cgi?id=1467518 "Convert Kuma to Python 3").

.travis.yml
peterbe: Actually, perhaps the name of the bash script(s) should be something like `ci-docs` instead of `travis-docs`.

Makefile
```diff
@@ -33,8 +33,11 @@ export DJANGO_SETTINGS_MODULE ?= kuma.settings.testing
 test:
 	py.test $(target)
 
+
```
ExE-Boss: Extraneous empty line:
```suggestion
```

scripts/travis-run
peterbe: First of all, my bad for explicitly running `make webpack` when it's part of `build-static`. 

Right now, the problem I have is quite luxurious. The tests passed but I noticed that the super-expensive `make build-static` was called twice. First by `kuma/Dockerfile` and then by this `docker-compose ...` command.

.travis.yml
escattone: 👍 Good points, let's stick with the bash scripts, and I'd vote for `ci-xxx` over `travis-xxx`, but whatever you prefer. It's fine the way it is as well, and we can change it if we ever change our CI provider.

scripts/travis-python
```diff
@@ -0,0 +1,50 @@
...
+time docker-compose up -d
+# Hang tight until MySQL appears to be working
+time docker-compose exec -T web urlwait mysql://root:kuma@mysql:3306/developer_mozilla_org 30
+# Hang tight until Elasticsearch appears to be working
+time docker-compose exec -T web urlwait http://elasticsearch:9200 30
+
+# echo "docker-compose exec -T web make webpack"
+# docker-compose exec -T web make webpack
+time docker-compose exec -T web make clean
+time docker-compose exec -T web make localecompile
```
escattone: Just for reference, https://github.com/mozilla/kuma/pull/5268#issuecomment-467682716  explains why we need the `make localecompile` before we run `make test` or `make coveragetest`.

.travis.yml
peterbe: Want me to change that now?

.travis.yml
```diff
@@ -1,89 +1,36 @@
...
+            - set -e
+            - scripts/travis-python
+            - scripts/travis-localerefresh
+          after_success:
+            - bash <(curl -s https://codecov.io/bash)
+        - language: node_js
+          node_js: 10
+          cache: npm
+          script:
+            - scripts/travis-js
```
ExE-Boss:
```suggestion
            - scripts/travis-js
        - language: python
          python: 3.6
          env:
            UID=0
            DOCKER_COMPOSE_VERSION=1.23.2
            _MAY_FAIL=1
          cache:
            directories:
              - downloads
              - build
              - static
          script:
            - set -e
            - scripts/travis-python
            - scripts/travis-localerefresh
    fast_finish: true
    allow_failures:
        - env: _MAY_FAIL=1
```

.travis.yml
escattone: It would be nice, but honestly whatever you prefer.

Makefile
```diff
@@ -34,7 +34,9 @@ test:
 	py.test $(target)
 
 coveragetest: clean
-	py.test --cov=$(target) $(target)
+	# py.test --cov=$(target) --no-cov-on-fail $(target)
+	@ echo "TEMPORARITLY RUNNING WITH -x"
+	@ py.test -x --cov=$(target) --no-cov-on-fail $(target)
```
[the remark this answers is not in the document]
escattone: I like the addition of `--no-cov-on-fail`.

.travis.yml
```diff
@@ -1,89 +1,36 @@
...
+            directories:
+              - downloads
+              - build
+              - static
+          script:
+            - set -e
+            - scripts/travis-python
+            - scripts/travis-localerefresh
+          after_success:
+            - bash <(curl -s https://codecov.io/bash)
```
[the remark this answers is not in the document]
ExE-Boss: This is similar to piping into bash, except that the file is first saved to disk into a temporary file, which is then executed.

This allows avoiding some piping into bash vulnerabilities where the fact that the file is streamed lets the server allow switching the payload midway through if it detects that it’s being streamed rather than downloaded ([which is possible](https://www.idontplaydarts.com/2016/04/detecting-curl-pipe-bash-server-side/)).

peterbe:
@escattone I believe I've addressed all things now. Exception one thing...

The reason I removed the allow-failures run of Python 3 is because I don't think it's valuable compared to the disadvantages. Number of disadvantage is noise and speed. If we're going to allow it to fail, I'm not sure it's worth keeping and I'm pretty confident that the only reason it's failing is because of the Celery dependencies and maybe some easy things that are best tackled when we tackle Python 3. 
My intention is to, within days, make a PR that rewrites everything in one sweep. Meaning, changing the Dockerfile, the .travis.yml, and all the dependencies that need to be upgraded. Switching a codebase from Python 2 to Python 3 is easy if you don't need to worry about making it work in both. What's hard is the upgrade of certain libs.

peterbe:
Another thing that bothers me is that I don't know how to address all of these warnings:

```
▶ travis lint .travis.yml
Warnings for .travis.yml:
[x] in matrix.include section: unexpected key language, dropping
[x] in matrix.include section: unexpected key cache, dropping
[x] in matrix.include section: unexpected key script, dropping
[x] in matrix.include section: specified node_js, but setting is not relevant for python
```

Am I not setting the language for each matrix node correctly?! Are the cache directories not working when done within a matrix??

.travis.yml
```diff
@@ -1,89 +1,36 @@
...
+          script:
+            - set -e
+            - scripts/ci-python
+            - scripts/ci-localerefresh
+        - language: node_js
+          node_js: 10
+          cache: npm
+          script:
+            - scripts/ci-js
+after_success:
```
ExE-Boss:
```suggestion
        - language: python
          python: 3.6
          env:
            UID=0
            DOCKER_COMPOSE_VERSION=1.23.2
            _MAY_FAIL=1
          cache:
            directories:
              - downloads
              - build
              - static
          script:
            - set -e
            - scripts/travis-python
            - scripts/travis-localerefresh
    fast_finish: true
    allow_failures:
        - env: _MAY_FAIL=1
after_success:
```

.travis.yml
```diff
@@ -1,89 +1,36 @@
...
+            directories:
+              - downloads
+              - build
+              - static
+          script:
+            - set -e
+            - scripts/ci-python
+            - scripts/ci-localerefresh
+        - language: node_js
+          node_js: 10
```
peterbe: This is a (performance) hack. The intention is to run npm stuff (i.e. `scripts/ci-js`) entirely separate. I think it would make more sense to use the Node and npm that gets installed in our Dockerfiles. But at the moment, the only thing it does is linting which doesn't affect runtime. 

Not sure how I feel about this. Perhaps we should move this into the Python 2.7 block above and execute it with `docker-compose exec -T web ...`. Hopefully we'll soon have more than just JavaScript linting.

peterbe:
(regarding travis cli warnings...)

> Another thing that bothers me is that I don't know how to address all of these warnings:

I sent a support ticket to TravisCI for help on this. The tricky thing is that I'm struggling to see if we're getting benefits from the Travis cache!

peterbe:
@escattone *If* you think the PR is good enough, can you please do a "Squash and rebase" (green button) or whatever it's called. Or ping me and I'll do one last rebase and squash manually before you merge.

ExE-Boss:
You should put the bug number into the PR tile like so:
> Fix bug 1548591: Use docker‑compose in Travis

peterbe:
A ha. [Here's the explanation](https://github.com/codecov/codecov-bash/issues/94) for the problems with codecov.

peterbe:
@escattone Now with working codecov!!

PR #5378, opened 2019-05-01
bug 1544482: reduce indexing default chunk size

escattone:
This PR reduces the default chunk size used when indexing for search, as well as makes it consistent whether indexing via the Django Admin console or via the `./manage.py reindex` management command. The default was 500 when using the Django Admin console, and 1000 when using the `./manage.py reindex` management command. Now, they are both 250.

The reason for this is that using a chunk size of 500 caused out-of-memory issues with the `celery-worker` pods on stage, even when the Kubernetes memory limit of the `celery-worker` pods was increased to match that of the pods on prod (`4Gi`).

A re-index on stage was successfully completed on stage with a chunk-size of 250.

### DIFF
--- a/kuma/search/management/commands/reindex.py
+++ b/kuma/search/management/commands/reindex.py
@@ -11,11 +11,11 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument(
             '-c', '--chunk',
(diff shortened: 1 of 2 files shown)


escattone:
I would say this PR is a prerequisite for the next push to production, since we'll need to re-create the search index on prod after the push (and it's easier to use the Admin console).

PR #5379, opened 2019-05-01
fixes bug 1548130 - make reindexing faster

peterbe:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1548130#c7 and https://bugzilla.mozilla.org/show_bug.cgi?id=1548130#c8

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -75,6 +75,9 @@ class Extractor(object):
         """
         Extract a unique set of KumaScript macro names used in the content
         """
+        if not self.document.html:
+            # No document.html, then there's no point bothering to parse it.
+            return []
         text_items = []
         for token in parse(self.document.html).stream:
             if token['type'] in ('Characters', 'SpaceCharacters'):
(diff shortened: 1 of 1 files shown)


kuma/wiki/content.py
```diff
@@ -217,10 +219,13 @@ def get_seo_description(content, locale=None, strip_markup=True):
         # Try constraining the search for summary to an explicit "Summary"
         # section, if any.
         # This line is ~20x times slower than doing the PyQuery analysis.
-        summary_section = (parse(content).extractSection('Summary')
-                           .serialize())
-        if summary_section:
-            content = summary_section
+        # Both `parse()` and `.serialize()` are slow and expensive.
+        # That's why we're careful to avoid it if we can.
+        if 'Summary' in content:
```
peterbe: This one line doubles the speed of reindexing.

kuma/wiki/content.py
```diff
@@ -1138,3 +1143,11 @@ def validate_src(self, src):
                     continue
             return True
         return False
+
+
+_content_section_tool = ContentSectionTool()
```
escattone: Good idea, seems safe from reading the `ContentSectionTool` code (i.e., no residue from a previous `parse` call can pollute a subsequent call).

PR #5380, opened 2019-05-01
test of using only docker-compose in travis

escattone:
This PR is intended only as an experiment for testing ideas within Travis.


escattone:
This is missing tests for the docs build and the `localerefresh`, and perhaps can be parallelized with a build matrix, but it works so far.

peterbe:
I *think* I've copied all of your tricks into https://github.com/mozilla/kuma/pull/5376
Let's crack on over there.

peterbe:
> perhaps can be parallelized with a build matrix, but it works so far.

Yes!! Simply, let's not *forget* that but we'll tackle it as soon as we get the basics working.

PR #5382, opened 2019-05-03
fixes bug 1548082 - hide ES_URLS creds in admin display

peterbe:
Any URL in `settings.ES_URLS`, *with credentials in the URL*, get scrubbed before being displayed like `https://username:password@....` (where the '...' is the same as before).

### DIFF
--- a/kuma/core/context_processors.py
+++ b/kuma/core/context_processors.py
@@ -1,8 +1,8 @@
 from __future__ import unicode_literals
-from six.moves.urllib.parse import urlparse
 
 from django.conf import settings
 from django.utils import translation
+from six.moves.urllib.parse import urlparse
 
 from .i18n import get_language_mapping
 
--- a/kuma/core/context_processors.py
+++ b/kuma/core/context_processors.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from six.moves.urllib.parse import urlparse
 
 from django.conf import settings
 from django.utils import translation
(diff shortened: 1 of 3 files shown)


peterbe:
@escattone Rebased and should hopefully work in CI now. r?

peterbe:
Ping @escattone for review

kuma/core/context_processors.py
```diff
@@ -2,13 +2,35 @@
...
     """Adds settings to the context."""
-    return {'settings': settings}
+
+    def clean_safe_url(url):
+        if '://' not in url:
+            # E.g. 'elasticsearch:9200'
+            url = 'http://' + url
+        parsed = urlparse(url)
+        if '@' in parsed.netloc:
+            parsed = parsed._replace(
```
escattone: I would prefer not using a "private" method on the parsed object. I was thinking that instead you could a standard string replacement on `parsed.netloc`, and then re-assemble the URL with `urlunparse`?

templates/admin/search/index/change_list.html
```diff
@@ -3,9 +3,11 @@
 
 {% block content_title %}
   {{ block.super }}
+
   {% if settings.ES_URLS %}
   <div class="g-d-c grp-rte">
-    <p>{% blocktrans with backend=settings.ES_URLS|join:", " %}Elasticsearch server: {{ backend }}{% endblocktrans %}</p>
+    {# 'safe_es_urls' comes from a context processors #}
```
escattone: Nit: drop the plural on `processors`, so just `a  context processor`

kuma/core/context_processors.py
peterbe: It's been too long since I used Python 2 :)

The `_replace` isn't really a private API. It's [fully documented](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) **but only in Python 3's documentation!**
But it's **not** documented in [python 2's urlparse](https://docs.python.org/2/library/urlparse.html) so your point is correct.

However, if I leave it like this, this code will be ready for Python 3 without modification. With that in mind, would you be OK with leaving it like this so we can keep it when we go to Python 3?

kuma/search/tests/test_views_admin.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.mark.django_db
+def test_view_indexes(client, settings, wiki_user):
+    """Test that looking at the list of Indexes doesn't reveal the
+    username:password from any 'settings.ES_URLS'.
+    """
+    wiki_user.is_staff = True
+    wiki_user.is_superuser = True
+    wiki_user.set_password('secret')
+    wiki_user.save()
+    client.login(username=wiki_user.username, password='secret')
```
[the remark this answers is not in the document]
peterbe: Thanks!

kuma/core/context_processors.py
escattone: Wow, I didn't realize that `_replace` was documented! Thanks for pointing that out. It's really strange that they kept the `_` prefix in Python 3! Anyway, I'm totally OK with it knowing this, thanks!

PR #5383, opened 2019-05-03
Sprints issue 1393, update developer needs banner copy

schalkneethling:
@peterbe r?

### DIFF
--- a/kuma/banners/jinja2/banners/includes/developer-needs.html
+++ b/kuma/banners/jinja2/banners/includes/developer-needs.html
@@ -1,9 +1,9 @@
 <div id="developer-needs" class="mdn-cta-content-container">
     <h2 class="mdn-cta-title">
-        {{ _('MDN Developer Survey') }}
+        {{ _('MDN survey') }}
     </h2>
     <p class="mdn-cta-copy">
-        {{ _('We want to learn more about the needs of developers') }}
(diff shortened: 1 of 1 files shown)


kuma/banners/jinja2/banners/includes/developer-needs.html
```diff
@@ -1,16 +1,16 @@
 <div id="developer-needs" class="mdn-cta-content-container">
     <h2 class="mdn-cta-title">
-        {{ _('MDN Developer Survey') }}
+        {{ _('MDN survey') }}
```
peterbe: I'm not an English expert but I really don't like this lowercasing. I think it should be "MDN Survey" since it's a header. 
Actually, I checked with a close friend (British) who is really good at this and he too thinks it should be "MDN Survey". 

I don't know the background of where this came from but I would love to have one last look at it @atopal 

Also, @davidflanagan you're pretty good with words :) WHat do you think?

atopal: Yeah, that's a typo. It was supposed to be "MDN Survey". Thanks for catching that, Peter.

schalkneethling: upated.

peterbe:
@escattone Can we merge this even though Travis is busted at the moment. I'd love to get this into Prod as of your next intention to ship.

PR #5386, opened 2019-05-06
Bug 1525719: add waffle flags to the /api/v1/whoami endpoint

davidflanagan:
For server-side rendered and cached versions of our pages, we need
to remove any user-specific data from the page. This includes waffle
flags, which can vary by user. This PR enables us to query the waffle
flags after a page has loaded, at the same time that we query the
user's name and profile picture.

### DIFF
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -226,20 +226,20 @@ def test_whoami_anonymous(client, api_settings, timezone):
     response = client.get(url, HTTP_HOST=api_settings.BETA_HOST)
     assert response.status_code == 200
     assert response['content-type'] == 'application/json'
-    json = response.json()
+    data = response.json()
 
     # pytest restrictions on using the db during tests means that
     # we don't know what flags, switches and samples we expect to
     # receive in the response. This test assumes that waffle works
     # correctly, so all we are doing here is validating that the
     # response includes dicts in the expected places.
-    assert type(json['waffle']['flags']) == dict
-    assert type(json['waffle']['switches']) == dict
-    assert type(json['waffle']['samples']) == dict
+    assert type(data['waffle']['flags']) == dict
+    assert type(data['waffle']['switches']) == dict
+    assert type(data['waffle']['samples']) == dict
 
     # Now check the rest of the response data
-    del json['waffle']
-    assert json == {
+    del data['waffle']
+    assert data == {
         'username': None,
         'timezone': timezone,
         'is_authenticated': False,
@@ -274,20 +274,20 @@ def test_whoami(user_client, api_settings, wiki_user, beta_testers_group,
     response = user_client.get(url, HTTP_HOST=api_settings.BETA_HOST)
     assert response.status_code == 200
     assert response['content-type'] == 'application/json'
-    json = response.json()
+    data = response.json()
 
     # pytest restrictions on using the db during tests means that
     # we don't know what flags, switches and samples we expect to
     # receive in the response. This test assumes that waffle works
     # correctly, so all we are doing here is validating that the
     # response includes dicts in the expected places.
-    assert type(json['waffle']['flags']) == dict
-    assert type(json['waffle']['switches']) == dict
-    assert type(json['waffle']['samples']) == dict
+    assert type(data['waffle']['flags']) == dict
+    assert type(data['waffle']['switches']) == dict
+    assert type(data['waffle']['samples']) == dict
 
     # Now check the rest of the response data
-    del json['waffle']
-    assert json == {
+    del data['waffle']
+    assert data == {
         'username': wiki_user.username,
         'timezone': timezone,
         'is_authenticated': True,
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -217,6 +217,7 @@ def test_whoami_disallowed_methods(client, api_settings, http_method):
     assert_no_cache_header(response)
 
 
+@pytest.mark.django_db
 @pytest.mark.parametrize('timezone', ('US/Eastern', 'US/Pacific'))
 def test_whoami_anonymous(client, api_settings, timezone):
     """Test response for anonymous users."""
@@ -225,7 +226,20 @@ def test_whoami_anonymous(client, api_settings, timezone):
     response = client.get(url, HTTP_HOST=api_settings.BETA_HOST)
     assert response.status_code == 200
     assert response['content-type'] == 'application/json'
-    assert response.json() == {
+    json = response.json()
+
+    # pytest restrictions on using the db during tests means that
+    # we don't know what flags, switches and samples we expect to
+    # receive in the response. This test assumes that waffle works
+    # correctly, so all we are doing here is validating that the
+    # response includes dicts in the expected places.
+    assert type(json['waffle']['flags']) == dict
+    assert type(json['waffle']['switches']) == dict
+    assert type(json['waffle']['samples']) == dict
+
+    # Now check the rest of the response data
+    del json['waffle']
+    assert json == {
         'username': None,
         'timezone': timezone,
         'is_authenticated': False,
@@ -240,6 +254,7 @@ def test_whoami_anonymous(client, api_settings, timezone):
     assert_no_cache_header(response)
 
 
+@pytest.mark.django_db
 @pytest.mark.parametrize(
     'timezone,is_staff,is_superuser,is_beta_tester',
     [('US/Eastern', False, False, False),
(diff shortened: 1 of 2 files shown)


davidflanagan:
@escattone this is my first step toward handling client side waffle flags in a new way.

The test is not passing yet... The request comes back with real waffle flag data in it, but when I try to query that data in the test itself, I don't get anything. So I think I need some mock waffle data or something.  I'm flagging for your review in its current state because I suspect you might have some idea about how to make the test work.

kuma/api/v1/views.py
```diff
@@ -133,6 +134,19 @@ def document_api_data(doc=None, ensure_contributors=False, redirect_url=None):
...
+# the whoami API will query the current value of each of them. This code
+# for querying the models and then querying the setting for each
+# model is copied from waffle.views._generate_waffle_js.
+#
+# Note that if we upgrade django-waffle, version 15 introduces a pluggable
+# flag model, and the approved way to query all flags will then become:
+#    all_waffle_flags = get_waffle_flag_model().get_all()
+all_waffle_flags = Flag.get_all()
+all_waffle_switches = Switch.get_all()
+all_waffle_samples = Sample.get_all()
```
peterbe: This here makes me nervous. First because it feels bad form to have a module global mutable sitting around simply be importing the module. 
Ideally, we shouldn't be touching the MySQL until Django decides to boot things up. 

Secondly, having data loaded into the module namespace makes it clunky to write tests that change flags, switches, and samples on a test-by-test basis.

peterbe: What do you think about using a memoizing function or something instead? Like django's lru_cache function.

kuma/api/v1/views.py
```diff
@@ -167,4 +181,12 @@ def whoami(request):
                 'large': None,
             }
         }
+
+    # Add waffle data to the data
+    data['waffle'] = {
+        'flags': {f.name: f.is_active(request) for f in all_waffle_flags},
+        'switches': {s.name: s.is_active() for s in all_waffle_switches},
+        'samples': {s.name: s.is_active() for s in all_waffle_samples},
```
peterbe: First of all, all *but* `all_waffle_flags` here could be heavily cached. Either in Redis or inside each gunicorn worker. 
This endpoint is going to see heavy traffic since it'll run for every single browser load. 
As it stands now we need to generate 3 dicts from 3 for-loops on every request. 

I don't have a better suggestion but I wanted to raise the point. Also, I'm not very experienced with django-waffle so I'm not entirely sure what I'm talking about. 

And we would certainly not want to break the Django admin. I.e. editing waffle stuff shouldn't require that we restart the servers for it to take effect. If it was hooked into the django cache framework, with a known cache key, we could write a `post_save` signal somewhere that's tied to these models.

peterbe:
The tests are failing: https://travis-ci.org/mozilla/kuma/jobs/529405669
And I think I know why. Or rather, a qualified guess. In Travis, we start up `docker-compose up -d` and as part of that command we immediately start the `web` service too which, due to Django importing all the Python files, triggers a need for a healthy Redis connection. 

Technically, I think something like `Flag.get_all()` triggers a code like this:
```python
def get_all(cls):
    cached = cache.get(CACHE_KEY)  # THIS is the line in the above mentioned Travis build failure Tracbeack
    if cached:
        return cached

    objs = cls.get_all_from_db()
    ...
```

I think the argument is strong that we shouldn't need or depend on Redis/MySQL until after all the python files have been fully loaded.

kuma/api/v1/views.py
davidflanagan: From a perf perspective, all of this code is already being run in the {{waffle.wafflejs()}} template tag in one of the files we import into our templates. So this is not adding any load to the server that the server isn't already handling.

Second: the only time this patch requires a server restart is when a new flag or switch or sample is added. But just adding one of those from the admin interface without deploying new code to use the flag or switch or sample won't have any impact. So, if a new flag is added, it will be useless until new code is deployed. And the deploy will cause the servers to restart. So I think we're okay here.

I guess this does mean that when deleting a flag, it would be important to first set it to false, and then delete it because existing frontend code won't see the deletion until the next deploy.

kuma/api/v1/views.py
davidflanagan: Yes, I can change this to call get_all() the first time these are needed, especially if doing it this way is causing a test failure.

kuma/api/v1/tests/test_views.py
```diff
@@ -225,7 +227,10 @@ def test_whoami_anonymous(client, api_settings, timezone):
     response = client.get(url, HTTP_HOST=api_settings.BETA_HOST)
     assert response.status_code == 200
     assert response['content-type'] == 'application/json'
-    assert response.json() == {
+    json = response.json()
```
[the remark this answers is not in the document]
davidflanagan: Good point. I'll fix that. Thanks.

kuma/api/v1/tests/test_views.py
```diff
@@ -272,3 +281,24 @@ def test_whoami(user_client, api_settings, wiki_user, beta_testers_group,
...
+    response = client.get(url, HTTP_HOST=api_settings.BETA_HOST)
+    assert response.status_code == 200
+    assert response['content-type'] == 'application/json'
+    json = response.json()
+
+    flags = json['waffle']['flags']
+    switches = json['waffle']['switches']
+    samples = json['waffle']['samples']
+
+    for flag in flags.iterkeys():
```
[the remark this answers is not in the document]
davidflanagan: Heh. I forgot that I could just loop over the keys directly. Thanks. The code is removed, in the most recent version, but thanks for pointing this out.

kuma/api/v1/tests/test_views.py
```diff
@@ -272,3 +281,24 @@ def test_whoami(user_client, api_settings, wiki_user, beta_testers_group,
...
     assert_no_cache_header(response)
+
+
+@pytest.mark.django_db
+def test_whoami_waffle(client, api_settings):
+    """Test that the response includes waffle information."""
+    url = reverse('api.v1.whoami')
+    response = client.get(url, HTTP_HOST=api_settings.BETA_HOST)
+    assert response.status_code == 200
+    assert response['content-type'] == 'application/json'
```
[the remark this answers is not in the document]
davidflanagan: I think it is just copied from other tests. Even though it is not necessary, it seems worth failing a test if the endpoint is changed to omit the content-type header. So I'm inclined to keep it as is.

kuma/api/v1/views.py
peterbe: > only time this patch requires a server restart is when a new flag or switch or sample is added

What I did was load the JSON API whoami endpoint in one tab, noticed the boolean on one of the switches, opened the Django admin and toggled one of the switches, back to the JSON API whoami endpoint and refresh. The boolean hadn't changed under `switches`. 

If we move the `Switches.get_all()` into the view function it would read from the cache every time. So if we can just trigger a cache purge (from the actions in Django admin) then I think this "staleness problem" would go away.

davidflanagan: Thanks for testing that. Sounds like I badly misundersood how waffle.models.Switch works. I assumed that because is_active() is a method it would be checking the current setting when I called it and not just returning the value when I called get_all().

It really seems like overkill to mess with the cache and the admin interface. I'll just move the get_all() calls into the function itself. As I said, I'm just duplicating code that is already called on every single request, so I don't believe there is a perf issue we need to address.

peterbe: > just duplicating code that is already called on every single request 

according to New Relic rendering a document (i.e. the jinja dance) is, by 69%, the most time consuming function. Within, it averages 24.1 Redis get requests!! Making up ~20% of those 69%. I bet waffle flag fetching stuff is a good chunk of that. Some other day we can review that. Or just not and rewrite stuff with React and Stumptown.

kuma/api/v1/tests/test_views.py
```diff
@@ -225,7 +226,20 @@ def test_whoami_anonymous(client, api_settings, timezone):
...
     assert response['content-type'] == 'application/json'
-    assert response.json() == {
+    data = response.json()
+
+    # pytest restrictions on using the db during tests means that
+    # we don't know what flags, switches and samples we expect to
+    # receive in the response. This test assumes that waffle works
+    # correctly, so all we are doing here is validating that the
+    # response includes dicts in the expected places.
+    assert type(data['waffle']['flags']) == dict
```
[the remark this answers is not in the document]
davidflanagan: So the basic problem was that I couldn't figure out a way to get test data into those flags. Possibly that was because I was initializing too soon. Now that I call get_all() later on, maybe I can get it to work.

But since I couldn't get any data, I was just verifying that the endpoint sends waffle.flags etc, and furthermore that waffle.flags is a dict and not a string, for example.

I'll make more attempt to get some actual test waffle data working. If that fails, I'll convert to isinstance sintead of using type()

davidflanagan:
@peterbe: tests fixed. Have I addressed all of your concerns now?

PR #5388, opened 2019-05-07
Fix jest tests to work in docker

davidflanagan:
I've been using jest and prettier locally and not really testing it
in docker. Ryan worked on a front-end patch recently and tried to
run tests and formatting from docker and uncovered some compatibility
issues. This PR fixes those issues, though in one case it simply
disables a test in docker.


kuma/javascript/src/__snapshots__/page.test.js.snap
```diff
@@ -775,9 +775,9 @@ Array [
           ike
            
           <time
-            dateTime="2019-01-02T03:04:05"
+            dateTime="2019-01-02T03:04:05Z"
```
davidflanagan: This snapshot test was failing because I was feeding it a date without a timezone, and my docker container is in a different timezone than my laptop is. By forcing the input data to UTC, the problem goes away.

package.json
```diff
@@ -20,7 +20,7 @@
         "eslint": "eslint --ext .js --ext .jsx kuma/",
         "optim:svg": "svgo jinja2/includes/icons/* --disable={removeViewBox,removeTitle} --multipass --pretty",
         "precommit": "npm run eslint && npm run stylelint",
-        "test": "jest",
+        "test": "jest --testPathIgnorePatterns tests/sass",
```
davidflanagan: The sass tests use node-sass, which requires a binary. These tests were failing in docker because I have a macos binary installed locally. If we've got the container set up correctly, they would presumably run correctly on travis and/or jenkins?  For now, though, I'm just excluding the sass tests from the `npm run test` target. We can still run them locally with `npx jest`, but now `docker-compose exec web npm run test` will not fail.

@schalkneethling: any ideas about better ways to solve this?

.prettierignore
```diff
@@ -0,0 +1 @@
+kuma/javascript/src/flow-typed/
```
davidflanagan: Running `npm run format` locally leaves the js files in flow-typed/ alone. But, for whatever reason, running `docker-compose exec web npm run format` runs prettier over those files. So I've added this config file to exclude that directory from formatting.

peterbe: Can't this be added to the existing `.prettierrc` file?

kuma/javascript/src/__snapshots__/page.test.js.snap
peterbe: I'm stunned that we're using snapshots. I did try it briefly for a project but I found it counterproductive. It added more overheads and just a chore to mass-update them without even looking at the details. To me, they became a test of the tests rather than a test of the functionality. 

By the way, what I ended up doing was revert all attempts at using snapshots and instead embarked to rewrite all enzyme tests with react-testing-library instead.

davidflanagan: I'm not yet convinced by the snapshot tests. But they're basically free, and haven't been too much of a headache so far. I don't know what a good alternative is. This is an example where having the snapshot test caught a fairly subtle bug, so I'm not ready to discard them.

.prettierignore
davidflanagan: Not as far as I can tell.

peterbe: Perhaps what you *can* do is specify the location of the file that tells with files to ignore :)

PR #5389, opened 2019-05-07
bug 1525719: wiki and beta domains (part 2)

escattone:
This PR addresses the gaps identified with the current implementation of the separation of domains (`beta` and `wiki` domains) listed in https://github.com/mdn/sprints/issues/1205#issuecomment-481456076. In summary, it adds missing endpoints to the `beta` domain and updates some Content Security Policies to account for the `wiki` domain. When adding the missing endpoints to the `beta` domain, some of the changes were made to allow the views that serve those endpoints to be wrapped in different `shared_cache_control` decorators depending on whether they served the `beta` domain or the `wiki` domain.

For testing, I used my usual `.env` file:
```
DEBUG=True
DOMAIN=mdn.localhost
ATTACHMENT_HOST=demos:8000
```

### DIFF
--- a/kuma/core/decorators.py
+++ b/kuma/core/decorators.py
@@ -17,6 +17,9 @@ from .urlresolvers import reverse
 from .utils import add_shared_cache_control
 
 
+HOUR = 60 * 60
+
+
 def shared_cache_control(func=None, **kwargs):
     """
     Decorator to set Cache-Control header for shared caches like CDNs.
@@ -42,6 +45,9 @@ def shared_cache_control(func=None, **kwargs):
     if func:
         return _shared_cache_controller(func)
     return _shared_cache_controller
+
+
+beta_shared_cache_control = shared_cache_control(s_maxage=HOUR)
 
 
 def user_access_decorator(redirect_func, redirect_url_func, deny_func=None,
--- a/kuma/landing/urls.py
+++ b/kuma/landing/urls.py
@@ -7,6 +7,9 @@ from kuma.core.decorators import shared_cache_control
 from . import views
 
 
+MONTH = 60 * 60 * 24 * 30
+
+
 lang_urlpatterns = [
     url(r'^$',
         views.home,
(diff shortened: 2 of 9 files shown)


kuma/landing/urls.py
```diff
@@ -24,12 +27,12 @@
 
 urlpatterns = [
     url(r'^contribute\.json$',
-        views.contribute_json,
+        shared_cache_control(views.contribute_json),
```
davidflanagan: I think it is worth adding a comment on line 27 to explain why you're using the decorators this way as a regular function instead of as an actual decorator in views.py.

kuma/landing/views.py
```diff
@@ -118,7 +117,6 @@ def promote_buttons(request):
 '''
 
 
-@shared_cache_control
 def robots_txt(request):
```
davidflanagan: Maybe replace the decorator with a comment indicating that it is cached, but the decorators are applied in the various urls.py files?

kuma/core/decorators.py
```diff
@@ -44,6 +47,9 @@ def _cache_controlled(request, *args, **kw):
     return _shared_cache_controller
 
 
+beta_shared_cache_control = shared_cache_control(s_maxage=HOUR)
```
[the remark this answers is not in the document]
escattone: My reasoning for not creating a new setting for the `beta` domain endpoints, like `CACHE_CONTROL_DEFAULT_SHARED_MAX_AGE`, is that the `beta` domain is a transition phase, and eventually `beta_shared_cache_control` would just become `shared_cache_control` and we'd use the `CACHE_CONTROL_DEFAULT_SHARED_MAX_AGE` setting (and the current `wiki` endpoints would no longer be decorated with `shared_cache_control`).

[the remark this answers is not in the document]
escattone: I think in a separate future PR we should decide what we want to do with the use of `beta` in filenames and code. From the start, my thought has been that it would eventually be replaced with something else. We could use a different name, as you said something like `read_only` or `react`, or we could even drop it and rename all of the other URL filenames from `urls.py` to `urls_wiki.py` and the `urls_beta.py` filenames then become simply `urls.py`.

kuma/landing/views.py
escattone: My thought was that it seemed overkill to define separate views for the `beta` endpoints just so we could decorate them differently, but I'm fine with doing that way as well.

kuma/core/decorators.py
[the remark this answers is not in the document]
peterbe: Oh I get it! Once the dust settles, the wiki site will be vastly simplified as it'll just be a plain Django app without any fancy attempts to CDN and stuff.

kuma/landing/views.py
peterbe: You do as you want. For me, it's not a blocker, but just a shared feeling/opinion.

PR #5390, opened 2019-05-09
use -env_file in docker-compose

peterbe:
What's remaining now is that this'll break Jenkins. Somewhere in the Groovy code, we need to do something similar to how we, now in this PR, do for Travis. i.e. `cp .env.travis .env`. 
@escattone is forcibly volunteering to make a suggestion to the Groovy code so that it can start `docker-compose` without failing on a missing `.env` file.


peterbe:
@escattone mentioned that something similar to `cp .env.travis .env` would need to happen for Jenkins. E.g. this file https://github.com/mozilla/kuma/blob/master/Jenkinsfiles/utils.groovy

peterbe:
Actually, @escattone would you mind making a PR against my PR to show how to do the `touch .env` in groovy?

peterbe:
The [documentation already says for people to create their on `.env` file](https://kuma.readthedocs.io/en/latest/installation.html#docker-setup)

peterbe:
@escattone The best thing about this is that it'll calm down the PYTHONWARNINGS which keep blathering about `/app/kuma/core/i18n.py:18: ImportWarning: Not importing directory '/app/kuma/core/jinja2': missing __init__.py`

callahad:
> @escattone would you mind making a PR against my PR to show how to do the touch `.env` in groovy?

@peterbe it'd just be this, right?

```diff
diff --git a/Jenkinsfiles/utils.groovy b/Jenkinsfiles/utils.groovy
index 5fc2aa277..71cbedfb1 100644
--- a/Jenkinsfiles/utils.groovy
+++ b/Jenkinsfiles/utils.groovy
@@ -242,6 +242,7 @@ def compose_test() {
     def es_url = 'http://elasticsearch:9200'
     def mysql_url = 'mysql://root:kuma@mysql:3306/developer_mozilla_org'
     def junit_results = '--junit-xml=/app/test_results/django.xml'
+    sh_with_notify("touch .env", 'Create an empty .env file')
     sh_with_notify("${dc} --version", 'Check the version')
     sh_with_notify(dc_down, 'Pre-test tear-down')
     sh_with_notify("${dc} up -d", 'Start the services detached')
```

Do we need to fill that `.env` file with anything in particular, like we do for Travis (`cp .env.travis .env`)?

peterbe:
> Do we need to fill that `.env` file with anything in particular, like we do for Travis (`cp .env.travis .env`)?

I suspect that question is directed towards @escattone but I suspect the answer is that instead of `touch .env` it should reuse the same stuff as used in `.travis.yml`. Right? 
I.e. `sh_with_notify("cp .env.travis .env", 'Create an .env file')` instead of `sh_with_notify("touch .env", 'Create an empty .env file')`.
I should probably remember this by now, but we run `docker-compose` to run tests in Travis AND Jenkins right? If this is the case should we change that default from `.env.travis` to `.env.ci` because it's not unique to TravisCI.

peterbe:
I rebased this and it still seems to work just fine. It's a joy to use this branch because it hides all the  Dostoevsky'esque walls of deprecation warnings which we'll keep getting until we upgrade to Django 2.

alshamiri5 (approved):

peterbe:
too old.

PR #5391, opened 2019-05-09
Add therapist as a pre-commit hook

MatonAnthony:
Therapist allows us to run linting on modified files before committing.
This is done to limit the amount of error in CI because of linting.

This therapist configuration lint:
    - "js" and "jsx" files using eslint
    - "css" and "scss" files using stylelint
    - "py" files using flake8

**Note: this needs a bugzilla bug number before merging.** 
**Note 2: If we think this is a good idea, I need to write a chunk of documentation about it for kuma.readthedocs.io**

_There has not been a decision about it as far as I am aware, but I found it interesting and it works quite well locally for me, so I thought it was worth talking about it._

@peterbe This was a wonderful idea imho.

### DIFF
--- a/.therapist.yml
+++ b/.therapist.yml
@@ -1,0 +1,26 @@
+actions:
+  flake8:
+    run: flake8 {files}
+    include: "*.py"
+
+  eslint:
+    run: $(npm bin)/eslint {files}
+    fix: $(npm bin)/eslint --fix {files}
+    include:
+      - "*.js"
+      - "*.jsx"
+
+  stylelint:
+    run: $(npm bin)/stylelint {files}
+    fix: $(npm bin)/stylelint --fix {files}
+    include:
+      - "*.css"
+      - "*.scss"
+
+shortcuts:
+  all:
+    flags:
+      - use-tracked-files
+      - include-untracked
+    options:
+      action: flake8


MatonAnthony:
The build error is not related to my pull request.

peterbe:
You forced my hand to escalate writing my first ADR: https://github.com/mdn/mdn/pull/72 :)

I haven't actually studied your first attempt of a `.therapist.yml` file but once the ADR is agreed upon I'd love to dig into this more in detail and make sure it works correctly in Travis. 

And as you said, this PR would need to be accompanied with the necessary changes to the development docs.

peterbe:
By the way, I think we should consider moving all linting OUT of docker. It's perhaps convenient to be able to run `docker-compose exec web make lint` and have it all be done in there instead of having to rely on external tools. 

But linting doesn't make sense going into the docker container. `flake8`, `flake8-import-order`, `pyflakes`, `codestyle`, `entrypoints`, `pycodestyle` and soon possibly also `black`. That's 6 packages already that we could stop having to wait for. 

Besides linting is only really applicable in two places:

* When doing local dev and about to prepare a pull request. 

* In CI

In TravisCI we'd just have another node in the matrix that does a linting check and it can run in parallel to the main tests. And if it lives outside of docker it integrates really well with the command line and an installed git pre-commit hook (or pre-push maybe). 

Also, with vscode it needs to know where the flake8 executable is and it can't be inside docker. 

I propose we extract `flake8` (and its dependencies) from `requirements/dev.txt` and then create a new file called something like `requirements/lint.txt` which you install locally in a virtualenv or something. 

We can still execute all of this with docker-compose without needing to have `Dockerfile` install it in the container. Mind you, what I had in mind is a tool that triggers `pip install -r requirements/lint.txt` inside docker.

peterbe:
This is too old and too much has changed. We've concluded that it's best not to put any of this into the `master` branch. Perhaps can share some best practices in the Kuma documentation as tips for developers who work with the repo a lot.

PR #5395, opened 2019-05-09
Bug 1525719: handle request locale more cleanly in beta pages

davidflanagan:
This patch creates a cleaner separation between document data and
request-related data in our code that does server-side rendering
and no longer mixes requestLocale into the document data object.

On the client-side locale is now handled with a LocaleProvider
component and corresponding context object.

### DIFF
--- a/kuma/javascript/src/document-provider.jsx
+++ b/kuma/javascript/src/document-provider.jsx
@@ -1,6 +1,8 @@
 // @flow
 import * as React from 'react';
-import { useState, useEffect } from 'react';
+import { useContext, useEffect, useState } from 'react';
+
+import LocaleProvider from './locale-provider.jsx';
 
 export type DocumentData = {
     locale: string,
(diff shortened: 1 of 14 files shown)


davidflanagan:
@escattone : Following up on your recent PR, this is my PR for fixing up the requestURL thing that we discussed. The changes on the server side are straightforward. The changes on the client side turned out to be bigger than I expected. I'm quite confident about the client-side changes, however, so don't feel like you need to review in any great detail.

kuma/javascript/src/document-provider.jsx
```diff
@@ -64,10 +48,11 @@ export default function DocumentProvider(
             throw new Error('DocumentProvider effect ran without body.');
         }
         const body = document.body;
+
         // This is the function that does client side navigation
-        function navigate(url, localeAndSlug) {
+        function navigate(url, slug) {
```
davidflanagan: A related change in this PR is that we never do client-side navigation if the locale is changing. So I can treat locale as a constant here, and just track the slug in the history mechanism. That's what a lot of the changes in this file are about.

kuma/wiki/views/document.py
```diff
@@ -877,13 +877,18 @@ def react_document(request, document_slug, document_locale):
...
     document_data = doc_api_data['documentData']
 
-    # And add another property to specify the requested locale
-    # (which may not be the same as the locale of the document)
-    document_data['requestLocale'] = document_locale
+    # Create another object of request-related data to pass to jinja
+    request_data = {
+        # Despite its name, the document_locale argument is the locale
+        # of the request, and is not always (in the case of missing
+        # translations) the same as the locale of the document.
```
escattone: Nice comment.

kuma/javascript/src/document-provider.jsx
```diff
@@ -157,24 +137,35 @@ export default function DocumentProvider(
...
+            // If the link is not to a /docs/ URL then we can't handle
+            // it via client-side navigation.
             if (parts[2] !== 'docs') {
                 return;
             }
 
+            // If the locale in the URL is not the same as the current
+            // locale then we can't do client-side navigation because
+            // we don't have the right strings loaded.
+            if (parts[1] !== locale) {
```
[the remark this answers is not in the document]
escattone: Wait, no I think I'm wrong here. We could still be on a Swahili page (`/sw/docs/...`) and that page theoretically could have a link to another document that doesn't exist in Swahili. I say theoretically because no example of this case may actually exist. I haven't found one yet.

PR #5396, opened 2019-05-09
test messing with package.json without cli

peterbe:
This is just a test. 

Here I manually edited the `package.json` without doing the legwork of updating the `package-lock.json` too. This is naughty and devs should use the cli to change packages to `npm` gets a chance to update `package-lock.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
         "react-svg-loader": "^2.1.0",
         "react-test-renderer": "^16.8.4",
         "sass-true": "4.0.0",
-        "stylelint": "9.3.0",
+        "stylelint": "9.4.0",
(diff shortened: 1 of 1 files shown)


peterbe:
I'm expecting this PR to fail. 

According to [the docs](https://docs.npmjs.com/cli/ci.html) ...

 > "If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock."

That's what I'm trying to prove. I think TravisCI automatically runs `npm ci`.

peterbe:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1546484#c9

PR #5397, opened 2019-05-10
bug 1548914: Run jest tests in Travis CI

MatonAnthony:
Add a matrix to Travis to run Jest tests in Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,13 @@ services:
 matrix:
     include:
         - language: python
+          name: "Documentation"
           python: 3.6
           cache: pip
           script:
             - scripts/ci-docs
         - language: python
+          name: "Python (lint + test)"
           python: 2.7
           env:
             UID=0
@@ -29,7 +31,14 @@ matrix:
             - scripts/ci-localerefresh
             - bash <(curl -s https://codecov.io/bash)
         - language: node_js
+          name: "Node (lint)"
           node_js: 10
           cache: npm
           script:
             - scripts/ci-jslint
+        - language: node_js
+          name: "Node (test)"
+          node_js: 10
+          cache: npm
+          script:
+            - scripts/ci-jest
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
           script:
             - scripts/ci-docs
         - language: python
-          name: "Python (lint + test)"
+          name: "Tests + Python linting"
           python: 2.7
           env:
             UID=0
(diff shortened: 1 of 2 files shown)


MatonAnthony:
The jest failure also happens locally.

MatonAnthony:
@davidflanagan @schalkneethling I think you are going to be the more interested in this bug :)

scripts/ci-jest
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e  # Exit on non-zero status
+set -u  # Treat unset variables as an error
+
+# Run test based on what's defined in `package.json`
+npm test
```
peterbe: Why are these run outside of Docker? 
I suspect this line should be `docker-compose exec -T web npm test`. Right @davidflanagan ?

MatonAnthony: @davidflanagan said in the Therapist ADR that Jest tests were 10x slower in Docker than on its local host.

I don't see any reasons why it should run inside the container, considering that Travis installs the npm packages anyway and that the result are identical.

peterbe: No, we want to unit test with the exact version of `node` (and `npm`). 

I think that's different for linting because any outcomes from linting does not affect runtime really. 

"Power users" of kuma dev might want to "cheat" and use their much faster host environment if they can but once they're done, the tests should be run in the correct environment.

MatonAnthony: The node version can be defined in travis.yml.

I don't think the `npm` version is of any importance, the worse that can happen is the build doesn't start at all.

I think it does not bring any added value to run it in Docker.

But ultimately it is your decision, if you would like me to change it, I can

peterbe: Would you mind moving it? Unit tests are very relevant to runtime (unlike linting).

MatonAnthony: Ok.

.travis.yml
```diff
@@ -29,7 +31,14 @@ matrix:
...
           node_js: 10
           cache: npm
           script:
             - scripts/ci-jslint
+        - language: node_js
+          name: "Node (test)"
+          node_js: 10
+          cache: npm
+          script:
+            - scripts/ci-jest
```
peterbe: Move this line to be between line 30 and 31. And remove this whole matrix node. It's not actually "python" (although it says that on line 17. It's just Python as the foundation before starting docker-compose.

.travis.yml
```diff
@@ -9,11 +9,13 @@ services:
 matrix:
     include:
         - language: python
+          name: "Documentation"
           python: 3.6
           cache: pip
           script:
             - scripts/ci-docs
         - language: python
+          name: "Python (lint + test)"
```
peterbe: Rename this to "Tests + Python linting" (or however you want to spell it) and it'll be accurate.

.travis.yml
MatonAnthony: Aren't we going to loose the ability to run those in parallel ?

peterbe: We are, but the "python block" already does all the docker-compose stuff. Remember, docker-compose isn't just python stuff. For example, you need to run the webpack build so that Django has the file it needs. 

We *could* do the docker-compose stuff too, in parallel, in this block but it'd probably take longer in total.

davidflanagan:
Thanks @peterbe. I'm good with whatever you and @MatonAnthony have come up with. Thanks for working on this, Anthony!

PR #5402, opened 2019-05-10
Bug 1525719: move user and waffle-related ga() calls to client side

davidflanagan:
We currently hardcode a bunch of ga() analytics calls into the
HTML rendered by our Jinja templates, even when the data in question
depends on the current user or the current waffle flags. This means
that the HTML we generate can't be cached for re-use.

This PR modifies the react_document.html template so that our
React-based beta pages are cacheable and do not reference the
current user or current waffle flags. Instead, user and waffle
related ga() calls are now made on the client side in user-provider.jsx
when that data is obtained from the /whoami API.

This PR also adds ga('send', 'pageload') calls for client-side navigation.

### DIFF
--- a/jinja2/includes/config.html
+++ b/jinja2/includes/config.html
@@ -5,20 +5,23 @@
     (function(win) {
         'use strict';
 
-        {{ waffle.wafflejs() }}
-
+        {#- This is stuff not needed by the React-based site #}
         {%- if not beta %}
-        // This needs to be set before ckeditor.js loads
-        {%- if settings.CKEDITOR_DEV %}
-          {%- set ckeditor_basepath = static('js/libs/ckeditor4/source/ckeditor/') %}
-        {%- else %}
-          {%- set ckeditor_basepath = static('js/libs/ckeditor4/build/ckeditor/') %}
-        {%- endif %}
-        window.CKEDITOR_BASEPATH = '{{ ckeditor_basepath }}';
+            {#- With React, waffle flags are returned by the /whoami API. #}
+            {{ waffle.wafflejs() }}
 
-        // Site configuration
-        win.mdn.ckeditor = {};
+            // This needs to be set before ckeditor.js loads
+            {%- if settings.CKEDITOR_DEV %}
+              {%- set ckeditor_basepath = static('js/libs/ckeditor4/source/ckeditor/') %}
+            {%- else %}
+              {%- set ckeditor_basepath = static('js/libs/ckeditor4/build/ckeditor/') %}
+            {%- endif %}
+            window.CKEDITOR_BASEPATH = '{{ ckeditor_basepath }}';
+
+            // Site configuration
+            win.mdn.ckeditor = {};
         {%- endif %}
+
         win.mdn.features = {};
         win.mdn.siteUrl = '{{ settings.SITE_URL }}';
         win.mdn.wikiSiteUrl = '{{ settings.WIKI_SITE_URL }}';
--- a/kuma/javascript/src/app.jsx
+++ b/kuma/javascript/src/app.jsx
@@ -2,6 +2,7 @@
 import * as React from 'react';
 
 import DocumentProvider from './document-provider.jsx';
+import GAProvider from './ga-provider.jsx';
 import LocaleProvider from './locale-provider.jsx';
 import Page from './page.jsx';
 import UserProvider from './user-provider.jsx';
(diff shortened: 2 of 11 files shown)


jinja2/includes/config.html
```diff
@@ -5,20 +5,23 @@
...
         {%- if not beta %}
-        // This needs to be set before ckeditor.js loads
-        {%- if settings.CKEDITOR_DEV %}
-          {%- set ckeditor_basepath = static('js/libs/ckeditor4/source/ckeditor/') %}
-        {%- else %}
-          {%- set ckeditor_basepath = static('js/libs/ckeditor4/build/ckeditor/') %}
-        {%- endif %}
-        window.CKEDITOR_BASEPATH = '{{ ckeditor_basepath }}';
+            {#- With React, waffle flags are returned by the /whoami API. #}
+            {{ waffle.wafflejs() }}
```
davidflanagan: This waffle.wafflejs() used to be outside the if not beta block. Now it is inside, since we don't want to cache waffle flags. Note that the changed lines below are just indentation changes.

kuma/javascript/src/app.jsx
```diff
@@ -14,12 +15,14 @@ type Props = { documentData: DocumentData, requestData: RequestData };
...
 export default function App(props: Props) {
     return (
-        <LocaleProvider locale={props.requestData.locale}>
-            <DocumentProvider initialDocumentData={props.documentData}>
-                <UserProvider>
-                    <Page />
-                </UserProvider>
-            </DocumentProvider>
-        </LocaleProvider>
+        <GAProvider>
```
davidflanagan: A new component that provides the ga() function to any components that want to send data to google analytics. If google analytics is not configured, the provider will just provide a dummy function that doesn't do anything.

kuma/javascript/src/locale-provider.test.js
```diff
@@ -0,0 +1,23 @@
...
+                <Consumer>{contextConsumer}</Consumer>
+            </LocaleProvider>
+        );
+
+        // We expect the Consumer to get the locale value from the provider
+        // and pass it to the contextConsumer function.
+        expect(contextConsumer.mock.calls.length).toBe(1);
+        expect(contextConsumer.mock.calls[0][0]).toEqual('foo-bar');
+    });
+});
```
davidflanagan: locale-provider.jsx was defined in my last PR, but I forgot to write a test for it then.

kuma/javascript/src/document-provider.jsx
```diff
@@ -89,6 +92,21 @@ export default function DocumentProvider(
...
+                        // If the document data includes enSlug, or if the
+                        // document is in english, then pass the slug to
+                        // google analytics as dimension 17.
+                        if (documentData.enSlug) {
+                            ga('set', 'dimension17', documentData.enSlug);
+                        } else if (documentData.locale === 'en-US') {
+                            ga('set', 'dimension17', documentData.slug);
+                        }
+
+                        ga('send', 'pageload');
```
escattone: I know very little about how we use `ga` events, so this is just a double check. You send a `pageview` event in the user provider component and in a comment within that component, you mention sending a `pageview` event in the document provider component, but here you send a `pageload` event. I assume that's what you wanted, but just wanted to make sure.

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -31,7 +31,37 @@
...
+        {#- TODO: delete or move to the client side #}
+        {%- if analytics_page_revision %}
+            // dimension12 == 'Page Revision'
+            ga('set', 'dimension12', '{{ analytics_page_revision }}');
+        {%- endif %}
+
+        {#- TODO: delete or move to the client side #}
+        {%- if analytics_en_slug %}
+            // dimension17 == 'English Slug'
+            ga('set', 'dimension17', '{{analytics_en_slug}}');
```
escattone: This is for sending `dimension17` for full page loads only right?

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -31,7 +31,37 @@
...
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+        })(window,document,'script','https://www.google-analytics.com/{{ GA_FILE }}.js','ga');
+        ga('create', '{{ config.GOOGLE_ANALYTICS_ACCOUNT }}', 'mozilla.org');
+        ga('set', 'anonymizeIp', true);
+
+        {#- TODO: delete or move to the client side #}
+        {%- if analytics_page_revision %}
+            // dimension12 == 'Page Revision'
+            ga('set', 'dimension12', '{{ analytics_page_revision }}');
```
escattone: I noticed that we're not setting this dimension for client-side navigation. Should we be?

kuma/wiki/jinja2/wiki/react_document.html
davidflanagan: Yes. I think I need to delete the TODO comment.

kuma/wiki/jinja2/wiki/react_document.html
davidflanagan: Kadir said he doesn't need it anymore after I wrote that TODO comment. So I should probably just delete it here for simplicity. Thanks for pointing this out.

kuma/javascript/src/document-provider.jsx
davidflanagan: I'll double-check and make sure I'm using the right string in both the comments and the code. Thanks for catching that!

PR #5404, opened 2019-05-13
react-svg-loader 3.0.1

peterbe:
[Changelog here](https://github.com/boopathi/react-svg-loader/blob/master/CHANGELOG.md#301) which basically points to an upgrade of `babel-plugin-react-svg` to 3.0.0 and `react-svg-core"` to 3.0.0. (along with some irrelevant changes of requiring at least node 8).


peterbe:
I've clicked around on the beta site and can't see any visual difference.

PR #5410, opened 2019-05-14
be more careful with PyQuery constructor

peterbe:


kuma/core/utils.py
```diff
@@ -458,3 +459,43 @@ def override_urlconf(new_urlconf):
...
+            args = (' {}'.format(args[0]),) + args[1:]
+    elif isinstance(args[0], str):
+        # If the input is a byte string, deal with it a byte string.
+        # Since this file is using __future__.unicode_literals and
+        # type and quoted string automatically becomes a unicode string.
+        # In this case force it all to stay as byte strings.
+        if args[0].split(force_bytes('://'), 1)[0] in (force_bytes('http'), force_bytes('https')):
+            args = (force_bytes(' ') + args[0],) + args[1:]
+
+    return pq(*tuple(args), **kwargs)
```
escattone: We can toss the `tuple`  call here. I think you just forgot to remove it.

kuma/scrape/sources/revision.py
```diff
@@ -171,3 +172,4 @@ def extract_data(self, html):
         data['content'] = source_elem.text
 
         return data
+        return data
```
escattone: Typo.

PR #5411, opened 2019-05-14
bug 1551701 - upgrade redis and django-redis

peterbe:
I need to tie this to a bug ID.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -162,9 +162,9 @@ django-redirect-urls==1.0 \
 # Provide a caching provider to use Redis as a cache in Django
 # Code: https://github.com/niwinz/django-redis
 # Docs: http://niwinz.github.io/django-redis/latest/
-django-redis==4.9.0 \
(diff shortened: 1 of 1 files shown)


escattone:
The upgrade to Redis 3 is going to have to be reverted. We're still using Celery 3 and Celery 3 doesn't support Redis 3. For example, as the stack trace above shows, Celery 3's Kombu package uses the Redis 2.x call signature for `zadd`:
```python
pipe.zadd(self.unacked_index_key, delivery_tag, time())
```
which ultimately leads to the `AttributeError: 'str' object has no attribute 'iteritems'` error. Redis 3 expects the call signature to be something like:
```python
pipe.zadd(self.unacked_index_key, {delivery_tag: time()})
```

See https://github.com/celery/kombu/blob/c1b1010daa6c8cee35230ce2f73b3f2c3bb7b10b/kombu/transport/redis.py#L150.

This PR is causing the Celery worker to fail. I'm going to submit a new PR that reverts back to Redis 2.

PR #5414, opened 2019-05-15
Bug 1525719: implement the task completion survey in React

davidflanagan:
This PR ports the task completion survey code task-completion.js
to a React component in task-completion-survey.jsx


davidflanagan:
Schalk is probably the right reviewer for this, since it is a non-trivial example of porting old-style JS code to a new React-based version.  But I've also tagged Ryan on it so he's aware of it.

@schalkneethling it was easy to duplicate the notification dialog in React, but it was surprisingly difficult to get the react hooks just right to handle the google analytics stuff and the code that hides notifications for a month. But I think I ended up with something that works well. 

Note that now that we have client-side navigations we have to decide how to deal with waffle flags. Currently we're getting them only on the first load, so if a notification is shown and the user navigates without dismissing it, we'll just keep showing the notification until they dismiss it or follow the link.  I think this is a reasonable thing to do, but let me know if you disagree.

Also: in order to implement this the React way, I had to break the string of text that we show up into three pieces, each of which will have to be translated separately. I started off trying to use the existing string so I could use the existing translations, but the could would have ended up being really hacky.

davidflanagan:
I should add that I intentionally made no attempt to port the mdn.Notifier utility for this PR, and instead just duplicated the part of it that was needed for this one feature. If and when we port other features over that use mdn.Notifier, we can write a more general utility and refactor this task-completion-survey.jsx component, if that looks like it makes sense at the time. For now though, I'm taking the approach of only porting over the minimal set of features.

kuma/javascript/src/icons/close.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="28" viewBox="0 0 24 28"><path d="M20.281 20.656c0 0.391-0.156 0.781-0.438 1.062l-2.125 2.125c-0.281 0.281-0.672 0.438-1.062 0.438s-0.781-0.156-1.062-0.438l-4.594-4.594-4.594 4.594c-0.281 0.281-0.672 0.438-1.062 0.438s-0.781-0.156-1.062-0.438l-2.125-2.125c-0.281-0.281-0.438-0.672-0.438-1.062s0.156-0.781 0.438-1.062l4.594-4.594-4.594-4.594c-0.281-0.281-0.438-0.672-0.438-1.062s0.156-0.781 0.438-1.062l2.125-2.125c0.281-0.281 0.672-0.438 1.062-0.438s0.781 0.156 1.062 0.438l4.594 4.594 4.594-4.594c0.281-0.281 0.672-0.438 1.062-0.438s0.781 0.156 1.062 0.438l2.125 2.125c0.281 0.281 0.438 0.672 0.438 1.062s-0.156 0.781-0.438 1.062l-4.594 4.594 4.594 4.594c0.281 0.281 0.438 0.672 0.438 1.062z"/></svg>
```
schalkneethling: You can remove the `width` and `height` attributes here as the `viewBox` is perfect on its own. I would also suggest that you add `role="presentation"` for accessibility. You then have two options with regards to giving this a title.

1. Add an `aria-label` attribute, with a text string such as "Close dialog", to the anchor element or button that wraps this. Then also add an `id` to the same element. Then, add an `aria-labelledby` to the SVG and set its value to that of the `id`.
2. Just before the first `path` element add `<title id="close-icon-title">Close dialog</title>`. On the SVG tag add `aria-labelledby="close-icon-title"`.

kuma/javascript/src/task-completion-survey.jsx
```diff
@@ -0,0 +1,177 @@
...
+const EMBARGO_TIMESTAMP = 'taskTracker';
+
+// This is how long the embargo lasts
+const EMBARGO_LENGTH = 1000 * 60 * 60 * 24 * 32; // 32 days
+
+// If the user dismisses or responds to the survey, then we call this
+// function to set the embargo timestamp. During this embargo period,
+// the notification will not be shown again to the user.
+function embargo() {
+    localStorage.setItem(
```
schalkneethling: Wondering whether it would be worth wrapping this in a `try/catch`? Not much we can do if it fails unless we want to fall back to using a cookie. I generally tend to do it, even if only to report the errors to myself in the `console` ;) There is also the issue with Safari in private mode - https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/

kuma/javascript/src/task-completion-survey.jsx
```diff
@@ -0,0 +1,177 @@
...
+        !isEmbargoed(); // and it is not embargoed
+
+    useEffect(() => {
+        if (shouldShow) {
+            // If we are showing the notification and asking the user to
+            // complete the survey, then tell Google Analytics about it.
+            // GA dimension14 is "Saw Survey Gizmo Task Completion survey".
+            //
+            // Note that since this code runs in useEffect() it only
+            // runs once each time that shoudShow becomes
```
schalkneethling: Minor typo:

```suggestion
            // runs once each time that shouldShow becomes
```

kuma/javascript/src/task-completion-survey.jsx
```diff
@@ -0,0 +1,177 @@
...
+                eventCategory: 'survey',
+                eventAction: 'prompt',
+                eventLabel: '',
+                eventValue: 'impression',
+                nonInteraction: true
+            });
+        }
+    }, [shouldShow]);
+
+    // If we're not going to show anything, there is nothing to display now
```
schalkneethling:
```suggestion
    // If we're not going to show anything, there is nothing to do further
```

kuma/javascript/src/task-completion-survey.jsx
```diff
@@ -0,0 +1,177 @@
...
+// This is how long the embargo lasts
+const EMBARGO_LENGTH = 1000 * 60 * 60 * 24 * 32; // 32 days
+
+// If the user dismisses or responds to the survey, then we call this
+// function to set the embargo timestamp. During this embargo period,
+// the notification will not be shown again to the user.
+function embargo() {
+    localStorage.setItem(
+        EMBARGO_TIMESTAMP,
+        String(Date.now() + EMBARGO_LENGTH)
```
peterbe: Is the `String()` really necessary? E.g.

```
> sessionStorage.foo = "4"
"4"
> sessionStorage.foo > 3
true
> sessionStorage.foo > 5
false
```

The browsers will force everything to a string anyway and JS will automatically cast strings to ints when you use the `>` operator?

Not trying to change anything, just curious to learn.

davidflanagan: String() is probably not necessary, but:

- flow might complain if I don't use it (I haven't checked, though)
- the original specs for localStorage said that they'd allow non-string values and serialize them in the way that things like postMessage() do. Browsers never implemented that, and maybe the specs have now changed to say that it is string only, but I'm still pissy about that. Forcing to String means that things won't break if the implementation later starts accepting other types
- using String() here helps to make it clear why we're using parseInt() when we later read the data out of localStorage

kuma/javascript/src/task-completion-survey.jsx
davidflanagan: Thanks, Schalk! I just checked that localStorage was supported in IE11 and figured that if it failed, it would fail silently. So I assumed that our current check whether it is supported was no longer necessary.

I'll probably just add try/catch here, and then consider converting to use a more generic storage wrapper class if needed later.

kuma/javascript/src/icons/close.svg
davidflanagan: This is perfect feedback. I knew I was leaving this out, but didn't know the right way to handle it. Thanks.

davidflanagan: Since I'm using role="presentation", I'd expect that to mean that the svg will just be ignored and not need to be labelled. So can't I just put the aria-label on the button and leave it at that?  (I'm updating the PR to do that, but you can tell me if I really need the id and aria-labelledby

schalkneethling: Sure, if the button has `aria-label` that will do. Thanks @davidflanagan

davidflanagan:
@schalkneethling I've addressed your review comments and have added a detailed test for the TaskCompletionSurvey component.

davidflanagan:
Here's what it looks like, by the way:

Screen Shot 2019-05-15 at 8 33 58 PM

And here is the version that is currently on developer.mozilla.org:

Screen Shot 2019-05-15 at 8 35 04 PM

kuma/javascript/src/task-completion-survey.jsx
```diff
@@ -0,0 +1,193 @@
...
+                {gettext(
+                    'and fill it out when you are done on the site. Thanks!'
+                )}
+            </div>
+            <button
+                css={styles.dismissButton}
+                onClick={dismiss}
+                aria-label={gettext('Dismiss')}
+            >
+                <CloseIcon css={styles.dismissIcon} role="presentation" />
```
schalkneethling: @davidflanagan - Is there a reason to not add the `role` directly to the SVG? Seems unlikely that we will ever use the SVG in a scenario where the `role` will not be valid.

kuma/javascript/src/task-completion-survey.test.js
```diff
@@ -0,0 +1,222 @@
...
+                    <DocumentProvider.context.Provider value={mockDocumentData}>
+                        <UserProvider.context.Provider value={mockUserData}>
+                            <TaskCompletionSurvey />
+                        </UserProvider.context.Provider>
+                    </DocumentProvider.context.Provider>
+                </LocaleProvider.context.Provider>
+            </GAProvider.context.Provider>
+        );
+
+        // It should not render anything
```
schalkneethling: I believe this comment is incorrect here?

kuma/javascript/src/task-completion-survey.jsx
davidflanagan: I tried that first, but the role attribute was getting stripped out, presumably by svg-react-loader. I didn't investigate why it was happening, and just move the role attribute here instead.

kuma/javascript/src/task-completion-survey.test.js
davidflanagan: Oops. I'll fix that.

davidflanagan: Fixed

kuma/javascript/src/task-completion-survey.jsx
davidflanagan: react-svg-loader invokes svgo, and it appears that svgo is stripping out the role attribute on the svg element.

davidflanagan: Despite that fact that react-svg-loader and svgo have terrible documentation, I was able to figure out how to pass svgo config settings to react-svg-loader in the webpack config file, and I disabled the svgo plugin that was stripping the role attribute, so I'm able to remove it here.

davidflanagan:
I somehow ended up with submodule changes in one of my commits. I've fixed that, and in the process ended up rebasing and squashing, so there's just one commit now. This is ready again for review, @schalkneethling

kuma/javascript/src/ga-provider.jsx
```diff
@@ -1,12 +1,8 @@
...
 import * as React from 'react';
+import { useContext, useEffect, useState } from 'react';
 
-type GAOptions = { [string]: any };
-type GAFunction = (
-    string,
-    string | GAOptions,
-    ...Array<string | number | GAOptions>
-) => void;
+type GAFunction = (...any) => void;
```
davidflanagan: It turned out to be too hard to accurately capture the type of the ga() function, and trying to do so was causing more flow problems than it was worth. So I took this easy way out.

PR #5415, opened 2019-05-15
make scrape_document accept multiple urls

peterbe:
I use this tool surprisingly often and it's so annoying to have to start it with one URL at a time. 

Also, in a perfect world, I'd rename the management command to be called `scrape_documents` instead but I didn't want "rock the boat" too much.

### DIFF
--- a/kuma/scrape/management/commands/scrape_document.py
+++ b/kuma/scrape/management/commands/scrape_document.py
@@ -1,4 +1,4 @@
-"""Scrape a wiki document from a Kuma website."""
+"""Scrape a wiki document(s) from a Kuma website."""
 from __future__ import unicode_literals
 
 from django.core.management.base import CommandError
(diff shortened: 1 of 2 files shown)


kuma/scrape/management/commands/scrape_document.py
```diff
@@ -33,19 +34,20 @@ def add_arguments(self, parser):
...
-            self.stderr.write('Document "%s" already exists. Use --force to update.' % path)
+        for url in options['urls']:
+            host, ssl, path = self.parse_url_or_path(url)
+            scraper = self.make_scraper(host=host, ssl=ssl)
+
+            params = {}
+            for param in ('force', 'translations', 'revisions', 'depth'):
+                if options[param]:
+                    params[param] = options[param]
+            scraper.add_source("document", path, **params)
```
jwhitlock: Instead of running multiple scrapers, I would add multiple sources to the single scraper.

peterbe: The `self.make_scraper(...)` call depends on the URL. You *could* do something like:

```python
host, ssl, path = self.parse_url_or_path(urls[0])
scraper = self.make_scraper(host=host, ssl=ssl)
for url in options['urls']:
    host, ssl, path = self.parse_url_or_path(url)
    ...
    scraper.add_source("document", path, **params)
    ...
```

But that feels clunky and it's not a huge cost with creating the scraper since the use case is running this rarely on your laptop. It's rare that you run `scrape_document` and it's even more rare that you have more than 1 URLs.

PR #5416, opened 2019-05-15
bug 1551968 - Exclude noindex documents from sitemaps

peterbe:
Regarding [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1551968#c1) I put in some lines like this:

```python
queries_before = len(connection.queries)

...

queries_after = len(connection.queries)
print("#QUERIES:", queries_after - queries_before)
```

inside the `build_locale_sitemap` function. Now, when I run `docker-compose exec web ./manage.py make_sitemaps` the print output looks something like this:

```
...
('#QUERIES:', 3)
('#QUERIES:', 3)
('#QUERIES:', 3)
('#QUERIES:', 2)
('#QUERIES:', 3)
('#QUERIES:', 3)
('#QUERIES:', 3)
('#QUERIES:', 2)
...
```

(it's one for every `settings.LANGUAGES`)

*Before*, with the `master` branch the same instrumentation looked like this:

```
('#QUERIES:', 26)
('#QUERIES:', 17)
('#QUERIES:', 23)
('#QUERIES:', 17)
('#QUERIES:', 36)
('#QUERIES:', 1)
('#QUERIES:', 18)
('#QUERIES:', 36)
('#QUERIES:', 25)
('#QUERIES:', 1)
('#QUERIES:', 14)
('#QUERIES:', 36)
('#QUERIES:', 6)
('#QUERIES:', 1)
('#QUERIES:', 1)
('#QUERIES:', 17)
('#QUERIES:', 21)
```

And **note**, this is using my dev mysql. In Prod we'd get linearly many more queries that matches how many matched documents we have per locale. This PR makes that stable instead. It's a marginally more expensive SQL query but that's many times faster than doing *many* SQL queries.

### DIFF
--- a/kuma/wiki/tests/test_tasks.py
+++ b/kuma/wiki/tests/test_tasks.py
@@ -1,15 +1,27 @@
 from __future__ import with_statement
 
 import os
+import random
 import re
 from datetime import datetime
+
+from six.moves.urllib.parse import urlparse
 
 from kuma.core.urlresolvers import reverse
 from kuma.users.models import User
 from kuma.users.tests import UserTestCase
+from kuma.wiki.constants import (
+    EXPERIMENT_TITLE_PREFIX,
+    LEGACY_MINDTOUCH_NAMESPACES
+)
 from kuma.wiki.templatetags.jinja_helpers import absolutify
 
-from ..models import Document, DocumentDeletionLog, DocumentSpamAttempt
+from ..models import (
+    Document,
+    DocumentDeletionLog,
+    DocumentSpamAttempt,
+    Revision
+)
 from ..tasks import (build_sitemaps,
                      delete_logs_for_purged_documents,
                      delete_old_documentspamattempt_data)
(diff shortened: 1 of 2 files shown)


kuma/wiki/tests/test_tasks.py
```diff
@@ -67,6 +78,115 @@ def test_sitemaps(tmpdir, settings, doc_hierarchy):
...
+            with open(os.path.join(root, file)) as f:
+                content = f.read()
+                all_locs.extend(re.findall('<loc>(.*?)</loc>', content))
+
+    # Exclude the inter-linking sitemaps
+    all_locs = [loc for loc in all_locs if not loc.endswith('.xml')]
+
+    # The all_locs will have one for each locale but we definitely shouldn't
+    # have any of the slugs that we know shouldn't be included
+    assert not [loc for loc in all_locs if legacy_slug in loc]
```
escattone: I would vote for removing not only line 178 (since `legacy_slug` is only the last of the legacy slugs defined above) but all of these `assert not [loc for loc in all_locs if XXX in loc]` lines (lines 178-181 inclusive), since your final assert really takes care of everything.

PR #5417, opened 2019-05-15
bug 1551941 - retry on SMTPServerDisconnected

peterbe:
Easy to do. Hard to test. Harder to actually user test. 

Actually, I have no idea how to actually cause a `SMTPServerDisconnected` exception but I made sure that email could be sent. To do that I had to manually trigger various functions from a little hack script. Emails still worked. 

The default in `redo` is to sleep 60 seconds. That was just too much. I made it 3 seconds. So if our SMTP server is in a really bad shape, it will try again after:

1. 3 seconds
2. 3 * 1.5 seconds
3. 3 * 1.5 * 1.5 seconds
4. 3 * 1.5 * 1.5 * 1.5 seconds
5. 3 * 1.5 * 1.5 * 1.5 * 1.5 seconds

and if it still failed on the last attempt the exception will bubble up as before.


tobinmori:
original filing: https://bugzilla.mozilla.org/show_bug.cgi?id=1551941

is this still an issue @peterbe?

peterbe:
> is this still an issue @peterbe?

Yes. It needs a code review from @escattone 
I think the SMTPServerDisconnected error isn't hugely important but this PR introduces a new library which I think makes us a lot more reliable when we use it. 

I don't think I wrote it down but this PR will introduce that `redo` lib which is great when you want to do retries of something that isn't simply a `requests.get()`.

callahad:

peterbe:
@escattone Rebased up against Python3 master.

peterbe:
@Gregoor You're the last one to touch sending emails. Without this change, sending a Stripe invoice might fail if, in that exact instance the webhook comes in, that our SMTP connection fails. And that does happen. This whole PR started because we saw those cryptic `SMTPServerDisconnected` exceptions happen in Sentry.

kuma/core/utils.py
```diff
@@ -539,3 +542,63 @@ def safer_pyquery(*args, **kwargs):
...
+        * smtplib.SMTPServerDisconnected
+
+    Only list exceptions that have been known to happen and are safe.
+    """
+
+    def send(self, *args, **kwargs):
+        # See https://github.com/mozilla-releng/redo
+        # for a list of the default options to the redo.retry function
+        # which the redo.retrying context manager wraps.
+        retry_options = kwargs.pop(
```
Gregoor: Might be a "I'm missing python subtleties"-kind of question: But why pop kwargs here and just specify retrying in 590 explicitly?

peterbe: Normally, when you override a method and want to add some extra keyword arguments you do something like:
```python
def send(self, *args, special="for this override", **kwargs):
```
that way you're not messing with `kwargs` with keys that are only relevant in the override. 
But the default this time is so much more complex (unlike the above example `"for this override"`). 

I guess it could be something like this instead:
```python
def send(self, *args, retrying=None, **kwargs):
    if retrying is None:
        retrying = { "retry_exceptions": .... }
```
Do you think that's prettier?

Gregoor: I think I do like that more, and I'd be even fine with having the default be that dict. Though a pitfall in Python is that those defaults are only initialied once, right? So all invocations of that functions would have shared mutable access to the same object, which might be too dangerous here?

PR #5420, opened 2019-05-16
Bug 1525719: change React localization to work with SSR

davidflanagan:
This patch modifies the compile-react-i18n Makefile target so
that it generates JSON files instead of JS files for the strings
that are needed in the React front-end. Then, it modifies the
render_react_app template tag in ssr.py to read the localization
data for the specified locale and pass that data (which is mostly
the strings catalog) to the SSR server and to the client.

On the client-side, this patch makes corresponding changes to the
gettext() function so that it uses this new string catalog.


davidflanagan:
This PR is not ready yet: I'm sure I've got failing tests. I'm flagging it for a preliminary review because I'm pretty stoked about this new approach to handling the React string catalog. But I'm also uncertain about the new get_localization_data() function in kuma/wiki/templatetags/ssr.py.

@escattone and @peterbe: does it seem okay to you to have this template tag loading and caching string catalogs each time the {{render_react_app()}} template tag is called for a new locale?  Is there a better way to handle this in django? It feels like brute force to actually just read data from a file.

The react.json files that are being read are typically under 10k, though there are a few that are bigger. The total size of all the files is about 300kb, so we're not talking about a lot of memory load on the servers. This will grow, of course, but still relatively small.

davidflanagan:
Note that the menus are localized along with the content. This hasn't been working, at least since we started doing SSR.

Screen Shot 2019-05-15 at 8 29 57 PM

kuma/wiki/templatetags/ssr.py
```diff
@@ -1,15 +1,32 @@
...
+    Read the frontend string catalog for the specified locale, parse
+    it as JSON, and return the resulting dict. The returned values
+    are cached so that we don't have to read files all the time.
+    """
+    try:
+        path = os.path.join(settings.BASE_DIR,
+                            'static', 'jsi18n', locale, 'react.json')
+        with open(path, 'r') as f:
+            return json.load(f)
+    except Exception as e:
```
peterbe: I'd rather not do this. In fact, I don't think we should accept any exceptions. Why wouldn't the file be there and valid JSON? 

If someone visits (beta.)developer.m.o/en-OS/docs/Foo isn't there already some code that checks `en-OS` against `settings.ACCEPTED_LOCALES` and throws out the request with a 404 Not Found, well before we get this deep. 

If the .json file doesn't exist or is corrupted, I'd rather we find out about it sooner. 

I could accept a broad and brutal defensive approach if this was a lot more fragile and depended on things like network or some other IO. And if our web server ephemeral disk is so fragile that reading disk IO is that fragile I still don't want to use this technique because that `return None` will get stuck in the Python processor memory until the Gunicorn worker is recycled (which is rare). 

Also, considering that this is for the beta site, I'd sooner rather than later learn about potential unexpected errors here.

Makefile
```diff
@@ -57,7 +57,7 @@ compilejsi18n:
 compile-react-i18n:
 	@ echo "## Generating React translation catalogs ##"
 	@ mkdir -p build/locale
-	@ python manage.py compilejsi18n -d react
+	@ python manage.py compilejsi18n -d react -f json
```
peterbe: I'm not sure when it's the right time but this command just *has* to work perfectly before every deployment. 

I can help. There should probably be a Django system check that asserts that for every locale in `settings.ACCEPTED_LANGUAGES` we expect there to be a non-corrupt `{settings.BASE_DIR}/static/jsi18n/{locale}/react.json`. Because if we have that, we can be confident the `get_localization_data` function never fails. 

Is that something you want me to help with.

kuma/wiki/templatetags/ssr.py
```diff
@@ -1,15 +1,32 @@
...
 import json
+import os
 
 import requests
 import requests.exceptions
 from django.conf import settings
+from django.utils import lru_cache
 from django_jinja import library
 
+@lru_cache.lru_cache()
```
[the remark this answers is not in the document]
peterbe: PS. That line needs to drop 1 line to be flake8 acceptable.

kuma/wiki/templatetags/ssr.py
peterbe: To be clear, I'm saying this in the argument strength that the context is the beta site. There, I think we can afford more learning-by-errors.

Makefile
davidflanagan: It sounds like a good idea to add something like this, and I'd welcome your help since I don't know what a "system check" is. But I hope we don't need to block this PR on adding that check.

davidflanagan:
This is updated now with tests and a more complete l10n implementation on the client-side. I took Peter's advice and removed any exception handling. 

Yesterday when I requested review it was because I just wanted to show it off and make sure you agreed with the direction. Now, it s ready for review for real.

davidflanagan:
The patch has 17 changed files, but I don't think the review will be that bad. Most of the scattered front-end changes are because we don't import locale-provider.jsx and gettext.js anymore (because those are deleted). And also because I previously called gettext() too early, so I've had to move the gettext() calls from load time to render time, which is probably the better way to do it, anyway.

peterbe:
I'll put actions to my words. I'm going to start adding "system checks". 
See https://docs.djangoproject.com/en/1.11/topics/checks/

It's a hugely practical tool. Django already has a bunch of these built in. I never know, by heart, which ones cause warnings and which ones cause errors but the complete list of built-ins are here: https://docs.djangoproject.com/en/1.11/ref/checks/

The Tecken project uses python-dockerflow because it's a requirement in CloudOps so the framework is slightly different there. But here's a good example of check: https://github.com/mozilla-services/tecken/blob/ce45bdcb66b64f49b2cddfc37f8f42249b0a7f04/tecken/dockerflow_extra.py#L50
It's simple function that opens your `django.conf.settings` and looks how you've configured your AWS S3 and Google Cloud Storage buckets. If the current settings means you don't have access to do a HEAD request on the bucket, it refuses to start. This is much more explicit and predictable than getting weird exceptions in runtime.

Makefile
peterbe: Let's move that discussion to https://github.com/mozilla/kuma/pull/5425

kuma/javascript/src/index.jsx
```diff
@@ -15,10 +16,11 @@ if (container) {
     // be garbage collected once it is no longer in use.
     window._react_data = null; // eslint-disable-line camelcase
 
+    // Store the string catalog so that l10n.gettext() can do translations
+    localize(data.requestData.locale, data.localizationData);
```
escattone: It seems like you could add `getLocale` to the import from `./l10n.js` above, and use `getLocale()` instead of `data.requestData.locale` in the call to `localize`. I would vote for doing that and getting rid of `requestData` altogether, since the `locale` is baked into the localization data (see comment below as well).

kuma/wiki/templatetags/ssr.py
```diff
@@ -1,15 +1,31 @@
...
+    path = os.path.join(settings.BASE_DIR,
+                        'static', 'jsi18n',
+                        locale, 'react.json')
+    with open(path, 'r') as f:
+        return json.load(f)
+
+
 @library.global_function
-def render_react_app(document_data, request_data, ssr=True):
+def render_react_app(locale, document_data, request_data, ssr=True):
```
escattone: I would vote for dumping `request_data` from the signature of this function, and then:
- change the creation of the `data` dictionary below to:
```
    data = {
        'localizationData': get_localization_data(locale),
        'documentData': document_data,
    }
```
- remove `request_data` altogether from the `react_document` view (and everywhere else it's used)
- only use `getLocale()` within the React components (dump all use of `data.requestData.locale`)

kuma/javascript/src/ssr.jsx
```diff
@@ -2,13 +2,15 @@ import React from 'react';
...
 /*
  * This function performs server-side rendering of our UI, given
  * a JSON object of document data. It is used by ../ssr-server.js
  */
 export default function ssr(data) {
-    return renderToString(
-        <App documentData={data.documentData} requestData={data.requestData} />
-    );
+    // Store the string catalog so that l10n.gettext() can do translations
+    localize(data.requestData.locale, data.localizationData);
```
escattone: Like before, I would vote for adding `getLocale` to the import from `./l10n.js` above, use `getLocale()` instead of `data.requestData.locale` in the call to `localize`, and get rid of `requestData` altogether.

kuma/javascript/src/index.jsx
davidflanagan: I can't call getLocale() here, because this is the call that is initializing that data. getLocale() won't work until localize() has been called.

And data.localizationData does not define the locale that it is for. I could modify the backend to add a locale property to the localizationData dict, but then it wouldn't just be a copy of the react.json file...

kuma/wiki/templatetags/ssr.py
davidflanagan: I thought about doing that, but worry that I'll end up needing other data (I don't know what) from the request and won't have anywhere to put it.

Also, in order to make this work, I'd have to do:

```
localizationData['locale'] = locale
```

davidflanagan:
I think I've addressed Peter's concerns for this PR, and Ryan has approved, so I'm going to go ahead and merge even though it is saying it is blocked.

PR #5423, opened 2019-05-17
Bug 1550503, remove payments experiment source code and assets

schalkneethling:
In the end, and because we have time, I removed all things payments except for the FAQ, terms and manage pages. That means, these URLs remain:

http://mdn.localhost:8000/en-US/payments/
http://mdn.localhost:8000/en-US/payments/terms/
http://mdn.localhost:8000/en-US/payments/recurring/management/

These also remain functional but simply load /payments/

http://mdn.localhost:8000/en-US/payments/recurring/
http://mdn.localhost:8000/en-US/payments/recurring/subscription

Not sure if that was the appropriate way to do it, or whether we need to create actual redirects.

These URLs no longer exist:

http://mdn.localhost:8000/en-US/payments/success
http://mdn.localhost:8000/en-US/payments/error
http://mdn.localhost:8000/en-US/payments/recurring/success/
http://mdn.localhost:8000/en-US/payments/recurring/error/

This also includes the copy change requested here:
https://github.com/mdn/sprints/issues/1467

Review of this is open to any/all of the following folks depending on their availability:

@peterbe @davidflanagan @escattone Thanks!


kuma/payments/urls.py
```diff
@@ -5,23 +5,11 @@
...
-    url(r'^recurring/error/?$',
-        views.confirmation, {'status': 'error', 'recurring': True},
-        name='recurring_payment_error'),
     url(r'^recurring/?$',
-        views.contribute_recurring_payment_initial,
+        views.contribute,
         name='recurring_payment_initial'),
     url(r'^recurring/subscription/?$',
-        views.contribute_recurring_payment_subscription,
+        views.contribute,
```
davidflanagan: It seems fine by me to do it this way without a redirect. If we ever turned contributions back on, these URLs would become live again. And if we discontinue the program for good, then view.contribute could redirect to a new "payments/discontinued" page.

kuma/settings/common.py
```diff
@@ -1025,22 +1024,10 @@ def pipeline_one_scss(slug, **kwargs):
     },
     'payments': {
         'source_filenames': (
-            'js/components/payments/focus-visible.js',
-            'js/components/payments/payments-handler-utils.js',
-            'js/components/payments/payments-handler.js',
-            'js/components/payments/payments-faq.js',
-            'js/tooltip.js',
             'js/components/payments/payments-manage.js',
         ),
```
davidflanagan: There's now just this one js file in the payments bundle, but the bundle still gets included on every document page (and the homepage, I assume). Maybe you can move it so that it is only included in the /payments/recurring/management page.

schalkneethling:
## Screenshot of payments page

Screenshot 2019-05-21 15 16 50

## Screenshot of terms page

Screenshot 2019-05-21 15 17 03

@davidflanagan ^^

schalkneethling:
@davidflanagan payments js is now only included on the `management.html` page.

PR #5424, opened 2019-05-17
1552277 GitHub complete login network requests should be retrying

peterbe:
(first of all, this PR is not ready because it includes two separate commits. One of which needs to be landed first in [this separate PR](https://github.com/mozilla/kuma/pull/5421). I didn't want to wait for #5421 to land before starting on this because I "had it fresh in my head". Once that other PR lands, I'll clean up this PR)

**This PR is important and not important at the same time.**

It's not important since it's rather harmless little change in something that happens relatively rarely. All it does is protect the completion-view of the signup flow by retrying `request.get` if GitHub was every to respond with a 500, 502 or 504 response. It's the right and a safe thing to do. 

But it's important because this PR introduces an important new utility function. Namely; `requests_retry_session`. It's a function I originally wrote for [tecken](https://github.com/mozilla-services/tecken/blob/master/tecken/base/utils.py#L14) about 2 years ago. It was very well researched and has been running very productively in symbols.mozilla.org for 2 years now. I've since reused in a bunch of places with success. And it's something [I've blogged about](https://www.peterbe.com/plog/best-practice-with-retries-with-requests) and it's a post that's been well received and had over 75k views. 

My hope and intention is to reuse this new utility function in a bunch of places in kuma. Almost all the places where we currently do `requests.get` or `requests.post`. This PR is its "introduction" to kuma.

### DIFF
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -13,6 +13,7 @@ from django.http import Http404
 from django.test import RequestFactory
 from pyquery import PyQuery as pq
 from pytz import timezone, utc
+from requests.exceptions import ProxyError, SSLError
 from waffle.models import Flag
 
 from kuma.core.tests import assert_no_cache_header
@@ -1129,6 +1130,39 @@ class KumaGitHubTests(UserTestCase, SocialTestMixin):
         assert resp.status_code == 200
         doc = pq(resp.content)
         assert 'Account Sign In Failure' in doc.find('h1').text()
+
+    def test_login_SSLError_on_getting_profile(self):
+        resp = self.github_login(profile_exc=SSLError)
+        # No redirect!
+        assert resp.status_code == 200
+        doc = pq(resp.content)
+        assert 'Account Sign In Failure' in doc.find('h1').text()
+
+    def test_login_ProxyError_on_getting_email_addresses(self):
+        resp = self.github_login(email_exc=ProxyError)
+        # No redirect!
+        assert resp.status_code == 200
+        doc = pq(resp.content)
+        assert 'Account Sign In Failure' in doc.find('h1').text()
+
+    @override_config(RECAPTCHA_PRIVATE_KEY='private_key',
+                     RECAPTCHA_PUBLIC_KEY='public_key')
+    def test_signin_captcha(self):
+        resp = self.github_login()
+        self.assertRedirects(resp, self.signup_url)
+
+        data = {'website': '',
+                'username': 'octocat',
+                'email': 'octo.cat@github-inc.com',
+                'terms': True,
+                'g-recaptcha-response': 'FAILED'}
+
+        with mock.patch('captcha.client.request') as request_mock:
+            request_mock.return_value.read.return_value = '{"success": null}'
+            response = self.client.post(self.signup_url, data=data, follow=True)
+        assert response.status_code == 200
+        assert (response.context['form'].errors ==
+                {'captcha': [u'Incorrect, please try again.']})
 
     def test_matching_user(self):
         self.github_login()
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -1145,25 +1145,6 @@ class KumaGitHubTests(UserTestCase, SocialTestMixin):
         doc = pq(resp.content)
         assert 'Account Sign In Failure' in doc.find('h1').text()
 
-    @override_config(RECAPTCHA_PRIVATE_KEY='private_key',
-                     RECAPTCHA_PUBLIC_KEY='public_key')
-    def test_signin_captcha(self):
-        resp = self.github_login()
-        self.assertRedirects(resp, self.signup_url)
-
-        data = {'website': '',
-                'username': 'octocat',
-                'email': 'octo.cat@github-inc.com',
-                'terms': True,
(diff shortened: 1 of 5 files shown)


peterbe:
@escattone I think this is ready now. The change is actually pretty trivial and I'm confident about the overall strategy. 

When this lands we can return to all sorts of places where we currently do `requests.get(...)` or `requests.post(...)` and replace them with `requests_retry_session().get(...)` etc.

peterbe:
> When this lands we can return to all sorts of places where we currently do `requests.get(...)`

Actually, I think the only remaining place is in `kumascript.py` but that's also the most interesting place. 

Today, if we ask kumascript (via HTTP) to render a piece of macro content, and kumascript, for some reason, times out or has some other sporadic network IO error, the whole rendering process will just accept that and move on. Instead, by using a retry approach there's a big chance that many many places it will just give kumascript another chance and fix itself. Exciting!

kuma/users/tests/test_views.py
```diff
@@ -1121,6 +1122,20 @@ def test_login_500_on_getting_email_addresses(self):
         doc = pq(resp.content)
         assert 'Account Sign In Failure' in doc.find('h1').text()
 
+    def test_login_SSLError_on_getting_profile(self):
```
escattone: I'm not sure what these two new tests are testing, since they don't really test whether `KumaGitHubOAuth2Adapter.complete_login` uses `requests_retry_session()` or not (both complete successfully whether I use `requests_retry_session()` or not). Maybe you added them simple to more fully flesh-out these tests?

escattone:
By the way, using GitHub's interface, I resolved the merge conflict that was blocking this PR.

kuma/users/tests/test_views.py
peterbe: I believe so. It's been a while but I think the intention was to just add some more coverage that didn't already exist.

tobinmori:
@peterbe is this one ready to go then, since #5421 has been merged?

peterbe:
Actually, I think the ball is in @escattone 's court. He had some questions about the unit tests. Yes, I had probably added some tests that didn't really pinpoint the code change but the tests are still good in that they cover something that wasn't covered before. 

Essentially, this is still awaiting review and I think it contains important platform fundamentals that allows us to do better retries when network is acting up.

callahad:

peterbe:
@escattone rebased for py3.

peterbe:
> @peterbe We still need a test for the `requests_retry_session()` function that checks it's essential functionality, that it actually retries. We need something where you mock a connection error for the first two responses, followed by a successful response for the third response, or something like that.

It's been a while since I wrote it but there's this: https://github.com/mdn/kuma/pull/5424/files#diff-fcf0c33dc6f2df263e24a5e0eaa87e98R76-R92

escattone:
@peterbe 

> > We still need a test for the `requests_retry_session()` function that checks it's essential functionality, that it actually retries. We need something where you mock a connection error for the first two responses, followed by a successful response for the third response, or something like that.
> 
> It's been a while since I wrote it but there's this: https://github.com/mdn/kuma/pull/5424/files#diff-fcf0c33dc6f2df263e24a5e0eaa87e98R76-R92

The specific test cases within that test don't verify the retry mechanism, since the response would be the same whether it retried or not. I'm thinking of this:

```python
@pytest.mark.parametrize('retry_condition', (
    dict(exc=ConnectionError),
    dict(text='gateway timeout!', status_code=504),
), ids=('connection-error', 'server-error'))
def test_requests_retry_session_2(mock_requests, retry_condition):
    url = 'http://example.com/oh/no/oh/ok'

    mock_requests.get(url, [
        retry_condition,
        retry_condition,
        dict(text='hi', status_code=200),
    ])

    response = requests_retry_session().get(url)
    assert response.text == 'hi'
```

which verifies that the request was retried for both cases. However, when I ran my parametrized test case above on the code in this PR, it failed in both cases, which makes me suspect that the retry mechanism is either not working or doesn't play well with `requests-mock`.

peterbe:
Ok. That was a timesink. Sigh. At least we're learning something!

So, this retry mechanism is internal to `requests`. If we wanted to unit test that it works, we'd be unit testing `requests` which I don't think is appropriate. 

With `mock_requests` you can specify a callback function instead of hardcoding the response. I did that an found out it only calls the callback function the first time. 
When you set up `mock_requests` to be a list of responses, it basically assumes that you're doing either of this.:

```python
requests.get('/')
requests.get('/')
requests.get('/')
...
```
or
```python
session = requests.Session()
session.get('/')
session.get('/')
session.get('/')
...
```

But how `HTTPAdapter` works is that it's only calling `session.get` **once** but then, it internalizes the retry functionality. 

I'm generally a bit disappointed to not have more control but I think it's fine as it is. 
Lots of eyeballs on the `requests` project and it's a highly maintained one so I'm included to trust that it works as expected. 

Perhaps [`responses`](https://pypi.org/project/responses/) works differently at this. I'm not sure. Not sure it's worth it either. 

By the way, years ago I tried some other library that went much deeper. Instead of mocking `requests` it mocked `socket` (or whatever it was called) in Python's standard library. It never worked sensibly. Perhaps the library was bad or it only works when you're doing unit tests of `requests` itself :) I confidently concluded it wasn't right and some other respectably Python peers expressed consent.

PR #5425, opened 2019-05-17
bug 1553162 - react_i18n_check

peterbe:
Looks like this:

```
▶ docker-compose exec web ./manage.py check
...
SystemCheckError: System check identified some issues:

ERRORS:
?: (kuma.core.E001) Locale file /app/static/jsi18n/sv-SE/react.json does not exist
	HINT: Run 'make compile-react-i18n'

System check identified 1 issue (1 silenced).
```

It's early days. The context of this is to commence writing more custom system checks. In https://github.com/mozilla/kuma/pull/5420 I said we should assume that react i18n should have been created when the app starts. WIth this in place, we can make bolder "assumptions" in runtime because the system isn't supposed to be allowed to start/deploy if you haven't done all the things you need to do first. 

At this very moment, this PR is expected to fail. In CI, we run `./manage.py migrate` which is one of the entry points for running checks. We *could* add `./manage.py check` to the CI script. 

Also, we might need to investigate the groovy scripts to make sure it doesn't deploy if a `./manage.py migrate` failed. 

This very first check is very verbose and it's perhaps an overstep to check the *content* of each `react.json` file but it's nice to test out the warning system too.


peterbe:
Note, if you had completely forgotten to run `make compile-react-i18n` before trying to start Django, you're going to get `len(settings.ACCEPTED_LOCALES)` errors which might not be very pleasant on the terminal. For example, we could exit on first or second file-missing error.

peterbe:
Another thing I need to do is inject this into `docker-compose.yml`  because running `docker-compose up` won't execute `./manage.py migrate` nor `./manage.py runserver` (we use `gunicorn`) there. 

PS. The way it works is that you can either manually run `./manage.py check` or by running `./manage.py migrate|runserver`  it does it automatically.

peterbe:
CC @escattone 

So one thing I don't like is that local devs won't notice system checks. 
See the latest commit added. Now when you start `docker-compose up` one of the outputs is:

```
checks_1         | SystemCheckError: System check identified some issues:
checks_1         |
checks_1         | ERRORS:
checks_1         | ?: (kuma.core.E001) Locale file /app/static/jsi18n/sv-SE/react.json does not exist
checks_1         | 	HINT: Run 'make compile-react-i18n'
checks_1         |
checks_1         | System check identified 1 issue (1 silenced).
```

But it doesn't stop `docker-compose up`. I *could* have used `docker-compose up --abort-on-container-exit` but who's going to do that?! 
And if you use `docker-compose up -d` it wouldn't be noticed anyway. 

I would much prefer that we change the `web: command:` step in `docker-compose.yml` to look something like this:

```
  web:
    <<: *worker
    command: ./scripts/start-web.sh
    ...
```
where `scripts/start-web.sh` would look like this:

```bash
set -ex

./manage.py check
gunicorn -w ${GUNICORN_WORKERS:-4} --bind 0.0.0.0:8000 --access-logfile=- --timeout=120 --worker-class=meinheld.gmeinheld.MeinheldWorker kuma.wsgi:application
``` 

How do you feel about that?

kuma/core/tests/test_checks.py
```diff
@@ -0,0 +1,70 @@
...
+        json.dump({
+            "formats": [],
+            "catalog": {},
+            "plural": "multiple"
+        }, f)
+    with open(sv_path, 'w') as f:
+        f.write('{{not valid JSON')
+    errors = checks.react_i18n_check(None)
+    assert len(errors) == 1
+    json_decode_error = 'Expecting property name: line 1 column 2 (char 1)'
```
davidflanagan: nit: seems a little brittle to make your test depend on a error message that is out of your control. Maybe just check the hint and the id?

peterbe: I agree. I guess I could use the `mock.ANY` or whatever it's called.

peterbe: Scratch that, it's a string inside a string. I'll think of something else.

kuma/core/checks.py
```diff
@@ -0,0 +1,70 @@
...
+                    key for key in ('catalog', 'plural', 'formats')
+                    if key not in data
+                ]
+                if missing:
+                    errors.append(Warning(
+                        'Locale file {} is missing keys {!r}'.format(
+                            path, missing
+                        ),
+                        hint=general_hint,
+                        id=WARNING_MISSING_I18N_DATA,
```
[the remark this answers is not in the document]
peterbe: Hmm... Your call. " I'd go ahead and treat this case as a data corruption error". So be it. 

Is there anything else you think should only yield a warning? 
To be honest, I just picked anything because I wanted to justify adding an example of `Warning` in this PR. I'd rather you direct more explicitly what exact you think is so busted that it should prevent the startup. 

Perhaps it's entirely silly to do any analysis of the *content* of the JSON files because it's so far away from where it matters. In principle, the checks should be lightweight. We could, for example, just check that the JSON isn't empty or something like that.

peterbe:
> But maybe warn for locales that have empty string catalogs?

We could use Error if the json file is empty and Warning if the json file's `catalog` key is empty. Again, I'm in your hands since you now understand the l18n much better than I do.

peterbe:
Also, thanks for the approval but shall we attempt to also tackle how these checks should be executed. Today, with docker-compose you would not notice the checks. Above I mentioned a solution that would change that.

kuma/core/checks.py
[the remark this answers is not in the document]
peterbe: @davidflanagan I'm ready to make a final call with the code. Please guide me. 
System checks should be lightweight in principle since their business logic is relatively far away from the where the data gets used. 

I'm probably in favor of dropping any `Warning` since it's so likely to not get noticed.

peterbe:
@escattone I'll keep the "not ready" flag on this till you've had a chance to share your mind about how we should be using this in `docker-compose`. On most other projects I've worked on, we use `./manage.py runserver` executed from docker-compose so I've never had this problem before. 

Another option, it never run `./manage.py check` anywhere but instead always run `./manage.py migrate` which might kill two birds with one stone. A) applies all outstanding migrations B) implicitly executes the system checks. 
For example, on my personal blog system, I don't use `docker-compose` but I also don't type `./manage.py runserver`. Instead I start it with `./bin/run.sh web` which triggers [these lines](https://github.com/peterbe/django-peterbecom/blob/c6f17146d4c0b294cba7805748654b3f1ef396f4/bin/run.sh#L33-L41).

peterbe:
@escattone And I discussed this in detail today. (The infra, not the react .json check stuff). In deployment, we always run `./manage.py migrate` but it's like this: `if not in_maintaintance_mode: ./manage.py migrate`. 
So we should change our Groovy stuff so we always run `./manage.py check`. Somewhere before the `./manage.py migrate` so that it gets run independent of maintenance mode.

peterbe:
@escattone Likes the idea of a one command that sums up all the things you need to run Django. I.e. the `migrate`, the `collectstatic` etc. and lastly starting `gunicorn`.

peterbe:
@escattone r?

First of all, we still need to make sure `./manage.py check` is called in the Groovy code. It needs to be done right before `if not maintenance_mode: ./manage.py migrate`. 

The idea of doing all sorts of "chores" in a separate docker-compose service sucked. 
The problem is that it takes so bloody long and 99% of the time, it's a waste of time. The rare few times when you do a `git pull origin master` after Schalk/David has made an edit are rare, you'll just have to remember to execute `make build-static`. Otherwise it just makes the fan spin since all those build-static steps are slow. It's fine, if you know there are things to do every now and then. Actually, I have a little script of my own that always makes sure now to run `docker-compose exec web make build-static` after a git pull. But it's not automated. 

The other key thing is the chicken and egg problem within the "web" container. If you allow the new script `./scripts/start-web` to exit because of a system check, it's possible that you can't rectify it unless the web service gets a chance to start. Case in hand; the fact that you have to run `docker-compose exec web make build-static` to rectify the problem. 

If you run `docker-compose up -d`  or `docker-compose` the `gunicorn` service will always start. Your only hope of noticing that something needs attention is by leveraging the output logging. So I made a rather odd bash function (`make_stink`) and you can read the code comment within.

peterbe:
Generally, I have this growing feeling that `docker-compose` should just be minimal and about starting things. It shouldn't be a myriad of scripts that do various chores. Those will need to be executed some other way but I don't think they should be part of `docker-compose up`. The only exception that I can think of right now is `./manage.py migrate`. Partly because it kills two birds with one stone and partly because it's so useful in that it quickly does something that the `gunicorn` command will greatly benefit from. 
I suspect there will be more little tasks that might go into `./scripts/start-web` but that's a story for another day. At least, now, there's a good place for it where you can type it in with bash syntax highlighting and with code comments.

escattone:
I'll get back to this after I review https://github.com/mozilla/kuma/pull/5469.

peterbe:
Hint hint @escattone https://github.com/mozilla/kuma/pull/5469 is merged now :)

peterbe:
@escattone We should still work on this one.

peterbe:
@escattone I unrequested your review and added @callahad
The reason is that I think he'd benefit from having this "framework" in place to complete the work on https://github.com/mozilla/kuma/pull/5998

Yeah, Dan, if you merge this, and go back to your S3 PR, you'd have to `cp kuma/core/checks.py kuma/attachments/checks.py` and edit accordingly so do some boto3 tricks instead of messing with JSON files.

callahad:

callahad:
Closing and re-opening to force Travis to re-run tests against the current `master`, which uses Python 3.

kuma/core/checks.py
```diff
@@ -0,0 +1,73 @@
...
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    # Still in Python 2
+    JSONDecodeError = ValueError
+try:
+    FileNotFoundError
+except NameError:
+    # Still in Python 2
+    FileNotFoundError = IOError
```
callahad: Now that we're on Py3, we can drop these conditionals

peterbe:
@callahad I think I've addressed your comments. Don't you need this to be able to finish https://github.com/mdn/kuma/pull/5998 so it can be used to take care of automatic bucket creation in dev, and automatic bucket access-validation in prod.

callahad:
> @callahad I think I've addressed your comments. Don't you need this to be able to finish #5998 so it can be used to take care of automatic bucket creation in dev, and automatic bucket access-validation in prod.

As mentioned [over there](https://github.com/mdn/kuma/pull/5998#issuecomment-559718280), Gregor and I agreed that neither were blockers. Hoping to review this tomorrow or Monday.

peterbe:
Ping @callahad

PR #5428, opened 2019-05-19
bug 1525719: cleanup react_document view

escattone:
I was double-checking how we handle contributors in the React-based views, and realized that there was some redundancy in the `react_document` view, as well as a number of no-longer-used context variables. This PR simplifies that view based on current usage. There is more here that we'll simplify later as we move more of the Jinja template code to the React side.

### DIFF
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -108,6 +108,10 @@ def document_api_data(doc=None, ensure_contributors=False, redirect_url=None):
     else:
         en_slug = ''
 
+    other_translations = doc.get_other_translations(
+        fields=('locale', 'slug', 'title'))
+    all_locales = set([doc.locale]) | set(t.locale for t in other_translations)
+
     return {
         'documentData': {
             'locale': doc.locale,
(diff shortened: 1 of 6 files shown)


kuma/api/v1/views.py
```diff
@@ -108,6 +108,10 @@ def document_api_data(doc=None, ensure_contributors=False, redirect_url=None):
     else:
         en_slug = ''
 
+    other_translations = doc.get_other_translations(
+        fields=('locale', 'slug', 'title'))
+    all_locales = set([doc.locale]) | set(t.locale for t in other_translations)
```
davidflanagan: nit: I had a moment where I thought you were using get_hreflang wrong and actually passing all locales to it. This would be slightly clearer if the variable name was translation_locales or available_locales or something like that.

davidflanagan:
It looks like fallback_reason is also unused, so you could delete that one, too!

kuma/api/v1/views.py
escattone: Good point, I like `available_locales` better, thanks!

PR #5429, opened 2019-05-19
bug 1525719: handle 404 in client-side nav

escattone:
This PR adds functionality to the `DocumentProvider` React component that allows client-side navigation even when no document JSON exists for a slug within a non-English locale. Currently, when document JSON is requested for a non-English document and nothing exists for that locale (the response is a 404), client-side navigation is abandoned. We instead request a full page load, and the `react_document` view that serves the full-page load will fall back to the English version of the document if the non-English version doesn't exist. This PR does that same thing but on the client-side.

For local testing, I used the following `.env`:
```
DEBUG=True
DOMAIN=mdn.localhost
ATTACHMENT_HOST=demos:8000
```
and performed the following steps:
- First, demonstrate the current behavior when requesting document JSON for a non-English slug:
    - checkout the master branch
    - docker-compose pull
    - docker-compose up -d
    - docker-compose exec web make build-static
    - load http://beta.mdn.localhost:8000/de/docs/Web
    - open dev tools and go to the `Network` tab
    - click on the `HTML` link under black `Grundlagen` header (the `Web/HTML` document has not been translated into German in the sample database, so the document API request will 404)
    - notice that a full-page load is performed

- Second, demonstrate the behavior of this PR when requesting document JSON for a non-English slug:
    - checkout the this branch
    - docker-compose down
    - docker-compose up -d
    - docker-compose exec web npm run webpack
    - load http://beta.mdn.localhost:8000/de/docs/Web
    - open dev tools and go to the `Network` tab (also select `XHR`)
    - click on the `HTML` link under black `Grundlagen` header
    - notice that two fetches are made: the first a 404, and the second for the English document
image

### DIFF
--- a/kuma/javascript/src/document-provider.jsx
+++ b/kuma/javascript/src/document-provider.jsx
@@ -52,10 +52,28 @@ export default function DocumentProvider(
         }
         const body = document.body;
 
+        function fetchWithFallback(url1: string, url2: string) {
+            const fetchConfig = { redirect: 'follow' };
+            return fetch(url1, fetchConfig).then(response => {
+                if (response.ok || !url2 || url2 === url1) {
+                    return response;
+                } else {
+                    return fetch(url2, fetchConfig);
+                }
+            });
+        }
+
         // This is the function that does client side navigation
         function navigate(url, slug) {
             body.style.opacity = '0.15';
-            fetch(`/api/v1/doc/${locale}/${slug}`, { redirect: 'follow' })
+            // The fallback is for the case when we request a non-English
+            // document that doesn't exist. In that case, before we abandon
+            // client-side navigation, we'll try falling-back to the English
+            // document.
+            fetchWithFallback(
+                `/api/v1/doc/${locale}/${slug}`,
+                `/api/v1/doc/en-US/${slug}`
+            )
                 .then(response => {
                     if (response.ok) {
                         return response.json();


kuma/javascript/src/document-provider.jsx
```diff
@@ -51,20 +51,32 @@ export default function DocumentProvider(
...
-                    if (response.ok) {
-                        return response.json();
+                    if (locale !== 'en-US' && response.status === 404) {
+                        // We requested a non-English document that didn't
+                        // exist, but before we abandon client-side navigation,
+                        // let's first try falling-back to the English document.
+                        return fetch(
+                            `/api/v1/doc/en-US/${slug}`,
+                            fetchConfig
+                        ).then(getJSON);
```
davidflanagan: JavaScript promises are awesome when they work well, but it can be really hard sometimes to figure out how to make them work well.  Ideally, all the .then() calls are in a single linear chain, and having to use a nested one like this is a code smell.

The code seems fine, and I'm not opposed to landing it this way. But factoring out the getJSON function doesn't seem ideal to me.

I think that the better approach might be to write a utility function fetchWithFallback() that can take one or two urls. With one url, it works just like fetch(). But with two URLs, if it gets a 404 on the first, it fetches the second.  If you had such a utility function, then this change would fit trivially into the existing promise chain and all you'd have to do is the logic of deciding to pass one url (when the locale is en-US) or two urls.

davidflanagan: I'm going to get some coffee and then try to write this utility function.

davidflanagan: This works for me:

```
function fetchWithFallback(url1: string, url2: string) {
  const fetchConfig = { redirect: "follow" };
  if (!url2 || url2 === url1) {
    return fetch(url1, fetchConfig);
  }

  return fetch(url1, fetchConfig).then(response => {
    if (response.ok) {
      return response;
    } else {
      return fetch(url2, fetchConfig);
    }
  });
}
```

davidflanagan: Actually, this is a more compact version that also seems to work:

```
function fetchWithFallback(url1: string, url2: string) {
  const fetchConfig = { redirect: "follow" };
  return fetch(url1, fetchConfig).then(response => {
    if (response.ok || !url2 || url2 === url1) {
      return response;
    } else {
      return fetch(url2, fetchConfig);
    }
  });
}
```

davidflanagan: Note that the way this function compares the two urls, you could simply call it with the original url and the same url forced to en-US and wouldn't even have to care about whether they were actually the same url.

escattone: Nice, I like it. I'll make that change, thanks @davidflanagan!

escattone:
@davidflanagan Great idea, thanks so much for your `fetchWithFallback`! The code is cleaner, simpler, and easier to understand.

PR #5431, opened 2019-05-20
Bug 1525719: measure and report client-side navigation timing.

davidflanagan:
This PR uses performance.mark() and performance.measure() to record
the fetch and the fetch+render times for each client-side navigation
and reports those times to Google Analytics with a 'timing' hitType
and a timingCategory of 'Client side navigation'.

### DIFF
--- a/kuma/javascript/src/document-provider.jsx
+++ b/kuma/javascript/src/document-provider.jsx
@@ -4,6 +4,7 @@ import { useContext, useEffect, useState } from 'react';
 
 import GAProvider from './ga-provider.jsx';
 import { getLocale } from './l10n.js';
+import { navigateStart, navigateFetchComplete } from './perf.js';
 
 export type DocumentData = {
     locale: string,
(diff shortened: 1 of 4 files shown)


davidflanagan:
The main part of this PR is the new perf.js file, and the calls to the methods it defines in document-provider.jsx and page.jsx.

In order to properly measure the document render time, I refactored page.jsx to create a new Document component that renders the document sub-components. And then instead of calling useContext() everywhere, I pass the document data via regular props. This simplifies the code a bit, but also causes more churn in the diff than I would have liked. Anyway: the important changes to page.jsx are at the top and the bottom, and you don't need to review the stuff in the middle very carefully.

escattone:
@davidflanagan I took the liberty of resolving the conflicts. I'm just waiting for the tests to pass and then I'll merge.

escattone:
@davidflanagan It looks like this needs to be re-based on master in order to get the tests/linting to pass.

PR #5432, opened 2019-05-20
Bug 1552481, implement final design for developer needs survey banner

schalkneethling:
This implements the final design for the developer needs survey.

## Desktop

Screenshot 2019-05-20 23 54 31

## Tablet

Screenshot 2019-05-20 23 55 08

## Mobile

Screenshot 2019-05-20 23 58 45

## Desktop RTL

Screenshot 2019-05-20 23 59 07

I am still doing some cross-browser testing, but the code is generally good for review. Thanks!


schalkneethling:
@davidflanagan r?

kuma/static/styles/components/banners/_common.scss
```diff
@@ -2,31 +2,60 @@
...
     z-index: 99;
 
+    @media #{$mq-small-desktop-and-up} {
+        height: 130px;
+    }
+
+    .slab-text {
+        background-color: #222;
+        color: #fff;
+        padding: 0 10px 0 20px;
```
davidflanagan: Is this an intentional design to have different padding on the left and right? Does it make sense in the mobile layout where it is centered?  Also, does it need an RTL variant?

schalkneethling: It is in the same vein as here:
https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#Values

I did not add quite as much left padding but wanted to keep the same general idea. We do not *need* RTL, but this is in `_common.scss` because future banners might.

schalkneethling:
@nickbrandt Let me know your thoughts on the implementation:
https://github.com/mozilla/kuma/pull/5432#issue-280559659

Thanks for the quick review @davidflanagan ☕️

nickbrandt:
@schalkneethling this is looking good, though I'm noticing some differences in the implementation vs. design.

- There is no box-shadow on the banner, which helps create separation on examples such as the mobile one you posted
- The heading "MDN Survey" is bold vs. it is regular in the design (matches other implementations of that heading style across MDN)
- I think we need a max width on the CTA button, as it gets pretty big in some instances
- Button font weight should be bold
- Icon is much smaller in implementation

You should be able to pull all of the values from the design from this link (clicking on elements shows you the css values), if you have any thoughts or questions on this, let me know!

https://zpl.io/beA7KkB

schalkneethling:
Thanks for the feedback @nickbrandt - I will address these items and update the PR

schalkneethling:
@nickbrandt updated.

## Desktop

Screenshot 2019-05-23 10 42 07

## Tablet

Screenshot 2019-05-23 10 42 37

## Mobile

Screenshot 2019-05-23 10 42 52

nickbrandt:
Looks great @schalkneethling , thanks for addressing those items!

PR #5434, opened 2019-05-21
Ssr on publish

davidflanagan:


davidflanagan:
@escattone: this is what I've got for the publish HTML on publish PR. It doesn't work yet because react_document.html still references the request object in a surprising number of places.

Also, I haven't handled the unpublish case yet, either.

The list of priority locales is from Kadir.

Let me know if this seems reasonable to you, and how much more work you think I should put into it. Should I leave the request cleanup to you, or shall I work on that some more?

escattone:
See https://github.com/mdn/sprints/issues/1587#issuecomment-500976932

PR #5437, opened 2019-05-21
bug 1553303: exclude kuma Dockerfile from renovate

escattone:
This PR adjusts the `renovate` configuration to exclude the Dockerfile for kuma (`docker/images/kuma/Dockerfile`). The base image in that file, `mdnwebdocs/kuma_base:latest` should never be pinned, since it is never acquired from the registry but rather always from a local build performed immediately previous (see our Jenkins and Travis flows).

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,7 +7,8 @@
     "assets/ckeditor4/source/plugins/descriptionlist/package.json",
     "kuma/static/js/libs/jquery-ui-1.10.3.custom/development-bundle/package.json",
     "kumascript/package.json",
-    "tests/performance/Dockerfile"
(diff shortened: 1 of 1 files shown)


peterbe:
Add fuel to the argument that JSON as a config language sucks. Not being able to put in comments there is annoying. 

I *think* you don't need to use a .json file for Renovate but all the examples and docs use JSON.

peterbe:
This doesn't need CI but it's nice to not have a CI check in the commit log view.

escattone:
@peterbe Thanks for the super quick review! (and I totally agree about JSON as a config language).

PR #5438, opened 2019-05-21
bug 1553333: fix ES port-binding failure in Jenkins

escattone:
Sometimes, when running `docker-compose -f docker-compose.yml -f docker-compose.test.yml run test` within Jenkins, the host system's bind to port 9200 fails when spinning-up the `elasticsearch` service:
```
+ docker-compose -f docker-compose.yml -f docker-compose.test.yml run test
Creating kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_elasticsearch_1 ... 
Creating kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_mysql_1         ... 
Creating kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_redis_1         ... 
Creating kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_elasticsearch_1 ... error
ERROR: for kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_elasticsearch_1  
Cannot start service elasticsearch: driver failed programming external connectivity on endpoint kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_elasticsearch_1 (ab17e49d1c083d1d03ac65474381a2c19a5d4f3c1d0bd950dc126dfcfe1a6946):
Bind for 0.0.0.0:9200 failed: port is already allocated
Creating kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_mysql_1         ... done
Creating kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_redis_1         ... done
ERROR: for elasticsearch 
Cannot start service elasticsearch: driver failed programming external connectivity on endpoint kuma_master-dag2lj3obypfho7nheyqt4rqhoc6v2dswzq2pndatnegq7lwo6na_elasticsearch_1 (ab17e49d1c083d1d03ac65474381a2c19a5d4f3c1d0bd950dc126dfcfe1a6946): 
Bind for 0.0.0.0:9200 failed: port is already allocated
Encountered errors while bringing up the project.
script returned exit code 1
```

This PR moves the host bind port for `elasticsearch` from `9200` to `9201`, which should keep it away from any commonly used ports while still keeping it available outside of docker.

I pushed the branch in my fork into a branch of the same name in the `kuma` repo just so Jenkins would pick it up and run the tests. It started and ran fine:
```
+ docker-compose -f docker-compose.yml -f docker-compose.test.yml run test
Creating elasticsearch-port-error-1553333_mysql_1 ... 
Creating elasticsearch-port-error-1553333_elasticsearch_1 ... 
Creating elasticsearch-port-error-1553333_redis_1         ...
============================= test session starts ==============================
platform linux2 -- Python 2.7.16, pytest-3.1.3, py-1.4.33, pluggy-0.4.0
Django settings: kuma.settings.testing (from ini file)
rootdir: /app, inifile: pytest.ini
plugins: cov-2.4.0, django-3.1.2
collected 1898 items
kuma/api/tests/test_signal_handlers.py ....
kuma/api/tests/test_tasks.py ............
kuma/api/v1/tests/test_views.py ...........................
...
```
but of course the real indicator of success will be when we've not seen any port-binding failures in Jenkins for a week or two.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,8 +102,6 @@ services:
       - discovery.type=single-node
       - xpack.security.enabled=false
       - ES_JAVA_OPTS=-Xms512m -Xmx512m
-    ports:
-      - "9200:9200"
     volumes:
       - esdata:/usr/share/elasticsearch/data
 


docker-compose.yml
```diff
@@ -103,7 +103,7 @@ services:
       - xpack.security.enabled=false
       - ES_JAVA_OPTS=-Xms512m -Xmx512m
     ports:
-      - "9200:9200"
+      - "9201:9200"
```
davidflanagan: IIUC, this means port 9201 inside the container, but localhost:9200 from the outside?

I see that port 9200 appears in kuma/settings/common.py and kuma/search/tests/test_views_admin.py  If neither of those need to change, then this looks good to me.

escattone: It's actually the opposite (see https://docs.docker.com/compose/compose-file/compose-file-v2/#ports). The first number is the port number on the host (so outside the docker-compose environment), and the second is the port number for that container inside the docker-compose environment.

peterbe:
Why is there already an Elasticsearch running on :9200?? ...in jenkins.

escattone:
@peterbe It baffles me too, because apparently nothing is or should be using port 9200 on the Jenkins box (I asked @limed about it).

peterbe:
I suspect we need to understand that first before we proceed with this change. 

I too have had that problem at times. It's because, in another terminal window on my mac, I had started some other project. The right solution wasn't to change the port number :)

escattone:
@peterbe I still think this is a good change. The port number I'm proposing to change is for access on the host:
- that "external" port exists solely for convenience/debugging -- purposes outside that of the docker-compose environment
- we shouldn't try to bind to a port on the host that could legitimately be used, since some Jenkins plugins use it (e.g., an ELK plugin)

In fact, it would be safest to use:
```
ports:
  - "9200"
```
and not expose the docker-compose Elasticsearch on the host at all.

It doesn't seem worthwhile to spend too much time determining why port 9200 is sometimes already in use, when we don't really need the external port at all, and even if we'd prefer to provide it for debugging/convenience, we shouldn't choose a port number that could legitimately be used already.

So given everything we have going on right now, I would vote for trying this as it stands, and if we encounter the problem again, we can dig into further.

peterbe:
> and not expose the docker-compose Elasticsearch to the host at all.

I think it was me who added that. I wanted it when I did the Elasticsearch 6 upgrade. Just so I could poke around directly in Elasticsearch. We're not doing this with the other services. 

Considering the shaky ground arguments, I think we should just retract that and not expose it at all. So I'd be in strong favor of simple deleting the two lines on elasticsearch about ports altogether. 

Yes, if we had more time and less on our plate we could debug this deeper.

escattone:
@peterbe 👍 I like that idea best as well. I'll remove the `ports` section altogether.

PR #5440, opened 2019-05-22
Bug 1525719: modify localization code to not violate CSP

davidflanagan:
In a recent patch I introduced a new l10n.js module for localization
of our React code. It used the Function() constructor to build a
plural function from the plural expression string that Django supplied.
Unfortunately, this violates the CSP policy that we would like to start
using.

So this PR modifies the backend to embed a valid plural function into
the window._react_data object that it includes in each HTML page. And
it also modifies ssr.jsx to use the Function() constructor to create
the plural function (in Node, so it does not violate CSP) when needed.

Because this patch decouples the plural expression from the string
catalog, it also removes the generic localizationData object and
uses the stringCatalog object directly.

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -17,7 +17,7 @@ if (container) {
     window._react_data = null; // eslint-disable-line camelcase
 
     // Store the string catalog so that l10n.gettext() can do translations
-    localize(data.requestData.locale, data.localizationData);
+    localize(data.requestData.locale, data.stringCatalog, data.pluralFunction);
 
     // This is the React UI for a page of documentation
     let app = <App documentData={data.documentData} />;
--- a/kuma/javascript/src/l10n.js
+++ b/kuma/javascript/src/l10n.js
@@ -14,71 +14,48 @@
  * @flow
  */
 
-// This is the type of the localization data that Django creates for a locale.
-type LocaleData = {
-    catalog: { [string]: string | Array<string> },
-    plural: ?string
-    // The data that Django passes to us also has a formats object
-    // containing template strings for date and time formatting and
-    // defining things like the thousands separator. We don't currently
-    // use them, so they are omitted from this type definition.
-};
+// This is the type of the string catalog that Django creates for a locale.
+type StringCatalog = { [string]: string | Array<string> };
 
 // The "plural index" function takes a count as its input and returns
 // a number that is an index into an array of plural forms. This default
 // function is suitable for English, but some languages have more than
 // one plural form and require more complicated rules.
-type PluralIndexFunc = number => number;
+type PluralFunction = number => number;
 
 /*
  * These are default values for the state variables that define the
  * current localization.
  */
 const defaultLocale = 'en-US';
-const defaultLocaleData: LocaleData = {
-    catalog: {},
-    plural: null
-};
-const defaultPluralIdx: PluralIndexFunc = n => (+n === 1 ? 0 : 1);
+const defaultStringCatalog: StringCatalog = {};
+const defaultPluralFunction: PluralFunction = n => (+n === 1 ? 0 : 1);
 
 /*
- * The currentLocale, currentLocaleData, and currentPluralIdx variables
- * hold the current state for this module. Their initial values are the
- * defaults above, which are suitable for untranslated English text.
- * The localize() function defines new values for each of them.
+ * The currentLocale, currentStringCatalog, and currentPluralFunction
+ * variables hold the current state for this module. Their initial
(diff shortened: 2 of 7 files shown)


davidflanagan:
This is a followup to PR #5420 to make the client-side l10n.js module work, even when we have a CSP policy in place. It avoids the use of the Function() constructor which isn't allowed without 'unsafe-eval' in CSP.

The patch is a little more complicated than I'd like, but it also includes some simplifications to the code, which is nice. It seems that we don't have a CSP policy active in prodution or staging, so it isn't urgent to get this merged.

kuma/javascript/src/index.jsx
```diff
@@ -17,7 +17,7 @@ if (container) {
     window._react_data = null; // eslint-disable-line camelcase
 
     // Store the string catalog so that l10n.gettext() can do translations
-    localize(data.requestData.locale, data.localizationData);
+    localize(data.requestData.locale, data.stringCatalog, data.pluralFunction);
```
davidflanagan: In the web browser, we're always going to get the plural function we need passed to us as part of the window._react_data object, so we don't have to construct it ourselves. That's how we avoid the CSP issue.

kuma/javascript/src/l10n.js
```diff
@@ -14,71 +14,48 @@
...
-type LocaleData = {
-    catalog: { [string]: string | Array<string> },
-    plural: ?string
-    // The data that Django passes to us also has a formats object
-    // containing template strings for date and time formatting and
-    // defining things like the thousands separator. We don't currently
-    // use them, so they are omitted from this type definition.
-};
+// This is the type of the string catalog that Django creates for a locale.
+type StringCatalog = { [string]: string | Array<string> };
```
davidflanagan: There are a lot of changes in this file, but most of them are variable and type name changes, and type simplifications like this one.  The change looks bigger that it actually is.

kuma/javascript/src/l10n.js
```diff
@@ -14,71 +14,48 @@
...
-    } else {
-        // Based on the JS code in
-        // https://github.com/django/django/blob/master/django/views/i18n.py
-        let f = new Function(
-            'n', // Don't change this! The data.plural expression uses it.
-            `let v = ${currentLocaleData.plural};
-             if (v === true) return 1;
-             else if (v === false) return 0;
-             else return v;`
-        );
```
davidflanagan: This was the problematic code before... The `new Function()` call gets us into CSP trouble.

kuma/javascript/src/ssr.jsx
```diff
@@ -9,8 +9,20 @@ import { localize } from './l10n.js';
...
+    let pluralFunction = null;
+    if (data.pluralExpression) {
+        // Creating a function like this in client-side JS will usually
+        // violate CSP. But we're running in Node here, so it is okay.
+        pluralFunction = new Function(
+            'n',
+            `let v=(${
+                data.pluralExpression
+            });return(v===true)?1:((v===false)?0:v);`
+        );
```
davidflanagan: During server-side rendering, we need to build the plural function explicitly, but we're running in Node, so we don't have to worry about CSP in this case.

kuma/wiki/templatetags/ssr.py
```diff
@@ -36,8 +36,11 @@ def render_react_app(locale, document_data, request_data, ssr=True):
     Note that we are not defining a generic Jinja template tag here.
     The code in this file is specific to Kuma's React-based UI.
     """
+    localization_data = get_localization_data(locale)
+
     data = {
-        'localizationData': get_localization_data(locale),
+        'stringCatalog': localization_data['catalog'],
+        'pluralExpression': localization_data['plural'],
```
davidflanagan: pluralExpression is the JS expression string that Django gives us. It is what we use in ssr.jsx to build the plural function. And we also use it below to build the plural function we embed in the HTML

kuma/wiki/templatetags/ssr.py
```diff
@@ -54,11 +57,36 @@ def _render(html, state):
...
+    if pluralExpression:
+        # A JavaScript function expression as a Python string
+        js_function_text = (
+            'function(n){{var v=({});return(v===true)?1:((v===false)?0:v);}}'
+            .format(pluralExpression)
+        )
+        # Splice it into the JSON-formatted data string
+        serializedState = (
+            '{pluralFunction:' + js_function_text + ',' + serializedState[1:]
+        )
```
davidflanagan: This is the most important part of the PR: a python string of syntactically correct JS code that we include in a JS data structure inside of an HTML script.

PR #5441, opened 2019-05-22
Make the language menu work for React-based pages

davidflanagan:
The main.js file is not included into our React-based pages, so
the language menu in the footer does not autosubmit like it is
supposed to. This PR adds the required event handler to make it
work again. For now, at least, this code is just tacked on at
the end of index.jsx so it can be transpiled and bundled with
webpack.

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -36,13 +36,13 @@ if (container) {
     }
 
     // Finally, here is some other miscellaneous code that we need to run.
-    // The idea is that this is code need to make the non-React parts of
+    // The idea is that this is code needed to make the non-React parts of
     // the UX work correctly. If we end up with more than a couple of things
-    // here we should refactor them into a separate module
+    // here, we should refactor them into a separate module
 
     // An event handler to make the language selector menu
     // in the footer work correctly.
-    // See also kuma/static/js/main.js where similar code appears ot
+    // See also kuma/static/js/main.js where similar code appears to
     // make the menu work on the wiki domain
     for (let menu of document.querySelectorAll('select.autosubmit')) {
         menu.addEventListener('change', function() {
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -34,4 +34,19 @@ if (container) {
         // server side rendering.
         ReactDOM.render(app, container);
     }
+
+    // Finally, here is some other miscellaneous code that we need to run.
+    // The idea is that this is code need to make the non-React parts of
+    // the UX work correctly. If we end up with more than a couple of things
+    // here we should refactor them into a separate module
+
+    // An event handler to make the language selector menu
+    // in the footer work correctly.
+    // See also kuma/static/js/main.js where similar code appears ot
+    // make the menu work on the wiki domain
+    for (let menu of document.querySelectorAll('select.autosubmit')) {
+        menu.addEventListener('change', function() {
+            this.form.submit();
+        });
+    }
 }


davidflanagan:
Schalk: this patch is trivial... more comments than code. But the tricky part was trying to decide where to put it. I like having it in the react dependency tree so that it gets transpiled and bundled by webpack.  It feels a little weird just tacking it on to index.jsx. But until there is more code tacked on, it didn't seem worth creating a utils directory for little snippets like this.

So this may not be the permanent home for this code. But it seemed as good as anywhere else for now. If you've got other ideas about where to put it, do let me know.

kuma/javascript/src/index.jsx
```diff
@@ -34,4 +34,19 @@ if (container) {
         // server side rendering.
         ReactDOM.render(app, container);
     }
+
+    // Finally, here is some other miscellaneous code that we need to run.
+    // The idea is that this is code need to make the non-React parts of
```
schalkneethling:
```suggestion
    // The idea is that this is code needed to make the non-React parts of
```

kuma/javascript/src/index.jsx
```diff
@@ -34,4 +34,19 @@ if (container) {
...
+    // the UX work correctly. If we end up with more than a couple of things
+    // here we should refactor them into a separate module
```
schalkneethling:
```suggestion
    // here, we should refactor them into a separate module
```

kuma/javascript/src/index.jsx
```diff
@@ -34,4 +34,19 @@ if (container) {
...
+
+    // An event handler to make the language selector menu
+    // in the footer work correctly.
+    // See also kuma/static/js/main.js where similar code appears ot
```
schalkneethling:
```suggestion
    // See also kuma/static/js/main.js where similar code appears that
```

kuma/javascript/src/index.jsx
```diff
@@ -34,4 +34,19 @@ if (container) {
...
+    // make the menu work on the wiki domain
+    for (let menu of document.querySelectorAll('select.autosubmit')) {
```
schalkneethling: Might it be a little better to do the query once and save the result in a variable, instead of doing the query each time we go through the `for` loop?

```suggestion
    let langSelector = document.querySelectorAll('select.autosubmit');
    for (let menu of langSelector) {
```

Then again, looking at the HTML, could we do:

```suggestion
    let langSelector = document.getElementById('language');
    langSelector.addEventListener('change', function() {
        this.form.submit();
    });
```

peterbe: Surely `for (let menu of document.querySelectorAll('select.autosubmit')) {` only executes the `document.querySelectorAll('select.autosubmit')` call just once, no?

schalkneethling: Sure, but it seems like we do not even need a `for` loop.

davidflanagan: The jQuery version of the code doesn't need a for loop because jQuery magically handles multiple items as if they were one item. As far as I can tell, there is only one 'select.autosubmit' element on the page, but I didn't want to rule out the possibility that we might have them elsewhere as well. But I could switch to querySelector() instead of querySelectorAll() and drop the loop.

But the loop is efficient as it is written. The querySelectorAll() call returns an iterable object, and the for/of loop then iterates that object. querySelectorAll() itself is only called once.

davidflanagan:
Thanks Schalk. I've fixed the typos, but I left the for/of loop alone. Let me know if you think I should just switch to querySelector() and make the code specific to this one language menu and not make it general for any other hypothetical 'select.autosubmit' elements.

PR #5448, opened 2019-05-24
Use React effect hooks to fix up article section headings.

davidflanagan:
Kuma runs two JavaScript modules on page load to add styles and
anchors to certain section heading within the main article content.
This PR adds those features to the React version of the site. It
moves the Article component from page.jsx into its own article.jsx
file, and then adds two useEffect() hooks to run code over the
article HTML content each time the article is re-rendered.

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -1,0 +1,140 @@
+// @flow
+import * as React from 'react';
+import { useEffect, useRef } from 'react';
+import { css } from '@emotion/core';
+
+import ClockIcon from './icons/clock.svg';
+import ContributorsIcon from './icons/contributors.svg';
+import { getLocale, gettext } from './l10n.js';
+
+import type { DocumentData } from './document-provider.jsx';
+type DocumentProps = {
+    document: DocumentData
+};
+
+// A media query that identifies screens narrower than a tablet
+const NARROW = '@media (max-width: 749px)';
+
+const styles = {
+    article: css({
+        gridArea: 'main',
+        boxSizing: 'border-box',
+        width: '100%',
+        overflowX: 'scroll',
+        // Less padding on the left because the sidebar also has
+        // padding on the right
+        padding: '30px 24px 30px 12px',
+        [NARROW]: {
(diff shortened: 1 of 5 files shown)


kuma/javascript/src/article.jsx
```diff
@@ -0,0 +1,140 @@
...
+// This is the React version of the pre-React code in
+// kuma/static/js/components/local-anchor.js
+function addAnchors(article) {
+    for (let heading of article.querySelectorAll('h2[id], h3[id]')) {
+        let anchor = document.createElement('a');
+        anchor.href = `#${heading.id}`;
+        anchor.classList.add('sectionLink');
+        anchor.textContent = ' \uD83D\uDD17'; // Unicode link emoji
+        heading.insertAdjacentElement('beforeend', anchor);
+    }
```
davidflanagan: Note that this is much simpler than the code it replaces because:
- it uses a font glyph for the link
- it drops the hidden <span>Section</span>. That was for screen readers, but I didn't understand why it was useful, so I cut it.
- it just uses bare hashes as relative URLs instead of appending them to the link
- it doesn't try to manage scrolling and just allows the browser to handle the hashes on its own

kuma/static/js/highlight.js
```diff
@@ -1,16 +1,19 @@
...
     'use strict';
 
     // find all targets and wrap appropriately
     function highlight(targets) {
         var $targets = $(targets);
         $targets.each(function() {
-            $(this).addClass('highlight-spanned').wrapInner('<span class="highlight-span"></span>');
+            $(this)
+                .addClass('highlight-spanned')
+                .wrapInner('<span class="highlight-span"></span>');
```
davidflanagan: prettier rewrapped this one line when I added the comment above. It seemed harmless so I left the change in. Let me know if you'd like me to avoid reformatting these older JS files.

davidflanagan:
Screen Shot 2019-05-24 at 5 01 45 PM

schalkneethling:
@davidflanagan Couple of conflicts here that needs a rebase. Thanks!

PR #5450, opened 2019-05-28
Bug 1555004: stop including the react.js strings file into beta pages

davidflanagan:
For reasons I do not understand we've all-of-a-sudden started getting
500 errors from the server because the react.js strings files aren't
properly in the django staticfiles manifest. Rather than figure out
the source cause, this PR just removes the react.js include from the
templates that use it, because with the new React localization system
we don't actually use it anymore, and it should have been removed
a couple of weeks ago.

This is basically a partial revert of commit
bf5fad61ac58f22f5f3b7fcb712cd2f8f0e2dab9 which is what set up
that react.js file in the first place. React localization is now
handled during SSR with the react.json file. So this PR reverts
react.js back to the original javascript.js which provides localization
for any strings used by non-React JavaScript modules.  (When we remove
those non-React JavaScript modules, or at least all of them that use
strings, we'll be able to remove this javascript.js strings file
as well.

### DIFF
--- a/jinja2/react_base.html
+++ b/jinja2/react_base.html
@@ -152,7 +152,7 @@
     <!--[if lte IE 8]>{% javascript 'selectivizr' %}<![endif]-->
 
     {{ providers_media_js() }}
-    <!-- strings used by javascript modules. Hopefully we can remove soon -->
+    {# Strings used by javascript modules. Hopefully we can remove soon. #}
     <script src="{{ statici18n(request.LANGUAGE_CODE) }}"></script>
     {% javascript 'react-main' %}
     <script>
--- a/jinja2/react_base.html
+++ b/jinja2/react_base.html
@@ -152,7 +152,8 @@
     <!--[if lte IE 8]>{% javascript 'selectivizr' %}<![endif]-->
 
     {{ providers_media_js() }}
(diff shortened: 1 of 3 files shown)


jinja2/react_base.html
```diff
@@ -152,7 +152,8 @@
     <!--[if lte IE 8]>{% javascript 'selectivizr' %}<![endif]-->
 
     {{ providers_media_js() }}
-    <script src="{{ react_i18n(request.LANGUAGE_CODE) }}"></script>
+    <!-- strings used by javascript modules. Hopefully we can remove soon -->
```
peterbe:
```suggestion
    {# strings used by javascript modules. Hopefully we can remove soon #}
```

Jinja comment instead since this is a comment that doesn't need to be seen beyond our code base.

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -144,9 +144,9 @@
     </div>
   </footer>
 
+  <!-- strings used by javascript modules. Hopefully we can remove soon -->
```
peterbe: ditto

davidflanagan:
Changes made. Thanks for the quick review @peterbe.

PR #5451, opened 2019-05-28
bug 1475383 - Remove Facebook link and icon from footer

peterbe:
The `.svg` file is still used on the user profile page. Can you think of any other references to it?

Screen Shot 2019-05-28 at 2 06 11 PM


schalkneethling:
@peterbe It is supposed to load on the profile edit view as well, but only the Github icon loads there at the moment.

PR #5452, opened 2019-05-28
Make the maximum width of React-based page layouts 1400px

davidflanagan:
See https://github.com/mdn/sprints/issues/1588


davidflanagan:
This PR just makes it so that once the window gets to 1400px wide, we stop making the titlebar, breadcrumbs, or article content any wider (though we still allow the header to grow).

It seemed like a simple change, but it required a surprising amount of changes to the css and document structure. Sorry if this is a pain to review.

Here's what the layout now looks like at 1600px wide:

Screen Shot 2019-05-28 at 1 45 06 PM

PR #5454, opened 2019-05-29
Convert the beta page menus back to hover behavior instead of click

davidflanagan:
This PR updates the design of our menus and converts them back to
the original hover behavior, with a smaller target area so that they
don't pop up as easily. And it restores the links that were previously
available when clicking directly on the menu labels.

The hover behavior works with simulated touch events in chrome dev tools
but not in firefox dev tools. I'd like to just deploy this as is and then
we can more easily test with a real touch device.

Because this new menu does not respond to click events, I've had to delete
some test cases that were testing clicks on the menu.

### DIFF
--- a/kuma/javascript/src/__snapshots__/page.test.js.snap
+++ b/kuma/javascript/src/__snapshots__/page.test.js.snap
@@ -98,8 +98,22 @@ Array [
 
 .emotion-9 {
   position: relative;
-  pointer-events: auto;
   margin-left: -5px;
+}
+
+.emotion-9:hover .dropdown-menu-label {
+  color: #3d7e9a;
+}
+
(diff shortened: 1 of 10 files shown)


davidflanagan:
Lots of snapshot files changed, but those should be ignorable. Mostly this is a simplification to dropdown.jsx.

Screen Shot 2019-05-28 at 5 39 27 PM

schalkneethling:
Oh snap ;p A conflict has made itself known @davidflanagan

davidflanagan:
Thanks Schalk. Those test snapshots are like honeypots for merge conflicts. If we decide to get rid of them, maybe that will be the reason. I've pushed an update with the latest page snapshot, resolving the conflict.

PR #5455, opened 2019-05-29
Bug 1553075, add speedcurve lux script

schalkneethling:
@davidflanagan r?


jinja2/base.html
```diff
@@ -176,5 +180,6 @@
   {% endblock %}
 
   {% block js %}{% endblock %}
+  <script src="https://cdn.speedcurve.com/js/lux.js?id=108906238" async defer crossorigin="anonymous"></script>
```
davidflanagan: Given that this script is async defer, can we just add it to the end of speedcurve-lux.html for simplicity? That would presumably mean a performance hit for browsers that don't support those things, but I'm not sure if we care?

davidflanagan: caniuse.com says that async and defer are both supported by IE10 and IE11...

jinja2/includes/speedcurve-lux.html
```diff
@@ -0,0 +1,3 @@
...
...s){r=s.startTime}else{if(k&&k.timing&&k.timing[t]){r=k.timing[t]-k.timing.navigationStart}else{return}}}if(n){var q=h(n);if(q){o=q.startTime}else{if(k&&k.timing&&k.timing[n]){o=k.timing[n]-k.timing.navigationStart}else{return}}}d.push({name:p,entryType:"measure",startTime:r,duration:(o-r)});return}function h(n){return c(n,g())}function c(p,o){for(i=o.length-1;i>=0;i--){var n=o[i];if(p===n.name){return n}}return undefined}function g(){if(k){if(k.getEntriesByType){return k.getEntriesByType("mark")}else{if(k.webkitGetEntriesByType){return k.webkitGetEntriesByType("mark")}}}return a}return{mark:b,measure:m,gaMarks:a,gaMeasures:d}})();LUX.ns=(Date.now?Date.now():+(new Date()));LUX.ac=[];LUX.cmd=function(a){LUX.ac.push(a)};LUX.init=function(){LUX.cmd(["init"])};LUX.send=function(){LUX.cmd(["send"])};LUX.addData=function(a,b){LUX.cmd(["addData",a,b])};LUX_ae=[];window.addEventListener("error",function(a){LUX_ae.push(a)});LUX_al=[];if("function"===typeof(PerformanceObserver)){var LongTaskObserver=new PerformanceObserver(function(c){var b=c.getEntries();for(var a=0;a<b.length;a++){var d=b[a];LUX_al.push(d)}});try{LongTaskObserver.observe({entryTypes:["longtask"],buffered:true})}catch(e){}};
```
davidflanagan: Should we be doing the !Mozilla.dntEnabled() thing here like we do in google_analytics.html? I suppose that might be a question for Kadir.

If so, that presumably means that the external script should also only be loaded in the no-dnt case, which means that we'd need to load that external script dynamiclly the way GA does it.

This code from the GA script could probably be repurposed: 

```
var a=s.createElement('script'),
var m=s.getElementsByTagName('script')[0];
a.async=1; // Don't know why GA isn't using defer, but we presumably would
a.src=LUX URL HERE
m.parentNode.insertBefore(a,m)
```

And if we're doing the DNT thing for both GA and LUX, then does it make sense to combine both into a single analytics.html include rather than having two separate files? (I don't have an opinion myself.)

jinja2/react_base.html
```diff
@@ -46,6 +46,10 @@
     {% include "includes/google_analytics.html" %}
   {% endblock %}
 
+  {% block speedcurve_lux %}
+    {% include "includes/speedcurve-lux.html" %}
+  {% endblock %}
```
davidflanagan: I find jinja templates confusing, so I removed as much of the inheritance as I could from kuma/wiki/jinja2/wiki/react_document.html.  Which means that it no longer inherits from react_base.html, and you'll need to add these lines explicitly to react_document.html.

react_base.html is still used by the react homepage in kuma/landing/jinja2/landing, but I might be changing that soon....

schalkneethling:
Thanks, @davidflanagan 

> And I think we need to decide what to do about LUX if DNT is enabled.

DNT is going away[1][2], so we do not have to be concerned about that here. 

[1] https://gizmodo.com/apple-is-removing-do-not-track-from-safari-1832400768
[2] https://www.w3.org/2011/tracking-protection/

I will make the other two updates as mentioned.

schalkneethling:
@davidflanagan updated.

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -49,6 +49,12 @@
...
+</script>
+<script src="https://cdn.speedcurve.com/js/lux.js?id=108906238" async defer crossorigin="anonymous"></script>
+
```
davidflanagan: I'd slightly prefer to use an include here rather than than repeating the snippet out in full.

I couldn't do that for the google analytics include files since it was sending user-specific data that we don't have here.

PR #5457, opened 2019-05-29
Remove mdn.utils.randomString()

davidflanagan:
PR #5445 causes a JS error in the developer console because a JS file
was is missing its the utils.js dependency. The problem is the
randomString() utility. Rather than include the entire utils.js file
this PR just removes the randomString() function and replaces the two
places it was called with a more efficient alternative.

### DIFF
--- a/kuma/static/js/interactive.js
+++ b/kuma/static/js/interactive.js
@@ -62,7 +62,11 @@
                     category: 'Interactive Examples',
                     action: 'Time to iframe fetch start',
                     label:
-                        mdn.utils.randomString(5) + '-' + new Date().getTime(),
+                        Math.random() // 5 random letters and digits
+                            .toString(36)
+                            .slice(-5) +
+                        '-' +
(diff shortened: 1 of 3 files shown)


davidflanagan:
This fixes a console error I caused with PR #5445.

randomString() was only used in two places, and my replacement should be faster because no loop is required. Less randomness, though, but 36**5 ought to be enough distinct values for our purposes.

davidflanagan:
Thanks for the review @peterbe. What do you mean by "repeating the code"? `Math.random().toString(36).slice(-5)` seems short enough to me that it is not worth defining a utility function for.

If you mean why are those two .js files so similar, I don't have an answer for you... Our analytics code seems to be a mess.

peterbe:
To be clear, I wasn't so worried about DRY in the traditional sense. It was more the fact that what if it matters. What if we in one place we realize it needs to be longer or mustn't start with a digit and then forget to update it in the other place.

PR #5459, opened 2019-05-29
Add Edit and History buttons to the titlebar for logged-in users.

davidflanagan:
Clicking these buttons takes the user to the wiki site. Currently
all logged-in users will see the buttons, but this PR includes some
code that makes a distinction between contributors and logged in
users who are not contributors. Once we've made that distinction
on the backend and in the whoami API, the frontend will show the
buttons to contributors onlyi.


davidflanagan:
Here's what it looks like:

Screen Shot 2019-05-29 at 3 27 47 PM

PR #5462, opened 2019-05-30
Move the TOC (the old black "jump bar") into the sidebar

davidflanagan:
This PR moves the table of contents for the page into the sidebar.
These are the links to internal sections that used to be in the
"jump bar". This is based on preliminary designs that Nick has
shown me and he's given approval to land it in this state.

It is still to be decided whether we want to show the TOC on all
pages or only on the Learn/ pages.


davidflanagan:
Here's what it looks like: 

Screen Shot 2019-05-30 at 10 19 52 AM

PR #5463, opened 2019-05-30
Convert language menu to display language name instead of icon

davidflanagan:
This PR deletes the language.svg icon and changes the LanguageMenu
component to display the name of the current language instead of
that icon.

The PR also fixes a bug in the Dropdown menu component that was
requiring the user to click directly on the text of a menu item.
With the new (and simplified) CSS, a click anywhere within the
highlighted rectangle will work to trigger the menu item


davidflanagan:
Sorry about all the changed snapshot files... I'm really beginning to have second thoughts about those tests. The only code changes that need review are in header/dropdown.jsx header/language-menu.jsx

davidflanagan:
Since the language menu is no longer part of the header, I thought about moving it up out of the header/ directory. But if I did that, then I'd also probably need to move dropdown.jsx out of the header/ directory. Plus moving tests and shapshots, that seemed like more churn than I wanted to create right now.

But let me know if you think I should do that move as part of this PR.

PR #5464, opened 2019-05-30
Add a React header to the beta homepage.

davidflanagan:
This is a complicated PR that, for the first time, sets up the
backend and frontend infrastructure to enable React to be used
on more than one endpoint.

The immediate goal of this PR is to add the React-based Header
component at the top of the homepage on the beta site. The rest
of the homepage is still rendered by Kuma based on Jinja templates.

In order to make this work, I've made a number of changes:

- in wiki/templatetags/ssr.py, the render_react_app() tag is
  renamed to render_react(). It takes a new component_name
  argument, and the unneeded request_data argument is removed.

- react_document.html is changed to call render_react with a
  component_name of "document". And react_homepage.html is
  modified to add a render_react tag with a component_name of
  "landing".

- The implementation of render_react uses the component name as
  a data attribute on the react container div so that client side
  code can use that name to decide which component to render into
  that container.

- And for server-side rendering, the render_react tag adds the
  component name to the SSR server URL that it POSTs to. This gives
  the SSR server a way to know which component to render.

- ssr-server.js is modified to expect the component name in the
  path and to pass that component name to the ssr() function in
  ssr.jsx.

- The app.jsx file and its App component are renamed to
  document-page.jsx and DocumentPage since it is no longer the
  only top-level component. Also, this PR adds a new LandingPage
  component in the landing-page.jsx file.

- ssr.jsx is modified so that it considers the component name.
  If it sees "document" it renders the DocumentPage component.
  If it sees "landing" it renders the LandingPage component.

- index.jsx is similarly modified. It gets the component name
  from the data-component-name attribute of the container div
  and uses it to decide whether to render a DocumentPage or
  a LandingPage. In the future, if we add new pages, we'll just
  have to add new cases to the switch statements in index.jsx
  and in ssr.jsx.

- Note that this component does not change the webpack config
  to create a separate bundle for the homepage and document pages.
  I think it makes more sense to have a single bundle that ends up
  in the user's browser cache.

- The homepage just renders a Header component and does not need
  any document data to do that. But Header and the Search and
  Login components that it uses all had minor dependencies on
  document data. So the PR also modifies those components to break
  the dependencies so that the Header can be displayed even when
  there is no document data.


davidflanagan:
I'm not sure who would be best to review this, but I'm flagging all three of you so that this is at least on your radar, since this touches a lot of the underlying architecture of our React setup.

See the long commit message for a guide to this patch. I'd start with react_document.html and react_homepage.html. Then review the changes to ssr.py  Then ssr-server.js, ssr.jsx and index.jsx. Next document-page.jsx and landing-page.jsx, and then finally the component changes in the header/ directory.

kuma/javascript/src/ssr.jsx
```diff
@@ -22,7 +24,53 @@ export default function ssr(data) {
...
+    // let stylesheets = [];
+    // html = html.replace(
+    //     /<style data-emotion-css="[^"]+">(.*?)<\/style>/g,
+    //     function(match, styles) {
+    //         stylesheets.push(styles);
+    //         return '';
+    //     }
+    // );
+    // return `<style>\n${stylesheets.join('\n')}\n</style>\n${html}`;
+    //
```
davidflanagan: @peterbe: this long comment is here for your benefit. I tried to consolidate all of the stylesheets, and found that it broke the hydration process. And in the process I realized that maybe lots of little stylesheets is not that inefficient anyway.  

But anyway, this comment and the comment code are preserved here in case you want to attempt the consolidation again in the future.

davidflanagan:
Homepage with header:

Screen Shot 2019-05-30 at 4 18 05 PM

I'm not sure why the default avatar is not showing up, but I don't think this PR broke it. I think it is something to do with my local environment.

kuma/javascript/src/ssr.jsx
schalkneethling: That should not be a blocker at this stage I don't think.

kuma/javascript/src/header/login.jsx
```diff
@@ -52,27 +51,22 @@ const styles = {
...
-    const WIKI_SITE_URL = editURL.substring(0, editURL.indexOf(absoluteURL));
-
     // if we don't have the user data yet, don't render anything
-    if (!userData) {
+    if (!userData || typeof window === 'undefined') {
         return null;
     }
 
+    // In order to render links properly, we need to know our own
+    // URL and the URL fo the editable wiki site. We get these from
```
schalkneethling:
```suggestion
    // URL and the URL of the editable wiki site. We get these from
```

kuma/javascript/src/header/search.jsx
```diff
@@ -39,25 +38,30 @@ const styles = {
 };
 
 export default function Search() {
+    // In order to render the action attribute of the form, we need
+    // to know the URL fo the wiki site (which has the /search endpoint).
```
schalkneethling:
```suggestion
    // to know the URL of the wiki site (which has the /search endpoint).
```

PR #5466, opened 2019-05-31
bug 1532736: increase rate limit on document views

escattone:
This PR simply increases the rate-limiting threshold for both the React and Non-React document views in order to avoid rate-limiting the Google bots. See https://github.com/mdn/sprints/issues/1629 for more context.

### DIFF
--- a/kuma/wiki/views/document.py
+++ b/kuma/wiki/views/document.py
@@ -604,7 +604,7 @@ def _document_raw(doc_html):
 @allow_CORS_GET
 @process_document_path
 @newrelic.agent.function_trace()
-@ratelimit(key='user_or_ip', rate='800/m', block=True)
+@ratelimit(key='user_or_ip', rate='1200/m', block=True)
 def document(request, document_slug, document_locale):
     """
     View a wiki document.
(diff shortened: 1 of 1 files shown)


peterbe:
I restarted Travis. Seemed transient.

peterbe:
Are there no better solutions to this that are much more performant that poor Django slinging back and forth to Redis? 
Ideally something on the AWS load balancer maybe. I appreciate that it might be hard for such a solution to have custom business logic such as adjusting the rate limit depending on a bunch of kuma specifics.

escattone:
Oops, I used the wrong bug number. I should have used 1532736 instead. I'll adjust.

peterbe:
> Oops, I used the wrong bug number. I should have used 1532736 instead. I'll adjust.

You might need to correct the comment made on 1518482 but you can, right now, just edit the PR title and it'll get right when we Squash and merge.

peterbe:
I editted the title. Feel free to merge away when you're back and Travis is happy again.

escattone:
Thanks @peterbe!

PR #5467, opened 2019-05-31
Implement search on the beta domain

davidflanagan:
This PR adds a simple new /api/v1/search?q=<query> API on the backend
and new frontend React components to display search results.

More importantly, however, this PR generalizes the React single page
app framework. Previously the DocumentProvider component did all the
work and supported client-side navigation between document pages.
With the introduction of search results pages, however, our single
page app now has two different kinds of pages to display and has to
be able to do client side navigation between them.

This PR removes DocumentProvider and replaces it with a generalized
Router component that can handle client side navigation among any
number of defined "routes". The PR defines a DocumentRoute for handling
document urls and pages and a SearchRoute for handling search result
urls and pages.

The Page component that was in page.jsx has been renamed Document and
moved to document.jsx, which also defines the DocumentRoute object.
Similarly, the SearchResultsPage component is defined along with the
SearchRoute object in search-results-page.jsx.

My goal is to encapsulate all functionality related to page loading
into the Router component. So the loading bar animation has moved from
the Header component into the Router component.

This PR touches many parts of the code, and there are many minor
changes throughout. The main chunk of new code, however is in router.jsx.

The search results page UX is preliminary, and I will continue to
refine it once the larger single-page-app framework has landed.


davidflanagan:
This is still a work in progress, but I figured I'd try out the draft PR feature.

Ryan: I've flagged you for review because we talked about this this morning.

Peter: I'm flagging you because you might have elasticsearch suggestions for me. I've created a new API that is much, much simpler than what is in kuma/search, and I'm hoping I can get away with it. But I barely understand ES... Maybe you can tell me if this looks reasonable.

davidflanagan:
This is turning into a much bigger frontend refactor than I anticipated. Don't look at the frontend code at this point.  Feedback on the backend search API would be very helpful, however.

kuma/api/v1/views.py
```diff
@@ -206,3 +207,19 @@ def whoami(request):
     }
 
     return JsonResponse(data)
+
+
+@never_cache
+@require_GET
+def search(request, locale):
+    qs = request.GET.get('q')
+    search = (WikiDocumentType.search()
```
escattone: I don't know enough about Elasticsearch to review this, so I'm going to defer to @peterbe here.

kuma/urls_beta.py
```diff
@@ -8,7 +8,7 @@
 from kuma.core.urlresolvers import i18n_patterns
 from kuma.landing.urls_beta import lang_urlpatterns as landing_lang_urlpatterns
 from kuma.views import serve_from_media_root
-
+import kuma.wiki.views.document
```
escattone: flake8 issue: this `import` can be removed.

peterbe:
This is what it looks like:
Screen Shot 2019-06-05 at 4 05 39 PM

@davidflanagan Do you want to clean up all those console loggings as part of the draft or shall we worry about that when the PR becomes a real PR?

peterbe:
Not sure if it's related but after having eyed the search results page, I clicked on a link in the top nav drop-down menus and this happened:
Screen Shot 2019-06-05 at 4 07 29 PM
Note the `TypeError: R.setLoading is not a function`

peterbe:
I really like how [in prod](https://developer.mozilla.org/en-US/search?q=web&topic=apps&topic=html&topic=css&topic=js&topic=api&topic=canvas&topic=svg&topic=webgl&topic=mobile&topic=webdev&topic=http&topic=webext&topic=standards) and [on Google](https://www.google.com/search?client=firefox-b-1-d&q=search+fragments) and [on Peterbe.com](https://www.peterbe.com/search?q=web) each search result has the absolute URL. It might seem a silly thing to output but it's such an affordance that without you don't really know what you're looking at. 

Also, I think the titles should be bold so that it better highlights the relationship.

peterbe:
I think whatever I typed in the search input widget should remain. That way I can easily edit my search to a better one. Also, it's just something very expected since that's how most site searches work. 
Screen Shot 2019-06-05 at 4 14 10 PM

peterbe:
There is no pagination. Feels like such "unimportant" feature that pointing it out is like pointing out a nit. Perhaps, where the "Next" button would be, make a code comment pointing to the Bugzilla URL about implementing React search pagination. Then I'd be happy *not* blocking this PR just because of lack of pagination. If that makes sense.

peterbe:
Not sure if it's deliberate (it's a draft PR after all) but the home page's search widget goes to the wiki still.

davidflanagan:
@peterbe: the frontend is completely unfinished, and also what is on github is now out of date with what is in my local tree. I would be very interested in what you think of the elastic search code in kuma/api/v1/views.py, however!

I think we're better off without pagination: I'd rather just show 10 or 20 results and be done with it. Unless actual users complain. But it would be pretty easy to add a second page if we actually want that.

And I don't understand your commet at at https://github.com/mozilla/kuma/pull/5467#issuecomment-499237121  Search URLs /search?q= work (or will work if they aren't in the version  you've got) on the beta site.

peterbe:
> I think we're better off without pagination

👍🏼

> And I don't understand your commet at https://github.com/mozilla/kuma/pull/5467#issuecomment-499237121

Perhaps moot since I was reviewing something that wasn't ready. Let's ignore me till this PR is ready for review.

peterbe:
@davidflanagan Regarding the Python side, do you want me to attempt a fix where the api view function leverages the existing Python code in `kuma/search/views*`?

davidflanagan:
@peterbe: I thought it was simpler to just use the ES API directly. After trying to figure out kuma/search/views, it was liberating to realize that I could just issue my own simpler queries to ES directly.

I've got an implicit hypothesis that the query in kuma/search/views has grown really complicated over the years as we've tried to tweak it, but that ES is a robust piece of software and that if we start over with a relatively simple query we can probably get search results that are just as useful as what we have now, and we can execute them more quickly. I've got no data to back that up. But I also don't have any reason to accept that we need to do a query as complicated as what we are doing now.  (Except, I think that I'm not currently handling searches for non-english locales correctly... In that case I think we should do two searches in parallel, one for the locale and one for english...)

Also, if we do implement some kind of autocomplete system for search, I think that will be a search against title only (or maybe title and slug) and we'll want it to be as simple and fast as possible. That query, at least, we would not want to use the stuff in kuma/search.

Having said all this, I've got no practical experience with elastic search, so tell me if I'm way off base here.

peterbe:
@davidflanagan Fair point but I don't think it would be that painful to extract it. I know I've seen bugs related to the boosting being done poorly. It would be nice to solve that only in one place. 
The one thing to watch out for is that the "wiki ES search" uses those filters which I still don't grok. 
Perhaps we just go ahead with it in this form for now and make a mental note (or better, a bug and a code comment pointing to that bug URL) for the near future.

davidflanagan:
@schalkneethling This has turned into a massive PR that generalizes all of our client-side-navigation code. I've updated the description comment above with details. But if you have time to start looking at this, the router.jsx file is the place to begin. Then document.jsx and search-results-page.jsx.

(Don't be scared off by the 4000 changed lines figure: most of that is snapshot files) But the big changes in this PR did cause lots of little changes throughout the code. Sorry about that.

davidflanagan:
I should add that I've gotten this code to the point where the tests are passing locally and eslint and flow are happy with it. But I have not written tests for router.jsx or for the two new Route objects or for the search features.

kuma/javascript/src/article.jsx
```diff
@@ -89,8 +89,16 @@ function addAnchors(article) {
...
+    useEffect(() => {
+        if (article.current) {
+            highlightSections(article.current);
+        }
+    }, [document]);
+    useEffect(() => {
+        if (article.current) {
+            addAnchors(article.current);
+        }
+    }, [document]);
```
jpetto: Wondering about the performance implications of comparing `document` after every render. Is there a smaller bit of data that could be compared to achieve the same result?

peterbe: It's not a deep comparison. Just checking if it's the same object. So dirt-cheap in terms of performance :)

kuma/javascript/src/document.jsx
```diff
@@ -314,41 +223,132 @@ function Content({ document }: DocumentProps) {
...
-     * rendered. And then it turns off the loading indicator for the page
-     *
-     * Effects don't run during server-side-rendering, but that is fine
-     * because we only want to send the GA event for client-side navigation.
-     * Calling navigateRenderComplete() will have no effect if
-     * navigateStart() was not previously called.
-     */
-    useEffect(() => {
-        navigateRenderComplete(ga);
+type Props = { data: ?DocumentData };
```
jpetto: Haven't seen this syntax before - `?` before the type. What's up with that? (Searching the internet was fruitless. 😄)

kuma/javascript/src/article.jsx
davidflanagan: I think Peter is right about this, but the official documentation is unclear. But googling turns of lots of hits for people trying to make useEffect() do deep comparison, so I assume this really is just testing reference equality.

But thank you for bringing this up Jon. I should arguable be comparing the actual strings of html that the Article component displays instead of the object that contains them. I don't believe that our architecture ever allows the html to change without creating a new documentdata object, though so it is probably okay as it stands.

kuma/javascript/src/document.jsx
davidflanagan: The ? means optionally null in Flow.

kuma/javascript/src/router.jsx
```diff
@@ -0,0 +1,570 @@
...
+    // This is an effect function that runs once and register the
+    // event listener that intercepts form submissions (such as on the
+    // MDN search box) to enable client side navigation.
+    function interceptFormSubmissions() {
+        // Flow complains if we don't verify this
+        if (!document || !document.body) {
+            return;
+        }
+
+        document.body.addEventListener('submit', (e: Event) => {
```
jpetto: Very cool (and convenient) that this works. 😄

kuma/javascript/src/article.jsx
jpetto: Excellent 😁

kuma/javascript/src/document.jsx
jpetto: Ah! I assumed it was something to do with nullable/optional values.

(I also had it in my head that you were using TypeScript, hence my internet search failure.)

jpetto:
I think this is looking great overall. Though I'm pretty on the outside of the project, it reads/feels like a very good balance of leveraging the React way and custom building the things that make sense for this particular use case (e.g. the router).

kuma/javascript/src/article.jsx
peterbe: The repetition just looks odd. Is there a strong reason to break up into multiple `useEffect` instead of one?

schalkneethling: I was wondering the same thing.

davidflanagan: I'm calling useEffect() twice because these are literally two different effects I'm using. We could combine them, but I think the perf difference would be insignificant.

Note that the two effects were already here in this component. The change here is bug fix where the effects were running too often (and so we were getting multiple link icons in each section heading)

kuma/api/v1/views.py
peterbe: I still think we should do a followup bug to see if we can extract this functionality somewhere from `kuma/search/views.py`. I don't think we need to file that bug right now.

kuma/javascript/src/search-results-page.jsx
```diff
@@ -0,0 +1,104 @@
...
+            <Header />
+            <Titlebar title={`${gettext('Search Results')}: ${query}`} />
+            {data &&
+                data.map(hit => {
+                    let url = `/${locale}/docs/${hit.slug}`;
+                    return (
+                        <div css={styles.searchResult} key={hit.slug}>
+                            <a css={styles.searchResultLink} href={url}>
+                                {hit.title}
+                            </a>
```
peterbe: I think it would be swell to display the URL of the search result here. Do you think we should add that later?

kuma/javascript/src/search-results-page.jsx
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        // If the path didn't match or we didn't find a query,
+        // we can't handle this URL
+        return null;
+    },
+
+    fetch({ locale, query }: SearchRouteMatchData): Promise<SearchResults> {
+        return fetch(`/api/v1/search/${locale}?q=${encodeURIComponent(query)}`)
+            .then(response => response.json())
```
peterbe: This is too optimistic. The API call might return a failed response. Either because of network errors so we need a `.catch` and a way to display that the search API call failed. 
Also, what if the `!response.ok` because something else is wrong. I can't think of an example but perhaps if the `query` is an empty string the search API might respond with `400 Bad Request: "search string can't be empty"`.

kuma/javascript/src/document.jsx
```diff
@@ -314,41 +223,132 @@ function Content({ document }: DocumentProps) {
...
-                <TaskCompletionSurvey />
-                <Titlebar document={document} />
-                <Breadcrumbs document={document} />
-                <Content document={document} />
-            </>
-        )
-    );
+// Like fetch(), but if the first URL doesn't work, try the second instead.
+function fetchWithFallback(url1: string, url2: string) {
+    const fetchConfig = { redirect: 'follow' };
```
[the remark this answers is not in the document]
davidflanagan: I think so. Ryan wasn't sure when he wrote that code. This PR just moves the code to a new location.

kuma/javascript/src/document.jsx
```diff
@@ -314,41 +223,132 @@ function Content({ document }: DocumentProps) {
...
+        return fetchWithFallback(
+            `/api/v1/doc/${locale}/${slug}`,
+            `/api/v1/doc/en-US/${slug}`
+        )
+            .then(response => {
+                if (response.ok) {
+                    return response.json();
+                } else {
+                    // If we didn't get a good response, throw an error
+                    // that we'll handle in the catch() function below.
```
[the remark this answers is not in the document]
davidflanagan: There used to be one below. Given the restructuring, the comment probably doesn't make sense anymore. Will fix.

kuma/javascript/src/search-results-page.jsx
davidflanagan: This PR started out about the search-results-page and then turned into a bigger one about the router component. I agree that there is still work to do for the search results page. And yes, since the existing non-React version displays the URL, I'll add that as well, though maybe in a followup PR.

Also since we're removing all the checkboxes to filter searches, I'm thinking it might be useful to display document tags along with each result. ElasticSearch will return them to us, so I'll probably try adding that too. Might be useful to distinguish the HTML Canvas page from the Web API Canvas page, for example.

kuma/javascript/src/search-results-page.jsx
davidflanagan: Ideally error handling would happen in the router. And I think there is code in there that falls back to doing a full page reload if we get a failure on client-side navigation. (I'm a little worried about causing an infinite loop of reloads, though.)

Maybe a better approach here would be to move this fetch out of the search route and into the search results component. That is: when you search we would always display an empty search results page right away, with a spinner, and then the search results page would do its own fetch, and could handle its own errors.

Are you okay with just adding a TODO comment for now and coming back to this after getting the router landed?

kuma/javascript/src/document.jsx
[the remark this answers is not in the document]
davidflanagan: I've taken it out. We shouldn't be getting redirects from an API endpoint anyway!

kuma/javascript/src/search-results-page.jsx
davidflanagan: Added a long TODO comment about these issues.

davidflanagan:
I still need to write tests. But I've made some tweaks, making Route into an actual JS class instead of just a Flow type, and in the process fixing the locale issue where it would route across locales when it should not have. I hope to wrap this up soon with tests.

davidflanagan:
I've addressed Peter's review comments, at least with TODOs, and have added tests, so I'm going to land this now, and then continue to improve the look and quality of the search results in follow-on PRs

davidflanagan:
(Assuming the transient test failure is resolved by restarting the job, that is.)

PR #5469, opened 2019-06-03
Browser compatibility data signal feature

johngian:


kuma/static/styles/components/compat-tables/bc-signal.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding-right: 15px;
+    padding-left: 15px;
+    min-width: 230px;
+}
+
+@media #{$mq-tablet-and-down} {
+    div.signal-link-container + section {
+        padding-top: 1.5 * $grid_spacing;
+    }
+}
```
schalkneethling: As this is Sass, I would suggest rewriting this as follows:

```
div.signal-link-container {
  position: absolute;
  width: 100%;
    
  & > hr {
    background-color: #3d7e9a;
    margin: 0;
    max-width: 100%;
  }
  
  & > a {
    float: right;
    background-color: #3d7e9a;
    color: #fff;
    padding-right: 15px;
    padding-left: 15px;
    border: 0;
    border-bottom-left-radius: 4px;
    border-bottom-right-radius: 4px;
    font-weight: bold;
    text-align: right;
    height: 35px;
    min-width: 230px;
  }
}

@media #{$mq-tablet-and-down} {
    div.signal-link-container + section {
        padding-top: 1.5 * $grid_spacing;
    }
}
```

schalkneethling: Some small stylistic changes included in the above as well.

kuma/static/js/wiki-compat-trigger.js
```diff
@@ -32,6 +32,20 @@
...
+
+                if (waffle.flag_is_active('bc-signals')) {
+                    // Load signaling feature
+                    var compatSignalJS = mdn.assets.js['wiki-compat-signal'][0];
+
+                    $.ajax({
+                        url: compatSignalJS,
+                        dataType: 'script',
+                        cache: true
+                    }).fail(function(jqXHR, textStatus, errorThrown) {
```
escattone: This fails to load when I enable our CSP middleware since it requires the addition of `'unsafe-eval'` to our CSP policy for `script-src`. Is there another way we could do this to avoid that? @schalkneethling @davidflanagan

kuma/static/js/wiki-compat-signal.js
```diff
@@ -0,0 +1,36 @@
...
+    function signalElem() {
+        var seperator = document.createElement('hr');
+        var container = document.createElement('div');
+        var signalLink = document.createElement('a');
+        var signalUrl = location.pathname + '$bc-signal';
+
+        signalLink.textContent = 'Flag as incorrect';
+
+        signalLink.addEventListener('click', function() {
+            $.get(signalUrl).done(function() {
```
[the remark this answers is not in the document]
schalkneethling: I agree. Let's not use jQuery for new code.

schalkneethling:
@escattone @davidflanagan @johngian I just hit me like a ton of bricks. 🙀 This should be built as a React component inside the new React codebase. If we merge this in as is(which is perhaps an option), we will have to port it to React within the next ~30 days if/when we switch to the React frontend for d.m.o

Thoughts?

schalkneethling:
I feel bad that this was communicated clearly so, @escattone if the rest of the code looks good to you, I am happy to merge this and do the rewrite. It would be a good opportunity for me to become more familiar with the new React codebase 😸

escattone:
@johngian After some more time to think about this, I would prefer a different endpoint than the document-based `/<locale>/docs/<slug>$bc-signal`. I would prefer a RESTful endpoint at `/api/v1/bc-signals` (or singular `/api/v1/bc-signal` if you prefer), to which, when we want to create a new signal, we POST a JSON blob (or multipart/form-data) that for now contains just the document, for example, something like `{document: "/en-US/docs/Web/CSS/-ms-scroll-snap-y"}`. I think this would give us several advantages:
- we don't use an idempotent HTTP method like GET to make a non-idempotent request
- as the client-side UX evolves, we can add more data to the JSON (or multipart/form-data) that we POST when creating a signal
- we could flesh-out the endpoint with other RESTful calls, like a GET to `/api/v1/bc-signals` that returns all of the signals (paginated) as JSON and/or a more specific GET for all of the signals for a specific document
- it's much easier to perform special tasks based on the path (like CDN behaviors that depend on the path) when the path is unique and doesn't share so much of the same structure as a document GET request

I've never like our `/<locale>/docs/<slug>$...` endpoints

escattone:
@johngian We're already using the `djangorestframework` a bit for search, so it's there to consider using for a `/api/v1/bc-signals` endpoint. It should take care of most if not all of the heavy-lifting and/or tedious work of creating that endpoint.

escattone:
@schalkneethling @davidflanagan I was thinking the same, that we'll have to get this working within the new React pages. @schalkneethling I like your idea of merging this, when it's ready, without the React version, and having you add that next.

escattone:
@schalkneethling @davidflanagan I'm not familiar enough with the front-end and how `kuma/static/js/wiki-compat-trigger.js` has been used with the compatibility tables and/or macro, but it seems that even before this PR it relied on dynamically loading Javascript and so would require the addition of `'unsafe-eval'` to our CSP policy for `script-src` in order to work (assuming that we enforced our current CSP policy, which we only do on stage). Can we avoid that somehow? Do we have to do it this way?

escattone:
@schalkneethling Is `kuma/static/js/wiki-compat-trigger.js` used for the new compat tables (generated by the `compat` macro`) as well as the old ones? Also it calls our `mozCompatTable()` (defined in `kuma/static/js/wiki-compat-tables.js`) on the compat table, but that seems to set-up a bunch of dynamic history stuff that doesn't seem necessary, even for the old tables. I'm out of my realm here though, so I'm probably missing something, but I'm wondering if `kuma/static/js/wiki-compat-trigger.js` and `kuma/static/js/wiki-compat-tables.js` are even necessary?

schalkneethling:
> but it seems that even before this PR it relied on dynamically loading Javascript and so would require the addition of 'unsafe-eval' to our CSP policy for script-src in order to work

@escattone Indeed, that "functionality" existed prior to this PR. I am not sure if this has changed at all on the React side, but I would not make it a blocker for this PR.

We definitely need to revisit it as it would not be great to have to add 'unsafe-eval' to our CSP policy

johngian:
> @johngian After some more time to think about this, I would prefer a different endpoint than the document-based `/<locale>/docs/<slug>$bc-signal`. I would prefer a RESTful endpoint at `/api/v1/bc-signals` (or singular `/api/v1/bc-signal` if you prefer), to which, when we want to create a new signal, we POST a JSON blob (or multipart/form-data) that for now contains just the document, for example, something like `{document: "/en-US/docs/Web/CSS/-ms-scroll-snap-y"}`. I think this would give us several advantages:
> 
>     * we don't use an idempotent HTTP method like GET to make a non-idempotent request
> 
>     * as the client-side UX evolves, we can add more data to the JSON (or multipart/form-data) that we POST when creating a signal
> 
>     * we could flesh-out the endpoint with other RESTful calls, like a GET to `/api/v1/bc-signals` that returns all of the signals (paginated) as JSON and/or a more specific GET for all of the signals for a specific document
> 
>     * it's much easier to make decisions based on the path (like CDN behaviors that depend on the path) when the path is unique and doesn't share so much of the same structure as a document GET request

It looks like ``/api/v1`` is only available to beta urls. How are you suggesting to handle it for all domains (not only beta, if i understood how this works) ?

escattone:
@johngian You're right on, we'd have to add it to `kuma.urls`. I was thinking you could define the endpoint/view within `kuma/api/v1/views.py`, and then we'd add something like the following to `kuma/urls.py`:
```
...
from kuma.api.v1.views import bc_signals
...
urlpatterns += [url(r'^/api/v1/bc-signals/?$', bc_signals, name='api.v1.bc-signals')]
...
```
and since we'd need it for the `beta` domain as well (which is using the new React code), you would also add it to `urlpatterns` within `kuma/api/v1/urls.py`.

kuma/static/js/wiki-compat-signal.js
```diff
@@ -0,0 +1,36 @@
...
+    'use strict';
+
+    var bcTable;
+
+    /**
+     * Creates and returns the signal element HTML
+     * @returns signal element HTML
+     */
+    function signalElem() {
+        var seperator = document.createElement('hr');
```
davidflanagan: Nit: misspelling of separator

kuma/static/js/wiki-compat-signal.js
```diff
@@ -0,0 +1,36 @@
...
+
+        signalLink.textContent = 'Flag as incorrect';
+
+        signalLink.addEventListener('click', function() {
+            $.get(signalUrl).done(function() {
+                signalLink.textContent = 'Thank you for letting us know!';
+            }).fail(function() {
+                signalLink.textContent = 'Something went wrong!';
+            }).always(function() {
+                $(container).children().delay(3000).slideUp('slow');
```
davidflanagan: Another use of jQuery that won't be quite so easy to replace...

kuma/static/js/wiki-compat-trigger.js
davidflanagan: This file is already doing the very same thing above at line 31, so it doesn't seem fair to ask John to avoid it here.  I'm not sure what jQuery is doing here. We might have better CSP luck just injecting a script tag. That might be a different CSP violation, I suppose...

kuma/static/js/wiki-compat-trigger.js
```diff
@@ -32,6 +32,20 @@
                 cache: true
             }).then(function() {
                 $('.bc-table').mozCompatTable();
+
+                if (waffle.flag_is_active('bc-signals')) {
```
davidflanagan: nit: I'd use `window.waffle` here, since waffle isn't actually declared as a variable.

Also, and more importantly, this code will fail on our React-based pages because waffle flags are no longer emitted into the HTML and are instead fetched via an API after the page loads.

kuma/static/js/wiki-compat-trigger.js
davidflanagan: Most pages have compat tables now, so this entire wiki-compat-trigger file isn't buying us much. It might be easier to just load the required CSS and JS unconditionally, which would avoid the CSP and $.ajax issue entirely.  For now, I guess I'd just leave it as is.

kuma/wiki/models.py
```diff
@@ -2012,3 +2012,28 @@ class DocumentSpamAttempt(SpamAttempt):
...
+    """Model to keep track of the BC signals."""
+    document = models.ForeignKey(
+        Document,
+        related_name='bc_signals',
+        null=True,
+        blank=True,
+        verbose_name=_('Document (optional)'),
+        on_delete=models.SET_NULL
+    )
+    user = models.ForeignKey(
```
davidflanagan: Have we thought carefully about whether we want to associate these signals with users? If we never collect the user id in the first place then we'll never have to worry about scrubbing user ids from the signal data before sharing it.

Given that most signals will come from non-logged in users, is there any real point in collecting the user data?

kuma/wiki/views/document.py
```diff
@@ -1005,3 +1006,28 @@ def _document_api_PUT(request, document_slug, document_locale):
...
+    if request.is_ajax():
+
+        user = request.user if request.user.is_authenticated else None
+
+        BCSignal.objects.create(
+            document=doc,
+            user=user
+        )
+
+        return HttpResponse()
```
davidflanagan: I assume the response is empty. Perhaps a 204 response would be most appropriate?

schalkneethling:
> * @schalkneethling: since this PR adds UX to the BCD tables, and the BCD tables are not handled in React, I think you'll find it difficult to make this one thing into a react component. So it seems to me that it is the code in wiki-compat-signal.js that will need to be turned into an effect hook in article.jsx which is also where we run the code that fixes up section headings.

Aha! Thanks for clearing that up @davidflanagan, that helps a lot.

peterbe:
@escattone Perhaps off topic but why is the API only available on the beta site? It doesn't feel like it'd be detrimental having it on the non-beta site. Right?

kuma/wiki/models.py
```diff
@@ -2012,3 +2012,28 @@ class DocumentSpamAttempt(SpamAttempt):
...
+        settings.AUTH_USER_MODEL,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL
+    )
+    submitted_at = models.DateTimeField(auto_now_add=True)
+
+    def __unicode__(self):
+        return u'BC Signal: %s' % (self.document.slug)
```
peterbe: I think it works with `__str__` just as well. One less thing to avoid having to rewrite when we upgrade to Python 3. 
Also, can you format the string with `.format` instead because I see that as a much more common tool.

peterbe:
```suggestion
        return u'BC Signal: {}'.format(self.document.slug)
```

kuma/wiki/views/document.py
```diff
@@ -1005,3 +1006,28 @@ def _document_api_PUT(request, document_slug, document_locale):
...
+    """Handle BC signals."""
+
+    doc, _ = _get_doc_and_fallback_reason(
+        document_locale, document_slug
+    )
+
+    if doc is None:
+        raise Http404
+
+    if request.is_ajax():
```
peterbe: Can this just be `if request.method == 'POST'` instead?

johngian:
I updated the PR after most of the comments. For now I changed the following:

* Use DRF for the signal API endpoint
* Add URL entry in both beta and default URLs
* Use JS fetch instead of jQuery
* Moved `getCookie` needed for Django CSRF token to `mdn.utils`
* Various nitpicks
* Various CSS fixes
* Removed user entry for model

TODO:
* Fix cursor on hover
* Replace jquery animations with CSS animations.

I am not familiar with the new React work you are doing so I am not sure how to address some comments regarding that. My understanding is that for now we are going forward with this iteration and the react stuff are going to be rewritten.

Thoughts ? 

Selection_216
Selection_217

escattone:
@johngian Sorry for the delay in getting to this! So many things going on at the same time! I hope to get to this within a day or two.

johngian:
Also here is the work to implement the same feature in react for beta after @davidflanagan suggestions: https://github.com/johngian/kuma/commit/c9a0d3cfd2c663fe77d95ab47fd39cb4e06f80ca

escattone:
@johngian I forgot to comment on the `bcd-signals` waffle flag and CloudFront. I suspect that waffle flag will be either ON or OFF for everyone at the same time, and if that's the case, I won't have to make any adjustments to our CloudFront configuration. We can simply turn ON the flag when we're all ready to release this. If that's not the case, the adjustment should be simple to do, so no worries there either.

Also, your changes to add this work into the React side of things for the `beta` domain look good as well (assuming the issue with getting the locale is fixed the same way there as well). Thanks for doing that. Once this lands, I'll fast-track your PR for that.

johngian:
PR updated

PR #5470, opened 2019-06-03
Bug 1556024 and Bug 1556469: TOC fixes

davidflanagan:
This PR tweaks the CSS for the "On this Page" TOC box in the sidebar
and changes the JS code so the box is not displayed if there is
no TOC for the page. This fixes bugs 1556024 and 1556469.

### DIFF
--- a/kuma/javascript/src/__snapshots__/page.test.js.snap
+++ b/kuma/javascript/src/__snapshots__/page.test.js.snap
@@ -1002,19 +1002,19 @@ Array [
 .emotion-6 {
   background-color: #fcfcfc;
   border: solid 1px #dce3e5;
-  padding: 8px 13px;
+  padding: 8px 8px 0px 13px;
   margin: 0 12px 20px -1px;
 }
 
 .emotion-6 ul {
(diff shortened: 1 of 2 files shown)


davidflanagan:
Screen Shot 2019-06-03 at 3 01 19 PM

davidflanagan:
Here's what it looks like when one of the titles wraps @nickbrandt:

Screen Shot 2019-06-03 at 5 01 34 PM

schalkneethling:
Waiting for feedback from @nickbrandt before merging.

nickbrandt:
@davidflanagan this is looking better, but I think we still need a bit more space between line items and potentially tighter line height to have clear separation (something like what is on the right of the attached). It's still a bit hard to differentiate on first glance.

Screen Shot 2019-06-04 at 8 37 11 AM

davidflanagan:
Adjusted line height from 21px to 20px and doubled the spacing between the list items to match Nick's image:

Screen Shot 2019-06-04 at 10 00 29 AM

PR #5384, opened 2019-05-03
bug 1548380 - Invalidate CloudFront on publish

peterbe:
Here's how I tested this locally...

I have my dev environment hooked up to my personal AWS S3 account. And the S3 bucket I've set up has a CloudFront distribution in front of it. Any edits to the wiki in my MySQL appears here, as an example: `curl -v https://d1jm0vr30jmdki.cloudfront.net/api/v1/doc/en-US/Learn/CSS`

So I wrote this script:

```python
import time
import datetime
import requests


def run(url):
    for _ in range(1000):
        r = requests.get(url)
        r.raise_for_status()
        print(datetime.datetime.now(), repr(r.json()['title']), r.headers.get('x-cache'), r.headers.get('cache-control'))
        time.sleep(3)

if __name__ == '__main__':
    import sys
    args = sys.argv[1:]
    url = args[0]
    run(url)
```

Started that in a terminal and went into my browser on `http://mdn.localhost:8000/en-US/docs/Web/CSS$edit` 
There, I edited the title and pressed "Publish". Then I went back to the terminal that keeps hitting the CDN and waited till its output changed. 

Hint: it did. and it was fast. :)

PR #5472, opened 2019-06-04
Invalidate cloudfront on publish/render

escattone:
This PR replaces https://github.com/mozilla/kuma/pull/5384 by building on top of it, adding the ability to skip the CDN cache invalidation when doing bulk document renders. Document renders don't have direct access to the `publish` command, only indirect access via the `render_done` signal.  So at its root, this PR changes the `render_done` signal such that it now provides/carries the additional information of whether or not to invalidate the CDN cache. It was important  to find a way to do that because we often run bulk re-renders (for example, all documents that use the `compat` macro).

I think this is complete (including tests), but I do still need to test this with the actual AWS keys, and I need to add the CloudFront distribution ID's for stage/production to the respective Kubernetes secrets as well as add the `MDN_CLOUDFRONT_API_DISTRIBUTIONID` to the stage/prod/standby environment configurations in mdn/infra (https://github.com/mdn/infra/tree/master/apps/mdn/mdn-aws/k8s/regions).

### DIFF
--- a/kuma/api/management/commands/publish.py
+++ b/kuma/api/management/commands/publish.py
@@ -39,17 +39,14 @@ class Command(BaseCommand):
             help='Partition the work into tasks, with each task handling this '
                  'many documents (default=1000)')
         parser.add_argument(
-            '--skip-cache-invaldation',
+            '--skip-cdn-invalidation',
             help=(
-                'No cache invalidation after publishing. By default True '
-                'if --all flag is used.'
+                'No CDN cache invalidation after publishing. Forced to True '
+                'if either the --all or --locale flag is used.'
             ),
             action='store_true')
 
     def handle(self, *args, **options):
-        skip_cache_invalidation = (
-            options['all'] or options['skip_cache_invalidation']
-        )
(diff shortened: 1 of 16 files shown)


kuma/api/management/commands/publish.py
```diff
@@ -39,17 +39,14 @@ def add_arguments(self, parser):
             help='Partition the work into tasks, with each task handling this '
                  'many documents (default=1000)')
         parser.add_argument(
-            '--skip-cache-invaldation',
+            '--skip-cdn-invalidation',
```
peterbe: 👍

escattone:
I tested this using the real stage AWS access keys. I had to add another policy in order to add permissions for CloudFront invalidation, but after that everything worked great. I also deployed a new secret for the CloudFront distribution ID to stage, prod, and standby, which is used by https://github.com/mdn/infra/pull/260 (which adds `MDN_API_CLOUDFRONT_DISTRIBUTIONID` to the environment of the `web` and `worker` Kubernetes deployments, among others).

PR #5475, opened 2019-06-05
Celery 4

peterbe:
Fixes bug 1228171
Fixes bug 1268281
Fixes bug 1268277
Fixes bug 1268254
Fixes bug 1268256
Fixes bug 1551701

Main difference:

* All the package updates to requirements. E.g. Celery 3 to 4.
* Note the deletion of `django-celery` which means that nothing is written to the MySQL database
* All periodic tasks are moved from MySQL into code. I manually "copied" them from https://developer.mozilla.org/admin/djcelery/periodictask/
* In local dev, the beat scheduler is still run within Celery on the same Python worker process. In prod we might need to change something because it's no longer `./manage.py celery --beat ...` but instead `celery -A kuma.celery:app worker --beat ...`

Other things to note:

* Celery 4 comes with a script that can rewrite your settings file to use the new names. I do worry that it might miss something but it's hard to track. 
* I personally still don't understand the kuma concept of listing queues. I've never used it before and I still don't really understand how it all fits into this upgrade or if it doesn't matter at all. Also, not sure I care as long as using the app works. 
* I *think* we get the true Redis config URL from an environment variable called `BROKER_URL` but in `django.conf.settings` we call it `CELERY_BROKER_URL`. That one might be fine but there's a chance that other such changes will break in production. Perhaps a bad example but if we used to have `CELERY_ALWAYS_EAGER=true` in the kubernetes environment variables then 
```diff
-CELERY_ALWAYS_EAGER = config('CELERY_ALWAYS_EAGER', False, cast=bool)
+CELERY_TASK_ALWAYS_EAGER = config('CELERY_TASK_ALWAYS_EAGER', False, cast=bool)
``` 
...will now "break" things. I think we just need to be vigilant and see what breaks in Stage and let it brew in stage for a while to make sure we test everything there properly. 
* ~~Note that in `ci-python` it now runs `docker-compose exec web celery -A kuma.celery:app inspect ping` which means that if the `worker` docker-compose thing is dead this script will fail.~~

### DIFF
--- a/kuma/humans/apps.py
+++ b/kuma/humans/apps.py
@@ -1,0 +1,26 @@
+from __future__ import unicode_literals
+
+from celery.schedules import crontab
+from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
+
+from kuma.celery import app
+
+
+class HumansConfig(AppConfig):
+    """
+    The Django App Config class to store information about the feeder app
+    and do startup time things.
+    """
+    name = 'kuma.humans'
+    verbose_name = _('Humans')
+
+    def ready(self):
+        """Configure kuma.humans after models are loaded."""
+
+        # Refresh humans.txt every day at 00:00
+        from kuma.humans.tasks import humans_txt
+        app.add_periodic_task(
+            crontab(minute=0, hour=0),
+            humans_txt.s()
+        )
--- a/kuma/humans/apps.py
+++ b/kuma/humans/apps.py
@@ -9,7 +9,7 @@ from kuma.celery import app
 
 class HumansConfig(AppConfig):
     """
-    The Django App Config class to store information about the feeder app
+    The Django App Config class to store information about the humans app
     and do startup time things.
     """
     name = 'kuma.humans'
--- a/kuma/wiki/tests/test_jobs.py
+++ b/kuma/wiki/tests/test_jobs.py
@@ -57,6 +57,3 @@ def test_contributors(db, cleared_cacheback_cache, settings, wiki_user_3,
     contributors = job.get(root_doc.pk)
     got = set(c['id'] for c in contributors)
     assert got == set([wiki_user_3.pk, banned_user.pk] + valid_contrib_ids)
-
-    assert ([c['id'] for c in job.get(root_doc.pk)] ==
-            ([wiki_user_3.pk, banned_user.pk] + valid_contrib_ids))
--- a/kuma/wiki/tests/test_jobs.py
+++ b/kuma/wiki/tests/test_jobs.py
@@ -13,6 +13,7 @@ def test_contributors(db, cleared_cacheback_cache, settings, wiki_user_3,
     Tests basic operation, ordering, caching, and handling of banned and
     inactive contributors.
     """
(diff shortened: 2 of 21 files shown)


peterbe:
I just removed the `docker-compose exec web celery -A kuma.celery:app inspect ping` from the `ci-python` script. It just doesn't work. Let's worry about making sure Celery starts in another step later. I.e. in a follow-up PR or something.

peterbe:
Actually @escattone this one would be nice to escalate because it'd be painful to unrot if it rots.

kuma/wiki/tests/test_jobs.py
```diff
@@ -52,5 +54,6 @@ def test_contributors(db, cleared_cacheback_cache, settings, wiki_user_3,
 
     # The new contributor shows up and is first, followed
     # by the freshly un-banned user, and then the rest.
+    contributors = job.get(root_doc.pk)
```
escattone: Just noticed that this is unused. I suspect you forgot to use it on the line below in place of `job.get(root_doc.pk)`?

peterbe:
Good catch about the `humans_txt` task. Added now.

kuma/wiki/tests/test_jobs.py
```diff
@@ -55,5 +55,8 @@ def test_contributors(db, cleared_cacheback_cache, settings, wiki_user_3,
     # The new contributor shows up and is first, followed
     # by the freshly un-banned user, and then the rest.
     contributors = job.get(root_doc.pk)
+    got = set(c['id'] for c in contributors)
+    assert wiki_user_3.pk in got
```
escattone: The check for `wiki_user_3.pk` is already covered by the `assert` below, which I was thinking would change to something like:
```
assert got == set([wiki_user_3.pk, banned_user.pk] + valid_contrib_ids)
```

kuma/humans/apps.py
```diff
@@ -0,0 +1,26 @@
...
+from celery.schedules import crontab
+from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
+
+from kuma.celery import app
+
+
+class HumansConfig(AppConfig):
+    """
+    The Django App Config class to store information about the feeder app
```
escattone: Copy leftovers: `feeder app` --> `humans app` (or something like that)

peterbe:
@escattone nits addressed. Let's see what Travis has to say.

kuma/wiki/tests/test_jobs.py
```diff
@@ -56,7 +56,7 @@ def test_contributors(db, cleared_cacheback_cache, settings, wiki_user_3,
     # by the freshly un-banned user, and then the rest.
     contributors = job.get(root_doc.pk)
     got = set(c['id'] for c in contributors)
-    assert wiki_user_3.pk in got
+    assert got == set([wiki_user_3.pk, banned_user.pk] + valid_contrib_ids)
```
escattone: Now that line 59 is what it is, lines 61-62 can be removed?

peterbe:
@escattone hopefully got it right this time.

peterbe:
🎉 🎊 🚢

PR #5479, opened 2019-06-05
bug 1518653: upgrade djangorestframework

escattone:
This is the last `djangorestframework` release that'll support Python 2 (https://github.com/encode/django-rest-framework/blob/master/docs/community/release-notes.md#39x-series).


peterbe:
I don't even know where we use djangorestframework :) But I'll try to figure that out tomorrow when I manually test it.

escattone:
@peterbe It's used only in the search app.

PR #5481, opened 2019-06-07
bug 1552159: fix code sample when no rendered HTML

escattone:
This fixes the bug described in https://bugzilla.mozilla.org/show_bug.cgi?id=1552159. The conditions for the bug are the following:
- `ENABLE_RESTRICTIONS_BY_HOST` is True (this is always true for stage and production, but false by default for local development)
- a request is made for a document's code sample using one of the attachment domains -- for production, these are either `mdn.mozillademos.org` or `demos-origin.mdn.mozit.cloud` 
- the document does not yet have any rendered HTML (its `rendered_html` attribute is falsy)

Under these conditions, an attempt is made to render the document, but fails when `document.get_absolute_url()` is called within `kuma.wiki.kumascript.get`, since the `wiki.document` endpoint is not defined for the attachments domains (for those domains, the root urlconf is set to `kuma.urls_untrusted`, which does not include that endpoint).

The fix is to explicitly specify the `urlconf` as `kuma.urls` when calling `reverse` within the `get_absolute_url` method.

I also modified an existing test so that it exposes the bug, preventing regressions.

### DIFF
--- a/kuma/wiki/tests/test_views_code.py
+++ b/kuma/wiki/tests/test_views_code.py
@@ -87,18 +87,19 @@ def test_code_sample_host_allowed(code_sample_doc, settings, client):
 
 # The pytest-django urls marker also resets urlconf caches after the test
 @pytest.mark.urls(settings.ROOT_URLCONF)
-def test_code_sample_host_restricted_host(code_sample_doc, settings, client):
+def test_code_sample_host_restricted_host(code_sample_doc, constance_config,
+                                          settings, client):
     """Users are allowed to view samples on the attachment domain."""
-    # Render the document, which will be a sync operation, before applying
-    # host restrictions.
-    code_sample_doc.get_rendered()
-
     url = reverse('wiki.code_sample',
                   args=[code_sample_doc.slug, 'sample1'])
     host = 'sampleserver'
     settings.ALLOWED_HOSTS.append(host)
     settings.ATTACHMENT_HOST = host
     settings.ENABLE_RESTRICTIONS_BY_HOST = True
+    # Setting the KUMASCRIPT_TIMEOUT to a non-zero value forces kumascript
+    # rendering so we ensure that path is tested for these requests that use
+    # a restricted urlconf environment.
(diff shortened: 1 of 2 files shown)


kuma/wiki/tests/test_views_code.py
```diff
@@ -87,18 +87,16 @@ def test_code_sample_host_allowed(code_sample_doc, settings, client):
...
-    # host restrictions.
-    code_sample_doc.get_rendered()
-
     url = reverse('wiki.code_sample',
                   args=[code_sample_doc.slug, 'sample1'])
     host = 'sampleserver'
     settings.ALLOWED_HOSTS.append(host)
     settings.ATTACHMENT_HOST = host
     settings.ENABLE_RESTRICTIONS_BY_HOST = True
+    constance_config.KUMASCRIPT_TIMEOUT = 1
```
peterbe: Why this change? Is that to deliberately cause a KS timeout?

escattone: Without this change, kumascript rendering is skipped (see https://github.com/mozilla/kuma/blob/aaabab98c62ebde4b4088055967568b4d9971b86/kuma/wiki/models.py#L572), so the call to `kuma.wiki.kumascript.get`, and eventually `document.get_absolute_url()`, is never made, and the problem is never exposed.

escattone:
Thanks for the review @peterbe.

> This trick of switching urlconf, I thought that was something exclusive to the beta vs non-beta. Is it used for attachments too (unrelated to beta)?

Yes, it's used for the attachments as well.

kuma/wiki/tests/test_views_code.py
peterbe: It's so odd looking so could you please explain what you just explained to me in a code comment? Or do you think that's dumb because it's something everyone knows?

escattone: I agree, and I've already added a comment, as you suggested, thanks!

PR #5483, opened 2019-06-07
Bug 1525719: make signin and signout work on the beta domain

davidflanagan:
Our login and logout endpoints require the ?next= redirect URL to
have the same origin as the endpoint. We were trying to use endpoints
on the wiki domain so that we didn't have to support them on the beta
domain. But that meant that when a user logged in or out, they were
taken away from the beta domain.

This PR fixes that by adding the auth endpoints (and also profile
viewing and editing) endpoints to the beta domain. It makes
corresponding frontend changes in the login.jsx component and
also tweaks the Jinja template for profile view so that we don't
show document editing activity on the beta domain--doing that
requires URLs that are not supported on the beta domain and were
causing 500 server errors due to reverse lookup failures.

### DIFF
--- a/kuma/urls_beta.py
+++ b/kuma/urls_beta.py
@@ -1,5 +1,7 @@
+from decorator_include import decorator_include
 from django.conf import settings
 from django.conf.urls import include, url
+from django.views.decorators.cache import never_cache
 from django.views.generic import RedirectView
 from django.views.static import serve
 
@@ -7,6 +9,7 @@ from kuma.core import views as core_views
 from kuma.core.decorators import beta_shared_cache_control, shared_cache_control
 from kuma.core.urlresolvers import i18n_patterns
 from kuma.landing.urls_beta import lang_urlpatterns as landing_lang_urlpatterns
+from kuma.users.urls_beta import lang_urlpatterns as users_lang_urlpatterns
 from kuma.views import serve_from_media_root
 
 
@@ -49,6 +52,13 @@ urlpatterns += [
     url(r'^(?P<path>@api/deki/files/.+)$', redirect_to_attachments_domain,
         name='attachments.mindtouch_file_redirect'),
 ]
+
+# Add the signin and signout urls
+urlpatterns += [url('users/', include('kuma.users.urls_beta'))]
+urlpatterns += i18n_patterns(url('',
+                                 decorator_include(never_cache,
+                                                   users_lang_urlpatterns)))
+
 
 if getattr(settings, 'DEBUG_TOOLBAR_INSTALLED', False):
     import debug_toolbar
--- a/kuma/users/urls_beta.py
+++ b/kuma/users/urls_beta.py
@@ -19,9 +19,6 @@ for provider in providers.registry.get_list():
         urlpatterns += prov_urlpatterns
 
 users_patterns = [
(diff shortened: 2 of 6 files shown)


davidflanagan:
Ryan: if you ignore the snapshot test snapshot changes (sorry about those!) this is a simple PR that should be easy to review. Are you okay with the way I just pulled in all of the user-related URLs into urls_beta.py? 

This PR finally motivated me to set up github auth for my dev server, and it all works like a charm now.

kuma/urls_beta.py
```diff
@@ -50,6 +53,16 @@
...
 ]
 
+# Add the signin and signout urls, and profile-related urls
+# TODO: These lines also pull in a number of endpoints that we don't
+# need on the beta site, so we should prune these down when the beta
+# site becomes the default.
+urlpatterns += [url('users/', include('kuma.users.urls'))]
+urlpatterns += i18n_patterns(url('',
+                                 decorator_include(never_cache,
+                                                   users_lang_urlpatterns)))
```
escattone: @davidflanagan As you say in your comment, I'd like to prune these down to what we actually need for handling logins/logouts. I'll look more into that next week.

escattone: After a quick look, I think we could prune the auth-key endpoints, the account recover endpoints, and the profile endpoints perhaps as well?

davidflanagan: The profile endpoints were a nice bonus, but they're not necessary and we could cut them.

As a practical matter, I find django's urlpatterns/include() system of registering routes impenetrable, and find myself incapable of doing anything other than copy and paste. And also kuma.users.urls seems like a particularly complex one, so I think I could use some pointers about how to enable just a subset of the users urls...

peterbe: And so much of it is changing in Django 2. Sigh. 

Would it help to revisit this a bit "deeper". Perhaps I've never understood it fully but currently we use a piece of middleware to dynamically switch the urlconf stuff around, in runtime, based on the domain used. Perhaps, if urlconf is clunky, we revisit that approach and instead of having "multiple urlconf" we have other tricks to swap between wiki and beta. Not sure what but perhaps we can reevaluate now that the beta project has come a longer way and is, at least mentally, more mature.

davidflanagan: So I solved my problem by just creating another urls_beta.py file in kuma/users/ and modifying this file to use that one instead of the original non-beta urls.

kuma/users/urls_beta.py
```diff
@@ -0,0 +1,35 @@
...
+    try:
+        prov_mod = importlib.import_module(provider.package + '.urls')
+    except ImportError:
+        continue
+    prov_urlpatterns = getattr(prov_mod, 'urlpatterns', None)
+    if prov_urlpatterns:
+        urlpatterns += prov_urlpatterns
+
+users_patterns = [
+    url(r'^signup/?$',
```
escattone: I don't think this one is used any longer? I'll look into it some more, but just wanted to document my thought for now. https://developer.mozilla.org/en-US/users/signup returns a page saying it has been disabled, but I don't know how one even gets to that page.

davidflanagan: Good point. I've removed it @escattone.

PR #5394, opened 2019-05-09
bug 1548669 - user timezone migration and ci

peterbe:
As expected, the first commit just added the `makemigrations --check --dry-run` in `ci-python` and it [broke the PR](https://travis-ci.org/mozilla/kuma/jobs/530428813).

PR #5486, opened 2019-06-10
user timezone migration

peterbe:
This is the second time we've had to make a migration like this year. [Previous one](https://github.com/mozilla/kuma/pull/5394). Unfortunately, there is no easy way out that isn't hackish. 
A much better option would be to NOT tie the option to the model. Instead it would be just done entirely different by making the model not tied to it and instead supply the options via a form class (e.g. admin model form). 

Without this change now it's breaking all other PRs. [For example](https://travis-ci.org/mozilla/kuma/jobs/543646160) so a quick merge would be appreciated.


peterbe:
Hmm... The choices come from `from sundial.zones import COMMON_GROUPED_CHOICES`
(in `kuma/users/models.py`). But why would they have changed?!

peterbe:
Are the timezones not static?? Here's the diff between the outputs of `sundial.zones.COMMON_GROUPED_CHOICES` of a month ago and now:

```diff
--- Old
+++ New
@@ -56,7 +56,7 @@
       ],
       [
         "Africa/Casablanca",
-        "Casablanca (GMT+0000)"
+        "Casablanca (GMT+0100)"
       ],
       [
         "Africa/Ceuta",
@@ -84,7 +84,7 @@
       ],
       [
         "Africa/El_Aaiun",
-        "El Aaiun (GMT+0000)"
+        "El Aaiun (GMT+0100)"
       ],
       [
         "Africa/Freetown",

```

So what does it mean? Is `sundial.zones.COMMON_GROUPED_CHOICES` a dynamic property that changes upon invokation? If so, this chore will just reappear again and again throughout the year.

peterbe:
For what it's worth, Casablanca Morocco changed to daylight savings on June 9th. 
https://www.timeanddate.com/time/change/morocco/casablanca
Could it be that that affects the names of the timezones.

peterbe:
I'm so confused! Something odd is going on with Casablanca. https://www.timeanddate.com/news/time/morocco-changes-clocks-2019.html

I think it's time to once and for all get rid of this "not so static" list of choices on that model. Instead we should supply it dynamically where it matters.

peterbe:
Note-to-self;
```
>>> tz = pytz.timezone('Africa/Casablanca')
>>> for i in range(3):
...   print(tz.localize(datetime.now() - timedelta(days=i)).strftime('%z'))
...
+0100
+0100
+0000
```

HOWEVER, the GMT offset will change for all locations. **It's not unique to Marocco in Africa**. 

It makes sense. For example, on March 9 the America/New_York timezone changes its offset compared to GMT because the US changes its Daylight Savings on a different day from Europe where Greenwich is. 

Blah! Let's just get out of the business of trying to put a GMT offset name to the user profile timezones.

PR #5492, opened 2019-06-10
Don't allow test strings in the l10n test to actually be localized

davidflanagan:
kuma/javascript/src/l10n.test.js tests our frontend gettext and
ngettext functions, and the way it was written the test strings
were actually being extracted into the react.pot file, which would
just confuse translators.

This PR changes the l10n.test.js file to invoke those functions
through alias names so that the strings can't be scraped and should
not appear in the react.pot file and react.po files.

### DIFF
--- a/kuma/javascript/src/l10n.test.js
+++ b/kuma/javascript/src/l10n.test.js
@@ -2,6 +2,12 @@
 
 import { localize, getLocale, gettext, ngettext, interpolate } from './l10n.js';
 
+// We don't actually want the strings in this file to be extracted
+// for localization, so we're going to use non-standard aliases
+// for gettext and ngettext.
+const gtext = gettext;
+const ngtext = ngettext;
+
 describe('getLocale', () => {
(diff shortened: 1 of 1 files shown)


kuma/javascript/src/l10n.test.js
```diff
@@ -2,6 +2,12 @@
 
 import { localize, getLocale, gettext, ngettext, interpolate } from './l10n.js';
 
+// We don't actually want the strings in this file to be extracted
+// for localization, so we're going to use non-standard aliases
+// for gettext and ngettext.
+const gtext = gettext;
+const ngtext = ngettext;
```
peterbe: Couldn't you have done `import { gettext as original_gettext } from './l10n.js'` so that the rest of the code gets the familiar looking function name.

peterbe: Oh wait. Silly me. I suspect the Duente (or whatever it's called) isn't *that* smart.

PR #5493, opened 2019-06-10
Bug 1557013: don't use zero width spaces for hyphenation

davidflanagan:
Some of our documents have really long one-word titles, like:

  CanvasRenderingContext2D.globalCompositeOperation

We want MDN to be able to break long strings like these reasonably.
In the past, we find the lower-to-upper case transitions and
inserted HTML <wbr> tags at those points to tell the browser that
they could break the word there. That was good user-facing UX, but
in bug 1523662 we discovered that for some reason Google was
sometimes showing the literal string "<wbr/>" in its search
results for our site.

So to fix that, we switched from <wbr> to zero width spaces
at the internal word transitions. This resulted in the same word
wrapping behavior, and google has been happy with it, I think.
But it caused this bug: that copy and paste is broken, especially
when trying to paste the title of a page into a code editor.

So in this iteration, we instead just using the CSS 'hyphens:auto'
property to tell browsers that they can break the long word. They
don't always break the word where we would like them to at the
internal word boundaries, but do seem to at least get syllable
boundaries (in English anyway). And, the benefit of this approach
is that browsers insert a hypen, so that users can more easily
tell that the line has been broken.

So we're running less code and we're getting a result that is
arguable better by just relying on the web platform to to the
right thing (or at least a reasonable thing).

Note that this PR adds the hyphens:auto change for both the new
React-based frontend and the old non-React frontend

### DIFF
--- a/kuma/javascript/src/__snapshots__/page.test.js.snap
+++ b/kuma/javascript/src/__snapshots__/page.test.js.snap
@@ -648,6 +648,10 @@ Array [
   font-family: x-locale-heading-primary,zillaslab,"Palatino","Palatino Linotype",x-locale-heading-secondary,serif;
   font-size: 45px;
   font-weight: bold;
+  -webkit-hyphens: auto;
+  -moz-hyphens: auto;
+  -ms-hyphens: auto;
+  hyphens: auto;
 }
 
 @media (max-width:749px) {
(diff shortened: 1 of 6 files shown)


davidflanagan:
These screenshots are the beta and wiki domains with this patch applied:

Screen Shot 2019-06-10 at 3 01 56 PM

Screen Shot 2019-06-10 at 3 03 33 PM

davidflanagan:
And here's what the same two pages look like without this patch (and with zero width spaces on the wiki domain and nothing on the  beta  domain)

Screen Shot 2019-06-10 at 3 07 40 PM

Screen Shot 2019-06-10 at 3 07 14 PM

davidflanagan:
The screenshots above are all Firefox Nightly. But things look substantially the same in Chrome

PR #5496, opened 2019-06-11
Bug 1558308: Always pass full HTML strings in document data

davidflanagan:
Typically when a React page is server-side rendered, the data associated
with that page ends up in the HTML page twice: once as HTML and once
as a JSON object. For MDN document pages I implemented an optimization
where I avoided passing the long strings of HTML content in the JSON
blob.

But then I modified the Sidebar component so that it would render the
TOC only if document.tocHTML was defined and by making the render
dependent on this variable that I had not sent in the JSON blob I
caused Bug 1558308 where hydration failed and the sidebar ended up
missing (it worked for client-side navigation, but the SSR'ed version
of the page would not work correctly.)

I'm reluctant to remove this optimization, but it seems likely to cause
confusing bugs later on, so I'm commenting it out for now. This is
something we can revisit later.

I suspect this won't have a noticable impact on page load time but
will track the stats on speedcurve to see.

### DIFF
--- a/kuma/wiki/templatetags/ssr.py
+++ b/kuma/wiki/templatetags/ssr.py
@@ -132,12 +132,30 @@ def server_side_render(component_name, data):
         # component_name is "document", we're going to make a copy of the
         # data (because the original belongs to our caller) and delete those
         # strings from the copy.
-        if component_name == 'document':
-            data = data.copy()
-            data['documentData'] = data['documentData'].copy()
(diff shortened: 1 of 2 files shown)


davidflanagan:
I'm asking Ryan for the review, but also tagging @schalkneethling and @peterbe because the optimization that I'm removing here was cool and something I probably mentioned to you, so you should know that it is gone now. :-(

PR #5503, opened 2019-06-14
Port various JS modules to work with React and to remove jQuery.

davidflanagan:
This large PR creates React-compatible versions of a number of
jQuery-based modules, and removes jQuery entirely from our beta pages.

This patch resolves bugs 1559480, 1560161, 1560135 and 1561037.

1) Bug 1559480: make BCD tables appear and behave correctly on beta site.

BCD table formatting and behavior depends on code in
wiki-compat-trigger.js running on page load, so the tables don't work
correctly with client-side navigation. This PR fixes that.

It makes the BCD stylesheet part of the main bundle that is loaded
into every React-based page, so the styles are always there and don't
have to be loaded when we find a page with a BCD table.

It also defines a new bcd.js file that defines an activateBCDTables()
function. We call this function with useEffect() from the Article
component so that it runs every time we display a new article that
might have a new BCD table in it.

activateBCDTables() is essentially a port of the jQuery code in
static/js/wiki-compat-table.js

2) Bug 1560161: syntax highlighting

In addition to patching up the BCD tables each time we display
a new article, we also need to perform syntax highlighting on
code blocks for each new article, so this commit adds a prism.js
file with a highlightSyntax() function that we can call from the
useEffect() hook of the Article component.

In addition it fixes what seems to be a regression in our SCSS files
where line highlighting was no longer happening on the beta site
or on the wiki site.

3) Bug 1560135: Open buttons for live examples

This PR ports the code from wiki-samples.js to live-examples.js so
that we can use an effect hook to add "Open in CodePen" and "Open in
JSFiddle" buttons to any documents that have live examples in them.

4) Bug 1561037: Narrow layouts for interactive examples

This PR ports the interactive example mediaQuery/narrowLayout code in
kuma/static/js/interactive.js to
kuma/javascript/src/interactive-examples.js and uses effect hooks in
article.jsx to get the code to run at the right time. This is so that
on narrow screens the interactive examples use their narrow layout
instead of their default wide layout.

5) Removing jQuery

This commit removes the last jQuery dependencies for our React-based pages.
It modifies two static/js/*.js files to convert simple jQuery calls
to the now widely-implemented standard alternatives. With those two
files modified (on both the wiki and beta domains) we no longer need
to include jQuery into the javascript bundle on the beta domain.

6) Other cleanup

Now that BCD and Prism JS is handled differently for the beta site
we no longer need window.mdn.assets output in our HTML, so that has
been removed for beta sites.

And, finally now that so much of the existing JS code has been ported
to the React codebase, there are no more gettext() calls in the
remaining legacy code used on the beta site, and so we can stop
outputting the javascript.js localization file.

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -3,9 +3,13 @@ import * as React from 'react';
 import { useEffect, useRef } from 'react';
 import { css } from '@emotion/core';
 
+import { activateBCDTables } from './bcd.js';
+import { addLiveExampleButtons } from './live-examples.js';
 import ClockIcon from './icons/clock.svg';
 import ContributorsIcon from './icons/contributors.svg';
 import { gettext } from './l10n.js';
+import { highlightSyntax } from './prism.js';
+import * as InteractiveExamples from './interactive-examples.js';
 import TagsIcon from './icons/tags.svg';
 
 import type { DocumentData } from './document.jsx';
@@ -36,6 +40,17 @@ const styles = {
         '& a.sectionLink': {
             fontSize: 24,
             textDecoration: 'none'
+        },
+
+        // Styles for BCD tables, overriding the stylesheets
+        '& table.bc-table button.bc-history-link': {
+            ':focus': {
+                outline: '#83d0f2 solid 3px',
+                outlineOffset: -3
+            },
+            ':hover': {
+                backgroundColor: '#83d0f2 !important'
+            }
         }
     }),
     metadata: css({
@@ -103,14 +118,25 @@ function addAnchors(article) {
 
 export default function Article({ document }: DocumentProps) {
     const article = useRef(null);
+
+    // This is a one-time effect we need to call the first time an article
+    // is rendered, to ensure that interactive examples resize themselves
+    // if the browser width changes.
+    useEffect(InteractiveExamples.makeResponsive, []);
+
+    // Each time we display an article we need to patch it up
+    // in various ways.
     useEffect(() => {
-        if (article.current) {
-            highlightSections(article.current);
-        }
-    }, [document]);
-    useEffect(() => {
-        if (article.current) {
-            addAnchors(article.current);
+        let rootElement = article.current;
+        if (rootElement) {
+            InteractiveExamples.setLayout(rootElement);
+            // Keep addLiveExampleButtons() before addAnchors() so the
+            // example title doesn't end up with a link in it on codepen.
+            addLiveExampleButtons(rootElement);
+            highlightSections(rootElement);
+            addAnchors(rootElement);
+            highlightSyntax(rootElement);
+            activateBCDTables(rootElement);
         }
     }, [document]);
 
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -1,4 +1,15 @@
-(function(win, doc, $) {
+/**
+ * This file creates a global window.mdn.analytics object that defines
+ * various functions that send data to google analytics. Some of these
+ * functions are used on both the wiki domain and the new readonly domain.
+ *
+ * TODO: some of the functions here are unused on the readonly domain
+ * so we should probably have a different, smaller, version of this file
+ * for use on that domain. Ideally we should review the data we are sending
+ * and if we still care about it, we should integrate the code into the
+ * new React codebase.
+ */
+(function(win, doc) {
     'use strict';
 
     // Adding to globally available mdn object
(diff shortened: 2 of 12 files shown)


davidflanagan:
Schalk: this is mostly a port of wiki-compat-tables.js to bcd.js, removing the jQuery dependency.

I've made a couple of intentional changes, though. Instead of having a big X button when we show implementation notes in a BCD table, I've gone with a more subtle thing where the down arrow in the button turns into an up arrow, indicating that clicking again will hide the note.

Also, the old code included an explicit keyboard handler for space and return instead of just relying on the default behavior of a button that has focus. I've changed so that keyboard focus seems to work (tested in Chrome and Firefox) without an event handler. But this does mean that the Escape key is no longer supported to dismiss a note.

These seem like reasonable changes to me, but let me know if you disagree. I actually wish we could just use tooltips for these implementation notes.  But maybe we can do that if we rewrite the entire table in React.

davidflanagan:
@schalkneethling I've added a second commit to this PR that makes a similar change to enable Prism syntax highlighting on beta pages. I didn't do that as a separate PR because the two PRs would have had overlapping changes in settings/common.py, and it seemed easier to just tack the new feature on here.

davidflanagan:
@schalkneethling: I've added one more commit here to complete the removal of jQuery. Sorry to keep adding new commits to this PR, but since they keep depending on each other, I think they will be easier to review in one place rather than as three interdependent PRs.

@escattone: I know you already started looking over this PR, but wanted to let you and @peterbe know that when we land this, we will no longer be including jquery in the javascript bundle for our single page app! 🎉

kuma/static/js/analytics.js
```diff
@@ -1,4 +1,15 @@
-(function(win, doc, $) {
+/**
```
davidflanagan: The changes in this file are to remove jQuery. Note that this file is also used by the wiki site, so if I've screwed anything up, it will also cause problems there.

kuma/static/js/newsletter.js
```diff
@@ -5,23 +5,18 @@
 
     var newsletterForm = doc.getElementById('newsletterForm');
     var isHidden = false;
-    var isArticle = false;
+    var isArticle = window.location.pathname.indexOf('/docs/') !== -1;
```
davidflanagan: The changes in this file are mostly a straightforward port of jquery functions to modern standards. Note that these changes affect both the wiki and the beta sites, so any errors here will affect the live site.

This particular line changes the way we detect whether we're on an article page to be compatible with React-based article pages.

kuma/static/js/newsletter.js
```diff
@@ -5,23 +5,18 @@
...
 
     if(!newsletterForm) {
         return;
     } else {
         // check for local storage
-        if(mdn.features.localStorage) {
+        if(mdn.features.localStorage !== false) {
```
davidflanagan: On the beta site, mdn.features.localStorage is undefined, so there is a change here and below to look explicity for false. On the beta site we simply assume that localStorage is available.

kuma/static/styles/components/content.scss
```diff
@@ -227,9 +227,12 @@ We cannot put these styles in the base/elements/ files because they don't apply
...
     ====================================================================== */
 
-    /* don't allow empty paragraphs by CKEditor */
+    /*
+     * Don't allow empty paragraphs by CKEditor.
+     * But do allow empty divs created by the Prism syntax highlighter.
+     */
     p:empty,
-    div:empty {
+    div:not(.line-highlight):empty {
```
davidflanagan: This change makes line highlighting work again. See http://beta.mdn.localhost:8000/en-US/docs/Web/CSS/gap#Examples for an example.  Line 5 of the CSS example should be highlighted, but it isn't without this change.

davidflanagan:
I pushed another commit, to resolve https://bugzilla.mozilla.org/show_bug.cgi?id=1561037

This would probably be better as a separate PR, but it would cause merge conflicts with this one, and since this one isn't reviewed yet, it seems easier to just add the code here.

davidflanagan:
I pushed yet another commit, to resolve https://bugzilla.mozilla.org/show_bug.cgi?id=1561035

As with the previous commits, this would probably be better as a separate PR, but it would cause merge conflicts with this one, and since this one isn't reviewed yet, it seems easier to just add the code here.

I think I'm done now, though.

kuma/javascript/src/article.jsx
```diff
@@ -36,6 +40,17 @@ const styles = {
         '& a.sectionLink': {
             fontSize: 24,
             textDecoration: 'none'
+        },
+
+        // Styles for BCD tables, overriding the stylesheets
+        '& table.bc-table button.bc-history-link': {
+            ':focus': {
+                outline: '#83d0f2 solid 3px',
+                outlineOffset: -3
```
escattone: The outlining seems a bit off (I'm using the latest release of Firefox):
image

escattone: As you told me in Slack @davidflanagan , it looks fine in Chrome:
image

escattone: @davidflanagan and I discussed this in Slack, and we're going to leave this for a future adjustment.

escattone:
@davidflanagan Thanks for the fix for the `ssr` bug. It works fine now.

PR #5506, opened 2019-06-21
Bug 1560573: Don't send two pageview events on the first page load

davidflanagan:
There is code in router.jsx that sends GA events for client-side
navigation. But it was also running on the initial page load as well.
This PR fixes that.

Also, this PR fixes a related bug in perf.js that I discovered while
working on the initial bug. When I added the LUX.init() call in
perf.js, I broke the custom performance measures I was collecting
because apparently LUX.init() clears all existing marks. I've fixed
that bug by calling LUX.init() first before I mark the start time for
the client-side navigation.

### DIFF
--- a/kuma/javascript/src/perf.js
+++ b/kuma/javascript/src/perf.js
@@ -11,11 +11,8 @@ let navigating = false;
 export function navigateStart() {
     navigating = true;
     try {
-        performance.clearMarks(START);
-        performance.clearMeasures(FETCH);
-        performance.clearMeasures(RENDER);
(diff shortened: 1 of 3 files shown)


davidflanagan:
Schalk and Ryan: this fixes the analytics bug we noticed while talking through the Router component yesterday.

Additionally, it fixes another bug I found while working on this. It turns out I broke the analytics in perf.js when I added LUX functionality in Bug 1525719, so this PR fixes that to by changing when I call LUX.init().

PR #5508, opened 2019-06-21
Bug 1560191, header menus are not localized on beta site

schalkneethling:
This is the menu in `fr` after this change:

Screenshot 2019-06-21 at 15 35 49

@davidflanagan r?


schalkneethling:
> Thanks, Schalk. This looks good to me.
> 
> Note that the screenshot you posted is for the non-beta site. 

Ha! So indeed it was ;)

PR #5510, opened 2019-06-24
Bug 1525719: add highlighted excerpts to search results

davidflanagan:
Our current search results include highlighted excerpts of
matching documents to provide more context about where the
search terms appear. This PR adds a similar feature to the
beta search results page.

### DIFF
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -246,7 +246,7 @@ def search(request, locale):
                       ])))
 
     # Add excerpts with search results highlighted
-    search = search.highlight(*WikiDocumentType.excerpt_fields)
+    search = search.highlight('content')
     search = search.highlight_options(order='score',
                                       pre_tags=['<mark>'],
                                       post_tags=['</mark>'])
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -245,6 +245,12 @@ def search(request, locale):
                           Q('term', locale={'value': locale, 'boost': 8}),
                       ])))
 
+    # Add excerpts with search results highlighted
+    search = search.highlight(*WikiDocumentType.excerpt_fields)
+    search = search.highlight_options(order='score',
+                                      pre_tags=['<mark>'],
+                                      post_tags=['</mark>'])
+
     # Return as many as 40 matches, since we're not implementing pagination yet
     response = search[0:40].execute()
     return JsonResponse(response.to_dict())
--- a/kuma/javascript/src/search-results-page.jsx
+++ b/kuma/javascript/src/search-results-page.jsx
@@ -19,7 +19,7 @@ export type SearchResults = {
         summary: string,
         tags: Array<string>,
         score: number,
-        excerpt: string
+        excerpts: Array<string>
     }>,
     error: ?any
 };
@@ -40,11 +40,24 @@ const styles = {
     link: css({
         fontWeight: 'bold'
     }),
-    summary: css({}),
+    summary: css({
+        marginBottom: 8
+    }),
     excerpt: css({
-        padding: 8,
+        padding: '2px 24px',
         fontStyle: 'italic',
-        fontSize: 12
+        fontSize: 12,
+
+        '& mark': {
+            fontWeight: 'bold',
+            backgroundColor: 'inherit'
+        },
+        ':before': {
+            content: '"..."'
+        },
+        ':after': {
+            content: '"..."'
+        }
     }),
     url: css({
         fontSize: 12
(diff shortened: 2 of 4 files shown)


davidflanagan:
Screen Shot 2019-06-24 at 9 49 28 AM

@peterbe: I'm asking for your review on this Peter, since you're the one who requested it.

kuma/javascript/src/search-results-page.jsx
```diff
@@ -96,6 +103,12 @@ export default function SearchResultsPage({ locale, query, data }: Props) {
                                     <div css={styles.summary}>
                                         {hit.summary}
                                     </div>
+                                    <div
+                                        css={styles.excerpt}
+                                        dangerouslySetInnerHTML={{
+                                            __html: hit.excerpt
```
peterbe: Is `hit.except` always a truthy thing? Just wondering what would happen if it's `null`. Then you get an empty useless div.

peterbe:
It's just... uh... not pretty. And that's coming from someone aesthetically challenged. 

It's not clear to me what I'm seeing. Consider this screenshot:
Screen Shot 2019-06-24 at 3 09 34 PM

For one, why isn't the word "STUFF" highlighted in the title of the first search result hit? (I know the technical explanation but imagine a user asking these questions).  Equally, the word "stuff" isn't highligted in the document summary even though it's there. Also, if only the 3rd document was able to find a highlight does it mean that the word "stuff" was not in the other two? Then why are they shown first? 

The display of the excerpt lacks an indicator that it *is* excerpts. 
On my personal blog I also use ES techniques like this (e.g. highlighting) but it's the only thing displayed underneath the title. And the little ellipses before and after makes it look more like excerpts because indicates that a *portion* of the sentences are extracted:
Screen Shot 2019-06-24 at 3 16 28 PM

Perhaps a personal preference but the use of `<mark>`'s default visual appearance (with the bright yellow background) makes it look like the Find-in-page in my browser. In my Nightly it uses this bright "purple" but either way, a bright color background highlighting feels "reserved" for find-in-page. 
Screen Shot 2019-06-24 at 3 23 14 PM
I think we should use `<b>` as the markers. That's what Google uses so that's high affordance. 

I'm wondering if it would be better to highlight on the title, the summary, *and* the body and do something like this:
```jsx
if (result.title_highlighted) {
  <a href={result.url} dangerouslySetInnerHTML={{__html result.title_highlighted}}/>
} else {
  <a href={result.url}>{result.title}</a>
}
```
...if you catch my drift. And then do the same for summary.

kuma/javascript/src/search-results-page.jsx
```diff
@@ -171,13 +184,27 @@ export class SearchRoute extends Route<SearchRouteParams, SearchResults> {
...
+                    ) {
+                        throw new Error('Search API returned unexpected data');
+                    }
+                    return {
+                        results: results.hits.hits.map(hit => {
+                            let score = hit._score;
+                            let excerpt =
+                                hit.highlight &&
+                                hit.highlight.content &&
+                                hit.highlight.content[0];
```
peterbe: First of all, we do ask ES to highlight on `['summary', 'content']` but we're never displaying any highlights from `summary`. 

Somewhere in here I think we should do something like:
```
const lines = []
['summary', 'highlight'].forEach(ns => {
  (hit.highligth[ns] || []).forEach(match => {
    if (lines.length < 3) {
      lines.push(<p key={ns + match} dangerouslySetInnerHTML={{ __html: match}}/>);
    }
  });  
});
```
It's not right but the point is that we display highlights from summary (first) and fill in from content up to a max point.

kuma/api/v1/views.py
```diff
@@ -245,6 +245,12 @@ def search(request, locale):
                           Q('term', locale={'value': locale, 'boost': 8}),
                       ])))
 
+    # Add excerpts with search results highlighted
+    search = search.highlight(*WikiDocumentType.excerpt_fields)
```
peterbe: I suppose it's outside the scope to have "title" added to that. I guess because of how it's indexed in ES.

kuma/api/v1/views.py
```diff
@@ -245,6 +245,12 @@ def search(request, locale):
...
+    search = search.highlight_options(order='score',
+                                      pre_tags=['<mark>'],
+                                      post_tags=['</mark>'])
```
peterbe: As mentioned in the general comment, I think this should be `<b>` and `</b>`. But feel free to ignore if you disagree.

davidflanagan:
@peterbe: there is no design for this: I'm just trying stuff out and seeing if it seems useful. This might be a feature that we hold off landing until we get feedback from @nickbrandt 

To respond to your specific comments:

- mark tags vs b tags: I think with the wiki site does is use mark tags but style them in some other way.  That is probably the semantically best thing to do. I was going to do that, but I was charmed by the default styling of mark exactly because it was so reminiscent of search in the browser highlighting. Easy to try different styling however.

- Making the excerpts stand out. I assert that the page summary is more useful information than the excerpts, so I don't want to just display the excerpts like we do on the existing search page. I tried to make the excerpts seem distinct by reducing the font size and using italics. Adding dots at the beginning and end seems like a good idea too.

- As for why there are not highlights in the title and summary. I'm not sure that's a good idea, but also I don't know how to make ElasticSearch do that. It just gives me an array of excerpts and I'm displaying the first one. I think it is possible for that excerpt to be from the summary, which would mean we'd repeat the same text twice. I guess that would be a reason to ask for excerpts from content only. (Though that might also include the summary)

peterbe:
Seems we're on the same page! (except minor preferences on some things)
Perhaps we can just pause on this PR and leave it open. It seems unimportant compared to so many other things. 
I think the new site search already trumps the existing kuma site-search with its confusing filtering checkboxes.

davidflanagan:
@peterbe: I think I've addressed most of your comments in this new version. If you've got a chance, please take another look. If the code seems sound and if the search results seem more useful this way, we might as well land it for beta testing and see what people think.

Screen Shot 2019-06-25 at 10 06 17 AM

PR #5511, opened 2019-06-24
Bug 1525719: add a "Give Feedback" link to beta pages

davidflanagan:
This PR adds a prominent request for feedback to all beta pages
and links to https://bugzilla.mozilla.org/show_bug.cgi?id=1561020
as a central location for collecting bugs and suggestions about
the beta site.

### DIFF
--- a/kuma/javascript/src/header/header.jsx
+++ b/kuma/javascript/src/header/header.jsx
@@ -214,46 +214,44 @@ export default function Header(props: Props): React.Node {
                 <Login />
             </div>
 
-            {/* Remove this div when we're done asking for feedback */}
-            <div
+            {
+                // Display a "Beta Feedback" banner.
+                // When we're done with beta testing, remove this <a> tag
+                // by reverting https://github.com/mozilla/kuma/pull/5511
+            }
+            <a
+                href="https://bugzilla.mozilla.org/show_bug.cgi?id=1561020"
+                target="_blank"
+                rel="noopener noreferrer nofollow"
                 css={css({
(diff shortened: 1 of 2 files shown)


davidflanagan:
Screen Shot 2019-06-24 at 11 21 14 AM

The idea here is that the yellow triangle in the upper left will make it obvious how to give feedback on the beta site. Clicking there opens https://bugzilla.mozilla.org/show_bug.cgi?id=1561020 in a new tab, and we can use that as a tracking bug...

davidflanagan:
Kadir and Nick have both approved the screenshot above after looking at it in Slack, so if code review looks positive, I think we can merge this.

kuma/javascript/src/header/header.jsx
```diff
@@ -213,6 +213,47 @@ export default function Header(props: Props): React.Node {
             <div css={styles.login}>
                 <Login />
             </div>
+
+            {/* Remove this div when we're done asking for feedback */}
```
peterbe: Thinking ahead, a fear is that in a couple of months, when we need to remove this little flap, we start here where the React element code is and then have to work backwards to figure out what else to change. 
Can you please mention this URL `https://github.com/mozilla/kuma/pull/5511/files` in that comment since the instruction should effectively be to **revert this PR**.

davidflanagan:
I updated the "how and when to remove this" comment to reference the PR. Thanks for that suggestion @peterbe.  Also, I added the word "Beta" to the banner at Sheppy's suggestion.

PR #5513, opened 2019-06-24
Bug 1561054: display document tags on the beta site

davidflanagan:
This PR adds document tags to the article metadata that we display
on beta pages, in the hope that it will be useful for SEO

Somewhat relatedly, it modifies the contributor metadata so
that usernames are not linked to profile pages, because the beta
site does not yet support profile pages.


davidflanagan:
@escattone I've asked for your review on this one because it touches the backend document api to add tags to the returned JSON. Tags are a bit like contributors in that there is some kind of job associated with them. For contributors, you added an ensure_contributors argument to document_api_data(). I don't think that something like that is necessary for tags, but want to check that with you.

davidflanagan:
Here's what the tags and contributors look like with this PR:

Screen Shot 2019-06-24 at 2 46 06 PM

PR #5514, opened 2019-06-24
Bug 1561034: fix BCD table layout for small-desktop widths

davidflanagan:
The CSS for BCD table layout is complicated and depends on the
presence of an element with class wiki-left-present to indicate that
there is a sidbar. That element had been removed on the beta pages,
breaking BCD table layout for widths 1023-1199. This PR adds
the class back to the element that contains the sidebar and
article, fixing BCD table layout.


escattone:
Before:
image

After:
image

PR #5517, opened 2019-06-27
Allow cdn.speedcurve.com in our CSP so the LUX script is not blocked

davidflanagan:
The LUX script is blocked by CSP on developer.allizom.org. This PR
allows that domain in our CSP configuration.


davidflanagan:
Ryan:  this is a 1 line change that would be nice to review and push if you're doing a push Thursday.

I can see the lux.js script being blocked on staging in the developer console, which means that even though my LUX SPA changes are on staging, I still am not able to verify that they send data to speedcurve.  I think they won't be blocked when we go to production, but still it seems worth keeping our CSP configuration updated to include this new analytics tool.

I'm not positive that this is the right way to do it... and maybe this is more of an mdn-infra task, but I'm asking for your review because I expect that you'll know how to handle this and it seems like something that should be on your radar.

PR #5518, opened 2019-06-27
1467518 convert kuma to python 3

peterbe:
When this lands we can do a bunch of clean up of bugs and issues and sprint issues. 

When switching between this branch and other Python2 based PRs you need to always delete the file `celerybeat-schedule` in the root directory. 

Also, these are good scripts to go back and forth between versions of MySQL:
https://gist.github.com/peterbe/8b25a06e6a1d614a78e1f93dc1a9c9a3
https://gist.github.com/peterbe/d34541d95c4625239eac6fadb1bfd415


kuma/wiki/forms.py
```diff
@@ -764,11 +766,18 @@ def akismet_parameters(self):
                     based_on = self.cleaned_data.get('based_on')
                     assert based_on, 'Expected a new translation.'
                     document = based_on.document
+                    # print("INPUT:", repr(document))
+                    # print("HTML______")
+                    # print(document.html)
+                    # print("HTML______")
+                    # print(document.html)
```
callahad: Can we remove these now?

kuma/wiki/forms.py
```diff
@@ -764,11 +766,18 @@ def akismet_parameters(self):
...
                     self._akismet_data = AkismetNewTranslationData(
                         self.request, self.cleaned_data, document,
                         self.data.get('locale'))
 
         parameters = self._akismet_data.parameters.copy()
+        # print(parameters)
+        # raise Exception
```
callahad: ...and these?

kuma/wiki/tests/test_admin.py
```diff
@@ -212,15 +216,15 @@ def test_spam_submission_filled(self):
         )
         self.client.login(username='admin', password='testpass')
         response = self.client.get(url)
-        self.assertEqual(response.status_code, 200)
+        assert response.status_code == 200
```
callahad: Is there a reason to move away from `self.assertEqual` here and in the following unit tests?

I tend to prefer it because it gives more descriptive error messages (showing the bad value) when it fails:

```
  File "test.py", line 13, in test_assertEquals
    self.assertEquals(response.status_code, 200)
AssertionError: 404 != 200
```

versus

```
  File "test.py", line 10, in test_assert
AssertionError
```

```

callahad:
So... reviewing this is gnarly and I really wish each of the commits told more of an isolated story. It's very hard to separate the harmless `u"foo"` to `"foo"` changes from changes that are more significant.

For example, I'd much prefer one commit that was just `result of running 2to3`, another that's functional cleanup of that, one that fixes anything turned up by running `pylint --py3k`, another for fixes to warnings produced by running Python 2 with the `-3` flag, etc. 

There's also the matter of how we're handling the transition. The changes to the Dockerfile and Travis config suggest that we're planning to go from 2 to 3 in a single, discrete change rather than first converting to code that would work in both 2 _and_ 3 and running that for a while, as recommended [by the docs](https://docs.python.org/3/howto/pyporting.html), perhaps with the help of a tool like [futurize](http://python-future.org/automatic_conversion.html).

Are we confident in our approach?

peterbe:
> Are we confident in our approach?

Yes.
I base that one the fact that here we are with a PR that has good bones. Switching from Python 2 to 3 is actually easy if you don't need to support both. Unfortunately, what makes it gnarly is the various libs that need to be upgraded. 
Another thing that makes it gnarly is the various database migrations you need to add which are one-way streets. A lot of the above-mentioned excuses could be resolved if we had more time. 
I've seen many Python projects for libs that individually just opt for a "rip the band-aid right off" approach and their justification is to bump the major part of the semver because Python2 folks can pin to an older version. E.g. Django. 

My plan is to revive this project once we launch the React site. I'll start by fixing the merge conflicts to a point where the tests pass locally. Then I'll write the database migrations so that tests pass in TravisCI. I'll also address the existing feedback (such as those many print statements that are left in). And then, when it's ready I would love that we set aside the necessary time to review "quickly" so it doesn't rot again. 

Note, with the Django 2 upgrade we're going for the other approach which is to do many small piecemeal upgrades where we can as separate PRs.

tobinmori:
- merge to master, and test on staging

kuma/wiki/tests/test_admin.py
callahad: Oh, hey, pytest rewrites bare asserts to get the same kind of nice output messages. Sold.

callahad:
Completed in #6084

PR #5519, opened 2019-06-27
Make the DNA banner appear on React-based document pages

davidflanagan:
Previously the banner was only on the home page, this adds it to
document pages (but not search results pages) for the beta site.

### DIFF
--- a/kuma/javascript/src/document.jsx
+++ b/kuma/javascript/src/document.jsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { css } from '@emotion/core';
 
 import Article from './article.jsx';
+import Banners from './banners.jsx';
 import { gettext } from './l10n.js';
 import LanguageMenu from './header/language-menu.jsx';
 import Header from './header/header.jsx';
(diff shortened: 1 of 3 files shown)


davidflanagan:
Screen Shot 2019-06-27 at 9 41 06 AM

kuma/javascript/src/banners-config.js
```diff
@@ -0,0 +1,23 @@
+//@flow
+import { gettext } from './l10n.js';
+import type BannerProps from './banners.jsx';
+
+// The landing page will display the first of the banners listed here that:
```
schalkneethling:
```suggestion
// The page will display the first of the banners listed here that:
```

kuma/javascript/src/document.jsx
```diff
@@ -237,6 +239,7 @@ function DocumentPage({ document }: DocumentProps) {
     return (
         <>
             <Header document={document} />
+            <Banners banners={banners} />
             <TaskCompletionSurvey document={document} />
```
schalkneethling: Looking at this again now, I reckon it would be better to have these two at the end of the list of components. From an accessibility perspective, it would be better to have the `Titlebar` and `Breadcrumbs` in the document structure and tab order before the two banners.

PR #5520, opened 2019-06-27
Bug 1561975: fix menu localization for beta pages.

davidflanagan:
The header component defines menus in a static data structure. Originally
I had gettext() calls directly in that data structure. They were visible
to Puente, but the gettext() ran before string catalogs were available
so translations didn't happen even though they were available in the
catalog.

Then we switched to calling gettext() when rendering. That worked for
translations, but meant that the actual string literals were no longer
visible to Puente and weren't being picked up. So when the react.pot
file went live, we lost those string.

So then we moved the gettext() calls back to the static data structure
and are back to the original problem of calling gettext() before strings
are available.

This PR finally gets it right (I hope). The data structure that defines
the menus is created inside a useMemo() hook, so that it runs during
rendering (after strings are available). But because we are using
useMemo() we don't have to re-localize the strings every time we render
a header.

### DIFF
--- a/kuma/javascript/src/header/header.jsx
+++ b/kuma/javascript/src/header/header.jsx
@@ -184,9 +184,8 @@ export default function Header(props: Props): React.Node {
                     {
                         label: gettext('Report a content problem'),
                         external: true,
-                        url: `https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title=${encodeURIComponent(
-                            props.document ? props.document.slug : '/' + locale
-                        )}`
+                        url:
+                            'https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title={{PATH}}'
                     },
                     {
                         label: gettext('Report a bug'),
@@ -197,6 +196,12 @@ export default function Header(props: Props): React.Node {
             }
         ],
         [locale]
+    );
+
+    // One of the menu items has a URL that we need to substitute
+    // the current document path into. Compute that now.
+    let path = encodeURIComponent(
+        `/${locale}` + (props.document ? `/docs/${props.document.slug}` : '')
     );
 
     return (
@@ -220,7 +225,10 @@ export default function Header(props: Props): React.Node {
                                         <a
                                             target="_blank"
                                             rel="noopener noreferrer"
-                                            href={item.url}
+                                            href={item.url.replace(
+                                                '{{PATH}}',
+                                                path
+                                            )}
                                         >
                                             {item.label} &#x1f310;
                                         </a>
--- a/kuma/javascript/src/header/header.jsx
+++ b/kuma/javascript/src/header/header.jsx
@@ -1,5 +1,6 @@
 //@flow
 import * as React from 'react';
+import { useMemo } from 'react';
 import { css } from '@emotion/core';
 
 import { getLocale, gettext } from '../l10n.js';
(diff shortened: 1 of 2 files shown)


davidflanagan:
One of the reasons I feel good about this patch is that it gets rid of the awkward fixurl() function that we were using to patch up the menus at render time.  Now that we just (efficiently) define the data structure at render time we no longer need that hack.

peterbe:
I was thiiiis close to approving the PR but I found a bug. 

First, I was here and just about to click on "Hjälpmedel" ("Accessability"):
Screen Shot 2019-06-27 at 3 18 56 PM

The link (on mouse hover) was: `http://beta.mdn.localhost:8000/sv-SE/docs/Web/Accessibility`

But when I clicked on it, this happened:
Screen Shot 2019-06-27 at 3 17 48 PM

The problem isn't that I don't have a translated "Accessibility" page, the problem is that it "redirected" me to the English version but the URL bar didn't update.

peterbe:
Actually, I'm not sure what I mentioned is related to this PR. 
If I go straight to http://beta.mdn.localhost:8000/sv-SE/docs/Web/Accessibility (typing the URL) it simply shows me the en-US content (but the menu remains in Swedish which is a bit odd). 

It's a bit odd also on the SSR side of things:
```
▶ curl -v http://beta.mdn.localhost:8000/sv-SE/docs/Web/Accessibility > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying ::1...
* TCP_NODELAY set
* Connected to beta.mdn.localhost (::1) port 8000 (#0)
> GET /sv-SE/docs/Web/Accessibility HTTP/1.1
> Host: beta.mdn.localhost:8000
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: meinheld/0.6.1
< Date: Thu, 27 Jun 2019 19:24:19 GMT
< Content-Length: 87626
< x-xss-protection: 1; mode=block
< x-content-type-options: nosniff
< Content-Language: sv-SE
< Vary: Cookie
< ETag: "36a66cf51c7084aac871cd43bafc4f95"
< X-kuma-revision: 1458816
< Cache-Control: s-maxage=3600, public, max-age=0
< Access-Control-Allow-Origin: *
< Content-Type: text/html; charset=utf-8
< X-Frame-Options: DENY
< Connection: Keep-Alive
<
```
Why is that allowed to be `200 OK`? Shouldn't it be `302`?

peterbe:
Actually, independent of whether it should be 200 vs 302, there is definitely something wrong with the SSR headers. 
```
Content-Language: sv-SE
```
No it's not :)

kuma/javascript/src/header/header.jsx
```diff
@@ -1,5 +1,6 @@
 //@flow
 import * as React from 'react';
+import { useMemo } from 'react';
```
peterbe: Why not use `React.useMemo` on line 101? Is this a necessary hack because of Flow?

kuma/javascript/src/header/header.jsx
```diff
@@ -153,18 +92,112 @@ type Props = {
...
-        }
-        return url;
-    }
+    // The menus array includes objects that define the set of
+    // menus displayed by this header component. The data structure
+    // is defined here during rendering so that the gettext() calls
+    // happen after the string catalog is available. And we're using
+    // useMemo() so that localization only happens when the locale
+    // changes.
+    const menus = useMemo(
```
peterbe: Perhaps this is nitpickery, if so I appologies in advance. 
The `<Header/>` component is used on every page mount, right. So the `props.searchQuery` is potentially different and internally (is that called a side-effect) the `locale` is different. And obviously `props.document` can change too. 

I noticed that the `<Header/>` component renders twice on every page. I put in some `console.log` of `locale` and `props.searchQuery` and it looks like this on every page click:
Screen Shot 2019-06-27 at 3 41 01 PM

(PS. I'm still grappling with why the prop specifies `searchQuery?: string` but it's allowed to be `undefined`)

Anyway, we could just memoize the whole component since the only inputs are easy to compare. E.g.

```
// Totally made up
export const Header = React.memo((props: Props): React.Node => {
  // will only render once per props passed
})
```
That way, the whole component would be memoized. And not just one smaller part of it inside. 
Immediate effect is that the whole component would only render *once* per page. 

I don't think we could get away from rendering the component for *every different* page because the URL to github issues depends on the document.

kuma/javascript/src/header/header.jsx
```diff
@@ -153,18 +92,112 @@ type Props = {
...
+                    },
+                    {
+                        url: `/${locale}/docs/MDN/Community`,
+                        label: gettext('Join the MDN community')
+                    },
+                    {
+                        label: gettext('Report a content problem'),
+                        external: true,
+                        url: `https://github.com/mdn/sprints/issues/new?template=issue-template.md&projects=mdn/sprints/2&labels=user-report&title=${encodeURIComponent(
+                            props.document ? props.document.slug : '/' + locale
```
peterbe: This does not match with line 199. But perhaps I'm missing the point anyway. 
I.e. if you change the page, the `useMemo` is re-run anyway. So there's no memoization across pages. 

Either way, it's just odd that the input hint on line 199 talks about `locale` but the bulk of the code actually depends on `[props.document, locale]`. Right?

peterbe: There's definitely something wrong. That link isn't updating as I changed pages. It sticks to whatever page I was on the first SSR load. 

By the way, potentially a new/different but, but this is what I get when I clicked on the "Rapportera a problem with the content" (but in Swedish):
Screen Shot 2019-06-27 at 3 57 02 PM
It's lacking information about the fact that I was on the Swedish page.

escattone:
@peterbe Regarding your first set of comments about requesting the Swedish version of a page and receiving an English body wrapped with a Swedish header and footer, that behavior has been the standard behavior on MDN for some time (see https://github.com/mozilla/kuma/blob/a316c32d5f8dab6c073d00ec7f6835128b95e9ce/kuma/wiki/views/document.py#L814).

kuma/javascript/src/header/header.jsx
davidflanagan: You're right that useMemo() should also be based on props.document. I'll fix that.

kuma/javascript/src/header/header.jsx
davidflanagan: It is convention that hooks are used "bare" like this. There may also be tools that depend on it

kuma/javascript/src/header/header.jsx
davidflanagan: That might be a good idea. I'm not sure why we're getting two renders. But it is unrelated to this localization issue, so I'd rather not tackle it as part of this PR.

kuma/javascript/src/header/header.jsx
davidflanagan: You're also right that the bug report form ought to include the locale. It is unrelated but I'll see if I can fix that now.

PR #5523, opened 2019-06-28
Wraps the SPA and Landing pages in React.StrictMode

schalkneethling:
Now that we are at a place where we have a *lot* of code in the React codebase, and we are going to be iterating and building on this a lot, I thought it would make sense to introduce some automated safeguards.

`React.StrictMode` is essentially `use strict;` for React ;p 
https://reactjs.org/docs/strict-mode.html

@davidflanagan r?


davidflanagan:
This seems like a reasonable thing to do. From the description of strict mode, it sounds like it is mostly looking out for legacy issues which we won't have in our codebase since we started out using hooks exclusively (and because we haven't yet had to import third party components). But still probably worth doing.

My only question before landing is about when the warnings are active: the docs say it only affects development, not production, but how does it know? Our webpack config is set up to use production mode even in development (which is a separate issues we should probably fix). If that is how it determines production vs development then we won't ever see any warnings.

I'm all for landing it but we should understand how to enable and disable it before we land it. (Maybe you already understand that...)

schalkneethling:
> My only question before landing is about when the warnings are active: the docs say it only affects development, not production, but how does it know? Our webpack config is set up to use production mode even in development (which is a separate issues we should probably fix).

+1 - I was not aware that that is the case.

> 
> I'm all for landing it but we should understand how to enable and disable it before we land it. (Maybe you already understand that...)

I believe it would be based on NODE_ENV and therefore it will currently just do nothing. Let me dig into our Webpack implementation.

peterbe:
> My only question before landing is about when the warnings are active: the docs say it only affects development, not production, but how does it know?

webpack can take a flag (`--mode=development|production`) which permeates to become available as `process.env.NODE_ENV` inside the React code. At least that's how it works in one of my side-projects. I use `webpack --config webpack.config.server.js --watch --mode=development` which produces a `dist/server.js` (ES5) file in which I can do things like `{process.env.NODE_ENV==='development' ? <p>Debug version: {version}</p> : null}`.

peterbe:
I don't see anything in that list that would be applicable to our code base. It's mostly legacy stuff and things that don't applicable to use because we're exclusively using hooks.

davidflanagan:
Using StrictMode was something that Dan recommended as well, IIRC, but I think he had forward-looking reasons for recommending it, not backward looking. 

I lean towards landing this, but maybe do it in conjunction with a patch that allows us to use development mode in development (which was another strong recommendation from Dan) and production mode for production.

The only reason we default to production mode is that I was afraid of shipping development bundles (which are ~10x bigger) accidentally and never took the time to figure out how to make our build system do the right thing. I'll ocassionally turn on development mode by editing the webpack config manually when I need to see a stack trace with real symbols in it, but this is really far from ideal.

Hopefully there is a webpack command-line option that can set production mode in a way that overrides the webpack config? If so, then presumably just using that command line option in the Makefile would suffice.

peterbe:
For that, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1562282

peterbe:
I think the PR change too much after David's approval. I don't know a way to undo that. But this PR is not approved. I think it's on my to review it next.

schalkneethling:
> I think the PR change too much after David's approval. I don't know a way to undo that. But this PR is not approved. I think it's on my to review it next.

+1 - I also marked it as not ready, as I still have a couple of changes/additions to the Webpack piece. That will be done tomorrow.

schalkneethling:
Ok, so here is a breakdown of what this does/adds.

## Local development mode

### Run

```
npm run webpack:dev
```

This builds our web bundle and our SSR bundle and starts a watch. When you change something on the React side, Webpack will rebuild only what is changed and update the relevant bundle.

This also starts a watch on the file system via chokidar. It will then restart our SSR server once Webpack is done rebuilding and emitting assets to ensure our client and server are in sync. This avoids problems with especially Emotion class names.

This then allows a workflow such as the below:

devmode


## Bundlesize

This adds a target that will provide bundlesize statistics on our `react.js` bundle in the form of the following:

Screenshot 2019-07-19 at 14 17 11

You can see this report by running `npm run webpack:bundlesize`

NOTE: I have set a limit of 2s just so that this task does not currently fail. It is not set in stone, and should be revised.

## Bundle Analyzer

To get an idea of where our bundle size comes from, I added the Webpack bundle analyzer plugin. To run this, do the following:

```
npm run webpack:analyze
```

This will produce a report like the following:

Screenshot 2019-07-19 at 14 14 06

## Production

For a production build, run:

```
npm run webpack:prod
```

This creates the bundles as before i.e. react.js and ssr.js but, it will now also create a compressed version of both as react.js.gz as well as ssr.js.gz(this one is probably not needed). It also generates hidden-source-maps for both of the files and stores it inside the `/dist/` folder along with all the other previously mentioned assets.

This is quite a number of things ;) @peterbe if you could take this for a spin and let me know your thought, that would be great. Thanks!

schalkneethling:
@escattone Would appreciate your eyes on this one as well. Thanks!

schalkneethling:
> I don't know what but this PR introduces `puppeteer` as a sub-dependency.
> It's HUUUGE and I am worried about adding something that large in the docker image.
> 
> You can find it inside:
> https://github.com/mozilla/kuma/blob/e04a89e9c10c3080181558641cb9ef9c634ff964/package-lock.json
> 
> ```
> ~/kuma  (e04a89e9c...) ✗
> ▶ npm-why puppeteer
> 
>   Who required puppeteer:
> 
>   kuma > size-limit > estimo > @condenast/perf-timeline-cli > puppeteer@1.19.0
> ```
> 
> Can you please remove `size-limit`? It's out of scope for this PR anyway.

Sure thing, that is a little cray indeed. We should probably opt for [bundlesize](https://github.com/siddharthkp/bundlesize) instead.

schalkneethling:
@peterbe `size-limit` has been removed.

peterbe:
By the way, the most important thing is: IT WORKS!! 
I ran `npm run webpack:dev` and it correctly restarted ssr when I made edits. I messed with some `.jsx` files (e.g. remove the `key=` attribute) and the warnings show up accordingly. 
Nice! 

I'm really looking forward to seeing this land.

schalkneethling:
> I hate being the annoying one to say this but we really ought to document these things. Right now, all we have is [your PR comment](https://github.com/mozilla/kuma/pull/5523#issuecomment-513219545).
> If we don't tackle that now we might never get around to tackle it.

No, you are 100% correct. I was thinking about that. Let me write down some docs on this and update the PR.

schalkneethling:
> Also, can you remove the thing that generates the .gz files? I think it's a mistake to do that with webpack. We should only ever do that kinda stuff with CloudFront.

Sure thing.

peterbe:
By the way, that .gz stuff bit me. When I last reviewed this branch it created a `react.js.gz` (and the .map). Anyway, I go back to the `master` branch and work on something and run the `npm run webpack-watch` which generated a new `react.js`. However, when I opened `http://beta.mdn.localhost:8000/static/react.js` in the browser, it would actually return the `react.js.gz` from disk. So every attempt I made to make `react.js` didn't show up. It sounds easy and obvious now but you have no idea how many different voodoo tricks I tried of restarting, build-static, etc. etc. I think, well, I *suspect*, that Whitenoise works a bit like Nginx in that it does something like this:
```
if existsOnDisk(path + '.gz') and request.supports('accept-encoding', 'gzip') {
    return serve(path + '.gz')
} else {
  return serve(path)
} 
```

schalkneethling:
Hey @peterbe, as discussed, I opened https://github.com/mozilla/kuma/issues/5743 to write the documentation portion of this. I have also removed the `compression` plugin and rebased against master.

With that, this is ready to merge so(hoping to not step on toes), I am going to go ahead and merge this. This would also be super useful for me as I am working on converting the newsletter.

PR #5524, opened 2019-06-28
Add rules-of-hooks linting

schalkneethling:
Adds [rules of hooks](https://reactjs.org/docs/hooks-rules.html) linting and fixes three linting errors that was highlighted post install.

@davidflanagan r?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,81 +1,64 @@
 module.exports = {
-    'env': {
-        'browser': true,
-        'es6': true,
-        "jest/globals": true,
-        'jquery': true,
-        "node": true
+    env: {
+        browser: true,
+        es6: true,
+        'jest/globals': true,
+        jquery: true,
+        node: true
     },
-    'parser': "babel-eslint",
-    'parserOptions': {
-        "ecmaVersion": 6,
+    parser: 'babel-eslint',
+    parserOptions: {
+        ecmaVersion: 6
     },
-    "plugins": ["jest", "flowtype", "import", "jsx-a11y"],
-    'extends': [
+    plugins: ['jest', 'flowtype', 'import', 'jsx-a11y', 'react-hooks'],
+    extends: [
         'eslint:recommended',
-        "plugin:import/errors",
-        "plugin:flowtype/recommended",
-        "plugin:jsx-a11y/recommended",
-        "prettier"
+        'plugin:import/errors',
+        'plugin:flowtype/recommended',
+        'plugin:jsx-a11y/recommended',
+        'prettier'
     ],
-    'rules': {
+    rules: {
         'no-global-assign': 'error',
-        'indent': [
-            'error',
-            4,
-            {SwitchCase: 1}
-        ],
-        'linebreak-style': [
-            'error',
-            'unix'
-        ],
+        indent: ['error', 4, { SwitchCase: 1 }],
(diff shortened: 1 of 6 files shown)


.eslintrc.js
```diff
@@ -1,81 +1,64 @@
...
-        'one-var-declaration-per-line': [
-            'error',
-            'always'
-        ],
-        'new-cap': 'error'
+        eqeqeq: ['error', 'smart'],
+        'one-var-declaration-per-line': ['error', 'always'],
+        'new-cap': 'error',
+        'react-hooks/rules-of-hooks': 'error',
+        'react-hooks/exhaustive-deps': 'warn'
```
schalkneethling: Prettier formatted this file on save. The only changes other than that, are the two lines above and line 13.

peterbe:
It's amazing that we haven't "suffered" from these problems earlier. 

I'm strongly in favor of that eslint rule! I've had much success with it in side-projects.

.eslintrc.js
peterbe: Haha! It happens. By the way, I suspect it was your editor that ran prettier on this file because they way we have set up `prettier` in `package.json` is on a strict path directory. 
I think this is fuel for the argument that we should use `.prettierignore` or something global instead.

kuma/javascript/src/user-provider.jsx
```diff
@@ -108,7 +108,7 @@ export default function UserProvider(props: {
                     }
                 });
             });
-    }, []); // empty array means we only fetch on mount, not on every render
+    }, [ga]); // only fetch on mount, or if `ga` changes
```
davidflanagan: So, I read about the exhaustive-deps eslint plugin and it sounded like a good thing. But seeing it in action, I've got mixed feelings.

The intent here really is that we only want to run this effect (to fetch user data) once.  ga is a constant, and we know it is not going to change.  The empty array made it clear that the effect would only run once. Having eslint force us to add ga to the array isn't going to change the behavior (since it is constant) but it reduces the clarity of the code.

On the other hand, this is the plugin that would have found the bug that Peter pointed out in code review yesterday in my pr to change menu localization.

Is there a way to configure it to tell it that 'ga' should be exempt? Or do we decide to adopt this lint rule but then supress its warnings on lines like these when we want to explicitly override it? Or do we just go with the simplicity of doing what the linter tells us even when it makes the code less clear?

kuma/javascript/src/header/search.jsx
```diff
@@ -50,7 +50,7 @@ export default function Search(props: Props) {
         if (inputfield.current && props.initialQuery) {
             inputfield.current.value = props.initialQuery;
         }
-    }, []);
+    }, [props.initialQuery]);
```
davidflanagan: This seems like it might be a legitimate actual bug that the linter found here. I don't think it was causing any problem in practice, but this change is a real improvement.

kuma/javascript/src/task-completion-survey.jsx
```diff
@@ -120,7 +120,7 @@ export default function TaskCompletionSurvey(props: Props) {
                 nonInteraction: true
             });
         }
-    }, [shouldShow]);
+    }, [shouldShow, ga]);
```
davidflanagan: It is harmless to add ga here, but not actually necessary since it is known to be constant, so here the linter is giving us a less useful signal.

PR #5526, opened 2019-06-28
bug 1562262 - run prettier check in ci

peterbe:
This should break the first time. I'll fix that in a second commit just to prove that it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
         "optim:svg": "svgo jinja2/includes/icons/* --disable={removeViewBox,removeTitle} --multipass --pretty",
         "precommit": "npm run eslint && npm run stylelint",
         "test": "jest --testPathIgnorePatterns tests/sass",
-        "format": "prettier --write kuma/javascript/src/**/*{.js,jsx}"
+        "format": "prettier --write kuma/javascript/src/**/*{.js,jsx}",
+        "pretty": "prettier --check kuma/javascript/src/**/*{.js,jsx}"
     },
     "dependencies": {
         "@emotion/core": "^10.0.7",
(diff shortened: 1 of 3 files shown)


peterbe:
Our current `precommit` *checks* but we *could* switch that to format. Here's what you get
```
▶ git commit -a -m "run 'prettier --check' in the precommit"
husky > npm run -s precommit (node v12.3.1)

Checking formatting...
kuma/javascript/src/header/language-menu.jsx
Code style issues found in the above file(s). Forgot to run Prettier?

husky > pre-commit hook failed (add --no-verify to bypass)
```

Actually, I think I'm in favor of switching that last one to `format`. Because the `precommit` is something that affects humans, not TravisCI. Suppose you did actually make a "prettier mistake" then why bother wasting the time explaining that problem?? That poor human would then need to either edit the `.jsx?` file or know to just run `npm run format`. All of these things can just be automated away.

package.json
```diff
@@ -19,9 +19,10 @@
         "stylelint": "stylelint \"kuma/static/styles/**/*.scss\"",
         "eslint": "eslint --ext .js --ext .jsx kuma/",
         "optim:svg": "svgo jinja2/includes/icons/* --disable={removeViewBox,removeTitle} --multipass --pretty",
-        "precommit": "npm run eslint && npm run stylelint",
+        "precommit": "npm run eslint && npm run stylelint && npm run pretty",
```
peterbe: How about changing that last word from `pretty` to `format`?
[More thoughts here](https://github.com/mozilla/kuma/pull/5526#issuecomment-506813790)

peterbe:
As expected! https://travis-ci.org/mozilla/kuma/jobs/551878653

peterbe:
Actually, I think it's better to make the `precommit` exit instead. Because that's what the other tools do in the `npm run precommit`. Consistent.

davidflanagan:
I just saw the related bug you filed, and that approach of not hardcoding the files to format in package.json sounds great to me. Hope you can make it work.

PR #5530, opened 2019-06-28
Bug 1562293: don't crash in Google Translate

davidflanagan:
When our React-based pages are loaded in translate.google.com
various JS errors caused by Google make React unmount the components
that display the (now-translated) content. This patch works around
the problem by simply never calling ReactDOM.hydrate() when we're
loaded into Google Translate

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -56,7 +56,15 @@ if (container) {
         if (container.firstElementChild) {
             // If the container element is not empty, then it was presumably
             // rendered on the server, and we just need to hydrate it now.
-            ReactDOM.hydrate(app, container);
+            //
+            // If we're running under google translate, however, we skip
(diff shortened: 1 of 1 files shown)


peterbe:
Actually. Wait, it doesn't :(
Screen Shot 2019-06-28 at 4 35 26 PM

peterbe:
Ignore my penultimate comment. I was just not on the right git branch.

peterbe:
Screen Shot 2019-06-28 at 4 41 15 PM

davidflanagan:
Thanks, Peter, for figuring out how to test this with the serveo.net proxy. Display inside google translate isn't ideal. The menus don't work, for example. But clicking on links does seem to work, and the beta site looks as good as or better under translation than the non-beta site does:

Screen Shot 2019-06-28 at 1 56 21 PM

PR #5531, opened 2019-07-08
Bug 1563511, ensure new page content is read on navigation complete

schalkneethling:
This improves the accessibility of our client-side navigation by setting a `live-region` and also moving focus to the logo. In a follow on PR I am going to reintroduce the skip links for documentation pages. When I do that, I will also move focus to the container for these links as opposed to the logo. For now, this is a noticeable improvement.

### Here is the before

https://send.firefox.com/download/56f9e4901abcc514/#rnmfD8S5RpLXmbiPtC7Nwg

### Here is the after

https://send.firefox.com/download/2f2b7d36ef9b99cc/#c0M47TqnXzSoQJDt4f3P1Q

@callahad-mozilla-owner r?

### DIFF
--- a/kuma/javascript/src/router.jsx
+++ b/kuma/javascript/src/router.jsx
@@ -146,6 +146,7 @@ export default function Router({
     // analytics and the other stops the loading animation. Both
     // functions are defined at the bottom of the component.
     useEffect(recordAndReportAnalytics, [pageState.data]);
+    useEffect(resetFocus, [pageState.data]);
     useEffect(stopLoading, [pageState.data]);
 
     // When the page is first loaded, and this function is called for
@@ -512,6 +513,16 @@ export default function Router({
         // browser to be able to go back to it by client-side navigation
         // we need to set its state.
         history.replaceState(initialURL, '', initialURL);
+    }
+
+    /**
+     * Reset focus to the window after navigation
+     */
+    function resetFocus() {
+        let documentTitle = document.querySelector('a[class*="logoContainer"]');
+        if (documentTitle) {
+            documentTitle.focus();
+        }
     }
 
     // This is an effect function that runs after a new page (with new
--- a/kuma/static/styles/components/structure/document-head.scss
+++ b/kuma/static/styles/components/structure/document-head.scss
@@ -6,6 +6,7 @@
     padding: $first-content-top-padding 0 $last-content-bottom-padding 0;
 }
 
+div[class*='titlebar'],
 .document-title {
     h1 {
         @include set-font-size($mobile-document-title-font-size);
--- a/kuma/static/styles/components/structure/page-header/main-header.scss
+++ b/kuma/static/styles/components/structure/page-header/main-header.scss
@@ -31,6 +31,17 @@ Main header, wraps all components of logo, main, and secondary navigation
     }
 }
 
+a[class*='logoContainer'] {
+    outline: none;
+
+    /* do not remove the outline for keyboard users */
+    /* stylelint-disable selector-pseudo-class-no-unknown */
+    &:focus-visible {
+        outline: unset;
+    }
+    /* stylelint-enable */
+}
+
 .logo {
     position: relative;
     z-index: $logo-z;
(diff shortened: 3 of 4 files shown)


callahad:
(Switched reviewer to this account from my Mozilla admin account; will review tomorrow as soon as the Fx68 announcement goes out.)

kuma/javascript/src/router.jsx
```diff
@@ -514,6 +515,16 @@ export default function Router({
         history.replaceState(initialURL, '', initialURL);
     }
 
+    /**
+     * Reset focus to the window after navigation
+     */
+    function resetFocus() {
+        let documentTitle = document.querySelector('a[class*="logoContainer"]');
```
callahad: This feels a tiny bit weird, but will be addressed by your follow on PR, so looks good to me.
>  In a follow on PR I am going to reintroduce the skip links for documentation pages. When I do that, I will also move focus to the container for these links as opposed to the logo

Though I do wonder if the variable should be named something other than `documentTitle` if we're going to jump to a container of skip links.

kuma/static/styles/components/structure/document-head.scss
```diff
@@ -6,6 +6,7 @@
     padding: $first-content-top-padding 0 $last-content-bottom-padding 0;
 }
 
+div[class*='titlebar'],
```
callahad: Looks like we have classes with names ending in `titlebar` and `titlebarContainer`, and this selects both of them.

1. Is that the expected targeting?
2. Is this change necessary for this issue? I'm not sure understand its the relevance to screen readers.

kuma/static/styles/components/structure/page-header/main-header.scss
```diff
@@ -31,6 +31,17 @@ Main header, wraps all components of logo, main, and secondary navigation
     }
 }
 
+a[class*='logoContainer'] {
+    outline: none;
```
callahad: Will this still be necessary when we change the focus to the skip-link container? Do we want to have any sort of visible styling when non `focus-visible` users are tabbing through the page?

kuma/javascript/src/router.jsx
schalkneethling: This will all go away once https://github.com/mozilla/kuma/pull/5533 lands.

kuma/static/styles/components/structure/document-head.scss
schalkneethling: This is one of the big reasons I am also working on moving most of the styles away from emotion and back into Sass. Those generated class names that start with `css-randomness-class-name` is forcing one to use really expensive CSS selectors.

Also in general, styles should only be tightly coupled with a component if they are truly unique to the component. For right now, this thankfully does not have unintended consequences(that I noticed).

kuma/static/styles/components/structure/page-header/main-header.scss
schalkneethling: > Will this still be necessary when we change the focus to the skip-link container? 

Nope

> Do we want to have any sort of visible styling when non focus-visible users are tabbing through the page?

Where it makes sense for sure, but the main focus(excuse the pun ;p) of these are for users using a keyboard. For the most part, knowing where focus is when using a mouse is generally intuitive. There are places where we can improve though. I often find that having an `active` state or, clear indication when a clicked was registered is super useful.

schalkneethling:
Thanks you for the review @callahad @callahad-mozilla-owner 

BTW, which of the two should I by using :)

schalkneethling:
> Switched reviewer to this account from my Mozilla admin account;

Gotcha, so the `callahad` one, right?

callahad:
> Gotcha, so the `callahad` one, right?

Yep! Per our [admin account standard](https://mana.mozilla.org/wiki/display/POLICIES/Standard%3A+Admin+accounts), anyone with admin access to Mozilla's GitHub organization must use that via a separate, dedicated account named `${username}-mozilla-owner`.

My normal writing-and-reviewing-code account is plain ol' @callahad.

PR #5532, opened 2019-07-08
Bug 1563256, horizontal scrolling on mobile view

schalkneethling:
~~Our mobile views are rather broken now. This adds a temporary band-aid while I work on a larger "rewrite" to fix mobile. I am also hiding the interactive examples on mobile(phone) as it has no user benefit being shown at these viewport sizes.~~

@peterbe r?

### DIFF
--- a/kuma/javascript/src/document.jsx
+++ b/kuma/javascript/src/document.jsx
@@ -99,7 +99,7 @@ const styles = {
         [NARROW]: {
             // If we're narrower than a tablet, put the sidebar at the
             // bottom and drop the toc line.
-            gridTemplateColumns: '1fr',
+            gridTemplateColumns: '100%',
             gridTemplateAreas: '"main" "side"'
         }
     }),
--- a/kuma/javascript/src/interactive-examples.js
+++ b/kuma/javascript/src/interactive-examples.js
@@ -54,24 +54,38 @@ export function setLayout(root: HTMLElement) {
     if (mediaQuery && !mediaQuery.matches) {
         for (let iframe of root.querySelectorAll(ieSelector)) {
             if (iframe instanceof HTMLIFrameElement) {
-                // NOTE: if we just do the postMessage() right away the
-                // message probably won't get through because the iframe
-                // won't have loaded enough to be listening for messages.
-                // So instead we wait until the iframe is ready.  It is
-                // possible (though it seems unlikely) that the iframe
-                // could finish loading before this effect function gets
-                // called. If that happens we would miss the load event
-                // and never send the message. This doesn't seem like a
-                // problem in local testing, but if we don't reliably get
-                // the right layout in production then we might want to
-                // combine this load event handler with a setTimeout() call
-                // so that we post the message on load or after 1 second.
+                let messagePosted = false;
+                /* NOTE: if we just do the postMessage() right away the
+                 * message probably won't get through because the iframe
+                 * won't have loaded enough to be listening for messages.
+                 * So instead we wait until the iframe is ready.
+                 */
                 iframe.addEventListener('load', () => {
-                    iframe.contentWindow.postMessage(
-                        { smallViewport: true },
-                        ieOrigin
-                    );
+                    if (!messagePosted) {
+                        iframe.contentWindow.postMessage(
+                            { smallViewport: true },
+                            ieOrigin
+                        );
+                        messagePosted = true;
+                    }
                 });
+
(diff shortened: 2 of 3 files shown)


peterbe:
The functionality we need is one that matches how it works on mobile on developer.m.o today. I.e. the interactive example requires manual scrolling but it shouldn't need to mess up the rest of the page. 
IMG-2427

schalkneethling:
> Completely hiding the interactive examples on mobile is a non-starter. We can't damage the page for mobile.

@peterbe We have been meaning to do this, and from the experiments we did, it seemed like setting `display: none` worked as well as any other option. It is of no value to someone on a mobile phone and simply clutters the page. It was not the cause caused of the bug though. That was caused by the word "Recommendation" inside the "standard" table.

The real fix, therefore, is the `hyphens: auto;` bit. @atopal Would you be opposed to hiding interactive examples on mobile? I can remove that change if need be.

peterbe:
Actually, I don't know if I know how to test this. I started a SSH proxy to my `beta.mdn.localhost:8000` with your branch checked out. This is what happens when I open with my iPhone in Safari:


* The header gets all "scrunched up"
* When you reload the page, it starts with the font being much bigger and zoomed in and then it changes to look like this. 
* If I read the diff in the PR, it's supposed to `display:none` for me since this is smaller than a tablet but the interactive example is still there.

schalkneethling:
> The header gets all "scrunched up"

Yup. Mobile is pretty broken at the moment.

> If I read the diff in the PR, it's supposed to display:none for me since this is smaller than a tablet but the interactive example is still there.

Indeed and it does for me. Will test this on a device and emulator again to be sure.

atopal:
Let's separate out the interactive example change. When we tested this back in January, it had zero effect on page load time. So, people would pay the price, but not even see an interactive example. We can decide whether we want to be smarter about that in a separate issue.

schalkneethling:
@peterbe Changed to only include the change that adds `hyphens: auto;` to the spec table.

peterbe:
Perhaps I'm not understanding the problem or I'm not sure if my dev environment isn't updating. But I've tried restarts and all sorts of build-static commands. 
It still looks like this:
IMG-2435

The text/font is too big and in Safari if I double click the text Safari will zoom out pretty nicely. Either way, it's not something I have to on developer.mozilla.org. How does it work there?

Are you seeing that it's resolved when you test with your iPhone?

schalkneethling:
@peterbe Alrighty so first thing. I misunderstood this problem initially. Having understood the problem, this was quite the thing to figure out but, in the end, it was partially staring me in the face.

The thing that was staring me in the face was that the interactive examples were not receiving the `postMessage` on load that informs the code that it should apply mobile styles. You, therefore, will notice that the CSS interactive examples input and output is shown side by side instead of being stacked.

Screenshot 2019-07-11 at 10 36 54

I found the code that manages these `postMessages` in the React portion and noticed that David left a comment which indicated that we might run into this problem. Seeing that we did indeed hit the issue, I added the `setTimeout` function with a delay of one second, as well as a variable to track whether the message was sent on `load`.

This resolved the problem with regards to the layout of the interactive examples. It did however not entirely fix the problem of the "zoomed" text. At the moment our layout "grid" is a combination of `flexbox`, `grid` as well as some `floats` from our previous codebase. Emotion also probably uses `matchMedia` to apply various styling at different breakpoints, and so it is not always clear when what is applied i.e. you do not see the normal tell of `@media screen and (min-width...`.

In the end, I found that the main content and sidebar's `grid-template-columns` is set to `1fr` on mobile which normally would work(I would think) but, perhaps because we have a bit of a mix of things at the moment, it was not working entirely as expected. Changing this to be `100%` instead of `1fr` fixed the problem.

While there is still horizontal scrolling, it now matches the behavior on production.

### Production

Screenshot 2019-07-11 at 11 56 39

### Beta

Screenshot 2019-07-11 at 11 57 07

I have tested this using devtools in Firefox, Chrome, and Safari as well as on a real iPhone 7 using Safari, Chrome(above), Safari on iPhone XS as well as Chrome on Android(below):

Screenshot 2019-07-11 at 12 43 37

atopal:
Thanks for the explanation, Schalk! You mention a 1s setTimeout delay. Do you thin that will have a performance impact?

schalkneethling:
> Do you thin that will have a performance impact?

I very much doubt that. It does not block execution, it simply waits before executing the postMessage.

PR #5533, opened 2019-07-08
Bug 1563466, add skip links to document header

schalkneethling:
Adds "Skip to" links back to all pages. @peterbe r?


schalkneethling:
@peterbe Once the document has loaded(best to use Chrome here, Firefox is a little finicky on Mac), hit the tab key and you should see the following:

Screenshot 2019-07-09 at 20 21 14

Press tab again

Screenshot 2019-07-09 at 20 21 22

Press tab again

Screenshot 2019-07-09 at 20 21 29

This should work the same on the landing page as well as document pages.

peterbe:
It appears now but it's not working. I press Tab once and "Skip to main content" appears but nothing happens when I hit Enter.
Screen Shot 2019-07-09 at 3 47 14 PM

peterbe:
If I press Tab twice (Skip to select language) it just focus the little select drop-down in the footer. 
Screen Shot 2019-07-09 at 3 48 26 PM

peterbe:
Nothing happens with "Skip to search either".

Actually, the "Skip to language" did work. It just was me not noticing what it did for me. I think I expected it to open the "Translations" thing that usually sits on the right-hand side underneath the search widget. 

By the way this is probably the explanation:

Screen Shot 2019-07-09 at 3 50 38 PM

kuma/javascript/src/components/a11y-nav.jsx
```diff
@@ -0,0 +1,26 @@
...
+                </a>
+            </li>
+            <li>
+                <a id="skip-search" href="#q">
+                    {gettext('Skip to search')}
+                </a>
+            </li>
+        </ul>
+    );
+}
```
peterbe: Perhaps it's nitpickery but I really don't like the location of this file. "components". 

I've seen projects that organize their files like this:

* "components"
  * "Home"
  * "About"
  * "Dashboard"
* "views"
  * "Home"
  * "About"
  * "Dashboard"
* "test"
  * "Home"
  * "About"
  * "Dashboard"

And I think that's a mistake because, if you decide to remove the "Dashboard" you have to search around for it all over the code base. A much better structure would be:

* "Home"
  * "component"
  * "view" 
  * "tests"
* "About"
  * "component"
  * "view" 
  * "tests"
* "Dashboard"
  * "component"
  * "view" 
  * "tests"

"Design for deletability" someone said and I can't find who. 

I'd much rather see a directory called "a11y" or something. Or, simply just call it "kuma/javascript/src/a11y-nav.jsx` which I think is a lot more matching to the kind of naming style that David started.

schalkneethling:
> Perhaps it's nitpickery but I really don't like the location of this file. "components".

Fair enough, that is probably not the best solution anymore. I would prefer a little more structure than just having everything in the root of the `javascript` folder. For this one, and then related accessibility components, I will use a folder called `a11y` inside `javascript`

> Perhaps I'm being a stickler but I don't think we should land this until it actually works. I.e. I think we need to make those DOM ID nodes available.

+1 My bad for assuming these still exist.

schalkneethling:
@peterbe Updated and tested with VoiceOver. Thanks!

PR #5534, opened 2019-07-08
Bug 1562861, svg icons are cropped and zoomed

schalkneethling:
@peterbe r?


kuma/static/styles/base/icons.scss
```diff
@@ -265,6 +267,15 @@ button[type='submit'] .icon-check-mark {
     vertical-align: text-bottom;
 }
 
+/* TODO: Need to remove the need for this be removing the
+   styling added from the React side */
```
peterbe: I don't understand the comment.

schalkneethling: There is styling on the React side that needs to be moved here. It just makes more sense to have all styling related to icons in one place. But, seeing that the history button is being removed, this is moot now ;)

schalkneethling:
> Looks great! 

Thanks!

> The "clock icon" looks a bit odd being taller than the "Edit" button but perhaps that clock is going away anyway.

Yeah, but seeing that it is being removed, I am not going to worry about it here. Thanks, Peter.

PR #5535, opened 2019-07-08
bug 1564038 - validation error on too long revision comment

peterbe:
When you make an edit, like on http://mdn.localhost:8000/en-US/docs/Tools$edit there's a "comment" field there *was* is `<input type="text" name="comment" maxlength="255">` which is now `<input type="text" name="comment">`. 

And on the http://mdn.localhost:8000/en-US/docs/Tools$revert/1555555 page at the bottom there used to be a `<textarea name="comment"></textarea>` which would/could invite people to type something longer than 255. **But!** It's weirder than that. Suppose that I typed in a 250 character comment into that textarea, you still get a hard ISE because MySQL barfs on it. It's because the comment it tries to save becomes `Revert to revision of 2019-02-26 09:45:01.060271 by irenesmith: <my 250 character comment>`

If it's going to be a textarea, make the database field a textfield I say. 

Now, what scares me a little is how will MySQL handle doing an alter like this on the prod database? When I ran it on my laptop it was blazing fast but I only have a couple of thousand revisions. Whereas in prod we'll have hundreds of thousands. 

I know that deep inside *PostgreSQL* there's no such thing as a "CharField" really. Deep down they're all textfields but with added guards (aka. constraints)

If you think this change is too dangerous; I can revert and do something like...:

```python
# add a sensible comment
revision.comment = ("Revert to revision of %s by %s" %
                                (revision.created, revision.creator))
if comment:
    revision.comment = u'%s: "%s"' % (revision.comment, comment)
    if len(revision.comment) > 255:
        revision.comment = revision.comment[:255 - 3] + '...'
```


peterbe:
Ping @escattone 
Don't want it to rot.

atopal:
@peterbe  Ryan tried this on staging and it was a pretty big change, running for almost half an hour, it would be even longer on production. I don't think this is a big enough issue to spend as much time on. Can we truncate the message before we save as your example suggests?

escattone:
@atopal I totally agree that this was not worth the effort. In hindsight, I should have recommended the less dangerous approach. I decided it was probably more work to revert it though (I'd have to migrate the stage DB back to the way it was), so I went ahead on production.

escattone:
The production DB migration took over 2 hours, and when I restarted the push (since the migration job lost its connection to the DB, the first run failed) it seems to have thought the migration was not done, so it started it all over again 😡 .

escattone:
Actually I was mistaken. The DB migration was being killed after 30 minutes by a daemon we run that kills long-running SQL commands, and then K8s would attempt another DB migration, and this went on 4 times. I had to scale down the long-running-SQL-killer (`kc -n percona-toolkit scale --replicas=0 deploy percona-toolkit-ptkill`), then re-launch the DB migration (which took ~1 hour), and finally scale the long-running-SQL-killer  back-up.

peterbe:
I guess it means alter tables on wiki_revision is not an option anymore. Those days are over. 
I'm never comfortable with mysql but I bet we *could*  potentially allow some operations that might be cheaper in terms of I/O such as removing columns. 
Or, just not bother anymore and learn to live with the db structure we have on those large tables. Sad though.

PR #5538, opened 2019-07-09
Bug 1562853, scrollbars at bottom of article, sidebar on windows

schalkneethling:
Fixes scrollbars at the bottom of doc pages.

## Before

Screenshot 2019-07-09 at 08 46 43

## After

Screenshot 2019-07-09 at 08 47 29

NOTE: This only happens on Windows - @callahad r?


peterbe:
If you can confidently test this fixes it in Window, then I think you should just merge it. The merge button would be disabled if you don't have a review. Dan is busy but I can approve it.

PR #5539, opened 2019-07-09
Bug 1563473, section links do not have descernable text

schalkneethling:
Uses our standard section link icon and adds descriptive text for screen readers. @peterbe r?


peterbe:
What do I need to do to update to get this CSS?
Screen Shot 2019-07-10 at 10 12 31 AM

kuma/javascript/src/components/section-link.jsx
```diff
@@ -0,0 +1,22 @@
...
+ * Returns a section anchor element with a nested SVG icon
+ * @param {Object} heading - The current heading DOM node
+ * @returns section anchor element as an HTML element
+ */
+export default function sectionAnchor(heading) {
+    const anchor = document.createElement('a');
+    anchor.href = `#${heading.id}`;
+    anchor.classList.add('section-link');
+    // heading.innerText is the current heading and will already be translated
+    anchor.setAttribute('aria-label', gettext('Link to ') + heading.innerText);
```
peterbe:
```suggestion
    anchor.setAttribute('aria-label', `${gettext('Link to')} ${heading.innerText}`);
```
Otherwise the localizers would have to be confronted with the extra whitespace and it might confuse or they might delete it.

kuma/javascript/src/article.jsx
```diff
@@ -108,11 +109,7 @@ function highlightSections(article) {
 // kuma/static/js/components/local-anchor.js
 function addAnchors(article) {
     for (let heading of article.querySelectorAll('h2[id], h3[id]')) {
-        let anchor = document.createElement('a');
-        anchor.href = `#${heading.id}`;
-        anchor.classList.add('sectionLink');
```
peterbe: [See line 41 in this file.](https://github.com/mozilla/kuma/blob/b6e10cf7607a41b2fad830dbb77cea50502ae9cd/kuma/javascript/src/article.jsx#L41)

It's kinda sad what's going on here. The whole reason for using css-in-js (`emotion` in our case) is that when you remove the JS it should imply the removal of the CSS. But that's not the case here. It's because we're using `DOMElement.classList(string)` and not `css={styles.sectionLink}`. Gross.

I wonder, do we really need React for any of this stuff?? From what I can see, the `local-anchor.js` is pure/vanilla JS. Its only flaw is that it's not callable on repeats. You could easily solve that by changing that file to:

```javascript
window.prependLocalAnchors = (function() {
    'use strict';

    function prependLocalAnchors() {
        /// ...the magic....
    }

    // `Array.from` is not supported in IE, progressively enhance
    if (typeof Array.from !== 'undefined') {
        window.addEventListener('load', function() {
            prependLocalAnchors();
        });
    }
    return prependLocalAnchors;
})();
```

Then, `article.jsx` would just look something like this:
```javascript
function addAnchors(article) {
    // SOME GOOD COMMENT EXPLANATION
    if (window.prependLocalAnchors) {
        window.prependLocalAnchors();
    }
}
```

I don't know what that means for django-pipeline etc to share a piece of JS code across both worlds but it would certainly solve a lot of things. Also, `local-anchor.js` seems to depend on `window.mdnIcons` which I don't know if we have or want to have in React land.

kuma/javascript/src/components/section-link.jsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
```
peterbe: Let's not use the folder "components". This file isn't even a component. It's a little helper function.

kuma/javascript/src/components/section-link.jsx
```diff
@@ -0,0 +1,22 @@
...
+ * @returns section anchor element as an HTML element
+ */
+export default function sectionAnchor(heading) {
+    const anchor = document.createElement('a');
+    anchor.href = `#${heading.id}`;
+    anchor.classList.add('section-link');
+    // heading.innerText is the current heading and will already be translated
+    anchor.setAttribute('aria-label', gettext('Link to ') + heading.innerText);
+    // render the SVG into the anchor element
+    render(<Anchor />, anchor);
```
peterbe: Any idea if this is expensive?

schalkneethling:
> What do I need to do to update to get this CSS?

@peterbe Running `docker-compose exec web make build-static` should do the trick

kuma/javascript/src/components/section-link.jsx
schalkneethling: Agreed. I missed this one when I changed some of this earlier today

kuma/javascript/src/article.jsx
schalkneethling: That should work. If you feel that is a better approach, we could go that route.

kuma/javascript/src/components/section-link.jsx
schalkneethling: I do not and was wondering about that. But this should happen on the server no? I did not notice anything that suggested this is not a good idea, but I honestly don't know.

peterbe:
I've been using this for a long time. Is it not right?
```
▶ cat peterbe_build-static.sh
#!/bin/bash

# Do what `make build-static` does but without `docker-compose up`

# webpack
docker-compose exec web npm run webpack

# compilejsi18n
docker-compose exec web mkdir -p build/locale
docker-compose exec web ./manage.py compilejsi18n

# compile-react-i18n
docker-compose exec web mkdir -p build/locale
docker-compose exec web ./manage.py compilejsi18n -d react -f json

# collectstatic
docker-compose exec web ./manage.py collectstatic --noinput
```

schalkneethling:
> I've been using this for a long time. Is it not right?

That probably works as well. I have just always done `docker-compose exec web make build-static` followed by a `docker-compose restart` just to be sure the latest is picked up.

kuma/javascript/src/components/section-link.jsx
peterbe: Effects aren't run in SSR React. Also, the reason it's affect is because you want to to trigger when you change the document. I.e. clicking on an internal link and doing a pushState move.

schalkneethling:
Just ran the docker-compose exec web make build-static and it looks correct for me locally:

Screenshot 2019-07-10 at 17 27 07

kuma/javascript/src/components/section-link.jsx
schalkneethling: Ah, I see. Not sure how we can test whether it will be/is expensive. Another options are https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup, but seemed like too much for what this is doing. I could definitely be wrong though

kuma/javascript/src/section-link.jsx
```diff
@@ -0,0 +1,25 @@
...
+    const anchor = document.createElement('a');
+    anchor.href = `#${heading.id}`;
+    anchor.classList.add('section-link');
+    // heading.innerText is the current heading and will already be translated
+    anchor.setAttribute(
+        'aria-label',
+        `${gettext('Link to')} ${heading.innerText}`
+    );
+    // render the SVG into the anchor element
+    render(<Anchor />, anchor);
```
davidflanagan: This seems scary to me. Since this is being called from an effect function that is touching up pre-defined HTML, it would be cleaner if we could do it without react.  I don't know if this is a performance issue or not, but I think that the only other place we call render() is at the toplevel in index.jsx.

I think it would be cleaner to define the svg icon as a string constant in this file and then use innerHTML to inject it into the anchor.  Or maybe better use innerHTML to define it the first time in an off-screen element, and then call cloneNode(true) to make a copy each time you need another icon.

kuma/javascript/src/section-link.jsx
```diff
@@ -0,0 +1,25 @@
...
+ */
+export default function sectionAnchor(heading) {
+    const anchor = document.createElement('a');
+    anchor.href = `#${heading.id}`;
+    anchor.classList.add('section-link');
+    // heading.innerText is the current heading and will already be translated
+    anchor.setAttribute(
+        'aria-label',
+        `${gettext('Link to')} ${heading.innerText}`
+    );
```
davidflanagan: The reason that I didn't put a label in here is that Sheppy had filed a bug about the "Link to" text looking like part of the section title to some tools.   Presumably using aria-label instead of a hidden label element fixes that problem.

kuma/javascript/src/section-link.jsx
peterbe: Both Schalk and I tried to find a reason not to use `render()`. It worked but it was hard to figure out if it was dangerous or slow. It's quite possibly async and relatively fast so you can't just...

```javascript
console.time('render');
render(<Anchor />, anchor);
console.timeEnd('render');
```
since it'll always be a tiny number and there's a chance there's something async and deep. 

Makes me wonder, it is perhaps better to log how often the article re-renders after a hydration and see if that number changes depending on this line.

schalkneethling:
> Makes me wonder, it is perhaps better to log how often the article re-renders after a hydration and see if that number changes depending on this line.

That might very well be worthwhile to do @peterbe, good idea.

PR #5540, opened 2019-07-09
Bug 1563461, logo svg missing title

schalkneethling:
Adds `aria-label` to MDN Web Docs logo link. @callahad-mozilla-owner r?


MarcoZehe:
I think the visually hidden labels also need the "Search MDN" text localized, since this is stuff read by screen readers, and having that in the language of the page would be much preferred. In the PR right now, only the visually displayed label has the gettext call.

PR #5541, opened 2019-07-09
Bug 1563464, missing label for search input field

schalkneethling:
@callahad-mozilla-owner r?


MarcoZehe:
I think the visually hidden labels also need their strings localized. Right now, only the visually visible label has a localized string. But the visually hidden label is what is being read out loud by the screen reader, and having that in the language of the page would be much preferred.

PR #5542, opened 2019-07-09
Implement beta version of signaling feature.

johngian:


johngian:
@escattone Here is the beta version of #5469 adapted after your last comments in the PR.

escattone:
@johngian There's a JS linting error:
```
/home/travis/build/mozilla/kuma/kuma/javascript/src/bcd.js
  279:9  error  Move function declaration to function body root  no-inner-declarations
```
But I don't know why `eslint` calls that an error. Seems like a design choice to me.

kuma/javascript/src/utils.js
```diff
@@ -0,0 +1,16 @@
+// Helper to get cookie values
+export function getCookie(name) {
```
schalkneethling: Add JSDoc style comment. The name is self-evident but, would be worth doing so we can clearly indicate what type `name` is expected to be, and what is returned.

```
/**
 * What does it do?
 * @params {type} name - ...
 * @returns What does it return?
 */
```

kuma/javascript/src/bcd.js
```diff
@@ -262,3 +263,62 @@ export function activateBCDTables(root: HTMLElement) {
...
+            const signalLink = document.createElement('a');
+            signalLink.textContent = 'Flag as incorrect';
+            const signalApiUrl = '/api/v1/bc-signal';
+
+            const payload = {
+                'slug': slug,
+                'locale': locale
+            };
+
+            signalLink.addEventListener('click', function() {
```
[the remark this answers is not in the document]
johngian: For now I suggest we use the same (or similar) codebase between the 2 versions. Thoughts ?

johngian:
@schalkneethling I updated the PR and now travis is green.

kuma/javascript/src/utils.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Helper function to get cookie value given a cookie name
+ * @params {string} cookie name
+ * @returns {string} cookie value
+ */
+export function getCookie(name) {
+    var cookieValue = null;
+    if (document.cookie && document.cookie !== '') {
```
escattone: Nit (and I realize I didn't catch this in https://github.com/mozilla/kuma/pull/5469): Since an empty string is falsy, I think this can be simplified to `if (document.cookie) {`

kuma/javascript/src/utils.js
```diff
@@ -0,0 +1,22 @@
...
+        var cookies = document.cookie.split(';');
+        for (var i = 0, l = cookies.length; i < l; i++) {
+            var cookie = cookies[i].trim();
```
escattone: Nit (and again I realize I didn't catch this in https://github.com/mozilla/kuma/pull/5469): I think these first two lines could be simplified to:
```javascript
for (var cookie of cookies) {
    cookie = cookie.trim()
    ...
}
```

kuma/javascript/src/bcd.js
```diff
@@ -262,3 +263,61 @@ export function activateBCDTables(root: HTMLElement) {
...
+    /**
+     * Creates and returns the signal element HTML
+     * @returns signal element HTML
+     */
+    function signalElem() {
+        const extContainer = document.createElement('div');
+        const separator = document.createElement('hr');
+        const container = document.createElement('div');
+        const signalLink = document.createElement('a');
+        signalLink.textContent = 'Flag as incorrect';
```
escattone: I think this should be wrapped in `gettext`, so:
```
signalLink.textContent = gettext('Flag as incorrect');
```

johngian:
@schalkneethling @escattone PR updated

PR #5543, opened 2019-07-09
bug 1561055 - /api/v1/search/en-US?q=stuff includes too much

peterbe:
http://beta.mdn.localhost:8000/en-US/search?q=stuff still seems to work at least. 

It started with that I just wanted to see how the `.to_dict()` stuff worked then I realized it would be easy to change and it would give me a chance to be challenged by jest and flow and just generally working on the post-David code. 
Then I noticed that there was no Python test coverage at all on the new API search so I spent some time on that. It's not perfect and it's a big fat test but it's a lot better than nothing.

### DIFF
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -370,7 +370,7 @@ class APISearchIntegration(ElasticTestCase):
         hits = response.json()['hits']
         assert hits
         expected_keys = set(
-            ['score', 'excerpts', 'summary', 'title', 'slug', 'tags'])
+            ['excerpts', 'summary', 'title', 'slug'])
         for hit in hits:
             assert set(hit.keys()) == expected_keys
 
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -3,12 +3,15 @@ from __future__ import unicode_literals
 from functools import partial
 
 import pytest
+from django.conf import settings
+from django.test import override_settings
 from waffle.models import Flag, Sample, Switch
 
 from kuma.api.v1.views import (document_api_data, get_content_based_redirect,
                                get_s3_key)
 from kuma.core.tests import assert_no_cache_header
 from kuma.core.urlresolvers import reverse as core_reverse
+from kuma.search.tests import ElasticTestCase
 from kuma.users.templatetags.jinja_helpers import gravatar_url
 from kuma.wiki.jobs import DocumentContributorsJob
 from kuma.wiki.templatetags.jinja_helpers import absolutify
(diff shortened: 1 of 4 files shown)


kuma/api/v1/views.py
```diff
@@ -268,7 +271,31 @@ def search(request, locale):
...
+    for hit in response:
+        # Be explicit about what to return.
+        # Only include things the React component might and will use.
+        hit_dict = hit.to_dict()
+        # Not every document has 'tags' so set this default.
+        hit_dict.setdefault('tags', [])
+        # 'locale' might get returned from hit.to_dict(). Remove it if there.
+        hit_dict.pop('locale', None)
+        # This one comes from the meta
+        hit_dict['score'] = hit.meta.score,
```
escattone: I noticed that `hit.meta.score` seems to always be a list with a single item, the actual score, and that score is different than the score that's being used currently, `hit._score`. Does it make sense to assign the actual score to `hit_dict['score']` rather than the list, like `hit_dict['score'] = hit.meta.score[0]`? Or should we use `hit._score` instead?

kuma/api/v1/tests/test_views.py
```diff
@@ -321,3 +324,53 @@ def test_whoami(user_client, api_settings, wiki_user, beta_testers_group,
...
+    def test_search_en_US(self):
+        url = reverse('api.v1.search', args=['en-US'])
+        response = self.client.get(url, HTTP_HOST=settings.BETA_HOST)
+        assert response.status_code == 400
+        assert response['content-type'] == 'application/json'
+        assert response.json() == {
+            'error': "No query string 'q'"
+        }
+
+        response = self.client.get(
```
escattone: Nit: I'd lean toward making each of these sections a separate test method, but feel free to ignore if you like it as is.

peterbe: Good point. I think I did it like that to over-optimize. It makes it faster by not having to have to do the expensive index set up and tear down and the fixture loading.

kuma/api/v1/views.py
peterbe: It's a list!? That's sad that I didn't even notice. Clearly, the score isn't being used.

peterbe:
Note-to-self; https://github.com/mozilla/kuma/pull/5574 changes the v1 search api to not include tags. Land that first before fixing these conflicts here.

kuma/api/v1/tests/test_views.py
peterbe: I was pondering it and decided it's not worth it. The setup/teardown is more expensive than the value in doing the tests "properly".

kuma/api/v1/views.py
peterbe: I'm feeling kinda aggressive today. Like with `tags`, if we're not displaying it, we shouldn't be sending it. APIs are hard to modify when they're established so it's always a good idea to work up slowly from the absolute minimal. 
So, the overall goal of this PR is to reduce the returned JSON to the absolute minimal that is used in the React UI view code.

peterbe:
Why isn't Travis starting?! I'm going to try to "restart" the PR.

peterbe:
[That worked!](https://travis-ci.org/mozilla/kuma/builds/562728240)

peterbe:
I'm in the midst of rewriting how site-search works anyway.

PR #5544, opened 2019-07-09
Bug 1563478, names in contributor list should be linked

schalkneethling:
This turns the contributors list into a functional component, uses semantic HTML, moves styling to Sass and links contributor names. For now, the names are linked to the wiki domain. Once we have profiles in React, we should update `profileBaseURL`

Also note, this depends on https://github.com/mozilla/kuma/pull/5534 for the icon to display correctly. @peterbe r?


kuma/javascript/src/document/contributors.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { gettext } from '../l10n.js';
+
+import ContributorsIcon from '../icons/contributors.svg';
+
+type Props = {
+    contributors: Array
+};
+
+export default function Contributors(props: Props) {
```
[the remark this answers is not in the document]
schalkneethling: Don't know. I followed David's example here:
https://github.com/mozilla/kuma/blob/master/kuma/javascript/src/header/search.jsx#L40

kuma/javascript/src/document/contributors.jsx
```diff
@@ -0,0 +1,40 @@
...
+    const profileBaseURL = 'https://wiki.developer.mozilla.org/profiles/';
```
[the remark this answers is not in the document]
schalkneethling: So, the `document` does contain the `editURL` but that points to the wiki. Assuming that is ok, I could do something like:

```
const url = new URL(document.editURL);
const profileBaseURL = `${url.protocol}//${url.hostname}/profiles/`;
```

Which would produce: `https://wiki.developer.mozilla.org/profiles/`

kuma/javascript/src/article.jsx
```diff
@@ -6,12 +6,13 @@ import { css } from '@emotion/core';
 import { activateBCDTables } from './bcd.js';
 import { addLiveExampleButtons } from './live-examples.js';
 import ClockIcon from './icons/clock.svg';
-import ContributorsIcon from './icons/contributors.svg';
 import { gettext } from './l10n.js';
 import { highlightSyntax } from './prism.js';
 import * as InteractiveExamples from './interactive-examples.js';
 import TagsIcon from './icons/tags.svg';
 
+import Contributors from './document/contributors.jsx';
```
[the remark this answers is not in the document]
schalkneethling: Me and my directories ;p

schalkneethling:
@peterbe Updated

schalkneethling:
Resolved conflicts:

Screenshot 2019-07-11 at 16 34 39

schalkneethling:
As I have rebased against master, it is all good now:

Screenshot 2019-07-11 at 17 38 49

kuma/javascript/src/article.jsx
```diff
@@ -155,6 +152,9 @@ export default function Article({ document }: DocumentProps) {
 }
 
 function ArticleMetadata({ document }: DocumentProps) {
```
peterbe: This demonstrates how to do destructing instead of `(props: Props)`. Care to take a stab?

kuma/javascript/src/contributors.jsx
```diff
@@ -0,0 +1,40 @@
...
+import { gettext } from './l10n.js';
+
+import ContributorsIcon from './icons/contributors.svg';
+
+type Props = {
+    contributors: Array,
+    profileBaseURL: String
+};
+
+export default function Contributors(props: Props) {
```
peterbe:
```suggestion
export default function Contributors({contributors, profileBaseURL}: ContributorProps) {
```
untested.

kuma/javascript/src/contributors.jsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+        <section className="contributors-sub">
+            <ContributorsIcon />
+            <header>
+                <h4>{gettext('Contributors to this page:')}</h4>
+            </header>
+            <ul>
+                {/*
+                 * TODO: These are temporarily pointing to the wiki domain,
```
peterbe: Not so temporary, is it?
After all, the profiles are and always will in the context of the wiki. 
I think you can simply delete this multi-line comment.

kuma/static/styles/components/wiki/contributors.scss
```diff
@@ -17,3 +19,29 @@ contributors block at the bottom of the article
         display: none;
     }
 }
+
+/*
+React contributors block at the bottom of the article
```
peterbe: I'm guessing we don't yet have a framework for separating SCSS/CSS for the wiki vs. the react. David had a "framework" where he'd eye over the JS in the jQuery and just rewrite in these new .jsx files.

peterbe:
Obviously, we need to figure out the jest errors too :) 
Let me know if you want help to share the load.

schalkneethling:
@peterbe updated. Looking at the Jest errors.

schalkneethling:
@peterbe updated

kuma/javascript/src/contributors.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+type ContributorProps = {
+    contributors: Array,
+    profileBaseURL: String
+};
+
+export default function Contributors({
+    contributors,
+    profileBaseURL
+}: ContributorProps) {
```
peterbe: 👍

kuma/javascript/src/last-modified-by.jsx
```diff
@@ -0,0 +1,36 @@
...
+                <h4>{gettext('Last updated by:')}</h4>
+            </header>
+            <a href={`${profileBaseURL}${lastModifiedBy}`} rel="nofollow">
+                {`${lastModifiedBy}, `}
+            </a>
+            <time dateTime={lastModified}>
+                {new Date(lastModified)
+                    .toISOString()
+                    .slice(0, -5)
+                    .replace('T', ' ')}
```
peterbe: On the old site:
```
 Last updated by: estelle, Apr 25, 2019, 10:30:38 AM
```
On the new site:
```
Last updated by:estelle, 2019-04-25 14:30:38
```
(ignore the whitespace)

Hmm... First of all, I generally prefer ISO dates. But...

Now I don't know when estelle made this change because there's no timezone information. Suppose that there was timezone information I might not have the mental capacity to figure out when that was. E.g. "2019-07-11T19:30:25.648Z" I think means 19:00 minus 6 hours so 3pm here on US eastern. Ugh. Tired. 

We *could* use `Dateobj.toString()` but then you'd get something like this:
```
Thu Jul 11 2019 15:31:32 GMT-0400 (Eastern Daylight Time)
```
That looks visually appealing to my eyes but it's going to end up always being in the timezone of the SSR server (US Pacific) and in a US-centric internationalization format. 

My preferred solution to these things is usually to do something like GitHub does:
```
<time title="2019-07-09T19:30:25.648Z">2 days ago</time>
```
which is super easy to parse/read and understand and if you so desperately need to know exactly the full ISO timezone-aware string is there. 

I would be happy to NOT block this PR just because of this but instead copy my comments into a fresh new bug and from there drag in more people for feedback (or bikeshedding).

peterbe: PS. I looked into my MySQL database. estelle made that change at 10:30 pacific time which means 13:30 by time (because I'm US Pacific + 3h). 
So, right now I have no idea where this 14:30 comes from.

peterbe: Oh I get it now! 
What your code is *trying* to display to me is the GMT time. 
```
new Date("2019-04-25T10:30:38.382093")
Date Thu Apr 25 2019 10:30:38 GMT-0400 (Eastern Daylight Time)

new Date("2019-04-25T10:30:38.382093").toISOString()
"2019-04-25T14:30:38.382Z"
```
That's incorrect. It's not the truth. The best solution now would be to try to display it in a somewhat realistic fashion. Perhaps best to just show it like the current wiki does. Then we can worry about it after this PR lands.

peterbe: Urgh. This is hard that I thought. In the wiki, the date formatting is done with [`babel`](https://pypi.org/project/django-babel/). Something like this:

```python
from babel import dates
# tzvalue is the settings.TIME_ZONE + the MySQL naive datetime.
# locale is the locale of the page (obiously)
return dates.format_datetime(tzvalue, locale=locale)
```

So we need to try to do something similar. I.e. we need the supply the document locale to JavaScript's date formatting function. So far so good:

```
d = new Date()
Date Thu Jul 11 2019 16:30:39 GMT-0400 (Eastern Daylight Time)

d.toLocaleDateString('sv-SE')
"2019-07-11"
d.toLocaleString('sv-SE')
"2019-07-11 16:30:39"
d.toLocaleString('fr')
"11/07/2019 à 16:30:39"
d.toLocaleString('en-US')
"7/11/2019, 4:30:39 PM"
```
But first we need to convert our stupid naive datetime string into a timezone aware one.

peterbe: The thing is, the timezone information is lost in the JSON document API. I can find it out when I'm inside Django by looking at `django.conf.settings.TIME_ZONE`. 

"So, it's 'US/Pacific' you say. Ok, that's -7h from UTC. Everyone knows that."
If you're equipped with this "-7" fact you can do this: 
```javascript
var d = new Date(`${lastModified}-07:00`);
d.toISOString()
"2019-04-25T17:30:38.382Z"
```

And that's correct. I checked. 10:30 US Pacific is 17:30 in UTC.

peterbe:
(Hopefully) for your convenience: https://github.com/schalkneethling/kuma/pull/1

schalkneethling:
Thanks, @peterbe, and @escattone for digging into this. The code that formats the DateTime is still that done by David. I do agree that there is a better solution but, that just displaying the raw format works for now and moves us forward :+1:

I will make the change and update the PR ⏲

peterbe:
For the record; here's the forked discussion about displaying dates: https://bugzilla.mozilla.org/show_bug.cgi?id=1565611

kuma/javascript/src/last-modified-by.jsx
```diff
@@ -0,0 +1,31 @@
...
+    return (
+        <section className="contributors-sub">
+            <ClockIcon />
+            <header>
+                <h4>{gettext('Last updated by:')}</h4>
+            </header>
+            <a href={`${profileBaseURL}${lastModifiedBy}`} rel="nofollow">
+                {`${lastModifiedBy}, `}
+            </a>
+            <time dateTime={lastModified}>{lastModified}</time>
```
peterbe: 👍

PR #5547, opened 2019-07-10
bug 1564961 - Stop using codecov on kuma PRs

peterbe:


peterbe:
DON'T MERGE just yet!

I've email Codecov.io and invite them to the constructive criticism. Perhaps we're doing it wrong. Perhaps there are options. If we can get the report only iff there are large new gaps in the total number of lines missing coverage.

hootener:
:wave: 

Hi I'm the CTO of codecov and was directed to this PR through the email you sent to us. I want to first understand this a little better:

> If we can get the report only iff there are large new gaps in the total number of lines missing coverage.


Codecov has several methods for coverage gating and PR comment gating. Full disclosure, none of these may be what you're looking for. But  at the very least I'd like to understand this use case a little better so that if we can't currently support your workflow, we can perhaps attempt to in the future. 

Directed questions:

1. Where are the large gaps you want to be alerted on?  I assume the code that's being added in the branch of the PR? Would `require_changes` potentially be helpful, here? https://docs.codecov.io/docs/pull-request-comments#section-requiring-changes
2. How large is a large gap?  For example, our commit statuses have a threshold option that will pass/fail your PR based on a threshold of the base's coverage: https://docs.codecov.io/docs/commit-status#section-threshold. If codecov is exceptionally valuable on large coverage swings, I'd suggest just disabling the PR comment altogether, adding a threshold status and then digging deeper on codecov when your `codecov/project` status fails due to missing the threshold. 
3. Is there anything of value for you in the comment for every PR? If there are parts of it that you find useful all the time, you could format the comment to only show the parts you find generally useful, with the `layout` parameter: https://docs.codecov.io/docs/pull-request-comments#section-layout

If I had to make a recommendation you all might find useful today, it would be to disable the PR comment altogether, and use codecov's threshold status checks so that PR's fail altogether when coverage is outside of a certain threshold. I *think* this would enompass your use case and remove the noisy PR comment from your PRs. Taking a rough stab at the codecov.yml you'd need, it'd be something like this:

**codecov.yml**

```
comment: off
coverage:
  status:
    project:
      default:
        target: auto
        threshold: 20% #this would only fail the PR if the coverage on the PR was 20% lower than base.
```

As an advanced feature, you could also add more status checks to split your code and have tighter restrictions on some areas than others. See: https://docs.codecov.io/docs/commit-status#section-splitting-up-projects-example-

Is any of this helpful? Perhaps not. If none of this seems like it can get Codecov to such a position where it is providing value to this repository, I totally understand turning it off. However, if this is the case, I would love any and all additional feedback you'd be willing to provide so we can understand your use case better.

peterbe:
@hootener I'm glad I reached out and I'm impressed with your response. Keep up the great work. 

First of all, I'm definitely not in favor of tying into the status checks. There are pull requests, often made by core contributors, that just know that decreased test coverage percentage is fine. For example, some temporary report or hack that is known to only be used by superusers, like, once or twice. 

So, let's keep with `comment:` in codecov.yaml and see if we can make it better for us. 

To start with, I think it'll be a huge improvement just to set `require_changes: true`. That'll avoid codecov comments when only a `.css` file has changed for example. 

However, I don't see how I can adjust a "threshold" percentage on https://docs.codecov.io/docs/pull-request-comments
But it's probably fine. 

What would be really cool is to have something like `require_changes: 'decrease'`. I.e. only post a comment if the test coverage drops. Because that's when you *really* need to wake up and scrutinize the PR and ask why is new code being added without tests?! @hootener Is this a feature that is currently lacking?

peterbe:
Let's [go here instead](https://github.com/mozilla/kuma/pull/5549)

PR #5549, opened 2019-07-11
bug 1565223 - tune codecov.yaml

peterbe:


hootener:
Answering some of your prior questions on the other PR

> What would be really cool is to have something like require_changes: 'decrease'. I.e. only post a comment if the test coverage drops. Because that's when you really need to wake up and scrutinize the PR and ask why is new code being added without tests?! @hootener Is this a feature that is currently lacking?

This currently isn't a feature. Although I can understand the utility of it. 

> First of all, I'm definitely not in favor of tying into the status checks. There are pull requests, often made by core contributors, that just know that decreased test coverage percentage is fine. For example, some temporary report or hack that is known to only be used by superusers, like, once or twice.


This is interesting. So you often expect PRs that greatly lower your coverage? Even below a known threshold? Is there any threshold level on an incoming PR where the test coverage is so low you absolutely wouldn't take it as is? Or is this always conditional?

peterbe:
> So you often expect PRs that greatly lower your coverage? 

Not really. I think it more stems from the fact that it happens and then you don't want it to block progress. Also, I think it used to be that you can't merge it if the checks don't pass which could be very annoying. But actually, I *just* now discovered that I, as an administrator, I *can* merge even though checks failed. 

This is what I see right above this comment ^

Screen Shot 2019-07-11 at 10 24 04 AM


That means, if we do make one those PRs, with sloppy Python code, that we know in our hearts, don't need test coverage we *can* still allow it though. With that in mind, we could definitely opt out of PR comments and tie into the checks instead.

peterbe:
@hootener Any idea why status checks yet? Does it matter than I edited the codecov.yaml after this PR was opened?

callahad:
> I think it used to be that you can't merge it if the checks don't pass which could be very annoying. But actually, I _just_ now discovered that I, as an administrator, I _can_ merge even though checks failed.

That's actually configurable in GitHub's [branch protection rules](https://github.com/mozilla/kuma/settings/branches), under "Require status checks to pass before merging."

You can even select which status checks are mandatory, e.g., Travis has to pass, but it's fine for CodeCov to fail.

callahad:

PR #5552, opened 2019-07-12
bug 1525719: remove history button on beta doc pages

escattone:
This PR removes the "History" button (clock icon) from the document pages on the beta domain.

Before:
image

After:
image

Note: This PR made me think that perhaps, within https://github.com/mozilla/kuma/pull/5548, I should have only added the `Add a translation` link within the language menu dropdown if the user was logged in and a contributor (the same logic used here for deciding whether or not to display the `Edit` button). I suppose so. What do you guys and @atopal and @nickbrandt think?

### DIFF
--- a/kuma/javascript/src/titlebar.jsx
+++ b/kuma/javascript/src/titlebar.jsx
@@ -5,7 +5,6 @@ import { css } from '@emotion/core';
 
 import { gettext } from './l10n.js';
 import EditIcon from './icons/pencil.svg';
-import HistoryIcon from './icons/clock.svg';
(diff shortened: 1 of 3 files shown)


schalkneethling:
🍒 Thanks @escattone - Only one thing, you can now remove the CSS here:
https://github.com/mozilla/kuma/blob/master/kuma/static/styles/base/icons.scss#L270

atopal:
I think it's okay to show this to everyone, Ryan.

escattone:
Thanks for the review and good catch @schalkneethling! I'll remove that CSS.

PR #5554, opened 2019-07-12
bug 1565618 - 'last updated by' date localized

peterbe:
On a `en-US` document: 
Screen Shot 2019-07-12 at 1 44 12 PM

On a `sv-SE` document:
Screen Shot 2019-07-12 at 1 44 27 PM

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -191,6 +191,7 @@ function ArticleMetadata({ document }: DocumentProps) {
                 lastModifiedBy={document.lastModifiedBy}
                 lastModified={document.lastModified}
                 profileBaseURL={profileBaseURL}
+                documentLocale={document.locale}
             />
         </div>
     );
--- a/kuma/javascript/src/last-modified-by.jsx
+++ b/kuma/javascript/src/last-modified-by.jsx
@@ -6,16 +6,30 @@ import { gettext } from './l10n.js';
 import ClockIcon from './icons/clock.svg';
 
 type LastModifiedByProps = {
+    documentLocale: String,
     lastModifiedBy: String,
     lastModified: String,
     profileBaseURL: String
 };
 
 export default function LastModifiedBy({
+    documentLocale,
     lastModifiedBy,
     lastModified,
     profileBaseURL
 }: LastModifiedByProps) {
+
+    // This fortunately works because the 'lastModified' date string is
+    // predictable and always of the same format. It's not a proper ISO
+    // string but it's close.
+    const lastModifiedDate = new Date(lastModified);
+    // Justification for these is to match the Wiki.
+    const dateStringOptions = {
+        year: 'numeric',
+        month: 'short',
+        day: 'numeric'
+    };
+
     return (
         <section className="contributors-sub">
             <ClockIcon />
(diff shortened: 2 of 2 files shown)


peterbe:
Note! I don't think Node supports `Date.toLocaleString()` because this is what you get in the SSR rendered HTML (using `sv-SE`): 
Screen Shot 2019-07-12 at 1 57 38 PM
As you can see the label prefix gets translated correctly but it falls back to the en-US formatting. 

So, it means that React will re-render this component upon load. Since it's a leaf node I don't think it causes a major repaint.

kuma/javascript/src/article.jsx
```diff
@@ -191,6 +191,7 @@ function ArticleMetadata({ document }: DocumentProps) {
                 lastModifiedBy={document.lastModifiedBy}
                 lastModified={document.lastModified}
                 profileBaseURL={profileBaseURL}
+                documentLocale={document.locale}
```
davidflanagan: Note that the document locale does not always match the user's locale when viewing untranslated documents. If you want his date to be treated as part of the "chrome" rather part of the article, consider using l10n.getLocale() here instead.

davidflanagan:
@peterbe: Perhaps the failure to localize the date on SSR has something to do with what is installed in our kuma-base image. Presumably we need some kind of locale data for each locale, and if the docker image is bare-bones maybe the data is just missing. I don't know how Node does this, but I'd guess that it gets data from the OS. So maybe we need to install an additional package into the container to get the right localization?

peterbe:
> So maybe we need to install an additional package into the container to get the right localization?

Mayhaps. I'm not too worried since the rerender after hydration is at a leaf node.

kuma/javascript/src/article.jsx
peterbe: We should do that! 
We've already seen that if you view a Swedish page that hasn't been translate the "chrome" is in Swedish. E.g. "Senast uppdaterad:" instead of "Last updated:". So it doesn't make sense to say "Senast uppdaterad: Oct 12, 2019" since the swedes would have expected that to be "Senast uppdaterad: 12 okt. 2019"

peterbe:
```
> date = new Date(Date.UTC(2019, 9, 12, 3, 0, 0));
Date Fri Oct 11 2019 23:00:00 GMT-0400 (Eastern Daylight Time)
> date.toLocaleString('sv-SE', {year: 'numeric', month: 'short', day: 'numeric'})
"11 okt. 2019"
```

peterbe: ^ That's from the Web Console here in Firefox.

peterbe: Actually, how do you even get to pages where the "chrome" is in, say, Swedish but the content is still in English?
For example, the "Tools/Eyedropper" page hasn't been translated to Swedish but you can manually mess with the URL to become: https://developer.mozilla.org/sv-SE/docs/Tools/Eyedropper
But that doesn't exist as a link. Kumascript and stuff makes sure that if you're o https://developer.mozilla.org/sv-SE/docs/Tools the link to "Eyedropper" is only available in English. 

So for the rare few people who mess with the URL and trigger page renders where the chrome language != content language, I think these people can put up with the date formatting not being perfect.

peterbe: I.e. I'm not going to bother fixing it to use `l10n.getLocale()`.

PR #5555, opened 2019-07-12
bug 1564277: use default USE_X_FORWARDED_HOST setting (False)

escattone:
Currently, we're setting the Django `USE_X_FORWARDED_HOST` setting to `True`. This is a relic of the past. This PR changes that Django setting to simply use the default value, which is `False`.


escattone:
@peterbe @limed By the way, this has been tested on stage and eliminates the issue raised by bug 1564277.

PR #5556, opened 2019-07-12
bug 1563210 - visually distinguish between current and obsolete

peterbe:
I'm clearly out of my depth here on the scss and stuff...

Basically, the wiki site relies on the fact that the `<head>` is only rendered once per page. That's why it conditionally inserts (or omits) a `<link rel=stylesheet href=archive.css>` into the DOM. That archive.css then overrides `.wiki-main-content`. 

This approach is not going to work with the React site because this kinda of business logic needs to depend not only on the initial server HTML rendering (aka. SSR) but also when you jump between pages. E.g. you might be on a non-archive page and click into an archive page and then click out to another non-archive page. 

So I changed approach. The `<article id="wikiArticle">` tag now becomes `<article id="wikiArticle" class="archive-content">` dynamically. *And*, now the CSS selector `.archive-content { ... }` is included on every page but used or not used dynamically. Yes, this bloats the total CSS even more but I can't think of a more convenient way. 

However, we COULD use emotion instead! The reason I didn't want to do that was because I didn't want to have to manually port that snippet of SCSS to emotion. Partially because I don't know how and partially because it's a form of DRY since you'd *have* to define that gradient color thing in two different places. 

The reason this PR isn't ready is because it doesn't work right. Here's what it looks like:
Screen Shot 2019-07-12 at 3 10 19 PM
For some reason, that thing has a negative margin which hides the color striped (red and orange) border.

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -117,20 +117,19 @@ export default function Article({ document }: DocumentProps) {
         }
     }, [document, userData]);
 
+    const isArchive =
+        document.slug === 'Archive' || document.slug.startsWith('Archive/');
+
     return (
         /*
          * The "text-content" class and "wikiArticle" id are required
          * because our stylesheets expect them and formatting isn't quite
          * right without them.
          */
-        <div
-            id="content"
-            ref={article}
-            className="text-content"
-            css={styles.article}
-        >
+        <div id="content" ref={article} css={styles.article}>
             <article
                 id="wikiArticle"
+                className={isArchive ? 'archive-content' : null}
                 dangerouslySetInnerHTML={{ __html: document.bodyHTML }}
             />
             <ArticleMetadata document={document} />
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -126,10 +126,16 @@ export default function Article({ document }: DocumentProps) {
          * because our stylesheets expect them and formatting isn't quite
          * right without them.
          */
-        <div id="content" ref={article} css={styles.article}>
+        <div
+            id="content"
+            ref={article}
+            className={
+                isArchive ? 'text-content archive-content' : 'text-content'
+            }
+            css={styles.article}
+        >
             <article
                 id="wikiArticle"
-                className={isArchive ? 'archive-content' : null}
                 dangerouslySetInnerHTML={{ __html: document.bodyHTML }}
             />
             <ArticleMetadata document={document} />
--- a/kuma/static/styles/components/structure/archive.scss
+++ b/kuma/static/styles/components/structure/archive.scss
@@ -1,15 +1,9 @@
-@import 'includes/vars';
-
 /*
-Archived pages
+For articles that are archive content. Makes them stand out more.
 ********************************************************************** */
 
-/*
-Red background & stripped border for archived pages
-- intentionally odd
-********************************************************************** */
 
-.wiki-main-content {
+.archive-content {
     background-color: $red-light;
     border: 10px solid $red;
     border-image: repeating-linear-gradient(-45deg, $red 0, $red 5px, $yellow 5px, $yellow 10px, $red 10px) 10 10 repeat;
@@ -19,7 +13,7 @@ Red background & stripped border for archived pages
 }
 
 @media #{$mq-tablet-and-up} {
-    .wiki-main-content {
+    .archive-content {
         border-width: 20px;
         border-image: repeating-linear-gradient(-45deg, $red 0, $red 10px, $yellow 10px, $yellow 20px, $red 20px) 20 20 repeat;
         // negative margins to make border flush to outside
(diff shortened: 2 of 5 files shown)


peterbe:
By the way, it took me a while to appreciate why my edits didn't take effect. I hadn't run `docker-compose restart ssr`.

kuma/javascript/src/article.jsx
```diff
@@ -144,20 +144,19 @@ export default function Article({ document }: DocumentProps) {
...
     return (
         /*
          * The "text-content" class and "wikiArticle" id are required
          * because our stylesheets expect them and formatting isn't quite
          * right without them.
          */
-        <div
-            id="content"
-            ref={article}
-            className="text-content"
```
davidflanagan: I think that taking out text-content will break stuff. I don't remember what. But I put the comment above to make sure that this wouldn't be removed.

kuma/javascript/src/article.jsx
```diff
@@ -144,20 +144,19 @@ export default function Article({ document }: DocumentProps) {
...
-            css={styles.article}
-        >
+        <div id="content" ref={article} css={styles.article}>
             <article
                 id="wikiArticle"
+                className={isArchive ? 'archive-content' : null}
```
davidflanagan: Given that the set of styles is pretty small for archived pages, how about just copying those styles here with emotion rather than messing around with the existing stylesheets?

davidflanagan: That way  you could fix this bug without having to change anything for the wiki site, which would be safer.

schalkneethling: I am a little hesitant to move CSS to JS to quickly. I would prefer to keep this in Sass.

kuma/javascript/src/article.jsx
peterbe: It's been days since this goldfish-brain made this edits but it's definitely wrong that I didn't also remove the code comment. I need to revisit this and question "Why did I remove the className?!?!"

kuma/javascript/src/article.jsx
peterbe: This is a big topic that mixes personal preferences together with the fact that we still have the wiki. 
By doing it all from 1 single .scss file we avoid DRY. Doing it all in SCSS means you can reuse variables for things like colors and media query tricks. 
Pragmatically I think emotion would be a better home because it should be easy peasy to tie it into the business logic conditional. 

How much other CSS things have been copy-n-pasted over from SCSS files into emotion in JS? 

Also, god forbid we decide to change the color from orange to maroon one day and we forget our DRY violation of July 2019. Then, how big a deal is it that it looks different in the wiki compared to the read-only site. Do the 1% care?

kuma/static/styles/components/structure/archive.scss
```diff
@@ -1,15 +1,9 @@
...
-/*
-Red background & stripped border for archived pages
-- intentionally odd
-********************************************************************** */
 
-.wiki-main-content {
+.archive-content {
     background-color: $red-light;
     border: 10px solid $red;
     border-image: repeating-linear-gradient(-45deg, $red 0, $red 5px, $yellow 5px, $yellow 10px, $red 10px) 10 10 repeat;
```
schalkneethling: ~~I would add a bottom margin of 20px to add some space between the content block and the article metadata.~~

Before

Screenshot 2019-07-19 at 08 47 35

After

Screenshot 2019-07-19 at 08 47 42

schalkneethling: ~~To get the bottom border to match the width of the rest we need to override the default 3px bottom border set on `#wikiArticle`. Using ids for styling is bad, but we are in this hot soup already ;p, you can create a higher specificity selector like so:~~

```
#wikiArticle.archive-content {
  border-bottom-width: 20px;
}
```

schalkneethling: Actually, scratch all of that. Move the `archive-content` class to the `div` with the `id` content, and remove the negative margins for the archive styling.

```
.archive-content {
    background-color: $red-light;
    border: 10px solid $red;
    border-image: repeating-linear-gradient(-45deg, $red 0, $red 5px, $yellow 5px, $yellow 10px, $red 10px) 10 10 repeat;
    padding: $mobile-center-spacing;
}

@media #{$mq-tablet-and-up} {
    .archive-content {
        border-width: 20px;
        border-image: repeating-linear-gradient(-45deg, $red 0, $red 10px, $yellow 10px, $yellow 20px, $red 20px) 20 20 repeat;
        padding: $gutter-width;
    }
}
```

kuma/javascript/src/article.jsx
```diff
@@ -144,20 +144,19 @@ export default function Article({ document }: DocumentProps) {
...
          * because our stylesheets expect them and formatting isn't quite
          * right without them.
          */
-        <div
-            id="content"
-            ref={article}
-            className="text-content"
-            css={styles.article}
-        >
+        <div id="content" ref={article} css={styles.article}>
```
schalkneethling: Move the `archive-content` class here.

peterbe:
here's what it looks like after your suggested changes. 
Screen Shot 2019-07-22 at 9 49 01 AM

and the wiki:
Screen Shot 2019-07-22 at 9 49 19 AM

kuma/javascript/src/article.jsx
peterbe: @schalkneethling In my latest push I took David's advice and made it so that the `text-content` className is always there.

PR #5557, opened 2019-07-12
bug 1562263 - Run flow check in CI

peterbe:
tl;dr; This way, `flow` is executed using the host, in CI. I.e. TravisCI's own Node environment. Pretty gross but it's the right thing to do. 

Flow goes away when the .jsx files are turned into .js files that are shipped to AWS. So, if it passes CI, it's no longer useful. It's the same story as linting. EC2 doesn't care about whitespace indentation. 

It's nuts that our Dockerfile says "We use Node version X.Y.Z" but then we CI check Flow with TravisCI's version. Which might be different. 
Flow isn't just "linting" but exists to take some of the burden that unit testing does. Basically, another deliberate inertia to slow down making releases (in a good way!). That's why I'm OK with using the "wrong Node environment" for executing Flow.


schalkneethling:
Thanks for the PR @peterbe - There are a bunch of errors being thrown related to `Cannot resolve module `@emotion/core`.`, `import styled from '@emotion/styled'`, and `import { create } from 'react-test-renderer'`

peterbe:
PS. @escattone and @schalkneethling I don't think this PR requires approval from both of you but it would be nice to have awareness of the issue from both of you.

escattone:
Thanks for getting this into our CI @peterbe!

PR #5560, opened 2019-07-15
Bug 1563467, ensure main navigation is keyboard accessible

schalkneethling:
@peterbe I am not completely done with this but, this is a big step forward and allows us to go to public beta. r?

I will keep iterating on this once this one has landed. Thanks!


schalkneethling:
@peterbe Not ready.  Still some more Flow problems, and issues related to client-side navigation. Will update once those are sorted out.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+        );
+        if (visibleSubMenu) {
+            visibleSubMenu.setAttribute('aria-hidden', 'true');
+            visibleSubMenu.classList.toggle('show');
+        }
+    }
+
+    useEffect(() => {
+        let menu = document.getElementById('header-main-nav');
+        var mediaQuery = window.matchMedia('(min-width: 47.9375em)');
```
peterbe: Let's not use `var`. But more importantly, `jest` doesn't know what `window` is. 
Also, why the 4 significant figures on the `em` there?

schalkneethling: Whoops, copy/paste fail. With regards to the value for `em`, that is `(767 / 16) * 1` which is our end-of-mobile breakpoint.

schalkneethling:
Oh, also realized I need to update the existing tests related to the navigation.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+            'li ul[aria-hidden="false"]'
+        );
+        if (visibleSubMenu) {
+            visibleSubMenu.setAttribute('aria-hidden', 'true');
+            visibleSubMenu.classList.toggle('show');
+        }
+    }
+
+    useEffect(() => {
+        let menu = document.getElementById('header-main-nav');
```
davidflanagan: defining a ref and using the useRef().current hook is probably the more react-friendly way to do this instead of getElementById().  I'm not sure precisely what difference it makes, though.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+        var mediaQuery = window.matchMedia('(min-width: 47.9375em)');
+        if (menu) {
+            if (mediaQuery.matches) {
+                desktopInteractionHandlers(menu);
```
davidflanagan: I'm pretty sure that this code is going to end up registering the same event handlers multiple times. Each time the menu is rendered, this effect function is going to run, but that doesn't mean that there will be a new menu element each time, so you're likely to end up with a memory leak here (and also below for the touch event handlers).

One approach would be to define the required event handlers directly in the rendering code, rather than adding them on with this effect.

The other approach would be to write corresponding functions for removing the event handlers.  If useEffect() returns a function  then that function will be used as the cleanup function. If you do that then the cleanup function will always be called before the effect function is called again.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+            } else {
+                /* Because the header is not rerendered during client-side
+                   navigation, we call this function to ensure the currrently
+                   visible submenu is hidden when a link it triggered */
+                hideSubMenus();
```
davidflanagan: Thank you for fixing this. I was looking into it myself and realized that the menus were badly broken on touch devices.   (It is so hard to do real touch device testing with the devserver...)

peterbe: Closing the menus based on navigation should be achieved by passing in the document location from the parent component.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+    // the current mdnDocument path into. Compute that now.
+    let path = encodeURIComponent(
+        `/${locale}` + (mdnDocument ? `/docs/${mdnDocument.slug}` : '')
+    );
+
+    return (
+        <nav id="header-main-nav" className="main-nav" role="navigation">
+            <ul>
+                {menus.map((menuEntry, index) => (
+                    <li key={index} className="top-level-entry-container">
```
peterbe: This makes me nervous. The `key` just needs to be something unique. What's guaranteed to be unique is the `menuEntry.url`.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+                                className="main-menu-arrow"
+                                aria-hidden="true"
+                            >
+                                ▼
+                            </span>
+                        </button>
+                        <ul aria-hidden="true">
+                            {menuEntry.items.map((item, index) => (
+                                <li
+                                    key={index}
```
peterbe: ditto

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,266 @@
...
+    });
+
+    // One of the menu items has a URL that we need to substitute
+    // the current mdnDocument path into. Compute that now.
+    let path = encodeURIComponent(
+        `/${locale}` + (mdnDocument ? `/docs/${mdnDocument.slug}` : '')
+    );
+
+    return (
+        <nav id="header-main-nav" className="main-nav" role="navigation">
```
peterbe: When we switch to synthetic events, we won't need the `id=` attribute I don't think.

kuma/javascript/src/header/main-menu.jsx
peterbe: Grr.. My bad. An `menuEntry` doesn't have a `.url`. That's what `menuEntry[0]` has. So this line should instead be `key={menuEntry.label}`.

schalkneethling:
@peterbe Ok, I am reworking the event bits. Thanks for the feedback.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    useEffect(() => {
+        const mediaQuery = window.matchMedia('(min-width: 47.9375em)');
+
+        if (!mediaQuery.matches) {
+            hideVisibleSubMenu();
+        }
+    }, [mdnDocument]);
```
schalkneethling: @peterbe This is the only portion I still have in the `useEffect`. I believe you have another option for how we can ensure the submenu closes upon navigation?

MarcoZehe:
I am very worried about the extensive use of aria-hidden. As [the spec notes](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden), it should be used with extreme caution and only if there is visually visible content that is intended to be hidden from assistive technologies to improve the user experience.

But if something gets hidden via CSS anyway, like with display:none; or visibility:hidden;, there is absolutely no need, and I mean *no* need, to use aria-hidden at all. Layout and the Accessibility engine will perfectly handle the hiding and showing of elements and their children if styled via CSS. So if the sub menu gets shown and hidden via CSS anyway, all the extra checks for aria-hidden can safely be removed without causing any damage.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,254 @@
+//@flow
+import * as React from 'react';
+import { useMemo, useEffect } from 'react';
+
+import { getLocale, gettext } from '../l10n.js';
+
+export default function MainMenu(mdnDocument: Object) {
```
schalkneethling: @peterbe Switched back to `Object` here as using `DocumentData` causes a ton of Flow exceptions at [this line](https://github.com/mozilla/kuma/pull/5560/files#diff-6b097a971a343a594f429782869d3df7R99). The reason those exceptions happen is that Flow treats all of the properties of `DocumentData` as mandatory. Therefore we would have to mark all properties as optional in order to use it here. If all properties are not actually optional, we then lose valuable checking from Flow. Perhaps I am overlooking something else, but that is what it seems like to me.

schalkneethling:
> I am very worried about the extensive use of aria-hidden. As [the spec notes](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden), it should be used with extreme caution and only if there is visually visible content that is intended to be hidden from assistive technologies to improve the user experience.
> 
> But if something gets hidden via CSS anyway, like with display:none; or visibility:hidden;, there is absolutely no need, and I mean _no_ need, to use aria-hidden at all. Layout and the Accessibility engine will perfectly handle the hiding and showing of elements and their children if styled via CSS. So if the sub menu gets shown and hidden via CSS anyway, all the extra checks for aria-hidden can safely be removed without causing any damage.

Thanks @MarcoZehe, I was not aware that the Accessibility engine will handle these cases so well. There are then quite a few places across MDN that we should not be using `aria-hidden`. Will update this PR.

schalkneethling:
@peterbe When testing keyboard functionality, either use Chrome or switch on the preference in macOS settings:

Before

Screenshot 2019-07-18 at 16 03 02


After

Screenshot 2019-07-18 at 16 03 10

schalkneethling:
@peterbe This is ready for re-review

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,229 @@
...
+
+        /* currentTarget instanceof 'HTMLButtonElement' is added
+           to keep Flow happy: https://github.com/facebook/flow/issues/218#issuecomment-74119319 */
+        if (
+            (!mediaQuery.matches,
+            event.type === 'touchstart' &&
+                currentTarget instanceof HTMLButtonElement &&
+                currentTarget.classList.contains('top-level-entry'))
+        ) {
+            if (currentTarget instanceof HTMLButtonElement) {
```
schalkneethling: Without this, Flow keeps inferring an error as it sees the variable as an `EventTarget` and not an HTML element. The fact that I check for this in the out conditional already seems to make no difference.

If I use `event.currentTarget` on line 127, it is happy but, that will always be the `nav` element that wraps the entire menu. So this is the only way I can keep flow happy.

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+// To avoid problems with flow and React.memo(), define the component
+// in this plain way first. See bottom of file for the final memo() and export.
+const _MainMenu = ({ document, locale }: Props) => {
+    // The CSS that supports this is sufficiently smart to understand
+    // hovering over the top-level menu items and stuff. But for mobile,
+    // there's no mouse hover so for that we use a piece of state to
+    // record onTouchStart events.
+    const [showSubMenu, setShowSubMenu] = useState(null);
+    // console.log('RENDERING _MainMenu', showSubMenu);
```
peterbe: Oops.

peterbe:
```suggestion
```

schalkneethling:
Merged o/\o ~ Thanks for all the help @peterbe 🥃

PR #5562, opened 2019-07-15
Bug 1564405: don't display long, unwrapped excerpts

davidflanagan:
This PR changes our search results page so that it ignores
result excerpts that are too long and so that it forces line
breaks even when excerpts don't have spaces. (There seems to
have been some kind of ElasticSearch change that made it start
sending us long excerpts without spaces.)

I'm sure there is still work to be done on the search results
page, but this fixes the particularly ugly problem.


peterbe:
I might take a "bigger stab" at attempting to solve the site search quite differently. So I'm delaying review of this.

PR #5563, opened 2019-07-16
Bug 1563481, toc component should be higher in source order

schalkneethling:
This ensures that the TOC component is in a more appropriate place in the DOM source order. As a result some `grid` styles needed updating as well. @callahad r?

### DIFF
--- a/kuma/javascript/src/document.jsx
+++ b/kuma/javascript/src/document.jsx
@@ -11,6 +11,7 @@ import Header from './header/header.jsx';
 import Route from './route.js';
 import TaskCompletionSurvey from './task-completion-survey.jsx';
 import Titlebar from './titlebar.jsx';
+import TOC from './toc.jsx';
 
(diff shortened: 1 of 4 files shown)


kuma/javascript/src/document.jsx
```diff
@@ -97,12 +98,13 @@ const styles = {
         maxWidth: 1400,
         margin: '0 auto',
         gridTemplateColumns: '25% 75%',
-        gridTemplateAreas: '"side main"',
+        grinTemplateRows: 'minmax(200px, 1fr) 75%',
```
MarcoZehe: Is this a typo? "grin" instead of "grid"?

schalkneethling: Oh boy 😄, it is indeed a typo. Thanks @MarcoZehe

peterbe:
Let's merge this if CI passes.

PR #5569, opened 2019-07-18
bug 1566302: sanitize name attr when override for id

escattone:
This PR addresses a loophole in the code that's used to inject section `id`'s as part of the `clean_content` function,  which is run after every render as well as when revisions are saved. Currently, any HTML elements which qualify as "section" tags (see https://github.com/mozilla/kuma/blob/c30762afdebfdf9bfe50077f526188073ffdefd5/kuma/wiki/content.py#L44) are automatically assigned an `id`. For example, this:
```html
<h1>the header text</h1>
```
will become this:
```html
<h1 id="the_header_text">the header text</h1>
```

This automatic selection of an `id` can be manually overridden by providing a `name` attribute, whose value will then be used in place of what would have been automatically assigned. For example, this:
```html
<h1 name="use_me_as_the_id">the header text</h1>
```
will become this:
```html
<h1 id="use_me_as_the_id" name="use_me_as_the_id">the header text</h1>
```

The problem is that when a name attribute is provided, its value is used without considering whether it contains troublesome characters (like spaces for example) that could cause problems when used as an `id` attribute intended for use in URL fragments (to jump to the section via the URL). So, for example, this:
```html
<h3 name="clear:_left">clear: left</h3>
```
currently becomes this:
```html
<h3 id="clear:_left" name="clear:_left">clear: left</h3>
```
and the colon in the `id` can lead to problems down the road like an invalid CSS selector (see https://bugzilla.mozilla.org/show_bug.cgi?id=1566302#c2).

This PR closes that loophole by sanitizing the value of the `name` attribute before assigning it as the value of the `id` attribute. So, taking the case immediately above, the new result would be:
```html
<h3 id="clear_left" name="clear:_left">clear: left</h3>
```

For local testing, I used these settings in my `.env` file (my usual settings):
```
DEBUG=True
DOMAIN=mdn.localhost
ATTACHMENT_HOST=demos:8000
```
```
##
# Host Database
#
##
255.255.255.255	broadcasthost
::1             localhost
```
and did something like this:
```
docker-compose pull
docker-compose up -d
docker-compose restart
docker-compose exec web ./manage.py scrape_document --revisions 5 https://developer.mozilla.org/en-US/docs/Web/CSS/clear
```

Then revert the document to one of the problematic revisions (one of the revisions before the most recent two revisions by `schalkneethling`), visit http://beta.mdn.localhost:8000/en-US/docs/Web/CSS/clear to see that React fails (most of the document is missing), log in, visit http://mdn.localhost:8000/en-US/docs/Web/CSS/clear, shift-reload to re-render, and then go back to http://beta.mdn.localhost:8000/en-US/docs/Web/CSS/clear and see that it loads fine this time.

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -681,7 +681,11 @@ class SectionIDFilter(html5lib_Filter):
                 # treat it as a manual override by the author and make
                 # that value be the ID.
                 if (None, 'name') in attrs:
-                    attrs[(None, u'id')] = attrs[(None, 'name')]
+                    # Sanitize the "name" attribute with self.slugify to
+                    # prevent the injection of spaces (which are illegal
+                    # for the "id" attribute) or any of the non-URL-safe
+                    # characters listed above.
+                    attrs[(None, u'id')] = self.slugify(attrs[(None, 'name')])
                     token['data'] = attrs
                     yield token
                     continue
--- a/kuma/wiki/tests/test_content.py
+++ b/kuma/wiki/tests/test_content.py
@@ -114,9 +114,10 @@ class InjectSectionIDsTests(TestCase):
             <p>test</p>
             <section>
                 <h1 class="header2">Header Two</h1>
+                <h1 name="Header: X" class="header3">Header Three</h1>
                 <p>test</p>
             </section>
-            <h2 name="Constants" class="hasname">This title does not match the name</h2>
+            <h2 name="C~o:n;s/t%a$n=t@s" class="hasname">This is ignored</h2>
             <p>test</p>
 
             <h1 id="i-already-have-an-id" class="hasid">This text clobbers the ID</h1>
@@ -140,6 +141,7 @@ class InjectSectionIDsTests(TestCase):
         expected = (
             ('header1', 'Header_One'),
             ('header2', 'Header_Two'),
+            ('header3', 'Header_X'),
             ('hasname', 'Constants'),
             ('hasid', 'This_text_clobbers_the_ID'),
             ('Quick_Links', 'Quick_Links'),


escattone:
@wbamberg Sorry to bother you, but I added you as a reviewer since you're an expert on MDN content and may see unintended consequences that we wouldn't.

kuma/wiki/content.py
```diff
@@ -681,7 +681,11 @@ def __iter__(self):
                 # treat it as a manual override by the author and make
                 # that value be the ID.
                 if (None, 'name') in attrs:
-                    attrs[(None, u'id')] = attrs[(None, 'name')]
+                    # Sanitize the "name" attribute with self.slugify to
+                    # prevent the injection of spaces (which are illegal
+                    # for the "id" attribute) or any of the non-URL-safe
+                    # characters listed above.
+                    attrs[(None, u'id')] = self.slugify(attrs[(None, 'name')])
```
schalkneethling: If we always dynamically get the value of `name` at runtime, this should be safe.

peterbe:
Hmm. Perhaps it's unrelated to this PR. 
Input before I hit publish:
Screen Shot 2019-07-18 at 8 21 08 AM

After I hit publish someone somewhere first pre-processes this input into `Document.html` and it becomes this:
```
>>> print Document.objects.get(locale='en-US', slug='Web/CSS/clear').html[:131]
<div>{{CSSRef}}</div>


<h3 id="MYNAME" name="MYNAME">This is my name</h3>

<h3 id="MYNAME" name="MYNAME">This is also my name</h3>
```
I.e. that `name` --> `id` happens on the `html`, not on the `rendered_html`.

peterbe:
I think I get it now. 
I *typed* this in:
```
<h3 id="MYNAME">This is my name</h3>

<h3 id="MYNAME">This is also my name</h3>
```
and the pre-processing turned that into:
```
<h3 id="This_is_my_name">This is my name</h3>

<h3 id="This_is_also_my_name">This is also my name</h3>
```

I guess, as you mentioned above, if the `name` is present it gets used as the ID "no questions asked" except the new difference, as of this PR, is that when it does that it now sneaks in a slugify transform so that the ID (made from the `name`) becomes safe enough.

peterbe:
@escattone I think this is safe to merge but I'd love to let you ponder my ramblings a little bit before just in case you have some immediate clever ideas.

wbamberg:
I can't think of a case where this would cause a problem but of course since this override "feature" is exposed to Wiki authors there might well be things that are assuming something about overridden `id` values.

The only concrete thing I can think of is live samples, where someone might have used `name` to override `id` and then use the overridden value in EmbedLiveSample. But sanitizing `name` would only break this if the name was already invalid, and then EmbedLiveSample wouldn't have worked anyway (probably: unless the sanitizing is more rigorous than it needs to be, so even valid CSS `id` values could be changed in this process).

But in general it seems that things like that might be where problems could occur. But then again, usages like this are so fragile that they are probably doomed to be broken anyway.

I think it's worth asking some other content team members if they can think of any obscure uses of this that might be broken by the change: @a2sheppy , @chrisdavidmills , @Elchi3 .

peterbe:
> usages like this are so fragile that they are probably doomed to be broken anyway.

😂 Truth!

a2sheppy:
This shouldn't be a problem. To be honest, I thought the ability to use `name` in this way was *removed* years ago by lorchard because it was violating standards in a way that was breaking something or other, so I for one haven't used it since then despite sometimes having a use for it.

PR #5570, opened 2019-07-18
Bug 1566429, prevent search input from expanding beyond viewport on iOS 10

schalkneethling:
This fixes the problem of the search input in the header expanding beyond the visible viewport in Safari on iOS 10. Tested on Firefox, Chrome using responsive mode, as well as on iPhone 7 using Safari running iOS 10.

Screenshot 2019-07-18 at 11 30 39

### DIFF
--- a/kuma/javascript/src/header/search.jsx
+++ b/kuma/javascript/src/header/search.jsx
@@ -32,6 +32,7 @@ const styles = {
         fontSize: 14,
         flex: '1 1 100px',
         minWidth: 60,
+        maxWidth: 320,
         color: '#000',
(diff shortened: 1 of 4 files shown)


kuma/javascript/src/header/search.jsx
```diff
@@ -32,6 +32,7 @@ const styles = {
         fontSize: 14,
         flex: '1 1 100px',
         minWidth: 60,
+        maxWidth: 320,
```
peterbe: Kinda wish there was a code comment here that explains this choice.

PR #5571, opened 2019-07-18
1542817 two different per post message handlerjs

peterbe:
Before this, we used to inject on every document page (react or wiki):
https://gist.github.com/peterbe/3092e982de1e053293e1e5e8bfe614f9
It's not huge (1.5KB uncompressed) but it's **blocking**. Page rendering has to pause when it reaches that `<script>` tag and switch to the javascript engine's main thread. Then it can return to the DOM rendering. 

Sometimes, that's a good thing because you genuinely don't want the DOM to be something unless a piece of JavaScript gets a chance to change the DOM (including event handlers). 

But, I think in this case, it's *not* a good case. 

Ultimately, what that script does is that it sets up [this](https://github.com/peterbe/kuma/blob/95357b746e54c01fda20bc055e5a8da80e359dd8/kuma/static/js/utils/perf-post-message-handler.js#L63):
```
window.addEventListener('message', perfMsgHandler, false);
```

That function (`perfMsgHandler`) adds itself to receive post messages from iframes. And it rejects all messages that are NOT from `interactive-examples.mdn.mozilla.net`. Those that are, it takes those messages and sends them to [Google Analytics's User Timings](https://analytics.google.com/analytics/web/#/report/content-site-speed-user-timings/a36116321w64965862p66726481/userTimingsExplorer-table.plotKeys=%5B%5D&_r.drilldown=analytics.userTimingCategory:RUM%20-%20Interactive%20Examples):
```javascript
window.ga('send', {
        hitType: 'timing',
        timingCategory: 'RUM - Interactive Examples',
        timingVar: perfData.measureName,
        timingValue:
            window.mdn.perf.getDuration(perfData.measureName) || 0
    });
```

So what does it mean? When switching to an external script, that is `async`, the browser will no longer block on rendering. Great. A worker thread will spawn to start downloading `/static/build/js/perf.654b849a6fd9.js` and parse and even execute it. 
**But** what can totally happen is that that async work on `perf.654b849a6fd9.js` takes a while. What *can* happen is that the `<iframe src="https://interactive-examples.mdn.mozilla.net/pages/tabbed/video.html?origin=http://mdn.localhost:8000">` tag loads sooner and inside that iframe the JavaScript in `https://interactive-examples.mdn.mozilla.net/pages/tabbed/video.html` starts sending messages to the opener and these messages aren't picked up by any handler. Bu huu! But, who cares?!
This event (no pun intended) is unlikely, because I think the external script will execute faster than the iframe's JavaScript most of the time, and most importantly; the worst that can happen is that we miss some Real User Measurements sent to GA. We have this just to get an insight into user timings of the iframes. But who even looks at that regularly?


peterbe:
By the way, I'm too lazy to find the exact source but I know that recent versions of Chrome at least will actually not really block rendering just because it encounters a piece of blocking JavaScript. It makes a bet that it won't matter and parallelizes the work so that it can continue with the DOM rendering whilst the javascript engine does its thing. 
I wouldn't be surprised that Firefox does something that like that too.

peterbe:
Best of all; ability to delete 2 `.min.js` files checked into git!

peterbe:
More fuel for why I think it's good to make that JavaScript not inline and blocking...

Many pages don't have interactive examples in an iframe. Those poor pages had to be slowed down rather "clunkily" just in case.

schalkneethling:
> the worst that can happen is that we miss some Real User Measurements sent to GA. We have this just to get an insight into user timings of the iframes. But who even looks at that regularly?

Now that we have SpeecurveLUX, the need to do this via GA is going away anyhow. So we are all good.

PR #5572, opened 2019-07-18
Bug 1563476, remove tags display from read-only site

schalkneethling:
As per this comment[https://bugzilla.mozilla.org/show_bug.cgi?id=1563476#c5] on Bugzilla, removing tags from the read-only site.


kuma/javascript/src/document.jsx
```diff
@@ -360,19 +359,6 @@ export class DocumentRoute extends Route<DocumentRouteParams, DocumentData> {
...
-                    // So if we receive data with no tags property, we
-                    // have to add one set to an empty array for backward
-                    // compatibility.
-                    //
-                    // NOTE: as a general matter, because of our S3
-                    // cache we need to do this every time we update
-                    // the doc API to ensure backward compatibility.
-                    if (!documentData.tags) {
-                        documentData.tags = [];
-                    }
```
peterbe: What about removing tags from the Document JSON API?
It's an API and it has a version so perhaps that'll be too big a task. But, who else is using the Document JSON API? Some scrapers maybe?

schalkneethling: @escattone Thoughts?

peterbe: Let's move this discussion to: https://bugzilla.mozilla.org/show_bug.cgi?id=1563476#c7

PR #5574, opened 2019-07-19
bug 1567482 - no tags in read-only site search results

peterbe:
I first edited `search-results-page.jsx` and removed all mention of tags etc. Then I ran the jest tests and everything passed. Kinda sad. Clearly something there is overly mocking or we were just not testing tags in any way.

### DIFF
--- a/kuma/javascript/src/search-results-page.test.js
+++ b/kuma/javascript/src/search-results-page.test.js
@@ -12,7 +12,6 @@ const fakeResults = [
         slug: 'slug1',
         title: 'title1',
         summary: 'summary1',
-        tags: ['tag1.1', 'tag1.2'],
         score: 10,
         excerpts: ['Test <mark>result</mark>', 'Excerpt <mark>2</mark>']
     },
@@ -20,7 +19,6 @@ const fakeResults = [
         slug: 'slug2',
         title: 'title2',
         summary: 'summary2',
-        tags: ['tag2.1', 'tag2.2'],
         score: 5,
         excerpts: []
     }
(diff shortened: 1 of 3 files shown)


kuma/javascript/src/search-results-page.test.js
```diff
@@ -12,15 +12,13 @@ const fakeResults = [
...
         summary: 'summary1',
-        tags: ['tag1.1', 'tag1.2'],
         score: 10,
         excerpts: ['Test <mark>result</mark>', 'Excerpt <mark>2</mark>']
     },
     {
         slug: 'slug2',
         title: 'title2',
         summary: 'summary2',
-        tags: ['tag2.1', 'tag2.2'],
```
escattone: Nit: As above, I wonder if we should keep the tags here as well?

kuma/javascript/src/search-results-page.test.js
```diff
@@ -154,8 +152,7 @@ describe('SearchRoute', () => {
                                 _source: {
                                     slug: r.slug,
                                     title: r.title,
-                                    summary: r.summary,
-                                    tags: r.tags
```
escattone: Nit: ...and here?

peterbe:
Note-to-self; this is not ready because the one and only consumer of the v1 api is this and the v1 API needs to stop sending tags. See https://github.com/mozilla/kuma/blob/a4eb20130e4f5b658bc6cfee33365c369ece0874/kuma/api/v1/views.py#L261

peterbe:
@escattone Radically, and cheekily, I just changed the v1 search API to *not* include tags. 
It hurts my soul a little to break a versioned API but I think the circumstances excuses me. We haven't spec'ed it, we haven't published about it, and most importantly of all we haven't encouraged its use. 

By the way, I just edited away the tags from the search and no tests noticed. THat's what https://github.com/mozilla/kuma/pull/5543 is all about.

PR #5575, opened 2019-07-19
bug 1567572 - gravatar_34 is inefficient and never used

peterbe:
I don't think this will conflict with https://github.com/mozilla/kuma/pull/5546 

Thing is, the `DocumentContributorsJob`cacheback job has a TTL of 12h and the `UserGravatarURLJob` job has a TTL of 24h. But basically, every 12h for every single document, we're calling `UserGravatarURLJob.get()` mostly as a waste of time.


peterbe:
Ooops! I forgot to check something. My first commit just removed the call. Perhaps I need to "clean deeper".

peterbe:
Ready for review!

PR #5586, opened 2019-07-24
bug 1566944 - going back scrolls to the top

peterbe:
See the comment about how to reproduce: https://bugzilla.mozilla.org/show_bug.cgi?id=1566944#c4

Removing this code solves the problem. And it makes sense. Forcing focus on that title DOM element will force the browser to scroll into view. 

That begs the question: What did that code really do? I.e. what was it for? Why was it necessary?
It was added in this commit: https://github.com/mozilla/kuma/commit/345687b46fb98d782c9256d78eb967acb21fbd4c#diff-17a8d9db184a0bff5b4b3e2054788150R518
I think it's related to a11y but I don't really get it.


peterbe:
Marking it has not ready because I don't really understood what functionality I just removed. Also, I haven't even looked at flow or jest yet.

PR #5590, opened 2019-07-25
Bug 1567218, add banner to wiki to distinguish it from devmo

schalkneethling:
Adds banner to `wiki.` with a link through to `d.m.o`

Screenshot 2019-07-25 at 19 03 35

@peterbe r?

### DIFF
--- a/kuma/wiki/jinja2/wiki/document.html
+++ b/kuma/wiki/jinja2/wiki/document.html
@@ -85,6 +85,12 @@
   {% endif %}
 {% endblock %}
 
+{% block wiki_notice %}
+    {% if request.get_host().startswith('wiki.') %}
+        {% include "wiki/includes/wiki-notice.html" %}
+    {% endif %}
+{% endblock %}
+
 {% block lang_switcher %}
   {% if (document.is_localizable or document.parent) and other_translations %}
     <form class="languages go" method="get" action="{{ wiki_url('Web') }}">
--- a/kuma/wiki/jinja2/wiki/includes/wiki-notice.html
+++ b/kuma/wiki/jinja2/wiki/includes/wiki-notice.html
@@ -1,4 +1,4 @@
-{% set url = 'https://developer.mozilla.org/' + request.path %}
+{% set url = settings.SITE_URL + request.path %}
 <div class="mdn-wiki-notice">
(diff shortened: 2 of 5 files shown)


kuma/wiki/jinja2/wiki/document.html
```diff
@@ -85,6 +85,12 @@
   {% endif %}
 {% endblock %}
 
+{% block wiki_notice %}
+    {% if request.get_host().startswith('wiki.') %}
```
peterbe: 👍🏼

kuma/wiki/jinja2/wiki/includes/wiki-notice.html
```diff
@@ -0,0 +1,7 @@
+{% set url = 'https://developer.mozilla.org/' + request.path %}
```
peterbe: I don't think so. First of all, this'll break on stage and local dev laptops. Also, `request.path` might have `...$edit` in it. What about search results? 
I suggest you move this into a jinja helper function and the initial solution could be:
```
def non_wiki_url():
    url = request.build_absolute_uri().replace('//wiki.', '//').split('$')[0]
    if ...:
    return url
```
  
```

schalkneethling: Perhaps, but unlikely as this is in the `document.html` template which is not used for edit pages.

schalkneethling:
> I don't agree with doing this only for when you're viewing a document. I think it should be there for everything. I realize now that my comment below about the jinja function assumes this idea that we have the banner always.
> 
> Please explain/convince me why we shouldn't always show it.
> 
> The new design of the banner is not very "screamy" and can thus always be shown.

Well, it is added to remind you when viewing a page, that perhaps you just edited, that you are on the editable wiki and not the main user-facing site. I suspect that when you are already editing a document, you do not need reminding that you are on the editable version ;)

Also, the editing UI is already so full of UI bits, it seems like this is just going to take up more UI real estate and add noise.

peterbe:
Ok. Fair enough. Don't let me hold you back. I still think you should make a thing that doesn't hardcode the `developer.mozilla.org` so that it at least works for Stage and for local development.

schalkneethling:
@peterbe updated to use `settings.SITE_URL` instead of the hard coded URL

PR #5595, opened 2019-07-26
Bug 1534911, page headings often loads with different font

schalkneethling:
@peterbe Finally got around to this one. While `swap` does seem like the property we want, and while it does solve the problem, it introduces a new one. As you can see from the Gif below when on a slow connection there is a period of time when the headings are invisible.

They are then rendered using a fallback, and finally the "correct" font swaps in.

swap

So we are swapping ;p one tradeoff for another in essence. I guess the period where the headings are not shown at all is the short block period, followed by fallback and then finally swap.

Thoughts? r?

### DIFF
--- a/kuma/static/styles/base/fonts/opensans.scss
+++ b/kuma/static/styles/base/fonts/opensans.scss
@@ -1,6 +1,6 @@
 @font-face {
     font-family: 'Open Sans';
-    font-display: 'fallback';
+    font-display: 'swap';
     src: url('#{$path-to-fonts}#{"OpenSans-Regular-webfont.woff2"}') format('woff2'),
          url('#{$path-to-fonts}#{"OpenSans-Regular-webfont.woff"}') format('woff');
     font-weight: normal;
@@ -9,7 +9,7 @@
 
 @font-face {
     font-family: 'Open Sans';
-    font-display: 'optional';
+    font-display: 'swap';
     src: url('#{$path-to-fonts}#{"OpenSans-Semibold-webfont.woff2"}') format('woff2'),
          url('#{$path-to-fonts}#{"OpenSans-Semibold-webfont.woff"}') format('woff');
     font-weight: bold;
@@ -18,7 +18,7 @@
 
 @font-face {
     font-family: 'Open Sans';
-    font-display: 'fallback';
+    font-display: 'swap';
     src: url('#{$path-to-fonts}#{"OpenSans-Italic-webfont.woff2"}') format('woff2'),
          url('#{$path-to-fonts}#{"OpenSans-Italic-webfont.woff"}') format('woff');
     font-weight: normal;
--- a/kuma/static/styles/locales/_zilla.scss
+++ b/kuma/static/styles/locales/_zilla.scss
@@ -2,7 +2,7 @@
 
 @font-face {
     font-family: zillaslab;
-    font-display: 'fallback';
+    font-display: 'swap';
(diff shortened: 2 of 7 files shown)


peterbe:
I don't understand what's going on in that GIF. Why are the headlines not showing immediately? Like the `<p>` tags below. 

If I read https://font-display.glitch.me/ right, `font-display:swap` has no delay (aka. invisible time). 

What I'm seeing in the GIF appears to be the behavior of `font-display: fallback`. I.e. sees if it can get the webfont within the first 100ms, if not use the system font and when the webfont finally comes in, switch to that.

schalkneethling:
> I don't understand what's going on in that GIF. Why are the headlines not showing immediately? Like the `<p>` tags below.
> 
> If I read https://font-display.glitch.me/ right, `font-display:swap` has no delay (aka. invisible time).
> 
> What I'm seeing in the GIF appears to be the behavior of `font-display: fallback`. I.e. sees if it can get the webfont within the first 100ms, if not use the system font and when the webfont finally comes in, switch to that.

Oh no, so fallback gives a ~100ms block period, then switches to the fallback. It then gives the custom font a ~3 second swap period at which point it abandons the custom font altogether and sticks with the fallback.

`swap` on the other hand is supposed to have a 0 second block period with an infinite swap. So the difference between `block` and `swap` is merely the block period.

I have to say I do find it very odd that with `swap` there is a period where the headings are invisible before it uses the "fallback".

peterbe:
> I have to say I do find it very odd that with swap there is a period where the headings are invisible before it uses the "fallback".

That's exactly why I'm doubting the GIF or my understanding of how this is supposed to work :)

schalkneethling:
> > I have to say I do find it very odd that with swap there is a period where the headings are invisible before it uses the "fallback".
> 
> That's exactly why I'm doubting the GIF or my understanding of how this is supposed to work :)

So, my thinking is that I need to rearrange the set of fallbacks. Currently we have quite the list. I am currently testing simply falling back to `serif` i.e.

Instead of doing:

```
font-family: x-locale-heading-primary,zillaslab,Palatino,"Palatino Linotype",x-locale-heading-secondary,serif;
```

I am going to do:

```
font-family: x-locale-heading-primary,zillaslab,serif;
```

peterbe:
If you make the simplest possible. E.g. `font-family: zillaslab,serif;` and reload, will it work as expected? 

Also, what the heck is `x-locale-heading-primary`??

peterbe:
I'm not going to review this just yet. Let me know when you're confident it's good and right and I'll check it out and do my own tests if needed.

schalkneethling:
@peterbe So, reading the spec again, this actually works as specified. The main difference between `swap` and `fallback` is the infinite swap time.

```
swap
Gives the font face an extremely small block period and an infinite swap period.

fallback
Gives the font face an extremely small block period and a short swap period.
```

What we are currently seeing with `fallback` is that the initial short(very short) block period happens, it then renders the heading with the fallback. Then a timer is started in which time it will swap in the custom font if it loads. For us, it looks like the timer ends before the font has successfully downloaded. So the browser sticks with the fallback.

Now with `swap` it will swap in the custom font whenever it is ready. The other thing is that the spec says:

```
Font block period
If the font face is not loaded, any element attempting to use it must render an *invisible* fallback font face.
```

So it seems like with `swap` browsers have elected to have a longer initial swap (aka invisible fallback font face) and thus the invisible font display is more evident.

schalkneethling:
So :) What `swap` gives us a surety that our headings will always(eventually) be displayed using our custom ZillaSlab font face.

PR #5598, opened 2019-07-27
Full Front End for MDN Signaling on Beta

munteanu210:
**UPDATE Sep 24, 2019**

- [x] [Schalk's advice](https://github.com/mozilla/kuma/pull/5598#issuecomment-534414291) about using the right aria attributes. 
- [x] Don't import `bcd.js` from `article.jsx`.
- [x] The function `activateBCDSignal` (or whatever it's called) should *not* concern itself with waffle flags at all. It's the callers job. 
  - [x] `bcd.js` needs to have its own entry in `PIPELINE_CSS` in `kuma/settings/common.py`
  - [x] It should, in `react_document.html` be invoked something like this: https://gist.github.com/peterbe/6e52d2243918a43bb0ea13e2c1582dd6
  - [x] Simply loading `<script src=/static/bcd-signal.js></script>` should put the `activateBCDSignal` function in the global `window` namespace.
- [x] `article.jsx` should call `window.activateBCDSignal(...)` and it needs to "memoize" so that it's safe to re-render the `<Article/>` component more than once without invoking `window.activateBCDSignal()` repeatedly. 
- [x] We don't need to worry about specifically lazy load the `bc-signal.scss`. It's going to be part of https://github.com/mozilla/kuma/issues/5669
- [x] (server-side) It should be impossible to run `curl -X POST -d ...` repeatedly. The django ratelimiter should protect it. 
- [x] (server-side) It should be impossible (400 Bad Request) to HTTP POST with any of the textarea fields *larger* than what the serializer allows. 
- [x] (server-side) Posting with `curl` without a valid CSRF cookie and header should yield a 403 Forbidden (or whatever the csrf middleware does). (Is this already done automatically?)

peterbe's note: To test this you first have to create a waffle Flag. This worked for me. 
Screen Shot 2019-09-18 at 4 32 11 PM


kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+     * Builds and creates brief explanation control which is basically a textarea
+     * @returns Brief explanation control as a `HTMLElement`
+     */
+    const createBriefExplanationControl = () => {
+        const headerText = gettext('Can you provide a brief explanation?');
+        const descriptionText = gettext(
+            'Briefly outline the issue you are highlighting.'
+        );
+
+        const textAreaControl = document.createElement('textarea');
```
escattone: For security reasons, I think we should set a `maxlength` on this, and check/enforce the length on the server-side.

escattone:
@munteanu210 I'm thinking about security concerns related to this expanded `bc-signals` endpoint (now accepting user-provided data). In addition to adding size/length validators on the `explanation` and `supporting_material` serializer fields, I'm thinking we should also limit this feature to logged-in users only, and that we should also add a rate limit (e.g., maybe even something quite restrictive like `@ratelimit(key='user_or_ip', rate='1/d', block=True)` -- one per ip/user per day).

atopal:
> I'm thinking we should also limit this feature to logged-in users only (e.g., using the `@login_required` decorator), and that we should also add a rate limit (e.g., maybe even something quite restrictive like `@ratelimit(key='user_or_ip', rate='1/d', block=True)` -- one per ip/user per day).

Unfortunately restricting it to logged in users would defeat the purpose of this tool, but rate limiting would be fine.

escattone:
@atopal Thanks for your input on the login restriction! I didn't know if that would fit with this project, and you've answered my question!

kuma/javascript/src/bcd.js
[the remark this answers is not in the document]
escattone: Just to be clearer here regarding the server side, I'd like to see size/length validators on the `explanation` and `supporting_material` serializer fields. If there is anything we can validate in terms of the content of those fields, that would be good to add as well.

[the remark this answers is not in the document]
peterbe: Maxlength on a `textarea` doesn't exist unless you have an `onchange` event or something that introspects its value. What really matters is the server-side validation so you can't maliciously HTTP POST in a monster string that bloats all memory we have.

schalkneethling:
@munteanu210 Thank you for the PR! Please ensure that you run all the SVG files through [SVGOMG](https://jakearchibald.github.io/svgomg/) and that there is an empty line at the end of the file. Thank you!

kuma/javascript/src/bcd.js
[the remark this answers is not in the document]
munteanu210: I added a maxlength of 200 characters. We can decrease/increase later if we decide to. I see maxlength defined [here](https://www.screencast.com/t/juWx1B48kl4) on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea). Will push this update along with a few others in the next coming days.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+     * Builds and creates supporting material control which is basically a textarea
+     * @returns Supporting material control as a `HTMLElement`
+     */
+    const createSupportingMaterialControl = () => {
+        const headerText = gettext('Do you have any supporting material?');
+        const descriptionText = gettext(
+            '<b>Browser documentation and release notes</b> are good supporting items to accompany your message. A demo hosted on services like <b>Codepen</b> or <b>JSBin</b> are perfect for providing real examples of your findings.'
+        );
+
+        const textAreaControl = document.createElement('textarea');
```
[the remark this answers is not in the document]
munteanu210: Following your feedback, I set a maxlength here as well.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+    /**
+     * Creates and returns the signal element HTML
+     * @returns signal element HTML
+     */
+    const signalElem = () => {
+        const extContainer = document.createElement('div');
+        const separator = document.createElement('hr');
+        const container = document.createElement('div');
+        const signalLink = document.createElement('a');
+        signalLink.textContent = gettext('What are we missing ?');
```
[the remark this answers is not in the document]
munteanu210: Following your feedback, I centered the text and removed the space.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+     * Moves user to a step, by settings displaying related blocks
+     * @param {Number} step - Step index to be shown
+     */
+    const toStep = step => {
+        if (step !== bcSignalStep) {
+            const prevStep = bcSignalStep;
+            bcSignalStep = step;
+            stepsInfoSpan.innerText = `${gettext(
+                'Step'
+            )} ${bcSignalStep} ${gettext('of')} ${bcSignalSteps}`;
```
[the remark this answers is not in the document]
munteanu210: Done. Thank you for your suggestion! Will push this update along with a few others in the next coming days.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+        leftBlockHeader.className = 'left-block-header';
+        leftBlockImage.className = 'left-block-image';
+
+        closeButton.addEventListener('click', toggleBcSignalBlock);
+
+        leftBlockHeader.innerText = gettext(
+            'Tell us what’s wrong with this table'
+        );
+        leftBlockDescription.innerHTML = gettext(
+            'Our goal is to provide accurate, real values for all our compatibility data tables. Notifying MDN of inaccurate data or supplying new data pushes us further towards our goal of providing <b>100% real values</b> to the developer community. <br /><b>Thank you for helping.</b>'
```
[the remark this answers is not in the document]
munteanu210: @peterbe, you are right. I solved the problem using l10n.js string interpolation to inject the HTML tag openings and closings. The text values will be translated, whereas the decorating tags will remain static. Do you think this is enough?

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+                )}
+            </p>
+            <b>${gettext('Can I keep track of my report?')}</b><br />
+            </p>
+                ${gettext(
+                    'You can join the GitHub repository to see updates and commits for this table data'
+                )}:
+            </p>
+            <p>
+                <a href="https://github.com/mdn/browser-compat-data" target="_blank">
```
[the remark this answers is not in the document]
munteanu210: Done. I added rel="noopener noreferrer" to the link, to eliminate the vulnerability.

munteanu210:
> @munteanu210 Thank you for the PR! Please ensure that you run all the SVG files through [SVGOMG](https://jakearchibald.github.io/svgomg/) and that there is an empty line at the end of the file. Thank you!

@schalkneethling I did this already. But SVGOMG removes the space at the end of the files, so I added this manually now, to solve the problem. Will push this update along with a few others in the next coming days.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+        closeButton.className = 'button close-btn';
+        closeButtonWrapper.className = 'close-button-wrapper';
+        leftBlockWrapper.className = 'column-5 left-block';
+        leftBlockHeader.className = 'left-block-header';
+        leftBlockImage.className = 'left-block-image';
+
+        closeButton.addEventListener('click', toggleBcSignalBlock);
+
+        leftBlockHeader.innerText = gettext(
+            'Tell us what’s wrong with this table'
```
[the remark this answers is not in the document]
munteanu210: @peterbe Thanks for you suggestion. I tried various approaches for this problem, but in the end none worked. First of all, I tried with the ascii code (&#x27;) and it didn't work. It seems that gettext escapes the ampersand and the code isn't converted as it should be. The same happens with the HTML entity code. 

The alternative way was escaping the apostrophe in the "normal way" using single quotes and a \', but the prettier --write task converts the wrappings to double quotes and linter doesn't like that. See [screenshot](https://www.screencast.com/t/h1FhgFW9). I noticed that in other JS files, the developers used this same alternative method (single quotes and escape). However, those files are not verified by linter.

These being said, let me know if you have any suggestions on how to overcome this problem?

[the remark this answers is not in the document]
schalkneethling: Hey @munteanu210, on the mozorg project we always used `’` for this purpose and it was actually a requirement from the l10n team. I would say what you have here is the correct approach.

kuma/wiki/migrations/0005_auto_20190805_0306.py
```diff
@@ -0,0 +1,35 @@
...
+# Generated by Django 1.11.22 on 2019-08-05 03:06
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wiki', '0004_auto_20190708_1314'),
```
peterbe:
```suggestion
        ('wiki', '0008_auto_20190914_2209'),
```
because new migrations have been added since.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+        )}</span>`;
+        completeRightBlockDescription.innerHTML = `
+            <b>${gettext('What happens next?')}</b><br />
+            <p>
+                ${gettext(
+                    'Our team will review your report. Once we verify the information you have supplied we will update this browser compatability table accordingly.'
+                )}
+            </p>
+            <b>${gettext('Can I keep track of my report?')}</b><br />
+            </p>
```
peterbe:
```suggestion
            <p>
```

peterbe: By the way, this would have been avoided if this was JSX instead.

peterbe:
@munteanu210 Do you know why this is behind a waffle flag? Or *where* the justification for that is?
The reason I ask is that we're already using way too many waffle flags and we tend to forget to clean them up. They incur a cost on every single request at the convenience of being able to do percentage rollouts or being able to switch it on/off quickly in the django admin. 
Another option would be to simply make it a `django.conf.setting` that is controlled by an environment variable. Or `django-constance` which is like waffle flags but without the ability to roll out selectively. 

Or, another option is to not bother and just have it without any feature flagging.

peterbe:
Here's what it looks like;
Screen Shot 2019-09-18 at 4 39 22 PM

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+        bcSignalCompleteBlock.classList.add('open');
+
+        fetch(signalApiUrl, {
+            method: 'POST',
+            body: JSON.stringify(payload),
+            headers: {
+                'X-CSRFToken': getCookie('csrftoken'),
+                'Content-Type': 'application/json'
+            }
+        });
```
peterbe: The final page "lies" :)
It said "Report Sent" when in fact the XHR post failed. (it failed on an 500 internal server error because my local mysql doesn't have the right database migration).

Essentially, I think we need to take care of that better. Getting a 500 from the server isn't completely unrealistic but what's very realistic is network failures or 504 gateway timeout or something.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+        finishButton.className = 'button neutral main-btn';
+        finishButton.innerText = gettext('Finish');
+
+        finishButton.addEventListener('click', finishReport);
+        completeImageTextBlock.innerText = gettext('Thank you!');
+        completeRightTextBlock.innerHTML = `<span>${gettext(
+            'Report sent'
+        )}</span>`;
+        completeRightBlockDescription.innerHTML = `
+            <b>${gettext('What happens next?')}</b><br />
```
schalkneethling: Do not use the `<b>` tag instead use relevant semantic heading tags. An `<h3>` or `<h4>` should be good.

kuma/javascript/src/bcd.js
```diff
@@ -271,45 +271,722 @@ export function activateBCDSignals(
...
+            <p>
+                ${gettext(
+                    'Our team will review your report. Once we verify the information you have supplied we will update this browser compatability table accordingly.'
+                )}
+            </p>
+            <b>${gettext('Can I keep track of my report?')}</b><br />
```
schalkneethling: Same here: Do not use the `<b>` tag instead use relevant semantic heading tags. An `<h3>` or `<h4>` should be good.

kuma/wiki/models.py
```diff
@@ -2037,6 +2037,12 @@ class BCSignal(models.Model):
         verbose_name=_('Document (optional)'),
         on_delete=models.SET_NULL
```
peterbe: Probably too late now but do you know what the reasoning was for this? If a document is ever deleted, what's the point of keeping a `BCSignal` row that you have no idea what browser feature it's talking about?

kuma/javascript/src/article.jsx
```diff
@@ -114,7 +114,9 @@ export default function Article({ document }: DocumentProps) {
     useEffect(() => {
         let rootElement = article.current;
         if (rootElement) {
-            activateBCDSignals(document.slug, document.locale, userData);
+            if (userData) {
```
peterbe: I'm confused. The use of this depends on being logged in but it's not enforced on the server.

kuma/javascript/src/bcd.js
munteanu210: Sounds good. Will add this update.

kuma/javascript/src/bcd.js
munteanu210: Sounds good. Will change this.

kuma/javascript/src/bcd.js
munteanu210: Sounds good. Will change this.

kuma/javascript/src/bcd.js
mcroud: Hello @peterbe, thank you for the comment here.
I will design a new screen to cover this scenario. When we say network failure, what would the user's actions be in this event? Would they be able to "try again" or would this likely be a permanent "no can do"? would there be more than one common "fail" state?  
Thank you!

peterbe: My preferred pattern is:
1. Submit is triggered, disable the submit button, and send the `fetch()` request
2. If it all worked, fine, do what we we already have
3. if it failed (either exception caught or `!response.ok), enable the submit button again, and display a message in red that the server submission failed.

mcroud: @peterbe something like this?:
Signalling - Step 2 error

peterbe: We can discuss the messaging when you turn that into code.

kuma/javascript/src/article.jsx
munteanu210: @peterbe, The userData object is created even if the user is not logged in. I tested this on a [browser & computer](https://www.screencast.com/t/UNSO6kobXY) where I didn't ever sign in before. 

As far as I remember, the reason why I added this control, was that the activateBCDSignals component was rendered twice due to userData changing. The component gets rendered with null, then again with the object having a new structure and value (see [screenshot](https://www.screencast.com/t/uQrUh6o9mR4)). Then, if the userData object is not passed, the waffle flag inside bcd.js is not even evaluated (see [screenshot](https://www.screencast.com/t/uQrUh6o9mR4)), so the "What are we missing?" button doesn't show.

johngian:
> [Kadir said:](https://github.com/mozilla/kuma/pull/5598#issuecomment-528492753)
> 
> > Unfortunately restricting it to logged in users would defeat the purpose of this tool, but rate limiting would be fine.
> 
> So I think we should make it publicly available. And to avoid getting too much crazy inbound server traffic we put a rate limit and a length validator on all textareas (in rest_framework serializer lingo).

I agree with that, it makes sense to deal with it in the seralizer level.

Also, answering inline to some of the questions:

> I'm a bit confused and concerned about a couple of things:
> 
>     * Why are we doing this? Where is the data going? Who needs and requested this?
This started as an idea to see how we (Open Innovation + MDN team) can improve the quality of BCD data. We tried a very lightweight approach with just a button to let users indicate if something is wrong and the experiment went well in terms of responses from the users. The next step is to let users indicate what exactly is wrong with the data.

The data is currently stored to the DB and we just show the entries in /admin but the idea is to somehow export spreadsheets. It's only for internal consumption for now. @Elchi3 can tell us more about how we are aiming to use the data.
> 
>     * Why are we storing some of the structured data as a (comma separated) string?
I guess this is just for the sake of simplicity since we are not currently aiming to use that somehow. We can always improve the DB model to be more clean. From what I see the only field that needs to be changed is the `browser`, right?

> 
>     * Why do we bother with waffle flags? They tend to be forgotten. The only advantage of waffle is that you can roll "in and out" between releases and you can set percentages (and other fancy tricks). If we're not planning to do that I'd rather we just add it without any feature flags. If we need to pull it, we'll make a git revert or a code change and make another release.
When we first started working on that we didn't want to expose the feature to all the users because it was just an experiment. If we think it's there to stay then I agree that waffle flags add complexity.

peterbe:
> Also, answering inline to some of the questions:
> 
> > I'm a bit confused and concerned about a couple of things:
> > ```
> > * Why are we doing this? Where is the data going? Who needs and requested this?
> > ```
> 
> This started as an idea to see how we (Open Innovation + MDN team) can improve the quality of BCD data. We tried a very lightweight approach with just a button to let users indicate if something is wrong and the experiment went well in terms of responses from the users. The next step is to let users indicate what exactly is wrong with the data.
> 
> The data is currently stored to the DB and we just show the entries in /admin but the idea is to somehow export spreadsheets. It's only for internal consumption for now. @Elchi3 can tell us more about how we are aiming to use the data.
> 

I'll chase Daniel and Florian to get more involved instead of just reviewing the implementation code without the full product context. 

> > ```
> > * Why are we storing some of the structured data as a (comma separated) string?
> > ```
> 
> I guess this is just for the sake of simplicity since we are not currently aiming to use that somehow. We can always improve the DB model to be more clean. From what I see the only field that needs to be changed is the `browser`, right?
> 

I think so. ...not that I understand the usage.
As long as we're storing the right identifier for the browser, I think it's fine actually, because a fancy `ArrayField` or `JSONField` isn't buying us that much. 

> > ```
> > * Why do we bother with waffle flags? They tend to be forgotten. The only advantage of waffle is that you can roll "in and out" between releases and you can set percentages (and other fancy tricks). If we're not planning to do that I'd rather we just add it without any feature flags. If we need to pull it, we'll make a git revert or a code change and make another release.
> > ```
> 
> When we first started working on that we didn't want to expose the feature to all the users because it was just an experiment. If we think it's there to stay then I agree that waffle flags add complexity.

Yeah, that makes sense. 

I still think we should reconsider. Waffle flags are for when you're very uncertain about deploying something. This here feels like a feature that should either be fully implemented or not implemented at all.

kuma/javascript/src/article.jsx
peterbe: So it's not really the user you're waiting for, it's the waffle flags. :)

I raised the topic of waffle elsewhere and I think we should redirect our attention to that discussion and when that's settled come back here. 

Also, the implementation should be fixed anyway because this effect is triggered by `[document, userData]` which makes me nervous. There might be something else that changes `document` or `userData` when in fact the only things that matter are `[document.slug, document.locale, userData.waffle.flags['bc-signals']]`.

To start with, remove any mention of "waffle" in `bcd.js` and make `activateBCDSignal` mean what it's called. E.g.
```javascript
bcTable = document && document.querySelector('.bc-table');
if (bcTable) {
    if (SOME WAY OF TELLING IF IT'S ALREADY *not* ACTIVATED) {
      const waffleEnabled = userData && userData.waffle.flags['bc-signals'];
      if (waffleEnabled) {
         activateBCDSignals(document.slug, document.locale)
      }
    }
}
```

peterbe: We can discuss the use of waffle elsewhere but the invokation needs to become idempotent so that it doesn't hurt to trigger this effect without accidentally activating it twice.

schalkneethling:
@munteanu210 It seems that all of the SVG images are decorative in nature. With that I would recommend that you add the following to all of them:

```
... role="img" aria-hidden="true />
```

kuma/javascript/src/bcd.js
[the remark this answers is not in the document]
atopal: can we start with 1000 characters? Just pasting a URL would otherwise max that out.

schalkneethling:
@munteanu210 Here is what needs to be done for the Webpack piece:

Create a new file inside `kuma/webpack-build-utils/presets` called `webpack.bcd.js`
Inside this file, add the following:

```
const path = require('path');

module.exports = () => ({
    target: 'bcd',
    entry: path.resolve(__dirname, './kuma/javascript/src/bcd-signal.jsx'),
    output: {
        filename: 'bcd-signal.js',
        path: path.resolve(__dirname, '../../kuma/javascript/dist/')
    }
});
```

Inside `package.json` add the following under `scripts`:

```
"webpack:bcd": "npm run webpack -- --env.presets bcd --env.presets svg",
...
"webpack:production:bcd": "npm run webpack:bcd -- --env.mode production",
...
"webpack:development:web": "npm run webpack:bcd -- --env.mode development --env.presets hardsource",
```

You can now run either of the following script targets:

```
npm run webpack:bcd // generates the bcd-signal.js file
npm run webpack:production // generates everything for production as well as the bcd-signal.js file
npm run webpack:web // generates everything in dev mode as well as the bcd-signal.js file
```

Let me know if you run into any problems.

webpack-build-utils/presets/webpack.bcd.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+
+module.exports = () => ({
+    target: 'web',
```
schalkneethling: This is 💯

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
...
+        }
+        if (controlObj.optional) {
+            controlHeader.className += ' with-optional-label';
+        }
+
+        controlDescription.className = 'control-description';
+
+        controlHeader.appendChild(createStepNumLabel(controlObj.index, true));
+        controlHeader.appendChild(document.createTextNode(controlObj.header));
+        controlDescription.innerHTML = controlObj.description;
```
peterbe: Here's another concern with security. That `controlObj.description` is potentially coming from the l10n service which is maybe out of our control. So someone can fake the translation and inject whatever they want in the DOM for unsuspecting users.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
...
+                bEnd: '</b>',
+                br: '<br />'
+            }
+        );
+
+        closeButtonWrapper.appendChild(closeButton);
+        leftBlockWrapper.appendChild(closeButtonWrapper);
+        leftBlockWrapper.appendChild(leftBlockImage);
+        leftBlockWrapper.appendChild(leftBlockHeader);
+        leftBlockWrapper.appendChild(leftBlockDescription);
```
peterbe: This is an example of why React is so much better. It has a scheduler that writes batch updates to the virtual DOM. Here instead, it will update the DOM 5 times in quick succession.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
...
+        extContainer.setAttribute('class', 'signal-link-ext-container');
+        container.appendChild(separator);
+        container.appendChild(signalLink);
+        extContainer.appendChild(container);
+        return extContainer;
+    };
+
+    bcTable = document && document.querySelector('.bc-table');
+
+    if (!bcSignalsRendered && bcTable && bcTable.insertAdjacentElement) {
```
peterbe: Is this necessary? Haven't you already checked for that on the 2nd line of this function? (referring to the `bcSignalsRendered` boolean variable)

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
+// @flow
+import { gettext, interpolate } from './l10n.js';
+import { getCookie } from './utils.js';
+
+let bcSignalsRendered = false;
+
+export function activateBCDSignals(slug: string, locale: string) {
+    if (bcSignalsRendered) {
+        return;
+    }
```
peterbe: Why not just do:
```
export function activateBCDSignals(slug: string, locale: string) {
    if (bcSignalsRendered) {
        return;
    }
    bcSignalsRendered = true;
```
Perhaps rename that variable to `bcSignalsInvoked`. 

I can't yet see if that's realistic but what if this function gets called twice within a short span of time, async. Since the setting of `bcSignalsRendered = true;` happens 800 lines away from here, that second invocation might come sooner.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
...
+
+    /**
+     * Resets all the controls to initial state
+     */
+    const resetControls = () => {
+        for (const browser of document.querySelectorAll('.browser.selected')) {
+            browser.classList.remove('selected');
+        }
+
+        const brief = document && document.getElementById('brief-explanation');
```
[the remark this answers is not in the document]
munteanu210: The condition is suggested by ci-flow. Flow requires explicit type checks, otherwise the code doesn't pass the CircleCI flow tests. There are other occurrences of this kind of checks, all for the same reason.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
...
+            selectedTableRow = selectRow.options[selectRow.selectedIndex].value;
+        }
+        const briefExplanation =
+            document && document.getElementById('brief-explanation');
+        if (
+            briefExplanation &&
+            briefExplanation instanceof HTMLTextAreaElement &&
+            briefExplanation.value &&
+            typeof briefExplanation.value === 'string'
+        ) {
```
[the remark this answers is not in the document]
munteanu210: My [answer from above](https://github.com/mozilla/kuma/pull/5598#discussion_r329282002) applies here as well.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,809 @@
...
+            'Tell us what’s wrong with this table'
+        );
+        leftBlockDescription.innerHTML = interpolate(
+            gettext(
+                'Our goal is to provide accurate, real values for all our compatibility data tables. Notifying MDN of inaccurate data or supplying new data pushes us further towards our goal of providing %(bStart)s100% real values%(bEnd)s to the developer community. %(br)s%(bStart)sThank you for helping.%(bEnd)s'
+            ),
+            {
+                bStart: '<b>',
+                bEnd: '</b>',
+                br: '<br />'
```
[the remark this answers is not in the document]
munteanu210: A similar conversation was brought up above. I [responded](
https://github.com/mozilla/kuma/pull/5598#discussion_r324394375) to it shortly after that. Let me know what your thoughts are on it. Also, as you said, it may be good to discuss this item with the rest of the team to see what suggestions they have.

kuma/javascript/src/bcd-signal.jsx
munteanu210: You have a point. As a general rule I prefer setting the boolean after the component is invoked and rendered (who knows what happens in the process), but in this case your concern is valid and probably moving it is a good idea.

munteanu210:
@peterbe Thank you for taking the time to review my updates, and send feedback! As always, I appreciate it! I will answer your [comment](https://github.com/mozilla/kuma/pull/5598#pullrequestreview-294459147) below. 

I agree with you that with React some things can be accomplished easier and with significantly less code.

I started the front end development on MDN Signaling at the end of June, and I asked the team what technologies I am allowed to use, and I specifically mentioned React and jQuery. I was told that none of these two can be used. Pure JS and existing technologies already in the project were the ones allowed, and making the codes compatible with a future React implementation was the requirement.

After doing a PR on the old codebase, we were asked to move the MDN Signaling code "as is" on the React side with some rework for everything to work, which we did. I guess we have to discuss with the rest of the team to see what's the best way to move forward: merge the MDN Signaling as is and as it was requested initially, or rewrite everything in React.

As for the webpack, it was and still is outside the areas that I feel comfortable and proficient with. Also, it is outside the framework that I was responsible for, but I was happy to jump on it for the greater cause of getting the MDN Signaling closer to completion. As you know, I discussed about the webpack with you and Schalk, and I very much relied on your input on how to do it best.

kuma/javascript/src/bcd-signal.jsx
munteanu210: I totally agree! But the initial requirement was different. Please see my [last comment](https://github.com/mozilla/kuma/pull/5598#issuecomment-536135278).

kuma/javascript/src/bcd-signal.jsx
munteanu210: I see what you say, and makes sense to me. May I kindly ask, what would be your approach on solving this?

peterbe:
Ok. We have to have forwards momentum. Let's ignore that it should have been written in React in the first place. And let's keep the current Webpack hack. Perhaps @Gregoor can rewrite that once this lands. 

The remaining big challenge is the security risk of rogue non-en-US translations that, if slipped through Pontoon, can inject XSS attacks. I think I have a solution!

First of all, @Gregoor is [working on a possible solution](https://github.com/mozilla/kuma/pull/5872/files) to be able to have HTML markup inside translated strings. But it's a bit overkill and it's honestly not ready. (Either we land his PR or we invest properly in `fluent-react` which will add kilobytes of JS code)

And the best solution might just be to fix our darn CSP! The reason it hasn't been fixed is because we rely on inline scripts too much that we can't easily move these over to external static scripts in time. 

But I reviewed all uses of `.innerHTML` in `bcd-signal.jsx` and I realized that the only single place where you need markup inside the string is here: https://github.com/mozilla/kuma/blob/360041240b0b6a60a2d08493bfe80d1703541ac2/kuma/javascript/src/bcd-signal.jsx#L504
Let's just rewrite that string so it doesn't depend on making parts of the text bold. Yes, it's compromise on the final result, but it'll definitely make it possible to land this. 

All other uses of `.innerHTML` can be replaced with `.innerText`. E.g.

```diff
-errorLabelEl.innerHTML = gettext('Connection error:');
+errorLabelEl.innerText = gettext('Connection error:');
```

And this just needs to be broken up into individual DOM node creations. 
https://github.com/mozilla/kuma/blob/360041240b0b6a60a2d08493bfe80d1703541ac2/kuma/javascript/src/bcd-signal.jsx#L673-L692

Does that feel like a good plan?

escattone:
> (server-side) Posting with curl without a valid CSRF cookie and header should yield a 403 Forbidden (or whatever the csrf middleware does). (Is this already done automatically?)

Sorry for getting to this late. It was something I thought about as well in my initial review (but didn't say anything), so just wanted to say that, yes, this is done automatically for **all** POST requests (actually all non-safe requests) via our `django.middleware.csrf.CsrfViewMiddleware`.

kuma/javascript/src/bcd-signal.jsx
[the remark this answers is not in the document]
peterbe: For the record, @Gregoor suggested your can solve it by defining a type. See https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAJgUwMYwIYBOmYAboWOnNgK4C2mAdgC4BcYA3tkXk5gKIxM9ZmwAUAZzbimBAJYMA5gEowAXgB8HXHnHiAMrOks8DAJ4AaMNt0A5PPTYnTAX2cBuVKkFMwmAgTgCAFlMXTwFTAB1AjwAB1i-D1RsOAZpK25ePn9AkLCIgAkTdEECNTBRFQ0OVDBfHODQ8XComPi-csoaYSYAOi5MHn5BHtEAcnRZEjGlDzq-AMb81riEgn78W3tiVTAxhEIGeQV6xYBaemWxj3dPFLSfInFMJgBhVJk4GHFyyrVNdi1U65JotaKrPwbHT6Qx9Ii0OAkTDjEiGWQAI0EMxuSQGQ2yizyzUKxVKlQ8Txe72YAW+5KAA
I think it can wait because perhaps a better solution altogether is to rewrite everything in React but that too can wait.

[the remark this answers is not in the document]
munteanu210: @peterbe thanks for your note. We'll keep this for the record as you said, and wait with the implementation until we decide that we really need this implemented, or everything is rewritten in React at a later date.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,829 @@
+// @flow
+import { gettext, interpolate } from './l10n.js';
+import { getCookie } from './utils.js';
+
+let bcSignalsInvoked = false;
+
+export function activateBCDSignals(slug: string, locale: string) {
```
peterbe: Would it be possible to just do:
```suggestion
window.activateBCDSignals = (slug: string, locale: string) => {
```
instead?

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,829 @@
...
+        const control = document.createElement('div');
+        const controlInnerWrapper = document.createElement('div');
+        const controlHeader = document.createElement('div');
+        const controlDescription = document.createElement('div');
+
+        control.className = 'control';
+        controlInnerWrapper.className = 'control-wrap';
+        controlHeader.className = 'control-header';
+        if (controlObj.inline) {
+            controlHeader.className += ' has-control';
```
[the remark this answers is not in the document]
munteanu210: OK, sounds good!

kuma/javascript/src/bcd-signal.jsx
munteanu210: This is doable. I will check if this works well in a bit.

kuma/javascript/src/bcd-signal.jsx
```diff
@@ -0,0 +1,829 @@
...
+        sendReportButton.innerText = gettext('Send report');
+        signalStepTwoBlock.className = 'inner-step';
+        controls.className = 'controls';
+        stepsButtonBlock.className = 'navigation-buttons reverse';
+        goBackButton.className = 'button prev-step-btn btn-dark';
+        goBackIcon.className = 'icon-back';
+        const backBtnLabel = document.createTextNode(gettext('Previous step'));
+
+        goBackButton.appendChild(goBackIcon);
+        goBackButton.appendChild(backBtnLabel);
```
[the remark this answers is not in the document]
munteanu210: It's due to the nature of the content, that needs to be appended to the button. In this case we had an icon and then a text to be appended to it. I couldn't alter the content of the button directly with innerText, since it has two elements inside. The createTextNode function allows to specify an initial text that is added to that node. It avoids multiple steps that createElement('whatever') needs and outputs a simple text node without a <span> or other tags to accompany it. The last step is appending the icon and the text node to the button. 

Do you prefer the other option, for consistency?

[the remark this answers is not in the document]
peterbe: Nah. Leave it as is.

kuma/api/v1/views.py
```diff
@@ -287,6 +288,7 @@ class APISearchView(SearchView):
 
 @waffle_flag('bc-signals')
 @api_view(['POST'])
+@throttle_classes([UserRateThrottle])
```
escattone: I don't see where you've set the `rate` limit for the `UserRateThrottle` class, so my understanding is that this will do nothing. You'll either need to do something like this:
```python
class OncePerDayUserThrottle(UserRateThrottle):
    rate = '1/day'
...
@throttle_classes([OncePerDayUserThrottle])
```
or instead of `throttle_classes`, just use the rate-limit decorator we use everywhere else in the kuma code:
```python
@ratelimit(key='user_or_ip', rate='1/d', block=True)
```

johngian: Isn't it too much to throttle to ones per day?

atopal: > Isn't it too much to throttle to ones per day?

yes, we should start with 10/day, and can adjust if necessary

johngian:
@escattone I updated the PR after your comments

munteanu210:
@escattone You can resolve the conflict easily by taking the file as is from the pull request. The file from master was more like a placeholder. The conflict is probably caused by the fact that the file was renamed by schalkneethling's commit yesterday. Or, as an alternative, I could rename the file to match schalkneethling's naming convention, adding the underscore in the filename. Let me know what you prefer.

escattone:
> @escattone I updated the PR after your comments

@johngian The changes look good and all of them worked for me locally when testing, thank you!

> @escattone You can resolve the conflict easily by taking the file as is from the pull request. The file from master was more like a placeholder. The conflict is probably caused by the fact that the file was renamed by schalkneethling's commit yesterday. Or, as an alternative, I could rename the file to match schalkneethling's naming convention, adding the underscore in the filename. Let me know what you prefer.

@munteanu210 Could you please resolve the merge conflict by renaming the file to match @schalkneethling 's naming convention? Once you do that and the tests pass, I think I can land this PR.

escattone:
@munteanu210 Actually, I can rename that file, so no worries.

escattone:
@munteanu210 Just FYI, I renamed the file to match @schalkneethling 's naming convention, and I also resolved the merge conflicts within that same file by taking the changes from this PR as is.

PR #5599, opened 2019-07-28
Bug 1568305, add specific styling for active state on in content links

schalkneethling:
@callahad This adds a specific style for the active state of in-content

active-style

r?

### DIFF
--- a/kuma/static/styles/components/wiki/elements.scss
+++ b/kuma/static/styles/components/wiki/elements.scss
@@ -11,8 +11,8 @@ article {
         }
 
         &:active {
-            color: $blue-dark;
-            font-weight: bold;
+            background-color: $blue-dark;
+            color: #fff;
         }
 
         &[name] {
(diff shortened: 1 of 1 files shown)


callahad:
Looks good, but I noticed that using bold can force other text on the line
to jump / reflow when you click -- maybe just a stronger color change would
be sufficient?

schalkneethling:
> Looks good, but I noticed that using bold can force other text on the line to jump / reflow when you click -- maybe just a stronger color change would be sufficient?

I was thinking that as well originally but, it is tricky to find a color combo that is also accessible for people with color blindness. Our current link color is blue, which is great but, it makes a different color kinda tricky unless we go for something like an orange/yellow.

I am a little wary of that direction because generally those colors indicate a warning, or that something is not quite right.

That is the reason I chose to indicate the interaction not only with color.

schalkneethling:
@callahad Here are two more options to consider:

### Using background-color

background-active

### Using text-deoration-style: double; as you suggested

double-active

callahad:
I really like the background-color one, would r+ :)

schalkneethling:
@callahad updated.

PR #5603, opened 2019-07-30
Bug 1569958, separate menu CSS for static site from wiki

schalkneethling:
This stops us from serving the main menu code for both the current and new menu to both front-ends. @peterbe r?

### DIFF
--- a/kuma/static/styles/components/structure-shared.scss
+++ b/kuma/static/styles/components/structure-shared.scss
@@ -43,7 +43,7 @@ Main Navigation
 - search bar
 -------------------------------------------------------------- */
 
-@import 'structure/page-header/nav-main';
+@import 'structure/page-header/static/main-nav';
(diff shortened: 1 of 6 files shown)


schalkneethling:
Need to resolve the conflict so, marked as not-ready

kuma/static/styles/components/structure-shared.scss
```diff
@@ -0,0 +1,105 @@
...
+
+/*
+Header
+====================================================================== */
+
+@import 'structure/page-header/main-header';
+@import 'structure/page-header/submenu';
+
+/*
+Man navigation and toolbox wrapper
```
peterbe:
```suggestion
Main navigation and toolbox wrapper
```

schalkneethling:
@peterbe This is ready for review.

PR #5611, opened 2019-08-01
Bug 1569816, switch highlight section to CSS only

schalkneethling:
This removes the need for the `highlightSections` JavaScript function that ensures that some of the document headings use the slab style dark backgrounds. It also adds a couple of small updates to the `addAnchors` function to:

1. Not add it to hidden headings
2. To add the link to a different location on the DOM based on whether the heading is an `h2` or `h3`

I tested this on all the top-level pages liked form the main menu, as well as the `Array` page.

@peterbe r?

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -63,23 +63,27 @@ const styles = {
     })
 };
 
-// This is an effect function that runs every time the article is rendered.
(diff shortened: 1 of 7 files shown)


schalkneethling:
Just double checking one piece of CSS I can most like remove. Will remove the not-ready label soon.

PR #5614, opened 2019-08-01
bug 1525719 - Beta supports wiki endpoints

escattone:
This PR enables the `beta` domain to support the `wiki` endpoints, which means it will either directly support the endpoint or redirect to the `wiki` domain. It also removes all references to the `beta` domain (e.g., `beta_shared_cache_control()` and `is_beta()`), except for the actual `beta` domain settings (`BETA_HOST`, `BETA_ORIGIN`, and `BETA_SITE_URL`). The PR that launches the `beta` domain, can remove these final vestiges of an intermediate state. Here are some other highlights:

- We no longer set the `urlconf` attribute on the incoming request to the separate the `beta` and `wiki` domain endpoints (so all of the `urls_beta.py` files have been removed).
- There is no longer any need to use the `urlconf` keyword argument to the `kuma.core.urlresolvers.reverse()` function unless we're in the un-trusted domain and want the URL for an endpoint in the `beta`/`wiki` domains (see https://github.com/mozilla/kuma/blob/8e31d237dc7252e5ea62090409bd9bd61f0af95b/kuma/wiki/models.py#L1275). When within the normal, `beta`, or `wiki` domains, all URLs can be acquired via reverse without worrying about setting the proper `urlconf`.
- It replaces the `is_beta()` utility function with `is_wiki()`
- Simplifies some of the `react_base.html` template, since it's only used when we're not in either of the `wiki` or `untrusted` domains.
- Removes the `beta_shared_cache_control()` function. Now all endpoints use `shared_cache_control()` (defaults to 5 minutes). Once we launch the `beta` domain, we can increase that value.
- For now, the `is_wiki()` function uses a "negative" test, meaning it returns `True` if the request is not explicitly for the `beta` domain (e.g., for both https://wiki.developer.mozilla.org/ and  https://developer.mozilla.org/ `is_wiki()` returns `True`). The PR to launch the `beta` domain should change this to a positive test, meaning only return `True` if the request is explicitly for the `wiki` domain (e.g., for https://wiki.developer.mozilla.org/  `is_wiki()` returns `True` but for https://developer.mozilla.org/ it returns `False`)
- When this PR is merged, the `beta` domain will either directly support every endpoint, or redirect it to the `wiki` domain. The endpoint is directly supported if it returns JSON or it returns HTML that is never cached, and the endpoint still has meaning outside of the `wiki` domain (e.g., the `/<locale>/docs/ckeditor_config.js` and `/<locale>/users/ban/<username>` endpoints do NOT have any meaning outside of the `wiki` domain).


peterbe:
This is what tidings sent to someone who was subscribing to changes:

```
worker_1         | To: IAMVECTOR@EXAMPLE.COM
worker_1         | Date: Fri, 02 Aug 2019 17:55:20 -0000
worker_1         | Message-ID: <20190802175520.15.55841@b4e5a5a2f74a>
worker_1         | X-Kuma-Document-Url: http://mdn.localhost:8000/en-US/docs/Web/HTML
worker_1         | X-Kuma-Document-Locale: en-US
worker_1         | X-Kuma-Editor-Username: peterbe
worker_1         | X-Kuma-Document-Title: HTML: Hypertext Markup Language
worker_1         |
worker_1         |
worker_1         |
worker_1         |
worker_1         | peterbe changed HTML: Hypertext Markup Language.
worker_1         |
worker_1         |
worker_1         |
worker_1         | --- [en-US] #1564219
worker_1         |
worker_1         | +++ [en-US] #1564220
worker_1         |
worker_1         | @@ -8,7 +8,7 @@
worker_1         |
worker_1         |        {{HTMLSidebar}}
worker_1         |      </div>
worker_1         |      <div>
worker_1         | -      Doing my own edit.
worker_1         | +      Doing my own edit..
worker_1         |      </div>
...
worker_1         |
worker_1         | --
worker_1         | View User Profile:
worker_1         |  http://mdn.localhost:8000/en-US/profiles/peterbe?utm_campaign=Wiki+Doc+Edits&utm_medium=email&utm_source=developer.mozilla.org
worker_1         |
worker_1         | Compare on MDN:
worker_1         |  http://mdn.localhost:8000/en-US/docs/Web/HTML$compare?from=1564219&to=1564220&utm_campaign=Wiki+Doc+Edits&utm_medium=email&utm_source=developer.mozilla.org
worker_1         |
worker_1         | View Article:
worker_1         |  http://mdn.localhost:8000/en-US/docs/Web/HTML?utm_campaign=Wiki+Doc+Edits&utm_medium=email&utm_source=developer.mozilla.org
worker_1         | Edit Article:
worker_1         |  http://mdn.localhost:8000/en-US/docs/Web/HTML$edit?utm_campaign=Wiki+Doc+Edits&utm_medium=email&utm_source=developer.mozilla.org
worker_1         | Article History:
worker_1         |  http://mdn.localhost:8000/en-US/docs/Web/HTML$history?utm_campaign=Wiki+Doc+Edits&utm_medium=email&utm_source=developer.mozilla.org
worker_1         | --
worker_1         | You are subscribed to edits on: HTML: Hypertext Markup Language.
worker_1         |
worker_1         |
worker_1         |
worker_1         |
worker_1         | Unsubscribe from these emails:
worker_1         | https://localhost:8000/en-US/unsubscribe/4?s=UnhYxMVUat
```

kuma/core/tests/test_middleware.py
```diff
@@ -74,11 +74,12 @@ def test_force_anonymous_session_middleware(rf, settings):
...
-     ('beta-origin', 'BETA_ORIGIN', 'kuma.urls_beta'),
+    (('beta', 'BETA_HOST', None),
+     ('wiki', 'WIKI_HOST', None),
+     ('beta-origin', 'BETA_ORIGIN', None),
      ('demos', 'ATTACHMENT_HOST', 'kuma.urls_untrusted'),
      ('demos-origin', 'ATTACHMENT_ORIGIN', 'kuma.urls_untrusted')),
-    ids=('beta', 'beta-origin', 'attachment', 'attachment-origin')
+    ids=('beta', 'wiki', 'beta-origin', 'attachment', 'attachment-origin')
 )
 def test_restricted_endpoints_middleware(rf, settings, host, key, expected):
```
peterbe: Note-to-Ryan: This test should be exclusively about the attachments (trusted/untrusted)

kuma/dashboards/urls.py
```diff
@@ -19,7 +22,7 @@
         views.topic_lookup,
         name='dashboards.topic_lookup'),
     url(r'^localization$',
-        shared_cache_control(RedirectView.as_view(
+        shared_cache_control(s_maxage=WEEK)(RedirectView.as_view(
```
peterbe: This could do with a code comment to explain. The truth is that it's exclusively an optimization thing that only works because of CloudFront.

kuma/urls.py
```diff
@@ -39,12 +40,14 @@
...
 urlpatterns = [url('', include('kuma.health.urls'))]
+# The non-locale-based landing URL's
 urlpatterns += [url('', include('kuma.landing.urls'))]
+# The locale-based landing URL's
 urlpatterns += i18n_patterns(url('', include(landing_lang_urlpatterns)))
 urlpatterns += i18n_patterns(
     url(
         r'^events',
-        shared_cache_control(RedirectView.as_view(
+        shared_cache_control(s_maxage=MONTH)(RedirectView.as_view(
```
peterbe: Needs code comment that explains this is only for CloudFront optimization.

kuma/wiki/views/document.py
```diff
@@ -787,18 +798,15 @@ def document(request, document_slug, document_locale):
...
-@process_document_path
-@newrelic.agent.function_trace()
-@ratelimit(key='user_or_ip', rate='1200/m', block=True)
 def react_document(request, document_slug, document_locale):
     """
     View a wiki document.
     """
+    # This view supports the "redirect" query parameter only. If any other
+    # query parameter is used, redirect to the wiki domain.
+    if any(param != 'redirect' for param in request.GET):
```
peterbe: Let's change this to be something like:
```python
if set(request.GET.keys()) && set(whitelist_query_params):
    return redirect_to_wiki()
```

kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -1,5 +1,5 @@
-{% set beta = settings.ENABLE_RESTRICTIONS_BY_HOST and is_beta(request) %}
-{% set untrusted = settings.ENABLE_RESTRICTIONS_BY_HOST and is_untrusted(request) %}
+{% set wiki = False %}
```
peterbe: Perhaps add a code comment here that explains that the reason we're doing this is for the benefit of the jinja includes that might need to know if we're on the wiki.

escattone:
@peterbe @schalkneethling I addressed everything in the review except for the React-based 4XX.html and 500.html templates. On further thought, I think those changes are outside the scope of this PR, and that we should create a tracking issue and separate PR to address that. What do you guys think?

schalkneethling:
> @peterbe @schalkneethling I addressed everything in the review except for the React-based 4XX.html and 500.html templates. On further thought, I think those changes are outside the scope of this PR, and that we should create a tracking issue and separate PR to address that. What do you guys think?

Makes sense to me. Let's get this one merged and then move on to that next. Currently, Travis is :(

escattone:
@schalkneethling Sounds good. It's strange that the test that failed in Travis didn't fail for me locally, but it has been fixed.

escattone:
I created https://github.com/mozilla/kuma/issues/5623.

PR #5625, opened 2019-08-06
update sub-modules

escattone:
Update the `kumascript` and `locale` sub-modules.


escattone:
Mistake, re-opening.

escattone:
Merging this since the failure is due to the https://codecov.io/bash service being unavailable. All tests passed otherwise.

PR #5629, opened 2019-08-09
Bug1570036 segment header styling for static and wiki sites

schalkneethling:
If I did this correctly, essentially nothing should change visually and the header should render exactly the same as it does currently. This applies both to `beta` as well as `wiki`.

This also enables `prettier` for the new style framework that is in the making i.e. `minimalist`

And finally fixes some formatting failures reported by `prettier`

@callahad r?


callahad:
Works as expected on my end.

I do find it odd that we're defining styles directly on `<header>` elements for the page versus an ID or class, but if you're happy enough with the structure then I'm happy.

I did note that these commits were signed with a GPG key that's not associated with your GitHub account, but I don't believe we enforce verified signatures, do we?

I still need to dig into those CI failures before hitting the approve button.

schalkneethling:
> Works as expected on my end.
> 
> I do find it odd that we're defining styles directly on `<header>` elements for the page versus an ID or class, but if you're happy enough with the structure then I'm happy.
> 

Agreed, let me be more explicit here.

> I did note that these commits were signed with a GPG key that's not associated with your GitHub account, but I don't believe we enforce verified signatures, do we?
> 

Not sure, probably not worth blocking this one but I will look into it.

> I still need to dig into those CI failures before hitting the approve button.

I know why those are happening. The `prettier` one's is odd as I did fix them locally. Anyhow, will update the PR soon.

PR #5631, opened 2019-08-12
use JSON.parse for hydration data

### LINKED ISSUE #5619: We should use JSON.parse instead of a JS literal for hydration state
See https://v8.dev/blog/cost-of-javascript-2019#json

We're currently doing this...
```
<script>window._react_data = {"locale": "en-US", "stringCatalog": {}, ... </script>
```

We should be doing...
```
<script>window._react_data = JSON.parse('{"locale": "en-US", "stringCatalog": {}, ... ')</script>
```

peterbe:
Fixes #5619

It used to render out this:
```
<script>window._react_data = {"locale": "en-US", "stri
```
now it renders out this:
```
<script id="react-data" type="application/json">{"locale": "en-U
```
To get to the same variable value, inside the client-side code instead of just accessing `window._react_data` it now does `JSON.parse(document.getElementById('react-data').textContent)`.

Same functionality, just packaing things (very) differently. 

This means, that when the browser parses the HTML document, as soon as it gets to that `<scrit` tag, instead of pausing the HTML/CSSOM parsing, to go into JavaScript parsing mode, it just moves on and ignores this as a piece of other-data. Then, whence the JavaScript code has been downloaded, parsed and started executing, it *now* turns that blob of data into real data. 

So I think it's got 3 big advantages:

1. `var stateData = JSON.parse(<state data as string>)` is faster than `var stateData = <state data>`
2. This way, the HTML parsing isn't blocked by the main thread having to switch over to JS parsing. Remember, inline script tags are blocking because it could be that there's a `document.write()` in there that would make the CSSOM different!
3. One less inline JS script that would block us from enabling CSP. (We still need to solve the `pluralFunction` stuff some day)

To test; please try some non-trivial documents that in in non-en-US.

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -13,14 +13,12 @@ if (container) {
     // The HTML page that loads this code is expected to have an inline
     // script that sets this window._document_data property to an object
     // with all the data needed to hydrate or render the UI.
-    let data = window._react_data;
-
-    // Remove the global reference to this data object so that it can
-    // be garbage collected once it is no longer in use.
-    window._react_data = null; // eslint-disable-line camelcase
+    // let data = window._react_data;
+    let data = JSON.parse(document.getElementById('react-data').textContent);
+    let pluralFunction = window._pluralFunction;
 
     // Store the string catalog so that l10n.gettext() can do translations
-    localize(data.locale, data.stringCatalog, data.pluralFunction);
+    localize(data.locale, data.stringCatalog, pluralFunction);
 
     let app = null;
     // This switch statement is duplicated in ssr.jsx. Anything changed
--- a/kuma/wiki/templatetags/ssr.py
+++ b/kuma/wiki/templatetags/ssr.py
@@ -58,14 +58,7 @@ def _render(component_name, html, state):
     form of the state dict, in the format expected by the client-side code
     in kuma/javascript/src/index.jsx.
     """
-    # We're going to need this below, but we don't want to keep it around
-    pluralExpression = state['pluralExpression']
-    del state['pluralExpression']
-
-    # Serialize the state object to JSON and be sure the string
-    # "</script>" does not appear in it, since we are going to embed it
-    # within an HTML <script> tag.
-    serializedState = json.dumps(state).replace('</', '<\\/')
+    inline_javascript = []
 
     # In addition to the JSON-serialized data structure, we also want
     # to pass the pluralForm() function required for the ngettext()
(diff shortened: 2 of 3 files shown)


peterbe:
One thing that I've seen mention around ideas like this https://twitter.com/HenrikJoreteg/status/1144273986647257088 is that reading from the DOM might be more time-consuming than you think. If you don't need to read the DOM to get the state data, you could basically "decouple" the Redux state (or just plain equivalent React state in our case) from the HTML/DOM. I.e. the JS is ready as soon as the .js files are read, parsed, and executed. 

However, suppose that reading from the DOM is slow, and ideally avoided, we're [already reading from the DOM anyway](https://github.com/mozilla/kuma/blob/a1ec0f7873d7cc22e04c894e7f7a42dfbef96dcc/kuma/javascript/src/index.jsx#L9-L11). And most importantly, the hydration should happen *after* the CSSOM has been fully computed and painted on screen. So, we're moving some of the CPU computational working away from the initial download and parsing phase into the document.onload phase.

peterbe:
Hmm... Seems Flow is rightfully complaining. I'll tinker with that tomorrow. I don't think it should block review though.

callahad:
If I'm reading [Henrik's blog post](https://joreteg.com/blog/improving-redux-state-transfer-performance) correctly, this is _"Option 2: Turn your state into an 'inert' script"_. Which didn't yield any meaningful performance gains compared to a normal JavaScript Object Literal.

Are you sure this is worth pursuing, versus jumping straight to fully inlining the state as a JS String literal?

peterbe:
> Are you sure this is worth pursuing, versus jumping straight to fully inlining the state as a JS String literal?

You're right! I was so bogged down on avoiding the JS literal that I didn't see how poorly (option #2 in his blog post) performed. 😢

I need to go back the drawing board a bit then. Especially in terms of being able to measure this stuff. 

However, I still don't know how we're ever going to be able to enable JS CSP that blocks inline scripts if we don't start with things like this.

peterbe:
Ugh. I'm losing patience with this. I ran Chrome Lighthouse on a page 5 times with this patch and 5 times on the master branch. The difference:

* On `master`: average TTI: 4.43s
* On patch branch: average TTI: 4.34s

The difference can just be a fluke. Let's assume it's the same. 

So then, I thought I'd try with what, in the blog post, he refers to as option 3. But it turns out to be horribly hard because of quotation escaping stuff.

schalkneethling:
> So then, I thought I'd try with what, in the blog post, he refers to as option 3. But it turns out to be horribly hard because of quotation escaping stuff.

Yup, I guess that is why he ended up using [jsesc](https://github.com/mathiasbynens/jsesc)

peterbe:
> Yup, I guess that is why he ended up using [jsesc](https://github.com/mathiasbynens/jsesc)

That's what I use on Song Search. I dare you to read the code. It's a iterating over the whole string. But the author is really good at JS so I'm feeling safe with it.

peterbe:
Just not worth it now given everything else (more interesting) going on.

PR #5639, opened 2019-08-14
Launch beta domain

escattone:
#### Description
This PR, in a nutshell, launches the `beta` domain, or in other words, the current `beta` domain becomes the main domain. The code that actually does that is tiny (see the changes in `kuma/core/utils.py`), but the consequences, especially for the unit and functional tests, are massive. Here are some highlights:
- the Selenium-based functional tests, and all of their related documentation and code, have been removed. This was a decision made by the development team along with product management. We felt that the effort required to maintain these tests outweighed their benefit. 
- the Locust-based performance tests have also been removed. We don't maintain these and never run them.
- the headless functional tests have significantly changed. For example, they now test nearly every endpoint that can be tested. I think only 5 endpoints are not "hit" (`/i18n/setlang/`, `/users/github/login`, `/en-US/users/account/recover/<key>/<key>`, `/csp-violation-capture`, and `/en-US/users/account/recover/send`).
- with this PR, all references to the `beta` domain have been removed from the code
- the two "lookup" dashboards, `/<locale>/dashboards/user_lookup` and `/<locale>/dashboards/topic_lookup`, are now only directly supported by the `wiki` domain (and so are now redirected from the main domain)

#### Testing
For local testing, I used these settings in my `.env` file:
```
DEBUG=True
DOMAIN=mdn.localhost
ATTACHMENT_HOST=demos:8000
```
```
##
# Host Database
#
##
127.0.0.1	localhost demos mdn.localhost wiki.mdn.localhost
255.255.255.255	broadcasthost
::1             localhost
```

and then at least do the following (assuming that you've activated a virtualenv loaded with `requirements/test.txt`):
```
cd <kuma-repo>
docker-compose pull
docker-compose up -d
docker-compose restart
pytest --base-url http://mdn.localhost:8000 tests
```


escattone:
The failures in Travis are extremely annoying in that I don't see them locally. For me, `npm run pretty` runs fine locally as well as `npm run format`. I'm not sure what's going on there. I need to look into the Python unit test failure (again I don't see that when I run locally). However, I'll do that tomorrow.

escattone:
Hm, when I run `docker-compose exec web npm run pretty`, I don't see the `prettier` check errors, but when I run the following on my Mac:
```
npm ci
npm run pretty
```
suddenly I do. I noticed it's using the newest version of node on my Mac, while the Docker image is using 10.14.2, but that's just a guess.

peterbe:
In travis we run `bin/ci-jslint` which runs `make jslint` which runs `npm run pretty` which ultimately means it doesn't run `docker-compose exec web npm run pretty`. Both approaches use (and should use) the same version of `prettier` in the `/package.json`. And we're using both Node 10 in docker and Node 10 in Travis.

MatonAnthony:
I love all the cleanup of code we don't use.

escattone:
@peterbe Just FYI, I do remove the beta feedback notice within this PR.

escattone:
NOTE: With a recent commit, this PR also changes the way Kuma tests are run within Jenkins. We no longer use `docker-compose.test.yml` (which has been removed), but instead run the Kuma tests in the same manner that we do within Travis, just using `docker-compose.yml`.

escattone:
just fyi, i'm keeping this from rotting by re-basing on `master` daily

escattone:
Re-based on `master`.

escattone:
FYI, just re-based this on `master` to keep it from rotting.

escattone:
Re-based on `master`. The only conflict was in `package-lock.json`, and what I do with those is simply re-generate the file using `npm install --package-lock-only` (which I ran both as `docker-compose exec web npm install --package-lock-only` and locally as `npm install --package-lock-only` to ensure there was no difference between the two generated files).

escattone:
Re-based on `master` in preparation for the possible `beta` launch tomorrow.

PR #5649, opened 2019-08-16
Fix Bug 1553512: blank paragraphs when pasting

callahad:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1553512

If `p:empty { display: none; }` is defined in the CKEditor stylesheet,
then pasting multiple lines of text also inserts an empty `<p></p>`
before the pasted content.

The "Source" button must be clicked twice to re-render the editor and
display the extraneous paragraphs.

I do not entirely understand why this resolves the issue, but it does.

### DIFF
--- a/kuma/static/styles/components/content.scss
+++ b/kuma/static/styles/components/content.scss
@@ -228,10 +228,8 @@ We cannot put these styles in the base/elements/ files because they don't apply 
     ====================================================================== */
 
     /*
-     * Don't allow empty paragraphs by CKEditor.
-     * But do allow empty divs created by the Prism syntax highlighter.
+     * Allow empty divs created by the Prism syntax highlighter.
      */
-    p:empty,
     div:not(.line-highlight):empty {
         display: none;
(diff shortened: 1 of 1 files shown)


callahad:
This CSS was added two years ago (Nov. 2017) in #4504 (32dcb8e30cf412622aa36793b847107af9894cf1) as part of [Bug 1414107 - Improve article scanability](https://bugzilla.mozilla.org/show_bug.cgi?id=1414107).

There have only been [five changes to ckeditor_config.js](https://github.com/mozilla/kuma/commits/master/kuma/wiki/jinja2/wiki/ckeditor_config.js) in that timeframe, none of which appear to have caused this bug.

Our version of CKEditor itself has also not changed in that timeframe.

...which makes me wonder how bug has somehow existed for nearly two years undetected.

The bug manifests in Firefox and Chrome on Linux (haven't tested Safari or other operating systems), so I don't think it's a change in browser behavior.

Issue #5645, opened 2019-08-15
As a user I want a streamlined account creation experience, so I can get started quickly


schalkneethling:
# User story
As a user, I want the instructions and help text to be simple and to the point when creating an account on MDN, so that I can get through this process quickly and easily.

# Tasks & Acceptance criteria

- [ ] Instruction content above fields is simplified to provided copy (Kadir to approve)
- [ ] Help text for trouble creating an account is moved to underneath submit button
- [ ] Remove the checkbox to make your Github URL public
- [ ] Update the help text for the username and email address fields (Kadir to approve)
- [ ] Simplify submit button text to "Create profile"

# Design
InVision: https://mozilla.invisionapp.com/share/Y8SGMKQBCW3#/368129372_Create_Account_-_MDN_Profile
Zeplin: https://zpl.io/2jYxqKp

Mobile Version (without header/footer): https://zpl.io/2v0ejgQ

peterbe:
I'm not sure I agree with putting sprints into kuma issues. 
We already have mdn/sprints for that. There, an issue is a bundle of functionality changes phrased around the idea of a "user story". Once you crack up your editor and roll up your sleeves the sprint issue is exploded into individual tasks. Aka. kuma issues that each have 1 PR each. 

What do you think?

And what do you think @tobinmori ?

schalkneethling:
> I'm not sure I agree with putting sprints into kuma issues.
> We already have mdn/sprints for that. There, an issue is a bundle of functionality changes phrased around the idea of a "user story". Once you crack up your editor and roll up your sleeves the sprint issue is exploded into individual tasks. Aka. kuma issues that each have 1 PR each.
> 
> What do you think?

I agree to some extent. I would say including the user story portion and the links to the design etc. was overkill. The reference to the original user story assigned to Nick was enough. This one could be focused on just the tasks/acceptance criteria.

Reasons I wanted the issue assigned to me here instead of in the sprints repo is so that I can connect my PR to the issue, have the issue closer to the code, and get the added benefit of Github project tracking.

jmswisher:
@schalkneethling Please update the milestone for this and other issues in mozilla/kuma. Nina Simone (Sprint 2) ended 2 weeks ago. The current sprint is Odetta (S3 Q3 2019), and the next one is Peppermint (S4 Q3 2019). (I would need write access to mozilla/kuma in order to update these issues.)

schalkneethling:
[...]

PR #5659, opened 2019-08-20
Fix issue #5645, update copy for new user registration

schalkneethling:
This is the copy updates only. This does not change the design/layout at all.

## Before

beta mdn localhost_8000_en-US_users_account_signup

## After

beta mdn localhost_8000_en-US_users_account_signup(updated)


callahad:
...you'll also need to adjust the `'having trouble'` to `'have trouble'` to get those template tests to pass.

PR #5664, opened 2019-08-20
Fix Bug 1574120 - Adding a Button For Copying Example Code

itai421:
This fix adds a button on every code example that copies the code. It looks like this:

image

(the icon only shows on hover )

This is my first PR for the project, and there a few things I am not sure regarding this change:

1. I used the icon found [here](https://www.flaticon.com/free-icon/copy_109492#term=copy&page=1&position=45) simply because I think it resembles the `jinja2/includes/icons/file-icons/file.svg` icon. I really do not know if it is a licensing issue, or even if this icon looks good.
2. Regarding mobile devices - I do not know if this functionality should be enabled when browsing from a mobile device, but for now it is.
3. The icon is set to shown on the right of the example box, I do not know if this will look strange with rtl languages. I think it will be fine, because as far as I know, code examples are always in English, but I did not checked it.


peterbe:
Thanks for your contribution!!
Mind you, we're swamped over here trying to launch the new read-only React site and be aware that it might be some time till we have time to review this. But I love the fact that you stepped up and "came with code" :)

jinja2/includes/icons/general/copy.svg
```diff
@@ -0,0 +1,80 @@
...
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+</svg>
```
schalkneethling: I would suggest that you use this icon:

```
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="9" y="9" width="13" height="13" rx="2" ry="2"/><path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"/></svg>
```

itai421:
@peterbe I am happy to contribute. Is the react read only site is a replacement for the current version of MDN? if so I think this PR can be closed.  

In any case, if one wishs to add a copy button to the new site, the logic is relatively simple and is all written at  `copyHandler`. The `this` at line 18 refers to the `pre` element that contains the code example.

Issue #5666, opened 2019-08-21
Remove search css from home page


schalkneethling:
On the homepage of the static site we no longer have the large search input in the hero area. Do we intend to bring this back? If not, there is some CSS we can segment and no longer serve to the static site.

atopal:
no plans for now

Issue #5613, opened 2019-08-01
Beta site HTML is larger


peterbe:
In beta.developer.mozilla.org the HTML payload is much larger. 

Comparing:

* https://beta.developer.mozilla.org/en-US/docs/Web/HTML/Element/link - 393K (59K gzipped)
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link - 225K (39K gzipped)

See:
https://htmltree.herokuapp.com/?url=https:%2F%2Fbeta.developer.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FHTML%2FElement%2Flink
https://htmltree.herokuapp.com/?url=https:%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FHTML%2FElement%2Flink

peterbe:
FYI, I dumped the whole `_react_data` to a JSON file and analyzed it:
```
 locale 5 0.0%
 url 20 0.0%
	 wikiURL 50 0.0%
	 absoluteURL 20 0.0%
	 enSlug 8 0.0%
	 language 12 0.0%
	 title 31 0.0%
	 locale 5 0.0%
	 lastModified 26 0.0%
	 slug 8 0.0%
	 bodyHTML 20,015 23.0%
	 hrefLang 2 0.0%
	 quickLinksHTML 62,998 75.0%
	 translateURL 58 0.0%
	 tocHTML 351 0.0%
	 summary 368 0.0%
```

Note! The `quickLinksHTML` part is 75% of the total payload!!
I *think* `quickLinksHTML` was added late in David's rush to finish. It would definitely account for why the *total* payload slows us down.

peterbe:
For reference, the page I did this on was: http://beta.mdn.localhost:8000/en-US/docs/Web/HTML

peterbe:
On http://beta.mdn.localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
the same numbers are:
```
 locale 5 0.0%
 url 65 0.0%
	 wikiURL 95 0.0%
	 absoluteURL 65 0.0%
	 enSlug 53 0.0%
	 language 12 0.0%
	 title 52 0.0%
	 locale 5 0.0%
	 lastModified 26 0.0%
	 slug 53 0.0%
	 bodyHTML 23,943 75.0%
	 hrefLang 2 0.0%
	 quickLinksHTML 6,705 21.0%
[...]

PR #5667, opened 2019-08-21
Fix issue #5666, remove hero search field CSS from static site

schalkneethling:
Removes the hero area search field CSS from the new front-end. I also took the opportunity to move the background SVG to a background image as we are not applying any styling to it from CSS.

@peterbe r?


peterbe:
Users still have to download the 4KB for the SVG but at least it can be done a little later and if someone has it browser cached, it means the HTML will be -4KB all win. 

So this'll make a tiny dent on https://github.com/mozilla/kuma/issues/5613

peterbe:
@schalkneethling Does this PR fix https://github.com/mozilla/kuma/issues/5666 ?
If so, can you edit the PR description to say `Fixes #5666\n\n...`

PR #5671, opened 2019-08-21
Remove prism-json.css from PIPELINE_CSS

peterbe:
There is no `js/prism-mdn/components/prism-json.css` file anywhere. 
There is a `prism-json.js` though. 
It's sad that django-pipeline doesn't complain louder about that. I should fix that.

Screen Shot 2019-08-21 at 11 07 41 AM


ExE-Boss:
Looks like I&nbsp;forgot to&nbsp;do&nbsp;this in&nbsp;#5189.

Issue #5668, opened 2019-08-21
Segment breadcrumb CSS


schalkneethling:
The styling for breadcrumbs are different between the static site and the wiki but, we serve all the CSS to both front-ends. We should segment and only serve the relevant CSS to the relevant front-end

schalkneethling:
Fixed in https://github.com/mozilla/kuma/commit/3da27ac44cc0511d76e6563e26ad8724fbb91372

PR #5674, opened 2019-08-22
Fixes issue #5668, segment breadcrumb CSS

schalkneethling:
This segments the CSS for breadcrumbs between the two different front-ends. I also cleaned up the HTML significantly and moved the CSS out of Emotion and into Sass. Visually, nothing should really have changed.

@callahad r?


schalkneethling:
Thanks, @callahad fixed:

Screenshot 2019-08-22 at 17 36 21

PR #5679, opened 2019-08-22
create print stylesheet and serve with media=print

### LINKED ISSUE #5678: Combine all print styles in print.css and serve with media=print
Currently, we have print specific styling peppered throughout the codebase. This is a maintenance problem and adds to the overall size of our CSS bundles.

Moving all our print specific styling into its own CSS file will improve maintainability but, now that it is its own "bundle", we can serve it separately with `media="print"`. When the site is viewed on screen, the `media` type of `print` won't match, and so the browser will:

1. Mark it is low priority in terms of download and parsing
2. When it does load it, it will be loaded in a non-blocking manner

schalkneethling:
Fixes #5678

This is in no way an attempt to improve our styling for print. This merely moves all print related styling into its own stylesheet and bundle and serves it using `media="print"`

@peterbe r?


peterbe:
Seems pretty significant. Can you please create an issue for it and then mentiont that issue in the PR description?

kuma/static/styles/minimalist/print.scss
```diff
@@ -0,0 +1,69 @@
+@import '../includes/vars';
+@import './mixins/print-mixins';
+
+@media print {
```
peterbe: What's the point of this? If this file is in a `<link href=print.css media=print>` it is already for printing.

kuma/static/styles/minimalist/print.scss
```diff
@@ -0,0 +1,69 @@
...
+    header {
+        display: none;
+    }
+
+    .center {
+        @include remove-center-spacing();
+    }
+
+    .notification-tray {
+        display: none;
```
peterbe: When you do it like this, the uglified output becomes:
```css
.notification-tray{display:none}.newsletter-box{display:none}.global-notice{display:none}.nav-footer{display:none} ...etc...
```
See [snapshot output](https://gist.github.com/peterbe/e686af33789999462485650075f3c9af)

The remedy would be to do something like this: https://github.com/peterbe/workon/blob/643ef0de2a3de09a0f1c460227b9c173ba91c36f/public/print.css#L1-L10

schalkneethling:
> Seems pretty significant. Can you please create an issue for it and then mentiont that issue in the PR description?

The issue is here https://github.com/mozilla/kuma/issues/5678 and is mentioned in the PR title(?)

kuma/static/styles/minimalist/print.scss
schalkneethling: Very true, there is no need for it here anymore.

kuma/static/styles/minimalist/print.scss
schalkneethling: Ah yes, I was not even thinking in that direction because I merely wanted to get everything into its own stylesheet, but this is *much* better.

kuma/settings/common.py
```diff
@@ -764,6 +763,15 @@ def pipeline_one_scss(slug, **kwargs):
         ),
         'output_filename': 'build/styles/home.css',
     },
+    'print': {
+        'source_filenames': (
+            'styles/minimalist/print.scss',
```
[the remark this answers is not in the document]
schalkneethling: Oh, no. I have a little side project called `mdn-minimalist` and this will become(if I get to suggest/decide) the new CSS that is used for MDN Web Docs. I am therefore slowly starting to mimic the directory structure inside Kuma ;p

schalkneethling:
> I'm not sure this is a good idea. It used to be that these selectors were scattered nicely into the files where they belong. If we one day decide, hypothetically, to kill the newsletter, you'd simply `git rm newsletter.scss` and it would take care of the print directives for newsletter too. With the approach here in this PR, it's very likely that you'll remember to edit `print.scss` and remove the `.newsletter-thing` too.

This is true but, if you look at the current state of the print styles, you will notice that it has not been kept up to date with the array of changes that have happened to the site, even though they were co-located.

If you apply the print stylesheet to the document, you will see that very little(other than the footer) is actually being hidden. For me, personally, it makes a lot more sense to have everything print related in `print.css`.

Currently, our front-end codebase is not documented, but that is changing with the work(on the side) I am doing with MDN-Fiori. As such, once this is out of my head and not only on my machine ;p, there will be much better documentation around these kinds of workflows/process.

> Also, what is the gain? The browser always downloads <link media=print href=/print.css> anyway. And now there's one more file it needs to download.

Yes but, as mentioned above, the current print styling is "not great", and to make it better will mean the actual size of the stylesheet will increase. If this happens where the style rules currently live, it will bloat the overall CSS bundle(s).

Having it separate means:

1. The `screen` bundles are not impacted
2. Print styling is isolated
3. Because it is served with `media="print"` the browser places a very low priority on it when viewed on a target that is not `print` and when it does download it, it does so in a non-blocking manner
4. We use HTTP2 right, so what is another low priority, non-blocking HTTP request between friend :)

schalkneethling:
@peterbe update.

kuma/static/styles/minimalist/print.scss
```diff
@@ -0,0 +1,30 @@
...
+.breadcrumbs,
+a.edit-section,
+.page-buttons,
+.open-in-host-container,
+.newsletter-box,
+#page-attachments,
+#offline-dialog-content,
+#wiki-left {
+    display: none;
+}
```
schalkneethling: And, with everything in the same print stylesheet, we can have this tiny optimization :)

peterbe:
> The issue is here #5678 and is mentioned in the PR title(?)

I edited the commit title and the PR description.

PR #5686, opened 2019-08-26
Fixes #5685, remove emotion and move styling to Sass

### LINKED ISSUE #5685: Remove @Emotion
We are seeing more and more potential long term maintenance problems with using Emotion in the way we currently do. It currently also complicates local development with server-side rendering to some extent but more importantly, the code that is generated adds a *lot* of bloat to our HTML via inlined `style` tags, sometimes as many as ~220.

There might be(there possibly is) ways to improve this but, the effort to investigate this does not seem to justify the end goal. Having all our CSS in CSS[read Sass] just makes a lot more sense.

With that, it is time to remove Emotion in favor of Sass.

schalkneethling:
This removes Emotion from the new front-end codebase in favor of keeping all styling in Sass/CSS. Beyond the general ergonomics and maintainability(which was not the main reasons for doing this), I had a look at the react.js file pre and post removing Emotion. The numbers are as follows:

## Pre

`react.js` 202.05kb uncompressed (63.31 gzipped)

## Post

`react.js` 171.94kb uncompressed (52.08 gzipped)

The styling that was moved to Sass adds back ~3.171kb gzipped which still shows a ~11.23kb saving overall. Through this, I have also found a number of additional areas where we can clean-up and optimize the styling delivered to the new front-end. These will follow shortly.

@peterbe r?

Peter, also note, I did not link up `search-results.scss` because I will leave that to you to determine the best place to hook that into the work you are currently doing on search. Thanks!


schalkneethling:
Ugh! This is frustrating. So it fails because of

```
ValueError: The file 'styles/general/flask.svg' could not be found with <kuma.core.pipeline.storage.ManifestPipelineStorage object at 0x7fd27c47d210>.
```

Problem is, I cannot duplicate this locally. I even ran the `docker-compose` commands used in Travis locally and they all pass :/ Still digging around to see why this is failing here.

schalkneethling:
I have no idea how the changes I initially made here is having this knock-on effect that breaks paths all over the places. This is no even going as far as PNG images in `kuma/static/styles/components/wiki/promos.scss` ¯\_(ツ)_/¯

schalkneethling:
@peterbe Just ran the following from `ci-python` locally, and it completes without error. Any idea why it keeps failing on Travis?

```
docker-compose exec -T web make clean

# Needed for Django's check_for_language
docker-compose exec -T web make localecompile
docker-compose exec -T web make build-static
```

schalkneethling:
As another attempt to reproduce the build failure locally I did:

```
manually delete the static directory
docker-compose down
docker-compose up -d

docker-compose exec -T web make clean

# Needed for Django's check_for_language
docker-compose exec -T web make localecompile
docker-compose exec -T web make build-static
```

Completes with no issues.

schalkneethling:
Aha! This one now fails for the same reason:
https://github.com/mozilla/kuma/pull/5698

Not sure what that means, but at least it suggests something common.

schalkneethling:
> I have no idea what is causing this issue yet, but I was able to reproduce it locally. The key is to use `DEBUG=False` in your `.env` file, **AND** to make sure to run `docker-compose exec web ./manage.py collectstatic` rather than `docker-compose exec web make collectstatic`, since the latter will not respect `DEBUG=False`. So to reproduce, do the following:

Thanks, @escattone!! 🍵  I have deduced much of this as well but, the killer here is being able to reproduce this locally. Thanks again for looking into this.

Thinking some more on this I reckon it would be better to do this as a couple of smaller pull requests. For one thing, it will be easier to review and for another, it should enable me to isolate where this problem starts to show up.

schalkneethling:
I have split this into multiple pull requests:

- https://github.com/mozilla/kuma/pull/5701
- https://github.com/mozilla/kuma/pull/5700
- https://github.com/mozilla/kuma/pull/5702
- https://github.com/mozilla/kuma/pull/5703
- https://github.com/mozilla/kuma/pull/5704
- https://github.com/mozilla/kuma/pull/5705
- https://github.com/mozilla/kuma/pull/5706
- https://github.com/mozilla/kuma/pull/5707
- https://github.com/mozilla/kuma/pull/5708
- https://github.com/mozilla/kuma/pull/5709
- https://github.com/mozilla/kuma/pull/5710

With that, I am closing this one and moving work and discussion to the individual pull requests. Thanks everyone for your assistance.

PR #5691, opened 2019-08-26
preload ZillaSlab-Regular

### LINKED ISSUE #5687: Preload Zillaslab-Regular
As mentioned here https://github.com/mozilla/kuma/issues/5617#issuecomment-524443741, seeing that we do depend on `Zillaslab-Regular` we should preload it as we do with the bold-variant.

schalkneethling:
Fixes #5687

Preloading `ZillaSlab-Regular`

### DIFF
--- a/jinja2/includes/preload.html
+++ b/jinja2/includes/preload.html
@@ -1,2 +1,5 @@
 {# because this is non-blocking, preload as early as possible #}
-<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Bold.subset.woff2') }}" as="font" type="font/woff2" crossorigin />
+<link rel="preload" href="{{ static('fonts/locales/ZillaSlab-Regular.subset.woff2') }}" as="font" type="font/woff2"
+    crossorigin />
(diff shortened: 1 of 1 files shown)


schalkneethling:
> So it seems that there is a benefit in preloading both web fonts.
> 
> That leaves two questions:
> 
> 1. How many people suffer from having to download the `.woff2` in vain?

The general strategy behind this is that if [preload](https://caniuse.com/#search=preload) is supported, [woff2](https://caniuse.com/#search=woff2) will be supported as well.

> 2. Would it be better if the preload was done with a header instead of a `<link ref=preload>`?

Here I believe `preloading` early in the head is non-blocking whereas just sticking the `@font-face` bits in the head might be blocking? It comes down to how the browser implements this I guess. A `preload` is a resource loading hint/strategy whereas adding `@font-face` is more explicit so, the browser might handle these scenarios differently.

peterbe:
I meant, either put it as an HTTP header or within the `<head>` of the HTML document.

schalkneethling:
> I meant, either put it as an HTTP header 

Perhaps this is a first step?

> or within the `<head>` of the HTML document.

This include is included in the `<head>` (https://github.com/mozilla/kuma/blob/master/jinja2/react_base.html#L22) or do I misunderstand?

PR #5698, opened 2019-08-26
Fixes #5697, use correct structure partial for new frontend

### LINKED ISSUE #5697: Use correct `structure` Sass partial for new frontend
At some point `kuma/static/styles/minimalist/document.scss` switched to using the old `structure.scss` partial. We should switch back to the correct partial.

schalkneethling:
@peterbe r?

### DIFF
--- a/kuma/static/styles/minimalist/document.scss
+++ b/kuma/static/styles/minimalist/document.scss
@@ -2,7 +2,7 @@
 Styling that is shared across the wiki and read-only sites
 ====================================================================== */
 @import '../includes/vars';
-@import 'vars/vars-paths';
+@import './vars/vars-paths';
 @import '../includes/mixins';
 @import '../includes/mixins_indicators';
 @import '../components/readable-line-length';
@@ -18,12 +18,12 @@ Utitlity classes (reused in a variety of contexts)
 ====================================================================== */
 @import '../components/wiki/document-list';
 @import '../components/wiki/external-icon';
-@import 'structure/containers';
+@import './structure/containers';
 
 /*
 Components that may appear on most wiki pages
 ====================================================================== */
-@import 'base/structure';
+@import './base/structure';
 @import '../components/wiki/wiki-block';
 @import '../components/wiki/quick-links';
 @import '../components/wiki/reviews';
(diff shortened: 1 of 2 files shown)


schalkneethling:
> ```
> 'styles/general/flask.svg' could not be found
> ```

Yup

> Where is the reference to `flask.svg` and did this break as of this change?
> Perhaps you're not noticing this locally because you have a filled `./static/` folder with lots of other stuff from other branches. Perhaps you can reproduce locally by first deleting the `static` directory.

See [here](https://github.com/mozilla/kuma/pull/5686#issuecomment-524806712) and further down.

schalkneethling:
> Does this just need to be rebased to pass in Travis now?

Yup. But I will wait on this one until all the Emotion stuff is done.

schalkneethling:
@peterbe rebased and ready for merge/review

kuma/static/styles/minimalist/document.scss
```diff
@@ -23,7 +23,7 @@ Utitlity classes (reused in a variety of contexts)
 /*
 Components that may appear on most wiki pages
 ====================================================================== */
-@import '../components/wiki/structure';
+@import 'base/structure';
```
peterbe: Would it be safer and saner to use `./base/structure` here? So it's always relative?

peterbe:
I'm not entirely sure how to test this? Which pages should I look for that *might* break as of this change?

schalkneethling:
> I'm not entirely sure how to test this? Which pages should I look for that _might_ break as of this change?

There is not really anything to test. The changes that were made in `_structure.scss` was reviewed separately. At some stage, I must have had a PR merge conflict and merged incorrectly causing us to switch back to the previous `structure.scss`. This then just switches us back to the correct one.

schalkneethling:
@peterbe updated. Thx

peterbe:
@tobinmori @schalkneethling Something interesting happened here. I blame Schalk :) ...but that's not the point. We've all done it. 

This here is the PR (that you're reading). 
And [this is the issue](https://github.com/mozilla/kuma/issues/5697).

Notice that they have very different labels. 

* Issue: `bug`
* PR: `cleanup`, `performance`, `frontend`

It's no harm but definitely confusing and potentially misleading. 
Should labels be *copied* to the PR from the issue?
Or, should labels be *avoided* on the PR?
Some labels are very uniquely fitting to PRs such as the `not ready` label which we've come to understand as "This is not ready to be merged but maybe review".

If PRs appear on the ZenHub dashboard, the labels are very useful but it's only really a problem or applicable if the PRs are assigned. 

Food for thought!

schalkneethling:
> Should labels be _copied_ to the PR from the issue?

Good point @peterbe, I reckon those labels do not really make sense on a pull request. We should probably have certain keywords reserved for pull requests only. 👍

PR #5707, opened 2019-08-27
Issue #5685, remove Emotion and move CSS to Sass - search results page

schalkneethling:
@peterbe r?

Also as mentioned in the original PR:

> Peter, also note, I did not link up search-results.scss because I will leave 
> that to you to determine the best place to hook that into the work you are 
> currently doing on search. Thanks!

### DIFF
--- a/kuma/javascript/src/search-results-page.jsx
+++ b/kuma/javascript/src/search-results-page.jsx
@@ -1,6 +1,5 @@
 // @flow
 import * as React from 'react';
-import { css } from '@emotion/core';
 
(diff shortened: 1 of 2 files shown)


Gregoor:
Is there a good way to see what it looks like or should I just review it "formally"?

schalkneethling:
> Is there a good way to see what it looks like or should I just review it "formally"?

"formally" is good thanks! @peterbe is working on updating the search page so, once this is merged he can incorporate this into the updated page.

PR #5708, opened 2019-08-27
Issue #5685, remove Emotion and move CSS to Sass - task completion survey

schalkneethling:
@peterbe r?

### DIFF
--- a/kuma/javascript/src/task-completion-survey.jsx
+++ b/kuma/javascript/src/task-completion-survey.jsx
@@ -1,7 +1,6 @@
 // @flow
 import * as React from 'react';
 import { useContext, useEffect, useState } from 'react';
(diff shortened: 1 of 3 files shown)


schalkneethling:
@gregoor to test this one you need to log into the Django admin at `mdn.localhost::8000/admin/`

Here you want to look for the `Waffle` heading and then `Waffle flags`. Find the flag titled `sg_task_completion` and click on it. 

1. Set `Everyone` to Yes
2. Set `Percent` to 99 (not sure if you need to do this but it won't hurt ;))
3. Check all four user types
4. Click `Save`

This should ensure that you always see the task completion survey whether logged in or not.

Go to any documentation page. The survey should show up in the top right corner of the page.

PR #5701, opened 2019-08-27
Issue #5685, remove Emotion and move CSS to Sass - Document and Sidebar

schalkneethling:
@peterbe r?

PR #5709, opened 2019-08-27
Issue #5685, remove Emotion and move CSS to Sass - titlebar

schalkneethling:
@peterbe r?

### DIFF
--- a/kuma/javascript/src/titlebar.jsx
+++ b/kuma/javascript/src/titlebar.jsx
@@ -1,58 +1,12 @@
 // @flow
 import * as React from 'react';
 import { useContext } from 'react';
-import { css } from '@emotion/core';
 
 import { gettext } from './l10n.js';
 import EditIcon from './icons/pencil.svg';
 import UserProvider from './user-provider.jsx';
 
 import type { DocumentData, DocumentProps } from './document.jsx';
-
-// TODO: define this in a global styles.js file
-// A media query that identifies screens narrower than a tablet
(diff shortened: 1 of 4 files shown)


Gregoor:
Some changes I'm seeing on my machine:
`.titlebar-container { background-color` from `blue` to `grey`
`h1 { font-size` from `45.3px` to `51px`
Also the size changes to `25px` (instead of `28px`) on small devices. And the breakpoint for that is now ~tablet, where it is ~phone on the prod beta site.

schalkneethling:
> Some changes I'm seeing on my machine:
> `.titlebar-container { background-color` from `blue` to `grey`
> `h1 { font-size` from `45.3px` to `51px`
> Also the size changes to `25px` (instead of `28px`) on small devices. And the breakpoint for that is now ~tablet, where it is ~phone on the prod beta site.

Thank you for the review @Gregoor - These changes are intentional. One of the downsides of having this decoupled from the Sass codebase was that a lot of values was not using the defined variables we have for font sizes, colors margins, padding, breakpoints etc. It was also not following a mobile-first approach.

You will, therefore, see a number of places in these pull requests where adjustments like these have been made.

schalkneethling:
Also, just a quick note. It will be best if we work through these in the order of the pull requests. I believe #5700 has been merged so, next in line should be #5701, then #5702, etc. Thanks!

Gregoor:
Great, thanks for the elaboration. Then it looks good to me! 🚀

PR #5715, opened 2019-08-27
chore(editorconfig): Fix `.jsx` indentation configuration

ExE-Boss:
**Q:** Should I set `indent_style` and `indent_size` for all files? And if so, what should I set it to?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,6 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 
 # Indentiation
-[*.{css,html,js,json,py,styl}]
+[*.{css,html,js,jsx,json,py,styl}]
 indent_style = space
 indent_size = 4


peterbe:
> **Q:** Should I set `indent_style` and `indent_size` for all files? And if so, what should I set it to?

Is it not already so?

Because I don't think I understand the question, I'll hold off on merging this.

ExE-Boss:
@peterbe
> > **Q:** Should I set `indent_style` and `indent_size` for all files? And if so, what should I set it to?
>
> Is it not already so?

No, it’s not:
https://github.com/mdn/kuma/blob/ca5062683e96f0782c5e7b3feaf7708a98574f87/.editorconfig#L4-L9

That just sets line ending behaviour.

schalkneethling:
@ExE-Boss @peterbe Do we even need this now that we have standardized on Prettier for this project?

ExE-Boss:
Yes, as Prettier (and **VS Code**) uses `.editorconfig` for indentation information.

peterbe:
> Yes, as **Prettier** (and **VS Code**) uses `.editorconfig` for indentation information.

It *can*. 

The two are somewhat different in that `.editorconfig` helps the *editor* how it should do stuff. But `prettier` in the final authority.

Issue #5718, opened 2019-08-28
Add section links without the need for JavaScript


schalkneethling:
From some testing, it would seem that the JavaScript DOM manipulation to add section links is causing layout thrashing. Could there be a way to process the Article contents and inject these when the document is being server-side rendered?

atopal:
I'd also be okay with removing these now that we have a table of contents again for all pages or alternatively, instrument this to see how many people are actually using it before we spend time optimizing (and we should probably instrument it either way).

peterbe:
> Could there be a way to process the Article contents and inject these when the document is being server-side rendered?

Yes, let me attempt this. Entirely in Python. If I get it right none of the (S)CSS should be affected.

schalkneethling:
Pull request(https://github.com/mozilla/kuma/pull/5719) started a nice discussion and the conclusion is: Do it on the back-backend

schalkneethling:
@atopal When this is done in Python and we have no JS anymore related to this, would you still want us to add JS to instrument or should we just safely assume people like this feature?

atopal:
We should still instrument it, but I'd rather do that before we do the work here to re-implement this feature.

peterbe:
@atopal Can you instrument right-clicks of people selecting to "Copy link location..."? 
I'd personally hate to see section anchor links go away. They are so useful. And the way they're currently implemented, pardon my French, suck. I'm close to a resolution already that will take a good chunk out of the DOM CPU bound work that clearly is killing our performance.

atopal:
@peterbe yeah, I think you can use onContextMenu for that. The purpose of instrumenting this was to not spend time optimizing it, in case we don't need to. Instrumenting is quick and we can get answer within a day. 

The reason we added this feature was because we had removed the TOC. Now that we have a TOC again, it's nice, but not a priority. You can right click the TOC to get the same result.

peterbe:
@escattone Note that https://github.com/mozilla/kuma/pull/5728 doesn't automatically resolve this. We'll still have a bunch of JavaScript to inject anchor links. Both in the wiki and in the React side. The reason is that on the being anonymous on the Wiki (if you can ever be anonymous any more!) or browsing the read-only site it relies on `Document.body_html` statically. That field only gets regenerated if you edit the document. 

We *could* run a massive re-render and once we've done that, return to this issue and make another PR that deletes all those JavaScript codes that adds anchor links by mutating the DOM. Thoughts?

peterbe:
[...]

PR #5728, opened 2019-08-29
Add section links without the need for JavaScript

peterbe:
Part of #5718

Weird to ask for a review and mark it as "not ready" at the same time. 

Here's why it's not ready; now that the `<a class="section-link">` tags are injected into the body HTML, the two pieces of client-side JavaScript that injects anchor links "conflict". Turns out, the `local-anchor.js` implements it differently from how `addAnchor` in `articles.jsx` did it. 
Screen Shot 2019-08-29 at 1 20 33 PM
This is what happens now when viewing in the Wiki. Schalk, can you do some magic to the Wiki's (S)CSS so that it works the same as the read-only site. 

Meanwhile, @escattone Can you take a look at the html5lib stuff?


schalkneethling:
> This is what happens now when viewing in the Wiki. Schalk, can you do some magic to the Wiki's (S)CSS so that it works the same as the read-only site.

Oh wait, will still need the JS after this, or not at all? If not, then we can merge this and I can update the CSS on the wiki side, open a PR, and merge that before we do another push. If we will still require the JS for a period(I remember seeing something about having to re-render all the pages and issues with logged-in vs. non-logged-in users), then I will also need to update `local-anchor.js` to inject the bits in the same way as on the new frontend.

Let me know. Thanks!

peterbe:
@schalkneethling Yes, we'll need the JS for a while. Until at least most of the pages have been re-rendered. But the good news is that the JS won't execute as much as it did before and most importantly, on pages that have been re-rendered, the JS will load but it won't mutate the DOM. 

As of this PR, the HTML is now always the same. I.e. 

```html
<h2 id="foo">Stuff <a href="#foo" class="section-link"><svg/></a></h2>
<p>...and...</p>
<h3 id="bar">Things</h2><a href="#foo" class="section-link"><svg/></a>
```
The beta site's (S)CSS can cope with this and the anchor link appears to the right. 
But the Wiki site's CSS can't cope with the `<a>` appearing *after* the `</h3>`. 

Why we have to treat them differently like this I don't know.

schalkneethling:
> Why we have to treat them differently like this I don't know.

The change came in when I removed the need for the JS that wraps the content of some headings in `span` elements to do the black background styling. I only did that for the new frontend and thus, the wiki frontend still assumes the old layout of the elements.

peterbe:
I'm losing faith in this one. In particular about the use of an inline `<svg>` tag. On http://beta.mdn.localhost:8000/en-US/docs/Web/HTML there are 5 `h2|h3`  tags. That means, the whole HTML contains that long SVG string 10 times! 5 times in the regular HTML and 5 times in the `_react_data` because it represents the `bodyHTML` which has those tags in it. 
That's roughly 14KB extra! Just for those little anchor links. 

(Yes, since they repeat the compress very well under gzip, I know, but more stuff for the browser to worry about)

I know that GitHub and npmjs.com server-side render their anchor links with a very similar look and they too use `<svg>`. 

How about this for a two-fold idea?

1) On GitHub and npmjs.com those little anchors don't appear until you mouse hover over them. On desktop. But on mobile they're visible immediately. 90% of MDN is Desktop users and you can make it so that the `h2` itself is clickable. So if you do, click the heading, the URL bar changes to `{currentURL}#clicked_id` which means that, on mobile, you can click the little Share icon to send a friend a link directly to a heading. 
If they only appear onHover, we can lazy-load them into the DOM. 

2) Replace the `<svg>` with (S)CSS. I know you can't style them, but they just always need to be dark blue and small. I'd save a lot!

peterbe:
More ideas...

1) Give up on this functionality and delete the whole concept. 

2) Instead of SSR or DOM mutations, just turn all the `<h2|h3>` headings into links. E.g. `<h2 id="see_also"><a href="#see_also">See also</a></h2>`. That would preserve the option to right-click and share a link to a heading. It would also remove any runtime DOM mutations and save a crap-tonne of HTML strings on the initial payload.

schalkneethling:
> 2\. Replace the `<svg>` with (S)CSS. I know you can't style them, but they just always need to be dark blue and small. I'd save a lot!

In this case, it is definitely an option to change the SVG to a background image. All that we will have to change is the following:

```
<h2 id="foo">Stuff <a href="#foo" class="section-link"><span class="offscreen">Link to Stuff</span></a></h2>
<p>...and...</p>
<h3 id="bar">Things</h2><a href="#foo" class="section-link"><span class="offscreen">Link to Things</span></a>
```

This is to ensure it remains accessible.

peterbe:
If we go ahead with this PR, and ignore the detrimental side-effects (emphasized by React hydration data serializing) I think it's still worth it. But if I replace the `<svg>` for the tag `<span class="offscreen">Link to Stuff</span>` will we not need to add some SCSS to do the `background-image:url(/path/to/anchor.svg)`?

schalkneethling:
> If we go ahead with this PR, and ignore the detrimental side-effects (emphasized by React hydration data serializing) I think it's still worth it. But if I replace the `<svg>` for the tag `<span class="offscreen">Link to Stuff</span>` will we not need to add some SCSS to do the `background-image:url(/path/to/anchor.svg)`?

Oh, we will definitely need that as well as some other tweaks on wiki side, but it is doable.

peterbe:
In https://github.com/mozilla/kuma/pull/5739 we (personally begrudgingly) removed the ability display the little section perma link anchors. That means that this PR is void. 

However, this PR has some nice bones to it. The anchor links are useful and it's something a lot of people expect to exist even if they don't use it very frequently. At least, the few times they need it, they *can* use the TOC to find the links. Or, they can view the HTML source of the page and figure out what the IDs are and manufacture the links manually from the HTML source. 

Let's close this one for now. If we want the feature back, we can re-open this PR, make the necessary fixes to the (S)CSS and let it be 100% server-side HTML + CSS.

PR #5739, opened 2019-09-03
Delete section anchor links

### LINKED ISSUE #5738: Delete section anchor links
They are nice but the way it's implemented it causes DOM mutations (an re-paints) for every single section header.  Let remove it. We can re-instate it some other day, ideally with server-side rendering.

peterbe:
Fixes #5738


schalkneethling:
@peterbe merge at will

PR #5747, opened 2019-09-04
Fixes #5746, adds hard-source-webpack-plugin

### LINKED ISSUE #5746: Add HardSource Webpack plugin
I read about the `hard-source-webpack-plugin` recently via Addy Osmani. I tried it locally and oh boy does it ever speed up module builds. I reckon we should add it to our dev tooling.

https://github.com/mzgoddard/hard-source-webpack-plugin

schalkneethling:
@Gregoor r?


webpack-build-utils/presets/webpack.hardsource.js
```diff
@@ -0,0 +1,6 @@
+const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
+
+// for more on this plugin see https://github.com/mzgoddard/hard-source-webpack-plugin
+module.exports = () => ({
+    plugins: [new HardSourceWebpackPlugin()]
```
Gregoor: Out of curiosity, why not put this in `webpack.development.js`? Or to take this to an extreme: do we want separate files for each plugin?

schalkneethling: It is a nice way to compose things into different script targets/environments. Also, this idea is something that will be coming to Webpack natively soonish https://gist.github.com/TheLarkInn/684bcdb5cf3a082c5a3cee907168c59a

It also makes it super simple to remove a specific plugin either entirely or only from some environment targets. It keeps the configs cleaner which is a blessing in itself ;p Once we know something is going to stick around for the long haul, we might decide to move it directly into the main dev config.

Gregoor: Cool, I like that two stage model 👍 Thanks for the elaboration!

PR #5756, opened 2019-09-05
5619 jsonparse  react data

peterbe:
I haven't done any Lighthouse testing locally to see if this makes a big difference. I know it's sane in principle. 

At first, I tried to do ALL of this in Python but after a lot of searching I couldn't find a library that not only JSON serializes but also one that escapes *everything else* that might appear in the string. At first I tried something like this:
```
templ = "<script>window._react_data = JSON.parse('{}');</script>"
return templ.format(
    json.dumps(data).replace('</', '<\\/').replace("'", "\\'") etc etc etc
)
```
But everything I tried just failed. There was always something weird. Especially when involving Unicode. It was just one problem after another. I tried searching for some decent Python libs that does what the npm package `jsesc` does but I couldn't find anything. And I really trust `jsesc` because ita author is really good at this stuff. 

So basically, what this PR does is that it moves the work of serializing the string over the Node instead of Python.

### DIFF
--- a/kuma/javascript/src/ssr.jsx
+++ b/kuma/javascript/src/ssr.jsx
@@ -1,10 +1,22 @@
 import React from 'react';
 import { renderToString } from 'react-dom/server';
-
+import jsesc from 'jsesc';
 import SinglePageApp from './single-page-app.jsx';
 import LandingPage from './landing-page.jsx';
 import { localize } from './l10n.js';
 
+/**
+ * Inspired by
+ * https://joreteg.com/blog/improving-redux-state-transfer-performance
+ * This function produces a string that you can inject into an HTML document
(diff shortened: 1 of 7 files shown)


peterbe:
>  thinning react_data

https://github.com/mozilla/kuma/compare/mozilla:master...peterbe:make-up-_react_data?expand=1

Gregoor:
> 
> 
> > thinning react_data
> 
> https://github.com/mozilla/kuma/compare/mozilla:master...peterbe:make-up-_react_data?expand=1

way ahead of me 🙈

peterbe:
The other option is to consider that hacky trick of using `dangeroulsyInsertSetHTML` with an empty string. But I don't know what that means to hydrating and hooking up to event handlers. 

I know that if you run my branch there, you'll see that it totally solves https://github.com/mozilla/kuma/issues/5613

PR #5757, opened 2019-09-06
Don't serialize what can be extacted from DOM from SSR

### LINKED ISSUE #5613: Beta site HTML is larger

peterbe:
Fixes #5613

An experiment that @Gregoor and I are playing with.

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -21,9 +21,14 @@ if (container) {
     if (quickLinksHTMLElement) {
         data.documentData.quickLinksHTML = quickLinksHTMLElement.innerHTML;
     }
-    let wikiArticleElement = document.querySelector('#wikiArticle')
+    let wikiArticleElement = document.querySelector('#wikiArticle');
     if (wikiArticleElement) {
         data.documentData.bodyHTML = wikiArticleElement.innerHTML;
+    }
+
+    let tocElement = document.querySelector('.document-toc ul');
+    if (tocElement) {
+        data.documentData.tocHTML = tocElement.innerHTML;
     }
 
     // Store the string catalog so that l10n.gettext() can do translations
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -17,9 +17,14 @@ if (container) {
     // with all the data needed to hydrate or render the UI.
     let data = window._react_data;
 
-    // Remove the global reference to this data object so that it can
-    // be garbage collected once it is no longer in use.
-    window._react_data = null; // eslint-disable-line camelcase
+    let quickLinksHTMLElement = document.querySelector('.sidebar .quick-links');
+    if (quickLinksHTMLElement) {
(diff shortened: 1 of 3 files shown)


peterbe:
Going with https://github.com/mozilla/kuma/pull/5771 instead.

peterbe:
Actually, this one does work. At least in theory. 

```
-rw-r--r--   1 peterbe  wheel    29K Oct  4 16:41 after.html.gz
-rw-r--r--   1 peterbe  wheel   115K Oct  4 16:41 after.html
-rw-r--r--   1 peterbe  wheel    39K Oct  4 16:43 mdn.html.gz
-rw-r--r--   1 peterbe  wheel   161K Oct  4 16:43 mdn.html
-rw-r--r--   1 peterbe  wheel    48K Oct  4 16:59 before.html.gz
-rw-r--r--   1 peterbe  wheel   202K Oct  4 16:59 before.html
```
(the `mdn.html` and `mdn.html.gz` is from developer.mozilla.org and it's run in not-DEBUG mode)

Basically, this PR saves 87KB on the size of the HTML. 
But comparing the gzipped versions it's only a difference of 19KB. 

But perhaps I'm just measuring things wrong but what I am seeing is that the `<Article/>` component renders more times. I don't know why because the `document` prop isn't different. 
I put in `console.log("RENDERING Article", new Date());` before the `return` inside `article.jsx` and noticed that it ran differently. Not sure if it's user error or an actual subtle effect of the hydration. 

So is it worth the risk for 19KB? I'm starting to doubt that. What do you think @Gregoor ?

The thing is that Kuma sucks in the sense that it's hard to measure these things. We don't have it yet in Stumptown-renderer but we will in a near future; the ability to make two versions of the same site in equal contexts. Then you can put both versions into a thing like Lighthouse or SpeedCurve and get real data.

Gregoor:
On average 3G speeds that'd be more than a quarter of a second, no? So that's imo significant enough to do it.

tobinmori:
@peterbe is there a user story for this one? looks like #5613 is closed.

peterbe:
> @peterbe is there a user story for this one? looks like #5613 is closed.

Yes, that was the user story. But that belonged to another user story which was that the React site was slow performing so we rushed at it with various little tricks to get it to be as fast as the Jinja based one. 

Generally, this PR is 90% of the way there but the last 10% is fraught with trouble. For one, we'rel not desperate for speed anymore and we can't decide if this is any good anyway (without a platform for experimentation).

PR #5765, opened 2019-09-08
Fixes #5641, convert newsletter to react

### LINKED ISSUE #5641: Convert newsletter signup form to React Land
While working on converting the [footer for docs pages](https://github.com/mozilla/kuma/issues/5640) to React I realized that we first need to convert the newsletter signup form.

This will only be applied to the `docs` pages and not the landing page currently. The component should be coded in a way that will allow us to reuse it easily in different places such as on the landing page, and later on the profile pages.

schalkneethling:
This code converts the newsletter sign-up form to a React component for the documentation pages(not the landing page).

### Scenario 1

1. Load up http://beta.mdn.localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
2. Enter an email address
3. When this field receives focus,
3.1 ensure that the privacy policy field is made visible
4. Check the privacy policy checkbox
5. Click on `Sign up now`

The form should be hidden and a `Thank you` message shown.

6. Refresh the page
7. The form should not be displayed

### Scenario 2

1. Load up http://beta.mdn.localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
2. Enter an email address
3. When this field receives focus,
3.1 ensure that the privacy policy field is made visible
4. Click on `Sign up now`

A message should be shown to the user as follows:
```
You must agree to the privacy notice
```

### Scenario 3

1. Load up http://beta.mdn.localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
2. Focus the email address field(do not enter an email address)
3. When this field receives focus,
3.1 ensure that the privacy policy field is made visible
4. Click on `Sign up now`

The following messages should be shown to the user:
```
Please enter a valid email address
You must agree to the privacy notice
```

### Scenario 4

1. Load up http://beta.mdn.localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
2. Click the close icon top right
3. The form should be hidden
4. Refresh the page
5. The form should not be displayed

### Scenario 5

1. Load up http://beta.mdn.localhost:8000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
2. Enter an email address
3. When this field receives focus,
3.1 ensure that the privacy policy field is made visible
4. Check the privacy policy checkbox
5. Open up the devtools
6. Find the form with `id=newsletter-form`
7. Add the following attribute to the form `data-skip-xhr`
8. Press enter
9. Click on `Sign up now`

Instead of submitting the form via Ajax, it should send you directly to https://www.mozilla.org/en-US/newsletter/

## Desktop display

Screenshot 2019-09-08 at 17 19 15

## Tablet and Mobile display

Screenshot 2019-09-08 at 17 19 29


kuma/javascript/src/utils.js
```diff
@@ -20,3 +20,60 @@ export function getCookie(name) {
...
+ * Example requestHeaders
+ * -----------------------
+ * {
+ *   'X-Requested-With': 'XMLHttpRequest',
+ *   'Content-type': 'application/x-www-form-urlencoded'
+ * }
+ *
+ * @returns new XMLHttpRequest object
+ */
+export function initAjaxRequest(method, url, requestHeaders) {
```
Gregoor: looks like we're using `fetch` in other parts of the codebase, which imo has a saner API, so I'd vote going with that instead. If older browser support is the issue, I'd vote for loading a polyfill on those (relatively few?) devices.

kuma/javascript/src/newsletter.jsx
```diff
@@ -0,0 +1,380 @@
...
+        let documentLocation = encodeURIComponent(document.location.href);
+        let emailInputValue =
+            emailInput && emailInput.value
+                ? encodeURIComponent(emailInput.value)
+                : '';
+        let params = `email=${emailInputValue}&newsletters=${newsletterType}${privacy}&fmt=${newsletterFormat}&source_url=${documentLocation}`;
+
+        ajaxRequest.send(params);
+
+        getAjaxResponse(ajaxRequest).then(response => {
```
Gregoor: Can we use `async/await`? The function complexity/nesting seems already a bit high, that'd help a lil

kuma/javascript/src/newsletter.jsx
```diff
@@ -0,0 +1,380 @@
...
+            let parsedResponse = JSON.parse(response);
+            if (parsedResponse.success !== 'success') {
+                let errorsArray = parsedResponse.errors;
+
+                // if there are erros and the array contain at least one item
+                if (errorsArray && errorsArray.length) {
+                    // show the errors
+                    let errorListContainer = errorsRef.current;
+
+                    if (errorListContainer) {
```
Gregoor: Can we go full-React here? 🙃 

Aka `setErrors(errorsArray)` and then later in render:
```jsx
<ul className="errorlist">{errorsArray.map(error => <li>yadayada
```
I think that'd make it more clear.

Gregoor:
No complaints from looking at it in action!

A few more code nits:
-  `saveNewsletterVisibleState`: I'd rename it to `permanentlyHideNewsletter` or sth similar
- we could cut down on some lines by constructing part of the query string through: `new URLSearchParams(new FormData(formRef.current)).toString()`

schalkneethling:
Thanks for the review @Gregoor - updated.

kuma/javascript/src/newsletter.jsx
```diff
@@ -0,0 +1,330 @@
...
+            error
+        );
+    }
+}
+
+export default function Newsletter() {
+    const emailRef = useRef(null);
+    const errorsRef = useRef(null);
+    const newsletterFormRef = useRef(null);
+    const privacyCheckboxRef = useRef(null);
```
Gregoor: this and the `emailRef` are now not needed anymore 🙂

PR #5766, opened 2019-09-08
Fixes 5640, render footer in react for docs pages

### LINKED ISSUE #5640: Move footer on document pages away from Jinja
Currently, the footer on the static site is still defined in the Jinja templates. While we cannot yet do this for the homepage(because it still contains the footer lang switcher), we can do this for the documentation pages. Seems like a worthy task.

https://bugzilla.mozilla.org/show_bug.cgi?id=1570040

schalkneethling:
Fixes #5640

This converts the footer to a React component for docs pages. This one does not depend on https://github.com/mozilla/kuma/pull/5765, but it would be best to review https://github.com/mozilla/kuma/pull/5765 before this one.

Both this PR and https://github.com/mozilla/kuma/pull/5765 should be merged before the next push to production.


kuma/javascript/src/footer.jsx
```diff
@@ -0,0 +1,118 @@
...
+                        <li>
+                            <a href="{{ wiki_url('Learn') }}">
+                                {gettext('Learn Web Development')}
+                            </a>
+                        </li>
+                        <li>
+                            <a href="/docs/MDN/About">{gettext('About MDN')}</a>
+                        </li>
+                        <li>
+                            <a href="{{ wiki_url('MDN/Feedback') }}">
```
Gregoor: I don't think this would work in a stable way. It might be replaced for SSR, but as soon as the component rerenders, we'd have this jinja tag as the href value.

schalkneethling: Whoops! My bad, overlooked.

schalkneethling:
Thanks for the review @Gregoor - updated

PR #5772, opened 2019-09-09
React newsletter fixes & refactor

### LINKED ISSUE #5782: React Newsletter is changing state during rendering
In https://sentry.prod.mozaws.net/operations/mdn-stage/issues/6247145/?query=is:unresolved we got an 500 error from `ssr-server.js`. But that's the underlying problem. The underlying problem was this:

```
Error: Minified React error #301; visit https://reactjs.org/docs/error-decoder.html?invariant=301 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Z (/app/kuma/javascript/dist/ssr.js:22:7658)
    at he (/app/kuma/javascript/dist/ssr.js:22:52879)
    at X (/app/kuma/javascript/dist/ssr.js:22:7139)
    at a (/app/kuma/javascript/dist/ssr.js:22:10878)
    at we (/app/kuma/javascript/dist/ssr.js:22:11874)
    at e.render (/app/kuma/javascript/dist/ssr.js:22:14460)
    at e.read (/app/kuma/javascript/dist/ssr.js:22:14084)
    at renderToString (/app/kuma/javascript/dist/ssr.js:22:19625)
    at gt (/app/kuma/javascript/dist/ssr.js:22:84247)
    at app.post (/app/kuma/javascript/ssr-server.js:68:14)
```

This happened because, during rendering, [state was changed](https://github.com/mozilla/kuma/blob/e1873a8914e1835e6c74774d6c764408fdf9e72d/kuma/javascript/src/newsletter.jsx#L159) which caused an infinite loop. 

The simplest solution is to fix `newsletter.jsx` to not change state during rendering but perhaps the best place to start is with the right tooling to prevent this from happening in the first place. 

Are there `eslint` plugins that catch state changes, in React hooks, during rendering? 

The error slipped through because both reviewer and test automation didn't catch it. 

I changed the code to:
```javascript
if (locale !== 'en-US') {
        throw new Error("BAL BLAB")
        // setShowNewsletterLang(true);
    }
```
and ran `npx jest` and it didn't fail. So, first of all we need much better test coverage on the `Newsletter` component. 

@Gregoor has [a PR](https://github.com/mozilla/kuma/pull/5772) that refactors `newsletter.jsx` but [it still has the bug](https://github.com/mozilla/kuma/blob/0533ef8d87acb3040e9221853c942820d41e59f6/kuma/javascript/src/newsletter.jsx#L154).

Gregoor:
Some of this might be opinionated, as in: I like...
- ...early returns
- ...destructuring

Happy to revert if that seems less readable to anyone.

There were some checks in there for whether `newsletterForm` exists, I dropped those because `submit` could only be called after the component was rendered at least once. And since the form isn't rendered conditionally `newsletterForm` should always exist. Unless I'm missing something... 🙃 

Fixes #5782

### DIFF
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -126,7 +126,15 @@ export default function Newsletter() {
         })
             .then(response => response.json())
             .then(({ success, errors }) => {
-                if (success === 'success') {
+                if (success) {
+                    permanentlyHideNewsletter();
+                    setShowSuccessfulSubscription(true);
+                    ga('send', {
+                        hitType: 'event',
+                        eventCategory: 'newsletter',
+                        eventAction: 'progression',
+                        eventLabel: 'complete'
+                    });
                     return;
                 }
 
@@ -138,16 +146,6 @@ export default function Newsletter() {
                     newsletterForm.dataset.skipFetch = 'true';
                     // and submit again for diagnoses on the server
                     newsletterForm.submit();
-                } else {
-                    setShowNewsletter(false);
-                    permanentlyHideNewsletter();
-                    setShowSuccessfulSubscription(true);
-                    ga('send', {
-                        hitType: 'event',
-                        eventCategory: 'newsletter',
-                        eventAction: 'progression',
-                        eventLabel: 'complete'
-                    });
                 }
             })
             .catch(e => {
@@ -166,6 +164,25 @@ export default function Newsletter() {
 
     if (!showNewsletter) {
         return null;
+    }
+
+    if (showSuccessfulSubscription) {
+        return (
+            <section className="newsletter-container">
+                <div id="newsletter-thanks" className="newsletter-thanks">
+                    <h2>
+                        {gettext(
+                            'Thanks! Please check your inbox to confirm your subscription.'
+                        )}
+                    </h2>
+                    <p>
+                        {gettext(
+                            'If you haven’t previously confirmed a subscription to a Mozilla - related newsletter you may have to do so. Please check your inbox or your spam filter for an email from us.'
+                        )}
+                    </p>
+                </div>
+            </section>
+        );
     }
 
     return (
@@ -300,24 +317,6 @@ export default function Newsletter() {
                     <span>{gettext('Hide Newsletter Sign-up')}</span>
                     <CloseIcon />
                 </button>
-            </div>
-            <div
-                id="newsletter-thanks"
-                className={
-                    showSuccessfulSubscription ? 'newsletter-thanks' : 'hidden'
-                }
-                aria-hidden={showSuccessfulSubscription ? false : true}
-            >
-                <h2>
-                    {gettext(
-                        'Thanks! Please check your inbox to confirm your subscription.'
-                    )}
-                </h2>
-                <p>
-                    {gettext(
-                        'If you haven’t previously confirmed a subscription to a Mozilla - related newsletter you may have to do so. Please check your inbox or your spam filter for an email from us.'
-                    )}
-                </p>
             </div>
         </section>
     );
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -94,7 +94,6 @@ export default function Newsletter() {
             return;
         }
 
-        // if skipFetch
         if (newsletterForm.dataset['skipFetch']) {
             /*
              * An error occured while attempting to subscribe
@@ -139,11 +138,10 @@ export default function Newsletter() {
 
                 if (errors && errors.length) {
                     setError(errors);
-                    // if there was an error, but there are no items in the array
                 } else if (errors && errors.length === 0) {
-                    // set the skip-fetch data attribute on the form
+                    // resubmit for diagnoses on the server (see skipFetch-
+                    // comment above)
                     newsletterForm.dataset.skipFetch = 'true';
-                    // and submit again for diagnoses on the server
                     newsletterForm.submit();
                 }
             })
@@ -153,9 +151,7 @@ export default function Newsletter() {
     };
 
     useEffect(() => {
-        // if `newsletterHide` is set
         if (localStorage.getItem('newsletterHide') === 'true') {
-            // do not show the newsletter
             setShowNewsletter(false);
         }
     }, []);
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -32,9 +32,10 @@ export default function Newsletter() {
     const newsletterType = 'app-dev';
     const newsletterFormat = 'H';
 
-    const [errors, setError] = useState([]);
     const [showNewsletter, setShowNewsletter] = useState(true);
     const [showPrivacyCheckbox, setShowPrivacyCheckbox] = useState(false);
+    const [submitToServer, setSubmitToServer] = useState(false);
+    const [errors, setError] = useState([]);
     const [
         showSuccessfulSubscription,
         setShowSuccessfulSubscription
@@ -93,7 +94,7 @@ export default function Newsletter() {
             return;
         }
 
-        if (newsletterForm.dataset['skipFetch']) {
+        if (submitToServer) {
             /*
              * An error occured while attempting to subscribe
              * the user via Ajax, but no specific error was provided. We
@@ -140,8 +141,7 @@ export default function Newsletter() {
                 } else if (errors && errors.length === 0) {
                     // resubmit for diagnoses on the server (see skipFetch-
                     // comment above)
-                    newsletterForm.dataset.skipFetch = 'true';
-                    newsletterForm.submit();
+                    setSubmitToServer(true);
                 }
             })
             .catch(e => {
@@ -154,6 +154,14 @@ export default function Newsletter() {
             setShowNewsletter(false);
         }
     }, []);
+
+    useEffect(() => {
+        const newsletterForm = newsletterFormRef.current;
+        if (submitToServer && newsletterForm) {
+            setSubmitToServer(false);
+            newsletterForm.submit();
+        }
+    }, [submitToServer]);
 
     if (!showNewsletter) {
         return null;
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -164,13 +164,13 @@ export default function Newsletter() {
         }
     }, []);
 
+    if (!showNewsletter) {
+        return null;
+    }
+
     return (
         <section className="newsletter-container">
-            <div
-                id="newsletter-form-container"
-                className={showNewsletter ? 'newsletter' : 'hidden'}
-                aria-hidden={showNewsletter ? false : true}
-            >
+            <div id="newsletter-form-container" className="newsletter">
                 <form
                     ref={newsletterFormRef}
                     id="newsletter-form"
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -23,7 +23,6 @@ function permanentlyHideNewsletter() {
 }
 
 export default function Newsletter() {
-    const errorsRef = useRef(null);
     const newsletterFormRef = useRef(null);
 
     const ga = useContext(GAProvider.context);
@@ -169,7 +168,7 @@ export default function Newsletter() {
     if (showSuccessfulSubscription) {
         return (
             <section className="newsletter-container">
-                <div id="newsletter-thanks" className="newsletter-thanks">
+                <div className="newsletter-thanks">
                     <h2>
                         {gettext(
                             'Thanks! Please check your inbox to confirm your subscription.'
@@ -190,7 +189,6 @@ export default function Newsletter() {
             <div id="newsletter-form-container" className="newsletter">
                 <form
                     ref={newsletterFormRef}
-                    id="newsletter-form"
                     className="newsletter-form nodisable"
                     name="newsletter-form"
                     action={newsletterSubscribeURL}
@@ -221,37 +219,27 @@ export default function Newsletter() {
                     <fieldset className="newsletter-fields">
                         <input
                             type="hidden"
-                            id="fmt"
                             name="fmt"
                             value={newsletterFormat}
                         />
                         <input
                             type="hidden"
-                            id="newsletterNewslettersInput"
                             name="newsletters"
                             value={newsletterType}
                         />
-                        <div
-                            ref={errorsRef}
-                            id="newsletter-errors"
-                            className={
-                                hasErrors ? 'newsletter-errors' : 'hidden'
-                            }
-                            aria-hidden={hasErrors}
-                        >
-                            <ul className="errorlist">
-                                {errors.map((error, index) => (
-                                    <li key={index}>{error}</li>
-                                ))}
-                            </ul>
-                        </div>
+                        {errors.length > 0 && (
+                            <div className="newsletter-errors">
+                                <ul className="errorlist">
+                                    {errors.map((error, index) => (
+                                        <li key={index}>{error}</li>
+                                    ))}
+                                </ul>
+                            </div>
+                        )}
 
-                        <div
-                            id="newsletterEmail"
-                            className="form-group newsletter-group-email"
-                        >
+                        <div className="form-group newsletter-group-email">
                             <label
-                                htmlFor="newsletterEmailInput"
+                                htmlFor="newsletter-email-input"
                                 className="form-label offscreen"
                             >
                                 {gettext('E-mail')}
@@ -292,10 +280,7 @@ export default function Newsletter() {
                                 .
                             </label>
                         </div>
-                        <div
-                            id="newsletterSubmit"
-                            className="newsletter-group-submit"
-                        >
+                        <div className="newsletter-group-submit">
                             <button
                                 onClick={submit}
                                 id="newsletter-submit"
@@ -309,7 +294,6 @@ export default function Newsletter() {
                 </form>
                 <button
                     onClick={closeNewsletter}
-                    id="newsletter-hide"
                     type="button"
                     className="only-icon newsletter-hide"
                     aria-controls="newsletter-form-container"
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -23,10 +23,8 @@ function permanentlyHideNewsletter() {
 }
 
 export default function Newsletter() {
-    const emailRef = useRef(null);
     const errorsRef = useRef(null);
     const newsletterFormRef = useRef(null);
-    const privacyCheckboxRef = useRef(null);
 
     const ga = useContext(GAProvider.context);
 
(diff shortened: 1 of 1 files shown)


schalkneethling:
> There were some checks in there for whether `newsletterForm` exists, I dropped those because `submit` could only be called after the component was rendered at least once.

Yup, I believe the main reason I had that is that Flow gets angry ;) see:
https://travis-ci.org/mozilla/kuma/jobs/582738096

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,68 +87,65 @@ export default function Newsletter() {
...
-                    eventCategory: 'newsletter',
-                    eventAction: 'progression',
-                    eventLabel: 'error-forward'
-                });
-                return true;
-            }
-
-            event.preventDefault();
+        if (!newsletterForm) {
+            return;
```
peterbe: Won't this just mean that the `onSubmit` is still firing? I.e. don't you need to `event.preventDefault()` first?

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,68 +87,65 @@ export default function Newsletter() {
...
+        ).toString();
+        fetch(newsletterSubscribeURL, {
+            method: 'POST',
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+                'Content-type': 'application/x-www-form-urlencoded'
+            },
+            body: params
+        })
+            .then(response => response.json())
```
peterbe: The response isn't always JSON. If `!response.ok` this'll fail cryptically.

kuma/javascript/src/newsletter.jsx
Gregoor: This is there for type refinement reasons, meaning the body of that if should never be executed. Unfortunately afaik there's no way to express in our type system that newsletterForm won't ever be null after the first render (which is the first time submit could ever be called).

And if we'd move `preventDefault` up we'd also prevent form submission for the `skipFetch` case where we afaik explicitly want to have it.

kuma/javascript/src/newsletter.jsx
Gregoor: Good point, I've added an error handler

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +91,71 @@ export default function Newsletter() {
...
+            .then(({ success, errors }) => {
+                if (success === 'success') {
+                    return;
+                }
```
schalkneethling: It looks like this is the piece of code that will run if the subscription completed successfully. If so, there are a couple of things we need to do. Essentially everything from https://github.com/mozilla/kuma/pull/5772/files#diff-92ad4e492a736caacb5293381494f20eR142 - 150 needs to be moved inside this conditional.

At the moment, once the `fetch` completes successfully, the newsletter is hidden, but no message is shown to the user.

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +91,71 @@ export default function Newsletter() {
...
+                } else {
+                    setShowNewsletter(false);
+                    permanentlyHideNewsletter();
+                    setShowSuccessfulSubscription(true);
+                    ga('send', {
+                        hitType: 'event',
+                        eventCategory: 'newsletter',
+                        eventAction: 'progression',
+                        eventLabel: 'complete'
+                    });
```
schalkneethling: I believe this needs to be moved to the `success` conditional higher up.

kuma/javascript/src/newsletter.jsx
Gregoor: Thanks, indeed. Apparently I only read the code with half a mind. I made those changes.

Also I noticed that `response.success` is a boolean (which imo also makes more sense), so that block would've never run. Or is there a case in which `response.success` would be the `success` string?

kuma/javascript/src/newsletter.jsx
peterbe: That `.catch` will only happen if the request completely failed. For example, on network errors. The *communication* with the server might be successful but the response from the server might be `403 Forbidden` or `504 Bad Gateway`. That would not trigger a `.catch()` because `fetch` did as well as it possibly can. It's just a problem between the browser and the server. I often use this:

```javascript
let response
try {
   response = await fetch(url)
} catch(err) {
   return this.setState({networkError: err})
}
if (!response.ok) {
  return this.setState({serverError: response}) 
}
const data = await response.json();
return this.setState({ data });
```

kuma/javascript/src/newsletter.jsx
```diff
@@ -167,13 +162,32 @@ export default function Newsletter() {
...
+    }
+
     return (
         <section className="newsletter-container">
-            <div
-                id="newsletter-form-container"
-                className={showNewsletter ? 'newsletter' : 'hidden'}
-                aria-hidden={showNewsletter ? false : true}
-            >
+            <div id="newsletter-form-container" className="newsletter">
```
peterbe: Why bother with an ID?

kuma/javascript/src/newsletter.jsx
```diff
@@ -221,15 +235,14 @@ export default function Newsletter() {
...
-                            aria-hidden={errorsArray ? true : false}
+                            aria-hidden={hasErrors}
                         >
                             <ul className="errorlist">
-                                {errorsArray &&
-                                    errorsArray.map((error, index) => (
-                                        <li key={index}>{error}</li>
-                                    ))}
+                                {errors.map((error, index) => (
+                                    <li key={index}>{error}</li>
```
peterbe: Don't use indexes for keys. If the first render, the `errors` are:
```
['Foo', 'Bar']
```
It will produce:
```html
<li>Foo</li>
<li>Bar</li>
```
and then a form submission happens and now `errors` is: 
```
['One', 'Two', 'Three']
```
It will produce (in the second render):

```html
<li>Foo</li>
<li>Bar</li>
<li>Three</li>
```

kuma/javascript/src/newsletter.jsx
```diff
@@ -221,15 +235,14 @@ export default function Newsletter() {
                             ref={errorsRef}
                             id="newsletter-errors"
                             className={
-                                errorsArray ? 'newsletter-errors' : 'hidden'
+                                hasErrors ? 'newsletter-errors' : 'hidden'
```
peterbe: The "React way" is to not use CSS to show or hide things. I think I would do it like this:

```jsx
{hasErrors && (<div id="newsletter-errors" className="newsletter-errors">
    <ul className="errorlist">
   ...
</div>)}
```

By the way, I see a lot of redundancy. Why do we need both an ID and a class? And can't you get to the `ul` with CSS simply by doing `.newsletter-errors ul`?

kuma/javascript/src/newsletter.jsx
```diff
@@ -167,13 +162,32 @@ export default function Newsletter() {
...
     return (
         <section className="newsletter-container">
-            <div
-                id="newsletter-form-container"
-                className={showNewsletter ? 'newsletter' : 'hidden'}
-                aria-hidden={showNewsletter ? false : true}
-            >
+            <div id="newsletter-form-container" className="newsletter">
                 <form
                     ref={newsletterFormRef}
```
peterbe: I still can't see why you need a ref for this? 
Can't it be done with state within the component?

kuma/javascript/src/newsletter.jsx
Gregoor: True, didn't even question those. I threw all the IDs that didn't serve a purpose out.

kuma/javascript/src/newsletter.jsx
```diff
@@ -167,13 +162,32 @@ export default function Newsletter() {
...
     }, []);
 
+    if (!showNewsletter) {
+        return null;
+    }
+
+    if (showSuccessfulSubscription) {
+        return (
+            <section className="newsletter-container">
+                <div id="newsletter-thanks" className="newsletter-thanks">
```
peterbe: So many IDs.

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +91,69 @@ export default function Newsletter() {
...
+                } else if (errors && errors.length === 0) {
+                    // set the skip-fetch data attribute on the form
+                    newsletterForm.dataset.skipFetch = 'true';
+                    // and submit again for diagnoses on the server
+                    newsletterForm.submit();
+                }
+            })
+            .catch(e => {
+                console.error('error while subscribing to newsletter', e);
+                setError(['An error occurred while submitting the form']);
```
peterbe: I don't know what you need to pay Flow to make it work but I think pretty print the error *here* isn't ideal. 
I would do:
```javascript
setErrors([e])
```

And then, somewhere in the "display part" of the error I'd do something like this:
```javascript
{errors.map(error => {
  <li key={error.toString()}>
    {error instanceof Error ? gettext('Network error bla bla')  : error}
  </li>
})
```
or something fancier.

kuma/javascript/src/newsletter.jsx
Gregoor: True, also prefer the React way and hence changed it just now. Didn't touch the CSS so far because I didnt feel strongly about it yet

kuma/javascript/src/newsletter.jsx
Gregoor: I don't think it would. Keys don't prevent reconciliation (i.e. finding out what changed) but rather make sure React doesn't lose track of which component is which. But this is indeed a potential source of future bugs if it's ever not a simple `li` element anymore but rather something with state (which could then lead to a state mixup). I'll change the keys to `index + error` which I think makes the most sense here.

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +90,69 @@ export default function Newsletter() {
...
-                                eventAction: 'progression',
-                                eventLabel: 'complete'
-                            });
-                        }
-                    }
-                });
+                if (errors && errors.length) {
+                    setError(errors);
+                    // if there was an error, but there are no items in the array
+                } else if (errors && errors.length === 0) {
```
peterbe: I'm struggling with this logic. What can `errors` be? Is either `null`, or `[]` or `['stuff']`?
On line 128, it exits early if `success` is truthy. 
So here `errors` is expected to be an empty array. Are we sure it's always an array?

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +90,69 @@ export default function Newsletter() {
...
-                                eventCategory: 'newsletter',
-                                eventAction: 'progression',
-                                eventLabel: 'complete'
-                            });
-                        }
-                    }
-                });
+                if (errors && errors.length) {
+                    setError(errors);
+                    // if there was an error, but there are no items in the array
```
peterbe: What line does this code comment belong to?

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +90,69 @@ export default function Newsletter() {
...
+                } else if (errors && errors.length === 0) {
+                    // set the skip-fetch data attribute on the form
```
peterbe: That comment says what the code *does*. Not *why* it does it.

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +90,69 @@ export default function Newsletter() {
...
+                    newsletterForm.dataset.skipFetch = 'true';
+                    // and submit again for diagnoses on the server
+                    newsletterForm.submit();
```
peterbe: This feels clunky and I think you should consider avoiding the DOM API when it's not necessary. 
Make a closure function called something like...

```javascript
function postForm(form) {
    fetch(...).then(...).then(
         ...
         setSkipFetch(true)
         postForm(form)
    )
   
}
```
And on the form's `onSubmit` call that instead. E.g. 
```
<form onSubmit={event => {
  postForm(newsletterForm)
}}>
```

It just feels very jQuery to manually trigger things on the DOM instead of containing it inside actual JavaScript code.

kuma/javascript/src/newsletter.jsx
Gregoor: Threw together a fiddle, hope it has more explanatory than confusatory power:
https://codesandbox.io/s/naughty-banach-txlxc

kuma/javascript/src/newsletter.jsx
```diff
@@ -207,44 +219,32 @@ export default function Newsletter() {
...
-                            id="newsletter-errors"
-                            className={
-                                errorsArray ? 'newsletter-errors' : 'hidden'
-                            }
-                            aria-hidden={errorsArray ? true : false}
-                        >
-                            <ul className="errorlist">
-                                {errorsArray &&
-                                    errorsArray.map((error, index) => (
+                        {errors.length > 0 && (
```
peterbe: You can also use `{!!errors.length && ( ... )}` but either is fine.

kuma/javascript/src/newsletter.jsx
```diff
@@ -207,44 +219,32 @@ export default function Newsletter() {
...
+                            <div className="newsletter-errors">
+                                <ul className="errorlist">
+                                    {errors.map((error, index) => (
                                         <li key={index}>{error}</li>
```
peterbe: I don't know where my comment went but this is wrong. It's not correct to use array indices as the key. It can trigger that the node isn't updated in the virtual DOM if the *content* changes.

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,76 +90,69 @@ export default function Newsletter() {
...
     };
 
-    /* If this is not en-US, show message informing the user
-       that newsletter is only acailable in English */
-    if (locale !== 'en-US') {
-        setShowNewsletterLang(true);
-    }
-
     useEffect(() => {
         // if `newsletterHide` is set
```
peterbe: Incomplete or "useless" code comment.

kuma/javascript/src/newsletter.jsx
```diff
@@ -36,15 +33,19 @@ export default function Newsletter() {
     const newsletterSubscribeURL = 'https://www.mozilla.org/en-US/newsletter/';
     const privacyPolicyURL = 'https://www.mozilla.org/privacy/';
```
peterbe: I don't know why these two are variables? Are they repeated? Can they don't just be part of the code directly? Or, if they're genuine constants, you might want to use the `const UPPER_CASE_NOTATION` on these to send a message to the code-reader that they never going to change.

kuma/javascript/src/newsletter.jsx
schalkneethling: Nope, not from what I can see.

kuma/javascript/src/newsletter.jsx
schalkneethling: Perhaps there *is* a different way to get it. Right now it used to pass the form into `new FormData(newsletterFormRef.current)`

kuma/javascript/src/newsletter.jsx
schalkneethling: Yup. If it is successful, it only has `{ success: true }`. If there are errors, it will always be an array of errors or and empty array

kuma/javascript/src/newsletter.jsx
schalkneethling: No, I guess they do not have to be. They are only used in one spot each.

kuma/javascript/src/newsletter.jsx
schalkneethling: We should probably move the comment from here https://github.com/mozilla/kuma/pull/5772/files#diff-92ad4e492a736caacb5293381494f20eR99 to here.

kuma/javascript/src/newsletter.jsx
schalkneethling: See line 159 ;)

kuma/javascript/src/newsletter.jsx
Gregoor: We could throw the values into the state, though that'd also mean a re-render on every keystroke. Not expensive here, but avoidable and I don't think it'd add much in terms of clarity of the code (which might be my lack of imagination).

kuma/javascript/src/newsletter.jsx
Gregoor: Fair enough, I changed it to show the stringified error!

kuma/javascript/src/newsletter.jsx
Gregoor: true, I got rid of those that were a bit thin and referenced the longer explanation fro skip-fetch.

kuma/javascript/src/newsletter.jsx
Gregoor: It's up here (as well as my response):
https://github.com/mozilla/kuma/pull/5772#discussion_r322723286

kuma/javascript/src/newsletter.jsx
Gregoor: I think this might get a little awkward given how many state setters we'd have to pass in (or several callbacks). But I did take it as an incentive to look at extracting at least some of it and I think it does look better now.

kuma/javascript/src/newsletter.jsx
peterbe: Ah, the old controlled vs not controlled component debate. 

> In most cases, we recommend using controlled components to implement forms. 
https://reactjs.org/docs/uncontrolled-components.html

I personally prefer to control inputs unless in some rare cases where the input might come from the browser rather than the keyboard.

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,97 +101,99 @@ export default function Newsletter() {
...
+                if (errors && errors.length) {
+                    setError(errors);
+                } else if (errors && errors.length === 0) {
+                    // resubmit for diagnoses on the server (see skipFetch-
+                    // comment above)
+                    setSubmitToServer(true);
+                }
+            })
+            .catch(e => {
+                setError([e.toString()]);
```
peterbe: Ok. So I concede that this line would catch the error if `!response.ok`
https://codesandbox.io/embed/hardcore-dewdney-j6k98 
but it's just a bit weird that the result here will be about a problem with the JSON decoding. 

One neat compromise would be to change line 36 to:
```javascript
}).then(response => {
  if (!response.ok) {
    throw new Error(`Bad response from server: ${response.statusText}`);
  }
  return response.json()
});
```

peterbe: E.g. https://codesandbox.io/embed/peaceful-chatterjee-nwc7f

Either way, I'm now losing confidence that this is worth discussing. There are so many other flaws in Kuma other than the handling of some relatively rare error strings.

kuma/javascript/src/newsletter.jsx
```diff
@@ -207,44 +221,32 @@ export default function Newsletter() {
...
-                        >
-                            <ul className="errorlist">
-                                {errorsArray &&
-                                    errorsArray.map((error, index) => (
-                                        <li key={index}>{error}</li>
+                        {hasErrors && (
+                            <div className="newsletter-errors">
+                                <ul className="errorlist">
+                                    {errors.map((error, index) => (
+                                        <li key={index + error}>{error}</li>
```
peterbe: 👍

kuma/javascript/src/newsletter.jsx
```diff
@@ -89,97 +101,99 @@ export default function Newsletter() {
...
-                                eventAction: 'progression',
-                                eventLabel: 'complete'
-                            });
-                        }
-                    }
-                });
+                if (errors && errors.length) {
+                    setError(errors);
+                } else if (errors && errors.length === 0) {
+                    // resubmit for diagnoses on the server (see skipFetch-
```
schalkneethling: @Gregoor `skipFetch` is not mentioned anywhere anymore so, perhaps we should just drop that portion of the comment?

PR #5773, opened 2019-09-09
Removes merge conflict comments that was left in HTML file

schalkneethling:
@escattone r?


peterbe:
Haha! 
We could easily add a line to `.travis.yml` that barfs immediately if `git grep` finds any patch markers. 
Perhaps the most pressing problem is that tests didn't fail. But it'd be hard to spot actually because those a just strings to go into the jinja template.

PR #5780, opened 2019-09-09
diff never again

### LINKED ISSUE #5781: Leftovers from git patch catching
In https://github.com/mozilla/kuma/commit/235c264d55cf5719b2505d30cbe836215ef74782#diff-77138e0db7f64114bd7795f66f0bd9ba we accidentally checked in a merge conflict and nobody noticed. Neither human review or test automation. It's easy to happen especially if you get an "Approve but you need to rebase and fix so and so merge conflict" and in a rush accidentally try to ship it. 
If the file was a `.py` file, you'd quickly get a `SyntaxError` but who knows what other damage it can cause that doesn't break things immediately.

peterbe:
Fixes #5781

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
           cache: pip
           script:
             - set -x
+            # Make sure there are no left-over git patch markers in
+            - git --no-pager grep '>>>>>>>' && (echo "Some bad merge!" && exit 1)
             # Make sure these match what's in requirements/dev.txt!
             - pip install flake8==3.7.3 flake8-import-order==0.18
             - flake8 kuma docs tests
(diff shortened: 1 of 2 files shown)


callahad:
Just noticed that this does the opposite of what we want.

`git grep` exits with `0` when it _does_ find matches, so this only passes when there _are_ conflict markers in our repo.

So where is it finding a conflict marker? In `travis.yml` where we define the `git grep` command 🤦‍♂ 

```
$ git --no-pager grep '>>>>>>>'
.travis.yml:            - git --no-pager grep '>>>>>>>'
The command "git --no-pager grep '>>>>>>>'" exited with 0.
```

I'll fix this in #6477.

PR #5785, opened 2019-09-10
use sentry for the ssr service

### LINKED ISSUE #5783: Sentry is not configured for the SSR service
Within the Node-based SSR service, uncaught exceptions are not captured by Sentry. Sentry has a Node-based SDK, so we should be using Sentry for the SSR service as well.

escattone:
Fixes #5783 

This PR enables Sentry for the `ssr` service, so uncaught exceptions are now captured by Sentry.

The `package.json` and `package-lock.json` files were auto-generated from doing `npm install @sentry/node`.

### DIFF
--- a/kuma/javascript/ssr-server.js
+++ b/kuma/javascript/ssr-server.js
@@ -13,6 +13,21 @@ if (process.env.NEW_RELIC_LICENSE_KEY && process.env.NEW_RELIC_APP_NAME) {
     require('newrelic');
 }
 
+if (process.env.SENTRY_DSN) {
+    console.log('Configuring Sentry for the SSR server.');
+    const Sentry = require('@sentry/node');
+    let options = {
+        dsn: process.env.SENTRY_DSN
+    };
+    if (process.env.REVISION_HASH) {
+        options.release = process.env.REVISION_HASH;
+    }
+    if (process.env.SENTRY_ENVIRONMENT) {
+        options.environment = process.env.SENTRY_ENVIRONMENT;
+    }
+    Sentry.init(options);
+}
+
 const express = require('express'); // Express server framework
 const morgan = require('morgan');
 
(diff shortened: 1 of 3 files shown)


escattone:
An `infra` companion to this PR has already been merged (see https://github.com/mdn/infra/pull/288), and this PR is already in place on stage (although I haven't actually triggered an uncaught exception yet on the stage `ssr` service).

kuma/javascript/ssr-server.js
```diff
@@ -13,6 +13,21 @@ if (process.env.NEW_RELIC_LICENSE_KEY && process.env.NEW_RELIC_APP_NAME) {
     require('newrelic');
 }
 
+if (process.env.SENTRY_DSN) {
+    console.log('Configuring Sentry for the SSR server.');
+    const Sentry = require('@sentry/node');
```
peterbe: I don't know how I like this. I'd rather this was move outside the if statement so that we know for sure that the `require` works even if you don't configure and call `.init(options)` on it.

peterbe: Actually, what I'd like to is move ALL of this code *except* the line `const Sentry = require('@sentry/node');` to somewhere further down after all the `require()` stuff has happened.

escattone: I'll do that.

peterbe:
Will it definitely use the same DSN that you use for Python/Django? Do I just need to add it `SENTRY_DSN` to my `.env` and then get Sentry for both the Django code *and* the Node code?

escattone:
> Will it definitely use the same DSN that you use for Python/Django? Do I just need to add it `SENTRY_DSN` to my `.env` and then get Sentry for both the Django code _and_ the Node code?

Yes.

escattone:
Successfully tested on stage. 🎉

PR #5793, opened 2019-09-11
Fixes #5786, content should be readable in IE

### LINKED ISSUE #5786: MDN should be readable in IE
Currently MDN's new front-end shows as a blank page in the latest version of IE. It should be at least readable.

schalkneethling:
There are still a *lot* of styling problems, but that is a separate problem I will address in separate pull requests. This prevents the white screen of death and removes the blocker.

Tested in IE10 on Windows 8 and IE11 on Windows 10

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -64,16 +64,28 @@ if (container) {
     }
 
     if (app) {
+        // NOTE: `document.all` is only available in lte IE10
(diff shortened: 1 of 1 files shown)


Gregoor:
Code LGTM, unfortunately the asset serving troubles I'm experiencing on FF & Chrome are even worse on IE which makes it impossible for me to test this on IE atm.

PR #5794, opened 2019-09-11
Removes the need to restart SSR after build

schalkneethling:
Whilst working on the IE bug I noticed the following comment:
https://github.com/mozilla/kuma/blob/master/kuma/javascript/src/index.jsx#L86

No need to restart SSR after every build \o/

This PR then:

1. Removes the "script" that restarted the SSR server
2. Adds a little util that can then be used to ensure we stop the SSR server before jumping into devmode
3. Removes chokidar as we no longer need it

I kept the `webpack:development:node` script in case someone wants to run this locally during development.

@peterbe r?


webpack-build-utils/terminal.js
```diff
@@ -0,0 +1,13 @@
+const { exec } = require('child_process');
+
+function runInTerminal() {
+    exec(process.argv[2], error => {
+        if (error) {
+            console.error(`exec error: ${error}`);
```
peterbe: Shouldn't that be `${error.toString()}` or else you might end up with `exec error: [Object]`.

schalkneethling: Good point

schalkneethling:
@peterbe updated.

schalkneethling:
> The `package-lock.json` diff is 8,271 lines just for removing `chokidar-cli`. [This is what I get](https://gist.github.com/peterbe/b95fe2a5b640d71e845e968badebe140) when I ran it.
> I don't trust `npm` at all any more. Whenever using `npm install --save` or `npm uninstall --save ...` do it within the Kuma docker image.

meh :( I will redo per your instructions via email.

schalkneethling:
Hmmmm wait. Should I just remove `package-lock.json` and then do a fresh `npm i` via Docker? I guess that is the way to go here right?

schalkneethling:
@peterbe updated with the `package.json` regenerated from within Docker. The diff size is now  1,150

schalkneethling:
@peterbe This one good to go?

peterbe:
No, I don't think it's right to "Generate package.lock from within Docker container". I think that's only something you do if you get some nasty merge conflicts from a rebase. Just take the diff from: 
https://gist.github.com/peterbe/b95fe2a5b640d71e845e968badebe140

schalkneethling:
> No, I don't think it's right to "Generate package.lock from within Docker container". I think that's only something you do if you get some nasty merge conflicts from a rebase. Just take the diff from:
> https://gist.github.com/peterbe/b95fe2a5b640d71e845e968badebe140

So, should I remove `package.lock` from this PR entirely and just leave the change in `package.json`?

peterbe:
Revert your current `package-lock.json` first. And manually undo the removal from `package.json`. So that your branch doesn't have any changes to it. Then, go into a `docker-compose exec web bash` session and run `npm uninstall --save chokadir`. 
Then you'll get the right changes to `package-lock.json`.

peterbe:
I don't know what you call your "origin" but my origin is the mozilla/kuma one. 
```
git checkout origin/master -- package-lock.json
```
So perhaps in your case it's 
```
git checkout upstrea/master -- package-lock.json
```

schalkneethling:
@peterbe updated 🤞

PR #5800, opened 2019-09-11
Prevent search from submitting when query is empty

### LINKED ISSUE #5790: Search shouldn't even start if the input is empty
Steps to reproduce:

1. Go to https://beta.developer.mozilla.org/
2. Put cursor into search input
3. Hit Enter

Expected result:

That nothing happens. 

Actual result: 

It goes to https://beta.developer.mozilla.org/en-US/search?q= with a blank page. 
Screen Shot 2019-09-10 at 3 04 42 PM

Gregoor:
Fixes #5790

### DIFF
--- a/kuma/javascript/src/header/__snapshots__/header.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/header.test.js.snap
@@ -337,13 +337,14 @@ exports[`Header snapshot 1`] = `
       id="main-q"
       name="q"
       placeholder="Search MDN"
+      required={true}
       type="search"
     />
   </form>
   <a
     className="signin-link"
     data-service="GitHub"
-    href="/users/github/login/?next=/[fake%20absolute%20url]"
+    href="/users/github/login/?next=/"
     rel="nofollow"
   >
     Sign in
--- a/kuma/javascript/src/header/__snapshots__/header.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/header.test.js.snap
@@ -345,7 +345,7 @@ exports[`Header snapshot 1`] = `
   <a
     className="signin-link"
     data-service="GitHub"
-    href="/users/github/login/?next=/"
+    href="/users/github/login/?next=/[fake%20absolute%20url]"
     rel="nofollow"
   >
     Sign in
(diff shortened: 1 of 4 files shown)


Gregoor:
I went with JS first but closed [that PR](https://github.com/mozilla/kuma/pull/5796) when my mind screamed "progressive enhancement" at me. This here would work in IE11 for example. I guess for trimming it looks a bit different, `pattern="someMagicRegex"` might do the trick, but that's more obscure. It does have similar good browser support though.

peterbe:
The problem with whitespace still needs to be solved. The server will do:
```
q = (request.GET.get('q') or '').strip()
if not q:
    raise Hell
```

If the field has `required` on it, and you type something then change your mind and remove it and blur the input field you get the ugly red border. And that's why you have to try to bend the browser's built-in functionality but trying to undo that. Who knows if that works in all browsers?

Why wouldn't `if (!input.value) { return event.preventDefault()} ` work in IE11?

Gregoor:
> The problem with whitespace still needs to be solved.

I've added a pattern to detect that. I personally am not big on regexes (I like readable code), but I don't think this one is too bad.

preventDefault wouldn't work in IE, because we decided [not to run the client-side app in it](https://github.com/mozilla/kuma/blob/558716f5efa0a8efc085f03e4cf0d96b1591081e/kuma/javascript/src/index.jsx#L86-L87).

peterbe:
What else is going to be completely broken in IE11? It's 1.56% of our audience (91.5% of them are IE 11). 
And are we doing all these CSS hacks and complicated `pattern` regex just for those users? 
If, instead of the complicated approach, we just accept that if one of those 1.56% of preople do submit the search with a blank string we simply just let it be. 

After all, as part of https://github.com/mozilla/kuma/pull/5770 I made it so that visiting the `/en-US/search` *without* a query term at least doesn't break anymore. 

What do you think @schalkneethling ?

schalkneethling:
@peterbe I believe this is fine. Search is currently one of the more underutilized features of the site. This also does not render it unusable, it is just a degraded experience.

peterbe:
@Gregoor Once you get a green from Travis, feel free to Squash and merge.

PR #5802, opened 2019-09-11
cleaner slate on ssr

peterbe:


peterbe:
Yeah, this won't work. It would need to be done *before* the line `VERSION=latest make build-kuma` if anything.

PR #5820, opened 2019-09-13
Add command to help with GitHub setup

### LINKED ISSUE #5822: GitHub setup is under-documented
There's a couple of things that aren't documented yet to have GitHub login set-up, namely:

- `SITE_ID` needs to bet set to the `localhost:8000` site
- `DOMAIN` needs to be set to `mdn.localhost`
- hosts file needs several entries

Gregoor:
Fixes #5822 

In addition to documenting the fields, this also includes a command (and documentation for it) that sets the fields and also creates (or updates) the SocialApp model for GitHub.

To test it run the command as described in the docs (maybe back up your `.env` file). Different scenarios you can test are permutations of pre-existing SocialApp and localhost-Site models, as well as with/without `SITE_ID` and `DOMAIN` fields in your `.env`.

Missing feature: it doesn't ping mdn.localhost to check if hosts are correctly set up. Why? Because our docker instances don't ship with ping 🙈

### DIFF
--- a/kuma/attachments/management/commands/configure_github_social.py
+++ b/kuma/attachments/management/commands/configure_github_social.py
@@ -46,7 +46,7 @@ class Command(BaseCommand):
 
             print(GITHUB_INFO)
             client_id = raw_input('Client ID: ').strip()
-            client_secret = raw_input('Client Secret: ')
+            client_secret = raw_input('Client Secret: ').strip()
 
             social_app, created = SocialApp.objects.update_or_create(
                 provider='github',
--- a/kuma/attachments/management/commands/configure_github_social.py
+++ b/kuma/attachments/management/commands/configure_github_social.py
@@ -1,24 +1,34 @@
-import fileinput, os, sys, urllib2
+import fileinput
+import os
+import sys
 
-from django.core.management.base import BaseCommand, CommandError
-from django.contrib.sites.models import Site
 from allauth.socialaccount.models import SocialApp
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.core.management.base import BaseCommand
+
+try:
+    input = raw_input
+except NameError:
+    # Python3's input behaves like raw_input
+    # TODO: Delete this block when we've migrated
+    pass
 
 LOCALHOST = 'localhost:8000'
 MDN_LOCALHOST = 'mdn.localhost'
 
 OVERWRITE_PROMPT = 'There\'s already a SocialApp for GitHub, if you want to overwrite it type "yes":'
-GITHUB_INFO = '\n'.join([
-    'Visit https://github.com/settings/developers and click "New OAuth App"',
+GITHUB_INFO = (
+    'Visit https://github.com/settings/developers and click "New OAuth App"\n'
     'Set "Homepage URL" to "http://mdn.localhost:8000/" and Authorization callback URL to ' +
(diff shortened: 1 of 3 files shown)


peterbe:
As a general comment and reminder, please file an issue first. The issue should talk about the problem exclusively (but you can *comment* about potential solutions). The PR should talk about the solution only. 
Also, the first line of the PR description should be 
```
Fixes #0000

(optional extra blather about the PR such as screenshots and hints for reviewers)
```

kuma/attachments/management/commands/configure_github_social.py
```diff
@@ -0,0 +1,71 @@
...
+OVERWRITE_PROMPT = 'There\'s already a SocialApp for GitHub, if you want to overwrite it type "yes":'
+GITHUB_INFO = '\n'.join([
+    'Visit https://github.com/settings/developers and click "New OAuth App"',
+    'Set "Homepage URL" to "http://mdn.localhost:8000/" and Authorization callback URL to ' +
+    '"http://mdn.localhost:8000/users/github/login/callback/" respectively'
+])
+ENV_INFO = 'Putting SITE_ID and DOMAIN into .env'
+HOSTS_INFO = '\n'.join([
+    'Make sure your hosts file contains these lines:'
+    '127.0.0.1 kubernetes.docker.internal localhost demos mdn.localhost beta.mdn.localhost wiki.mdn.localhost',
```
peterbe: Ditto about the `kubernetes.docker.internal` line. Also, `localhost` should already be there.

kuma/attachments/management/commands/configure_github_social.py
```diff
@@ -0,0 +1,71 @@
...
+        site, created = Site.objects.update_or_create(
+            domain=LOCALHOST,
+            defaults={'name': LOCALHOST}
+        )
+        social_app.sites.add(site)
+
+        print('\n')
+
+        print(ENV_INFO)
+        overwrite_or_create_env_vars({'SITE_ID': site.id, 'DOMAIN': MDN_LOCALHOST})
```
peterbe: I don't know what's wrong with that function but it shouldn't have edited my file. 
```
▶ diff .env .env.backup
65a66
> DOMAIN=mdn.localhost
78a80
> DOMAIN=mdn.localhost
145,147d146
<
< DOMAIN=mdn.localhost
< SITE_ID=1
\ No newline at end of file
```

The `SITE_ID=site.id` should only happen if it's not already a key called `SITE_ID`. Because `SITE_ID = 1` is [already the default](https://github.com/mozilla/kuma/blob/6be5a95002bfe9e6750e36524421c9021a2ad50c/kuma/settings/common.py#L365).

Also, as you can see from by backup, I already had `DOMAIN=mdn.localhost` in mine yet it got added again.

peterbe: Also, when it does (correctly) add new entries the file fails to have 1 trailing linebreak.

peterbe:
> Missing feature: it doesn't ping mdn.localhost to check if hosts are correctly set up. Why? Because our docker instances don't ship with ping

Ha! Yeah, that's never going to be possible. I think you nailed it with the instruction to check that your `/etc/hosts` has certain lines.

Gregoor:
Thank you very much for taking the time! I think I've addressed all of the issues.

PR #5823, opened 2019-09-13
cache docker builds

peterbe:
An experiment.


peterbe:
This did not work and the effort to make it worth is unlikely to be worth it considering all the other things we could work on.

PR #5834, opened 2019-09-16
Fixes #5828, use react header on new signup flow

### LINKED ISSUE #5828: Use React header on new accounts signup flow
**Summary**

As part of the work in https://github.com/mdn/sprints/issues/1857 we need to start moving the page to our new frontend. The first step is to use the React header as we do for the landing page.

**Audience**

Users signing up for a new account

**Rationale**

See Epic https://github.com/mdn/sprints/issues/1857

**Workaround**

N/A

**Proposal**

Use new React header as we do for the landing page

schalkneethling:
This changes the signup page to use the new header. I opted to create a separate page type because I see this page adopting the new frontend tech much sooner than the landing page.

## Testing

The simplest way is to:

1. Open up the landing page in an incognito page
2. Click on the sign in link
3. Now, here you might get derailed because you have already "signed up" with your Github user.
4. This is where you would then use the test account I refered to in the email sent earlier.
5. Using this account to sign-in, should bring you to the new user signup view
6. Here you want to be sure that the new frontend header is being used as apposed to the old frontend header

## Current

Screenshot 2019-09-16 at 21 12 12

## New

Screenshot 2019-09-16 at 21 11 09


schalkneethling:
@Gregoor I have changed this so where I call `render_react` in Jinja, I pass `False` as the last argument. This tells `ssr` to not server-side render.

As this only adds the header portion, perhaps it is not really that big of a deal but, we do, do server-side rendering on the landing page.

The reason I opted out of server-side rendering is, when it was on, it causes a bunch of Python test failures all with the following outcome:

```bash

  File "/app/kuma/users/tests/test_templates.py", line 53, in test_signup_page_disabled

    response = self.github_login()

  File "/app/kuma/users/tests/__init__.py", line 183, in github_login

    response = self.client.get(callback_url, data, follow=True)
...

  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 116, in post

    return request('post', url, data=data, json=json, **kwargs)

  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 60, in request

    return session.request(method=method, url=url, **kwargs)

  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 533, in request

    resp = self.send(prep, **send_kwargs)

  File "/usr/local/lib/python2.7/site-packages/requests_mock/mocker.py", line 69, in _fake_send

    return self._real_send(session, request, **kwargs)

  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 646, in send

    r = adapter.send(request, **kwargs)

  File "/usr/local/lib/python2.7/site-packages/requests_mock/adapter.py", line 255, in send

    raise exceptions.NoMockAddress(request)

NoMockAddress: No mock address: POST http://ssr:8000/ssr/signup
```

schalkneethling:
Thanks @Gregoor and @callahad - Yes indeed, the indentation is Prettier. I am thinking perhaps we should either exclude `HTML` files or files with a `.jinja` extension. We do not currently use `.jinja` and perhaps it is not worth the trouble.

I will manually improve the indentation and just not "format on save" this file.

schalkneethling:
> Looks good to me, though I also couldn't figure out why SSR doesn't work right. Is it maybe that the landing page doesn't even really test for it? I.e. this is the first time tests are running with it?

The landing page definitely have a *lot* fewer tests than the user account portion.

PR #5835, opened 2019-09-16
drop preconnect to interactive examples

### LINKED ISSUE #5680: dns-prefetch or preconnect or both?
On view-source:https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video you'll see:
```html
 <link rel="dns-prefetch" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
 <link rel="preconnect" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
```

I keep forgetting how this stuff works but it certainly feels like one is a sub-set of the other. Is it a matter of that this is how stuff was done back in the day?

peterbe:
Fixes #5680

It's pointless to do `dns-prefetch` *and* `preconnect` because one is a subset of the other. 
What I don't like is that so many pages don't have a iframe whose source is `https://interactive-examples.mdn.mozilla.net/...`. These for example:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array 
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math

For all those, the user's browser is doing HTTP stuff all in vain. 

I started a deeper rabbit hole where it would analyze the HTML about to be rendered and only for the pages with an iframe (whose domain is in a whitelist) bother. Maybe I should finish that. 

Thoughts @atopal ?

### DIFF
--- a/kuma/wiki/jinja2/wiki/document.html
+++ b/kuma/wiki/jinja2/wiki/document.html
@@ -59,7 +59,6 @@
 {% endblock %}
 
 {% block extrahead %}
-  <link rel="dns-prefetch" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
   <link rel="preconnect" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
 
   <link rel="alternate" type="application/json" href="{{ url('wiki.json_slug', document.slug) | absolutify }}">
--- a/kuma/wiki/jinja2/wiki/react_document.html
+++ b/kuma/wiki/jinja2/wiki/react_document.html
@@ -85,7 +85,6 @@
     {% stylesheet 'banners' %}
   {% endif %}
 
-  <link rel="dns-prefetch" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
   <link rel="preconnect" href="https://interactive-examples.mdn.mozilla.net" pr="0.75" />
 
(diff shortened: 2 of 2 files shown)


peterbe:
Also, the pages that do have an interactive examples iframe, they usually appear above the fold and will start to download very soon anyway. I don't know what the point of this tag is but I know it's potentially sensitive since it might affect web perf KPIs.

atopal:
Yes, of course, it only makes sense to do that for pages where we have interactive examples, but how important is that relative to all the other things, considering impact on performance? I'd guess it's fairly low in priority. So, I'd leave it for now.

peterbe:
> I'd guess it's fairly low in priority. So, I'd leave it for now.

My interpretation of that is to not bother with the more complex but refined solution. I'll just let that branch die slowly on my disk.

PR #5836, opened 2019-09-16
remove rarely-used locales (28)

### LINKED ISSUE #5788: Delete localizations with less than 10 page views per day
https://bugzilla.mozilla.org/show_bug.cgi?id=1469079

https://bugzilla.mozilla.org/show_bug.cgi?id=1483763

https://bugzilla.mozilla.org/show_bug.cgi?id=1528618

escattone:
Fixes #5788

This PR removes the following 28 locales from MDN:
- `af`
- `az`
- `cs`
- `ee`
- `ff`
- `fy-NL`
- `ga-IE`
- `ha`
- `hr`
- `ig`
- `ka`
- `ln`
- `mg`
- `ml`
- `ro`
- `son`
- `sq`
- `sr`
- `sr-Latn`
- `sw`
- `ta`
- `te`
- `tl`
- `tn`
- `wo`
- `xh`
- `yo`
- `zu`

each of which has less than 10 page views per day.

As part of the database migration that this PR performs, all of the documents and document-deletion logs within each of the locales above will be permanently deleted, and also all users that have selected one of the above locales as their preferred locale will have their `locale` field nullified.

Once this PR is merged to `master`, the following steps should be performed:
- each of the 28 locales above should be put into `read-only` mode within MDN's Pontoon settings
- each of the 28 locales above should be removed from https://github.com/mozilla-l10n/mdn-l10n


kuma/core/tests/test_form_fields.py
```diff
@@ -27,11 +27,11 @@ def test_xx_YY_locale(self):
         num = _format_decimal(1234.567)
         assert '1,234.567' == num
 
-    def test_fy_NL_locale(self):
+    def test_pt_BR_locale(self):
```
escattone: NOTE: `babel` doesn't know about the `pt-XX` locales. If it did, the decimal would be formatted as `1.234,567` rather than `1,234.567`.

peterbe:
> each of which has less than 10 page views per day.

10 x 28 = 280 Page Not Found per day. Googlebot might not like that. 
Can we make these redirect to something? E.g. if `/en-US/docs/foo/bar` exists it would be a good candidate for `/fy-NL/docs/foo/bar`.

Perhaps it's not that important since our robots.txt and sitemaps should be correct within the next cycle.

kuma/users/migrations/0012_auto_20190914_2209.py
```diff
@@ -0,0 +1,31 @@
...
+    'tl', 'tn', 'wo', 'xh', 'yo', 'zu'
+)
+
+
+def update_users_with_discarded_locales(apps, schema_editor):
+    """
+    Nullify the preferred locale of all users with one of the DISCARDED_LOCALES.
+    """
+    User = apps.get_model('users', 'User')
+    User.objects.filter(locale__in=DISCARDED_LOCALES).update(locale=None)
```
peterbe: Instead of listing these, as this'll forever keep these strings in git, you could do:
```python
User.objects.exclude(locale__in=settings.ACCEPTED_LOCALES).update(locale=None)
```

kuma/wiki/migrations/0008_auto_20190914_2209.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Delete all documents and document-deletion logs within the
+    DISCARDED_LOCALES.
+    """
+    Document = apps.get_model('wiki', 'Document')
+    docs_to_delete = Document.objects.filter(locale__in=DISCARDED_LOCALES)
+    docs_to_delete.update(parent_topic=None)
+    docs_to_delete.delete()
+    DocumentDeletionLog = apps.get_model('wiki', 'DocumentDeletionLog')
+    DocumentDeletionLog.objects.filter(locale__in=DISCARDED_LOCALES).delete()
```
peterbe: Ditto here about the idea about using `.exclude(locale__in=settings.AVAILABLE_LOCALES)`

peterbe:
`s/AVAILABLE/ACCEPTED_LOCALES/`

peterbe:
ACTUALLY! At least for the `User`, the locale can not be `null`. So the update needs to be 
```
....update(locale='')
```

peterbe:
Some notes from prod:
```
>>> User.objects.all().count()
421111
>>> qs = User.objects.exclude(locale__in=settings.ACCEPTED_LOCALES)
>>> qs.count()
12
>>> for user in qs:
...   print(user.username, user.locale)
...
(u'DELETED', u'')
(u's911415', u'')
(u'adiccb', u'')
(u'belkacem77', u'')
(u'patrick.claes', u'')
(u'salasamui', u'')
(u'Howard.Chen', u'')
(u'nicolasolmos', u'')
(u'AndreyWeb-ID', u'')
(u'Rodrigo-Ludgero', u'')
(u'shmel3', u'')
(u'Xander7913', u'')
>>> User.objects.filter(locale__isnull=True).count()
0
```

So lemme revise what I suggested for the user migration one:

```python
User.objects.exclude(locale__in=settings.ACCEPTED_LOCALES).update(locale='')
```
that means that the 12 users that had `locale=''` they'd be included but that's fine. For all the other ones whose locale is `not in settings.ACCEPTED_LOCALES` they'd get `locale=''`.

kuma/users/migrations/0012_auto_20190914_2209.py
escattone: I'd prefer to keep it as is, since it captures exactly what this data migration should do at this point in time. The value of `settings.ACCEPTED_LOCALES` will change over time, changing what this migration does (although it's probably safe as long as we don't add any of these locales back?).

kuma/wiki/migrations/0008_auto_20190914_2209.py
escattone: Same here. I'd prefer to keep it as is, since it captures exactly what this data migration should do at this point in time.

escattone:
@peterbe Good catch on using `User.objects.update(locale='')` rather than `User.objects.update(locale=None)`! I'll make that change.

escattone:
@perterbe I let my fingers do the walking, and by force of habit re-based my `User.objects...update(locale='')` into the original commit. Sorry!

escattone:
> > each of which has less than 10 page views per day.
> 
> 10 x 28 = 280 Page Not Found per day. Googlebot might not like that.
> Can we make these redirect to something? E.g. if `/en-US/docs/foo/bar` exists it would be a good candidate for `/fy-NL/docs/foo/bar`.
> 
> Perhaps it's not that important since our robots.txt and sitemaps should be correct within the next cycle.

Yes, I think that's the case.

atopal:
> > > each of which has less than 10 page views per day.
> > 
> > 
> > 10 x 28 = 280 Page Not Found per day. Googlebot might not like that.
> > Can we make these redirect to something? E.g. if `/en-US/docs/foo/bar` exists it would be a good candidate for `/fy-NL/docs/foo/bar`.
> > Perhaps it's not that important since our robots.txt and sitemaps should be correct within the next cycle.
> 
> Yes, I think we don't have to worry about it since we'll re-generate a new sitemap and `robots.txt`.

Yeah, no worries, as long as the sitemap is up to date. We have almost 60,000 404s per month.

kuma/users/migrations/0012_auto_20190914_2209.py
peterbe: I'm totally cool with that. Mind you, `settings.ACCEPTED_LOCALES` does change but the migration will never ever run again in stage or prod because they're remembered as run (past tense).

PR #5844, opened 2019-09-18
Update Docker docs to cover default memory limits

callahad:
I lost a day trying to figure out why my containers kept dying on macOS. This note would've saved me.


peterbe:
THANK YOU! Much needed. Here's mine. 
Screen Shot 2019-09-19 at 11 17 52 AM

PR #5847, opened 2019-09-18
add message when falling-back to English

### LINKED ISSUE #5806: New frontend: no call to action for missing translation
Previously, when a user landed on `https://developer.mozilla.org/<localeX>/slug/` because they clicked on a link using `https://developer.mozilla.org/slug` (this is supposed to redirect to the user's locale), a call to action was displayed to help with the translation.

Steps to reproduce:
* Ran the accessibility controls from the Fx devtools
* Clicked on a link : https://developer.mozilla.org/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast 
* Was redirected to https://developer.mozilla.org/fr/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast

Previous/expected result:
* a CTA box as displayed in https://wiki.developer.mozilla.org/fr/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast
image

Current result:
* No indication that I'm reading the en-US content though the URL contains the /fr/ fragment on https://developer.mozilla.org/fr/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast
image

Aside:
I know that localization may be not the strongest point of MDN (both content and tooling/feature wise) at the moment but that's not helping getting traction here.

escattone:
Fixes #5806 

This PR adds a simplified `beta` domain version of the translation message that is displayed within the `wiki` domain when falling-back to the English version of a document that has been requested in a non-English locale. The goal here was to show a clear message that was simple to implement in the React world (e.g., no changes would be required to the `document_api_data` function in the back-end). cc @SphinxKnight 

For example, this is the **`wiki`** version of the message (http://wiki.mdn.localhost:8000/ca/docs/Web/HTTP):

image

This is the `beta` version **before** this PR (http://beta.mdn.localhost:8000/ca/docs/Web/HTTP):

image

This is the `beta` version **after** this PR (http://beta.mdn.localhost:8000/ca/docs/Web/HTTP):

image

For local testing, I used these settings in my `.env` file (my usual settings):
```
DEBUG=True
DOMAIN=mdn.localhost
WIKI_HOST=wiki.mdn.localhost:800
```
```
##
# Host Database
#
##
255.255.255.255	broadcasthost
::1             localhost
```
and did something like this:
```
docker-compose pull
docker-compose up -d
docker-compose restart
```

Then try:
- http://beta.mdn.localhost:8000/en-US/docs/Web/HTTP (no fallback, so no message)
- http://beta.mdn.localhost:8000/es/docs/Web/HTTP (no fallback, so no message)
- http://beta.mdn.localhost:8000/ca/docs/Web/HTTP (fallback, so show message)
- http://wiki.mdn.localhost:8000/ca/docs/Web/HTTP (for comparison)

The `Help us translate this article!` link should go to `http://wiki.mdn.localhost:8000/en-US/docs/Web/HTTP$locales`.

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -66,7 +66,7 @@ export default function Article({ document }: DocumentProps) {
             }
         >
             {document.locale !== locale && (
-                <div id="doc-pending-fallback" class="warning">
+                <div id="doc-pending-fallback" className="warning">
                     <p>
                         <bdi>
                             {gettext(
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -4,6 +4,7 @@ import { useContext, useEffect, useRef } from 'react';
 
 import { activateBCDSignals, activateBCDTables } from './bcd.js';
 import { addLiveExampleButtons } from './live-examples.js';
+import { getLocale, gettext } from './l10n.js';
 import { highlightSyntax } from './prism.js';
 import * as InteractiveExamples from './interactive-examples.js';
 import UserProvider from './user-provider.jsx';
@@ -47,6 +48,7 @@ export default function Article({ document }: DocumentProps) {
 
     const isArchive =
         document.slug === 'Archive' || document.slug.startsWith('Archive/');
+    const locale = getLocale();
 
     return (
         /*
@@ -63,6 +65,21 @@ export default function Article({ document }: DocumentProps) {
                     : 'article text-content'
             }
         >
+            {document.locale !== locale && (
+                <div id="doc-pending-fallback" class="warning">
+                    <p>
+                        <bdi>
+                            {gettext(
+                                "You're reading the English version of this content, since no translation exists yet for this locale."
+                            )}
+                            &nbsp;
+                            <a href={document.translateURL} rel="nofollow">
+                                {gettext('Help us translate this article!')}
+                            </a>
+                        </bdi>
+                    </p>
+                </div>
+            )}
             <article
                 id="wikiArticle"
                 dangerouslySetInnerHTML={{ __html: document.bodyHTML }}
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -66,7 +66,7 @@ export default function Article({ document }: DocumentProps) {
             }
         >
             {document.locale !== locale && (
-                <div id="doc-pending-fallback" className="warning">
+                <div className="warning">
                     <p>
                         <bdi>
                             {gettext(
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -70,7 +70,7 @@ export default function Article({ document }: DocumentProps) {
                     <p>
                         <bdi>
                             {gettext(
-                                "You're reading the English version of this content, since no translation exists yet for this locale."
+                                'You’re reading the English version of this content since no translation exists yet for this locale.'
                             )}
                             &nbsp;
                             <a href={document.translateURL} rel="nofollow">


escattone:
By the way, the front-end team should feel free to close this and replace with something better. 😃

kuma/javascript/src/article.jsx
```diff
@@ -63,6 +65,21 @@ export default function Article({ document }: DocumentProps) {
                     : 'article text-content'
             }
         >
+            {document.locale !== locale && (
+                <div id="doc-pending-fallback" class="warning">
```
schalkneethling: Use `className` here. Also, it feels more like an `info` message as opposed to a warning message. Will double check if we have styling for this, else I can provide you with some Sass do add.

Note to self: I am still not sure why our warning messages are red, feels like it should be yellow instead.

kuma/javascript/src/article.jsx
```diff
@@ -63,6 +65,21 @@ export default function Article({ document }: DocumentProps) {
...
+                    <p>
+                        <bdi>
+                            {gettext(
+                                "You're reading the English version of this content, since no translation exists yet for this locale."
```
schalkneethling: Surprised that `es-lint` is not complaining about the `"` here. You could avoid that by doing

```suggestion
                                'You’re reading the English version of this content since no translation exists yet for this locale.'
```

kuma/javascript/src/article.jsx
escattone: Will do! Thanks for the review @schalkneethling !

kuma/javascript/src/article.jsx
```diff
@@ -63,6 +65,21 @@ export default function Article({ document }: DocumentProps) {
                     : 'article text-content'
             }
         >
+            {document.locale !== locale && (
+                <div id="doc-pending-fallback" className="warning">
```
peterbe:
```suggestion
                <div className="warning">
```
That ID is not used in CSS. On the [wiki it's used to trigger these lines](https://github.com/mozilla/kuma/blob/59ca3f7d0e508e063239057745c9190a533a6c74/kuma/static/js/wiki.js#L224-L228):
```
mdn.analytics.trackError('Translation Pending', 'displayed');
```
I guess it simply means we need to have that analytics track thing on the React side too.

escattone:
Good catch, thanks @peterbe!

peterbe:
But don't we need to make the `mdn.analytics.trackError` thing too? Like the wiki did.

escattone:
> But don't we need to make the `mdn.analytics.trackError` thing too? Like the wiki did.

Sorry, that was hasty of me. I was thinking it wasn't important, but maybe it is. Let's see what @atopal says about the `mdn.analytics.trackError('Translation Pending', 'displayed');` code.

atopal:
If we can keep it without too much hassle, we should do it. We are currently not using it for decision making, but since L10n will soon be a thing we have to make fundamental decisions about, it would be good to have that data ready.

peterbe:
@escattone Perhaps you didn't notice it but if you merge https://github.com/escattone/kuma/pull/1 into this we can finish the review here on this PR.

escattone:
@peterbe Thanks! I didn't notice it until late last night when i checked my personal email. 😄

kuma/javascript/src/article.jsx
```diff
@@ -47,6 +48,13 @@ export default function Article({ document }: DocumentProps) {
 
     const isArchive =
         document.slug === 'Archive' || document.slug.startsWith('Archive/');
+    const locale = getLocale();
+
+    useEffect(() => {
+        if (document.locale !== locale) {
+            mdn.analytics.trackError('Translation Pending', 'displayed');
```
schalkneethling: I wonder if we should not instead use the `ga` functionality in the new frontend? One reason is that this will most likely fail when server-side rendered as `mdn.analytics` will not be defined.

Looking at the code in `/kuma/static/js/analytics.js` the `trackError` function calls `trackEvent` and passes the following in addition to what is passed into the function:

```
{
  'category': category, // 'Translation Pending'
  'action': action, // 'displayed'
  'label': label // ''
}
```

`trackEvent` calls `send` which then does the final call:

```
ga('send', {
  'eventCategory' : 'Translation Pending',
  'eventAction'   : 'displayed',
  'eventLabel'    : '',
  'eventValue'    : 0,
  'hitCallback'   : null,
  'hitType': 'event'
});
```

So for this we can do:

```
...
import GAProvider from './ga-provider.jsx';
...
const ga = useContext(GAProvider.context);
...
ga('send', {
  hitType: 'event',
  eventCategory: 'Translation Pending',
  eventAction: 'displayed',
  eventLabel: ''
});
```

peterbe: You're right @schalkneethling that `const ga = useContext(GAProvider.context);` is the new better API. 

However, in its current state `mdn.analytics.trackError` is only ever reached for from an effect (`useEffect`) and they only execute when in the browser.

schalkneethling: Ah yes, that is true. But, would it not seem a little out of place to have this one place where we can into the "old" JS codebase, but everywhere else we use the new frontend GA bits?

peterbe: THANK YOU @schalkneethling for that reminder. I put this up: https://github.com/escattone/kuma/pull/2
@escattone ping ^

escattone: Merged. Thanks @schalkneethling and @peterbe!

PR #5855, opened 2019-09-20
make breadcrumb schema validate

### LINKED ISSUE #5853: Escalated number of errors from Google Search Console: Breadcrumbs
[Google Search Console](https://search.google.com/search-console/breadcrumbs/drilldown?resource_id=https%3A%2F%2Fdeveloper.mozilla.org%2F&item_key=CgwICRoIcG9zaXRpb24QAxgP&hl=en) emailed us about a big increase in indexing "errors" about breadcrumbs.
Screen Shot 2019-09-20 at 1 47 54 PM

The code that produces our breadcrumbs (on the Wiki) [hasn't changed in years](https://github.com/mozilla/kuma/blame/master/kuma/wiki/jinja2/wiki/includes/document_macros.html).

peterbe:
Fixes #5853

This one was tricky. I know the issue is a handful but ponder what this PR results in...:

This is what it looks like:
Screen Shot 2019-09-20 at 3 12 58 PM

And if I open view-source:http://wiki.mdn.localhost:8000/en-US/docs/Web/CSS/animation-delay and paste in the relevant `<ol>` tag it validates:
Screen Shot 2019-09-20 at 3 17 55 PM

### DIFF
--- a/kuma/static/styles/components/wiki/crumbs.scss
+++ b/kuma/static/styles/components/wiki/crumbs.scss
@@ -35,6 +35,13 @@ $crumb-vertical-spacing-desktop: $grid-spacing / 4;
         }
     }
 
+    a.crumb-current-page {
+        &:link,
+        &:visited {
+            color: $text-color;
+        }
+    }
+
     span {
         display: inline-block;
         position: relative;
@@ -66,10 +73,5 @@ $crumb-vertical-spacing-desktop: $grid-spacing / 4;
 
     li:last-child span:after {
         display: none;
-    }
-
-    li:last-child a {
-        // Make the last-child <a> appear like regular text
-        color: #333;
     }
 }
--- a/kuma/static/styles/components/wiki/crumbs.scss
+++ b/kuma/static/styles/components/wiki/crumbs.scss
@@ -67,4 +67,9 @@ $crumb-vertical-spacing-desktop: $grid-spacing / 4;
     li:last-child span:after {
         display: none;
     }
+
+    li:last-child a {
+        // Make the last-child <a> appear like regular text
+        color: #333;
+    }
 }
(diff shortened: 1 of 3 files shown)


kuma/static/styles/components/wiki/crumbs.scss
```diff
@@ -67,4 +67,9 @@ $crumb-vertical-spacing-desktop: $grid-spacing / 4;
     li:last-child span:after {
         display: none;
     }
+
+    li:last-child a {
```
schalkneethling: Seeing that we know when we output the last list item in `document_macros.html`, I would suggest we use a class here instead, for example:

```
...
<a href="{{ document.get_absolute_url() }}" class="crumb-current-page" property="item" typeof="WebPage">
  <span property="name" aria-current="page">{{ document.title }}</span>
</a>
<meta property="position" content="{{ document.parents|length_plus_one }}">
```

```
a.crumb-current-page {
  &:link,
  &:visited {
    color: $text-color;
  }

  &:hover,
  &:focus {
    text-decoration: underline;
  }
  
}
```

peterbe: Sorry for being lazy but why is that better? Is `:last-child` a bad technique? E.g. slower.
Also, with `:last-child` I can have less stuff in the HTML. 
Please educate me whilst I just make the change.

peterbe: Also, note that we already used `:last-child` for [some other thing](https://github.com/mozilla/kuma/blob/3e5c2a8b6c800f908d07de302d4b8d04a13ec3a9/kuma/static/styles/components/wiki/crumbs.scss#L67).

schalkneethling: I would not say it is bad, it definitely has its uses, and I do use it. In this case, it just feels better to be explicit with the class name i.e. It is clear what the styling is aiming to accomplish without having to write explanatory comments ;)

PR #5863, opened 2019-09-24
Fixes #5859, improve accessibility of custom lang dropdown

### LINKED ISSUE #5859: Improve the accessibility of language drop-down
**Summary**

From https://bugzilla.mozilla.org/show_bug.cgi?id=1583146 reported by @marcozehe

## What did you do?

1. With NVDA on Windows, visit any MDN documentation page.
2. Find the heading 1, and down arrow a few lines until you get to the button that speaks your language as the name, e. g. "English".
3. DownArrow once more. You'll hear "button" spoken again. But there is no label.
4. Try to open the button menu with the keyboard. 

### What happened?

After step 3, No label was spoken, and the button does not indicate that it will open a popup for a language selection.
After step 4, the menu isn't keyboard operable, only drops down on mouse hover.

## What should have happened?

There should not be an empty portion of the button, and it should indicate that it has a popup. The menu should be keyboard operable, even without a screen reader.

**Additional context**

The solution is probably to make the span and its children that shows the dropdown arrow aria-hidden="true", since this is purely presentational. The button accessible doesn't need this, since it laready has the language name as its label.

Second, it should have aria-haspopup="menu" or similarly appropriate, to indicate that there is a menu attached toit. And because this is initially collapsed, it should also have aria-expanded="false".. The latter, of course, changes to "true" once the button gets expanded, and the menu appears.

And of course, this must be made keyboard accessible. It's a menu, like other menus on MDN, too, so the same user interaction applies, like focusing the first menu item when expanding with the keyboard, up and down arrow selects, escape closes and puts focus back on the language button, etc. The list is a menu, the li is presentation, the links are menuitem roles.

schalkneethling:
This does not entirely resolve issue #5859 but, it is now on par with the rest of our menus on the new frontend. The next steps would be to add the additional required JavaScript to ensure that:

1. the menu is navigable via the arrow keys
2. When the menu is opened and closed toggle `aria-hidden` (current `aria-hidden` is not set at all)
3. Ensure focus moves to the first item in the list
4. Pressing the escape key closes the menu

All of this is relevant to this custom menu as well as the main navigation of the site. I will create separate issues for these so that it can be triaged, prioritized, and moved into our sprint work.

In this pull request, I do set a custom label for the `button` element that triggers the menu so that instead of the assistive technology simply reading `English pop-up button`, it will now read `Choose your preferred language pop-up button`

@MarcoZehe let me know your thoughts on this pull request as well as my comments above. Thank you.

### DIFF
--- a/kuma/javascript/src/header/language-menu.jsx
+++ b/kuma/javascript/src/header/language-menu.jsx
@@ -2,7 +2,7 @@
 import * as React from 'react';
 
 import Dropdown from './dropdown.jsx';
-import { gettext } from '../l10n.js';
+import { gettext, interpolate } from '../l10n.js';
 
 import type { DocumentData } from '../document.jsx';
 
@@ -28,11 +28,20 @@ export default function LanguageMenu({ document }: Props): React.Node {
     // and we can just use the document language string without translation.
     let label =
         document.locale === 'en-US' ? gettext('English') : document.language;
+    let chooseLanguageString = interpolate(
+        gettext('Current language is %s. Choose your preferred language.'),
+        [label]
+    );
 
     return (
-        <Dropdown label={label} right={true}>
+        <Dropdown
+            label={label}
+            right={true}
+            ariaOwns="language-menu"
+            ariaLabel={chooseLanguageString}
+        >
             {document.translations.map(t => (
-                <li key={t.locale} lang={t.locale}>
+                <li key={t.locale} lang={t.locale} role="menuitem">
                     <a href={t.url} title={t.localizedLanguage}>
                         <bdi>{t.language}</bdi>
                     </a>
--- a/kuma/static/styles/minimalist/components/_dropdown.scss
+++ b/kuma/static/styles/minimalist/components/_dropdown.scss
@@ -1,5 +1,3 @@
-@import '../vars/vars-nav';
-
 .dropdown-container {
     /* We use relative here not because we need to tweak positioning
        but because we need to position the menu itself relative to the
@@ -9,15 +7,16 @@
     /* This adjusts for the 5px left padding */
     margin-left: -5px;
 
-    &:hover {
-        &.dropdown-menu-label,
-        &.dropdown-menu-label a {
-            color: $dropdown-item-background-color;
-        }
-
-        &.dropdown-menu-items {
-            display: flex;
-        }
+    .dropdown-menu-label {
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+        color: $text-color;
+        padding: 0 5px 5px;
+        border: none;
+        white-space: nowrap;
+        font-size: $tiny-font-size;
+        font-weight: bold;
     }
 
     .dropdown-menu-items {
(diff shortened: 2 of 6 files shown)


MarcoZehe:
Two thoughts:

1. If the menu gets shown and hidden via CSS, like toggling display:none; to block and vice Versa, aria-hidden is not needed at all. Accessibility will recognize the CSS toggle.
2. If you override the language label, make sure to not lose current information. The way you propose this now, the info that the current language is English, gets lost. So change it to something like „Toggle preferred language (currently English)“ or such.

schalkneethling:
> Two thoughts:
> 
> 1. If the menu gets shown and hidden via CSS, like toggling display:none; to block and vice Versa, aria-hidden is not needed at all. Accessibility will recognize the CSS toggle.

Perfect, then this should be all good.

> 2. If you override the language label, make sure to not lose current information. The way you propose this now, the info that the current language is English, gets lost. So change it to something like „Toggle preferred language (currently English)“ or such.

Agreed. That was my initial intent. The challenge is to not mangle the strings that are exposed for localization. I will think about it a bit more and see how I can make it work. Thanks Marco!

kuma/javascript/src/header/language-menu.jsx
```diff
@@ -28,11 +28,18 @@ export default function LanguageMenu({ document }: Props): React.Node {
...
     let label =
         document.locale === 'en-US' ? gettext('English') : document.language;
+    let chooseLanguageString = gettext('Choose your preferred language.');
+    let currentLangPre = gettext('Current language is');
 
     return (
-        <Dropdown label={label} right={true}>
+        <Dropdown
+            label={label}
+            right={true}
```
joedarc: Can't this just be `right` without the `right={true}`?

kuma/static/styles/minimalist/components/_dropdown.scss
```diff
@@ -54,25 +55,17 @@
         }
     }
 
-    &:hover .dropdown-menu-items {
-        display: flex;
+    &:hover .dropdown-menu-label,
+    &:hover .dropdown-menu-label a,
+    &:focus .dropdown-menu-label,
+    &:focus .dropdown-menu-label a {
+        color: $dropdown-item-background-color;
```
Gregoor: nit: if you don't think this is too much nesting:
```scss
&:hover, &:focus {
  .dropdown-menu-label a {
    answer: 42;
  }
}
```

kuma/javascript/src/header/language-menu.jsx
```diff
@@ -28,11 +28,18 @@ export default function LanguageMenu({ document }: Props): React.Node {
     // and we can just use the document language string without translation.
     let label =
         document.locale === 'en-US' ? gettext('English') : document.language;
+    let chooseLanguageString = gettext('Choose your preferred language.');
+    let currentLangPre = gettext('Current language is');
```
Gregoor: Reminder from our meeting that we might want to use `interpolate` here

kuma/javascript/src/header/language-menu.jsx
schalkneethling: Thank you for your comment here @joedarc. Not sure if we should change this in this PR as it is:

1. Unrelated to the work here
2. Affects more components than the lang selector

Perhaps file an issue for this. It seems like a great `good-first-issue` to me. 🎉

PR #5871, opened 2019-09-25
replace gravatar with github avatar

### LINKED ISSUE #5848: Expose the GitHub avatar to Jinja as well as via the "whoami" endpoint
**Summary**
Expose the GitHub avatar to the front-end code as both a Jinja helper as well as an attribute of the `whomai` endpoint.

**Rationale**
Required as part of #5830 

**Audience**
All MDN members

escattone:
Fixes #5848 

#### Summary
This PR replaces the use of Gravatar with the user's GitHub avatar URL.

- For Jinja templates, you can now use `get_avatar_url(user)`, which also supports an optional `provider` argument, defaulting to `'github'`, but accommodating the future by supporting `get_avatar_url(user, provider='google')`.

- The `whoami` endpoint has also been updated to provide the GitHub avatar URL (and no longer provides the Gravatar large/small URL's).

##### Beta header
image

##### Wiki header
image

##### Wiki/beta user profile
image

image

##### Wiki/beta user profile (edit)
image


kuma/humans/fixtures/contributors.json
```diff
@@ -1,154 +1,170 @@
 [
```
escattone: This file has been updated with the latest response from https://api.github.com/repos/mozilla/kuma/contributors, limited to the top 8 contributors for brevity sake.

kuma/users/templatetags/jinja_helpers.py
```diff
@@ -12,15 +12,22 @@
...
-    return job.get(email, secure=secure, size=size,
-                   rating=rating, default=default)
+def get_avatar_url(user, provider='github'):
+    """
+    Get the user's avatar URL for the specified provider. Returns None if the
+    user is not logged-in, or the default avatar if the user is logged-in but
+    has no avatar for the specified provider.
+    """
+    if user and user.is_authenticated:
+        for account in user.socialaccount_set.filter(provider=provider):
```
peterbe: Some day, when you can log in with either (github or google), there'd need to be a smart way to merge accounts. That day, how do you know which socialaccount to use? Perhaps whichever one you most recently logged in with??

kuma/users/templatetags/jinja_helpers.py
```diff
@@ -12,15 +12,22 @@
...
-    job = UserGravatarURLJob()
-    return job.get(email, secure=secure, size=size,
-                   rating=rating, default=default)
+def get_avatar_url(user, provider='github'):
+    """
+    Get the user's avatar URL for the specified provider. Returns None if the
+    user is not logged-in, or the default avatar if the user is logged-in but
+    has no avatar for the specified provider.
+    """
+    if user and user.is_authenticated:
```
peterbe: This will break the syndication feed. 

All places except one already do a check that `user` is a thing and `.is_authenticated`. E.g. `login.html` does:
```django
{% if user.is_authenticated %}
        <a class="user-url" href="{{ user.get_absolute_url() }}">
            <img src="{{ get_avatar_url(user) }}" 
...
```
Same with the `whoami` view:
```python
user = request.user
    if user.is_authenticated:
        data = {
            'username': user.username,
            'timezone': user.timezone,
            'is_authenticated': True,
            'is_staff': user.is_staff,
            'is_superuser': user.is_superuser,
            'is_beta_tester': user.is_beta_tester,
            'avatar_url': get_avatar_url(user),
```

However, one place where it doesn't is the `DocumentJSONFeedGenerator.write` method:

```python
if revision.creator:
    item_out['author_avatar'] = get_avatar_url(revision.creator)
```
In this case, the `user` isn't from a `request` but rather from a regular `wiki.Document.current_revision` lookup.

I think we should simply delete this if-statement and it should work in both scenarios. The feed would get avatar URLs.

kuma/users/templatetags/jinja_helpers.py
```diff
@@ -12,15 +12,22 @@
...
+        for account in user.socialaccount_set.filter(provider=provider):
+            avatar_url = account.get_avatar_url()
+            if avatar_url:
+                return avatar_url
+        else:
+            return settings.DEFAULT_AVATAR
```
peterbe: then we can remove that `else:` and dedent this line.

kuma/wiki/feeds.py
```diff
@@ -141,8 +141,8 @@ def write(self, outfile, encoding):
             else:
                 revision = document
 
-            if revision.creator.email:
-                item_out['author_avatar'] = gravatar_url(revision.creator.email)
+            if revision.creator:
+                item_out['author_avatar'] = get_avatar_url(revision.creator)
```
peterbe: Because of this line, with `django-querycount` enabled, this happens:

```
web_1            | http://wiki.mdn.localhost:8000/en-US/docs/feeds/json/all/
web_1            | |------|-----------|----------|----------|----------|------------|
web_1            | | Type | Database  |   Reads  |  Writes  |  Totals  | Duplicates |
web_1            | |------|-----------|----------|----------|----------|------------|
web_1            | | RESP |  default  |   505    |    0     |   505    |     65     |
web_1            | |------|-----------|----------|----------|----------|------------|
web_1            | Total queries: 505 in 2.8313s
web_1            |
web_1            |
web_1            | Repeated 65 times.
web_1            | SELECT `socialaccount_socialaccount`.`id`,
web_1            | `socialaccount_socialaccount`.`user_id`,
web_1            | `socialaccount_socialaccount`.`provider`,
web_1            | `socialaccount_socialaccount`.`uid`,
web_1            | `socialaccount_socialaccount`.`last_login`,
web_1            | `socialaccount_socialaccount`.`date_joined`,
web_1            | `socialaccount_socialaccount`.`extra_data` FROM
web_1            | `socialaccount_socialaccount` WHERE
web_1            | (`socialaccount_socialaccount`.`user_id` = 9 AND
web_1            | `socialaccount_socialaccount`.`provider` = 'github')
web_1            |
web_1            | Repeated 51 times.
web_1            | SELECT `socialaccount_socialaccount`.`id`,
web_1            | `socialaccount_socialaccount`.`user_id`,
web_1            | `socialaccount_socialaccount`.`provider`,
web_1            | `socialaccount_socialaccount`.`uid`,
web_1            | `socialaccount_socialaccount`.`last_login`,
web_1            | `socialaccount_socialaccount`.`date_joined`,
web_1            | `socialaccount_socialaccount`.`extra_data` FROM
web_1            | `socialaccount_socialaccount` WHERE
web_1            | (`socialaccount_socialaccount`.`user_id` = 853 AND
web_1            | `socialaccount_socialaccount`.`provider` = 'github')
```

See https://github.com/escattone/kuma/pull/3

kuma/users/templatetags/jinja_helpers.py
escattone: Good question. I guess we'll cross that bridge when we get there.

kuma/users/templatetags/jinja_helpers.py
escattone: Good point. Although the `if user and user.is_authenticated:` is probably harmless (since it's really meant to filter-out anonymous users), I can't think of any case where we'd call `get_avatar_url` without testing for the user anyway (since we'd have to handle the `None` returned otherwise). I'll remove that check. Thanks @peterbe!

kuma/wiki/feeds.py
escattone: Nice analysis and good idea, thanks @peterbe !

escattone: Merged https://github.com/escattone/kuma/pull/3

PR #5872, opened 2019-09-25
Issue#5819 add interpolated with elements

### LINKED ISSUE #5819: [L10n] avoid hardcoded puncture marks in the page content
Hi, 

In the latest batch of page contents exposed to l10n, the string `newsletter-privacy-input` is broken into `I’m okay with Mozilla handling my info as explained in this `, `Privacy Policy` with hyperlink, and a hard-coded period[1].

Please avoid such situation, due to a) in some languages, the link might not always in the end of sentence b) the full stop of a sentence might not be a half-width dot like English. It applies to other puncture marks too.

We should probably a) break this into multiple segments with appropriate comment so localizers can notice, or b) merge in to one like what we have previously. See the old version[2] here for an example.

Thanks!

[1] https://github.com/mozilla/kuma/blob/d4572628db6dcac43dcb0340d62d0ef33e51a565/kuma/javascript/src/newsletter.jsx#L282

[2] https://github.com/mozilla-l10n/mdn-l10n/blob/f9e3bbe67d68738db5c11a4f79b27f2881da7885/zh_TW/LC_MESSAGES/django.po#L342

Gregoor:
Fixes #5819 

I added a Component that works similar to `interpolate`, except that it returns an array with `React.Node`s in it. That way you can also use it to interpolate elements into a string (without risking XSS).

Longer term I'd like to investigate using [Fluent](https://github.com/projectfluent/fluent.js/) which we used in Common Voice and will soon be the No 1 localization system inside of FF. The things we do in `l10n.jsx` are similar but less powerful to what Fluent offers (pluralization comes to mind).

### DIFF
--- a/kuma/javascript/src/l10n.js
+++ b/kuma/javascript/src/l10n.js
@@ -13,6 +13,7 @@
  *
  * @flow
  */
+import * as React from 'react';
 
 // This is the type of the string catalog that Django creates for a locale.
 type StringCatalog = { [string]: string | Array<string> };
@@ -149,4 +150,33 @@ export function interpolate(s: string, args: Array<any> | { [string]: any }) {
             String(typedArgs[match.slice(2, -2)])
         );
     }
+}
+
+const ELEMENT_REGEXP = /(<\w+\s*\/>)/g;
+/**
+ * Similar to interpolate() it interpolates the given id-string with the given
+ * properties, but it can also work with React Elements. It will replace strings
+ * of the form `<name/>` and return an Array of React Elements.
+ */
+export function Interpolated({
+    id,
+    ...args
+}: {
+    id: string,
+    [key: string]: React.Node
+}): React.Node[] {
+    return id.split(ELEMENT_REGEXP).map((str, i) => {
+        if (!ELEMENT_REGEXP.test(str)) {
+            return str;
+        }
+
+        const element = args[str.slice(1, -2).trim()];
+        if (!element) {
+            return str;
+        }
+
+        return React.cloneElement(element, {
+            key: str + i
+        });
+    });
 }
--- a/kuma/javascript/src/l10n.js
+++ b/kuma/javascript/src/l10n.js
@@ -175,6 +175,13 @@ export function Interpolated({
             return str;
         }
 
+        /**
+         *  This string refers to an element, but is there a corresponding
+         *  element in args?
+         *  `args` contain: privacyLink={<a href="#here">Hello</a>}
+         *  args["<footer/>".slice(1, -2).trim()]; // false
+         *  args["<privacyLink/>".slice(1, -2).trim()]; // true
+         */
         const element = args[str.slice(1, -2).trim()];
         if (!element) {
             return str;
--- a/kuma/javascript/src/l10n.js
+++ b/kuma/javascript/src/l10n.js
@@ -152,6 +152,11 @@ export function interpolate(s: string, args: Array<any> | { [string]: any }) {
     }
 }
 
+/**
+ * ELEMENT_REGEXP.test("This is just a string"); // false
+ * ELEMENT_REGEXP.test("<elem/>"); // true
+ * ELEMENT_REGEXP.test("<elemWithSpace />"); // true
+ */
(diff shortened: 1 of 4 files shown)


kuma/javascript/src/l10n.js
```diff
@@ -136,3 +137,32 @@ export function interpolate(s: string, args: Array<any> | { [string]: any }) {
         );
     }
 }
+
+const ELEMENT_REGEXP = /(<\w+\s*\/>)/g;
+/**
+ * Similar to interpolate() it interpolates the given id-string with the given
+ * properties, but it can also work with React Elements. It will replace strings
+ * of the form `<name/>` and return an Array of React Elements.
```
schalkneethling: Could we add an example here in the comment of a simple use case? Until I saw it in action in the newsletter component I was not 100% sure of how to use it.

kuma/javascript/src/l10n.js
```diff
@@ -136,3 +137,32 @@ export function interpolate(s: string, args: Array<any> | { [string]: any }) {
...
+ */
+export function Interpolated({
+    id,
+    ...args
+}: {
+    id: string,
+    [key: string]: React.Node
+}): React.Node[] {
+    return id.split(ELEMENT_REGEXP).map((str, i) => {
+        if (!ELEMENT_REGEXP.test(str)) {
```
schalkneethling: It took some reasoning for me to completely understand what is going on here and in the next conditional :) Maybe it is just me but, perhaps it is worth adding some comments to explain what it does. Something like:

```
/* Tests whether there is an element in the string, for example
   ELEMENT_REGEXP.test("This is just a string"); // false 
   ELEMENT_REGEXP.test("<elem/>"); // true */
```

kuma/javascript/src/l10n.js
```diff
@@ -136,3 +137,32 @@ export function interpolate(s: string, args: Array<any> | { [string]: any }) {
...
+            return str;
+        }
+
+        const element = args[str.slice(1, -2).trim()];
```
schalkneethling:
```
/* This string refers to an element, but is there a corresponding element in args?
   `args` contain: privacyLink={<a href="#here">{gettext('Hello')}</a>}
   args["<footer/>".slice(1, -2).trim()]; // false
   args["<privacyLink/>".slice(1, -2).trim()]; // true
```

kuma/javascript/src/l10n.test.jsx
```diff
@@ -152,3 +160,28 @@ describe('interpolate()', () => {
...
+        const tree2 = renderer
+            .create(
+                <Interpolated
+                    id="wow <head/> such <body /> many <foot />"
+                    head={<h1>Beautiful</h1>}
+                    foot={<p className="something">more</p>}
+                />
+            )
+            .toJSON();
+        expect(tree2).toMatchSnapshot();
```
schalkneethling: Perhaps we can add another test where just a string is passed and we can then confirm that the string is returned unchanged?

Gregoor:
Great, thank you very muh for the thoughtful comments, I agree with all of them and worked them in (pretty much took over your trim-comment verbatim).

peterbe:
Don't merge this yet! I'm reaching out to Adrian on the L10n team for his opinion.

peterbe:
> @peterbe wdyt of us merging this one in? I see that it doesn't cover all the bases but imo it covers enough use-cases to be worth it.

Let's do it!

Suppose we get our act together and fix the validation of translated strings, then we can safely sleep at night when the `gettext` string contains HTML like `gettext('Check <b>this</b> out!')`. 

However, how would that work when the HTML is non-trivial? E.g.
```
<div dangerouslySetInnerHTML={{__html: interpolate('See our <a href="%{url}">Privacy Policy</a>.', {url: 'https://www.mozilla.org'})}}/>
```
But with *your* PR here, independent of safety, translators would be seeing this in Pontoon instead:
```
See our <link>Privacy Policy</link>
```
right? That seems worth it.

Gregoor:
Correcto, that's what they'd be seeing! I'll resolve the conflict and merge it in then.

PR #5873, opened 2019-09-25
Componentize SearchResultsPage

Gregoor:
[As requested](https://github.com/mozilla/kuma/pull/5770#issuecomment-534527385)

This extracts a lot of the SearchResultsPage render code into separate components, making the control fow in it easier to follow.

### DIFF
--- a/kuma/javascript/src/search-results-page.jsx
+++ b/kuma/javascript/src/search-results-page.jsx
@@ -47,7 +47,7 @@ function makePaginationPageURL(uri) {
 
 function ResultsMeta({
     locale,
-    results
+    results: { count, previous, next, query, start, end }
 }: {
     locale: string,
     results: SearchResults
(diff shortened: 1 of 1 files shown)


kuma/javascript/src/search-results-page.jsx
```diff
@@ -45,160 +45,162 @@ function makePaginationPageURL(uri) {
...
+                    ),
+                    {
+                        count: results.count.toLocaleString(),
+                        // XXX this 'locale' is something like 'en-US'
+                        // need to turn that into "English (US)".
+                        locale: locale,
+                        query: results.query
+                    }
+                )}{' '}
+                {!!results.count &&
```
peterbe: Can you do the same technique here. So...
```jsx
{!!results.count && (
  <>
  </>
)}
```

peterbe: Also, I think the logic should be:
```
if (results.count) {
  if (results.previous || results.next) {
      "Showing results ..."
  } else {
     "Show all results."
  }
}
```
See how nicer it is with "traditional" if statement blocks :)

Gregoor:
True, good catch! Thanks. It does look a little convoluted now, maybe I will pivot to an if-else in that case

PR #5874, opened 2019-09-25
Submodule: locale c6a3df5...3cb75f0

peterbe:


peterbe:
@escattone This was made using `kuma-deployer ~/kuma l10n`
but because I stumbled some obstacles so to be able to re-run it I had to comment out stuff that can't be done twice and ran it till it worked.

PR #5877, opened 2019-09-26
React error boundaries

### LINKED ISSUE #5720: React error boundaries
We won't catch every runtime client-side React bug in advance. 
When it happens, we should ideally be sending the exception to something like Sentry, but we should also have a "pretty" error page like we have in Django for 500.html errors. 

This is what it looks like if an error happens (in this example, in `article.jsx`)

Screen Shot 2019-08-28 at 2 08 43 PM

If it's not weird enough already, it'll be even worse once the footer is a React component.

### LINKED ISSUE #5913: Unit test error boundaries
https://github.com/mozilla/kuma/issues/5720 is about adding an error boundary. We should have some unit tests of these.

peterbe:
Fixes #5720 
Fixes #5913

When the whole app fails:
Screen Shot 2019-09-26 at 12 04 31 PM

When the `Content` component fails:
Screen Shot 2019-09-26 at 12 04 45 PM

### DIFF
--- a/kuma/javascript/src/error-boundaries.jsx
+++ b/kuma/javascript/src/error-boundaries.jsx
@@ -1,0 +1,95 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+import { gettext } from './l10n.js';
+
+export class AppErrorBoundary extends React.Component {
+    state = { error: null };
+
+    static getDerivedStateFromError(error) {
+        return { error };
+    }
+
+    logError() {
+        // https://github.com/mozilla/kuma/issues/5833
+        // When we start to use Sentry, change the method signature to
+        // logError(error, errorInfo) {
+    }
+
+    componentDidCatch(error, errorInfo) {
+        document.title = 'Application rendering Error';
+        this.logError(error, errorInfo);
+    }
+
+    render() {
+        if (this.state.error) {
+            return (
+                <ErrorWrapper title={gettext('Application rendering error')}>
+                    <p>
+                        {gettext(
+                            'An unhandled error occurred in the application. The error has been logged and an administrator was notified. We apologize for the inconvenience!'
+                        )}
+                    </p>
+                </ErrorWrapper>
+            );
+        }
+
+        return this.props.children;
+    }
+}
+
+AppErrorBoundary.propTypes = {
+    children: PropTypes.element.isRequired
+};
+
+export class ContentErrorBoundary extends AppErrorBoundary {
+    componentDidCatch(error, errorInfo) {
+        document.title = 'Content rendering error';
+        this.logError(error, errorInfo);
+    }
+    render() {
+        if (this.state.error) {
+            return (
+                <ErrorWrapper title={gettext('Content rendering error')}>
+                    <p>
+                        {gettext(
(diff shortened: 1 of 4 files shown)


peterbe:
Note, I didn't bother with Flow. Perhaps this can be a follow-up issue. 
Also, no unit tests. Also a follow-up issue.

atopal:
Can we also get a GA error for these cases? It's much easier to work with that for me.

peterbe:
> Can we also get a GA error for these cases? It's much easier to work with that for me.

Just like we have a separate issue just for Sentry sending, I can make a separate issue for doing a GA sending thing.

peterbe:
I think I cut corners on this one and I don't like it. I shouldn't give up on Flow and jest testing. We have a deadline but I don't know if it's fair to block that on better error reporting. 

I'm going to take this back and try to add...

* Flow
* jest tests
* GA tracker

peterbe:
@Gregoor We had the MDN Dev sprint planning meeting and chatted about this. Time is of the essence and the team had me convinced. 

I'm going to write follow-up issues that worries about using GA tracker and one that worries about Flow and one that  worries about a jest unit test of the component. 

The truth is, without this PR landed as soon as possible, we won't have any Sentry on client-side errors. And we generally feel like we need tthis to launch the React site with confidence.

kuma/javascript/src/index.jsx
```diff
@@ -33,88 +34,104 @@ if (container) {
...
-                        initialURL={data.url}
-                        initialData={data.documentData}
-                    />
-                </React.StrictMode>
+                <SinglePageApp
+                    initialURL={data.url}
+                    initialData={data.documentData}
+                />
             );
+            // app = (
```
Gregoor: I like what you've done with moving StrictMode-wrapping downstairs, but I'd vote for also deleting the commented out code.

kuma/javascript/src/error-boundaries.jsx
```diff
@@ -0,0 +1,95 @@
...
+    componentDidCatch(error, errorInfo) {
+        document.title = 'Application rendering Error';
+        this.logError(error, errorInfo);
+    }
+
+    render() {
+        if (this.state.error) {
+            return (
+                <ErrorWrapper title={gettext('Application rendering error')}>
+                    <p>
```
Gregoor: nit: Is there a reason for why we're supplying children to this component when they're the same in both instances?

peterbe:
> Can we also get a GA error for these cases? It's much easier to work with that for me.

https://github.com/mozilla/kuma/issues/5929

kuma/javascript/src/error-boundaries.jsx
peterbe: No reason. I think I had bigger ideas and intentions but I've run out if ideas :)
Also, using the same English sentence for application errors and content errors makes for *fewer* things that need to be translated. I'm going to move this to the `<ErrorWrapper/>`

By the way, what do you think about the name `ErrorWrapper`?

kuma/javascript/src/index.jsx
peterbe: This is one of the many reasons why code reviews are good.

Issue #5798, opened 2019-09-11
Fix uploaded file attachments in MDN.


tobinmori:
https://bugzilla.mozilla.org/show_bug.cgi?id=1110801

This can and has bit us. If attachments are served via Django it can kill our server.

Currently attached files land in EFS and Django is forced to serve these files (a process killer); if the uploaded files landed in S3, they could be served by S3. 

The offending code is here:
https://github.com/mozilla/kuma/blob/4e84a8c2a0d1fd59c736a4f88fbecc80e1b01cb6/kuma/attachments/views.py#L46

- [ ] Not using EFS, but using S3 instead
- [ ] Fix the CloudFront TTL to something other than 15min (much less important)
- [ ] Migrate existing EFS content from EFS to S3

jmswisher:
To be continued

jmswisher:
@callahad-zenhub What is this waiting on?

callahad:
@jmswisher There's a checklist in the first comment on the associated pull request (https://github.com/mdn/kuma/pull/5998). The main blocker is (https://github.com/mdn/infra/issues/336).

tobinmori:
Can i close this one out @callahad?

PR #5882, opened 2019-09-27
Add Constance flag for disabling attachment upload

callahad:
This defines a new Constance flag (`WIKI_ATTACHMENTS_DISABLE_UPLOAD`)
which disables adding or editing attachments via the Wiki. Changes may
still be made via the Django Admin.

We need this capability for the EFS-to-S3 migration (#5798).

### DIFF
--- a/kuma/attachments/tests/test_models.py
+++ b/kuma/attachments/tests/test_models.py
@@ -1,5 +1,6 @@
 import datetime
 
+from constance.test import override_config
 from django.contrib.auth.models import Group, Permission
 from django.contrib.contenttypes.models import ContentType
 from django.core.files.base import ContentFile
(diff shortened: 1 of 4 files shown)


peterbe:
@callahad Would you mind using your fork to send PRs from? ...instead of this repo. I'm sure your diligent to delete your branches when you're done but if you miss, it'll linger and nobody's going to remember "What was this branch for???"

callahad:
@peterbe 

> Would you mind using your fork to send PRs from?

Sure, I'll switch to using a personal fork.

> it's kinda harsh that it doesn't say something like "File attachment upload is currently disabled for a migration thing we're working on"

Yeah, that's just our default behavior whenever a user isn't allowed to add attachments: [don't show the button](https://github.com/mozilla/kuma/blob/8d9d31e634040e0def62fb1ab5b915934914cbd4/kuma/attachments/jinja2/attachments/includes/attachment_list.html#L1). If we think we'll use this more than once or twice in the future, then we could update that template to have a disabled button with a message... but I suspect it will only be used very rarely.

PR #5884, opened 2019-09-30
Remove unused settings constants

callahad:
I couldn't find any reference to these in our code, or our dependencies.

Might as well remove them, right? :)


callahad:
I would absolutely appreciate additional review before merging.

This was basically a series of "huh, really?" moments while reading `common.py`, followed by a bunch of `rg --no-ignore CONSTANT_NAME` to search both the Kuma repo and my virtualenv.

PR #5885, opened 2019-09-30
Fix a few deprecation warnings for Django 2.0

callahad:
"Direct assignment to a reverse foreign key or many-to-many relation" [was deprecated](https://docs.djangoproject.com/en/1.11/releases/1.10/#features-deprecated-in-1-10) in Django 1.10.

Switching to `.set(...)` clears up half a dozen test warnings.

### DIFF
--- a/kuma/attachments/tests/test_models.py
+++ b/kuma/attachments/tests/test_models.py
@@ -136,12 +136,12 @@ class AttachmentModelTests(UserTestCase):
 
         # Create a group with the negative permission.
         g1, created = Group.objects.get_or_create(name='cannot_attach')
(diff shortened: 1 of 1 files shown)


peterbe:
FYI, I've been using this bug to kill a couple of other warnings. 
https://bugzilla.mozilla.org/show_bug.cgi?id=1467529

But generally, all of that is moot until we switch to Python 3 :)

callahad:
> Was that all?!

Yep. Takes us down to 9 warnings from 21. Was getting tired of scrolling past all that logging spew when testing the S3 work. :)

escattone:
Thanks @callahad !

PR #5887, opened 2019-09-30
Improve display of URLs in search results

### LINKED ISSUE #5658: Full URLs in search results make it difficult to read
Right now our search results take the form:

> __Title__
> *https://developer.mozilla.org/en-US/docs/Example*
> Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
> *...Duis aute irure __dolor__ in reprehenderit in voluptate velit...*
> *...esse __cillum__ dolore eu fugiat nulla pariatur. Excepteur sint...*
> *...occaecat cupidatat non __proident__, sunt in culpa qui officia...*

For whatever reason, I find the long, full URLs to be distracting and difficult to read. Especially since the first 41 characters (`https://developer.mozilla.org/en-US/docs/`) are the same for every result.

I suggest we abbreviate the URLs to emphasize just the document slug (e.g., `/Web/API/WebRTC_API`, the only part of the URL that varies between results.

Comparison screenshot:

Full URL versus Document Slug

Our options are:

1. Full URL *(status quo):*
    `https://beta.developer.mozilla.org/en-US/docs/Web/API/WebRTC_API`

2. Schemaless URL:
    `beta.developer.mozilla.org/en-US/docs/Web/API/WebRTC_API`

3. Absolute Path:
    `/en-US/docs/Web/API/WebRTC_API`

4. Slug Only:
    `/Web/API/WebRTC_API`

In all cases, the patch is a completely trivial modification to [search-results-page.jsx](https://github.com/mozilla/kuma/blob/d51048f1a8c5a608266fd798a7643f462acb42cc/kuma/javascript/src/search-results-page.jsx#L108).

callahad:
Omits the schema / host / path prefix that is common to all search
results in favor of just highlighting the part of the URL that changes
from result to result: the document slug.

Fixes #5658

Before: 
> Screenshot of status quo, showing full URLs in search results

After:
> Screenshot of proposed modification

### DIFF
--- a/kuma/javascript/src/search-results-page.jsx
+++ b/kuma/javascript/src/search-results-page.jsx
@@ -95,7 +95,6 @@ function Results({
 }) {
     return (results.documents || []).map(result => {
         const path = `/${locale}/docs/${result.slug}`;
-        const url = window && window.origin ? `${window.origin}${path}` : path;
         return (
             <div className="result-container" key={result.slug}>
                 <div className="result">
@@ -111,7 +110,7 @@ function Results({
                         }}
                     />
                     <div className="result-url">
-                        <a href={path}>{url}</a>
+                        <a href={path}>{path}</a>
                     </div>
                 </div>
             </div>


kuma/javascript/src/search-results-page.jsx
```diff
@@ -111,7 +110,7 @@ function Results({
                         }}
                     />
                     <div className="result-url">
-                        <a href={path}>{url}</a>
+                        <a href={path}>{result.slug}</a>
```
peterbe:
```suggestion
                        <a href={path}>{path}</a>
```
Otherwise there might be to identical strings there. E.g. `Web/HTML/Css` and `Web/HTML/Css`. But if you hover over the links you see that the actual URLs are `/en-US/docs/Web/HTML/Css` and `/sv-SE/docs/Web/HTML/Css`.
The Wiki site-search has slightly better messaging but basically, if you're using https://developer.mozilla.org/sv-SE/search?q=stuff&topic=apps&topic=html&topic=css&topic=js&topic=api&topic=canvas&topic=svg&topic=webgl&topic=mobile&topic=webdev&topic=http&topic=webext&topic=standards it will search Elastisearch for `{q: "stuff", locale: 'sv-SE' || 'en-US'}` so you'll get a little for each locale.

callahad: Done :) Hopefully breadcrumbs are somewhere in our future, but this is a small improvement in the meantime.

peterbe:
I think you accidentally checked in a movement of the git submodule pointers. If you need help, I can rubberduck with you to untangle that.

callahad:
This was fixed in f305d59c68a3a2a46145f9163b1c5c6af57e9ec9

PR #5899, opened 2019-10-01
Server-side generate site-search

### LINKED ISSUE #5824: Server-side generate site-search
Originally from: https://bugzilla.mozilla.org/show_bug.cgi?id=1567487

The site-search page is slow because it depends on first downloading a "blank skeleton", then downloading the React bundle, and then React can paint `<h3>Results: {searchTerm}</h3>` whilst it waits for the actual XHR to do the fetch. 

Screen-Recording-2019-09-13-at-12 09 03-PM

joedarc:
Fixes #5824 

When the search results page is rendered with react, the results are pre-loaded using SSR.

### DIFF
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -1,3 +1,6 @@
+from collections import OrderedDict
+from operator import attrgetter
+
 from django.conf import settings
 from django.http import HttpResponsePermanentRedirect, JsonResponse
 from django.shortcuts import get_object_or_404
@@ -6,6 +9,7 @@ from django.views.decorators.cache import never_cache
 from django.views.decorators.http import require_GET
 from rest_framework import serializers, status
 from rest_framework.decorators import api_view
+from rest_framework.generics import ListAPIView
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
 from waffle.decorators import waffle_flag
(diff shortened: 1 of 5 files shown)


peterbe:
I haven't fully started reviewing this yet but here's how it looks like:

Screen Shot 2019-10-08 at 12 21 32 PM

Here's when I use `links https://beta.developer.mozilla.org/en-US/search\?q\=search`:
Screen Shot 2019-10-08 at 12 23 10 PM

peterbe:
@escattone Are you OK with the reshuffle of the files? He did what had to be done to be able to have **1** Django view that can deal with `/en-US/search?q=css` and within it needs to "fork off" to the respective implementations depending on wiki or not wiki. 
It looks a bit weird from there because the actual search implementation isn't a plain and pure Python function but rather a Django view class. Technically, the wiki has at its core a view function class that takes in the request and combines it with elasticsearch-dsl to produce a response. 

Either way, I'm a little be hesitant to merge it today in case we need to keep `master` ready for some quick fail-forwards.

escattone:
> @escattone Are you OK with the reshuffle of the files?

Yes, no problem, thanks @peterbe.

schalkneethling:
I have nothing to add here, this looks all good! Thanks @joedarc 🍕

PR #5907, opened 2019-10-01
Remove use of Django 1's `allow_tags` attribute

callahad:
The `allow_tags` attribute on `ModelAdmin` methods was removed in Django 2.0
(https://docs.djangoproject.com/en/dev/releases/2.0/#deprecated-features-2-0)

This replaces all instances of `allow_tags` in our codebase, eliminating
the associated RemovedInDjango20Warning. I have manually verified that
every call to an affected method continues to render correctly.

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1467529

### DIFF
--- a/kuma/wiki/admin.py
+++ b/kuma/wiki/admin.py
@@ -9,7 +9,7 @@ from django.http import HttpResponseRedirect
 from django.template.defaultfilters import truncatechars
 from django.template.response import TemplateResponse
 from django.utils import timezone
-from django.utils.html import escape, format_html, format_html_join
+from django.utils.html import format_html, format_html_join
 from django.utils.safestring import mark_safe
 from django.utils.text import Truncator
 from django.views.decorators.cache import never_cache
@@ -138,10 +138,9 @@ def related_revisions_link(obj):
     )
     count = obj.revisions.count()
     what = (count == 1) and 'revision' or 'revisions'
-    return '<a href="%s">%s&nbsp;%s</a>' % (link, count, what)
+    return format_html('<a href="{}">{}&nbsp;{}</a>', link, count, what)
 
 
-related_revisions_link.allow_tags = True
 related_revisions_link.short_description = "All Revisions"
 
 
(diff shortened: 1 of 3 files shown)


callahad:
This drops us from 9 warnings to 7 when running our Python test suite.

kuma/wiki/admin.py
```diff
@@ -211,30 +212,36 @@ def topic_sibling_documents_link(obj):
...
-    out = (x(obj) for x in funcs)
-    return '<ul>%s</ul>' % ''.join('<li>%s</li>' % x for x in out if x)
+    results = []
+
+    for func in funcs:
+        result = func(obj)
+        if result:
+            results.append(format_html('<li>{}</li>', result))
+
+    return format_html('<ul>{}</ul>', mark_safe(''.join(results)))
```
callahad: It should be possible to significantly condense this function by using `format_html_join`, but that function is not working as expected for me. So I took a more imperative approach here. :(

The following should work, but doesn't.

```python
def combine_funcs(obj, funcs):
    """Combine several field functions into one block of lines"""
    items = (func(obj) for func in funcs)
    list_body = format_html_join('', '<li>{}</li>', (item for item in items if item))
    return format_html('<ul>{}</ul>', list_body)
```

Using that snippet, each list item is rendered as `<li>&lt;</li>` for God knows why. The input (`items`) looks correct. Each item has an appropriate `__html__` attribute which produces the expected string. But running them through `format_html_join` breaks everything. 🤷‍♂.

callahad: It should be possible to significantly condense this function by using `format_html_join`, but that function is not working as expected for me. So I took a more imperative approach here. :(

The following should work, but doesn't.

```python
def combine_funcs(obj, funcs):
    """Combine several field functions into one block of lines"""
    items = (func(obj) for func in funcs)
    list_body = format_html_join('', '<li>{}</li>', (item for item in items if item))
    return format_html('<ul>{}</ul>', list_body)
```

Using that snippet, each list item is rendered as `<li>&lt;</li>` for God knows why. The input (`items`) looks correct. Each item has an appropriate `__html__` attribute which produces the expected string. But running them through `format_html_join` breaks everything. 🤷‍♂.

callahad: Oh! Figured it out. One sec...

PR #5910, opened 2019-10-02
Avoid deprecated `Manager.use_for_related_fields`

callahad:
Fixes another `RemovedInDjango20Warning`: use_for_related_fields was
deprecated in Django 1.10, and removed in Django 2.0:
https://docs.djangoproject.com/en/dev/releases/1.10/#manager-use-for-related-fields-and-inheritance-changes


callahad:
Taken together with #5907, this drops us down to 5 warnings in our test suite, all in external dependencies.

Those can be resolved by:
- Upgrading [django-recaptcha](https://github.com/praekelt/django-recaptcha)
- Upgrading [django-decorator-include](https://github.com/twidi/django-decorator-include) (Must happen at the same time as the move to Django 2)
- Patching [django-soapbox](https://github.com/ubernostrum/django-soapbox)

callahad:
Oh, cool. This is should be a simple, mechanical transformation of the code to resolve... but it breaks a database migration, so something's wrong.

Will revisit later (or anyone else is welcome to pick it up)

PR #5928, opened 2019-10-03
Fixes #5927, remove compact.scss from codebase

### LINKED ISSUE #5927: Remove components/compact.scss from codebase
The Sass file `kuma/static/styles/components/compact.scss` is hanging around the codebase with no purpose. It is one less Sass file I(@schalkneethling) have to think about ;p

schalkneethling:
Delete unused Sass file @callahad r?


schalkneethling:
Somehow I missed the two `main.scss` files where this file was imported but, doing a search across production, there is nothing that uses this class any longer:

https://developer.mozilla.org/en-US/search?css_classnames=compact&locale=*

This is still good for review.

callahad:
Reads fine. Will spin it up locally and review tonight or first thing
tomorrow morning

schalkneethling:
> Looks like there are still two places where we need to remove the `compact` class (and verify that they render correctly without it).
> 
> ```
> kuma/wiki/jinja2/wiki/includes/kumascript_errors.html
> 26:          <dl class="compact">
> 
> kuma/dashboards/jinja2/dashboards/includes/revision_dashboard_body.html
> 53:                                <dl class="revision_ip compact hidden">
> ```

Good catch @callahad 

Hmmm, the stock standard way I used to find instances where this is used clearly failed me(need to improve that). Looking at this now, I am only going to remove it from the new frontend where we will not see there KumaScript errors.

callahad:
Sounds good. Honestly not sure we need it for the KumaScript errors; they're collapsed by default:

Minimized KumaScript error with prominent button labeled "More information about this error"

And I'm not sure the `compact` class really makes enough of a difference to justify keeping it around.

<details><summary>Expand for screenshots</summary>
With `compact` (default):

Screenshot

Without:

Screenshot
</details>

The revisions dashboard, however, becomes more difficult to read. Though we could easily address that by tweaking the markup instead. Or inlining that stylesheet just on the dashboard pages? (Do we have a good, established way to do that?) It's a tiny thing, but no sense shipping it to everyone as part of the main CSS bundle.

<details><summary>Expand for screenshots</summary>

With `compact` (default):

Screenshot


Without:

Screenshot

</details>

schalkneethling:
Thank you for the review @callahad - I brought `compact.scss` back, but I am now only including it for the dashboards bundle. I removed the class from the KumaScript error include and kept the rest as it was before.

PR #5934, opened 2019-10-04
Load BCD table CSS after content

### LINKED ISSUE #5669: Lazy load wiki-compat-tables.css
Instead of making a big bundle, we should lazy load just the CSS needed for pages with a BCD table.

Gregoor:
Fixes #5669 

Wasn't sure if there's a preferred way of deferring this load, so I just moved it into its own style bundle that'll be included after content (but before the JS bundle).

### DIFF
--- a/kuma/wiki/jinja2/wiki/react_document.html
+++ b/kuma/wiki/jinja2/wiki/react_document.html
@@ -118,7 +118,7 @@
   {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(),
                   document_data)|safe }}
 
-  <!-- site js -->
+  {% stylesheet 'react-bcd' %}
   {% javascript 'react-main' %}
   <script>
     if (window.mdn && mdn.analytics) mdn.analytics.trackOutboundLinks();
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -690,9 +690,6 @@ PIPELINE_CSS = {
             'styles/libs/prism/prism-line-numbers.css',
 
             'styles/wiki-syntax.scss',
-
-            # Styles for BCD tables
-            'styles/wiki-compat-tables.scss',
         ),
         'output_filename': 'build/styles/react-mdn.css',
     },
@@ -701,6 +698,12 @@ PIPELINE_CSS = {
             'styles/minimalist/search-page.scss',
         ),
         'output_filename': 'build/styles/react-search.css',
+    },
+    'react-bcd': {
+        'source_filenames': (
+            'styles/wiki-compat-tables.scss',
+        ),
+        'output_filename': 'build/styles/react-bcd.css'
     },
     'mdn': {
         'source_filenames': (


kuma/wiki/jinja2/wiki/react_document.html
```diff
@@ -118,7 +118,7 @@
   {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(),
                   document_data)|safe }}
 
-  <!-- site js -->
+  {% stylesheet 'react-bcd' %}
```
peterbe: Won't this just put a `<link rel=stylesheet>` within the body? So it'll block rendering, no?

Gregoor: Yes it would block for everything after, but since all the content is above this should be fine, no?

peterbe: Not so sure. It depends on the browser I think. I'll have to refresh my memory on the details of it.

Gregoor: Built a little example. Interesting, Chrome indeed behaves differently than FF, namely it doesn't render/load what comes after until the request is resolved:
https://glitch.com/edit/#!/melodious-globe?path=web/index.html:20:4

Gregoor: Was the concern that we're blocking loading of the js bundle?

peterbe: What makes me nervous is that this just doesn't feel natural. In soccer/football, the goalie is critical (like a `<link rel=stylesheet>`) but at kick-off, the ball is far away from the goal area. The coach could say "Our goalie isn't here yet. But go ahead and kick off the game without him." Some observers of the game might think "To me, the game hasn't started till the goalie is here so I'm not going to even look until it's all there."

There's also the rather strange effect that the footer is below your prosed `<link rel=stylesheet>` so that'll be blank too in FF. Just because of the BCD table. 

Who knows what other stuff the browsers might hold back until the CSSOM is complete?? Perhaps it refuses to load some images or iframes. It doesn't feel like we know how all these things are going to work universally. It might be one thing to test in latest FF and latest Chrome on desktops but there are so many more browsers. 

The ideal thing is to NOT display the BCD table until AFTER its CSS has loaded and that CSS should be loaded long after the CSSOM is finished rendering. 

We *could* do this:

```html
<link rel="preload" href="/static/bcd.css">
<p id="bcd-table-loading">Loading BCD table</p>
<table style="display:none" id="bcd-table"> ... </table>
```
```js
window.onload = () => {
  link = document.querySelector('link[rel="preload"]')
  link.onload = () => {
      document.querySelector('#bcd-table-loading').style.display= 'none';
      document.querySelector('#bcd-table').style.display= 'block';
  }
}
```
In principle (ie. theory), this is the best practice. 
* It's self-contained to just the BCD thing.
* It doesn't mess with the CSSOM in any way we're not confident about. 
* It's also not displaying an unstyle BCD table.
* It's definitely removing `bcd.css` from the critical rendering path. (aka. Critical Request Chain as they call it in Chrome Lighthouse)
* We're using the DOM events to take our place in the queue instead of hoping for attention.

peterbe: I'm not convinced my solution is ideal either but for the sake of completeness here is a version of it: https://careful-coelurus.glitch.me/  (I changed the `/styles.css` to load in 4s)

It works ideal in FF and Chrome but not in Safari. I need to investigate that. What does it do in the latest IE? (I don't know what IE was called before it switched to Chrome)

peterbe: Actually, this is getting ugly in Safari. The only way to make work exactly like FF and Chrome is to put the `<link rel="preload" href="style.css" as="style" />` right at the end of the body before the `<script>` tags. Ugly!

peterbe: Yikes! I even tried https://master-origin-loadcss.fgview.com/test/preload.html but that doesn't work either.

peterbe: Oh! I got it! Now it works ideal in Safari too. This latest iteration was that instead of doing
```javascript
const link = document.querySelector('link[rel="preload"]');
link.rel = 'stylesheet';
```
we instead do this:
```javascript
const inspiration = document.querySelector('link[rel="preload"]');
const link = document.createElement('link')
link.href = inspiration.href;
link.rel = 'stylesheet';
document.body.appendChild(link);
```

Now you get the same wonderful and desired effect in all my 3 browsers.

Gregoor: Great stuff, thanks Peter! Also works in IE11.

I see your point about a `<link>` at this position in the DOM seeming unnatural (despite me being more of a handball-guy). 

>     * It doesn't mess with the CSSOM in any way we're not confident about.

Don't quite understand how this version is different in this regard with Non-JS one

>     * We're using the DOM events to take our place in the queue instead of hoping for attention.

Well, we're using the DOM position to queue. Imo the two don't differ in this regard.

A downside of both approaches: They relayout. But only something like CSS modules / Emotion could help for optimally spitting out styles and elements to fix that I suppose.

I do like that the original version doesn't require JS to work. I'm unsure which I like better overall atm.

schalkneethling: I had a stab at it using essentially what Peter has done here:
https://capable-flannel.glitch.me

This works beautifully in Chrome and Safari but does not work in Firefox. It seems like Firefox does not yet support `preload` and as such the `onload` is never fired. A polyfill will solve this though:
https://github.com/filamentgroup/loadCSS/blob/v2.0.1/src/cssrelpreload.js

As can be seen here https://capable-flannel.glitch.me/polyfilled

peterbe: > https://capable-flannel.glitch.me

That one seems scary to me. You're relying on hoisting in a DOM parsing. I.e. `activateBCD(this)` happens on line 8 but `function activateBCD(style) {` happens on line 67. 
I'm surprised it's not triggering an error in the web console.

peterbe: @schalkneethling I think in both of your solutions, the `styles.css` is loaded twice in FF and Chrome. (haven't tested with Safari)

schalkneethling: Oh, and can be moved anywhere really. It does not have to be at the bottom. Also, would be really odd if it does actually download twice. The entire thing is based on this technique here:
https://www.filamentgroup.com/lab/async-css.html#a-modern-approach

schalkneethling: Should it not be preloaded, then cached so, even if the browser does try to request it again when the `rel` attribute value changes, it should grab it from the browser cache and not do another network request.

peterbe: Doing two downloads kinda makes sense. The `preload` is not a guarantee. The browser's promise is "I'll try to download it in the background when I have some time available so that if you ever need it I'll have it cached."
After all, the preload is just an optimization bonus in a sense. Like a hint.

schalkneethling: well true, but the function should only be called once the CSS file has loaded. If that happens twice, the function should be called twice no?

Gregoor: It took a while, but I now appreciate `preload`. In the example glitch page it doesn't make a difference for me in terms of speed (in Chrome), but I can see that prioritization does not affect a one-ressource situation.
Something I still don't get:
Why the whole `<noscript>` and `style.rel = "stylesheet";` dance? Couldn't we just have the two link-tags, one with `rel="preload"` and one with `rel="stylesheet"`, like it is in the MDN example (plus the shim for non-blink/webkit browsers)?

And wrt the table being hidden by default, I guess there's two options
- do it on the server side, which would require some HTML parsing, element finding and inline style setting. On the client, to make it work on noscript browsers we'd have to add a `display: block !important;` to our CSS
- do it on the client side, attach a MutationObserver that hides the BCD table initially when it appears and unhides it when its CSS has loaded.

I prefer option two because it feels more straightforwad

schalkneethling: > Why the whole <noscript> and style.rel = "stylesheet"; dance?

So, one of the main benefits of `preload` on a CSS file is, from the docs:
https://w3c.github.io/preload/

> initiates an early fetch and separates fetching from resource execution.

So, until we do the switcheroo from `rel="preload"` to `rel="stylesheet"` the browser does not execute the CSS and thus there is no render-blocking. To also answer @peterbe question/comment earlier, unlike `prefetch`, `preload` is not merely a hint to the browser, from the docs:

> the application can use the preload keyword to initiate early, high-priority, and non-render-blocking fetch of a CSS resource that can then be applied by the application at appropriate time 

However, seeing that we rely on JS to do the switcheroo at the appropriate time, a user with JS disabled will never have the CSS applied and so we need the "normal" `link` inside the `noscript` tag

> And wrt the table being hidden by default

I would also vote for option 2 but, what is problematic with regards to using `display: none;` and then `display: block;`?

Gregoor: > So, until we do the switcheroo from `rel="preload"` to `rel="stylesheet"` the browser does not execute the CSS and thus there is no render-blocking.

I guess what I was trying to get at is:
Why do all that with JS and `noscript` when just having both the `rel="preload" and the `rel="stylesheet"` tags behaves the same (like in this example: https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content)? Or maybe I'm missing some nuance here?

> I would also vote for option 2 but, what is problematic with regards to using `display: none;` and then `display: block;`?

I was thinking we'd need `!important` which would make it less preferable but thinking about it more I guess there's other, less-dirty more-cascady ways of achieving the display-un-noneing.

peterbe:
My current feeling is that we take a big pause on *this* issue/PR here and move the topic over to Stumptown-renderer which'll soon need a similar solution. stumptown-renderer doesn't even have lazy loading BCD table but I'm confident it will. Just a matter of time. 

Kuma is now fast enough. I doubt this here is considered a low-hanging fruit fix. Clearly, we've demonstrated arguments for and against that. 

The reason it's important is that if we *don't* have a really solid solution to *full* lazy loading where we lazy load CSS *and* JS is if we don't we're going to have to fear writing exactly what (S)CSS and JS we want for fear of bloating bundles that everyone has to download. Stumptown is a place to try to get things right from the start even though at this stage it might feel like premature optimization. 

In a not-so-distant future, Stumptown-renderer is going to have the ability to ship a full solution and really taking it for a spin. We'd be able to test these things in Speedcurve, Webpagetest, etc.

Gregoor:
I read through all the threads again to refresh my memory and looked at the savings. @peterbe's conclusion still coheres to me, so I'm gonna close this PR. Shall we also close the related issues then?

PR #5937, opened 2019-10-05
Remove html5shiv from template (#5905)

### LINKED ISSUE #5905: Stop shipping html5shiv
In the head of all document we have the following that targets all versions of IE:

```
<!--[if IE]>
  <meta http-equiv="imagetoolbar" content="no">
  {% javascript 'html5shiv' %}
<![endif]-->
```

We *really* do not need this any longer, especially once we have switched to a "reader mode" for `lte IE10`.

nsiregar:
Remove
```
<!--[if IE]>
  <meta http-equiv="imagetoolbar" content="no">
  {% javascript 'html5shiv' %}
<![endif]-->
```
from jinja template as in issue #5905 

Fixes #5905


schalkneethling:
Thanks for the pull request @nsiregar - You can also go ahead and remove the html5shiv file from the codebase here:
https://github.com/mozilla/kuma/tree/master/kuma/static/js/libs/html5shiv

You can go ahead and remove the entire directory.

nsiregar:
Updated, html5shiv removed from codebase

schalkneethling:
@nsiregar Found another place in the codebase that needs cleanup. You can remove the following from `common.py`

```
'html5shiv': {
        'source_filenames': (
            'js/libs/html5shiv/html5shiv.js',
        ),
        'output_filename': 'build/js/html5shiv.js',
    },
```

https://github.com/mozilla/kuma/blob/master/kuma/settings/common.py#L1102

PR #5938, opened 2019-10-05
Remove wiki/article-head from new frontend (#5924)

### LINKED ISSUE #5924: Remove wiki/article-head from new frontend
In `kuma/static/styles/minimalist/document.scss` we currently include `kuma/static/styles/components/wiki/article-head.scss`. This has no relevance to the new frontend and should thus be removed.

## Potential saving

+--------------------------------------------------------------+
| Size         | 73 bytes                                      |
|--------------------------------------------------------------|
| Gzipped      | 80 bytes                                      |
|--------------------------------------------------------------|

nsiregar:
Remove `wiki/article-head.scss` from `kuma/static/styles/minimalist/document.scss`

Fixes #5924

### DIFF
--- a/kuma/static/styles/minimalist/document.scss
+++ b/kuma/static/styles/minimalist/document.scss
@@ -44,7 +44,6 @@ Components that may appear on most wiki pages
 Article meta
 - contains breadcrumbs and page buttons.
 -------------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


MatonAnthony:
I think a yellow or orange colour would look great.

PR #5944, opened 2019-10-07
GA track errors from error boundaries

### LINKED ISSUE #5929: GA track errors from error boundaries
Once https://github.com/mozilla/kuma/issues/5720 lands we should add a Google Analytics `ga.trackError` thing. 

Look at https://github.com/mozilla/kuma/pull/5914 for possible inspiration.

~~I dare to argue that using error boundaries to [send errors to Sentry](https://github.com/mozilla/kuma/issues/5833) first.~~

peterbe:
Fixes #5929

Truth be told, I don't know if this works. I'm not even sure how to test it. If I fake it by manually making sure [these lines](https://github.com/mozilla/kuma/blob/6df857e40c19011fe41b00347d3e7f3de51349cb/kuma/wiki/jinja2/wiki/react_document.html#L41-L46) really do render then `window.ga` *is* defined by when the call to it happens, I see no XHR requests in my browser. 
I don't know if it's because my fake UA is `12345`. Or perhaps there's a race condition. The way we set up GA is that sometime during the `window.onload` it injects a script tag with src `'https://www.google-analytics.com/{{ GA_FILE }}.js'`. What if, milliseconds before that, our React hydration code happens and comes across this `logError` method.

All in all, I'm just nervous about all of this working. Would love some extra review and perhaps a slow rollout.

### DIFF
--- a/kuma/javascript/src/error-boundaries.jsx
+++ b/kuma/javascript/src/error-boundaries.jsx
@@ -1,24 +1,31 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 
+import GAProvider from './ga-provider.jsx';
 import { gettext } from './l10n.js';
 
 export class AppErrorBoundary extends React.Component {
     state = { error: null };
 
+    static contextType = GAProvider.context;
+
     static getDerivedStateFromError(error) {
         return { error };
     }
 
-    logError() {
-        // https://github.com/mozilla/kuma/issues/5833
-        // When we start to use Sentry, change the method signature to
-        // logError(error, errorInfo) {
+    logError(boundaryName) {
+        // https://developers.google.com/analytics/devguides/collection/analyticsjs/events#event_fields
+        this.context('send', {
+            hitType: 'event',
+            eventCategory: 'errorboundary',
+            eventAction: boundaryName,
+            eventLabel: document.location.href
+        });
     }
 
-    componentDidCatch(error, errorInfo) {
+    componentDidCatch() {
         document.title = 'Application rendering Error';
-        this.logError(error, errorInfo);
+        this.logError('application');
     }
 
     render() {
@@ -39,9 +46,9 @@ AppErrorBoundary.propTypes = {
 };
 
 export class ContentErrorBoundary extends AppErrorBoundary {
-    componentDidCatch(error, errorInfo) {
+    componentDidCatch() {
         document.title = 'Content rendering error';
-        this.logError(error, errorInfo);
+        this.logError('content');
     }
     render() {
         if (this.state.error) {
@@ -65,7 +72,7 @@ function ErrorMessage({ title, children }) {
                         <h1 className="page-title">{title}</h1>
                         <p>
                             {gettext(
-                                'An unhandled error occurred in the application. The error has been logged and an administrator was notified. We apologize for the inconvenience!'
+                                'An unhandled error occurred in the application. We apologize for the inconvenience!'
                             )}
                         </p>
                         {children}
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -6,6 +6,7 @@ import SinglePageApp from './single-page-app.jsx';
 import LandingPage from './landing-page.jsx';
 import SignupFlow from './signup-flow.jsx';
 import { AppErrorBoundary } from './error-boundaries.jsx';
+import GAProvider from './ga-provider.jsx';
 import { localize } from './l10n.js';
 
 let container = document.getElementById('react-container');
@@ -67,10 +68,13 @@ if (container) {
        render any visible UI. It activates additional checks and
        warnings for its descendants.
        @see https://reactjs.org/docs/strict-mode.html */
-    app = <React.StrictMode>{app}</React.StrictMode>;
-
-    /* Need docstring*/
-    app = <AppErrorBoundary>{app}</AppErrorBoundary>;
+    app = (
+        <GAProvider>
+            <AppErrorBoundary>
+                <React.StrictMode>{app}</React.StrictMode>
+            </AppErrorBoundary>
+        </GAProvider>
+    );
 
     // NOTE: `document.all` is only available in lte IE10
     // $FlowFixMe
--- a/kuma/javascript/src/landing-page.jsx
+++ b/kuma/javascript/src/landing-page.jsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 
 import A11yNav from './a11y/a11y-nav.jsx';
 import Banners from './banners.jsx';
-import GAProvider from './ga-provider.jsx';
 import Header from './header/header.jsx';
 import UserProvider from './user-provider.jsx';
 
@@ -15,12 +14,10 @@ import UserProvider from './user-provider.jsx';
  */
 export default function LandingPage() {
     return (
-        <GAProvider>
-            <UserProvider>
-                <A11yNav />
-                <Header />
-                <Banners />
-            </UserProvider>
-        </GAProvider>
+        <UserProvider>
+            <A11yNav />
+            <Header />
+            <Banners />
+        </UserProvider>
     );
 }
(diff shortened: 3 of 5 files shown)


kuma/javascript/src/error-boundaries.jsx
```diff
@@ -1,24 +1,31 @@
...
-        // https://github.com/mozilla/kuma/issues/5833
-        // When we start to use Sentry, change the method signature to
-        // logError(error, errorInfo) {
+    logError(boundaryName) {
+        // https://developers.google.com/analytics/devguides/collection/analyticsjs/events#event_fields
+        this.context('send', {
+            hitType: 'event',
+            eventCategory: 'errorboundary',
+            eventAction: boundaryName,
+            eventLabel: document.location.href
```
escattone: Will `document` always be defined? I don't know how this React component gets access to the `document` object. Must be some React magic under the hood?

kuma/javascript/src/error-boundaries.jsx
```diff
@@ -1,24 +1,31 @@
...
         return { error };
     }
 
-    logError() {
-        // https://github.com/mozilla/kuma/issues/5833
-        // When we start to use Sentry, change the method signature to
-        // logError(error, errorInfo) {
+    logError(boundaryName) {
+        // https://developers.google.com/analytics/devguides/collection/analyticsjs/events#event_fields
+        this.context('send', {
```
escattone: This `this.context` confuses me as well. It must be that your `contextType` definition above tells React to call `useContext` on that object in order to return the value of `this.context`?

kuma/javascript/src/error-boundaries.jsx
```diff
@@ -1,24 +1,31 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 
+import GAProvider from './ga-provider.jsx';
 import { gettext } from './l10n.js';
 
 export class AppErrorBoundary extends React.Component {
     state = { error: null };
 
+    static contextType = GAProvider.context;
```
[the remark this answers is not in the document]
peterbe: That's for hooks. This component is a class. I don't think you can use hooks for error boundaries. Either that or their docs are out of date.

kuma/javascript/src/error-boundaries.jsx
peterbe: An error boundary can only ever happen in a browser. There, the `document` and `window` globals are always available. At least I hope so :)

kuma/javascript/src/error-boundaries.jsx
peterbe: All these years and I'm still unsure how to use contexts. Because I almost never like them and very rarely think they're needed. I *think* this is how you use it.

kuma/javascript/src/error-boundaries.jsx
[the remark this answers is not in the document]
schalkneethling: Could/should this be `const ga = useContext(GAProvider.context);`, perhaps this is just another way of accomplishing the same end result?

kuma/javascript/src/error-boundaries.jsx
schalkneethling: If my comment on line 10 above makes sense/is valid, then we can do the following here:

```
ga('send', {
  hitType: 'event',
  eventCategory: 'errorboundary',
  eventAction: boundaryName,
  eventLabel: document.location.href
});
```

kuma/javascript/src/error-boundaries.jsx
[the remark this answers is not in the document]
Gregoor: Yeah unfortunately there's no error boundary hooks (yet). And hooks can't be used from class components (hence no `useContext()`).
So the use of contextType makes sense to me here. Another way would be to wrap this component in a [context consumer](https://reactjs.org/docs/context.html#contextconsumer) and receive the context as a prop. But I think Peter's version is more concise and better than that here.

kuma/javascript/src/error-boundaries.jsx
Gregoor: Before the API I didn't like them either. I.e. I also think the `contextType` thing is not great and not very explicit. But simultaneously it's the right choice here (until we get error boundary hooks).

schalkneethling:
@peterbe @Gregoor 

I did some testing with the GA debugger extension and when I run the code as is here and changing `const locale = getLocale();` to `const locale = geLocale();` in `header.jsx` I do not see a GA call that reflects what is in `this.context(...`

What I do see is the following:

```
Running analytics_debug.js. This script is intended for testing and debugging only.
...
analytics_debug.js:16 Running command: ga("set", "anonymizeIp", true)
analytics_debug.js:16 Running command: ga("send", "pageview", {hitCallback: [function]})
analytics_debug.js:16 Storage not available. Aborting hit.
analytics_debug.js:16 Registered new plugin: ga(provide, "render", Function)
```

Now, if I add the following line to `logError`:

```
window.mdn.analytics.trackError('test', 'trackError');
```

I see the following:

```
Running analytics_debug.js. This script is intended for testing and debugging only.
...
analytics_debug.js:16 Running command: ga("set", "anonymizeIp", true)
analytics_debug.js:16 Running command: ga("send", "pageview", {hitCallback: [function]})
analytics_debug.js:16 Storage not available. Aborting hit.
analytics_debug.js:16 Running command: ga("send", {hitType: "event", eventCategory: "test", eventAction: "trackError", eventLabel: "", eventValue: 0, hitCallback: null})
analytics_debug.js:16 Storage not available. Aborting hit.
analytics_debug.js:16 Registered new plugin: ga(provide, "render", Function)
```

From this it seems that the `send` call made inside `this.context` is not reaching the Analytics code and the request is never sent. Is my test case a valid one?

schalkneethling:
I also see this warning in the console:

```
checkPropTypes.js:20 Warning: Failed prop type: The prop `children` is marked as required in `ErrorMessage`, but its value is `undefined`.
    in ErrorMessage (created by AppErrorBoundary)
    in AppErrorBoundary
```

schalkneethling:
Ah ok, I logged out `this.context` and I get `ƒ noop()`. Reading the code in `ga-provider.jsx` I see the following:

```
// If there is a window object that defines a ga() function, then
// that ga function is the value we will provide. Otherwise we just
// provide a dummy function that does nothing.
if (typeof window === 'object' && typeof window.ga === 'function') {
  ga = window.ga;
} else {
  ga = noop;
}
```

schalkneethling:
Logging out `window.ga` inside the `logError` function does indeed suggest that `window.ga` exists at this point in code execution:

Screenshot 2019-10-08 at 13 48 23

So, perhaps it has something to do with line 10 i.e. the point in execution when the `static` variable is set?

Gregoor:
> Or perhaps there's a race condition.

Shouldn't the script order in the HTML ensure that analytics is loaded before react? From my understanding analytics also makes sure to collect all the calls, to it before the actual script is loaded. I.e. there's code in our analytics script tag that basically says

```javascript
window.ga = window.ga || function() {
  (window.ga.q = window.ga.q || []).push(arguments);
};
```

Unfortunately I'm fighting local build problems again, so might be another hour until I have something more conrete to say about the things Schalk is discovering right now.

Gregoor:
An unrelated comment to what Schalk has unearthed here: do we maybe want to track the error message and the offending line as well? I guess it'd still need some source mapping to then figure out what exactly happened, but happy to look at that if we want that!

schalkneethling:
@peterbe @Gregoor It Works!

```
Initializing Google Analytics.
analytics_debug.js:16 Running command: ga("create", "UA-123456-1", "mozilla.org")
...

analytics_debug.js:16 Running command: ga("send", {hitType: "event", eventCategory: "errorboundary", eventAction: "application", eventLabel: "http://beta.mdn.localhost:8000/en-US/"})

....
analytics_debug.js:16 Storage not available. Aborting hit.
analytics_debug.js:16 Registered new plugin: ga(provide, "render", Function)
```

peterbe:
> do we maybe want to track the error message and the offending line as well?

Maybe. That's what Sentry is for. My bet is that 99% of errors will be cryptic as hell and near impossible to reproduce (e.g. [`script error`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#Notes)). And 99% of those you won't be able to reduce it down to a specific line of .js code. It might be worth experimenting with but let's not block this for that.

PR #5951, opened 2019-10-09
Fixes #5950, removes iex-test.js

### LINKED ISSUE #5950: Stop serving bug1522937-iex-test.js
As part of the `wiki` JavaScript bundle, we are still including 'js/utils/bug1522937-iex-test.js'. This is no longer needed so we can safely remove the inclusion inside `common.py` and delete the `.js` file.

TremaineNeethling:
removed reference from common.py and removed relevant .js files


TremaineNeethling:
@Gregoor could you review this please?

PR #5953, opened 2019-10-09
Performace Fixes #5672

### LINKED ISSUE #5672: JavaScript that can be deferred should be deferred
We have JS that loads like `<script src="/static/react-stuff.sj">` but we might benefit from adding `defer` and load them the `<head>` instead. 
E.g. https://flaviocopes.com/javascript-async-defer/

ankiiitraj:
added `defer` in `PIPELINE_JS` to `extra_context` and removed `async`.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -953,6 +953,7 @@ PIPELINE_JS = {
         'output_filename': 'build/js/banners.js',
         'extra_context': {
             'async': True,
+            'defer': True,
         },
     },
     'users': {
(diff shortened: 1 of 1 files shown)


ankiiitraj:
hi @peterbe. Can you please assign yourself as the reviewer.

ankiiitraj:
@peterbe I will be more careful now on. Thanks for the guidance.

peterbe:
I don't feel comfortable with this. The changes are too broad and too unplanned. This stuff matters and there's a reason why `async` was set on some of them and this PR just changes that. It'd make more sense to individually turn each `async` into `defer`, one at a time and then it can come with an explanation and justificiation. This PR is just a mass-edit. Sorry @ankiiitraj but perhaps you can find another open issue to contribute to that isn't so complicated.

PR #5956, opened 2019-10-09
Better footer article metadata display

### LINKED ISSUE #5954: Strange footer article metadata display
**Summary**
Screen Shot 2019-10-09 at 1 55 56 PM

**Steps To Reproduce (STR)**

1. Visit https://developer.mozilla.org/en-US/docs/Games/Anatomy
2.  Scroll down
3. Read about when it was last modified

**Actual behavior**
_What actually happened?_

The HTML looks like this:

```html
<section class="contributors-sub">
  <svg class="icon icon-clock" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 28" aria-hidden="true"><path d="...."></path></svg>

  <header><h4>Last modified:</h4></header>

  <time datetime="2019-03-23T19:24:26.976056">Mar 23, 2019</time>,
  <a href="https://wiki.developer.mozilla.org/en-US/docs/Games/Anatomy$history" rel="nofollow">by MDN contributors</a></section>
```

**Expected behavior**
The `<header>` should probably not be used there. Nor the use if `<h4>`.

**Additional context**

The new React site launched 5 min ago.

joedarc:
Fix #5954

Screen Shot 2019-10-09 at 5 59 43 PM

### DIFF
--- a/kuma/javascript/src/last-modified.jsx
+++ b/kuma/javascript/src/last-modified.jsx
@@ -30,7 +30,9 @@ export default function LastModified({
     return (
         <section className="contributors-sub">
             <ClockIcon />
-            <strong>{gettext('Last modified: ')}</strong>
+            <span className="contributors-sub-label">
+                {gettext('Last modified:')}
+            </span>
             <time dateTime={lastModified}>
                 {lastModifiedDate.toLocaleString(
                     documentLocale,
--- a/kuma/javascript/src/last-modified.jsx
+++ b/kuma/javascript/src/last-modified.jsx
@@ -30,9 +30,7 @@ export default function LastModified({
     return (
         <section className="contributors-sub">
             <ClockIcon />
-            <header>
-                <h4>{gettext('Last modified:')}</h4>
-            </header>
+            <strong>{gettext('Last modified: ')}</strong>
             <time dateTime={lastModified}>
                 {lastModifiedDate.toLocaleString(
                     documentLocale,
--- a/kuma/javascript/src/last-modified.jsx
+++ b/kuma/javascript/src/last-modified.jsx
@@ -30,9 +30,9 @@ export default function LastModified({
     return (
         <section className="contributors-sub">
             <ClockIcon />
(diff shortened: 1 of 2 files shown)


joedarc:
@escattone I ended up swapping to use a `<span>` instead and added a little styling. It looks pretty much the same as the above image. Let me know if I need to make any adjustments!

kuma/javascript/src/last-modified.jsx
```diff
@@ -30,9 +30,9 @@ export default function LastModified({
     return (
         <section className="contributors-sub">
             <ClockIcon />
-            <header>
-                <h4>{gettext('Last modified:')}</h4>
-            </header>
+            <span className="contributors-sub-label">
+                {gettext('Last modified:')}
+            </span>
```
schalkneethling: We should not change this HTML structure as it is specifically done for semantic and accessibility reasons. I will also add the CSS to resolve this problem in a separate comment.

kuma/static/styles/minimalist/structure/_article.scss
```diff
@@ -20,5 +20,10 @@ Main wiki article content
         margin-top: 32px;
         font-size: 0.88889rem;
         color: $grey;
+
+        .contributors-sub-label {
+            margin: 5px;
+            font-weight: bold;
+        }
```
schalkneethling: You can replace the above with the following:

```
.icon-clock {
    vertical-align: unset;
}

header {
    display: inline-block;
    padding: 4px 0;
}
```

joedarc: Would you be alright with the header padding just being `4px` so that it gives a little space between the elements?

Screen Shot 2019-10-10 at 9 47 12 AM

joedarc: With `padding: 4px;`

Screen Shot 2019-10-10 at 9 55 01 AM

joedarc:
Changes made, thanks for the feedback @schalkneethling!

Issue #5959, opened 2019-10-10
DOMException on some pages due to invalid selectors


peterbe:
E.g. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Basic_Concepts_of_Grid_Layout

Error:
```
DOMException: "'#Track_sizing_and_minmax() ~ pre[class*=html], #Track_sizing_and_minmax() ~ * pre[class*=html]' is not a valid selector"
```

peterbe:
Same error here: https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

peterbe:
The curious thing is that the error is "caused" by HTML like this:
```html
<h3 id="capitalize_(General)" name="capitalize_(General)"><code>capitalize</code> (General)</h3>
```
But if you go to https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/text-transform and, when logged in, and Shift refresh the new HTML becomes:
```html
<h3 id="capitalize_General" name="capitalize_(General)"><code>capitalize</code> (General)</h3>
```

which the React code should be able to cope with.

peterbe:
After having done that, https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform now works.

peterbe:
Some pages just need to be re-rendered (on the wiki) and then they work. 
Other pages don't work. Like this Korean page: https://developer.mozilla.org/ko/docs/Web/CSS/linear-gradient
I think it's failing because of the non-ascii characters in the DOM ID.

peterbe:
Oops! Didn't mean to press the close button.

schalkneethling:
> Some pages just need to be re-rendered (on the wiki) and then they work.

Yup, I have seen that as well. Wonder if we can kick-off a re-render of all the pages. Should solve the majority of the errors no?

peterbe:
> Wonder if we can kick-off a re-render of all the pages. Should solve the majority of the errors no?

I think we should do that. But it won't solve all pages. E.g. [this Korean page](https://developer.mozilla.org/ko/docs/Web/CSS/linear-gradient)

peterbe:
Another page where it doesn't help to re-render: https://developer.mozilla.org/zh-CN/docs/Web/CSS/linear-gradient

peterbe:
[...]

PR #5960, opened 2019-10-10
DOMException on some pages due to invalid selectors

peterbe:
Part of #5959

This is a band-aid solution to get certain pages to not throw. Some pages have stuff like:
```html
<h3 id="45도_기울어진_그레이디언트">45도 기울어진 그레이디언트</h3>
```

The flaw here isn't the non-ascii as an ID but the fact that it starts with a number which makes it an **invalid** ID. 

What this PR does it swallows those exceptions and just moves on and thus you don't get the "Open in Codepen" buttons. 

Screen Shot 2019-10-10 at 11 07 55 AM

### DIFF
--- a/kuma/javascript/src/live-examples.js
+++ b/kuma/javascript/src/live-examples.js
@@ -23,7 +23,6 @@ export function addLiveExampleButtons(rootElement) {
         // We expect the iframe to be in a section with an id related
         // to the frame id.
         let sectid = frame.id.replace(idPrefix, '');
-
         // It *used* to be that the ' character was allowed as a safe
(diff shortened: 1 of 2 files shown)


peterbe:
I've figured out a more complete solution. Hang on.

peterbe:
So, in this PR there's a half-way solution. There's Python code that makes sure that the `h3` ID doesn't start with a number. That unbreaks the "Open in Codepen" stuff but it now breaks the iframe.

peterbe:
This PR tries too many different things in the same problem space. Going to break it up.

PR #5978, opened 2019-10-14
5949 expose the raw kuma html in the document json api

peterbe:
@escattone this is NOT urgent. But it's important as the basis of a discussion. Also, I wouldn't mind if we land this in production so I can continue my work with testing this "migration strategy".

This PR is part of a project that spans Kuma, stumptown-content, and stumptown-renderer. 
With this change, I was able to run `render_document --all` and the that populates an S3 bucket. 
Now, because every file gets a unique name (no more `Web` the file and `Web` the "folder" because the file becomes `Web.json`), I can run: `s3cmd sync s3://peterbe-mdn-documents/ kumadocs/ --recursive` and be able to run a Node script that walks this `kumadocs/` directory without problems. 

Part of the [Stumptown Render Kuma AND Stumptown Semi-Dynamically](https://docs.google.com/document/d/1yNJCNJ2iv4H7f8E0OcJlLxDDrP90Pv6os85yMAnX0mI/edit?usp=sharing) project.

### DIFF
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -119,8 +119,6 @@ def document_api_data(doc=None, redirect_url=None):
     """
     Returns the JSON data for the document for the document API.
     """
-    print("DOC:", doc)
-    print("redirect_url", redirect_url)
     if redirect_url:
         return {
             'documentData': None,
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -61,7 +61,7 @@ def get_s3_key(doc=None, locale=None, slug=None,
                prefix_with_forward_slash=False):
     if doc:
         locale, slug = doc.locale, doc.slug
-    key = reverse('api.v1.doc', args=(locale, slug))
+    key = reverse('api.v1.doc', args=(locale, slug)) + '.json'
     if prefix_with_forward_slash:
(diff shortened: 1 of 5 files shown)


peterbe:
@escattone Good to good me thinks!

kuma/api/v1/views.py
```diff
@@ -109,6 +119,8 @@ def document_api_data(doc=None, redirect_url=None):
     """
     Returns the JSON data for the document for the document API.
     """
+    print("DOC:", doc)
+    print("redirect_url", redirect_url)
```
escattone: Forgot to remove these?

peterbe:
When this goes into production, can we trigger a mass publish? (doesn't have to happen immediately) 
But when we do, can we delete all existing keys first otherwise we're going to have one called ".../Web" and one called ".../Web.json" which'll still cause trouble when using the aws cli to download this bucket to disk.

escattone:
> When this goes into production, can we trigger a mass publish? (doesn't have to happen immediately)
> But when we do, can we delete all existing keys first otherwise we're going to have one called ".../Web" and one called ".../Web.json" which'll still cause trouble when using the aws cli to download this bucket to disk.

Definitely, I can do both of those things after we deploy this.

PR #5989, opened 2019-10-14
live-examples trusts section IDs too much

### LINKED ISSUE #5959: DOMException on some pages due to invalid selectors

peterbe:
Fixes #5959

Screen Shot 2019-10-14 at 3 56 10 PM

Note that there are *no* "Open in Codepen" buttons. At *first*, I did this:

```javascript
// in article.jsx
addLiveExampleButtons(rootElement);

// in live-examples.js
for (let frame of iframes) {
  ...
  try {
      let html = rootElement.querySelector(
            `#${sectid} ~ pre[class*=html], #${sectid} ~ * pre[class*=html]`
        );
   } catch(error) {
     ...
   }
```

But then I realized that in that context, we don't have the GAProvider context because `addLiveExampleButtons` isn't a React component. It's just a function. 
So then I changed it to:

```javascript
// in article.jsx
try { 
    addLiveExampleButtons(rootElement);
} catch(error) {
    console.error(error);
   ga('send', ...)
}

// in live-examples.js
for (let frame of iframes) {
  ...
  let html = rootElement.querySelector(
        `#${sectid} ~ pre[class*=html], #${sectid} ~ * pre[class*=html]`
    );
```

Yes, it means that it bails on the first `iframe` failing in the loop. But this feels still cleaner and it's also a bit more defensive in that it captures anything that can go wrong anywhere inside `live-examples.js`.

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -35,11 +35,21 @@ export default function Article({ document }: DocumentProps) {
             InteractiveExamples.setLayout(rootElement);
             // Keep addLiveExampleButtons() before addAnchors() so the
             // example title doesn't end up with a link in it on codepen.
-            addLiveExampleButtons(rootElement);
+            try {
+                addLiveExampleButtons(rootElement);
+            } catch (error) {
+                console.error(error);
+                ga('send', {
+                    hitType: 'event',
+                    eventCategory: 'article-effect-error',
+                    eventAction: 'addLiveExampleButtons',
+                    eventLabel: error.toString()
+                });
+            }
             highlightSyntax(rootElement);
             activateBCDTables(rootElement);
         }
-    }, [document]);
+    }, [document, ga]);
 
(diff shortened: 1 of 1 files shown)


peterbe:
So once this lands, the next thing to do is to perhaps do something similar to the other functions:
https://github.com/mozilla/kuma/blob/b670667355a647cc76fd279eae78ce0538e8451f/kuma/javascript/src/article.jsx#L35-L47

Basically, instead of the draconian content error boundaries, we wrap each and every function call from a `useEffect` and start hawkishly monitor GA for these `article-effect-error` categories.

PR #5996, opened 2019-10-16
inject js polyfills if needed

### LINKED ISSUE #5993: Application Rendering Error: edge 20
**Summary**
Getting 
Application Rendering Error
An unhandled error occured in the application. We aplogize for the inconvenience!
<button> reload page
In Edge 20 (trying to update to 44) on Windows 10.

**Steps To Reproduce (STR)**
Visit any page on developer.mozilla.org with Edge 20 on windows 10

**Actual behavior**
Home page loaded with error on top.
Clicking on "web technologies" I go to developer.mozilla.org/en-US/docs/Web/ but the page is blank other than the error.

**Expected behavior**
The page to load without error

**Additional context**
Filing this bug from my mac, new to the pc, so no screen shot know-how. yet. And not logged into anything on that device yet. sorry no screen shot attached.

### LINKED ISSUE #5995: Site should work even if you don't have window.fetch
We're getting a lot of error boundaries happen because people's browser don't have `window.fetch`.

peterbe:
Fixes #5995
Fixes #5993

I really think this is the smoothest solution to the problem of legacy browsers. 
Let's not try to solve this with changing the `babel.config.js` so that it transpiles to much older browsers. 

For my testing, I used Firefox 38esr since it was the last browser before `fetch` was introduced. 
I downloaded it from https://ftp.mozilla.org/pub/firefox/releases/38.8.0esr/mac/en-US/
Then I found it on the command line and started it like this:

```bash
cd /Applications
# First time
./Firefox\ 2.app/Contents/MacOS/firefox --new-instance --ProfileManager
# Any other time
./Firefox\ 2.app/Contents/MacOS/firefox
```
(I called my `Firefox 2.app` because there already as a release version of Firefox in my `/Applications` folder)

CC @joedarc


jinja2/includes/polyfill_scripts.html
```diff
@@ -0,0 +1,14 @@
+<script>
+  window.fetch ||
+    document.write(
+      '<script src="https://unpkg.com/unfetch/polyfill"><\/script>'
+    );
+</script>
+<script>
+  // Only include the fat polyfill for browsers that seem to not have
+  // certain JS features. E.g. Firefox 38.
+  Array.prototype.includes ||
```
peterbe: I think I picked this test because it was the first error I got. We just need *something* that gives us a decent indication that the browser is old. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

jinja2/includes/polyfill_scripts.html
```diff
@@ -0,0 +1,14 @@
+<script>
+  window.fetch ||
+    document.write(
+      '<script src="https://unpkg.com/unfetch/polyfill"><\/script>'
```
peterbe: [`unfetch`](https://github.com/developit/unfetch) is a great `fetch` polyfill. It's the smallest one I know and has sufficient features. It's maintained and developed by Jason Miller @developit who started the Preact project amongst many other great libraries.

peterbe:
We are not done with the old browsers but as of these hacks, it no longer keeps falling into the error boundaries. 

The header still isn't working, which leads me to the next point...

Another thing to work on is the CSS for old browsers. I don't know how you'd do feature detection but perhaps it can be assumed that if the browser doesn't have `fetch` it probably doesn't support our stylesheet. So perhaps we can do something like this:

```html
window.fetch ||
    (document.write('{% stylesheet "legacy_css" %}"/>')
    && document.write(
      '<script src="https://unpkg.com/unfetch/polyfill"><\/script>'
    ))
```

Can we fix the header with CSS for legacy browsers? Or do we need something like?:

```html
window.fetch ||
    (document.write('{% stylesheet "legacy_css" %}"/>')
    && document.write(
      '<script src="https://unpkg.com/unfetch/polyfill"><\/script>'
    )
    && document.write(
      '<script src="/static/legacy_header.js"><\/script>'
    ))
```

schalkneethling:
Thanks for the PR @peterbe. As mentioned on Slack there are a couple of problems here.

The first thing is that we should probably [not rely on unpkg for production](https://medium.com/@david.gilbertson/unpkg-is-great-but-youd-be-mad-to-use-if-for-a-production-app-5bde7f3fc4f4) usage, although that might have changed.

If we do decide to go this route, which again I believe we need to reconsider, something like https://polyfill.io/v3/ might be a better solution?

Either way, we will be adding third party resources into our code so we would need to request a rapid risk analysis(RRA) before we are able to deploy this to production. We also had to do this previously when we were considering adding the Codepen API.

peterbe:
RRA sounds like a lot of red tape blocking progress. Also, I appreciate it because why should we trust those anyway. 
How about we simply vendor these in? I.e. download them, and serve them kinda like we serve jquery. It would also mean slightly better perf for those poor suckers whose browsers cause these to be downloaded since they'd need to do 2 more DNS and SSL lookups. 

Shall I do that?

schalkneethling:
> download them, and serve them kinda like we serve jquery

Yup, if we serve them then we do not need the RRA. :+1:

peterbe:
@schalkneethling I actually don't know how unpkg and cdn.polyfill.io works but I learned the hard way that you can't just download these with `curl` or `wget` because I think they do some dynamic serving based on your user-agent or something. So I fired up my old Firefox 30 and typed in `https://cdn.polyfill.io/v3/polyfill.min.js` and `https://unpkg.com/unfetch/polyfill`, as it got served to Firefox 38, and then I copied that JS into files on disk and checked them in.

Bisrat-Aregawi:
[https://www.youtube.com/watch?v=pweGksqjIZo](url) following this have worked for me.

PR #5998, opened 2019-10-16
Support serving files from S3 instead of EFS

callahad:
This is the backend work necessary for #5798.

We'll still have a bit of operations work to do after this is merged to actually close out the issue.

This isn't a huge diff, but each commit is self-contained and it's likely easier to review than working through the whole thing as a single unit.

Before this is ready:
- [x] Preliminary review
- [x] Get prod / stage infrastructure set up (mdn/infra#336)
- [x] ~~Write docs~~ (No action needed by developers, so not blocking on docs)
- [x] ~~Figure out what needs to happen to get new `kuma_base` images published to the Docker Hub~~ (happens automatically)
- [x] ~~Investigate cleaning up temporary attachments created during runs of the test suite
    (we don't do this currently for either storage backend)~~

After this is merged and deployed:
- [ ] File bug to clean up old EFS volumes
- [ ] File bug to follow up on `get_file_url()`
- [ ] Remove `USE_S3` configuration switch
- [ ] Remove filesystem serving code from `raw_file` view
- [ ] File wishlist bugs for decommissioning the old attachments CDN (which would probably require pre-resolving attachment redirects to point directly at the new CDN.)

Setup notes:
- You'll probably have to build your own kuma_base image. Instructions in [docker/images/kuma_base/README.rst](https://github.com/mozilla/kuma/tree/master/docker/images/kuma_base)
- You can migrate existing attachments by ~~copying the `media/attachments` folder itself into `media/s3/mdn-attachments/`~~ running this command from the root of the `kuma` checkout:

        docker-compose run --rm -v "$(pwd)/media/attachments:/mnt/attachments" \
          --entrypoint=/bin/sh minio -c "mkdir -p /data/mdn-attachments && cp -r /mnt/attachments /data/mdn-attachments/"
- ~~Minio doesn't support AWS bucket policies, so you have to make attachments publicly readable manually:~~
    1. ~~Visit http://localhost:9000~~
    2. ~~Create the `mdn-attachments` bucket if not already present~~
    3. ~~Click the "..." menu next to the bucket name, choose "Edit Policy"~~
    4. ~~Click "Add" to create a public Read-Only policy with no prefix (or a `*` prefix; it's the same)~~


callahad:
Pytest passes, but Travis currently fails when we run:

```sh
# This checks that running `./manage.py makemigrations` wasn't forgotten
docker-compose exec -T web ./manage.py makemigrations --check --dry-run
```

Because we're changing the Model's storage, `makemigrations` wants to generate the following migration. Which seems fragile / nonsensical / dangerous to me. 

<details><summary>Generated migration</summary>

```python
# -*- coding: utf-8 -*-
# Generated by Django 1.11.23 on 2019-10-16 07:16
from __future__ import unicode_literals

from django.db import migrations, models
import kuma.attachments.utils
import storages.backends.s3boto3


class Migration(migrations.Migration):

    dependencies = [
        ('attachments', '0001_squashed_0008_attachment_on_delete'),
    ]

    operations = [
        migrations.AlterField(
            model_name='attachmentrevision',
            name='file',
            field=models.FileField(max_length=500, storage=storages.backends.s3boto3.S3Boto3Storage(access_key=b'minio', auto_create_bucket=True, bucket_acl=b'public-read', bucket_name=b'mdn-attachments', custom_domain=b'localhost:9000/mdn-attachments', default_acl=b'public-read', endpoint_url=b'http://minio:9000', object_parameters={b'CacheControl': b'public, max-age=31536000, immutable'}, querystring_auth=False, region_name=b'us-east-1', secret_key=b'minio123', secure_urls=False), upload_to=kuma.attachments.utils.attachment_upload_to),
        ),
        migrations.AlterField(
            model_name='trashedattachment',
            name='file',
            field=models.FileField(help_text='The attachment file that was trashed', max_length=500, storage=storages.backends.s3boto3.S3Boto3Storage(access_key=b'minio', auto_create_bucket=True, bucket_acl=b'public-read', bucket_name=b'mdn-attachments', custom_domain=b'localhost:9000/mdn-attachments', default_acl=b'public-read', endpoint_url=b'http://minio:9000', object_parameters={b'CacheControl': b'public, max-age=31536000, immutable'}, querystring_auth=False, region_name=b'us-east-1', secret_key=b'minio123', secure_urls=False), upload_to=kuma.attachments.utils.attachment_upload_to),
        ),
    ]
```
</details>

Suggestions for a way forward?

It's been nearly a decade since I last really wrestled with Django migrations 😬

callahad:
Right, the SQL generated for this migration is literally a no-op:

```
$ ./manage.py sqlmigrate attachments 0002_auto_20191016_0739
BEGIN;
--
-- Alter field file on attachmentrevision
--
--
-- Alter field file on trashedattachment
--
COMMIT;
```

Committing this migration as generated is almost certainly the wrong thing.

- The actual `storage` value is configuration-dependent
- It exposes secrets and other configuration variables (esp. the AWS `secret_key`)

Which I guess isn't a problem in the CI environment -- it's all test data anyway -- but if you alter any of those values locally and run `makemigrations` you'll get another useless migration with a side of potentially leaking sensitive information.

kuma/attachments/models.py
```diff
@@ -2,16 +2,39 @@
...
+        bucket_acl='public-read',  # Only used if bucket is auto-created
+        default_acl='public-read',
+        querystring_auth=False,
+        custom_domain=settings.ATTACHMENTS_AWS_S3_CUSTOM_DOMAIN,
+        secure_urls=settings.ATTACHMENTS_AWS_S3_SECURE_URLS,
+        region_name=settings.ATTACHMENTS_AWS_S3_REGION_NAME,
+        endpoint_url=settings.ATTACHMENTS_AWS_S3_ENDPOINT_URL,
+    )
+else:
+    storage = default_storage
```
peterbe: You don't need to do this. You can leave it to `None` since what you've got here is already the default (no pun intended) behavior inside `FileField`.

callahad:
...going and re-reading the Migrations docs. Clearly a lot has changed since I cut my teeth on South.

kuma/attachments/models.py
```diff
@@ -49,6 +72,9 @@ def __str__(self):
         return self.title
 
     def get_file_url(self):
+        if self.current_revision is None:
+            return ""
```
peterbe: I know you didn't write the `on_delete=models.SET_NULL` on line 52 but doesn't that seem very odd?
How can a file attachment not belong to a revision? Won't it be unfreed spaces that's floating without an owner? I.e. bad in Rust. 

Arguably, returning an empty string won't cause an exception but I'd favor something like:
```python
if not self.current_revision:
    raise ValueError("Attachments without a revision don't have a file URL")
```

This won't necessarily need a unit test but whatever code that calls and depends on `get_file_url` should be aware of this and deal with it accordingly. I think that'd be a lot crisper than dealing with the surprising effect of a URL that is an empty string. Because an empty string definitely isn't a "URL".

kuma/attachments/models.py
```diff
@@ -2,16 +2,39 @@
...
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 from django_mysql.models import Model as MySQLModel
+from storages.backends.s3boto3 import S3Boto3Storage
 
 from .utils import attachment_upload_to, full_attachment_url
 
 
+if settings.ATTACHMENTS_USE_S3:
+    storage = S3Boto3Storage(
```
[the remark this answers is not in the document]
peterbe: > We don't want that. Attachments are a special case; they should specify their own storage system.

Mind you, `models.FileField` is only used by `kuma/attachments/models.py`. 

Suppose that we someday add something elsewhere we have files coming in. Wouldn't we want that on S3 too? We definitely don't want regular fs disk in this era of clouds. Yes, perhaps it *could* be a thing to have some stuff uploaded to GCS and some to S3 but it's unlikely. 

I'm too lazy to read up but surely there's a way with `django-storages` to control the destinations for different ORM things.

All in all, having migrations that depend on overrideable config seems a bigger negative than to have a broad default storage solution for all models. What do you think?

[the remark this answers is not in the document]
joedarc: Looks like one way to do this is to create a subclass of S3BotoStorage and override the bucket name. Something like: 

```
class MyS3Storage(S3BotoStorage):
        kwargs['bucket'] = getattr(settings, 'MY_AWS_STORAGE_BUCKET_NAME')
        super(MyS3Storage, self).__init__(*args, **kwargs)
```

Then you can set `DEFAULT_FILE_STORAGE` to that subclass.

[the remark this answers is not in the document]
joedarc: > Attachments are a special case; they should specify their own storage system.

Implementing how I mentioned I suppose you could have a subclass of `AttachmentS3Storage` that overrides what is necessary to specify it's own storage without having to grab the bucket name from the ENV?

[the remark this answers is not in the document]
callahad: > Mind you, `models.FileField` is only used by `kuma/attachments/models.py`.

We use Boto and access S3 elsewhere in Kuma, unrelated to attachments, and we do so using global environment variables. Attachments needs different settings from those.

> Suppose that we someday add something elsewhere we have files coming in. Wouldn't we want that on S3 too?

We almost certainly don't want it to be in the exact same bucket, with the exact same object properties, and the exact same CDN configuration as for attachments. So we're going to have to special case something somewhere.

> I'm too lazy to read up but surely there's a way with `django-storages` to control the destinations for different ORM things.

If you're going to push this point, please do read the docs or present a working alternative. Django's defined extension point for file operations is passing differently configured Storage instances to specific models, or changing the global default.

kuma/attachments/models.py
callahad: So you'd prefer this?

```diff
--- a/kuma/attachments/models.py
+++ b/kuma/attachments/models.py
@@ -2,7 +2,6 @@ import os
 from datetime import datetime
 
 from django.conf import settings
-from django.core.files.storage import default_storage
 from django.db import models
 from django.db.utils import IntegrityError
 from django.utils.encoding import python_2_unicode_compatible
@@ -13,7 +12,7 @@ from storages.backends.s3boto3 import S3Boto3Storage
 
 from .utils import attachment_upload_to, full_attachment_url
 
-
+storage = None
 if settings.ATTACHMENTS_USE_S3:
     storage = S3Boto3Storage(
         access_key=settings.ATTACHMENTS_AWS_ACCESS_KEY_ID,
@@ -31,8 +30,6 @@ if settings.ATTACHMENTS_USE_S3:
         region_name=settings.ATTACHMENTS_AWS_S3_REGION_NAME,
         endpoint_url=settings.ATTACHMENTS_AWS_S3_ENDPOINT_URL,
     )
-else:
-    storage = default_storage
 
 
 @python_2_unicode_compatible
```

callahad: FWIW, the FileField constructor looks like this:

```python
    def __init__(self, verbose_name=None, name=None, upload_to='', storage=None, **kwargs):
        self._primary_key_set_explicitly = 'primary_key' in kwargs

        self.storage = storage or default_storage
        self.upload_to = upload_to

        kwargs['max_length'] = kwargs.get('max_length', 100)
        super(FileField, self).__init__(verbose_name, name, **kwargs)
```

So falsey values get resolved to `default_storage`

kuma/attachments/models.py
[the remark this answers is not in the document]
callahad: @joedarc's idea of creating a S3Boto3Storage subclass and building the configuration in its constructor sounds promising. I'll give that a shot and see if it improves the migration.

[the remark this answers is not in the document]
peterbe: What I meant was that there is no other use of `models.FileField` in any of the other apps. The only django app that have that uses file storage is [the `attachment` one](https://github.com/mozilla/kuma/tree/af446fe757a1fe87f80e452179ab734b33d0fb06/kuma/attachments).

Also, `ripgrep` only finds:

```
▶ rg FileField
kuma/attachments/models.py
106:    file = models.FileField(upload_to=attachment_upload_to, max_length=500)
216:    file = models.FileField(

kuma/attachments/migrations/0001_squashed_0008_attachment_on_delete.py
34:                ('file', models.FileField(max_length=500, upload_to=kuma.attachments.utils.attachment_upload_to)),
80:                ('file', models.FileField(help_text='The attachment file that was trashed', max_length=500, upload_to=kuma.attachments.utils.attachment_upload_to)),
```

Actually, it appears that `AWS_STORAGE_BUCKET_NAME` rules and there's no easy way to have a different bucket for a different `FileField`. But perhaps it's moot because I'm guessing `django-storages` gives the keys sane name. 

So, to avoid the whole migration problem, if we just set
```python
DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
```
in `settings/common.py` 
the only thing that needs to be done is to set the various [AWS S3 related settings](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html). 
And if any of that clashes with how we use AWS and S3 for other stuff, you'd change it to something like

```python
DEFAULT_FILE_STORAGE = 'kuma.core.storages.KumaFileStorage'
```
and `kuma/core.storages.py` something like this:
```python
from storages.backends.s3boto3 import S3Boto3Storage


class KumaFileStorage(S3Boto3Storage):
    # something like @joedarc suggested
```

That should avoid the migrations problem. It should also give you full control over the environment variables. 

Also, perhaps by doing that we'd gain some other capabilities to better control the uploads and stuff depending on prod or local dev in docker-compose.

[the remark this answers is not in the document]
callahad: @peterbe How do you feel about https://github.com/mozilla/kuma/pull/5998/commits/3668f2c00d807999c14c309ef7c097da93b5a6b2? (@joedarc's suggestion was great!)

And do you still have a strong preference for `DEFAULT_FILE_STORAGE`? It would avoid a migration altogether, but at the risk of implicitly throwing future FileField/ImageField/etc's into the same bucket as attachments.

I prefer to keep it totally explicit to avoid that risk, but I could be convinced.

callahad:
> I'll see if there's an easy way to automate setting the policy

So we've got some options.

Looks like MinIO's canned Read-Only policy is equivalent to:

```javascript
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": ["s3:GetBucketLocation", "s3:ListBucket"],
            "Effect": "Allow",
            "Principal": {"AWS": ["*"]},
            "Resource": ["arn:aws:s3:::mdn-attachments"]
        },
        {
            "Action": ["s3:GetObject"],
            "Effect": "Allow",
            "Principal": {"AWS": ["*"]},
            "Resource": ["arn:aws:s3:::mdn-attachments/*"]
        }
    ]
}
```

We technically don't need that first statement (it allows anonymous users to enumerate the bucket contents), but it's what you get with `mc policy set download kuma/mdn-attachments`. We should stick with it as that canned policy is recognized as such in the MinIO GUI, while the more minimal policy is not. Plus, listing the bucket's contents can be useful during development.

We can set the policy using `boto3`:

```python
# Assume read_only_policy is set to the JSON string above
s3client = boto3.client('s3', aws_access_key_id='minio', aws_secret_access_key='minio123', endpoint_url='http://localhost:9000')
s3client.put_bucket_policy(Bucket='mdn-attachments', Policy=read_only_policy)

# Alternatively, using boto3's higher-level Resource API
s3 = boto3.resource('s3', aws_client_access_key='minio', aws_secret_access_key='minio123', endpoint='http://localhost:9000')
bucket_policy = s3.BucketPolicy('mdn-attachments')
bucket_policy.put(Policy=read_only_policy)
```

Or we could grab the MinIO client and run it:

    env MC_HOSTS_kuma=http://minio:minio123@localhost:9000 mc policy set download kuma/mdn-attachments

Or we could do that with Docker, but it gets a little bit gnarly since a standalone invocation of `docker run` won't necessarily be able to see things inside the network that `docker-compose` creates for Kuma. We'd have to do something like:

    docker run --rm --net kuma_default -e MC_HOST_kuma=http://minio:minio123@minio:9000 minio/mc policy set download kuma/mdn-attachments

...which assumes that your checkout of the repo is inside a folder literally named 'kuma' (resulting in the `kuma_default` network name).

peterbe:
@callahad and I chatted about this over the phone and concluded...

* I'm fine with NOT using that overly broad solution of `DEFAULT_FILE_STORAGE` if this new trick that Dan and Joe figured out can set explicit `storage` attributes without causing new weird migrations if you change the config later. 

* Dan is going to see if he can configure `minio` in such a way that the existing `media/attachments` folder that devs have can be used as the location where `minio` will use. And does that mean 0 manual efforts.

* I suggested that having the bucket creation codifed somewhere in Django is useful. Not only does it alliviate a manual instruction. By putting it into a checks.py files (far away from models.py) it can serve two purposes:
  1. Just automatically create the buckets for local development
  2. When used in production it properly sanity checks that the bucket truly matches what's been configured. I.e. "Then env var says the bucket name is Foo, but does it *really* exist and do I have access to read it?"

* To speed up the framework of Django system checks in Kuma I'm going to ask Dan to review so we can get our own framework rolling for making system checks of our own.

kuma/attachments/models.py
```diff
@@ -8,10 +8,37 @@
...
 from django.utils.translation import ugettext_lazy as _
 from django_mysql.models import Model as MySQLModel
+from storages.backends.s3boto3 import S3Boto3Storage
 
 from .utils import attachment_upload_to, full_attachment_url
 
 
+class AttachmentStorage(S3Boto3Storage):
+    def __init__(self, *args, **kwargs):
+        configuration = dict(
```
Gregoor: Just for my understanding: Why are we subclassing S3Boto3Storage instead of passing the configuration in when instantiating?

callahad: You can find the whole details in the comments marked "resolved" above, but TL;DR is that whatever you pass directly to the FileField gets included as part of the model migration, which seemed super awful. E.g., it ends up embedding the AWS credentials in the migration.

Subclassing avoids that, since those settings are hidden inside the constructor.

Gregoor: Ah oh, I see. Thanks for the context, makes sense!

escattone:
@callahad I see you're way ahead of me on the infrastructure side with https://github.com/mdn/infra/issues/336 😄

callahad:

callahad:
Closing and re-opening to force Travis to re-run tests against the current `master`, which uses Python 3.

callahad:
Hm... doesn't seem to be rebuilding on Travis. But that's fine, since there are more commits I need to add anyway :)

callahad:
AFAICT, the remaining issues are:

1. Migrating local content for developers. This is unavoidable, but there are very few Kuma devs, and I suspect most of us don't have any local attachments to migrate anyway.
2. Configuring Minio for local development (creating the bucket and setting permissions). I've addressed this by adding an ephemeral container to `docker-compose.yml`, and I'm rather happy with that solution.

I'll clean up the commit history and re-push.

callahad:
**I believe this is ready for final review and merge.**

Using S3 is controlled by `ATTACHMENTS_USE_S3`, which is off by default in production, but on by default in local development environments. I'll coordinate switching it on in production after this code is merged and deployed.

Also, if you have pre-existing local attachments you'd like to import into the new, local Minio instance, you'll need to run the following snippet from the root of your `kuma` repo:

    docker-compose run --rm -v "$(pwd)/media/attachments:/mnt/attachments" \
        --entrypoint=/bin/sh minio -c "mkdir -p /data/mdn-attachments && cp -r /mnt/attachments /data/mdn-attachments/"

This is such a rare situation that I'm hoping this comment + the commit message + a mention in Slack and an email to the team mailing list will be sufficient documentation.

At one point @peterbe suggested "having the bucket creation codifed somewhere in Django", specifically:
>  1. Just automatically create the buckets for local development
>  2. When used in production it properly sanity checks that the bucket truly matches what's been configured. I.e. "The env var says the bucket name is Foo, but does it *really* exist and do I have access to read it?"

Upon further reflection, I don't think application code is the right place to address those concerns.

I've handled the first point by adding an ephemeral `minio_setup` service that runs with `docker-compose up` and ensures that local buckets are created and properly configured.

For the second, we should absolutely make sure that AWS configuration has been passed in, but actually accessing S3 over the network to validate the presence and functioning of S3 feels a bit out of scope for the Django application. Or at least, something that we can add later, if we decide we need it.

kuma/attachments/models.py
callahad: Will file a separate bug to track resolving this. The need to file that bug is captured in the post-merge checklist at the top of this issue.

Gregoor:
It looks ready to me as well 🚀 

I agree that we don't need to codify the bucket creation for the reasons you outlined. What I do wonder is whether we might want to inform people who are running a local kuma. For example, we could check whether they have local attachments and print a warning and the import command you posted above.
But imo that's more of a nice to have, you've got my thumb-up for bringing this in as is :+1:

Thanks for the great PR, I really liked reading through your commit messages, I think I'm inspired to put more words into mine now as well.

kuma/attachments/views.py
```diff
@@ -38,18 +40,33 @@ def raw_file(request, attachment_id, filename):
...
-        response = StreamingHttpResponse(rev.file, content_type=rev.mime_type)
-        response['Content-Length'] = rev.file.size
+    # Attachments must be served from safe (untrusted) domains
+    if not is_untrusted(request):
+        return redirect(attachment.get_file_url(), permanent=True)
+
+    # NOTE: All of this, just to support conditional requests (last-modified / if-modified-since)
+    # Very important while we're potentially serving attachments from disk.
+    # Far less important when we're just redirecting to S3.
+    # Consider removing?
```
peterbe: I'd much rather simplify it and "outsource" as much as possible to S3. Your code is good but theirs is *really* good. 
So something like this:

```python
if settings.ATTACHMENTS_USE_S3:
    return redirect(rev.file.url)

if_modified_since = parse_http_date_safe(request.META.get('HTTP_IF_MODIFIED_SINCE'))
if if_modified_since and if_modified_since >= calendar.timegm(rev.created.utctimetuple()):
    response = HttpResponseNotModified()
else:
    response = StreamingHttpResponse(rev.file, content_type=rev.mime_type)
    response['Content-Length'] = rev.file.size
response['Last-Modified'] = convert_to_http_date(rev.created)
response['X-Frame-Options'] = f'ALLOW-FROM {settings.DOMAIN}'
return response
```

It's your call but I get a bit nervous about messing with the headers of a file that's going to be redirected anyway.

callahad: This will need further revision once we're on S3, but here's my current thinking:

The attachment URLs we embed in documents are keyed to an `Attachment` ID, rather than a more specific `AttachmentRevision`.

That means that users need to make two requests for every attachment on a page:

1. Discover the current redirect
2. Fetch the actual attachment

The exchange ends up looking like this:

redirect

Because the CDN URL is unique for each `AttachmentRevision`, we can instruct browsers to cache them immutably. Users with a warm cache can then completely skip the second request.

The problem is the first request, used to discover the CDN URL. Is it worth caching that? How and for how long?

(I think it is: if we don't, then we have to handle resolving every attachment. And that could add significant latency for a user with an otherwise warm cache.)

Right now I believe it will just be cached for however long our default TTL is in CloudFront, but we should probably be more explicit about the caching headers once we're using S3 in production.

peterbe: I think my suggestion was to get out of the business of `last-modified-if` stuff if it is a redirect. 
The caching of the 302 will be cached by CloudFront too. 
```
curl -v https://developer.m.o/files/123/fw.png
< HTTP/2 302
< x-cache: Hit from cloudfront
< cache-control: max-age=900, public
...
```
(the 900 comes from https://github.com/mdn/kuma/blob/2d8ead0cff04e395366637aa7aaebf277d797e75/kuma/attachments/views.py#L36)
which means that even the redirect could be cached for 15 min. So for 99% of the time, in a given 15min period, our Django doesn't even need to bother with any of this stuff since CloudFront will take care of that too. 

I think we should just increase the number `15*60` to something much larger so that S3 caches the redirects for much longer. But that's rather immaterial. Having to figure out the redirect (in our Django) every 15min is peanuts compared to having to actually serve the file. 

So, perhaps leave your fancy if-modified-since stuff, as is, since it's rarely going to matter much anyway. And if so, perhaps remove the code comment about the note.

PR #6000, opened 2019-10-16
Provide valid aria role for logo

### LINKED ISSUE #5999: Provide valid aria role for logo
**Summary**
The logo in the header has an aria role of "image". That's [not a valid role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques)

**Expected behavior**
The logo should have a valid role. "img" in this case.

atopal:
Fixes #5999 

Replacing the invalid role "image" with the valid role "img"


atopal:
Thank you for the super fast review :D

PR #6020, opened 2019-10-21
Use semantic HTML for 'Last modified' piece

### LINKED ISSUE #5975: Use semantic HTML for "Last modified" piece
This is a follow-up to https://github.com/mozilla/kuma/pull/5974

Here's what the HTML looks like:

```html
<section class="contributors-sub"><svg class="icon icon-clock" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 28" aria-hidden="true"><path d="M14 8.5v7c0 .281-.219.5-.5.5h-5a.494.494 0 0 1-.5-.5v-1c0-.281.219-.5.5-.5H12V8.5c0-.281.219-.5.5-.5h1c.281 0 .5.219.5.5zm6.5 5.5c0-4.688-3.813-8.5-8.5-8.5S3.5 9.313 3.5 14s3.813 8.5 8.5 8.5 8.5-3.813 8.5-8.5zm3.5 0c0 6.625-5.375 12-12 12S0 20.625 0 14 5.375 2 12 2s12 5.375 12 12z"></path></svg><header><h4>最后修改：</h4></header><time datetime="2019-10-11T02:30:33.778708">2019年10月11日</time>,<!-- --> <a href="https://wiki.developer.mozilla.org/zh-CN/docs/Web/CSS/linear-gradient$history" rel="nofollow">by MDN contributors</a></section>
```
Screen Shot 2019-10-11 at 11 34 47 AM

The problem is that `h4` is naturally a block element so it requires CSS to force it to be `display:inline-block`. 

Here's what it looks like in "screen reader" like `links`:

Screen Shot 2019-10-11 at 11 12 35 AM

All we need is this:
```html
<p class="last-modified"><b>Last modified:</b> Sep 29, 2019</p>
```
Alternatively this:
```html
<p class="last-modified">
  <svg>...</svg>
  <b>Last modified:</b> Sep 29, 2019
</p>
```
so that we can control the svg icon from CSS if we have to. 

In general, in Kuma we use inline `<svg>` which means that we can control the SVGs with CSS if we have to.

joedarc:
Fixes #5975

Looks like the styling to the existing structure was removed [here](https://github.com/mdn/kuma/pull/5952/) and it looks odd on production again. With the changes here's how it looks:

Screen Shot 2019-10-21 at 11 36 03 AM


joedarc:
@schalkneethling, @peterbe thoughts?

kuma/javascript/src/last-modified.jsx
```diff
@@ -28,21 +26,26 @@ export default function LastModified({
...
-                    dateStringOptions
-                )}
-            </time>
-            ,{' '}
-            <a href={`${wikiRevisionHistoryURL}`} rel="nofollow">
-                {gettext('by MDN contributors')}
-            </a>
+            <ul>
+                <li className="last-modified">
+                    {' '}
```
peterbe: This is surely not needed.

joedarc:
@peterbe thanks for the review. I removed that extra space(was just used to pad the icon) and added a little more styling, i think it looks better. Let me know!

Screen Shot 2019-10-21 at 10 12 28 PM

I know it's a subtle difference, but it lines up the SVG with the text a little better and looks more similar to how it looks in the wiki.

kuma/static/img/clock.svg
```diff
@@ -0,0 +1,3 @@
+<svg class="icon icon-clock" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 28" aria-hidden="true">
+    <path d="M14 8.5v7c0 .281-.219.5-.5.5h-5a.494.494 0 0 1-.5-.5v-1c0-.281.219-.5.5-.5H12V8.5c0-.281.219-.5.5-.5h1c.281 0 .5.219.5.5zm6.5 5.5c0-4.688-3.813-8.5-8.5-8.5S3.5 9.313 3.5 14s3.813 8.5 8.5 8.5 8.5-3.813 8.5-8.5zm3.5 0c0 6.625-5.375 12-12 12S0 20.625 0 14 5.375 2 12 2s12 5.375 12 12z"/>
+</svg>
```
schalkneethling: We do already have this file located here:
https://github.com/mdn/kuma/blob/master/jinja2/includes/icons/general/clock.svg#L1

kuma/static/styles/minimalist/structure/_article.scss
```diff
@@ -1,6 +1,9 @@
 /*
 Main wiki article content
 ====================================================================== */
+
+$clock-url: '../img/clock.svg';
```
schalkneethling: Changing this to the following should work here:

```suggestion
$clock-url: '../../../general/clock.svg';
```

kuma/javascript/src/last-modified.jsx
```diff
@@ -28,21 +26,25 @@ export default function LastModified({
...
-                    dateStringOptions
-                )}
-            </time>
-            ,{' '}
-            <a href={`${wikiRevisionHistoryURL}`} rel="nofollow">
-                {gettext('by MDN contributors')}
-            </a>
+            <ul>
+                <li className="last-modified">
+                    <b>{gettext('Last modified:')}</b>{' '}
```
schalkneethling: It is unfortunate that we have all of those `{' '}`, would be a terrible idea to include the space inside the `gettext` call, for example:

```suggestion
                    <b>{gettext('Last modified: ')}</b>
```

kuma/static/styles/minimalist/structure/_article.scss
joedarc: @schalkneethling it seems like it's not finding the icon located there. I tried to just do:

`$clock-url: '../../../general/clock.svg';`
and 
`$path-to-svg-general-icons + 'clock.svg';` because that variable was defined. The reason i moved the clock to the img folder was to follow [this](https://github.com/mdn/kuma/blob/6b6cb9f7648aba08ee797a1190973929cf13401a/kuma/static/styles/minimalist/structure/_footer.scss#L1) which works well. 

Let me know if you have any other input.

kuma/javascript/src/last-modified.jsx
joedarc: Unfortunately if i add the space it doesn't translate properly.

peterbe: Yeah, the `{' '}` after the `</b>` is better otherwise the translators will have to deal with a whitespace.

peterbe: > would be a terrible idea to include the space inside the `gettext` call, for example:

@schalkneethling ?? Yes, it would be terrible. Was it a mistake to make the suggested edit?

schalkneethling: It was totally just a suggestion, it can be safely ignored ;)

kuma/static/styles/minimalist/structure/_article.scss
schalkneethling: So, this will work: `../general/clock.svg`

I noticed that the current SVG has the `width` and `height` attributes on it as well as `viewBox`. The problem was not introduced here, but I would ask that you remove the `width` and `height` attributes as well. 

Our icon set should also have a base color of `#333` but that is just a note to self. I also notice that this icon does *not* do well at `16px` so I should redraw this one as well.

Perhaps to make it render a bit better, for now, let’s set the font size of `.metadata ul` to `@include set-font-size(16px);` and then set the size of the icon to `18px`

joedarc: @schalkneethling 

> So, this will work: ../general/clock.svg

Correct, thank you 👍 

I didn't adjust the size of the icon because it looked good to me, when i made it `18px` it looked a little too big in my opinion.

16px:

Screen Shot 2019-10-24 at 9 39 13 AM
 
18px:

Screen Shot 2019-10-24 at 9 45 21 AM

However, feel free to look over the changes and if you still want that size change just let me know!

kuma/static/styles/minimalist/structure/_article.scss
```diff
@@ -20,5 +23,24 @@ Main wiki article content
         margin-top: 32px;
         font-size: 0.88889rem;
         color: $grey;
+
+        header {
+            @include visually-hidden;
+        }
```
schalkneethling: The way this is written we do not actually need the class `visually-hidden` on the `header` element. I am in two minds to whether we should just remove the class from the header, or change this to `.visually-hidden` 🤔 

Having the class on the `header` element makes the intent explicitly clear by just looking at the HTML. On the other hand, if you highlight the element in devtools and look at the CSS, it will also be clear what is happening. 🤔 

Perhaps doing something like:

```
header.visually-hidden {
  @include visually-hidden;
}
```

What do you think @joedarc?

joedarc: > Having the class on the header element makes the intent explicitly clear by just looking at the HTML.

@schalkneethling, I definitely agree. It's minor but it makes more sense when inspecting the HTML. Made the requested change. 👍

PR #6027, opened 2019-10-22
Legacy namespace documents not indexed/followed

### LINKED ISSUE #6022: Legacy namespace documents are indexed/followed in production
**Summary**
Legacy namespace documents, like https://developer.mozilla.org/en-US/docs/User:anonymous:uitest, are given a `robots` `meta` tag of `index, follow`. This may not be important, but I just wanted to record it since it's a deviation from the Wiki view. I discovered this when running the new and improved headless integration tests against production ([the test `test_user_document`](https://github.com/mdn/kuma/blob/b94d91211ed959121b019fc31bf6c63087cac5ff/tests/headless/test_endpoints.py#L126) fails).

**Steps To Reproduce (STR)**
1. Visit https://developer.mozilla.org/en-US/docs/User:anonymous:uitest
2. View the source HTML

**Actual behavior**
image

**Expected behavior**
```
...
...
```

**Additional context**
See the [new document template](https://github.com/mdn/kuma/blob/b94d91211ed959121b019fc31bf6c63087cac5ff/kuma/wiki/jinja2/wiki/react_document.html#L15) as opposed to the [Wiki document template](https://github.com/mdn/kuma/blob/b94d91211ed959121b019fc31bf6c63087cac5ff/kuma/wiki/jinja2/wiki/document.html#L41).

peterbe:
Fixes #6022

### DIFF
--- a/kuma/wiki/jinja2/wiki/react_document.html
+++ b/kuma/wiki/jinja2/wiki/react_document.html
@@ -12,13 +12,7 @@
   <title>{{ page_title(document_data['title'] + seo_parent_title) }}</title>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <meta name="robots" content=
-    {%- if request.get_host() in settings.ALLOW_ROBOTS_WEB_DOMAINS -%}
-        "index, follow"
-    {%- else -%}
-        "noindex, nofollow"
-    {%- endif -%}
-  >
+  <meta name="robots" content="{{ robots_meta_content }}">
 
   {% include "includes/preload.html" %}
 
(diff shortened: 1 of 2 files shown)


peterbe:
If I retrace the steps how this was done in the wiki there are more possibilities of that `doc_html` variable being falsy but I don't know how realistic they are. 

Also, I haven't compared this logic to the logic used when building the sitemaps. Perhaps it would be nice if they could share a common (fast!) function. Mind you, the sitemaps might be building based on an ORM query whereas viewing an individual query is based on it already being extracted from the ORM. So, perhaps an overkill ambition.

PR #6038, opened 2019-10-24
Add Google OAuth as another authentication provider

### LINKED ISSUE #6009: Google Auth Workflow: Backend OAuth Flow
Wire up Google OAuth logins.

I would suggest this, but it looks like it does not support 2.7:
https://github.com/googleapis/google-api-python-client

So might have to go with this:
https://pypi.org/project/google-auth/

Front-end HTML/CSS is here:
https://app.zenhub.com/workspaces/mdn-sprint-board-5ab3d23fd0f4ea53b0022a61/issues/mozilla/kuma/6008

To get this work rolling, can this work be started with undesigned pages/no graphics and just links? After the front-end is done, then implementing this on the front-end should be able to happen quickly. 

Acceptance:
- [ ] user can log in with Google Acct
- [ ] handle both cases of profile image showing (1) default anonymous profile image (2) existing profile image. See design comps for this:

no profile pic: https://mozilla.invisionapp.com/share/Y8SGMKQBCW3#/screens/374829955
has profile pic: https://mozilla.invisionapp.com/share/Y8SGMKQBCW3#/screens/374095626

### LINKED ISSUE #6023: Google Auth Workflow: Backend OAuth Flow Tests
Depends on #6009 

**Acceptance criteria**
Create a suite of tests that are at least as exhaustive as the ones we have for GitHub OAuth.

Gregoor:
Fixes #6009, Fixes #6023 

What it does:
- adds all the auth routes for Google (analogous to the GitHub routes we already had)
  - **Update:** Since Schalk's PR has been merged, signup has changed. If you want to test the first step, visit localhost.org:8000/users/google/login
- defaults the username to the local-part of the email address (the thing before the @) if no username is given
- retrieves avatar from any provider given

Now for some venting (grown-up word for whining):
This took longer than it should, for what it ultimately is. One reason that I failed to notice that the `.env` `DOMAIN` key SHALL NOT contain a port, which just silently fails. allauth throws a PermissionDenied error, which didn't help much. What helped much was asking Ryan who pointed the problem out, thanks again! ~So one of the remaining **TODO**s for me is figuring out if I can make that throw a more descriptive error so that no one has to bang their head against that particular wall again.~

~The other **TODO** is documenting how to set-up a Google Oauth account and the required DOMAIN changes (Google requires a valid TLD, so no more mdn.localhost). I'd also like to extend the `configure_github_social` command to take care of "automatically" configuring this.~

~I've put tests into a separate issue (#6023), I'll work on that one next.~


Gregoor:
@schalkneethling r?

schalkneethling:
Thanks, @Gregoor. Here is what I would like to do, and then we can take it from there if it makes sense to you as well

- [ ] 1. Get https://github.com/mdn/kuma/pull/6057 merged. Currently, there is something strange going on with CI, or perhaps it is the Python tests themselves. I am going to poke at it a little more but, would like to get @escattone's input
- [ ] 2. See if merging that causes any conflicts here and have @escattone review the Python bits here. I am pinging Ryan specifically as he has had the most/latest interaction with the accounts work so, it might be fresher in his mind.
- [ ] 3. Get this merged(behind a flag if there is a possibility that this can cause problems if pushed to production)
- [ ] 4. Work on the front-end portion and let you know of any stumbling blocks I run into during that process.

If that makes sense/works for you, then let’s take it from the top :)

Gregoor:
Thanks for lining out a plan, Schalk! Makes sense to me. I could put the google auth stuff behind a flag though atm I don't think there's a damage in exposing the endpoint early. I.e. people would have to explicity type the route and the "damage done" is that they then have an account through Google, which they'll be able to have a week or two later anyway. Wdyt?

callahad:

kuma/attachments/management/commands/configure_social_auth.py
```diff
@@ -0,0 +1,88 @@
+import fileinput
```
peterbe: Why is this script in the `kuma/attachments` app? Perhaps a mistake?

peterbe:
Very strangely, here's what happened:
```
▶ docker-compose exec web ./manage.py configure_social_auth


There's already a SocialApp for github, if you want to overwrite it type "yes":


Visit https://console.developers.google.com/projectcreate and create a new project
After that visit https://console.developers.google.com/apis/credentials and add the following URLs:
Authorized JavaScript origins: http://localhost.org:8000
Authorized redirect URIs: http://localhost.org:8000/users/google/login/callback/

Client ID:
```
I didn't actually type `yes`. I just hit Enter when it was waiting for me on that first question about github. I guess that means it skipped the `github` provider. So I don't think I understood that what was next was setting up Google. I think you should add `Setting up SocialApp for Google...` to the `GOOGLE_INFO` constant. 

But what was really strange was what happened next

I clicked the little "Copy to clipboard" on:
Screen Shot 2019-11-15 at 12 29 26 PM

But that actually put this into my clipboard:
```
\n867917246374-j7us63g68e3e3oei5iju2jtk6lmk20eb.apps.googleusercontent.com\n
```
So by pasting that into the `input()` it actually thinks I don't want to type in anything there. 
The solution to this is to wrap the input() in a loop. 

```python
while True:
   client_id = input('Client ID:').strip()
    if client_id:
        break
```

If someone decides "Uh? I don't have one of those. I wanna bail." then they run Ctrl-C which raises a `KeyboardInterupt` exception and exists as expected.

peterbe:
Perhaps this is a matter of the flux of change but this is what I get after the consent screen:
Screen Shot 2019-11-15 at 12 59 27 PM

peterbe:
Again, sorry if this is unrelated but a bug's a bug. It used to say "Username  Peter Bengtsson" but when I clicked "change username" I get this:
Screen Shot 2019-11-15 at 1 01 15 PM

peterbe:
Note, it used to be "Peter Bengtsson" (which is a strange username, but whatever) and now the default text in the input is just "p".

peterbe:
So, I restarted and this time after the Google consent screen I just pressed the blue "Create profile" button and didn't touch anything else and ended up with this:
Screen Shot 2019-11-15 at 1 04 37 PM

(The TOCs checkbox I get)

There seems to be some confusion as to what the display username and the input username is.

peterbe:
Something strange is happening with the avatar display on the right. It shows my mug for brief second and when it realizes there's a form validation error (the TOCs) it goes back to the dino head:
Screen-Recording-2019-11-15-at-1 06 45-PM

docs/installation.rst
```diff
@@ -208,34 +208,48 @@ http://localhost:8000/admin/login/
...
+* Authorization callback URL: http://localhost.org:8000/users/github/login/callback/.
+
+To enable Google authentication, you'll need to first `create an API project on Google`_.
+After that we'll need to `configure credentials for that project`_ with settings like:
 
-To automate setting Django up for Github auth you can run
-``docker-compose exec web ./manage.py configure_github_social`` and follow its steps.
+* Name: MDN Development for (<username>).
+* Authorized JavaScript origins: http://localhost.org:8000.
+* Authorized redirect URIs: http://localhost.org:8000/users/google/login/callback/.
```
peterbe:
```suggestion
* Authorized redirect URIs: http://localhost.org:8000/users/google/login/callback/
```

docs/installation.rst
```diff
@@ -208,34 +208,48 @@ http://localhost:8000/admin/login/
...
-* Authorization callback URL: http://localhost:8000/users/github/login/callback/.
+* Authorization callback URL: http://localhost.org:8000/users/github/login/callback/.
+
+To enable Google authentication, you'll need to first `create an API project on Google`_.
+After that we'll need to `configure credentials for that project`_ with settings like:
 
-To automate setting Django up for Github auth you can run
-``docker-compose exec web ./manage.py configure_github_social`` and follow its steps.
+* Name: MDN Development for (<username>).
+* Authorized JavaScript origins: http://localhost.org:8000.
```
peterbe:
```suggestion
* Authorized JavaScript origins: http://localhost.org:8000
```
to avoid the risk of accidentally copy-n-pasting that.

kuma/users/views.py
```diff
@@ -477,11 +477,16 @@ def get_form(self, form_class=None):
         self.email_addresses = {}
         form = super(SignupView, self).get_form(form_class)
         form.fields['email'].label = _('Email address')
+
+        # When no username is provided, default to the local-part of the email address
+        if form.initial.get('username', '') == '':
+            form.initial['username'] = form.initial.get('email', ('',))[0].split('@')[0]
```
peterbe: Ah! This is why my default username input became "p". I speculate that `form.initial['emai'] == 'pbengtsso@mozilla.com'` but this code here expects it to be `form.initial['emai'] == ['pbengtsso@mozilla.com']`.

peterbe: Yeah, I put a print on `form.initial` here, deleted myself, and started over. It yields:
```
{'email': 'pbengtsson@mozilla.com', 'username': '', 'first_name': 'Peter', 'last_name': 'Bengtsson'}
```
therefore `form.initial.get('email', ('',))[0].split('@')[0]` will yield `'p'`.

Gregoor: Weird, I can see that behavior now too. I actually changed that to a tuple after `.get(` started returning a tuple after the Python 3 upgrade, which I thought was an API change. But now I'm getting the same behavior as you, so I reverted it.

Gregoor:
Thank you very much for the thorough review Peter, and sorry you had to suffer some of those Google pains too, I could've done a better job shielding you from it. I hope with those changes you recommended (and I just pushed) at least the pain of a future reviewer will be reduced.

Wrt frontend, yes that's not behaving quite right yet but I think Schalk and I agreed that this will be tackled in another issue, leaving this to be (mostly) about the backend.

kuma/attachments/management/commands/configure_social_auth.py
Gregoor: Moved! Thanks for catching that mistake

kuma/attachments/management/commands/configure_social_auth.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        for social_app in social_apps:
+            social_app.sites.add(site)
+
+        print('\n')
+
+        print(ENV_INFO)
+        overwrite_or_create_env_vars(
+            {'SITE_ID': site.id, 'DOMAIN': DOMAIN} if site.id != settings.SITE_ID else
+            {'DOMAIN': DOMAIN})
```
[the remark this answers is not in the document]
escattone: Yes, I agree here. When you add something to `.env`, you'll have to do a `docker-compose down` followed by a `docker-compose up`.

kuma/attachments/management/commands/configure_social_auth.py
```diff
@@ -0,0 +1,88 @@
...
+        print('\n')
+
+        social_apps = []
+        for provider in ['github', 'google']:
+            social_app = SocialApp.objects.filter(provider=provider).first()
+            if social_app is None or input(OVERWRITE_PROMPT % provider) == 'yes':
+                print('\n')
+
+                print(GITHUB_INFO if provider == 'github' else GOOGLE_INFO)
+                client_id = input('Client ID: ').strip()
```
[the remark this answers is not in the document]
escattone: Sweet! I've always wanted something like the "walrus operator", and most often when I was writing `while` loops. So nice that we're using the latest Python now!

kuma/users/templatetags/jinja_helpers.py
```diff
@@ -14,13 +14,13 @@
 
 
 @library.global_function
-def get_avatar_url(user, provider='github'):
+def get_avatar_url(user):
```
escattone: I guess if the user has both GitHub and Google social accounts, we don't care which avatar we return? Might we want to guarantee the order of the results from the `socialaccount_set` query so it's predictable? For example, perhaps we should order the `socialaccount_set` query by `date_joined` so that we always select the avatar from the first or last social account the user signed-up with?

kuma/users/management/commands/configure_social_auth.py
```diff
@@ -0,0 +1,92 @@
...
+from django.contrib.sites.models import Site
+from django.core.management.base import BaseCommand
+
+LOCALHOST = 'localhost:8000'
+DOMAIN = 'localhost.org'
+
+OVERWRITE_PROMPT = 'There\'s already a SocialApp for %s, if you want to overwrite it type "yes":'
+GITHUB_INFO = (
+    'Visit https://github.com/settings/developers and click "New OAuth App"\n'
+    'Set "Homepage URL" to "http://mdn.localhost:8000/" and Authorization callback URL to ' +
```
escattone: Should this be `"http://localhost.org:8000/"` instead of `"http://mdn.localhost:8000/"`? I think we're going to settle on `localhost.org` for everything instead of `mdn.localhost`?

kuma/users/management/commands/configure_social_auth.py
```diff
@@ -0,0 +1,92 @@
+import fileinput
+import os
+import sys
+
+from allauth.socialaccount.models import SocialApp
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.core.management.base import BaseCommand
+
+LOCALHOST = 'localhost:8000'
```
escattone: Should this be `localhost.org:8000` as well? I'm not sure anymore.

schalkneethling:
@Gregoor @peterbe @escattone 

Indeed, thank you everyone for your reviews. Anything front-end related that seems or is flaky is to be expected. The start of the work to add the actual front-end is here https://github.com/mdn/kuma/pull/6094

Even that does not address everything as it is blocked on this work being merged. I will address all of that once the back-end code has landed.

The one thing we need to be sure of is that when this is merged, it does not negatively impact any of the existing sign-up flow, why I was thinking that we should probably have this new feature behind Constance and/or Waffle.

kuma/attachments/management/commands/configure_social_auth.py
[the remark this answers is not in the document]
peterbe: Not `docker-compose down`!! Won't that wipe the persistent layers?  :)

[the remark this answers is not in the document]
escattone: I've always used `docker-compose down`, which not only stops the containers but removes them (but it doesn't remove the volumes storing the DB data, for that you'd have to do a `docker-compose down --volumes`). My thinking was that new containers will be created anyway when the `docker-compose up` is run, but maybe `docker-compose stop` is good enough?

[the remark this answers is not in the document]
Gregoor: Whoops, stop should be enough here indeed. Thanks for the feedbacks!

Gregoor:
@escattone Thankslots for the review and the considerations, I tried to put all ! I have to admit I didn't fully grasp the Persona stuff, but I think your filtering for provider proposal makes sense, so I went with that.
And also very much share your allauth pain, don't think it's easy to follow their flows as it does have a case of lasagna code, too many layers. Thanks for enduring the pain!

---

@peterbe I vaguely remember us opting for separate accounts per social registration, but can't find it atm. Anyway I'm with you that what you outlined makes more sense, for Common Voice we did do it like that. Two things come to mind:

This might lead to account loss, when you change your address on GitHub. How do you get your account back? For Common Voice this worked because we also had passwordless-login (i.e. getting a login-link via email), thanks to Auth0.

There's also a general security angle here, i.e. account take over through providers that don't do proper verification, suddenly the weakest link defines our general account security. Though given our two providers that's probably not a concern atm. While GitHub does allow you to add any address and afaik also sends that over on social-auth, it also has a verified flag we're already checking in other places.

Shall we move this discussion into a separate issue? I do see the value, just don't think it should block this.

escattone:
> a case of lasagna code, too many layers

@Gregoor Love that phrase "lasagna code"! ❤️ 😆

escattone:
@Gregoor 

> For Common Voice this worked because we also had passwordless-login (i.e. getting a login-link via email), thanks to Auth0.

I wish we could switch to using Auth0, but my guess is that it would be too costly.

> There's also a general security angle here, i.e. account take over through providers that don't do proper verification, suddenly the weakest link defines our general account security. Though given our two providers that's probably not a concern atm.

That's a really good point that we should keep in mind.

escattone:
> If it's possible, and if the human uses the same email on Google as on GitHub it should definitely re-use the existing profile. Anything else would suck.
> 
> I have a Google account on [pbengtsson@mozilla.com](mailto:pbengtsson@mozilla.com) but my GitHub is on [mail@peterbe.com](mailto:mail@peterbe.com). However, one of my verified email aliases on GitHub is my Google email. If done correctly, if there's overlap in the emails it should be just 1 user account and it means it wouldn't matter which provider I use.

Yes! I'm hoping that we can do it this way as well. Actually, I'm going to do something very much like this for `persona`-related user accounts for https://github.com/mdn/kuma/issues/6037, with the difference being that the `persona` social account would be dropped from the "recovered" user once the GitHub/Google social account has been associated.

atopal:
One question we should answer: What happens when you have different emails for Google/Github and end up with two MDN user accounts? Should we give you the option to merge them down the line when you notice?

Gregoor:
I just had a productive meeting with @escattone where some allauth things clicked for me. He is tackling the log-me-in-when-I-already-have-an-account-for-that-email aspect of all this, which is great! Thanks again!

For merging accounts after the fact we don't have an answer yet. My personal vote be to push this out for when we got rid of the wiki since migrating contributions seems fairly complex to me. But if this feature is important enough, that might trump my worry!

Gregoor:
Flake8 is throwing an exception and my docker engine keeps dying when I try to update update my image (after upgrading flake8). @peterbe @escattone looking at the CI error, it does seem like a flake8 bug that might just be fixed through an upgrade, doesn't it?

atopal:
Update after talking to Mustafa: Yes, let's do the same thing we do with Persona accounts. Log the user in and add that new log-in as the second social auth method.

kuma/users/views.py
```diff
@@ -477,11 +477,19 @@ def get_form(self, form_class=None):
         self.email_addresses = {}
         form = super(SignupView, self).get_form(form_class)
         form.fields['email'].label = _('Email address')
+
+        # When no username is provided, default to the local-part of the email address
+        if form.initial.get('username', '') == '':
+            email = form.initial.get('email', '')
+            if isinstance(email, tuple):
```
peterbe: I'm not sure if it helps but you can do:
```suggestion
            if isinstance(email, (tuple, list)):
```
which is useful sometimes when you're not entirely sure what type it's going to be but it's some iterable.

Gregoor:
@peterbe @escattone any change requests or can we merge this in?

escattone:
@Gregoor I'll take a final look through this tomorrow or Friday, and then hopefully we can merge this. After several more days of banging my head against the wall of `django-allauth` (as part of my work on a PR that allows us to re-use an existing account, for example, if in the past we've signed-up with GitHub but now we're signing-in with Google, or vice-versa), I feel like I know the code even better than I did when I first reviewed this.

kuma/users/views.py
```diff
@@ -584,15 +592,16 @@ def form_valid(self, form):
...
+        # For GitHub/Google users, find matching legacy Persona social accounts
+        assert self.sociallogin.account.provider in ('github', 'google')
         or_query = []
         for email_address in self.email_addresses.values():
             if email_address['verified']:
                 or_query.append(Q(uid=email_address['email']))
         if or_query:
             reduced_or_query = reduce(operator.or_, or_query)
             matching_accounts = (SocialAccount.objects
+                                              .filter(provider='persona')
```
peterbe: You explained this to me on Zoom. Can you also write this down in a code comment so we can remember the why. (The why is that odd thing about UID being a int and sometimes an email)

kuma/users/management/commands/configure_social_auth.py
```diff
@@ -0,0 +1,98 @@
...
+)
+GOOGLE_INFO = (
+    'Visit https://console.developers.google.com/projectcreate and create a new project\n'
+    'After that visit https://console.developers.google.com/apis/credentials and add the following URLs:\n'
+    'Authorized JavaScript origins: http://localhost.org:8000\n'
+    'Authorized redirect URIs: http://localhost.org:8000/users/google/login/callback/\n'
+)
+ENV_INFO = 'Putting SITE_ID and DOMAIN into .env'
+HOSTS_INFO = (
+    'You should should run "docker-compose restart" now, so that the .env updates are applied\n.'
```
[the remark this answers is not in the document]
Gregoor: I just tested it with `restart`, a print in `common.py` and all three python processes logged the new value.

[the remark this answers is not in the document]
escattone: Hm, I'm surprised that `docker-compose restart` worked, as it was my understanding from past experience that a `docker-compose stop`+`docker-compose up` was required as well.

kuma/users/templatetags/jinja_helpers.py
```diff
@@ -16,13 +16,13 @@
...
     """
     Get the user's avatar URL for the specified provider. Assumes that the user
```
escattone: Nit: Just realized that the comment should be updated.

kuma/users/views.py
```diff
@@ -584,12 +595,14 @@ def form_valid(self, form):
...
-        assert self.sociallogin.account.provider == 'github'
+        # For GitHub/Google users, find matching legacy Persona social accounts
+        assert self.sociallogin.account.provider in ('github', 'google')
         or_query = []
         for email_address in self.email_addresses.values():
             if email_address['verified']:
-                or_query.append(Q(uid=email_address['email']))
+                # only persona accounts have emails as UIDs
+                # but adding the provider criteria makes this explicit and future-proof
+                or_query.append(Q(uid=email_address['email']) & Q(provider='persona'))
```
escattone: Nit: This multi-line chunk could be simplified to:
```python
uids = Q()
for email_address in self.email_addresses.values():
    if email_address['verified']:
        uids |= Q(uid=email_address['email'])
if uids:
    matching_accounts = SocialAccount.objects.filter(uids, provider='persona')
else:
    ...
```

kuma/users/management/commands/configure_social_auth.py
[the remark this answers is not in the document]
peterbe: I think I get it. It depends on which env vars it is. 
For example, I don't have `SENTRY_DSN` mentioned at all in `docker-compose.yml`. 
So, if I edit the `.env` file and then trigger a restart, Python (via `./manage.py') will start from scratch and re-read the `.env` file from disk. 
However, take `DEBUG` for example. If I mess with that in my `.env` it will never change until I do a `docker-compose stop ; docker-compose up`. 
That's because of how `docker-compose` works and `python-decouple` doesn't read this value from the file on disk, but from the environment. 
This is because we have this in `docker-compose.yml`:
```
DEBUG=${DEBUG:-True}
```

I genuinely believe we should go for https://github.com/mdn/kuma/pull/5390 instead of listing every possible env var in `docker-compose.yml`. 

Truth be told, I can't guarantee that that'll solve the problem because maybe what `docker-compose` does is that it reads the whole file when you `docker-compose up` and pushes it into memory and only `docker-compose stop` can clear that. 

Either way, the correct instruction is that you have to (as of today) run `docker-compose stop; docker-compose up` if you edit `DOMAIN` because that's listed in `docker-compose.yml`.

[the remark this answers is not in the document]
Gregoor: Thanks for the deep-dive, Peter! That makes some sense (in the unfortunate sense of the word).

I've changed it to stop'n'up instead of restart.

kuma/users/views.py
Gregoor: Thanks, this is much better indeed!

PR #6043, opened 2019-10-24
./manage.py runserver instead of gunicorn for dev

peterbe:
If you put a `--reload` at the end of the existing `gunicorn` line, it'll actually trigger as soon as you edit a .py file in your editor. It'll say:

```
web_1            | [2019-10-24 11:28:25 +0000] [11] [INFO] Worker reloading: /app/kuma/wiki/views/document.py modified
web_1            | [2019-10-24 11:28:25 +0000] [9] [INFO] Worker reloading: /app/kuma/wiki/views/document.py modified
```
in `docker-compose up`.
But it doesn't actually work. I don't know if it's because of the worker class or something else docker related. 

Anyway, this stuff works. It starts faster too. And any edit to a .py file reloads so you don't have to run `docker-compose restart web` anymore. 

I guess we're running `gunicorn` in docker-compose to be more similar to what we run in production, but that's already out of whack with the fact that some much is so different when you're switching `settings.DEBUG` on and off. 

What do you think @escattone ?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,7 +50,7 @@ services:
   # Web is based on worker b/c you cannot clear the "ports" with docker-compose.
   web:
     <<: *worker
-    command: gunicorn -w ${GUNICORN_WORKERS:-4} --bind 0.0.0.0:8000 --access-logfile=- --timeout=120 --worker-class=meinheld.gmeinheld.MeinheldWorker kuma.wsgi:application
+    command: ./manage.py runserver 0.0.0.0:8000
     depends_on:
       - mysql
       - elasticsearch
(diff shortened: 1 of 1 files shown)


escattone:
@peterbe I'd prefer to keep using `gunicorn` as the default, and let users customize their local development with something like you've described. We also have something like you've described in the docs (see https://kuma.readthedocs.io/en/latest/development.html#customizing-the-docker-environment). However, your comments made me think that perhaps we should consider running `make build-static` and the tests in Travis and Jenkins with `DEBUG=False`?

peterbe:
`make build-static` isn't run with `DEBUG=False`?? How then does `django-pipeline` know to fully minimize and concatenate the files? 
In `settings/common.py` `PIPELINE` is set up like this:
```
    'COMPILERS': (
        ('kuma.core.pipeline.sass.DebugSassCompiler'
            if DEBUG else
            'pipeline.compilers.sass.SASSCompiler'),
    ),
```
So surely DEBUG is False otherwise we wouldn't get the production grade .css files compiled from the .scss files. 

Also, tests are always automatically run with `DEBUG=False`. 

Thanks for the tip about `docker-compose.dev.yml`. I had completely forgotten about that!

PR #6056, opened 2019-10-27
fixes #6018 

### LINKED ISSUE #6018: text doesn't wrap on the search results page
OS: Windows7
Browser: Chrome 77.0.3865.75
text doesn't wrap on the search results page.

https://developer.mozilla.org/zh-CN/search?q=stopPropagation

issue-screenshot

and I think you can fix it by adding 
`overflow-wrap: break-word;`
in the container 
solution

ankiiitraj:


ankiiitraj:
Hi @schalkneethling can you review this PR. Thanks.

PR #6066, opened 2019-10-29
don't depend on window.ga till it's actually needed

### LINKED ISSUE #5986: Drop in GA traffic
**Summary**
We are seeing a drop in GA after the react front-end launch which is inconsistent with other data sources (search console in this case)

**Steps To Reproduce (STR)**
1. Go to GA
2. compare traffic before and after the launch

How can we ensure that we are still collecting as much telemetry as before?

peterbe:
Fixes #5986

### DIFF
--- a/kuma/javascript/src/ga-provider.jsx
+++ b/kuma/javascript/src/ga-provider.jsx
@@ -4,9 +4,13 @@ import { useContext, useEffect, useState } from 'react';
 
 export type GAFunction = (...any) => void;
 
-const noop: GAFunction = () => {};
+function ga(...args) {
+    if (typeof window === 'object' && typeof window.ga === 'function') {
+        window.ga(...args);
+    }
+}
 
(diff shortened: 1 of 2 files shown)


kuma/javascript/src/ga-provider.jsx
```diff
@@ -4,9 +4,13 @@ import { useContext, useEffect, useState } from 'react';
 
 export type GAFunction = (...any) => void;
 
-const noop: GAFunction = () => {};
+function ga(...args) {
+    if (typeof window === 'object' && typeof window.ga === 'function') {
```
escattone: I'm wondering, if in the future we should consider making this "Ghostery"-proof like we do [here](https://github.com/mdn/kuma/blob/ea715ffdce74c9016854e74b5f28a25ef3f4d830/kuma/static/js/analytics.js#L40) in `kuma/static/js/analytics.js`? Something to think about, but not for this PR.

PR #6094, opened 2019-11-11
Fixes #6024, add multi auth sign-in ui

### LINKED ISSUE #6024: Sign-In Landing page
**Summary**

We are moving to a new sign-in/sign-up flow where we will be offering two options to users of MDN Web Docs. The first is the existing Github auth and the second new option will be Google Auth.

To facilitate this, we need a new account auth landing page that presents these options to the user(see below):

Screenshot 2019-10-22 at 15 10 50

This issue is related to the creation of the new landing page and changing the current account flow to point to this intermediary page, instead of directly sending users to Github.

**Audience**

All new users of MDN Web Docs

**Workaround**

Currently, we offer only one option for sign-up which possibly narrows our reach.

## Acceptance Criteria

- [ ] When a user clicks the `signin` link in the header, they are sent to a new intermediary page as discussed under `summary` above.
- [ ] Page offers the option between our two auth provider i.e. Github or Google
- [ ] Clicking either will send the user to the third party for authentication
- [ ] Show relevant bullet points

schalkneethling:
This adds a general modal to the codebase and then also adds an auth specific modal. I need to make a few tiny tweaks to some of the files and then write up a proper test case.

@Gregoor feel free to take a look in the mean time. I will also document how to go about using the modal "framework" for other modals, but that will happen on `mdn-fiori` which is where our pattern library/design sytem lives.


schalkneethling:
I reckon the bigger challenge of all of this is going to be to orchestrating how we review, merge and deploy all of these:

- [ ] https://github.com/mdn/kuma/pull/6038
- [ ] https://github.com/mdn/kuma/pull/6083
- [ ] https://github.com/mdn/kuma/pull/5518
- [ ] https://github.com/mdn/kuma/pull/6084
- [ ] https://github.com/mdn/kuma/pull/6094 (this one)

I am thinking some waffles, priorities and/or Constance is going to be in order 😸

tobinmori:
Thanks @schalkneethling. this order looks good to me, but agree that we probably need to integrate waffle flags. 

I can create an issue for this. @Gregoor is this something you can take on?

kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -0,0 +1,36 @@
+(function() {
+    'use strict';
+
+    var authModalContainer = document.getElementById('auth-modal');
+
+    if (authModalContainer) {
```
Gregoor: nit: I'd go for early return here to reduce indentation:
```javascript
if (!authModalContainer) return;
```

kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -0,0 +1,36 @@
...
+        /* The React header takes just a couple of milliseconds to
+           long to load so, we need to wrap the code below in a 
+           setTiemout */
+        setTimeout(function() {
```
Gregoor: why not call this from React then?

Gregoor: oh, because it is also used on non-react controlled pages?

schalkneethling:
@Gregoor 

> Just some nits from my first read, will do a more thorough review later. In general I do wonder if we might want to move more of this code to react-land

Eventually yes, for now it will be used in many places that will still be Jinja for some time so, it makes sense to keep it "agnostic"

kuma/static/js/components/user-signup/auth-modal.js
schalkneethling: Exactly :+1:

schalkneethling:
I am also thinking now that on mobile phones, we do not want to show the modal at all. There we want the user to simply navigate to the auth landing page. I, therefore, need to add a conditional and do an early return if we are on a mobile device.

Would be ideal if we could target only actual mobile phones but, for now, using a `matchMedia` query based on viewport size will have to do. I mean we could do a query on `device-pixel-ratio` but, not all mobile phones have @2x ratio.

schalkneethling:
@Gregoor I have addressed the initial comments and also added some more conditional logic around an early return when:

1. The auth modal is not present or,
2. We are on a mobile phone.

kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -0,0 +1,41 @@
...
+    'use strict';
+
+    var authModalContainer = document.getElementById('auth-modal');
+
+    if (
+        !authModalContainer ||
+        (window.matchMedia && window.matchMedia('(max-width:47.9385em)').matches)
+    ) {
+        return;
+    }
```
schalkneethling: @Gregoor I am thinking now that this piece(the core of it) can(should?) be moved to `modal.js` so it is available to all modal implementations in the same manner. Should we want to change the logic for it, we can do it in one place and it will be inherited by all modal. I am thinking something like:

```
...
shouldShowModal: function(modal) {
  if (
    !modal ||
    (window.matchMedia && window.matchMedia('(max-width:47.9385em)').matches)
  ) {
      return false;
  } else {
      return true;
  }
}
...
```

Then inside a modal implementation such as this:

```
var modal = document.getElementById('auth-modal');

if (!window.mdn.modalDialog.shouldShowModal(modal)) {
  return;
}
```

Thoughts?

jinja2/includes/auth-modal.html
```diff
@@ -0,0 +1,19 @@
...
+      </a>
+      <a href="{{ provider_login_url('github') }}" class="google-auth">
+        {{ _('Sign in with Google') }}
+      </a>
+    </div>
+    <button id="close-modal" class="close-modal" data-last-focusable="true">
+      <span>{{ _('Close modal') }}</span>
+    </button>
+  </section>
+</div>
```
schalkneethling: Thinking over all of this some more, I am thinking that this can be made into a base template that you can then extend when creating a modal. The base template might look something like this:

```
<div id="{{ modal_id }}" class="modal hidden">
  <section class="{{ modal_content_class }}" tabindex="-1" role="dialog" aria-modal="true" aria-labelledby="modal-main-heading">
    <header>
      <h2 id="modal-main-heading">{{ modal_heading }}</h2>
	</header>
    {% block modal_content %}{% endblock %}
    <button id="close-modal" class="close-modal" data-last-focusable="true">
      <span>{{ _('Close modal') }}</span>
    </button>
  </section>
</div>
```

Something like that. Thoughts?

schalkneethling: So, I went ahead and did what I talked about above :) Let me know if this was not a good move ;p

kuma/static/js/components/user-signup/auth-modal.js
Gregoor: Yes, I think that makes a lot of sense given that probably all modals should share that behavior!

kuma/settings/common.py
```diff
@@ -836,6 +836,12 @@ def pipeline_one_scss(slug, **kwargs):
         ),
         'output_filename': 'build/styles/signup-flow.css',
     },
+    'auth-modal': {
+        'source_filenames': (
+            'styles/minimalist/components/auth-modal.scss',
```
peterbe: Why is there a generic `modal.js` that is separate from `auth-modal.js` if that's not the case for the SCSS files?

kuma/settings/common.py
```diff
@@ -1831,6 +1847,8 @@ def get_user_url(user):
 # Whether or not to enable the BCD signalling feature.
 # Affects loading of CSS (statically) and JS (in runtime).
 ENABLE_BCD_SIGNAL = config('ENABLE_BCD_SIGNAL', default=True, cast=bool)
+# Multi Auth Switch
```
peterbe: That comment isn't adding any value.

kuma/static/js/components/modal.js
```diff
@@ -0,0 +1,88 @@
+(function() {
+    'use strict';
+
+    if (!window.mdn) {
+        window.mdn = {};
```
peterbe: This code would not be needed if the modal.js wasn't `async` loaded.

kuma/static/js/components/modal.js
```diff
@@ -0,0 +1,88 @@
...
+    var modalDialog = {
+        /**
+         * Returns `false` if either of the following conditions are true else, returns `true`:
+         * 1. Multi auth is not turned on
+         * 2. The modal we expect is not present
+         * 3. We are on a mobile phone
+         * @param {Object} modal - The modal as an `HTMLElement`
+         * @returns {Boolean}
+         */
+        showModal: function (modal) {
```
peterbe: This is poorly named. the `verbNoun` makes me think it's going to *do something* with the noun.

kuma/static/js/components/modal.js
```diff
@@ -0,0 +1,88 @@
...
+            if (
+                !window.mdn.multiAuthEnabled ||
+                !modal ||
+                (window.matchMedia &&
+                    window.matchMedia('(max-width:47.9385em)').matches)
```
peterbe: So, this means that as soon as this JS code is loaded (which is basically as soon as the JS has been downloaded or cached) it executes a `window.matchMedia()` even though the DOM might not have been painted yet. 
Also, can't you just use `clientWidth` instead? It's bound to be [order of magnitude faster](https://www.measurethat.net/Benchmarks/Show/1213/3/matchmedia-vs-clientwidth#latest_results_block)

peterbe: Also, 99.9% of uses don't click "Sign in". For all of these 99.9% of people, they have to delay their page rendering for this DOM operation to complete first. 

What it ought to be is that this stuff is figured out in runtime. I.e. 
```javascript
document.querySelector('a.signin-linlk').addEventListener('click', event=> {
  if (isModalEnabled()) {
      event.preventDefault()
     showModal();
  }
});
```

kuma/static/js/components/modal.js
```diff
@@ -0,0 +1,88 @@
...
+                    event.shiftKey &&
+                    event.key === 'Tab' &&
+                    event.target === firstFocusable
+                ) {
+                    event.preventDefault();
+                    lastFocusable.focus();
+                }
+            });
+
+            modal.addEventListener('keyup', function(event) {
```
peterbe: That should be `document.body.addEventListener` so you can press `Esc` anywhere on the page.

kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -0,0 +1,40 @@
...
+    var authModalContainer = document.getElementById('auth-modal');
+
+    if (!window.mdn.modalDialog.showModal(authModalContainer)) {
+        return;
+    }
+
+    /* The React header takes just a couple of milliseconds to
+       long to load so, we need to wrap the code below in a 
+       setTimeout */
+    setTimeout(function() {
```
peterbe: Can be removed if not `async` loaded. 

PS. Tease me all you like but I still don't understand why all of this code stuff is done in ES5 vanilla JS and not in react.

kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -0,0 +1,40 @@
...
+                event.preventDefault();
+                var modalContentContainer = authModalContainer.querySelector(
+                    'section'
+                );
+                var closeModalButton = document.getElementById('close-modal');
+
+                authModalContainer.classList.remove('hidden');
+                modalContentContainer.focus();
+
+                closeModalButton.addEventListener('click', function() {
```
peterbe: Why can't the modal itself be responsible for this kind of stuff?

kuma/settings/common.py
```diff
@@ -973,6 +979,16 @@ def pipeline_one_scss(slug, **kwargs):
...
         },
     },
+    'auth-modal': {
+        'source_filenames': (
+            'js/components/modal.js',
+            'js/components/user-signup/auth-modal.js',
+        ),
+        'output_filename': 'build/js/auth-modal.js',
+        'extra_context': {
+            'async': True,
```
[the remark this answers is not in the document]
schalkneethling: Yup, this should be `defer` instead

schalkneethling:
Oh noes! Ah well, nothing that cannot be fixed and or improved :) I will break these out into an issue, or issues, as appropriate. We can then discuss the details in each issue, as appropriate :)

Thanks @peterbe

schalkneethling:
> We're using a modal at all. 99% of users don't click sign in. Either because they're not interested or they've already signed in. But now they have to pay for ~3KB of CSS and 1.7KB on every page load that blocks rendering.

You can see the details around this decision here https://github.com/mdn/kuma/issues/6024#issuecomment-545900792 and here https://github.com/mdn/kuma/issues/6024#issuecomment-545902712

schalkneethling:
> It's weird to not use modals for mobiles. Why can't they get a modal? Granted, the modal will be almost as wide as the width of the screen but this has been solved before.

It simply felt like the benefit of the modal was a little lost on mobile and simply sending them to the dedicated page made more sense.

With that said, we will be sending code down to mobile users for the modal, but knowing we will never execute it, which is not great at all. Perhaps we can revisit this. Opened https://github.com/mdn/kuma/issues/6274

schalkneethling:
> Where in the modal does it present a reminder to what the user picked last time?

There was no requirement for this in the current design/spec. But if we could do that, that would be great to add.

schalkneethling:
> Why is this not implemented in React? ES5 is dead to me :)

It is used across pages that have not been converted to React yet. We could have done it as a separate React bundle/component and injected it as we do for BCD Signaling, but there was a time/deadline aspect to this as well.

kuma/settings/common.py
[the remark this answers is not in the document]
schalkneethling: https://github.com/mdn/kuma/issues/6275

kuma/settings/common.py
schalkneethling: It is separate as ` kuma/static/styles/minimalist/components/_modal.scss`. Currently `auth-modal` is the only one using it so it pulls it in directly as a dependency here `kuma/static/styles/minimalist/components/auth-modal.scss`. Once we have two different modals using `_modal.scss` we can/should load it as a separate bundle so we do not include the same CSS twice.

kuma/settings/common.py
schalkneethling: I added it more as a quick way to find the switch but, the name of the property is just as easy to search for :) I will add a bit more context around it.

kuma/static/js/components/modal.js
schalkneethling: True, but I was also following a pattern that is used in other places in the codebase. Perhaps it is moot at this stage and really simply is not needed. Once we mark it as `defer` would you vote to simply remove these lines?

kuma/static/js/components/modal.js
schalkneethling: You are 100% correct. Perhaps `shouldShowModal`?

kuma/static/js/components/user-signup/auth-modal.js
schalkneethling: It definitely can handle adding the event listener to the `closeModalButton`. 100%

peterbe:
> > Where in the modal does it present a reminder to what the user picked last time?
> 
> There was no requirement for this in the current design/spec. But if we could do that, that would be great to add. Should we create an issue to discuss this?

https://github.com/mdn/kuma/issues/6279

peterbe:
> > Why is this not implemented in React? ES5 is dead to me :)
> 
> It is used across pages that have not been converted to React yet. We could have done it as a separate React bundle/component and injected it as we do for BCD Signaling, but there was a time/deadline aspect to this as well.

I vaguely remember the debate but I also remember that I kinda gave up on trying to stay in the debate. Permission to not understand everything™.

Using Jinja, has the advantage that we don't first need to solve proper code-splitting in Kuma's front-end. 

Perhaps it doesn't matter. Shipping is better than no shipping. 
When we re-implement this all over again in Yari we'll do it properly instead. (I.e. full React implementation and no bloat to the generic JS bundle)

peterbe:
> > We're using a modal at all. 99% of users don't click sign in. Either because they're not interested or they've already signed in. But now they have to pay for ~3KB of CSS and 1.7KB on every page load that blocks rendering.
> 
> You can see the details around this decision here [#6024 (comment)](https://github.com/mdn/kuma/issues/6024#issuecomment-545900792) and here [#6024 (comment)](https://github.com/mdn/kuma/issues/6024#issuecomment-545902712)

The first comment just talks about the link is determined by a waffle flag. (Eventually it wasn't a waffle flag but `settings.MULTI_AUTH_ENABLED`)

The second comment is a suggestion that hints that it would maybe be less work to *only* bother to implement the modal functionality. 

However, what if you're trying to access a page that you're not allowed to access. (At the moment these don't exist but we can image `/en-US/settings/subscription`) Where is it going to redirect to? 
We could redirect the homepage and have something like this:
```
window.onload = () => {
  if (document.location.search.includes('?next=')) {
    triggerAuthModal();
  }
}
```
but that doesn't taste great. 

Truth is that Kuma's front-end is inflexible in that we can't use a modal without paying the heavy tax of forcing every single page load the extra bytes of JS, CSS, and HTML. 

If a modal is bestest we'll keep it and simply pay the price of reducing perf on every page. When we move to Yari we'll have proper code-splitting so that left-clicking "Sign-in" can trigger a lazy load of the JS and CSS that's needed. 

My personal preference is that we ONLY have the sign in landing page because it's so super rare that you'd click it and benefit from that sweet modal CSS & JS. 
If it was a shopping carts content thing I think a modal would make a lot more sense. 

Perhaps we just leave it like it is and when we go to Yari we can defer this to @malqinneh and code based on what he thinks is the idealest bestest thing for users.

peterbe:
To back up the "argument" with facts, here's what the so called bloat is:

* HTML - 847B (430B gzipped (but hard to compare))
* JS - 1.7K  (628B gzipped)
* CSS - 2.3K (856B gzipped)

**Note** This is bytes I have to pay for as **an already signed-in user too**.

peterbe:
Perhaps @atopal is missing out on this nerdy discussion. This mention will ping him. :)

PR #6096, opened 2019-11-11
Defer newsletter.js

### LINKED ISSUE #6077: Newsletter JS should load deferred
We're currently loading the newsletter.js bundle with `async`. There's probably no reason to do that. 
Thankfully the script tag for it is far down in the HTML so the chances are smaller that it'll cause harm. 

`async` script tags *execute* as soon as they're ready. Unlike `defer` which always execute after the rendering phase. This means that `async` script tags can cause the rendering to have to pause as it CPU processes the `async` script. 

One common scenario is if you click on a page, then click an internal link. This time, the script tag gets noticed in the HTML and the browser realizes that it has it in the browser cache, so it executes it immediately. This might mean it's blocking other important rendering and DOM event handlers.

joedarc:
Fix #6077

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1096,7 +1096,7 @@ PIPELINE_JS = {
         ),
         'output_filename': 'build/js/newsletter.js',
         'extra_context': {
-            'async': True,
+            'defer': True,
         },
     },
     'perf': {


schalkneethling:
@joedarc Whoops, we already have this one[https://github.com/mdn/kuma/pull/6079] from @peterbe

joedarc:
@schalkneethling Whoops! Sorry about that.

PR #6098, opened 2019-11-12
don't 500 on overly big ?page in search

### LINKED ISSUE #6092: search_phase_execution_exception in ES on huuuge ?page params
https://sentry.prod.mozaws.net/operations/mdn-prod/issues/6620806/

```
TransportError: TransportError(500, u'search_phase_execution_exception', u'Result window is too large, from + size must be less than or equal to: [10000] but was [33010]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.')
(24 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/generics.py", line 201, in get
    return self.list(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 42, in list
    page = self.paginate_queryset(queryset)
  File "rest_framework/generics.py", line 173, in paginate_queryset
    return self.paginator.paginate_queryset(queryset, self.request, view=self)
  File "rest_framework/pagination.py", line 204, in paginate_queryset
    self.page = paginator.page(page_number)

TransportError: TransportError(500, u'search_phase_execution_exception', u'Result window is too large, from + size must be less than or equal to: [10000] but was [33010]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.')
```

peterbe:
Fixes #6092

If you add `?page=20` to the `rest_framework`, which, by default, is for the SQL ORM you get this:

```sql
...  OFFSET 200 LIMIT 10
```
(because the default page size is 10 so the offset is 20 * 10).

But if that `page` number is crazy like, `?page=2000` instead you get:
```sql
...  OFFSET 20000 LIMIT 10
```

which MySQL can handle and won't choke. However, Elasticsearch is quick to judge. It'll crash on an offset that large. 

Normally, the way pagination works in `rest_framework` is that it evaluates and validates the `page` query string parameter pretty late and that might be happening AFTER then SQL/ES query is attempted.

### DIFF
--- a/kuma/search/paginator.py
+++ b/kuma/search/paginator.py
@@ -1,4 +1,5 @@
-from django.core.paginator import EmptyPage, Page, PageNotAnInteger, Paginator
+from django.core.paginator import (
+    EmptyPage, InvalidPage, Page, PageNotAnInteger, Paginator)
 from django.utils.functional import cached_property
 
 
@@ -37,6 +38,18 @@ class SearchPaginator(Paginator):
         assigns the count from the ES result to the Paginator.
         """
         number = self.validate_number(number)
+        if number >= 1000:
+            # Anything >=1,000 will result in a hard error in
+            # Elasticsearch which would happen before we even get a chance
+            # to validate that the range is too big. The error you would
+            # get from Elasticsearch 6.x is something like this:
+            #
+            #     Result window is too large, from + size must be less
+            #     than or equal to: [10000] but was [11000].
+            #
+            # See https://github.com/mdn/kuma/issues/6092
+            raise InvalidPage('Page number too large')
+
         bottom = (number - 1) * self.per_page
         top = bottom + self.per_page
 
--- a/kuma/search/paginator.py
+++ b/kuma/search/paginator.py
@@ -20,7 +20,7 @@ class SearchPaginator(Paginator):
         """
         Validates the given 1-based page number.
 
-        This class overrides the default behavior and ignores the upper bound.
+        We also check that the number isn't too large.
         """
         try:
             number = int(number)
@@ -28,16 +28,7 @@ class SearchPaginator(Paginator):
             raise PageNotAnInteger('That page number is not an integer')
         if number < 1:
             raise EmptyPage('That page number is less than 1')
-        return number
 
-    def page(self, number):
-        """
-        Returns a page object.
-
-        This class overrides the default behavior and ignores "orphans" and
-        assigns the count from the ES result to the Paginator.
-        """
-        number = self.validate_number(number)
         if number >= 1000:
             # Anything >=1,000 will result in a hard error in
             # Elasticsearch which would happen before we even get a chance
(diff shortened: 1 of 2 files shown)


kuma/search/tests/test_views.py
```diff
@@ -163,6 +163,14 @@ class PaginationSearchView(SearchView):
         response = view(request)
         assert response.data['pages'] == 2
 
+    def test_paginate_too_large(self):
+        response = self.client.get('/en-US/search', {'page': 1000},
+                                   HTTP_HOST=settings.WIKI_HOST)
+        # It's a bit odd that an invalid query string parameter
+        # causes a 404 Not Found but that's how the paginator for
+        # rest_framework works.
```
joedarc: Agreed..that is a bit odd.

peterbe:
@joedarc You get the same effect as if the `?page=` query string was invalid. E.g. 
Screen Shot 2019-11-12 at 9 31 10 AM

joedarc:
@peterbe awesome, thanks!

kuma/search/paginator.py
```diff
@@ -37,6 +38,18 @@ def page(self, number):
...
+            # Anything >=1,000 will result in a hard error in
+            # Elasticsearch which would happen before we even get a chance
+            # to validate that the range is too big. The error you would
+            # get from Elasticsearch 6.x is something like this:
+            #
+            #     Result window is too large, from + size must be less
+            #     than or equal to: [10000] but was [11000].
+            #
+            # See https://github.com/mdn/kuma/issues/6092
+            raise InvalidPage('Page number too large')
```
escattone: Nit: I would say a better spot for this would be within the `validate_number` method above:

```python
    def validate_number(self, number):
        """
        Validates the given 1-based page number.
        This class overrides the default behavior and ignores the upper bound.
        """
        try:
            number = int(number)
        except (TypeError, ValueError):
            raise PageNotAnInteger('That page number is not an integer')
        if number < 1:
            raise EmptyPage('That page number is less than 1')
        if number >= 1000:
            # Anything >=1,000 will result in a hard error in
            # Elasticsearch which would happen before we even get a chance
            # to validate that the range is too big. The error you would
            # get from Elasticsearch 6.x is something like this:
            #
            #     Result window is too large, from + size must be less
            #     than or equal to: [10000] but was [11000].
            #
            # See https://github.com/mdn/kuma/issues/6092
            raise InvalidPage('Page number too large')
        return number
```

kuma/search/tests/test_views.py
escattone: I agree too. An HTTP 400 (bad request) with some more detail in the explanation seems more appropriate.

kuma/search/paginator.py
peterbe: Got it!

PR #6103, opened 2019-11-13
remove humans.txt generation

### LINKED ISSUE #6099: humans.txt is outdated and arguably a bit pointless
The [humans.txt project](http://humanstxt.org/) isn't particularly important. We're currently only doing it to list past developer's GitHub handles. 
There are a bunch of other potential things we could be putting in there but most of those are already covered by the [contribute.json](https://developer.mozilla.org/contribute.json).

peterbe:
Fixes #6099


kuma/core/views.py
```diff
@@ -56,3 +58,11 @@ def rate_limited(request, exception):
...
     response['Retry-After'] = '60'
     return response
+
+
+@shared_cache_control(s_maxage=60 * 60 * 24 * 30)
+def humans_txt(request):
+    """We no longer maintain an actual /humans.txt endpoint but to avoid the
+    sad 404 we instead now just encourage people to go and use the GitHub
+    UI to see the contributors."""
+    return HttpResponse("See https://github.com/mdn/kuma/graphs/contributors\n")
```
escattone: I was thinking we should make this a 301 redirect. What do you think?

peterbe: I prefer this because it's less disruptive and also explains the transition for those that might care.

peterbe:
@escattone if you're bored, feel free to take a look as Dan is away.

kuma/core/views.py
escattone: That's fine, but then I think we should make the URL a link for convenience.
```suggestion
    return HttpResponse('See <a href="https://github.com/mdn/kuma/graphs/contributors">https://github.com/mdn/kuma/graphs/contributors</a>\n')
```

peterbe: That'd require me to change the content-type to `text/html` which doesn't feel right considering it's `.txt`.

PR #6108, opened 2019-11-13
6039 conditional css loading based on waffle flags

### LINKED ISSUE #6039: jinja2/react_base.html template uses request.user

peterbe:
Fixes #6039

Basically, since jinja is always run anonymously, unlike the React code which uses `/api/v1/whoami` the Waffle conditions can be different in the two contexts. I.e. the base template might say there's no need for the `banners/base.css` but the React (e.g. `banners.jsx`) might say the opposite. 

This PR mimics what `waffle.flag(flag_name)` does but without a `request` so that it instead strives for the "least common denominator".

### DIFF
--- a/kuma/core/templatetags/jinja_helpers.py
+++ b/kuma/core/templatetags/jinja_helpers.py
@@ -203,7 +203,7 @@ def possible_waffle_flag(*names):
     """Return true if there is any Waffle flag by this name that MIGHT be
     true for at least one user."""
 
-    def _any_waffle_flag_chance(name):
+    def any_waffle_flag_chance(name):
         # This is basically the `Flag.is_active()` method as an ORM statement,
         # but instead of using a `request` object it assumes *any* request
         # which means the least common denominator.
@@ -215,4 +215,4 @@ def possible_waffle_flag(*names):
             Q(authenticated=True) |
             Q(testing=True)).exists()
 
-    return any(_any_waffle_flag_chance(name) for name in names)
+    return any(any_waffle_flag_chance(name) for name in names)
--- a/kuma/core/templatetags/jinja_helpers.py
+++ b/kuma/core/templatetags/jinja_helpers.py
@@ -1,13 +1,13 @@
-
-
 import datetime
 import json
+from functools import lru_cache
 from html.parser import HTMLParser
 
 import jinja2
 from django.conf import settings
 from django.contrib.messages.storage.base import LEVEL_TAGS
 from django.contrib.staticfiles.storage import staticfiles_storage
+from django.db.models import Q
 from django.template import defaultfilters
 from django.template.loader import get_template
 from django.utils.encoding import force_text
@@ -18,6 +18,7 @@ from pytz import timezone, utc
 from soapbox.models import Message
 from statici18n.templatetags.statici18n import statici18n
 from urlobject import URLObject
+from waffle.models import Flag
 
 from ..urlresolvers import reverse, split_path
 from ..utils import (format_date_time, is_untrusted, is_wiki, order_params,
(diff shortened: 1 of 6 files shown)


peterbe:
Note, it could be that the Users and Groups in a waffle flag which I don't bother checking. We don't use it for anything at the moment and the chances are small we will again. And perhaps the day we do that we will no longer be using jinja (hopefully).

kuma/core/templatetags/jinja_helpers.py
```diff
@@ -194,3 +195,24 @@ def get_locale_localized(locale):
     language = settings.LOCALES[locale].english
 
     return _(language)
+
+
+@library.global_function
+@lru_cache()
```
escattone: This is a problem, since once the value for a particular flag name enters the cache, it's never cleared until it's pushed out via the LRU mechanism, and then that only happens per Django instance. So in other words, if the Waffle flag is changed via the admin console, the change most likely won't be picked-up by the Django instances, and possibly not ever until this particular LRU cache is manually cleared in each and every Django instance.

peterbe:
I'm going to retry this whole solution but with Option 3 as per your comment.

peterbe:
@escattone I was going to jump straight into option 3 whereby we introduce something like `ENABLE_DEVELOPER_NEEDS` in settings/common.py but then I realized this is clunky. It means that the release of the developer needs survey has to be perfectly timed with a DevOps operation. Actually, I don't even know what the pain or procedure is of just editing an env var and restarting nodes. I'm only familar with editing an env var and then doing a new release. 

Either way, why don't we use Waffle Switches?! They're never dependent on user. If we make the presence of `banners.css` in the CSS bundle dependent on a Waffle Switch it means we 

1. Can't have a disconnect in it's computed value in JS vs. SSR. 
2. Someone like @atopal can (within the expiration time of the CDN) enable something like the Developer Needs survey at his own time without having to do a DevOps communication. 

To be honest, I'm starting to wonder, if you use Waffle Switches (dumb versions of Waffle Flag), how is that different from Constance? And which is preferred?


@atopal Boils down to this; for Developer Needs and BCD Signaling, **do you need the full power of Waffle Flags?** I.e. do you need the **ability to start it for a sub-set of users?**

peterbe:
@atopal Here's the thing. For the loading of the CSS we *must* switch away from waffle Flags because they don't make sense any more since the CSS loading is always without a user. 

But by switching the CSS loading from waffle Flag to waffle Switch is just the CSS loading. You can still use waffle Flags (e.g. 5% of users) for the JS stuff. That way you have full power, all via the Django admin. But it's perhaps annoying and error prone to have to remember that you need to go to https://developer.mozilla.org/admin/waffle/flag/ **and** https://developer.mozilla.org/admin/waffle/switch/ and do so in the right order. 

What I can do is move it all and entirely over to waffle Switches. What do you prefer Kadir?

Note-to-self; branch is called `6039-use-waffle-switch-for-conditional-css`

atopal:
Hey Peter, there are things that have to be for subset of our users, and things that we want to show to all users. I don't mind having to specify which is which and following a procedure.

* For the developer needs survey we can use a switch (on/off)
* For the BCD signals we don't need a switch or flag anymore. 
* For the task completion survey we need to be able to show it to a subset of users
* For the soapbox, we can use a switch (on/off)

escattone:
@peterbe 👍 I'm totally fine with Waffle switches. They're definitely way more convenient in terms of ease of deployment. My only reason for preferring environment variables was that they'd be easier to use when moving into the JAMStack world, but we could instead provide a Django endpoint that the JAMStack build could use to get these settings.

peterbe:
>     * For the soapbox, we can use a switch (on/off)

It's currently not dependent on any flag. It's simply dependent on there being a soapbox message.

atopal:
I don't know how it's implemented, but there is functionality to turn it on and off from the django admin

peterbe:
> I don't know how it's implemented, but there is functionality to turn it on and off from the django admin

I think you're referring to the "Is active" on https://developer.mozilla.org/admin/soapbox/message/84/change/ which has nothing to do with waffle flags. 

And waffle flags is the ultimate problem because they're fundamentally based on the user (yes, its possible to use independent of the user) and that's a thing we *can't* do in the era where the SSR HTML is always unconditionally built without any particular user in mind.

PR #6112, opened 2019-11-14
add endpoint and Jinja helper for username check

### LINKED ISSUE #6021: Backend work to support User Acct creation
### Summary
Create the Kuma back-end required to support the new account signup flow (see #5911).

### Acceptance Criteria
- [ ] Create a new function `is_username_taken(user)` available to Jinja, as well as its associated tests.
- [ ] Create a new API endpoint for doing the same, as well as its associated tests.
- [ ] (Not certain of this one yet) To the endpoint above, add the ability to check if one or more email addresses are already taken by another account, as well as tests.

escattone:
Fixes #6021

### DIFF
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -332,11 +332,11 @@ class SearchViewTests(ElasticTestCase):
         assert data['locale'] == 'fr'
 
 
-@pytest.mark.parametrize(
-    'http_method', ['put', 'post', 'delete', 'options', 'head'])
+@pytest.mark.parametrize('http_method', ('put', 'post', 'delete', 'options'))
 def test_get_user_disallowed_methods(client, wiki_user, http_method):
     """
-    HTTP methods other than GET are not allowed on the api.v1.get_user endpoint.
+    HTTP methods other than GET and HEAD are not allowed on the api.v1.get_user
+    endpoint.
     """
     url = reverse('api.v1.get_user', args=(wiki_user.username,))
     response = getattr(client, http_method)(url)
@@ -345,27 +345,34 @@ def test_get_user_disallowed_methods(client, wiki_user, http_method):
 
 
 @pytest.mark.parametrize('case', ('upper', 'lower'))
-def test_get_existing_user(client, wiki_user, wiki_user_github_account, case):
+@pytest.mark.parametrize('http_method', ('get', 'head'))
+def test_get_existing_user(client, wiki_user, wiki_user_github_account,
+                           http_method, case):
     """
-    Test GET on the api.v1.get_user endpoint for an existing user, and also
-    that the username is case insensitive.
+    Test GET and HEAD on the api.v1.get_user endpoint for an existing user, and
+    also that the username is case insensitive.
     """
     username = getattr(str, case)(wiki_user.username)
     url = reverse('api.v1.get_user', args=(username,))
-    response = client.get(url)
+    response = getattr(client, http_method)(url)
     assert response.status_code == 200
     assert response['content-type'] == 'application/json'
-    data = response.json()
-    assert data['username'] == wiki_user.username
-    assert data['avatar_url'] == wiki_user_github_account.get_avatar_url()
-    for field in ('title', 'fullname', 'timezone', 'locale'):
-        assert field in data
     assert_no_cache_header(response)
+    if http_method == 'get':
+        data = response.json()
+        assert data['username'] == wiki_user.username
+        assert data['avatar_url'] == wiki_user_github_account.get_avatar_url()
+        for field in ('title', 'fullname', 'organization', 'location',
+                      'timezone', 'locale'):
+            assert data[field] == getattr(wiki_user, field)
 
 
-def test_get_nonexisting_user(client, wiki_user, wiki_user_github_account):
-    """Test GET on the api.v1.get_user endpoint for a non-existing user."""
+@pytest.mark.parametrize('http_method', ('get', 'head'))
+def test_get_nonexisting_user(db, client, http_method):
+    """
+    Test GET and HEAD on the api.v1.get_user endpoint for a non-existing user.
+    """
     url = reverse('api.v1.get_user', args=('nonexistent',))
-    response = client.get(url)
+    response = getattr(client, http_method)(url)
     assert response.status_code == 404
     assert_no_cache_header(response)
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -330,3 +330,42 @@ class SearchViewTests(ElasticTestCase):
         assert data['documents']
         assert data['count'] == 5
         assert data['locale'] == 'fr'
+
+
+@pytest.mark.parametrize(
+    'http_method', ['put', 'post', 'delete', 'options', 'head'])
+def test_get_user_disallowed_methods(client, wiki_user, http_method):
+    """
+    HTTP methods other than GET are not allowed on the api.v1.get_user endpoint.
+    """
+    url = reverse('api.v1.get_user', args=(wiki_user.username,))
+    response = getattr(client, http_method)(url)
+    assert response.status_code == 405
+    assert_no_cache_header(response)
+
+
+@pytest.mark.parametrize('case', ('upper', 'lower'))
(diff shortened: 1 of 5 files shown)


kuma/api/v1/views.py
```diff
@@ -300,3 +301,22 @@ def bc_signal(request):
...
+    Returns a JSON response with a small subset of public information if a
+    user with the given username exists, otherwise returns a status code of
+    404. The case of the username is not important, since the collation of
+    the username column of the user table in MySQL is case-insensitive.
+    """
+    fields = ('username', 'title', 'fullname', 'timezone', 'locale')
+    try:
+        user = User.objects.only(*fields).get(username=username)
+    except User.DoesNotExist:
+        raise Http404('No user exists with the username "{}".'.format(username))
```
peterbe:
```suggestion
        raise Http404(f'No user exists with the username "{username}".')
```

kuma/api/v1/views.py
```diff
@@ -300,3 +301,22 @@ def bc_signal(request):
...
+@never_cache
+@require_GET
+def get_user(request, username):
+    """
+    Returns a JSON response with a small subset of public information if a
+    user with the given username exists, otherwise returns a status code of
+    404. The case of the username is not important, since the collation of
+    the username column of the user table in MySQL is case-insensitive.
+    """
+    fields = ('username', 'title', 'fullname', 'timezone', 'locale')
```
peterbe: Why these? Why not `organization` and `location` as well?

kuma/api/v1/tests/test_views.py
```diff
@@ -330,3 +330,42 @@ def test_search_basic(self):
...
+    """
+    HTTP methods other than GET are not allowed on the api.v1.get_user endpoint.
+    """
+    url = reverse('api.v1.get_user', args=(wiki_user.username,))
+    response = getattr(client, http_method)(url)
+    assert response.status_code == 405
+    assert_no_cache_header(response)
+
+
+@pytest.mark.parametrize('case', ('upper', 'lower'))
```
peterbe: I'm not sure this is worth it. There is no way we can make it NOT case-insensitive (double-negative) so by doing this you're basically unit testing the MySQL collation. This testing isn't testing your code, but testing implementation details in the system.

kuma/api/v1/tests/test_views.py
```diff
@@ -330,3 +330,42 @@ def test_search_basic(self):
...
+    assert response['content-type'] == 'application/json'
+    data = response.json()
+    assert data['username'] == wiki_user.username
+    assert data['avatar_url'] == wiki_user_github_account.get_avatar_url()
+    for field in ('title', 'fullname', 'timezone', 'locale'):
+        assert field in data
+    assert_no_cache_header(response)
+
+
+def test_get_nonexisting_user(client, wiki_user, wiki_user_github_account):
```
peterbe:
```suggestion
def test_get_nonexisting_user(client):
```
??

kuma/api/v1/tests/test_views.py
```diff
@@ -330,3 +330,42 @@ def test_search_basic(self):
...
+    username = getattr(str, case)(wiki_user.username)
+    url = reverse('api.v1.get_user', args=(username,))
+    response = client.get(url)
+    assert response.status_code == 200
+    assert response['content-type'] == 'application/json'
+    data = response.json()
+    assert data['username'] == wiki_user.username
+    assert data['avatar_url'] == wiki_user_github_account.get_avatar_url()
+    for field in ('title', 'fullname', 'timezone', 'locale'):
+        assert field in data
```
peterbe:
```suggestion
        assert data[field] == getattr(wiki_user, field)
```
??

kuma/users/tests/test_helpers.py
```diff
@@ -19,3 +21,18 @@ def test_public_email():
...
+    Test is_username_taken for an existing user, and also that the username is
+    case insensitive.
+    """
+    assert is_username_taken(getattr(str, case)(wiki_user.username))
+
+
+@pytest.mark.parametrize('username', ('nonexistent', None))
+def test_is_username_taken_nonexisting_user(wiki_user, username):
+    """Test is_username_taken for an non-existing user."""
+    assert not is_username_taken(username)
```
peterbe: These two tests above are not wrong but they're pretty verbose too. It could be simplified to just:
```python
def test_is_username_taken(wiki_user):
    assert is_username_taken(wiki_user.username)
    assert is_username_taken(wiki_user.username.upper())  
    assert not is_username_taken('nonexistent')  
```

kuma/api/v1/tests/test_views.py
```diff
@@ -330,3 +330,42 @@ def test_search_basic(self):
         assert data['documents']
         assert data['count'] == 5
         assert data['locale'] == 'fr'
+
+
+@pytest.mark.parametrize(
+    'http_method', ['put', 'post', 'delete', 'options', 'head'])
```
peterbe: This is pretty overkill. I don't see a huge benefit in testing that the `require_GET` decorator works. If it's really important that you can't use anything other than `GET` we should probably write and explain the reasoning of that where we use the `require_GET` decorator in views.py

kuma/api/v1/views.py
```diff
@@ -300,3 +301,22 @@ def bc_signal(request):
         serializer.save()
         return Response(serializer.validated_data, status=status.HTTP_201_CREATED)
     return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@never_cache
+@require_GET
```
peterbe: What is the reasoning here? I'm not really *questioning*, I'm trying to learn. For example, what's the harm in using `HEAD`. Like, I could use `HEAD` to query if a user exists like this:

```
▶ curl -I http://mdn.localhost:8000/api/v1/users/doesthisusernameexist
HTTP/1.0 405 Method Not Allowed
```

kuma/api/v1/views.py
escattone: Oh man, I forgot we're in the wonderful new world of Python 3 and its f-strings! 🎉

kuma/api/v1/views.py
escattone: No particular reason other than I chose a subset that I thought would be essential. I'm happy to add `organization` and `location`  and others as well.

kuma/api/v1/tests/test_views.py
escattone: My thought here is that this is testing the "contract" of the endpoint, and case insensitivity is part of that contract, no matter how it's implemented. I would say it's important that the tests detect a change in that "contract", so for example, if we inadvertently modified the MySQL collation or we moved to PostgreSQL and didn't account for the collations.

kuma/api/v1/tests/test_views.py
escattone: Oops! Good catch, thanks!

kuma/api/v1/views.py
escattone: Good point, I think we should allow `HEAD` as well.

kuma/api/v1/tests/test_views.py
escattone: My thinking here is again to test the "contract" of the endpoint, so that if that contract changes, the tests will fail. I think `HEAD` makes sense here though, so I'm going to allow that.

kuma/users/tests/test_helpers.py
escattone: Yes, that's fine too. In fact, I originally wrote the test that way. Either way is fine to me, but I'm happy to move back to this since the function being tested is so simple.

escattone:
> One slight thing that worries me is that this PR introduces two new implementations (one public (the API) and one private (the jinja helper)) but without context.
> Perhaps I've been living under a rock but would it not make sense to introduce these things when we know they're going to be actually used/needed by the user-facing stuff.

This work arose directly from @schalkneethling's needs for the new user signup flow (see https://github.com/mdn/kuma/issues/5911 for context).

kuma/api/v1/tests/test_views.py
escattone: The reason I didn't do it that way from the beginning was that all of those fields are blank for the `wiki_user`, but even then it's better this way, thanks.

PR #6113, opened 2019-11-14
black

peterbe:
Early days. But a great start!

This'll be hard to land. Chicken and egg. If we land this now, it'll break every single TravisCI build until all files have been converted. I'd rather try to get full consensus from the team that this is what we want to do and then I can take care of gradually fixing old files. 

I need to figure out how to tell Travis that it's OK with this part of the matrix failing. Also, I'd rather NOT let this linger forever but instead aggressively get it over and done with so if that means some red CI builds (just on the python linting part!) for a couple of days I'm OK with that. 

I have a tool that I can use to analyze which .py files are mentioned in any of the open PRs. I'll be slowly by slowly using that to upgrade files in batches. My intention is piece-meal this until all pages are formatted with `black`. 

- [ ] Agree on the config (line-length and stuff)
- [ ] Agree on a "level of aggression" to land this
- [ ] Update developer documentation
- [x] Incorporate into `make pythonlint`
- [ ] Actually reformat all existing files


Gregoor:
I like black's defaults and I'm definitely in favor of using it!

Thanks for getting the ball rolling Peter!

escattone:
👍 I'm definitely in favor of using `black` as well. I've mentioned this in conversation many times, but just wanted to record that here as well.

escattone:
Not sure if you'd still like to use this PR? I imagine not. Perhaps we should close it?

callahad:
I feel like we should just have a flag day and run *everything* through
Black, add it to our CI, and be done with it. :)

peterbe:
> Not sure if you'd still like to use this PR? I imagine not. Perhaps we should close it?

Nah. I just need to refresh it a bit into the new Era of poetry and stuff. 

Another thing is that I think I want to enable `black` checking in CI but only make it WARN. For now. Then, when it has changed enough fiiles, we can switch from WARN to exit non-0. 

I have some tools available that makes it easy to incrementally update files. I'll get around to it.

peterbe:
Closing in favor of https://github.com/mdn/kuma/pull/6436

PR #6134, opened 2019-11-18
Fixes #6067, remove contributor messaging post signup

### LINKED ISSUE #6067: Remove contributor notification post account creation
Once a user has successfully signed up, we show a banner similar to the one below either just below the page header, or generally at the top of the page.

Screenshot 2019-10-30 at 16 06 16 16 41 01

Because of the changes to account roles, these no longer makes sense and should be removed.

schalkneethling:
This removes the messaging around contributions post-sign-up.

## Before

67869535-3ce5a700-fb36-11e9-930b-66b82f6fd141

## After

Screenshot 2019-11-18 at 15 19 05

@peterbe r?

### DIFF
--- a/kuma/users/signal_handlers.py
+++ b/kuma/users/signal_handlers.py
@@ -1,14 +1,11 @@
 from allauth.account.signals import email_confirmed, user_signed_up
 from allauth.socialaccount.signals import social_account_removed
-from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
 from django.db.models.signals import post_delete, post_save, pre_delete
 from django.dispatch import receiver
-from django.utils.translation import ugettext_lazy as _
 from waffle import switch_is_active
 
-from kuma.core.urlresolvers import reverse
 from kuma.payments.utils import cancel_stripe_customer_subscription
 from kuma.wiki.jobs import DocumentContributorsJob
 
--- a/kuma/users/signal_handlers.py
+++ b/kuma/users/signal_handlers.py
@@ -21,10 +21,6 @@ def on_user_signed_up(sender, request, user, **kwargs):
     """
     Signal handler to be called when a given user has signed up.
     """
-    url = reverse('wiki.document', args=['MDN/Getting_started'])
-    msg = _('You have completed the first step of '
-            '<a href="%s">getting started with MDN</a>') % url
-    messages.success(request, msg)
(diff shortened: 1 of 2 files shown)


schalkneethling:
My apologies, I neglected to add the steps to test. Here they are:

1. Click on the `Sign In` link in the header
2. Once redirected to MDN, click on the `change email address` link(button to be honest)
3. In the "other" field, enter an email address that has not been verified
4. Tick the Agree checkbox and click the "Create Profile" button.

This should send you to the view shown in https://github.com/mdn/kuma/pull/6134#issue-342158441

kuma/users/signal_handlers.py
```diff
@@ -21,10 +21,6 @@ def on_user_signed_up(sender, request, user, **kwargs):
     """
     Signal handler to be called when a given user has signed up.
     """
-    url = reverse('wiki.document', args=['MDN/Getting_started'])
-    msg = _('You have completed the first step of '
-            '<a href="%s">getting started with MDN</a>') % url
-    messages.success(request, msg)
```
peterbe: As you might have noticed, you'll have to run `make pythonlint` on this file.

schalkneethling: Yup indeed. Which extension do you use in VSCode to auto PEP8/Flake8 all the Python things?

peterbe: First of all, you need a virtualenv which has our version of `flake8` installed. Soon, you'll also need it so you get our version of `Black`. 
The virtualenv needs to be at least Python 3.7. You can try to run to run `pip install -r requirements/dev.txt` into this virtualenv. 

Once you have this virtualenv activated locally on the command like you should be able to run `make pythonlint` on the command line (outside of docker). 

Next you need to put the path to the `python` and stuff in the `.vscode/settings.json`. Here's mine:
```json
{
    "python.linting.pylintEnabled": false,
    "python.linting.flake8Enabled": true,
    "python.linting.enabled": true,
    "restructuredtext.confPath": "${workspaceFolder}/docs",
    "python.pythonPath": "/Users/peterbe/virtualenvs/kuma/bin/python"
}
```

I have no idea what that `restructuredtext.confPath` thing is. 

Also, in VSCode you need to select this virtualenv. Do you know how to do that? I'm no t actually sure if you need to once you've set up the `python.pythonPath` in the `.vscode/settings.json` file.

PR #6149, opened 2019-11-22
self-destruction with choices

### LINKED ISSUE #5862: User acct deletion: endpoint that allows MDN users to delete acct/archive their contribs
**Related stories:**
Parent: https://github.com/mdn/sprints/issues/1922
Parent: https://github.com/mdn/sprints/issues/1947
Frontend: https://app.zenhub.com/workspaces/mdn-dev-5d6ea9fb3fe11605a63cea86/issues/mdn/kuma/6071

**Acceptance:**
- [ ] Use case 1: Endpoint allows the ability for non-contributor users to delete their account from the Modal 
- [ ] Use case 2: Endpoint allows the ability for contributor users to delete their account from the Modal, BUT retain attributions on contributions with username intact
- [ ] Use case 3: Endpoint allows the ability for contributor users to delete their account from the Modal, BUT retain attributions as 'anonymous'

**Hypothetical Incoming Data Args**: 
- MDN uid, attributions_prefers_anon, action

peterbe:
Fixes #5862

I don't think you need to go through the same trouble I did to manually test this. But here's what I did:

1. First I log in as `peterbe` and click myself to http://wiki.mdn.localhost:8000/en-US/profiles/peterbe/delete
I pick the "Switch all my attributions for page changes to "Anonymous", and delete my account." option. 
Screen Shot 2019-11-22 at 2 11 32 PM

When I'm done, I get signed out. I can sign in again but it's back into the same account but now with 0 revisions. 

2. Make my way back into the user delete page but now select the other option:
Screen Shot 2019-11-22 at 3 05 03 PM

When I'm done, I get signed out.
In a separate browser I viewed my public profile before and after.

Before:
Screen Shot 2019-11-22 at 3 21 18 PM

After:
Screen Shot 2019-11-22 at 3 22 13 PM


peterbe:
I took the liberty to change the page title. Thanks @atopal !
Screen Shot 2019-11-25 at 9 10 34 AM

kuma/users/tests/test_views.py
```diff
@@ -1295,26 +1297,203 @@ def test_account_refresh_token_saved_next_login(self):
...
+    # Pretend the user logged in with GitHub
+    SocialAccount.objects.create(
+        user=wiki_user,
+        provider='github',
+        extra_data=dict(
+            email=wiki_user.email,
+            avatar_url='https://avatars0.githubusercontent.com/yada/yada',
+            html_url="https://github.com/{}".format(wiki_user.username)
+        )
+    )
```
escattone: You can use the `wiki_user_github_account` pytest fixture instead of creating this yourself.

kuma/users/tests/test_views.py
```diff
@@ -1295,26 +1297,203 @@ def test_account_refresh_token_saved_next_login(self):
...
+
+    document = create_document(save=True)
+    revision = create_revision(
+        title='My First And Only Revision',
+        document=document,
+        creator=wiki_user,
+        save=True)
+    assert Revision.objects.filter(creator=wiki_user).exists()
```
escattone: All of this can be avoided by using the `create_revision` pytest fixture.

kuma/users/tests/test_views.py
```diff
@@ -1295,26 +1297,203 @@ def test_account_refresh_token_saved_next_login(self):
...
+    # Pretend the user logged in with GitHub
+    SocialAccount.objects.create(
+        user=wiki_user,
+        provider='github',
+        extra_data=dict(
+            email=wiki_user.email,
+            avatar_url='https://avatars0.githubusercontent.com/yada/yada',
+            html_url="https://github.com/{}".format(wiki_user.username)
+        )
+    )
```
escattone: Same here, you can use the `wiki_user_github_account` pytest fixture instead of creating this yourself.

kuma/users/tests/test_views.py
```diff
@@ -1295,26 +1297,203 @@ def test_account_refresh_token_saved_next_login(self):
...
+        stripe_customer_id='123456',
+    )
+
+    document = create_document(save=True)
+    revision = create_revision(
+        title='My First And Only Revision',
+        document=document,
+        creator=wiki_user,
+        save=True)
+    assert Revision.objects.filter(creator=wiki_user).exists()
```
escattone: Same here, you could use the `create_revision` pytest fixture instead of this chunk of code.

kuma/users/jinja2/users/user_delete.html
```diff
@@ -28,46 +28,21 @@ <h1>{{ title }}</h1>
...
-          (page changes) you've made, and our content license requires that your
-          revisions have an attribution.
-          {% endtrans %}
-        </p>
-        <p>
-          {% trans %}
-          Please file a bug to request closing your account, and ensure one of
-          these options is in the the description field:
+          {% trans count=revisions.count() %}
+          You have {{ count }} wiki revisions.
```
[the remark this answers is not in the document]
peterbe: Both are extremely fast and extremely rare. I'll opt for clunkiness for now.

kuma/users/views.py
```diff
@@ -450,14 +455,87 @@ def user_edit(request, username):
...
+        user.stripe_customer_id = ''
+        user.save()
+
+        user.socialaccount_set.all().delete()
+
+    def force_logout():
+        request.session.clear()
+
+    def delete_user():
+        user.revisionakismetsubmission_set.all().delete()
```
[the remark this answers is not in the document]
peterbe: Why keep them? The `AkismetSubmission` is ["used as a log of sorts"](https://github.com/mdn/kuma/blob/3ea95bb7d3c53a6d1a08eaf45b5f6b66b80c0e39/kuma/spam/models.py#L22) and if we replace the user with `anon`, what point is it to keep. All it will tell you is that 'ham' or 'spam' was submitted at {date} by some anonymous user. 

I'm just not convinced there's any value in keep the data. Granted, if it is a log, we also don't open the /var/log/someapp/things.log file and rewrite it.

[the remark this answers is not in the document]
escattone: I'm thinking it's like removing an entry from a log when there's still value in the entry. We have the information that an Akismet "ham" or "spam" submission was made at a specific time/date on a specific document revision, and that information still seems valuable even though the sender is now anonymous.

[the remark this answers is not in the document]
peterbe: Nevermind, I changed so that they're moved over to the anonymous user.

kuma/users/tests/test_views.py
peterbe: I don't undertstand that fixture. It's a fixture that returns an instance but doesn't create anything. 
However, `from kuma.wiki.tests import revision as create_revision` gets me a function that creates Revisions.

kuma/users/tests/test_views.py
peterbe: I don't understand that. The `create_revision` fixture is a function that returns an instance. It doesn't create anything.

peterbe:
Wow! This was a lot more work than I had envisioned. Either way, I think I've addressed all your feedback now. 
Basically, it a LOT more careful with all the protected foreign key fields to the user before it does the `user.delete()` thing.

peterbe:
Also, bullet point removed:
Screen Shot 2019-12-11 at 3 58 15 PM

kuma/users/tests/test_views.py
escattone: Oops, sorry, you could just use the `root_doc` fixture, which provides a doc and its revision.

kuma/users/tests/test_views.py
escattone: Oops, sorry, you could just use the `root_doc` fixture, which provides a doc and its revision.

peterbe:
I finally get it! As mentioned in Slack, the fixture `create_revision` was poorly named. I guess it should be called `root_doc_revision`. 
But I didn't want to correct that too. So I just switched to `root_doc` and used `root_doc.revisions.first()` to get my `revision` instance. 
I suppose we can tackle that fixture name at a later day.

PR #6157, opened 2019-11-25
flake8 3.7.9

peterbe:


peterbe:
> ...maybe even add a comment to the flake8 line in `dev.txt` pointing back to that, too?

Mayhaps but I never opened `dev.txt` to edit so I wouldn't have noticed. 

It sucks that we have the version of `flake8` in two places. It's like that because we don't want to have to "install the world" in the `linting` part of the TravisCI matrix just to get the flake8 version. Perhaps we can edit our `dev.txt` and add something like:
```diff
+-r linting.txt

-flake8==.3.7.9
   ...
```
That way, in `.travis.yml` we'd be able to do `pip install -r requirements/linting.txt` and for people who do the full `docker-compose` dance, they get the right version of `flake8` too. 

Yeah, I'll do that once this lands.

callahad:
> Perhaps we can edit our `dev.txt` and add something like:

Yep, definitely the better long-term solution.

peterbe:
Follow-up PR: https://github.com/mdn/kuma/pull/6158

PR #6158, opened 2019-11-25
linting.txt to break away from dev.txt

peterbe:


callahad:
LGTM, pending Travis.

Should probably update #6113 to put Black into `linting.txt` as well.

PR #6189, opened 2019-12-04
Add style for highlighting active sidebar-item dynamically

### LINKED ISSUE #6101: Quick links for active list elements should be bolded in the distinguish way for better UX
**Summary**
When user move around the MDN pages and read e.g. tutorial parts, he needs additional effort to localize his place on the website. Of course there is breadcrumb, but main localization component which is `class="quick-links"` doesn't mark user current location in the distinguish way. There is "next" and "previous" CTA, however user wants jump into right section, which can be rooted not in order and this is why bolding with different style than  `li data-default-state` could be beneficial especially when user is in the middle of long articles. (note! CR -Quick Links could be scrollable)

**Steps To Reproduce (STR)**

 Given I'm on the https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/A_first_splash
 When I would like to use Quick Links on the left hand side
 And move to the another page 
 Then I don't know which links is active one and which can lead to next/previous articles.
 And I need to scroll up to see breadcrumb or CTA which makes additional unnecessary effort
 

**Actual behavior**
Current page is not visible via `class="quick-links"`
obraz

**Expected behavior**
Active page is bolded in the distinguish way with different style than  `li data-default-state` 

**Additional context**
Better UX across the page, together with improvement "Quick Links could be scrollable" for long articles could be really helpfull.
https://github.com/mdn/kuma/issues/6102

Gregoor:
Fixes #6101 

This adds a function to the client bootstrap script which _dynamically_ creates a stylesheet, selecting the sidebar link for the current slug and styles it as recommended in the issue.

Okay, no doubt this is a little hacky, BUT people didn't seem excited when I asked how they'd feel about reviewing a new KS macro and arguably this is the least amount of code that can accomplish what's wanted.
I also don't feel too bad about it because while not-best-practice, I don't think this is the hack on which we'll build kingdoms, i.e. it's pretty standalone.

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -9,6 +9,21 @@ import { AppErrorBoundary } from './error-boundaries.jsx';
 import GAProvider from './ga-provider.jsx';
 import { localize } from './l10n.js';
 
+function addActiveSidebarItemStyle() {
+    const selector = `.sidebar a[href="${location.pathname}"]`;
+    const style = `
+        ${selector} { font-style: italic; color: black; }
+        ${selector}:hover { text-decoration: none }
+    `;
+    const node = document.createElement('style');
+    node.innerHTML = style;
+    if (document.body) {
+        document.body.appendChild(node);
+    }
+}
+
+addActiveSidebarItemStyle();
+
 let container = document.getElementById('react-container');
 if (container) {
     let componentName = container.dataset.componentName;


Gregoor:
Woops meant for this to be a draft, main Todo is for me to check more different page types, especially ones that already have an active sidebar item style

peterbe:
Let's be a lot more careful with this because it's pretty scary to me. You'd potentially triggering a complete repaint when adding a `style` to the DOM. 

Another thing is that we're now splitting our CSS between some inline stuff and some SASS files. 

Another, simpler option would be to do something like this:
```javascript
[...document.querySelectorAll('#quicklinks a[href]')].filter(a => a.href.patname === location.pathname).forEach(a => a.classList.add('current'))
```
then, you can just put a 1 liner in the main SASS file like:
```css
#quicklinks a.current {
    font-weight: bold;
}
```
Whichever we chose we ought to add it carefully and test that it's not a performance hit.

atopal:
I understand there was a mixup when this was pulled into the sprint, but I think it's low priority, and not worth spending more time on since we'll be redoing side bars soon.

peterbe:
Then, considering that this is already [nearly solved in stumptown-renderer](https://github.com/mdn/stumptown-renderer/pull/283) I think it examplifies the importance of putting the energy in the right sustainable places.

peterbe:
By the way, I know we're close to a solution possibly with that hack `[...document.querySelectorAll('#quicklinks a[href]')].filter(a => a.href.patname === location.pathname).forEach(a => a.classList.add('current'))` but it's no guarantee that it'll work awesomely for every type of side bar and there's a web performance risk of a post-render re-rendering by doing it in clien-side which I'd rather not take lightly. In `stumptown-renderer` it gets taken care of on the server so it's so fast that it doesn't even need to be web perf empirically measured to justify inclusion.

Gregoor:
I'm happy with closing it and handling in stumptown.

What I don't see yet is how adding a class vs a style would make a difference, perf-wise. I'd wager adding a style is faster. The browser does have to do extra work of parsing and integrating a new rule, but there's less JS looping and GCing. And the resulting repaint should be the same.
For either, if we do want to reconsider this pre-yari, we could use the good-ol-idle-callback.

Anyway, I'm stipulating when I should be providing data for an issue we're addressing more fundamenta-yaril-i :)

PR #6194, opened 2019-12-04
scoop up the right pre elements' text content

### LINKED ISSUE #6161: Discrepancy between w.d.m.o and d.m.o for embedded live examples with hidden fragments when opening in CodePen / JSFiddle
**Summary**
When a live sample has hidden parts (cf. the first live sample of https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout ), clicking "Open in CodePen" (or "Open in JSFiddle") does not produce the same result whether we click from developer.mozilla.org ([this page](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout)) or from wiki.developer.mozilla.org ([this page](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout)).

**Steps To Reproduce (STR)**

1. Open https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout
2. Click on the first "Open in CodePen" button

**Actual behavior**

The embedded pen (in CodePen) does not take into account the `<pre>`s that had `class="hidden"`.
Result from opening from d.m.o on CodePen: https://codepen.io/SphinxKnight/pen/jOOjwLG

**Expected behavior**

* The result in CodePen / JSFiddle should be the same than the one on the MDN page
* Result from opening from w.d.m.o on CodePen: https://codepen.io/SphinxKnight/pen/gOONRGM

* There should not be any discrepancy for live samples between d.m.o and w.d.m.o

**Additional context**
* Reported earlier through https://github.com/mdn/sprints/issues/2376
* Macro: https://github.com/mdn/kumascript/blob/master/macros/EmbedLiveSample.ejs
* Kuma: https://github.com/mdn/kuma/blob/master/kuma/javascript/src/live-examples.js

peterbe:
Fixes #6161

Jump straight to [this comment](https://github.com/mdn/kuma/issues/6161#issuecomment-561805170) to see what the problem is. The [next comment](https://github.com/mdn/kuma/issues/6161#issuecomment-561805517) demonstrates the solution.

### DIFF
--- a/kuma/javascript/src/live-examples.js
+++ b/kuma/javascript/src/live-examples.js
@@ -48,20 +48,14 @@ export function addLiveExampleButtons(rootElement) {
         // direct siblings and also descendants of direct siblings
         // (because sometimes some of the source code is tucked inside
         // a hidden div).
-        let html = rootElement.querySelector(
-            `#${sectid} ~ pre[class*=html], #${sectid} ~ * pre[class*=html]`
-        );
-        let css = rootElement.querySelector(
-            `#${sectid} ~ pre[class*=css], #${sectid} ~ * pre[class*=css]`
-        );
-        let js = rootElement.querySelector(
-            `#${sectid} ~ pre[class*=js], #${sectid} ~ * pre[class*=js]`
-        );
+        let html = section.querySelectorAll('pre[class*=html]');
+        let css = section.querySelectorAll('pre[class*=css]');
+        let js = section.querySelectorAll('pre[class*=js]');
 
         // Now get the source code out of those pre elements
-        let htmlCode = html ? html.textContent : '';
-        let jsCode = js ? js.textContent : '';
-        let cssCode = css ? css.textContent : '';
+        let htmlCode = [...html].map(p => p.textContent).join('');
+        let jsCode = [...js].map(p => p.textContent).join('');
+        let cssCode = [...css].map(p => p.textContent).join('');
 
         // If we found any source code, then add buttons
         if (htmlCode || cssCode || jsCode) {


kuma/javascript/src/live-examples.js
```diff
@@ -48,20 +48,14 @@ export function addLiveExampleButtons(rootElement) {
...
+        let css = section.querySelectorAll('pre[class*=css]');
+        let js = section.querySelectorAll('pre[class*=js]');
 
         // Now get the source code out of those pre elements
-        let htmlCode = html ? html.textContent : '';
-        let jsCode = js ? js.textContent : '';
-        let cssCode = css ? css.textContent : '';
+        let htmlCode = [...html].map(p => p.textContent).join('');
+        let jsCode = [...js].map(p => p.textContent).join('');
+        let cssCode = [...css].map(p => p.textContent).join('');
```
schalkneethling: This is so clean and nice to read, my only concern is IE :( Perhaps doing it more IE friendly would be the better option, thanks @peterbe

peterbe: I don't get it. Why wouldn't this work in IE?

schalkneethling: The spread operator is not supported in IE - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility

and neither is arrow functions - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility

Perhaps I am not thinking about it correctly?

peterbe: Hehe. You forgot that this is converted to ES5  by Babel :)

schalkneethling: Doh! For some reason, I had it in my head that this was part of the old JS :p

PR #6207, opened 2019-12-06
Move developer_needs to waffle Switch

### LINKED ISSUE #6202: Move developer_needs to waffle Switch
Spawned from: https://github.com/mdn/kuma/pull/6108#issuecomment-561241527

We're currently doing this:
```
  {% if waffle.flag('developer_needs') %}
    {% stylesheet 'banners' %}
  {% endif %}
```
...to determine loading the banners CSS. 

It doesn't need to be a waffle FLAG and it can't be either because a Flag doesn't work for CSS. It only works for the `/api/v1/whoami` which triggers the loading of `banners.jsx` in runtime. 

It should become this:
```diff
- {% if waffle.flag('developer_needs') %}
+ {% if waffle.switch('developer_needs') %}
    {% stylesheet 'banners' %}
  {% endif %}
```

Also, *today* the only factor that determines if you need the banners CSS is `developer_needs` but it can become more than that. So it might one day evolve to:

```
 {% if waffle.switch('developer_needs') or waffle.switch('other_thing') %}
    {% stylesheet 'banners' %}
  {% endif %}
```
...but that's for another day. 

Making this switch should do two other things:

1. The JavaScript code (that runs based on `GET /api/v1/whoami`) should be looking at the waffle Switch. 
2. We should migrate our existing waffle Flag to waffle Switch.

peterbe:
Fixes #6202

**Before:**
Screen Shot 2019-12-06 at 11 16 51 AM

**After:**
Screen Shot 2019-12-06 at 11 17 21 AM
and
Screen Shot 2019-12-06 at 11 19 11 AM


peterbe:
For what it's worth; this is pretty similar to: https://github.com/mdn/kuma/pull/6204/files

peterbe:
> be nice to convert more of the Waffle flags that are only used as on/off switches to Waffle switches

We're currently doing a MySQL look for every flag and every switch, right?
Since Flags will still be around, we *could* cache it all in Redis down to just 1 key, per `request.user.id`. 
But that assumes none of the Flags use a `random()` I guess.

PR #6213, opened 2019-12-06
re-use existing user accounts on signup

### LINKED ISSUE #6182: Implement automatic merging of Persona/GitHub/Google accounts on signup
### Summary
When creating an MDN account via GitHub/Google social login, if one of the verified GitHub/Google email addresses matches an existing Persona/GitHub/Google social account associated with an existing user, automatically recover that user instead of creating a new one, skipping the usual `/<locale>/users/account/signup` page step (where the user selects a username and email address).

See https://github.com/mdn/kuma/issues/6037 for context.

### Acceptance Criteria
- [ ] When signing-up for an MDN account via GitHub/Google social login, a user with an existing Persona/GitHub/Google account (i.e., an existing MDN user account with an associated Persona/GitHub/Google social account) that matches one of the verified GitHub/Google email addresses is automatically recovered without having to go through the `/<locale>/users/account/signup` page step in order to select a username and email address.
- [ ] After recovery, if the user has any associated Personal social accounts, they are deleted, so that only any associated GitHub/Google social accounts remain.

escattone:
Fixes #6182 

It was a long and winding road to this PR. The `django-allauth` package doesn't make it easy to do this kind of thing, with their poor/sparse documentation and source code that's very difficult to read/understand (the OAuth2-based sign-up flow is spread across multiple classes, sub-classes, adapters, and utility functions). With that said however, I'm really happy with the final results.

This PR handles the automatic detection and reuse of an existing user when performing a social-provider signup, for example via GitHub or Google. The detection of an existing user is made via a match of one of the existing user's **verified** email addresses with one of the **verified** email addresses harvested from the social provider (i.e. GitHub or Google). If a matching existing user is found, the GitHub/Google social account will automatically be associated with that user, and that user will be logged-in. The account signup step (`/<locale>/users/account/signup`), where the GitHub/Google social account is associated with a new user created with the username/email-address from the signup form, is skipped.

For testing within your local environment, you might want to try what I did:
- Login to the admin console using `test-user`/`test-password`.
- Go to the `Social accounts` section and delete any existing GitHub social account for your own user (my username is `escattone`).
- Let's also manually associate a Persona social account with your user. You'll have to do this from the command line. First do a `docker-compose exec web ./manage.py shell_plus`, and then:
   ```python
   user = User.objects.get(username='escattone')
   SocialAccount.objects.create(user=user, provider='persona')
   ```
- Log out of the admin console.
- Go to http://localhost.org:8000/users/github/login (or just go to http://localhost.org:8000 and click the `Sign in` button)
- Assuming you're already logged-in to GitHub, you should be automatically logged-in, since your existing user will be found, a new GitHub social account will be associated with it, and the `/en-US/users/account/signup` step will be skipped.
- Log out.
- Log back into the admin console using `test-user`/`test-password`.
- Go to the `Social accounts` section and verify that your Persona social account has been deleted.

Also the tests for the new code are thorough and run through the exact same flow that `django-allauth` runs after the OAuth2 dance has completed (the `complete_social_login()` call), so they should provide confidence as well.

### DIFF
--- a/kuma/users/adapters.py
+++ b/kuma/users/adapters.py
@@ -259,9 +259,7 @@ class KumaSocialAccountAdapter(DefaultSocialAccountAdapter):
         """
         We allow "auto-signup" (basically skipping the sign-up form page) only
         if there is an existing user that we can re-use instead of creating
-        a new one. An existing user is found if at least one of the user's
-        verified email addresses matches one of the verified email addresses
-        within the social login object.
+        a new one.
         """
         return bool(get_existing_user(sociallogin))
 
--- a/kuma/users/adapters.py
+++ b/kuma/users/adapters.py
@@ -1,10 +1,13 @@
 from allauth.account.adapter import DefaultAccountAdapter, get_adapter
+from allauth.account.models import EmailAddress
+from allauth.account.utils import cleanup_email_addresses
 from allauth.exceptions import ImmediateHttpResponse
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 from allauth.socialaccount.models import SocialLogin
 from django import forms
 from django.contrib import messages
 from django.contrib.auth import get_user_model
+from django.db.models import Q
 from django.shortcuts import redirect, render
 from django.utils.cache import add_never_cache_headers
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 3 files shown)


kuma/users/providers/github/provider.py
```diff
@@ -29,17 +29,17 @@ class KumaGitHubProvider(GitHubProvider):
...
-            # let's ignore all email address that have not been verified at
-            # GitHub's side
-            if (not email_address.get('verified', False) or
-                    not email_address.get('email', '').strip()):
-                continue
-            email_addresses.append(EmailAddress(email=email_address['email'],
-                                                verified=True,
-                                                primary=email_address['primary']))
-        return email_addresses
+        result = []
```
escattone: This rewrite wasn't necessary, but I hated the old code so much that I decided to clean it up.

escattone:
Codecov is way off in its analysis of `kuma/users/tests/test_adapters.py`. Something is really off there.

escattone:
@atopal I thought I'd ping you just so you had an early chance to read my initial comment and make sure this matches your expectations at the product level.

atopal:
Thanks Ryan, yes, exactly like this. Sorry for the late response, catching up after some travel.

kuma/users/adapters.py
```diff
@@ -207,17 +210,91 @@ def get_connect_redirect_url(self, request, socialaccount):
...
+    if emails:
+        # Users can have multiple associated EmailAddress objects, so
+        # let's use "distinct()" to remove any duplicate users.
+        users = list(get_user_model().objects
+                                     .filter(emails,
+                                             emailaddress__verified=True)
+                                     .distinct())
+        # For now, we're only going to return a user if there's only one.
+        if len(users) == 1:
+            return users[0]
```
Gregoor: This does make it impossible for users to choose which email address of their social_account they want to use if they already have an account with any of their email addresses, right? So we are sacrificing some freedom for some sanity? Which might be the right trade-off to make, just want to make sure I'm being conscious of it.

If I were to dream of the optimal UX, it'd probably be something like:
- if you only have one verified email and there exists only one account for it, do what we're doing here
- if you have multiple verified emails, pick one
  - if there's only one account account for it, do what we're doing here
  - if there's multiple, pick one

Is that really optimal? I don't know, it clutters login/signup and puts a lot more work in our lap. I think this trade-off might be good enough for now, until people start complaing. Just wanted to think this through for a second 🤔

Gregoor:
It reads great so far, Ryan. Thanks for spending so much brain juice on it, I wanna do it justice, so I'm gonna spend the after-noon running it with various scenarios. Your code is super readable to me, but as you are well-aware, there's mountains of allauth underneeth that I need to consider.

kuma/users/adapters.py
escattone: Great points and questions @Gregoor.

> This does make it impossible for users to choose which email address of their social_account they want to use if they already have an account with any of their email addresses, right? So we are sacrificing some freedom for some sanity?

Yes, you're right on. I had originally gone down the path of assuming that even for the case when we find an existing user, we should still go to the signup page so that the user has the chance to change her/his username/email, but that's more challenging to implement (for various reasons) and it wasn't what @atopal wanted. However, even given the approach of this PR, the user always has the option of editing his/her profile to create/select a new primary email address.

escattone: @Gregoor I wasn't sure what to do for the case when we find multiple user accounts that are associated with one or more verified social-provider emails, so I decided to not pick one at all. The downside, as you see, is that means another, brand-new user account is created for the user. Perhaps that's the wrong decision, and that instead we should allow the user to select one (and perhaps optionally delete the others?) via a separate page? What do you think @atopal?

atopal: @escattone that's a legacy issue, right? From here on, if you authenticate with emails we already have an account for, we'll add that to the existing account. Is that correct? I'm not too worried about the case with multiple accounts with the same email. We can handle those cases manually at first and then automate if necessary. I just wonder what account we log you in with when you log in (just out of curiosity).

escattone: @atopal Thanks for taking some time to comment, as I know you're busy with your trip! Yes, it's a legacy issue, and from here on we'll add to an existing account. Sounds good, i think you're right that this is a rare case, if it exists at all, and that we can manually handle it if an issue arises.

escattone:
@Gregoor Sorry that I didn't get to your comment earlier!

> I had another idea of how to express it, but I'm not convinced that it's better than what you have. Here it is anyways:
> 
> ```python
> @pytest.mark.parametrize('provider', ('github', 'google'))
> @pytest.mark.parametrize('invalid_fields', (
>   {},
>   {'social_email_address': 'user@nomatch.com'},
>   {'user_email_verified': False},
>   {'social_email_verified': False},
> ))
> def test_user_reuse_on_social_login(wiki_user, client, rf, invalid_fields, provider):
>   should_match = not invalid_fields
>   field_defaults = {'social_email_address': wiki_user.email, 'user_email_verified': True, 'social_email_verified': True}
>   # I was a little lazy/ignorant here
>   social_email_address, user_email_verified, social_email_verified = make_unpackable({**field_defaults, **invalid_fields})
> ```
> 
> Now I fear that I've hid what makes this most likely worse, by making up the function `make_unpackable` (I'm so used to JS object destructuring, not sure what the idiomatic python equivalent is).

Thanks for really digging into the details of this and suggesting an alternative! I like your idea! I really like the de-structuring options in JS, and I wish we had something like that in Python for dictionaries. Still, it would be easy, although not quite as elegant as JS, to do what you've suggested, since in Python 3 the order of the dict is guaranteed to be the same as the insertion order, which you've determined by using `**field_defaults` as the first unpacking for the `{**field_defaults, **invalid_fields}` dict comprehension, and since that matches the order of the target variables `social_email_address, user_email_verified, social_email_verified = ...`. So in the end, there's no need for a `make_unpackable` function, and you could just do:
```python
social_email_address, user_email_verified, social_email_verified = {**field_defaults, **invalid_fields}.values()
```
which I think is a really elegant way to set the values. If we decided to go with this approach, it'd probably be best to add `ids` like this:
```python
@pytest.mark.parametrize('invalid_fields', (
  {},
  {'social_email_address': 'user@nomatch.com'},
  {'user_email_verified': False},
  {'social_email_verified': False}),
  ids=('match', 'nomatch_address', 'nomatch_user_unverified', 'nomatch_social_unverified')
)
```
since the `invalid_fields` values as strings would make less intelligible test case id's.

PR #6219, opened 2019-12-07
class="new" is inserted on links that could work

### LINKED ISSUE #6217: class="new" is inserted on links that could work
See https://github.com/mdn/kuma/issues/6163#issuecomment-561802945

The document has this raw HTML:
```
<a href="/en-US/docs/Web/Houdini/">Houdini</a>
```
but when you save this, the post-process rendering turns that into:
```
<a class="new" href="/en-US/docs/Web/Houdini/">Houdini</a>
```
But the only reason it claims it doesn't exist is because someone accidentally put in an extra trailing `/`. If you'd just remove that, it should work fine.

peterbe:
Fixes #6217

To test; before you check out my branch, take a perfectly fine slug URL and make somet HTML like this:
```html
<a href="/en-US/docs/Web/Learn/">Learn!</a>
```
After edit and publish you'll notice that the rendered HTML becomes
```html
<a class="new" rel="nofollow" href="/en-US/docs/Web/Learn/">Learn!</a>
```
Now check out my branch and edit again.

### DIFF
--- a/kuma/wiki/tests/test_content.py
+++ b/kuma/wiki/tests/test_content.py
@@ -893,7 +893,6 @@ def test_dont_annotate_links_with_trailing_slashes(db, root_doc):
         </ul>
     """
     actual_raw = parse(html).annotateLinks(base_url=AL_BASE_URL).serialize()
-    expected_raw = 'xxx'
     expected_raw = """
         <ul>
         <li><a href="/en-US/docs/Root/">Root</a></li>
--- a/kuma/wiki/tests/test_content.py
+++ b/kuma/wiki/tests/test_content.py
@@ -879,6 +879,31 @@ def test_annotate_links_nonexisting_doc(db, anchor, full_url, has_class):
     assert normalize_html(actual_raw) == normalize_html(expected_raw)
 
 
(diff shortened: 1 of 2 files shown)


peterbe:
I was thinking of doing something fancier where I'd fix/rewrite the `href` so it becomes
```html
<a href="/en-US/docs/Web/Learn">Learn!</a>
```
but it actually got really convoluted and complex and considering that the links now work perfectly fine I'm satisfied with not bother any further/deeper.

kuma/wiki/tests/test_content.py
```diff
@@ -879,6 +879,31 @@ def test_annotate_links_nonexisting_doc(db, anchor, full_url, has_class):
...
+    # Sanity check our fixture
+    assert Document.objects.filter(locale='en-US', slug='Root').exists()
+    html = """
+        <ul>
+        <li><a href="/en-US/docs/Root/">Root</a></li>
+        <li><a href="/en-US/docs/Hopeless/">Hopeless</a></li>
+        </ul>
+    """
+    actual_raw = parse(html).annotateLinks(base_url=AL_BASE_URL).serialize()
+    expected_raw = 'xxx'
```
escattone: Nit: This line can be deleted.

PR #6220, opened 2019-12-07
disable a.new links in read-only mode

### LINKED ISSUE #6163: Add a way to indicate links are "to do" items
## Current situation
Currently, when an article needs to reference a page that has not yet been written, we still add the `<a>` tags, linking to the intended article's URL. Sometimes these get written right away, sometimes years pass before someone has the opportunity to write that content.

This works fairly well, and gives contributors or potential contributors quick access to the editor through these links.

## The problem
On the down side, some users are confused by this, thinking there's an error of some kind, so we get bug reports about them from time to time.

For multiple reasons, it would be preferable to have a way to label a selection of content as requiring future work to be done. This would let us track and find these to-do items without confusing users.

## Proposed method to mark to-do items
My thinking is that we could use `<mark>` to label content that needs work, with a style that indicates that it's a "to-do" marker:

```
<mark class="mark-todo"
       title="We need to write this landing page">
   <a href="/en-US/docs/Web/API/Some_new_API">
      Some new API
   </a>
</mark>
```
These would be invisible to everyday users (perhaps), but would be highlighted in some way, with a tooltip taken from the `title`, which would be provided while writing the content, to explain the work needing to be done.

## Finding to-do items
Once this is in place, it would be fairly easy to locate and monitor the state of tasks waiting to be done that are labeled in this way. Simply search for uses of the `<mark>` element, and for each of them, you can get the details from the `title` attribute. With that data, you can put it into a spreadsheet or, preferably, a dashboard presenting the most useful information and statistics, with links to create these pages.

peterbe:
Fixes #6163

CC @chrisdavidmills 

We already had a KS solution for links made via [`jsapixref`](https://github.com/mdn/kumascript/blob/9a44f78ea18a807681d1aeb37b7eb42ec3c8dd3a/macros/jsapixref.ejs#L15) that puts a title into these links. It makes these:
Screen Shot 2019-12-06 at 9 05 05 PM

I suppose that's only for JS docs. (Yay KumaScript!)
So my PR is gentle and only adds the tooltip for those that don't already have a tooltip. 

Also, the screenshot won't show it but visually there's no difference but now when you hover over it it doesn't make a `pointer` cursor.

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -114,6 +114,28 @@ export default function Article({ document }: DocumentProps) {
         }
     }, [document, userData, ga]);
 
+    /** Any link inside the article that matches `a.new` means it's a
+     * wiki page that hasn't yet been created. Clicking on it will
+     * 404, if you're on the read-only site. If you're viewing HTML
+     * like that on the Wiki and click it, it will take you to the
+     * form to *create* the page. Disable all of that here in the read-only
+     * site.
+     */
+    useEffect(() => {
+        let rootElement = article.current;
+        if (rootElement) {
+            for (let link of rootElement.querySelectorAll('a.new')) {
+                // Makes it not be clickable and no "pointer" cursor when
+                // hovering. Better than clicking on it and being
+                // disappointed.
+                link.removeAttribute('href');
+                if (!link.title) {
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Cool, let's try this out.

PR #6226, opened 2019-12-10
SSR Sentry error reporting

### LINKED ISSUE #5981: SSR Sentry error reporting
At the time of writing, 7 times this has happened:
```
HTTPError: 500 Server Error: Internal Server Error for url: http://ssr/ssr/SPA
  File "django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 544, in wrapper
  File "django/views/decorators/cache.py", line 57, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "ratelimit/decorators.py", line 30, in _wrapped
    return fn(*args, **kw)
  File "kuma/search/views.py", line 34, in search
    return render(request, 'search/react.html', context)
  File "django/shortcuts.py", line 30, in render
    content = loader.render_to_string(template_name, context, request, using=using)
  File "django/template/loader.py", line 68, in render_to_string
    return template.render(context, request)
  File "django_jinja/backend.py", line 106, in render
    return mark_safe(self.template.render(context))
  File "newrelic/api/function_trace.py", line 121, in dynamic_wrapper
  File "jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/app/kuma/search/jinja2/search/react.html", line 3, in top-level template code
    {% set query = request.GET.get('q') %}
  File "/app/jinja2/react_base.html", line 103, in top-level template code
    {% block document_head %}{% endblock %}
  File "/app/kuma/search/jinja2/search/react.html", line 20, in block "document_head"
    results)|safe }}
  File "kuma/wiki/templatetags/ssr.py", line 50, in render_react
    return server_side_render(component_name, data)
  File "kuma/wiki/templatetags/ssr.py", line 133, in server_side_render
    response.raise_for_status()
  File "requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
```
https://sentry.prod.mozaws.net/operations/mdn-prod/issues/6448273/?query=is:unresolved

It's "fine". The SSR service is bound to be imperfect. What's worrying is that there are no Node errors in Sentry. It's just too unlikely that there have been no exceptions in the SSR Node service.

peterbe:
Fixes #5981

This was a lot tricker than I first envision. Sorry. 
Also, all of this will go away with Yari. For now, with this fix, it'll tie us over quite a bit in the many months of Kuma still doing the SSR. 

Ultimately, the problem is that if something went horribly wrong inside `ssr-server.jsx` (the express server) it would fail but never report anything to Sentry. All you'd get is a Sentry error in Python for the possible `HTTPError`. E.g.https://github.com/mdn/kuma/issues/5981#issue-506747552

I don't really understand how or why but it seems that if you don't use...
```
try {
  somethignThatCanGoWrong()
} catch (err) {
  Sentry.captureException(err);
}
```
...it wouldn't send to Sentry. I don't see why the [SDK Quickstart for Express](https://sentry.io/for/express/) doesn't mention this. 
Any idea @Gregoor ?

Now, with this change you'll get errors like this: https://sentry.prod.mozaws.net/operations/mdn-stage/issues/6842634/
And nothing in Sentry for Python, because why would you want 2 errors for one and the same problem. Plus any Python errors sent to Sentry would be rather useless anyway. 

Now, what I hate is how our SSR integration code in Python works is that if there's any operational error talking to the Express server it just swallows it and does a client-side rendering. Blah!

Also, where we're currently doing `request.post(...)` there in `template_tags/ssr.py` we ought to use that `requests_retry_session` trick mentioned in some other PR. That way, if there's an unfortunate temporary network error it at least tries again. 

I'll try to comment and justify some changes within the code below.

### DIFF
--- a/kuma/javascript/ssr-server.js
+++ b/kuma/javascript/ssr-server.js
@@ -15,24 +15,9 @@ if (process.env.NEW_RELIC_LICENSE_KEY && process.env.NEW_RELIC_APP_NAME) {
 
 const express = require('express'); // Express server framework
 const morgan = require('morgan');
-const sentry = require('@sentry/node');
+const Sentry = require('@sentry/node');
 
 const ssr = require('./dist/ssr.js'); // Function to do server-side rendering
-
-// Configure and initialize Sentry if a DSN has been provided.
-if (process.env.SENTRY_DSN) {
-    console.log('Configuring Sentry for the SSR server.');
-    let options = {
-        dsn: process.env.SENTRY_DSN
-    };
-    if (process.env.REVISION_HASH) {
-        options.release = process.env.REVISION_HASH;
-    }
-    if (process.env.SENTRY_ENVIRONMENT) {
-        options.environment = process.env.SENTRY_ENVIRONMENT;
-    }
-    sentry.init(options);
-}
 
 // Configuration
 const PID = process.pid;
@@ -47,6 +32,28 @@ const MAX_BODY_SIZE = 1024 * 1024 * 5; // 5 megabyte max JSON payload
 
 // We're using the Express server framework
 const app = express();
+
+// Configure and initialize Sentry if a DSN has been provided.
+if (process.env.SENTRY_DSN) {
+    console.log('Configuring Sentry for the SSR server.');
+    let options = {
+        dsn: process.env.SENTRY_DSN
+    };
+    if (process.env.REVISION_HASH) {
+        options.release = process.env.REVISION_HASH;
+    }
+    if (process.env.SENTRY_ENVIRONMENT) {
+        options.environment = process.env.SENTRY_ENVIRONMENT;
+    }
+    Sentry.init(options);
+
+    // The request handler must be the first middleware on the app
+    app.use(Sentry.Handlers.requestHandler());
+    // The error handler must be before any other error middleware
+    app.use(Sentry.Handlers.errorHandler());
+} else {
+    console.warn('SENTRY_DSN is not available so sentry is not initialized.');
+}
 
 // Log all requests, so we get timing data for SSR.
 app.use(morgan('tiny'));
@@ -81,7 +88,13 @@ app.get('/readiness/?', (req, res) => {
  * not be parsed, escaped, or displayed as HTML.
  */
 app.post('/ssr/:componentName', (req, res) => {
-    res.json(ssr(req.params.componentName, req.body));
+    try {
+        res.json(ssr(req.params.componentName, req.body));
+    } catch (err) {
+        console.error(`Error running ssr(): ${err.toString()}`);
+        Sentry.captureException(err);
+        res.status(500).json({ error: err.toString() });
+    }
 });
 
 if (require.main === module) {
--- a/kuma/javascript/ssr-server.js
+++ b/kuma/javascript/ssr-server.js
@@ -33,28 +33,6 @@ const MAX_BODY_SIZE = 1024 * 1024 * 5; // 5 megabyte max JSON payload
 // We're using the Express server framework
 const app = express();
 
-// Configure and initialize Sentry if a DSN has been provided.
-if (process.env.SENTRY_DSN) {
-    console.log('Configuring Sentry for the SSR server.');
-    let options = {
-        dsn: process.env.SENTRY_DSN
-    };
-    if (process.env.REVISION_HASH) {
-        options.release = process.env.REVISION_HASH;
-    }
-    if (process.env.SENTRY_ENVIRONMENT) {
-        options.environment = process.env.SENTRY_ENVIRONMENT;
-    }
-    Sentry.init(options);
-
-    // The request handler must be the first middleware on the app
-    app.use(Sentry.Handlers.requestHandler());
-    // The error handler must be before any other error middleware
-    app.use(Sentry.Handlers.errorHandler());
-} else {
-    console.warn('SENTRY_DSN is not available so sentry is not initialized.');
-}
-
 // Log all requests, so we get timing data for SSR.
 app.use(morgan('tiny'));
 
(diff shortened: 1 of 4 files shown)


kuma/javascript/ssr-server.js
```diff
@@ -15,25 +15,10 @@ if (process.env.NEW_RELIC_LICENSE_KEY && process.env.NEW_RELIC_APP_NAME) {
 
 const express = require('express'); // Express server framework
 const morgan = require('morgan');
-const sentry = require('@sentry/node');
+const Sentry = require('@sentry/node');
```
peterbe: All the code examples in the documentation uses this notation for the default export. So I changed to that. Ultimately no functional difference but now our code looks like the docs.

kuma/javascript/ssr-server.js
```diff
@@ -48,6 +33,28 @@ const MAX_BODY_SIZE = 1024 * 1024 * 5; // 5 megabyte max JSON payload
...
+    }
+    if (process.env.SENTRY_ENVIRONMENT) {
+        options.environment = process.env.SENTRY_ENVIRONMENT;
+    }
+    Sentry.init(options);
+
+    // The request handler must be the first middleware on the app
+    app.use(Sentry.Handlers.requestHandler());
+    // The error handler must be before any other error middleware
+    app.use(Sentry.Handlers.errorHandler());
```
peterbe: These 4 lines right here (above) is the only new thing. The rest is that I just moved this stuff to be *AFTER* the `const app = express();` (line 34). 
If you don't do this you don't get information in Sentry about what what posted to SSR.

kuma/javascript/ssr-server.js
```diff
@@ -48,6 +33,28 @@ const MAX_BODY_SIZE = 1024 * 1024 * 5; // 5 megabyte max JSON payload
...
+} else {
+    console.warn('SENTRY_DSN is not available so sentry is not initialized.');
```
peterbe: This was my own personal debugging when I was trying to understand why SENTRY_DSN wasn't enabled. 
After I figured out what was going on I just decided to keep this line because I think it's useful.

peterbe: By the way, to enable Sentry in local development, it's *not* good enough to put it into your `.env`. That'll only help your Python image in docker-compose. It's because the `ssr` service in `docker-compose.yml` doesn't use anything like `dotenv` to read from a `.env` file on disk.

kuma/javascript/ssr-server.js
```diff
@@ -81,7 +88,13 @@ app.get('/readiness/?', (req, res) => {
  * not be parsed, escaped, or displayed as HTML.
  */
 app.post('/ssr/:componentName', (req, res) => {
-    res.json(ssr(req.params.componentName, req.body));
+    try {
+        res.json(ssr(req.params.componentName, req.body));
+    } catch (err) {
+        console.error(`Error running ssr(): ${err.toString()}`);
+        Sentry.captureException(err);
```
peterbe: The core of this whole PR is that I have to manually trigger this or else nothing's sent to Sentry. 
I don't really understand why the `@sentry/node` and `express` configuration can't do this automatically.

kuma/settings/testing.py
```diff
@@ -85,3 +85,6 @@
 
 # Use a dedicated minio bucket for tests
 ATTACHMENTS_AWS_STORAGE_BUCKET_NAME = 'test'
+
+# Never enabled in tests.
+SENTRY_DSN = None
```
peterbe: This was a total timesink. I had put the `mdn-stage` Sentry DSN into my personal `.env` and it broke all sorts of tests when running `pytest`.

kuma/wiki/tests/test_ssr.py
```diff
@@ -109,7 +110,7 @@ def test_failed_server_side_render(mock_get_l10n_data,
     localization_data = {'catalog': {'s': 't'}, 'plural': None}
     mock_get_l10n_data.side_effect = lambda l: localization_data
     mock_dumps.side_effect = sorted_json_dumps
-    url = '{}/{}'.format(settings.SSR_URL, 'page')
+    url = f'{settings.SSR_URL}/page'
```
peterbe: Stylistic and unimportant change. Sorry.

kuma/wiki/templatetags/ssr.py
```diff
@@ -134,11 +134,7 @@ def server_side_render(component_name, data):
         result = response.json()
         return _render(component_name, result['html'], result['script'])
 
-    except requests.exceptions.ConnectionError:
-        print("Connection error contacting SSR server.")
-        print("Falling back to client side rendering.")
-        return client_side_render(component_name, data)
-    except requests.exceptions.ReadTimeout:
-        print("Timeout contacting SSR server.")
+    except requests.exceptions.RequestException as exception:
```
peterbe: A `ConnectionError` and a `ReadTimeout` exception are both subclasses of `RequestException`. So is `HTTPError` which would happen if the network talk to the URL worked but the server responded with a 500 error. 

This will NOT raise a Python exception:
```python
url = "https://httpbin.org/status/500"
r = requests.get(url)
```
But this will:
```python
url = "https://httpbin.org/status/500"
r = requests.get(url)
r.raise_for_status()
```
In fact, you get:
```
...
requests.exceptions.HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: https://httpbin.org/status/500
```

peterbe:
I don't know what else is lurking here. If you get any HTTP related error from the SSR service, it'll just swallow it in `template_tags/ssr.py` and instead return 200 OK with the client-side React code. 

Actually, what a browser user would get is an "empty" server HTML (aka. the SPA skeleton) and then the client-side React would give it a try and then it'll probably cause an ErrorBoundary catch. 

Generally, I don't like any of this. If the SSR service isn't working, we should instead respond to the user that there was a 500 because there's something wicked going on. Instead, we're responding with `200 OK` and then our client-side React code has to deal with the problem. Gross!

Can we just work on Yari instead?

kuma/javascript/ssr-server.js
escattone: As you say in the comment, this must be before any other error middleware, but it also must be **after** any Express controllers you'd like it to monitor/handle. If you move this below the definition of the `app.post` controller below, you'd no longer need to explicitly capture exceptions and send them to Sentry (I tested it locally, and it works). The downside, though, is that it would monitor/handle all of the other controllers as well, unless you define the `app.post` controller first, do this `app.use(Sentry.Handlers.errorHandler());` next, and finally add all of the remaining Express controllers (e.g., the `healthz` and `readiness` endpoints) that we don't really care to have Sentry handle.

peterbe: Resolved by discussion.

kuma/javascript/ssr-server.js
```diff
@@ -84,6 +86,15 @@ app.post('/ssr/:componentName', (req, res) => {
     res.json(ssr(req.params.componentName, req.body));
 });
 
+// Important that this is defined *after* the request handlers have been
+// added otherwise Sentry won't automatically hook in.
+if (process.env.SENTRY_DSN) {
+    // The request handler must be the first middleware on the app
+    app.use(Sentry.Handlers.requestHandler());
```
escattone: Sorry to make you update this again @peterbe! This has to go back up to the top, just after the `Sentry.init(options);`, since it has to be the first middleware for the `app` (before the `morgan` and `express.json` middleware).

kuma/javascript/ssr-server.js
```diff
@@ -84,6 +86,15 @@ app.post('/ssr/:componentName', (req, res) => {
...
+    // The error handler must be before any other error middleware
```
escattone: I'm thinking we should copy the comment from https://docs.sentry.io/platforms/node/express/, so this would become:
```
// The error handler must be before any other error middleware and after all controllers
```

PR #6240, opened 2019-12-13
exclude Persona social accounts when getting avatar URL

### LINKED ISSUE #6235: Viewing profile of user with Persona account generates an ISE
**Summary**
It seems that viewing the profile of a user who has a Persona social account triggers an ISE.

**Steps To Reproduce (STR)**
1. Visit https://developer.allizom.org/en-US/profiles/sheppy
2. It will generate a 500 error
3. Visit https://developer.allizom.org/en-US/profiles/escattone
4. Displays fine

**Actual behavior**
Returns 500 (internal server error)

**Expected behavior**
Success display of user's profile page

**Additional context**
This seems to be related to one or more of the account PR's that were merged prior to deploying to stage.

escattone:
Fixes #6235 

This is blocking pushes.

### DIFF
--- a/kuma/users/templatetags/jinja_helpers.py
+++ b/kuma/users/templatetags/jinja_helpers.py
@@ -18,11 +18,14 @@ from ..models import User
 @library.global_function
 def get_avatar_url(user):
     """
-    Get the user's avatar URL for the first social account provider that has one.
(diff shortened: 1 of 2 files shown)


escattone:
Wow, what's wrong with Codecov? It's totally off!

PR #6242, opened 2019-12-13
Update Installation instructions

mindy:
- Change `localhost:8000` to `localhost.org:8000`
- Update command to add social auth 
- Add Docker instructions

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -181,7 +181,7 @@ A few thousand lines will be printed, like::
 
 Visit the homepage
 ==================
-Open the homepage at http://localhost:8000 . You've installed Kuma!
+Open the homepage at http://localhost.org:8000 . You've installed Kuma!
 
 Create an admin user
 ====================
@@ -204,9 +204,24 @@ management command. Replace ``YOUR_USERNAME`` with your username and
     --password YOUR_PASSWORD
 
 With a password-enabled admin account, you can log into Django admin at
-http://localhost:8000/admin/login/
+http://localhost.org:8000/admin/login
 
 .. _enable-github-auth:
+
+Update the Sites section
+=======================================
+#. After logging in to the Django admin (an alternative is using the login ``test-super`` 
+   with password ``test-password``), scroll down to the Sites section. 
+
+#. Click on "Change". 
+
+#. Click on the entry that says ``localhost:8000``. 
+
+#. Change both the domain and display name from ``localhost:8000`` to ``localhost.org:8000``. 
+
+#. Click "Save".
+
+
 
 Enable GitHub/Google authentication (optional)
 =======================================
@@ -214,8 +229,8 @@ Since Google's OAuth requires us to use a valid top-level-domain, we're going to
 http://localhost.org:8000 as an example for every URL here.
 
 To automate setting Django up for social auth you can run
-``docker-compose exec web ./manage.py configure_social_auth`` and follow its steps (and
-ignore the rest of this section).
+``docker-compose exec web ./manage.py configure_github_social`` and follow its steps (and
+ignore the rest of this section). 
 
 If you want to do it manually, follow these steps:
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -204,18 +204,33 @@ management command. Replace ``YOUR_USERNAME`` with your username and
...
+
 Enable GitHub/Google authentication (optional)
 =======================================
 Since Google's OAuth requires us to use a valid top-level-domain, we're going to use
 http://localhost.org:8000 as an example for every URL here.
 
 To automate setting Django up for social auth you can run
-``docker-compose exec web ./manage.py configure_social_auth`` and follow its steps (and
-ignore the rest of this section).
+``docker-compose exec web ./manage.py configure_github_social`` and follow its steps (and
```
escattone: We don't have a Django management command named `configure_github_social`, so this should remain `configure_social_auth`.

docs/installation.rst
```diff
@@ -204,18 +204,33 @@ management command. Replace ``YOUR_USERNAME`` with your username and
...
+#. After logging in to the Django admin (an alternative is using the login ``test-super`` 
+   with password ``test-password``), scroll down to the Sites section. 
+
+#. Click on "Change". 
+
+#. Click on the entry that says ``localhost:8000``. 
+
+#. Change both the domain and display name from ``localhost:8000`` to ``localhost.org:8000``. 
+
+#. Click "Save".
```
escattone: This whole section is important until I can create a new sample database, after which we could remove it.

docs/installation.rst
```diff
@@ -306,6 +322,14 @@ To stop the containers::
 
     docker-compose stop
 
+If you have made changes to the .env or /etc/hosts file, it's a good idea to run::
+
+    docker-compose down
+    docker-compose pull
+    git pull --ff-only origin master
+    docker-compose up -d
```
escattone: The `docker-compose pull` and `git pull --ff-only origin master` commands shouldn't be part of this, since they're not necessary after making changes to the `.env` and/or `/etc/hosts` files. I would suggest just this:
```
docker-compose down
docker-compose up -d
```
or `docker-compose stop` could be used instead of `docker-compose down`.

peterbe: I would strongly prefer `docker-compose stop` since otherwise you're losing your mounted databases and you have to start over. There are times you want to do that, but it's not related to making a change to the `.env` file. 

Also, I'm still struggling to see the value in `-d` since it can easily hide important error messages and warnings on the running processes.

docs/installation.rst
```diff
@@ -306,6 +322,14 @@ To stop the containers::
 
     docker-compose stop
 
+If you have made changes to the .env or /etc/hosts file, it's a good idea to run::
+
+    docker-compose down
```
tobinmori: per @peterbe's point, can we change 'down' to 'stop'?

docs/installation.rst
escattone: Just wanted to clear-up something here. Your mounted volumes storing your database data are **not** lost when using `docker-compose down`. You'd have to use `docker-compose down --volumes` to do that. It really doesn't matter if you use `docker-compose down` or `docker-compose stop` after updating your `.env` file. I always use `docker-compose down` out of habit (it also covers the case of running new images after doing a `docker-compose pull`), but in this case of updating the `.env` file, `docker-compose stop` does just what's needed and nothing else (like removing networks), so we can go with that.

tobinmori: ahh. thanks for shedding light on that @escattone! ok, nevermind my comment then. :)

docs/installation.rst
mindy: 👍

docs/installation.rst
mindy: changed to 'stop' 👍

docs/installation.rst
mindy: Thank you for the clarity! 

I also removed the `-d` per Peter's comment.

mindy:
Ready for another 👀 @escattone @peterbe  @tobinmori

docs/installation.rst
```diff
@@ -204,18 +204,33 @@ management command. Replace ``YOUR_USERNAME`` with your username and
...
+
 Enable GitHub/Google authentication (optional)
 =======================================
 Since Google's OAuth requires us to use a valid top-level-domain, we're going to use
 http://localhost.org:8000 as an example for every URL here.
 
 To automate setting Django up for social auth you can run
 ``docker-compose exec web ./manage.py configure_social_auth`` and follow its steps (and
-ignore the rest of this section).
+ignore the rest of this section).
```
peterbe: Your editor doesn't seem to trim trailing spaces upon save. If you have VSCode or Atom, it should be easy to set up.

docs/installation.rst
```diff
@@ -306,6 +322,12 @@ To stop the containers::
 
     docker-compose stop
 
+If you have made changes to the .env or /etc/hosts file, it's a good idea to run::
```
peterbe:
```suggestion
If you have made changes to the ``.env`` or ``/etc/hosts`` file, it's a good idea to run::
```

docs/installation.rst
mindy: Removed extra space-- I was editing in GitHub's UI, so that's probably why it snuck in there. I'll make sure VSCode has this setting set.

PR #6250, opened 2019-12-17
right locale for wiki site search

### LINKED ISSUE #6146: Macro search results are mangled for non-en-US locales
See for example https://wiki.developer.mozilla.org/en-US/search?locale=*&kumascript_macros=WebExtAllExamples&topic=none

This lists all pages that call WebExtAllExamples, across all locales. One entry looks like:

Screen Shot 2019-11-21 at 4 30 25 PM

This is the German translation of the https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Examples page.

But the first link, "**Beispiele für Erweiterungen**", has the en-US locale in the URL, like this: https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Beispiele - note the translated slug but the en-US locale. If I click it, I get "Create a new page", because that page doesn't exist.

After the short description, the entry is supposed to have "`${url} Score: 82.20941 translated from ${original}`, where `url` is the localized page, and `original` is the en-US version. But these are wrong too:

* `url`: https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Beispiele - nonexistent page with en-US locale but de slug
* `original`: https://developer.mozilla.org/de/docs/Mozilla/Add-ons/WebExtensions/Beispiele - the proper value for `url`

 I've seen some cases where the "`${url} Score: 82.20941 translated from ${original}` bit doesn't appear, and then there is no usable link to the actual page, and I have to guess what the locale is, to be able to fix the link.

peterbe:
Fixes #6146

You can't argue with the results!
Screen Shot 2019-12-16 at 8 25 02 PM

To test:

1. Pick a rarely used macro from wiki. For example `DOMAttributeMethods`
2. Then go to a wiki page that has a custom translated slug. E.g. `http://wiki.mdn.localhost:8000/fr/docs/Apprendre/HTML/Introduction_%C3%A0_HTML` if you have it. 
3. Edit it and put in `{{DOMAttributeMethods('stuff')}}`
4. Now search for `http://wiki.mdn.localhost:8000/en-US/search?locale=*&kumascript_macros=DOMAttributeMethods&topic=none`
5. What you might see, before switching to my branch, is the link of that search result goes to `http://wiki.mdn.localhost:8000/en-US/docs/Apprendre/HTML/Introduction_%C3%A0_HTML` which is the wrong locale!
6. Check out my branch and the search should wowkr again.

### DIFF
--- a/kuma/search/fields.py
+++ b/kuma/search/fields.py
@@ -37,5 +37,4 @@ class SiteURLField(serializers.ReadOnlyField):
         args = [getattr(value, arg) for arg in self.args]
         kwargs = {arg: getattr(value, arg) for arg in self.kwargs}
         locale = getattr(value, 'locale', settings.LANGUAGE_CODE)
-        path = reverse(self.url_name, locale=locale, args=args, kwargs=kwargs)
-        return '%s%s' % (settings.SITE_URL, path)
+        return reverse(self.url_name, locale=locale, args=args, kwargs=kwargs)
--- a/kuma/search/jinja2/search/results.html
+++ b/kuma/search/jinja2/search/results.html
@@ -22,17 +22,16 @@
 {% block bodyclass %}search-results search-results-no-demo search-navigator-flush{% endblock %}
 
 {% macro doc_block(doc, index) -%}
-  {% set doc_url = wiki_url(doc.slug) %}
   <li class="result-{{ index }}">
     <div class="column-container">
       <div class="column-1">
           {% include 'includes/icons/file-icons/plain-text.svg' %}
       </div>
       <div class="column-5 result-list-item">
-        <h4><a href="{{ doc_url }}"{% if index == 1 %} tabindex="1"{% endif %}>{{ doc.title }}</a></h4>
+        <h4><a href="{{ doc.url }}"{% if index == 1 %} tabindex="1"{% endif %}>{{ doc.title }}</a></h4>
         <p>{{ doc.excerpt|safe }}</p>
-        <p class="search-meta"><a href="{{ doc_url }}">{{ doc_url }}</a>
-        {% if request.user.is_superuser %}Score: {{ doc.score }}{% endif %}
+        <p class="search-meta"><a href="{{ doc.url }}">/{{ doc.locale }}/{{ doc.slug }}</a>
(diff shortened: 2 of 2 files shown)


kuma/search/jinja2/search/results.html
```diff
@@ -22,17 +22,16 @@
...
           {% include 'includes/icons/file-icons/plain-text.svg' %}
       </div>
       <div class="column-5 result-list-item">
-        <h4><a href="{{ doc_url }}"{% if index == 1 %} tabindex="1"{% endif %}>{{ doc.title }}</a></h4>
+        <h4><a href="{{ doc.url }}"{% if index == 1 %} tabindex="1"{% endif %}>{{ doc.title }}</a></h4>
         <p>{{ doc.excerpt|safe }}</p>
-        <p class="search-meta"><a href="{{ doc_url }}">{{ doc_url }}</a>
-        {% if request.user.is_superuser %}Score: {{ doc.score }}{% endif %}
+        <p class="search-meta"><a href="{{ doc.url }}">/{{ doc.locale }}/{{ doc.slug }}</a>
+        {% if request.user.is_superuser %}Score: {{ doc.score|floatformat }}{% endif %}
```
peterbe: Unrelated but I thought it stinks to have to see those massive floating point numbers when 99% of the time you don't care.

kuma/search/fields.py
```diff
@@ -37,5 +37,4 @@ def to_representation(self, value):
         args = [getattr(value, arg) for arg in self.args]
         kwargs = {arg: getattr(value, arg) for arg in self.kwargs}
         locale = getattr(value, 'locale', settings.LANGUAGE_CODE)
-        path = reverse(self.url_name, locale=locale, args=args, kwargs=kwargs)
-        return '%s%s' % (settings.SITE_URL, path)
+        return reverse(self.url_name, locale=locale, args=args, kwargs=kwargs)
```
peterbe: The only time absolute URLs should ever be used is when you high up in the view code. This is a low level code and should *not* concern itself with how the URL should be presented.

PR #6252, opened 2019-12-17
instrumentation for header elements

### LINKED ISSUE #6199: Turn back on instrumentation of MDN header
**Summary**
When we switched to the react header, we disabled sending events for usage of the header, the breadcrumbs, login, etc.

**Steps To Reproduce (STR)**
Hover or click on a header item in the wiki, it will send an event, do the same on the developer.mozilla.org and it will not.

schalkneethling:
Fixes #6199

Hey @peterbe,

I have completed instrumentation for Sign In, breadcrumbs, and the accessibility menu. The one I have not done yet is the main menu. The reason I have skipped this one is that I am not sure how to best approach this.

The `MainMenu` is defined in `kuma/javascript/src/header/main-menu.jsx` but, it is neither a functional nor a `class` component. I am not sure if we have used the `GAProvider` in such a context.

I did some searching around the code and it seems that we have not. I would, therefore, appreciate your review of what I have done here, and we can then have a discussion on how to handle the main menu. Thanks!


peterbe:
@schalkneethling I fixed the PR so it doesn't mention "Fixes #..." in the title but on the first line of the description instead. 
Do you want to fix up the jest tests first?

schalkneethling:
> @schalkneethling I fixed the PR so it doesn't mention "Fixes #..." in the title but on the first line of the description instead.

Ah, I did not know that was our convention, thanks, Peter.

> Do you want to fix up the jest tests first?

Sure, probably just snapshot tests

schalkneethling:
@peterbe Some extra context. So, I have added some items and questions here around what I could see that is tracked: https://github.com/mdn/kuma/issues/6199#issuecomment-566464998

Then the main files to look at are:

`kuma/static/js/wiki.js` ~ lines 92 - 140
`kuma/static/js/auth.js` ~ lines 12 - 20

You will notice here is also a custom event that is fired and listened for on lines 27-33 as well as 36-42 but, I reckon these will still be logged.

kuma/javascript/src/a11y/a11y-nav.jsx
```diff
@@ -1,23 +1,64 @@
...
+                    id="skip-main"
+                    href="#content"
+                    onClick={() => {
+                        sendAccessMenuItemClick(
+                            'Skip to main content',
+                            '#content'
+                        );
+                    }}
+                >
                     {gettext('Skip to main content')}
```
peterbe: In [wiki.js](https://github.com/mdn/kuma/blob/a3d6cb8e7314b70e6c07a54a7c8d86d6c7682448/kuma/static/js/wiki.js#L95) we're only tracking `contextmenu` clicks. You have it on regular clicks. 
However, I put some `console.log` in on that event handler and it doesn't fire.

peterbe:
You're missing tracking on the buttons in `main-menu.jsx`. We no longer use `onclick` for these because they use CSS `:hover` to open up. But I think we need something like:
```diff
                        <button
                            type="button"
                            className="top-level-entry"
                            aria-haspopup="true"
+                           onMouseOver={event => {
+                               sendMenuOpen();
+                           }}
                            onTouchStart={() => {
                                // Ultimately, because there's no :hover on
                                // mobile, we have to compensate for that using
                                // JavaScript.
+                              sendMenuOpen();
                                setShowSubMenu(
                                    showSubMenu === menuEntry.label
                                        ? null
                                        : menuEntry.label
                                );
                            }}
                        >
                            {menuEntry.label}
                            <span
                                className="main-menu-arrow"
                                aria-hidden="true"
                            >
                                ▼
                            </span>
                        </button>
```

kuma/javascript/src/breadcrumbs.jsx
```diff
@@ -8,6 +11,22 @@ type DocumentProps = {
...
 
 export default function Breadcrumbs({ document }: DocumentProps) {
+    const ga = useContext(GAProvider.context);
+
+    /**
+     * Send a signal to GA when there is an interaction on one
+     * of the breadcrumb menu links.
+     * @param {String} label  - The label for the GA event
+     */
+    const sendBreadcrumbItemClick = label => {
```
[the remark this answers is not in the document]
peterbe: I think we should do the same here. Instead of creating a new function, we should just do:
```javascript
function sendBreadcrumbItemClick(event) {
  const label = event.target.href;
  ga('send', ...)
}  
```

kuma/javascript/src/header/login.jsx
```diff
@@ -61,13 +63,29 @@ export default function Login(): React.Node {
...
+        };
         // Otherwise, show a login prompt
         return (
             <a
                 href={`/users/github/login/?next=${LOCATION}`}
                 data-service="GitHub"
                 rel="nofollow"
                 className="signin-link"
+                onClick={() => {
+                    sendSignInEvent('Github');
```
peterbe: Same point here. Instead of spelling `Github` twice (it's already misspelled) we can use `event.target.dataset.service` to get the same string.

schalkneethling:
> You're missing tracking on the buttons in `main-menu.jsx`. We no longer use `onclick` for these because they use CSS `:hover` to open up. But I think we need something like:

Yup, I am not sure how to use `GAProvider` inside main-menu as it is neither a functional nor class component. Any suggestions on how we can use it in these cases? Perhaps a custom hook that we can then re-use for everything?

kuma/javascript/src/a11y/a11y-nav.jsx
```diff
@@ -1,23 +1,64 @@
...
+     * @param {String} label  - The label for the GA event
+     */
+    const sendAccessMenuItemClick = (action, label) => {
+        ga('send', {
+            hitType: 'event',
+            eventCategory: 'Access Links',
+            eventAction: action,
+            eventLabel: label
+        });
+    };
```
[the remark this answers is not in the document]
schalkneethling: That was my default intent but then I thought no, wait, React works different with the `onClick`s :( Just when I think I understand "thinking in React" :)

schalkneethling:
> Remember, please don't rebase. Merge in `master` if you need to resolve conflicts.

So no, `git rebase -i master`? I am so used to that, have been doing it forever :)  So, do you mean I should instead do, `git merge master` on the feature branch? Thanks Peter!

kuma/javascript/src/a11y/a11y-nav.jsx
schalkneethling: Interesting, I definitely saw the GA events in the console. I was using the enter key, but should not make a difference right?

kuma/javascript/src/a11y/a11y-nav.jsx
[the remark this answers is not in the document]
schalkneethling: Oh wait, I misread. It is something in the middle of what I was thinking :)

kuma/javascript/src/a11y/a11y-nav.jsx
schalkneethling: > In [wiki.js](https://github.com/mdn/kuma/blob/a3d6cb8e7314b70e6c07a54a7c8d86d6c7682448/kuma/static/js/wiki.js#L95) we're only tracking `contextmenu` clicks.

Hmmm, does that not track both? i.e. click and contextmenu(i.e. right click)

schalkneethling: Looks to be the case: https://api.jquery.com/on/

schalkneethling: Made the changes here and tested again using the enter key as well as clicking the links, and the GA events are definitely being fired:

Screenshot 2019-12-18 at 12 23 05

kuma/javascript/src/header/login.jsx
schalkneethling: >  (it's already misspelled)

You mean because it is GitHub? That is based on their brand guidelines.

schalkneethling: Whoops, I made the 💩 💩 ;p

schalkneethling:
Thanks for the review @peterbe, updated.

peterbe:
> So, do you mean I should instead do, `git merge master` on the feature branch?

Yes. 
Then your merge conflict resolution becomes its own commit that can be seen in this git history which helps the PR. 

The only time you should use `git rebase -i master` is if you want to remove a junk commit message. E.g. "debug this f**ing thing"

You should consider using https://pypi.org/project/gg/
If you have that installed you can do things like:

    gg mastermerge

which will hop over to the `master` branch, git pull the later from origin, go back to your topic branch and run `git merge master` for you.

schalkneethling:
> which will hop over to the `master` branch, git pull the later from origin, go back to your topic branch and run `git merge master` for you.

Ooooooh, nice! Thanks Peter

schalkneethling:
> > You're missing tracking on the buttons in `main-menu.jsx`. We no longer use `onclick` for these because they use CSS `:hover` to open up. But I think we need something like:
> 
> Yup, I am not sure how to use `GAProvider` inside main-menu as it is neither a functional nor class component. Any suggestions on how we can use it in these cases? Perhaps a custom hook that we can then re-use for everything?

@peterbe Any thoughts on the above?

kuma/javascript/src/breadcrumbs.jsx
```diff
@@ -8,6 +11,22 @@ type DocumentProps = {
...
 
 export default function Breadcrumbs({ document }: DocumentProps) {
+    const ga = useContext(GAProvider.context);
+
+    /**
+     * Send a signal to GA when there is an interaction on one
+     * of the breadcrumb menu links.
+     * @param {String} label  - The label for the GA event
+     */
+    function sendBreadcrumbItemClick(label) {
```
peterbe: Can this also please take `event` instead and on line 26 do `event.target.href` instead? That way you're not creating new functions on ever render.

schalkneethling:
> How is `main-menu` not a component?

I have no idea what happened :( Initially when I tried to use `useContext` inside `main-menu` it got an eslint error that I cannot use it outside a React component or custom hook. Which was a little confusing `export default MainMenu;` which is a function.

That is neither here nor there though. It works now ¯\_(ツ)_/¯

schalkneethling:
@peterbe Updates made and instrumentation added to main menu. Not sure what to do about the conflicts in those snapshot tests. Hmmm, will take a look.

schalkneethling:
@peterbe Also added signaling to `Dropdown` components. Currently that would be the mugshot as well as the language menu in the header.

schalkneethling:
> That is neither here nor there though. It works now ¯_(ツ)_/¯

Now I know why. It was because I was doing:

`const ga = useContext(GAProvider.context);`

inside for example `sendDropdownInteraction` as opposed to inside `export default function Dropdown` 🤕 Well, you live you learn 😸

schalkneethling:
@peterbe Ok, also updated so clicks on items in the TOC will also be sent to GA

kuma/javascript/src/toc.jsx
```diff
@@ -1,12 +1,44 @@
...
+                hitType: 'event',
+                eventCategory: 'MozMenu',
+                eventAction: action,
+                eventLabel: label
+            });
+        }
+    }
+
+    useEffect(() => {
+        const documentToc = document.querySelector('.document-toc');
```
peterbe: You should only use `.querySelector` on stuff you don't have access to in the JSX. 
I demonstrated this here: https://codesandbox.io/s/proud-resonance-i542y

schalkneethling: Dang it, you are correct. I can totally do this using a `ref` on the section. Meh. Thanks, @peterbe

schalkneethling:
> Generally, it seems to be working. I can take one final review when you've tackled the git conflicts and fixed the use of `document.querySelector`.

Updated. I found this piece in the sandbox code interesting:

```
return () => {
      if (tocMenu) {
        tocMenu.removeEventListener("click", sendTOCClicks);
      }
};
```

When would this piece of code run? Would this happen after this first `onClick` has taken place? Thanks again!

peterbe:
> When would this piece of code run?

It doesn't run but it's good practice.

schalkneethling:
@peterbe updated and rebased.

peterbe:
Please don't rebase and squash. Now I can't see what has changed incrementally so I have to kinda start over :)
Also, did you notice that it broke the jest tests?

kuma/javascript/src/header/main-menu.jsx
```diff
@@ -218,11 +240,23 @@ const _MainMenu = ({ document, locale }: Props) => {
...
-                                        <a href={item.url}>{item.label}</a>
+                                        <a
+                                            href={item.url}
+                                            onClick={sendMenuItemInteraction}
+                                            onContextMenu={
+                                                sendMenuItemInteraction
+                                            }
+                                        >
+                                            {item.label}
+                                        </a>
```
peterbe: This is what happens when I hover over the menu a bit:
Screen Shot 2019-12-23 at 2 52 54 PM
(using Chrome and my debug build)

THere's something weird going on with that `onMouseOver` that's over-firing.

schalkneethling:
> Please don't rebase and squash. Now I can't see what has changed incrementally so I have to kinda start over :)

Meh, apologies for that. I was being premature ;p

> Also, did you notice that it broke the jest tests?

So strange to get these conflicts as these things are generated by Jest. I notice when looking at your PR that there was a `/` that sneaked in along the way ☹️

schalkneethling:
> Merge in [this](https://github.com/schalkneethling/kuma/pull/4) and I think it's good to go.

Merged.

PR #6256, opened 2019-12-17
make welcome email more user focused

escattone:
Fixes https://github.com/mdn/sprints/issues/2425

The `Subject:` line of the email remains `Take the next step to get involved on MDN!`. Is that acceptable?


atopal:
Thanks for catching that Ryan.

Let's go with: Getting started with your new MDN account

escattone:
Updated the `Subject` of the email as you suggested, thanks for the quick review @atopal!

PR #6257, opened 2019-12-17
do not provide a default S3 endpoint URL

escattone:
This simple change allows us to use `vhost` S3-bucket addressing on stage/prod, while maintaining `path` addressing for local development using `minio`. Thanks to @callahad for pointing me in the right direction for this change.


escattone:
This work was part of https://github.com/mdn/infra/issues/354.

PR #6271, opened 2019-12-20
Use correct message for outdated content

### LINKED ISSUE #6251: Russian translated page says it's not translated to Russian yet
# Request type

- [ ] Please close this issue, I accidentally submitted it without adding any details
- [ ] New documentation
- [x] Correction or update

# Details

🤭I have occasionally followed the 🔖bookmark from my handy and suddenly appeared on the 🇷🇺Russian version of the actual article (thought my browser is configured for 🇬🇧 En locale) , which declares itself as it's an English version of the documentation, and that it is missing any translations to the rest of the languages. 

🤔 Well... I guess the message is too tacky and in fact is outdated, so probably, either author of the original text, or the one who wrote the translation must update the content of the warning message according to the actual state of the main article's content. 

**EDIT** URL for the page is: https://developer.mozilla.org/ru/docs/Web/Web_Components/%D0%98%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5_%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D1%82%D0%B5%D0%BB%D1%8C%D1%81%D0%BA%D0%B8%D1%85_%D1%8D%D0%BB%D0%B5%D0%BC%D0%B5%D0%BD%D1%82%D0%BE%D0%B2

### LINKED ISSUE #6308: Message about translation status differ in wiki vs read-only
**Summary**
Compare https://developer.mozilla.org/de/docs/Web/HTML/Element/form and https://wiki.developer.mozilla.org/de/docs/Web/HTML/Element/form

On the read-only, it says:

*"You are reading the English version of this content because there is no translation for this locale yet. Help us translate this article!"*

But clearly the text is in German. 

On the wiki it says:

*"This translation is incomplete. Please help translate this article from English."*

which I can believe to be accurate.

Gregoor:
Fixes #6251 
Fixes #6308

Looks like I was erroneously showing the message for non-existent content before!

### DIFF
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -27,12 +27,10 @@ function TranslationStatus({
     } else if (translationStatus === 'outdated') {
         content = (
             <>
-                {gettext(
-                    'You’re reading the English version of this content since no translation exists yet for this locale.'
-                )}
+                {gettext('This translation is incomplete.')}
                 &nbsp;
                 <a href={translateURL} rel="nofollow">
-                    {gettext('Help us translate this article!')}
+                    {gettext('Please help translate this article from English')}
                 </a>
             </>
         );


kuma/javascript/src/article.jsx
```diff
@@ -27,12 +27,10 @@ function TranslationStatus({
...
         content = (
             <>
-                {gettext(
-                    'You’re reading the English version of this content since no translation exists yet for this locale.'
-                )}
+                {gettext('This translation is incomplete.')}
                 &nbsp;
                 <a href={translateURL} rel="nofollow">
-                    {gettext('Help us translate this article!')}
+                    {gettext('Please help translate this article from English')}
```
schalkneethling: I wonder if we should mention English here. Was that not part of the confusion reported in the issue? Perhaps we can just say `Please help translate this article`?

sirartemis: These articles are translate completely. Yes there are mistakes in them ,but to understand what it is possible.
Problem

schalkneethling: Ah, I see @sirartemis, so the problem is not the text of the message, it is that the message is shown on a page that has already been translated?

sirartemis: Exactly. Whole problem in URL's. Need to change URL paths to be on English language only.
In places where main sections on Russian, no messages, but there on English - there are messages.
Need change something in code, so that there is no translation check on the translation path and do another check.
Messages appear because of this check

Gregoor: > I wonder if we should mention English here. Was that not part of the confusion reported in the issue? Perhaps we can just say `Please help translate this article`?

From my understanding, translations always come from english, which is why this string was also used originally (I got both strings from the old code, just made a mistake when picking the string).

> Ah, I see @sirartemis, so the problem is not the text of the message, it is that the message is shown on a page that has already been translated?

The message is shown when a translated revision is older than 10 days and tagged as inprogress. I can't speak to how reasonable that model is, but it shouldn't have changed from the old version.

Gregoor:
Thanks for the review Schalk! And @sirartemis could you create a new issue if you also think this is not the same one as #6251 (which is what I think). Thanks!

PR #6281, opened 2019-12-24
Load modal js as defer

### LINKED ISSUE #6275: Load modal javascript as defer
Currently, the JavaScript for the auth modal is marked as `async`. As @peterbe mentions here https://github.com/mdn/kuma/pull/6094#discussion_r344937847, that is not ideal. To be honest, I believe mentally my intent was to mark this as `defer` and not `async`.

We should update it to be `defer`

schalkneethling:
@peterbe Switched to `defer` and updates comment.

Fixes #6275

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1847,6 +1847,7 @@ ENABLE_NEWSLETTER_SIGNUP = config('ENABLE_NEWSLETTER_SIGNUP', default=False, cas
 # Whether or not to enable the BCD signalling feature.
 # Affects loading of CSS (statically) and JS (in runtime).
 ENABLE_BCD_SIGNAL = config('ENABLE_BCD_SIGNAL', default=True, cast=bool)
+
 # Enable or disable the multi auth(Google and Github) sign-in flow
 # When disabled, Github will be the default and only Auth provider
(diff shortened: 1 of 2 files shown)


kuma/settings/common.py
```diff
@@ -1847,7 +1847,8 @@ def get_user_url(user):
 # Whether or not to enable the BCD signalling feature.
 # Affects loading of CSS (statically) and JS (in runtime).
 ENABLE_BCD_SIGNAL = config('ENABLE_BCD_SIGNAL', default=True, cast=bool)
-# Multi Auth Switch
+# Enable or disable the multi auth(Google and Github) sign-in flow
```
peterbe: Put a space between this and the previous section above.

kuma/settings/common.py
```diff
@@ -1847,7 +1847,8 @@ def get_user_url(user):
...
+# When disabled, Github will be the default and only Auth provider
 MULTI_AUTH_ENABLED = config('MULTI_AUTH_ENABLED', default=False, cast=bool)
```
peterbe: Too late now but it would be nice to have the same pattern for naming these kinds of Boolean constants. 
So, this should be ENABLE_MULTI_AUTH

PR #6296, opened 2019-12-31
esc key should close auth modal

### LINKED ISSUE #6272: auth modal doesn't respect Esc key
I'm not a modal expert but it seems counter-intuitive that you can't click anywhere outside the modal to close. Either way, the `Esc` key handler only works if the modal is focused which it won't be unless you click into it. 

Also, in Chrome there's a rather ugly blue border for focus on it. 
Screen Shot 2019-12-20 at 2 51 49 PM

### LINKED ISSUE #6273: modalDialog is tied to multi-auth
Nit!

`mdn.modalDialog` is supposed to be generic but it [depends on `multi-auth` being enabled](https://github.com/mdn/kuma/blob/55a1f34e4c33022ada5628b06285e04bfe2c5268/kuma/static/js/components/modal.js#L23). 

Move that check over to [auth-modal.js](https://github.com/mdn/kuma/blob/55a1f34e4c33022ada5628b06285e04bfe2c5268/kuma/static/js/components/user-signup/auth-modal.js#L6) instead. 

Also, it's confusing my little brain that one function is called `showModal` and it's a function that returns a boolean without side-effects. But there's another function called `closeModal` which doesn't return anything and is only side-effects.

peterbe:
Fixes #6272
Fixes #6273

### DIFF
--- a/kuma/static/js/components/user-signup/auth-modal.js
+++ b/kuma/static/js/components/user-signup/auth-modal.js
@@ -3,38 +3,44 @@
 
     var authModalContainer = document.getElementById('auth-modal');
 
-    if (!window.mdn.modalDialog.showModal(authModalContainer)) {
+    if (
+        !authModalContainer ||
+        !window.mdn.multiAuthEnabled ||
+        !window.mdn.modalDialog.shouldShowModal()
+    ) {
         return;
     }
 
-    /* The React header takes just a couple of milliseconds to
-       long to load so, we need to wrap the code below in a 
-       setTimeout */
-    setTimeout(function() {
-        var pageHeader = document.querySelector('.page-header');
-        pageHeader.addEventListener('click', function(event) {
-            if (event.target.classList.contains('signin-link')) {
-                event.preventDefault();
-                var modalContentContainer = authModalContainer.querySelector(
-                    'section'
-                );
-                var closeModalButton = document.getElementById('close-modal');
+    var pageHeader = document.querySelector('.page-header');
+    pageHeader.addEventListener('click', function(event) {
+        function handleKeyup(event) {
+            if (event.key === 'Escape') {
+                closeModalButton.click();
+            }
+        }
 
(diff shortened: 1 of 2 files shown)


kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -3,38 +3,44 @@
...
+        !window.mdn.multiAuthEnabled ||
+        !window.mdn.modalDialog.shouldShowModal()
+    ) {
         return;
     }
 
-    /* The React header takes just a couple of milliseconds to
-       long to load so, we need to wrap the code below in a 
-       setTimeout */
-    setTimeout(function() {
```
peterbe: Much of this file's diff gets very complicated. It's related to my deleting the line  `setTimeout(function() {` and `}, 100);` which don't make any sense to have any more now that this .js is loaded with `defer`.

kuma/static/js/components/user-signup/auth-modal.js
```diff
@@ -3,38 +3,44 @@
...
                 );
-            }
-        });
-    }, 100);
+                document.removeEventListener('keyup', handleKeyup);
+            });
+
+            window.mdn.modalDialog.handleKeyboardEvents(authModalContainer);
+
+            document.addEventListener('keyup', handleKeyup);
```
schalkneethling: I am not sure about moving this here? It seems the functionality we would want with modals is to always listen on the `body` so, would it not be better to have `modal.js` take care of that?

peterbe: You technically can't. You need to make sure you execute the `document.body.removeEventListener('keyup', sameHandlerFunction)` when the modal closes. You don't get access to that inside `modal.js` because the closing is controlled from this file. 

It's not like `React.Component` which gets a `componentWillUnmount` so it itself would know when it's closing down.

PR #6300, opened 2020-01-02
allow api search without 'q'

### LINKED ISSUE #6265: Feature regression: "Advanced" search no longer allows query-less search
Previously, using the advanced search functionality (through the special URLs documented on https://developer.mozilla.org/en-US/docs/MDN/Contribute/Tools/Search) did not require the `q=` term specifying a search term.

This allowed us to do things like use `css_classnames=` to find everywhere a given CSS is used in documentation content, `html_attributes=` to locate every use of a given HTML attribute, and `kumascript_macros=` to find every use of one of a set of one or more KumaScript macros.

These searches were often used during refactoring and cleanup work, as well as when doing style and layout improvements to suites of articles. When we need them, we usually need them pretty badly.

While I have no *immediate*, **urgent** need for these search abilities, when the time comes that I do, it will probably be pretty urgent, so I'm filing this now in hopes of getting a fix for this onto the schedule.

peterbe:
Fixes #6265

With this change, the v1 API now allows a search with absolutely no parameters. 
That's how it used to work when it was only the Wiki search and if you added the `.json` suffix to the URL. 
Screen Shot 2020-01-02 at 9 55 22 AM

However, search the search input field (upper right-hand corner on every page) disallows submission without a query, I think it's fair to keep the web UI to respect that. So now you (still) get:
Screen Shot 2020-01-02 at 9 21 52 AM


kuma/core/tests/test_views.py
```diff
@@ -95,31 +95,12 @@ def test_global_home(self):
...
-        doc = pq(r.content)
-        assert m.message == doc.find('div.global-notice').text()
-
-        url = reverse('home')
-        r = self.client.get(url, follow=True)
-        assert 200 == r.status_code
-
-        doc = pq(r.content)
-        assert not doc.find('div.global-notice')
-
```
peterbe: Lemme explain why I deleted this test. 
It relies on going to two different pages. Pages that aren't a wiki document. 
But you can't go to the `/search/` page any more and get a 200 if you don't have a `?q=....` so that page can't be used to test the functionality of Soapbox.
Second, who cares about Soapbox. Pretty sure it works and is never being worked on. 
Third, actually, the error that happens isn't that `request.GET.get('q')` was falsy but that the Elasticsearch query failed hard because the Elasticsearch index doesn't exist in the `elasticsearch` in `docker-compose`. It's a wrong to have any test rely on that database unless it's under the fixtures in the `ElasticTestCase` base class.

peterbe:
@Gregoor Basically, as of this change you can open `http://localhost.org:8000/api/v1/search/en-US` and before you couldn't.  `http://localhost.org:8000/en-US/search?q=` and `http://localhost.org:8000/en-US/search` still throw up a nice error as before. More info in the PR description.

peterbe:
Ping @Gregoor 
If you feel swamped with PRs, consider deflecting to someone else.

a2sheppy:
Does the UI provide a means of accessing this still? I haven't looked recently.

peterbe:
> Does the UI provide a means of accessing this still? I haven't looked recently.

No changes to the UI.

PR #6301, opened 2020-01-02
don't index html_attributes and css_classnames

### LINKED ISSUE #6264: Create Lightweight Elasticsearch Dev Instance
About 60% of the bytes we transfer from MySQL into Elasticsearch is a field called `html_attributes`. Actually, here's a breakdown:
```
59.7%     html_attributes
35.0%     content
3.5%      css_classnames
0.8%      summary
0.3%      kumascript_macros
0.3%      tags
0.1%      slug
0.1%      modified
0.1%      title
0.0%      id
0.0%      locale
0.0%      boost
0.0%      parent
```

The only way to search on this `html_attributes` (or `css_classnames` for that matter) is if you find this page: https://developer.mozilla.org/en-US/docs/MDN/Contribute/Tools/Search 
but I'm not even sure any of that is working. 
https://developer.mozilla.org/en-US/search?locale=en-US&html_attributes=style will not work and https://wiki.developer.mozilla.org/en-US/search?locale=en-US&html_attributes=style doesn't find anything. 

Having that much data into Elasticsearch matters for developers using the `docker-compose` tooling. If you have over 5,000 documents indexed in Elasticsearch need to massively increase how much RAM you're giving to docker. 

It would be nice to not have to index that stuff for your local dev instance.

peterbe:
Fixes #6264

Before:
```
kuma@de9e7356b28f:/app$ curl http://elasticsearch:9200/_cat/indices?v
health status index          uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   mdn-main_index O6qoannIRMCbCV6vQwmO_Q   1   0       3895            0     63.2mb         63.2mb
```

After:
```
kuma@6beed7f29583:/app$ curl http://elasticsearch:9200/_cat/indices?v
health status index          uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   mdn-main_index rrp9t4pET2quU5uo4Ng5Gw   1   0       3895            1     36.6mb         36.6mb
```

Doesn't look that much actually but I only have ~4k documents indexed. And until this change you keep getting those annoying errors about "High watermark" from the log output of the elasticsearch container in `docker-compose`.


kuma/wiki/search.py
```diff
@@ -286,13 +287,8 @@ def reindex_all(cls, chunk_size=500, index=None, percent=100):
...
-            'Indexing %(total)d documents into %(total_chunks)d chunks of '
-            'size %(size)d into index %(index)s.' % {
-                'total': total,
-                'total_chunks': total_chunks,
-                'size': chunk_size,
-                'index': index.prefixed_name
-            }
+        message = (
+            f'Indexing {total:,} documents into {total_chunks} chunks of '
+            f'size {chunk_size} into index {index.prefixed_name}.'
```
peterbe: This change is unrelated but it hurt my eyes to see that legacy use of string interpolation. Also, I don't think these kinds of management commands would never need to be translated away from en-US.

peterbe:
@escattone This got rotten in the `black` update. It's a really simple fix that only relates to local development as the default doesn't change so nothing changes unless you set the extra settings. 
If you think this is a waste of your review time that I can accept that.

Issue #6283, opened 2019-11-19
E - User flow instrumented


atopal:
# User story
As the PM I want our new user flow to be instrumented so we can identify bottle necks

# Context
We have a new user flow that will hopefully allow more users to easily get an MDN account. We have already talked about some potential confusing screens. Instrumenting the flow should allow us to identify issues and check whether our optimizations are working. 

## My questions:
1. How many people have tried to sign up to MDN on any given day (not sign in to an account)?
2. How many people have tried to sign in to MDN on any given day?
3. How many users chose Google to sign up? 
4. How many users chose Github to sign up? 
5. How many users signed in with Google? 
6. How many users signed in with Github? 
7. How many Google users came back to create their profile?
8. How many Github users came back to create their profile?
9. How many people clicked to edit their username?
10. How many people clicked to change their email address?
11. How many people saw a username error?
12. How many people saw an email error?
13. How many people opted in to receive news?
14. How many people opted out of receiving news?
15. How many people ended up creating an account?
16. How many people ended up creating an account authenticated with Google?
17. How many people ended up creating an account authenticated with Github?

## Breakdown for events
category | action | label
3. ~~signup-flow | sign-up | Google~~ --> `signup-flow | auth-started | Google` (Google button in multi-auth modal pushed)
4. ~~signup-flow | sign-up | Github~~ --> `signup-flow | auth-started | GitHub` (GitHub button in multi-auth modal pushed)
5. ~~signup-flow | sign-in | Google~~ --> `signup-flow | returning-user-signin | Google`
6. ~~signup-flow | sign-in | Github~~ --> `signup-flow | returning-user-signin | GitHub`
7. `signup-flow | profile-audit | Google`
8. `signup-flow | profile-audit | Github`
9. `signup-flow | profile-edit | username edit`
10. `signup-flow | profile-edit | email edit`
11. `signup-flow | profile-edit-error | username`
12. `signup-flow | profile-edit-error | email`
13. `signup-flow | free-newsletter | opt-in`
14. `signup-flow | free-newsletter | opt-out`
16. `signup-flow | profile-created | Google`
17. `signup-flow | profile-created | Github`

# Acceptance criteria
- [x] sign-up events are implemented
- [x] sign-in events are implemented
- [x] profile events are implemented
- [x] newsletter events are implemented
- [x] test run successfully documented
- [x] Google analytics dashboard created for analysis

# Tasks
*If relevant, you can make a checklist for tasks.*

[...]

PR #6311, opened 2020-01-07
auth ga instrumentation

peterbe:
Part of #6283


docker-compose.yml
```diff
@@ -156,6 +158,14 @@ services:
     entrypoint: ""
     command: /bin/sh -c "for BUCKET in 'test' 'mdn-attachments'; do mc mb -p minio/$$BUCKET && mc policy set download minio/$$BUCKET; done"
 
+  mockga:
+    build:
+      context: mockga
+      dockerfile: Dockerfile
+    # command: server /data
```
peterbe: oops.

mockga/index.js
```diff
@@ -0,0 +1,40 @@
...
+    const parts = [
+        chalk.bold('ec: ') + chalk.green(showValue(ec, 20)),
+        chalk.bold('ea: ') + chalk.green(showValue(ea, 30)),
+        chalk.bold('el: ') + chalk.green(showValue(el, 50))
+    ];
+    const output = `[${new Date().toISOString()}] GA TRACK EVENT cid=${showValue(
+        cid,
+        10
+    )} | ${parts.join(' ')}`;
+    console.log(output);
```
peterbe: So for now, I'm going to venture to say it's enough to just console log print these. It looks like this, in the `docker-compose` logs (or foreground): 
```
[2020-01-07T20:25:31.197Z] GA TRACK EVENT cid=1161ff2e-… | ec: CATEGORY             ea: MYACTION                        el: MY LABEL
```

It's not perfect but it'll at least give you something. 
We could consider mounting a filesystem and have it write to a .log file that we add to `.gitignore`. Let's perhaps deal with that the day it matters.

escattone:
@peterbe Looks good to me so far. I'm going to look into how to determine the social OAuth provider used on sign-in. I suspect that information is available in the session, but let's see.

escattone:
@peterbe Just FYI, the social OAuth provider used for the login/sign-in is available via the `session` object, so for example:
```python
provider = request.session.get('sociallogin_provider')
# The provider will be None if the user is not authenticated.
assert provider in (None, 'github', 'google')
```

peterbe:
> going to look into how to determine the social OAuth provider used on sign-in

I already have that now. I've actually already implemented those. Just need to figure out how to test it. Thing is; I don't like writing tests for implementation details. I like using high level APIs in tests (i.e. to actually sign in) and use mock_requests etc. to see if something was sent to tracking.

kuma/users/signal_handlers.py
```diff
@@ -27,6 +37,54 @@ def on_user_signed_up(sender, request, user, **kwargs):
...
+        #       last_login=timezone.now(),
+        #       ...
+        #
+        # NOTE! The SocialAccount (`sociallogin.account` in this context) has
+        # its own `.last_login` and `.date_joined` which is *different* from
+        # the same fields as they're stored in the `auth_user` database.
+        # (peterbe's note): Not sure how that works or why it's so.
+        # So, use the date stored on the user and not on the SocialAccount
+        # instance.
+        if is_almost_same_dates(
```
escattone: I don't feel entirely comfortable with this approach, for two reasons:
- there is still a chance that the `on_user_logged_in` signal is never fired after the `on_user_signed_up` signal (due to the lack of a verified email on the user account, which I was able to demonstrate locally), so we can't always count on the `on_user_logged_in` to cover both cases
- the `is_almost_same_dates` function doesn't feel bullet-proof

So I tested a different approach that seems simpler and worked for me locally. I was thinking that we should send the `ACTION_SIGN_UP` GA event within the `on_user_signed_up` handler, set a custom attribute on the request object indicating that this was a sign-up (`is_signup` or something like that), and then only send the `ACTION_SIGN_IN` GA event within the `on_user_logged_in` handler if the `request` object does not have the `is_signup` attribute. Something like this:

```python
@receiver(user_signed_up, dispatch_uid='users.user_signed_up')
def on_user_signed_up(sender, request, user, **kwargs):
    """
    Signal handler to be called when a given user has signed-up.
    """
    request.is_signup = True
    track_event(
        CATEGORY_SIGNUP_FLOW,
        ACTION_SIGN_UP,
        request.session.get('sociallogin_provider')
    )
    if switch_is_active('welcome_email'):
        ....

@receiver(user_logged_in, dispatch_uid='users.user_logged_in')
def on_user_logged_in(sender, request, user, **kwargs):
    """
    Signal handler to be called when a given user has signed-in.
    """
    if not hasattr(request, 'is_signup'):
        track_event(
            CATEGORY_SIGNUP_FLOW,
            ACTION_SIGN_IN,
            request.session.get('sociallogin_provider')
        )
```

What do you think of that approach?

kuma/settings/common.py
```diff
@@ -1665,6 +1665,34 @@ def parse_iframe_url(url):
...
+# you might want to substitute the events tracking URL to a local dev server.
+# https://developers.google.com/analytics/devguides/collection/protocol/v1/reference
+GOOGLE_ANALYTICS_TRACKING_URL = config(
+    'GOOGLE_ANALYTICS_TRACKING_URL',
+    default='https://www.google-analytics.com/collect'
+)
+GOOGLE_ANALYTICS_TRACKING_URL = config(
+    'GOOGLE_ANALYTICS_TRACKING_URL',
+    default='https://www.google-analytics.com/collect'
+)
```
escattone: This code should be removed, since it's a duplicate of the code immediately above.

kuma/wiki/content.py
```diff
@@ -238,6 +238,9 @@ def get_seo_description(content, locale=None, strip_markup=True):
             if strip_markup:
                 seo_summary = summaryClasses.text()
             else:
+                from pprint import pprint
+                pprint([
+                    (item, to_html(item)) for item in summaryClasses.items()])
```
escattone: This was probably left in by mistake, and should be removed.

kuma/users/signal_handlers.py
peterbe: They're both hacks. Your solution and my solution. But A) your solution appears to resolve the flaw you found in mind and B) that `is_almost_same_dates()` trick never felt good. 
I'll just copy your solution but attempt to add some code comments to explain the reasoning. (similar to what I did to justify the `is_almost_same_dates` hack)

peterbe:
@escattone For the record, I believe this is complete now. r?

kuma/users/adapters.py
```diff
@@ -247,6 +251,14 @@ def save_user(self, request, sociallogin, form=None):
...
             # by a unique email address).
             user.socialaccount_set.filter(provider='persona').delete()
+            # Send an event to Google Analytics that the authentication
+            # from one provider could be added to an account created by a
+            # *different* provider (originally).
+            track_event(
+                CATEGORY_SIGNUP_FLOW,
+                ACTION_SOCIAL_AUTH_ADD,
+                f'{sociallogin.account.provider}-added'
+            )
```
escattone: We need to add something like the following after this `track_event()` call:
```python
request.social_auth_added = True
```
which we'll check for when handling the `user_signed_up` and `user_logged_in` signals (see further comments below).

kuma/users/signal_handlers.py
```diff
@@ -1,23 +1,67 @@
...
+            CATEGORY_SIGNUP_FLOW,
+            ACTION_AUTH_SUCCESSFUL,
+            sociallogin.account.provider)
+
+
 @receiver(user_signed_up, dispatch_uid='users.user_signed_up')
 def on_user_signed_up(sender, request, user, **kwargs):
     """
     Signal handler to be called when a given user has signed up.
     """
```
escattone: Here we need to add something like:
```python
if getattr(request, 'social_auth_added', False):
    return
```
since we don't want to do any of the things we do in this signal handler (i.e., sending GA events for `profile-created` and `free-newsletter` as well as sending the welcome email) when the user is simply adding another social provider to their existing MDN account.

kuma/users/signal_handlers.py
```diff
@@ -27,6 +71,16 @@ def on_user_signed_up(sender, request, user, **kwargs):
             )
 
 
+@receiver(user_logged_in, dispatch_uid='users.user_logged_in')
+def on_user_logged_in(sender, request, user, **kwargs):
+    sociallogin = kwargs.get('sociallogin')
+    if sociallogin and not getattr(request, 'signed_up', False):
```
escattone: We need to check for the `social_auth_added` attribute here as well, so these two lines should change to something like:
```python
if getattr(request, 'signed_up', False) or getattr(request, 'social_auth_added', False):
    return
sociallogin = kwargs.get('sociallogin')
if sociallogin:
```

kuma/users/signal_handlers.py
```diff
@@ -1,23 +1,67 @@
...
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
 from django.db.models.signals import post_delete, post_save, pre_delete
 from django.dispatch import receiver
 from waffle import switch_is_active
 
+from kuma.core.ga_tracking import (
+    ACTION_AUTH_SUCCESSFUL,
+    ACTION_FREE_NEWSLETTER,
+    ACTION_PROFILE_ALREADY_CREATED,
```
escattone:
```suggestion
    ACTION_RETURNING_USER_SIGNIN,
```

kuma/core/ga_tracking.py
```diff
@@ -0,0 +1,107 @@
...
+#
+# Here are some of those useful constants...
+
+CATEGORY_SIGNUP_FLOW = 'signup-flow'
+# Right before redirecting to the auth provider.
+ACTION_AUTH_STARTED = 'auth-started'
+# When redirected back from auth provider and it worked.
+ACTION_AUTH_SUCCESSFUL = 'auth-successful'
+# When we don't need to ask the user to create a profile.
+ACTION_PROFILE_ALREADY_CREATED = 'profile-already-created'
```
escattone:
```suggestion
ACTION_RETURNING_USER_SIGNIN = 'returning-user-signin'
```

kuma/users/signal_handlers.py
```diff
@@ -27,6 +71,16 @@ def on_user_signed_up(sender, request, user, **kwargs):
             )
 
 
+@receiver(user_logged_in, dispatch_uid='users.user_logged_in')
+def on_user_logged_in(sender, request, user, **kwargs):
+    sociallogin = kwargs.get('sociallogin')
+    if sociallogin and not getattr(request, 'signed_up', False):
+        track_event(
+            CATEGORY_SIGNUP_FLOW,
+            ACTION_PROFILE_ALREADY_CREATED,
```
escattone:
```suggestion
            ACTION_RETURNING_USER_SIGNIN,
```

kuma/users/tests/test_views.py
```diff
@@ -18,6 +19,14 @@
 
 from kuma.attachments.models import Attachment, AttachmentRevision
 from kuma.authkeys.models import Key
+from kuma.core.ga_tracking import (
+    ACTION_AUTH_STARTED,
+    ACTION_AUTH_SUCCESSFUL,
+    ACTION_FREE_NEWSLETTER,
+    ACTION_PROFILE_ALREADY_CREATED,
```
escattone:
```suggestion
    ACTION_RETURNING_USER_SIGNIN,
```

kuma/users/tests/test_views.py
```diff
@@ -1264,6 +1273,89 @@ def test_signup_public_github(self, is_public=True):
...
+                assert response.status_code == 302
+
+                track_event_mock_signals.assert_has_calls([
+                    mock.call(
+                        CATEGORY_SIGNUP_FLOW,
+                        ACTION_AUTH_SUCCESSFUL,
+                        'github'),
+                    mock.call(
+                        CATEGORY_SIGNUP_FLOW,
+                        ACTION_PROFILE_ALREADY_CREATED,
```
escattone:
```suggestion
                        ACTION_RETURNING_USER_SIGNIN,
```

kuma/users/views.py
```diff
@@ -716,7 +720,16 @@ def dispatch(self, request, *args, **kwargs):
...
+signup_as_view = SignupView.as_view()
+@redirect_in_maintenance_mode
+def signup(request, *args, **kwargs):
+    """Wrapper function just so we can trigger a track_event() call"""
+    if request.session.get('sociallogin_provider'):
+        track_event(
+            CATEGORY_SIGNUP_FLOW,
+            ACTION_PROFILE_AUDIT,
+            request.session['sociallogin_provider'])
+    return signup_as_view(request, *args, **kwargs)
```
escattone: The problem with this code is that this view is called for both GET and POST (POST when the `Create profile` button is clicked), so we will see multiple events when the user tries to create a profile (at least two events, but more if there are errors on the submitted profile page, for example, forgetting to agree to Mozilla's terms and privacy policy). So I'd recommend replacing the code above with the following changes (which I tested locally):
```python
    def get(self, request, *args, **kwargs):
        if request.session.get('sociallogin_provider'):
            track_event(
                CATEGORY_SIGNUP_FLOW,
                ACTION_PROFILE_AUDIT,
                request.session['sociallogin_provider'])
        return super().get(request, *args, **kwargs)


signup = redirect_in_maintenance_mode(SignupView.as_view())
```
so restoring the `signup` function to its original state, and adding a `get` method to the `SignupView` class so we only trigger the `profile-audit` event on GET.

peterbe:
@escattone r? again

So, one thing I noticed was that the ACTION_PROFILE_AUDIT was never covered in tests. Or perhaps, I'm misremembering that wrong. Anyway, check out the new `test_signup_github_event_tracking` test. It used to be that it did the `self.github_login` before the `settings.GOOGLE_ANALYTICS_ACCOUNT='UA-XXXX-1'` gets set up. Now, if you look at this test, it does a lock of mock patching, but it does run the whole gambit. It triggers all of these:

* `ACTION_AUTH_STARTED`
* `ACTION_AUTH_SUCCESSFUL`
* `ACTION_PROFILE_AUDIT`
* `ACTION_PROFILE_CREATED`
* `ACTION_FREE_NEWSLETTER`

It seems a bit nuts that we have to do 5 external network calls just for one single user journey but signing in and having to create a profile is pretty darn rare.

kuma/users/signal_handlers.py
```diff
@@ -1,23 +1,73 @@
...
+    authenticated with a provider but not necessarily fully logged in on
+    our site. For example, if the user hasn't created a profile (e.g.
+    agreeing to terms and conditions) the 'user_logged_in' won't fire until
+    then.
+    """
+    # If the user did the "social_auth_add" they already logged in and
+    # all we needed to do was to "combine" their user (from two different
+    # providers) can be reduce to the same.
+    if getattr(request, 'social_auth_added', False):
+        return
```
escattone: This is the right thing in the wrong place. It should be moved to the beginning of the `on_user_signed_up()` handler immediately below.

kuma/users/signal_handlers.py
```diff
@@ -1,23 +1,73 @@
...
+    """
+    Signal handler to be called when a given user has at least successfully
+    authenticated with a provider but not necessarily fully logged in on
+    our site. For example, if the user hasn't created a profile (e.g.
+    agreeing to terms and conditions) the 'user_logged_in' won't fire until
+    then.
+    """
+    # If the user did the "social_auth_add" they already logged in and
+    # all we needed to do was to "combine" their user (from two different
+    # providers) can be reduce to the same.
```
escattone: This comment reads strangely here at the end.

kuma/users/signal_handlers.py
```diff
@@ -27,6 +77,26 @@ def on_user_signed_up(sender, request, user, **kwargs):
             )
 
 
+@receiver(user_logged_in, dispatch_uid='users.user_logged_in')
+def on_user_logged_in(sender, request, user, **kwargs):
+    # We've already recorded that the have signed up. No point sending one
```
escattone:
```suggestion
    # We've already recorded that they have signed up. No point sending one
```

PR #6313, opened 2020-01-08
Update search index instructions

mindy:
Update instructions per our Slack convo: https://mozilla.slack.com/archives/CLTBV3W9X/p1578358816108100

### DIFF
--- a/docs/elasticsearch.rst
+++ b/docs/elasticsearch.rst
@@ -37,20 +37,14 @@ Using the Admin
 This process works both in a development environment and in production:
 
 - Open the Django admin search index list view. On Docker, this is located
-  at http://localhost:8000/admin/search/index/.
+  at http://localhost.org:8000/admin/search/index/.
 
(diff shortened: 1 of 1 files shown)


docs/elasticsearch.rst
```diff
@@ -37,20 +37,14 @@ Using the Admin
...
 
-- On the search index list, select the checkbox next to the newly created index
-  (the top most) and select "Populate selected search index via Celery" from
-  the Action dropdown menu. Click "Go" to start indexing asynchronously.
+- Inside the ``web`` Docker container::
 
-- In the development environment, you can watch the indexing process with::
-
-    docker-compose logs -f elasticsearch worker
+    ./manage.py reindex
```
peterbe: I'd rather use the full command so it's easy to copy-n-paste. I only ever use a bash shell when I'm debugging. I think you can just...
```suggestion
    docker-compose exec web ./manage.py reindex
```

PR #6330, opened 2020-01-10
Add newsletter subscription toggle to signup

### LINKED ISSUE #5792: Implement Newsletter integration to track opt-ins (without Salesforce for now)
**Parent Issue:**
https://github.com/mdn/sprints/issues/2044

**Email Design comps are here:**
https://github.com/mdn/sprints/issues/1814

**Technical Overview**:

We'll be using the Mozilla Basket API to implement this:
https://github.com/mozmeao/basket

We do not have to support Subscription management. Once a user has subscribed they can unsubscribe via a link at the bottom of the email. 

Contact @tobinmori for Dev API URL and credentials. 

**Tasks & Acceptance criteria:**

- asked @atopal for context as to where the subscription actually occurs.
- create an API endpoint in Kuma that (that uses our own Basket API key) wraps Basket's subscription creation
- no need to handle unsubscribe (this will be handled by user via unsubscribe link at bottom of email)
- need to get API key for staging to test
- be aware that Staging Basket API subscriptions go to prod Salesforce (so actual Salesforce subscriptions are created)

Gregoor:
Fixes #5792 
I mean it doesn't literally, because that issue is a about Salesforce integration which we're foregoing for now. Shall we rename it @tobinmori?

What this PR has in common with its prequel (#6141) is that it adds a newsletter subscription toggle to the sign-up form. What it does differently is that it's only persisting that decision in the DB and not doing any basket/salesforce sign-up.

Once we have figured out the salesforce stuff, we'll have to make sure to import every user who already opted-in.


tobinmori:
thanks @Gregoor. I renamed the parent ticket to be more accurate

Issue #6307, opened 2020-01-06
interactive examples are broken (empty) on "open in new tab"


wbamberg:
**UPDATE (Jan 10, peterbe)** 
Let's make this issue exclusively about the JavaScript `TypeError` that only happens when using "open in new tab". The problem with the line number gutter is for #6310. 

---

* Visit a page, say: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

* Right-click a link to a docs page from the sidebar, say https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asIntN, and select "Open Link in New Tab"

* Switch to the new tab.

In Chrome 80, I get an interactive example where the line numbers overlap the example:

<img width="800" alt="Screen Shot 2020-01-06 at 2 45 51 PM" src="https://user-images.githubusercontent.com/432915/71854463-193d6080-3093-11ea-9cd9-20c623bd9c2c.png">

In Firefox Nightly (73, adblockers disabled), I get an empty interactive example:

<img width="844" alt="Screen Shot 2020-01-06 at 2 48 39 PM" src="https://user-images.githubusercontent.com/432915/71854605-7cc78e00-3093-11ea-8446-3f11932c5fd9.png">

In both these cases, reloading the page fixes it.

In Firefox release (71) it works as expected.

peterbe:
I can't reproduce. In either browser. It seems sporadic. Perhaps there's some other JS error that's blocking something. Are you saying it's only happening after it has been opened in a new tab? If you can reproduce it, can you open the Web Console and share what it says. Or the Browser Console even?

I have a vague theory. We have a bunch of badly configured JS that's set to execute `async` which means that if the browser cache is hot, some of these scripts might fire too early. 

With regards to the messed up line numbers, I think we can focus on https://github.com/mdn/kuma/issues/6310 for that.

peterbe:
Possibly related: https://bugzilla.mozilla.org/show_bug.cgi?id=1607442

wbamberg:
**For the Chrome problem:**

https://github.com/mdn/kuma/issues/6310 looks totally like a duplicate of the Chrome problem I'm talking about here. I get this all the time when opening in a new tab, and never/hardly ever when I just click a link.

I don't see anything interesting in the Chrome console:

[...]

PR #6336, opened 2020-01-13
Issue #6307, interactive examples broken, Failed to execute measure

schalkneethling:
@peterbe This is part 1 of the items mentioned here:
https://github.com/mdn/kuma/issues/6307#issuecomment-573764829

Related to this error:

```
perf.654b849a6fd9.js:1 Uncaught DOMException: Failed to execute 'measure' on 'Performance': The mark 'interactive-editor-loading' does not exist.
    at Object.setMeasure (https://developer.mozilla.org/static/build/js/perf.654b849a6fd9.js:1:703)
    at handlePerfMarks (https://developer.mozilla.org/static/build/js/perf.654b849a6fd9.js:1:948)
    at perfMsgHandler (https://developer.mozilla.org/static/build/js/perf.654b849a6fd9.js:1:1557)
setMeasure @ perf.654b849a6fd9.js:1
handlePerfMarks @ perf.654b849a6fd9.js:1
perfMsgHandler @ perf.654b849a6fd9.js:1
perf.654b849a6fd9.js:1 Uncaught DOMException: Failed to execute 'measure' on 'Performance': The mark 'interactive-editor-loading' does not exist.
    at Object.setMeasure (https://developer.mozilla.org/static/build/js/perf.654b849a6fd9.js:1:703)
    at handlePerfMarks (https://developer.mozilla.org/static/build/js/perf.654b849a6fd9.js:1:948)
    at perfMsgHandler (https://developer.mozilla.org/static/build/js/perf.654b849a6fd9.js:1:1557)
setMeasure @ perf.654b849a6fd9.js:1
handlePerfMarks @ perf.654b849a6fd9.js:1
perfMsgHandler @ perf.654b849a6fd9.js:1
```

### DIFF
--- a/kuma/static/js/utils/perf.js
+++ b/kuma/static/js/utils/perf.js
@@ -47,10 +47,15 @@ mdn.perf = {
             return;
         }
 
-        performance.measure(
-            measureData.measureName,
-            measureData.startMark,
-            measureData.endMark
-        );
+        try {
+            performance.measure(
+                measureData.measureName,
+                measureData.startMark,
+                measureData.endMark
+            );
+        } catch (error) {
+            console.error('Error while setting performance measure: ', error);
+        }
+
     }
 };


kuma/static/js/utils/perf.js
```diff
@@ -47,10 +47,15 @@ mdn.perf = {
...
-            measureData.endMark
-        );
+        try {
+            performance.measure(
+                measureData.measureName,
+                measureData.startMark,
+                measureData.endMark
+            );
+        } catch (error) {
+            console.error('Error while setting performance measure: ', error);
```
peterbe:
```suggestion
            console.error(`Error while setting performance measure: ${error.toString()}`);
```
Your code is fine but I wanna take this opportunity to better understand Interactive Examples and to carefully see if this change fixes things.

peterbe:
I'm starting to have doubts about this. The only point of using `try/catch` is if you want to prevent the browser to give up on the rest of the file. If you have:
```html
<script>throw new Error('something')</script>
<script src="foo.js"></script>
```
the browser will happily download, parse, and execute `foo.js` even though an unexpected error was thrown/crashed in the inline scripts. If you have this:

```html
<script>throw new Error('something')</script>
<script>alert('This will still run')</script>
```
If you run that, you'll get an error in your Browser Console **and** you'll get an Window Alert box that appears. I.e. the second `<script>` block will go ahead. 
See: https://innovative-newsboy.glitch.me/

In other words, `try/catch` help us for anything outside the performance measurement stuff. 

The only time you should use try/catch is when there's code that you want should run AFTER something can go wrong. E.g. 
```html
<script>
try {   throw new Error('something')  }
catch (err) { console.log('That sucked') }
alert("This will run")
</script>
```
See: https://materialistic-passenger-77sp7p6bl.glitch.me/

So, the point is, I see no point trying to suppress any crashes within that performance stuff. It's fundamentally wrong to use `try/catch` to hope that it'll make "other things" work better.

kuma/static/js/utils/perf.js
schalkneethling: Also, I would love to use JS template strings, but our pipeline does not allow us at the moment. If you run `build-static` you will see it throws an error:

```
pipeline.exceptions.CompressorError: b"Parse error at -:57,26\n            console.error(`Error while setting performance measure: ${error.toSt\n                          ^\nSyntaxError: Unexpected character '`'\n    at JS_Parse_Error.get (eval at <anonymous> (/tools/node_modules/uglify-js/tools/node.js:27:1), <anonymous>:86:23)\n    at /tools/node_modules/uglify-js/bin/uglifyjs:384:40\n    at time_it (/tools/node_modules/uglify-js/bin/uglifyjs:620:15)\n    at /tools/node_modules/uglify-js/bin/uglifyjs:345:9\n    at Socket.<anonymous> (/tools/node_modules/uglify-js/bin/uglifyjs:601:13)\n    at Socket.emit (events.js:215:7)\n    at endReadableNT (_stream_readable.js:1184:12)\n    at processTicksAndRejections (internal/process/task_queues.js:80:21)\n"
```

schalkneethling:
> It's fundamentally wrong to use try/catch to hope that it'll make "other things" work better.

For sure but, I did not do this just because I was hoping that it will solve the problem with interactive examples. I also did it so we have a little more context around the error when it happens inside our minified/uglified code.

peterbe:
> so we have a little more context around the error when it happens inside our minified/uglified code.

What more would we find out? I would argue that the "natural crash" is better than attempting to put it into `console.error`.

schalkneethling:
> > so we have a little more context around the error when it happens inside our minified/uglified code.
> 
> What more would we find out? I would argue that the "natural crash" is better than attempting to put it into `console.error`.

Fair enough, so you reckon we should leave this as is?

peterbe:
That is my review. That I think we should close this.

PR #6346, opened 2020-01-15
Remove structure.scss from documents.scss 

### LINKED ISSUE #6255: _nav-access.scss is included twice
At the time of writing https://developer.mozilla.org/static/build/styles/react-mdn.6ee517b0c3a5.css looks like this if you pretty-print it: 
https://gist.github.com/peterbe/58d13ed197fa5f05fc93def81f432606
(search for `#nav-access {` and you'll see it appears twice)

A tool that minifies won't be able to clean this because the order might matter I suppose. 

It's included [here](https://github.com/mdn/kuma/blob/a3d6cb8e7314b70e6c07a54a7c8d86d6c7682448/kuma/static/styles/components/_structure.scss#L17) and [here](https://github.com/mdn/kuma/blob/a3d6cb8e7314b70e6c07a54a7c8d86d6c7682448/kuma/static/styles/minimalist/base/_structure.scss#L17)

mindy:
`structure.scss` was being included in two places: `documents.scss` and `main.scss`. Both files are included in `react-mdn.scss`. 

This change removes `structure.scss` from `documents.scss`, so it will only live in `main.scss`. 

Closes #6255

### DIFF
--- a/kuma/static/styles/minimalist/document.scss
+++ b/kuma/static/styles/minimalist/document.scss
@@ -25,7 +25,6 @@ Utitlity classes (reused in a variety of contexts)
 /*
 Components that may appear on most wiki pages
 ====================================================================== */
-@import './base/_structure.scss';
 @import '../components/wiki/_wiki-block.scss';
 @import '../components/wiki/_quick-links.scss';
 @import '../components/wiki/_toc.scss';


peterbe:
@mindy If you know any good linting tools or whatever that can automate this checking, then I think we should definitely add that to the CI. 

One idea I had was to write a CSS parser (well, *use* a CSS parser to be exact) and have it loop over the CSS selectors one by one and if it finds a selector + body that has already been seen, it's considered a duplicate and it should warn.

mindy:
@peterbe Initially I wanted to adopt a few plugins I used in the past that worked well-- [Mini CSS Extract](https://www.npmjs.com/package/mini-css-extract-plugin) and [Optimize CSS Assets](https://www.npmjs.com/package/optimize-css-assets-webpack-plugin) (config was copied almost verbatim from Create React App if you wanted to see an [example](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js)). But then I saw that our CSS is loaded through the Django pipeline and using these plugins would require that our CSS is imported in JS files and processed through Webpack. 

Tobin tried to explain how the pipeline works, but I didn't feel comfortable right now changing up how our CSS is loaded, and wasn't sure if it was worth it since there's Yari. I'm happy to look into it if it's something we want to prioritize!

peterbe:
I don't know if you already knew this but we won't be able to (ever) get sourcemaps working with our React and SCSS code. Between the raw ES6 .jsx files and the user sits Webpack (which includes Babel), and then `django-pipeline`. So, there's a Webpack process that turns `*.jsx` into `dist/react.js` and `dist/react.js.map`. THen `django-pipeline` takes over and bundles that `dist/react.js` with a bunch of other .js files into `build/js/react-main-d91fe101a.js` and that file is all concatenated and minified and would never map back to the `react.js.map` file. 

That's the JS side but I think we're up against something similar with SCSS and stuff. 
There are less tools for SCSS compared to CSS, but linting the CSS might be silly because it's impossible to learn from it. 
However, if you can think of some/any tools that can help a little we're all ears. For example, what would have immediately caught the understandable mistake of accidentally including that `_structure.scss` twice? 

`stumptown-renderer` is all different. We haven't decided exactly how we want to do CSS but I think the team is generally in favor of something like SCSS instead of css-in-js but there are many nuances to that equation. We can also, for example, do a general grid/normalize/base CSS (from a SCSS file) but have individual extra CSS within the component as css-in-js or something. 

One thing we're certain of, for `stumptown-renderer` is that we must have a good story for smart code splitting so that you can freely type whatever in `stylesexclusivetoonesinglepage.(s)css` without having to worry about bloating all other pages. 

CC @Gregoor

schalkneethling:
I would also not invest too much time on this inside the current Kuma codebase and instead focus our efforts on making sure all of this is baked into `mdn-minimalist` and `Yari`

peterbe:
> I would also not invest too much time on this inside the current Kuma codebase and instead focus our efforts on making sure all of this is baked into `mdn-minimalist` and `Yari`

You're absolutely right. But, let me plant one little idea in your heads to consider...
It might make sense to not have the same CSS authoring experience on the base as you do on "components". 
Similarly, how some project use `bootstrap.min.css` and `normalize.css` but then *within* the page they use `styled-components` or `emotion`. We might want to consider doing something like that. 
I.e. split it distinctly into two separate modules that don't cross paths. One stylesheet solution for the basic rendering and a different solution for stylesheets that have more "page specific" jobs.

PR #6352, opened 2020-01-17
Add tests for signin landing page

### LINKED ISSUE #6024: Sign-In Landing page

schalkneethling:
@peterbe This adds some basic tests for the new signin landing page r?

fix #6024

### DIFF
--- a/kuma/users/tests/test_templates.py
+++ b/kuma/users/tests/test_templates.py
@@ -22,23 +22,6 @@ from . import SampleRevisionsMixin, SocialTestMixin, UserTestCase
 from ..models import User, UserBan
 
 
-def test_signin_landing(db, client, settings):
-    settings.MULTI_AUTH_ENABLED = True
-    response = client.get(reverse('socialaccount_signin'))
-    github_login_url = '/users/github/login/'
-    google_login_url = '/users/google/login/'
-    # first, make sure that the page loads
-    assert response.status_code == 200
-    doc = pq(response.content)
-    # ensure the title is accurate
-    assert 'Create your MDN Web Docs Account' in doc.find('h1').text()
-    # ensure that both auth buttons are present
-    assert doc('.auth-button-container a').length == 2
-    # ensure each button links to the appropriate endpoint
-    assert doc('.github-auth').attr.href == github_login_url
-    assert doc('.google-auth').attr.href == google_login_url
-
-
 class SignupTests(UserTestCase, SocialTestMixin):
     profile_create_strings = (
         'Create your MDN profile',
--- a/kuma/users/tests/test_templates.py
+++ b/kuma/users/tests/test_templates.py
@@ -22,6 +22,22 @@ from . import SampleRevisionsMixin, SocialTestMixin, UserTestCase
 from ..models import User, UserBan
 
 
+def test_signin_landing(db, client):
+    response = client.get(reverse('socialaccount_signin'))
+    github_login_url = '/users/github/login/'
+    google_login_url = '/users/google/login/'
+    # first, make sure that the page loads
+    assert response.status_code == 200
+    doc = pq(response.content)
+    # ensure the title is accurate
+    assert 'Create your MDN Web Docs Account' in doc.find('h1').text()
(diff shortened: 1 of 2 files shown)


kuma/users/tests/test_templates.py
```diff
@@ -22,6 +22,23 @@
...
+def test_signin_landing(db, client, settings):
+    settings.MULTI_AUTH_ENABLED = True
+    response = client.get(reverse('socialaccount_signin'))
+    github_login_url = '/users/github/login/'
+    google_login_url = '/users/google/login/'
+    # first, make sure that the page loads
+    assert response.status_code == 200
+    doc = pq(response.content)
+    # ensure the title is accurate
+    assert 'Create your MDN Web Docs Account' in doc.find('h1').text()
```
peterbe: The title is not dynamic based on some business logic. I think this assertion should be avoided because generally text should be avoid to be tested unless the test is around some important phrasing that might or might not show up.

kuma/users/tests/test_views.py
```diff
@@ -1602,3 +1602,16 @@ def test_invalid_uid_fails(wiki_user, client):
...
     assert_no_cache_header(response)
     assert b'This link is no longer valid.' in response.content
+
+
+def test_signin_landing(db, client, settings):
+    settings.MULTI_AUTH_ENABLED = True
+    response = client.get(reverse('socialaccount_signin'))
+    assert response.status_code == 200
+    doc = pq(response.content)
+    assert 'Create your MDN Web Docs Account' in doc.find('h1').text()
```
peterbe: How is this test different from the one in `test_templates.py`?

Actually, I'm not entirely sure what `test_TEMPLATES.py` mean because what really matters is what users SEE and that API is the view. Not the template. 
I'd vote you *move* that nice `test_signin_landing` into this file here.

kuma/users/tests/test_views.py
```diff
@@ -1602,3 +1602,16 @@ def test_invalid_uid_fails(wiki_user, client):
...
+
+
+def test_signin_landing_multi_auth_disabled(db, client):
```
peterbe: At some point soon, we're going to switch the default of `MULTI_AUTH_ENABLED` to true. It's false right now just because of timing and us being careful. 
So, I'd put in a `settings.MULTI_AUTH_ENABLED = False` on the first line. It reads better that way.

kuma/users/tests/test_templates.py
schalkneethling: Yeah, I was not sold on doing this but, I saw some other tests were testing similar scenarios so I through it in there.

kuma/users/tests/test_views.py
schalkneethling: ¯\_(ツ)_/¯ - I was doing some searching around the tests and stumbled on one of these `test_template` files. My thinking then was, perhaps the `view` tests, in this case, should just test that the appropriate response is received based on the `MULTI_AUTH_ENABLED` setting.

The actual specifics of what should be present in the page is then tested in the `template` tests. Perhaps these are just a holdover or are used for specific instances. Not sure. I will move the contents of that test here.

schalkneethling:
@peterbe Thank you for the review. Updated.

PR #6356, opened 2020-01-17
Limit wiki syntax highlighting to #wikiArticle

### LINKED ISSUE #6355: Syntax Highlighting shouldn't be applied while localizing
**Summary**
Wiki:
While localizing a doc with code blocks, English document ("_origin_") gets syntax highlighted.

This behaviour complicates localization as one can't simply copy and paste said code blocks. All those Prism `<span>`s gets copied to the CKEditor. ("_target_")

Note: Edit page doesn't even have any Prism stylesheets.

**Steps To Reproduce (STR)**

1. Open any doc with one or more example code blocks.
2. Go to any localized versions. Press edit and copy one of the code blocks from _origin_.
3. Paste to _target_ doc. View _target_ doc as source.

**Actual behavior**
Unnecessary `<span>`s are added to _target_ source.
주석 2020-01-18 020836
주석 2020-01-18 020836-2

**Expected behavior**
Prism not applied to _origin_.

alattalatta:
Fixes #6355

This should limit unnecessary highlighting applied while localizing a document with code examples.

It will also reduce this left padding below, which is for line numbers.

주석 2020-01-18 022925

### DIFF
--- a/kuma/static/js/wiki.js
+++ b/kuma/static/js/wiki.js
@@ -62,7 +62,7 @@
     /*
         Syntax highlighting scripts
     */
-    if ($('article pre').length && ('querySelectorAll' in doc)) {
+    if ($('#wikiArticle pre').length && ('querySelectorAll' in doc)) {
         if (mdn.assets && mdn.assets.js.hasOwnProperty('syntax-prism')) {
             (function() {
                 mdn.assets.js['syntax-prism'].forEach(function(url, index, array) {


peterbe:
Do you have a URL to demonstrate this on?

peterbe:
Yeah, you end up with HTML like this:
Screen Shot 2020-01-17 at 4 37 26 PM

When it should be more like this:
Screen Shot 2020-01-17 at 4 38 09 PM

Your fix fixed it.

PR #6358, opened 2020-01-19
Fix gap left from Edge Mobile in compat data table

vinyldarkscratch:
Last year, BCD had made the decision to remove Edge Mobile from the compatibility data tables, as a) the platform it runs on is deprecated, and b) many users have been confused about whether it referred to the Windows Mobile version (the correct answer), or the iOS/Android versions.  We have already purged Edge Mobile from the entire repository, however a small column-sized gap was left in the table:

Screen Shot 2020-01-18 at 18 35 58

This PR aims to fix this styling issue by reducing the column count by one for the "web" and "js" table types (essentially, the two table types that initially displayed Edge Mobile):

Screen Shot 2020-01-18 at 18 40 15

### DIFF
--- a/kuma/static/styles/components/compat-tables/_vars.scss
+++ b/kuma/static/styles/components/compat-tables/_vars.scss
@@ -43,12 +43,12 @@ $bc-mobile-browser-width: 220px;
 
 $table-types: (
 	web: (
-		size: 13,
+		size: 12,
 		feature-name-width: 200px,
 		platform-border: (8)
 	),
 	js: (
-		size: 14,
+		size: 13,
 		feature-name-width: 200px,
 		platform-border: (8, 15)
 	),


peterbe:
@vinyldarkscratch Pardon my ignorance but how does this correlate to pages whose BCD table includes NodeJS?
Like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#Browser_compatibility

vinyldarkscratch:
Edge Mobile was also included in JavaScript tables, so there's a gap present on them as well:

image

(Note: you can see the gap if you resize your browser window to be between 768px and 1200px -- smaller and larger screens render the table fine.)

peterbe:
How do I reproduce the problem? I clicked around bunch of pages to see the wide gap which I then would expect to disappear when I try your git branch.

vinyldarkscratch:
To reproduce:

- Head over to any page that contains a BCD table (aside from webextensions)
- Resize your browser so its width is in between 960px and 1200px
- You should see a small column-sized gap in the table at the end

PR #6360, opened 2020-01-20
Switch to async GA tag

### LINKED ISSUE #6253: P1 - Switch to async GA tag
From https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tag

*"The alternative async tag below adds support for preloading, which will provide a small performance boost on modern browsers, but can degrade to synchronous loading and execution on IE 9 and older mobile browsers that do not recognize the async script attribute. Only use this tag configuration if your visitors primarily use modern browsers to access your site."*

That's us!

So, instead of 
```
<script>
(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
</script>
```

we switch to 
```
<script>
window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
</script>
<script async src='https://www.google-analytics.com/analytics.js'></script>
```
which means that when the browser sees that second script tag it can start to download the file before it starts executing the (second) script tag.

mindy:
Not sure if this is the best way, but ended up having to insert this dynamically because I wanted to respect `Mozilla.dntEnabled()`.

Closes #6253

### DIFF
--- a/jinja2/includes/google_analytics.html
+++ b/jinja2/includes/google_analytics.html
@@ -5,13 +5,15 @@
     {%- set GA_FILE = 'analytics_debug' if settings.DEBUG else 'analytics' %}
     // only load GA if DNT is not enabled
     if (Mozilla && !Mozilla.dntEnabled()) {
-        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-        })(window,document,'script','https://www.google-analytics.com/{{ GA_FILE }}.js','ga');
-
+        window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
         ga('create', '{{ settings.GOOGLE_ANALYTICS_ACCOUNT }}', 'mozilla.org');
         ga('set', 'anonymizeIp', true);
+
+        const gaSrc = 'https://www.google-analytics.com/{{ GA_FILE }}.js';
+        const gaScript = document.createElement('script');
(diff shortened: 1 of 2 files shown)


peterbe:
This is not my review but I do wonder if this PR is "missing the point" :)
The difference between:
```html
<script async src="/foo.js"></script>
```
and 
```html
<script>
var s = document.createElement('script')
s.src = '/foo.js'
s.async = true
document.head.appendChild(s);
</script>
```
is that the browser can't see, before it starts executing things, that it's *going* to need to download `$currenurl/foo.js`. 

Browsers will look over the HTML it's about to execute. If the HTML references things it can start a thread to prepare the download of those things so that when it comes time to "execute the HTML" some of those things that will be needed might already have been downloaded and prepared. 

In the second example, where it runtime injects a script tag dependency, that downloading and parsing won't start to happen until all the HTML has been fully parsed and the browser is now "pre-rendering" the page. 

Would you mind exploring this a little deeper? I love the intent that we do Google Analytics the right way so that DNT is fully respected. It's honestly a topic I'm less confident around. 

But it's important to get these bits right because the correct way to do GA tracking is very important to us. And the day we switch away from Jinja templates and have a more "jamstack" platform we won't be able to use runtime server variables in the script code.

peterbe:
To follow up from Slack; if you wrap the `ga('create', '{{ settings.GOOGLE_ANALYTICS_ACCOUNT }}', 'mozilla.org')` line with a `dntEnabled` check, you get the best of both worlds.

PR #6369, opened 2020-01-22
Remove focus outline on modal

### LINKED ISSUE #6339: Hide focus ring on modal - Auth
In general, I would oppose removing the focus ring that browsers add to the item that is currently focused or active but, in the case of the modal "window" itself when opened, I would not be so opposed ;p

With that said, we should use https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible so that in supporting browsers at least, the focus ring would still be drawn for keyboard users.

mindy:
Closes #6339

### DIFF
--- a/kuma/static/styles/minimalist/components/_auth-providers.scss
+++ b/kuma/static/styles/minimalist/components/_auth-providers.scss
@@ -6,6 +6,15 @@
         margin: 60px auto;
     }
 
+    &:focus {
+        outline-style: none;
+    }
+
+    &:focus-visible {
+        outline: 5px auto #e59700; // linux and windows
+        outline: 5px auto -webkit-focus-ring-color; // everyone else
+    }
+
     .auth-button-container {
         margin: 0 auto;
         width: 350px;
--- a/kuma/static/styles/minimalist/components/_auth-providers.scss
+++ b/kuma/static/styles/minimalist/components/_auth-providers.scss
@@ -12,7 +12,7 @@
 
     &:focus-visible {
         outline: 5px auto #e59700; // linux and windows
-        outline: 5px auto -webkit-focus-ring-color; // everyone else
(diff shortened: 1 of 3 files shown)


kuma/static/styles/minimalist/components/_auth-providers.scss
```diff
@@ -6,6 +6,15 @@
...
     }
 
+    &:focus {
+        outline-style: none;
+    }
+
+    &:focus-visible {
+        outline: 5px auto #e59700; // linux and windows
+        outline-color: -webkit-focus-ring-color; // everyone else
+    }
```
schalkneethling: Thanks so much for doing this @mindy. In this instance, perhaps it is better to just remove that outlined style on the modal entirely. For one thing, the browser support for `:focus-visible` is not great yet https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#Browser_compatibility, although generally I would be in favor of adding it anyhow.

Perhaps it is not needed on the modal itself as much seeing that the interaction flow suggested that, that is where the focus is moved. The other thing is `stylelint` does not know about the pseudo class(yet) and so linting fails(https://travis-ci.org/mdn/kuma/jobs/640659152#L293). We can get around that by disabling `stylelint` for this block, but perhaps all of that is simply not needed.

We do have proper focus states on the buttons as well as the close icon when they receive focus, and once the modal is open, tab order is trapped inside the modal so, we should be good.

What are your thoughts?

mindy: I think that's a good call, especially since, as you mentioned, we have proper focus states on the elements within the modal. 

I upgraded stylelint to the latest version (13.0.0) so we should be okay (at least running it locally) if we do want to add the focus-visible class in the future!

PR #6371, opened 2020-01-23
Do not show avatar transition on signup

### LINKED ISSUE #6367: Mugshot only shown during submission
To explain what's happening...
When I press "Create Profile" it starts submitting the form and you can't see it (in this screencap) but the browser indicates that it's loading. 
Right after pressing that button, the dino head changes to my mugshot from my Google account. Just for a brief moment until the form realizes I didn't fill it in properly so it needs to reload but with the form validation errors highlighted in red. 
Screen-Recording-2020-01-22-at-4 08 00-PM

Perhaps *my* face is NOT supposed to appear on the sign up landing page at all. And only for that split second when I'm submitting form. If so, it's weird because it flashes.

schalkneethling:
Fix #6367


schalkneethling:
@peterbe Will update this based on your last comment on the issue. I guess the current implementation was based on the original design that showed the Auth providers logo used for the image of step 2. It does make sense to just show your avatar when we have it, and we do have it.

peterbe:
I'm not sure what's supposed to happen. This is what I see after I come back from Google's Account page:
Screen Shot 2020-01-24 at 2 11 04 PM

Still no checkmark or my Gmail mugshot.

peterbe:
If I press the "Create Profile" button this happens:
Screen Shot 2020-01-24 at 2 13 01 PM

The checkbox got added on the dino icon. No idea what that means (... from a user point of view).

peterbe:
Ultimately, I think I should see my Gmail mugshot. That would be a nice indicator demonstrate that my Google authentication at least recognized me.

schalkneethling:
> ACTUALLY! I wonder if the bug I saw was related to this change. Perhaps not but I'd rather be careful.

How is it that you have not username? That is problematic though for sure.

schalkneethling:
> How is it that you have not username?

I know the answer to this now :)

schalkneethling:
@peterbe updated.

schalkneethling:
Also addresses items in https://github.com/mdn/kuma/issues/6370

schalkneethling:
@peterbe We good here?

kuma/static/js/components/user-signup/signup.js
```diff
@@ -4,74 +4,73 @@
...
-            mdnProfileImgContainer.classList.add('checked');
-            mdnProfileImg.classList.add('animate');
-
-            mdnProfileImg.onanimationend = function() {
-                form.submit();
-            };
-        });
+        for (var i = 0, l = errors.length; i < l; i++) {
+            errors[i].classList.add('hidden');
+        }
```
peterbe: Man, I miss jQuery. But you should probably just do `errors.forEach(function(error) { error.classList.add('hidden') }` instead of this loop.

peterbe: Actually. Ignore that. It wouldn't really change anything.

kuma/users/jinja2/socialaccount/includes/change_email_address.html
```diff
@@ -1,4 +1,5 @@
 <div id="change-email-container" class="hidden">
     <label for="{{ form.email.id_for_label }}" class="visually-hidden">{{ default_email }}</label>
     {{ form.email }}
+    <div id="{{ form.email.id_for_label }}-error" class="error hidden">{{ _('The email field is required.') }}</div>
```
peterbe: I don't understand why this is needed. First of all, doesn't the `{{ form.email }}` produce something like `<input name="email" required ...>` so that should take care of protection against omission. 
Second, if the user did type something that gets sent (e.g. `''` or `'usernamethatsalreadytaken'`) the `{{ form.email }}` 's own error handling should be able to display that.

schalkneethling:
> I'm sorry for being an ass but what does even mean that the Google logo becomes greyscale?
> The little green checkbox feels like an accomplishment, but I'm still not seeing the point of making the Google logo greyscale.
> 
> 1. Sign in and arrive on the "Create profile" page.
> 2. Click "change username"
> 3. Change the username to an empty string.
> 4. Click "Create profile"
> 
> This happens:
> Screen Shot 2020-02-07 at 12 25 22 PM
> 
> I.e. the whole username piece disappears.
> 
> Here's a bug though.

The greyscale was a request from Kadir/original design that I overlooked originally:
https://github.com/mdn/kuma/issues/6370#issuecomment-578500625

> The little green checkbox feels like an accomplishment

Kinda why I suggested the copy under the logo should read `1. Authorized with Google` and not `1. Authorize with Google`, but is a minor detail perhaps.

kuma/users/jinja2/socialaccount/includes/change_email_address.html
schalkneethling: You are correct, and I was considering removing the email portion of the validation. Perhaps I was trying to be consistent or had an edge case in mind. That must have been a super sharp edge though. Let me think through this again and remove it if it serves no real purpose.

schalkneethling:
> Here's a bug though.

Dang it! :)

schalkneethling:
> * Sign in and arrive on the "Create profile" page.
> 
> * Click "change username"
> 
> * Change the username to an empty string.
> 
> * Click "Create profile"

@peterbe I tried these steps to reproduce but this is what I get:

<details>
<summary>Screen recording</summary>
create-account
</details>

peterbe:
I don't know what but I can't reproduce it myself either anymore.

kuma/users/jinja2/socialaccount/includes/change_username.html
```diff
@@ -1,4 +1,5 @@
 <div id="change-username-container" {% if is_hidden %}class="hidden"{% endif %}>
     <label for="{{ form.username.id_for_label }}" class="visually-hidden">{{ form.username.label }}</label>
     {{ form.username }}
+    <div id="{{ form.username.id_for_label }}-error" class="error hidden">{{ _('A username is required.') }}</div>
```
peterbe: I still don't get it with these. If I try to submit the form with an empty username this happens:
Screen Shot 2020-02-10 at 11 43 39 AM
So how do you possibly benefit from this line here?

schalkneethling: How do you even get into that state 🤣 Did you add `novalidate` to the form? Also, how is the descriptive copy below the input field red? Hmm actually, it seems like the label, (server validation)error message, input field, and the descriptive copy is all inside a container with an `error` class.

Could you add the steps to reproduce? Thanks, Peter

peterbe: The native web form `required` on the `<input>` is easy to bypass. Just type an empty string (or even many spaces to empty string) into the form and submit. That'll send it into Django and thankfully, somewhere in our Django form handling, we do `username = username.strip()` before the validation kicks in.

schalkneethling: Gotcha!

schalkneethling: Ok, this is resolved as well. I removed the `error` class that was adding to the containing list element so, the styling will not be correct when a server error is returned.

With regards to the JS and the error message you are referring to here, I added a `trim` to the string that I should have done from the start so, now it should let you submit an empty username field.

As for the reason this was all added. There is/was? a scenario where you can land on the signup page with an empty username. In those instances, clicking "create profile" just seemed to do nothing as the actual input field was hidden.

This ensures that the field is made visible and an error message is presented to the user.

schalkneethling:
@peterbe updated. I believe we are ready to roll 🎢

schalkneethling:
> I think a lot of these problems stem from the fact that the forms can either be fully hidden, half-hidden or fully visible depending on the user having clicked "change username/email". That's why we have these bipolar attempts at hiding and unhiding and adding and removing .error to class lists.

I agree and also think this might have been over-designed. Having the input field visible and editable by default makes it just so much easier to change the value if you care. If you don't, then whether it is displayed as plain text or not makes no difference.

> Like, why does there even need to be a hyperlink that isn't a hyperlink that I have to press just to see the username and email form? Why can't it just always be plainly visible?

By design.

> By the way, I dislike using links for what should be buttons. (I know, from inspecting the DOM that's actually a <button> styled to appear like a hyperlink). It makes me have to think "Oh, do I need to go to some other page to change my username/email? Why can't I do that here? Why can't I open that link in a new tab????!"

I agree and also have a strong preference for buttons being and looking like buttons. I have made some calls on other aspects but was asked(told) not to so...

> We still don't even know if https://bugzilla.mozilla.org/show_bug.cgi?id=1614273 is related to this

I believe a minor portion of it does actually. This PR has really gotten ugly though. At this point, I am thinking we should abandon it and I can split this out into a couple of smaller, more focused changes(user stories) and then open separate pull requests for those.

That way we can also question some of the design decisions. Let me know your thoughts on this.

schalkneethling:
As mentioned here[https://github.com/mdn/kuma/pull/6371#issuecomment-585575003] and then discussed on Slack, let's close out this PR and refocus around some smaller issues.

I will go ahead and create those and start the process of creating new pull requests for review. Thanks so much for all of your feedback here @peterbe

tobinmori:
@atopal and @malqinneh you might be interested in - there's some design feedback here that we can circle back to later, but in the interest of expediency we're closing this out, and we'll document those in a sep. ticket.

atopal:
@tobinmori who is creating those tickets? Can we link them from here?

schalkneethling:
> @tobinmori who is creating those tickets? Can we link them from here?

I am :) I will be sure to link them here. Below is the first one:

- https://github.com/mdn/kuma/pull/6487 for issue
- https://github.com/mdn/kuma/issues/6483

schalkneethling:
Auth provider logos should be grayscale
- https://github.com/mdn/kuma/issues/6500

tobinmori:
many thanks @schalkneethling!

schalkneethling:
And here are the last couple of spinoff issues:

- https://github.com/mdn/kuma/issues/6511
- https://github.com/mdn/kuma/issues/6512
- https://github.com/mdn/kuma/issues/6513
- https://github.com/mdn/kuma/issues/6514

Some require input others just need doing 😃

PR #6384, opened 2020-01-31
Update Jinja-related dependencies

callahad:
All changelogs indicate minor updates; should be good if tests pass.

Unpinned `markupsafe` from `pyproject.toml` since we don't use it
directly; it's only present as a `jinja2` dependency.

- Updating babel (2.2.0 -> 2.8.0)
  https://github.com/python-babel/babel/blob/v2.8.0/CHANGES

- Updating django-jinja (2.4.1 -> 2.5.0)
  https://github.com/niwinz/django-jinja/blob/2.5.0/CHANGES.adoc
  (commits at https://github.com/niwinz/django-jinja/compare/2.4.1...2.5.0)

- Updating jinja2 (2.10.1 -> 2.11.1)
  https://github.com/pallets/jinja/blob/2.11.1/CHANGES.rst

- Updating markupsafe (1.0 -> 1.1.1)
  https://github.com/pallets/markupsafe/blob/1.1.1/CHANGES.rst


callahad:
Updated to fix one test failure.

Babel now formats French decimals by using U+202F (NARROW NO-BREAK SPACE) for grouping instead of U+00A0 (NO-BREAK SPACE).

PR #6401, opened 2020-02-03
always clear cache between pytests

### LINKED ISSUE #6399: Cache is not cleared between tests
When working on https://github.com/mdn/kuma/pull/6143 I ran into a problem that tests failed or worked depending on the order you run them. 
The reason was that some tests make assumptions about the *state* of the cache. Most likely, the assumptions are that the cache is empty. But it can sometimes NOT be.

peterbe:
Fixes #6399

### DIFF
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -75,7 +75,7 @@ def test_doc_api_404(client, root_doc):
     assert_no_cache_header(response)
 
 
-def test_doc_api(client, trans_doc, cleared_cacheback_cache):
+def test_doc_api(client, trans_doc):
     """On success we get document details in a JSON response."""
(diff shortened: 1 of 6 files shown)


peterbe:
@Gregoor @escattone I actually only need 1 of you to review this one. But whoever doesn't review you probably should notice it. 

I cheated a little bit. I just happened to have noticed that the tests always assume that `settings.SITE_ID == 1` so if you put anything (like `SITE_ID=3`) in your personal `.env` file it would break the tests. Even on `master`. With this error:
```
allauth.socialaccount.models.DoesNotExist: SocialApp matching query does not exist.
```

I actually no longer remember which `pytest` would break if run in a particular order. Either way, this PR is good and the preconditions for tests should always be deterministically equal and not be at whim of happenstance.

escattone:
@peterbe Totally agree with @Gregoor, nice change in all respects! Thanks!

PR #6402, opened 2020-02-03
Fix hover styling on BCD signaling section

### LINKED ISSUE #6064: Hover styling should affect icon border width [BCD signalling]
On live, when a browser icon is hovered over or has focus, the border changes to MDN blue.

Screenshot 2019-10-17 at 10 24 47

Signifying focus with just a colour change isn't sufficient for accessibility so may we have the border change thickness similar to how it is on Zeplin:

Screenshot 2019-10-17 at 10 27 31

In the design the hover state changes the border form default `2px` width to `3px`, I'm anticipating that making this animate/transition may look a bit odd, so I'm happy to have no transition between these hover styles. Also happy to see `4px` if that works better!

mindy:
- Apply a thicker border width when hovering over browser icon

**Before:** 
hover-before

**After:** 
hover-after

Closes #6064

### DIFF
--- a/kuma/static/styles/components/compat-tables/_bc-signal.scss
+++ b/kuma/static/styles/components/compat-tables/_bc-signal.scss
@@ -463,9 +463,11 @@ div.bc-signal-block {
             transition: border-color .1s;
         }
 
-        &:hover {
+        &:hover:not(.selected) {
             .browser-logo {
                 border-color: #3d7e9a;
+                outline: 3px $blue solid;
+                outline-offset: -3px;
             }
         }
 
--- a/kuma/static/styles/components/compat-tables/_bc-signal.scss
+++ b/kuma/static/styles/components/compat-tables/_bc-signal.scss
@@ -11,7 +11,7 @@ div.signal-link-container {
     transition: all 1s ease-out;
 
     & > hr {
-        background-color: #3d7e9a;
+        background-color: $blue;
         margin: 0;
         max-width: 100%;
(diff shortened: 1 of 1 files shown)


schalkneethling:
@mindy I notice that the `kumascript` and `locale` submodule updates sneaked into the PR. Because this is such a small change, it is probably easiest to just do the following:

```
git checkout master
git branch -D bcd-border-6064
git pull origin master
git submodule update --init --recursive
git checkout -b bcd-border-6064

-- apply your change

git add .
git commit -m 'Message'
git push origin bcd-border-6064 (might/probably need a force push)
```

mindy:
Thanks @schalkneethling ! I followed your instructions and pushed up a new version without the submodules.

kuma/static/styles/components/compat-tables/_bc-signal.scss
```diff
@@ -463,9 +463,11 @@ div.bc-signal-block {
             transition: border-color .1s;
         }
 
-        &:hover {
+        &:hover:not(.selected) {
             .browser-logo {
                 border-color: #3d7e9a;
```
schalkneethling: While you are working on this file, mind changing this to

```suggestion
                border-color: $blue;
```

mindy: Thanks for the suggestion @schalkneethling! Did a search within the file and replaced all instances (only 4) with `$blue`.

PR #6423, opened 2020-02-05
use Authorization header for passing access token for GitHub

### LINKED ISSUE #6421: Use of access tokens as query parameters in the GitHub OAuth API is deprecated
**Summary**
We use GitHub's OAuth API as one of MDN's sign-up/sign-in OAuth providers. We're starting to receive emails from GitHub that using the https://api.github.com/user API with the access token as a query parameter has been deprecated and that the `Authorization` header should be used instead. This occurs within `GitHubOAuth2Adapter.complete_login` method provided by `django-allauth`, but `django-allauth` has not yet fixed this (although a PR has been submitted that does -- see https://github.com/pennersr/django-allauth/pull/2458). Even if `django-allauth` fixes the issue, it wouldn't help in our case since we override this method (https://github.com/mdn/kuma/blob/266bd9d8ebf24c950037a1965b1967022fca233f/kuma/users/providers/github/views.py#L20). We need to update our overridden method to pass the token via the `Authorization` header rather than via a query parameter.

**Rationale**
We no longer have to concern ourselves with using a deprecated approach.

**Audience**
All users who sign-up/sign-in to MDN via GitHub.

escattone:
Fixes #6421 

I tested this locally and it works great.

For additional reference, see https://github.com/pennersr/django-allauth/pull/2458

### DIFF
--- a/kuma/users/providers/github/views.py
+++ b/kuma/users/providers/github/views.py
@@ -19,11 +19,11 @@ class KumaGitHubOAuth2Adapter(GitHubOAuth2Adapter):
 
     def complete_login(self, request, app, token, **kwargs):
         session = requests_retry_session()
-        params = {'access_token': token.token}
-        profile_data = session.get(self.profile_url, params=params)
+        headers = {'Authorization': f'token {token.token}'}
+        profile_data = session.get(self.profile_url, headers=headers)
         profile_data.raise_for_status()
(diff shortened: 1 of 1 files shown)


peterbe:
Yay! Do you think we should stall a little bit until some of the maintainers of `django-allauth` get a chance to review that PR?

PR #6426, opened 2020-02-05
Fix dropdown menu position on RTL languages 

### LINKED ISSUE #6354: Language list dropdown is cut off for RTL languages
For example, https://developer.mozilla.org/ar/docs/Web :

Screen Shot 2020-01-17 at 9 10 32 AM

mindy:
- Move positioning of dropdown menu to SCSS file, so we can use the `bidi` mixin to handle LTR and RTL languages. 
- Affects both the User Avatar dropdown in the header and the language dropdown

**Sample URL:** 
http://localhost.org:8000/ar/docs/Web/CSS/:focus-visible

**Before:**
menu-before

**After:** 
menu-after

Closes #6354

### DIFF
--- a/kuma/javascript/src/header/__snapshots__/language-menu.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/language-menu.test.js.snap
@@ -23,14 +23,9 @@ exports[`LanguageMenu snapshot 1`] = `
     </span>
   </button>
   <ul
-    className="dropdown-menu-items"
+    className="dropdown-menu-items right"
     id="language-menu"
     role="menu"
-    style={
-      Object {
-        "right": 0,
-      }
-    }
   >
     <li
       lang="es"
--- a/kuma/javascript/src/header/__snapshots__/login.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/login.test.js.snap
@@ -23,13 +23,8 @@ exports[`Login component when user is logged in 1`] = `
       />
     </button>
     <ul
-      className="dropdown-menu-items"
+      className="dropdown-menu-items right"
       role="menu"
-      style={
-        Object {
-          "right": 0,
-        }
-      }
     >
       <li>
         <a
--- a/kuma/javascript/src/header/dropdown.jsx
+++ b/kuma/javascript/src/header/dropdown.jsx
@@ -69,8 +69,7 @@ export default function Dropdown(props: DropdownProps) {
             </button>
             <ul
                 id={props.ariaOwns}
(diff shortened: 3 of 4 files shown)


schalkneethling:
Hey @mindy, thank you for the PR. The Jest snapshot tests are failing because of the change in the `jsx` file. To update those snapshot tests, run the following and then add and commit the result:

```
npm t -- --u
```

https://jestjs.io/docs/en/snapshot-testing.html#updating-snapshots

mindy:
> Hey @mindy, thank you for the PR. The Jest snapshot tests are failing because of the change in the `jsx` file. To update those snapshot tests, run the following and then add and commit the result:
> 
> ```
> npm t -- --u
> ```
> 
> https://jestjs.io/docs/en/snapshot-testing.html#updating-snapshots

Thanks @schalkneethling ! Do you think it's a good idea to add `npm run test` to the precommit hook to catch this sooner?

schalkneethling:
> > Hey @mindy, thank you for the PR. The Jest snapshot tests are failing because of the change in the `jsx` file. To update those snapshot tests, run the following and then add and commit the result:
> > ```
> > npm t -- --u
> > ```
> > 
> > 
> > https://jestjs.io/docs/en/snapshot-testing.html#updating-snapshots
> 
> Thanks @schalkneethling ! Updated the snapshots. Do you think it's a good idea to add `npm run test` to the precommit hook to catch this sooner?

We can surely bring this up for a vote. We need to make sure everyone is ok with that. I know people have been a little concerned about adding too much, as it might over time make the commit process tedious.

schalkneethling:
> > > Hey @mindy, thank you for the PR. The Jest snapshot tests are failing because of the change in the `jsx` file. To update those snapshot tests, run the following and then add and commit the result:
> > > ```
> > > npm t -- --u
> > > ```
> > > 
> > > 
> > > [jestjs.io/docs/en/snapshot-testing.html#updating-snapshots](https://jestjs.io/docs/en/snapshot-testing.html#updating-snapshots)
> > 
> > 
> > Thanks @schalkneethling ! Updated the snapshots. Do you think it's a good idea to add `npm run test` to the precommit hook to catch this sooner?
> 
> We can surely bring this up for a vote. We need to make sure everyone is ok with that. I know people have been a little concerned about adding too much, as it might over time make the commit process tedious.

@mindy I would suggest opening an issue and asking for input from the team. Thanks!

mindy:
Thanks @schalkneethling ! Created https://github.com/mdn/kuma/issues/6437

Issue #6420, opened 2020-02-05
U - As a wiki editor I want to log into MDN using a Google account


atopal:
# User story
As a wiki editor I want to log into MDN using a Google account

# Acceptance criteria

- [ ] When a user clicks on sign in on the wiki, she is able to select Google auth in addition to Github auth
- [ ] After signing in, the user is redirected back to the page she started on.

# Tasks
*If relevant, you can make a checklist for tasks.*

1. - [ ] This is an incomplete task.
2. - [ ] Add redirect to sign-in link on Wiki to developer.mozilla.org

peterbe:
Being able to do `?next=` redirect links that contain the domain *might* be really tricky. Tricky, because it might entail digging deep into either `django-allauth` or even `django` itself.

Would it be an option to simply forfeit it. 

Suppose that you're on `https://wiki.developer.mozilla.org/en-US/docs/Foo/Bar` and click "Sign in" it MIGHT take you to `https://developer.mozilla.org/en-US/docs/Foo/Bar` (same path different domain) so you're completely lost at least. The "Edit in wiki" button will still be there.

peterbe:
@atopal @tobinmori I started looking at this and unfortunately it's not as easy as I had hoped. The github sign in link appears in other places other than the upper right hand corner of the Wiki pages. 

I think we can figure it all out but I just wanted to let you know that I think it's harder than I had first imagined.

peterbe:
[This first PR](https://github.com/mdn/kuma/pull/6435) isn't complete solution. At least it takes you from the Wiki to the ability to sign in with Google. I think, if we land this, this issue will be massively reduced in severity.

escattone:
Actually we can do this without leaving the wiki domain. See https://github.com/mdn/kuma/pull/6435#pullrequestreview-356373423

tobinmori:
So @peterbe, are you good with going with Ryan's proposed solution?

peterbe:
> So @peterbe, are you good with going with Ryan's proposed solution?

Yes. We had a meeting about it and it has been implemented rather pleasantly in the PR. 
It was a bit hairy because there's so many more places other than the upper right-hand corner linkt. Pluse everything needs to be implemented condtionally since `MULTI_AUTH_ENABLED` is off by default so the places in the templates where it matters need both solutions.

peterbe:
I can't think of any other piece of code I'd like/could fix to make this issue any more resolved. 
[...]

PR #6435, opened 2020-02-07
link wiki's sign-in to read-only site

peterbe:
Part of #6420

Now, when I'm on `http://wiki.localhost.org:8000/sv-SE/docs/Web/JavaScript` the link in the upper right-hand corner (GitHub icon removed also) is now: 
```
http://localhost.org:8000/sv-SE/users/account/signup-landing?next=%2Fsv-SE%2Fdocs%2FWeb%2FJavaScript
```

This is an incomplete compromise. [Based on this self-comment](https://github.com/mdn/kuma/issues/6420#issuecomment-583046995).
What's happening is that instead of linking to the GitHub "jump page" it instead links to the signup landing page. The `?next=` could have been `http%2E%2F%2Fwiki.localhost.org%2F8000%2Fsv-SE%2Fdocs%2FWeb%2FJavaScript` but instead it's `%2Fsv-SE%2Fdocs%2FWeb%2FJavaScript`.
If the `?next=` contains an absolute URL, it gets completely ignored. I don't know if that's our code, `allauth`, or Django but I know these kinds of things are very security sensitive and there's probably a *very* good reason that some code somewhere freaks out on absolute `?next` urls. 

So, if you click that link, it takes to the read-only site, and then it takes you back to the read-only *equivalent* of the page you where on. In my example case, that would be: `http://localhost.org:8000/sv-SE/docs/Web/JavaScript`
The compromise is that at least I get somewhat back to where I was and from there, I can get back to the Wiki by pressing the Edit in wiki button. 

Note that this PR does *not* close the issue. That's because of the `?next=` + absolute URL compromise. 
Also, on the `user_edit.html` page there's a link to GitHub on the Wiki. We'll just have to deal with that in a separate PR.


kuma/users/templatetags/jinja_helpers.py
```diff
@@ -132,6 +134,24 @@ def provider_login_url(context, provider_id, **params):
...
+@contextfunction
+def remote_login_url(context, **params):
+    """Redirects to the read-only site and makes sure the 'next' query
+    string is fully domained."""
+    request = context['request']
+    next = None
+    if 'next' not in params:
+        next = get_request_param(request, 'next')
+    if not next:
+        next = request.get_full_path()
```
escattone: Lines 143-147 don't seem right to me (although it all works in the end due to lines 146-147). You're never using the `next` keyword argument if one is provided. It seems like you'd want something like this instead:
```python
next = params.get('next') or get_request_param(request, 'next') or request.get_full_path()
```

kuma/users/templatetags/jinja_helpers.py
```diff
@@ -132,6 +134,24 @@ def provider_login_url(context, provider_id, **params):
...
+    url = request.build_absolute_uri(reverse('socialaccount_signin')).replace(
+        '/wiki.', '/'
+    )
```
escattone: It's safer to explicitly use the settings, something like this:
```python
from urllib.parse import urljoin
...
...
url = urljoin(settings.SITE_URL, reverse('socialaccount_signin'))
```

kuma/users/tests/test_templates.py
```diff
@@ -152,14 +152,28 @@ def test_signin(self):
...
+
+        with self.settings(MULTI_AUTH_ENABLED=True):
+            # http:// because it's always that in test client.
+            wiki_url_base = f'http://{settings.WIKI_HOST}'
+            response = self.client.get(all_docs_url, follow=True, HTTP_HOST=settings.WIKI_HOST)
+            assert response.status_code == 200
+            parsed = pq(response.content)
+            link = parsed.find("a.login-link")[0]
+            url = urlparams(reverse('socialaccount_signin'), next=all_docs_url)
+            expect_url = wiki_url_base.replace('/wiki.', '/') + url
```
escattone:
```suggestion
            expect_url = settings.SITE_URL + url
```

kuma/users/tests/test_templates.py
```diff
@@ -152,14 +152,28 @@ def test_signin(self):
...
+            response = self.client.get(all_docs_url, follow=True, HTTP_HOST=settings.WIKI_HOST)
+            assert response.status_code == 200
+            parsed = pq(response.content)
+            github_link = parsed.find("a.login-link[data-service='GitHub']")[0]
+            github_url = urlparams(reverse('github_login'), next=all_docs_url)
+            assert github_link.attrib['href'] == github_url
+
+        with self.settings(MULTI_AUTH_ENABLED=True):
+            # http:// because it's always that in test client.
+            wiki_url_base = f'http://{settings.WIKI_HOST}'
```
escattone: These lines could then be deleted.

peterbe:
Ryan and I chatted about this and came to the following conclusion...

1. On the Wiki, go straight to `/en-US/users/signup-landing` (or whatever it's called)
2. On the "Edit Profile" page, remove the "Click to connect your GitHub account"
3. If you try to reach a page you must be signed in to see, e.g. `/en-US/users/account/email` instead of sending the user to `/en-US/users/signin` send them to `/en-US/users/signup-landing`
4. On the 404 template, remove the "Sign in with GitHub" button. 
5. Finally, test that we can use BOTH the wiki. and the read-only domain with both Google and GitHub.

peterbe:
Note to self; here's what the 404.html looks like now if you have settings.MULTI_AUTH_ENABLED:
Screen Shot 2020-02-11 at 3 22 31 PM
It's hard to see in the screenshot but the "Sign in" link goes to `http://wiki.localhost.org:8000/en-US/users/account/signup-landing?next=%2Fen-US%2Fxxxsearch`

peterbe:
Same deal here:
Screen Shot 2020-02-11 at 3 23 59 PM

peterbe:
Now, I'm changing `settings.LOGIN_URL` to respect `settings.MULTI_AUTH_ENABLED` too. 
But if you end up on http://wiki.localhost.org:8000/en-US/users/signin anyway, I've changed that template too:
Screen Shot 2020-02-11 at 3 29 48 PM

peterbe:
So, if you do try to go to `http://wiki.localhost.org:8000/en-US/users/account/email` it will now instead redirect straight to `http://wiki.localhost.org:8000/en-US/users/account/signup-landing?next=/en-US/users/account/email`.

peterbe:
As a matter of fact, except that `recover_done.html` template, which I don't really understand how it works, I would dare to say that this PR now **fixes** the issue.

kuma/users/jinja2/account/login.html
```diff
@@ -36,9 +36,15 @@ <h1>{{ _("Please sign in") }}</h1>
       {% endif %}
 
       {% if socialaccount_providers  %}
+        {% if settings.MULTI_AUTH_ENABLED %}
+          {% trans login_url=url('socialaccount_signin')|urlparams(next=request.get_full_path()) %}
```
escattone: Likewise, you could simplify by using your new `login_url` function here:
```suggestion
          {% trans login_url=login_url() %}
```
which, like above, worked for me when I tested it locally.

Also, the text in line 33 above needs to change to include Google, so something like:
```jinja
      <p>{% trans %}
      In order to contribute, you need to sign in to MDN. You can sign in using
      GitHub or Google and a MDN profile will be created for you if you don't already
      have one.
      {% endtrans %}</p>
```

peterbe: No, because that copy won't matter any more. Since the new `settings.LOGIN_URL` means you don't end up this template at all (unless it's bookmarked or browser history). And that `settings.LOGIN_URL` is set depending on `settings.MULTI_AUTH_ENABLED`.

PR #6436, opened 2020-02-07
add black as dev dependency

peterbe:
My master plan for "black'ing" everything is:

1. Get `black` into `pyproject.toml` and run it in `.travis.yml` in a way that it just *lists* files but doesn't break the build.
2. Using the same technique I used for `stumptown-content` to slowly introduce `prettier`, there'll be a smattering of batches of `.py` files that I have "black'ified". 
3. Repeat until there are no more non-blacked files that can be blacked in `master`.
4. Remove that hack that would make Travis actually break if a single .py file isn't black'ed. 

So, if I can just get this file landed, between Dan's myriad of poetry upgrades, it'd make things easier.


peterbe:
@callahad @escattone I'm asking for review from both, but I suspect it's enough that just one of you review and land it. I just really wanted to get your attention and your eyes on my above mentioned rough plan. 

Kuma (Python) is the single only project I work on where I have to manually correct indentation and stuff. Absolutely every other side-project or active MDN project that I work on has this stuff automated.

callahad:
Looks like this also picked up a few minor updates. Glancing at the lockfile, I see:

- boto3 (1.11.11 -> 1.11.12)
- botocore (1.14.11 -> 1.14.12)
- django-extensions (2.2.6 -> 2.2.7)
- s3transfer (0.3.2 -> 0.3.3)

Maybe you ran `poetry update` at some point, which pulled those in? If the need arises, you can limit it to specific packages, e.g., `poetry update pytest-django`.

For the sake of keeping the history clean, I'd want to split this into two PRs, one for Black and one for the other updates. On the other hand, there's nothing objectionable about the end result, so it's your call if you want to split it up or not.

peterbe:
What the heck! I have no idea how that happened. I didn't see your comments until now after it's been merged. :(

I did not notice that it upgraded unrelated stuff. I definitely didn't run `poetry update`. All I did was run `poetry add --dev black` and that's it. 

However, I did run into a git conflict. You we both had `poetry` related PRs open and yours got merged first. That made my `poetry.lock` conflict which I resolved like this: https://www.peterbe.com/plog/how-to-resolve-a-git-conflict-in-poetry.lock

Would that cause an upgrade of stuff?

Gee. This is sad and annoying.

callahad:
No worries, I tweaked the commit message to cover both changes ;)

As for Poetry, I don't think any of those commands should have updated
unrelated dependencies in the lockfile, but I'll try to reproduce it on
Monday.

PR #6440, opened 2020-02-07
big black superformat bang

peterbe:


peterbe:
Here's how I made this. I ran my `gitfilesinflight` command to get a list of .py files in open PRs. That gave me [a list of 18 files](https://gist.github.com/peterbe/3d86769876aa60d4d5f5dcc1639a6ba2) and then I used `git ls-files | rg '\.py'` to get a list of all possible files. Then I took like this and `' '.join(paths)` print out:
```
docker-compose exec web poetry run black docs/conf.py kuma/__init__.py kuma/api/__init__.py kuma/api/apps.py ...
```

callahad:
Awesome! Once CI passes I'll follow your steps and confirm I get the same results. If so, we can ship it :)

peterbe:
So, I had to do a bunch of extra tuneups. I don't really get why. For example, the `W504` error came up which I couldn't make any sense of. Then, out of frustration, I just ran `black` on those files again and it fixed itself. Perhaps a bug in `black`. 

Also, it broke a bunch of flake8 things due to this bug: https://github.com/PyCQA/pycodestyle/issues/815

So, I ran `flake8` on all those files looking for `E231` errors. THen I manually opened each and every file and manually removed those excess trailing commas. 
To test that my manual intervention worked, I ran, for example:
```
black kuma/search/tasks.py && flake8 kuma/search/tasks.py
```
until it was all well. 

Perhaps it's Friday afternoon but I don't quite get why I had to do this. Either way, I now lost track of my noble attempt to not mess with files in flight in open PRs. Truth be told, so many of those open PRs are depressingly stale and not all of them will become conflicting after this monster of a change.

peterbe:
Let's use https://github.com/mdn/kuma/pull/6447 instead.

PR #6448, opened 2020-02-10
don't index MDN/Doc_status

### LINKED ISSUE #6422: E - What to do about Doc_status when going to git?
E.g. https://developer.mozilla.org/fr/docs/MDN/Doc_status/API/WebRTC

This page has `<meta name="robots" content="index, follow">` which I find a bit baffling to start with, but here we are. It's also included in [our sitemaps](https://developer.mozilla.org/sitemaps/fr/sitemap.xml)

There are, across all locales, 3,428 pages under `Doc_status`. 236 of them in `en-US`. 

I'm not even sure if the kinda KumaScript it does (the `DocStatus` macro) would even work in Yari because we're not exporting *everything* that we used to have in MySQL. 

We're going to need features like this when we're in Git too. Various dashboards and search tools that help translators. But we'll have to tackle that *after* we manage the most critical work which is to migration from Wiki to Git. 

**Note!** Google hates when you break the web with 404s. A "remedy" to that is to stop Google indexing it in the first place. But, when we switch to Yari, we don't want to have to remove it from sitemaps AND 404 the page on the same day. 
**It would be ideal to slowly phase these kinds of pages out of Google** by making theming `noindex, nofollow` and removing them from the sitemap XML files.

peterbe:
Fixes #6422


peterbe:
@Gregoor @callahad @escattone Whoever feels compelled to review. r?

kuma/wiki/models.py
```diff
@@ -1484,6 +1485,14 @@ def has_legacy_namespace(self):
         namespace, separator, _ = self.slug.partition(":")
         return namespace in LEGACY_MINDTOUCH_NAMESPACES if separator else False
 
+    @property
+    def has_noindex_slug(self):
+        """Return true if the slug is in hand-picked list of prefixes."""
+        for slug_start in NOINDEX_SLUG_STARTS:
+            if self.slug.startswith(slug_start):
+                return True
+        return False
```
callahad: I really want to code golf that down to:

```python
@property
def has_noindex_slug(self):
    """Return true if the slug is in hand-picked list of prefixes."""
    return any(self.slug.startswith(slug_start) for slug_start in NOINDEX_SLUG_STARTS)
```

...but this is Python, and so that's probably a bad idea. :)

kuma/wiki/constants.py
```diff
@@ -661,3 +661,6 @@
 WIKI_ONLY_DOCUMENT_QUERY_PARAMS = frozenset(
     ("edit_links", "include", "macros", "nomacros", "raw", "section", "summary",)
 )
+
+
+NOINDEX_SLUG_STARTS = ("MDN/Doc_status",)
```
callahad: Nit: is `SLUG_STARTS` a good name here, or would `SLUG_PREFIXES` be more clear?

Possible premature optimization: Are we sure that we'll only be concerned with prefixes, or might we be better served by a list of patterns instead, say, globs or regexps?

kuma/wiki/tests/test_models.py
```diff
@@ -150,6 +150,14 @@ def test_document_has_legacy_namespace(slug, legacy):
     assert Document(slug=slug).has_legacy_namespace == legacy
 
 
+@pytest.mark.parametrize(
+    "slug,should_noindex",
+    [("MDN/Doc_status", True), ("MDN/Doc_status/API", True), ("Web/CSS", False)],
```
callahad: What should `MDN/Doc_Status` return? (Case sensitivity)

What about something like `MDN/Doc_status_dashboard` (Should we only match at path boundaries, or should it be any arbitrary string prefix?)

kuma/wiki/models.py
peterbe: I'm not even convinced that's better that the "long form". It's "hipper" to use `any()` but I think it reads well as a "dumb" for loop.

kuma/wiki/constants.py
peterbe: Yeah, "PREFIXES" might be more accurate. 
I can't imagine there being a need for something that isn't just prefixes. If I'm wrong, we'll deal with it the day it matters. Ok?

kuma/wiki/tests/test_models.py
peterbe: It's fine to be case sensitive. There are no `Doc_StatUsEs`. 
There does exist a slug called `MDN/Doc_status_clone` but I'm confidently fine with that getting excluded too.

peterbe:
I renamed it to `NOINDEX_SLUG_PREFIXES` instead.

kuma/wiki/tests/test_models.py
callahad: Okay, that's fine with me. I find case sensitivity a bit surprising here, so I'd really appreciate two changes:

1. Add a test here to demonstrate case sensitivity, `("mdn/doc_status", False)` or similar.
2. Add a comment where we define `NOINDEX_SLUG_STARTS` noting that it's case sensitive. Just a `# case sensitive` at the end of the line would be enough.

Otherwise, 👍

peterbe: Done.

peterbe: I'll merge this when it goes green now.

PR #6454, opened 2020-02-10
don't start the 'api' service in docker-compose

peterbe:
I don't know else to test it but things seem to work just fine.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -63,23 +63,11 @@ services:
       - mysql
       - elasticsearch
       - redis
-      - kumascript
       - ssr
       - minio
       - mockga
     ports:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -63,24 +63,12 @@ services:
       - mysql
       - elasticsearch
       - redis
-      - kumascript
```
peterbe: This *was* weird. `docker-compose` starts all things by default, no? And the `kumascript` service depends on the `web` one. 
Actually, I could see the point of going the other way around. That instead, we say that `web` depends on `kumascript` but that the `kumascript` service doesn't depend on `web`.

PR #6458, opened 2020-02-10
Update coverage (4.5.4 -> 5.0.3)

callahad:
- Updating coverage (4.5.4 -> 5.0.3)
  https://github.com/nedbat/coveragepy/blob/coverage-5.0.3/CHANGES.rst

Lots of internals changed with the move to 5.0 (`.coverage` is now a
SQLite database!), but nothing that requires changes to our code.

Coverage 5 can also read settings from `pyproject.toml`, per PEP 518:
https://www.python.org/dev/peps/pep-0518/#tool-table

This lets us remove the separate `.coveragerc` file.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,0 @@
-[run]
-source = kuma
-branch = 1
-
-[report]
-omit = *migrations*,*/management/commands/*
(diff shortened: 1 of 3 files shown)


callahad:
And our Codecov integration still works 😌

PR #6461, opened 2020-02-11
Remove unused custom pytest markers

callahad:


callahad:
Whoops, needed to also remove some now unused `import pystest` statements to make flake8 happy.

PR #6469, opened 2020-02-12
Include Django migrations in linting

callahad:


tests/headless/map_301.py
```diff
@@ -701,7 +701,7 @@
                 url_test(
                     path + "/" + child_path + "?raw&macros",
                     redirect_path + "/" + child_path + "?raw&macros",
-                    **zone_url_test_kwargs
+                    **zone_url_test_kwargs,
```
peterbe: Why does black put this in? Sometimes. 
Isn't `flake8` going to complain?
But then you remove it and `black` doesn't put it back. 

Mind you, perhaps I'm confusing this with trailing commas in long literal dicts.

callahad: Oh, I talk about that in the first commit message: we're now explicitly telling Black (via pyproject.toml) that we want to target Python 3.8, which means it can use trailing commas in arg lists.

Before this, it'd use per-file heuristics to try and guess which version of Python we were targeting, and thus which syntactic features to enable when reformatting code.

callahad: Specifically, from the Black docs:

> One exception to adding trailing commas is function signatures containing `*`, `*args`, or `**kwargs`. In this case a trailing comma is only safe to use on Python 3.6. Black will detect if your file is already 3.6+ only and use trailing commas in this situation. If you wonder how it knows, it looks for f-strings and existing use of trailing commas in function signatures that have stars in them. In other words, if you’d like a trailing comma in this situation and Black didn’t recognize it was safe to do so, put it there manually and Black will keep it.

PR #6480, opened 2020-02-13
Re-enable stripe subscription form when modal is closed

### LINKED ISSUE #6472: After cancelling the stripe modal, the payment option is grayed out
**Summary**
When I cancel the stripe modal by clicking the x in the top right corner, the modal goes away and the payment option is grayed out. I have to reload the page to be able to pay.

Gregoor:
Fixes #6472 

I also had to add code to re-disable the form when the form is submitted again, after the modal has been closed once. The original jQuery code sets some semi-hidden data, that I could only access if I made the code jQuery dependent:
https://github.com/mdn/kuma/blob/master/kuma/static/js/main.js#L69

I decided adding a line for explicitly disabling the form is the lesser evil.

## How to test it
1. Enable Stripe in Kuma: https://kuma.readthedocs.io/en/latest/installation.html#enable-stripe-payments-optional
2. Sign-in/up as a staff user (or make a user staff through Django Admin)
3. Go to user edit and scroll to the bottom
4. (optional) if you want to go through the flow again there's multiple options
   1. delete your user (can also be found in user edit)
   1. run `docker-compose exec web python manage.py shell -c "from kuma.users.models import User; User.objects.filter(username='<YOUR_USERNAME>').update(stripe_customer_id='')"`
   1. go into the database and unset the stripe_customer_id

### DIFF
--- a/kuma/static/js/stripe-subscription.js
+++ b/kuma/static/js/stripe-subscription.js
@@ -36,11 +36,17 @@
         token: function(response) {
             stripeTokenInput.value = response.id;
             stripeForm.submit();
+        },
+        closed: function() {
+            if (!stripeTokenInput.value) {
+                stripeForm.classList.remove('disabled');
+            }
         }
     });
 
     stripeForm.addEventListener('submit', function(e) {
         if (!stripeTokenInput.value) {
+            this.classList.add('disabled');
             handler.open();
             e.preventDefault();
         }


peterbe:
How do I test this? In my mysql inside my docker-compose, this user already has a subscription set up?
Screen Shot 2020-02-13 at 3 16 00 PM

I'm guessing perhaps I can use the `./manage.py shell` to find the user and do some `stripe` magic to cancel the `user.stripe_customer_id` thing. 

Perhaps there could be a Django management command to cancel existing subscriptions for a test user.

Gregoor:
Good point, thanks Peter. I've added info on how to test it to the initial comment.

schalkneethling:
@Gregoor I tested this locally. I keep getting the Stripe error, but that might be because there is something wrong with the keys in my `.env` - I cannot verify my account on Stripe because South Africa is not one of the "supported" countries.

I believe in the past I used someone else's keys. With that said, the functional change here works as expected 🎉

Gregoor:
Great, thanks Schalk!

Issue #6479, opened 2020-02-13
Server error when trying to delete my account on stage


Gregoor:
It looks like there is a foreign key constraint blocking deletion of users in some cases. Here's the related sentry error:
> "Cannot delete some instances of model 'User' because they are referenced through a protected foreign key: 'AttachmentRevision.creator'", <QuerySet [<AttachmentRevision: test (file: "Screenshot_6.png", ID: #17085)>]>

https://sentry.prod.mozaws.net/operations/mdn-stage/issues/7271436

peterbe:
What's so bizarre is that there is already a test for this: https://github.com/mdn/kuma/blob/3df115a9ba8d3130094a9007f4e9d668a6f274fe/kuma/users/tests/test_views.py#L1533

peterbe:
I wonder, what happened is that you didn't have any Wiki document revisions, so it didn't even ask if you wanted to donate or scrub your account. But for some reason, you had an AttachmentRevision to your name.

Gregoor:
Also I think I just created the account yesterday and not doing anything in the wiki with it. So I don't even know where that Attachment would've come from 🤔

peterbe:
> Also I think I just created the account yesterday and not doing anything in the wiki with it. So I don't even know where that Attachment would've come from 🤔

Now that's very curious!
Here's the attachment revision that supposedly belongs to your account: https://developer.allizom.org/admin/attachments/attachmentrevision/17085/change/
And here's its attachment itself: https://developer.allizom.org/admin/attachments/attachment/16885/change/
Doesn't say in the Django Admin who the creator is.

Gregoor:
Okay maybe I got things mixed up and I had an account before, that is certainly my screenshot 😕

peterbe:
Either way, it's a bug. And I think I can write a unit test to reproduce it. 

By the way, it's weird that you can't look at who the creator was of an attachment revision in the Django admin. 

I tried going into the Kubernetes Stage cluster but https://github.com/mdn/infra/issues/374

peterbe:
Crap. It's not trivial. If a user has attachments, they need to be treated just like document revisions. I.e. "Do you want to donate them to 'Anonymous' or just scrub your account?"

That means that the UI needs to change too. Annoyingly, that UI needs to change in two places :(

CC @schalkneethling @atopal

peterbe:
[...]

PR #6482, opened 2020-02-13
delete user with AttachmentRevisions

peterbe:
Part of #6479

And CC @escattone 

This PR is only half the solution to #6479. Basically, our strategy is to confront the user, if they had revisions, so that they have to make a decision about them. Great. However, a user might not have done any DOCUMENT revisions. They might just have ATTACHMENT revisions. As in the case with [Gregor on Stage](https://sentry.prod.mozaws.net/operations/mdn-stage/issues/7271436/).

This PR just protects the Django side of things so the error can't happen. If you have EITHER document OR attachment revisions you now need to have a POST variable `attributions`. 

Once this is merge (or, can I perhaps start on it already??) I'll make it so that the two UIs (I hate that we have to maintain this in two different places!) doesn't decide whether or not to show the "What do do with you contributions" form piece only on the *document* revisions.

### DIFF
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -1611,9 +1611,6 @@ def test_delete_user_no_revisions_but_attachment_revisions_donate(
     The user didn't have any revisions to confront the legacy of, but there might be
     other things attached to the user.
     """
-    other_user = django_user_model.objects.create(
-        username="other", email="other@example.com"
-    )
     assert not Revision.objects.filter(creator=wiki_user).exists()
 
     attachment_revision = AttachmentRevision(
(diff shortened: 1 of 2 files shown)


kuma/users/tests/test_views.py
```diff
@@ -1601,6 +1601,48 @@ def test_delete_user_keep_attributions(
...
+    url = reverse("users.user_delete", kwargs={"username": wiki_user.username})
+    response = user_client.post(url, HTTP_HOST=settings.WIKI_HOST)
+    # This means it didn't work! The form rejects.
+    assert response.status_code == 200
+
+    # Ok, let's donate the attachment revisions to "Anonymous"
+    response = user_client.post(
+        url, {"attributions": "donate"}, HTTP_HOST=settings.WIKI_HOST
+    )
+    # This means it didn't work! The form rejects.
```
escattone: This comment is off, maybe something like this:
```suggestion
    # This means it worked! The user's attributions have been donated to the Anonymous user.
```

PR #6486, opened 2020-02-13
Stop depending on .settings.testing in Makefile

Gregoor:
Peter & my thinking on this was that the testing settings shouldn't be used outside of pytest as that dual-use leads to some weirdness, like the CI running system checks with those settings.


peterbe:
An interesting test is to remove that check for “testing” in the system checks. Because, I contend, it shouldn’t even be there.

Gregoor:
> 
> 
> An interesting test is to remove that check for “testing” in the system checks. Because, I contend, it shouldn’t even be there.

agreed, merged and made the change. Let's see what Travis thinks about that.

Gregoor:
Travis agrees with you @peterbe !

kuma/users/checks.py
```diff
@@ -0,0 +1,35 @@
+import stripe
+from django.conf import settings
+from django.core.checks import Error, register
+
+
+STRIPE_ERROR_ID = "kuma.users.E001"
```
peterbe: nit: It doesn't need the `_ID` suffix, because they're all IDs. I think just `STRIPE_ERROR` is good enough. If we had a bunch of other things that could go wrong, it ought to be called something more description like `STRIPE_PLAN_UNEXPECTED_ERROR`.

.flake8
```diff
@@ -10,7 +10,7 @@ ignore = E203, E501, W503
 per-file-ignores =
     kuma/settings/local.py:F403,F405
     kuma/settings/prod.py:F403,F405
-    kuma/settings/testing.py:F403,F405
+    kuma/settings/pytest.py:F403,F405
```
peterbe: I don't know why but that name makes me nervous. It's called exactly like the [package](https://pypi.org/project/pytest/) and it might just work but it feels doesn't feel solid. How about `pytesting.py` which rings a bit better. You're using these settings when pytest'ing :)

peterbe:
I'm still not entirely comfortable with the check itself not being unit tested. It does have some business logic now. And payment related kuma code is here to stay and it matters. Yes, it'd be very mocked but I think a simple happy-path test that at least executes the check sets a good tone. I wouldn't block a merge on it but I do want to share my feelings.

peterbe:
@Gregoor Any reason not to make this an open PR? I think it's a great idea. 
Perhaps @callahad ought to review it.

peterbe:
@Gregoor Can you `master merge` this so it's ready for when @callahad comes back.

peterbe:
@Gregoor The curse of checking in git submodules seems to have struck here too :)

Gregoor:
So many force-pushes, but I finally got it right 🙈

PR #6487, opened 2020-02-13
Do not show avatar transition on signup

### LINKED ISSUE #6483: Do not show avatar transition on signup
It seemed like animating/transitioning in the new avatar on click of the "Create account" button. Turns out this introduces more problems than it is worth. Let's just show the relevant avatar by default with no fancy business.

This will also resolve https://bugzilla.mozilla.org/show_bug.cgi?id=1614273#c2

schalkneethling:
Fixes #6483 

Tested with and without multi auth enabled.

### DIFF
--- a/kuma/users/jinja2/socialaccount/signup.html
+++ b/kuma/users/jinja2/socialaccount/signup.html
@@ -27,17 +27,17 @@
     <ol class="auth-service-header">
       {% if provider.name == "GitHub" %}
         <li class="github-auth checked">
-          <img src="{{ static('auth/github-auth.svg') }}" width="90px" alt="Github" />
+          <img src="{{ static('auth/github-auth.svg') }}" width="90" alt="Github" />
           {{ _('Authorize with Github') }}
         </li>
       {% elif provider.name == "Google" %}
         <li class="google-auth checked">
-          <img src="{{ static('auth/google-auth.svg') }}" width="90px" alt="Google" />
+          <img src="{{ static('auth/google-auth.svg') }}" width="90" alt="Google" />
           {{ _('Authorize with Google') }}
         </li>
       {% endif %}
       <li class="mdn-profile">
-        <img src="{{ account.get_avatar_url() }}" width="90px" alt="{{ _('User avatar') }}" />
+        <img src="{{ account.get_avatar_url() }}" width="90" alt="{{ _('User avatar') }}" />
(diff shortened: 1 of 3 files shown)


schalkneethling:
@peterbe This is just so weird. Why are we seeing such different things? I do not see the newsletter sign-up checkbox at all, is that a `.env` setting?

Also, the avatar not loading, what is the `src` of the HTML element?

schalkneethling:
Here is what I see:

create-account

schalkneethling:
I am going to open an issue to just not bother with that second check-mark. It makes things way too complicated for very little(if any) user benefit. But that is a separate issue with a separate PR :)

peterbe:
> Also, the avatar not loading, what is the `src` of the HTML element?

`undefined`.

schalkneethling:
@peterbe I also notice that the Google logo in your screenshot is grayed out. That is not implemented yet in this PR so, perhaps you need to `docker-compose exec web make build-static`?

schalkneethling:
> > Also, the avatar not loading, what is the `src` of the HTML element?
> 
> `undefined`.

Hmmm, so for some reason for you `{{ account.get_avatar_url() }}` returns `undefined`.

schalkneethling:
> > > Also, the avatar not loading, what is the `src` of the HTML element?
> > 
> > 
> > `undefined`.
> 
> Hmmm, so for some reason for you `{{ account.get_avatar_url() }}` returns `undefined`.

This makes me think it is JS trying to set the `src` and not finding the `data-` attribute.

kuma/users/jinja2/socialaccount/signup.html
```diff
@@ -37,7 +37,7 @@ <h1>{{ _('Create your MDN account') }}</h1>
         </li>
       {% endif %}
       <li class="mdn-profile">
-        <img src="{{ static('auth/mdn-auth.svg') }}" width="90px" alt="{{ _('User avatar') }}" />
+        <img src="{{ account.get_avatar_url() }}" width="90px" alt="{{ _('User avatar') }}" />
```
peterbe: Perhaps I'm old but I learned back in the ancient days of decades ago that the `width` was just a number. Can you use `px` in them these days?

schalkneethling: Ha! weird. It will probably just be ignored. Not sure why it has `px`

PR #6502, opened 2020-02-14
Add noindex for the whole wiki but allow it in robots.txt

### LINKED ISSUE #6340: T - Update robots.tx (Warnings have spiked on Search Console)
**Summary**
We are seeing a large spike in warnings on Google Search Console. The spike started on December 22, 2019.

Coverage

**Steps To Reproduce (STR)**
1. Go to Search Console
2. Open Coverage Report
3. open "Indexed, though blocked by robots.txt"

**Actual behavior**
It looks like a bunch of URLs are blocked from indexing by Google that should not be blocked. Examples are:
* https://developer.mozilla.org/it/docs/Web/API/GlobalEventHandlers/onresize$translate
* https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/end%24edit
* https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateInit/sdpMid$translate?tolocale=de

When I test these URLs in Search Console's robots.txt tester it tells me that those URLs are allowed, and they should be. $translate, or $edit are not in robots.txt

**Expected behavior**
Those URLs should not be blocked, so that Google can crawl the pages, discover the no-index meta tag and remove them from its index.

**Additional context**
I assume that something is broken when it comes to handling the $ chracter in the URL. Maybe we just remove everything with the $url from our robots.txt. As far as I can tell, almost all of those URLs now live under wiki.developer.mozilla.org anyway.

Gregoor:
Fixes #6340

Sets `noindex, nofollow` for all wiki pages, hence also removes the block from all those jinja files. That also contained a check for `has_legacy_namespace` so I removed all related code. But I'm not super sure about that change, maybe there's future use for it?

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1140,7 +1140,11 @@ ENABLE_RESTRICTIONS_BY_HOST = config(
 # Allow robots, but restrict some paths
 # If the domain is a CDN, the CDN origin should be included.
 ALLOW_ROBOTS_WEB_DOMAINS = set(
-    config("ALLOW_ROBOTS_WEB_DOMAINS", default="developer.mozilla.org", cast=Csv())
+    config(
+        "ALLOW_ROBOTS_WEB_DOMAINS",
+        default="developer.mozilla.org,wiki.developer.mozilla.org",
+        cast=Csv(),
+    )
 )
 
 # Allow robots, no path restrictions
--- a/jinja2/base.html
+++ b/jinja2/base.html
@@ -11,13 +11,7 @@
   <title>{% block title %}{{ _('MDN Web Docs') }}{% endblock %}</title>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <meta name="robots" content=
-    {%- if request.get_host() in settings.ALLOW_ROBOTS_WEB_DOMAINS -%}
-        "{% block robots_value%}index, follow{% endblock %}"
-    {%- else -%}
-        "noindex, nofollow"
-    {%- endif -%}
-  >
+  <meta name="robots" content="noindex, nofollow">
 
   {% include "includes/preload.html" %}
 
--- a/kuma/attachments/jinja2/attachments/edit_attachment.html
+++ b/kuma/attachments/jinja2/attachments/edit_attachment.html
@@ -2,7 +2,6 @@
 {# L10n: 'title' is the title of the document. #}
 {% set title = _('Attachments | %(title)s', title=document.title) %}
 {% block title %}{{ page_title(document.title) }}{% endblock %}
-{% block robots_value %}noindex, nofollow{% endblock %}
 {% set classes = 'document' %}
 {% from 'includes/common_macros.html' import li_field %}
 {% from "includes/error_list.html" import errorlist %}
--- a/kuma/dashboards/jinja2/dashboards/macros.html
+++ b/kuma/dashboards/jinja2/dashboards/macros.html
@@ -4,7 +4,6 @@
 {% set github_macro_list = "https://github.com/mdn/kumascript/tree/master/macros/" %}
 {% set github_macro_view_prefix = "https://github.com/mdn/kumascript/blob/master/macros/" %}
 {% block title %}{{ page_title(title) }}{% endblock %}
-{% block robots_value %}noindex, nofollow{% endblock %}
 
 {% block content %}
 <h1>{{ title }}</h1>
--- a/kuma/dashboards/jinja2/dashboards/revisions.html
+++ b/kuma/dashboards/jinja2/dashboards/revisions.html
@@ -4,7 +4,6 @@
 {% set classes = 'compare' %}
 
 {% block title %}{{ page_title(title) }}{% endblock %}
-{% block robots_value %}noindex, nofollow{% endblock %}
 
 {% block document_head %}
   <!-- heading -->
--- a/kuma/dashboards/jinja2/dashboards/spam.html
+++ b/kuma/dashboards/jinja2/dashboards/spam.html
@@ -3,7 +3,6 @@
 {% set title = _('Spam Dashboard') %}
 
 {% block title %}{{ page_title(title) }}{% endblock %}
-{% block robots_value %}noindex, nofollow{% endblock %}
 
 {% set stat_ids=(('spam_viewers_change', '% Change', 'rate'),
                  ('spam_viewers', 'Spam Viewers', 'data'),
--- a/kuma/landing/jinja2/landing/maintenance-mode.html
+++ b/kuma/landing/jinja2/landing/maintenance-mode.html
@@ -3,7 +3,6 @@
 {% set styles = ('maintenance-mode',) %}
 
 {% block title %}{{ page_title(_('Maintenance Mode')) }}{% endblock %}
-{% block robots_value %}noindex, nofollow{% endblock %}
 
 {% block content %}
 
--- a/kuma/search/jinja2/search/results.html
+++ b/kuma/search/jinja2/search/results.html
@@ -17,8 +17,6 @@
   {% endif %}
 {% endblock %}
 
-{% block robots_value %}noindex, nofollow{% endblock %}
-
 {% block bodyclass %}search-results search-results-no-demo search-navigator-flush{% endblock %}
 
 {% macro doc_block(doc, index) -%}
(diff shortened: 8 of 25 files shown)


Gregoor:
I'm not super-sure about https://github.com/mdn/kuma/pull/6502/commits/28dab3d161df809d5321f3f07442acfec046063b @peterbe. My understanding is we want to serve the allow-robots for the wiki domain as well. And for the react page we always want `index, follow`. Hence the config `ALLOW_ROBOTS_WEB_DOMAINS` serves no more purpose and I dropped it.
I fear I might be missing something here though.

kuma/landing/views.py
```diff
@@ -106,22 +106,15 @@ def promote_buttons(request):
...
 def robots_txt(request):
     """Serve robots.txt that allows or forbids robots."""
     host = request.get_host()
     if host in settings.ALLOW_ROBOTS_DOMAINS:
         robots = ""
-    elif host in settings.ALLOW_ROBOTS_WEB_DOMAINS:
-        robots = ROBOTS_ALLOWED_TXT
     else:
-        robots = ROBOTS_GO_AWAY_TXT
+        robots = ROBOTS_ALLOWED_TXT
```
escattone: I think we need to keep this exactly as it was.

kuma/landing/views.py
```diff
@@ -106,22 +106,15 @@ def promote_buttons(request):
     for locale in settings.ENABLED_LOCALES
 )
 
-ROBOTS_GO_AWAY_TXT = """\
-User-Agent: *
-Disallow: /
-"""
-
```
escattone: Keep this as well.

kuma/settings/common.py
```diff
@@ -1137,11 +1137,6 @@ def pipeline_one_scss(slug, **kwargs):
     "ENABLE_RESTRICTIONS_BY_HOST", default=True, cast=bool
 )
 
-# Allow robots, but restrict some paths
-# If the domain is a CDN, the CDN origin should be included.
-ALLOW_ROBOTS_WEB_DOMAINS = set(
-    config("ALLOW_ROBOTS_WEB_DOMAINS", default="developer.mozilla.org", cast=Csv())
-)
```
escattone: We need to keep this too.

kuma/search/jinja2/search/react.html
```diff
@@ -13,8 +13,6 @@
   {% stylesheet 'react-search' %}
 {% endblock %}
 
-{% block robots_value %}noindex, nofollow{% endblock %}
-
```
escattone: It seems to me that this means we're changing this to `index, follow` (since it uses `react_base.html`), which seems like a mistake to me. It seems to me we should never be indexing/following any search results page.

Gregoor:
Thank you Ryan, I expected that I would be glossing over some realities here, but I did not expect that it would be staging. That's a great point, I'll revert those bits.

kuma/search/jinja2/search/react.html
Gregoor: The base wiki site now always comes with `noindex, nofollow`, so I just removed that override from all the subpages

Gregoor:
@escattone @peterbe after reflecting about it some more I'm unsure again. Given that the wiki now has `noindex, nofollow` on all pages, isn't it fine that the robots.txt doesn't disallow no more?

atopal:
@Gregoor yes, from a Google search ranking perspective it's fine to set the robots.txt to allow for all after setting no-index, no-follow on all of staging and all of the wiki. on the read-only version only the wiki documents and the start page should be set to index.

Gregoor:
Thanks that's good context. I just realized that my last comment was confused though. While it is true for the wiki on stage, stage is ofc also serving the react frontend for which my PR erroneously changed the robots meta to `index` 🤦‍♂ Will fix that!

kuma/search/jinja2/search/react.html
peterbe: I'm not so sure about that. It's a big topic but one shouldn't necessarily set `noindex,nofollow` unless you have a strong reason. A search results page (in developer.mozilla.org) is actually a perfectly fine page for Google to index. It might not be super helpful but it's not a bad page. 
It's actually not unusual to google for something and you arrive on some sort of site search results page.

escattone: I still disagree that the search results page should be indexed/followed, but I'll happily defer to our SEO people. Here my point was simply that we've always used `noindex, nofollow` for both our read-only and wiki search results pages, and by removing this code it will inherit the robots `meta` tag from `jinja2/react_base.html` which in this PR changes to a fixed `index, follow` (the `robots_value` Jinja block has been removed).

escattone:
> Thanks that's good context. I just realized that my last comment was confused though. While it is true for the wiki on stage, stage is ofc also serving the react frontend for which my PR erroneously changed the robots meta-tag to `index` 🤦‍♂ Will fix that!

Perfect, thanks @Gregoor!

Still, I want discuss with @atopal that it seems strange to allow robots to crawl stage (https://developer.allizom.org and https://wiki.developer.allizom.org). I know, as @Gregoor says above, that on stage we'd be specifying `noindex, nofollow` for all pages of both the read-only and wiki stage site, but why not just block the crawling of stage altogether as we've always done?
I don't think any of the stage read-only or wiki URL's are included in the Google Search warnings are they? I can see how it makes sense to allow crawling of the production site (and have the pages specify either `index, follow` or `noindex, nofollow` as appropriate), to eliminate any warnings on the production read-only or wiki sites, but assuming we don't have this problem with any of the stage URL's, it doesn't make sense to me to allow stage to be crawled at all.

So, to summarize, I was thinking that this PR should do the following:
- always allow crawling of the production read-only and wiki sites
- always block crawling of the stage read-only and wiki sites
- always use `index, follow` for all of the read-only pages (although I think we should except the search-results page here, as we've always done in the past, since that will just eat crawl budget given all of the query parameter variations, and for very little/no benefit it seems to me)
- always use `noindex, nofollow` for all of the wiki pages

but instead it seems that the following will be done instead:
- always allow crawling of the stage and production read-only and wiki sites
- use `index, follow` for all of the production read-only site pages
- use `noindex, nofollow` for all of the stage read-only site pages **and** of course all of the stage and production wiki site pages

escattone:
@Gregoor @atopal After further reflection, I strongly prefer that we take the first approach I mentioned above with this PR, since it avoids allowing robots to crawl stage which we've never allowed before (plus, it's simpler to implement I think):

- always allow crawling of the production read-only and wiki sites
- always block crawling of the stage read-only and wiki sites
- always use `index, follow` for all of the read-only pages (although I think we should except the search-results page here, as we've always done in the past, since that will just eat crawl budget given all of the query parameter variations, and for very little/no benefit it seems to me -- what do you think?)
- always use `noindex, nofollow` for all of the wiki pages

atopal:
I really don't care about stage as long as we don't ask Google to crawl it. Whatever works for you guys there. Search console works per domain, so I only care about useless error messages on developer.mozilla.org crowding out actual important errors. Stage has no impact on that.

atopal:
re search results pages: should be no-index, no-follow. We don't want to send people to those (I don't feel too strongly about that though).

atopal:
Also, this really ballooned beyond what the initial scope of this issue was: the wiki. I'd be okay to limit the scope to that again, at least for this sprint. I'll leave it up to the dev team to decide whether that's a good idea here.

escattone:
Thanks for your comments @atopal!

Gregoor:
Thanks for that exchange, Ryan and Kadir. I think that helped me get a better grasp and sorry for blowing this out of scope&proportion, I was a little over-eager in ""oversimplifiying"" things it seems.

@escattone I reset the branch back to master and made a change commit that only sets "noindex, nofollow" for all the wiki pages. I think that makes it in line with your recommendations. Thanks again for thinking it through thoroughly!

Gregoor:
Right, I've changed that accordingly. I wasn't sure about my config (over?)simplifications anymore, as I also wasn't sure whether I've grokked the attachment host stuff completely.

Thanks again!

escattone:
@Gregoor @atopal I think we can safely release what is currently here and see how that affects the search warnings, but one thing I'm unsure of is whether we might need to allow the crawling of **all** MDN URL's. Currently our production `robots.txt` disallows a list of URL's, but should we stop doing that?

atopal:
We need to remove disallowed pages from robots.txt on the wiki domain.

Gregoor:
We want
```
User-agent: *
Disallow:
```
everywhere

atopal:
We need to remove disallowed pages from robots.txt on the wiki domain.

PR #6517, opened 2020-02-17
Show username for keep attribution option

### LINKED ISSUE #6418: T - Add username to "Keep my attribution for my page changes"
When a user chooses to delete their account we have two scenarios:

1. The user has no contributions
2. The user has at least one

For the second case we show two radio buttons with the following text:

- Keep my attribution for my page changes
- Switch my attribution for my page changes to "Anonymous"

Originally it was designed that the first option should also include the user's username, for example:

- Keep my attribution for my page changes(pietpompies)

The strings used for this are defined inside the following Python file:
https://github.com/mdn/kuma/blob/master/kuma/users/forms.py#L211

In this context we have no idea what the username is. To remedy this, we need to update the following Jinja:

https://github.com/mdn/kuma/blob/master/kuma/users/jinja2/users/includes/delete-user-content.html#L13

To not use the shorthand for including all the form field but, explicitly code the fields. In this way, we can have something like(pseudo-code) for the first option field:

```
{{ form.label_for form.username.value() }}
```

schalkneethling:
This shows the current user's username in parentheses next to the "Keep my attribution for my page changes" string

Screenshot 2020-02-17 at 21 19 36

Fix #6418

### DIFF
--- a/kuma/users/jinja2/users/includes/delete-user-content.html
+++ b/kuma/users/jinja2/users/includes/delete-user-content.html
@@ -7,7 +7,20 @@
             <p>{{ _('Deleting your account loses any preferences you have set.') }}</p>
             <p>{{ _('Although we can delete your account, any revisions you have made to pages on MDN will remain. Please select how would like us to handle the attribution for your changes.') }}</p>
             {{ form.attributions.errors }}
-            {{ form.attributions }}
+            <ul>
+                {% for field in form.attributions %}
+                    <li>
+                        <label for="{{ field.id_for_label }}">
+                            {{ field.tag() }}
+                            {% if "Anonymous" not in field.choice_label %}
+                                {{ field.choice_label }} ({{ username }})
+                            {% else %}
+                                {{ field.choice_label }}
+                            {% endif %}
+                        </label>
+                    </li>
+                {% endfor %}
+            </ul>
         {% else %}
             <p>{{ _('Deleting your account loses any preferences you have set, as well as makes your username available for others to use.') }}</p>
         {% endif %}


kuma/users/jinja2/users/includes/delete-user-content.html
```diff
@@ -7,7 +7,20 @@
             <p>{{ _('Deleting your account loses any preferences you have set.') }}</p>
             <p>{{ _('Although we can delete your account, any revisions you have made to pages on MDN will remain. Please select how would like us to handle the attribution for your changes.') }}</p>
             {{ form.attributions.errors }}
-            {{ form.attributions }}
+            <ul>
+                {% for field in form.attributions %}
+                    <li>
+                        <label for="{{ field.id_for_label }}">
+                            {{ field.tag() }}
```
schalkneethling: @callahad btw. this is what tripped me up. In the Django docs[1] it shows this to be `{{ field.tag }}`. When I did that I got `<bound method BoundWidget.tag of <django.forms.boundfield.BoundWidget object at 0x7fb96803b970>>`. I went reading and searching all over the place to try and figure out why the heck I got this output.

In the end the answer was staring me in the face i.e. `bound method BoundWidget.tag`. Method being the keyword here :)

[1] https://docs.djangoproject.com/en/3.0/ref/forms/widgets/#radioselect

kuma/users/jinja2/users/includes/delete-user-content.html
```diff
@@ -7,7 +7,20 @@
...
+                            {% if "Anonymous" not in field.choice_label %}
```
callahad: This feels a bit brittle. I wonder if there's a way we could do this inside `kuma/users/forms.py` instead.

It's been years since I've really worked with Django forms; I'll re-read the docs and see.

schalkneethling: It does for sure. Currently, we do not have the context to know who the user is within the Form class. If we could do it that way, that would definitely be much better.

callahad: Maybe the approach in #6523 is better, in that case? 🤷‍♂

schalkneethling:
Closing in favor of https://github.com/mdn/kuma/pull/6523

PR #6518, opened 2020-02-17
Always show checkbox on newsletter form

### LINKED ISSUE #5977: [Newsletter] Privacy checkbox hidden when JS disabled
**Summary**

The privacy checkbox is hidden for users with JS disabled on the newsletter signup form.

**Steps To Reproduce (STR)**

Go to any page with a newsletter signup form with JS disabled in the browser

1. The privacy checkbox is hidden
2. Clicking on submit will no submit the form as the privacy field is required, but the user will not be aware of the missing field

**Expected behavior**

Privacy checkbox should be visible for users with JS disabled. Generally, in these cases the preferred pattern is to make everything essential visible and then visually hide elements via JS by toggling the `hidden` class

mindy:
Visual changes approved by Mustafa on Slack: https://mozilla.slack.com/archives/CLTBV3W9X/p1581702471117700 

**BEFORE**
Homepage:
homepage-before

Footer: 
footer-before

**AFTER**
Homepage:
homepage-after

Footer:
footer-after

Closes #5977

### DIFF
--- a/jinja2/includes/newsletter.html
+++ b/jinja2/includes/newsletter.html
@@ -1,24 +1,8 @@
 <div class="newsletter">
     <form id="newsletterForm" class="newsletter-form nodisable" name="newsletter-form" action="https://www.mozilla.org/en-US/newsletter/" method="post">
         <div class="newsletter-head">
-            <h2 class="newsletter-teaser">{{ _('Learn the best of web development') }}</h2>
-            <p class="newsletter-description">{{ _('Get the latest and greatest from MDN delivered straight to your inbox.') }}</p>
-            {% if request.LANGUAGE_CODE != 'en-US' %}
-            <p class="newsletter-lang">{{ _('The newsletter is offered in English only at the moment.') }}</p>
-            {% endif %}
-        </div>
-
-        <div class="newsletter-fields">
-            <input type="hidden" id="fmt" name="fmt" value="H">
-            <input type="hidden" id="newsletterNewslettersInput" name="newsletters" value="app-dev" />
-            <div id="newsletterErrors" class="newsletter-errors"></div>
-
-            <div id="newsletterEmail" class="form-group newsletter-group-email">
-                <label for="newsletterEmailInput" class="form-label offscreen">{{ _('E-mail') }}</label>
-                <input type="email" id="newsletterEmailInput" name="email" class="form-input newsletter-input-email" required placeholder="{{ _('you@example.com') }}" size="30">
-            </div>
-
-            <div id="newsletterPrivacy" class="form-group form-group-agree newsletter-group-privacy hidden">
+            <h2 class="newsletter-teaser">{{ _('Learn the best of web development. Get the latest and greatest from MDN delivered to your inbox.') }}</h2>
+            <div id="newsletterPrivacy" class="form-group form-group-agree newsletter-group-privacy">
                 <input type="checkbox" id="newsletterPrivacyInput" name="privacy" required>
                 <label for="newsletterPrivacyInput">
                 {% trans policy_url="https://www.mozilla.org/privacy/" %}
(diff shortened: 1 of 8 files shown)


jinja2/includes/newsletter.html
```diff
@@ -1,31 +1,30 @@
...
         </div>
 
         <div class="newsletter-fields">
             <input type="hidden" id="fmt" name="fmt" value="H">
             <input type="hidden" id="newsletterNewslettersInput" name="newsletters" value="app-dev" />
-            <div id="newsletterErrors" class="newsletter-errors"></div>
 
             <div id="newsletterEmail" class="form-group newsletter-group-email">
                 <label for="newsletterEmailInput" class="form-label offscreen">{{ _('E-mail') }}</label>
                 <input type="email" id="newsletterEmailInput" name="email" class="form-input newsletter-input-email" required placeholder="{{ _('you@example.com') }}" size="30">
```
schalkneethling: Not introduced here but I noticed that this `input` has a `size` attribute with a limit of 30. For consistency, we should remove it here. Also, some people have really looooong email addresses especially if they use something like myname+mdnwebdocs@gmail.com

jinja2/includes/newsletter.html
```diff
@@ -1,31 +1,30 @@
 <div class="newsletter">
     <form id="newsletterForm" class="newsletter-form nodisable" name="newsletter-form" action="https://www.mozilla.org/en-US/newsletter/" method="post">
         <div class="newsletter-head">
-            <h2 class="newsletter-teaser">{{ _('Learn the best of web development') }}</h2>
-            <p class="newsletter-description">{{ _('Get the latest and greatest from MDN delivered straight to your inbox.') }}</p>
+            <h2 class="newsletter-teaser">{{ _('Learn the best of web development. Get the latest and greatest from MDN delivered to your inbox.') }}</h2>
+            <div id="newsletterPrivacy" class="form-group form-group-agree newsletter-group-privacy">
+                <input type="checkbox" id="newsletterPrivacyInput" name="privacy" required>
+                <label for="newsletterPrivacyInput">
```
schalkneethling: I have seen it be pretty standard practice to nest the `checkbox` input inside the `label` tag. What do you think? If you do decide to do it here, then we should do it in the `jsx` as well.

schalkneethling: I noticed that on mobile the alignment of the checkbox with the text is a little off:

Screenshot 2020-02-18 at 09 37 36

Moving the `checkbox` inside the `label` resolves that problem so, let's make that change.

schalkneethling:
Not to throw a spanner in the works but the location of the privacy checkbox field here feels very out of context:

Screenshot 2020-02-18 at 09 40 46

I would recommend moving it to in between the input field and the submit button.

schalkneethling:
Also, another reason to change this is that the HTML structure is not as expected. This causes problems with tab order and accessibility. If you focus the email input field and then press tab the privacy checkbox field is skipped.

schalkneethling:
> Visual changes approved by Mustafa on Slack: https://mozilla.slack.com/archives/CLTBV3W9X/p1581702471117700

Not to be the grumpy uncle (😺 ), but we should ensure that if we discuss something on Slack that we record the conversations and any decisions made in the relevant issue. This is to make sure that it is easily accessible, and easy to reference by everyone. Thank you.

kuma/static/styles/components/_newsletter.scss
```diff
@@ -82,6 +90,19 @@ input.newsletter-input-email {
     }
 }
 
+.newsletter-group-privacy {
+    display: flex;
+    margin: $grid-spacing 0;
+    input {
+        margin: 3px 0 0;
+    }
+    label {
```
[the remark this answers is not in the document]
mindy: ✅

jinja2/includes/newsletter.html
mindy: Removed. Good catch!

malqinneh:
> Not to throw a spanner in the works but the location of the privacy checkbox field here feels very out of context:
> 
> Screenshot 2020-02-18 at 09 40 46
> 
> I would recommend moving it to in between the input field and the submit button.

No strong preference here. @mindy if it's relatively easy to place the checkbox on top of the email input - that would suffice. @schalkneethling I would prefer to redesign that newsletter call-out to address communication issues users have had with becoming a monthly subscriber **vs** signing up for an account **vs** signing up for a newsletter.

schalkneethling:
> No strong preference here. @mindy if it's relatively easy to place the checkbox on top of the email input

I was actually thinking in between the input and the submit button.

> @schalkneethling I would prefer to redesign that newsletter call-out to address communication issues users have had with becoming a monthly subscriber vs signing up for an account vs signing up for a newsletter.

Sounds good. Do we have an issue describing the problem and possible ways to address it?

> Again, @mindy's implementation resolves the issue at hand "Always show checkbox on newsletter form" and would not delay for optimizations.

Agreed. I would want to make sure we do not break accessibility and or expectation for people using assistive tech or the keyboard. And as mentioned, for me at least, there is a visual disconnect on desktop viewports between the checkbox and the form fields.

malqinneh:
> Sounds good. Do we have an issue describing the problem and possible ways to address it?

Not yet, we are not prioritizing it for this/next sprint but the team is aware.

mindy:
Following an earlier discussion, Kadir has asked to abandon this diff.

PR #6520, opened 2020-02-18
Record which test functions cover each line

callahad:
Prior to this patch, our coverage reports only indicated whether a line
was covered or not. Now, they include additional metadata showing which
tests cover each line. Ned Batchelder calls this "Who Tests What:"
https://nedbatchelder.com/blog/201810/who_tests_what_is_here.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,9 +105,13 @@ target-version = ["py38"]
 [tool.coverage.run]
 source = ["kuma"]
 branch = true
+dynamic_context = "test_function"
 
 [tool.coverage.report]
 omit = ["*migrations*", "*/management/commands/*"]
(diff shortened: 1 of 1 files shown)


callahad:
Before:
> Screenshot_2020-02-18 Coverage for kuma core urlresolvers py 100%(1)

After:
> Screenshot_2020-02-18 Coverage for kuma core urlresolvers py 100%

peterbe:
Does this slow down tests in CI though?

callahad:
> Does this slow down tests in CI though?

Nope. For example, this finished the Docker tests in 13:33, while #6523, without this change, finished in 14:19. Well within normal Travis variation.

PR #6523, opened 2020-02-18
Show username when deleting an account with edits

### LINKED ISSUE #6418: T - Add username to "Keep my attribution for my page changes"

callahad:
When a user deletes their account, we ask if we should retain or
anonymize their prior contributions to the Wiki. This change displays
their current username as part of the deletion / anonymization prompt.

Fixes #6418

Alternative approach to #6517

### DIFF
--- a/kuma/users/forms.py
+++ b/kuma/users/forms.py
@@ -204,17 +204,17 @@ class UserRecoveryEmailForm(forms.Form):
 
 
 class UserDeleteForm(forms.Form):
+    def __init__(self, *args, username, **kwargs):
+        super().__init__(*args, **kwargs)
 
-    ATTRIBUTIONS_KEEP = "keep"
-    ATTRIBUTIONS_DONATE = "donate"
-    ATTRIBUTIONS_CHOICES = (
-        (ATTRIBUTIONS_KEEP, _("Keep my attribution for my page changes")),
-        (
-            ATTRIBUTIONS_DONATE,
-            _('Switch my attribution for my page changes to "Anonymous"'),
-        ),
-    )
+        choices = {
+            "keep": (
+                _("Keep my attribution for my page changes (%(username)s)")
+                % {"username": username}
+            ),
+            "donate": _('Switch my attribution for my page changes to "Anonymous"'),
+        }
 
-    attributions = forms.ChoiceField(
-        required=True, choices=ATTRIBUTIONS_CHOICES, widget=forms.widgets.RadioSelect()
-    )
+        self.fields["attributions"].choices = choices.items()
+
+    attributions = forms.ChoiceField(required=True, widget=forms.widgets.RadioSelect())
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -470,7 +470,7 @@ def user_edit(request, username):
         "edit_user": edit_user,
         "user_form": user_form,
         "username": user_form["username"].value(),
-        "form": UserDeleteForm(),
+        "form": UserDeleteForm(username=username),
         "revisions": revisions,
         "attachment_revisions": attachment_revisions,
         "subscription_info": retrieve_stripe_subscription_info(edit_user,),
@@ -548,7 +548,7 @@ def user_delete(request, username):
     if request.method == "POST":
         # If the user has no revisions there's not choices on the form.
         if revisions.exists() or attachment_revisions.exists():
-            form = UserDeleteForm(request.POST)
+            form = UserDeleteForm(request.POST, username=username)
             if form.is_valid():
                 if form.cleaned_data["attributions"] == "donate":
                     donate_attributions()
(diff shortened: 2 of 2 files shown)


kuma/users/forms.py
```diff
@@ -204,17 +204,17 @@ def save(self, request):
...
+        choices = {
+            "keep": _("Keep my attribution for my page changes"),
+            "donate": _('Switch my attribution for my page changes to "Anonymous"'),
+        }
 
-    attributions = forms.ChoiceField(
-        required=True, choices=ATTRIBUTIONS_CHOICES, widget=forms.widgets.RadioSelect()
-    )
+        if username:
+            choices["keep"] += f" ({username})"
```
callahad: This reifies the `ugettext_lazy` instance into a real string. Not sure if that's problematic or not; I'm a bit unfamiliar with localization in Django.

callahad:
I'm not entirely certain of this approach, but it does avoid the need to manually render the form as in #6517

schalkneethling:
This is **much** better than unpacking the form and replying on a string that might change. I am going to close mine in favor of this one. Thanks, @callahad 🌮

kuma/users/forms.py
callahad: Perhaps more ideally, the username would be a token in the localized string, since just slamming it onto the end here might not be correct in, say, RTL locales. But it should still be intelligible regardless. E.g.,:

> (callahad) Keep my attribution for my page changes

escattone: @callahad Yes, the best practice is to include the `username` as a placeholder in the localized string (so translators can reorder the placeholder as befits their language). I'll suggest a change.

kuma/users/forms.py
```diff
@@ -204,17 +204,17 @@ def save(self, request):
...
-    ATTRIBUTIONS_DONATE = "donate"
-    ATTRIBUTIONS_CHOICES = (
-        (ATTRIBUTIONS_KEEP, _("Keep my attribution for my page changes")),
-        (
-            ATTRIBUTIONS_DONATE,
-            _('Switch my attribution for my page changes to "Anonymous"'),
-        ),
-    )
+        choices = {
+            "keep": _("Keep my attribution for my page changes"),
```
escattone:
```suggestion
            "keep": _("Keep my attribution for my page changes (%(username)s)") % {"username": username},
```

escattone: I think we'll always have a truthy `username`?

kuma/users/forms.py
callahad: Other strings we use have `%s` and similar placeholders in them. Maybe use one here?

kuma/users/forms.py
callahad: If we want to require a username, I wonder if we should enforce it by changing the constructor's signature to `def __init__(self, *args, *, username, **kwargs):`, where the `*,` indicates the beginning of keyword-only arguments. Thus, if `username` wasn't passed, we'd blow up with:

    TypeError: UserDeleteForm() missing 1 required keyword-only argument: 'username'

Or maybe that's too much new-python black magic?

escattone: Interesting, I didn't know about that! Nice to learn something new about Python 3, thanks @callahad! Your idea sounds good to me!

peterbe: The simpler approach is to do:
```python
    username = kwargs.pop('username')
    super().__init__(*args, **kwargs)
    ...
    ...as is...
```
that way, it's always required but doesn't have to be truthy.

callahad: If we're going to enforce this, then I have a strong preference for `def __init__(self, *args, *, username, **kwargs)`; it does not require truthiness, it puts the args in the signature where they're more visible, and it gives a better error message: `missing 1 required keyword-only argument: 'username'` versus `KeyError: 'username'`

callahad:
I'll update the signature and re-push later tonight.

kuma/users/forms.py
callahad: Oh, cool, I misunderstood how this works. The correct syntax is just `def __init__(self, *args, username, **kwargs)`.

You only need the `*, ` if you're not otherwise collecting all the positional args.

PR #6535, opened 2020-02-19
Old payment management with new Stripe subscriptions

### LINKED ISSUE #6471: U - As a potential subscriber I want an FAQ page to know why MDN wants my money and what will happen with it, so I can make an informed decision
AC:
- [ ] FAQ page is accessible to everyone
- [ ] FAQ page text is updated with new text

### LINKED ISSUE #6473: U - As a paying subscriber I want a link in my profile, so I can cancel my subscription
AC:
- [ ] working link on profile that goes to https://wiki.developer.mozilla.org/en-US/payments/recurring/management

Gregoor:
Fixes #6473, Fixes #6471 

## What it does:
- reuse old payment management view
  - enable it based on a waffle flag instead of the env var
  - handle multiple stripe source types when retrieving them. This is the yin to [this PR](https://github.com/mdn/kuma/pull/6528)'s yang
- link to the payment management view

## How to test it:
1. go to the user edit view
1. create a subscription (at the bottom) if you don't already have one
1. click on the cancel subscription link
1. verify that you can see your subscription on the next page
1. cancel it
1. use the link on the page to go back to user edit
1. check if you can create a new subscription

### DIFF
--- a/kuma/users/jinja2/users/includes/stripe_subscription.html
+++ b/kuma/users/jinja2/users/includes/stripe_subscription.html
@@ -64,7 +64,7 @@
     </div>
     <hr>
     <p>
-      {{ _('To cancel your subscription, please <a href="%(support_email)s">contact support</a>. <br>If you have questions, please read the <a href="%(terms_url)s">FAQ</a> or you can also contact support.', support_email='mailto:mdn-support@mozilla.com', terms_url=url('recurring_payment_initial')) }}
+      {{ _('To cancel your subscription <a href="%(manage_url)s">click here</a>. <br>If you have questions, please read the <a href="%(terms_url)s">FAQ</a> or you can also contact support.', manage_url=url('recurring_payment_management'), terms_url=url('recurring_payment_initial')) }}
     </p>
   {% endif %}
   {% javascript "stripe-subscription" %}
--- a/kuma/users/jinja2/users/includes/stripe_subscription.html
+++ b/kuma/users/jinja2/users/includes/stripe_subscription.html
@@ -64,7 +64,10 @@
     </div>
     <hr>
     <p>
-      {{ _('To cancel your subscription <a href="%(manage_url)s">click here</a>. <br>If you have questions, please read the <a href="%(terms_url)s">FAQ</a> or you can also contact support.', manage_url=url('recurring_payment_management'), terms_url=url('recurring_payment_initial')) }}
+      {% trans manage_url=url('recurring_payment_management'), terms_url=url('recurring_payment_initial'), support_email='mailto:' + settings.CONTRIBUTION_SUPPORT_EMAIL -%}
+        To cancel your subscription <a href="{{ manage_url }}">click here</a>. <br>If you have questions, please read
+        the <a href="{{ terms_url }}">FAQ</a> or you can also <a href="{{ support_email }}">contact support</a>.
+      {% endtrans -%}
     </p>
   {% endif %}
   {% javascript "stripe-subscription" %}
--- a/kuma/users/jinja2/users/includes/stripe_subscription.html
+++ b/kuma/users/jinja2/users/includes/stripe_subscription.html
@@ -6,7 +6,7 @@
   </style>
 </noscript>
 
-<section class="stripe-subscription">
+<section class="stripe-subscription" id="subscription">
   {% stylesheet 'stripe-subscription' %}
 
   {% if not subscription_info %}
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -50,6 +50,7 @@ from kuma.core.ga_tracking import (
     CATEGORY_SIGNUP_FLOW,
     track_event,
 )
+from kuma.core.utils import urlparams
 from kuma.wiki.forms import RevisionAkismetSubmissionSpamForm
 from kuma.wiki.models import (
     Document,
@@ -820,8 +821,13 @@ def create_stripe_subscription(request):
         raven_client.captureException()
         has_stripe_error = True
 
-    query_params = "?" + urlencode({"has_stripe_error": has_stripe_error})
-    return redirect(reverse("users.user_edit", args=[user.username]) + query_params)
+    return redirect(
+        urlparams(
+            reverse("users.user_edit", args=[user.username]),
+            has_stripe_error=has_stripe_error,
+        )
+        + "#subscription"
+    )
 
 
 recovery_email_sent = TemplateView.as_view(
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -28,7 +28,7 @@ from django.shortcuts import get_object_or_404, redirect, render
 from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.encoding import force_text
-from django.utils.http import urlencode, urlsafe_base64_decode
+from django.utils.http import urlsafe_base64_decode
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.cache import never_cache
 from django.views.decorators.http import require_POST
--- a/kuma/payments/jinja2/payments/management.html
+++ b/kuma/payments/jinja2/payments/management.html
@@ -22,7 +22,7 @@
     <div class="manage-payments">
         <article class="column-container">
             <header class="manage-payments-header">
-                <h1>{{_('Your MDN profile')}}</h1>
+                <h1>{{_('MDN subscription management')}}</h1>
             </header>
             <div>
                 <a href="{{ url('users.user_edit', username=user.username) }}" class="button neutral back">{% include 'includes/icons/arrows/arrow-left.svg' %} {{_('Back to profile')}}</a>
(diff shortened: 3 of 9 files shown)


peterbe:
Why does the "Why not _set on up_?" link (`http://wiki.localhost.org:8000/en-US/profiles/peterbe/edit?has_stripe_error=False`) have all that `?has_stripe_error=False` stuff in it?
Screen Shot 2020-02-26 at 11 18 23 AM
Just seeming it makes me nervous. "Stripe??" "error??" What?!

peterbe:
The email link has a weird subject line prefilled:
`mailto:mdn-support@mozilla.com?Subject=Recurring%20payment%20support`
"Payment"?! What am I *paying* for? What did I buy?? I thought I set up a *subscription*. "How are they different?" "Oh no is that for something else?"
Screen Shot 2020-02-26 at 11 20 16 AM

I'm unhappy with this hack of trying to revive some of the old code just because it happened to be about Stripe and stuff. 
There might also be existing MDN users who remember the old project (from outside Mozilla) who will get confused. For example, the current URL is `http://wiki.localhost.org:8000/en-US/payments/recurring/management`

peterbe:
Screen Shot 2020-02-26 at 11 23 31 AM

Inconsistencies here. 

1. the document title is `MDN subscription management` but the page title is `Your MDN profile`
2. The sub-title "Subscriptions" is somewhat correct but it's what the page title should be (?)
3. The address bar talks about "payments recurring" which doesn't match the document title or the page sub-title.

peterbe:
Minor, but the layout of this confirmation box is just weird. :)
Screen Shot 2020-02-26 at 11 07 25 AM
I like that I'm forced to move the mouse cursor so that I can't confirm by simply clicking two times in the same spot. 

I still think the message could be improved. 

**Are you sure**
*This will cancel your monthly subscription to MDN Web Docs*

peterbe:
Screen Shot 2020-02-26 at 11 07 43 AM
"Your active subscriptions"

Not necessary to say because it has that block boxed thing below with its own sub-title of "Monthly Subscriptions"

peterbe:
By the way, where do I see my billing receipts? 
If a corporation wants to set up a subscription (made by an individual within the corporation) they need to have the receipts for tax purposes.

Gregoor:
> But I did click the FAQ link and it 404ed. :(

As mentioned on Slack, FAQ is not part of the issue the PR was originally addressing. But I might pull it in and get rid of the env var once and for all.


> By the way, where do I see my billing receipts?

That'd be this issue #6562, which we're afaik also tackling this sprint.


I'm gonna let @atopal reply to the other concerns.

Gregoor:
> Why does the "Why not _set on up_?" link (`http://wiki.localhost.org:8000/en-US/profiles/peterbe/edit?has_stripe_error=False`) have all that `?has_stripe_error=False` stuff in it?

A lil UX gimmick I threw in there. When that param is present in the user edit view, the page is scrolled to the subscription section:
https://github.com/mdn/kuma/blob/4bd367a058efd55540428ae1ebb12fbb9030dbce/kuma/static/js/stripe-subscription.js#L6-L13
I agree that it's a misnomer here though. I'll try to think of a better name.


Thanks for the review Peter. I've decided to do the FAQ-enabling and MDN_CONTRIBUTION-removing in this PR as well. So with this everything is tied to the subscription waffle.
I think I've addressed all the technical points, I'll leave the rest for @atopal to address.

peterbe:
> > Why does the "Why not _set on up_?" link (`http://wiki.localhost.org:8000/en-US/profiles/peterbe/edit?has_stripe_error=False`) have all that `?has_stripe_error=False` stuff in it?
> 
> A lil UX gimmick I threw in there. When that param is present in the user edit view, the page is scrolled to the subscription section:
> 
> https://github.com/mdn/kuma/blob/4bd367a058efd55540428ae1ebb12fbb9030dbce/kuma/static/js/stripe-subscription.js#L6-L13
> 
> I agree that it's a misnomer here though. I'll try to think of a better name.
> 
> Thanks for the review Peter. I've decided to do the FAQ-enabling and MDN_CONTRIBUTION-removing in this PR as well. So with this everything is tied to the subscription waffle.
> I think I've addressed all the technical points, I'll leave the rest for @atopal to address.

Then it should be an anchor. Less "weird". Less fragile. Less JS code. That's what anchors are for.

Gregoor:
Thanks Peter, good point. That makes for cleaner and more expressive code.

kuma/users/views.py
```diff
@@ -820,8 +821,13 @@ def create_stripe_subscription(request):
...
         has_stripe_error = True
 
-    query_params = "?" + urlencode({"has_stripe_error": has_stripe_error})
-    return redirect(reverse("users.user_edit", args=[user.username]) + query_params)
+    return redirect(
+        urlparams(
+            reverse("users.user_edit", args=[user.username]),
+            has_stripe_error=has_stripe_error,
+        )
+        + "#subscription"
```
peterbe: Thank you!!

peterbe:
Thank you for employing the anchor instead of the JS, but why do I still get this after I have signed up?
http://wiki.localhost.org:8000/en-US/profiles/peterbe/edit?has_stripe_error=False#subscription

Gregoor:
> 
> 
> Thank you for employing the anchor instead of the JS, but why do I still get this after I have signed up?
> http://wiki.localhost.org:8000/en-US/profiles/peterbe/edit?has_stripe_error=False#subscription

The user edit view still needs it to determine whether to show the stripe error:
https://github.com/Gregoor/kuma/blob/9b5266516f8a8c39e5c3dfe16c99d6962db0b2ec/kuma/users/views.py#L431
**After reading some more:** I didn't realize that django had another way of passing data between routes, using [redirect's kwargs](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#redirect). I'm gonna switch to that!

peterbe:
There are inconsistencies in the `?subject=...` email links across http://wiki.localhost.org:8000/en-US/payments/recurring/management vs. http://wiki.localhost.org:8000/en-US/payments/recurring

Screen Shot 2020-03-03 at 10 27 14 AM
No it's not. :) This is the "Subscription management" page.

kuma/users/jinja2/users/includes/stripe_subscription.html
```diff
@@ -64,7 +64,7 @@ <h2>{{ _("You're helping MDN grow") }}</h2>
     </div>
     <hr>
     <p>
-      {{ _('To cancel your subscription, please <a href="%(support_email)s">contact support</a>. <br>If you have questions, please read the <a href="%(terms_url)s">FAQ</a> or you can also contact support.', support_email='mailto:mdn-support@mozilla.com', terms_url=url('recurring_payment_initial')) }}
+      {{ _('To cancel your subscription <a href="%(manage_url)s">click here</a>. <br>If you have questions, please read the <a href="%(terms_url)s">FAQ</a> or you can also contact support.', manage_url=url('recurring_payment_management'), terms_url=url('recurring_payment_initial')) }}
```
peterbe: Can't the `contact support` here still be a mailto: link like it was before?

Gregoor:
> **After reading some more:** I didn't realize that django had another way of passing data between routes, using [redirect's kwargs](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#redirect). I'm gonna switch to that!

Okay I think I misunderstood that, redirect's kwargs would need to be present in the URL as well. So I guess we need to stick to the query param after all. Is that correct @peterbe?

peterbe:
> > **After reading some more:** I didn't realize that django had another way of passing data between routes, using [redirect's kwargs](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#redirect). I'm gonna switch to that!
> 
> Okay I think I misunderstood that, redirect's kwargs would need to be present in the URL as well. So I guess we need to stick to the query param after all. Is that correct @peterbe?

I guess it's fine. It's just always a bit weird when you mix multiple "things" on the same page. It's already so very weird that that super important little form appears BELOW where you type in your LinkedIn URL. So best just leave it as is for now under the circumstances.

kuma/users/jinja2/users/includes/stripe_subscription.html
Gregoor: Review comment for myself: reuse `settings.CONTRIBUTION_SUPPORT_EMAIL`

atopal:
Thanks @peterbe !

@Gregoor sorry for the late reply. Let me know if I missed anything, and if you're still blocked here. 

> Why does the "Why not set on up?" link (http://wiki.localhost.org:8000/en-US/profiles/peterbe/edit?has_stripe_error=False) have all that ?has_stripe_error=False stuff in it?

No idea, Gregor?


> The email link has a weird subject line prefilled:
mailto:mdn-support@mozilla.com?Subject=Recurring%20payment%20support
"Payment"?! What am I paying for? What did I buy?? I thought I set up a subscription. "How are they different?" "Oh no is that for something else?"

Yeah, email subject line should be changed to subscription support

> Inconsistencies here.
> the document title is MDN subscription management but the page title is Your MDN profile

Yeah, should be „MDN subscription management“

PR #6537, opened 2020-02-19
Remove our custom StrippedCharField

callahad:
As of Django 1.9, the built-in CharField strips whitespace by default


callahad:
...I'll need to commit a migration for this, but I kinda wanted to make Travis generate it for me since my Docker environment is otherwise occupied by the Django 2 work 😉

callahad:
Huh, no migrations necessary! Verified locally. Weird. My mental model of when Django needs them or not isn't totally accurate yet.

```
$ docker-compose exec web ./manage.py makemigrations --check --dry-run
No changes detected
```

peterbe:
I'm kinda trusting that you've grepped for `StrippedCharField` and that it's not mentioned anywhere else. If not, we'll fix it some other Tuesday.

PR #6544, opened 2020-02-20
Replace `onTouchStart` events for main navigation

erkde:
Disclaimer:

A first (non-trivial) attempt at a PR in this project, addressing an issue I filed yesterday, and was curious if I could find a solution.

--

This commit essentially just replaces the `onTouchStart` event for opening and closing the main navigation menus, to using the same pseudo selectors as used in `dropdown.scss`, so navigation works in both touch and non-touch browsers, regardless of the viewport size.

One little differences, the selector `:hover` is intentionally only applied to viewports `@mq-tablet-and-up`, in smaller viewports, the menus render vertically, and mousing through the options, just makes 
the UX really jittery.

A snapshot of `MainMenu` and a simple test for it is also added.

--

https://bugzilla.mozilla.org/show_bug.cgi?id=1616397


erkde:
Seems like a few other people are reporting this same issue #6570 (under Additional Context) & #6571

Looking at https://caniuse.com/#search=focus-within, the pseudo class that this PR implements (based on the implementation in Dropdown.js) doesn't seem to have particularly wide support for such a critical aspect of the site, perhaps just using simple click events to open and close the navigation (and the Dropdown perhaps) is going to have less issues?

schalkneethling:
@erkde Tha you for your PR and highlighting the additional logged issues. I do think our problem goes deeper and we need to rethink our interaction model. I am going to recommend that we implement consistent interactions across mobile and desktop. the one that makes the most sense, is to always open the submenu items based on a click/touch event.

erkde:
That makes sense, thanks for feedback, I'll close for now then.

PR #6546, opened 2020-02-20
Add CSS fallback for MathML

### LINKED ISSUE #6391: Regression: Math (MathJax/MathML) widgets broken in new platform
The new platform (developer.mozilla.org) cannot correctly present the CKEditor math components that work correctly on wikil.developer.mozilla.org. In particular, I'm using Chrome, and while Chrome doesn't support MathML, the platform is able to render matrices using a simplified form that while it's not ideal is quite readable.

On the new platform, though, the underlying Latex source is output (CKEditor uses Latex input and generates MathML or other syntax as appropriate depending on the browser).

This needs to be fixed because it is rendering pages unreadable, including many key pages in the WebXR and WebGL documentation, plus in a couple of places in the WebRTC docs.

schalkneethling:
This loads a CSS polyfill if MathML is not supported in the browser.

Tested with the following URLs:

- https://developer.mozilla.org/en-US/docs/Web/API/XRRigidTransform/matrix
- https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getTransform

Firefox supports this natively, Chrome does not.

fix #6391

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -882,6 +882,7 @@ PIPELINE_JS = {
             "js/libs/prism/prism-wasm.js",
             "js/libs/prism/prism-line-highlight.js",
             "js/libs/prism/prism-line-numbers.js",
+            "js/components/mathml.js",
             # The react.js file is created by webpack and
             # placed in the kuma/javascript/dist/ directory.
             "react.js",
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -882,7 +882,6 @@ PIPELINE_JS = {
             "js/libs/prism/prism-wasm.js",
             "js/libs/prism/prism-line-highlight.js",
             "js/libs/prism/prism-line-numbers.js",
-            "js/components/mathml.js",
             # The react.js file is created by webpack and
             # placed in the kuma/javascript/dist/ directory.
             "react.js",
@@ -903,6 +902,13 @@ PIPELINE_JS = {
         ),
         "output_filename": "build/js/banners.js",
         "extra_context": {"async": True},
+    },
+    "mathml": {
+        "source_filenames": (
+            "js/components/mathml.js",
+        ),
+        "output_filename": "build/js/mathml.js",
+        "extra_context": {"defer": True},
     },
     "users": {
         "source_filenames": ("js/users.js",),
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -904,9 +904,7 @@ PIPELINE_JS = {
         "extra_context": {"async": True},
     },
     "mathml": {
-        "source_filenames": (
-            "js/components/mathml.js",
-        ),
+        "source_filenames": ("js/components/mathml.js",),
         "output_filename": "build/js/mathml.js",
         "extra_context": {"defer": True},
     },
--- a/kuma/static/js/components/mathml.js
+++ b/kuma/static/js/components/mathml.js
@@ -1,0 +1,73 @@
+/**
+ * vanilla version of the function at:
+ * https://github.com/mdn/kuma/blob/master/kuma/static/js/wiki.js#L252
+ */
+(function() {
+    'use strict';
+
+    /*
+     * Bug 981409 - Add some CSS fallback for browsers without MathML support.
+     *
+     * This is based on:
+     * https://developer.mozilla.org/en-US/docs/Web/MathML/Authoring#Fallback_for_Browsers_without_MathML_support
+     * and https://github.com/fred-wang/mathml.css
+     */
+    var mathElements = document.querySelectorAll('math');
+    // if no `math` elements on the page, bail
+    if (!mathElements.length) {
+        return;
+    }
+
+    /**
+     * Returns a `math` element wrapped in a `div` that is positioned offscreen
+     * @returns `div` element
+     */
+    function getMathElement() {
+        var offscreenContainer = document.createElement('div');
+        var mathMLNamespace = 'http://www.w3.org/1998/Math/MathML';
+        var mathElement = document.createElementNS(mathMLNamespace, 'math');
+        var mspaceElement = document.createElementNS(mathMLNamespace, 'mspace');
+
+        mspaceElement.setAttribute('height', '23px');
+        mspaceElement.setAttribute('width', '77px');
+
+        mathElement.append(mspaceElement);
+        offscreenContainer.append(mathElement);
+        offscreenContainer.classList.add('offscreen');
+
+        return offscreenContainer;
+    }
+
+    // Test for MathML support
+    var mathMLTestElement = document.body.appendChild(getMathElement());
+    var box = mathMLTestElement.querySelector('mspace').getBoundingClientRect();
+    document.body.removeChild(mathMLTestElement);
+
+    var supportsMathML =
+        Math.abs(box.height - 23) <= 1 && Math.abs(box.width - 77) <= 1;
+    if (!supportsMathML) {
+        // Add CSS fallback
+        var polyfill = document.createElement('link');
+        polyfill.href = mdn.staticPath + 'styles/libs/mathml.css';
+        polyfill.rel = 'stylesheet';
+        polyfill.type = 'text/css';
+
+        document.head.append(polyfill);
+
+        // if we do not do this, React hydrate is unhappy and removes the element
+        // Warning: Did not expect server HTML to contain a <div> in <div>.
+        setTimeout(function() {
+            // Add notification
+            var wikiArticleContainer = document.getElementById('wikiArticle');
+            var notice = document.createElement('div');
+            var messageContainer = document.createElement('p');
+
+            messageContainer.textContent =
+                'Your browser does not support MathML. A CSS fallback has been used instead.';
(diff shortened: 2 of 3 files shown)


kuma/static/js/components/mathml.js
```diff
@@ -0,0 +1,73 @@
...
+        // if we do not do this, React hydrate is unhappy and removes the element
+        // Warning: Did not expect server HTML to contain a <div> in <div>.
+        setTimeout(function() {
+            // Add notification
+            var wikiArticleContainer = document.getElementById('wikiArticle');
+            var notice = document.createElement('div');
+            var messageContainer = document.createElement('p');
+
+            messageContainer.textContent =
+                'Your browser does not support MathML. A CSS fallback has been used instead.';
```
schalkneethling: NOTE: This is not wrapped in a `gettext` call because non-React code does not have access to these i18n functions. The reason for that is because we do not include the following on the React-based frontend:

```
<script src="{{ statici18n(request.LANGUAGE_CODE) }}"></script>
```

https://github.com/mdn/kuma/blob/master/jinja2/base.html#L170

@atopal Are you ok with this?

kuma/settings/common.py
```diff
@@ -882,6 +882,7 @@ def pipeline_one_scss(slug, **kwargs):
             "js/libs/prism/prism-wasm.js",
             "js/libs/prism/prism-line-highlight.js",
             "js/libs/prism/prism-line-numbers.js",
+            "js/components/mathml.js",
```
Gregoor: A recent conversation with Peter made me have thoughts about this:
Our current Kuma way of writing JS (including what's already in the codebase, my PRs and this PR) is brittle, i.e. it's barely tested, non-type checked and has implicit dependencies (also called `querySelector`s). Now any synchronous code that gets added to the main bundle has the potential of erroring and thereby preventing any other script to run. The cherry on top is that this behavior is release-mode exclusive as these scripts listed here get separate `<script>`-tags in dev.

So given all that my recommendation for this PR is to have mathml be a separate script. The drawback is that it takes longer for MathML to be displayed. But given that it is also dynamically loading in CSS, I think we haven't chosen the fast-path for this anyway. So imo we can take the hit.

kuma/static/js/components/mathml.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * vanilla version of the function at:
+ * https://github.com/mdn/kuma/blob/master/kuma/static/js/wiki.js#L252
```
Gregoor: When linking to lines on GitHub I recommend linking to a specific hash instead of linking to the tip of a branch. If the moves lines in the file, the link still points to the intended statement. Downside is ofc that it might be outdated. But at least you know what to look for.
To turn a branch-link into a commit sha-link press `y` while viewing that file within GitHub

kuma/static/js/components/mathml.js
```diff
@@ -0,0 +1,73 @@
...
+        mathElement.append(mspaceElement);
+        offscreenContainer.append(mathElement);
+        offscreenContainer.classList.add('offscreen');
+
+        return offscreenContainer;
+    }
+
+    // Test for MathML support
+    var mathMLTestElement = document.body.appendChild(getMathElement());
+    var box = mathMLTestElement.querySelector('mspace').getBoundingClientRect();
```
Gregoor: To my point of the earlier comment: this would break all script execution if a page contains a `<math>` element but not a `<mspace>` element. Now I admit that seems unlikely, but it's still scary that an extra element _somewhere_ could break the whole page's JS. I wouldn't make this code defensive though, once it's separated from the main bundle it can just error out.

What I'd do though (in general) is move `querySelector` & co to the top of the block. That way it's at least clear what the template "requirements" are for each script.

kuma/static/js/components/mathml.js
```diff
@@ -0,0 +1,73 @@
...
+        // Add CSS fallback
+        var polyfill = document.createElement('link');
+        polyfill.href = mdn.staticPath + 'styles/libs/mathml.css';
+        polyfill.rel = 'stylesheet';
+        polyfill.type = 'text/css';
+
+        document.head.append(polyfill);
+
+        // if we do not do this, React hydrate is unhappy and removes the element
+        // Warning: Did not expect server HTML to contain a <div> in <div>.
```
peterbe: Why is this code not written in React? Mixing custom DOM manipulation stuff with React is dangerously fragile. 
For example, if we decide to change the React component...
```diff
-id="wikiArticle"
+id="wiki-article"
````
who's going to remember to look up every other .js file to see if it depends on it being called exactly that?

If the only reason for doing it in vanilla JS is to try to hack it so it works in the Wiki AND in the read-only site, then remember that the Wiki is definitely going away. Can we not live without this for another couple of months? Or, fix it in React and just accept previewing in the Wiki is a bit different. 

Sorry for sounding negative but I wonder if this PR just introduces more fragility than it's worth.

File: kuma/static/js/components/mathml.js (hunk 3 above)
Comment (schalkneethling) [reply kuma/static/js/components/mathml.js]: Agreed. To be honest, though, I should have just added this as a comment to the PR. This will not be very useful for very long.

**Comment by schalkneethling** [comment]
Thank you for looking at the PR @peterbe and @Gregoor 

> Why is this code not written in React?

First, this is a P2 and not part of one of our main key results. We are also going to redo this in ~6months right? 😺 We have always said to not fuss too much with making Kuma better and instead, focus our energy on making sure Yari is the best it can be.

This PR does not make Kuma better nor worse but, fixes a regression that our Chrome users face. It also helps out our content team who need to use the `math` element in the documentation.

While this is by no means perfect, it solves our immediate problem.

> Our current Kuma way of writing JS (including what's already in the codebase, my PRs and this PR) is brittle, i.e. it's barely tested, non-type checked and has implicit dependencies (also called querySelectors). 

True, but that is the nature of the beat. Our pipeline will not allow us to make this better and again, we should not focus energy on making that better in Kuma.

> So given all that my recommendation for this PR is to have mathml be a separate script.

I agree that making this a separate, deferred script makes sense. I toyed with the idea and honestly do not know why I chose not to go that route in the end :)

> To my point of the earlier comment: this would break all script execution if a page contains a <math> element but not a <mspace> element.

In this case, we **know** that it will always have a `mspace` element as it is called against a piece of HTML we control. I am going to refactor this a little so that is clear.

File: kuma/static/js/components/mathml.js (hunk 1 above)
Comment (atopal) [reply kuma/static/js/components/mathml.js]: I don't understand why we are showing that message at all. Who does that help?

**Comment by atopal** [comment]
I think it's fine to have a degraded wiki experience for MathML editing for a short while, if the alternative is costly and fragile.

File: kuma/static/js/components/mathml.js (hunk 1 above)
Comment (schalkneethling) [reply kuma/static/js/components/mathml.js]: We do not have to, it is just how the current code functions. Easy to remove. It would then also be less code and less DOM interaction. Should I go ahead and remove @atopal?

**Comment by schalkneethling** [comment]
> I think it's fine to have a degraded wiki experience for MathML editing for a short while, if the alternative is costly and fragile.

This currently works on the `wiki.` side, it is one the "new" front end that it does not work.

File: kuma/static/js/components/mathml.js (hunk 1 above)
Comment (atopal) [reply kuma/static/js/components/mathml.js]: Yeah, go ahead, (take that Chesterton's fence ;)

**Comment by atopal** [comment]
> > I think it's fine to have a degraded wiki experience for MathML editing for a short while, if the alternative is costly and fragile.
> 
> This currently works on the `wiki.` side, it is one the "new" front end that it does not work.

Sure, my comment was related to the previous ones by Peter and Gregor. If this leads to a maintenance issue, I'd be okay with degrading the experience for a while on the wiki. It's a general statement to give you options. You have more information than I to make the right call here.

**Comment by peterbe** [comment]
> > I think it's fine to have a degraded wiki experience for MathML editing for a short while, if the alternative is costly and fragile.
> 
> This currently works on the `wiki.` side, it is one the "new" front end that it does not work.

So is this PR to help fix a bug in the read-only site? 
The Wiki is definitely on a rapid deprecation path, but the read-only front-end is (sadly) here to stay for a lot longer until Yari 2 becomes solid.

File: kuma/wiki/jinja2/wiki/react_document.html [hunk 6]
```diff
@@ -102,6 +102,7 @@
 
   <!-- site js -->
   {% javascript 'react-main' %}
+  {% javascript 'mathml' %}
```
peterbe: This is good and a nice tradeoff. This way, a crash won't block the rest of the JS (bundle) file. 

I still think it should be done with React instead of fragile DOM manipulations but if we just leave this as is, and that frees you up to work on what's more important, then I'm all for greenlighting the review of this. Thank you!

schalkneethling:
> So is this PR to help fix a bug in the read-only site?

Yup, the read-only side

kuma/static/js/components/mathml.js
schalkneethling: Aha! https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence

schalkneethling:
thank you for the review and feedback @peterbe, @Gregoor and @atopal - Updated with the latest.

PR #6561, opened 2020-02-21
Remove option to change email in sign up form

### LINKED ISSUE #6513: Remove hidden unnecessary hidden fields from signup.html
In socialaccount/signup.html there are three hidden fields for the username and email fields that unnecessary. These can be safely removed.

### LINKED ISSUE #6548: T - Remove option to *change* email in sign up form
Basically, Option 1 in https://docs.google.com/document/d/17FUhSfPQ5uPyA5VvctYFZZ0tPhc205gr4nZM8Nsu2KI/edit#

* No "change username" hyperlink.
* No `<input>` elements at all for email. Hidden or could-be hidden. 
* No form validation of email at all. 
* Remove the last two words on the paragraph line above "... choose username and email." to "... choose username."

Notes-to-self;

* There might be a slight conflict with the necessary work on https://github.com/mdn/kuma/issues/6541 since the form, before save, should display the *correct* email in the form-preview.
* Working on this issue, might "automatically" cover all of https://github.com/mdn/kuma/issues/6513 too.

peterbe:
Fixes #6548
Fixes #6513

Wow that became a big diff. All I wanted to do was just remove the "change email" thing and edit the "...chose username AND EMAIL ADDRESS" line. 
But it cut deep! 

To test, I highly recommend signing in to GitHub with the `mdnwebdocs-tester` (email `mdnwebdocs.tester@gmail.com`) account and notice that it never asks for any other email address choices and that ultimately it works and that the final `user.email` is that of https://github.com/settings/emails.mdnwebdocs-tester

It could very well be that there are some other minor more things I could have cut away. I'm not sure I want to obsess over that in this current time. 

@schalkneethling Can you check this PR fixes https://github.com/mdn/kuma/issues/6513 too?


kuma/users/signup.py
```diff
@@ -24,31 +24,20 @@ class SignupForm(BaseSignupForm):
...
 
-    It has an additional other_email form field to handle the case of GitHub
-    which may deliver a number of emails for users to choose from upon signup.
-
     The heavy lifting happens in the view.
     """
 
-    email = forms.CharField(
-        required=False, widget=forms.TextInput(attrs={"type": "email"})
-    )
```
peterbe: Removed because now it *is* required and what's left if you change that is that this is just want the parent class set already.

kuma/users/signup.py
```diff
@@ -57,39 +46,8 @@ def __init__(self, *args, **kwargs):
...
-        # if the selected email address is "other" we cut things short
-        # and clean the value in the form's clean method instead
-        if value == self.other_email_value:
-            return value
-
-        # otherwise we emmulate the functionality of the EmailField here
-        # stripping whitespaces, validating the content and then
-        # run allauth's own email value cleanup
-        self.cleaned_data["email"] = value.strip()
         validators.validate_email(value)
```
peterbe: This is now the only thing left and it might be that the parent form (i.e. the parent class) already does this and does it better. However, the parent form also messes with checking for dupes on the email which we don't care about in MDN. (...unfortunately, if you ask me)

kuma/users/tests/test_views.py
```diff
@@ -1132,75 +1131,44 @@ def test_email_addresses(self):
         profile_data = self.github_profile_data.copy()
         profile_data["email"] = public_email
         email_data = [
-            {"email": private_email, "verified": True, "primary": True},
+            # It might be unrealistic but let's make sure the primary email
+            # is NOT first in the list. Just to prove that pick that email not
+            # on it coming first but that's the primary verified one.
             {"email": unverified_email, "verified": False, "primary": False},
+            {"email": private_email, "verified": True, "primary": True},
```
peterbe: As of this change, I **think** this automatically answers https://github.com/mdn/kuma/issues/6541
What do you think @escattone ?

kuma/users/tests/test_views.py
```diff
@@ -1132,75 +1131,44 @@ def test_email_addresses(self):
...
-        profile_data["email"] = None
-        email_data = self.github_email_data[:]
-        private_email = "octocat.private@example.com"
-        email_data[0]["email"] = private_email
-        self.github_login(profile_data=profile_data, email_data=email_data)
-        response = self.client.get(self.signup_url)
-        assert response.status_code == 200
-        assert_no_cache_header(response)
-        assert response.context["form"].initial["email"] == private_email
-
```
peterbe: Death to all tests that test implementation details! It doesn't matter HOW the form works. Just that it works and the inputs match the expected outputs.

kuma/users/tests/test_views.py
```diff
@@ -1209,7 +1177,7 @@ def test_signup_public_github(self, is_public=True):
         data = {
             "website": "",
             "username": "octocat",
-            "email": "octo.cat@github-inc.com",
+            "email": "octocat-private@example.com",
```
peterbe: I know this test wasn't related but it became. Now that you're not allowed to POST in an email that isn't one of your OAuth emails (from the scope), all the tests that didn't adhere needed to change too.

kuma/users/tests/test_views.py
```diff
@@ -1307,12 +1275,29 @@ def test_signup_username_and_email_errors_event_tracking(self):
                         mock.call(
                             CATEGORY_SIGNUP_FLOW, ACTION_PROFILE_EDIT_ERROR, "username"
                         ),
-                        mock.call(
```
peterbe: As of this, your PR from last week is now going to be compliant with the new way of things. So, no need to reopen that issue/PR.

kuma/users/views.py
```diff
@@ -687,27 +670,21 @@ def form_valid(self, form):
...
-        if form.other_email_used:
-            data = {
-                "email": selected_email,
-                "verified": False,
-                "primary": True,
-            }
-        else:
-            data = self.email_addresses.get(selected_email, None)
+        if selected_email not in self.email_addresses:
+            return HttpResponseBadRequest("email not the default suggested one")
```
peterbe: This is basically a solution to "hackers" not allowing that it said in the `<input type="hidden" name="email" value="{whatwegotfrom oauth}">` to be what's submitted. So, I don't care how rough the error is. Just needs to reject and not trigger a Sentry crash or bother with a UI pleasant form validation message.

kuma/users/views.py
```diff
@@ -687,27 +670,21 @@ def form_valid(self, form):
...
+            email=data["email"], verified=data["verified"], primary=True
+        )
+        form.sociallogin.email_addresses = self.sociallogin.email_addresses = [
+            primary_email_address
+        ]
+        if data["verified"]:
+            # we have to stash the selected email address here
+            # so that no email verification is sent again
+            # this is done by adding the email address to the session
+            get_adapter().stash_verified_email(self.request, data["email"])
```
peterbe: I honestly still don't understand these lines :(
Does it matter?

kuma/users/views.py
```diff
@@ -687,27 +670,21 @@ def form_valid(self, form):
...
         selected_email = form.cleaned_data["email"]
-        if form.other_email_used:
-            data = {
-                "email": selected_email,
-                "verified": False,
-                "primary": True,
-            }
-        else:
-            data = self.email_addresses.get(selected_email, None)
+        if selected_email not in self.email_addresses:
```
escattone: As written, this doesn't work for signup via Google, because `self.email_addresses` is always an empty list for Google signups. In the `get_form` method above, the results for these two lines:
```python
        email = self.sociallogin.account.extra_data.get("email") or None
        email_data = (self.sociallogin.account.extra_data.get("email_addresses")) or []
```
are very different depending upon whether you signup with GitHub or Google. With GitHub, `email` will typically(always?) be `None` and `email_data` will be a non-empty list, while with Google it's the opposite, i.e. `email` will be the user's verified Google email and `email_data` will be an empty list.

I'm doubting that it still makes sense to keep the `self.default_email` idea, the distinction between `email` and `extra_email_addresses`. I'm wondering if we could just merge the `email` and `email_data` values into a single list of email addresses, use that for `self.email_addresses`, and drop `self.default_email`?

Also, right now we're just testing this with GitHub login, and it seems we should add an additional test for Google login since they're so different in their social-login data.

schalkneethling:
> @schalkneethling Can you check this PR fixes #6513 too?

There is this one that might also not be needed:
https://github.com/mdn/kuma/blob/master/kuma/users/jinja2/socialaccount/signup.html#L80

kuma/users/views.py
peterbe: I'll work on that. 

I'm disgusted that my change would have broken logging in with Gmail and I didn't notice. I.e. that the tests passed. Not sure if we're mocking incorrectly or we're not even testing it. 

I know I tested signing up with Gmail in this branch in my local browser. But perhaps I did  some code refactorings *after* having done a full end-to-end testing. 
Gross!

peterbe: Oh it gets worse!
I know I kinda rushed this PR on Friday when all I had was a half-day but guess what; not only did I managed to break signup for Google, it also breaks for GitHub!!

So, that means that my (bad) code changes, which had great intentions, broke both Google AND GitHub and none of our tests noticed.

peterbe:
> > @schalkneethling Can you check this PR fixes #6513 too?
> 
> There is this one that might also not be needed:
> https://github.com/mdn/kuma/blob/master/kuma/users/jinja2/socialaccount/signup.html#L80

Would you mind revising that comment? I think your link is saying we don't need the hidden username input. 

I would also strongly recommend you permalinks by linking to a specific line in a specific branch in a specific commit. When viewing the source, click the line "Copy permalink"
Screen Shot 2020-02-24 at 3 45 35 PM

If you're on a specific page, in the GitHub tree browser, and want to link to the whole page, at the current git sha, simply press `y` on your keyword and GitHub will change the current URL to the specific one.

peterbe:
@escattone I'm pleased with this now. 
The first thing I did was to add a happy-path test that uses Google and fully signs up. I can't believe we didn't have that before. 
Once in place, I better understood the relationship with the context data vs. the form submission. 

Also, this time, right before making the git commit, I started two Incognito browsers on `http://localhost.org:8000` and did fully sign up with GitHub *and* GitHub. It worked!

peterbe:
Note-to-self; as soon as this lands, return to https://github.com/mdn/kuma/pull/6530

schalkneethling:
@peterbe Don't worry about https://github.com/mdn/kuma/issues/6513, this cleans up all the pieces I was thinking off. Thanks

PR #6585, opened 2020-02-25
translateURL should be based on parent

### LINKED ISSUE #6536: "Help us translate this document" link not working
**Summary**
Screen Shot 2020-02-19 at 9 29 52 AM

**Steps To Reproduce (STR)**

Try https://developer.mozilla.org/sv-SE/docs/Web/JavaScript/Reference/Functions
You might need to be signed in. 

**Actual behavior**
That the`Snälla hjälp oss att översätta denna artikel från engelska` text is a clicklable link. 

**Expected behavior**
It's not clickable because the `<a>` element lacks a `href`. 

**Additional context**
Seems to happen in FirefoxNightly and latest stable Chrome.

peterbe:
Fixes #6536

If you're on a non-en-US page that is not fully translated, it now links to the wiki `$edit` URL, not the URL for choosing locale. Because, surely, if you're reading a non-en-US page (e.g. `sv-SE` in the current URL) then you shouldn't need to choose locale :)
This fix makes it so that that link works the same as the Wiki except in the read-only the link hops over to the Wiki.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,4 @@ kuma/static/js/libs/ckeditor/source/ckbuilder/
 
 # Emacs backup files
 *~
+junit.xml
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -165,10 +165,10 @@ def document_api_data(doc=None, redirect_url=None):
             "wikiURL": absolutify(doc_absolute_url, for_wiki_site=True),
             "translateURL": (
                 absolutify(
-                    reverse("wiki.select_locale", args=(doc.slug,), locale=doc.locale,),
+                    reverse("wiki.edit", args=(doc.slug,), locale=doc.locale,),
                     for_wiki_site=True,
                 )
-                if doc.is_localizable
+                if doc.parent and doc.parent.is_localizable
                 else None
             ),
             "translationStatus": translation_status,
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -171,15 +171,10 @@ def document_api_data(doc=None, redirect_url=None):
             ),
             "translateURL": (
                 absolutify(
-                    reverse(
-                        "wiki.select_locale",
-                        args=(doc.parent.slug,),
-                        locale=doc.parent.locale,
-                    ),
+                    reverse("wiki.select_locale", args=(doc.slug,), locale=doc.locale),
                     for_wiki_site=True,
                 )
-                + f"?{urlencode(dict(tolocale=doc.locale))}"
-                if doc.parent and doc.parent.is_localizable
+                if doc.is_localizable
                 else None
             ),
             "translationStatus": translation_status,
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -171,7 +171,11 @@ def document_api_data(doc=None, redirect_url=None):
             ),
             "translateURL": (
                 absolutify(
-                    reverse("wiki.select_locale", args=(doc.slug,), locale=doc.locale),
+                    reverse(
+                        "wiki.select_locale",
+                        args=(doc.parent.slug,),
+                        locale=doc.parent.locale,
+                    ),
                     for_wiki_site=True,
                 )
                 + f"?{urlencode(dict(tolocale=doc.locale))}"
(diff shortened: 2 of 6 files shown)


kuma/api/v1/views.py
```diff
@@ -165,10 +165,10 @@ def document_api_data(doc=None, redirect_url=None):
             "wikiURL": absolutify(doc_absolute_url, for_wiki_site=True),
             "translateURL": (
                 absolutify(
-                    reverse("wiki.select_locale", args=(doc.slug,), locale=doc.locale,),
+                    reverse("wiki.edit", args=(doc.slug,), locale=doc.locale,),
```
escattone: Good point, but I think this should use the `wiki.translate` endpoint with the `tolocale` query parameter instead. Something like:
```python
from urllib.parse import urlencode
...
    reverse("wiki.translate", args=(doc.parent.slug,), locale=doc.parent.locale,) + f"?{urlencode(dict(tolocale=doc.locale))}",
```

peterbe: Hmm.... I think it goes deeper than that. In the wiki, [it uses the "edit URL"](https://github.com/mdn/kuma/blob/f41bbd1428d97c6ebed9b0e247379bc7d0999cfd/kuma/wiki/jinja2/wiki/document.html#L207) but it does that because it knows that that document translation has already started. So, I think you're right in that the "translate URL" should be what you point out but I think the overall code needs to a little smarter.

peterbe: Actually, are you sure? The Wiki uses `...$edit` for the same piece of functionality. 

[The Wiki presents that](https://github.com/mdn/kuma/blob/f41bbd1428d97c6ebed9b0e247379bc7d0999cfd/kuma/wiki/jinja2/wiki/document.html#L207) 
based on these conditions:
1. `document.parent_id and document.current_revision` and 
2. `document.current_revision.localization_in_progress` and 
3. `document.current_revision.translation_age >= 10`

That's basically how it figures out that the (localized) document is "This translation is incomplete."

The equivalent in the `document_api_data` is these lines of code:
https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/api/v1/views.py#L150-L153
which means it ultimately sets ` translation_status == "outdated"`, which, when you look at its usage: https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/javascript/src/article.jsx#L27-L34

I.e. The Wiki presents the `$edit` URL if the localization is considered "outdated" (alias "in progress")
Therefore I think the read-only site should do the same?

**NOW** That's the *naive* thing. The document's "translate URL" is `...$translate?tolocale=sv-SE` and it gets used in the read-only site correctly in the [language dropdown menu](https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/javascript/src/header/language-menu.jsx#L51-L61)

**SO** In conclusion, what we need to do is go back to `article.jsx` and NOT use the `translateURL` key at all. But instead, use the Wiki edit URL. I.e.:


1. This English page (that you're reading) hasn't been translated to French, click `wiki.d.m.o/fr/docs/Slug?tolocale=fr`
2. This French page (that you're reading right now) has been translated but it's in progress, click `wiki.d.m.o/fr/docs/Slug$edit`

Yeah. That's got to be correct. Then means we need another URL in the document_api_data. The `$edit` URL! We have the `wikiURL` but not the `editURL`. So I'll add that.

peterbe:
@escattone Finally, I'm pleased with this. 
Now the read-only message and link matches how the Wiki works. One could argue that if the localized content is "in progress" there ought to be a link too to keep working on that, but there isn't. Again, this matches how the Wiki work(ed). :)

I *could* have used `editURL = wikiURL + '$edit'` but that felt fragile. 

WHY DID THIS SEEMINGLY HARMLESS LITTLE ISSUE TAKE TWO DEVELOPERS TWO DAYS!!?

.gitignore
```diff
@@ -104,3 +104,4 @@ kuma/static/js/libs/ckeditor/source/ckbuilder/
 
 # Emacs backup files
 *~
+junit.xml
```
peterbe: This change is for when you run `jest` and it fails to execute. For example, on my macOS I now have Node 13 which has compatibility problems with the `node-sass` package. So every time I try to run `jest -b` on the host it leaves this file.

kuma/api/v1/views.py
escattone: @peterbe Thanks for your excellent, detailed analysis! I totally agree with the action plan in your last comment, except for one thing, the `translateURL` should use the locale and slug of the parent document (really, always `en-US` and the English slug), so `wiki.d.m.o/en-US/docs/Web/Tutorials?tolocale=fr` instead of `wiki.d.m.o/fr/docs/Web/Tutoriels?tolocale=fr`, so the same code I suggested in https://github.com/mdn/kuma/pull/6585#discussion_r384197148.

Issue #6573, opened 2020-02-24
"Submit as: Spam" shouldn't be exposed to users


alattalatta:
**Summary**
_What is the problem?_
The revision dashboard has an Akismet column with "Submit as: Spam" row-buttons. It doesn't work, which implies I don't have permissions to use it. (I saw @chrisdavidmills had marked one for [Getting started with the Web](https://wiki.developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web$history))

**Steps To Reproduce (STR)**
_How can we reproduce the problem?_

1. Visit Revision Dashboard without administrative permissions.

**Actual behavior**
_What actually happened?_
Spam tagging buttons are there, and nothing happens when I click. No HTTP requests are initiated.

**Expected behavior**
_What did you expect to happen?_
No spam tagging buttons exposed. (or, do something when clicked)

peterbe:
According to the code, the "Submit as: Spam" row depends on this permission check:
```python
    show_spam_submission = request.user.is_authenticated and request.user.has_perm(
        "wiki.add_revisionakismetsubmission"
    )
```
So, if you see it, you must have permission to add a AkismetSubmission.

alattalatta:
That... is strange. I didn't know I had such permission. Sorry!

For the 'button not initiating any requests' part - there's an error in the console.
`TypeError: mdn.utils is undefined` (it's trying to use `getCookie` function from it)

`mdn.utils` is defined in `js/utils/utils.js` file, which is bundled with `wiki.js`. And dashboard doesn't use `wiki.js`.

I suspect the fix would be to add `js/utils/utils.js` to `common.py`'s `'dashboard'` bundle setting. Should I change this one, or open another issue?

peterbe:
Thank you @alattalatta 
The problem you've stumbled on now is actually the same as https://github.com/mdn/kuma/issues/6550 is about. That issue is a P2 issue so it'll get done very soon. 

Can we close this now? I think so.

alattalatta:
Sure! Thank you for your confirmation.

PR #6587, opened 2020-02-25
don't use async scripts

### LINKED ISSUE #6550: I set default language english in profile, but site's language always my local.
@hzx commented on [Thu Feb 20 2020](https://github.com/mdn/sprints/issues/2837)

# Request type

- [ ] Show site in language I set in my profile. In addition even after selecting default language in popup.

# Details
I set default language english in profile, but site's language always my local.
Screen Shot 2020-02-21 at 02 34 40

Screen Shot 2020-02-21 at 02 35 22

Screen Shot 2020-02-21 at 02 40 17

Screen Shot 2020-02-21 at 02 41 51

---

@chrisdavidmills commented on [Fri Feb 21 2020](https://github.com/mdn/sprints/issues/2837#issuecomment-589607478)

This is more of a platform issue, so I'll move this over to the kuma repo.

peterbe:
Fixes #6550

It's a miracle this hasn't caused more bugs. Perhaps it has caused hundreds of thousands of bugs  due to the progressive-enhancement nature of non-reactive JS, people have simply not noticed and they just refresh and get luckier the next time. 

You should almost never use `async` on a script tag. The only time you should is when you're 101% comfortable with it potentially executing before other .js file has NOT been executed. 

For example:
```html
<script src="/jquery.js"></script>
<script src="/mdn.js"></script>
<script async src="/main.js"></script>
```
is dangerous, **if** the contents of `main.js` is something like this:
```javascript
$(function() { // hi! 
});
```
or...
```javascript
if (window.mdn.utils.getCookie('foo')) {
  ...
}
```

The only time you should use `async` is if...

1) You're sure it doesn't depend any JS inline, or script tags. Doesn't matter if the `<script>` tag appears before or after the `<script async>` one. 
2) You're ok with it blocking the rendering process of the page. 
3) You're ok with the page not being fully rendered and/or `DOMContentLoaded` event signals have fired. 

It's a rare optimization trick that has deep implications. For the Wiki I can't imagine any reason any `<script>` tag should be `async` (other than `perf.js` or `analytics_debug.js`) .

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -901,7 +901,6 @@ PIPELINE_JS = {
             "js/components/banners/banners.js",
         ),
         "output_filename": "build/js/banners.js",
-        "extra_context": {"async": True},
     },
     "mathml": {
         "source_filenames": ("js/components/mathml.js",),
@@ -911,12 +910,10 @@ PIPELINE_JS = {
     "users": {
         "source_filenames": ("js/users.js",),
         "output_filename": "build/js/users.js",
-        "extra_context": {"async": True},
     },
     "user-signup": {
         "source_filenames": ("js/components/user-signup/signup.js",),
         "output_filename": "build/js/signup.js",
-        "extra_context": {"async": True},
     },
     "delete-user-page": {
         "source_filenames": (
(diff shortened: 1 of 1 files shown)


sideshowbarker:
I’m still running into a bug that seems to have been caused by the problem this change was made to fix — specifically, I still get this:

> TypeError: can't access property "getCookie", mdn.utils is undefined

…when trying to use the **Submit as: spam** button on any Revision Dashboard page (same bug that was reported in #6573)

PR #6598, opened 2020-02-26
Possibly to prefill Reason textarea on $delete

### LINKED ISSUE #6597: Possibly to prefill Reason textarea on $delete
If you're going to have some human-helped automation that deletes the 20% or so non-en-US documents that aren't actually translated, it would be nice if you don't have to type in the same reason every time. 
Screen Shot 2020-02-26 at 11 56 40 AM

Would be neat if you could control it with something like `?reason=Sample%20reason`

peterbe:
Fixes #6597

### DIFF
--- a/kuma/wiki/views/delete.py
+++ b/kuma/wiki/views/delete.py
@@ -100,7 +100,8 @@ def delete_document(request, document_slug, document_locale):
             document.delete()
             return redirect(document)
     else:
-        form = DocumentDeletionForm()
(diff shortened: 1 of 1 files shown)


peterbe:
Screen Shot 2020-02-26 at 12 19 03 PM

Most importantly, the context is that email thread with the subject `How to get rid of these Untranslated translations on MDN???` which this enables.

PR #6463, opened 2020-02-11
Switch to async GA tag

mindy:
Decided to load the async script tag for everyone. Conclusion reached in https://github.com/mdn/kuma/issues/6253#issuecomment-584340062

Closes #6253

PR #6601, opened 2020-02-26
Revert "Switch to async GA tag (#6463)"

mindy:
We suspect that this change is causing a drop in performance, from 3.4s to 3.7s for the 90th percentile. [Slack convo for reference.](https://mozilla.slack.com/archives/CLTBV3W9X/p1582711416043200)

This reverts commit 669ec0a87ddc44045d4593c986b4d102b6717668.


callahad:
Worth at least trying this to see if it is indeed the culprit

PR #6608, opened 2020-02-27
Update componentName in SSR test

erkde:
Commit fcac45bcbb changed the switch statement on `componentName` in index.jsx and ssr.jsx from having a case for `document` to `SPA`.


schalkneethling:
@peterbe Can you take a look?

PR #6612, opened 2020-02-28
Fix a ResourceWarning: unclosed file in our tests

callahad:
Failing to consume or close a StreamingHttpResponse raises a
ResourceWarning: unclosed file.

This fixes that.


callahad:
Yep! Python hides ResourceWarning by default, but they're visible if `PYTHONWARNINGS` is set... Just feels nice to be as warning-free as possible :)

PR #6628, opened 2020-03-03
Add `isLoading` prop to React Router state 

mindy:
Currently, the visibility of the loading animation is tied to the component's data. It shows by default, and then is hidden when either of these cases occur: 1) a Promise resolves with a non-null value, or 2) initial data is passed to the component. 

In some cases, such as the Subscriptions landing page (https://github.com/mdn/kuma/issues/6582), neither case will be true, so the loading animation always shows. 

The suggestion in the comments from David was:
> // TODO: update Router to handle the null return case

But I think that data and the loading state should be de-coupled, and therefore added an additional `isLoading` prop to `PageState`.  It is set to `true` and changes to `false` when the Promise resolves or if we have the initial data we need to render the component. 

To test:
- Check out the landing page branch (https://github.com/mindy/kuma/tree/landing-6582) and go to /payments. Without these routing changes, the animation persists, but with the changes, the animation is removed, as expected. I thought it would be easier to review if the routing changes were separate from Subscriptions, so that's why it's in a separate diff. 
- For this branch, I checked to make sure Search and Documents still loads correctly, since those are the only two sections that use this. 

Parent issue:
https://github.com/mdn/kuma/issues/6581

Context (thanks to @schalkneethling for all of this info!):
https://github.com/mdn/kuma/issues/6581#issuecomment-591378120

### DIFF
--- a/kuma/javascript/src/router.jsx
+++ b/kuma/javascript/src/router.jsx
@@ -73,7 +73,6 @@ export default function Router({
     };
 
     // Router state: this is the data we'll use below to render the page
-    //
     let [pageState: PageState, setPageState] = useState({
         url: null,
         route: null,
--- a/kuma/javascript/src/router.jsx
+++ b/kuma/javascript/src/router.jsx
@@ -66,16 +66,21 @@ export default function Router({
         params: ?RouteParams,
         // Data fetched asynchronously by the Route.fetch() function.
         // This value will be passed to the component as the data prop.
-        data: ?RouteData
+        data: ?RouteData,
+        // The loading state of the page. Defaults to `true`, as it should
+        // always appear until the component finishes rendering
+        isLoading: ?Boolean
     };
 
     // Router state: this is the data we'll use below to render the page
+    //
     let [pageState: PageState, setPageState] = useState({
         url: null,
         route: null,
         component: null,
         params: null,
-        data: null
+        data: null,
+        isLoading: true
     });
 
     // We also need to access the current page state from our event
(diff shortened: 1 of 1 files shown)


kuma/javascript/src/router.jsx
```diff
@@ -66,16 +66,21 @@ export default function Router({
...
         // This value will be passed to the component as the data prop.
-        data: ?RouteData
+        data: ?RouteData,
+        // The loading state of the page. Defaults to `true`, as it should
+        // always appear until the component finishes rendering
+        isLoading: ?Boolean
     };
 
     // Router state: this is the data we'll use below to render the page
+    //
```
schalkneethling: Whoops :)

mindy: Fixed :)

kuma/javascript/src/router.jsx
```diff
@@ -66,7 +66,10 @@ export default function Router({
         params: ?RouteParams,
         // Data fetched asynchronously by the Route.fetch() function.
         // This value will be passed to the component as the data prop.
-        data: ?RouteData
+        data: ?RouteData,
+        // The loading state of the page. Defaults to `true`, as it should
+        // always appear until the component finishes rendering
```
peterbe: That line isn't correct. The "loading" state has nothing to do with rendering. It has to do with fetching.

peterbe:
After Zoom-chatting to Mindy our joint plan is to redirect our attention on https://github.com/mdn/kuma/pull/6635 which is part optimization and part necessary warm-up for her Subscription landing page component.

Issue #6579, opened 2020-02-24
T - Track successful subscription signups in whoami


tobinmori:
**Description:** 

We need to track successful signups so that we don't bother subscribing users with the banner. 

**Acceptance Criteria**
- [ ] Upon successful Stripe payment, store in whoami that the user is currently a subscriber
- [ ] need a new field in User obj e.g. is_current_subscriber

**Possible solutions**
- regularly ping Stripe API for subscription status
- use Stripe webhooks for cancellation, charge failure events

PR #6633, opened 2020-03-04
Add subscriptions banner

### LINKED ISSUE #6577: T - Subscription Banner
**Parent User Story:** #6554

**Description:**

The Subscription banner advertises requests for contributions to MDN

**Acceptance Criteria:**
- [x] banner shows on MDN Doc pages and homepage
- [x] banner only displays to logged in users (waffle)
- [x] banner only shows to subset of users (e.g. 50%)
- [x] banner only shows on desktop (not mobile)
- [x] banner is shown on top of the content at the bottom of the view port. (note: The newsletter CTA is shown at the bottom of the page. We might change the newsletter CTA to be a CTA for the monthly subscription, but that is independent of the banner.)
- [x] banner links to FAQ
- [x] banner links to edit profile page where user can sign up for subscription
- [x] banner has option for dismissal
- [x] dismissing banner keeps it quiet for 7 days (using local storage to track this is fine)

**Design**
https://app.abstract.com/projects/9decc480-70fc-11e9-9453-4f3fbffea7ce/branches/64547606-339e-494d-8d45-559aa506d80a/commits/6ae7cc2467cf36a609b9b2bc5189dabf9c5c4e7e/files/96E76C7B-2AF8-4473-A39E-E81890F91955/layers/E3B8DB66-C6A4-4BDC-A2FF-A80A02DAC2E8?collectionId=0750a16f-5979-4db7-9b1b-d4d676607dcd&collectionLayerId=2536d7fc-cd8e-4b9b-9580-615dc1f76e70&mode=design

**NOTE**
- ideally build this simply first so that the dependent work can get started (pre-Stripe modal)

schalkneethling:
This adds the new subscription banner to the static site. The banner will be displayed on the homepage and documentation pages.

The banner is hidden behind the `subscription_banner` waffle flag and is only shown to logged-in users.

fix #6577


kuma/javascript/src/banners.jsx
```diff
@@ -114,43 +114,53 @@ export type BannerProps = {
...
                     <h2 className="mdn-cta-title slab-text">{props.title}</h2>
                     <p className="mdn-cta-copy">{props.copy}</p>
                 </div>
                 <p className="mdn-cta-button-container">
                     <a
                         href={props.url}
                         className="mdn-cta-button"
-                        target="_blank"
-                        rel="noopener noreferrer"
+                        {...(props.newWindow && {
```
schalkneethling: @Gregoor Flow is going to complain about this. Not sure if there is a better way to accomplish this that Flow will also be happy with or, whether we should quiet Flow for this one.

I tried various approaches. Changing the prop definition to `Boolean` removes the error here but then introduces it further down at line 112. Changing that line to be say `Boolean(true)` moves the error to line 258 :)

peterbe:
I added the `do not merge` because this PR is blocked by...

* https://github.com/mdn/kuma/pull/6535
* https://github.com/mdn/kuma/issues/6579

the second is important. The code in this PR will happily display the banner to already subscribing users. Having `userData.isSubscriber` or something in the whoami would be critical to the inclusion of the banner's presence.

kuma/landing/jinja2/landing/react_homepage.html
```diff
@@ -11,7 +11,14 @@
 
   {% if waffle.switch('developer_needs') %}
     {% stylesheet 'banners' %}
+    {% stylesheet 'banner_developer_needs' %}
   {% endif %}
+
+  {% if waffle.flag('mdn_subscriptions') %}
```
peterbe: This will never work. You can't use `waffle.flag` in any Jinja template that is used on the read-only site. It was always be false. The only decent solution, for now, is to have this:
```
{% stylesheet 'banners' %}

{% if waffle.switch('developer_needs') %}
{% stylesheet 'banner_developer_needs' %}
{% endif %}

{% stylesheet 'banner_mdn_subscriptions' %}
```
Yes, it means every single page, authenticated or not, waffled on or not, will have `styles/components/banners/mdn-subscriptions.css` added to it. So, **all pages will become slower when this is landed in master and in production**.

peterbe: So actually, as of this, every page now needs to ship with:

* `http://localhost.org:8000/static/build/styles/banners.92084d28db45.css`  **2.0K**, 701B gzipped
* `http://localhost.org:8000/static/build/styles/mdn-subscriptions.bf637faaf5b0.css` **868B**, 393 gzipped

So it's not a huge increase by any means but it's definitely something to be aware of since we've often talked about the bloat of our CSS be one of the biggest culprits of inferior First Paint Time.

schalkneethling:
> I added the `do not merge` because this PR is blocked by...
> 
>     * #6535
> 
>     * #6579

Very true and I did mention this. Not sure what exactly happened to that in the end but, my understanding is that this iteration does not need to be perfect/complete.

The idea here, for this sprint, is to have the base implemented behind a waffle(also the reason why we need to separate flags/switches) and then iterate. While those two items are very important, they should not block this from being merged.

kuma/javascript/src/banners.jsx
```diff
@@ -211,6 +242,18 @@ export default function Banners(props: BannersProps) {
         let bannersList = props.banners || defaultBanners;
         for (const banner of bannersList) {
             if (userData.waffle.flags[banner.id]) {
+                // if this banner is to be shown to authenticated users only,
+                // and the current user is not authenticated, return `null`
+                if (banner.authenticated && !userData.isAuthenticated) {
+                    return null;
```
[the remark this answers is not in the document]
schalkneethling: After merging in the changes and working on resolving the test failures I realized that:

1. I needed to add back the `authenticated` property or else, all banners would only be shown to authenticated users.
2. While we will only be showing the subscriptions banner now, we will later in the year have at least the developer needs banner as well. At that point, the subscription banner will most likely be shown to everyone but, we might have a different banner we only want to show to authenticated users(perhaps something around a new feature only available to logged-in users/subscribed users). I am therefore suggesting we use a more general property to indicate that a banner is to be shown to authenticated users only. Who knows, we might even need a `subscriber` property as well later on 😸.
3. When a banner is active but should not be shown to the current user, I do not believe this to be an error, and we can simply return `null` and return to the loop, so I removed the `Error` and replaced it with the previous `return null`

kuma/javascript/src/banners.jsx
```diff
@@ -211,6 +245,18 @@ export default function Banners(props: BannersProps) {
...
+                }
+
+                // special case for the subscriptions banner as we need access
+                // to the username as it forms part of the `url`
+                if (banner.id === 'mdn_subscription' && userData.username) {
```
schalkneethling: Partially in reference to https://github.com/mdn/kuma/issues/6577#issuecomment-595774644 I switched this back to using the `mdn_subscriptions` flag as there will be a period where we will show the form on edit profile, but not yet show the banner on other pages. This means we have a need for two independant switches.

kuma/javascript/src/banners.jsx
```diff
@@ -114,43 +114,51 @@ export type BannerProps = {
...
     id: string,
+    // class name used on main banner container. Exclusively used
+    // for styling purposes.
+    classname: string,
     // The banner title. e.g. "MDN Survey"
     title: string,
     // The banner description. e.g. "Help us understand the top 10 needs..."
-    copy: string,
+    // Could also be a React Element such as that returned by `<Interpolated />`
+    copy: Object | string,
```
Gregoor:
```suggestion
    copy: React.ReactNode | string,
```
I think that should also work and narrows the type down a bit more.

kuma/javascript/src/banners.jsx
```diff
@@ -189,20 +200,43 @@ export default function Banners(props: BannersProps) {
         () => [
             {
```
Gregoor: Given the static nature of this information I'd recommend moving it outside of the component fn and to the top-level of this module.

kuma/javascript/src/banners.jsx
Gregoor: Oh I think my mind was in TypeScript mode. It might be `React.Node` in Flow

peterbe:
> > I added the `do not merge` because this PR is blocked by...
> > ```
> > * #6535
> > 
> > * #6579
> > ```
> 
> Very true and I did mention this. Not sure what exactly happened to that in the end but, my understanding is that this iteration does not need to be perfect/complete.
> 

I didn't see it in the PR description. Sorry. 

What is it that can be not-perfect or not-complete? I don't quite understand where we are able to compromise? I can see that there might be a need to not perfect unit tests for example. But links and functionality and stuff ought to work, no?

If there are areas within the code that is "deliberately" imperfect it needs to be prefixed with a code comment that explains the justification for the imperfection. 

> The idea here, for this sprint, is to have the base implemented behind a waffle(also the reason why we need to separate flags/switches) and then iterate. While those two items are very important, they should not block this from being merged.

I'd rather not merge until it works. If it's broken, what's the point of putting it in `master`. And if the PR isn't intended to be merged into master we should make it a draft PR instead. 


To me, the pressing issues to be resolved are:

1. If we desire to have the banner launched separately from the basic Subscription form, then THIS waffle flag should be called `subscription_banner` not `mdn_subscriptions`. 
2. Sorry for being dim but I still can't understand why the business logic needs to do an additional check if you're authenticated. The waffle FLAG is already dependent on being logged in. 
3. If we're ever going to allow the banner to be for anonymous users, then it's weird to link to the wiki edit-profile URL and expect the user to understand that they need to sign up first. I'd much rather we leverage the Subscription landing page or something.
4. The links need to be made absolute. Not the `/$locale/payments` link because that stays on the read-only site and that is blocked by @mindy 's ["Subscription landing page"](https://github.com/mdn/kuma/pull/6647) So that's the blocker for this PR.

schalkneethling:
>     2\. Sorry for being dim but I still can't understand why the business logic needs to do an additional check if you're authenticated. The waffle FLAG is already dependent on being logged in.

So, the way it was here https://github.com/schalkneethling/kuma/pull/5/files#diff-fe2457a03c177a0968f418fa6371dd34R246 banners would only ever be shown to users who are logged in. Seeing that we have two potential banners, we need a way to indicate that the one(subscriptions) is for authenticated users only, and the other(devneeds) is not.

So, only knowing that a flag is active does not tell us whether the banner is meant for authenticated eyes only.

schalkneethling:
> I didn't see it in the PR description. Sorry.

No worries. This was not in the PR description, it was most likely in the big Google Doc we used to iterate over some of the open questions.

schalkneethling:
> What is it that can be not-perfect or not-complete? I don't quite understand where we are able to compromise? I can see that there might be a need to not perfect unit tests for example. But links and functionality and stuff ought to work, no?

What I mean here was that _this_ PR does not have to be perfect as we will not switch on this banner as soon as it is merged. This is the first portion of the work the get the foundation in place.

With that said, we could let this one sit for a while while the blockers are still in the works, and once merged, this one can be updated as needed.

schalkneethling:
>     3\. If we're ever going to allow the banner to be for anonymous users, then it's weird to link to the wiki edit-profile URL and expect the user to understand that they need to sign up first. I'd much rather we leverage the Subscription landing page or something.

This will not be the case for unauthenticated users. Once we are at that stage of the evolution, the payment flow will be done via a modal. With that said, before the user can actually signup for a subscription, they will have to create/have an account :)

i.e. the flow will be more complex/involved for users without an account.

schalkneethling:
>     1. If we desire to have the banner launched separately from the basic Subscription form, then THIS waffle flag should be called `subscription_banner` not `mdn_subscriptions`.

It has been done.

peterbe:
What this PR needs is a default waffle flag as a db migration. I can help with that. 
Make it entirely disabled for everyone or enabled for all staff users?

schalkneethling:
> What this PR needs is a default waffle flag as a db migration. I can help with that.

Let me take a stab at it. I believe I have done one before :)

> Make it entirely disabled for everyone or enabled for all staff users?

Let's just make it disabled for everyone by default. Kadir can then tweak it as needed via the Django admin. Thanks, Peter!

kuma/javascript/src/banners.jsx
```diff
@@ -211,6 +245,18 @@ export default function Banners(props: BannersProps) {
...
+                // if this banner is to be shown to authenticated users only,
+                // and the current user is not authenticated, return `null`
+                if (banner.authenticated && !userData.isAuthenticated) {
+                    return null;
+                }
+
+                // special case for the subscriptions banner as we need access
+                // to the username as it forms part of the `url`
+                if (banner.id === 'subscription_banner' && userData.username) {
+                    banner.url = `/${locale}/profiles/${userData.username}/edit#subscription`;
```
peterbe: This needs to link to the Wiki.

schalkneethling: So, something like:

```
const wikiURL = "https://wiki.developer.mozilla.org";
...
banner.url = `${wikiURL}/${locale}/profiles/${userData.username}/edit#subscription`;
```

escattone: @schalkneethling Is `win.mdn.wikiSiteUrl` available from the React pages?

escattone:
> I noticed that I can happily go to https://developer.mozilla.org/en-US/profiles/peterbe/edit and it doesn't require a redirect to the Wiki. It's just that I get the Wiki header.
> But it does make me nervous. Because what if the form there, which is Wiki functionality, doesn't work if the HOST header isn't right.

That entire page, not just the header, is non-React code, even when displayed on the read-only site, which isn't an issue since those pages are **never** cached by our CDN (since they're always specific to a user). So we're safe here to use this page with either the read-only or Wiki site.

> Perhaps @escattone or @Gregoor knows if it's safe to use the Stripe stuff on the read-only domain.

I don't know enough about the Stripe code to answer this, so I'll defer to @Gregoor there.

schalkneethling:
@Gregoor @peterbe based on @escattone's comment here https://github.com/mdn/kuma/pull/6633#issuecomment-597252676, are we good to go?

kuma/javascript/src/banners.jsx
schalkneethling: Let me give that a go. I was unsure of which portions of this happen on the server and which parts happen on the client. If on the server we will not but, the call to get `userData` happens inside a `useEffect` so we might be safe here.

schalkneethling: > @schalkneethling Is `win.mdn.wikiSiteUrl` available from the React pages?

It works o/\o

kuma/core/migrations/0004_create_subscription_banner_flag.py
```diff
@@ -0,0 +1,23 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def create_subscription_banner_flag(apps, schema_editor):
+    Flag = apps.get_model("waffle", "Flag")
+    Flag.objects.get_or_create(
+        name="subscription_banner", staff=True, note="Shows the subscription banner"
+    )
```
peterbe: If this created the flag, `Flag.get_all()` will not pick it up and you have to wait 1 hour. 
So what you do is:
```python
flag, created = Flag.objects.get_or_create(
        name="subscription_banner", staff=True, note="Shows the subscription banner"
    )
flag.flush()
```

peterbe: Totally a nit but good practice. 
Also, I think we should have do it something like this:
```
if not Flag.objects.filter(name="subscription_banner").exists():
    flag = Flag.objects.create(
        name="subscription_banner", staff=True, note="Shows the subscription banner"
    )
    flag.flush()
```

kuma/javascript/src/banners.jsx
peterbe: Be careful there. Banners are server-side rendered. What's happening here is that they are *rendered* twice. Once on the server and once in the client. Actually, if you're logged in it's twice on the client. I.e. we're failing to react hydrate. This might have some performance degradations but if so very little. After all, in the server it rendered nothing and in the client it renders something. This is the same we're doing with the "Sign in" link. So the DOM, that the virtual DOM needs to *mutate*, is possibly a harmless (and necessary!) mutation. 
What I'm not entirely sure of is what React does when it freaks out during hydration. If it notices that the DOM doesn't match what the virtual DOM ought to be, does it trigger a full new rendering? I know, in development, it can lead to `console.warn` at most. 

Just be careful when using globals that can never exist in SSR. 

I think it's overkill right now but I think we should never do banners in server-side because of the nature of them potentially being 100% client-side. Any banner that depends on walffe FLAGs (not switches) or user ought to be started from a `useEffect` instead of the plain rendering path.

schalkneethling:
Python migration updated. Thank you. @peterbe, @Gregoor, and @escattone

peterbe:
Weird. How come this works:
```
>>> flag = Flag.objects.all().first()
>>> flag.flush()
>>>
```

peterbe:
Also, why does this work:
```
>>> flag = Flag.objects.create(
...             name="subscription_bannerXXX", staff=True, note="Shows the subscription banner"
...         )
>>> flag.flush()
>>>
```

peterbe:
I honestly don't know why that is how it works. It's the model is a `<class '__fake__.Flag'>` thing when outside of migrations it's `<class 'django.db.models.base.ModelBase'>`.
I just don't know how to proceed and thankfully it's not worth it. We just have to accept the caching as is. But the pattern of `if not exists(name=name): create(name=name, other=things)` is a good pattern. 
So just delete the line `flag.flush()`

kuma/core/migrations/0004_create_subscription_banner_flag.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import migrations
+
+
+def create_subscription_banner_flag(apps, schema_editor):
+    Flag = apps.get_model("waffle", "Flag")
+    if not Flag.objects.filter(name="subscription_banner").exists():
+        flag = Flag.objects.create(
+            name="subscription_banner", staff=True, note="Shows the subscription banner"
+        )
+        flag.flush()
```
peterbe:
```suggestion
```

let's just drop that. We'll take that battle some other day.

schalkneethling:
Thanks, @peterbe - updated ☮️

kuma/core/migrations/0004_create_subscription_banner_flag.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def create_subscription_banner_flag(apps, schema_editor):
+    Flag = apps.get_model("waffle", "Flag")
+    if not Flag.objects.filter(name="subscription_banner").exists():
+        flag = Flag.objects.create(
```
peterbe:
```suggestion
        Flag.objects.create(
```

peterbe:
@Gregoor If you're OK with people using the Subscription form, using the Wiki header& footer, on the read-only domain, then please go ahead and merge this.

peterbe:
@Gregoor  Actually, perhaps not merge it! But approve it. I'll leave it to @schalkneethling and @atopal  to figure out if this is OK to have in `master`.

atopal:
> @Gregoor If you're OK with people using the Subscription form, using the Wiki header& footer, on the read-only domain, then please go ahead and merge this.

I didn't read all of the conversation up to this point. Can you expand on that a bit? I wouldn't mind that on the profile, but are you saying, they'd see the wiki header on every page?

schalkneethling:
> > @Gregoor If you're OK with people using the Subscription form, using the Wiki header& footer, on the read-only domain, then please go ahead and merge this.
> 
> I didn't read all of the conversation up to this point. Can you expand on that a bit? I wouldn't mind that on the profile, but are you saying, they'd see the wiki header on every page?

The only items I see that are “missing” is that on the edit profile page we have the wiki header. Then currently, we will show the banner to people even if they already have a subscription. With that said, it is behind a waffle flag for staff only.

It also links to the current FAQ(wiki), but I do not think there is anything new with regards to the FAQ Mindy is working on other than the fact that it is in React, for right now anyhow.

There is also some code refactoring we would like to do, but that will not impact any of the user-visible aspects

schalkneethling:
Merging this now as:

1. There are two approving reviews
2. This is becoming a blocker for further work
3. It will be behind a flag and therefore not shown to user even after a production push

Thanks you for all the hard work on reviewing all of this @peterbe @Gregoor @escattone 🍵

Issue #6623, opened 2020-03-03
KeyError username on signup with GitHub


peterbe:
https://sentry.prod.mozaws.net/operations/mdn-prod/issues/7399307/

```
[...]

PR #6637, opened 2020-03-05
fix username keyerror (github kuma issue #6623)

escattone:
@peterbe This was a thought experiment for partly addressing #6623 (it doesn't address the failure to recover existing accounts, just the immediate server error) in a way that is more radical and cleans-up some code that has been bothering me for some time, but I'm not sure I like it. I'm submitting it as an example for discussion.

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -599,31 +599,31 @@ class SignupView(BaseSignupView):
         form = super(SignupView, self).get_form(form_class)
         form.fields["email"].label = _("Email address")
 
-        User = get_user_model()
-
-        # When no username is provided, default to the local-part of the email address
-        if form.initial.get("username", "") == "":
-            email = form.initial.get("email", "")
-            if isinstance(email, tuple):
-                email = email[0]
-            suggested_username = suggested_username_base = email.split("@")[0]
-            increment = 1
-            while User.objects.filter(username__iexact=suggested_username).exists():
-                increment += 1
-                suggested_username = f"{suggested_username_base}{increment}"
-            form.initial["username"] = suggested_username
-
-        self.matching_user = None
-        initial_username = form.initial.get("username", None)
-
-        # For GitHub/Google users, see if we can find matching user by username
+        # We should only see GitHub/Google users.
         assert self.sociallogin.account.provider in ("github", "google")
-        try:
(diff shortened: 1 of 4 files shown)


kuma/users/views.py
```diff
@@ -599,31 +599,31 @@ def get_form(self, form_class=None):
...
-        # When no username is provided, default to the local-part of the email address
-        if form.initial.get("username", "") == "":
-            email = form.initial.get("email", "")
-            if isinstance(email, tuple):
-                email = email[0]
-            suggested_username = suggested_username_base = email.split("@")[0]
+        # We should only see GitHub/Google users.
+        assert self.sociallogin.account.provider in ("github", "google")
+
+        initial_username = form.initial.get("username")
```
peterbe: Down on line 624 there's that strange...
```python
if initial_username:
    ...
else:
    initial_username = ""
```
I think a more idiomatic syntax would be to use 
```suggestion
        initial_username = form.initial.get("username") or ""
```

kuma/users/views.py
```diff
@@ -599,31 +599,31 @@ def get_form(self, form_class=None):
...
+                initial_username = f"{initial_username_base}{increment}"
+        else:
+            initial_username = ""
 
-        # For GitHub/Google users, see if we can find matching user by username
-        assert self.sociallogin.account.provider in ("github", "google")
-        try:
-            self.matching_user = User.objects.get(username=initial_username)
-            # deleting the initial username because we found a matching user
-            del form.initial["username"]
```
peterbe: With this, you can never get a KeyError on `form.initial['username']` any more. 👍🏼

peterbe:
Start with this :) https://github.com/escattone/kuma/pull/4

kuma/users/views.py
escattone: Yep, that's better @peterbe, thanks. I think instead but almost the same I'll use:
```python
initial_username = form.initial.get("username", "")
```

peterbe: But then `intial_username` might become `None` something else falsy. 
But storm in a teacup, I don't think it's a risk in the current context.

escattone: Ah, yes, I suppose the actual value of the `username` key could be `None` so yours is safer. Sorry, I'll adjust that in my clean-up PR.

PR #6645, opened 2020-03-05
enable multi-auth

escattone:
Enables multi-auth (the addition of Google login) for everything (stage, production, etc.).

@peterbe I went down the road of exploring doing the clean-up (resolving all code that depends-on `settings.MULTI_AUTH_ENABLED`) in this PR as well (just to get it over with), but decided it was best to do that in a later PR as [you originally suggested](https://github.com/mdn/infra/pull/378#issuecomment-595394057). 😄


kuma/settings/common.py
```diff
@@ -1694,10 +1694,10 @@ def get_user_url(user):
 
 # Enable or disable the multi auth(Google and Github) sign-in flow
 # When disabled, Github will be the default and only Auth provider
```
peterbe: This code comment is not making sense.

escattone: Oops, good catch.

PR #6650, opened 2020-03-06
Multi auth cleanup

escattone:
Clean-up after enabling multi-auth (Google authentication) in production. See https://github.com/mdn/kuma/pull/6645 and https://github.com/mdn/infra/pull/378#issuecomment-595394057.

### DIFF
--- a/kuma/users/tests/test_views.py
+++ b/kuma/users/tests/test_views.py
@@ -1036,16 +1036,17 @@ def test_user_edit_github_is_public(wiki_user, wiki_user_github_account, user_cl
 
 @pytest.mark.parametrize("case", ("DOMAIN", "WIKI_HOST"))
 def test_404_logins(db, client, case):
-    """The login buttons should display on the 404 page."""
+    """A login link should display within the body on the wiki 404 page."""
     response = client.get(
         "/something-doesnt-exist", follow=True, HTTP_HOST=getattr(settings, case)
     )
     assert response.status_code == 404
-    providers_shown = pq(response.content).find(".socialaccount-providers")
+    signin_url = reverse("socialaccount_signin")
+    signin_shown = pq(response.content).find(f'#content a[href^="{signin_url}"]')
     if case == "WIKI_HOST":
-        assert providers_shown
+        assert signin_shown
     else:
-        assert not providers_shown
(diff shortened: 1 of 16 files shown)


kuma/users/tests/test_views.py
```diff
@@ -1036,16 +1036,17 @@ def test_user_edit_github_is_public(wiki_user, wiki_user_github_account, user_cl
 
 @pytest.mark.parametrize("case", ("DOMAIN", "WIKI_HOST"))
 def test_404_logins(db, client, case):
```
escattone: Not sure this test was worth the effort of refactoring, but there it is.

kuma/users/views.py
```diff
@@ -600,7 +597,7 @@ def get_form(self, form_class=None):
         # We should only see GitHub/Google users.
         assert self.sociallogin.account.provider in ("github", "google")
 
-        initial_username = form.initial.get("username", "")
+        initial_username = form.initial.get("username") or ""
```
peterbe: hehe. a sneak-in aftereffect :)

peterbe:
I think you need to run `jest --updateSnapshot` to update the snapshots.

By the way, have I recently reflected how much I detest jest snapshots? I think we should delete them entirely and I'm going to advocate we never enable them on any new MDN projects. 
CC @schalkneethling 
They never help us avoid mistakes because every time we get snapshot errors we just run `jest --updateSnapshot` and don't reflect. They become a test of the tests which just causes distractions.

escattone:
> I think you need to run `jest --updateSnapshot` to update the snapshots.
> 
> By the way, have I recently reflected how much I detest jest snapshots? I think we should delete them entirely and I'm going to advocate we never enable them on any new MDN projects.
> CC @schalkneethling
> They never help us avoid mistakes because every time we get snapshot errors we just run `jest --updateSnapshot` and don't reflect. They become a test of the tests which just causes distractions.

@peterbe I hear you, but I can honestly say that in the few times I've had to adjust the snapshots, I've never run `jest --updateSnapshot` but have always updated the snapshots manually. I guess it depends on how us developers use/abuse the tool. As you said, if we're not reflective, then they're ultimately not helpful.

PR #6652, opened 2020-03-06
Use two spaces for yaml, reindent .travis.yml

callahad:
This matches the most common practice, including what we do in mdn/infra.


callahad:
This is ridiculously trivial, but it's been bugging me all week

PR #6658, opened 2020-03-08
Extract duplicate switch statement into component

erkde:
This commit just extracts the duplicate switch statement in index.jsx and ssr.jsx into it’s own thing, an App component.

As the default handling for an unknown prop `componentName` is different in index.jsx (throws an error) than in ssr.jsx (writes to console), the App component just returns null to indicate this condition. 

App also takes an optional second prop `data`, in index.jsx it comes from `window._react_data`, I can’t find a type for it anywhere, so have left it as `any` for now.

### DIFF
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -23,13 +23,8 @@ if (container) {
     // Store the string catalog so that l10n.gettext() can do translations
     localize(data.locale, data.stringCatalog, data.pluralFunction);
 
+    // Will throw an error for an unknown componentName.
     let app = <App componentName={componentName} data={data} />;
-
-    if (app === null) {
-        throw new Error(
-            `Cannot render or hydrate unknown component: ${componentName}`
-        );
-    }
 
     /* StrictMode is a tool for highlighting potential problems
        in an application. Like Fragment, StrictMode does not
--- a/kuma/javascript/src/index.jsx
+++ b/kuma/javascript/src/index.jsx
@@ -2,11 +2,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-import SinglePageApp from './single-page-app.jsx';
-import LandingPage from './landing-page.jsx';
-import SignupFlow from './signup-flow.jsx';
-import UserAccount from './user-account/user-account.jsx';
-
+import App from './app.jsx';
 import { AppErrorBoundary } from './error-boundaries.jsx';
 import GAProvider from './ga-provider.jsx';
 import { localize } from './l10n.js';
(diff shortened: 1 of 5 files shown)


peterbe:
@Gregoor Can you take a look at this. I know you mentioned something similar and perhaps you can better visualize what's going on with on-going open pull requests that would conflict with this.

kuma/javascript/src/index.jsx
```diff
@@ -25,49 +23,20 @@ if (container) {
...
-            // to handle it as a special case here.
-            app = <SignupFlow />;
-            break;
-        default:
-            throw new Error(
-                `Cannot render or hydrate unknown component: ${componentName}`
-            );
+    let app = <App componentName={componentName} data={data} />;
+
+    if (app === null) {
```
Gregoor: `React.createElement` for which JSX is syntactical sugar always returns an object, describing the component to be rendered (and its props). What will be null is the renderer output. E.g.:
Screenshot (2)

One option would be to let the component itself throw and then add an error boundary for the SSR component. It'd also be great if we had tests that made sure it's correctly catching and logging the error in that case then.

erkde: @Gregoor thanks for spotting this, and the comments. 

I've just added a second commit c878df1ec so it throws, along with a test to check ssr.jsx writes the caught error to console.

With adding an error boundary to the SSR component, I notice in index.jsx - the throw happens outside of the AppErrorBoundary, as it did before with the switch statement, I guess this is intentional, or should it actually be inside?

erkde:
Note - has additional commit(s) based on review, may need a squash/edit.

kuma/javascript/src/index.jsx
erkde: Hmm, I think my comment above is partially incorrect, it seems in index.jsx, the error is not thrown when the App component is assigned, e.g.

```js
let app = <App componentName={componentName} data={data} />;
```

or when it is re-assigned to within the AppErrorBoundary, e.g.

```js
let app = (
  <GAProvider>
    <AppErrorBoundary>
      <React.StrictMode>{app}</React.StrictMode>
    </AppErrorBoundary>
  </GAProvider>
);
```

but when app is passed to render, e.g.
 
```js
ReactDOM.render(app, container);
```

So, I guess this PR's moves the error from being thrown outside of the AppErrorBoundary, to inside.

I'm not sure I understand why you would want to throw the error outside of it, but this does seem to change that.

erkde: One more thing, using error boundaries in ssr.jsx, doesn't seem to be currently supported with `renderToString`: https://github.com/facebook/react/issues/10442

Gregoor:
Sorry for taking so long to get back to you and thanks for making the changes and adding tests!

I would've had one more nit, which is the fact that we're checking for the error message, instead of introducing a new error type. [Then I learned that babel doesn't properly subclass](https://github.com/babel/babel/issues/5524) and became sad. But now we have tests, so at least a mindless change to the error message blows CI up. So I guess we're good on that front :)

Regarding your notes:
Fortunately we don't rely on error boundaries in SSR atm, so as long as `renderToString` throws (which it does), we're good. And with `index.jsx` (aka the client bootstrap) the error throwing happens within the boundary (as we noticed before on-render, not on-create-element), so we can still control what users see as well as report the error to our analytics.

Thanks again for the patience with me and with Kuma (I just needed some time to figure out how to re-run the SSR piece myself, so double kudos for you).

Gregoor:
Oh an if you want to look at another consolidation task (though more Python-heavy), here's one: https://github.com/mdn/kuma/issues/6678

erkde:
Cool, thanks @Gregoor - happy to take a squiz at #6678

Issue #6581, opened 2020-02-25
T - Convert Subscription 'Thank You' Page to React


tobinmori:
## 💬 Description
Update design, content, and functionality for Subscription "Thank You" page and convert to React. Current version is here: https://wiki.developer.mozilla.org/en-US/payments/recurring

## 👷‍♀ Tasks
- [x] Add route and view in Django for Thank You page 
- [x] Create React component for Thank You page
- [x] Create feedback form which, upon submission, records feedback by sending a GA event 

## 🎨 Design
https://mozilla.invisionapp.com/share/TDVED747SB9#/screens/399165369

schalkneethling:
Hey @mindy,

The information here will apply to both the work here as well as the work on the FAQ so, this will be a central location for reference. Hope you find this useful.

As @escattone mentioned on Slack, there is a utility[1] that determines whether we are on the wiki or read-only site, and based on that, either load the wiki or React based document[2] for the documentation pages.

For quick reference here are the two places in the codebase referenced to above:

[1] https://github.com/mdn/kuma/blob/e9fa3ec4c70a8cbcf31e62706dc05cf6aaa1ddfa/kuma/core/utils.py#L45
[2] https://github.com/mdn/kuma/blob/e9fa3ec4c70a8cbcf31e62706dc05cf6aaa1ddfa/kuma/wiki/views/document.py#L660

For the "thank you" and "FAQ" pages everything will be React and so, these pages will be what is known as a `SPA(SinglePageApp)` component. https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/javascript/src/index.jsx#L32 and here https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/javascript/src/ssr.jsx#L45

Before we get to that however we need to get some things set-up in Django. Seeing that we do not currently have a dedicated "Thank you" page, you would need to add a new `url` and `view`

The `url` would be defined in https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/payments/urls.py#L6

```
url(r"^recurring/thank-you?$", views.thank_you, name="thank_you"),
```

The `view` will go in https://github.com/mdn/kuma/blob/074250459d0425c96158bec6e2e37c59f2a7efb0/kuma/payments/views.py

```
@skip_if_disabled
@ensure_wiki_domain
@never_cache
def thank_you(request):
[...]

PR #6660, opened 2020-03-09
Create subscription page (copy of #6647 but with a clean commit history)

### LINKED ISSUE #6582: T - Create landing page (FAQ) for Subscriptions
## 💬 Description
Create a landing page for Subscriptions in React, so we can reuse components. The page will live at /payments. 

## ✅ Acceptance Criteria
- [ ] User can navigate to /payments page
- [ ] User can read FAQ about why we are doing this and what to expect 
- [ ] User viewing on a smaller screen does not have a terrible experience

_Note: The "Help grow MDN" callout in the header is **not** expected to be completed during this sprint_
 
## 👷‍♀ Tasks
- [x] Add a new route and view in Django for /payments
- [x] Create React component, borrowing content from existing [Thank You page](https://wiki.developer.mozilla.org/en-US/payments/recurring)
- [x] Update copy to match what's in [latest copy edits](https://docs.google.com/document/d/1TH-P-hcZY1UzBtXDGGZnF3jTCy7b-cgC3TUo4Vwxfdc/edit?usp=sharing )

## 🎨 Design
https://app.abstract.com/projects/9decc480-70fc-11e9-9453-4f3fbffea7ce/branches/master/commits/5eb3004bb0e0aeb362aae44b3da2fa7d15d893cb/files/96E76C7B-2AF8-4473-A39E-E81890F91955/layers/0A8C3971-7D07-40BF-96B5-3B467D450CC4?mode=build&selected=734966551-1AF136E1-5788-4D35-A3CF-0E452D009A7B

~~https://mozilla.invisionapp.com/share/TDVED747SB9#/screens/399165366~~

## 🔦 Context
Please see Schalk's comment https://github.com/mdn/kuma/issues/6581#issuecomment-591376908

mindy:
**Summary:** 
Convert existing FAQ page to the Subscription landing page in React. 

**To do:** 
- [ ] Translations are not working (see screenshots below, where locale is `FR` and header is translated, but not the main content.

**URLs** 
**/payments** is changing from wiki to non-wiki
**/contribute** currently redirects to /payments, is changing from wiki to non-wiki 
**/payments/recurring** is still on the wiki domain, same old Jinja page for now. Header says "Thank you for supporting MDN."

**To test:** 
Go to /payments (http://localhost.org:8000/en-US/payments): 
Screen Shot 2020-03-05 at 5 29 40 PM

Make sure former /payments/recurring page still works by going to /payments/recurring:
Screen Shot 2020-03-05 at 5 29 47 PM
 
Closes #6582

### DIFF
--- a/jinja2/react_base.html
+++ b/jinja2/react_base.html
@@ -98,7 +98,6 @@
   <main id="content">
       {% block document_head %}{% endblock %}
       {% block content %}{% endblock %}
-      </div>
   </main>
 
   {%- block contrib_popup %}{# Inject contributions popup where needed #}{% endblock %}
--- a/jinja2/react_base.html
+++ b/jinja2/react_base.html
@@ -104,33 +104,35 @@
 
   {%- block contrib_popup %}{# Inject contributions popup where needed #}{% endblock %}
 
-  <!-- Footer -->
-  <footer id="nav-footer" class="nav-footer">
-    <div class="center">
-        {% include "includes/footer/mdn-footer.html" %}
+  {% block footer %}
+    <!-- Footer -->
+    <footer id="nav-footer" class="nav-footer">
+      <div class="center">
+          {% include "includes/footer/mdn-footer.html" %}
 
-        {% include "includes/footer/moz-footer.html" %}
+          {% include "includes/footer/moz-footer.html" %}
 
-        {% block lang_switcher %}
-          {% include "includes/lang_switcher.html" %}
-        {% endblock %}
+          {% block lang_switcher %}
+            {% include "includes/lang_switcher.html" %}
+          {% endblock %}
 
-        {% block footer_copyright %}
-          <ul class="footer-tos">
-            <li><a href="https://www.mozilla.org/about/legal/terms/mozilla">{{ _('Terms') }}</a></li>
-            <li><a href="https://www.mozilla.org/privacy/websites/">{{ _('Privacy') }}</a></li>
-            <li><a href="https://www.mozilla.org/privacy/websites/#cookies">{{ _('Cookies') }}</a></li>
-          </ul>
+          {% block footer_copyright %}
+            <ul class="footer-tos">
+              <li><a href="https://www.mozilla.org/about/legal/terms/mozilla">{{ _('Terms') }}</a></li>
+              <li><a href="https://www.mozilla.org/privacy/websites/">{{ _('Privacy') }}</a></li>
+              <li><a href="https://www.mozilla.org/privacy/websites/#cookies">{{ _('Cookies') }}</a></li>
+            </ul>
 
-          <div id="license" class="contentinfo">
-            {% trans copyright_url='/en-US/docs/MDN/About#Copyrights_and_licenses', thisyear=thisyear() %}
-              <p>&copy; 2005-{{ thisyear }} Mozilla and individual contributors.</p>
-              <p>Content is available under <a href="{{ copyright_url }}">these licenses</a>.</p>
-            {% endtrans %}
-          </div>
-        {% endblock %}
-    </div>
-  </footer>
+            <div id="license" class="contentinfo">
+              {% trans copyright_url='/en-US/docs/MDN/About#Copyrights_and_licenses', thisyear=thisyear() %}
+                <p>&copy; 2005-{{ thisyear }} Mozilla and individual contributors.</p>
+                <p>Content is available under <a href="{{ copyright_url }}">these licenses</a>.</p>
+              {% endtrans %}
+            </div>
+          {% endblock %}
+      </div>
+    </footer>
+  {% endblock %}
 
   {% block auth_modal %}
     {% include "includes/auth-modal.html" %}
--- a/jinja2/react_base.html
+++ b/jinja2/react_base.html
@@ -97,7 +97,6 @@
   <!-- Content will go here -->
   <main id="content">
       {% block document_head %}{% endblock %}
-      <div class="center clear">
       {% block content %}{% endblock %}
       </div>
   </main>
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { getLocale, gettext, Interpolated } from '../l10n.js';
 import A11yNav from '../a11y/a11y-nav.jsx';
 import Header from '../header/header.jsx';
+import Footer from '../footer.jsx';
 import Route from '../route.js';
 import SubHeader from './subheader.jsx';
 import ListItem from './list-item.jsx';
@@ -318,6 +319,7 @@ export default function PaymentsLandingPage() {
                     </ol>
                 </section>
             </main>
+            <Footer />
         </>
     );
 }
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -345,11 +345,10 @@ export class PaymentsRoute extends Route<PaymentsRouteParams, null> {
     }
 
     match(url: string): ?PaymentsRouteParams {
-        const path = new URL(url, BASEURL).pathname;
+        const currentPath = new URL(url, BASEURL).pathname;
         const paymentsPath = `/${this.locale}/payments`;
-        const regex = new RegExp(paymentsPath, 'g');
 
-        if (regex.test(path)) {
+        if (currentPath.startsWith(paymentsPath)) {
             return {
                 locale: this.locale
             };
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -4,7 +4,6 @@ import * as React from 'react';
 import { getLocale, gettext, Interpolated } from '../l10n.js';
 import A11yNav from '../a11y/a11y-nav.jsx';
 import Header from '../header/header.jsx';
-import Footer from '../footer.jsx';
 import Route from '../route.js';
 import SubHeader from './subheader.jsx';
 import ListItem from './list-item.jsx';
@@ -319,7 +318,6 @@ export default function PaymentsLandingPage() {
                     </ol>
                 </section>
             </main>
-            <Footer />
         </>
     );
 }
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -1,0 +1,367 @@
+// @flow
+import * as React from 'react';
+
+import { getLocale, gettext, Interpolated } from '../l10n.js';
+import A11yNav from '../a11y/a11y-nav.jsx';
+import Header from '../header/header.jsx';
+import Footer from '../footer.jsx';
+import UserProvider from '../user-provider.jsx';
+import Route from '../route.js';
+import SubHeader from './subheader.jsx';
+import ListItem from './list-item.jsx';
+
+type PaymentsRouteParams = {
+    locale: string
+};
+
+export default function PaymentsLandingPage() {
+    const locale = getLocale();
+    const urls = {
+        annualReport:
+            'https://www.mozilla.org/en-US/foundation/annualreport/2019/',
+        email:
+            'mailto:mdn-support@mozilla.com?Subject=Manage%20monthly%20subscription',
+        moco: 'https://www.mozilla.org/foundation/moco/',
+        mozillaFoundation: 'https://www.mozilla.org/foundation/',
+        managePayments: `/${locale}/payments/recurring/management`,
+        stripe: 'https://stripe.com/',
+        taxDeductible: 'https://donate.mozilla.org/faq#item_tax_a',
+        terms: `/${locale}/payments/terms`
+    };
+
+    return (
+        <UserProvider>
+            <A11yNav />
+            <Header />
+            <SubHeader
+                title="Become a monthly supporter"
+                description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
+                columnWidth="7"
+            />
+            <main
+                id="contributions-page"
+                className="contributions-page"
+                role="main"
+            >
+                <section className="section">
+                    <header>
+                        <h2>{gettext('FAQs')}</h2>
+                    </header>
+                    <ol id="contribute-faqs" className="faqs clear">
+                        <ListItem
+                            title={gettext('Why is MDN asking me for money?')}
+                            number="1"
+                        >
+                            <p>
+                                {gettext(
+                                    'MDN is seeking direct support from our users. We’d like to accelerate our growth and extend the maintenance of MDN’s content and platform, with assistance from those who use it.'
+                                )}
+                            </p>
+                            <p>
+                                {gettext(
+                                    'Our user base has grown exponentially in the last few years and we have a large list of improvements we’d like to make. While MDN is currently wholly funded by Mozilla, and has been from the beginning, we’d like to create a closer, more collaborative relationship between our audience (that’s you!), our content (written for you and sometimes by you), and our supporters (also, you, again)-- to accelerate those improvements.'
+                                )}
+                            </p>
+                        </ListItem>
+                        <ListItem
+                            title="How is my payment handled? Is it secure?"
+                            number="2"
+                        >
+                            <p>
+                                <Interpolated
+                                    id={gettext(
+                                        'All payment information goes through payment processor <stripeLink />, and a record of your payment will be stored by Mozilla. Mozilla does not receive or store your credit card number.'
+                                    )}
+                                    stripeLink={
+                                        <a
+                                            href={urls.stripe}
+                                            target="_blank"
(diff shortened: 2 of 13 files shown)


schalkneethling:
@mindy Thank you for the PR!

There is one Python test that is currently failing:

```
=================================== FAILURES ===================================
___________________________ test_redirect[payments] ____________________________
  File "/app/kuma/payments/tests/test_views.py", line 181, in test_redirect
    assert_redirect_to_wiki(response, url)
  File "/app/kuma/core/tests/__init__.py", line 14, in assert_redirect_to_wiki
    assert response.status_code == 301
AssertionError
----------------------------- Captured stdout call -----------------------------
<class 'requests.exceptions.ConnectionError'> error contacting SSR server.
Falling back to client side rendering.
================== 1 failed, 1967 passed in 193.13s (0:03:13) ==================
make: *** [Makefile:36: coveragetest] Error 1
```

You can run these by following the steps here:
https://kuma.readthedocs.io/en/latest/tests.html#running-the-test-suite

NOTE: The first step is to do `make bash` and then run the rest of the commands from the resulting bash. To avoid having to run all of the tests, you can do the following for the above error:

```
py.test kuma/payments/
```

mindy:
@schalkneethling Thank you so much for your help! I fixed the test by removing "payments" from that array, since /payments will be not on the wiki so I don't think the test applies to that page any longer.

kuma/urls.py
```diff
@@ -105,6 +102,9 @@
     ),
 )
 urlpatterns += i18n_patterns(re_path(r"^payments/", include(payments_lang_urlpatterns)))
+urlpatterns += i18n_patterns(
+    re_path(r"payments$", payment_views.index, name="payments")
+)
```
peterbe: I think we need to think deeper about this. It's nice that we're being consistent and not using a trailing slash if other pages don't use that. However, `/payments/` is actually a namespace. Look at the line above (line 104). 
So what you have name is a clash. 
Basically, I think the `payments_view.index` needs to be moved into `payments/urls.py`.

peterbe:
I don't like that the document title is "Contribute ..." when the page isn't about contributing. It's about frequently asked questions around *subscriptions*.
Screen Shot 2020-03-10 at 11 36 51 AM

peterbe:
I think this PR needs to also clean up the old payments FAQ. 
If this PR merges we'll have the same information written twice. Once in a jinja template and once in a React .jsx file.

peterbe:
There's something strange going on and it's bizarre that nothing visually is happening. 
This is what I get when I visit http://localhost.org:8000/fr/payments?a=d
Screen Shot 2020-03-10 at 11 44 12 AM

peterbe:
Now I know what's going on. That page is actually never server-rendered. This is a nasty trick that Kuma's React does and I've never understood or liked it. 
If you curl it this is what happens:
```
▶ curl -I http://localhost.org:8000/fr/payments
HTTP/1.1 404 Not Found
Date: Tue, 10 Mar 2020 15:47:51 GMT
Server: WSGIServer/0.2 CPython/3.8.0
Content-Type: text/html
X-Frame-Options: DENY
Content-Length: 1709
Content-Language: fr
Vary: Cookie
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
```

As if the page doesn't exist. Fair. It means we have a problem but you don't see a loud "Sorry! Page Does Not Exist". Instead, what it does it that it falls back on doing a client-side render. 
Let's not do that. SSR render these pages. (which very much might explain the confusion about never-localized strings)

Also, how come no test noticed this? I think that's the first thing to look into. Opening `client.get('/en-US/payments')` in a test should be 200 OK and the content of it should have stuff about the FAQ.

mindy:
> I don't like that the document title is "Contribute ..." when the page isn't about contributing. It's about frequently asked questions around _subscriptions_.

Changed it to "Become a monthly supporter". Also eventually this page will have a CTA in the header to contribute: 
contribute

mindy:
> I think this PR needs to also clean up the old payments FAQ.
> If this PR merges we'll have the same information written twice. Once in a jinja template and once in a React .jsx file.

I see that, but I think it's fine to leave it for now-- in the next sprint I'm going to convert the existing Jinja page to something else: https://github.com/mdn/kuma/issues/6581

mindy:
@peterbe I need some guidance on how to do this-- do you have an example of an SSR page I can look at for an example?

peterbe:
> > I think this PR needs to also clean up the old payments FAQ.
> > If this PR merges we'll have the same information written twice. Once in a jinja template and once in a React .jsx file.
> 
> I see that, but I think it's fine to leave it for now-- in the next sprint I'm going to convert the existing Jinja page to something else: #6581

In that case, I'd rather you put a note in the old stuff to say that it's going to be deleted.

peterbe:
> @peterbe I need some guidance on how to do this-- do you have an example of an SSR page I can look at for an example?

The site-search thing. Do what it does. But you don't have a `result` so it might need some work. I'm going to attempt to do it too, in parallel, and we can compare results.

mindy:
> > > I think this PR needs to also clean up the old payments FAQ.
> > > If this PR merges we'll have the same information written twice. Once in a jinja template and once in a React .jsx file.
> > 
> > 
> > I see that, but I think it's fine to leave it for now-- in the next sprint I'm going to convert the existing Jinja page to something else: #6581
> 
> In that case, I'd rather you put a note in the old stuff to say that it's going to be deleted.

Added a comment at the top of payments.html

peterbe:
Another thing, before I forget, I don't like that the "payment terms" page is on the Wiki only. I can appreciate that you have to go to the Wiki to do your Stripe thing, for now, but I think the terms and the faq ought to be on the same site.

jinja2/react_base.html
```diff
@@ -104,33 +104,35 @@
...
+            <div id="license" class="contentinfo">
+              {% trans copyright_url='/en-US/docs/MDN/About#Copyrights_and_licenses', thisyear=thisyear() %}
+                <p>&copy; 2005-{{ thisyear }} Mozilla and individual contributors.</p>
+                <p>Content is available under <a href="{{ copyright_url }}">these licenses</a>.</p>
+              {% endtrans %}
+            </div>
+          {% endblock %}
+      </div>
+    </footer>
+  {% endblock %}
```
peterbe: I still cannot understand why this change is necessary? 
As far as I can tell, the block of code was put in a `{% block footer %} ... {% endblock %}` But is there any reason to do it?

kuma/javascript/src/payments/index.jsx
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    getComponent() {
+        return PaymentsLandingPage;
+    }
+
+    match(url: string): ?PaymentsRouteParams {
+        let path = new URL(url, BASEURL).pathname;
+
+        if (path !== `/${this.locale}/payments`) {
```
peterbe: We just need to be very careful that this works when we figure out the correct URL trailing slash thing in `urls.py`.

kuma/javascript/src/payments/subheader.jsx
```diff
@@ -0,0 +1,35 @@
...
+import * as React from 'react';
+
+import { gettext } from '../l10n.js';
+
+type Props = {
+    title: string,
+    subtitle?: string,
+    description?: string,
+    children?: React.Node,
+    columnWidth?: string
```
peterbe: Is it really optional?

kuma/payments/jinja2/payments/index.html
```diff
@@ -1,9 +1,14 @@
 {% extends "react_base.html" %}
 
-{% block title %}{{ _('Become a monthly supporter') }}{% endblock %}
+{% block title %}{{ page_title(_('Become a monthly supporter')) }}{% endblock %}
 
 {% block document_head %}
-    {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(), None)|safe }}
+    <!--
```
peterbe: This needs to be `{#` and `#}` otherwise it becomes a HTML comment.

kuma/payments/jinja2/payments/index.html
```diff
@@ -1,9 +1,14 @@
...
+        This page doesn't need to fetch data, so we pass `True`
+        as an argument to indicate that we have the data we need, which
+        allows the component to load properly.
```
peterbe:
```suggestion
        allows the component render in the server. 
```

kuma/urls.py
mindy: Will work on this in the next few minutes 👍

kuma/javascript/src/payments/subheader.jsx
mindy: Yeah, because we set a default value for `columnWidth` ("all") below

jinja2/react_base.html
mindy: If we don't have this, it renders two footers. So this allows us to override this footer in `payments/index.html`:

two-footers

mindy: I could remove the React Footer and leave the one here, but when do we know which one to use? The React Footer is only used in documents right now.

kuma/payments/jinja2/payments/index.html
mindy: Fixed. Thank you!

kuma/urls.py
mindy: This is now fixed-- the index page is now being handled in `payments/urls.py`. Thanks for the suggestion!

schalkneethling:
Hey @mindy,

Not to throw a spanner in the works but what was the reason for not using the React-based footer here?

The build is failing because of this:

```
kuma/urls.py:17:1: F401 'kuma.payments.views as payment_views' imported but unused
```

You can safely remove that `import`.

Also as a side note. If any/all of the "instructions" I provided was incorrect or contributed to these problems, let's update it so we have a nice how-to for when we need to do this next.

Thanks so much @mindy! 🍵

mindy:
Thanks @schalkneethling ! I removed the React footer because I wasn't sure which one to use (https://github.com/mdn/kuma/pull/6660#discussion_r390573927) and the React footer left a gap at the bottom of the page (which is fixed now): 

Screen Shot 2020-03-11 at 11 24 35 AM

I added the React footer back in, and moved the `<div class="center clear"></div>` from `react_base.html` to the specific components that use `block content` (it was only the homepage). 

Great idea-- I'll add a page to the Kuma wiki with your instructions and some edits.

jinja2/react_base.html
peterbe: I give up on all this footer stuff. I think you an @schalkneethling should push back on the influx of feature additions and convince management that you have to refactor this mess up. 
As mentioned in Slack, we have **4** `<html ...` tags, when it should be 2 (one for the Wiki and one for the read-only site). 
```
▶ rg '<html ' | rg -v jquery | rg -v ckeditor | rg -v tests | rg -v '\.ltxt'
jinja2/base.html:<html lang="{% block lang %}{{ LANG }}{% endblock %}" dir="{{ DIR }}" class="no-js">
jinja2/react_base.html:<html lang="{% block lang %}{{ LANG }}{% endblock %}" dir="{{ DIR }}" class="no-js">
kuma/wiki/jinja2/wiki/react_document.html:<html lang="{{ document_data['hrefLang'] }}" dir="{{ DIR }}">
kuma/users/jinja2/socialaccount/base_signup.html:<html lang="{% block lang %}{{ LANG }}{% endblock %}" dir="{{ DIR }}" class="no-js">
```
And **4** `<footer` tags when there should be 2. 
```
▶ rg '<footer' | rg -v jquery | rg -v ckeditor | rg -v tests | rg -v '\.ltxt' | rg -v l10n.js
jinja2/base.html:    <footer id="nav-footer" class="nav-footer">
jinja2/react_base.html:    <footer id="nav-footer" class="nav-footer">
kuma/javascript/src/footer.jsx:        <footer id="nav-footer" className="nav-footer">
kuma/users/jinja2/socialaccount/base_signup.html:    <footer id="nav-footer" class="nav-footer">
```

kuma/javascript/src/payments/subheader.jsx
peterbe: bah! I meant the line above. `children` Sorry. Really not sure it matters at all.

kuma/payments/jinja2/payments/payments.html
```diff
@@ -1,3 +1,5 @@
+<!-- This page will be replaced in the next sprint by https://github.com/mdn/kuma/issues/6581 -->
```
peterbe:
```suggestion
{# This page will be replaced in the next sprint by https://github.com/mdn/kuma/issues/6581 #}
```

kuma/javascript/src/payments/index.jsx
```diff
@@ -0,0 +1,363 @@
...
+        if (regex.test(path)) {
+            return {
+                locale: this.locale
+            };
+        }
+        return null;
+    }
+
+    fetch() {
+        return Promise.resolve(null);
```
peterbe: I think this should never run because we never need to do client-side fetch. So you can replace this with 
```suggestion
        throw new Error('Payments should never need to post-fetch more data');
```

kuma/javascript/src/payments/index.jsx
```diff
@@ -0,0 +1,363 @@
...
+    getComponent() {
+        return PaymentsLandingPage;
+    }
+
+    match(url: string): ?PaymentsRouteParams {
+        const path = new URL(url, BASEURL).pathname;
+        const paymentsPath = `/${this.locale}/payments`;
+        const regex = new RegExp(paymentsPath, 'g');
+
+        if (regex.test(path)) {
```
peterbe: Pardon my ignorance but why is this necessary? Why can't it be `if (path.startsWith(`/${this.locale}/payments`) {`?

peterbe:
It might seem silly but it's hugely beneficial for possible future refactoring if you at least render the view. 
https://github.com/mindy/kuma/pull/1

jinja2/react_base.html
schalkneethling: Ah ok, I am on the same page here now. The change was indeed simple to wrap the `footer` HTML inside a `footer` Jinja block. This is required because `payments/index.html` extends `react_base.html`. If the `footer` is not wrapped in a `block` there is no way for us to override the `block` inside `payments/index.html`

The result would be two footers. One from React and one from the extended `react_base.html`. The reason `search/react.html` did not run into this scenario first is that it does not use the React footer but the one inherited from `react_base.html`

This means that as long as we have pages that need to extend `react_base.html` the following will always be the template used by said page:

```
{% extends "react_base.html" %}

{% block title %}{{ page_title(_('Become a monthly supporter')) }}{% endblock %}

{% block document_head %}
   {#
        This page doesn't need to fetch data, so we pass `True`
        as an argument to indicate that we have the data we need, which
        allows the component to render in the server.
    #}
    {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(), True)|safe }}
{% endblock %}

{% block content %}{% endblock %}

{% block footer %}{% endblock %}
```

SPA is probably not the best naming scheme anymore inside Kuma anyway. RPC(React Page Component) might be more appropriate :)

schalkneethling: With that said, if all pages that extend `react_base.html` do not need the footer any longer, we can remove the footer from `react_base.html` and in so doing, the need for the Jinja `footer block`.

Currently, though, three pages extend it:

1. `landing/react_homepage.html`
2. `payments/index.html`
3. `search/react.html`

Numbers 2 and 3 do not need the footer but, until the homepage is converted to React, it will still need it.

kuma/static/styles/components/payments/_page.scss
```diff
@@ -117,6 +119,10 @@
                 padding-left: 1em;
```
schalkneethling: I would suggest that we add `margin-top: 20px;` here to create a bit of space between the paragraph and the list.

kuma/javascript/src/payments/subheader.jsx
mindy: I see-- when I remove `children`, it complains that that it wasn't in the Props declaration, so we need it here.

mindy: It does-- I got a complaint when I removed the `?` saying that children was required in `<SubHeader />` but not provided

kuma/static/styles/components/payments/_page.scss
mindy: Changed the `p:last-of-type` rule to `p:last-child` rule instead, so we still have the margin between list and paragraph. Thank you for pointing this out!

kuma/javascript/src/payments/index.jsx
mindy: Good point-- changed to `startsWith` instead. Thanks Peter!

mindy:
Ready for another 👀 when you get a chance, @schalkneethling

jinja2/react_base.html
mindy: Opened up https://github.com/mdn/kuma/issues/6678 so we don't loose track of this.

PR #6674, opened 2020-03-12
Add instructions for adding a React page

mindy:
Based on @schalkneethling 's very helpful instructions here: https://github.com/mdn/kuma/issues/6581#issuecomment-591376908


docs/development.rst
```diff
@@ -133,6 +133,137 @@ To get an idea about the size distribution of our react codebase you can run::
 
 This will open an interactively explorable report in your browser.
 
+Add a React page
+----------------
+
+First we need to set up a few things in Django:
+
+#. Create a new Jinja template that our view will render (we'll use ``example.html``). Paste the following code in ``example.html`` to get started::
```
schalkneethling: Perhaps be a bit more specific as to where a file such as this would be added. Say for example we are adding a new page to the user management "app", where would `example.html` go?

docs/development.rst
```diff
@@ -133,6 +133,137 @@ To get an idea about the size distribution of our react codebase you can run::
...
+    {% extends "react_base.html" %}
+
+    {% block title %}{{ page_title(_('My Page Title')) }}{% endblock %}
+
+    {% block content %}This works!{% endblock %}
+
+    {# This overrides the Jinja footer, since we will be using the React footer instead. Without this line, two footers will render. #}
+    {% block footer %}{% endblock %}
+
+#. In ``views.py``, render the Jinja template we just created::
```
schalkneethling: Continuing on from the above comment, let's be specific as to which `views.py` we are editing.

Maybe we can use the `users` app under `kuma/` in this example?

docs/development.rst
```diff
@@ -133,6 +133,137 @@ To get an idea about the size distribution of our react codebase you can run::
...
+
+    @never_cache
+    def example(request):
+        return render(request, "example.html")
+
+#. In ``kuma/urls.py``, add a new url that will render our view. This code will change slightly depending on your url and file structure::
```
schalkneethling: Here also that would be 
```suggestion
#. In ``kuma/users/urls.py``, add a new url that will render our view. This code will change slightly depending on your url and file structure::
```

docs/development.rst
```diff
@@ -133,6 +133,137 @@ To get an idea about the size distribution of our react codebase you can run::
...
+            if (regex.test(path)) {
+                return {
+                    locale: this.locale
+                };
+            }
+            return null;
+        }
+
+        fetch() {
+            return Promise.resolve(null);
```
schalkneethling: Perhaps we should add a note about the fact that this component uses static data and that is why this instance of the `fetch` function retuns `null`. For now, we can then just point to an example in the codebase of a component that does actually need initial data.

At a later stage we can then add another example of a component that does implement `fetch`.

schalkneethling:
@mindy Let me know if you have a moment to have a look over the items here. It would be great to have this as part of our docs. Thank you

docs/development.rst
mindy: Great idea-- I think being specific will help with clarity. I'm going to use `kuma/dashboards` as an example.

docs/development.rst
mindy: I'm going to use `/dashboards/`, because the folder structure is a bit simpler than `/users/` (specifically the jinja2 folder)

docs/development.rst
mindy: I ended up going down a rabbit hole trying to understand the Router component! I ended up adding examples for `fetch` and initial data. It turns out static pages do not need the `fetch()` method at all.

PR #6680, opened 2020-03-14
record subscriptions

### LINKED ISSUE #6579: T - Track successful subscription signups in whoami

### LINKED ISSUE #6684: Add subscription reminder to account cancellation message
**Summary**
The account cancellation message should remind people that their subscription will be cancelled too.

**Rationale**
Currently this is implicit, and might surprise a small number of people.

**Audience**
Subscribers who cancel their account.

**Proposal**
Currently we have two messages: 
"Deleting your account loses any preferences you have set." and "Deleting your account loses any preferences you have set, as well as makes your username available for others to use."

Please add this sentence after either one of them: "It will also cancel any MDN subscription you might have."

peterbe:
Fixes #6579
Fixes #6684

- [x] Needs test for new feature in `api/v1/whoami`
- [x] Add new Stripe webhook for cancellations
- [x] Explain, in the PR, why certain changes were made (e.g. refactoring of checks, breaking up and moving `retrieve_stripe_subscription_info`, why the cancel method is a generator)

I really don't like some of the APIs in `payments/utils.py` that were about payments. For example, when you press to cancel your monthly subscription(s), it deletes all subscriptions you have. It'd be more sensible if it would list each subscription on the `payments/recurring/monthly` and there's a cancel for each one. 
Another thing is that in payments/recurring/management there's no information about each subscription so you can't sync-check the `UserSubscriptions`. Possibly joint with the point above. 
Also, I get nervous when I see all the unit tests for `payments/utils.py`, with heavy mocking, that probably shouldn't be tests at all. Instead, there should just be things in `payments/tests/test_views.py`

## How to test

Check out the branch and `docker-compose exec web ./manage.py migrate`
In one browser tab, go to http://localhost.org:8000/admin/users/usersubscription/
In another browser tab, go to http://localhost.org:8000/api/v1/whoami (note the `is_subscriber`)
In another tab go to `http://wiki.localhost.org:8000/en-US/profiles/$username/edit` and if you had a subscription already, go back to http://localhost.org:8000/admin/users/usersubscription/ and http://localhost.org:8000/api/v1/whoami and refresh. 
If you cancel your subscription, reload the above-mentioned pages and observe what happens. 
**Note** If you cancel your subscription and set up a new one, this'll be a *different* subscription so you'll get two records like this:
Screen Shot 2020-03-16 at 11 29 51 AM


kuma/payments/tests/test_utils.py
```diff
@@ -29,7 +29,7 @@ def test_get_stripe_customer_data(mock_retrieve):
 @mock.patch("stripe.Subscription.retrieve", return_value=mock.Mock(spec_set=["delete"]))
 def test_cancel_stripe_customer_subscription(mock_sub, mock_cust):
     """A stripe customer's subscriptions can be cancelled."""
-    cancel_stripe_customer_subscription("cust_id")
+    list(cancel_stripe_customer_subscription("cust_id"))
```
peterbe: This is rather ugly but there's a good reason for it. Lemme explain. 

First of, look at `payments/views.py` around line 51. If you don't do that pattern you don't get information about which subscriptions get canceled. It just cancels everything based on a customer :(
So that's the primary use case. for why `cancel_stripe_customer_subscription` is now a generator. And because it's a generator, nothing happens unless you "exhaust" it. 

What I'd *ideally* like is two-fold, and I don't know what should come first. 
On the "MDN subscription management" page the cancel button is for ALL your subscription. Another way to phrase it is that it cancels the whole customer. It would be better if that UI was one cancel button per subscription. It's silly that someone would have more than 1 subscriptions. If they have a $5.00 and a $7.50 something's screwy because it ought to be just one for $12.50. 
Either way, the Django view or the template UX should be done on a per-subscription basis. So what I *could* do is rewrite `payments/views.py` so that canceling a subscription triggers a Django signal. 

However, and here's the point! I didn't want to change `payments/views.py` too much. Not right now at least. Hence this one little change only. Just to get my code unblocked.

kuma/payments/utils.py
```diff
@@ -53,4 +53,5 @@ def cancel_stripe_customer_subscription(stripe_customer_id):
     customer = stripe.Customer.retrieve(stripe_customer_id)
     for sub in customer["subscriptions"]["data"]:
         s = stripe.Subscription.retrieve(sub["id"])
+        yield s
```
peterbe: See longer comment about why I made this change. I'm still not satisfied with this. I think what it ought to do instead is to trigger Django signals. Custom ones. One for subscriptions having been created and one for subscriptions about to be canceled. I don't mind working on refactoring like that but my PR-spidy senses tells me this PR is already large and complex enough. I'll file some followup issues.

kuma/payments/views.py
```diff
@@ -48,7 +49,10 @@ def recurring_payment_management(request):
         context["cancel_request"] = True
         cancel_success = False
         try:
-            cancel_stripe_customer_subscription(request.user.stripe_customer_id)
+            for subscription in cancel_stripe_customer_subscription(
+                request.user.stripe_customer_id
+            ):
+                UserSubscription.set_canceled(request.user, subscription.id)
```
peterbe: Note my boldness of not swallowing errors. This operation is so "low-level" that my reasoning is that if the MySQL writes fail, then almost all hope is lost.

kuma/users/admin.py
```diff
@@ -50,3 +50,12 @@ def revisions(self, obj):
...
         return format_html('<a href="{}"><strong>{}</strong></a>', link, count)
+
+
+@admin.register(UserSubscription)
+class UserSubscriptionAdmin(admin.ModelAdmin):
+    readonly_fields = ("user", "updated", "created", "stripe_subscription_id")
+    list_display = ("user", "canceled", "updated", "created")
+    search_fields = ("user__username",)
+    list_filter = ("canceled", "created")
+    ordering = ("updated",)
```
peterbe: All of this is a bit rough and not thought-through, but does it matter much?

kuma/users/apps.py
```diff
@@ -1,11 +1,9 @@
 import stripe
 from django.conf import settings
 from django.contrib.auth.apps import AuthConfig
+from django.core.checks import register
 from django.utils.translation import gettext_lazy as _
 
-# the checks self-register so we don't need to use anything from the import
-import kuma.users.checks  # noqa: F401
-
```
peterbe: I know we had a debate about this when you originally implemented it. So don't think I'm using this chance to rewrite it in a style/fashion that *I* prefer. 
No, the reason it has to be like this is that if not done like this, the models aren't loaded by the way this *import* happens. the `register()` wrapper would make sure the check isn't actually executed until after the `UserConfig.ready()` is executed, but it's different about import-time stuff. I think you can work through the exactly order and logic. If you simply put in a temporary `import kuma.users.checks` here and start Django you'll see the error for yourself.

kuma/users/checks.py
```diff
@@ -31,11 +30,16 @@ def stripe_check(app_configs, **kwargs):
...
+        errors.append(
+            Error(f"unable to retrieve stripe plan: {error}", id=STRIPE_PLAN_ERROR)
+        )
 
+    try:
         create_missing_stripe_webhook()
     except stripe.error.StripeError as error:
         errors.append(
-            Error(f"unable to retrieve stripe plan: {error}", id=STRIPE_PLAN_ERROR)
+            Error(f"unable get or create strip webhooks: {error}", id=STRIPE_PLAN_ERROR)
```
peterbe: Technically, the function should be renamed to `get_or_create_missing_stripe_webhook` or something better because it's not 100% creating. But one thing at a time. At least now this error message is slightly more accurate.

kuma/users/models.py
```diff
@@ -191,3 +192,30 @@ def get_recovery_url(self):
         token = default_token_generator.make_token(self)
         link = reverse("users.recover", kwargs={"token": token, "uidb64": uidb64})
         return link
+
+
+class UserSubscription(models.Model):
+    user = models.ForeignKey(User, on_delete=models.PROTECT)
```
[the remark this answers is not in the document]
peterbe: It'll be changed in the next push. I'm working on adding some tests to the `delete_user` Django views when a user has an active subscription.

kuma/users/stripe_utils.py
```diff
@@ -29,47 +27,24 @@ def create_stripe_customer_and_subscription_for_user(user, email, stripe_token):
...
-        return {
-            "next_payment_at": datetime.fromtimestamp(
-                stripe_subscription.current_period_end
-            ),
-            "brand": card.brand,
-            "expires_at": f"{card.exp_month}/{card.exp_year}",
-            "last4": card.last4,
-            # Cards that are part of a "source" don't have a zip
-            "zip": card.get("address_zip", None),
-        }
```
peterbe: The explanation for this is that I first of all needed to break it up. I need to retrieve the subscription so I can take its `.id` and stick it into `UserSubscription`. 
Technically, it might be possible to keep this function but I just thought it'd be better to A) break it up and B) this stuff of making a little dict to is exclusive to the view function. It's not a utility function. It belongs to the view only. 
I hope you can muster my refactoring.

kuma/users/tests/test_stripe_subscription.py
```diff
@@ -13,18 +15,44 @@
...
-    }
+@dataclass
+class StripeCustomerSource:
+    object: str
+    brand: str
+    exp_month: int
+    exp_year: int
+    last4: int
+
+    def get(self, key, default=None):
```
peterbe: This right here is what I switched to dataclasses instead of the `SimpleNamespace`. The mock of the customer needs a `get` method.

kuma/users/views.py
```diff
@@ -467,14 +468,45 @@ def user_edit(request, username):
 
             return redirect(edit_user)
 
+    subscription_info = None
+    stripe_customer = get_stripe_customer(edit_user)
+    # XXX Might be nice to explain WHY we're doublechecking the user email.
```
peterbe: If you can explain to me why we're doing this double-check, I can replace that comment.

kuma/users/views.py
```diff
@@ -796,7 +828,11 @@ def create_stripe_subscription(request):
     try:
         email = request.POST.get("stripe_email", "")
         stripe_token = request.POST.get("stripe_token", "")
-        create_stripe_customer_and_subscription_for_user(user, email, stripe_token)
+        subscription = create_stripe_customer_and_subscription_for_user(
+            user, email, stripe_token
+        )
+        UserSubscription.set_active(user, subscription.id)
```
peterbe: What I'd *like* to do is to not do it like this but instead depend on signals. But I'm not sure it's a great win either way.

peterbe:
@Gregoor I know this is a beast but I tried to break it up a little.

peterbe:
177 new lines of test coverage!

kuma/users/tests/test_stripe_subscription.py
```diff
@@ -78,43 +124,89 @@ def test_create_stripe_subscription_fail(mock1, mock2, test_user):
...
-                }
-            ),
-        }
-    )
-
-
-@patch(
-    "stripe.Event.construct_from", side_effect=mock_stripe_payment_event,
-)
+@patch("stripe.Event.construct_from")
```
Gregoor: much nicer :)

kuma/users/views.py
Gregoor: I think it was you who asked for it, to make sure that people who signed up in the past subscription iteration would get the option to create a new stripe customer if their past one used a different email.
Tbh I can't quite strongly defend the decision anymore and I think it should probably be one of the things we drop once we move stripe accounts.

kuma/users/views.py
```diff
@@ -467,14 +470,54 @@ def user_edit(request, username):
 
             return redirect(edit_user)
 
+    subscription_info = None
+    stripe_customer = get_stripe_customer(edit_user)
+    # XXX Might be nice to explain WHY we're doublechecking the user email.
+    if stripe_customer and stripe_customer.email == edit_user.email:
```
Gregoor: can we extract that into a separate function again? I find user_edit to be already a bit much, readability wise.

kuma/payments/tests/test_utils.py
Gregoor: What speaks against having `cancel_stripe_customer_subscription` just return a list of IDs for the cancelled subscriptions? Generators seem a little overblown to me here.

peterbe: Because I couldn't trust the `stripe` Python API. What if this happens:

```
>>> s = stripe.Subscription.retrieve('myid')
>>> s.id
'myid'
>>> s.delete()
>>> s.id
None
```
Maybe. Like in the Django ORM, if you delete an object, I think its primary key becomes `None` afterwards before Python "garbage collects". 

Also, I think I had in mind the idea that perhaps we want to trigger a signal BEFORE we delete it and one AFTER. 

BUT! Now I realize something stupid. The generator doesn't return `s.id` after the `s = stripe.Subscription.retrieve(()` call. It returns the `sub['id']` which most likely won't mutate. 

In conclusion, I think I definitely overcomplicated things. In the heat of the moment, I think I had grand ideas about refactoring some of the `payments` stuff but you're right, this can just return a list. I'll do that.

peterbe: Well, actually my intentions weren't that bad. You give it the customer ID and get a list of subscription *objects*. 
Either way, as of right now, none of the users of `cancel_stripe_customer_subscription` actually need the subscription *objects*. They're fine with it just being the IDs.

kuma/users/views.py
peterbe: I really don't understand the history or the current state. For example, what if I set up a subscription on Monday, change my primary MDN email on Tuesday, and revisit the Edit Profile page on Wednesday. Then the `stripe_customer_id` and `user.email` is going to diverge and it'll appear as if I don't have a subscription. 

It's tempting to delete this second check because emails aren't first-class citizens in Kuma. 

The thing is, what might be the case is that we know that about 20 people are still actively paying subscriptions through Stripe to MDN. Those subscriptions were set up when Kuma worked differently. That's a very different case.

kuma/users/views.py
peterbe: Actually. I'm just going to kill this. Mostly because we no longer have a record of WHY we're doing it and the simple above-mentioned use case where a MDN user changes her primary email and suddenly it appears as if she doesn't have a subscription.

kuma/users/views.py
Gregoor: I think I'm also in favor of changing the current behavior, especially given that we'll get a new stripe account soon anyway. All the points you mention make sense to me and I can't defend my past decision to make it email-dependent anymore.

kuma/users/views.py
```diff
@@ -467,14 +470,53 @@ def user_edit(request, username):
 
             return redirect(edit_user)
 
+    subscription_info = None
+    stripe_customer = get_stripe_customer(edit_user)
```
Gregoor: What I meant to ask with my comment before, can we extract that whole block 474-510 into its own method, similar to before? i.e. something like `retrieve_and_synchronize_subscriptions`?
Personally I like shorter methods which extract out what can be meaningfully named and extracted, just want to get a feel for if you feel similar or whether it's my personal preference.

kuma/users/views.py
```diff
@@ -467,14 +470,53 @@ def user_edit(request, username):
...
+                "brand": card.brand,
+                "expires_at": f"{card.exp_month}/{card.exp_year}",
+                "last4": card.last4,
+                # Cards that are part of a "source" don't have a zip
+                "zip": card.get("address_zip", None),
+            }
+
+            # To perfect the synchronization, take this opportunity to make sure
+            # we have an up-to-date record of this.
+            UserSubscription.set_active(edit_user, stripe_subscription_info.id)
```
Gregoor: Curious if we might want to record [stripe's subscription created](https://stripe.com/docs/api/subscriptions/object#subscription_object-created) date in our DB as well?

peterbe: I don't see that. The `UserSubscription` record (unless it's `canceled`) means the user is an active subscriber. I.e. that they're actively paying us $$$. 
The `UserSubscription.created` is already a record of when the subscription started.

kuma/users/views.py
peterbe: Yep yep. I gotcha. 
What I didn't like before was that view related stuff was put into the `stripe_utils.py` when it wasn't a utility, but actually totally related to the view part. Now, the edit user view function has become huge which isn't great. I'll refactor it a bit.

peterbe: Addressed in the latest commit.

PR #6683, opened 2020-03-16
Profile - Use React header and keep user on read only

schalkneethling:
This adds the React header to the edit profile page. This is another step towards converting the page to be 100% React-based.

This now also keeps the user on the read-only site as they navigate between their profile and the rest of MDN. This is particularly important once we open up subscriptions to a wider audience.

This _does_ introduce an additional `case` statement but, this is temporary and will be changed to `SPA` once the page is all React.

I have also removed the `Banner` component and we will not be using those on profile pages.

r?

fix #6683


kuma/static/styles/minimalist/main.scss
```diff
@@ -23,7 +23,6 @@ Main stylesheet for the site, standalone
 @import '../components/_tags.scss';
 @import '../components/_errorlist.scss';
 @import '../components/_newsletter.scss';
-@import './components/_newsletter.scss';
```
schalkneethling: Removed this duplicate `newsletter` import

Gregoor:
Does also converting the user detail page as part of this PR also make sense? Given that saving redirects there I thought it might be a good combo.

schalkneethling:
> Does also converting the user detail page as part of this PR also make sense? Given that saving redirects there I thought it might be a good combo.

Makes sense but, as I sneaked this in as is 😺, let's spin up a separate issue for that and merge this one as is. Thanks for the review and suggestion @Gregoor 🌮

PR #6688, opened 2020-03-17
Hide subscription form when stripe throws an exception

Gregoor:
In the spirit of the chaos monkey, let's at least let users edit their profile when stripe throws an unforeseen error, like in the case of: #6690  (which is still mysterious to me)

### DIFF
--- a/kuma/users/jinja2/users/user_edit.html
+++ b/kuma/users/jinja2/users/user_edit.html
@@ -128,7 +128,7 @@
         </li>
       </ul>
     </form>
-    {% if is_me and waffle.flag('subscription') and settings.STRIPE_PUBLIC_KEY %}
+    {% if is_me and waffle.flag('subscription') and settings.STRIPE_PUBLIC_KEY and not has_stripe_crashed %}
       {% include "users/includes/stripe_subscription.html" %}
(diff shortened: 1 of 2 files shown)


peterbe:
I'd MUCH rather we do something like...
```
except stripe.InvalidRequestError as error:
    ...
```
But more realistically, perhaps before we look at this hack, we dig deeper into the actual Stripe API error and perhaps review our `user.stripe_customer_id` etc. compared to what we have in Stripe.

Gregoor:
While I agree that we should also investigate the original error, I think we should also be prepared for any curve balls Stripe or someone using the over-eagerly using the dashboard might throw us.

Wrt narrowing the error: good point, will change it to that.

peterbe:
I say we close this. 
I think the original thought that started this draft was the idea that originally stemmed from the fact that we cross prod data with stage config. 
Also, the Stripe subscription form is soon moving from the wiki Edit Profile to the Subscription landing page. But I'll leave you to close the draft if you agree.

Gregoor:
If this was a more important view I might argue more for the chaos monkey, but given this is hopefully going away soon, probably not worth the effort.

PR #6692, opened 2020-03-17
translation over a deleted translation

### LINKED ISSUE #6673: 内部服务器错误(Internal Server Error)
**Summary**
_What is the problem?_

Each time I try to submit the translation,It shows error in specific page--`/docs/Learn/CSS/Building_blocks/Selectors`

The problem never happend when I try to submit in other pages.
**Steps To Reproduce (STR)**

1. try to translate the page /docs/Learn/CSS/Building_blocks/Selectors into zh-CN
2. which would open page https://wiki.developer.mozilla.org/zh-CN/docs/Learn/CSS/Building_blocks/Selectors$translate?tolocale=zh-CN
3. submit and it jumped to the "内部服务器错误"page.(Internal Server Error)

**Actual behavior**
_What actually happened?_
内部服务器错误Internal Server Error

**Expected behavior**
It should jumped to the page revised.

**Additional context**
None,thanks.

peterbe:
Fixes #6673

Here's how I tested it. First I went to a localized piece of content that works and exists. 
E.g. http://wiki.localhost.org:8000/sv-SE/docs/Mozilla/Add-ons/WebExtensions
Then I (soft) deleted it on http://wiki.localhost.org:8000/sv-SE/docs/Mozilla/Add-ons/WebExtensions$delete
Now, if I go back to http://wiki.localhost.org:8000/sv-SE/docs/Mozilla/Add-ons/WebExtensions it actually (rather sneakily) draws from the en-US document despite the locale in the URL. On that now English document (but with Swedish header & footer) I use the language drop-down and pick "Add translation". Then pick Swedish in the list. Then I just go ahead and try to create the new swedish translation. Should not crash :)

### DIFF
--- a/kuma/wiki/tests/test_views_translate.py
+++ b/kuma/wiki/tests/test_views_translate.py
@@ -71,3 +71,33 @@ def test_translate_post(root_doc, trans_doc_client):
         len(Document.objects.filter(title=root_doc.title + " Redirect 1", locale="fr"))
         == 0
     )
+
+
+def test_translate_over_deleted_document(root_doc, trans_doc_client):
+    """Trying to create a new translation when there already existed a document
+    for that locale, but it was soft-deleted.
+    If this is the case, it should fully delete the soft-deleted document and
+    proceed with the new translation.
+
+    Much of this is to assert that the fix in https://github.com/mdn/kuma/issues/6673
+    worked.
+    """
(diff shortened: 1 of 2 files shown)


peterbe:
@escattone I would love to get this in some time in the next release. The Chinese translators are doing an excellent job and with all the recent deletions of incomplete and abandoned translations, this patch is going to be important.

PR #6693, opened 2020-03-17
TypeError in search-results-page on overly big page

### LINKED ISSUE #6350: T - TypeError 'count.toLocaleString' in SSR
https://sentry.prod.mozaws.net/operations/mdn-prod/issues/7090931/

```
TypeError: Cannot read property 'toLocaleString' of undefined
  File "/app/kuma/javascript/dist/ssr.js", line 22, in kt
    '{snip} ocuments found for "%(query)s" in %(locale)s.',a),{count:a.toLocaleString(),locale:n,query:c})," ",t))}function jt(e){var t=e.locale;return( {snip}
  File "/app/kuma/javascript/dist/ssr.js", line 22, in a
    '{snip} .state);null!=d&&(f.state=r({},f.state,d))}}else if(D={},f=a(o.props,i,s),null==(f=G(a,o.props,f,i))||null==f.render)return void he(e=f,a);i {snip}
  File "/app/kuma/javascript/dist/ssr.js", line 22, in ye
    '{snip} lement(e);){var i=e,c=i.type;if("function"!=typeof c)break;a(i,c)}return{child:e,context:t}}var ve=function(){function e(t,n){if(!(this inst {snip}
  File "/app/kuma/javascript/dist/ssr.js", line 22, in e.render
    '{snip} -- --\x3e"+I(n):(this.previousWasTextNode=!0,I(n));if(e=(t=ye(e,t,this.threadID)).child,t=t.context,null===e||!1===e)return"";if(!o.isValidE {snip}
  File "/app/kuma/javascript/dist/ssr.js", line 22, in e.read
    '{snip} +=c}else{var f=i.children[i.childIndex++],m="";try{m+=this.render(f,i.context,i.domNamespace)}catch(e){throw e}r.length<=this.suspenseDepth& {snip}
...
(5 additional frame(s) were not displayed)
```

peterbe:
Fixes #6350

To reproduce, make a search, on the read-only site, that yields few (less than 100) search results. You may or may not see a "Next" button at the bottom of the page. Click that till the "Next" button doesn't appear anymore. At the point increment the `page=NUMBER` to `page=NUMBER+1` 
E.g. http://localhost.org:8000/en-US/search?q=intersectionobserver&page=2
Before you would get an error inside SSR because the React code assumes there is a count. 
With this change you get this:
Screen Shot 2020-03-17 at 4 44 42 PM

### DIFF
--- a/kuma/search/tests/test_views.py
+++ b/kuma/search/tests/test_views.py
@@ -166,6 +166,14 @@ class ViewTests(ElasticTestCase):
         # rest_framework works.
         assert response.status_code == 404
 
+    def test_page_too_large(self):
+        # Note that this doesn't use the Wiki host
+        response = self.client.get("/en-US/search", {"page": 2, "q": "anything"})
+        # The search 'q' is fine and the page isn't too large on its own, but
+        # the pagination doesn't go beyond the first page so the rest_frameworks
+        # pagination will trigger an error.
+        assert response.status_code == 400
+
     def test_tokenize_camelcase_titles(self):
         for q in ("get", "element", "by", "id"):
             response = self.client.get(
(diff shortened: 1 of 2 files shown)


peterbe:
@Gregoor ping for a review

PR #6694, opened 2020-03-18
refactor banners.jsx

### LINKED ISSUE #6659: T - Refactor Banners-Component
Copied from [my comment](https://github.com/mdn/kuma/pull/6633#pullrequestreview-371057988) over in #6633

> The comments in banners.jsx feel a little outdated, could use some touch-up. I also think the Banners-Component isn't greatly named and could be more straight-forward. Sorry, I know this is not really what this PR is about, but since we're already touching it maybe sth worth thinking about. I.e. what I'd do:
> Instead of `useMemo` and defining a list of objects, I'd so something like (fairly incomplete version, just showing the idea):
> 
> ```js-jsx
> export default function ActiveBanner() {
>     const userData = useContext(UserProvider.context);
> 
>     const bannerId = userData ? userData.waffle.flags[banner.id] : null;
>     if (!bannerId) {
>         return null;
>     }
> 
>     switch (bannerId) {
>         case 'developer_needs':
>             return (
>                 <Banner
>                     id="developer_needs"
>                     className="developer-needs"
>                     title={gettext('MDN Survey')}
>                 />
>             );
> 
>         case 'mdn_subscriptions':
>             if (!userData.isAuthenticated) {
>                 return null;
>             }
> 
>             return (
>                 <Banner
>                     id="mdn_subscriptions"
>                     className="mdn-subscriptions"
>                     title={gettext('Become a monthly supporter')}
>                     url={`/${locale}/profiles/${userData.username}/edit#subscription`}
>                 />
>             );
> 
>       default:
>           console.warn("Unknown banner for id, bannerId"); // or even better: throw it in sentry
>           return null;
>     }
> }
> ```
> 
> Imo it currently suffers from premature abstraction rn which hurts readability, straightforwardness and adds typing noise.
> 
> I'd also move `isEmbargoed` into Banner itself, since this is also where it's being set.

Gregoor:
Fix #6659

I found the indirection added by making the render data-structure based hindered readability and modifiability (is that a word) of the code. And it leaked which lead to a declarative-imperative mishmash. Imo it's more straightforwad and readable now.

Specifics:
- rename it to active-banner.jsx, to more actively reflect what it will render
- change the declarative-ish data structure based flow to a more imperative one
- remove banners prop which was only used for testing
  - tie tests closer to the actual banners being shown


kuma/javascript/src/active-banner.jsx
```diff
@@ -182,86 +179,80 @@ function Banner(props: BannerProps) {
...
-                // The userData context isn't available yet so
-                // this is set when the banner is used.
-                url: '',
-                authenticated: true,
-                embargoDays: 7
-            }
-        ],
-        [locale] // This tells useMemo() to only compute this once.
-    );
+    if (!userData) {
```
peterbe: Why this?

peterbe: Wait! I get it now. The `userData` is the result of `whoami` which doesn't necessarily require that the user is authenticated. It's just a shortcut for checking if we know about the waffle flags yet.

kuma/javascript/src/active-banner.jsx
```diff
@@ -182,86 +179,80 @@ function Banner(props: BannerProps) {
...
+                    url={`${
+                        window.mdn ? window.mdn.wikiSiteUrl : ''
+                    }/${locale}/profiles/${userData.username ||
+                        ''}/edit#subscription`}
+                    embargoDays={7}
+                />
+            );
+
+        default:
+            // TODO: throw it into Sentry once that's set-up for the frontend
```
peterbe: I don't think we want to set up Sentry for the front-end. :)

kuma/javascript/src/active-banner.jsx
```diff
@@ -182,86 +179,80 @@ function Banner(props: BannerProps) {
...
+            // eslint-disable-next-line no-console
+            console.warn('Unknown banner for id', bannerId);
+            return null;
```
peterbe: I'd much rather this was `throw new Error(`Unknown banner '${bannerId}'`)` but I could go either way.

peterbe:
Yep. It works.

kuma/javascript/src/active-banner.jsx
```diff
@@ -182,86 +179,80 @@ function Banner(props: BannerProps) {
...
+                    cta={gettext('Subscribe')}
+                    url={`${
+                        window.mdn ? window.mdn.wikiSiteUrl : ''
+                    }/${locale}/profiles/${userData.username ||
+                        ''}/edit#subscription`}
+                    embargoDays={7}
+                />
+            );
+
+        default:
```
peterbe: How could this ever happen? The `bannerId` will never be anything that is NOT inside `BANNER_IDS`.
The only reason we could end up down here is if the `BANNER_IDS` array isn't the same as the list of blocks in this switch statement. 

It is, after all, kinda strange that we're actually maintaining two lists. One is the `BANNER_IDS` array. The other one is the list of case statements in the `switch` code. So that's not gorgeous but it'll be fine. 

Another way of structuring the code would be to put it like this:

```javascript
const BANNERS = {
  developer_needs: id => {
     return <Banner id={id} className="developer-needs" ... />
  },
  subscription_banner: id => {
      return <Banner id={id} className="subscription-banner" ... />
  }
}

for (const ([key, value]) of Object.entries(userData.waffle.flags)) {
  if (key in BANNERS && value && !isEmbargoed(key)) {
     return BANNERS[key](key);
  }
}
return null;
```

That way, the list of waffle flag keys is only listed 1 single time.

Gregoor:
Excellent point(s) thank you Peter, my initial refactor was apparently a bit short-sighted. I adjusted it similar to your recommendation, though I stuck with the switch (instead moved it inside of a for loop, which does make the whole thing a lil over-nested to be fair).

PR #6706, opened 2020-03-19
Subscription banner to be hidden to active subscribers

### LINKED ISSUE #6702: Subscription banner to be hidden to active subscribers
**Summary**
If you are an active subscriber you shouldn't need to see the subscription banner. 

Depends on https://github.com/mdn/kuma/pull/6680 which exposes `user.is_subscriber` in the `/api/v1/whoami` 

Today the subscription banner depends on the `subscription_banner` waffle flag and if the user is authenticated. The reason it depends on the user being authenticated is because it links to the Edit Profile which anonymous users won't have. 

**Audience**
Everyone.

**Rationale**
Don't want to sell to sold people. 

**Workaround**
Live with the banner appear to people already on board. 

**Proposal**
Where we do the waffle flag check *and* the `isEmbargoed('subscription_banner')` we also add another layer business logic. 

**Additional context**
https://github.com/mdn/kuma/issues/6701 will remove the need to be authenticated.

https://github.com/mdn/kuma/pull/6694 is refactoring how the JS code works in the banners in general. 

But note that this means the subscription banner will need to depend on custom logic that is unique to the subscription banner. I.e. the waffle flag logic and the `isEmbargoed` logic are both generic and apply equally to the subscription banner and the developer needs banner. So we might need to cut a bit depper.

peterbe:
Fixes #6702

I didn't realize this but in our jsx files, we're converting the result of `/api/v1/whoami` into a brand new object and I didn't update that when I added the `is_subscriber` to Django. And the tests didn't notice :(

### DIFF
--- a/kuma/javascript/src/active-banner.jsx
+++ b/kuma/javascript/src/active-banner.jsx
@@ -55,6 +55,9 @@ import CloseIcon from './icons/close.svg';
 import { getLocale, gettext, Interpolated } from './l10n.js';
 import UserProvider from './user-provider.jsx';
 
+export const DEVELOPER_NEEDS_ID = 'developer_needs';
+export const SUBSCRIPTION_ID = 'subscription_banner';
+
 // Set a localStorage key with a timestamp the specified number of
 // days into the future. When the user dismisses a banner we use this
 // to prevent the redisplay of the banner for a while.
@@ -98,6 +101,21 @@ function isEmbargoed(id) {
         // just won't work
         return false;
     }
+}
+
+// Certain banners don't apply to the user based on other reasons.
+// For example, the Subscription banner should not be displayed
+// if the user is authenticated and has 'isSubscriber'.
+function notForThisUser(id, userData) {
+    // If the banner is the subscription banner, and the user has a truthy
+    // 'isSubscriber' they don't get this banner.
+    if (id === SUBSCRIPTION_ID) {
+        if (userData.isSubscriber) {
+            // This user will NOT get this banner.
+            return true;
+        }
+    }
+    return false;
 }
 
 // The <Banner> component displays a simple call-to-action banner at
@@ -179,9 +197,6 @@ function Banner(props: BannerProps) {
     );
 }
 
-export const DEVELOPER_NEEDS_ID = 'developer_needs';
-export const SUBSCRIPTION_ID = 'subscription_banner';
-
 export default function ActiveBanner() {
     const userData = useContext(UserProvider.context);
     const locale = getLocale();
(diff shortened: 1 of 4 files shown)


kuma/javascript/src/active-banner.jsx
```diff
@@ -55,6 +55,9 @@ import CloseIcon from './icons/close.svg';
 import { getLocale, gettext, Interpolated } from './l10n.js';
 import UserProvider from './user-provider.jsx';
 
+export const DEVELOPER_NEEDS_ID = 'developer_needs';
+export const SUBSCRIPTION_ID = 'subscription_banner';
```
peterbe: The reason I *moved* these to up here was because it felt better that way. Now, there's a new utility function which depends on it and it didn't feel right to have the constants somewhere below.

kuma/javascript/src/active-banner.jsx
```diff
@@ -100,6 +103,21 @@ function isEmbargoed(id) {
     }
 }
 
+// Certain banners don't apply to the user based on other reasons.
+// For example, the Subscription banner should not be displayed
+// if the user is authenticated and has 'isSubscriber'.
+function notForThisUser(id, userData) {
```
peterbe: Not in love with this name. Is it good enough?

kuma/javascript/src/active-banner.jsx
```diff
@@ -191,7 +206,11 @@ export default function ActiveBanner() {
     }
 
     for (const id in userData.waffle.flags) {
-        if (!userData.waffle.flags[id] || isEmbargoed(id)) {
+        if (
+            !userData.waffle.flags[id] ||
+            isEmbargoed(id) ||
+            notForThisUser(id, userData)
```
peterbe: Since that function only does one thing. I could move its logic to somewhere here. Something like this:

```diff
if (!userData.waffle.flags[id] || isEmbargoed(id)) {
    continue;
}
+// Code comment...
+if (banner === SUBSCRIPTION_ID && userData.isSubscriber)  {
+   continue;
+}
```

would you prefer that?

kuma/javascript/src/active-banner.test.js
```diff
@@ -7,11 +7,11 @@ import ActiveBanner, {
 } from './active-banner.jsx';
 import UserProvider from './user-provider.jsx';
 
-const mockUserData = { ...UserProvider.defaultUserData };
-
 describe('ActiveBanner', () => {
+    let mockUserData;
     beforeEach(() => {
         localStorage.clear();
+        mockUserData = { ...UserProvider.defaultUserData };
```
peterbe: Before, the `mockUserData` was created once and mutated across each test making it so that that mutation is applied to each and every test making the tests indeterministic. 

What I *could* do, if you prefer, is to simply put `const mockUserData = { ...UserProvider.defaultUserData };` into every test. Is that neater?

kuma/javascript/src/active-banner.jsx
Gregoor: I think it's simple enough that I'd actually prefer that, yup!

peterbe:
@Gregoor I suppose, since you approved + nit, and I tackled the nit. That means I can merge this when it goes green. Thanks!

Gregoor:
yup, that's exactly how I meant it! 🚀

PR #6709, opened 2020-03-20
redirect to parent of deleted documents if possible

### LINKED ISSUE #6708: Redirect to parent on deleted documents that can't be found
**Summary**
Take https://developer.mozilla.org/es/docs/Tools/Page_Inspector/Atajos_teclado  for example. 
It's a translation of https://developer.mozilla.org/es/docs/Tools/Page_Inspector/Keyboard_shortcuts but the only thing they've translated is the document title and document slug. 
If we delete this document now, on the grounds that it's actually not translated and potentially out of date with the risk of sending Spanish speaking MDN users to see old content. Then, what will happen is two things:

1. It'll cease to be included in sitemap XML files (for `es`). A good thing!
2. People will get a `404 Not Found` on that URL if they, had it bookmarked or linked to.

**Rationale**
If we don't gracefully deal with deletion of untranslated translations, where the slug is translated, we're running the risk of making sadness happen. 

**Audience**
Hmm..

**Proposal**
We leverage the soft-deleted documents! The page would still exist in our database (until it's purged, aka. Yari) so we can look it up like this:
```python
# somewhere in kuma/wiki/views/document.py

def react_document(request, document_slug, document_locale):
    ...
    doc, fallback_reason = _get_doc_and_fallback_reason(document_locale, document_slug)
    if doc is None:
        fallback_doc, fallback_reason, redirect_url = _default_locale_fallback(
            request, document_slug, document_locale
        )
       if fallback_doc:
            redirect(redirect_url) 
      else:
         
          # THIS IS WHAT I'M PROPOSING
          deleted_document = Document.deleted_objects.get(locale=document_locale, slug=document_slug):
          if deleted_document.parent:
                 return redirect(deleted_document.parent.get_absolute_url())
         # /END OF PROPOSAL

         raise Http404

    ...
```

**Additional context**
This will really help the effort to get rid of unmaintained translations that are of really low quality. Be that from not actually being translated or just horribly out of date for some other reason.

peterbe:
Fixes #6708

Steps to reproduce. 
Go to any en-US page on wiki.localhost.org:8000 and click to make a translation of it. Pick a language. When you edit, the only thing you change is the title and publish. Make a page like this for example: http://wiki.localhost.org:8000/sv-SE/docs/Web/HTML/Element/l%C3%A4nk
Now, once published, use the Wiki to delete it. That means you'll get a 404, in `curl`, on http://localhost.org:8000/sv-SE/docs/Web/HTML/Element/l%C3%A4nk
Now, enable this branch and that 404 will turn into a 302.

### DIFF
--- a/kuma/wiki/views/document.py
+++ b/kuma/wiki/views/document.py
@@ -894,7 +894,7 @@ def react_document(request, document_slug, document_locale):
                 return redirect(redirect_url)
         else:
             # It could be that the document you're trying to view was deleted and
-            # the reason we're not finding a fallback is the because the slug
+            # the reason we're not finding a fallback is because the slug
             # doesn't match.
             # E.g. you're trying to view `/sv-SE/docs/Foö/Bår` but that document
             # was deleted and as a soft-delete its parent was `/en-US/docs/Foo/Bar`
--- a/kuma/wiki/views/document.py
+++ b/kuma/wiki/views/document.py
@@ -897,7 +897,7 @@ def react_document(request, document_slug, document_locale):
             # the reason we're not finding a fallback is because the slug
             # doesn't match.
             # E.g. you're trying to view `/sv-SE/docs/Foö/Bår` but that document
-            # was deleted and as a soft-delete its parent was `/en-US/docs/Foo/Bar`
+            # was soft-deleted and its parent was `/en-US/docs/Foo/Bar`
             if document_locale != settings.LANGUAGE_CODE:  # Not in English!
                 redirect_url = _get_deleted_parent_redirect_url(
                     document_locale, document_slug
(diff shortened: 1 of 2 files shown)


kuma/wiki/views/document.py
```diff
@@ -885,6 +893,18 @@ def react_document(request, document_slug, document_locale):
             if redirect_url is not None:
                 return redirect(redirect_url)
         else:
+            # It could be that the document you're trying to view was deleted and
+            # the reason we're not finding a fallback is the because the slug
```
escattone: Typo:
```suggestion
            # the reason we're not finding a fallback is because the slug
```

kuma/wiki/views/document.py
```diff
@@ -885,6 +893,18 @@ def react_document(request, document_slug, document_locale):
...
+            # doesn't match.
+            # E.g. you're trying to view `/sv-SE/docs/Foö/Bår` but that document
+            # was deleted and as a soft-delete its parent was `/en-US/docs/Foo/Bar`
```
escattone: I think this reads better?
```suggestion
            # was soft-deleted and its parent was `/en-US/docs/Foo/Bar`
```

Issue #6578, opened 2020-02-24
T - Subscription: 'Become a monthly supporter' embedded on FAQ page


tobinmori:
**Description**:
Logged in users who click on subscribe in the banner see embedded widget on FAQ page to become monthly subscribers

**Acceptance Criteria**:
- [x] when logged in users click on subscribe in the banner they see the FAQ page to become monthly subscribers
- [x] has checkbox to verify that user wants to become a subscriber and acts as a gate to Stripe modal (no data needs to be stored upon confirmation)
- [x] Text in widget is identical to design for legal reasons (text in task issue)
- [x] If user clicks subscribe without having checked the box, an error message is shown
- [x] If the checkbox is checked and the user clicks on subscribe, a stripe modal is shown

**Design**:
https://app.abstract.com/projects/9decc480-70fc-11e9-9453-4f3fbffea7ce/branches/64547606-339e-494d-8d45-559aa506d80a/commits/6ae7cc2467cf36a609b9b2bc5189dabf9c5c4e7e/files/96E76C7B-2AF8-4473-A39E-E81890F91955/layers/E3B8DB66-C6A4-4BDC-A2FF-A80A02DAC2E8?collectionId=0750a16f-5979-4db7-9b1b-d4d676607dcd&collectionLayerId=2536d7fc-cd8e-4b9b-9580-615dc1f76e70&mode=design

**Notes**:

tobinmori:
@Gregoor this issue can began after @schalkneethling finishes the 'bare bones' version of #6577

schalkneethling:
> @Gregoor this issue can began after @schalkneethling finishes the 'bare bones' version of #6577

Also, just a heads up @Gregoor. This may be deferred to the next sprint as we will not be adding this interaction in this sprint. Instead, we will merely send users to the edit profile page where they will complete the sign up using the feature you have already implemented there.

Gregoor:
Good to know, thanks! I'll hold my horses then 🐴

tobinmori:
@Gregoor I've moved this one out of sprint Bravo for now. We can move it in once the bare bones banner is in.

tobinmori:
We decided to eliminate this step, so closing this.

schalkneethling:
![Screenshot 2020-03-23 at 12 25 04](https://user-images.githubusercontent.com/10350960/77307099-5663d700-6d01-11ea-9cf6-5bb21c307646.png)

Confirming that the above is the design and wording that is being implemented.

[...]

PR #6723, opened 2020-03-24
Add embedded subscription form to subscription landing page

schalkneethling:
This concludes the UI portion of #6578 

NOTE: We will not show the component for mobile and tablet

## Mobile

Screenshot 2020-03-24 at 14 42 48

## Tablet

Screenshot 2020-03-24 at 14 42 56

## Standard Desktop

Screenshot 2020-03-24 at 14 43 04

## Wide Desktop

Screenshot 2020-03-24 at 14 43 12

@Gregoor let's take this thing over the finish line :)

- [x] I have not placed this behind any form of waffle.
- [ ] ~~The logged in users email address needs to be exposed so we can set it~~
- [ ] On successful subscribe event the user needs to be directed to the Thank you page.

I _think_ that might be the only missing pieces.

### DIFF
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -7,6 +7,7 @@ import Header from '../header/header.jsx';
 import Footer from '../footer.jsx';
 import Route from '../route.js';
 import SubHeader from './subheader.jsx';
+import SubscriptionForm from './subscription-form.jsx';
 import ListItem from './list-item.jsx';
 
 type PaymentsRouteParams = {
@@ -32,11 +33,14 @@ export default function PaymentsLandingPage() {
         <>
             <A11yNav />
             <Header />
-            <SubHeader
-                title="Become a monthly supporter"
-                description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
-                columnWidth="7"
-            />
+            <div className="subscriptions subheader-container">
+                <SubHeader
+                    title="Become a monthly supporter"
+                    description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
+                    columnWidth="7"
+                />
+                <SubscriptionForm />
+            </div>
             <main
                 id="contributions-page"
                 className="contributions-page"
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -33,12 +33,8 @@ export default function PaymentsLandingPage() {
         terms: `/${locale}/payments/terms`
     };
 
-    let showSubscriptionForm = false;
-    // if the subscription banner is not enabled,
-    // do not show the <SubscriptionForm />
-    if (userData && userData.waffle.flags['subscription_banner']) {
-        showSubscriptionForm = true;
-    }
+    const showSubscriptionForm =
+        userData && userData.waffle.flags.subscription_banner;
 
     return (
         <>
@@ -50,7 +46,7 @@ export default function PaymentsLandingPage() {
                     description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
                     columnWidth="7"
                 />
-                <SubscriptionForm showSubscriptionForm={showSubscriptionForm} />
+                {showSubscriptionForm && <SubscriptionForm />}
             </div>
             <main
                 id="contributions-page"
--- a/kuma/javascript/src/payments/index.jsx
+++ b/kuma/javascript/src/payments/index.jsx
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import { useContext } from 'react';
 
 import { getLocale, gettext, Interpolated } from '../l10n.js';
 import A11yNav from '../a11y/a11y-nav.jsx';
@@ -10,11 +11,14 @@ import SubHeader from './subheader.jsx';
 import SubscriptionForm from './subscription-form.jsx';
 import ListItem from './list-item.jsx';
 
+import UserProvider from '../user-provider.jsx';
+
 type PaymentsRouteParams = {
     locale: string
 };
 
 export default function PaymentsLandingPage() {
+    const userData = useContext(UserProvider.context);
     const locale = getLocale();
     const urls = {
         annualReport:
@@ -29,6 +33,13 @@ export default function PaymentsLandingPage() {
         terms: `/${locale}/payments/terms`
     };
 
+    let showSubscriptionForm = false;
+    // if the subscription banner is not enabled,
+    // do not show the <SubscriptionForm />
+    if (userData && userData.waffle.flags['subscription_banner']) {
+        showSubscriptionForm = true;
+    }
+
     return (
         <>
             <A11yNav />
@@ -39,7 +50,7 @@ export default function PaymentsLandingPage() {
                     description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
                     columnWidth="7"
                 />
-                <SubscriptionForm />
+                <SubscriptionForm showSubscriptionForm={showSubscriptionForm} />
             </div>
             <main
                 id="contributions-page"
--- a/kuma/javascript/src/payments/subscription-form.jsx
+++ b/kuma/javascript/src/payments/subscription-form.jsx
@@ -64,7 +64,7 @@ export default function SubscriptionForm() {
                 name="subscription-form"
                 method="post"
                 onSubmit={submit}
-                disabled={subscribeFormDisabled ? 'disabled' : ''}
+                disabled={subscribeFormDisabled}
             >
                 <label className="payment-opt-in">
                     <input
@@ -81,7 +81,7 @@ export default function SubscriptionForm() {
                                 <a
                                     target="_blank"
                                     rel="noopener noreferrer"
-                                    href={`/${locale}/payments/`}
+                                    href={`/${locale}/payments/terms/`}
                                 >
                                     {gettext('Payment Terms')}
                                 </a>
@@ -94,7 +94,7 @@ export default function SubscriptionForm() {
                 <button
                     type="submit"
                     className="button cta primary"
-                    disabled={subscribeButtonEnabled ? '' : 'disabled'}
+                    disabled={subscribeButtonEnabled}
                 >
                     {gettext('Continue')}
                 </button>
--- a/kuma/javascript/src/payments/subscription-form.jsx
+++ b/kuma/javascript/src/payments/subscription-form.jsx
@@ -4,19 +4,11 @@ import { useRef, useState } from 'react';
 
 import { getLocale, gettext, Interpolated } from '../l10n.js';
 
-type Props = {
-    showSubscriptionForm: boolean
-};
-
-export default function SubscriptionForm({ showSubscriptionForm }: Props) {
+export default function SubscriptionForm() {
     const locale = getLocale();
     const subscriptionFormRef = useRef(null);
     const [subscribeFormDisabled, setSubscribeFormDisabled] = useState(false);
     const [subscribeButtonEnabled, setSubscribeButtonEnabled] = useState(false);
-
-    if (!showSubscriptionForm) {
-        return null;
-    }
 
     const toggleButton = () => {
         setSubscribeButtonEnabled(!subscribeButtonEnabled);
--- a/kuma/javascript/src/payments/subscription-form.jsx
+++ b/kuma/javascript/src/payments/subscription-form.jsx
@@ -105,6 +105,9 @@ export default function SubscriptionForm() {
                 >
                     {gettext('Continue')}
                 </button>
+                <small className="subtext">
+                    {gettext('Payments are not tax deductible')}
+                </small>
             </form>
         </div>
     );
--- a/kuma/javascript/src/payments/subscription-form.jsx
+++ b/kuma/javascript/src/payments/subscription-form.jsx
@@ -1,0 +1,111 @@
+import * as React from 'react';
+import { useContext, useRef, useState } from 'react';
+
+import { getLocale, gettext, Interpolated } from '../l10n.js';
+import UserProvider from '../user-provider.jsx';
+
+export default function SubscriptionForm() {
+    const locale = getLocale();
+    const subscriptionFormRef = useRef(null);
+    const userData = useContext(UserProvider.context);
+    const [subscribeFormDisabled, setSubscribeFormDisabled] = useState(false);
+    const [enableSubscribeButton, setEnableSubscribeButton] = useState(false);
+
+    /**
+     * Toggles the enabled state of the subscription button
+     */
+    const enableButton = () => {
+        if (!enableSubscribeButton) {
+            setEnableSubscribeButton(true);
+        } else {
+            setEnableSubscribeButton(false);
+        }
+    };
+
+    /**
+     * Opens Stripe modal allowing a user to complete their subscription.
+     * @param {Object} event - The form submit event
+     */
+    const submit = event => {
+        event.preventDefault();
+
+        const subscriptionForm = subscriptionFormRef.current;
+        const formData = new FormData(subscriptionForm);
+
+        setSubscribeFormDisabled(true);
+
+        const stripeHandler = window.StripeCheckout.configure({
+            key: window.mdn.stripePublicKey,
+            locale: 'auto',
+            name: 'MDN Web Docs',
+            zipCode: true,
+            currency: 'usd',
+            amount: 500,
+            email: userData ? userData.email : '', // TODO: email is not currently exposed
+            token: function(response) {
+                formData.set('stripe_token', response.id);
+                subscriptionForm.submit();
+            },
+            closed: function() {
+                if (!formData.get('stripe_token')) {
+                    setSubscribeFormDisabled(false);
+                }
+            }
+        });
+
+        stripeHandler.open();
+    };
+
+    return (
+        <div className="subscriptions-form">
+            <header className="subscriptions-form-header">
+                <h2>
+                    <Interpolated
+                        id={gettext('$5 <perMontSub />')}
+                        perMontSub={<sub>{gettext('/mo')}</sub>}
+                    />
+                </h2>
+            </header>
+            <form
+                ref={subscriptionFormRef}
+                name="subscription-form"
+                method="post"
+                onSubmit={submit}
+                disabled={subscribeFormDisabled ? 'disabled' : ''}
+            >
+                <label className="payment-opt-in">
+                    <input
+                        type="checkbox"
+                        required="required"
+                        onClick={enableButton}
+                    />
+                    <small>
+                        <Interpolated
+                            id={gettext(
+                                'By clicking this button, I authorize Mozilla to charge this payment method each month, according to the <paymentTermsLink />, until I cancel my subscription.'
+                            )}
+                            paymentTermsLink={
+                                <a
+                                    target="_blank"
+                                    rel="noopener noreferrer"
+                                    href={`/${locale}/payments/`}
+                                >
+                                    {gettext('Payment Terms')}
+                                </a>
+                            }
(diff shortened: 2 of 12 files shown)


kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,111 @@
...
+        setSubscribeFormDisabled(true);
+
+        const stripeHandler = window.StripeCheckout.configure({
+            key: window.mdn.stripePublicKey,
+            locale: 'auto',
+            name: 'MDN Web Docs',
+            zipCode: true,
+            currency: 'usd',
+            amount: 500,
+            email: userData ? userData.email : '', // TODO: email is not currently exposed
```
schalkneethling: I did not include the `label` attribute here as I get the following warning in the console:

```
StripeCheckout.configure: Unrecognized option 'label'. Did you mean 'name'? You can learn about the available configuration options in the Checkout docs: https://stripe.com/docs/checkout
```

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+        stripeHandler.open();
+    };
+
+    return (
+        <div className="subscriptions-form">
+            <header className="subscriptions-form-header">
+                <h2>
+                    <Interpolated
+                        id={gettext('$5 <perMontSub />')}
```
schalkneethling: Here is another hardcoded `$5`

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Toggles the enabled state of the subscription button
+     */
+    const enableButton = () => {
+        if (!enableSubscribeButton) {
+            setEnableSubscribeButton(true);
+        } else {
+            setEnableSubscribeButton(false);
+        }
```
peterbe: `setEnableSubscribeButton(!enableSubscribeButton)` ?

schalkneethling: Doh! of course :)

schalkneethling:
@Gregoor Just had a talk with @atopal 

> I have not placed this behind any form of waffle.

I can go ahead and wrap this component in a waffle flag. I will use the same one that is used for the banner. That way, when you turn off the banner, you also turn off this component.

> The logged in users email address needs to be exposed so we can set it

Kadir said that this should not be a blocker for this PR. We should instead create an issue for this work(adding it to the API and setting it on the form) and then improve this after this one is merged.

> On successful subscribe event the user needs to be directed to the Thank you page.

I can then also take this one.

Gregoor:
@schalkneethling that works for me too :+1: Then I'll just review it for now

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+import { getLocale, gettext, Interpolated } from '../l10n.js';
+import UserProvider from '../user-provider.jsx';
+
+export default function SubscriptionForm() {
+    const locale = getLocale();
+    const subscriptionFormRef = useRef(null);
+    const userData = useContext(UserProvider.context);
+    const [subscribeFormDisabled, setSubscribeFormDisabled] = useState(false);
+    const [enableSubscribeButton, setEnableSubscribeButton] = useState(false);
```
Gregoor: In the spirit of consistency with L11, I'd recommend going with `subscribeButtonEnabled`

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Toggles the enabled state of the subscription button
+     */
+    const enableButton = () => {
```
Gregoor: `toggleButton` would be more fitting (and imo then the comment is also superfluous).

schalkneethling:
> On successful subscribe event the user needs to be directed to the Thank you page.

@Gregoor The above is done on the Stripe side correct? i.e. it is a config not something we do in the code?

Gregoor:
> 
> 
> > On successful subscribe event the user needs to be directed to the Thank you page.
> 
> @Gregoor The above is done on the Stripe side correct? i.e. it is a config not something we do in the code?

Yes.. no.. ish.. 😁 The flow is: stripe-client-modal gives us a token, we post that to our backend, which uses it to create a customer and a subscription, and responds to our frontend with the new user data. So this form will need post to our backend ultimately and after that we could redirect (or we do it optimistically while submission is in progress).

kuma/javascript/src/payments/index.jsx
```diff
@@ -28,15 +33,25 @@ export default function PaymentsLandingPage() {
...
-                description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
-                columnWidth="7"
-            />
+            <div className="subscriptions subheader-container">
+                <SubHeader
+                    title="Become a monthly supporter"
+                    description="Support MDN with a $5 monthly subscription and get back more of the knowledge and tools you rely on for when your work has to work."
+                    columnWidth="7"
+                />
+                <SubscriptionForm showSubscriptionForm={showSubscriptionForm} />
```
schalkneethling: @Gregoor not sure is this is the best way to do this. Also not sure if it would be better to instead use `useState` instead of a variable.

With the current set-up, things work as expected but, I do see the following in the developer console:

```
Warning: Did not expect server HTML to contain a <div> in <div>.
```

Note: I am testing this after running `npm test` so the SSR server is not running.

What I also notice is that if I add a `console.log` at line 16 of `subscription-form.jsx` it is logged 3 times on a single page load. Perhaps this is normal, but I thought I would call it out.

schalkneethling: Also, if it _is_ better to use `useState`, I am curious whether it is a case of, "in these kinds of cases, _always_ use useState"

kuma/javascript/src/payments/index.jsx
```diff
@@ -28,15 +33,25 @@ export default function PaymentsLandingPage() {
         terms: `/${locale}/payments/terms`
     };
 
+    let showSubscriptionForm = false;
+    // if the subscription banner is not enabled,
+    // do not show the <SubscriptionForm />
+    if (userData && userData.waffle.flags['subscription_banner']) {
+        showSubscriptionForm = true;
+    }
```
Gregoor: I'd go with

```jsx
const showSubscriptionForm = userData && userData.waffle.flags.subscription_banner;
```

kuma/javascript/src/payments/index.jsx
Gregoor: Since the component doesn't need to do anything when not shown (apart from not showing up), a more React-y way of doing this would be:

```jsx
return (
  <stuff>
  {showSubscriptionForm && <SubscriptionForm />}
  </stuff>
);
```

Then you can also forego having `showSubscriptionForm` as a prop.

Wrt client and server rendering differing as well as render running multiple times, will look at it in action, but here's some wild guesses:
- SSR is at times quite stubborn for me locally with rendering new changes, it might also still be running on an old version of the code for you
- render does need to be re-run at least once, given that userData only comes in later (after the first render)

Gregoor: > 
> 
> Also, if it _is_ better to use `useState`, I am curious whether it is a case of, "in these kinds of cases, _always_ use useState"

`useState` would not be better here, since the value of the variable is wholly dependdent on the context's state and can be computed from it with no larger costs. If it was something computationally heavy it might be justified to use state with a function initializer, i.e.
```js
const showSubscriptionForm = useState(() => someHeavyComputation());
```

schalkneethling: > Since the component doesn't need to do anything when not shown (apart from not showing up), a more React-y way of doing this would be:

I completely forgot that I could do that! Thanks Gregor!

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,107 @@
...
+                key: window.mdn.stripePublicKey,
+                locale: 'auto',
+                name: 'MDN Web Docs',
+                zipCode: true,
+                currency: 'usd',
+                amount: 500,
+                email: '',
+                token: function(response) {
+                    formData.set('stripe_token', response.id);
+                    subscriptionForm.submit();
```
schalkneethling: > Yes.. no.. ish.. 😁 The flow is: stripe-client-modal gives us a token, we post that to our backend, which it uses to create a customer and a subscription via the Stripe API. Then our backend responds to our frontend with the new user data. So this form will need post to our backend ultimately and after that we could redirect (or we do it optimistically while submission is in progress).

That is what is happening here correct? So the Python code that recieves this form submission is the piece that needs to do the redirect to the Thank you page.

kuma/javascript/src/payments/index.jsx
schalkneethling: Perhaps you can send a PR to update the docs here :)
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,107 @@
...
+                        perMontSub={<sub>{gettext('/mo')}</sub>}
+                    />
+                </h2>
+            </header>
+            <form
+                ref={subscriptionFormRef}
+                name="subscription-form"
+                method="post"
+                onSubmit={submit}
+                disabled={subscribeFormDisabled ? 'disabled' : ''}
```
peterbe: Use a plain boolean here instead:
```suggestion
                disabled={subscribeFormDisabled}
```

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,107 @@
...
+                            }
+                        />
+                    </small>
+                </label>
+                <input type="hidden" name="stripe_token" />
+                <input type="hidden" name="stripe_email" />
+                <button
+                    type="submit"
+                    className="button cta primary"
+                    disabled={subscribeButtonEnabled ? '' : 'disabled'}
```
peterbe: ditto
```suggestion
                    disabled={subscribeButtonEnabled}
```

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,107 @@
...
+     */
+    const submit = event => {
+        event.preventDefault();
+
+        const subscriptionForm = subscriptionFormRef.current;
+        if (subscriptionForm) {
+            const formData = new FormData(subscriptionForm);
+            setSubscribeFormDisabled(true);
+
+            const stripeHandler = window.StripeCheckout.configure({
```
peterbe: This assumes that the stripe SDK .js has fully loaded. Is that safe?

peterbe:
This is what it looks like at this screen size:
Screen Shot 2020-03-24 at 3 28 59 PM

If I make it a couple pixels narrower this happens:
Screen Shot 2020-03-24 at 3 30 38 PM

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,107 @@
...
+                    <small>
+                        <Interpolated
+                            id={gettext(
+                                'By clicking this button, I authorize Mozilla to charge this payment method each month, according to the <paymentTermsLink />, until I cancel my subscription.'
+                            )}
+                            paymentTermsLink={
+                                <a
+                                    target="_blank"
+                                    rel="noopener noreferrer"
+                                    href={`/${locale}/payments/`}
```
peterbe: Shouldn't that be 
```suggestion
                                    href={`/${locale}/payments/terms/`}
```
Which is still a Wiki only page.

kuma/payments/jinja2/payments/index.html
```diff
@@ -14,3 +14,7 @@
 {% block content %}{% endblock %}
 
 {% block footer %}{% endblock %}
+
+{% block js %}
+<script src="https://checkout.stripe.com/checkout.js" defer></script>
```
peterbe: Pardon my ignorance but won't this happen:
```html
<script defer type="text/javascript" src="/static/react.js" charset="utf-8"></script>
<script src="https://checkout.stripe.com/checkout.js" defer></script>
```
If we can just use Jinja and the `js` block in a way so that it's the other way around then there's no chance that the React `useEffect` happens after the checkout SDK is loaded. 

The bestest solution would be to embed the SDK with something like this: https://github.com/stripe/stripe-js#installation but that's bound to bloat the bundle for all other pages :(

kuma/javascript/src/payments/subscription-form.jsx
Gregoor: Yup, we can do it like that!

schalkneethling:
> This is what it looks like at this screen size:
> Screen Shot 2020-03-24 at 3 28 59 PM
> 
> If I make it a couple pixels narrower this happens:
> Screen Shot 2020-03-24 at 3 30 38 PM

Yup, that is expected. At that breakpoint and smaller the component is hidden.

kuma/javascript/src/payments/subscription-form.jsx
schalkneethling: Could very well be, this is what it is currently on edit profile. @Gregoor thoughts?

kuma/payments/jinja2/payments/index.html
schalkneethling: Yup, I initially looked at a React-based Stripe implementation but yes, bundle size.

schalkneethling: So `index.html` extends `react_base.html` which has the following blocks:

```
{% block site_js %}
  {{ providers_media_js() }}
  {% javascript 'react-main' %}
  {% javascript 'auth-modal' %}
  {% for script in scripts %}
    {% javascript script %}
  {% endfor %}
{% endblock %}

{% block js %}{% endblock %}
```

We are not overriding `site_js` inside `index.html` but we simply add Stripe as additional JS using the `js` block:

```
{% block js %}
<script src="https://checkout.stripe.com/checkout.js" defer></script>
{% endblock %}
```

What we want, and what I believe you are suggesting, is that the load order ends up being:

```
<script src="https://checkout.stripe.com/checkout.js" defer></script>
<script defer type="text/javascript" src="/static/react.js" charset="utf-8"></script>
```

and not the current:

```
<script defer type="text/javascript" src="/static/react.js" charset="utf-8"></script>
<script src="https://checkout.stripe.com/checkout.js" defer></script>
```

I wonder if we can safely refactor the above to something like this:

```
{% block site_js %}
  {{ providers_media_js() }}
  {% javascript 'auth-modal' %}
  {% for script in scripts %}
    {% javascript script %}
  {% endfor %}
{% endblock %}

{% block js %}{% endblock %}

{% block react_js %}
  {% javascript 'react-main' %}
{% endblock %}
```

Thinking out loudly here. No testing or experimentation has been done yet 😄

kuma/javascript/src/payments/subscription-form.jsx
schalkneethling: I do not know whether I would call it safe, but the user will have to have noticed the component, checked the terms checkbox and then click `Continue` :)

With a slow connection, that might not be enough time though.

What do you see as our options? Do we disable the form and perhaps do some kind of polling until `window.StripCheckout` is not undefined? I guess if we do that, we should present some message to the user, or show a spinner or else they will just think the form is broken.

schalkneethling:
> CSRF token is missing in the form submission. You get a 403 when you submit the Stripe modal.
> Perhaps you knew that already :)

Good point! How do we get at a valid `csrf` token inside React? I looked at search in the header and the newsletter form but neither of those seems to use a `csrf` token.

kuma/javascript/src/payments/subscription-form.jsx
schalkneethling: Nope, nope, my bad. Updating.

Gregoor:
This would be the way to go, I think: https://docs.djangoproject.com/en/3.0/ref/csrf/#ajax

Happy to do that in the next PR. We'll also want to create a new view for this. The existing user_edit view is already fairly overloaded.

schalkneethling:
> This would be the way to go, I think: https://docs.djangoproject.com/en/3.0/ref/csrf/#ajax
> 

Ah yes, that should do the trick.

> Happy to do that in the next PR. We'll also want to create a new view for this. The existing user_edit view is already fairly overloaded.

Sounds good to me, thanks Gregor.

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -0,0 +1,107 @@
...
+                name="subscription-form"
+                method="post"
+                onSubmit={submit}
+                disabled={subscribeFormDisabled}
+            >
+                <label className="payment-opt-in">
+                    <input
+                        type="checkbox"
+                        required="required"
+                        onClick={toggleButton}
```
Gregoor: We should go full [controlled component](https://reactjs.org/docs/forms.html#controlled-components), i.e. also set the `<input>`'s `value=` based on `subscribeButtonEnabled`. It is also behaving opposite of what is expected for me, i.e. when checking the box the button is disabled.

Also my renaming advice was a little shortsighted for that one. I'd go with something more meaningful in this context like: `paymentAuthorized`.
Similar nit/advice for the other flag: I'd go with `isSubmitting`.
The upside of these renames is that the state now reflects user intent and not what UI might result from it. Now when we want to make other UI changes based on it, we don't have to rename it.

kuma/payments/jinja2/payments/index.html
Gregoor: Another idea: Load the script dynamically inside the SubscriptionForm component, when it gets mounted (i.e. inside of `useEffect(() => { loadStripe() }, []`). Then you can also track whether stripe is already loaded, if the user manages to get to the stripe modal step before it's ready.

peterbe: Yeah, what @Gregoor said. 

By the way, would you could also have done is drop the `defer` here because then you'd get this:
```
<script defer src="/static/react.js"></script>
<script       src="https://checkout.stripe.com/checkout.js"></script>
```
and then, the browser will load https://checkout.stripe.com/checkout.js before it loads `/static/react.js` but that's not guarantee that it `https://checkout.stripe.com/checkout.js` is ready by the time the `useEffect` triggers deep inside `/static/react.js`.

kuma/javascript/src/payments/subscription-form.jsx
schalkneethling: I had a look at the `checkbox` in order to set a value on it but noticed that this field in the current implementation does not have a `name` attribute:

```
<input type="checkbox" required="required">
```

This means that the data of the field is not sent to the server, no? Looking at the Python for `create_stripe_subscription` it also does not look like the backend is checking for the field or a specific value.

I am guessing we want to change this?

Perhaps something like:

```
<input type="checkbox" name="payment_authorized" required="required">
```

With that, do we want to do validation on the backend if the box was not checked? I mean, technically, someone should not be able to submit the form if they did not check the checkbox.

We also are not storing this anywhere so, should we be?

schalkneethling:
> Another idea: Load the script dynamically inside the SubscriptionForm component, when it gets mounted (i.e. inside of useEffect(() => { loadStripe() }, []). Then you can also track whether stripe is already loaded, if the user manages to get to the stripe modal step before it's ready.

Could we delay it even further and only load this once the user checks the checkbox?

schalkneethling:
> Happy to do that in the next PR. We'll also want to create a new view for this. The existing user_edit view is already fairly overloaded.

Agreed. Seeing that the same functionality is going to continue to live on the edit profile page, we should probably create this view so that both these instances can use the same backend code.

kuma/javascript/src/payments/index.jsx
```diff
@@ -28,15 +33,21 @@ export default function PaymentsLandingPage() {
         terms: `/${locale}/payments/terms`
     };
 
+    const showSubscriptionForm =
+        userData && userData.waffle.flags.subscription_banner;
```
peterbe: Why `subscription_banner` and not `subscription`?

PR #6760, opened 2020-03-27
allow all for robots on prod wiki

### LINKED ISSUE #6340: T - Update robots.tx (Warnings have spiked on Search Console)

escattone:
Fixes #6340 

This PR opens up all of the endpoints on the production wiki domain (https://wiki.developer.mozilla.org) to robots. We already put `noindex, nofollow` in place for all of the wiki endpoints via #6502.

### DIFF
--- a/kuma/landing/tests/test_views.py
+++ b/kuma/landing/tests/test_views.py
@@ -101,6 +101,23 @@ def test_robots_allowed_main_website(client, settings):
     assert b"Disallow: /admin/\n" in content
 
 
+def test_robots_all_allowed_wiki_website(client, settings):
+    """On the wiki website, allow robots with NO restrictions."""
+    host = "main.mdn.moz.works"
+    wiki_host = "wiki." + host
+    settings.WIKI_HOST = wiki_host
+    settings.ALLOWED_HOSTS = [host, wiki_host]
+    settings.ALLOW_ROBOTS_WEB_DOMAINS = [host, wiki_host]
+    response = client.get(reverse("robots_txt"), HTTP_HOST=wiki_host)
+    assert response.status_code == 200
+    assert_shared_cache_header(response)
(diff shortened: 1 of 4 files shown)


peterbe:
@escattone @Gregoor I don't think it's worth our combined time for me to try to understand what's going on here. If you really think my review is of importance, that's fine, and I'll do that then. But for me to fully understand what went wrong, what we need, how this stuff works, I'd need to spend a good chunk of time on it and it'd have to come later. 

So, let me know or decide to own this between just the two of you.

escattone:
Thanks @peterbe, you have more than enough on your plate, so don't worry about this one!

PR #6767, opened 2020-03-30
Privacy checkbox hidden with JS disabled

### LINKED ISSUE #5977: [Newsletter] Privacy checkbox hidden when JS disabled

schalkneethling:
When JavaScript is disabled or failed to execute the privacy checkbox field is hidden. This makes submitting the form impossible as this is a required field, but is not accessible via the UI.

The pull requests address the issue:

### With JS disabled

Screenshot 2020-03-30 at 14 43 10

### With JS enabled

Screenshot 2020-03-30 at 14 44 03

### With JS disabled

Screenshot 2020-03-30 at 14 39 58

### With JS enabled

Screenshot 2020-03-30 at 14 40 44

@mindy r?

fixes #5977

### DIFF
--- a/kuma/javascript/src/newsletter.jsx
+++ b/kuma/javascript/src/newsletter.jsx
@@ -44,7 +44,7 @@ export default function Newsletter() {
     const newsletterFormat = 'H';
 
     const [showNewsletter, setShowNewsletter] = useState(true);
-    const [showPrivacyCheckbox, setShowPrivacyCheckbox] = useState(false);
+    const [showPrivacyCheckbox, setShowPrivacyCheckbox] = useState(true);
     const [errors, setError] = useState([]);
     const [
         showSuccessfulSubscription,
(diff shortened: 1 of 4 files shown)


kuma/javascript/src/newsletter.jsx
```diff
@@ -145,6 +145,9 @@ export default function Newsletter() {
         if (disableNewsletter) {
             setShowNewsletter(false);
         }
+        // if JS is enabled client-side, hide the privacy checkbox until
+        // the email input field receive focus
+        setShowPrivacyCheckbox(false);
```
peterbe: Beyond the checkbox, does signing up work at all with JS disabled?

schalkneethling: Technically yes, as it will post to `https://www.mozilla.org/en-US/newsletter/`

PR #6768, opened 2020-03-30
subscriber number

### LINKED ISSUE #6731: T - Provide subscribers with a static customer number that is incremented upon each new subscription.
**Description**

Upon each new subscription signup we need to store a customer number that increments +1 over the previous customer's (most-recent) number. 

This number needs to be static (and not dynamically computed based on a count of customers) for a few reasons:
- the customer should be able to rely on their number as her own.
- X customer's unsubscribe or deletion should not affect other customer Y's number. 
- if we ever gave rewards/prizes for customer no. 100, we would not want the number to be dynamic, as that might disastrous

**Acceptance Criteria**
- [ ] upon customer subscription, a number is calculated and stored in our database
- [ ] this customer number is immutable and unaffected by other customers' subscription cancellations or acct deletions

peterbe:
Fixes #6731


peterbe:
The reason I didn't add it to `/api/v1/whoami` is because it'll be easy to when it's time to actually use it. Also, we have some interesting PRs in mid-air that would conflict on this anyway.

kuma/users/migrations/0018_user_subscriber_number.py
```diff
@@ -0,0 +1,30 @@
...
+
+    dependencies = [
+        ("users", "0017_usersubscription"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="subscriber_number",
+            field=models.PositiveIntegerField(blank=True, null=True),
```
Gregoor: Alternative idea, make this a blank auto increment field, and then set it to its default value, which would simplify the updating code (aka outsource it to MySQLs data structures).

Tbh it's probably 100x eaiser to have this idea than to ORM-wrangle Django into making it work, so feel free to ignore it.

peterbe: I don't know what a "blank auto increment field" is. 
I'm hesitant to step outside the ORM if it's possible. My sticking to the ORM API you become more independent of the database vendor. 
Auto-incrementing things and stuff like sequences in PostgreSQL are generally scary because of transactions. If you manually maintain it instead you can make sure it doesn't skip.

Gregoor: Wrt db vendor independentness: I haven't seen a single DB technology switch in later stage products yet, so I consider that ORM feature to be over-sold. But I also share your hope that MDN will someday run on PostgreSQL, so I'll take that point.

Wrt transaction failure: I don't foresee how 
```sql
UPDATE auth_users SET subscriber_no = NEXTVAL('subscriber_numbers') WHERE id = ?
```
could fail in a way that would bump the sequence. I.e. we wouldn't even need a transaction for use-case.

Though it probably works differently (for the worse 🙊) in MySQL which might make this more pain than gain indeed.

peterbe: I thought `NEXTVAL` was a Postgres-only thing since Postgres uses sequences as an underlying thing to generate primary key IDs.
My first iteration was to use something like this:
```python
    cursor.execute("""
        UPDATE auth_user 
        SET subscriber_number = (SELECT MAX(subscriber_number) FROM auth_user) 
        WHERE id=%s
    """, [user.id])
```
But you're not allowed to do edits and selects on the same table in MySQL.

peterbe: Hmm... 
```
django.db.utils.OperationalError: (1305, 'FUNCTION test_developer_mozilla_org.NEXTVAL does not exist')
```

Gregoor: Oh wow, I wasn't aware of that limitation in MySQL.

And I also just learned that "Incorrect table definition; there can be only one auto column". So there's literally no other way to do it, than the way you did do it.

I just learned something. It's mostly that I miss postgres.

Gregoor: And yup I was thining of PG when I said `NEXTVAL`, thought we were in a similar mindspace since you mentioned PG sequences before.

peterbe: I'm not entirely sure it's worth going into the trouble, considering how rarely people will become subscribers. So to have two people sign up within the same millisecond is sorry unlikely. 
But, before merging this, I need to think about the use of locking. There's no point locking the user's row because that's not where the risk lies. It's about the relationship to _other_ users. 
So, technically you should lock the whole _table_ but that's silly because why need with it for the other 99.9% when they don't care about this number. 
No, the best solution has to be to use a global lock specifically for this.

peterbe:
@Gregoor I changed my mind. See discussion above. Basically, it didn't make sense to "lock" the user row because it actually depends all other users. However, it's not sensible to [lock the whole table](https://dev.mysql.com/doc/refman/5.6/en/lock-tables.html), which I don't even know if it's feasible. 
For a simple global locking technique you can either use `django-redis` or [`django-mysql`](https://django-mysql.readthedocs.io/en/latest/locks.html) which I'm less familiar with. 
Yes, it's unlikely that we can upgrade to Postgres but I trust Redis more than I trust MySQL and I 'm well aware that `django-redis` lockin works really well.

Gregoor:
Great, this reads better to me as well, since I'm a little confused by isolation levels in MySQL. :+1:

peterbe:
> I'm (still) a little confused by isolation levels in MySQL.

Tell me about it! Isolation level stuff in Postgres is hard enough. 

What might be a better option is to make the `subscriber_number` a unique field. Then you'd be able to do things like perhaps do something equally hacky like this:
```python
while True:
    highest_number = User.objects.filter(subscriber_number__isnull=False).aggregate(...)
    try:
        User.objects.filter(id=self.id).update(subscriber_number=highest_number + 1)
        break
    except ProgrammingError as exception:
       if not is_duplicate_subscriber_number_error(exception):
           raise
```

But first of all, can you combine `unique=True, null=True` on a model field in Django ORM??
I think what we've got so far is safe under all the circumstances.

peterbe:
@Gregoor As of [the last commit](https://github.com/mdn/kuma/pull/6768/commits/d8316f93f9fd7380d44aed7685d23560305a8156) here's what happens now when you do 
```python
UserSubscription.set_active(user, subscription_id)
```

### Before
```
SELECT `users_usersubscription`.`id`, `users_usersubscription`.`user_id`, `users_usersubscription`.`stripe_subscription_id`, `users_usersubscription`.`canceled`
UPDATE `users_usersubscription` SET `user_id` = 848, `stripe_subscription_id` = 'sub_H0okT3ahzkjqK2', `canceled` = NULL, `created` = '2020-03-31 09:38:47.941686
SELECT `auth_user`.`id`, `auth_user`.`password`, `auth_user`.`last_login`, `auth_user`.`is_superuser`, `auth_user`.`username`, `auth_user`.`first_name`, `auth_u
SELECT MAX(`auth_user`.`subscriber_number`) AS `number` FROM `auth_user` WHERE `auth_user`.`subscriber_number` IS NULL
UPDATE `auth_user` SET `subscriber_number` = 1 WHERE `auth_user`.`id` = 848
```

### After
```
SAVEPOINT `s140451756050240_x1`
SELECT `users_usersubscription`.`id`, `users_usersubscription`.`user_id`, `users_usersubscription`.`stripe_subscription_id`, `users_usersubscription`.`canceled`
UPDATE `users_usersubscription` SET `user_id` = 848, `stripe_subscription_id` = 'sub_H0okT3ahzkjqK2', `canceled` = NULL, `created` = '2020-03-31 09:38:47.941686
SELECT `auth_user`.`id`, `auth_user`.`password`, `auth_user`.`last_login`, `auth_user`.`is_superuser`, `auth_user`.`username`, `auth_user`.`first_name`, `auth_u
SELECT MAX(`auth_user`.`subscriber_number`) AS `number` FROM `auth_user` WHERE `auth_user`.`subscriber_number` IS NULL
UPDATE `auth_user` SET `subscriber_number` = 1 WHERE `auth_user`.`id` = 848
RELEASE SAVEPOINT `s140451756050240_x1`
```

So I'll merge this now if tests pass. Ok?

Gregoor:
Sounds good 👌

Issue #6759, opened 2020-03-27
U - Back-end telemetry in GA creates bogus traffic


atopal:
**Summary**
The event "auth-started" has gone up 3600% in the last two weeks, all other events have been below 15%+. It's already problematic that back-end generated events don't send a client ID to correlate with front-end usage, but as long as the numbers are low it doesn't skew the numbers much. In this case though we are counting an additional 250k uses per week that don't actually exist and it skews the numbers.

**Steps To Reproduce (STR)**
1. check numbers on GA. [Direct Link](https://analytics.google.com/analytics/web/#/report/content-event-events/a36116321w64965862p66726481/_u.date00=20200322&_u.date01=20200326&_u.date10=20200308&_u.date11=20200312&_.useg=user1bmFc6fPS0SM8DRRHnP3NA&_r.drilldown=analytics.eventCategory:signup-flow)

**Actual behavior**
A large number of bogus events was logged (maybe from bots?)

**Expected behavior**
Every event should send a client ID, where possible, additionally log-in links should have a [no-follow attribute](https://support.google.com/webmasters/answer/96569?hl=en).

tobinmori:
Every event without a GA client Id, counts as a new user. 

Solutions?
- Inject client_id into links?
- new task: drop event (for now), p1, 0.5
- next sprint: next task: look at mitigation

peterbe:
Another "mitigation" is to check the HTTP referer. If you use `curl http://localhost.org:8000/users/github/login/` it means the `request.META.get('HTTP_REFERER')` becomes `None`. 
If you actually click the GitHub/Google button in a browser the `HTTP_REFERER` becomes a URL whose hostname is the same as ours. 

Also, I tested that if I click to open the modal but then either ⌘-click or if I right-click and select "Open link in a new tab" the HTTP_REFERER gets set. 

I'll make a PR to demonstrate.

atopal:
Can we please just comment out the event collection for this already? It's massively messing up our numbers.

atopal:
Nevermind, it's in production already, so I should hopefully see the effects tomorrow.

chinikes:
@tobinmori , @atopal - Wanted to confirm - what epic is this user story part of?

peterbe:
It *is* already an epic. 
[...]

PR #6770, opened 2020-03-30
comment out auth-started GA event for now

peterbe:
Part of #6759

I hope the code comment is sufficient to explain the rationale of this PR. 
Note that this doesn't close the issue.

### DIFF
--- a/kuma/users/providers/github/views.py
+++ b/kuma/users/providers/github/views.py
@@ -1,3 +1,5 @@
+from urllib.parse import urlparse
+
 from allauth.account.utils import get_next_redirect_url
 from allauth.socialaccount.providers.github.views import GitHubOAuth2Adapter
 from allauth.socialaccount.providers.oauth2.views import (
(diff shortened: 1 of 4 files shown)


atopal:
Hey Peter, if I understand correctly, this doesn't actually do what the title says? I'd much rather turn event collection off for now and then turn it on again with client ID if we can, and in that case I don't much about bots, because GA will discount them anyway. The referrer solution would still count every event sent as an extra user on MDN.

peterbe:
@atopal I updated the PR title. 
We can do another one where we actually turn it off. I guess you haven't had a chance yet to evaluate if this solved the problem of "rogue events".

atopal:
Data just came in for April 1st. Down to 1k per day. Thanks Peter!

It still distorts the data, since without the client ID every event is counted as a separate users, but for the log in flow that should not be more than 30k "users" per month. We can live with that.

peterbe:
> Data just came in for April 1st. Down to 1k per day. Thanks Peter!
> 
> It still distorts the data, since without the client ID every event is counted as a separate user, but for the log in flow that should not be more than 30k "users" per month. We can live with that.

@tobinmori @atopal So what does this mean? Shall we bother removing the instrumentation entirely? ...until we have that thing that puts the GA clientID into the web request?

atopal:
No, it's fine, we can live with that for now and can decide later whether we still want to implement the client ID solution. If numbers stay this way it's just not enough noise to matter.

PR #6773, opened 2020-03-31
Update payment received mail copy & attachment

### LINKED ISSUE #6743: T - Update copy on text email receipt and add receipt attachment.
https://docs.google.com/document/d/1Gx-2cniZlWTPhsbYBNJj7n36lmnvPZ-vNEFOoAA5ksQ/edit?usp=sharing

Gregoor:
Fixes #6743

Changes the copy of the email and adds the invoice as an attachment


Gregoor:
Two tests are still failing, apparently due to mocking not doing its thing. @peterbe I think I would benefit from your experienced eye on this, when you have the time.

kuma/users/tasks.py
```diff
@@ -68,19 +73,37 @@ def send_welcome_email(user_pk, locale):
...
         plain = render_email("users/email/payment_received/plain.ltxt", context)
-        send_mail_retrying(subject, plain, settings.DEFAULT_FROM_EMAIL, [user.email])
+        send_mail_retrying(
+            subject,
+            plain,
+            settings.DEFAULT_FROM_EMAIL,
+            [user.email],
+            attachment={
+                "name": os.path.basename(urlparse(payment_intent.invoice_pdf).path),
+                "bytes": requests.get(payment_intent.invoice_pdf).content,
```
peterbe: That's very optimistic :)

peterbe:
> Two tests are still failing, apparently due to mocking not doing its thing. @peterbe I think I would benefit from your experienced eye on this, when you have the time.

Because of the python linting? 
Or the whole PR in general?

peterbe:
By the way, when we built Mozilla Symbol Server, we did it this way. It's not really a Webhook but Mozilla's TaskCluster has a myriad of "tasks" that it needs to do and if any of them fail, that build of Firefox gets flagged as a failed build and alarms go off. One of the tasks it needed to do was to upload a 100MB-2GB .zip file full of debug symbols to https://symbols.mozilla.org. It's basically just doing this:
```python
attempts = 0
while True:
    r = requests.post('https://symbols.mozilla.org/api/upload', files=zip_file)
    if r.ok:
         break
    attempts += 1
    if attempts > MAX_ATTEMPTS:
        raise Hell("Break the build!")
```
Yes, it takes a long time but TaskCluster is concurrency built-in. But at least you can know for certain that all is taken care of and you don't get Firefox builds shipped to users for which there are all debug symbols stored somewhere.

Gregoor:
Thanks for the detailed review (as always), Peter. I'll go through the comments one-by-one in a bit, but first a question wrt moving it to the view (and my ignorance might show here): Don't we risk clogging up the web workers in the (unlikely) case multiple invoices are sent out at once?

peterbe:
> Don't we risk clogging up the web workers in the (unlikely) case multiple invoices are sent out at once, given the queue-less blocking nature of the beast?

Yes and no. We have Gunicorn in front which works like a queue system too with WSGI. If all threads and processes are busy, Gunicorn will hold off the work until the next "web worker" is ready, willing, and able. 
And if Gunicorn is so incomprehensibly busy that all the web workers are failing, it will bail with a `SystemExit` which means that whatever we have in front of Gunicorn will turn that into a 5xx error. 
This is what happened when we had EFS instead of S3 to serve up file attachment payloads.

Gregoor:
Right, and we expect the traffic for this to be so miniscule in comparison that we wouldn't run into the problems we ran into with EFS?

peterbe:
> Right, and we expect the traffic for this to be so miniscule in comparison that we wouldn't run into the problems we ran into with EFS?

True and most importantly is that we're changing how we do things. We cram to do as much as possibly in the "Gunicorn space" because we know that Stripe will be fully responsible for the retrying functionality. 

Alternatively, we'd implement our own retry-queue functionality. And that's not trivial.

Gregoor:
Though even with stripe's great retry mechanism the thread still wouldn't be released, it'd be blocked [sleeping](https://github.com/stripe/stripe-python/blob/8629add6274c6058eed14c27a9ed1b86999ee22f/stripe/http_client.py#L144). Which I'm gonna assume is fine as well, since it'd happen rarely and we'd always have enough active threads to handle the load.

I'll go ahead with changing the code. Thanks Peter!

kuma/users/jinja2/users/email/payment_received/plain.ltxt
```diff
@@ -1,10 +1,20 @@
...
+Thank you.
 
-{{ invoice_pdf }}
+You supported MDN with a ${{ dollar_amount }} monthly subscription.
+
+Amount paid: ${{ dollar_amount }}.00
+Payment method: {{ credit_card_brand }}
+
+Invoice number: {{ invoice_number }}
+Date of issue: {{ payment_date }}
```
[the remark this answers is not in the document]
Gregoor: I went with that initially but kept getting a syntax error and abandoned it. Just looked at it again and realized that I tried django templating filter syntax, instead of jinja's 🙈

Gregoor:
(I marked this ready for review so that CI would run, but apparently that didn't do the trick yet)

peterbe:
> (I marked this ready for review so that CI would run, but apparently that didn't do the trick yet)

CI will run on draft PRs too. If it's not starting it's because it gets stuck in TravisCI.

peterbe:
Let's try again.

kuma/users/views.py
```diff
@@ -935,3 +936,35 @@ def stripe_hooks(request):
...
+        subject = "".join(subject.splitlines())
+        plain = render_email("users/email/payment_received/plain.ltxt", context)
+        send_mail_retrying(
+            subject,
+            plain,
+            settings.DEFAULT_FROM_EMAIL,
+            [user.email],
+            attachment={
+                "name": os.path.basename(urlparse(payment_intent.invoice_pdf).path),
+                "bytes": requests.get(payment_intent.invoice_pdf).content,
```
peterbe, line 967: Perhaps use `requests_retry_session().get(` instead?

kuma/users/views.py
```diff
@@ -945,7 +945,7 @@ def _send_payment_received_email(payment_intent, locale):
         "payment_date": datetime.fromtimestamp(payment_intent.created),
         "next_payment_date": subscription_info["next_payment_at"],
         "invoice_number": payment_intent.id,
-        "dollar_amount": settings.CONTRIBUTION_AMOUNT_USD,
+        "cost": settings.CONTRIBUTION_AMOUNT_USD,
```
Gregoor, line 948: @peterbe I renamed it after all, given that we now have a formatting function that could work with other locales, I'd find the naming misleading. Since the english original doesn't contain the currency symbol, is it too hopeful to assume that the localizers don't introduce one themselves?

kuma/core/templatetags/jinja_helpers.py
```diff
@@ -187,3 +188,8 @@ def get_locale_localized(locale):
     language = settings.LOCALES[locale].english
 
     return _(language)
+
+
+@library.filter
+def money(value, currency="USD"):
+    return format_currency(value, currency)
```
Gregoor, line 195: So this doesn't respect the used locale yet. But it could be extended to do that, once we go multi-national!

peterbe: One thing at a time I guess. Having this here is powerful because it'll be just one single point for ruling all formatting. 
That `format_currency` does take a `locale` option but we can worry about that some other day.

kuma/users/tests/test_stripe_subscription.py
```diff
@@ -157,7 +160,7 @@ def test_stripe_payment_succeeded_sends_invoice_mail(mock1, client):
     payment_email = mail.outbox[0]
     assert payment_email.to == [testuser.email]
     assert "manage monthly subscriptions" in payment_email.body
-    assert "invoice" in payment_email.subject
+    assert "Receipt" in payment_email.subject
```
peterbe, line 164: I think you should check for some of the language in the email too. 
For example 
```python
assert 'Invoice number test_invoice_001' in payment_email.body
assert 'You supported MDN with a $5.00 monthly subscription' in payment_email.body
```

kuma/users/views.py
```diff
@@ -935,3 +935,37 @@ def stripe_hooks(request):
...
+
+
+def _send_payment_received_email(payment_intent, locale):
+    user = get_user_model().objects.get(stripe_customer_id=payment_intent.customer)
+    subscription_info = _retrieve_and_synchronize_subscription_info(user)
+    locale = locale or settings.WIKI_DEFAULT_LANGUAGE
+    context = {
+        "payment_date": datetime.fromtimestamp(payment_intent.created),
+        "next_payment_date": subscription_info["next_payment_at"],
+        "invoice_number": payment_intent.id,
```
peterbe, line 947: This is incorrect. That ID is a Stripe ID. The correct Invoice number is this:
Screen Shot 2020-04-03 at 9 56 30 AM

You can see it on the PDF too:
Screen Shot 2020-04-03 at 9 57 05 AM

malqinneh:
@peterbe those are good suggestions but pretty sure we use some instances of the current wording in the user profile as well. I think it's fine as is (albeit _potentially_ awkward). Also, think @Gregoor said we can only link to the invoice (not attach)? I think either or is enough. The email itself is structured as a receipt.

peterbe:
> we can only link to the invoice (not attach)? I think either or is enough. The email itself is structured as a receipt.

Not sure what you mean. We're actually NOT linking to the invoice. We are attaching it into the email actually. 
But I was suggesting we include a link to the (web view version) of the invoice *too*. So you can choose: Download & save the attached PDF, or go to the web to read it.

atopal:
@peterbe we'll do that possibly in a follow up when we do HTML emails. The issue is that this is a text email, and the link would be a long string of random letters and numbers. Something that looks very off.

kuma/users/views.py
```diff
@@ -935,3 +935,37 @@ def stripe_hooks(request):
...
+        subject = "".join(subject.splitlines())
+        plain = render_email("users/email/payment_received/plain.ltxt", context)
+        send_mail_retrying(
+            subject,
+            plain,
+            settings.DEFAULT_FROM_EMAIL,
+            [user.email],
+            attachment={
+                "name": os.path.basename(urlparse(payment_intent.invoice_pdf).path),
+                "bytes": requests_retry_session()
```
peterbe, line 966: nit: Can you move this to *above* the `send_mail_retrying` otherwise, if the SMTP retrying actually happens, you'll be downloading the invoice over and over. 

Also, although it's unlikely, I think you should always assert something about HTTP responses. Because they can fail in two completely different ways `RequestError` and `not response.ok`.

So, perhaps something like this:

```python
download = requests_retry_session().get(payment_intent.invoice_pdf)
download.raise_for_status()
send_mail_retrying(
   ...,
   attachment={
       "bytes": download.content,
        ...
  }
```

kuma/users/tests/test_stripe_subscription.py
```diff
@@ -156,8 +159,10 @@ def test_stripe_payment_succeeded_sends_invoice_mail(mock1, client):
     assert len(mail.outbox) == 1
     payment_email = mail.outbox[0]
     assert payment_email.to == [testuser.email]
+    assert "Receipt" in payment_email.subject
+    assert "Invoice number: test_invoice_001" in payment_email.body
+    assert "You supported MDN with a $4.99 monthly subscription" in payment_email.body
```
peterbe, line 164: wonderful!

kuma/users/views.py
Gregoor: Great point wrt raising the error, I changed it accordingly.

Wrt retrying-triggering-multiple-downloads, I don't think that would've happened, as the bytes were read before it's being passed to the mail call, right?

peterbe: > I don't think that would've happened, as the bytes were read before it's being passed to the mail call, right?

You are absolutely right! It's not a function, it's the result of the function.

PR #6775, opened 2020-03-31
fix all outbound links when Tracking Protection is off

### LINKED ISSUE #6758: "Edit in wiki" link does not work if Firefox "Do Not Track" signal is enabled
**Summary**
_What is the problem?_

Cannot click "Edit in wiki" link directly. cmd+click or right click works.

**Steps To Reproduce (STR)**

1. Use Firefox.
1. Visit about:preferences#privacy
1. Enable "Do Not Track"
1. Visit https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/Planned_changes.
1. Ensure you're logged in.
1. Click "Edit in wiki"

**Actual behavior**
_What actually happened?_

Nothing.

**Expected behavior**

Go the wiki.

**Additional context**

The amount of code executed for this click seems over the top.

peterbe:
Fixes #6758

Before checking out this branch; go to the home page. Change your Firefox to disable tracking protection. I.e. clicking the little toggle so it looks like this:
Screen Shot 2020-03-31 at 1 27 45 PM
(doing that will reload the page)
Now, try to click any link to a Hacks post on the home page.

### DIFF
--- a/kuma/static/js/analytics.js
+++ b/kuma/static/js/analytics.js
@@ -41,10 +41,6 @@
             if(ga && ga.create) {
                 // Send event to GA
                 ga('send', data);
-                // execute callback now
-                if(callback) {
-                    callback();
-                }
             }
             else if(ga && !ga.create) {
                 // GA blocked or not yet initialized
@@ -52,14 +48,9 @@
                 data.hitCallback = null;
                 // add to queue without callback
                 ga('send', data);
-                // execute callback now
-                if(callback) {
-                    callback();
-                }
             }
-            else if(callback) {
-                // GA disabled or blocked or something, make sure we still
-                // call the caller's callback:
+
+            if(callback) {
                 callback();
             }
         },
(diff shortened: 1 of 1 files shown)


kuma/static/js/analytics.js
```diff
@@ -41,6 +41,10 @@
             if(ga && ga.create) {
                 // Send event to GA
                 ga('send', data);
+                // execute callback now
+                if(callback) {
```
Gregoor, line 45: given that all branches of the outer conditional call the callback, I'd recommend reverting this, removing L56-58 and removing the else from 60

kuma/static/js/analytics.js
```diff
@@ -48,14 +48,9 @@
...
                 ga('send', data);
-                // execute callback now
-                if(callback) {
-                    callback();
-                }
             }
-            else if(callback) {
-                // GA disabled or blocked or something, make sure we still
-                // call the caller's callback:
+
```
peterbe, line 52: I put in an extra blank line to make extra separated.

Issue #6557, opened 2020-02-21
U - As a potential subscriber I want the option to log in during the payment process, so I don't have to navigate the page for that


atopal:
# Acceptance Criteria
- [x] not-signed-in users who click "Learn more" on the banner are taken to the FAQ page.
- [x] not-signed-in users who click "continue" on the FAQ page are shown the sign-in overlay with specific text about supporting MDN.
- [x] After they are authenticated, users are taken to the FAQ page.
- [x] The banner is behind a waffle flag, so we can turn it on and off for non-logged in users as necessary.

# Context
The goal is to give people who weren't signed in at the time they clicked the "subscribe" button an option to sign in without taking them out of the payment flow.

# Design

- Signed in user:
 https://share.goabstract.com/857f8a74-8fdd-41e6-9d78-d5c838d4054a?mode=design
- **Non-signed in users (but with an account):** 
 https://share.goabstract.com/857f8a74-8fdd-41e6-9d78-d5c838d4054a?collectionLayerId=1c005660-47c6-4d98-aac5-bc533fa9ae26&mode=design
- Non-signed in users (without an account): 
https://share.goabstract.com/857f8a74-8fdd-41e6-9d78-d5c838d4054a?collectionLayerId=0083a875-bdc1-49e2-9215-9a0c59a4cab6&mode=design
- And then the specific use-case where you have the banner take you to the user profile/settings page(?): https://share.goabstract.com/80aff1b8-a8da-4250-ab33-f6426d45b2e1?mode=design

tobinmori:
Some questions for @atopal @malqinneh :

**Re: "Registered MDN User, logged in/Non-MDN user, not logged-in" modal**

- What specific event(s) trigger this modal? 
- The messaging is a bit confusing here in that these sound more like a statements of one's logged-in status, rather than a CTA
- Where does the user land upon clicking either option?

atopal:
@tobinmori not sure what modal you mean? There is no modal in this flow. This is for when a non-logged in users clicks "subscribe" on the payments banner.

chinikes:
@schalkneethling , @atopal , @tobinmori - This story has been marked done on the spreadsheet. Can you confirm the ACs have been met? Thanks!

chinikes:
Hi @tobinmori - this story doesn't have any tasks set up to implement it although it's in the current sprint. Is the Team considering taking it out of the sprint / rolling over to Delta? If so we'll need the story "tasked out" in time before delta's planning session

[...]

PR #6780, opened 2020-04-01
Update subscription banner copy and link

### LINKED ISSUE #6772: T - Update subscription banner copy
The banner currently in production has the following copy:

Become a monthly supporter
Support MDN with a $5 monthly subscription <Learn more. />
[ Subscribe ]

We need to update this to the following

Become a monthly supporter
Support MDN with a $5 monthly subscription.
[ Learn more ]

schalkneethling:
Updates copy for the subscription banner. Also updates the link to point to the subscription landing page.

@Gregoor r?

fixes #6772

### DIFF
--- a/kuma/javascript/src/active-banner.jsx
+++ b/kuma/javascript/src/active-banner.jsx
@@ -52,7 +52,7 @@ import * as React from 'react';
 import { useContext, useState } from 'react';
 
 import CloseIcon from './icons/close.svg';
-import { getLocale, gettext, Interpolated } from './l10n.js';
+import { getLocale, gettext } from './l10n.js';
 import UserProvider from './user-provider.jsx';
 
 // Set a localStorage key with a timestamp the specified number of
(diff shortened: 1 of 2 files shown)


kuma/javascript/src/active-banner.jsx
```diff
@@ -232,23 +232,11 @@ export default function ActiveBanner() {
...
-                                }
-                            />
-                        }
-                        cta={gettext('Subscribe')}
-                        url={`${
-                            window.mdn ? window.mdn.wikiSiteUrl : ''
-                        }/${locale}/profiles/${userData.username ||
-                            ''}/edit#subscription`}
+                        copy={gettext(
+                            'Support MDN with a $5 monthly subscription'
```
peterbe, line 236: @schalkneethling Since you're going to have to merge conflict update this anyway, do you think you can squeeze in a jsx code comment that mentions `https://github.com/mdn/kuma/issues/6654` here?

peterbe:
Pardon my ignorance but why do we still have 
```
if (!userData.isAuthenticated) {
     return null;
}
```

Isn't it enough to depend on the waffle flag? Now that we don't depend on every visitor having an Edit Profile, what's the point of this additional check? 

By the way, I'm working on https://github.com/mdn/kuma/issues/6557 which will make the payment form, in the subscription landing page, "work" for anonymous too. So should that work involve lifting this additional check?

schalkneethling:
> By the way, I'm working on #6557 which will make the payment form, in the subscription landing page, "work" for anonymous too. So should that work involve lifting this additional check?

Yup, makes sense to handle that there. Thanks Peter

schalkneethling:
The requested change was because I still had Prettier 1.x installed when doing a commit and it overwrites the formatting of a bunch of 2.x formatted files. This has been reverted so, merging :)

PR #6789, opened 2020-04-01
display your future and current subscriber_number

### LINKED ISSUE #6733: T - Display (auto-incremented) customer number to subscribers
This issue is dependent upon the completion of: #6731

**Acceptance Criteria:**

[See user story for design and AC. ](https://github.com/mdn/kuma/issues/6732)

peterbe:
Fixes #6733

### What a non-subscriber would see on the Subscription landing page
Screen Shot 2020-04-02 at 3 23 09 PM
It says "2" because user `peterbe` has already snagged the number 1. 
You can only make a screenshot like this if you relax the waffle flag to all users which is a bit contrived at the moment. 

### What a non-subscriber would see on the Edit Profile page
Screen Shot 2020-04-02 at 3 34 09 PM

### What a subscriber sees would see on the Edit Profile page
Screen Shot 2020-04-03 at 6 48 14 AM


peterbe:
@mindy (or @schalkneethling ) can you help me style the `description` in the `SubHeader` component. Here's what it looks like as of this draft:
Screen Shot 2020-04-01 at 3 26 17 PM

Here's what it should look like:
Screen Shot 2020-04-01 at 3 14 06 PM

peterbe:
ACTUALLY! Hang on. It's actually not a `<p>` tag under the `<h1>`. It's probably a new thing. 
Based on this:
 
Screen Shot 2020-04-01 at 3 32 56 PM

...which is how it'll look on the `/$locale/payments/` page.

peterbe:
@mindy Silly me! The `SubHeader` component already accepts a `subtitle`. Here's how it looks like. 
Screen Shot 2020-04-01 at 4 01 41 PM
Screen Shot 2020-04-01 at 4 01 49 PM

Looks good?

mindy:
It's almost there, it just looks a little too big. I think if you reduce the font-size to 24px then you're good.

peterbe:
> It's almost there, it just looks a little too big. I think if you reduce the font-size to 24px then you're good.

Would you mind making a "counter PR" against my branch?

Also, what's with the massive gap between the text on the left and the subscription form on the right?

mindy:
> Would you mind making a "counter PR" against my branch?
> 
> Also, what's with the massive gap between the text on the left and the subscription form on the right?

Can do! I'll look into the gap as well.

peterbe:
@malqinneh @atopal Look at the screenshots above in the PR description. 
In particular the sentence on the Edit Profile for subscribers. 
It reads weirdly. That sentence doesn't sound right.

malqinneh:
> That sentence doesn't sound right.

Ah, should be "You are MDN member number [x]." We updated those in the designs - thank you for catching Peter!

kuma/payments/jinja2/payments/index.html
```diff
@@ -8,7 +8,10 @@
         as an argument to indicate that we have the data we need, which
         allows the component to render in the server.
     #}
-    {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(), {"email": settings.CONTRIBUTION_SUPPORT_EMAIL})|safe }}
+    {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(), {
+        "email": settings.CONTRIBUTION_SUPPORT_EMAIL,
+        "next_subscriber_number": next_subscriber_number,
```
Gregoor, line 13: How about adding the actual `subscriberNumber` to this data as well, instead of (or in addition to) getting it from `userData`? Would make the thank-you page re-layout less, which is nice.

peterbe: You can't. Actually you can, but I'm not sure it's right. 
At the moment, the `kuma.payments.views.thank_you` view has the `@never_cache` page and there's a hole punched in the CDN for `/*/payments/` so you could get the session cookie on these pages but I don't like that. I think the payments page and the thank you page should be cached in the CDN and we use the `whoami` to update it after initial rendering. Like we're doing with the subscription banner. 

The only reason we currently can't do this yet is because these Django views depend on `@waffle_flag("subscription")` which is a temporary thing. I hope :)

Gregoor: I see! Then I have the opposite question: Shouldn't `next_subscriber_number` then also come from the API?

peterbe: `next_subscriber_number` is not tied to a user in any way. It's a "global" number.

Gregoor: Right, but if its cached wouldn't the number become fairly inaccurate over time?

peterbe: Right now, we don't cache it. It's a cheap computation so using Redis instead of MySQL wouldn't be a big win. 
The harsh reality is that it's not perfectly sustainable but it'll work in Kuma. 
If we decide to put a `cache-control` on the `/en-us/payments/` page it means that that number will be stale in a CDN for 5 minutes. 

It definitely is something weird about this whole design. Because, caching or not, if some other person manages to sign up between you loading the payments page and you filling out the Stripe form that message will no longer be accurate. I.e. it *said* "You will become member number 100" but when I finally typed in the credit card and zip code and everything, I later find out that I actually became 101. 
I see no way around that other than locking the number. Like they do with selling concert tickets or airplane tickets but that's crazy because you can't lock the *one* and only number there is. I guess, an alternative is to generate the next number for you each time you hit the page and include it in the submission. 
I.e. when *I* load the landing page it says "you will be number 123" and when *you* load the page it says "you will be number 124" and then we (almost) simultaneously start filling the credit card, maybe even at different speeds, and additionally, suppose that I bail, then there'll be one member number 122 and one 124 (you) but no member number 123. 

Therefore I think it's safe, under the circumstances, to do it like this. 
Having said that, it wouldn't perhaps be that horribly hard to add a `/api/v1/subscriptions/next-subscriber-number`, called from a `useEffect`, which can *additionally* overlay whatever might have been rendered in the CDN and can be stale. 

How about we wait with that until we have hundreds of new subscribers every day?

Gregoor: Oh I misread before, I thought we would cache it (for a potentially long time). Agreed, I don't think being a number or two off every now and then would be be bad. But being off by a lot consistently, would've been strange. But since this isn't the case, that's good.

What I liked about having it in the API is also that it makes it more Yari2 compatible. But yeah, this is probably good enough for now :+1:

Issue #6747, opened 2020-03-26
GA instrumentation on clicking "Sign in" doesn't make sense any more


peterbe:
**Summary**
In https://github.com/mdn/kuma/commit/865deab60d6a97966e13eff6f1bbf492e1631afe 
https://github.com/mdn/kuma/pull/6252 we re-enabled the same GA instrumentation we used to have on the Wiki onto the read-only site. 
Later we enabled multi-auth. Now, clicking "Sign in" doesn't means you're going to GitHub immediately. No, now it's a modal that appears and the user has to make a choice. 

**Steps To Reproduce (STR)**
Go to http://localhost.org:8000/en-US/ with some GA debugging in web console. 
Press "Sign in" and notice that it sends
```
Running command: ga("send", {hitType: "event", eventCategory: "Authentication", eventAction: "Started sign-in", eventLabel: "GitHub"})
```

Granted, we might someday have an instrumentation event for people who click to see the modal but never follow through. 

**Additional context**

I can't `git revert` [the commit](https://github.com/mdn/kuma/commit/865deab60d6a97966e13eff6f1bbf492e1631afe) because it does so much more.

peterbe:
CC @atopal if you're wondering about those events.

atopal:
Peter, I wish you hadn't removed this. Can you turn it on again and just remove the "github" label? I know the meaning has changed slightly, but this is a metric we have been collecting since before 2015. It's a simple way to see the impact of our design decisions, and I don't think there are any side effects, are there?

peterbe:
The side effect is that I needed that `onclick` for triggering the modal. That way I can use the modal not only in the header but also on the Subscription landing page. So it was in the way. I'll see if I can combine the two.

atopal:
Ah, alright. Thanks Peter!

atopal:
If it's not possible, let me know, we may sacrifice this then.

peterbe:
> If it's not possible, let me know, we may sacrifice this then.

It's possible. I just [made a function](https://github.com/mdn/kuma/pull/6795/files#diff-ad95ab66b0d421a930f76e10485e2418R109) that calls one function at a time. Hopefully, it'll work. 
I'd rather we don't have to do this considering that the old code that got left behind was for something entirely different which is no longer the case. Problem is that I just moved too quickly.

PR #6795, opened 2020-04-02
re-enable ga instrumentation on modal click

peterbe:
Part of #6747

This puts it back. @atopal Do you want to fix the fact that it says `GitHub` when in fact the user hasn't made a choice yet?
Screen Shot 2020-04-02 at 7 07 25 AM

### DIFF
--- a/kuma/javascript/src/header/__snapshots__/header.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/header.test.js.snap
@@ -397,7 +397,6 @@ exports[`Header snapshot 1`] = `
   </form>
   <a
     className="signin-link"
-    data-service="GitHub"
     href="/en-US/users/account/signup-landing?next=/[fake%20absolute%20url]"
     onClick={[Function]}
     rel="nofollow"
--- a/kuma/javascript/src/header/__snapshots__/header.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/header.test.js.snap
@@ -397,6 +397,7 @@ exports[`Header snapshot 1`] = `
   </form>
   <a
     className="signin-link"
+    data-service="GitHub"
     href="/en-US/users/account/signup-landing?next=/[fake%20absolute%20url]"
     onClick={[Function]}
     rel="nofollow"
--- a/kuma/javascript/src/header/__snapshots__/login.test.js.snap
+++ b/kuma/javascript/src/header/__snapshots__/login.test.js.snap
@@ -66,7 +66,6 @@ exports[`Login component when user is logged in 1`] = `
 exports[`Login component when user is not logged in 1`] = `
 <a
   className="signin-link"
-  data-service="GitHub"
   href="/en-US/users/account/signup-landing?next=/"
   onClick={[Function]}
   rel="nofollow"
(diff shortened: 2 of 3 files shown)


peterbe:
Actually @atopal you said ["just remove the "github" label?"](https://github.com/mdn/kuma/issues/6747#issuecomment-607529716)
But what should it be instead?

atopal:
> Actually @atopal you said ["just remove the "github" label?"](https://github.com/mdn/kuma/issues/6747#issuecomment-607529716)
> But what should it be instead?
Nothing. The label is optional

peterbe:
It's still a bit of a "lie". You haven't actually "Started sign-in" just because you clicked to trigger the modal to appear. :)

What we could consider though, is to fix the modal. Hear me out...
It used to be that clicking "Sign in" was the same as going to github.com/auth.... and that's what that GA instrumentation piece was for. That click has actually been moved to the modal now. 
In the flurry of building that modal and the "multi-auth" we missed this instrumentation piece. 
So the functionality still exists! It's just been moved. And we now have another sign in link we could instrument on and this other one says "Google". 

So what we could do is rewrite that old `sendSignInEvent` stuff so it's based on the actual click *inside* the modal. That way we get almost the same functionality as before multi-auth but we'd get more. 

What do you think?

atopal:
> What we could consider though, is to fix the modal. Hear me out...
> It used to be that clicking "Sign in" was the same as going to github.com/auth.... and that's what that GA instrumentation piece was for. 

No, that instrumentation has been around for more than 5 years. Back then we had Github and Persona as authentication options, via  a modal no less ;) That's why that event still has a label. With only one option you don't actually need a label. It shows interest in logging in.

> That click has actually been moved to the modal now.
> In the flurry of building that modal and the "multi-auth" we missed this instrumentation piece.
> So the functionality still exists! It's just been moved. And we now have another sign in link we could instrument on and this other one says "Google".
> 
> So what we could do is rewrite that old `sendSignInEvent` stuff so it's based on the actual click _inside_ the modal. That way we get almost the same functionality as before multi-auth but we'd get more.

The idea is to see how many people click to sign in. I'd like to keep this metric so we can see over time how much more we have incentivized people, especially if we start working on increasing logged in user numbers again. That said, I'm not religious about it. And if this stands in the way of something useful, we can sacrifice it.

peterbe:
What I'm trying say is that as it now goes, clicking the "Sign in" doesn't mean you're bothering to sign in. You could potentially close the modal and you haven't started signing in. 

Either way, once I can get the tests to pass we're back to how it used to be. r?

atopal:
Yes, and clicking on that link before, when you weren't signed in to Gitub, didn't mean that either. It's a potential sign-in-intent. Whether they are sucessful was mesaured with the "Finished sign-in" event.

peterbe:
> Yes, and clicking on that link before, when you weren't signed in to Gitub, didn't mean that either. It's a potential sign-in-intent. Whether they are sucessful was mesaured with the "Finished sign-in" event.

Before we launched multi-auth, clicking "Sign in" started the redirect to Github. It doesn't anymore. Now it's just an intent to display the modal. 

Either way, semantics. Are you ok with me merging this which undoes the removal of that client-side GA event on the click?

atopal:
Yes 👍

PR #6796, opened 2020-04-02
Add .gitattributes for keeping LF EOL for all text

Gregoor:
I once again ran into the problem that kuma will fail to run fairly cryptically if one has CRLF as line endings. Unfortunately git for Windows defaults to applying those for text on checkout (aka `autocrlf`). If one unsets it and updates git, it'll be set again. So this adds a `.gitattributes` file that tells git to use LF for all files, independently of system settings.

This should only affect Windows users and all editors I know of know how to read LFs, so currently I don't foresee any downsides to this.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*	text eol=lf


peterbe:
Where did you get this recommendation from? 
I did a quick Google search and stumbled on https://www.edwardthomson.com/blog/git_for_windows_line_endings.html which recommends
```
* text=auto
```
...instead. 
And that's actually what they have in the default example here too: https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

Gregoor:
What the setting described in the link does is 

> With this set, Windows users will have text files converted from Windows style line endings (\r\n) to Unix style line endings (\n) when they’re added to the repository.

but with that

> kuma will fail to run fairly cryptically if one has CRLF as line endings.

Now it might be the case that there are tools which can't handle LF on Windows, but so far I only had the opposite problem.

PR #6798, opened 2020-04-02
start ssr servie after 'make build-static' for CI

peterbe:
Fixes https://github.com/mdn/kuma/issues/6797


peterbe:
It works!! I'm really pleased with this fix. It's not costing us any extra time and you do get a chance now to test the rendering of SSR via Django. 
If we can land this soon, I can update my ongoing work on the "Subscriber number" to actually make assertions on the rendered HTML *content*.

PR #6806, opened 2020-04-03
open stripe after auth

### LINKED ISSUE #6741: T - Update redirect so that after signup flow user lands on Stripe modal

peterbe:
Fixes #6741 

Screen Shot 2020-04-03 at 2 37 01 PM

Here's how to test it. 

1. Enable `subscription` and `subscription_banner` for everyone.
2. Sign out
3. Go to http://localhost.org:8000/en-US/payments (use your imagination that the Subscription banner took you there)
4. Check checkbox and click "Continue"
5. The auth modal should appear. Click to sign in. 
6. Due to [this bug](https://github.com/mdn/kuma/issues/6805) after you've signed, it takes you to `/en-US/` instead of `/en-US/payments/` so imagine that bug is fixed by manually adding `payments/` to the URL.

Expect the checkbox to automatically be checked and that Stripe starts as soon as the script has loaded. 

This solution uses `sessionStorage` to remember that you've at least once agreed to the checkbox. That means it's tied to the one and same tab. 
If you want to quickly test it, when you on http://localhost.org:8000/en-US/payments/ you can type `sessionStorage.setItem('stripe-form-continue','true')` into the Web Console and refresh the page. 

CC @schalkneethling Because of the SCSS change I had to make. 
In Firefox, if you have a controlled React component like this:
```javascript
<input type="checkbox" checked={myBool} onChange={handlerFunction} required/>
```
(which is the correct way to do checkboxes in React)
what happens is that you get that red glow around it. This only happens in Firefox (only other browser I tested was Chrome). I don't like that Firefox! It makes it look like something's wrong or that the user has done something wrong. 
Hence the SCSS hack. You still get the native effect of the checkbox that prevents a form submission if it's not checked. And that's enough for me. 
You can play with it here: https://codesandbox.io/s/sad-bose-zkvnz 

Also @schalkneethling pay attention to the changes I made to the auth modal stuff. It seems to work :)


kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -72,55 +72,118 @@ export default function SubscriptionForm() {
...
+     * If you arrived on this page, being anonymous, you'd have to first sign in.
+     * Suppose that you do that, we will make sure to send you back to this page
+     * with the sessionStorage key set.
+     * Basically, if you have that sessionStorage key, it will, for you, check the
+     * checkbox and press the "Continue" button.
+     */
+
+    useEffect(() => {
+        if (userData && userData.isAuthenticated) {
+            let autoTriggerStripe = false;
```
Gregoor, line 87: can we move that sessionStorage value extraction & deletion into a separate function? I think it would aid readability.

Gregoor: Could look like this:
```javascript
useEffect(() => {
    if (userData && userData.isAuthenticated && popStripeContinuation()) {
        setPaymentAuthorized(true);
        setOpenStripeModal(true);
    }
}, [userData]);
```
(there's probably a better name somewhere out there)

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -72,55 +72,118 @@ export default function SubscriptionForm() {
...
+
+    function submitHandler(event) {
+        event.preventDefault();
+        // Not so fast! If you're not authenticated yet, trigger the
+        // authentication modal instead.
+        if (userData && userData.isAuthenticated) {
+            setOpenStripeModal(true);
+        } else {
+            try {
+                sessionStorage.setItem(
```
Gregoor, line 165: if you do extract the getter (as recommended above), I might also extract that right next to it which has the nice benefit of grouping serialization and deserialization.

kuma/javascript/src/payments/subscription-form.jsx
```diff
@@ -42,6 +42,42 @@ function useScriptLoading(url) {
...
 
+const STRIPE_CONTINUE_SESSIONSTORAGE_KEY = 'stripe-form-continue';
+
+/**
+ * Return true if you have a sessionStorage key that says you can go straight
+ * to continue the Stripe subscription form.
+ * This also has the side-effect that once you call it, it demolishes that
+ * sessionStorage.
+ */
+function getStripeContinuation() {
```
Gregoor, line 53: nit(ish): Imo `get` is the wrong verb for a destructive read. I don't know what's the best one, but here's some ideas: pop, retrieve, take.
Alternatively I'm also down for explicitness like `getAndUnsetStripeContinuation`

peterbe: Yeah, pop is better. I actually had it like that and then changed my mine. The reason I changed my mind is that I first thought it didn't make sense to use:
```
var value = popSomething()
```
but then I realized, late, that that's how Python works:
```python
>>> list=list('abc')
>>> value = list.pop()
>>> value
'c'
```

peterbe:
@Gregoor I opted for `popStripeContinuation` and `pushStripeContinuation` and fixed a conflict. 
I'll merge this as soon as it goes green.
At the time of writing, the integration between this PR and Travis doesn't appear to work but it definitely started a build: https://travis-ci.com/github/mdn/kuma/builds/158845201
So I'll keep an eye on that one.

PR #6807, opened 2020-04-03
Drop 'timezone' from whoami

### LINKED ISSUE #6750: Drop 'timezone' from whoami
**Summary**
In [`api/v1/whoami`](https://github.com/mdn/kuma/blob/d29cd36431459030afc9145b906dcfad88b398ac/kuma/api/v1/views.py#L216) we're including a `timezone` value. 

No code appears to use it for anything. It's mention in `user-provider.jsx` but only to build up the `userData` context but nothing beyond that that actually uses it. 

**Audience**
Developers

**Rationale**
It's wasting space. 

**Workaround**
Our "v1" API isn't a public API because it's based on browser cookies. So we don't have to worry about breaking any third-party apps that depend on this. 
Not really a workaround.

**Proposal**
Delete it and clean up around it. 

**Additional context**
Let's block this on https://github.com/mdn/kuma/issues/6742

peterbe:
Fixes #6750


peterbe:
@Gregoor I fixed the conflict now.

PR #6809, opened 2020-04-06
Add task completion survey to landing page

### LINKED ISSUE #6522: Add task completion survey to start page
**Summary**
We currently can't show the task completion survey on the start page, because it requires jqery and we don't have jquery on the start page anymore. So we either add jquery to the start page or we remove the dependency altogether. 

**Rationale**
We should be showing the task completion survey prompt on all pages we expect MDN users to be on, so we don't bias the results and get a full picture of the issues being experienced.

schalkneethling:
This turned out _much_ simpler than I anticipated 😸 

@atopal I only added this to the modern front-end as it does not really make much sense to me to have this on the retired `wiki`. Let me know if that was the wrong call.

@Gregoor r?

Fixes #6522

### DIFF
--- a/kuma/javascript/src/landing-page.jsx
+++ b/kuma/javascript/src/landing-page.jsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 import A11yNav from './a11y/a11y-nav.jsx';
 import ActiveBanner from './active-banner.jsx';
 import Header from './header/header.jsx';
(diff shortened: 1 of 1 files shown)


atopal:
> @atopal I only added this to the modern front-end as it does not really make much sense to me to have this on the retired `wiki`. Let me know if that was the wrong call.

Yeah, that was the intention. 👍

PR #6826, opened 2020-04-07
Client-side tracking for subscriptions (phase 1)

### LINKED ISSUE #6697: T - Subscriptions: client-side GA tracking (phase 1)
Description: This is phase 1 (of possibly several) for implementing _client-side_ GA tracking of Subscriptions. A different task will address server-side GA tracking. 

1. Measure banners shown on front-end 
`monthly payments | CTA shown | banner`
2. Measure clicks on "learn more" on the banner (front-end).
`monthly payments | subscribe intent | banner` 
3. Measure views of the FAQ page/subscription landing page (front-end)
`monthly payments | CTA shown | subscription-landing-page`
4. Measure clicks on "continue" on the FAQ page (front-end).
`monthly payments | subscribe intent | subscription-landing-page` 
5. Measure successful subscription on thank you page load (front-end)
`monthly payments | successful subscription | successful-subscription-front-end`

Gregoor:
Fixes #6697 

**Update:** The comments at the end of the page should reflect the current thinking. This is just here for posterity

___

But I'm not sure if I'm happy with it just yet. Here's my gripe:
- we need to send the ga event for clicking the banner CTA & successful subscription before triggering their associated navigation action (`/payments` for the former, `/payments/thank-you` for the latter)
- so we need to pause navigation until the ga event submission successfully went through
  - or until a timeout is reached, which I've set at 300ms for now. Most people who'll hit that timeout will likely be people who have adblockers or are on flaky networks. The lower we set it, the fewer flaky network people we track
- we could just not track these events, as what we're tracking with them is also (somewhat) tracked by others:
  - subscription creation is also tracked in the backend
    - which would have the downside of not capturing the flow of that user?
  - clicking the banner CTA takes one to the `/payments` page and atm it's the only way to get there. So the event we have there (`CTA shown | subscription-landing-page`) should track a simliar thing?

_Longer term, probably yari-2-longer-term, we could also bring client-side navigation back, which would allow us to navigate within the page, without waiting for GA._

cc @atopal @peterbe 
Would be curious to hear your thoughts on my comments above


atopal:
> Fixes #6697
> 
> But I'm not sure if I'm happy with it just yet. Here's my gripe:
> 
>     * we need to send the ga event for clicking the banner CTA & successful subscription before triggering their associated navigation action (`/payments` for the former, `/payments/thank-you` for the latter)
> 
>     * so we need to pause navigation until the ga event submission successfully went through
>       
>       * or until a timeout is reached, which I've set at 300ms for now. Most people who'll hit that timeout will likely be people who have adblockers or are on flaky networks. The lower we set it, the fewer flaky network people we track

I think some flakiness is fine, we are talking about a lot of events here.


>     * we could just not track these events, as what we're tracking with them is also (somewhat) tracked by others:
>       
>       * subscription creation is also tracked in the backend
>         
>         * which would have the downside of not capturing the flow of that user?

We don't care about the flow at this point, only the conversion rate. The issue is that one measure would come from the front-end and the other from the back-end, and while front-end collects incompletely, the back end collects pretty much every event, it would massively skew the numbers.

>       * clicking the banner CTA takes one to the `/payments` page and atm it's the only way to get there. So the event we have there (`CTA shown | subscription-landing-page`) should track a simliar thing?

That won't be the only way to get to that page. We'll have links in our newsletter every week, on social media and many other places.

kuma/javascript/src/active-banner.jsx
```diff
@@ -182,9 +190,48 @@ function Banner(props: BannerProps) {
...
+    }, [ga]);
+
+    return (
+        <Banner
+            id={SUBSCRIPTION_ID}
+            classname="mdn-subscriptions"
+            title={gettext('Become a monthly supporter')}
+            // do not hardcode dollar amount, use CONTRIBUTION_AMOUNT_USD
+            // https://github.com/mdn/kuma/issues/6654
+            copy={gettext('Support MDN with a $5 monthly subscription')}
```
peterbe, line 213: It is kinda a bummer that we keep pushing this as now, what will happen is that translators will today get `'Support MDN with a $5 monthly subscription'` and soon it'll become `'Support MDN with a %(amount)s monthly subscription'` and that gettext will have to start over.

Gregoor:
Some TODOs from my meeting with Peter that I need to work on before we can put this back into review:

1. Remove the click hijacking code and go with URL query parameters for analytics instead
2. If not possible, at least factor our the GA existence check, and use it to let adblock users pass through immediately
3. Use interpolated for the dollar amount for the banner

Gregoor:
@peterbe changed it, based on our call yesterday. Thanks again for the inspiration!

kuma/javascript/src/ga-provider.jsx
```diff
@@ -28,6 +51,25 @@ const context = React.createContext<GAFunction>(ga);
...
+     * analytics data.
+     */
+    useEffect(() => {
+        const ids = window.location.search
+            .substr(1)
+            .split('&')
+            .map((part) => part.split('='))
+            .filter(
+                ([key, id]) => key === GA_QUERY_KEY && id in QUERY_PARAM_GA_DATA
+            );
```
peterbe, line 66: Nit: You could just do:
```javascript
const sp = new URLSearchParams(window.location.search.substr(1));
if (sp.has(GA_QUERY_KEY) && sp.get(GA_QUERY_KEY) in QUERY_PARAM_GA_DATA) {
  ga('send', QUERY_PARAM_GA_DATA[sp.get(GA_QUERY_KEY)]);
}
```
It's marginally safer since you're letting the browser APIs do the "string parsing".

Gregoor:
Thanks Peter, great points! I've addressed them both.

PR #6829, opened 2020-04-07
link to auth with next

### LINKED ISSUE #6805: T - Fix post-sign-ins redirect so that user lands on correct page
**Summary**
You're not taken back to where you were when you sign in. At least via the modal and GitHub.

Screen-Recording-2020-04-03-at-2 31 33-PM

**Steps To Reproduce (STR)**

1. Go to some page other than the home page. Be not-signed in.
2. Click sign in. 
3. Click GitHub and complete your GitHub auth stuff

**Actual behavior**
I ended up back on the home page `/en-US/`. :(

**Expected behavior**
To be taken to the page I was originally on. 

**Additional context**
Pretty sure this is regression but don't know if it started when we switched to a modal.

peterbe:
Fixes #6805 

To test, start by being anonymous, then...

1. go to a page like http://localhost.org:8000/en-US/docs/Web/JavaScript and click "Sign in" clicking on either Google or GitHub will have the `?next=%2Fen-US%2Fdocs%2FWeb%2FJavaScript` thing on it/ 
2. instead of single-clicking the "Sign in" link, right-click and "Open link in a new tab". It should take you to a page like http://localhost.org:8000/en-US/users/account/signup-landing?next=/en-US/docs/Web/JavaScript Observe that the GitHub and Google buttons have `?next=%2Fen-US%2Fdocs%2FWeb%2FJavaScript` on it. 
3. go straight to the sign in landing page http://localhost.org:8000/en-US/users/account/signup-landing and note that the Google and GitHub buttons *don't* have a `?next=...` on them because it doesn't want to redirect back to this page and it should instead redirect to whatever the default is. 
4. type in `http://localhost.org:8000/en-US/profiles/$username/edit` (replace `$username` for something that exists). It should have a "Sign in" link with `?next=%2Fen-US%2Fprofiles%2Fpeterbe%2Fedit` on it.

### DIFF
--- a/kuma/core/context_processors.py
+++ b/kuma/core/context_processors.py
@@ -4,6 +4,8 @@ from constance import config
 from django.conf import settings
 from django.utils import translation
 
+from kuma.core.urlresolvers import reverse
+
 from .i18n import get_language_mapping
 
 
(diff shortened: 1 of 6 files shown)


peterbe:
@schalkneethling Do you think you can take a look at this one too from a testing point of view?

peterbe:
Great news! I'm pretty sure this is going to work for sign UPs as well (CC @atopal and @tobinmori )

## 1
Screen Shot 2020-04-09 at 8 27 38 AM

## 2
Screen Shot 2020-04-09 at 8 26 31 AM

## 3
Screen Shot 2020-04-09 at 8 28 33 AM

I'll call that a win!

## 4
Screen Shot 2020-04-09 at 8 28 38 AM

tobinmori:
A good news @peterbe . I like it. Thanks!

kuma/core/context_processors.py
```diff
@@ -54,10 +56,30 @@ def i18n(request):
...
+    """Return a function by the same name as the context processor.
+    That means, in the jinja templates, instead of doing
+
+        {% set url = next_url %}
+
+    you just have to do:
+
+        {% set url = next_url() %}
+
+    which means that the actual content processor function isn't executed
```
schalkneethling, line 68:
```suggestion
    which means that the actual context processor function isn't executed
```
?

PR #6858, opened 2020-04-13
check yarn.lock git state

### LINKED ISSUE #6857: We should check that yarn.lock isn't changing in CI
**Summary**
It's all too easy that you accidentally edit `package.json` dependencies instead of using the relevant `yarn add/remove ...` command.
If you do that and accidentally check that into `master` every developer that runs `make build-base` are going to have a git change to the `yarn.lock` file. 

**Audience**
Developers

**Rationale**
We should, in CI, check that the `yarn.lock` file doesn't change. 

**Workaround**
If it happens again, someone has to manually fix it with a PR and stuff. But it's tricky to fix because you need to retrace where `package.json` and `yarn.lock` *were* in sync and work from there. 

**Proposal**
We don't have git inside the docker container but our `docker-compose.yml` uses a mounted file system in Travis. So we can use that. 

**Additional context**
In [stumptown-renderer we use `cksum` before and after](https://github.com/mdn/stumptown-renderer/blob/master/scripts/ci-yarn-lockfile-check.sh) running `yarn install`. Not sure if that's applicable in docker.

peterbe:
Fixes #6857

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ jobs:
       node_js: 12
       cache: yarn
       script:
-        - scripts/ci-yarn-lock-check
         - scripts/ci-jslint
         - scripts/ci-flow
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ jobs:
       node_js: 12
       cache: yarn
       script:
+        - scripts/ci-yarn-lock-check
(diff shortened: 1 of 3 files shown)


peterbe:
https://github.com/mdn/kuma/issues/6857#issuecomment-612879579

PR #6871, opened 2020-04-13
Visual edits to subscription form

mindy:
**Changes**
- Decrease form font size to 12px
- Set width of form to 470px
- Disallow wrapping of text around checkbox 

**To test**
Visit http://localhost.org:8000/en-US/payments/

**Before**
Screen Shot 2020-04-13 at 3 39 44 PM

**After**
Screen Shot 2020-04-13 at 3 19 53 PM

### DIFF
--- a/kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
+++ b/kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
@@ -27,7 +27,6 @@
         section,
         form {
             font-size: 12px;
-
             padding: 20px 20px 5px;
 
             &.error {
@@ -40,9 +39,6 @@
 
             label {
                 display: flex;
-                input {
-                    margin: 0 10px 0 0;
-                }
             }
 
             button {
@@ -53,7 +49,6 @@
 
                 &::after {
                     float: right;
-                    display: inline-block;
                     content: '';
                     background: transparent
                         url('../../../../arrows/arrow-right.svg') 0 0 no-repeat;
@@ -68,8 +63,11 @@
             // and haven't even started paying attention to the form, the checkbox,
             // in Firefox, will appear red. Like something is wrong!
             // This will undo that.
-            input[type='checkbox']:required {
-                box-shadow: none;
+            input[type='checkbox'] {
+                margin: 0 10px 0 0;
+                &:required {
+                    box-shadow: none;
+                }
             }
         }
 
--- a/kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
+++ b/kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
@@ -21,12 +21,14 @@
         background-color: #fff;
         border-radius: 4px;
         box-shadow: $box-shadow-blue;
-        max-width: 460px;
-        min-height: 320px;
+        min-width: 470px;
+        width: 470px;
 
         section,
         form {
-            padding: 20px 20px 0;
+            font-size: 12px;
+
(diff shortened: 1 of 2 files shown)


kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
```diff
@@ -29,6 +30,10 @@
                 margin: 0;
             }
 
+            label {
+                display: flex;
+            }
```
schalkneethling, line 35: I think this might cause more problems than it solves. With this set to `flex`, the alignment of the `checkbox` is a bit off in Chrome:

Screenshot 2020-04-15 at 12 58 08

The alignment in Firefox is also very different and feels odd being centered with the copy:

Screenshot 2020-04-15 at 12 58 21

If we forego `flex` and set the `font-size` to `$tiny-font-size` for the label I feel we get a better end result that is consistent between browsers:

Screenshot 2020-04-15 at 12 58 29

kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
```diff
@@ -52,8 +56,11 @@
             // and haven't even started paying attention to the form, the checkbox,
             // in Firefox, will appear red. Like something is wrong!
             // This will undo that.
-            input[type='checkbox']:required {
-                box-shadow: none;
+            input[type='checkbox'] {
+                margin: 0 10px 0 0;
+                &:required {
+                    box-shadow: none;
```
schalkneethling, line 62: So weird, wonder why Firefox has implemented it this way 🤔

kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
```diff
@@ -14,12 +14,13 @@
...
         box-shadow: $box-shadow-blue;
-        max-width: 460px;
-        min-height: 320px;
+        min-width: 470px;
+        width: 470px;
 
         section,
         form {
-            padding: 20px 20px 0;
+            font-size: 12px;
```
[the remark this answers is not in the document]
malqinneh: We're using the same type size from the legal copy in the footer (which is also set in negative contrast polarity).

We need to overhaul type styles on MDN so that different H styles work well together but, for now, I think it's OK to maintain all legal text to one style.

Also no issues came up about legal text size during user testing.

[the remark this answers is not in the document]
schalkneethling: > We're using the same type size from the legal copy in the footer

To be honest, that copy is also too small.

kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
mindy: It looks like it just needed an `align-items` set to make it look right in Firefox:
ff

schalkneethling: That works!

kuma/static/styles/minimalist/components/subscriptions/_subscription-form.scss
```diff
@@ -21,12 +21,13 @@
...
         box-shadow: $box-shadow-blue;
-        max-width: 460px;
-        min-height: 320px;
+        min-width: 470px;
+        width: 470px;
 
         section,
         form {
-            padding: 20px 20px 0;
+            font-size: 14.5px; // this sets the <small /> text to about 12px
```
schalkneethling, line 29: :+1:

PR #6873, opened 2020-04-14
Revert "Add .gitattributes for keeping LF EOL for all text"

Gregoor:
Reverts mdn/kuma#6796


peterbe:
@mindy I doubt you need to actually test this. Just asking you for review because I think you should be involved.

PR #6877, opened 2020-04-14
Events for (un)authenticated user subscription conversion

### LINKED ISSUE #6827: T - Track subscription conversion of signed-in & not signed-in users
We need two sets of numbers.  
1a. Absolute number of people who clicked on "continue" on the landing page and were already sign in before their current session.
1b. Absolute number of people who successfully subscribed and were in group 1a.

2a. Absolute number of people who clicked on "continue" on the landing page and were not signed in.
2b. Absolute number of people who successfully subscribed and were in group 2a.

That gives us to conversation rates
1b/1a = conversation rate for people who were already signed in when they made the choice to pay

2b/2a = conversation rate for people who had to sign in after they made the choice to pay.

_Originally posted by @atopal in https://github.com/mdn/kuma/issues/6524#issuecomment-610427221_

___

Here's my rewording of it, to make sure nothing is lost in translation:

In the context of users who express an intent to go through the subscription flow (i.e. press continue on the payments landing page), we're looking to subdivide them into two buckets, signed-in and not signed-in. For each bucket we want to know how many were in the bucket to begin with, and how many went through with it and actually created a subscription.

What an implementation could look like:
When a user clicks continue on the payments landing page, we both session-store and send an event to analytics indicating whether they were already logged in.
Once the user successfully creates a subscription, we read from the session-store whether they started of logged-in or not, and send a subscription creation even specific to that to analytics.

cc @peterbe

Gregoor:
Fixes #6827

## What changed
- creating the analytics search params now also uses `URLSearchParams` (thanks for the reminder, Peter!) and the function accepts multiple params
- send new analytic events:
  - `subscribe intent (unauthenticated)`
  - `subscribe intent (authenticated)`
  - `successful subscription (unauthenticated)`
  - `successful subscription (authenticated)`
-  rename the `xStripeContinuation` functions using `sessionStorage` to `xStartedUnauthenticated` to
   - reflect what users did and not what we're using it for
   - reuse it to decide which of the new analytics events to send

## How to test it:
Run through the subscription flow on http://localhost.org:8000/en-US/payments, once already logged-in and once logged-out (which will force you to login during the process).

### DIFF
--- a/kuma/javascript/src/ga-provider.jsx
+++ b/kuma/javascript/src/ga-provider.jsx
@@ -9,9 +9,13 @@ export const CATEGORY_MONTHLY_PAYMENTS = 'monthly payments';
 const GA_SESSION_STORAGE_KEY = 'ga';
 
 function getPostponedEvents() {
-    return JSON.parse(
-        sessionStorage.getItem(GA_SESSION_STORAGE_KEY) || JSON.stringify([])
-    );
+    try {
+        return JSON.parse(
+            sessionStorage.getItem(GA_SESSION_STORAGE_KEY) || JSON.stringify([])
+        );
+    } catch (e) {
+        // No sessionStorage support
+    }
 }
 
 /**
@@ -21,11 +25,15 @@ function getPostponedEvents() {
  * requests).
  */
 export function gaSendOnNextPage(newEvents: any[]) {
-    const events = getPostponedEvents();
-    sessionStorage.setItem(
-        GA_SESSION_STORAGE_KEY,
-        JSON.stringify(events.concat(newEvents))
-    );
+    try {
+        const events = getPostponedEvents();
+        sessionStorage.setItem(
+            GA_SESSION_STORAGE_KEY,
+            JSON.stringify(events.concat(newEvents))
+        );
+    } catch (e) {
+        // No sessionStorage support
+    }
 }
 
 function ga(...args) {
@@ -58,7 +66,11 @@ export default function GAProvider(props: {
      */
     useEffect(() => {
         const events = getPostponedEvents();
-        sessionStorage.removeItem(GA_SESSION_STORAGE_KEY);
+        try {
+            sessionStorage.removeItem(GA_SESSION_STORAGE_KEY);
+        } catch (e) {
+            // No sessionStorage support
+        }
         for (const event of events) {
             ga('send', event);
         }
--- a/kuma/javascript/src/ga-provider.jsx
+++ b/kuma/javascript/src/ga-provider.jsx
@@ -15,6 +15,18 @@ const QUERY_PARAM_GA_DATA = {
         eventAction: 'successful subscription',
         eventLabel: 'subscription-landing-page',
     },
+    'subscription-success-authenticated': {
+        hitType: 'event',
+        eventCategory: CATEGORY_MONTHLY_PAYMENTS,
+        eventAction: 'successful subscription (authenticated)',
+        eventLabel: 'subscription-landing-page',
+    },
+    'subscription-success-unauthenticated': {
+        hitType: 'event',
+        eventCategory: CATEGORY_MONTHLY_PAYMENTS,
+        eventAction: 'successful subscription (unauthenticated)',
+        eventLabel: 'subscription-landing-page',
+    },
     'banner-cta': {
         hitType: 'event',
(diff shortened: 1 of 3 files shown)


Gregoor:
@atopal FYI, these are the new events:

* `subscribe intent (unauthenticated)`
* `subscribe intent (authenticated)`
* `successful subscription (unauthenticated)`
* `successful subscription (authenticated)`

The pre-existing events `subscribe intent` and `successful subscription` will still be sent, for both user buckets. But we COULD remove them since they'll just be the sum of the two. But I wanted to ask you first, if that's desired.

atopal:
Thanks Gregor, let's keep the old ones, it can be cumbersome to sum events, so that might be useful. And we can remove it later if it turns out that we don't need it after all.

tobinmori:
@atopal just double-checking that this is something we want done for Sprint Echo. correct? (I had assumed so)

atopal:
Yes, this story is for the current sprint.

Gregoor:
Peter call feedback:
- drop a comment into the code explaining why we're sending duplicate events
- ga sending for success bugs (and probably any other URL param GA event as well)

kuma/javascript/src/ga-provider.jsx
```diff
@@ -6,25 +6,26 @@ export type GAFunction = (...any) => void;
...
-    'banner-cta': {
-        hitType: 'event',
-        eventCategory: CATEGORY_MONTHLY_PAYMENTS,
-        eventAction: 'subscribe intent',
-        eventLabel: 'banner',
-    },
-};
+function getPostponedEvents() {
+    return JSON.parse(
+        sessionStorage.getItem(GA_SESSION_STORAGE_KEY) || JSON.stringify([])
```
peterbe, line 13: What's wrong with 
```suggestion
        sessionStorage.getItem(GA_SESSION_STORAGE_KEY) || '[]'
```
I'm not trying to change your mind; just trying to learn from you.

Gregoor: I like the former better because it seems more resistant to typos. The latter is "just some string which happens to be the serialized form of an array". I guess this example doesn't drive the point home, but I like erring using the (de-)serialization functions for even the simplest of cases, to not create precedence.

peterbe:
I don't know how you tested it but I started with the path of already being authenticated. 
First I went to `http://localhost.org:8000/en-US/payments` (manually) when clearly the GA events are happening when I press the "Continue" button:
Screen Shot 2020-04-16 at 12 07 08 PM
That's the expected right?

But when I press the "Pay $5.00" button, which redirects(?) to the Thank You page, I see no events taking place:
Screen Shot 2020-04-16 at 12 11 06 PM

Shouldn't I be seeing `successful subscription` and `successful subscription (authenticated)`?

peterbe:
Also, check this out. I start on the home page. I'm authenticated but I don't have a subscription. So the banner appears for me. Good. But when I click the "Learn more" button on the banner, there's no ` eventAction: 'subscribe intent',  eventLabel: 'banner',`
Screen Shot 2020-04-16 at 12 17 10 PM

Did it work when you tested it?

kuma/javascript/src/ga-provider.jsx
```diff
@@ -6,25 +6,35 @@ export type GAFunction = (...any) => void;
...
+ */
+export function gaSendOnNextPage(newEvents: any[]) {
+    try {
+        const events = getPostponedEvents();
+        sessionStorage.setItem(
+            GA_SESSION_STORAGE_KEY,
+            JSON.stringify(events.concat(newEvents))
+        );
+    } catch (e) {
+        // No sessionStorage support
```
peterbe, line 36: It's a nit but now you're conflating the forgiveness of `sessionStorage.setItem` with a bunch of other things. For example, `const events = getPostponedEvents();` could crash. The `JSON.stringify(` could fail. Or the `events.concat(` could fail. 
I think it ought to be:
```javascript
export function gaSendOnNextPage(newEvents: any[]) {
    const events = getPostponedEvents();
    const value = JSON.stringify(events.concat(newEvents));
    try {
        sessionStorage.setItem(GA_SESSION_STORAGE_KEY, value);
    } catch (e) {
        // No sessionStorage support
    }
}
```

Gregoor:
Hm, I can't seem to reproduce either of the issues. Can you add some logs to the code I added to make sure the build pipeline is not screwing things up?

Issue #6848, opened 2020-04-10
Replace Constance with plain settings


peterbe:
**Summary**
Remove `django-constance` as a dependency and move everything to `django.conf.settings` so it's controlled by environment variables instead. 
If anything *can't* be environment variables, but needs to be controlled by the Django Admin, between releases, then we can use waffle switches. 

**Rationale**
The origin of Constance was to be able to edit run-time settings without having to wait for a new release which, historically, was controlled outside the MDN team. So they needed it to be able edit settings faster. 

These days we (the MDN Dev team) control the releases and we release more frequently and "at a whim" if we need to. We also have waffle switches which we're more actively using so we can change MDN between releases via the Django Admin. 

Any use of Constance requires a database lookup since that's where the source of truth is stored. (That and the default values as per our .py code)
If they could be in `django.conf.settings` instead it requires 0 I/O during runtime so there's a slight performance benefit. 

There's also the simple but powerful benefit of having fewer parts that can go wrong. 

**Audience**
Developers

**Proposal**
Search and destroy all uses of Constance in the .py code. And for each key, use the latest MySQL production dumps to make sure whatever default we set (in `settings/common.py`) matches the production value if overwritten in prod MySQL

**Additional context**
We can do it in many little PRs rather than doing them all at once. And when enough has been done, one final PR can delete `django-constance` from `pyproject.toml`.

peterbe:
One complication is that some of those values aren't easily expressed as one-liners in an environment variable. E.g. the `GOOGLE_ANALYTICS_CREDENTIALS`

peterbe:
Note-to-self; having the MySQL prod dumb locally is no good because the values are pickled strings which depend on the version of Python or however `picklefield` stores thing.

PR #6878, opened 2020-04-14
move WELCOME_EMAIL_FROM to settings

peterbe:
Part of #6848

Here's how I checked the default value from prod:
```
>>> from constance import config
>>> config.WELCOME_EMAIL_FROM
'Janet Swisher <no-reply@mozilla.org>'
```

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1507,11 +1507,12 @@ CONSTANCE_CONFIG = dict(
         1000,
         "Number of expired sessions to cleanup up in one go.",
     ),
-    WELCOME_EMAIL_FROM=(
-        "Janet Swisher <no-reply@mozilla.org>",
-        "Email address from which welcome emails will be sent",
-    ),
     AKISMET_KEY=("", "API key for Akismet spam checks, leave empty to disable"),
+)
+
+# Email address from which welcome emails will be sent
+WELCOME_EMAIL_FROM = config(
+    "WELCOME_EMAIL_FROM", default="Janet Swisher <no-reply@mozilla.org>"
 )
 
 # Email address to request admin intervention
--- a/kuma/users/tasks.py
+++ b/kuma/users/tasks.py
@@ -1,7 +1,6 @@
 import logging
 
 from celery import task
-from constance import config
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.utils import translation
@@ -56,7 +55,7 @@ def send_welcome_email(user_pk, locale):
             email = EmailMultiAlternativesRetrying(
                 _("Getting started with your new MDN account"),
                 content_plain,
-                config.WELCOME_EMAIL_FROM,
+                settings.WELCOME_EMAIL_FROM,
                 [user.email],
             )
             email.attach_alternative(content_html, "text/html")


kuma/settings/common.py
```diff
@@ -1507,13 +1507,14 @@ def parse_iframe_url(url):
...
-    WELCOME_EMAIL_FROM=(
-        "Janet Swisher <no-reply@mozilla.org>",
-        "Email address from which welcome emails will be sent",
-    ),
     AKISMET_KEY=("", "API key for Akismet spam checks, leave empty to disable"),
 )
 
+# Email address from which welcome emails will be sent
+WELCOME_EMAIL_FROM = config(
+    "WELCOME_EMAIL_FROM", default="Janet Swisher <no-reply@mozilla.org>"
```
callahad, line 1515: Do we still want to force Janet to be the human-readable face of the automated email whenever anyone signs up?

callahad: Actually, we should defer that decision, since the welcome emails are written from Janet's first person perspective.

...though they also mention IRC, which doesn't exist anymore. I'll file a separate bug.

escattone: Good point. This is the current setting on stage and prod, but maybe we should change it to something like `MDN Web Docs Team` since the community aspect of MDN is no longer Janet's role? @jmswisher Do you have a recommendation here?

escattone: @callahad Oops, missed your previous comment while creating mine!

peterbe: https://github.com/mdn/kuma/issues/6365#issuecomment-577371250

*This* PR doesn't actually take a stand. It just rearranges the Python code (which I know you know)
But to consider;

* Should the welcome email come from "MDN" or "Mozilla Developer Network" or "Mozilla MDN" instead of "Janet Swisher"??
* Having a real persons full name hardcoded in the Python code feels a bit "unprofessional"
* Can anybody remember WHY we picked a human name instead of a more "corporate name"? CC @atopal 

PS. I'm going to go ahead with this PR but we can still discuss the topic on a tagent.

jmswisher: I wanted the email to appear to come from a human, to make it more personal. In fact, I wanted it to come from a reply-able email alias like mdn-community-mgr@mozilla.org, instead of an impersonal no-reply address, but that was deemed infeasible. 

I'm fine with continuing to have my name appear on the email, though it does sound like the message needs to be revisited, if it mentions IRC. Something like "Janet Swisher, MDN Community Lead" would be better than just "Janet Swisher", which leads to users wondering WTH "Janet Swisher" is.

peterbe: I remember starting a branch to tidy up that welcome email the day they shut down the IRC but then I got sidetracked because I didn't have a Matrix URL to replace it with. 

I for one think that "Janet Swisher, MDN Community Lead" is a huge improvement with or without fixing the content of the email itself. 
And, let's leave to entirely to you Janet to decide what email you want. I would strongly recommend a new change request issue explicitly for the *value* of the `WELCOME_EMAIL_FROM` config.

peterbe: Separately, I think @atopal and @malqinneh should also scrutinize that email to consider a link to the Subscription landing page. Every $ counts and with some thinking time on the right wording, it's really easy to put in an edit to that email body. And we can actually use waffle flags for it too I think.

callahad: Separate issue filed at #6883

PR #6897, opened 2020-04-17
subscription api to list subscriptions

### LINKED ISSUE #6892: T - the API to support subscription management
The backend pieces needed for https://github.com/mdn/kuma/issues/6703

That way we, can Reactify the subscription management page. 

What we need is to endpoints:

1. Getting your subscriptions (plural but it's probably never more than 1)
2. Cancel a subscription (or cancel them all if that's easier)

peterbe:
Part of #6892

Here's how I test it,
I open `payments/pages/index.jsx` and inject these lines:

```jsx
    // TEMPORARY HACKY CODE TO TEST THINGS
    useEffect(() => {
        if (isSubscriber) {
            fetch('/api/v1/subscriptions/').then((r) => {
                console.log('RESPONSE:', r);
                if (r.ok) {
                    r.json().then((data) => {
                        console.log(
                            'SUBSCRIPTION DATA:',
                            JSON.stringify(data, null, 2)
                        );
                    });
                }
            });
        }
    }, [isSubscriber]);
```

But you can also do the same with `curl` if you copy the cURL call from the dev tools Network panel as that will have the `Set-Cookie` in it.

I know the PR looks big. It's not that big. The changes to `kuma/api/v1/views.py` and `kuma/api/v1/urls.py` are the most important and most relevant. Everything else is just refactoring now that these subscription stuff things are spread across `kuma.payments`, `kuma.users`, and `kuma.api.v1`. 

I really hope we can see retire EVERYTHING related to subscriptions from `kuma.users` (including `git mv kuma/users/stripe_utils.py kuma/payments`) when we remove the subscription stuff from the Edit Profile page. 

PS. I'm make it a draft for now to break up the introduction of the `GET /api/v1/subscriptions` from the API stuff for cancelling a subscription.

### DIFF
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -37,7 +37,10 @@ from kuma.search.filters import (
 )
 from kuma.search.search import SearchView
 from kuma.users.models import User, UserSubscription
-from kuma.users.stripe_utils import create_stripe_customer_and_subscription_for_user
+from kuma.users.stripe_utils import (
+    create_stripe_customer_and_subscription_for_user,
+    retrieve_and_synchronize_subscription_info,
+)
 from kuma.users.templatetags.jinja_helpers import get_avatar_url
 from kuma.wiki.models import Document
 from kuma.wiki.templatetags.jinja_helpers import absolutify
@@ -365,13 +368,20 @@ def send_subscriptions_feedback(request):
     return HttpResponse(status=204)
 
 
-@api_view(["POST"])
-def create_subscription(request):
+@api_view(["POST", "GET"])
+def subscriptions(request):
     if not request.user.is_authenticated or not flag_is_active(request, "subscription"):
-        return Response(None, status=status.HTTP_403_FORBIDDEN,)
+        return Response(None, status=status.HTTP_403_FORBIDDEN)
 
-    user = request.user
-    create_stripe_customer_and_subscription_for_user(
-        user, user.email, request.data["stripe_token"]
-    )
-    return Response(None, status=status.HTTP_201_CREATED)
+    if request.method == "POST":
+        create_stripe_customer_and_subscription_for_user(
+            request.user, request.user.email, request.data["stripe_token"]
+        )
+        return Response(None, status=status.HTTP_201_CREATED)
+
+    all_subscriptions = []
+    subscription_info = retrieve_and_synchronize_subscription_info(request.user)
+    if subscription_info:
+        all_subscriptions.append(subscription_info)
+
+    return Response({"subscriptions": all_subscriptions})
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -521,23 +521,66 @@ def test_send_subscriptions_feedback_failure(client, settings):
 @patch("kuma.api.v1.views.create_stripe_customer_and_subscription_for_user")
 def test_create_subscription_success(mock, user_client):
     response = user_client.post(
-        reverse("api.v1.create_subscription"),
+        reverse("api.v1.subscriptions"),
         content_type="application/json",
         data={"stripe_token": "tok_visa"},
     )
-
     assert response.status_code == 201
 
 
 @pytest.mark.django_db
 @override_flag("subscription", True)
-def test_create_subscription_failure_without_login(client):
-    response = client.post(reverse("api.v1.create_subscription"))
+def test_subscriptions_without_login(client):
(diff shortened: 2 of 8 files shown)


kuma/api/v1/views.py
```diff
@@ -37,7 +37,10 @@
 )
 from kuma.search.search import SearchView
 from kuma.users.models import User, UserSubscription
-from kuma.users.stripe_utils import create_stripe_customer_and_subscription_for_user
+from kuma.users.stripe_utils import (
+    create_stripe_customer_and_subscription_for_user,
+    retrieve_and_synchronize_subscription_info,
```
peterbe, line 42: This is the big deal. `kuma/users/stripe_utils.py` is now the new central home for these big convenience functions. They get used from `kuma/users/views.py` and `kuma/payments/views.py` and `kuma/api/v1/views.py`. (plus some other things like `users/signal_handlers.py`)

It's an unfinished book what to do exactly with these things and where the ideal place is. I actually thing `kuma.users` is **the wrong place** but I think it all kinda started (a couple of months ago) because we added the Subscriptions 2020 to Edit Profile initially. 

Perhaps when this last batch of re-arrangements is done, we go in and move `stripe_utils.py` from `kuma/users/` to `kuma/payments/`. That way, `kuma/users` gets smaller and `kuma/payments/` can be the home for all things Stripe related.

kuma/api/v1/views.py
```diff
@@ -365,13 +368,20 @@ def send_subscriptions_feedback(request):
     return HttpResponse(status=204)
 
 
-@api_view(["POST"])
-def create_subscription(request):
+@api_view(["POST", "GET"])
```
Gregoor, line 371: any reason to have one fn for both? Imo they don't share enough functionality that I would've opted for that.

peterbe: How else are you supposed to do it? There is one option, but I'm not convinced it's better. 
One option would be to use class-based views so that can have one method called `post` and another method called `get`. 
But we don't use class-based views in Kuma. ...except in some areas where we integrate with django_restframework. 

What would be "ideal" would be if Django allowed you to specify this in `urls.py`. Like this:

```python
urlpatterns = [
    path("subscriptions/", methods.GET, views.get_subscriptions),
    path("subscriptions/", methods.POST, views.create_subscription),
]
```
but that's not how Django works. 

It's never going to get sane because you can change two variables at the same time: the URL and the method. Sometimes, you want the same URL and a list of methods. Sometimes the other way around. 

What would you have opted for?

peterbe: By the way, I think I can see why Django works like this. Back in the day, when there wasn't any SPAs and where Django got to do everything and almost always with Django forms. You'd see this pattern a lot:

```
def some_view_that_shows_and_creates(request):
    if request.method == "POST":
        form = FeedbackForm(request.POST)
        if form.is_valid():
            form.save()
            # redirect to prevent browser refreshes submitting
           return redirect(url() + "?created=true")
    else:
         form = FeedbackForm()
    return render(request, 'template.html', {'form': form})
```
this way, the `form`, when used inside `template.html` it's sometimes bound by data and validation and sometimes unbound.

Gregoor: Ah dang, didn't consider Django being in the way of doing it right, sorry 🙈 

django-rest-framework has a feature for handling it, but that seems to necessitate using classes, which I find even less exciting:
https://www.django-rest-framework.org/api-guide/viewsets/

So yeah, sorry, not as actionable I'd have hoped this suggestion would be.

Gregoor: Oh right, I remember that pattern. It does make sense for that. Though I think a router (aka urlpatterns) where you'd also declare request method would make more sense (which is how Rails does it). Then you can have both the classic server rendered and the SPA cake 🤷‍♂️ But I guess we have to work with the Django we have, not the one I want 😁

peterbe: I'm pretty confident that the REST'y use of 
```
GET  /api/v1/subscriptions/
POST /api/v1/subscriptions/
```
is sane. I don't think Kuma (or API v1) is ever trying to be accurate about how it does REST but it's a pretty common pattern to use the same URL but different methods.
...instead of doing something like:
```
GET  /api/v1/subscriptions/list/
POST /api/v1/subscriptions/create/
```

Gregoor: oh yeah no, I'm in complete agreement with what the URLs look like, it's just the view unification where I think django does the wrong thing.

PR #6915, opened 2020-04-20
cancel subscription v1 api

### LINKED ISSUE #6892: T - the API to support subscription management

peterbe:
Fixes #6892

This PR is blocked by: https://github.com/mdn/kuma/pull/6897

Here's how I tested it; in `payments/pages/index.js` I injected this mess:

```javascript

    // TEMPORARY HACKY CODE TO TEST THINGS
    useEffect(() => {
        if (isSubscriber) {
            fetch('/api/v1/subscriptions/').then((r) => {
                if (r.ok) {
                    r.json().then((data) => {
                        const [subscription] = data.subscriptions;
                        console.log(
                            'SUBSCRIPTION DATA:',
                            JSON.stringify(subscription, null, 2)
                        );
                        const u = `/api/v1/subscriptions/`;
                        fetch(u, {
                            method: 'DELETE',
                            headers: {
                                'X-CSRFToken': getCookie('csrftoken'),
                            },
                        }).then((r) => {
                            console.log('RESPONSE', r);
                        });
                    });
                }
            });
        }
    }, [isSubscriber]);
```
and it canceled by subscription. It's something like that that @mindy will implement in https://github.com/mdn/kuma/issues/6896

I know the diff looks huge but it's actually not too bad. The most important change is that the `kuma.api.v1.views.subscriptions` now can take `DELETE` requests and that'll cancel all your subscriptions. 

I thought a lot about this. At first, I thought I'd want to split it up so you'd be able to cancel individual subscriptions, just to be future-proof. But is it??? Will we ever have the ability to cancel B but leave A, C subscriptions still on? If so, can't we just deal with it then? It feels like a huge deal to add support for having multiple subscriptions. 
There's a certain truth to the fact that there might exist old users, from the Subscriptions v0 days, that have old subscriptions. But if so, there is still only 1 `settings.STRIPE_PLAN_ID` so you can't have multiple subscriptions unless we severely refactor how all of this is working. Therefore, I decided to keep it simple and make it so that cancelling your subscription means to iterate over all subscriptions by the user and cancel each. Like it used to work like.


peterbe:
Note-to-self; when https://github.com/mdn/kuma/pull/6897 lands in `master` I'll merge master into this PR and make it ready for review.

peterbe:
Gha!! I forgot to add the new test to `api/v1/tests/test_views.py`

kuma/api/v1/urls.py
```diff
@@ -9,10 +9,10 @@
...
     re_path(r"^users/(?P<username>[^/]+)/?$", views.get_user, name="api.v1.get_user"),
-    re_path(
-        r"^subscriptions/feedback/$",
+    path(
+        "subscriptions/feedback/",
         views.send_subscriptions_feedback,
         name="api.v1.send_subscriptions_feedback",
     ),
-    re_path(r"^subscriptions/$", views.subscriptions, name="api.v1.subscriptions"),
+    path("subscriptions/", views.subscriptions, name="api.v1.subscriptions"),
```
peterbe, line 17: So much more convenient. We should be using `path()` instead of `re_path()` wherever we can instead. 
The reason we're not already using it everywhere is because we didn't have the option when we upgraded to Django 2. Had we attempted to upgrade to Django 2 and fix all the `urlpatterns` we'd be taking on too much of a risk.

kuma/javascript/src/payments/components/subscription-form.jsx
```diff
@@ -10,7 +10,7 @@ import GAProvider, {
 import UserProvider from '../../user-provider.jsx';
 import { getCookie } from '../../utils';
 
-const SUBSCRIPTION_URL = '/api/v1/subscriptions';
+const SUBSCRIPTION_URL = '/api/v1/subscriptions/';
```
peterbe, line 13: Hugely important!
Actually, without this fix, it led to a very dangerous problem. 
If you do...
```javascript
fetch('/api/v1/subscriptions', {method:"POST", ...}).then(r => {
  if (response.ok) {
    ....
})
```
what would happen is that you'd see, in the Django server logs:
```
POST /api/v1/subscriptions => 301
GET /api/v1/subscriptions/ => 200
```

kuma/payments/tests/test_utils.py
```diff
@@ -1,60 +0,0 @@
...
-    cancel_stripe_customer_subscription("cust_id")
-    assert mock_sub.call_count == 2
-
-
-@mock.patch("stripe.Customer.retrieve", return_value={"subscriptions": {"data": []}})
-@mock.patch("stripe.Subscription.retrieve", side_effect=Exception("Oops"))
-def test_cancel_stripe_customer_no_subscriptions(mock_sub, mock_cust):
-    """A stripe customer with no subscriptions returns True."""
-    cancel_stripe_customer_subscription("cust_id")
-    assert not mock_sub.called
```
peterbe, line 60: The reason I deleted this was because I don't think any of it isn't already touched by other more comprehensive tests. Also, we're moving towards `kuma/users/stripe_utils.py` instead for our core workers. And https://github.com/mdn/kuma/issues/6916 would be about the same amount of work.

peterbe: I should add, a lot of the stuff that was being tested in here got moved to `stripe_utils.py` anyway.

kuma/users/stripe_utils.py
```diff
@@ -95,6 +95,26 @@ def create_stripe_customer_and_subscription_for_user(user, email, stripe_token):
...
+    """Delete all subscriptions for a Stripe customer."""
+    assert user.stripe_customer_id
+    cancelled = []
+    for subscription_id in cancel_stripe_customer_subscription(user.stripe_customer_id):
+        UserSubscription.set_canceled(user, subscription_id)
+        cancelled.append(subscription_id)
+    return cancelled
+
+
+def cancel_stripe_customer_subscription(stripe_customer_id):
```
Gregoor, line 108: I'd reconsider renaming those two fns, imo they don't differ on the fact they operate on single/multiple subscriptions as the names imply, but that one does it over at stripe, and one uses that to deactive our local ones. Here's an idea that's probably not optimal either: `cancel_stripe_subscriptions` and `cancel_subscriptions`. Where the latter calls the former and keeps up the pretense that we could switch out subscription service providers.

peterbe: Gee. I can't remember why `cancel_stripe_customer_subscription` even exists! The `api.v1` needs only the `cancel_stripe_customer_subscriptionS` one. Where or which place, outside, this file, needs to dig in specifically to `cancel_stripe_customer_subscription`??

PR #6923, opened 2020-04-21
update robots.txt for developer.mozilla.org

### LINKED ISSUE #6812: T - Update robots.txt (remove sites)
**Summary**
There are several page types that we have in our robots.txt, which means that Google can't crawl it to see that we don't want those pages to be indexed.

**Steps To Reproduce (STR)**
1. Go to Search Console
2. go to Coverage > Valid with Warnings > Indexed, though blocked by robots.txt
3. Alternatively: https://developer.mozilla.org/robots.txt

**Actual behavior**
Google has a link to https://developer.mozilla.org/en-US/docs/tag/SpeechRecognitionResult (for example), but that URL is blocked in robots.txt, so it can't follow the link to see that it redirects to https://wiki.developer.mozilla.org/en-US/docs/tag/SpeechRecognitionResult which has a `no-index` directive.

**Expected behavior**
All page types that redirect to the wiki should be removed from the robots.txt of the read-only site, since the wiki has a `no-index` directive on all content already.

**Additional context**
The reason to do this is so we can see actually problematic content show up in our search console reports, instead of this noise. Search console only shows up to 1000 pages as problematic, but there are currently more than 10k warnings, so we might be missing large issues.

escattone:
Fixes #6812 

This turned-out to be much more work than just removing lines from developer.mozilla.org's `robots.txt`, since many of the URL's that would have been exposed to Google bot would NOT have returned `noindex,nofollow` in either the `robots` `meta` tag or the `X-Robots-Tag` header.

This is a draft PR since there is more work to do.


atopal:
Ah, sorry @escattone . We can leave those out in this first iteration and only remove those that would return at least no-index.

escattone:
> Ah, sorry @escattone . We can leave those out in this first iteration and only remove those that would return at least no-index.

Thanks @atopal, I think that's better. I'll put together a simple PR to remove the ones that would return `noindex` for this first iteration, and then we can address more in the next iterations.

PR #6926, opened 2020-04-21
CI python without coverage

peterbe:
This makes it so that coverage tests are only used if the PR build is a PR and has any changes to `.py` files. 

See comment below about the potential (hint: 1min faster builds)

### DIFF
--- a/scripts/ci-python
+++ b/scripts/ci-python
@@ -87,12 +87,7 @@ echo "TRAVIS_PULL_REQUEST:"
 echo $TRAVIS_PULL_REQUEST
 
 is_pullrequest_build() {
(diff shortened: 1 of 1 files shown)


peterbe:
When running `docker-compose exec -T web make test` you get:
```
1999 passed in 135.48s (0:02:15)
```

When running `docker-compose exec -T web make coveragetest` you get:
```
1999 passed in 206.95s (0:03:26)
```

So, if we can avoid coverage tests (for good reasons!) we stand to save 71 seconds. I.e. **feedback on the tests passing 1 minute sooner**. Not a sneeze!

peterbe:
@callahad Not asking for a review from you but perhaps you can share some impressions and thoughts.

PR #6927, opened 2020-04-21
update robots.txt for developer.mozilla.org

### LINKED ISSUE #6812: T - Update robots.txt (remove sites)

escattone:
Partially fixes #6812 

This PR removes some paths listed in #6812 from `robots.txt` that return `noindex` in either their `robots` `meta` tag or their `X-Robots-Tag` header. Some other paths listed in #6812 do not yet return `noindex`, namely:
```
Disallow: /api/
Disallow: /*docs/get-documents
Disallow: /*docs/Experiment:*
Disallow: /*$children
Disallow: /*docs.json
Disallow: /*/files/
Disallow: /media
Disallow: /*profiles*/edit
Disallow: /*users/
```
but we can address those in another iteration.


escattone:
I'm going to merge this one without review since it's simple and I'd like to get it into the push to stage/production later today.

PR #6928, opened 2020-04-21
Responsive toc

### LINKED ISSUE #6817: T - Implement mobile-optimized Table of Contents
See User Story for design and description: #6785

# Acceptance Criteria 
- [x] Essential: New responsive table of contents implemented on mobile (only)
- [x] Non-essential: new responsive table of contents implemented on desktop, tablet
- [x] [Instrumentation](https://docs.google.com/spreadsheets/d/1plvnK5gBcrpmSwVwaVMFVvPK8ihrDL4LzObhZOtlj2Y/edit?usp=sharing) on [tap] events for table of contents (if applicable events exist on desktop)

schalkneethling:
Makes the table of contents responsive across devices

## Mobile initial state

Screenshot 2020-04-21 at 19 14 56

## Mobile after interaction

Screenshot 2020-04-21 at 19 15 13

## Tablet initial state

Screenshot 2020-04-21 at 19 15 26

## Tablet after interaction

Screenshot 2020-04-21 at 19 15 36

## Desktop

Screenshot 2020-04-21 at 19 16 23

@mindy @Gregoor @malqinneh r?

fixes #6817


kuma/javascript/src/toc.jsx
```diff
@@ -48,11 +62,20 @@ export default function TOC({ html }: Props) {
 
     return (
         <aside className="document-toc-container">
-            <section className="document-toc" ref={tocRef}>
+            <section className="document-toc" ref={documentTOCRef}>
```
peterbe, line 65: The React-way of doing this would be to use React state and then reflect that state in `className={...}`

kuma/static/styles/components/wiki/_toc.scss
```diff
@@ -126,36 +126,109 @@ On This Page component
...
 
     ul {
-        @include bidi(((padding-left, 20px, padding-right, 0),));
+        display: none;
+        position: absolute;
+        top: 50px;
+        background-color: #fff;
+        padding: 0 20px;
+        border: 2px solid $blue;
+        width: 90%;
```
mindy, line 182: I think something like `width: calc(100% - 40px)` would match the design a little more closely. And then that would  eliminate the need for the `width: 94%;` part below.

schalkneethling: Good idea, let me give this a whirl.

kuma/javascript/src/toc.jsx
```diff
@@ -48,11 +59,20 @@ export default function TOC({ html }: Props) {
...
```
Gregoor, line 62: Sorry for nitting this here, given that the code existed before this PR, but I think instead of using ref&effect we should do
```jsx
<section onClick={sendTOCClicks}/>
```
Should do the same thing, with less code in this context, no?

schalkneethling:
Thanks for the review @peterbe @mindy @Gregoor 🍵

PR #6929, opened 2020-04-21
better test coverage on stripe utils

### LINKED ISSUE #6916: T - improve test coverage for stripe_utils.py
**Summary**
All the red in this report says it all:
Screen Shot 2020-04-20 at 11 45 17 AM

I don't think it was always this bad. I suspect it was never perfect but we've moved stuff around and rearrange things. We have 3 Django apps: `users`, `payments`, and `api.v1` that all "reach in" to do stuff with Subscriptions in different ways. 

Our other "excuse" is that we quickly injected the new Stripe Subscription MVP into the Edit Profile page instead of reviving the old `payments` app. 

Either way, test coverage should still be high level and not test implementation details but. We just need to refactor how we're doing tests. 

**Rationale**
Be sustainable and bug free in the area of subscriptions on the server side. 

**Audience**
Devs.

**Proposal**
Review and revisit some test functions so that we get much improved test coverage. 

**Additional context**
I noticed this when I was working on https://github.com/mdn/kuma/pull/6915 and that PR (which at the time of writing is a DRAFT) changes how a lot of subscription view and utility functions work. So I'd much rather tackle *this* PR once *that* PR has landed.

peterbe:
Fixes #6916

How to test it:
```
pytest --cov=kuma --no-cov-on-fail kuma/api/v1/tests/test_views.py -x --ff && coverage html
```
then `open htmlcov/kuma_users_stripe_utils_py.html` and see that it's mostly covered. 

This turned out to be a lot messier than I had hoped. And we are potentially taking on some duplication of tests. 

The point is, the v1 JSON API is the only "API" that matters. What these tests are trying to simulate is a human QA person trying to do things like listing subscriptions, canceling subscriptions, and creating subscriptions all using `curl`. That QA person wouldn't care how it's done as long as it doesn't actually use Stripe over the network. 

The mocking got a little hairy and it's as fragile as all mocking always is. But it walks important business logic code we've written. 

Now, I fear we still have a bunch of duplicate tests somewhere in `kuma.payments` and `kuma.users` but those will soon *not* have a public API. Once we get rid of subscription stuff on the Edit Profile page, we can tackle more refactoring. That day (when we get rid of subscriptions on the Edit Profile, and the Subscription management page has been rewritten in React) then we can revisit some tests in `users/tests/*.py` and `payments/tests/*.py`.


kuma/users/stripe_utils.py
```diff
@@ -99,10 +99,13 @@ def cancel_stripe_customer_subscriptions(user):
     """Delete all subscriptions for a Stripe customer."""
     assert user.stripe_customer_id
     customer = stripe.Customer.retrieve(user.stripe_customer_id)
+    canceled = []
     for sub in customer.subscriptions.data:
         s = stripe.Subscription.retrieve(sub.id)
         UserSubscription.set_canceled(user, s.id)
         s.delete()
+        canceled.append(s)
+    return canceled
```
Gregoor, line 108: smuggled a fix for an unrelated problem in, huh? 😏 I'm all for fixes

peterbe: I have no memory of that! Gah!! I wonder if I accidentally worked on the wrong branch or something.

Gregoor: Oh I guess it would've been more prudent for me to postpone merging after you replied whether it was intentional. Will do so next time!

PR #6940, opened 2020-04-23
Convert Subscription Management page to React

### LINKED ISSUE #6896: T - Front-end for Subscriptions Management
Front-end portion of #6703 

Items to do: 
- [x] Move Subscriptions Management to /payments/management
- [x] Convert content to React
- [x] Update design
- [x] Write tests

mindy:
**Changes:** 
- Move location from /payments/recurring/management to /payments/management/
- Re-style page with updated confirmation message
- Add function to format dates 
- Organize API routes on front-end by consolidating to one file 
- Extract Payments error message to separate component so other pages can use

**To test:**
- Go to http://localhost.org:8000/en-US/payments/management/ 
- Try canceling a subscription and signing up again. 

Closes #6896


package.json
```diff
@@ -37,7 +37,7 @@
         "process:sass": "node-sass-chokidar ./kuma/static/styles/ -o ./static/styles -w",
         "start": "run-p process:sass webpack:dev",
         "test": "jest --testPathIgnorePatterns tests/sass",
-        "test-ci": "JEST_JUNIT_OUTPUT_DIR='./test_results' JEST_JUNIT_OUTPUT_NAME='jest.xml' jest --ci --reporters=default --reporters=jest-junit --testPathIgnorePatterns tests/sass",
+        "test-ci": "JEST_JUNIT_OUTPUT_DIR='./test_results' JEST_JUNIT_OUTPUT_NAME='jest.xml' NODE_ICU_DATA=node_modules/full-icu jest --ci --reporters=default --reporters=jest-junit --testPathIgnorePatterns tests/sass",
```
mindy, line 40: The tests are passing locally, but in Docker these two are failing, I think because there is no `Intl` support in the Docker container:
```
FAIL kuma/javascript/src/payments/formatDate.test.js
  ● formatDate() › formats date according to locale
    expect(received).toEqual(expected) // deep equality
    Expected: "22 de maio de 2020"
    Received: "May 22, 2020"
      10 |         const mockDate = 1590149080000;
      11 |         const expected = '22 de maio de 2020';
    > 12 |         expect(formatDate('pt-PT', mockDate)).toEqual(expected);
         |                                               ^
      13 |     });
      14 |     it('formats date according to options', () => {
      15 |         const mockDate = '2020-05-23T08:04:40';
      at Object.<anonymous> (kuma/javascript/src/payments/formatDate.test.js:12:47)
  ● formatDate() › formats date according to options
    expect(received).toEqual(expected) // deep equality
    Expected: "05.20"
    Received: "05/20"
      19 |         };
      20 |         const expected = '05.20';
    > 21 |         expect(formatDate('tr', mockDate, mockOptions)).toEqual(expected);
         |                                                         ^
      22 |     });
      23 |     it('returns empty string for invalid date', () => {
      24 |         const mockDate = '11/9999';
      at Object.<anonymous> (kuma/javascript/src/payments/formatDate.test.js:21:57)
```
I added a devDependency `full-icu` to fix it, but it seems like there could be a better solution. I don't like having to add a package just to get tests to pass on Docker-- is there a way to add `Intl` support to the container somehow?

kuma/users/stripe_utils.py
```diff
@@ -47,13 +47,14 @@ def retrieve_and_synchronize_subscription_info(user):
...
-                "amount": stripe_subscription_info.plan.amount,
+                # Amount is in cents, so divide by 100
+                "amount": stripe_subscription_info.plan.amount / 100,
                 "brand": card.brand,
                 "expires_at": f"{card.exp_month}/{card.exp_year}",
                 "last4": card.last4,
                 # Cards that are part of a "source" don't have a zip
                 "zip": card.get("address_zip", None),
-                # TODO: Deprecated. Only used in the Edit Profile view
+                # TODO: Deprecated. Only used in the Edit Profile and Payments Management view
```
peterbe, line 57: Then it's no longer deprecated.

peterbe: Remove the comment entirely?

kuma/users/stripe_utils.py
```diff
@@ -101,8 +102,13 @@ def cancel_stripe_customer_subscriptions(user):
     customer = stripe.Customer.retrieve(user.stripe_customer_id)
     for sub in customer.subscriptions.data:
         s = stripe.Subscription.retrieve(sub.id)
-        UserSubscription.set_canceled(user, s.id)
-        s.delete()
+        try:
+            UserSubscription.set_canceled(user, s.id)
+            s.delete()
```
peterbe, line 107: First of all, since this isn't wrapped in an atomic transaction what's going to happen is that `UserSubscription.set_canceled(user, s.id)` is called and fine but the next line doesn't actually do anything. 
The simplest solution, from a quick skim, is to move `UserSubscription.set_canceled(user, s.id)` to be right before the `return True`

package.json
peterbe: This seems worthwhile to fix. So you can't use `new Intl()` inside the docker container??
What does Google have to say on this?

kuma/users/stripe_utils.py
```diff
@@ -47,13 +47,14 @@ def retrieve_and_synchronize_subscription_info(user):
 
             subscription_info = {
                 "id": stripe_subscription_info.id,
-                "amount": stripe_subscription_info.plan.amount,
+                # Amount is in cents, so divide by 100
+                "amount": stripe_subscription_info.plan.amount / 100,
```
peterbe, line 51: Can we move this to the view layer instead? Sending floating-point numbers via JSON is bad because it's unpredictable. Integers are not. I'd rather we communicate in cents and if we prefer to display it in dollars, we make that a formatting thing in our React code.

kuma/payments/urls.py
```diff
@@ -5,10 +5,6 @@
 lang_urlpatterns = [
     path("terms/", views.payment_terms, name="payment_terms"),
     path("thank-you/", views.thank_you, name="thank_you"),
-    path(
-        "recurring/management/",
-        views.recurring_payment_management,
-        name="recurring_payment_management",
-    ),
```
peterbe, line 12: I guess this is for me to fix in https://github.com/mdn/kuma/issues/6898 
Let's just not forget to do it as soon as we land this PR.

kuma/payments/jinja2/payments/management.html
```diff
@@ -1,125 +1,23 @@
...
-                            please contact <a href="{{ url }}">{{ email }}</a>
-                            and provide your customer reference number {{ reference_number }}.
-                        {% endtrans %}
-                    </b>
-                </div>
-            {% endif %}
-        {% endif %}
-    </div>
+{% block document_head %}
+    {{ render_react("SPA", request.LANGUAGE_CODE, request.get_full_path(), True)|safe }}
```
peterbe, line 17: No need for the support email on the management page??

kuma/javascript/src/payments/routes.jsx
```diff
@@ -24,6 +26,8 @@ export function PaymentPage(props: PaymentPageProps) {
     const { locale, slug, data } = props;
     const getPage = () => {
         switch (true) {
+            case slug.includes(PAYMENT_PATHS.MANAGEMENT):
+                return <ManagementPage locale={locale} />;
```
peterbe, line 30: Nice!

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+type Props = {
+    locale: string,
+};
+export const title = gettext('Manage monthly subscriptions');
+export const successMsg = gettext(
+    'Your monthly subscription has been successfully canceled.'
+);
+
+const ManagementPage = ({ locale }: Props) => {
```
peterbe, line 31: I don't yet know what's causing this but there's definitely something strange going on with the loading order on this page. 

This is what I get when I refresh the page. 
Screen-Recording-2020-04-24-at-2 38 22-PM

kuma/javascript/src/payments/formatDate.js
```diff
@@ -0,0 +1,44 @@
...
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+    }
+): string => {
+    const dateObj = new Date(date);
+    if (dateObj instanceof Date && isNaN(dateObj.valueOf())) {
+        return '';
+    }
+    return new Intl.DateTimeFormat(locale, options).format(dateObj);
```
peterbe, line 43: So is this what's behaving totally different in Node in Docker vs Node on your host machine??

kuma/javascript/src/payments/formatDate.js
```diff
@@ -0,0 +1,44 @@
...
+export const formatDate = (
+    locale: string,
+    date: Date | number | string,
+    options?: DateTimeFormatOptions = {
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+    }
+): string => {
+    const dateObj = new Date(date);
```
peterbe, line 39: If `date` can be a `Date` object, why convert it a second time? I'm referring to the flow type on line 32.

kuma/javascript/src/payments/formatDate.js
```diff
@@ -0,0 +1,44 @@
...
+    if (dateObj instanceof Date && isNaN(dateObj.valueOf())) {
+        return '';
```
peterbe, line 41: This is strange. If you passed in something so invalid/bad that it fails to make that type conversion you should probably throw an error instead of masking to an empty string.

kuma/javascript/src/payments/components/error-message.jsx
```diff
@@ -0,0 +1,21 @@
...
+        id={gettext(
+            "We're sorry, something went wrong. Please contact <emailLink />."
+        )}
+        emailLink={
+            <a
+                target="_blank"
+                rel="noopener noreferrer"
+                href={`mailto:${window.mdn.contributionSupportEmail}`}
+            >
+                {gettext(window.mdn.contributionSupportEmail)}
```
peterbe, line 15:
```suggestion
                {window.mdn.contributionSupportEmail}
```
because we don't want our localizers to localize an email address.

kuma/javascript/src/payments/api.js
```diff
@@ -0,0 +1,87 @@
...
+        body,
+        headers: {
+            'Content-Type': 'application/json',
+            'X-CSRFToken': getCookie('csrftoken'),
+        },
+    })
+        .then((res) => {
+            if (!res.ok) {
+                throw new Error(
+                    `${res.status} - ${res.statusText} posting ${SUBSCRIPTIONS_FEEDBACK_URL}`
```
peterbe, line 78: Nice!

kuma/javascript/src/payments/api.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/**
+ * Submit user feedback to Google Analytics
+ *
+ * @param {{ "feedback": string }} body - body of request, data type JSON
+ * @param {function(Object)} onSuccess - The callback that handles a successful response.
+ * @param {function(string)} onError - The callback that handles an error.
+ */
+export function sendFeedback(body, onSuccess, onError) {
```
peterbe, line 66: Perhaps I'm a hipster but have you considered using async/await instead? You can still use the callbacks. E.g.

```javascript
export async function sendFeedback(body, onSuccess, onError) {
    try {
        const res = await fetch(...);
        if (!res.ok) {
           throw new Error(...)
        }
        onSuccess(await res.json())

    } catch(err) {
        onError(err);
    }
}
```

kuma/javascript/src/payments/api.js
```diff
@@ -0,0 +1,87 @@
...
+        );
+}
+
+/**
+ * Cancel all user's subscriptions
+ *
+ * @param {function(Object):void} onSuccess - The callback that handles a successful response.
+ * @param {function(string):void} onError - The callback that handles an error.
+ */
+export function deleteSubscriptions(onSuccess, onError) {
```
peterbe, line 38: Ditto about using async/await. (see comment below)

kuma/javascript/src/payments/formatDate.js
mindy: Thanks for the suggestion. I changed this so it throws an error instead.

kuma/javascript/src/payments/api.js
mindy: Updated these to use async/await

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,188 @@
...
+    const [subscription, setSubscription] = useState<?SubscriptionData>(null);
+    const [isSubscriber, setIsSubscriber] = useState<?boolean>(null);
+    const userData: ?UserData = useContext(UserProvider.context);
+
+    // Currently we don't have a way to update the UserProvider context, so
+    // we are saving the context value `isSubscriber` to local state. It will be
+    // out-of-sync with context upon a successful delete, but I think it's ok
+    // for now, since actions after deleting are limited-- nothing builds upon
+    // `isSubscriber`. The only thing you can do from this point is go to
+    // /payments which will refresh the UserProvider context.
```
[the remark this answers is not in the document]
mindy: Thanks Peter, I adopted your suggestion! 

Looking forward, I'd still prefer if we had a way to update context so it accurately reflects `isSubscriber`. I looked into it briefly but it would be a change bigger than I feel comfortable with in this PR, since `UserProvider` is used in many components. Probably not worth it just for this one page-- maybe one day if we find we need to update context in more than one place.

mindy:
> Thing is `userData` will already have `userData.isSubscriber` and if that's falsy there's no point fetching subscription data.

That's how it's currently set up-- it will not fetch subscription data if `isSubscriber` is falsey.

babel.config.js
```diff
@@ -32,6 +32,7 @@ module.exports = (api) => {
     ];
     const plugins = [
         path.join(nodePath, '@babel/plugin-proposal-class-properties'),
+        path.join(nodePath, '@babel/plugin-transform-runtime'),
```
mindy, line 35: In order to use `async` and `await`, I had to add this plugin, following Babel's recommendation https://babeljs.io/docs/en/babel-plugin-transform-runtime. 

Another way to fix it, which is how I fixed it in the tests, was to include `regenerator-runtime` directly in the file (https://www.npmjs.com/package/regenerator-runtime), but I read a few posts that said the Babel way was preferred to avoid polluting the global scope. 

Note: I originally incorrectly thought the `regenerator-runtime` error was caused by using `fetch` and Jest, but it turned out to be an issue with using `async` and `await`.

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
```diff
@@ -0,0 +1,75 @@
...
+            <p className="alert error" data-testid="error-msg">
+                <ErrorMessage />
+            </p>
+        );
+    }
+
+    return (
+        <form onSubmit={handleSubmit}>
+            <div>
+                <strong>{title}</strong>
```
schalkneethling, line 47: Consider using an `h4` or `h5` as appropriate here instead of `strong`.

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
```diff
@@ -0,0 +1,75 @@
...
+                <p>
+                    {interpolate(
+                        'Your monthly subscription will end on %(date)s. You will have to set up a new subscription if you wish to resume making payments to MDN Web Docs.',
```
schalkneethling, line 50: Do we have this scenario still? I mean a subscription end/expiration date?

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
```diff
@@ -0,0 +1,75 @@
...
+                        { date }
+                    )}
+                </p>
+                <div className="form-footer">
+                    <button
+                        type="button"
+                        className="cta keep-membership"
```
schalkneethling, line 57: Could we instead use the `.positive` class here?

kuma/javascript/src/payments/pages/index.jsx
```diff
@@ -263,7 +263,7 @@ export default function LandingPage({ data, locale }: RouteComponentProps) {
                                     subscriptionsLink={
                                         <a href={urls.managePayments}>
                                             {gettext(
-                                                'manage monthly subscription page'
+                                                'manage monthly subscriptions page'
```
schalkneethling, line 266: Maybe it is just me 😸 but "subscriptions" suggests that you can have more than one? Perhaps "subscription" is more appropriate in this case?

kuma/javascript/src/payments/pages/index.jsx
```diff
@@ -291,7 +291,7 @@ export default function LandingPage({ data, locale }: RouteComponentProps) {
...
```
schalkneethling, line 294: same comment here as above

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,200 @@
...
+                <p>
+                    {interpolate(
+                        'Next payment occurs on %(nextPaymentDate)s.',
+                        {
+                            nextPaymentDate,
+                        }
+                    )}
+                </p>
+                <div className="active-subscriptions">
+                    <ul>
```
schalkneethling, line 111: Do we need this wrapped in a `div`? Could we instead simply add the `active-subscriptions` class directly to the `ul`?

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,200 @@
...
+                        <li>
+                            <span className="label amount">
+                                {gettext('Amount')}
+                            </span>
+                            <span className="value">
+                                {/* amount is in cents, so divide by 100 to get dollars */}
+                                {formatMoney(locale, subscription.amount / 100)}
+                            </span>
```
schalkneethling, line 119: This looks like a good candidate for a description list. You can then have something like:

```
<dl className="active-subscriptions">
  <dt className="label amount">{gettext('Amount')}</dt>
  <dd className="value">
    {/* amount is in cents, so divide by 100 to get dollars */}
    {formatMoney(locale, subscription.amount / 100)}
  </dd>
...
</dl>
```

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,200 @@
...
+                return (
+                    <Interpolated
+                        id={gettext('Please <signInLink /> to continue.')}
+                        signInLink={<SignInLink text={gettext('sign in')} />}
+                    />
+                );
+            case !userData.isSubscriber || canceled:
+                // Not a subscriber or just canceled their subscription
+                return renderInvitation();
+            case userData.isSubscriber && !!subscription:
```
schalkneethling, line 168: Just curious here. Could you be subscriber and not have any subscriptions?

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,200 @@
...
+                        setShowForm={setShowForm}
+                        date={lastActiveDate}
+                    />
+                )}
+            </>
+        );
+    };
+
+    const renderContent = () => {
+        switch (true) {
```
schalkneethling, line 156: Another curiosity question :) This is an interesting use of the `switch` statement, is there a specific benefit to this over using `if/else`

kuma/javascript/src/payments/pages/management.jsx
```diff
@@ -0,0 +1,200 @@
...
+
+    return (
+        <>
+            <Subheader title={title} />
+            <main
+                className="contributions-page manage-subscriptions"
+                role="main"
+                data-testid="management-page"
+            >
+                <section>
```
schalkneethling, line 188: If there is no styling attached to this `section` element, I would recommend removing it. We do not get any accessibility or semantic benefit.

kuma/javascript/src/payments/pages/thank-you.jsx
```diff
@@ -42,10 +42,10 @@ const ThankYouPage = ({ locale }: Props) => {
                                     )}
                                     subscriptionLink={
                                         <a
-                                            href={`/${locale}/payments/recurring/management/`}
+                                            href={`/${locale}/payments/management/`}
                                         >
                                             {gettext(
-                                                'manage monthly subscription page'
+                                                'manage monthly subscriptions page'
```
schalkneethling, line 48: Another instance of the "subscriptions" vs "subscription" question.

kuma/javascript/src/signin-link.jsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export default function SignInLink({
+    className,
+    text = gettext('Sign in'),
+}: Props): React.Node {
+    const ga = useContext(GAProvider.context);
+    const locale = getLocale();
+
+    if (typeof window === 'undefined') {
```
schalkneethling, line 20: I believe we used to do this test this way because in the past you could do:

`undefined = "monkey-wrench";` or even `undefined = "";`

In strict mode(which we are in) `undefined` is immutable(non-writable global) and so one can simply do:

```
if (window === undefined)
```

kuma/javascript/src/signin-link.jsx
```diff
@@ -0,0 +1,70 @@
...
+        return null;
+    }
+
+    const LOCATION = window.location.pathname;
+    /**
+     * Send a signal to GA when a user clicks on the Sing In
+     * lnk in the header.
+     * @param {Object} event - The event object that was triggered
```
schalkneethling, line 28: A holdover from a previous version of the function?

kuma/static/styles/components/payments/_page.scss
```diff
@@ -214,4 +196,145 @@
...
+                button:last-child {
+                    margin-bottom: 10px;
+                    order: -1;
+                }
+                @media #{$mq-tablet-and-up} {
+                    flex-direction: row;
+                    justify-content: flex-end;
+                    button:last-child {
+                        margin: 0 0 0 10px;
+                        order: unset;
```
schalkneethling, line 273: `unset` is not supported in IE, use `inherit` instead.

kuma/static/styles/components/payments/_page.scss
```diff
@@ -214,4 +196,145 @@
...
+                display: flex;
+                flex-direction: column;
+                button:last-child {
+                    margin-bottom: 10px;
+                    order: -1;
+                }
+                @media #{$mq-tablet-and-up} {
+                    flex-direction: row;
+                    justify-content: flex-end;
+                    button:last-child {
```
schalkneethling, line 271: Nit: Not sure if prettier did not kick in but, there should be an empty line above this one.

kuma/static/styles/components/payments/_page.scss
```diff
@@ -214,4 +196,145 @@
...
+                        margin: 0 0 0 10px;
+                        order: unset;
+                    }
+                }
+            }
+        }
+    }
+
+    button {
+        box-sizing: border-box;
```
schalkneethling, line 281: Meh, I cannot wait for us to have this defined in the reset for all the things :) Soon...

kuma/static/styles/components/payments/_page.scss
```diff
@@ -214,4 +196,145 @@
...
+                border-color: darken($border-color, 50%);
+                color: darken($text-color, 50%);
+            }
+        }
+    }
+
+    .alert {
+        clear: both;
+        padding: 20px;
+        &.error {
```
schalkneethling, line 329: Nit: empty line above this line.

schalkneethling:
A complete side note. I was looking at something in `incentives.jsx`[1] and noticed that somehow Git seems to think that the only people who worked on that file are @mindy and @peterbe. Any idea why my name has been stricken from the record :)

[1] https://github.com/mdn/kuma/blob/9658817b766654da4a70f7409b609123287fabb8/kuma/javascript/src/payments/components/incentives.jsx

peterbe:
> Any idea why my name has been stricken from the record :)

Curious that!
Your name is all over the [`git blame` view](https://github.com/mdn/kuma/commits/master/kuma/javascript/src/payments/components/incentives.jsx)
But not in the [commits view](https://github.com/mdn/kuma/commits/master/kuma/javascript/src/payments/components/incentives.jsx).

peterbe:
Screen Shot 2020-04-28 at 1 02 28 PM

@schalkneethling brought this up too. I know the API is returning a *list* of subscriptions but we shouldn't be wedded to that. If it's more appropriate to say "subscription" instead of "subscriptions" then I think we should fix it. Even if it means we need to file a new issue. 

Do you have a strong preference @malqinneh @atopal ?

kuma/javascript/src/formatters.js
```diff
@@ -0,0 +1,76 @@
...
+export const formatDate = (
+    locale: string,
+    date: Date | number | string,
+    options?: DateTimeFormatOptions = {
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+    }
+): string => {
+    const dateObj = new Date(date);
```
peterbe, line 51: Now I get it! If `date` already was a `Date` object, doing `new Date(aDateObj)` is harmless. It just returns a new Date object instance.

kuma/javascript/src/signin-link.jsx
```diff
@@ -0,0 +1,70 @@
...
+export default function SignInLink({
+    className,
+    text = gettext('Sign in'),
+}: Props): React.Node {
+    const ga = useContext(GAProvider.context);
+    const locale = getLocale();
+
+    if (typeof window === 'undefined') {
+        return null;
+    }
```
peterbe, line 22: So this is a new thing? How does it relate to subscription management?

Either way, the `Login` component already does this check so `<SigninLink/>` won't be called if `typeof window === 'undefined'` anyway.

atopal:
Yeah, I think we should go with singular, but I don't feel strongly about that.

kuma/javascript/src/payments/pages/management.jsx
peterbe: I'm indifferent. Personally *I* prefer plain `if/else if` blocks but I don't think you should change it if you prefer this style here.

malqinneh:
@peterbe Singular is fine! If Mindy leaves them out - I'd be happy to jump into code and make that update.

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
mindy: Good question. I'm not sure, it was in the updated mockup so I added it. I asked Kadir in Slack and will update with the latest.

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
mindy: Unfortunately not-- the button in the mockup is outlined in dark gray.

mindy: On second thought I'll leave as is-- `light-border-color` seems too light and I didn't see any grays that were a better fit.

kuma/javascript/src/payments/pages/index.jsx
mindy: Fixed and searched for all instances of "monthly subscription" to make sure it's singular (except one case where it didn't apply).

kuma/javascript/src/payments/pages/management.jsx
mindy: Changed-- thank you!

kuma/javascript/src/payments/pages/management.jsx
mindy: Originally I had it as a description list but it was such a headache to style. I will try again tomorrow.

kuma/javascript/src/payments/pages/management.jsx
mindy: I don't believe so. This logic is here because while it's waiting for the request to get subscriptions to complete, it can fall through to show the "Loading" text.

kuma/javascript/src/payments/pages/management.jsx
mindy: I usually use an if/else unless it's more than 3 blocks, then I use a `switch` statement for faster performance (this is what I have read, I've never actually done a speed test comparing the two).

kuma/javascript/src/payments/pages/management.jsx
mindy: There is styling attached to this that adds padding around the edges. I tried to move it to `<main>` but it doesn't look quite right.

kuma/javascript/src/payments/pages/thank-you.jsx
mindy: Changed to singular

kuma/static/styles/components/payments/_page.scss
mindy: Same!

mindy:
> A complete side note. I was looking at something in `incentives.jsx`[1] and noticed that somehow Git seems to think that the only people who worked on that file are @mindy and @peterbe. Any idea why my name has been stricken from the record :)
> 
> [1] https://github.com/mdn/kuma/blob/9658817b766654da4a70f7409b609123287fabb8/kuma/javascript/src/payments/components/incentives.jsx

This happened because I used `git mv` to move files into the `components` folder, which loses the history. I have `git config log.follow true` set in my local git config so I still see the full history locally, but I know that doesn't help this situation when viewing in on GitHub :( 

It has been a requested feature on GitHub though, opened 3 years ago and still active: https://github.com/isaacs/github/issues/900

kuma/javascript/src/signin-link.jsx
mindy: Good point-- removed this. And since the management page is not SSR, window will always be defined.

mindy: This relates to the subscription management page, because when you're not logged in, you see this message: 
Screen Shot 2020-04-28 at 9 25 01 PM

And I wanted to re-use what the sign in link in the header was doing, so I moved it into a separate signin-link component.

kuma/javascript/src/signin-link.jsx
mindy: I imagine we still want to track the sign in click events, but maybe add a different label to differentiate between the header sign in link and the subscriptions management page. I will check with Kadir on this.

kuma/javascript/src/signin-link.jsx
mindy: Good to know! I ended up removing this altogether per Peter's comment below.

kuma/payments/jinja2/payments/management.html
mindy: It's removed from the new mockups-- the only time you would see it is if the error screen renders like so (this is the generic error screen on all payments pages):
Screen Shot 2020-04-28 at 9 44 58 PM

kuma/javascript/src/formatters.js
mindy: > I still don't understand or see what the `SignInLink` has to do with anything and why it does the extrac check of `typeof window`.

I removed the `typeof window` check, but just so you know, I copied it from the original login code. It looked like an important check so I copied it over! https://github.com/mdn/kuma/blob/master/kuma/javascript/src/header/login.jsx#L16

mindy:
The last thing I need to do for this PR is convert this piece to a `<dl />`, will update when it's done. 
Screen Shot 2020-04-28 at 9 50 53 PM

kuma/javascript/src/formatters.js
schalkneethling: 😸 It is for sure an important check, we should probably change the other instance as well :) Not in this PR though ;)

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
schalkneethling: Buttons are next on my list for the design system :)

kuma/javascript/src/payments/pages/management.jsx
schalkneethling: Thank you for the feedback Mindy.

kuma/javascript/src/payments/pages/management.jsx
schalkneethling: I guess what I was asking is whether this can simply be:

```suggestion
            case userData.isSubscriber:
```

kuma/javascript/src/payments/pages/management.jsx
schalkneethling: Something to look into in future. Not worth spending more time on it now. Thank you Mindy!

kuma/javascript/src/signin-link.jsx
schalkneethling: I also ask seeing that the function does not currently accept any arguments.

kuma/static/img/subscriptions/credit-card.svg
```diff
@@ -0,0 +1,3 @@
+<svg class="icon icon-credit-card" xmlns="http://www.w3.org/2000/svg" width="21" height="16" aria-hidden="true">
```
schalkneethling, line 1: We should remove the `width` and `height` and replace with `viewBox` for responsiveness(especially in IE11)

kuma/static/img/subscriptions/money.svg
```diff
@@ -0,0 +1,3 @@
+<svg class="icon icon-money-bill" xmlns="http://www.w3.org/2000/svg" width="23" height="16" aria-hidden="true">
```
schalkneethling, line 1: We should remove the `width` and `height` and replace with `viewBox` for responsiveness(especially in IE11) - Also, here and the one above, we should add `role="img"`

kuma/static/img/subscriptions/credit-card.svg
schalkneethling:
```
<svg class="icon icon-credit-card" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 21 16" aria-hidden="true" role="img"><path d="M18.85 0H1.72C.77 0 0 .77 0 1.71V14.3c0 .93.77 1.7 1.72 1.7h17.13c.95 0 1.72-.77 1.72-1.71V1.7c0-.93-.77-1.7-1.72-1.7zM1.93 1.71h16.7c.13 0 .22.1.22.22v1.5H1.72v-1.5c0-.12.1-.22.21-.22zm16.7 12.58H1.94a.21.21 0 01-.21-.22V8h17.13v6.07c0 .12-.1.22-.21.22zM6.87 10.7v1.43c0 .24-.2.43-.43.43H3.86a.43.43 0 01-.43-.43V10.7c0-.23.2-.42.43-.42h2.57c.23 0 .43.19.43.42zm6.85 0v1.43c0 .24-.19.43-.42.43H8.43a.43.43 0 01-.43-.43V10.7c0-.23.2-.42.43-.42h4.86c.23 0 .42.19.42.42z"/></svg>
```

kuma/javascript/src/formatters.js
peterbe: I think the commenting is getting criss-crossed. 

So it *used* to be this:
```javascript
function Login() {
  const userData = useContext(UserProvider.context);
  if (!userData || typeof window === 'undefined') {
    return null;
  }
  if (userData) {
    return <img src={userData.mugshot}>
  } else {
    return <a href="signin" onClick={openModal}>Sign in</a>
  }
```
Then you extracted the `return <a href="signin" onClick={openModal}>Sign in</a>` into its own new component and inside that one you did the check on `typeof window !== 'undefined'` **again**. 

I think that confused me. 
Also, it confused me why you refactored `<Login/>` but you, in some other comment, explained that you wanna reuse `<SigninLink/>` in more places than just within `<Login/>`.

All good here.

kuma/javascript/src/payments/components/cancel-subscription-form.jsx
mindy: Updated the wording: "Your monthly subscription will end on %(date)s, but if you cancel now it will end immediately."

kuma/javascript/src/signin-link.jsx
mindy: Kadir responded and said we can keep the same click events.

kuma/javascript/src/payments/pages/management.jsx
mindy: If we change it to that, then it will no longer show the "Loading..." text for a few seconds while it waits for the subscriptions call to complete, so this is what you'll see:
Screen Shot 2020-04-29 at 11 44 02 AM

mindy:
Ready for another 👀 @schalkneethling. Thank you!

PR #6960, opened 2020-04-27
Make breadcrumbs mobile and tablet-friendly

### LINKED ISSUE #6822: T - Implement mobile-optimized breadcrumbs.
See User Story issue for comps and description: #6784.

# Acceptance Criteria 
- [ ] New attribute-based breadcrumb implemented on mobile (only)

mindy:
**Changes**
- On mobile and tablet screens, hide all but the immediate parent crumb 
- On mobile and tablet screens, breadcrumbs appear above title
- New arrow styles between breadcrumbs
- Add Sass mixin for caret-like arrows 

**To test**
Go to http://localhost.org:8000/en-US/docs/Web/HTML/Element/article and view breadcrumbs on various screen sizes.

**Before:**
_Desktop_
before-desktop

_Mobile_
before-mobile

**After:** 
_Desktop_
after-desktop

_Mobile_
after-mobile

Closes #6822

### DIFF
--- a/kuma/javascript/src/breadcrumbs.jsx
+++ b/kuma/javascript/src/breadcrumbs.jsx
@@ -1,7 +1,7 @@
 // @flow
 import * as React from 'react';
 import { useContext } from 'react';
-
+import { gettext } from './l10n.js';
 import GAProvider from './ga-provider.jsx';
 
 import type { DocumentData } from './document.jsx';
@@ -54,6 +54,9 @@ export default function Breadcrumbs({ document }: DocumentProps) {
                             onClick={sendBreadcrumbItemClick}
                             onContextMenu={sendBreadcrumbItemClick}
                         >
+                            <span className="pre-text">
+                                {gettext('See')}&nbsp;
+                            </span>
                             <span property="name">{p.title}</span>
                         </a>
                         <meta property="position" content={i + 1} />
--- a/kuma/javascript/src/breadcrumbs.jsx
+++ b/kuma/javascript/src/breadcrumbs.jsx
@@ -36,7 +36,7 @@ export default function Breadcrumbs({ document }: DocumentProps) {
                 vocab="https://schema.org/"
                 aria-label="breadcrumbs"
             >
-                {document.parents.map((p, i) => (
+                {document.parents.map((p, i, { length }) => (
                     <li
                         key={p.url}
                         property="itemListElement"
@@ -44,7 +44,11 @@ export default function Breadcrumbs({ document }: DocumentProps) {
                     >
                         <a
                             href={p.url}
-                            className="breadcrumb-chevron"
+                            className={
+                                i === length - 1
+                                    ? 'breadcrumb-previous'
+                                    : 'breadcrumb-chevron'
+                            }
                             property="item"
                             typeof="WebPage"
                             onClick={sendBreadcrumbItemClick}
--- a/kuma/javascript/src/document.jsx
+++ b/kuma/javascript/src/document.jsx
@@ -100,13 +100,15 @@ function DocumentPage({ document }: DocumentProps) {
             <A11yNav />
             <Header document={document} />
             <main role="main">
-                <Titlebar title={document.title} document={document} />
-                <div className="full-width-row-container">
-                    <div className="max-content-width-container">
-                        <Breadcrumbs document={document} />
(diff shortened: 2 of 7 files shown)


kuma/static/styles/minimalist/utils/_utility-mixins.scss
```diff
@@ -10,3 +10,34 @@
...
+    clip: initial;
+    height: initial;
+    margin: initial;
+    overflow: initial;
+    padding: initial;
+    position: initial !important; /* stylelint-disable-line declaration-no-important */
+    width: initial;
+}
+
+@mixin caret-arrow($color, $size, $direction: right, $thickness: 2px) {
```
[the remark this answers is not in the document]
mindy: I like this format! Thanks Schalk

mindy:
Ready for another 👀 @schalkneethling

kuma/javascript/src/document.jsx
```diff
@@ -100,13 +100,15 @@ function DocumentPage({ document }: DocumentProps) {
...
+                <header>
+                    <Titlebar title={document.title} document={document} />
+                    <div className="full-width-row-container">
+                        <div className="max-content-width-container">
+                            <Breadcrumbs document={document} />
+                            <LanguageMenu document={document} />
+                        </div>
                     </div>
-                </div>
+                </header>
```
schalkneethling, line 111: I am torn about whether we should mark this up as a `header`. I kinda feel like in this case, the header is the section before this i.e. the section that contains the logo, navigation, authentication etc.

I am thinking we should instead mark this up with something like:

```
<div class="subheader">
...
</div>
```

What do you think?

kuma/static/styles/minimalist/components/_breadcrumbs.scss
```diff
@@ -17,17 +19,65 @@
...
+        .breadcrumb-previous {
+            &::before {
+                @include caret-arrow($blue, left);
+                margin: 0 10px 0 0;
+                position: relative;
+                top: -1px;
+            }
+            span::before {
+                display: inline-block;
+                content: 'See\00a0';
```
schalkneethling, line 31: I am not entirely sure why we have this. I know it is per design but, it seems a little odd to me. for one thing, if this is important to provide context, we should not add this via generated content as it will not be accessible[1].

If it is not important, I would vote to not include it.

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/content#Accessibility_concerns

schalkneethling: @malqinneh Thoughts?

malqinneh: It's to communicate to users that the breadcrumb label is meant for a section that they are *not* currently on. If that makes sense.

kuma/javascript/src/document.jsx
mindy: I hear you, but there can be multiple headers per page, e.g. article headers. I vote for leaving it as a header.

kuma/static/styles/minimalist/components/_breadcrumbs.scss
mindy: In that case, I can move the "See " text into the JSX file and hide it for desktop.

kuma/javascript/src/breadcrumbs.jsx
```diff
@@ -36,20 +36,27 @@ export default function Breadcrumbs({ document }: DocumentProps) {
...
+                                i === length - 1
+                                    ? 'breadcrumb-previous'
+                                    : 'breadcrumb-chevron'
+                            }
                             property="item"
                             typeof="WebPage"
                             onClick={sendBreadcrumbItemClick}
                             onContextMenu={sendBreadcrumbItemClick}
                         >
+                            <span className="pre-text">
```
mindy, line 57: I was concerned about disrupting the breadcrumb schema, so I wanted to leave `<span property="name">` as is.

mindy:
Ready for another look @schalkneethling. Thank you!

kuma/static/styles/minimalist/components/_breadcrumbs.scss
schalkneethling: Thank you Mindy and Mustafa

PR #6966, opened 2020-04-29
fix robots and endpoints headless integration tests

escattone:
Fixes https://github.com/mdn/kuma/issues/6918

Already tested against stage and prod using:
```python
docker-compose exec web pytest --base-url https://developer.allizom.org tests
docker-compose exec web pytest --base-url https://developer.mozilla.org tests
```
so with this PR all integration tests pass again.


escattone:
Since it has already been tested against stage and prod, it's not worth wasting @Gregoor's time reviewing this.

PR #6967, opened 2020-04-29
Add command for exporting subscribers to sendinblue

### LINKED ISSUE #6933: T - Prototype sendinblue marketing email integration.
**Description**

We still do not have a finalized sendinblue contract but would like to get a head start by prototyping some of the functionality that we know that we'll need.

**Acceptance Criteria**
- [x] import TEST users (do not use actual MDN users for this due to privacy concerns) from Kuma into sendinblue
- [x] filter Kuma TEST users into sendinblue as 2 audiences (2 groups: paying subscribers and non-subscribers)

Gregoor:
Fixes #6933 

## What it does
Adds a command to export users who've opted-in to the "new" newsletter signup checkbox into sendinblue. It's currently done in two requests, since we want to bucket users into paying and non-paying lists. The docs for the API used are here: https://developers.sendinblue.com/reference#importcontacts-1

~I chose to not add the API key to our config files it yet, as we're still in the early testing phase of this. So in the future this command might have no arguments, but for now it has three (more below).~

Also there's a limit to the import CSV string size, which is ~8MB. My napkin math says we'll reach that once we try to import 400k users at once. Currently we're only importing the once who've checkedthe fairly new (signup-only) opt-in. So I opted for not complicating the code for the 400k case. My expectation is that we'll have proper syncing set-up before reach that.

~Oh and I went with raw SQL since I spent half an hour reading docs trying to find out how to do what I already knew how to do in SQL. Frustrated I gave up and just went raw. Happy to do the Django thing if anyone prefers that (and has some guidance for me).~

## How to test it
1. Create a sendinblue account over at https://www.sendinblue.com (you can skip a lot of the profile set-up, look for skip in the upper right)
1. Get your v3 API key at https://account.sendinblue.com/advanced/api
1. Create two lists at https://my.sendinblue.com/lists/new/parent_id/1 - one for paying and one for non-paying users
1. Add the sendinblue config keys to your .env (see below for key naming)
1. Run the following and check after that you receive emails informing you about successful imports (might take a min)
https://github.com/mdn/kuma/blob/fa26f1861e0d089b3c2ba6a7c3195e17d0d8761e/kuma/settings/common.py#L1844-L1848
```bash
docker-compose exec web ./manage.py export_to_sendinblue
```

### DIFF
--- a/kuma/users/checks.py
+++ b/kuma/users/checks.py
@@ -2,10 +2,13 @@ import stripe
 from django.conf import settings
 from django.core.checks import Error
 
+from . import sendinblue
 from .stripe_utils import create_missing_stripe_webhook
 
 STRIPE_PLAN_ERROR = "kuma.users.E001"
 STRIPE_PLAN_INACTIVE_ERROR = "kuma.users.E002"
+
+SENDINBLUE_LIST_ERROR = "kuma.users.E003"
 
 
 def stripe_check(app_configs, **kwargs):
@@ -23,20 +26,44 @@ def stripe_check(app_configs, **kwargs):
         if not plan.active:
             errors.append(
                 Error(
-                    f"{settings.STRIPE_PLAN_ID} points to an inactive Strip plan",
+                    f"{settings.STRIPE_PLAN_ID} points to an inactive Stripe plan",
                     id=STRIPE_PLAN_INACTIVE_ERROR,
                 )
             )
     except stripe.error.StripeError as error:
         errors.append(
-            Error(f"unable to retrieve stripe plan: {error}", id=STRIPE_PLAN_ERROR)
+            Error(f"unable to retrieve Stripe plan: {error}", id=STRIPE_PLAN_ERROR)
         )
 
     try:
         create_missing_stripe_webhook()
     except stripe.error.StripeError as error:
         errors.append(
-            Error(f"unable get or create strip webhooks: {error}", id=STRIPE_PLAN_ERROR)
+            Error(
+                f"unable get or create Stripe webhooks: {error}", id=STRIPE_PLAN_ERROR
+            )
         )
+
+    return errors
+
+
+def sendinblue_check(app_configs, **kwargs):
+    errors = []
+
+    if not settings.SENDINBLUE_API_KEY:
+        return errors
+
+    for id in [
+        settings.SENDINBLUE_PAYING_LIST_ID,
+        settings.SENDINBLUE_NOT_PAYING_LIST_ID,
+    ]:
+        response = sendinblue.request("GET", f"contacts/lists/{id}")
+        if not response.ok:
+            errors.append(
+                Error(
+                    f"Error when checking sendinblue list #{id}: {response.json()['message']}",
+                    id=SENDINBLUE_LIST_ERROR,
+                )
+            )
 
     return errors
--- a/kuma/users/management/commands/export_to_sendinblue.py
+++ b/kuma/users/management/commands/export_to_sendinblue.py
@@ -28,14 +28,15 @@ class Command(BaseCommand):
         non_paying_users = []
         for user in users.only("email", "first_name", "last_name"):
             is_paying = user.id in active_subscriber_user_ids
-            list = paying_users if is_paying else non_paying_users
-            list.append(
-                {
-                    "EMAIL": user.email,
-                    "FIRSTNAME": user.first_name,
-                    "LASTNAME": user.last_name,
-                }
-            )
+            row = {
+                "EMAIL": user.email,
+                "FIRSTNAME": user.first_name,
+                "LASTNAME": user.last_name,
+            }
+            if is_paying:
+                paying_users.append(row)
+            else:
+                non_paying_users.append(row)
 
         def export_users(users, list_id):
             csv_out = StringIO()
@@ -58,7 +59,9 @@ class Command(BaseCommand):
                 "content-type": "application/json",
             }
 
-            response = requests_retry_session().request("POST", URL, json=payload, headers=headers)
+            response = requests_retry_session().request(
+                "POST", URL, json=payload, headers=headers
+            )
             response.raise_for_status()
 
         print(f"Exporting {len(paying_users)} paying user(s)")
--- a/kuma/users/management/commands/export_to_sendinblue.py
+++ b/kuma/users/management/commands/export_to_sendinblue.py
@@ -4,7 +4,8 @@ from io import StringIO
 
 import requests
 from django.core.management.base import BaseCommand
-from django.db import connection
+
+from kuma.users.models import User, UserSubscription
 
 URL = "https://api.sendinblue.com/v3/contacts/import"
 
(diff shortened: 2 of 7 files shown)


peterbe:
Have you tried that SQL on a prod dump? There are over 7k users with `is_newsletter_subscribed` true. I'm less familiar with how MySQL would cope with sub-selects like that but it could be too expensive to be safe. By the way, I have another idea.

kuma/users/management/commands/export_to_sendinblue.py
```diff
@@ -0,0 +1,89 @@
...
+                last_name,
+                EXISTS(
+                    SELECT 1
+                    FROM users_usersubscription
+                    WHERE user_id = auth_user.id AND canceled IS NULL
+                ) AS is_paying
+            FROM auth_user
+            WHERE email <> '' AND is_newsletter_subscribed
+        """
+        )
```
peterbe, line 42: An extremely simple solution would be this:
```python
active_subscriber_user_ids = set(
    UserSubscription.objects.filter(canceled__isnull=True).values_list('user_id', flat=True)
)
users = User.objects.filter(is_newsletter_subscribed=True).exclude(email='')
for user in users.only('email', 'first_name', 'last_name'):
    is_paying = user.id in active_subscriber_user_ids
    print(user.email, is_paying)
```

Since the `active_subscriber_user_ids` set is unlikely to be in the tens of thousands, it's fine to hold it all in memory. 
The rest is about as fast and memory-efficient as can possibly be.

kuma/users/management/commands/export_to_sendinblue.py
```diff
@@ -0,0 +1,89 @@
...
+                "accept": "application/json",
+                "content-type": "application/json",
+            }
+
+            response = requests.request("POST", URL, data=payload, headers=headers)
+            response.raise_for_status()
+
+        print(f"Exporting {len(paying_users)} paying user(s)")
+        export_users(paying_users, options["paying_users_list_id"])
+        print(f"Exporting {len(non_paying_users)} non-paying user(s)")
```
peterbe, line 88:
```suggestion
        print(f"Exporting {len(non_paying_users)} non-paying users")
```
:)

Gregoor:
> 
> 
> Have you tried that SQL on a prod dump? There are over 7k users with `is_newsletter_subscribed` true. I'm less familiar with how MySQL would cope with sub-selects like that but it could be too expensive to be safe. By the way, I have another idea.

Not the exact query, since I haven an older dump but I ran a similar (arguably more expensive) query for all 400k users I have, namely:
```mysql
select
   id,
   exists(select 1 from wiki_revision where creator_id = auth_user.id)
from auth_user```
which took ~3sec

File: kuma/users/management/commands/export_to_sendinblue.py (hunk 1 above)
Comment (Gregoor) [reply kuma/users/management/commands/export_to_sendinblue.py:42]: neat, thanks for django-fying it, that's definitely better than the things I came up with. You and @escattone  and most django-istas would find that preferable over the raw SQL, right?

Comment (peterbe) [reply kuma/users/management/commands/export_to_sendinblue.py:42]: Any day!
Raw SQL should be avoided at all costs since the ORM makes us agnostic. However, easier said than done. 

Also, your command is run once and then cleaned up. But when you use `connection.cursor()` you should use a context manager so it closes itself.

```python
    cursor.execute(sql_string)
```

peterbe:
By the way, I **much** appreciate your "honesty" of A) seeking a prototype instead of spending too much time trying to figure it out and B) explaining this reasoning in the PR description.

Gregoor:
Thank you for always reaching out a helping hand! I'll try to keep the habit up.

And to my own surprise I had to realize that I actually like the djangoified version as well. Maybe even more. My SQL bias sits too deep for me to have a real honest answer to it just yet 😁

kuma/users/management/commands/export_to_sendinblue.py
```diff
@@ -0,0 +1,77 @@
...
+
+import requests
+from django.core.management.base import BaseCommand
+
+from kuma.users.models import User, UserSubscription
+
+URL = "https://api.sendinblue.com/v3/contacts/import"
+
+
+def _dictfetchall(cursor):
```
peterbe, line 13: You can delete this now.

kuma/users/management/commands/export_to_sendinblue.py
```diff
@@ -0,0 +1,77 @@
...
+                },
+                separators=(",", ":"),
+            )
+            headers = {
+                "api-key": options["api_key"],
+                "accept": "application/json",
+                "content-type": "application/json",
+            }
+
+            response = requests.request("POST", URL, data=payload, headers=headers)
```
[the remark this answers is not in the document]
Gregoor: That indeed works, thanks!

kuma/users/management/commands/export_to_sendinblue.py
```diff
@@ -0,0 +1,77 @@
...
+            UserSubscription.objects.filter(canceled__isnull=True).values_list(
+                "user_id", flat=True
+            )
+        )
+        users = User.objects.filter(is_newsletter_subscribed=True).exclude(email="")
+        paying_users = []
+        non_paying_users = []
+        for user in users.only("email", "first_name", "last_name"):
+            is_paying = user.id in active_subscriber_user_ids
+            list = paying_users if is_paying else non_paying_users
```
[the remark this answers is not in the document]
Gregoor: good point and imo more than a nit!

Gregoor:
Actually I can't find the reasons in me for only running it one time, once it's in prod. As long as the checkboxes for per-user (un)subscribing properly work, re-running this command should effectively be a no-op (i.e. adding contacts to lists in which they already exists, which the import API turns into a merge without a change).
Now in theory there could be a race condition if a user un/subscribes in between the command querying for user data and sending the requests to sendinblue. I'm not sure if this is likely enough to be worth worrying about though 🤔

peterbe:
So, we merge it into master and delete it from master once it’s been run and the incremental sync is working?

Gregoor:
I felt this was a good place to add the initial pieces for the other upcoming sendinblue PRs. Now this PR includes config keys for sendinblue, a system check and its own module. Hope you're up for a re-review @peterbe. I've updated the initial post with new instructions for how to run it.

kuma/users/checks.py
```diff
@@ -23,20 +26,44 @@ def stripe_check(app_configs, **kwargs):
...
+    ]:
+        response = sendinblue.request("GET", f"contacts/lists/{id}")
+        if not response.ok:
+            errors.append(
+                Error(
+                    f"Error when checking sendinblue list #{id}: {response.json()['message']}",
+                    id=SENDINBLUE_LIST_ERROR,
+                )
+            )
+
```
peterbe, line 68: I only typed in the API key and one of the list IDs. Silly me. This happened:
Screen Shot 2020-05-07 at 8 30 03 AM

Why is it complaining about `Listfoo1`? After all, I did create *one* of them:
Screen Shot 2020-05-07 at 8 31 04 AM

Gregoor:
TODO: add some docs

PR #6969, opened 2020-04-29
Fix Jest error in user-provider test

mindy:
**Issue:** 
Running `jest` gives you this error in `user-provider.test.js`: 
```
  console.error node_modules/react-test-renderer/cjs/react-test-renderer.development.js:120
    Warning: An update to UserProvider inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
        in UserProvider
        in GAProvider
```

**Changes:**
Update test so it does not cause this error.

**To test:**
Run `jest` and ensure that no `console.errors` are logged and tests pass.

### DIFF
--- a/kuma/javascript/src/user-provider.test.js
+++ b/kuma/javascript/src/user-provider.test.js
@@ -1,10 +1,13 @@
 //@flow
 /* eslint-disable camelcase */
 import React from 'react';
-import { act, create } from 'react-test-renderer';
+import { render, waitFor } from '@testing-library/react';
 
 import GAProvider from './ga-provider.jsx';
 import UserProvider from './user-provider.jsx';
+
+// This can be removed once https://github.com/mdn/kuma/pull/6940 lands
+require('regenerator-runtime/runtime');
 
 describe('UserProvider', () => {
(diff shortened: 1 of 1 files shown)


kuma/javascript/src/user-provider.test.js
```diff
@@ -1,11 +1,14 @@
...
 /* eslint-disable camelcase */
 import React from 'react';
-import { act, create } from 'react-test-renderer';
+import { render, waitFor } from '@testing-library/react';
 
 import GAProvider from './ga-provider.jsx';
 import UserProvider from './user-provider.jsx';
 
+// This can be removed once https://github.com/mdn/kuma/pull/6940 lands
+require('regenerator-runtime/runtime');
```
peterbe, line 10: Let's wait till #6940 lands so we don't have to remember to make another PR just to remove this.

PR #6972, opened 2020-04-30
Move stripe webhooks to /api/v1/stripe_hooks

Gregoor:


kuma/api/v1/views.py
```diff
@@ -1,17 +1,25 @@
 import json
+import os
```
peterbe, line 2: This file is getting so huge now! We might need to break it up into `api/v1/views/subscriptions.py` and `api/v1/views/search.py` etc.

Gregoor:
CI is confusing me. I'm getting
```
$ docker-compose exec web black kuma --check --diff kuma docs tests
All done! ✨ � ✨
448 files would be left unchanged.
```

Gregoor:
back on master 🚀

PR #6978, opened 2020-05-01
limit response in $local/docs/get-documents

### LINKED ISSUE #6754: SystemError: <method 'get' of 'dict' objects> returned a result with an error set
https://sentry.prod.mozaws.net/operations/mdn-prod/issues/6659909/

```
timeout: timeout
SystemError: <method 'get' of 'dict' objects> returned a result with an error set
  File "meinheld/mlogging.py", line 187, in _access
    'h': environ.get('REMOTE_ADDR', '-'),

SystemError: <method 'get' of 'dict' objects> returned a result with an error set
```

Low priority: 12x in 2yrs, but might worth looking into with spare time.

peterbe:
Fixes #6754

See email from me about this.

### DIFF
--- a/kuma/wiki/views/misc.py
+++ b/kuma/wiki/views/misc.py
@@ -84,7 +84,7 @@ def autosuggest_documents(request):
 
     # Generates a list of acceptable docs
     docs_list = []
-    for doc in docs:
+    for doc in docs[:100]:
         data = doc.get_json_data()
         data["label"] += " [" + doc.locale + "]"
         docs_list.append(data)


kuma/wiki/views/misc.py
```diff
@@ -84,7 +84,7 @@ def autosuggest_documents(request):
 
     # Generates a list of acceptable docs
     docs_list = []
-    for doc in docs:
+    for doc in docs[:100]:
```
escattone, line 87: I would say we could make this even smaller, say somewhere between `25` and `50`.

peterbe: You're not wrong but it's not important enough at the moment. :) 
Thanks!

PR #6992, opened 2020-05-04
github new issue link

### LINKED ISSUE #6911: T- Fix github feedback link, to /choose not /new
**Summary**
The Wiki and the read-only site has a "Report an issue"
It links to https://github.com/mdn/kuma/issues/new but if you use that it bypasses the nice issue templates and issue don't automatically get the `status: needs triage` label so they're likely to miss our weekly itrage. 

**Audience**
Users of MDN.

**Rationale**
Every now and then we get issues filed that don't have the `status: needs triage` label.

**Workaround**
Do nothing.

**Proposal**
Change URL from https://github.com/mdn/kuma/issues/new
to https://github.com/mdn/kuma/issues/new/choose in both the WIki header and the read-only site header. 

**Additional context**
Perhaps our chooser template thing should have a section for **content**. 
All too often people writing in and complain about something not being right in the content. No problem with doing that. But the so often fail to include a URL of which page they're referring to. If we use a better template we can strongly press for this before we throw the issue over to `mdn/sprints`. 
(I know there's a "Report a content problem" in the Feedback nav but clearly people aren't seeing this)

peterbe:
Fixes #6911

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -21,7 +21,7 @@
     },
     "bugs": {
         "list": "https://mzl.la/2onLvZ8",
-        "report": "https://github.com/mdn/kuma/issues/new",
+        "report": "https://github.com/mdn/kuma/issues/new/choose",
         "mentored": "https://mzl.la/2BFx4Ie",
         "goodfirstbug": "https://mzl.la/2ojEGaW"
     },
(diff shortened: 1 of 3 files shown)


peterbe:
It's human to forget that I had [already created the PR](https://github.com/mdn/kuma/pull/6985) but why did GitHub not help me at all with that duplication?!

PR #6993, opened 2020-05-04
Enhanced responsive footer

### LINKED ISSUE #6821: T - Implement mobile-optimized footer.
See User Story issue for comps and description: #6784.

# Acceptance Criteria 
- [ ] New responsive footer implemented on desktop, tablet, mobile

schalkneethling:
This updates the footer to be fully responsive across devices. One thing to note, the lang selector has temporarily been removed from the landing page.

I will create an issue to address this topic as there is some discussion needed around how to best implement this on the React side.

With that out of the way, here is what to look for.

## Footer on mobile

Screenshot 2020-05-04 at 15 34 46

## Footer on tablet

Screenshot 2020-05-04 at 15 34 57

## Footer on desktop

Screenshot 2020-05-04 at 15 34 25

NOTE: There are some very slight deviations from the design in https://github.com/mdn/kuma/issues/6784 due to technical constraints.

Also NOTE: The footer on `wiki.` should still be as it is currently in production.

@mindy r?

@malqinneh for design review.

fix #6821


malqinneh:
Thank you @schalkneethling!

### For Desktop and Tablet:
- Spacing can be tighter specifically around legal note/links (refer to [Mozilla's footer for reference](https://www.mozilla.org/en-US/))
- No need to use Zilla for [MDN] & [Mozilla] labels (one font should suffice)

### For Desktop, Tablet, and Mobile:
- Background can be #000 instead of #333 (MDN logo in the header and background of footer can be 100% black for brand guide alignment and contrast legibility)

jinja2/react_base.html
```diff
@@ -105,29 +105,73 @@
...
+          <li><a
+              href="https://www.mozilla.org/firefox/?utm_source=developer.mozilla.org&utm_campaign=footer&utm_medium=referral">Firefox</a>
+          </li>
+        </ul>
+        
+        <div class="social social-mdn">
+          <h4>{{ _('MDN') }}</h4>
+          <ul>
+            <li>
+              <a href="https://twitter.com/mozdevnet">
```
mindy, line 130: Suggestion: I think we should make the external links open in a new tab.

schalkneethling: I was thinking the same but, they were not currently so, I left it as is. Happy to change it though.

kuma/static/styles/minimalist/structure/_footer.scss
```diff
@@ -1,183 +1,144 @@
...
     }
+}
 
-    .languages {
-        @include clearfix();
-        padding-bottom: $gutter-width;
-        margin-bottom: $gutter-width;
+@supports (display: grid) {
+    .content-container {
+        display: grid;
```
mindy, line 95: I don't have much experience with CSS grids, so this is very neat to see how it's implemented !

jinja2/react_base.html
mindy: Since we were both thinking it, I think we should go with our instincts and change it

malqinneh: Third-ed!

schalkneethling: I will make it so.

schalkneethling:
Also, no worries with regards to the lang selector. We are going to standardize on no lang selector across footers:

https://github.com/mdn/kuma/issues/6784#issuecomment-623449432

schalkneethling:
>     * No need to use Zilla for [MDN] & [Mozilla] labels (one font should suffice)

Yeah, that is somehow happening on the landing page. Will have to override it :(

schalkneethling:
>     * Background can be #000 instead of #333 (MDN logo in the header and background of footer can be 100% black for styleguide alignment and contrast legibility)

I would not recommend this. Having a foregroud color of `#fff`, both `#000` and `#333` passes WCAG AAA as a background color in terms of contrast. The problem with pure white on pure black is that it can cause general eye strain and, is particularly bad for people with dyslexia - https://uxmovement.com/content/why-you-should-never-use-pure-black-for-text-or-backgrounds/

schalkneethling:
> Thank you @schalkneethling! This looks good to me-- just had one question about the external links.

Updated!

malqinneh:
> I would not recommend this. Having a foregroud color of `#fff`, both `#000` and `#333` passes WCAG AAA as a background color in terms of contrast. The problem with pure white on pure black is that it can cause general eye strain and, is particularly bad for people with dyslexia - https://uxmovement.com/content/why-you-should-never-use-pure-black-for-text-or-backgrounds/

In context of the footer (ie. size, location, copy/links, priority on page), do you think changing it to #000 to maintain standardization as per Mozilla's brand guide will cause users eye strain?

schalkneethling:
> > I would not recommend this. Having a foregroud color of `#fff`, both `#000` and `#333` passes WCAG AAA as a background color in terms of contrast. The problem with pure white on pure black is that it can cause general eye strain and, is particularly bad for people with dyslexia - [uxmovement.com/content/why-you-should-never-use-pure-black-for-text-or-backgrounds](https://uxmovement.com/content/why-you-should-never-use-pure-black-for-text-or-backgrounds/)
> 
> In context of the footer (ie. size, location, copy/links, priority on page), do you think changing it to #000 to maintain standardization as per [Mozilla's brand guide](https://mozilla.design/) will cause users eye strain?

I am not a fan, I personally find(especially small text) almost impossible to read:

Screenshot 2020-05-05 at 19 14 35

Also, MDN does not have to match Mozilla's brand identity 1:1, we are the rebel alliance :)

schalkneethling:
Thank you for the reviews @mindy and @malqinneh

PR #7038, opened 2020-05-11
Switch to single interaction model for main navigation

### LINKED ISSUE #7008: T - Standardize on single interaction model - main navigation
In the past the top-level entries for each main menu linked to a landing page and so, it was no possible to trigger the drop-down menus via a click event.

This is no longer the case. On mobile, we do trigger the drop-down via a tap event. I propose that we standardize the menu interaction on all viewports to be expanded via click event as opposed to hover on desktop and click on mobile.

## Acceptance Criteria

- [ ] Main menu expands on click on all viewports and devices
- [ ] instrumentation matches desktop events or implements new events

schalkneethling:
This switches us to a single interaction model with regards to the main menu for all viewport sizes and devices.

@mindy r?

fix #7008


mindy:
I don't see anything in this diff that could've affected it, but the wiki header dropdown isn't showing for me on hover or click: 
wiki-nav

Let me know if I should move this to a separate issue.

schalkneethling:
> I think there needs to be a `pointer-events: none` added to the dropdown arrow. Currently clicking on `.main-menu-arrow` does not open up the dropdown.

I moved the icon to a background image. One less DOM element per top-level entry o/\\o

> I don't see anything in this diff that could've affected it, but the wiki header dropdown isn't showing for me on hover or click

Hmm, still works for me

schalkneethling:
> I think there needs to be a `pointer-events: none` added to the dropdown arrow. Currently clicking on `.main-menu-arrow` does not open up the dropdown.

I moved the icon to a background image. One less DOM element per top-level entry o/\\o

> I don't see anything in this diff that could've affected it, but the wiki header dropdown isn't showing for me on hover or click

Hmm, still works for me

schalkneethling:
Thank you for the review @mindy - updated

mindy:
> I moved the icon to a background image. One less DOM element per top-level entry o/\o
> 
Even better!

> > I don't see anything in this diff that could've affected it, but the wiki header dropdown isn't showing for me on hover or click
> 
> Hmm, still works for me
This is working for me now, sorry about the false alarm! Weirdly when I turn off "cross-site trackers blocked" it works 🤷‍♀️

mindy:
LGTM. Thank you Schalk! I think you may need to pull the latest version of master to get the build to pass, per Peter's email (I'm in the same boat and need to do so myself :))

PR #7045, opened 2020-05-11
Remove document v1 API endpoints

### LINKED ISSUE #7000: Remove document v1 API endpoints
**Summary**
We built `/api/v1/doc/:locale/:slug` when we started the React re-write in hopes that subsequent clicks would load the new necessary document state via an XHR request alone. 
Turns out we never used it. And it's now extremely unlikely that we ever will. ...in Kuma. 

**Rationale**
The v1 API is here to stay and it helps to keep it not slowly deteriorating. And faster CI builds helps everyone. 

**Audience**
Devs.

**Proposal**
Things like https://developer.mozilla.org/api/v1/doc/en-US/Web/HTML/Element/progress will 404.

**Additional context**
Perhaps it's used by the linter in `stumptown-content`??
Acting on this issue is bound to clash with https://github.com/mdn/kuma/issues/6999 if done in parallel.

peterbe:
Fixes #7000

It's almost bittersweet how much work went into this and now gets deleted. It really wasn't that long ago that we wrote it. Well, that's life. 

I think this is worth doing because the v1 API is here to stay indefinitely. The Kuma front-end that this would have supported, is not. Yari is coming. 

Ultimately, this means less code which is always good and it means slightly faster completion in CI since it touched on quite a few unit tests.


kuma/api/apps.py
```diff
@@ -12,5 +12,3 @@ class APIConfig(AppConfig):
 
     def ready(self):
```
escattone, line 13: We should probably just get rid of the `ready` method altogether.

peterbe: done. Thanks!

PR #7052, opened 2020-05-12
Update mobile search interaction

### LINKED ISSUE #7011: T - Implement header search component
In https://github.com/mdn/kuma/issues/6784 a new interaction model has been designed for the search input field in the header.

## On Mobile

79573164-315a4e00-808c-11ea-82b7-86dfc6bab7c6

1. The search input is initially collapsed and only shows the magnifying glass at the top right.
2. Once a user taps/activates the search input, it expands to fill the entire header area
3. The user can exit out of the active state by either click/tapping the `x`
4. This will then collapse the input back to the icon only and reveal the rest of the header
5. Optional: The back arrow would clear the search that was recently submitted and take the user back to the previous page, before the search query was submitted (essentially behaving like a browser's back button). Example: https://material.io/ 

## On Tablet

1. Keep styling and interaction as on desktop

NOTE: Because users often resize their browsers we need to ensure that the search input is keyboard accessible at tablet viewport sizes.

## On Desktop

79573189-38815c00-808c-11ea-8062-81543f561b87

Styling and interaction are similar to the current desktop experience.

mindy:
Changes
- Add toggle button that handles showing and hiding the search form
- Update styles so it looks like the new mobile design, but keeping the tablet and desktop the same 

Closes #7011 

Note: This PR does not address item 5 in the ticket, which is adding the back button. Mustafa said he was fine with leaving it out, and our current router does not appear to give us the ability to go back in history. I'm going to take a look at the router code tomorrow, as there seem to be functions that indicate this was possible in the past, to see how difficult this would be. But any how, it will be a separate PR if it does end up happening. 

To test: 
1. Using a small screen width, go to any of the React pages. Click on magnifying glass in top right corner:
Screen Shot 2020-05-11 at 10 40 55 PM

2. Ensure that scrolling down covers content on the page: 
Screen Shot 2020-05-11 at 10 41 18 PM

3. Type a search term and press "Enter", should be taken to search page with search input pre-filled:
Screen Shot 2020-05-11 at 10 50 54 PM

4. Make window size larger, should see the rounded search input that is on the production site: 
Screen Shot 2020-05-11 at 10 52 07 PM


kuma/javascript/src/header/search.jsx
```diff
@@ -4,45 +4,75 @@ import { useEffect, useRef } from 'react';
...
+                    id="main-q"
+                    name="q"
+                    placeholder={gettext('Search MDN')}
+                    pattern="(.|\s)*\S(.|\s)*"
+                    required
+                />
+            </form>
+            <button className="toggle-form" onClick={handleClick}>
+                <CloseIcon className="close-icon" />
+                <SearchIcon className="search-icon" />
```
schalkneethling, line 69: Not sure if this might be expensive DOM wise but, could we dynamically switch between the icons as you are currently doing with the text?

kuma/static/styles/minimalist/components/_search-widget.scss
```diff
@@ -1,26 +1,38 @@
...
+    top: -$mobile-search-height;
+    right: 0;
+    width: 100vw;
+    height: $mobile-search-height;
+
+    form {
+        background: $neutral-100;
+        grid-area: S;
+        display: flex;
+        flex: 2 1;
```
schalkneethling, line 17: The above rule and the rule `grid-area: S` has no effect as the form is not a grid item. We can safely remove it. `grid-area: S` should be on `.header-search` instead. Our grid and general layout is... a little scattered at the moment :smile\_cat:

mindy: Removed `flex: 2 1;` and moved `grid-area: S` to `.header-search`. Thank you!

kuma/javascript/src/header/search.jsx
mindy: It seems like they both need to be on the page in order to animate, otherwise it just swaps out the icons. I will experiment with this some more.

mindy: I couldn't get this to work with keeping the animation and dynamically switching icons, unless we wanted to introduce a library like React Transition Group. Let me know if you have suggestions on how to handle this.

schalkneethling: Minor premature optimization :) I am fine with keeping it this way. Thanks for looking into it @mindy

kuma/static/styles/minimalist/components/_search-widget.scss
```diff
@@ -30,4 +41,92 @@
...
+            .search-icon {
+                display: inline-block;
+                flex: 0 0 25px;
+                fill: $text-color;
+            }
+        }
+
+        .toggle-form {
+            display: none;
+        }
```
schalkneethling, line 130: I prefer that we keep these in context so, for example, I would move this to the block at line 45 like so:

```
.toggle-form {
    border: inherit;
    height: inherit;
    position: inherit;
    display: flex;
    align-items: center;
    justify-content: center;
    overflow: hidden;
    width: $mobile-search-height;
    top: $mobile-search-height;
    right: 0;

    @media #{$mq-tablet-and-up} {
        display: none;    
    }
}
```

And then the same with the other instances such as `header-search` etc.

mindy: Thanks Schalk! Updated and ready for another look 👀

PR #7059, opened 2020-05-12
Front-end testing cleanup and improvements

mindy:
Changes:
- Add `eslint-plugin-testing-library` (thanks @peterbe for sharing this [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)!) 
- Import `jest-dom` before test files - I think we use it enough that it makes sense to always have it available. 
- Clean up existing tests that imported jest-dom matchers individually
- Fix eslint errors

To test:
Run `jest`

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -121,7 +121,7 @@ module.exports = {
     // setupFiles: [],
 
     // The path to a module that runs some code to configure or set up the testing framework before each test
-    // setupTestFrameworkScriptFile: null,
+    setupFilesAfterEnv: ['./jest.setup.js'],
 
     // A list of paths to snapshot serializer modules Jest should use for snapshot testing
     // snapshotSerializers: []
--- a/kuma/javascript/src/article.jsx
+++ b/kuma/javascript/src/article.jsx
@@ -55,7 +55,9 @@ export default function Article({ document }: DocumentProps) {
     // This is a one-time effect we need to call the first time an article
     // is rendered, to ensure that interactive examples resize themselves
     // if the browser width changes.
-    useEffect(InteractiveExamples.makeResponsive, []);
+    useEffect(() => {
+        InteractiveExamples.makeResponsive();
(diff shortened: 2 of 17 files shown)


jest.config.js
```diff
@@ -121,7 +121,7 @@ module.exports = {
     // setupFiles: [],
 
     // The path to a module that runs some code to configure or set up the testing framework before each test
-    // setupTestFrameworkScriptFile: null,
```
mindy, line 124: From the Jest docs:

> Note: setupTestFrameworkScriptFile is deprecated in favor of setupFilesAfterEnv.

kuma/javascript/src/article.jsx
```diff
@@ -55,7 +55,9 @@ export default function Article({ document }: DocumentProps) {
     // This is a one-time effect we need to call the first time an article
     // is rendered, to ensure that interactive examples resize themselves
     // if the browser width changes.
-    useEffect(InteractiveExamples.makeResponsive, []);
+    useEffect(() => {
```
mindy, line 58: This is to fix an eslint error from `react-hooks`, it said to make this an inline function.

kuma/javascript/src/header/search.jsx
```diff
@@ -9,16 +9,16 @@ type Props = {
...
     // After our first render, set the input field's initial value
     const inputfield = useRef(null);
     useEffect(() => {
-        if (inputfield.current && props.initialQuery) {
-            inputfield.current.value = props.initialQuery;
+        if (inputfield.current && initialQuery) {
+            inputfield.current.value = initialQuery;
         }
-    }, [props.initialQuery]);
+    }, [initialQuery]);
```
mindy, line 21: Another eslint error fix-- `props` will change when any props change (in this case there's only one prop right now so it doesn't really matter) so the fix is to destructure props and refer to the specific props inside `useEffect`.

PR #7061, opened 2020-05-12
Add ability to push to multiple clusters

### LINKED ISSUE #7049: Modify jenkins pipeline to be able to push to multiple clusters
**Summary**
Currently we can only push deployments to a singular kubernetes cluster, in order to help push forward issue #7034 it makes it easier to be able to push deployments to multiple clusters so that we can get Kuma and kumascript running in parallel before we switch over fully to the EKS cluster

**Rationale**
Helps facilitate migrating  kuma and kumascript over to the new EKS cluster 

**Audience**
This is a backend only change so no one will be able to see this change

limed:
Add ability to push to multiple cluster for just the `eu-central-1` region, the `get_target_script` function in `utils.groovy` needed to be updated to be able to accept an argument. This allows the function to be able to use the  appropriate config file during deployments.

This goes hand in hand with mdn/infra#404 and this PR resolves #7049

### DIFF
--- a/Jenkinsfiles/standby-push.groovy
+++ b/Jenkinsfiles/standby-push.groovy
@@ -31,8 +31,8 @@ stage('Push') {
 }
 
 stage('Push EKS') {
-    def current_revision_hash = utils.get_revision_hash('prod.eks.mm')
     dir('infra/apps/mdn/mdn-aws/k8s') {
+        def current_revision_hash = utils.get_revision_hash('prod.eks.mm')
         withEnv(["TO_REVISION_HASH=${env.GIT_COMMIT}",
                  "FROM_REVISION_HASH=${current_revision_hash}"]) {
             utils.migrate_db('prod.eks.mm')
--- a/Jenkinsfiles/standby-push.groovy
+++ b/Jenkinsfiles/standby-push.groovy
@@ -1,0 +1,47 @@
+stage("Announce") {
+    utils.announce_push()
+}
+
+stage("Check Pull") {
+    // Check that the image can be successfully pulled from the registry.
+    utils.ensure_pull()
+}
+
+stage("Prepare Infra") {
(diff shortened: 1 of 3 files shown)


Jenkinsfiles/standby-push.groovy
```diff
@@ -0,0 +1,47 @@
...
+            // Monitor the rollout until it has completed.
+            utils.monitor_rollout()
+            // Record the rollout in external services like New-Relic.
+            utils.record_rollout()
+        }
+    }
+}
+
+stage('Push EKS') {
+    def current_revision_hash = utils.get_revision_hash('prod.eks.mm')
```
escattone, line 34: I think this needs to be moved below the `dir('infra/apps/mdn/mdn-aws/k8s') {` line below, or else it'll fail since it assumes that it's run from the `infra/apps/mdn/mdn-aws/k8s` directory.

Jenkinsfiles/utils.groovy
```diff
@@ -166,37 +169,37 @@ def is_read_only_db() {
     }
 }
 
-def migrate_db() {
+def migrate_db(target='') {
     /*
      * Migrate the database (only for kuma and writeable databases).
      */
     if ((get_repo_name() == 'kuma') && !is_read_only_db()) {
```
escattone, line 176: The function `get_target_script()` is also used within `is_read_only_db()`, so that function's signature needs to change, and `target` needs to be passed into its call here.

Jenkinsfiles/standby-push.groovy
limed: Good catch, fixed this in ab12738810e72d1d05fbc200d26f8f519484e5d4

Jenkinsfiles/utils.groovy
limed: Fixed in ab12738810e72d1d05fbc200d26f8f519484e5d4

PR #7062, opened 2020-05-13
Sync with sendinblue

### LINKED ISSUE #7024: T - synchronise email address between MDN and sendinblue
AC
- [ ] when a user sets a new primary email, the email address in sendinblue is updated to that new email address

Gregoor:
Fixes #7024 

I might've done a bit of a controversial thing here, but I like it:

Instead of calling sendinblue related functions from all over the kuma.users app, I added 4 new signals, representing what is happening in those places.
https://github.com/Gregoor/kuma/blob/a4c3cf839fb79d9f74d68b0ade038e836e0c3191/kuma/users/signals.py#L3-L7

Yes, it adds indirection (boo), but it allows us to neatly store all the sendinblue stuff in one place. I initially started appending all the required functionality to `kuma.users.{tasks, signal_handlers}` but didn't like how I was contributing to these files growing more and more with less and less related functionality.

Anyway, while I'm fairly happy with the base functionality there's still two things missing:
- testing
- a cron syncing everything, like we talked about in our call last week
  - while ideally not necessary, I found sendinblue to error on some requests for reasons I didn't grasp yet. So I might want to invest in a fallback

Before I go on with that, I'd like to get a thumb up/down whether the approach is generally sound. Then I'll work on the rest and ask for a real review.


Gregoor:
Moved the updating contact logic into the contact, as we talked about. It didn't make sense for deleting the contact, as one of the signals only gets the user's email after it has already been changed for that user. But I like it for the updating and I think it made for nicer code.

## How to test it:
Create a sendinblue list, put the id and API key in the config (see docs for details). Keep sendinblue's dsahboard open and run through these scenarios:
1. sign up (with/out newsletter subscription checked)
2. start a subscription
3. cancel a subscription
4. change your primary email
5. delete your account

Refresh sendinblue in between to see how the contact changes.

peterbe:
Note-to-self; The [CI passed](https://travis-ci.com/github/mdn/kuma/builds/166313971) but it seems stuck here in the PR.

peterbe:
First thing I tested was to unset my newsletter subscriber on the profile page. And checked it and pressed save. Yes, now I appear:
Screen Shot 2020-05-15 at 3 52 13 PM

However, I went back to my profile http://localhost.org:8000/en-US/profiles/peterbe2/edit and changed my name. But that didn't update my line in Sendinblue. Same output as the screenshot above. Is that a bug?

kuma/users/newsletter/checks.py
```diff
@@ -0,0 +1,36 @@
...
+
+def sendinblue_check(app_configs, **kwargs):
+    if not settings.SENDINBLUE_API_KEY:
+        return []
+
+    response = sendinblue.request("GET", "contacts/attributes")
+    if not response.ok:
+        return [
+            Error(
+                f"Error when creating sendinblue attribute: {response.json()['message']}",
```
peterbe, line 17: That string isn't really right. Also, I don't think you can assume that it's JSON. 
```suggestion
                f"Error getting sendinblue attributes: {response.status_code}",
```

kuma/users/newsletter/sendinblue.py
```diff
@@ -2,8 +2,8 @@
 
 from kuma.core.utils import requests_retry_session
 
-
 API_URL = "https://api.sendinblue.com/v3/"
+LIST_ID = settings.SENDINBLUE_LIST_ID
```
peterbe, line 6: I don't see the point in this. 
Also, it's generally better to "delay" the use of `settings.` until it's needed. It's probably not applicable here but if you don't do that you can write a test like this:

```python
def test_something(settings):
    settings.SENDINBLUE_LIST_ID = 'test'
    ...
```

kuma/users/newsletter/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+        "contacts",
+        json={
+            "updateEnabled": True,
+            "email": user.email,
+            "attributes": {
+                "IS_PAYING": UserSubscription.objects.filter(
+                    user=user, canceled__isnull=True
+                ).exists()
+            },
+            "listIds": [int(sendinblue.LIST_ID)],
```
peterbe, line 28:
```suggestion
            "listIds": [int(settings.SENDINBLUE_LIST_ID)],
```

kuma/users/newsletter/sendinblue.py
Gregoor: I like that it defers the knowledge of where the list ID comes from to the sendinblue module. Couldn't you also write:

```python
def test_something(settings):
    sendinblue.LIST_ID = 'test'
    ...
```

There's a textual aesthetic thing here, that I also think it reads better, so maybe that is biasing me.

Gregoor: oh no I guess it would be more like:

```python
@patch('sendinblue.LIST_ID', 'test')
def test_something():
  ...
```

kuma/users/newsletter/tasks.py
Gregoor: I'm still torn but I like consistency and your way is more consistent with what we're doing in other places. So I'll go with it! (not merging in the suggestions, because it needs some importing and moving yada yada)

kuma/users/newsletter/sendinblue.py
peterbe: Yeah, you could and it would probably work. But sticking to the convention of using `settings` is generally better. The `settings` pytest fixture is great because it's guaranteed to not leak across different tests and now you don't need to know exactly what the `mock.patch` *path* is. 

All in all, it's fine. Just a nit. Or perhaps a recommendation from my experience.

PR #7065, opened 2020-05-14
Fix user menu and lang menu on mobile and standardize interaction

### LINKED ISSUE #7009: T - Fix user menu on mobile and standardize interaction model
Similar to https://github.com/mdn/kuma/issues/7008 but related to the user menu(aka avatar/mugshot).

Currently, on desktop, the submenu for the user menu is triggered via a hover event. On mobile, this functionality does not work at all. The user can sign in but, once signed in, cannot access their profile nor logout.

This task than covers both fixing the broken functionality on mobile and standardizing on a consistent interaction model i.e. user menu is expanded via a click event on all viewports and devices.

## Acceptance Criteria

- [ ] The user menu is exandable on all viewports and devices
- [ ] User menu is expanded via a click/tap event on all viewports and devices
- [ ] instrumentation matches desktop events or implements new events

schalkneethling:
Turns out the user(avatar) menu and language menu did not work on touch devices :scream: 

This PR resolves this and then also standardizes the interaction to match that of the main menu. All menus are now triggered via tap or click interactions across all devices.

@mindy r?

fix #7009


kuma/javascript/src/header/dropdown.jsx
```diff
@@ -69,7 +104,9 @@ export default function Dropdown(props: DropdownProps) {
             </button>
             <ul
                 id={props.ariaOwns}
-                className={`dropdown-menu-items${props.right ? ' right' : ''}`}
+                className={`dropdown-menu-items${props.right ? ' right' : ''} ${
```
mindy, line 107: Extra space here `''} ${`

kuma/javascript/src/header/dropdown.jsx
```diff
@@ -48,16 +53,46 @@ export default function Dropdown(props: DropdownProps) {
...
+
+        document.addEventListener('click', (event: MouseEvent) => {
+            if (
+                event.target &&
+                event.target instanceof HTMLElement &&
+                !event.target.classList.contains('dropdown-menu-label')
+            ) {
+                hideDropdownMenuIfVisible();
+            }
+        });
```
mindy, line 77: Would it be simpler to add an `onBlur` event to the button? Maybe something like
```
onBlur={() => {
  setShowDropdownMenu(false);
}}
```

kuma/javascript/src/header/dropdown.jsx
schalkneethling: I agree, the space is actually added inside the `true` portion but, I am not a fan :) I would rather just do:

```suggestion
className={`dropdown-menu-items ${props.right ? 'right' : ''} ${
```

as I did for the next `class` :+1:

kuma/javascript/src/header/dropdown.jsx
schalkneethling: I don't believe so. When you use the keyboard and tab from the `button` to one of the dropdown menu items, it will fire the `blur` event and close the dropdown.

kuma/javascript/src/header/dropdown.jsx
schalkneethling: Oh doh! I misread what you were referring to. Either way, updated it to the latter so it is consistent with other places in the codebase.

kuma/javascript/src/header/dropdown.jsx
mindy: Ah didn't think about that! Makes sense to keep as is then 👍

PR #7072, opened 2020-05-14
Set default argument for make when there are only 3 arguments

limed:
Also sneaked in some changes to notify slack that its pushing to EKS and also removed the record deployment job for EKS

### DIFF
--- a/Jenkinsfiles/standby-push.groovy
+++ b/Jenkinsfiles/standby-push.groovy
@@ -31,6 +31,13 @@ stage('Push') {
 }
 
 stage('Push EKS') {
+
+    // add additional notify
+    notify_slack([
+        status: "Pushing to EKS",
+        message: "${repo} image ${tag}"
+    ])
+
     dir('infra/apps/mdn/mdn-aws/k8s') {
         def current_revision_hash = utils.get_revision_hash('prod.eks.mm')
         withEnv(["TO_REVISION_HASH=${env.GIT_COMMIT}",
@@ -40,8 +47,6 @@ stage('Push EKS') {
             utils.rollout('prod.eks.mm')
             // Monitor the rollout until it has completed.
             utils.monitor_rollout('prod.eks.mm')
-            // Record the rollout in external services like New-Relic.
-            utils.record_rollout('prod.eks.mm')
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Jenkinsfiles/standby-push.groovy
```diff
@@ -31,6 +31,13 @@ stage('Push') {
 }
 
 stage('Push EKS') {
+
+    // add additional notify
+    notify_slack([
+        status: "Pushing to EKS",
+        message: "${repo} image ${tag}"
+    ])
```
escattone, line 39: I don't think this will work as it is. First of all, you'd need to use `utils.notify_slack`, but also `repo` and `tag` are undefined. I think it'd be better to use `utils.announce_push("EKS")` here and then also change the `announce_push` function within `Jenkinsfiles/utils.groovy` to something like:
```groovy
def announce_push(target=null) {
    /*
     * Announce the push.
     */
    if (target == null) {
        target = get_target_name()
    }
    def repo = get_repo_name()
    def tag = get_commit_tag()
    notify_slack([
        status: "Pushing to ${target}",
        message: "${repo} image ${tag}"
    ])
}
```

limed: Oh good catch, I took a slightly different approach with this and instead just modified the `get_target_name` function instead to accept an argument. See commit bb604699920d28b2fe39cede91c6fefd23a3e34b for details

Jenkinsfiles/utils.groovy
```diff
@@ -12,7 +12,10 @@ def get_commit_tag() {
     return env.GIT_COMMIT.take(7)
 }
 
-def get_target_name() {
+def get_target_name(target_name='') {
+    if (target_name != null && !target_name.trim().isEmpty()){
+        return target_name
+    {
```
escattone, line 18: I just noticed that this is missing the closing bracket `}` (you have another `{` instead), but it can be simplified as well, since a `null` value as well as an empty string are both **falsy** in Groovy:
```suggestion
    if (target_name) {
        return target_name
    }
```

limed: Good catch! Fix in 4befc0ff6d6d953b8149da901e9902770922c1c1

PR #7130, opened 2020-05-25
Add base for new accounts pages

### LINKED ISSUE #7073: T - Acct Settings: layout / framework
# Acceptance Criteria

- [ ] base Acct Setting landing page (React)
- [ ] sidebar navigation component
- [ ] button styling
- [ ] title bar/name

# sidebar nav changes from design comps-
- update link 'email'->'email preferences' / slide 3
- add link 'email addresses' link->slide 9

schalkneethling:
This sets up the basis for the new account management experience. It is ready for review in general but there is still at least one open question(https://github.com/mdn/kuma/issues/7073#issuecomment-633524429)

Also, I have a test on the Python side to ensure the account page is not accessible by anonymous users but, it will be good to also have a happy path for logged in users. I need some assistance on getting that test setup/written.

The new base account page is accessible at http://localhost.org:8000/en-US/accounts/

Note the `s`. This is not ideal but, `account` clashes with an existing route we have. I did not want to block the account work on that so, moved ahead with `accounts` for now. We can resolve that while the rest of the work happens.

I am also separating out the sidebar nav into a separate PR, again, to not block work on other components etc.

@mindy @peterbe r?

fix #7073


jinja2/react_base.html
```diff
@@ -95,10 +95,8 @@
   {% block pageheader %}{% endblock %}
 
   <!-- Content will go here -->
-  <main id="content">
```
mindy, line 98: The a11y nav links to `#content`, so I think we should still keep the ID (whether it be here or one of the children) or update the link to point at a new ID.

kuma/javascript/src/account/routes.jsx
```diff
@@ -0,0 +1,65 @@
+// @flow
+import * as React from 'react';
+import Route from '../route.js';
+import Page from '../base/page.jsx';
+import LandingPage from './pages/index.jsx';
+
+type AccountRoutesParams = {
```
mindy, line 7: What do you think about abstracting `AccountRoutesParams` and `AccountPageProps` to a generic type that can be shared with the `/payments` pages?

kuma/javascript/src/account/routes.jsx
```diff
@@ -0,0 +1,65 @@
...
+    MANAGE_EMAIL: 'manage-email',
+    SUBSCRIPTION: 'subscription',
+};
+
+export function AccountPage(props: AccountPageProps) {
+    const { locale } = props;
+    const getPage = () => {
+        switch (true) {
+            default:
+                return <LandingPage locale={locale} />;
```
mindy, line 26: Missing a prop type `data` here, which the build is complaining about.

jinja2/react_base.html
schalkneethling: Doh, good catch. Will update.

schalkneethling:
@mindy updated

kuma/accounts/views.py
```diff
@@ -0,0 +1,12 @@
+import logging
+
+from django.shortcuts import render
+
+from kuma.core.decorators import login_required
+
+log = logging.getLogger("kuma.accounts.views")
+
+
+@login_required
```
peterbe, line 10: I'd rather we don't do this. Once we go to Yari this will never be a thing. You can't have the server redirect based on cookies unless it's `/api/v1/*` endpoints. See what we did for [the subscription management page](https://github.com/mdn/kuma/blob/3371363ed90751dcb5c495111d34d2c61a98537f/kuma/payments/views.py#L33-L35).
What the main React page component should do is:

1) Wait for `/api/v1/whoami`
2) If no user in `userData`, display [something about not being logged in](https://github.com/mdn/kuma/blob/3371363ed90751dcb5c495111d34d2c61a98537f/kuma/javascript/src/payments/pages/management.jsx#L151-L158).
3) If logged in, load the `/api/v1/account` (or whatever we chose)

This is the pattern of the future.

kuma/urls.py
```diff
@@ -100,6 +101,7 @@
         name="redirect-to-payments",
     ),
 )
+urlpatterns += i18n_patterns(re_path(r"^accounts/", include(accounts_lang_urlpatterns)))
```
peterbe, line 104: I don't understand this. 

* On line 48 of route.jsx it says `/${this.locale}/account` and here's it's `accounts/`
* In my browser address bar I don't want to see "accounts" because that sounds to me like that's where I manage my different accounts. Like a list of my different providers. A place where it lists my different google and github logins. 
* What is already there? "account clashes with an existing route we have". 

One last time, are we sure "settings" isn't better? I think it's rather important to nail this. A word that works for the basic stuff like setting your name, managing your subscription, administer your browser sets, etc. What other future things are going to go there and do they sound better with "settings" or "account"?

schalkneethling: > On line 48 of route.jsx it says `/${this.locale}/account` and here's it's `accounts/`

My bad, it should be `accounts`

> In my browser address bar I don't want to see "accounts" because that sounds to me like that's where I manage my different accounts. Like a list of my different providers. A place where it lists my different google and github logins.

I completely agree

> What is already there? "account clashes with an existing route we have".

I went off an error thrown by `web` in the logs but, I think it might be this here: https://github.com/mdn/kuma/blob/53780fa48f38456ffaa4a5e234426a05eddf0ba0/kuma/users/urls.py#L57

> One last time, are we sure "settings" isn't better? I think it's rather important to nail this. A word that works for the basic stuff like setting your name, managing your subscription, administer your browser sets, etc. What other future things are going to go there and do they sound better with "settings" or "account"?

😬 I think `account` would work but, the more I think about it, and seeing we already ran into a problem, I have been convinced to switch to using `settings` 🎉

schalkneethling: Ah :hankey: we already have `settings` as well :laughing:

schalkneethling: `preferences` worked out 🍭

schalkneethling:
@peterbe and @mindy updated. Thank you!!

schalkneethling:
I have to say I am a bit baffled by this error:

```
==================================== ERRORS ====================================

____________ ERROR collecting kuma/preferences/tests/test_views.py _____________

import file mismatch:

imported module 'test_views' has this __file__ attribute:

  /app/kuma/payments/tests/test_views.py

which is not the same as the test file we want to collect:

  /app/kuma/preferences/tests/test_views.py

HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules

!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!

=============================== 1 error in 9.98s ===============================

make: *** [Makefile:34: coveragetest] Error 2
```

@peterbe Any ideas?

peterbe:
Delete any *.pyc files in that new folder and try again?

peterbe:
Delete any *.pyc files in that new folder and try again?

schalkneethling:
> Delete any \*.pyc files in that new folder and try again?

I did `find . -name "*.pyc"` at the root of Kuma and it found no `.pyc` files

mindy:
Looks great @schalkneethling ! Waiting for the build to finish then I'll merge 👍

schalkneethling:
Thanks so much for the review and assistance @mindy and @peterbe :tea:

Issue #7076, opened 2020-05-15
T - Acct Settings: subscription component


tobinmori:
# Acceptance Criteria

- [ ] An active subscriber can see the status of their subscription on account settings
- [ ] A user can cancel their subscription from the account settings page

# Design 

![Screenshot 2020-06-01 at 09 50 28](https://user-images.githubusercontent.com/10350960/83387798-9fb16080-a3ed-11ea-9f67-857c07613a03.png)

mindy:
@atopal if the user does not have the waffle flag set, should we hide the subscription link in the sidebar nav?

atopal:
No, that's okay, since the subscription button is supposed to link to the subscription landing page anyway, and we won't launch this new account settings page without having launched the subscription option.

tobinmori:
status: 1/2 complete: component and logic, but based on original designs.

schalkneethling:
Acceptance criteria and design has been updated:
https://github.com/mdn/kuma/issues/7076#issue-618982588

schalkneethling:
Currently, on the payment management page the detail and layout is as follows:

![Screenshot 2020-06-18 at 14 22 35](https://user-images.githubusercontent.com/10350960/85019887-ce5e6380-b16f-11ea-80b3-ec97fbb39b1c.png)

On the same section on account details, the designs call for the following:

![Screenshot 2020-06-18 at 14 22 49](https://user-images.githubusercontent.com/10350960/85019903-d4ecdb00-b16f-11ea-894c-a827d3ea796b.png)

Seeing that we can share components and functionality here, it makes sense to me for these to match from a UI perspective as well. The alternative is that we need to create two variants based on the main component that uses the sub-components.

We _can_ do that, but it seems wasteful. Will the component on account setting not replace the payment management page in the end anyway?

Thoughts @malqinneh and @atopal?

atopal:
That's a good question, I don't think it's a good idea to cancel immediately after people click that button, that can easily happen by accident. For cancelling your account we have an overlay for you to confirm, but I haven't seen that for subscriptions. Mustafa?

schalkneethling:
> That's a good question, I don't think it's a good idea to cancel immediately after people click that button, that can easily happen by accident. For cancelling your account we have an overlay for you to confirm, but I haven't seen that for subscriptions. Mustafa?

The management page does do that and so, sharing the components and as such the functionality, we would inherit the same flow.

malqinneh:
[...]

PR #7131, opened 2020-05-25
Create hook to fetch subscription data

mindy:
Subscription data is also needed on the account settings Subscriptions component, so rather than duplicate the code, create a hook that both of them can use. The data needed is for this view:

Screen Shot 2020-05-26 at 10 29 46 AM

Changes:
- Move code that fetches subscription data into a custom hook 

To test:
- Management page should work just as it used to
- Tests in `payments/pages/management.test.jsx` should still pass 

Part of #7076

### DIFF
--- a/kuma/javascript/src/hooks/useSubscriptionData.js
+++ b/kuma/javascript/src/hooks/useSubscriptionData.js
@@ -1,0 +1,34 @@
+// @flow
+import { useEffect, useState } from 'react';
+import { getSubscriptions } from '../payments/api.js';
+import { type UserData } from '../user-provider.jsx';
+
+export type SubscriptionData = {
+    id: string,
+    amount: number,
+    brand: string,
+    expires_at: string,
+    last4: string,
+    zip: string,
+    next_payment_at: string,
+};
+
+function useSubscriptionData(userData: ?UserData) {
+    const [subscription, setSubscription] = useState<?SubscriptionData>(null);
+    const [error, setError] = useState<?string>(null);
+
+    useEffect(() => {
+        if (userData && userData.isSubscriber) {
+            const handleSuccess = (data) => {
+                const [subscription] = data.subscriptions;
+                setSubscription(subscription);
+            };
+            const handleError = (error) => setError(error);
+            getSubscriptions(handleSuccess, handleError);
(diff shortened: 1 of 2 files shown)


peterbe:
What is the current verdict on the confusing reality that you'd get your subscriptions stuff from two different places? 

I'd rather we iron out that flaw before we pile on my code. Where "iron out that flaw" means pushing back on the designs :)

It's not a just matter of DRY on code, but it's confusing as hell for users if there are two different places to do the same thing.

peterbe:
CC @escattone @Gregoor @tobinmori 
Dragging everyone in here. Perhaps an email would be a better place but I'm genuinely worried that we accept the new UX screens without pushing back. It's really bad as it is that, right now, we have the subscriptions form in the Edit Profile page AND on the Subscription landing page. And that ought to be cleaned up before we work on the new accounts settings. 
Having the same functionality in two different places is really bad for users because they'll worry about using the wrong thing and it confuses with their memory as to where to find things.

schalkneethling:
I believe the idea is that in general, the subscription page on `/accounts` will show you the current status of your subscription and allow you to cancel(manage) it. However, should a user not have a subscription, they could then sign up for one from this page as well.

I do agree with you though @peterbe. The page should point users to the subscription landing page should they not have a subscription instead of duplicating the functionality. So, if you have a subscription, show the subscription management tools else, show a link to the subscription landing page.

mindy:
> I do agree with you though @peterbe. The page should point users to the subscription landing page should they not have a subscription instead of duplicating the functionality. So, if you have a subscription, show the subscription management tools else, show a link to the subscription landing page.

That is the plan--  the "Continue" button will link users to the subscription landing page if they are not a subscriber. So there will only be one path to sign up for a subscription. 

Regarding the subscriptions data, I think having the data displayed here makes sense to me. The other way they could get this info is by going to `/payments/management`, which isn't advertised in many places. I do think cancelling a subscription should be done in account settings instead of taking the user to `/payments/management/` though, but not sure if that would make this more confusing to the user.

peterbe:
Ok. So it's just a tiny cross-pollination between the payments pages and the account settings pages. Good.

PR #7142, opened 2020-05-27
fix legacy samples files

escattone:
This PR updates the way we handle "legacy" files, all of the really old, rarely accessed files under `/diagrams/*`, `/presentations/*`, and `/samples/*`. However, this PR is really the second step.

The first step, which is already in place, is that all of the files under `/diagrams/*`, `/presentations/*`, and `/samples/*` have been moved to the stage and production media S3 buckets and made available via `https://media.prod.mdn.mozit.cloud` and `https://media.stage.mdn.mozit.cloud`, but the files are exactly the same between stage and production, and always will be since the legacy files are frozen. Also, we've added two new behaviors to the primary stage and production CDN's that serve traffic intended for `/diagrams/*` and `/presentations/*` directly from the appropriate S3 bucket. **You will notice that there is no CDN behavior for `/samples/*`.** This is because many of the `/samples/*` URL's are redirected by Django -- some to specific documents, some to `http://mdn.github.io`, and some to specific file attachments at `https://mdn.mozillademos.org/files/*` -- so all of the traffic for `/samples/*` has to first be routed to Django and then redirected from there.

So now that this first step is understood, we can better talk about what this PR does. It basically does three things:
- Adds a catch-all redirect for `/samples/*`, after all of the previous special cases, that simply redirects to `https://media.prod.mdn.mozit.cloud/samples/*`. I think it's fine to always redirect to the production media domain for samples because they are frozen, and will always be the same between stage and production, so there's no need to distinguish between the stage and production media domains. This is already how the special-case samples redirects work, for example `/samples/canvas-tutorial/images/backdrop.png` always redirects to `https://mdn.mozillademos.org/files/5395/backdrop.png`, which is the production domain (the stage domain is `https://files-stage.mdn.mozit.cloud/files/5395/backdrop.png`).
- Removes the code related to serving legacy files, since Django does not do that any longer.
- Adds a few new headless integration tests.

### DIFF
--- a/kuma/redirects/redirects.py
+++ b/kuma/redirects/redirects.py
@@ -1,5 +1,6 @@
 from functools import partial
 
+from django.conf import settings
 from redirect_urls import redirect as lib_redirect
 
 from kuma.core.decorators import shared_cache_control
@@ -704,6 +705,14 @@ scl3_redirectpatterns = [
         "http://mdn.github.io/webgl-examples/tutorial/webgl.css",
         re_flags="i",
         permanent=True,
+    ),
+    # All of the remaining "samples/" URL's are redirected to the
+    # the media domain (ATTACHMENTS_AWS_S3_CUSTOM_DOMAIN).
+    redirect(
+        r"^samples/(?P<sample_path>.*)$",
+        f"{settings.ATTACHMENTS_AWS_S3_CUSTOM_URL}/samples/{{sample_path}}",
+        re_flags="i",
+        permanent=False,
     ),
     # Bug 887428 - Misprinted URL in promo materials
     # RewriteRule ^Firefox_OS/Security$ docs/Mozilla/Firefox_OS/Security
--- a/tests/headless/map_301.py
+++ b/tests/headless/map_301.py
@@ -517,6 +517,13 @@ MOZILLADEMOS_URLS = list(
     )
 )
 
+# These are the remaining "samples/*" URL's, the ones that haven't been redirected
+# by any of the special cases, which we expect to be redirected to the media CDN.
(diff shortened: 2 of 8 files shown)


kuma/redirects/redirects.py
```diff
@@ -705,6 +705,16 @@ def locale_redirect(pattern, to, prepend_locale=True, **kwargs):
         re_flags="i",
         permanent=True,
     ),
+    # All of the remaining "samples/" URL's are redirected to the
+    # the production media domain ("media.prod.mdn.mozit.cloud"),
+    # since the "samples/" legacy files on that domain are the same,
+    # and will always be the same, between stage and production.
+    redirect(
+        r"^samples/(?P<sample_path>.*)$",
+        f"https://media.prod.mdn.mozit.cloud/samples/{{sample_path}}",
```
limed, line 714: Should we consider using the `ATTACHMENTS_AWS_S3_CUSTOM_DOMAIN` config instead? That should point to the attachments domain instead of actually hard coding the redirect for different environments to a singular domain

tests/headless/map_301.py
```diff
@@ -517,6 +517,25 @@
...
 )
 
+# These are the remaining "samples/*" URL's, the ones that haven't been redirected
+# by any of the special cases, which we redirect to the media CDN.
+DEFAULT_SAMPLES_URLS = list(
+    flatten(
+        (
+            url_test(
+                "/samples/cssref/background.html",
+                "https://media.prod.mdn.mozit.cloud/samples/cssref/background.html",
```
limed, line 527: Same as comment about, maybe use `ATTACHMENTS_AWS_S3_CUSTOM_DOMAIN` instead?

PR #7170, opened 2020-05-29
add ability to save/serve sitemaps to/from S3

escattone:
This PR provides and enables the ability to save the sitemap files to S3. When enabled, it also serves the sitemap files from S3, but for stage and production, those requests will never reach Django but instead be routed (by the primary CDN) directly to S3. This is the last step in eliminating our reliance on EFS. Here are the steps forward regarding EFS:
- [x] review this PR
- [x] test on stage
- [x] merge this PR
- [ ] create and merge an https://github.com/mdn/infra PR that removes EFS from our K8s resources

This is related to our work on https://github.com/mdn/kuma/issues/7034 because that effort to move to EKS was intended to not only reduce the cluster-management burden but also to eliminate the common problem of "stuck" cluster nodes, which we've discovered to be related to our use of EFS.


escattone:
I tested this on stage and it worked like a charm! 🎉

PR #7188, opened 2020-06-02
Add contributions linkto usernav

### LINKED ISSUE #7077: T - Acct Settings: Contributions component
# Acceptance Criteria

- [ ] subnav 'Contributions' link displays Recent docs activity

# Design

https://app.abstract.com/projects/9decc480-70fc-11e9-9453-4f3fbffea7ce/branches/10ce2c1c-129f-4379-877f-133962b86641/commits/9c73ba49eb0260cdd4fb2e99ee71d1ec1da90e57/files/4ffa74d7-ce7c-4f32-a6ea-7f1baa52572f/layers/817A00F5-7A93-4FB4-9735-17B280B45B9D?collectionId=885f7b0e-f211-4ba1-8442-7fb435e5312b&collectionLayerId=bed4c149-32eb-45a7-aae2-9118fd8034c4

schalkneethling:
"It will be simple they said..." 😄 

So, adding the link to the usernav is indeed simple, quick, and easy. But...

The page this links to is `http://wiki.localhost.org:8000/en-US/dashboards/revisions?user=username` - Note that `wiki` portion.

This page does not use the new header and as such, all links will from hereon out keep you on the wiki. This is probably not what we want.

What we probably want to do is to update the template this page uses to use the new header but, do we also want to prevent this from going to the `wiki.` domain? Is there a different option here?

Would love your thoughts @peterbe and @atopal 

/cc @tobinmori 

fix #7077

### DIFF
--- a/kuma/javascript/src/header/login.jsx
+++ b/kuma/javascript/src/header/login.jsx
@@ -37,6 +37,7 @@ export default function Login(): React.Node {
             </>
         );
         let viewProfileLink = `/${locale}/profiles/${userData.username}`;
+        let contributionsLink = `/${locale}/dashboards/revisions?user=${userData.username}`;
         let editProfileLink = `${viewProfileLink}/edit`;
 
         return (
@@ -47,6 +48,11 @@ export default function Login(): React.Node {
                     right={true}
                     hideArrow={true}
                 >
+                    <li>
+                        <a href={contributionsLink}>
+                            {gettext('Contributions')}
+                        </a>
+                    </li>
                     <li>
                         <a href={viewProfileLink}>{gettext('View profile')}</a>
                     </li>


atopal:
Any chance we could display that only for people who have contributions? They have to switch between wiki and the read-only site all the time anyway. If not, I don't mind much, this is going away soon anyway. I wouldn't change the header or anything.

schalkneethling:
@atopal Another option is to not bother with the contributions link at all. I doubt we have the info with regards to contribution status on the React side.

peterbe:
> Any chance we could display that only for people who have contributions?

Yes, this is possible. It would trivial to add to the `whoami` endpoint. Since it's a MySQL query we only need to do if `userData.isAuthenticated` it's a cheap way to avoid the query for most of the requests.

peterbe:
We'd obviously need to remember to undo that from the `/api/v1/whoami` once we switch to Yari.

kuma/javascript/src/header/login.jsx
```diff
@@ -47,6 +48,11 @@ export default function Login(): React.Node {
                     right={true}
                     hideArrow={true}
                 >
+                    <li>
+                        <a href={contributionsLink}>
+                            {gettext('Contributions')}
```
peterbe, line 53: Would it be advantageous to call this "Your contributions" instead? After all, we say "Your profile".

kuma/javascript/src/header/login.jsx
```diff
@@ -37,6 +37,7 @@ export default function Login(): React.Node {
             </>
         );
         let viewProfileLink = `/${locale}/profiles/${userData.username}`;
+        let contributionsLink = `/${locale}/dashboards/revisions?user=${userData.username}`;
```
peterbe, line 40: Does it work if your username has a space in it? (...if that's even possible to have!)
Does it work if your username has a non-ascii or a control character in it?

peterbe: I know there's a LOT of users with usernames that `[^a-z0-9_-]` in them. E.g. `Peter Wennblom` and `madeup@gmail.com`.
I think the safest thing is to use something like 
```suggestion
        let contributionsLink = `/${locale}/dashboards/revisions?user=${encodeURIComponent(userData.username)}`;
```
...but I haven't tested it.

schalkneethling: This is how we currently do it on the Jinja side:

```
<a href="{{ url('dashboards.revisions') }}?user={{ detail_user.username }}" class="button">{{ _('View all activity') }}</a>
```

kuma/javascript/src/header/login.jsx
schalkneethling: Well, this is per design and also, `View profile` and `Edit profile` will be replaced by `Account Settings` so I reckon just `Contributions` is fine.

kuma/javascript/src/header/login.jsx
schalkneethling: Shall I go ahead and encode the username here @peterbe?

peterbe: Yes please. You never know what might go wrong if you don't encode it. I think it's safest to do so. For example, what if you come up with a crafty username that causes a URL that does unintentional things we couldn't have foreseen. Security or just breakage, either way it's easy to prevent. 

Had you done this:
```javascript
let sp = new URLSearchParams();
sp.set('username', userData.username);
let contributionsLink = `/${locale}/dashboards/revisions?${sp.toString()}`;
```
that would have produced a string of `/en-US/dashboards/revisions?user=p%20%C3%A9ter%40gmail.com`. And you'd get exactly the same result with:

```javascript
let contributionsLink = `/${locale}/dashboards/revisions?user=${encodeURIComponent(
    userData.username
)}`;
```
I checked and confirmed that. 

For the record, I tested it all by changing my username on the Django shell to `u.username = 'p éter@gmail.com'` and things still worked without encoding:
Screen Shot 2020-06-03 at 4 25 06 PM

peterbe:
So @schalkneethling @tobinmori @atopal What's up with this? Shall we add "is_contributor" to the `whoami` API first or not?
This PR, nor [the issue](https://github.com/mdn/kuma/issues/7077), have a ZenHub blocker on that. 
I'd rather we implement that first before proceeding with *this* PR.

schalkneethling:
> So @schalkneethling @tobinmori @atopal What's up with this? Shall we add "is_contributor" to the `whoami` API first or not?
> This PR, nor [the issue](https://github.com/mdn/kuma/issues/7077), have a ZenHub blocker on that.
> I'd rather we implement that first before proceeding with _this_ PR.

No there is 😄 https://github.com/mdn/kuma/issues/7077

atopal:
@peterbe  Yeah, let's do it.

peterbe:
Let’s file a new issue. And let’s make it a count instead of a Boolean.

schalkneethling:
> Let’s file a new issue. And let’s make it a count instead of a Boolean.

Here is the issue: https://github.com/mdn/kuma/issues/7194

schalkneethling:
@peterbe updated.

PR #7190, opened 2020-06-03
tell google translate to avoid pre tags

### LINKED ISSUE #7101: Code snippets gets translated by Google Translate
This should not happen. 
Screen Shot 2020-05-18 at 2 28 21 PM

From: https://translate.google.com/translate?sl=auto&tl=sv&u=https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FLearn%2FCSS%2FFirst_steps%2FWhat_is_CSS

peterbe:
Fixes #7101

I've tested this quite extensively using the Chrome extension and using https://translate.google.com and it works. Very satisfying. 
But what to be tested is the HTML our Python code produces. You can either sacrifice one of your Wiki docs in your docker-compose and just paste HTML snippets into it and hit save and see what it produces. Basically, these:
```html
<pre>something</pre>
<pre class="syntaxbox">something</pre>
```
...should become:
```html
<pre class="notranslate>something</pre>
<pre class="syntaxbox notranslate">something</pre>
```
Arguably, you can't argue with the unit test I wrote :)

What's important to observe is if the `pre` tags still look right. I.e. this new class doesn't interfere with Prism.

### DIFF
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -193,6 +193,7 @@ def clean_content(content):
     parsed.injectSectionIDs()
     if not settings.ALLOW_ALL_IFRAMES:
         parsed.filterIframeHosts(settings.ALLOWED_IFRAME_PATTERNS)
+    parsed.annotatePreTags()
     return parsed.serialize()
 
 
@@ -365,6 +366,11 @@ class ContentSectionTool(object):
     @newrelic.agent.function_trace()
     def annotateLinks(self, base_url):
         self.stream = LinkAnnotationFilter(self.stream, base_url)
+        return self
+
+    @newrelic.agent.function_trace()
+    def annotatePreTags(self):
+        self.stream = PreNotranslateClassFilter(self.stream)
         return self
 
     @newrelic.agent.function_trace()
@@ -553,6 +559,27 @@ class LinkAnnotationFilter(html5lib_Filter):
 
                         add_to_attr("class", links[href]["classes"])
                         add_to_attr("rel", links[href]["rel"])
+
+                token["data"] = attrs
+
+            yield token
+
+
+class PreNotranslateClassFilter(html5lib_Filter):
+    """
+    Filter which annotates <pre> tags with the 'notranslate' class.
+    """
+
+    def __iter__(self):
+        input = html5lib_Filter.__iter__(self)
+        for token in input:
+            if token["type"] == "StartTag" and token["name"] == "pre":
+                attrs = dict(token["data"]) or {(None, "class"): ""}
+                for (namespace, name), value in attrs.copy().items():
+                    if name == "class" and "notranslate" not in value:
+                        before = attrs.get((namespace, "class")) or ""
+                        after = f"{before} notranslate".strip()
+                        attrs[(namespace, "class")] = after
 
                 token["data"] = attrs
 
--- a/kuma/wiki/content.py
+++ b/kuma/wiki/content.py
@@ -576,7 +576,7 @@ class PreNotranslateClassFilter(html5lib_Filter):
             if token["type"] == "StartTag" and token["name"] == "pre":
                 attrs = dict(token["data"]) or {(None, "class"): ""}
                 for (namespace, name), value in attrs.copy().items():
-                    if name == "class" and "notranslate" not in value.lower().split():
+                    if name == "class" and "notranslate" not in value.split():
                         before = attrs.get((namespace, "class")) or ""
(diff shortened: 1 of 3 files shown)


peterbe:
Man that was hard! It wasn't that hard to write the core python code that loops over the tokens. It was just plain Python workfun. But man it's a bundled ness of madness the way the extractor works and the changes to the unit tests might look simple here in the final form but it required a serious taking-apart-everything kinda debugging session.

kuma/wiki/content.py
```diff
@@ -559,6 +565,27 @@ def add_to_attr(attr_name, add_list):
...
+    Filter which annotates <pre> tags with the 'notranslate' class.
+    """
+
+    def __iter__(self):
+        input = html5lib_Filter.__iter__(self)
+        for token in input:
+            if token["type"] == "StartTag" and token["name"] == "pre":
+                attrs = dict(token["data"]) or {(None, "class"): ""}
+                for (namespace, name), value in attrs.copy().items():
+                    if name == "class" and "notranslate" not in value:
```
escattone, line 579: Nit: This would miss annotating `pre` tags with a custom class name that is a superset of `notranslate`, like `notranslate:yada`, so it would be better to be exact with:
```suggestion
                    if name == "class" and "notranslate" not in value.split():
```

kuma/wiki/tests/test_content.py
```diff
@@ -994,6 +994,35 @@ def test_annotate_links_external_link():
...
+            "<pre>h1 { color:pink }</pre>",
+            '<pre class="notranslate">h1 { color:pink }</pre>',
+        ],
+        # 'notranslate' already set
+        [
+            '<pre class="notranslate brush:js">h1 { color:brown }</pre>',
+            # Note, no difference!
+            '<pre class="notranslate brush:js">h1 { color:brown }</pre>',
+        ],
+    ),
```
escattone, line 1016: Nit: Would be good to add human-readable `ids` here for easy identification of the case in the `pytest` output on failure, something like:
```suggestion
    ),
    ids=("simple", "noclass", "nochange"),
```

kuma/wiki/content.py
peterbe: Fixed.

kuma/wiki/tests/test_content.py
peterbe: Thanks! Fixed.

kuma/wiki/content.py
```diff
@@ -559,6 +565,27 @@ def add_to_attr(attr_name, add_list):
...
+    Filter which annotates <pre> tags with the 'notranslate' class.
+    """
+
+    def __iter__(self):
+        input = html5lib_Filter.__iter__(self)
+        for token in input:
+            if token["type"] == "StartTag" and token["name"] == "pre":
+                attrs = dict(token["data"]) or {(None, "class"): ""}
+                for (namespace, name), value in attrs.copy().items():
+                    if name == "class" and "notranslate" not in value.lower().split():
```
escattone, line 579: I was worried about the use of `.lower()` here, since I wasn't sure if something like `class="NOTRANSLATE"` would work the same as `class="notranslate"`, so I checked this using the Google Translate Chrome extension. It turns out that `class="NOTRANSLATE"` and `class="noTranslate"` do **not** work, but interestingly `class="NoTranslate"` does. Your call on whether we should also check for `NoTranslate`, but I think we should remove `.lower()` here:
```suggestion
                    if name == "class" and "notranslate" not in value.split():
```

peterbe: Good catch!
Also, the chances that someone already put in `<pre class="NOTransLate">` are extremely rare. Let's do your suggestion.

kuma/wiki/tests/test_content.py
```diff
@@ -994,6 +994,45 @@ def test_annotate_links_external_link():
...
+        ],
+        # 'notranslate' already set
+        [
+            '<pre class="notranslate brush:js">h1 { color:brown }</pre>',
+            # Note, no difference!
+            '<pre class="notranslate brush:js">h1 { color:brown }</pre>',
+        ],
+        [
+            '<pre class="NOTranslate">h1 { color:brown }</pre>',
+            '<pre class="NOTranslate">h1 { color:brown }</pre>',
```
escattone, line 1018: It looks like you forgot to add the `notranslate` in the expected result here:
```suggestion
            '<pre class="NOTranslate notranslate">h1 { color:brown }</pre>',
```

And, by the way, thanks for adding a test for that case!

PR #7196, opened 2020-06-04
Update to mobile header

### LINKED ISSUE #7146: T - Update to header on mobile
With the initial implementation of the new responsive header, we made a compromise with regards to the placement of the sign-in/user-menu link on mobile.

Screenshot 2020-05-28 at 10 46 05

Screenshot 2020-05-28 at 10 46 38

After speaking to @malqinneh, the following update was proposed[https://github.com/mdn/kuma/issues/7010#issuecomment-633739965] as a better option.

82847484-4e204780-9ebc-11ea-9320-0b2fbe9c2d0e

schalkneethling:
Update positioning of the sign in link, search icon, and user avatar on
mobile devices to match updated design

fix #7146


kuma/javascript/src/header/search.jsx
```diff
@@ -65,8 +65,12 @@ export default function Search({ initialQuery }: Props) {
...
             </form>
             <button className="toggle-form" onClick={handleClick}>
-                <CloseIcon className="close-icon" />
-                <SearchIcon className="search-icon" />
+                <CloseIcon
+                    className={showForm ? 'close-icon' : 'close-icon hide'}
+                />
+                <SearchIcon
+                    className={showForm ? 'search-icon hidden' : 'search-icon'}
+                />
```
peterbe, line 73: Why are we doing this? Why have it in the DOM if the CSS sets `visibility: hidden;`?
There's probably an explanation. A11y maybe? or something. Either way, we need a code comment that explains in to the future reader why we're not using React of this. 

Secondly, if there's a good reason for using CSS instead of React, you probably have a typo in there. `hide  != hidden`.

peterbe: Also, have you considered doing `<SearchIcon style={!showForm && {visibility:'hidden'}}>` instead? 
Or, perhaps put this SCSS in a dedicate `search.scss` so that when/if we refactor this code we can't forget to undo the hack from such a fundamental file as `base/_utilities.scss` where it's likely to get stuck forever.

schalkneethling: > Why are we doing this? Why have it in the DOM if the CSS sets visibility: hidden;?

I cannot remember 100% why we kept it like this. I and Mindy did discuss this and there was a good reason for not swapping the actual icons. I will try and dig up the reasoning. It was not for accessibility reasons because we have the text 'Close search' and 'Open search'. I believe it has to do with the animation on interaction.

> Secondly, if there's a good reason for using CSS instead of React, you probably have a typo in there. hide != hidden.

This is intentional. The close icon is animated along with the form and so, we cannot use `display: none;` as transitions will not work. I, therefore, set the one to hidden(`display:none`) and the other to hide(`visibility:hidden`)

The utility classes `hidden` and `hide` will be/is being used elsewhere so, keeping them in utils makes sense.

schalkneethling: Yup, it was to enable animation:
https://github.com/mdn/kuma/pull/7052/files#r424314565

schalkneethling:
@peterbe Updated. You will see from the comments that I now add/remove the search icon SVG as it is not being animated.

I also added the following to the precommit hook:

```
"test:precommit": "jest --testPathIgnorePatterns tests/sass -o --u",
```

This will run Jest tests and update snapshots, but only if there are relevant changed files since the last commit. https://jestjs.io/docs/en/cli.html#running-from-the-command-line

peterbe:
> I also added the following to the precommit hook:

Let's never do snapshots again :) ...in yari and any other JS project.

peterbe:
BEFORE:
IMG_40F5F1CC885E-1

AFTER:

peterbe:
Now, both the logo and the word "Sign in" is cut off.

peterbe:
At least the mugshot isn't cut off when you're signed in. 
IMG_6E76C4B012A1-1

If you're wondering how I faked that, I injected this hack in `kuma/api/v1/views.py`:
```diff
diff --git a/kuma/api/v1/views.py b/kuma/api/v1/views.py
index 390308ce8..62f55cba4 100644
--- a/kuma/api/v1/views.py
+++ b/kuma/api/v1/views.py
@@ -62,7 +62,8 @@ def whoami(request):
     authenticated or anonymous.
     """
     user = request.user
-    if user.is_authenticated:
+    user = User.objects.get(username="peterbe")
+    if user.is_authenticated or 1:
         data = {
             "username": user.username,
             "is_authenticated": True,
```

schalkneethling:
Bah! More often than not responsive design does not paint the entire picture.

> note to self: take more time and test on all the things.

for one thing, we should shrink the logo somewhat on mobile devices. Space up there is getting cramped and I do not want to decrease the touch area just to get everything to fit.

schalkneethling:
Thanks again for the review @peterbe. Updated. Tested now also on iPhone 7 and 8

## iPhone 8

Screenshot 2020-06-07 at 17 31 45

## iPhone 7

Screenshot 2020-06-07 at 17 32 20

peterbe:
image

That’s what it looks like with a soapbox on.

peterbe:
But this is without soapbox. 

I saw we ignore the first problem. 
Clearly it’s working now. 
image

peterbe:
When logged in. 
image

peterbe:
Shall we worry about the soapbox?

schalkneethling:
> Shall we worry about the soapbox?

For sure, but I will spin up another issue for that one. Thanks @peterbe 🍵

peterbe:
Then feel free to merge this if you think we can live without it until we solve it for soapbox. 
My worry is that we forget and it gets deprioritized on the basis that soapboxes don't always appear. 
So perhaps, it might be worth keep plugging away here on *this* PR.

PR #7218, opened 2020-06-10
Add user details component

### LINKED ISSUE #7097: T - Acct Settings: user details component form
# Acceptance criteria

- [ ] user form with name/static email/username/lang (slide 2)

# Q:
- [ ] check language dropdown size

schalkneethling:
Add user details component to account settings

fix #7097

### DIFF
--- a/kuma/accountsettings/jinja2/accountsettings/index.html
+++ b/kuma/accountsettings/jinja2/accountsettings/index.html
@@ -6,15 +6,13 @@
 {{ super() }}
 {% stylesheet 'accountsettings' %}
 {% stylesheet 'accountsettings-titlebar' %}
+{% stylesheet 'accountsettings-user-details' %}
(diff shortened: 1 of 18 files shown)


schalkneethling:
> I'm not reviewing but I think you should switch to controlled components and state instead.
It's much more predictive and generally universally encouraged in the React community.
Then you'd drop the name and defaultValue attributes and instead add value and onChange.

> I guess there's supposed to be a fetch phase to get the existing data. Is that blocking this PR?

100% Agree with you @peterbe but, as @Gregoor also mentioned, I wanted to get the UI portion out the door so the focus can be on wiring up the back/middle portions in the next sprint

> Also, where is the left-hand side menu?

After all the discussions we had, account settings have been vastly simplified and in this new state, there is no need for a left hand nav.

PR #7264, opened 2020-06-22
Sendinblue unsubscribe webhook

### LINKED ISSUE #7171: T - Sendinblue webhook for users unsubscribing through the mail
Schalk made me aware that this wasn't implemented yet. It means the user profile wouldn't accurately show a user's subscription status, if they unsubscribe through the link in the email. And worse, they'd get re-subscribed if they changed their email / payment status.

API docs (relevant for implementation):
https://developers.sendinblue.com/reference#createwebhook

Gregoor:
Fixes #7171 

## What it does
- renames `STRIPE_WEBHOOK_HOSTNAME` to `CUSTOM_WEBHOOK_HOSTNAME`, to make it reusable with other webhooks
- adds a system check which creates a sendinblue webhook, listening to the `unsubscribe` event (analogous to what we did for Stripe)
- adds an associated view route for handling the webhook
  - it differs from stripe in that sendinblue doesn't offer signing of the events, so we only interpret the hook being (correctly) called as a nudge, which we then use to contact sendinblue and check whether something did change
  - currently we only do that for checking whether a given email is still in the system (and in the correct sendinblue list) and then set the associated user's `is_newsletter_subscribed` toggle accordingly

## How to test it
- start ngrok (proxiying to whatever port you run kuma on)
- add your custom hostname to your .env (`CUSTOM_WEBHOOK_HOSTNAME`)
- start kuma
- 🔍 after it says something like "system checks found no problems", check https://my.sendinblue.com/advanced/webhook whether a new webhook was created (assuming you've already configured sendinblue)

and then
- have a user in your db with `is_newsletter_subscribed` set to **true** and their email to `example@example.com`
  - you can set the former through the UI, but doing the latter through the UI will lead to that email being in sendinblue, which defeats the purpose of what you're about to test (the webhook event will lead to a refresh, which finds a user still in the system, hence no change)
- go back to https://my.sendinblue.com/advanced/webhook, click Actions and then "send a test"
- 🔎 check your DB / user profile whether the user is now marked as not subscribed

### DIFF
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -633,7 +633,7 @@ def test_sendinblue_unsubscribe(mock_check_sendinblue, client):
 
     email = "testuser@example.com"
 
-    create_user(
+    user = create_user(
         save=True, username="testuser", email=email, is_newsletter_subscribed=True,
     )
 
@@ -644,4 +644,5 @@ def test_sendinblue_unsubscribe(mock_check_sendinblue, client):
     )
     assert response.status_code == 200
 
-    assert not User.objects.get(email=email).is_newsletter_subscribed
+    user.refresh_from_db()
+    assert not user.is_newsletter_subscribed
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -3,7 +3,6 @@ from types import SimpleNamespace
 from unittest import mock
 
 import pytest
-
 from django.conf import settings
 from django.core import mail
 from django.test import Client
@@ -20,7 +19,7 @@ from kuma.core.ga_tracking import (
 from kuma.core.tests import assert_no_cache_header
 from kuma.core.urlresolvers import reverse
 from kuma.search.tests import ElasticTestCase
-from kuma.users.models import User, UserSubscription
+from kuma.users.models import UserSubscription
 from kuma.users.tests import create_user
 from kuma.wiki.models import BCSignal
 
--- a/kuma/api/v1/tests/test_views.py
+++ b/kuma/api/v1/tests/test_views.py
@@ -21,7 +21,7 @@ from kuma.core.tests import assert_no_cache_header
 from kuma.core.urlresolvers import reverse
 from kuma.search.tests import ElasticTestCase
 from kuma.users.models import User, UserSubscription
-from kuma.users.tests import user
+from kuma.users.tests import create_user
 from kuma.wiki.models import BCSignal
 
 
@@ -483,7 +483,7 @@ def test_stripe_payment_succeeded_sends_invoice_mail(
     }
     download_url.return_value = bytes("totally not a pdf", "utf-8")
 
-    testuser = user(
+    testuser = create_user(
         save=True,
         username="testuser",
         email="testuser@example.com",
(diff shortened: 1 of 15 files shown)


peterbe:
Before:
```
>>> u = User.objects.get(username='peterbe2')
>>> u.is_newsletter_subscribed
True
>>> u.email
'example@example.com'
```
After ("Send a Test"):
```
>>> u.refresh_from_db()
>>> u.is_newsletter_subscribed
False
>>>
```

🍾

kuma/api/v1/tests/test_views.py
```diff
@@ -624,3 +624,24 @@ def test_stripe_subscription_canceled_sends_ga_tracking(
...
+    )
+
+    response = client.post(
+        reverse("api.v1.sendinblue_hooks"),
+        content_type="application/json",
+        data={"event": "unsubscribe", "email": email},
+    )
+    assert response.status_code == 200
+
+    assert not User.objects.get(email=email).is_newsletter_subscribed
```
peterbe, line 647: First of all, `user` is a terrible name for a function. If you want to search-and-replace it with `create_user` I'd approve that. 
Either way, I think you should hold on to the User instance and replace this line here with:
```suggestion
    my_user.refresh_from_db()
    assert not my_user.is_newsletter_subscribed
```
It's a nit but it means you're absolutely certain it's the same object and not someone else.

kuma/users/newsletter/checks.py
```diff
@@ -29,8 +29,62 @@ def sendinblue_check(app_configs, **kwargs):
...
+        return [
+            Error(
+                f"Error getting sendinblue webhooks: {response.status_code}",
+                id=SENDINBLUE_API_ERROR,
+            )
+        ]
+
+    if not no_webhooks_found:
+        for webhook in response.json()["webhooks"]:
+            if webhook["url"] == url and set(events) == set(webhook["events"]):
```
peterbe, line 69: Shouldn't that be:
```suggestion
            if webhook["url"] == url and set(events) & set(webhook["events"]):
```
I.e. an overlap.

kuma/api/v1/views.py
```diff
@@ -284,3 +285,22 @@ def _download_from_url(url):
...
+@require_POST
+@never_cache
+def sendinblue_hooks(request):
+    # Sendinblue does not sign its webhook requests, hence the event handlers following
+    # are different from the Stripe ones, in that they treat the event as a notification
+    # of a _potential_ change, while still needing to contact sendinblue to verify that
+    # it actually happened.
+    event = json.loads(request.body)
+
+    if event["event"] == "unsubscribe":
```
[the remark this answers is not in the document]
Gregoor: Thank you, my intuition went against this but after some consideration I overcame my intuition and agree with you that this is strictly an improvement.

kuma/users/newsletter/checks.py
Gregoor: my bad reason for doing it: I did the same thing we do when setting up the stripe webhook. But I agree that a subset of the events would be enough. Kinda. Then again, we would then also tolerate webhooks that send more events than we can't handle, so maybe equality checking is better in that regard?

Gregoor: I'm gonna merge the PR because I'm not leaning either way on this one but I feel like it's a discussion worth having (or keeping in mind)

PR #7269, opened 2020-06-23
fix(compat‑tables): Update platform borders and merge selector lists

ExE-Boss:
The **JavaScript** platform borders became outdated when **IE Mobile** was removed, and the **WebExtension** platform borders became outdated when **Safari 14** added support for **WebExtensions**.

I also merged the selector lists of the border declarations.

---

Fixes <https://github.com/mdn/kuma/issues/7274>


peterbe:
@Gregoor I don't know anything about this PR but it definitely makes me nervous that if this is right to merge, we're going to need to learn from it in Yari too.

PR #7303, opened 2020-07-01
redis docker image version should match production

### LINKED ISSUE #7270: T - docker-compose.yml should use the matching Redis version
**Summary**
See https://github.com/mdn/kuma/blob/1f294850f523788ff80fcb9fe8b60d279ece6ec2/docker-compose.yml#L108
We're not using the same version as we use in AWS ElastiCache. 

**Rationale**
That means that we get a false sense of reality in local development. 

**Audience**
Dev

**Proposal**
If we use Redis 5.1.2 in ElasiCache we should make that line look like this:
```yaml
  redis:
    image: redis:5.1.2
```

**Additional context**
We should confirm that our [version of mysql](https://github.com/mdn/kuma/blob/1f294850f523788ff80fcb9fe8b60d279ece6ec2/docker-compose.yml#L82) and [version of elasticsearch](https://github.com/mdn/kuma/blob/1f294850f523788ff80fcb9fe8b60d279ece6ec2/docker-compose.yml#L99) also matches what we use in RDS and Elastico.

escattone:
Fixes #7270 

Currently, the stage and production Redis instances are running with version `5.0.6`. Let's use the same version within `docker-compose.yml`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -105,7 +105,10 @@ services:
       - esdata:/usr/share/elasticsearch/data
 
   redis:
-    image: redis
+    # The version of this Redis image should be kept synchronized with production. It
+    # should match the "redis_engine_version" variable in the following Terraform file:
+    # https://github.com/mdn/infra/blob/master/apps/mdn/mdn-aws/infra/modules/multi_region/redis/variables.tf
+    image: redis:5.0.6
 
   kumascript:
     image: mdnwebdocs/kumascript


docker-compose.yml
```diff
@@ -105,7 +105,8 @@ services:
       - esdata:/usr/share/elasticsearch/data
 
   redis:
-    image: redis
+    # The version of this Redis image should be kept synchronized with production.
```
peterbe, line 108: Would it be worth linking to what that is?
E.g. a link to https://github.com/mdn/infra/blob/1c842b5fb81b0eccf8624852e0dffca8ac13ab65/apps/mdn/mdn-aws/infra/modules/multi_region/redis/variables.tf#L41
Or, perhaps https://github.com/mdn/infra/blob/master/apps/mdn/mdn-aws/infra/modules/multi_region/redis/variables.tf in general. 
That way, it's easier to remember where to look that up.

escattone: 👍 I'll do that.

PR #7330, opened 2020-07-09
Add middleware for redirecting to a dev origin

Gregoor:
I don't really know what to say that I didn't already try to capture in the middleware's code comment. So I'll just link to that:

https://github.com/Gregoor/kuma/blob/e958abb4f66b9845f4ed9cb4d83bd16e24c9c7d6/kuma/core/middleware.py#L358-L366

### DIFF
--- a/kuma/core/middleware.py
+++ b/kuma/core/middleware.py
@@ -1,4 +1,4 @@
-from urllib.parse import urlsplit
+from urllib.parse import urlencode, urlparse, urlsplit
 
 from django.conf import settings
 from django.contrib.sessions.middleware import SessionMiddleware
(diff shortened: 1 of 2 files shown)


peterbe:
After much discussion, we opted for a simpler solution. 
https://github.com/mdn/kuma/pull/7332 makes it so that Kuma can be OK with redirecting back to Yari's dev server after successfully authenticating. But it requires no state. But it also means that, unlike the solution in *this* PR, if you "fall into" kuma (aka. `localhost.org:8000`) you're stuck and you'd have to manually change the address bar to `s/8000/3000` to get back into Yari-land.

PR #7332, opened 2020-07-09
allow localhost.org:3000 in the 'next' URL

### LINKED ISSUE #7331: Make ?next=... respect the Yari local dev hostname
**Summary**
When you go to `http://localhost.org:8000/users/github/login/?next=$NEXT` that `$NEXT` can't contain a hostname that is different from `localhost.org:8000` or else it gets dropped. For (good) security reasons. 
We need to be able to override that for when you're doing local development in Yari with authentication.

**Audience**
Dev. 

**Rationale**
We want to be very strict,  careful, and deliberate. Aka. it's tied down hard to very specific things and ONLY when doing *local* Kuma development. 

**Workaround**
If this doesn't work, when you start up Yari (http://localhost.org:3000) and click sign in, you have to manually edit the addressbar to go back to `:3000` instead of `:8000`. 

**Proposal**
I have a hack in mind already that I know works. 

**Additional context**
It doesn't have to be perfect because it's going to be made very explicit and exclusively only for the select few "hardcore" developers who work on Yari with authentication.

peterbe:
Fixes #7331

Simply set 
```
ADDITIONAL_NEXT_URL_ALLOWED_HOSTS=localhost.org:3000
```
in your `.env` and start up `docker-compose`. Now, you can use `http://localhost.org:8000/users/github/login/?next=http://localhost.org:3000/en-US/docs/Web/HTML/Element/audio` and when it's done doing the auth-provider dance it should redirect to `http://localhost.org:3000/en-US/docs/Web/HTML/Element/audio`.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1892,3 +1892,12 @@ SENDINBLUE_API_KEY = config("SENDINBLUE_API_KEY", default=None)
 SENDINBLUE_LIST_ID = config("SENDINBLUE_LIST_ID", default=None)
 
 SITEMAP_USE_S3 = config("SITEMAP_USE_S3", cast=bool, default=True)
+
+# When doing local development with Yari, if you want to have `?next=...` redirects
+# work when you sign in on Yari, this needs to be set to `localhost.org:3000` in your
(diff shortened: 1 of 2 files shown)


Gregoor:
So cool that you found that option 🎉

PR #7403, opened 2020-08-06
link to existing close account functionality from account settings

### LINKED ISSUE #7402: T - Close Account UI
In https://github.com/mdn/kuma/issues/7012#issuecomment-669250828 it has been decided that for now, we will reuse the functionality already present in Kuma for user account deletion. As such, the close account work can be completed by changing the close account button to an anchor element and linking it to: https://developer.mozilla.org/en-US/profiles/%7BUSERNAME%7D/delete

schalkneethling:
Update the close account component on new account setting so that it
links to the existing delete account functionality.

@peterbe r?

fix #7402


kuma/javascript/src/accountsettings/components/close-account.jsx
```diff
@@ -1,13 +1,16 @@
...
-import CloseAccountForm from './close-account-form.jsx';
+type Props = {
+    locale: string,
+    userData: UserData,
+};
 
-const CloseAccount = () => {
-    const [showForm, setShowForm] = useState<boolean>(false);
+const CloseAccount = ({ locale, userData }: Props) => {
+    const { username }: any = userData;
```
peterbe, line 13: I don't understand this. Also, you can put this into the line above.

peterbe: I appreciate that it's a bit weird that type `UserData` makes the `username` optional. It makes sense and we should fix that, in Yari, so that you have an interface for authenticated and not authenticated users. E.g.
```typescript
interface BasicUser {
  isAuthenticated: boolean,
  waffle: ...,
  ...
}
interface AuthenticatedUser extends BasicUser {
  username: string;
  email: string; 
}
```

kuma/javascript/src/accountsettings/components/close-account.jsx
```diff
@@ -1,13 +1,16 @@
...
 import * as React from 'react';
-import { useState } from 'react';
 
+import { type UserData } from '../../user-provider.jsx';
 import { gettext } from '../../l10n.js';
 
-import CloseAccountForm from './close-account-form.jsx';
+type Props = {
+    locale: string,
+    userData: UserData,
```
peterbe, line 9:
```suggestion
    username: string,
```

kuma/javascript/src/accountsettings/components/close-account.jsx
```diff
@@ -1,13 +1,16 @@
...
+};
 
-const CloseAccount = () => {
-    const [showForm, setShowForm] = useState<boolean>(false);
+const CloseAccount = ({ locale, userData }: Props) => {
```
peterbe, line 12:
```suggestion
const CloseAccount = ({ locale, username }: Props) => {
```

kuma/javascript/src/accountsettings/components/close-account.jsx
peterbe, line 13:
```suggestion

```

kuma/javascript/src/accountsettings/pages/index.jsx
```diff
@@ -49,7 +49,7 @@ export default function LandingPage({ data, locale }: RouteComponentProps) {
                         userData={userData}
                         contributionSupportEmail={contributionSupportEmail}
                     />
-                    <CloseAccount />
+                    <CloseAccount locale={locale} userData={userData} />
```
peterbe, line 52:
```suggestion
                    <CloseAccount locale={locale} username={userData.username} />
```

kuma/javascript/src/accountsettings/components/close-account.jsx
schalkneethling: The above was for that exact reason. Flow does not like the fact that `username` can be null or undefined and so, does not allow me to coerce it into a string

schalkneethling:
@peterbe I did experiment with exactly this i.e. passing `username` as opposed to `userData` but, then I need to deal with essentially the same problem in `index.jsx`

![Screen shot showing the message "Cannot create `CloseAccount` element because  null or undefined [1] is incompatible with  string [2] in property `username`."](https://user-images.githubusercontent.com/10350960/89622409-4b19de00-d893-11ea-9840-5aeb4d460677.png)

So, in `pages/index.jsx` I could have the following to overcome the above problem:

```
export default function LandingPage({ data, locale }: RouteComponentProps) {
    const contributionSupportEmail = data.contributionSupportEmail;
    const userData = useContext(UserProvider.context);
    const { username }: any = userData;

    ...
    <CloseAccount locale={locale} username={username} />
```

schalkneethling:
I have updated this to be as in https://github.com/mdn/kuma/pull/7403#issuecomment-670383254

peterbe:
See https://github.com/schalkneethling/kuma/pull/7

schalkneethling:
> See [schalkneethling#7](https://github.com/schalkneethling/kuma/pull/7)

Merged.

PR #7408, opened 2020-08-10
correctly hide and show bcd signaling panels

### LINKED ISSUE #7309: T - Thank you section shown after sending browser compatibility report should be properly hidden from screen readers, not just visually hidden
**Summary**

The section shown (the h2 that says "thank you for your report") just after the browser compatibility table should not be visible screen readers when a report has not been sent but still is. It is only visually hidden, not completely hidden.

**Steps To Reproduce (STR)**
How can we reproduce the problem?_

1. With the NVDA screen reader and Firefox, visit any page that has the browser compatibility table such as https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
2. With NVDA, navigate to just after the browser compatibility table

**Actual behavior**

This section appears, starting with the h2 that thanks you for your report

**Expected behavior**

This should not appear at all

**Additional context**

schalkneethling:
Correctly hide and show the different panels of the BCD signaling component
for users using assistive technologies.

There are other much larger problems we need to address when implementing this in Yari, but this solve our immediate, ugly problem 😸 

@peterbe r?

fix #7309

### DIFF
--- a/kuma/static/styles/components/compat-tables/_bc-signal.scss
+++ b/kuma/static/styles/components/compat-tables/_bc-signal.scss
@@ -116,11 +116,10 @@ div.signal-link-container {
 }
 
 div.bc-signal-block {
+    display: none;
     background-color: $blue-light;
     box-sizing: border-box;
-    max-height: 0;
     overflow: hidden;
-    transition: max-height .3s ease-out, padding-top .1s ease .3s;
 
     &:not(.complete) {
         padding-top: 0;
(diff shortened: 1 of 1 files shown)


peterbe:
This disables the animation of the flap opening and closing. Are we OK with that? It's not a big loss if you ask me. Animations can be cool but here's probably not the right place for it. 

So, if we're going to go down this route, is the removal of the animation definitive or just a necessary evil for the benefit of a11y? I believe @Gregoor has volunteered to re-enable the BCD Signalling feature on Yari, in React. But even there, it might be worth to keeping it simple and avoiding animations.

schalkneethling:
> This disables the animation of the flap opening and closing. Are we OK with that?

Yup, we are.

> So, if we're going to go down this route, is the removal of the animation definitive or just a necessary evil for the benefit of a11y? I believe @Gregoor has volunteered to re-enable the BCD Signalling feature on Yari, in React. But even there, it might be worth to keeping it simple and avoiding animations.

We can have both, we just need to implement it a little differently 😸 There is also a larger problem here that we will solve which is source order.

schalkneethling:
Thank you for the quick review @peterbe!

PR #7413, opened 2020-08-11
updates to the main menu

### LINKED ISSUE #7220: T - Reduce height of mobile navigation
**Summary**

Originally reported in https://github.com/mdn/kuma/issues/7203

When the navigation menu is shown on mobile it takes over the entire height of the viewport. Our menu has very few initial entries when shown and so we can do without all the additional white space.

Interaction with the menu is also a bit jarring on a slow connection as it is not immediately clear that a new page load has been triggered.

**Steps To Reproduce (STR)**

1. Load any page on a mobile device
2. Expand the navigation
3. Expand one of the top-level entries
4. Tap on an individual entry

**Actual behavior**
The submenu collapses but, the overall overlay remain and it is not immediately clear that the tap event was captured.

**Expected behavior**
Reducing the amount of whitespace initially shown and improving the behavior of the menu on tap events will make for better overall user experience.

schalkneethling:
Updates to the main menu backported from the design system.

This:

1. Improves accessibility be resolving the HTML structure and ARIA roles
2. Moves focus to the top-level menu item when a submenu is closed
3. Correctly label submenus
4. Reduces the height of the navigation on mobile and shows an overlay that greys out the underlying content

@peterbe R?

fix #7220


schalkneethling:
> I used https://ace.accessibe.com/?website=https://dae25b9f30b9.ngrok.io/en-US/ and found some a11y warnings but I don't think any of it is regressions.

thanks for checking @peterbe - we have more accessibility issues to address for sure. Those are mostly known

schalkneethling:
> I don't like that it's not using React patterns (i.e. using that `document.querySelector('.page-overlay')` stuff) but I don't think it matters in the sinking ship that is Kuma.

Out of curiosity and learning, how would should we do this in a more React way. I thought because there was no way for me to get at it using `useRef`, this was the way to go?

PR #7420, opened 2020-08-14
update default WELCOME_EMAIL_FROM

### LINKED ISSUE #7419: Remove Janet Swisher's name from account welcome emails
**Summary**
When a user signs up, the welcome email message they receive should not contain Janet Swisher's name.

**Rationale**
Janet will no longer be on MDN staff.

**Audience**
Users signing up for accounts

**Proposal**
Remove Janet's name from the "From" field, and from the body of the message. Your call on what to do instead. You could use someone else's name, or "MDN team", or something else. Might want to switch back to a no-reply address if no one will be responding to people who reply to the welcome email.

**Additional context**
I'll miss y'all very much.

peterbe:
Fixes #7419

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -1578,8 +1578,7 @@ SESSION_CLEANUP_CHUNK_SIZE = config(
 
 # Email address from which welcome emails will be sent
 WELCOME_EMAIL_FROM = config(
-    "WELCOME_EMAIL_FROM",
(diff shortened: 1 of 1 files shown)


peterbe:
Seems we're still using the default from .py in prod:
```
>>> from django.conf import settings
>>> settings.WELCOME_EMAIL_FROM
'Janet Swisher - MDN Community Lead <mdn-admins@mozilla.org>'
>>> import os
>>> os.environ.get('WELCOME_EMAIL_FROM')
```

PR #7434, opened 2020-08-19
run python lint in GHA

peterbe:
Let's see.


peterbe:
I have to merge it to master so that the tests will start to run so I can optimize it from there.

PR #7438, opened 2020-08-19
js lints in GHA

peterbe:


.github/workflows/js-lints.yml
```diff
@@ -0,0 +1,59 @@
+name: JavaScript Lints
```
schalkneethling, line 1: Minor change as this seems to also lint SCSS, perhaps this should read:

```suggestion
name: JavaScript and SASS Lints
```

schalkneethling (approved):
Other than the one suggestions, this is an r+ Thanks @peterbe

.github/workflows/js-lints.yml
peterbe: Why not "SCSS"?

PR #7457, opened 2020-08-25
Fix search widget's border being cut off by input white background

joaopaulobdac:
# Current visual

Tested on Chromium 84.0.4147.135 (Official Build) Arch Linux (64-bit)

(notice the border)

# The fix

This PR changes the search widget's input background to transparent, but it's safe because the parent `.header-search form` has a `#fff` background as well.

### DIFF
--- a/kuma/static/styles/minimalist/components/_search-widget.scss
+++ b/kuma/static/styles/minimalist/components/_search-widget.scss
@@ -64,6 +64,7 @@
         color: $text-color;
         border-width: 0;
         min-width: 60px;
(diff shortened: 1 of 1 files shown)


peterbe:
Sorry about the linting CI error. It was a mistake our end of how we test things. 
I think we can simply ignore it as it's being fixed by https://github.com/mdn/kuma/pull/7459

schalkneethling (approved):
r+ Thanks @joaopaulobdac 🥳

PR #7459, opened 2020-08-26
fix eslint error

peterbe:
https://eslint.org/docs/rules/no-prototype-builtins


peterbe:
Sigh. The strategy of only running JS Lints if any of the .js, .jsx, or .scss files change means we don't find out that our existing code breaks when new `eslint` upgrades come out etc. 
I think we need to reassess that strategy.

schalkneethling (approved):
r+ Thanks @peterbe

PR #7463, opened 2020-08-27
run js lints when yarn.lock gets upgraded

peterbe:
If a PR comes in that edits any `kuma/**/*.(js|jsx|scss)` the "JavaScript and SASS Lints" workflow kicks in. 
But if Dependabot upgrades something like `prettier` or `eslint` in the `yarn.lock` we wouldn't notice if our code is compatible with this. This change hopefully fixes it. 

I learned about this technique here: https://github.com/technote-space/get-diff-action/issues/111#issuecomment-681387855

### DIFF
--- a/.github/workflows/js-lints.yml
+++ b/.github/workflows/js-lints.yml
@@ -21,6 +21,9 @@ jobs:
             .jsx
             .scss
           PREFIX_FILTER: kuma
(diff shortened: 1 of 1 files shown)


schalkneethling (approved):
r+

PR #7486, opened 2020-09-08
GitHub action that runs docker compose

peterbe:
The only difference between *this* PR and https://github.com/mdn/kuma/pull/7456 is that this one is made on the `mdn` repo rather than from a forked repo.


peterbe:
@escattone Here's a status update. The original feedback you mentioned was that it never builds the docker image. That was easily fixed by introducing a `docker-compose build` plus the fact that the `docker-compose.yml` now never mentioned the image. This basically makes us (at least of `kuma`) independent of DockerHub.com. 
The problem with this is that we're never getting any caching hits so the whole GHA takes about the same amount of time as it did when we use `.travis.yml`. 
At this point now they're "equal". 

We've established that `docker-compose build` is fast on a laptop because the docker you have has caching built in. Basically, on a clean clone of kuma running `docker-compose build` once is slow, but the second time it's fast. 

That's where I started to see if we can use that fancy GitHub Action that's using GitHub Container Registry. I'm not, yet, fully able to make that work due to permissions. But **if it that caching problem can't be fixed, it's still attractive since it's getting rid of TravisCI**. 

I haven't given up yet that's the escape hatch. In this branch, we roll back the attempts to push caching layers and we're still looking at ~15min but it's all GitHub Action.

peterbe:
@escattone With TravisCI going down, it's probably worth getting off it sooner rather than later. 

If I recall correctly, here's how this work went/is going:

1. Switch to GHA (but use Dockerhub). It's awesomely fast.
2. Ryan points out we should run the docker-compose build so we test new changes to poetry or Dockerfile that hasn't been pushed to the registry. 
3. I put those back, now it's 100% GHA but the same total build time as Travis. At least it's not any Travis anymore.
4. I try to use the new GitHub registry as a caching strategy for Docker. I couldn't figure out how to get it to work. Also not super engaged and focused. Plus a lot of that tech is new and a lot of people keep pestering GitHub to add it to `actions/cache` so it's seamless. 
5. I remove all attempts to be smart with the docker layer caching with the GitHub registry. 

I think we'll soon be starting to shut down much of the Kuma code as we piece-meal destroy the Wiki parts. 
Soon enough our Dockerfile and our pyproject.toml will become sufficiently small that it doesn't have to take 15 minutes to `docker-compose build`!

Since this change is basically just a switch from one provider to another, we can land this whenever. So, r?

escattone (approved):
Thanks @peterbe!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,69 @@
...
+          # We might be able to use something
+          # like `if [ ${{ github.head_ref}} eq 'master']` but not sure.
+          # Slower...
+          docker-compose run testing make coveragetest
+          # Faster...
+          #docker-compose run testing make test
+
+      - name: Submit code coverage
+        run: |
+          bash <(curl -s --retry 3 --retry-connrefused https://codecov.io/bash)
```
jezdez, line 54: I've had more luck with using their own GitHub action btw!

peterbe: Didn't even know that was a thing. 
Truth be told, I've advocated we remove Codecov anyway, but it wasn't entirely agreed upon. I never look at it and it's often "too noisy" in the PR UI. 
Hopefully, soon enough Kuma will shrink down to so little code that it might give us a sense of fresh air to revisit things.

PR #7495, opened 2020-09-10
use black 20.8b1

peterbe:
To update the lock file *and* the `pyproject.toml` I used: `poetry add --dev black@20.8b1`


schalkneethling (approved):
r+

PR #7509, opened 2020-09-21
add mdn browser compat report banner

### LINKED ISSUE #7505: Soapbox banner required for MDN browser comatibility report
# Content Issue

- [ ] Please close this issue, I accidentally submitted it without adding any details
- [ ] New documentation
- [x] Correction or update

# URL of page you are seeing the problem on:

All MDN pages ;-)

# Details

To accompany the launch of the MDN browser compatibility report (see https://github.com/mdn/insights/issues/42), we'd like to put up a soapbox banner to draw people's attention to it.

The wording should be "Check out the MDN Browser Compatibility Report, a deep dive into web compatibility frustrations, with useful insights into how they could be overcome.", and "MDN Browser Compatibility Report" should be linked to the report.

We'd like this banner to run until the day before the soapbox banner for the 2020 Web DNA Survey is published (see https://github.com/mdn/kuma/issues/7492)

Does this sound OK?

schalkneethling:
Add banner to notify people of the MDN browser compatibility report.
Waiting to hear whether we are linking to the site or the report directly.

@peterbe r? /cc @chrisdavidmills ^

fix #7505

### DIFF
--- a/kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
+++ b/kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def create_browser_compat_report_banner_switch(apps, schema_editor):
+    Switch = apps.get_model("waffle", "Switch")
+    if not Switch.objects.filter(name="mdn_browser_compat_report").exists():
+        Switch.objects.create(
+            name="mdn_browser_compat_report", active=False, note="Enable/disable the MDN browser compat report banner"
+        )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0009_remove_user_interviews_banner_flag"),
+        # This is needed otherwise `apps.get_model('waffle', 'Flag')`
+        # will raise a Django app LookupError.
+        ("waffle", "0001_initial"),
+    ]
+
+    operations = [migrations.RunPython(create_browser_compat_report_banner_switch)]
--- a/kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
+++ b/kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
@@ -10,7 +10,7 @@ def create_browser_compat_report_banner_switch(apps, schema_editor):
         Switch.objects.create(
             name="mdn_browser_compat_report",
             active=False,
-            note="Enable/disable the MDN browser compat report banner",
+            note="The MDN Browser Compatibility Report",
         )
 
 
--- a/kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
+++ b/kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
@@ -8,7 +8,9 @@ def create_browser_compat_report_banner_switch(apps, schema_editor):
     Switch = apps.get_model("waffle", "Switch")
     if not Switch.objects.filter(name="mdn_browser_compat_report").exists():
         Switch.objects.create(
-            name="mdn_browser_compat_report", active=False, note="Enable/disable the MDN browser compat report banner"
+            name="mdn_browser_compat_report",
+            active=False,
+            note="Enable/disable the MDN browser compat report banner",
         )
 
 
--- a/kuma/javascript/src/active-banner.jsx
+++ b/kuma/javascript/src/active-banner.jsx
@@ -203,7 +203,9 @@ function MDNBrowserCompatReportBanner() {
                 'A deep dive into web compatibility frustrations, with useful insights into how they could be overcome.'
             )}
             cta={gettext('Read the report')}
-            url={'https://insights.developer.mozilla.org/'}
+            url={
+                'https://mdn-web-dna.s3-us-west-2.amazonaws.com/MDN-Browser-Compatibility-Report-2020.pdf'
+            }
             newWindow
         />
     );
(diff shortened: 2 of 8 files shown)


chrisdavidmills:
@schalkneethling I think it would make way more sense to link directly to the report in the banner as the main CTA*, although we could also link the word "insights" in the banner wording to the Insights site.

* "download this new interesting report" is a much more obvious actionable CTA than "go to our insights site, and check out the new report we've got available to download".

schalkneethling:
> @schalkneethling I think it would make way more sense to link directly to the report in the banner as the main CTA*, although we could also link the word "insights" in the banner wording to the Insights site.
> 
> * "download this new interesting report" is a much more obvious actionable CTA than "go to our insights site, and check out the new report we've got available to download".

Alrighty, CTA is pointing at the report now directly.

robnyman:
For the PDF report I think it's sort of bad practice linking directly to a PDF, unless the text clearly states it links to a PDF and ideally also shows the size of the PDF file.

Can you have it link both to the Insights section and to the PDF?

E.g. "As part of our Insights section [link] we have now published the MDN Browser Compatibility Report (PDF, 1.3MB) [link to PDF]: A deep dive into web compatibility frustrations, with useful insights into how they could be overcome."

foolip:
I agree with @robnyman  that it'd be good to do something about linking unknowing people directly to a large PDF. Perhaps adding just "(PDF, 1.9 MB)" after the link would do the trick, though? On https://insights.developer.mozilla.org/ it will be fairly easy to find/click the wrong thing, since the 2019 report will still be at the top and most prominent.

robnyman:
My only small worry there is that only a PDF link might deter some people from downloading it directly, while seeing the Insights section and the bigger project might make it seem worth it. That's why I thought both would be good.

But I might be overthinking this…

kuma/javascript/src/active-banner.jsx
```diff
@@ -189,9 +189,28 @@ function Banner(props: BannerProps) {
...
 
+export const MDN_BROWSER_COMPAT_REPORT_ID = 'mdn_browser_compat_report';
 export const DEVELOPER_NEEDS_ID = 'developer_needs';
 export const SUBSCRIPTION_ID = 'subscription_banner';
 
+function MDNBrowserCompatReportBanner() {
+    return (
+        <Banner
+            id={MDN_BROWSER_COMPAT_REPORT_ID}
+            classname="developer-needs"
```
peterbe, line 200: ??

peterbe: Also, `classname` when it should be `className`. Apparently, linting CI didn't notice :(

kuma/core/migrations/0010_create_browser_compat_report_banner_switch.py
```diff
@@ -0,0 +1,26 @@
...
+from django.db import migrations
+
+
+def create_browser_compat_report_banner_switch(apps, schema_editor):
+    Switch = apps.get_model("waffle", "Switch")
+    if not Switch.objects.filter(name="mdn_browser_compat_report").exists():
+        Switch.objects.create(
+            name="mdn_browser_compat_report",
+            active=False,
+            note="Enable/disable the MDN browser compat report banner",
```
peterbe, line 13:
```suggestion
            note="The MDN Browser Compatibility Report",
```

peterbe:
If driving traffic to insights.developer.mozilla.org is important, let's use that as the CTA URL. If not, let's modify the CTA text from `Read the report` to `Read the report (PDF, 1.8mb)`. 
Is there not a dedicated permalink for this report that is *not* the homepage for insights.developer.m.o?

kuma/static/styles/components/banners/browser-compat-report.scss
```diff
@@ -0,0 +1,170 @@
+@import '../../includes/vars';
+@import '../../includes/mixins';
+@import 'vars';
+
+.developer-needs {
```
peterbe, line 5: Obviously, this is related to that `classname` being a bit "wrong". I wouldn't mind you taking the time to come up with a generic name instead so this block of scss can be shared. Or, let this new banner use `developer-needs.scss` but I noticed that the SCSS isn't actually identical.

robnyman:
"Read the report (PDF, 1.8mb)" sounds just fine to me, in particular if we want to get this out tomorrow morning.

kuma/javascript/src/active-banner.jsx
```diff
@@ -189,9 +189,28 @@ function Banner(props: BannerProps) {
...
+            title={gettext('MDN Browser Compatibility Report')}
+            copy={gettext(
+                'A deep dive into web compatibility frustrations, with useful insights into how they could be overcome.'
+            )}
+            cta={gettext('Read the report')}
```
peterbe, line 205:
```suggestion
            cta={gettext('Read the report (PDF, 1.8mb)')}
```

As per @robnyman 's suggestion.

kuma/javascript/src/active-banner.jsx
peterbe, line 200:
```suggestion
            className="developer-needs"
```

kuma/javascript/src/active-banner.jsx
```diff
@@ -189,9 +189,28 @@ function Banner(props: BannerProps) {
...
 
+export const MDN_BROWSER_COMPAT_REPORT_ID = 'mdn_browser_compat_report';
 export const DEVELOPER_NEEDS_ID = 'developer_needs';
 export const SUBSCRIPTION_ID = 'subscription_banner';
 
+function MDNBrowserCompatReportBanner() {
+    return (
+        <Banner
+            id={MDN_BROWSER_COMPAT_REPORT_ID}
+            className="developer-needs"
```
peterbe, line 200:
```suggestion
            classname="developer-needs"
```

Oops. I didn't realize this hack uses that typical React prop name but in lowercase. Ugly. :(

escattone:
Closing this since it was folded into https://github.com/mdn/kuma/pull/7514 along with some changes.

PR #7517, opened 2020-09-22
Add GA to browser compat CTA

### LINKED ISSUE #7516: Add GA to Report download CTA
There is a request to add GA to track the number of clicks on the CTA to download the MDN Browser Compat Report.

schalkneethling:
This adds a call to GA when CTA button to download the report PDF is clicked. Confirmed with GA testing that the following will be sent on click:

```
Running command: ga("send", {hitType: "event", eventCategory: "mdn_browser_compat_report", eventAction: "Browser Compat Report CTA clicked", eventLabel: "banner"})
```

@peterbe r?

fix #7516

### DIFF
--- a/kuma/javascript/src/active-banner.jsx
+++ b/kuma/javascript/src/active-banner.jsx
@@ -194,6 +194,8 @@ export const DEVELOPER_NEEDS_ID = 'developer_needs';
 export const SUBSCRIPTION_ID = 'subscription_banner';
 
 function MDNBrowserCompatReportBanner() {
+    const ga = useContext(GAProvider.context);
+
     return (
         <Banner
             id={MDN_BROWSER_COMPAT_REPORT_ID}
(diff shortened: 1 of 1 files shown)


kuma/javascript/src/active-banner.jsx
```diff
@@ -216,6 +218,14 @@ function MDNBrowserCompatReportBanner() {
             url={
                 'https://mdn-web-dna.s3-us-west-2.amazonaws.com/MDN-Browser-Compatibility-Report-2020.pdf'
             }
+            onCTAClick={() => {
+                ga('send', {
+                    hitType: 'event',
+                    eventCategory: MDN_BROWSER_COMPAT_REPORT_ID,
+                    eventAction: 'Browser Compat Report CTA clicked',
```
peterbe, line 225: What is the context for how we've done this in the past? The reason I ask is; should we perhaps include **the year** in that action?

peterbe (approved):
It works. 
Screen Shot 2020-09-22 at 8 40 17 AM

But I think having the year in there might be a good idea.

kuma/javascript/src/active-banner.jsx
schalkneethling: We have not had this one in the past but, good idea 👍

PR #7539, opened 2020-10-01
Disable the payments endpoint

### LINKED ISSUE #7537: Disable the payments endpoint
Requested by @chrisdavidmills via Slack:

> Hey folks! Could we please hide this page — https://developer.mozilla.org/en-US/payments/ ? People keep finding it and asking how they can use it to support us

peterbe:
Fixes #7537

Pardon the intrusion @escattone but I think this is a much simpler solution. Basically, a rock-solid simple Django setting that, by default, disables subscriptions. Combined with waffle flags, all traffic to anything "payments" will 404. And this way the tests don't have to change.

### DIFF
--- a/kuma/payments/tests/test_views.py
+++ b/kuma/payments/tests/test_views.py
@@ -17,6 +17,15 @@ def test_payments_index(client, settings):
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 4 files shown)


escattone (approved):
@peterbe Love it, so much better, thank you!

Issue #7412, opened 2020-08-10
Migrate Selected MDN Pages


Emilfs:
**Summary**
_What is the problem?_
These pages needed to be migrated to in-tree (discussed with @sylvestre):
https://developer.mozilla.org/en-US/docs/Mozilla/Memory_Sanitizer
have redirects to
https://firefox-source-docs.mozilla.org/tools/sanitizer/memory_sanitizer.html#memory-sanitizer

https://developer.mozilla.org/en-US/docs/Debugging_Mozilla_with_gdb
have redirects to
https://firefox-source-docs.mozilla.org/contributing/debugging/debugging_mozilla_with_gdb.html

https://developer.mozilla.org/en-US/docs/Debugging_Mozilla_with_lldb
have redirects to
https://firefox-source-docs.mozilla.org/contributing/debugging/debugging_mozilla_with_lldb.html

https://developer.mozilla.org/en-US/docs/Understanding_crash_reports
have redirects to
https://firefox-source-docs.mozilla.org/contributing/debugging/understanding_crash_reports.html

https://developer.mozilla.org/en-US/docs/Debugging_a_minidump
have redirects to
https://firefox-source-docs.mozilla.org/contributing/debugging/debugging_a_minidump.html

https://developer.mozilla.org/en-US/docs/Debugging_Mozilla_with_Valgrind
have redirects to
https://firefox-source-docs.mozilla.org/contributing/debugging/debugging_mozilla_with_valgrind.html

https://developer.mozilla.org/en-US/docs/Debugging/Record_and_Replay_Debugging_Firefox
have redirects to
https://firefox-source-docs.mozilla.org/contributing/debugging/record_and_replay_debugging_firefox.html

**Steps To Reproduce (STR)**
_How can we reproduce the problem?_

1. 
2. 
3. 

**Actual behavior**
_What actually happened?_

**Expected behavior**
_What did you expect to happen?_

**Additional context**
_Is there anything else we should know?_

peterbe:
@escattone It's assigned to you now. Does it still need to be triaged?

escattone:
@peterbe I think it needs to be triaged in terms of priority. I suppose I will just do it as if it's a P1, but I'd like Hermina to know about these kinds of requests just so she's aware of the kinds of support requests that we get.

sylvestre:
Sphinx removed a few of these pages already i think

Le lun. 24 août 2020 à 23:42, Ryan Johnson <notifications@github.com> a
écrit :

[...]

PR #7540, opened 2020-10-01
Add firefox source docs redirects (#7412)

fiji-flo:
* add firefox source docs redirects


escattone (changes requested):
@fiji-flo This looks and works great, except for two things. I've submitted them as a PR against this PR because I had to test them in order to get to the bottom of what went wrong. See https://github.com/fiji-flo/kuma/pull/1.

escattone (approved):
@fiji-flo Awesome, thank you!

PR #7661, opened 2020-12-07
fix: profile view - submenu obscured

### LINKED ISSUE #7620: Error on MDN profile page: Sub menu not visible upon hover at viewport widths larger than 1024px
@tannerdolby commented on [Mon Nov 16 2020](https://github.com/mdn/sprints/issues/3893)

I noticed this issue on my [profile page](https://developer.mozilla.org/en-US/profiles/tannerdolby) at `/en-US/profiles/username` where the sub menu column is only 50% visible as its content is overflowing the parent container. 

Within the `<div class="user-cards">` container, when I hover over `<button id="advanced-menu">..</button>` the `submenu-column` container appears but the second `<li>` item isn't visible since the content overflows and does not provide a scroll option. 

The scrolling mechanism is disabled for the underlying parent container `<section id="user-head">..</section>` since it has `overflow: hidden;` declared. 

After I added `overflow: auto` or `overflow-y:auto` to `.user-head` the content from the dropdown menu could be viewed after scrolling down a bit to uncover the second dropdown item "Email Addresses".

Here is a screenshot of the issue, reproduced for viewports > 1024px.

Screen Shot 2020-11-16 at 12 41 39 AM

The dropdown menu works fine for viewports smaller than 1024px. After the media query happens at 1024px and the layout shifts, both of the drop down buttons are visible but the second item is a bit cut off at the bottom. 

Potential Fixes:
1. Remove `overflow: hidden` from `.user-head` class to allow the `.submenu-column` container to have a scrolling mechanism so both the `<li>` items can be viewed/accessed.
2. OR Allow the `.user-head` container to grow and accommodate the space created by `.submenu-column`. That way each list item from the drop down will be visible when hovering the "settings" button in `.user-cards`. But then there will be some unwanted space under the profile picture since `.user-head` would indeed be increasing in height to make room for both the list items in the dropdown. 

Note: I couldn't find a template issue for MDN Profiles at `/en-US/Profiles/` so I created a blank issue. Thanks!

schalkneethling:
Ensure submenu is fully visible on view profile page

### Before

Screenshot 2020-12-07 at 14 03 36

### After

Screenshot 2020-12-07 at 14 04 30

fix #7620

### DIFF
--- a/kuma/static/styles/components/users/_user-head.scss
+++ b/kuma/static/styles/components/users/_user-head.scss
@@ -1,7 +1,6 @@
 .user-head {
     background: $light-background-color;
     margin-bottom: $grid-spacing;
-    overflow: hidden;
     padding: $grid-spacing;
     position: relative;
 }


escattone (approved):
Just in case Chris doesn't have time to get to this!

tannerdolby:
@schalkneethling It looks much better. Nice work!

PR #7662, opened 2020-12-07
fix: header search ui change

### LINKED ISSUE #7648: The search box in the Kuma header (e.g., on the home page) looks very different from a Yari document page
In Yari, the search box on the home page looks much different from the search box on a Yari document page.

schalkneethling:
Change the UI of the search input in the header to match Yari

Screenshot 2020-12-07 at 17 08 51

fix #7648

### DIFF
--- a/kuma/static/styles/minimalist/components/_search-widget.scss
+++ b/kuma/static/styles/minimalist/components/_search-widget.scss
@@ -43,7 +43,6 @@
             transform: translateY(0);
             padding: 2px 8px;
             border: 2px solid $text-color;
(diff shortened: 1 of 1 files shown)


escattone (approved):
@schalkneethling This is great, but I'm wondering if we should either remove the `svg.search-icon` element from the Kuma search box, or add it to the Yari search box?

schalkneethling:
> @schalkneethling This is great, but I'm wondering if we should either remove the `svg.search-icon` element from the Kuma search box, or add it to the Yari search box?

Going to add it to Yari

PR #7670, opened 2020-12-10
change language cookie to match yari

escattone:
Change the language cookie name that we respect in Kuma to match what we use in Yari (`preferredlocale`), and also change the language cookie expiration period to match Yari as well (3 years). This PR allows locale-less URLs not handled by Yari's Lambda@Edge function, to be handled the same way when they hit Django. So this would cover requests like `/search`, `/profile`, `/profiles/...`, `/users/...`, `/account`, and `/events`, redirecting them by first trying the `preferredlocale` language cookie, then the `accept-language` header, and finally falling back to `en-US`.

This PR fills in the missing pieces from https://github.com/mdn/yari/issues/1889.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -310,9 +310,10 @@ MT_TO_KUMA_LOCALE_MAP = {
     "fa": "fa",
 }
 
+LANGUAGE_COOKIE_NAME = "preferredlocale"
 LANGUAGE_COOKIE_DOMAIN = DOMAIN
 # The number of seconds we are keeping the language preference cookie. (1 year)
-LANGUAGE_COOKIE_AGE = 365 * 24 * 60 * 60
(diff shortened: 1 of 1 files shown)


peterbe (approved):
Works like a charm!!

```
▶ curl -I -H 'Accept-language: sv-SE' -H 'Cookie: preferredlocale=de' http://localhost.org:8000/profiles/peterbe
HTTP/1.1 302 Found
Date: Thu, 10 Dec 2020 19:02:23 GMT
Server: WSGIServer/0.2 CPython/3.8.0
Content-Type: text/html; charset=utf-8
Location: /de/profiles/peterbe
Cache-Control: public, max-age=0, s-maxage=300
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Connection: close
```

PR #7673, opened 2020-12-10
change preferredlocale in the edit profile view

### LINKED ISSUE #7672: Changing your locale on the Edit Profile should set a preferredlocale cookie
**Summary**
When you click to "Edit profile" and change your language, it shouldn't just update your session cookie profile, but it should also set a non-HTTP cookie which that works like [we're doing it in Yari](https://github.com/mdn/yari/blob/70f3bef86d5563501627a7b598215b7d21c6eb3b/client/src/ui/molecules/language-menu/index.tsx#L46-L52).

**Audience**
Anybody using MDN and has an account, GitHub or Google. 

**Rationale**
It makes sense. You can always change your preferred language by using the language drop-down. But what if the user doesn't realize that and they click their "Edit profile" to change it. And can you imagine the disappointment if that doesn't work? :)

**Workaround**
Wait until we rewrite the View and Edit profile as Yari SPAs.

**Proposal**
A little one-liner in the `users/views.py` that sets the cookie in the same way that Yari's language-dropdown does it.

**Additional context**
https://github.com/mdn/kuma/pull/7670

peterbe:
Fixes #7672

That was easy! And it totally works. 
Screen Shot 2020-12-10 at 2 31 29 PM

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -473,7 +473,16 @@ def user_edit(request, username):
                 # If there's no Beta Testers group, ignore that logic
                 pass
 
-            return redirect(edit_user)
+            response = redirect(edit_user)
+            if edit_user.locale:
+                response.set_cookie(
+                    key=settings.LANGUAGE_COOKIE_NAME,
+                    value=edit_user.locale,
+                    max_age=settings.LANGUAGE_COOKIE_AGE,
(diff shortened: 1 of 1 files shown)


escattone (approved):
❤️ it, thanks @peterbe!

PR #7674, opened 2020-12-10
use Django default cookie domain so it matches Yari

escattone:
If we just go with Django's default `LANGUAGE_COOKIE_DOMAIN` (which is `None`) we match the behavior of Yari (i.e., the current domain is specified as the cookie's domain, without any leading `.`).


escattone:
I also realized that we need to match Yari's approach to making the language cookie secure (or not if doing local development).

PR #7689, opened 2020-12-17
use the search results' locale, not the current page

### LINKED ISSUE #7688: Search results' URLs provide incorrect information about localized page being unavailable / absence of redirect
Steps to reproduce: 
1. Go to https://developer.mozilla.org/fr/search?q=startsWith (or start a search with "startsWith" from https://developer.mozilla.org/fr/ ) 
2. Click the first link (I got "Getting started with the Web" / https://developer.mozilla.org/fr/docs/Learn/Getting_started_with_the_web

Actual result: 

*  Read the page stating "Page not found ... The page you requested doesn't exist in French but it exists in English" 

Expected result: 

* If we go to https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web and use the locale selector, we can access https://developer.mozilla.org/fr/docs/Apprendre/Commencer_avec_le_web
* Thus, https://developer.mozilla.org/fr/docs/Apprendre/Commencer_avec_le_web should be displayed as available or https://developer.mozilla.org/fr/docs/Learn/Getting_started_with_the_web should redirect

Can also be reproduced with the second link (https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith  which should redirect to https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/startsWith ) 

Screenshots: 

mdn_links_01
mdn_links_02
mdn_links_03

peterbe:
Fixes #7688

To understand this fix, read: https://github.com/mdn/kuma/issues/7688#issuecomment-747663872

### DIFF
--- a/kuma/javascript/src/search-results-page.jsx
+++ b/kuma/javascript/src/search-results-page.jsx
@@ -86,15 +86,9 @@ function ResultsMeta({
     );
 }
 
-function Results({
-    locale,
-    results,
-}: {
-    locale: string,
-    results: SearchResults,
-}) {
+function Results({ results }: { results: SearchResults }) {
     return (results.documents || []).map((result) => {
-        const path = `/${locale}/docs/${result.slug}`;
+        const path = `/${result.locale}/docs/${result.slug}`;
         return (
             <div className="result-container" key={result.slug}>
                 <div className="result">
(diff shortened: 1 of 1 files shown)


peterbe:
Unfortunately, my local kuma is pretty busted. I can't write to the Elasticseach inside docker any more. Because I've run out of disk space. That happened when I ran that `scrape_document` one too many times and now Docker for mac is telling the elasticsearch container inside `docker-compose` that there's not much disk space left, so Elasticsearch goes into read-only mode. Yuck! But if you guys have some non-en-US content inside you localhost.org:8000. then please test out the PR. Make a search with `/ANYTHING_OTHER_THAN_EN_US/search?q=AnythING` and you'll find that it gives you links to `Document` combos that don't actually exist nor is indexed in Elasticsearch.

escattone (approved):

PR #7699, opened 2020-12-29
auth-modal-content: fix typo "GitHub"

simon04:


peterbe (changes requested):
Thank you for this!! 
But it's doubly bad because now your improvement broke a unit test. 
```python
assert b"Sign in with Github" in response.content
```
in `kuma/users/tests/test_templates.py`
Can you update that too please?

simon04:
Done. Sorry for not checking the unit tests (and GitHub checks).

peterbe:
> Done. Sorry for not checking the unit tests (and GitHub checks).

It's not your fault that it takes so bloody for the tests to run :)

By the way, please refrain from using rebase and squash. We do that when we merge.

peterbe (approved):

PR #7705, opened 2021-01-04
Add support for Cross-Origin search requests

### LINKED ISSUE #7703: Add support for Cross-Origin search requests
## Description

I would like to be able to query the MDN Docs as documented [here](https://developer.mozilla.org/en-US/docs/MDN/Tools/Search#Search_query_format)

## Reason

I am currently developing a front-end component which would be able to search the MDN docs, and as it's hosted on a different domain from the MDN docs, a CORS error ensues

## Steps to re-produce:

- Open Dev Console on a page that isn't under `developer.mozilla.org`
- Run `fetch("https://developer.mozilla.org/api/v1/search/en-US?q=History%20Api")`

#### What should happen

A successful request should go through and return JSON data

#### What happens

An error in the request with the classic CORS error message

```
Access to fetch at 'https://developer.mozilla.org/api/v1/search/en-US?q=History%20Api' from origin 'https://matsu.fi' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

peterbe:
Fixes #7703

Screen Shot 2021-01-04 at 2 30 30 PM

### DIFF
--- a/kuma/api/v1/decorators.py
+++ b/kuma/api/v1/decorators.py
@@ -1,0 +1,14 @@
+from functools import wraps
+
+
+def allow_CORS_GET(func):
+    """Decorator to allow CORS for GET requests"""
+
+    @wraps(func)
+    def inner(request, *args, **kwargs):
+        response = func(request, *args, **kwargs)
+        if "GET" == request.method:
+            response["Access-Control-Allow-Origin"] = "*"
+        return response
+
+    return inner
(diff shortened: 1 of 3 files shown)


peterbe:
I looked into installing `django-cors-headers` which was what many people recommended on Stackoverflow and on the Django rest_framework documentation. 
But, I can only think of a single place where I would want to enable this. 
With `django-cors-headers` you can set the [`CORS_URLS_REGEX`](https://github.com/adamchainz/django-cors-headers#cors_urls_regex) but that means we have to specifically pick only the one endpoint we want it. For the record, I don't yet feel comfortable enabling this for any other v1 API views without more careful thought.

escattone (approved):
Thanks @peterbe!

PR #7706, opened 2021-01-04
get rid of 'application_ACAO' switch

peterbe:
Every single `/api/v1/whoami` call will return the `application_ACAO` (and it's [always](https://bugzilla.mozilla.org/show_bug.cgi?id=1104260#c17) anyway).
I'm just deleting it wherever I can find it. It's useless now. 

Once this lands, as soon as migrations are run, it should make [this](https://developer.mozilla.org/admin/waffle/switch/21/change/) be 404.


escattone (approved):
Thanks @peterbe !

PR #7709, opened 2021-01-05
get rid of unused waffle switches

peterbe:
I looked through the [list of switches](https://developer.mozilla.org/admin/waffle/switch/) and then brutally removed all their uses. It broke a bunch of tests but I just deleted the tests where it was appropriate.


escattone (approved):
Thanks @peterbe!

PR #7710, opened 2021-01-05
longer caching on file attachments

peterbe:
There are lots of images that we can't easily download and make as part of mdn/content, so we still rely on a Kuma serving URLs like `/files/3710/Test_Form_2.jpg` that are served via Kuma, which returns a 302 to a S3 URL. 

Increasing the cache-control will significantly increase the probability of an end-user getting a warm cache hit on those images.


kuma/settings/common.py
```diff
@@ -1153,6 +1153,16 @@ def pipeline_one_scss(slug, **kwargs):
 _PROD_ATTACHMENT_ORIGIN = "demos-origin.mdn.mozit.cloud"
 ATTACHMENT_ORIGIN = config("ATTACHMENT_ORIGIN", default=_PROD_ATTACHMENT_ORIGIN)
 
+# Primary use case if for file attachments that are still served via Kuma.
```
escattone, line 1156:
```suggestion
# Primary use case is for file attachments that are still served via Kuma.
```

escattone (approved):

PR #7716, opened 2021-01-07
remove waffle flags kumaediting, page_move, section_edit

peterbe:
I feel lighter already. At least a little bit.


kuma/api/v1/tests/test_views.py
```diff
@@ -37,7 +37,7 @@ def test_whoami_disallowed_methods(client, http_method):
 def test_whoami_anonymous(client, settings):
     """Test response for anonymous users."""
     # Create some fake waffle objects
-    Flag.objects.create(name="section_edit", authenticated=True)
+    Flag.objects.create(name="vip_only", authenticated=True)
```
escattone, line 40: Nice attention to detail in maintaining the check for the `authenticated=True` case!

kuma/api/v1/tests/test_views.py
```diff
@@ -79,7 +79,7 @@ def test_whoami(
     Flag.objects.all().delete()
 
     # Create some fake waffle objects
-    Flag.objects.create(name="section_edit", authenticated=True)
+    Flag.objects.create(name="vip_only", authenticated=True)
```
escattone, line 82: ...and here as well.

escattone (approved):
Thanks @peterbe!

PR #7718, opened 2021-01-08
remove all dashboard endpoints

escattone:
Removes **all** of the dashboard endpoints. As part of that, it removes the `Contributions` link in the logged-in user's dropdown menu, fixing what is currently broken.

This is a large but logically clean PR. I first attempted to break this up into smaller chunks, but found it more difficult to review, so this is actually easier than it seems.


peterbe:
I have no idea how to review this :)

peterbe (approved):
Well, view and edit profile still works. 
I'm not sure how or what else to review. There's basically no addition of written code. 
The tests pass and generally we know we don't need to be very careful because 90% of the code isn't running because requests can't come in to it.

Issue #7734, opened 2021-01-14
Calls to Redis is consuming 53% of our busiest part (api/v1/whoami)


peterbe:
**Summary**
<img width="1250" alt="Screen Shot 2021-01-14 at 1 04 06 PM" src="https://user-images.githubusercontent.com/26739/104633900-db8ff200-566d-11eb-8c57-e7e039a54432.png">

**Rationale**
The reason is happens is because very single pageview of MDN triggers a `GET /api/v1/whoami` and that includes this code:
```python
{f.name: True for f in Flag.get_all() if f.is_active(request)}
```
If you have 10 flags, this results in ~15 Redis `GET` requests. It depends because some calls to `flag.is_active()` can exit early. A lot of the time, it will trigger a Redis `GET` to list the specific users a flag is for. 

**Audience**
Our infra :)
And users who have to wait unnecessarily long to see if they're signed in. 

**Proposal**
We can do two things:

1. Optimize how we get a list of flags that avoid unnecessary network IO (without replacing or patching `django-waffle`)
2. Have less waffle Flags in our database

**Additional context**
The assumption is that if we keep calling `/api/v1/whoami` on everyone's pageview, and if we can reduce the overheads it'll make less stress on the system and we'll need less resources which means we can save electricity and reduce our resource needs which leads to saving money.

peterbe:
This is what we have today.
<img width="838" alt="Screen Shot 2021-01-14 at 2 04 24 PM" src="https://user-images.githubusercontent.com/26739/104636484-6aead480-5671-11eb-9c99-50f79a47d731.png">

I know we can do a more optimized loop over the flags to avoid even bothering to call those `cache.get(...)` for a bunch of our flags.

peterbe:
Now I remember, we talked about this (at length) about a year ago. We concluded two things:

1. Let's use `waffle.Switch` instead of `waffle.Flag` if at all possible. One is more powerful than the other but if you're just going to have it plainly on or off, then Switch is better. 
2. Let's not use a waffle thing at all. It's easy to just have it hardcoded in the Python code. If you want to be able to toggle something faster than submitting a PR, you can make it depend on `django.conf.settings` so it can be toggled without needing a code change.

peterbe:
After all of the above mentioned PRs are merged, let's go back to https://developer.mozilla.org/admin/waffle/flag/ and scrutinize whether we need 

1. `contrib_beta`
2. `recurring_payment_beta`
3. `sg_task_completion`

[...]

PR #7737, opened 2021-01-14
remove use of spam_* waffle flags

peterbe:
Part of #7734


etc/sample_db.json
```diff
@@ -321,21 +321,6 @@
         "name": "spam_submisions_enabled",
```
escattone, line 321: Oops, missed this one.

escattone (changes requested):
@peterbe Missing line deletions in `etc/sample_db.json`.

escattone (approved):

PR #7753, opened 2021-01-27
reinstate search health checks

peterbe:
Now http://localhost.org:8000/_kuma_status.json includes search. 
I tested it by doing things like turning off the elasticsearch server, or messing with the index name. 

This "undoes" some of the stuff that got dropped in https://github.com/mdn/kuma/pull/7747/files#diff-6ab9a23c58ba8e2eb4faf8a0031834d663a26064c6f03284f9b7e5145bed941a

But the old unit tests used to not just test the happy path, but also test things like `ConnectionError` and `NotFoundError` etc. I tried to do that too. But after 1h of trial-and-error and hair-tearing I couldn't get it to work. My mock-fu must be a bit lost. I'm not 100% sure but I suspect it's something related to the connection pooling. What I would find is that if I use `pytest .. -k test_status_missing_index` (for example), it would work and my mocking would take effect. And individually each of the 4 (new and relevant) unit tests work. But whenever trying to run all the tests one of them would always fail. Or rather, when I run the whole suite, 1 would work and the other 3 would fail. Order didn't matter. So I think the connection pooling could be related. 
Then I thought, instead of using the connection pooling, let's create a new connection each time. But after 30 min of more debugging and swearing, I still couldn't get it to work. 
At this point, I had spent 10min fully implementing and manually testing a really nice solution. The other 60min was spent endlessly trial-and-error deep debugging mocking and it was just getting me frustrated. 
Given the full context, that the core functionality certainly has been implemented, I just can't justify more time on the mocking unit tests. I don't think it's important enough in the current context to nail those. It was pretty easy to manually test it. Once. And it's not critical that these error handlings work since it's not a functionality that our users would ever see.

### DIFF
--- a/kuma/health/tests/test_views.py
+++ b/kuma/health/tests/test_views.py
@@ -10,6 +10,15 @@ from kuma.core.tests import assert_no_cache_header
 from kuma.users.models import User
 
 
+@pytest.fixture
+def mock_search_count():
+    with mock.patch("elasticsearch_dsl.connections.Elasticsearch") as search:
+        instance = search()
+        instance.cluster.health.return_value = {"status": "pink"}
+        instance.count.return_value = {"count": 90}
+        yield instance
+
+
 @pytest.mark.parametrize("http_method", ["put", "post", "delete", "options"])
 @pytest.mark.parametrize("endpoint", ["liveness", "readiness", "status"])
 def test_disallowed_methods(client, http_method, endpoint):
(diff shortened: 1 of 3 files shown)


peterbe:
@escattone I resolved the merge conflict manually.

escattone (approved):
@peterbe I'm sorry for the frustration you went though on the tests, and you definitely made the right decision to cut your losses. It's not at all important enough. Thanks for adding this back!

PR #7755, opened 2021-01-27
search locale should be case insenstive

### LINKED ISSUE #7754: locale=en-US query parameter returns an error: "not one of the available choices"
:wave: Not sure if this is the correct repo (should be https://github.com/mdn/yari? maybe https://github.com/mdn/kuma?). 

I'm seeing this behavior in both the web and api. Searching with [`?locale=en-US`](https://developer.mozilla.org/en-US/search?q=foo&locale=en-US) yields an error, while using [`?locale=en-us`](https://developer.mozilla.org/en-US/search?q=foo&locale=en-us) succeeds. My understanding is that "en-US" is the accurate casing but this should be case insensitive?

❌ Search Results for  foo  | MDN 2021-01-26 18-48-12

✅ Search Results for  foo  | MDN 2021-01-26 18-49-00

peterbe:
Fixes #7754

Now, these work:
* http://localhost.org:8000/en-US/search?q=foo&locale=en-US
* http://localhost.org:8000/en-US/search?q=foo&locale=en-US&locale=Fr

(they didn't before)

and http://localhost.org:8000/en-US/search?q=foo still works.

### DIFF
--- a/kuma/api/v1/search/__init__.py
+++ b/kuma/api/v1/search/__init__.py
@@ -153,6 +153,8 @@ def _find(params, total_only=False, make_suggestions=False, min_suggestion_score
             # This can happen if the search happened right as the index had
             # just been deleted due to a fresh re-indexing happening in Yari.
             exceptions.NotFoundError,
+            # This can happen when the index simply isn't ready yet.
+            exceptions.TransportError,
         ),
         # The default in redo is 60 seconds. Let's tone that down.
         "sleeptime": settings.ES_RETRY_SLEEPTIME,
--- a/kuma/api/v1/search/__init__.py
+++ b/kuma/api/v1/search/__init__.py
@@ -155,8 +155,9 @@ def _find(params, total_only=False, make_suggestions=False, min_suggestion_score
             exceptions.NotFoundError,
         ),
         # The default in redo is 60 seconds. Let's tone that down.
-        "sleeptime": 1,
-        "attempts": 5,
+        "sleeptime": settings.ES_RETRY_SLEEPTIME,
+        "attempts": settings.ES_RETRY_ATTEMPTS,
+        "jitter": settings.ES_RETRY_JITTER,
     }
     with retrying(search_query.execute, **retry_options) as retrying_function:
         response = retrying_function()
--- a/kuma/api/v1/search/forms.py
+++ b/kuma/api/v1/search/forms.py
@@ -34,6 +34,18 @@ class SearchForm(forms.Form):
         # See https://www.peterbe.com/plog/initial-values-bound-django-form-rendered
         data = MultiValueDict({**{k: [v] for k, v in initial.items()}, **data})
 
+        # Because the `?locale=en-US&locale=Fr` might come in from the `request.GET`
+        # we can't edit it there. So instead, we mutate it here in the `data`
+        if "locale" in data:
+            # Always force it to lowercase, because that's what the ChoiceField
+            # is configured to. And the searches should always be in lower case.
+            # Remember, Django forms will allow this to be a single string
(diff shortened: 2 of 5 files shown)


kuma/api/v1/search/forms.py
```diff
@@ -34,6 +34,18 @@ def __init__(self, data, **kwargs):
...
+        if "locale" in data:
+            # Always force it to lowercase, because that's what the ChoiceField
+            # is configured to. And the searches should always be in lower case.
+            # Remember, Django forms will allow this to be a single string
+            # (e.g. `?locale=Fr`) or a multi-value (`?locale=fr&locale=En-US`).
+            if isinstance(data["locale"], str):
+                data["locale"] = data["locale"].lower()
+            else:
+                data["locale"] = [x.lower() for x in data["locale"]]
+
```
peterbe, line 48: This is the magic sauce that the PR was really about solving.

kuma/api/v1/search/__init__.py
```diff
@@ -153,10 +153,13 @@ def _find(params, total_only=False, make_suggestions=False, min_suggestion_score
             # This can happen if the search happened right as the index had
             # just been deleted due to a fresh re-indexing happening in Yari.
             exceptions.NotFoundError,
+            # This can happen when the index simply isn't ready yet.
+            exceptions.TransportError,
```
peterbe, line 157: I've seen this happen quite a lot when I do my manual testing. It's not unrealistic that you get:
```
elasticsearch.exceptions.TransportError: TransportError(503, 'search_phase_execution_exception')
```
that happens when index isn't ready. There's some window of time between when the *server* is ready to accept connections and the *index* being ready. 
I've seen it many times that you can just wait a little, and refresh and the problem goes away. So I added it to the list of retry exceptions. 

Yes, this is unrelated to the (original) issue at hand.

kuma/api/v1/search/__init__.py
```diff
@@ -153,10 +153,13 @@ def _find(params, total_only=False, make_suggestions=False, min_suggestion_score
...
         ),
         # The default in redo is 60 seconds. Let's tone that down.
-        "sleeptime": 1,
-        "attempts": 5,
+        "sleeptime": settings.ES_RETRY_SLEEPTIME,
+        "attempts": settings.ES_RETRY_ATTEMPTS,
+        "jitter": settings.ES_RETRY_JITTER,
```
peterbe, line 162: I should have done this the first time around. 

Yes, this is unrelated to the (original) issue at hand.

kuma/api/v1/tests/test_search.py
```diff
@@ -57,45 +57,68 @@ def search(self, *args, **kwargs):
         return result
 
 
-def test_search_basic_match(user_client, settings):
+@pytest.fixture
+def mock_elasticsearch():
     fake_elasticsearch = FindEverythingFakeElasticsearch()
+    with patch("elasticsearch_dsl.search.get_connection") as get_connection:
+        get_connection.return_value = fake_elasticsearch
+        yield fake_elasticsearch
```
peterbe, line 65: So I noticed, the way I was doing it before was busted. 
When I added the now 2nd unit test with mocks, the second test would get the results from the fixture mocking in the first test. 
There are two unit tests (now): `test_search_basic_match` and `test_search_case_insensitive_locale` and because of connection pooling, what happens is that in the second test the mocking is already done for you and it reuses the first mocking. 
So I discovered a new, better way, to mock the connection. This feels much better.

escattone (approved):
Looks and works great. I like your improvements as well! Thanks @peterbe!

southgate:
Thanks y'all 🙌🏼

PR #7759, opened 2021-01-27
Make /$locale/search.json redirect

### LINKED ISSUE #7758: Make /$locale/search.json redirect
**Summary**
We get a lot of `NoReverseMatch` on URLs like https://developer.mozilla.org/en-US/search.json
That endpoint disappeared when we switched to the new search API. 
Let's make it redirect. 

**Additional context**
https://sentry.prod.mozaws.net/operations/mdn-prod/issues/10482841/

peterbe:
Fixes #7758

Because the testing is removed for this legacy Django app, I didn't bother with a new unit test. But to sanity check myself, I wrote a little script to check:
```python
import requests

def T(pathname, expect):
    r = requests.get(f"http://localhost.org:8000{pathname}", allow_redirects=False)
    assert r.status_code == 301, (pathname, r.status_code)
    assert (
        r.headers["location"] == expect
    ), f"WANTED: {expect}\tGOT: {r.headers['location']}"


T("/en-US/search.json?q=bar", "/api/v1/search?q=bar")
T("/en-US/search.json?q=bar&locale=fr", "/api/v1/search?q=bar&locale=fr")
T("/en-US/search.json", "/api/v1/search")
T("/en-US/search.json?locale=fr", "/api/v1/search?locale=fr")
T("/fr/search.json", "/api/v1/search?locale=fr")
```

### DIFF
--- a/kuma/search/views.py
+++ b/kuma/search/views.py
@@ -1,5 +1,6 @@
 from urllib.parse import parse_qs, urlencode
 
+from django.conf import settings
 from django.shortcuts import render
 from django.urls import reverse_lazy
 from django.views.decorators.cache import never_cache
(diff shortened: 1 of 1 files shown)


peterbe:
What's interesting, but only a little bit, is that if you look at the actual URL used in https://sentry.prod.mozaws.net/operations/mdn-prod/issues/10482841/ it was: https://developer.mozilla.org/en-US/search.json?q=canvas%20circ&topic=js
That `topic` thing is gone.

kuma/search/views.py
```diff
@@ -115,11 +116,14 @@ class SearchRedirectView(RedirectView):
...
-        url = reverse_lazy(
-            "api.v1.search", kwargs={"locale": self.request.LANGUAGE_CODE}
-        )
-        if query_string:
-            url += "?" + query_string
+        url = reverse_lazy("api.v1.search")
+        qs = parse_qs(query_string)
+        # If you used `/en-Us/search.json` you can skip the `?locale=`
+        # because the default locale in `/api/v1/search` is `en-US`.
+        if self.request.LANGUAGE_CODE.lower() != settings.LANGUAGE_CODE.lower():
```
escattone, line 123: This also ensures that the locale in the URL (the request) takes precedent if a locale is also specified in the query string and they're different, which seems fine to me.

escattone (approved):
Looks and works great. Users will just have to adjust to the new JSON format.

PR #7765, opened 2021-01-28
Let Dependabot upgrade poetry packages

### LINKED ISSUE #7745: Let Dependabot upgrade poetry packages
We're currently doing Dependabot the old fashioned way. 
We should do what we're doing [in yari](https://github.com/mdn/yari/blob/master/.github/dependabot.yml) and drive it from a config file rather than from Dependabot's now deprecated app.

peterbe:
Fixes #7745

I just find `poetry` updating bewildering. I struggle to update just one package without it upgrading a bunch of other things. 
Considering that the worst that can happen is that Dependabot makes a PR that fails, let's give this a good start.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "pip"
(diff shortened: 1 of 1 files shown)


peterbe:
Note-to-self; Documentation is here: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

escattone (approved):

PR #7778, opened 2021-02-04
slug_prefixes

peterbe:
This is the precursor to be able to filter the site-search to specific tree(s). E.g. searching for the word `float` in the CSS reference doc. Or the word `fetch` in the Learning area. 

There's nothing using this yet. We won't add it to Kuma's UI and it's not the right time to add it to the not-yet-launch Yari site-search. But now it becomes possible. 

One interesting potential with this is that this could be a solution to those monster `Index` pages [E.g.](https://developer.mozilla.org/en-US/docs/Web/API/Index) For example, that macro could be made to only show the top 10 sub-pages but with a link at the bottom that is a search. Granted, our search currently always requires a search free-text input, but that's something we could build on. 

My hope is to add to the site-search UI some advanced field that appears and the user can just click things like "CSS" or "JavaScript" and behind the scenes we hardcode those to mean things like `&slug_prefix=Web/CSS`

### DIFF
--- a/kuma/api/v1/search/__init__.py
+++ b/kuma/api/v1/search/__init__.py
@@ -50,7 +50,8 @@ def search(request, locale=None):
         "size": form.cleaned_data["size"],
         "page": form.cleaned_data["page"],
         "sort": form.cleaned_data["sort"],
-        "slug_prefixes": form.cleaned_data["slug_prefix"],
+        # The `slug` is always stored, as a Keyword index, in lowercase.
+        "slug_prefixes": [x.lower() for x in form.cleaned_data["slug_prefix"]],
     }
     results = _find(
         params,
--- a/kuma/api/v1/search/__init__.py
+++ b/kuma/api/v1/search/__init__.py
@@ -50,6 +50,7 @@ def search(request, locale=None):
         "size": form.cleaned_data["size"],
         "page": form.cleaned_data["page"],
         "sort": form.cleaned_data["sort"],
+        "slug_prefixes": form.cleaned_data["slug_prefix"],
     }
     results = _find(
         params,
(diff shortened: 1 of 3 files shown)


peterbe:
To test, go to http://localhost.org:8000/api/v1/search?q=float&locale=en-US&slug_prefix=Web/HTML 
Or, you can use multiple like http://localhost.org:8000/api/v1/search?q=foreach&locale=en-US&slug_prefix=Web/HTML&slug_prefix=Web/CSS and observe that none of the popular JavaScript documents are found.

escattone (changes requested):
The `slug_prefix` seems to be case sensitive. For example, http://localhost.org:8000/api/v1/search?q=float&locale=en-US&slug_prefix=web/css gives me nothing:
image

but http://localhost.org:8000/api/v1/search?q=float&locale=en-US&slug_prefix=Web/CSS works fine:
image

None of the other search query parameters seem to be case sensitive, so it seems to me `slug_prefix` shouldn't be either.

Otherwise, this looks good!

peterbe:
It's deliberate that it doesn't return anything. It's because the `slug` is a `Keyword` index. So it's case sensitive. 
You're not supposed to type in a `slug_prefix`. 
I had envisioned that the Yari site-search UI might someday have something like this:
```html
<b>Filter by tree:</b>
...
  <li><a href="?slug_prefix=Web/Progressive_web_apps">Progress web apps (PWAs)</a></li>
  <li><a href="?slug_prefix=Web/Guide/Graphics">Graphics</a></li>
  <li><a href="?slug_prefix=Web/CSS">CSS</a></li>
...
```
Perhaps a bit of an extreme example, but the idea is that you don't type these things in free-form. It's selected for you but we overlay it with a more "human" label (if applicable). 

The other thing is that the `slug` is indexed in ES but we're not using it anywhere for anything. 
The current (hopefully only another week or so) Kuma search UI actually gets the `locale` and `slug` from the `mdn_url`. And in the new Yari site-search (currently only PR) we don't use the slug for anything. 

Having said all of that, **I eventually agree**. Since we're never using the `slug` for anything. I.e. we're not including its stored value in the search results (backend or front-end). It makes no sense to be so strict. 

**In conclusion**, I'll make it all lowercase but that's going to require a little change to the deployer so that it uploads it into Elasticsearch in lowercase too.

peterbe:
Now, for this to work we'll need to depend on https://github.com/mdn/yari/pull/2842 first. 
Thankfully, the timing of this is pretty harmless since that `slug` field isn't yet used for anything in production.

peterbe:
Now http://localhost.org:8000/api/v1/search?q=floating&locale=en-US&slug_prefix=Web/HTmL&slug_prefix=web/CSs works. 
I tested it locally.

escattone (approved):

PR #7783, opened 2021-02-08
remove elasticsearch from docker-compose

peterbe:
Hopefully self-explanatory from the PR title and the git diff :)


peterbe:
> @peterbe I'm thinking we should also remove the default value of `elasticsearch:9200` for `ES_URLS`?
> 
> https://github.com/mdn/kuma/blob/4d29ea4d67bebe2df9f5de54025e39485a8cee6c/docker-compose.yml#L40

You're right! I tested it. It still respects my URL that I have in my `.env` file.

escattone (approved):

PR #7790, opened 2021-02-09
return total for suggestions correctly

### LINKED ISSUE #7789: The "Did you mean..." suggestions' total is wrong
The "Did you mean..." says there are supposed to be 103 matches on this suggestion:
Screen Shot 2021-02-09 at 5 21 50 PM

But if you apply it, you get a different total:
Screen Shot 2021-02-09 at 5 22 00 PM

Not sure if it's an error in Kuma or in how the results are presented in Yari.

peterbe:
Fixes #7789

Look at http://localhost.org:8000/api/v1/search?q=finding&locale=en-US before and after this branch. Collapse the `documents` key and look at the `suggestions` key.


escattone (approved):
Thanks @peterbe, looks and works great!

PR #7794, opened 2021-02-18
support sign-up from Yari

peterbe:
Once this lands, we can start testing the `/en-US/sign-up` SPA in Yari. Even from prod. 
However, one thing that's a bit weird is that Django will still respond with a HTML page if your CSRF cookie expires or you manually muck around with it and not follow the happy-path.

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -687,6 +687,7 @@ class SignupView(BaseSignupView):
         # Once Kuma id divorced from doing any
         is_yari_signup = self.request.session.get("yari_signup", False)
         if is_yari_signup:
+            form.data = form.data.copy()
             for key in form.initial:
                 if key not in form.data and form.initial[key]:
                     form.data[key] = form.initial[key]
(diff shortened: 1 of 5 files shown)


peterbe:
@escattone I don't know what I was waiting for. I don't want to kill myself over writing a bunch of unit tests. 
First of all, a lot of this stuff will have to change once we stop using Kuma's front-end to sign in/up. Basically, that's as soon as we have landed https://github.com/mdn/yari/pull/2975, changed the "Sign in" link, and done whatever's needed to the CDN (which I don't know if it's any). 
And who knows, we have so much unhappiness around `django-allauth` that we should seriously consider alternatives like OIDC. 

To test this out, you need to start this in a terminal and then run https://github.com/mdn/yari/pull/2975 in another. There are instructions within that PR about how to test it.

kuma/users/views.py
```diff
@@ -680,6 +681,17 @@ def get_form(self, form_class=None):
                     "verified": False,
                     "primary": False,
                 }
+
+        # This magic sauce only exists whilst this Django view is being used
+        # both by Yari (in testing) and by old Kuma (production).
+        # Once Kuma id divorced from doing any
```
escattone, line 687: Delete this comment line?

peterbe: No, we're going to have to come back to this view function soon. I do worry about remembering to do so due to being human and that sickness of not wanting to sink too much time in kuma things for something as unimportant as sign-up.

PR #7801, opened 2021-02-26
Let's CDN cache most site-search queries

### LINKED ISSUE #7799: Let's CDN cache most site-search queries
I downloaded the 1,000 most frequent URLs when people do a site-search in `en-US`. 
In en-US alone we have 7,586 searches per day. And that's based only on the top 1,000 search terms. Remember, this only counts people who have GA tracking not blocked. 

And I estimate that roughly 59% of all our `/$locale/docs/` traffic is `en-US`. That means that **we get 13.5k site-searches per day**. And remember, this is just for the top 1,000 search terms. And since site-searches today is never cached by the CDN, that's how many times Kuma has to process 13.5k Elasticsearch queries per day (and equally all the Django code that wraps Elasticsearch into a request/(JSON)response). Perhaps roughly 20k site-searches per when you include all search terms and the people who don't let GA track.
Also, each request that goes into Django also triggers the rate limiter which means it has to use Redis too. 

As of today, site-search triggers a full Django+React SSR render and this accounts for 18% of the most time-consuming work our EC2 has to do:
Screen Shot 2021-02-25 at 4 41 05 PM
That number will shrink by itself once Yari takes over the search and it's just a `/api/v1/search` XHR request.

I speculate that we can leverage the CDN to:

- **Make users get faster search results**
- **Reduce the stress on our Kubernetes cluster and be able to use a lot less EC2 power and save money and electricity**

Many searches are a single word (i.e. 0 spaces)
```
0 spaces: 91.3%
1 spaces: 8.5%
2 spaces: 0.2%
```

Most searches are short single words. A complete list of the [top 1,000 most common searches are here](https://gist.github.com/peterbe/7cc2c8fa1cc72e976b9ac0caa0852055)

If you analyze the length of the search terms you get:
```
0-10       85.1%
11-20      14.6%
21+        0.3%
```

Since all traffic always goes through the CDN anyway, we can set a `Cache-Control` on these. Say, 24h. I don't think it's expected that a search for `splice` should get documents that have recently changed within the last day. 

We could potentially make the setting of the `Cache-Control` only on short and simple searches without any advanced filtering (e.g `?locale=ja&locale=en-us&slug_prefix=Web/CSS`). That way our CDN won't have to store too many things that aren't worth caching.

peterbe:
Fixes #7799

### DIFF
--- a/kuma/api/v1/search/__init__.py
+++ b/kuma/api/v1/search/__init__.py
@@ -1,5 +1,6 @@
 from django import http
 from django.conf import settings
+from django.utils.cache import patch_cache_control
 from elasticsearch import exceptions
 from elasticsearch_dsl import Q, query, Search
 from redo import retrying
@@ -7,6 +8,12 @@ from redo import retrying
 from kuma.api.v1.decorators import allow_CORS_GET
 
 from .forms import SearchForm
+
+# This is the number of seconds to be put into the Cache-Control max-age header
+# if the search is successful.
+# We can increase the number as we feel more and more comfortable with how
+# the `/api/v1/search` works.
+SEARCH_CACHE_CONTROL_MAX_AGE = 60 * 60 * 12
 
 
 class JsonResponse(http.JsonResponse):
@@ -67,7 +74,25 @@ def search(request, locale=None):
         params,
         make_suggestions=make_suggestions,
     )
-    return JsonResponse(results)
+    response = JsonResponse(results)
+
+    # The reasoning for caching is that most of the time, the seaches people make
+    # are short and often stand a high chance of being reused by other users
+    # in the CDN.
+    # The worst that can happen is that we fill up the CDN with cached responses
+    # that end up being stored there and never reused by another user.
+    # We could consider only bothering with this based on looking at the parameters.
+    # For example, if someone made a search with "complex parameters" we could skip
+    # cache-control because it'll just be a waste to store it (CDN and client).
+    # The reasoning for not using a "shared" cache-control, i.e. `s-max-age` is
+    # because the cache-control seconds we intend to set are appropriate for both
+    # the client and the CDN. If the value set is 3600 seconds, that means that
+    # clients might potentially re-use their own browser cache if they trigger
+    # a repeated search. And it's an appropriate number for the CDN too.
+    # For more info about how our search patterns behave,
+    # see https://github.com/mdn/kuma/issues/7799
+    patch_cache_control(response, public=True, max_age=SEARCH_CACHE_CONTROL_MAX_AGE)
+    return response
 
 
 def _find(params, total_only=False, make_suggestions=False, min_suggestion_score=0.8):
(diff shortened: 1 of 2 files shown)


peterbe:
@escattone I think this is going to make a big difference. The new site-search is "inferior" to the old site-search in that the new one requires 2 phases. 1) download and execute the lazy-loaded JS which starts the XHR, and 2) XHR to Kuma and render its output. 
But hey, that's what you've got to do in the Jamstack world. It makes the site-search a bit more "flickering". So if whatever we can do to make that better will help. 

Once this lands, something I'd be very much interested in doing is to temporarily inject some GA event sending on the site-search app that reports back to use how it's working out for users. For example, we can measure how many people get a CDN cache hit/miss on the XHR, if they change the sort order, if they change the language (for non-en-US), if they go to page 2, etc. 
But to do that, I should make myself a Google Doc, have that signed off by you and the team, and then have a plan for analysis and rolling back the experiment instrumentation. One thing at a time.

kuma/api/v1/search/__init__.py
```diff
@@ -67,7 +74,25 @@ def search(request, locale=None):
...
+    # For example, if someone made a search with "complex parameters" we could skip
+    # cache-control because it'll just be a waste to store it (CDN and client).
+    # The reasoning for not using a "shared" cache-control, i.e. `s-max-age` is
+    # because the cache-control seconds we intend to set are appropriate for both
+    # the client and the CDN. If the value set is 3600 seconds, that means that
+    # clients might potentially re-use their own browser cache if they trigger
+    # a repeated search. And it's an appropriate number for the CDN too.
+    # For more info about how our search patterns behave,
+    # see https://github.com/mdn/kuma/issues/7799
+    patch_cache_control(response, public=True, max_age=SEARCH_CACHE_CONTROL_MAX_AGE)
```
peterbe, line 94: This might cause confusion when you're doing local development. E.g. localhost.org:8000/api/v1/search 
What we can do is something like:
```suggestion
    patch_cache_control(
        response, public=True, 
        max_age=0 if settings.DEBUG else SEARCH_CACHE_CONTROL_MAX_AGE
    )
```

peterbe:
> @peterbe Just some comment nits. Let's try this! I'll need to adjust the caching policy on the CDN behavior that handles all of the `api` requests. In fact, what I'll probably do, is create a new behavior just for `/api/v1/search` and ensure that it's caching policy accounts for the query parameters when creating the cache key. That's alll easy, but just wanted to let you know that merging and deploying this will have no effect until I do that AWS work.

You do what you have to do :)
We'll have to wait for this to hit Stage first but once there and you've changed the CloudFront for Yari Stage, then let me know and I'll test it. 

Other things we can do...

1. Perhaps create a temporary experiment with the GA event sending to phone home what the headers are for `X-Cache` on that `fetch` call that the search-results does. Then we can see, over a period of time, how many people are getting warm and cold caches on XHR site-searches. 
2. Perhaps it's smart and cheap to trigger a cache invalidation, via boto3, at the bottom of every `*-build.yml` just for something like `/api/v1/search*`. Just to reduce the worst stale caching for users and if that works, we can bump up the TTL from 12h to 24h.

PR #7814, opened 2021-03-23
ProtectedError trying to close account through /api/v1/settings

### LINKED ISSUE #7813: ProtectedError trying to close account through /api/v1/settings
https://sentry.prod.mozaws.net/operations/mdn-stage/issues/10922237/?referrer=github_plugin

peterbe:
Fixes #7813

I copied these, without much deeper thought, from the existing `kuma/users/views.py` for last time we had to solve this. 

Another solution would/could have been to instead starting to delete these old Wiki models so the referenced foreign keys stop even being a thing. But I think for now this is quicker.


escattone (approved):
@peterbe I agree that this approach is quicker and safer for now. Thanks!

PR #7815, opened 2021-03-25
correctly generate csrftoken for sign-up

peterbe:
it used to work *sometimes* before if you had a cookie already. 

Now, I tested it with a fresh new private window and used an email that would need to sign-up.


escattone (approved):
@peterbe Ah, yes this fits what I was seeing. We weren't generating a CSRF token if there wasn't already an existing one.

PR #7818, opened 2021-03-30
invalid form should yield a JSON 400 response

peterbe:
Part of https://github.com/mdn/yari/issues/3377

At least the errors don't fly by silently in the redirect now. 

Screen Shot 2021-03-29 at 10 03 06 PM

There's a PR coming for this on the Yari side too.


schalkneethling (approved):
🎉

PR #7821, opened 2021-03-30
make sign-up work for private email GH accounts

### LINKED ISSUE #7820: Sign-up doesn't work for private GitHub emails
We can use the mdnwebdocs.tester GitHub account to try to replicate the bug.

peterbe:
Fixes #7820

### DIFF
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -689,7 +689,7 @@ class SignupView(BaseSignupView):
         is_yari_signup = self.request.session.get("yari_signup", False)
         if is_yari_signup:
 
-            if not email:
+            if not email and extra_email_addresses:
                 # Pick the first primary email.
                 for data in extra_email_addresses:
                     if data["primary"]:
@@ -697,8 +697,7 @@ class SignupView(BaseSignupView):
                         break
                 else:
                     # Pick the first non-primary email.
-                    for data in extra_email_addresses:
-                        email = data["email"]
+                    email = extra_email_addresses[0]["email"]
 
             form.initial["email"] = email
 
--- a/kuma/users/views.py
+++ b/kuma/users/views.py
@@ -690,10 +690,16 @@ class SignupView(BaseSignupView):
         if is_yari_signup:
 
             if not email:
+                # Pick the first primary email.
                 for data in extra_email_addresses:
-                    if data["primary"] or not email:
+                    if data["primary"]:
                         email = data["email"]
                         break
+                else:
+                    # Pick the first non-primary email.
+                    for data in extra_email_addresses:
+                        email = data["email"]
+
             form.initial["email"] = email
 
             form.data = form.data.copy()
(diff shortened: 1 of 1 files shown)


kuma/users/views.py
```diff
@@ -685,9 +685,17 @@ def get_form(self, form_class=None):
...
-        # Once Kuma id divorced from doing any
+        # Once Kuma is divorced from doing any HTML rendering, we can refactor this.
         is_yari_signup = self.request.session.get("yari_signup", False)
         if is_yari_signup:
+
+            if not email:
+                for data in extra_email_addresses:
+                    if data["primary"] or not email:
+                        email = data["email"]
+                        break
```
escattone, line 696: This confuses me. As written, `if data["primary"] or not email` is effectively the same as `if True` since `not email` will always be `True`. Did you want to prioritize the primary email, otherwise take the first one? If so, I suspect this should be:
```suggestion
                    if data["primary"] or not email:
                        email = data["email"]
                        if data["primary"]:
                            break
```

escattone (changes requested):

kuma/users/views.py
peterbe: What about just 
```suggestion
                    email = data["email"]
                    if data["primary"]:
                        break
```

escattone: I like it! 👍

escattone: It'll always take the last email though if there isn't a primary, but I suppose we don't care?

peterbe: No, we do care. Let's try this again.
The spec has to be this, right?

1. Pick the *first* `primary` email
2. If none of them are `primary`, pick the first one

peterbe: It's awfully verbose now but the business logic is correct.

kuma/users/views.py
```diff
@@ -685,9 +685,23 @@ def get_form(self, form_class=None):
...
+            if not email:
+                # Pick the first primary email.
+                for data in extra_email_addresses:
+                    if data["primary"]:
+                        email = data["email"]
+                        break
+                else:
+                    # Pick the first non-primary email.
+                    for data in extra_email_addresses:
+                        email = data["email"]
```
escattone, line 701: This will select the last non-primary email. It'd be simpler to do something like:
```suggestion
                    email = extra_email_addresses[0]["email"] if extra_email_addresses else None
```

escattone, line 701: Or we could do something like this instead:
```suggestion
            if not email and extra_email_addresses:
                # Pick the first primary email.
                for data in extra_email_addresses:
                    if data["primary"]:
                        email = data["email"]
                        break
                else:
                    # Pick the first non-primary email.
                    email = extra_email_addresses[0]["email"]
```

escattone (approved):
Thanks @peterbe!

PR #7826, opened 2021-04-11
Delete python-lints.yml

ScottB111:


escattone:
@ScottB111 What is your reason for this change? We want to keep flake8 and black linting.

PR #7836, opened 2021-05-03
fix for docker-compose on m1 macs

escattone:
I just recently upgraded to a MacBook pro with an M1 chip, and needed to do this.


peterbe (approved):
I've seen this around. Can you add a comment to the YAML line above to explain *why* that's there?

PR #7838, opened 2021-05-04
only run documentation-build if documents change

peterbe:
- Use `on.pull_request.paths` instead of using get-diff-action because we don't care about which files changed once the workflow starts. 
- No point messing around with caching the python pip stuff.


peterbe:
@nschonni r?

peterbe:
I started wanting to [upgrade get-diff-action to v4](https://github.com/mdn/kuma/pull/7833/files) but realized it doesn't make sense if you can use `on.pull_request.paths` to do all the hard work.

nschonni (approved):

escattone (approved):
Nice! Thanks @peterbe!

PR #7849, opened 2021-05-07
Add cloudfront geo country to whoami

### LINKED ISSUE #7848: Add country geo information to /api/v1/whoami
We can use this immediately to be able to determine if a user should be seeing certain banners in Yari.

peterbe:
Fixes #7848

Once this lands, we can immediately start to do things like this in Yari:

```javascript
const userData = useUserData();
if (userData.geo && userData.geo.country) {
  console.log(`This user is view from ${userData.geo.country}`);
}
```


escattone (approved):
Thanks @peterbe!

PR #7850, opened 2021-05-07
use get_token for csrf instead of header

peterbe:
Remember this? https://github.com/mdn/kuma/pull/7815
You can't rely on `request.META['CSRF_COOKIE']` on the initial load but technically this view function clearly checks, on line 129, that you can't even get that far if you're not signed in. 

But it just doesn't feel right relying on returning something that might be null. `get_token` feels like the more appropriate API to use.


escattone (approved):
@peterbe Good call, thanks!

Issue #7851, opened 2021-05-07
Get rid of the feeder app


peterbe:
We don't use it for anything. It's just in the way. Yari as its own, much simpler, way to consume the Hacks blog for the MDN home page.

PR #7852, opened 2021-05-07
Get rid of the feeder app

peterbe:
Part of #7851

Once this lands and we're certain everyone (prod in particular) has run the migration, we can remove the app as well and close the issue. I just don't know any other way to remove something from Kuma where you first want the migration to kick in so it can delete from MySQL.


escattone (approved):
Thanks @peterbe! Once you get it green, let's merge!

peterbe:
@escattone So to generate the relevant migration 
1. Edited the `feeder/models.py` to be an entirely *empty* file
2. Ran `docker-compose exec web python manage.py makemigrations feeder`
3. Checked in the `kuma/feeder/migrations/0003_auto_20210507_1205.py` file so it has somewhere to sit. 
4. Remove almost all the code
5. Left the app.

But what if I *move* the `0003_auto_20210507_1205.py` file into `kuma/core/migrations/`? Then I can do `git rm -r kuma/feeder` and all other instances (e.g. Prod, Stage, Gregor's machine) will have that `DROP TABLE kuma_feeder ...` code next time they run `docker-compose exec web python manage.py migrate`, right? Right?

Cause if I proceed with this PR as it stands, I have to make another PR and if people don't run `./manage.py migrate` before the whole `feeder/migrations` folder disappears, they will have orphaned data in their MySQL.

PR #7866, opened 2021-05-12
updating welcome email

chrisdavidmills:
As promised, I've had a go at updating the MDN signup welcome mail, to better match today's MDN.

I have also tried to update the corresponding plain.ltxt file, although I'm not 100% sure if I've got that right.

LMK what you think. Thanks!

### DIFF
--- a/kuma/users/jinja2/users/email/welcome/html.ltxt
+++ b/kuma/users/jinja2/users/email/welcome/html.ltxt
@@ -14,16 +14,17 @@
         </p>
         <p>
           {% trans
-              words_link=add_utm('https://developer.mozilla.org/en-US/docs/MDN/Getting_started#Option_1_I_like_words', 'welcome'),
-              code_link=add_utm('https://developer.mozilla.org/en-US/docs/MDN/Getting_started#Option_2_I_like_code', 'welcome'),
-              both_link=add_utm('https://developer.mozilla.org/en-US/docs/MDN/Getting_started#Option_3_I_like_both_words_and_code', 'welcome'),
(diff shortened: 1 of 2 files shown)


peterbe:
I think it's a mistake to have 2 files. We should only be maintaining the HTML file (aka. template). 
Within the Python `send_welcome_email()` function we can generate the plain text email from the HTML one by running it through `html2text` or something. 
The chances of someone still preferring plain text over HTML in multipart emails is now extremely low. And for those few that still do, that generated plain text email would be OK in terms of content. 

I'd be happy to tackle that later.

peterbe (approved):
Here's what the email HTML becomes: https://gist.github.com/peterbe/64d7642ce4434cd288eda20ef61dd221

It's still pretty geared towards contributing to the content but it's better now than it was before. And this will tie us over until we reconsider this for the benefit of using (or not using!) the welcome email for the mdn++ stuff. 

I think we should remove those pesky `utm` tracker query string things. 
Nobody's around to care to track those. Perhaps it's just me but they look a bit "icky" and make me, as a sign-upee feel like a product of some tracking stats. 

What do you think? 
If you want to, we can merge this as is and I'll follow-up with a new PR for the deletion of the `add_utm()` stuff.

peterbe:
PS. I tested all the URLs and they all 200 OK.

escattone (approved):
Thank you @chrisdavidmills!

escattone:
> What do you think?
> If you want to, we can merge this as is and I'll follow-up with a new PR for the deletion of the `add_utm()` stuff.

@peterbe Let's do that, thank you!

PR #7869, opened 2021-05-12
refactorings for welcome html email

### LINKED ISSUE #7868: Refactorings for welcome HTML email
Based on https://github.com/mdn/kuma/pull/7866 
we'll drop the `utm` query string things and we'll get rid of the plain text email template.

peterbe:
Fixes #7868

As I was hacking on it, I printed the `content_plain` (and the `content_html`) to stdout to see what it would look like:

Screen Shot 2021-05-12 at 2 29 05 PM

The argument is here: https://github.com/mdn/kuma/pull/7866#issuecomment-839905330 about not maintaining two files when 99% of people read HTML emails and the 1% can get theirs automatically generated from the HTML.


escattone (approved):
I like it, thanks @peterbe!

PR #7876, opened 2021-05-19
landing page variants

peterbe:


kuma/api/v1/plus.py
```diff
@@ -62,3 +64,17 @@ def landing_page_survey(request):
         context["csrfmiddlewaretoken"] = get_token(request)
 
     return JsonResponse(context)
+
+
+@ratelimit(key="user_or_ip", rate="100/m", block=True)
+def landing_page_variant(request):
```
escattone, line 70: I would add the following:

```suggestion
@require_GET
def landing_page_variant(request):
```

kuma/api/v1/plus.py
```diff
@@ -62,3 +64,17 @@ def landing_page_survey(request):
...
+    variants = settings.PLUS_PRICE_VARIANTS
+    assert isinstance(variants, list)
+    session_key = "plus-landing-page-variant-v1"
+    variant = request.session.get(session_key)
```
escattone, line 74: I'm confused. If we're getting the variant from the session (via a session cookie) here, why not also get it from the session in the `landing_page_survey` view above, because on the client side won't you always do a GET from this view (to get the price and the variant) **before** you do a GET from the `landing_page_survey` view?

kuma/api/v1/urls.py
```diff
@@ -12,6 +12,11 @@
         views.send_subscriptions_feedback,
         name="api.v1.send_subscriptions_feedback",
     ),
+    path(
+        "plus/landing-page/",
```
escattone, line 16:
```suggestion
        "plus/landing-page-variant/",
```

kuma/api/v1/tests/test_plus.py
```diff
@@ -93,3 +93,22 @@ def test_ping_landing_page_survey_authenticated(user_client, wiki_user):
     assert response.status_code == 200
     (result,) = LandingPageSurvey.objects.all()
     assert result.user == wiki_user
+
+
+@pytest.mark.django_db
+def test_landing_page_variant_happy_path(client, settings):
+    settings.PLUS_PRICE_VARIANTS = ["Foo", "Bar", "Buzz"]
```
escattone, line 100: Nit. I would suggest doing the following, just so we also test that we've configured the casting/conversion of the env var properly. 
```suggestion
    settings.PLUS_PRICE_VARIANTS_JSON = '["Foo", "Bar", "Buzz"]'
```

kuma/settings/common.py
```diff
@@ -1830,3 +1830,15 @@ def get_user_url(user):
...
+    config(
+        "PLUS_PRICE_VARIANTS_JSON",
+        default=json.dumps(
+            [
+                "$X a month or $XX a year",
+                "$Y a month or $YY a year",
+            ]
+        ),
+    )
+)
```
escattone, line 1844: Nit. This works fine of course, but it's our convention to use the `cast` feature of `config` for this kind of thing:
```suggestion
PLUS_PRICE_VARIANTS = config(
    "PLUS_PRICE_VARIANTS_JSON",
    default=json.dumps(
        [
            "$X a month or $XX a year",
            "$Y a month or $YY a year",
        ]
    ),
    cast=json.loads
)
```

kuma/api/v1/urls.py
peterbe: I'm going to override that and make this change:
```diff
diff --git a/kuma/api/v1/urls.py b/kuma/api/v1/urls.py
index 8504daa8d..182652583 100644
--- a/kuma/api/v1/urls.py
+++ b/kuma/api/v1/urls.py
@@ -13,12 +13,12 @@ urlpatterns = [
         name="api.v1.send_subscriptions_feedback",
     ),
     path(
-        "plus/landing-page/",
+        "plus/landing-page/variant/",
         plus.landing_page_variant,
         name="api.v1.plus.landing_page_variant",
     ),
     path(
-        "plus/landing-page-survey/",
+        "plus/landing-page/survey/",
         plus.landing_page_survey,
         name="api.v1.plus.landing_page_survey",
     ),

```
Now they're more consistent and neater.

kuma/settings/common.py
peterbe: That's much better!

kuma/api/v1/tests/test_plus.py
peterbe: I'm actually more in favor of using `kuma/settings/pytest.py` then the test suite can never be at the mercy of individual people's overrides of that setting.

kuma/api/v1/plus.py
peterbe: Excellent point! I didn't really mean to implement it like this. It used to be that the landing page's React code would conjure up the variant. Then that number would be passed to the `<LandingPageSurvey/>` component that just sends it along. 
But now that the variant decision is in Kuma you have no business submitting the survey at all if you haven't already run through the `landing_page_variant` view function!
So we can simply stop *sending* the variant in the XHR for the survey. And inside `landing_page_survey` we can get the `variant`, not from `request.GET` but from `request.session`. I'll implement that instead.

peterbe:
Tests fixed.

kuma/api/v1/urls.py
escattone: Nice, I like that even better!

peterbe:
@escattone I had yet another mistake in my tests. Fixed that. But I also renamed the setting. The `price` part of it is actually less relevant. It's about the `variant` number. So I prefer to call it `PLUS_VARIANTS`. I know it's still a list of price variants but not sure it's important.

escattone (approved):
Thanks @peterbe!

PR #7877, opened 2021-05-19
delete 'scrape' app

peterbe:
I feel lighter already. Just trying to break up things into chunks. 
Also, you can't argue with the fact that this deletes 142 unit tests that probably none matter anymore.


peterbe:
Not much surprising, but when I compare this CI run with before, you get: `Run Python tests   1m 56s`  vs. `Run Python tests   2m 34s`
So it's a welcome addition until we can delete all of the `wiki`.

escattone (approved):
@peterbe I already did this in my branch, but that's OK.

peterbe:
> @peterbe I already did this in my branch, but that's OK.

I bet! Dang. I thought you only took care of the `wiki` one. 
Either way, we'll have a bit faster CI runs until your thing finally lands.

PR #7886, opened 2021-06-10
remove SSR settings

escattone:
Forgot to remove these two unused SSR settings in https://github.com/mdn/kuma/pull/7825.


escattone:
And also apparently forgot to remove the now non-existent `tests` directory from the `flake8` calls.

PR #7887, opened 2021-06-10
Stripe (again)

Gregoor:
The Yari partner PR is here: https://github.com/mdn/yari/pull/4007

## How to test
1. Create a stripe account
2. [Get yourself private & public API keys](https://dashboard.stripe.com/test/apikeys). Set them as `STRIPE_PUBLIC_KEY` and `STRIPE_SECRET_KEY` respectively.
3. [Create a new product in stripe's dashboard](https://dashboard.stripe.com/test/products)
  1. Create a price for that product and set it as `STRIPE_PRICE_IDS`
4. start both yari and kuma up, go to http://localhost.org:3000/en-US/settings to start the subscription flow

## TODOs
- [x] (re)integrate with stripe webhooks to keep subscriptions updated
- [x] remove newsletter module (sendinblue related code)?
- [x] add some tests, remove the old ones


kuma/api/v1/urls.py
```diff
@@ -22,6 +22,15 @@
         plus.landing_page_survey,
         name="api.v1.plus.landing_page_survey",
     ),
-    path("subscriptions/", views.subscriptions, name="api.v1.subscriptions"),
+    path(
+        "subscriptions/checkout",
```
peterbe, line 26:
```suggestion
        "subscriptions/checkout/",
```
I've thought a lot about this and come to conclude that trailing slashes are better. 
Yes, there are some endpoints above that don't use a trailing slash and that's unfortunate and something we should eventually fix, too. But all the new `landing-page` related endpoints all have trailing slashes. They were easy because there was 0 need do worry about parity about old kuma. 
If you agree and go for this, we're going to need to remember to update yari too but it's not time-critical. 

The advantages with trailing slashes are:
- If you failed to include the `/`, Django has well-tested redirects that can save you. But you can't do that when the trailing slash is *not* there. For example `https://developer.mozilla.org/api/v1/settings/` will 404, but if line 7 (above) had a trailing slash both `https://developer.mozilla.org/api/v1/settings/` and `https://developer.mozilla.org/api/v1/settings` would work. `fetch` would just follow the redirect. 
- If you decide to extend the URL, it feels more right to build on. E.g. someday we decide to add `/subscriptions/checkout/ping/`.
- Trailing slashes is how URLs are supposed to work and the only reason for not using it is vanity in the address bar or when you put a link in an email or in some marketing. The vanity argument doesn't make sense of an API. 

I can think of counter-arguments too. And therefore I'm pretty soft in my conviction. But I would definitely appreciate a discussion first and foremost.

kuma/api/v1/views.py
```diff
@@ -149,30 +138,50 @@ def send_subscriptions_feedback(request):
...
 
-@api_view(["POST", "GET", "DELETE"])
+@require_POST
 @never_cache
-def subscriptions(request):
-    if not request.user.is_authenticated or not flag_is_active(request, "subscription"):
+def subscription_checkout(request):
+    user = request.user
+    if not user.is_authenticated or not flag_is_active(request, "subscription"):
         return Response(None, status=status.HTTP_403_FORBIDDEN)
```
peterbe, line 146:
```suggestion
        return HttpResponseForbidden("'subscription' not enabled")
```

kuma/api/v1/views.py
```diff
@@ -149,30 +138,50 @@ def send_subscriptions_feedback(request):
...
-    all_subscriptions = []
-    subscription_info = retrieve_and_synchronize_subscription_info(request.user)
-    if subscription_info:
-        all_subscriptions.append(subscription_info)
+@require_POST
+@never_cache
+def subscription_customer_portal(request):
+    user = request.user
+    if not user.is_authenticated or not flag_is_active(request, "subscription"):
+        return Response(None, status=status.HTTP_403_FORBIDDEN)
```
peterbe, line 176:
```suggestion
        return HttpResponseForbidden("'subscription' not enabled")
```

kuma/api/v1/views.py
```diff
@@ -149,30 +138,50 @@ def send_subscriptions_feedback(request):
...
+
+    assert user.stripe_customer_id
+
+    session = stripe.billing_portal.Session.create(
+        customer=user.stripe_customer_id, return_url=request.headers.get("Referer")
+    )
 
-    return Response({"subscriptions": all_subscriptions})
+    return JsonResponse({"url": session.url}, status=status.HTTP_200_OK)
```
peterbe, line 184:
```suggestion
    return JsonResponse({"url": session.url})
```

kuma/users/stripe_utils.py
```diff
@@ -1,141 +1,48 @@
...
-                "last4": card.last4,
-                # Cards that are part of a "source" don't have a zip
-                "zip": card.get("address_zip", None),
-                "next_payment_at": datetime.fromtimestamp(
-                    stripe_subscription_info.current_period_end
-                ),
-            }
+def retrieve_and_synchronize_stripe_subscription(user):
+    if not user.stripe_customer_id:
+        return None
```
peterbe, line 13: I plead ignorance here but I suspect the only time this can happen is when we debug. I.e. you mess with the database and switch between test accounts etc. 
If some code calls this function on a user that doesn't have a `stripe_customer_id` it sounds like something is broken in the caller. Perhaps we can inject, before the `return None`, something like `warnings.warn(f"Tried to sync stripe subscription on a user ({user!r}) that doesn't have a stripe_customer_id")`. Or do you think we should throw an error just to feel the most explicit and safe?

kuma/api/v1/urls.py
Gregoor: I see your points and I don't have recent experiences pushing me in either direction so I'm going to go with your good arguments.

kuma/users/stripe_utils.py
Gregoor: atm the fn was written with the assumption that it could be called from places where a user has not started setting up a subscription yet and hence also has no customer attached. Explicitly that call site is when fetching the subscription for the /settings endpoint

peterbe: If you have time and are in there in those files, I would suggest you turn that `if` into `assert user.stripe_customer_id` and then conditionally call it like `user.stripe_customer_id and retrieve_and_synchronize_stripe_subscription(user)` at the call site.

Gregoor:
I have added a test for successful subscription creation, deletion was already in there. Plus there are docs for hooks now. Lmkwyt!

peterbe:
As of https://github.com/Gregoor/kuma/pull/4 I'm happy with all the django view tests and some nits. 

If you can review and merge that I can do one last final review and merge, but I've tested it pretty extensively locally and I think it's ready to go.

Gregoor:
Also gonna put a thanks here, I think your PR to my PR is a big improvement for readability and code navigatability!

kuma/users/checks.py
```diff
@@ -2,43 +2,48 @@
...
-        or settings.STRIPE_PLAN_ID
-    ):
+    if not settings.STRIPE_SECRET_KEY:
         return errors
 
     try:
-        plan = stripe.Plan.retrieve(settings.STRIPE_PLAN_ID)
-        if not plan.active:
+        prices = retrieve_stripe_prices()
+        if not prices:
```
peterbe, line 21: I like this! If you have set `STRIPE_SECRET_KEY` you *have* to have set prices too.

peterbe (approved):
I don't know much more I can test it. I did test it quite extensively when I made my "counter-PR" and just because this lands doesn't mean it goes into prod on the next release.

PR #7893, opened 2021-06-11
return waffle.flags to whoami temporarily

peterbe:
As discussed in Slack.


peterbe:
[We decided to fix the archived-content instead](https://github.com/mdn/archived-content/pull/21)

PR #7902, opened 2021-06-11
drop gpg from Dockerfile installs

peterbe:


peterbe:
I suspect we only needed this because we used to need it for manually installing Node.

peterbe:
Tests in CI are passing. I'm able to `docker-compose build` and `docker-compose up` and then use this from Yari (search, sign-in, whoami etc).

escattone (approved):

PR #7903, opened 2021-06-14
Plus bookmarks

peterbe:
This is the Kuma portion of https://github.com/mdn/yari/pull/4021

At the moment, it does not use the ORM. So for local development and testing, it just populates a global mutable. 

What's important is the proposed API and the new file organization for the files inside `kuma/ap1/v1/plus/`. 

- [x] Use ORM with migration
- [x] Django unit tests
- [x] Some sort of tooling to track all saved URLs so that if we check if something has become a redirect, we only need to do it once per URL rather than for every users' bookmark

---

**July 21** Here's how you test this

First, checkout this branch of Yari: https://github.com/mdn/yari/pull/4021 make sure you `yarn install && yarn dev`

1. Check out the branch, 
2. Add this to your kuma `.env` file:
```
# TEMPORARY
FAKE_USER_SUBSCRIBER_NUMBER=123
FAKE_USER_AVATAR_URL=https://www.peterbe.com/avatar.random.png
```
3. `docker-compose up`, 
4. run `docker-compose exec web python manage.py migrate`
5. run `docker-compose exec web python manage.py createsuperuser` (e.g. username `admin`  password `admin`)
6. Go to <http://localhost:8000/admin> and sign in 
7. Go to <http://localhost:3000/> and you should now see that you're signed in 
8. Click into a document and note the little bookmarking star to the right of the bread crumbs
9. Once you've bookmarked a couple, go to <http://localhost:3000/en-US/plus/bookmarks>


peterbe:
@escattone Just FYI, I fixed the recent merge conflicts. Should be open for review now again.

peterbe:
@escattone One feature that I think would be really nice to add is the ability to search in your bookmarks. 
It'd be more of a free-text filter than a search but we're on Postgres now so we can set up a GIN index on the document's title and/or breadcrumb which would make the filtering fast and smart. 
We should definitely wait till we know we have the basics working. I don't know if @dalexsy has a design for a filter option. At the moment, the only "feature" we have for handling your list of bookmarks is pagination. 

Also, if time allows we can add the "notes" field to that search. In fact, we already have that as part of the ORM but we're not exposing it in the UI. (Again, it's smart to get the basics working first).

And if we do add search it might be really smart to include [results highlighting](https://www.postgresql.org/docs/13/textsearch-controls.html#TEXTSEARCH-HEADLINE) and for that to be rendered in the listing.

kuma/api/v1/plus/bookmarks.py
```diff
@@ -0,0 +1,190 @@
...
+        except DocumentURL.DoesNotExist:
+            response = download_url(absolute_url)
+            metadata = response.json()["doc"]
+            # Because it's so big, don't store certain fields that are
+            # not helpful.
+            metadata.pop("body", None)
+            metadata.pop("toc", None)
+            metadata.pop("sidebarHTML", None)
+            metadata.pop("other_translations", None)
+            metadata.pop("flaws", None)
```
escattone, line 121: We're only using the `title`, `mdn_url`, and `parents` from `metadata`. I think we should only store what we're going to use. Also, how do `DocumentURL` objects get updated, for example if the document they represent is moved?

kuma/api/v1/plus/bookmarks.py
```diff
@@ -0,0 +1,190 @@
...
+                # to be the one they deleted.
+                was_undo = False
+                for most_recently_modified in Bookmark.objects.filter(
+                    user_id=request.user.id,
+                ).order_by("-modified")[:1]:
+                    was_undo = most_recently_modified.id == bookmark.id
+                if not was_undo:
+                    # When you undo, it doesn't change the 'created' date.
+                    # But if you've re-bookmarked it after some time
+                    bookmark.created = timezone.now()
```
escattone, line 150: Hm, I wonder if we could make this simpler. What do you think if we just compare `bookmark.deleted` with the current time, and if that difference is less than some short period of time (like an hour or less, or maybe just a few minutes or less?), we won't change `bookmark.created`, otherwise we will?

escattone (commented):
@peterbe I've been looking through this PR and its companion (https://github.com/mdn/yari/pull/4021), and overall it looks really well done. My comments so far are more like suggestions for organizational adjustments. More tomorrow.

kuma/documenturls/models.py
```diff
@@ -0,0 +1,90 @@
...
+@receiver(pre_save, sender=DocumentURL)
+def assert_lowercase_uri(sender, instance, **kwargs):
+    # Because it's so important that the `uri` is lowercased, this makes
+    # absolutely sure it's always so.
+    # Ideally, the client should check this at upsert, but if it's missed,
+    # this last resort will make sure it doesn't get in in
+    instance.uri = DocumentURL.normalize_uri(instance.uri)
+
+
+class DocumentURLCheck(models.Model):
```
escattone, line 60: I don't see where this model is used other than the admin interface, and I don't see that its `check_uri` class method is ever used anywhere. Perhaps this is the first part of a system to be added in the future that will keep the `DocumentURL` objects up-to-date (e.g., when documents have been moved), or am I missing something?

kuma/documenturls/models.py
```diff
@@ -0,0 +1,90 @@
...
+class DocumentURL(models.Model):
+    """There are documents we look up for things like bookmarks and notes.
+    These are not legacy Wiki documents but rather remote URLs.
+    """
+
+    # E.g. /en-us/docs/web/javascript (note that it's lowercased!)
+    # (the longest URI in all our of content as of mid-2021 is 121 characters)
+    uri = models.CharField(max_length=250, unique=True, verbose_name="URI")
+    # E.g. https://developer.allizom.org/en-US/docs/Web/JavaScript
+    absolute_url = models.URLField(verbose_name="Absolute URL")
```
escattone, line 31: The `absolute_url` field is always constructed from the incoming URI right? It seems to me that we don't need to save it. Can't we just construct it from the `uri`  (e.g. `f"{settings.BOOKMARKS_BASE_URL}{uri}/index.json"`) whenever we need it? I realize that the `uri` is always lower-case, but that shouldn't matter when using the `absolute_url` for fetching.

kuma/settings/pytest.py
```diff
@@ -35,10 +35,9 @@
     current_prefix = cache_settings.get("KEY_PREFIX", "")
     cache_settings["KEY_PREFIX"] = "test." + current_prefix
 
-# This avoids reading the static folder for each test client request, for
-# a 10x speedup on Docker on MacOS.
-WHITENOISE_AUTOREFRESH = True
-WHITENOISE_USE_FINDERS = True
+
+# Always assume we prefer https.
+PROTOCOL = "https://"
```
escattone, line 40: This is redundant, must have been added by mistake.

kuma/bookmarks/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+from kuma.documenturls.models import DocumentURL
+
+
+class Bookmark(models.Model):
+    documenturl = models.ForeignKey(
+        DocumentURL, on_delete=models.CASCADE, verbose_name="Document URL"
+    )
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    notes = models.JSONField(default=list)
```
escattone, line 12: 👍 I like that you're preparing for the future. 😄

kuma/api/v1/plus/bookmarks.py
```diff
@@ -0,0 +1,190 @@
...
+
+    return inner
+
+
+@never_cache
+@require_http_methods(["GET"])
+@is_subscriber
+def bookmarks(request):
+    try:
+        page = int(request.GET.get("page") or "1")
```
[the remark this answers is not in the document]
peterbe: I prefer the `or` because if the the URL is `?page=` what `request.GET.get('page')` is then is `''` (empty string). 
In the this context, it can't be None but imagine it was:
`request.GET['page'] = None`
you'd ultimately end up doing `int(None)`
But using `or` will ignore the falsy things like `''` or `None`.

kuma/api/v1/plus/bookmarks.py
```diff
@@ -0,0 +1,190 @@
...
+    return JsonResponse(context)
+
+
+@never_cache
+@require_http_methods(["GET", "POST"])
+@is_subscriber
+def bookmarked(request):
+
+    url = (
+        (request.GET.get("url") or "")
```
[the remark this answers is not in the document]
peterbe: ditto as mentioned above.

kuma/api/v1/plus/bookmarks.py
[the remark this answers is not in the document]
escattone: @peterbe Ah, good point! 👍

kuma/api/v1/plus/bookmarks.py
```diff
@@ -0,0 +1,190 @@
...
+@is_subscriber
+def bookmarks(request):
+    try:
+        page = int(request.GET.get("page") or "1")
+        assert page > 0 and page < 100
+    except (ValueError, AssertionError):
+        return HttpResponseBadRequest("invalid 'page'")
+    try:
+        per_page = int(
+            request.GET.get("per_page") or settings.API_V1_BOOKMARKS_PAGE_SIZE
```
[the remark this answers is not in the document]
escattone: @peterbe's way is better, because it covers the case when someone uses `per_page=` (empty string).

kuma/api/v1/plus/bookmarks.py
peterbe: The idea is that if you un-soft-delete something, only give it a fresh new `created` if it wasn't the last thing you soft-deleted. 

I know the current implementation is a bit complex but it works, is unit tested, and is time agnostic. 

The overall intention is that if you un-soft-delete something, let it stay in the position it was instead of going back to the most recent. Try un-bookmarking something on <HTTP://localhost:3000/en-US/plus/bookmarks> that is NOT the top-most. Then press the Undo button. It reappears where it was before. 

But, suppose you bookmark something new almost every day. Back in last month you *un-*bookmarked `/en-US/docs/Foo` and today you return to that page. When viewed, it clearly doesn't show as bookmarked in the breadcrumb bar. Now, suppose you do click to bookmark it and then immediately after you navigate to <HTTP://localhost:3000/en-US/plus/bookmarks>, now you do expect `/en-US/docs/Foo` to appear first in the list because it was the most recently bookmarked URL even if it was technically un-soft-deleted.

kuma/api/v1/plus/bookmarks.py
peterbe: I changed it to an "allowlist" of keys instead. 



> Also, how do `DocumentURL` objects get updated, for example if the document they represent is moved?

They don't in this first version. 
I was hoping we add a something on the message queue to improve that. 
For example, we could have a periodic job that loops over the least recently checked `DocumentURL` rows and updates them. 
Another option is that we do something like this:

```diff
if not DocumentURL.objects.filter(url=ur).exists():
    return DocumentURL.objects.create(...)
-return DocumentURL.objects.get(url=url)
+else:
+    doc = DocumentURL.objects.get(url=url)
+    if (timezone.now() - doc.updated_at).days > 24 * 60 * 60:
+        update_documenturl(doc.id)  # send to a Celery task
+    return doc
```

Or, if we want to avoid Celery entirely, we can boldly go ahead and make it sync like this:

```diff
if not DocumentURL.objects.filter(url=ur).exists():
    return DocumentURL.objects.create(...)
-return DocumentURL.objects.get(url=url)
+else:
+    doc = DocumentURL.objects.get(url=url)
+    if (timezone.now() - doc.updated_at).days > 24 * 60 * 60:
+        doc.update()  # sync refresh 
+        doc.refresh_from_db()
+    return doc
```

if we do that, it's yet another reason why we *don't* need a message queue. But to be honest, I think it's only a matter of time till we need our message queue for something, so I'm not super motivated to get rid of Celery.

kuma/documenturls/models.py
peterbe: I'm torn. The intention of this field is more as an "audit trail". 
The field is for: **What URL was used (and should continue to be used) for download the metadata?**
The code comment there was wrong, so I changed it. 
It's basically the URL we're sending to `requests.get(HERE)` when we fetch the metadata. 
Yes, its value *can* be reconstructed based on `settings.BOOKMARKS_BASE_URL` and `uri`, but there's some sort of comfort to know that we have a record of which URL was used to `requests.get()` in case we ever change `settings.BOOKMARKS_BASE_URL` or what if it's related to Deep-dives, then you can't so easily "reverse" from the `uri` value. Perhaps the `/api/v1` endpoint has changed for fetching deep-dives' metadata. 

Perhaps, before I jump into deleting it, you and I sleep on it a smidge longer.

kuma/documenturls/models.py
peterbe: Exactly. It's scattered forward-thinking. I'm basically using the ORM as an audit log of the refresh/updates. Perhaps it's silly. Perhaps we can use the Python logging we could spew from our message queue workers. 

Again, not perfectly settled on how the refreshing/updating should work. Perhaps we don't bother. Like the point above, sleep on it? We could also write down these mumbling thoughts into code comments so as to remind the future self what the intention was and what dreams we had.

peterbe:
@escattone I don't know where the comment went but I changed the API endpoint to be just 1 endpoint. 
If you do `GET /api/v1/bookmarks/` it will behave very differently if you supplied a truthy `?url=...` or not. Thanks for that feedback!

kuma/api/v1/plus/bookmarks.py
```diff
@@ -0,0 +1,197 @@
...
+@require_http_methods(["GET", "POST"])
+@require_subscriber
+def bookmarks(request):
+    # E.g. POST -d url=https://... /api/v1/bookmarks/
+    # or   GET /api/v1/bookmarks/?url=https://...
+    if request.method == "POST" or request.GET.get("url"):
+        return _toggle_bookmark(request)
+
+    # E.g. GET /api/v1/bookmarks/
+    return _get_bookmarks(request)
```
escattone, line 52: I was thinking that we'd always use `_get_bookmarks` for both of the `GET` cases:

1.  When a `url` param is provided, we'd return either an empty list if the subscriber hasn't bookmarked that `url`, or a list of that one bookmark if they have. For this case we'd ignore any `page` and `per_page` params if provided.
2. When a `url` param is **not** provided, we'd respect the `page` and `per_page` params if provided and return a list of all of that subscriber's bookmarks.

but if you'd rather keep it the way it is currently, where the `GET` with a `url` param returns a differently-formatted JSON response than the `GET` without a `url` param, could you move the chunk of code within `_toggle_bookmark` that handles the `GET` case into its own new function called `_get_bookmark` (or something like that)? I think it would make things more readable to see something like this:

```python
    if request.method == "POST":
        return _toggle_bookmark(request)

    if request.GET.get("url"):
        return _get_bookmark(request)

    return _get_bookmarks(request)
```

so readers of the code don't mistakenly think that we're going to toggle the bookmark for `GET`s with a `url` because the function is named `_toggle_bookmark`.

peterbe: > readers of the code don't mistakenly think that we're going to toggle the bookmark for `GET`s

Fair point but "toggle" is a write-operation so GET is never appropriate for that. 
Having said that, it's weird that the function is called `_toggle_bookmark()` when it can take a GET request :)

My reasoning, perhaps flawed, was that finding out if the page you're on is bookmarked or not is possibly going to be very different in nature. I.e. you don't need metadata stuff like:
```python
"metadata": {
    "total": users_bookmarks.count(),
    "page": page,
    "per_page": per_page,
},
```

And what if we decide to need very different stuff from that `items` list if it's *not* an empty list. 

Now, you're making a slightly different suggestion that we break it out into *3* private functions. I.e. `_get_bookmark`. 
The reasoning I lumped what *you* call `_toggle_bookmark` and `_get_bookmark` is that they do a lot of stuff in common. But the more I look at it, perhaps it's sufficiently easy to make that functionality shared across 2 different functions.

kuma/documenturls/models.py
peterbe: Here's another thought... If we don't have a periodic job that keeps checking and updating previously saved `DocumentURL` objects, and only rely on updating (documents from the remote index.json) when there's a db write then the data can get stale until some "unlucky" user forces a refresh. 

So all in all it's not a trivial problem. Networks can be flaky and we don't want to punish an unlucky user with errors and delays, in their act of *using* their bookmark state. 

I think the more I think about it, a periodic Celery job is the best course of action. It's something we're familiar with and have routines for. 
Long story short, let's chat about it but definitely treat it as a follow-up from *this* PR.

kuma/api/v1/plus/bookmarks.py
escattone: @peterbe Yeah, my main concern was that the code reads like a `GET` with a `url` param will toggle the bookmark, but of course you see that it doesn't once you look inside the `_toggle_bookmark` function. It's just that the name of the function is misleading for the `GET` case.

peterbe: Implemented now. Thanks for the patience and feedback!

escattone (approved):
Awesome, thanks @peterbe!

PR #7904, opened 2021-06-14
drop libtidy-dev from apt install

peterbe:
I suspect in the olden days we used to depend on a python lib, like [django-tidy](https://pypi.org/project/django-tidy/) to post-process HTML in the Wiki.


peterbe:
@escattone This time, I did not exhaustively test this by doing all the usual things I do from Yari (e.g. site-search, sign up, whoami). But given what `tidy` does, I doubt it can have any bearing.

escattone (approved):

PR #7923, opened 2021-06-23
use waffle switch to protect subscription

### LINKED ISSUE #7921: Use waffle Switch to protect the subscription config URL
In https://github.com/mdn/kuma/commit/5f09827177eb895c3fa1b277fa2edaa077953651 we made it so that `/api/v1/subscriptions/config/` is protected by the `subscriptions` waffle **flag** but that means it's dependent on the user (e.g. staff users only). That means we can't reach it from the Prod or Stage builds in GitHub Action and thus never expose it in Yari. 

If we switch to a Switch instead, we can make the config exposed on Stage without allowing people to use the checkout or customer portal until we're ready.

peterbe:
Fixes #7921

Now all 3 subscription-related endpoints are "protected" by one and the same waffle **Switch** and all the old waffle Flags are cleaned up. 

Here's my reasoning:

1. The `/api/v1/subscriptions/config/` can't ever depend on the user. Therefore you can't use a Flag there. It will never make sense. 
2. We *could* have a Switch for the config and then a Flag for the checkout/customer portal. But that's dangerously complicated. 
3. If you want to do a thing like: Only allow Staff members to sign up, in prod, initially, then set the Switch to **true** and then write a little bit of JS React code to control the presence of the button. E.g. `{settingsData?.subscriptionConfig && userData.isStaff && <Subscription ...>}`
4. In general, flexibility can be dangerous because it's a footgun. So keeping it at a minimal is safer.
5. An env var might be a better alternative to `waffle.Switch` but when performance overheads don't matter at all, the `waffle.Switch` is actually a bit better than env vars because it won't require a new Kuma release the day we want to switch it all on.


peterbe:
@Gregoor With Ryan on PTO this week and next this most probably won't go out in Prod. In fact, Prod is still not upgraded so it doesn't even yet have the `/api/v1/subscriptions/config/` endpoint. 
I think it's safe to merge this. Everyone who wants to use it need to remember to run `docker-compose exec web python manage.py migrate` afterwards. 
So I think this is safe to merge assuming you approve.

Gregoor (approved):
Both the reasons and the code make sense to me. Thanks for putting this together Peter!

PR #7929, opened 2021-07-06
fully deprecated the $sample endpoint

peterbe:


kuma/wiki/views/code.py
```diff
@@ -25,11 +25,13 @@ def code_sample(request, document_slug, document_locale, sample_name):
     if settings.DOMAIN in request.get_host():
         return HttpResponseForbidden()
 
-    document = get_object_or_404(Document, slug=document_slug, locale=document_locale)
-    data = document.extract.code_sample(sample_name)
-    data["document"] = document
-    data["sample_name"] = sample_name
-    return render(request, "wiki/code_sample.html", data)
+    return HttpResponse(
+        "Legacy ($sample) URLs for live samples are now fully "
```
escattone, line 29:
```suggestion
        "Legacy ($samples) URLs for live samples are now fully "
```

kuma/wiki/views/code.py
```diff
@@ -25,11 +25,13 @@ def code_sample(request, document_slug, document_locale, sample_name):
...
+        "deprecated and will not work. If you have a document that relies "
+        "on a URL with '$sample' in it, switch to using the EmbedLiveSample() "
```
escattone, line 31:
```suggestion
        "on a URL with '$samples' in it, switch to using the EmbedLiveSample() "
```

kuma/wiki/tests/test_views_code.py
```diff
@@ -105,40 +55,3 @@ def test_code_sample_host_restricted_host(code_sample_doc, settings, client):
     assert response.status_code == 200
     assert "public" in response["Cache-Control"]
     assert "max-age=31536000" in response["Cache-Control"]
-
-
-def test_raw_code_sample_file(
```
escattone, line 110: I think we want to keep this test, since the endpoint will remain at least a while longer.

escattone (changes requested):

escattone (approved):

PR #7930, opened 2021-07-06
unbreak black and Dependabot upgrades

peterbe:
At some point, `pyproject.toml` and `poetry.lock` got a new version of black. 
E.g. https://github.com/mdn/kuma/commit/c705739a2596b0d8af7e47ad35a3c79f865bb4d6
What happened was that `.github/workflows/python-lints.yml` never ran. Now what happens is that the version of `black` you get inside the docker container *doesn't match the version* of `black` used in `python-lints.yml`. 
Worse, because it wouldn't even have noticed because a change to `pyproject.toml` our `.py` files no longer match that version that you get inside the docker container. So, when you run `black` inside `bash` on your docker container, it says some files aren't formatted even though nobody's touched them for ages. 
This PR solves both problems.

### DIFF
--- a/.github/workflows/python-lints.yml
+++ b/.github/workflows/python-lints.yml
@@ -5,6 +5,8 @@ on:
     paths:
       - "kuma/**/*.py"
       - .github/workflows/python-lints.yml
+      # This is in case Dependabot updates 'black'
+      - pyproject.toml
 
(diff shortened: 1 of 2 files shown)


escattone (approved):
Thanks @peterbe!

PR #7932, opened 2021-07-09
Big butcher axe

peterbe:
🪓 🥰 🗑


peterbe:
[Upgraded from `raven` to `sentry_sdk` too](https://sentry.prod.mozaws.net/operations/mdn-stage/issues/11847272/?query=is%3Aunresolved)

peterbe:
@escattone This is so cool! Me for example, I have a local `redis` and local `postgres` 13 running here on my MacBook host. Now I can run Kuma without `docker-compose`!
I created a `Procfile` that looks like this:

```
▶ cat Procfile
celery: CELERY_BROKER_URL=redis://localhost:6379/0 ./scripts/start-worker
web: ./manage.py runserver
```
and in *my* `.env` file I have:
```
DATABASE_URL=postgresql://localhost/kuma
```

And now running `poetry run honcho start`
it starts `celery` and `runserver` all and entirely without docker or docker-compose. 

Two immediate cute advantages;

1. I can put in print statements into Django's core library itself. (e.g. `emacs /Users/peterbe/Library/Caches/pypoetry/virtualenvs/kuma-zSqACmTg-py3.8/lib/python3.8/site-packages/django/__init__.py`) 
2. I can enable [full logging of every single SQL statement in Postgres](https://www.peterbe.com/plog/log-all-postgressql-sql) which is fantastic for debugging or profiling SQL traffic. 
3. It starts a `localhost:8000` in 3-4 seconds. 
4. Simply running `poetry install` on the host is much faster than doing it inside the docker container.

escattone (commented):
@peterbe An incomplete first-pass review:
- I think we should add the `postgresql-client` to the kuma base docker build
- I think we need to reinstate `make build-static` (and add back whitenoise) only because we need it to style the Django admin, right?

docker/images/kuma_base/Dockerfile
```diff
@@ -12,15 +12,10 @@ RUN set -x \
...
     libsasl2-modules \
     gettext \
-    build-essential \
-    libxml2-dev \
-    libxslt1-dev \
-    libffi-dev \
-    libmagic-dev \
-    default-libmysqlclient-dev \
-    default-mysql-client
+    build-essential
```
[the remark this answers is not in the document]
peterbe: Why should we do that? We're using `psycopg2-binary` in `pyproject.toml` and clearly the container works after the existing `Dockerfiler` has been built which includes the `poetry install`. 
Is `postgresql-client` making this better or faster?

peterbe:
>     * I think we need to reinstate `make build-static` (and add back whitenoise) only because we need it to style the Django admin, right?

Sigh. You're right. I first thought we could use `staticfiles` from Django but it is hardcoded'ly disabled when `DEBUG=False`. 

I'll work on putting it `whitenoise` back.

peterbe:
@escattone I put `whitenoise` back in. 
Now, I can set `echo 'DEBUG=false' >> .env` and `docker-compose up` and `docker-compose exec web make build-static` and then go to <http://localhost:8000/admin> and it works.

docker/images/kuma_base/Dockerfile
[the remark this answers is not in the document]
escattone: @peterbe and I discussed this earlier today over a video call, but I'll summarize the key points here. The motivation behind my request was for the convenience of running the `psql` client from any of the Kubernetes pods that run the Kuma image. In the end, we decided to leave it out because for those times when I need to run `psql` within Kubernetes, I can `kubectl exec` into the "admin" pod, on which I've already installed the `psql` client.

escattone (changes requested):
@peterbe I'm just going through the code that exists, and it looks good so far. I found some minor changes we need to make in `Jenkinsfiles/utils.groovy`:

- The line https://github.com/mdn/kuma/blob/a00db1b5f2ba96a4a04a27969de833710b6f700d/Jenkinsfiles/utils.groovy#L221 needs to change to:
    ```groovy
    def db_url = 'postgres://kuma:kuma@postgres:5432/developer_mozilla_org'
    ```
- And because of that change we need to update https://github.com/mdn/kuma/blob/a00db1b5f2ba96a4a04a27969de833710b6f700d/Jenkinsfiles/utils.groovy#L226 to:
    ```groovy
    sh_with_notify("${dc_exec} urlwait ${db_url} 30", 'Wait for Postgres')
    ```



I also found some easy deletions that we can add to this PR if you'd like:

- The entire `kuma/core/management/commands` directory. They all fail and we don't use them.
- `scripts/anonymize.sql`
- `scripts/check_anonymize.sql`
- `scripts/create_sample_db.sh`
- `scripts/create_read_only_user.sql`
- `settings.py` (the one within the root directory) Wow, I don't know how this has lasted this long. This hasn't been used since the pre-AWS, Santa Clara data center days. Delete with extreme prejudice! 😆

pyproject.toml
```diff
@@ -11,59 +11,31 @@ documentation = "https://kuma.readthedocs.io"
...
 elasticsearch-dsl = "^7.3.0"
-enmerkar = "^0.7.1" # Maintained fork of django-babel
-google-api-python-client = "^2.10.0"
 gunicorn = "^20.1.0"
-html5lib = "1.0.1"
-httplib2 = "^0.17.0"
-jinja2 = "^3.0.1"
-lxml = "^4.6.3"
-mysqlclient = "^1.4.6" # (Django database driver)
 newrelic = "^5.14.0.142"
```
escattone, line 26: We need to upgrade `newrelic` to the latest so it stays in sync with Django (see https://github.com/newrelic/newrelic-python-agent/commit/2db7dc607cf8717c212f2839e426a9d8d820e80d). I've already tested this on stage, and it fixes the 500 errors that we get when we should be returning 404's.
```suggestion
newrelic = "^6.4.4.161"
```

kuma/urls.py
```diff
@@ -19,23 +16,21 @@
...
-        # has been requested by a client, all other client requests will be
-        # redirected by the CDN instead of this Django service.
-        shared_cache_control(s_maxage=MONTH)(
-            RedirectView.as_view(
-                url="https://community.mozilla.org/events/", permanent=False
-            )
-        ),
-        name="events",
-    ),
-)
```
escattone, line 38: I'm thinking we should keep this redirect. I don't feel strongly about it though.

escattone (changes requested):
@peterbe I pushed this to stage and ran our [yari integration tests](https://github.com/mdn/yari/tree/main/testing/integration). They uncovered an issue where our `newrelic` package is out-of-sync with Django (which caused 404's to become 500's), and that we're no longer redirecting the `/{locale}/events` endpoint. Other than that, it's looking really good! The db migration (on our new stage Postgres RDS instance) ran great!

escattone:
@peterbe Just a reminder that we're going to have to update our Yari integration tests to match this PR, but I can take care of that.

escattone:
@peterbe Here's the Yari integration-test companion for this PR: https://github.com/mdn/yari/pull/4234

It assumes that you'll add back the `/{locale}/events` redirect.

peterbe:
@escattone `newrelic` upgraded and all the URLs relating to `$samples/...` now hard-404 instead of that legacy plain text message.

r?

peterbe:
@escattone Do you want me to merge in https://github.com/mdn/yari/pull/4234 into this PR so it's all together?

kuma/urls.py
```diff
@@ -19,23 +14,21 @@
...
 
 urlpatterns = [re_path("", include("kuma.health.urls"))]
-# The non-locale-based landing URL's
-urlpatterns += [re_path("", include("kuma.landing.urls"))]
-urlpatterns += i18n_patterns(
-    re_path(
-        r"^events",
+urlpatterns += [
+    path(
+        "<locale>/events/",
```
escattone, line 19: We need to remove the trailing slash, or if we don't, we'll need to add `events` to:

 https://github.com/mdn/yari/blob/c8c5860bf9331690a3fb17874c663d7eafea0add/deployer/aws-lambda/content-origin-request/index.js#L28.

As it is, it works locally, but not when deployed to stage. On stage, a request for `/en-US/events` is redirected to `/en-US/events/`, but then that request is redirected back to `/en-US/events` by the content-origin-request lambda code:

https://github.com/mdn/yari/blob/c8c5860bf9331690a3fb17874c663d7eafea0add/deployer/aws-lambda/content-origin-request/index.js#L182

```suggestion
        "<locale>/events",
```

escattone: @peterbe On second thought, I think a better way to handle this `events` redirect is via the fundamental redirects, which I've gone ahead and done in https://github.com/mdn/yari/pull/4234. So let's just remove this endpoint (and its tests) altogether from Kuma.

escattone (changes requested):
@peterbe Also, the entire `kuma/wiki` folder can be deleted. None of it is used any longer.

escattone:
@peterbe Once you make those changes (remove the `events` endpoint/tests, and the `kuma/wiki` folder), I think this PR is ready to go! I've already tested it on stage using the Yari integration tests, and other than the changes requested, it looks good!

kuma/urls.py
peterbe: I don't think so. Now, with regular `path()` we get the best of both worlds. 
A GET to `/en-US/events` is a redirect to `https://community.mozilla.org/events/` but that's because it does **2 redirects**. 
```
▶ xh http://localhost:8000/en-US/events
HTTP/1.1 301 Moved Permanently
location: /en-US/events/

▶ xh http://localhost:8000/en-US/events/
HTTP/1.1 302 Found
location: https://community.mozilla.org/events/
```

So the net effect is that our URL works with or without a slash. 

I think the code here in `urls.py` is correct but I'll need to re-read your comment about the headless tests.

peterbe: Oh I think I see now. What Lambda does matters. 
The big question is; can't we just move this redirect to the fundamental redirects instead of even having any of it here in Kuma??

peterbe: If you can merge https://github.com/mdn/yari/pull/4269 we can totally rip out all of this code from kuma.

peterbe:
If we merge https://github.com/mdn/yari/pull/4269 we'll never have to worry about `/:locale/events/?` ever again in the kuma context.

escattone (approved):
Let's swing the butcher axe! 🪓  Thanks @peterbe!

PR #7938, opened 2021-07-21
delete nginx and docker-compose.ssl.yml

peterbe:
If you really need to test `https://localhost:8000` there are other ways (e.g. `ngrok`), but I don't think Kuma needs to keep this since nobody's using it.


peterbe:
@escattone Perhaps I'm just on a momentum from the butcher-axe work :)

escattone (approved):

PR #7943, opened 2021-07-23
support market research survey (july 2021)

peterbe:
In the new July 2021 landing page survey, we're no longer bothering with variants first. It's just plain survey submissions. We no longer submit or record `email` as part of the survey. And since we don't even have user auth at the moment, let's not store that either. 

This PR basically deletes all things related to `variant` and it does a couple of big simplifications which is best seen by looking at the edits to the `models.py` file.


peterbe:
Let's not bother with this. It's not going to be used or needed at all for the landing page survey 2.0. I think.

escattone (approved):

PR #7962, opened 2021-07-27
calm down Dependabot

peterbe:
It's just too chatty at the moment. 
There's too little value to upgrade these more frequently than weekly. 
And for the record, one can always push for a refresh by going to https://github.com/mdn/kuma/network/updates/175308387


peterbe:
There's nothing to review here @fiji-flo @escattone so I'm going to merge it as is.

PR #7967, opened 2021-08-05
periodically refresh documents

### LINKED ISSUE #7965: Periodically refresh DocumentURL rows
**Summary**
Now that https://github.com/mdn/kuma/pull/7903 has landed, we should periodically re-fetch the `DocumentURL` in a cron-like way.

peterbe:
Fixes #7965

~I'm making it a draft for now because I think there's a mismatch between calling it `document_url` and `documenturl` and it would be nice to tidy that up.~ 
Also, it would be great to have a think about the use case of a 30x redirect and how that should effect.


peterbe:
The only possible bug about this is that a document might become a redirect to a new URL that isn't a Yari document. 
And if an existing document becomes a redirect to another document, we should build some sort of "merge functionality". But it might make sense to stall on that functionality. We don't desperately need it yet and if it starts to cause problems, it'll become a LOT easier to debug and test once it's a real deal.

PR #7972, opened 2021-08-17
subplat integration

peterbe:
As of this, imagine you tell the sub plat that our "download URL" is `https://developer.mozilla.org/users/subplat/download/` then, after a user has paid, they click that green ("Click here to download") button. 
What that will do is force the user to sign in again, but with "silent authentication" which means it redirects them to `/users/fxa/login/authenticate/?prompt=none` (it should be fast) and this time, in the FxA OIDC callback it should get a "second chance" to see that the user has indeed signed in and has paid. 

To test, add this to your kuma `.env`:

```
SUBSCRIPTION_SUBSCRIBE_URL=https://accounts.stage.mozaws.net/subscriptions/products/prod_Jtbg9tyGyLRuB0?plan=price_1JFoTYKb9q6OnNsLalexa03p
SUBSCRIPTION_SETTINGS_URL=https://accounts.stage.mozaws.net/subscriptions

SUBPLAT_CONFIGURATION_CHECK=true
```
You also, need to check out https://github.com/mdn/yari/pull/4498 in Yari so you can get the link. 
But you can also ignore Yari entirely and just manually type in <http://localhost:8000/users/subplat/subscribe/> independent of being signed in or not.

### DIFF
--- a/kuma/settings/common.py
+++ b/kuma/settings/common.py
@@ -336,6 +336,22 @@ OIDC_RP_CLIENT_SECRET = config("OIDC_RP_CLIENT_SECRET", default=None)
 # Function that gets called
 OIDC_OP_LOGOUT_URL_METHOD = "kuma.users.auth.logout_url"
 
+# This is the default URL that mozilla_django_oidc will use if there is no
+# appropriate `?next=...` it can use.
+LOGIN_REDIRECT_URL = "/"
+LOGIN_REDIRECT_URL_FAILURE = "/?login=failure"
+
+# This will HEAD request `SUBSCRIPTION_SUBSCRIBE_URL` and
+# `SUBSCRIPTION_SETTINGS_URL` to see if they're reachable at all.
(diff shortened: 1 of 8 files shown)


peterbe:
@escattone One thing that occurs to me now is that one of the settings is called `SUBSCRIPTION_SUBSCRIBE_URL` but the other one is called `SUBPLAT_CONFIGURATION_CHECK`. If the first word (before the `_`) is the "namespace", clearly it's different. Naming things is hard. Perhaps that imperfection is a good reason to pause and contemplate what it really should be (consistently) named.

peterbe:
I still can't get this to work, in practice. I think we might need some help. 
https://mozilla.slack.com/archives/C4D36CAJW/p1629389458083800 for those with access

I tried messing with `mozilla_django_oidc` so that I force-add `&prompt=none` to the authentication URL. Now, there's no prompt but it always comes back (redirects) back with `?error=invalid_request`.

PR #7997, opened 2021-10-21
WIP: Notifications

nicolaslara:
This is a WIP for the MDN plus notifications feature. This isn't ready to merge.


kuma/notifications/extract.py
```diff
@@ -0,0 +1,113 @@
...
+        elif len(notifications) > 0:
+            return [self.text + ", ".join(notifications)]
+        return [self.text + i for i in notifications]
+
+
+class NowSupported(NotificationGenerator):
+    text = "Now supported in "
+
+    def compare(self, old, new):
+        if new.get("version_added") and not old.get("version_added"):
```
ddbeck, line 86: I want to alert you to something potentially hazardous here:

Setting aside lists (where a feature is represented by multiple support statements  as in `[{ "version_added": "X" }, …]`), it's also common for BCD to record data for future versions. That is, "now supported" is a question you can't answer from a support object alone. I suspect you'll need some sort of utility that looks up the current release of a browser (from the `browsers` part of the data), then uses that value to figure out whether a set of support statements shows support for that version.

The logic for this thing will be a bit complex: it'll have to be able to resolve versions in ranges (e.g., current version 92 in the range represented by `"version_added": "37", "version_removed": "99"`, multiple statements, and to ignore statements with flags or `"preview"` values).

nicolaslara: Thanks for the explanation! This is very useful! 

What I have right now is the naive implementation so I can run through the process of generating notifications for the whole bcd object.

Part of the reason I wanted to split the notification generators into classes is to be able to account for this kind of behaviour. I can now take what you wrote (plus anything else I can gather from the spec) and build a better NowSupported generator. I'll ping you when it's done so we can review it together

ddbeck: Great, thanks @nicolaslara! Glad to hear that this is more of a placeholder than the real thing. I'm not surprised it was already on your mind, but I'm relieved to have it said explicitly.

PR #8006, opened 2021-11-09
Validate access_token every hour.

akatsoulas:


kuma/settings/common.py
```diff
@@ -332,11 +334,18 @@ def _get_locales():
...
+# Store the access token in session
+OIDC_STORE_ACCESS_TOKEN = config("OIDC_STORE_ACCESS_TOKEN", cast=bool, default=True)
 # Set the issuer of the token.
 FXA_TOKEN_ISSUER = config("FXA_TOKEN_ISSUER", default="https://accounts.firefox.com")
+# Validate access token endpoint.
+FXA_VERIFY_URL = config(
+    "FXA_VERIFY_URL", default="https://oauth.accounts.firefox.com/v1/verify"
+)
+# Set token re-check time to an hour in seconds
+FXA_TOKEN_EXPIRY = config("FXA_TOKEN_EXPIRY", default=3600)
```
escattone, line 348: This doesn't seem to be used anywhere.

kuma/users/middleware.py
```diff
@@ -0,0 +1,42 @@
...
+            response_token_info = (
+                requests.post(settings.FXA_VERIFY_URL, data={"token": access_token})
+            ).json()
+
+            # if the token is not verified, log the user out
+            if (
+                response_token_info.get("code") == 400
+                and response_token_info.get("message") == "Invalid token"
+            ):
+                logout(request)
```
escattone, line 42: @akatsoulas and I discussed this over Zoom, but documenting it here as well. We need to add an `else` clause here and within that `else` clause reset the `oidc_id_token_expiration` for another hour from now.

akatsoulas:
@escattone pr updated

kuma/users/middleware.py
```diff
@@ -0,0 +1,46 @@
...
+        else:
```
escattone, line 43: I think this `else` clause should be relative to
```python
            if (
                response_token_info.get("code") == 400
                and response_token_info.get("message") == "Invalid token"
            ):
```
and not
```python
        if access_token and expiration < now:
```
otherwise any request for a refreshable URL where the one-hour time limit hasn't been reached yet, will keep extending that time limit, and we could theoretically never check for an expired token, or maybe I'm missing something?

escattone (changes requested):

escattone (approved):

PR #8013, opened 2021-11-24
Remove refresh token

akatsoulas:


escattone (approved):
Thanks @akatsoulas, and also thanks for fixing my PR from yesterday!

PR #8027, opened 2022-01-13
Notifications

fiji-flo:


kuma/notifications/utils.py
```diff
@@ -1,6 +1,11 @@
+import json
+
 from kuma.notifications.models import Watch, Notification, NotificationData
 
 
+browsers: dict = json.loads(open("browsers.json").read())
```
fiji-flo, line 6: @nicolaslara that file is not checked in (and probably shouldn't be). However, this is fails the migration on stage.

PR #8036, opened 2022-02-21
Changed pagination strategy. Added bulk edit endpoints

Guyzeroth:


kuma/core/tests/__init__.py
```diff
@@ -10,11 +10,12 @@
...
-    assert "no-cache" in response["Cache-Control"]
-    assert "no-store" in response["Cache-Control"]
-    assert "must-revalidate" in response["Cache-Control"]
-    assert "s-maxage" not in response["Cache-Control"]
+    if "Cache-Control" in response:
+        assert "max-age=0" in response["Cache-Control"]
+        assert "no-cache" in response["Cache-Control"]
+        assert "no-store" in response["Cache-Control"]
+        assert "must-revalidate" in response["Cache-Control"]
+        assert "s-maxage" not in response["Cache-Control"]
```
fiji-flo, line 18:
```suggestion
    assert "max-age=0" in response["Cache-Control"]
    assert "no-cache" in response["Cache-Control"]
    assert "no-store" in response["Cache-Control"]
    assert "must-revalidate" in response["Cache-Control"]
```

@Guyzeroth That should work and we should assert we set `Cache-Control`

PR #8045, opened 2022-03-09
Merge bulk and individual watched api

Guyzeroth:
- For easier use on the client PWA merge the single and multiple API endpoint for watched items.
- Default watch settings are now returned at top level of response and each item in the multiple GET returns it's own preferences. 
- Having this context enables us to filter on specific updates in the notifications tab.

### DIFF
--- a/kuma/api/v1/plus/notifications.py
+++ b/kuma/api/v1/plus/notifications.py
@@ -8,10 +8,10 @@ from typing import Optional
 import requests
 from django.conf import settings
 from django.db.models import Q
+from django.middleware.csrf import get_token
 from ninja import Field, Router
 from ninja.pagination import paginate
 
-from kuma.api.v1.decorators import require_subscriber
 from kuma.documenturls.models import DocumentURL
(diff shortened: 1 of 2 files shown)


kuma/api/v1/plus/notifications.py
```diff
@@ -229,9 +232,8 @@ class UpdateWatch(Schema):
     update_custom_default: bool = False
 
 
-@watch_router.post("/watch{path:raw_url}", response={200: Ok, 400: NotOk, 400: NotOk})
-def update_watch(request, raw_url, data: UpdateWatch):
-    url = DocumentURL.normalize_uri(raw_url)
+@watch_router.post("/watching/", response={200: Ok, 400: NotOk, 400: NotOk})
+def update_watch(request, url: str, data: UpdateWatch):
```
fiji-flo, line 236: Just maybe double check that we don't have any url encode issues / that the `normalize_uri` did nothing crucial.

fiji-flo (approved):

PR #8050, opened 2022-03-17
918 Support subscription limits

Guyzeroth:


caugner (approved):
LGTM, although I don't have the context for all changes

PR #8055, opened 2022-03-21
Fix:Unwatch many null case

Guyzeroth:


kuma/api/v1/plus/notifications.py
```diff
@@ -343,6 +342,7 @@ def unwatch(request, data: UnwatchMany):
         watch__url__in=data.unwatch
     ).delete()
     profile: UserProfile = request.auth
+    subscription_limit_reached = False
     if not profile.is_subscriber:
```
caugner, line 346:
```suggestion
    if profile.is_subscriber:
        subscription_limit_reached = False
    else:
```

caugner (approved):
I have a suggestion, but approve the changes anyways.

PR #8059, opened 2022-03-23
Fix subscriber status being ignored when updating collections

Guyzeroth:


schalkneethling:
lgtm. Previously everyone had a 5 collection size limit. After this change, only non-subscribers will have the limit enforced.

schalkneethling (approved):

PR #8060, opened 2022-03-24
Fix delete limits

Guyzeroth:


Guyzeroth:
Fixes subscription limit being incorrectly true on collection item delete for subscribers.

schalkneethling (approved):

PR #8071, opened 2022-04-06
user with is_staff=True should be subscribers

fiji-flo:


kuma/users/tasks.py
```diff
@@ -46,7 +46,7 @@ def process_event_subscription_state_change(event_id):
             event.save()
             return
 
-    if "mdn_plus" in payload["capabilities"]:
+    if "mdn_plus" in payload["capabilities"] and not user.is_staff:
```
fiji-flo, line 49: Don't update capabilities for `is_staff`. (We need to clean this manually).

fiji-flo, line 49: Don't update capabilities for `is_staff`. (We need to clean this manually).

